PR #150, opened 2013-03-27
TrueCrypt support for OpenPGP

hongquan:
Hello,

Here is the patch of OpenPGP to support the use of pkcs15's data object by TrueCrypt.

The method is to map data object to OpenPGP private DOs (0101 - 0104). However, currently, only DO 0101 can be used. Because other DOs (0102 - 0104) cannot be accessed at pkcs15 emulation initialization. Ref: http://sourceforge.net/mailarchive/message.php?msg_id=30646373

Another limitation is the inconsistence of filename between TrueCrypt own file list and token's. Please see these screenshot:
http://storage2.static.itmages.com/i/13/0327/h_1364371394_2877566_f6a08508dd.png
http://storage5.static.itmages.com/i/13/0327/h_1364371606_1091578_409c011341.png
There is no way to store the filename to OpenPGP card, it has to be hard-coded.
 
So, the note for user is: After importing keyfile to token, user should restart TrueCrypt to get the right filename.


frankmorgner:
any reason why you re-opened this? It's not compiling, see the results of travis or appveyor

hongquan:
Hi. I open this because some OpenPGP users want to use it with TrueCrypt and derivatives.

frankmorgner:
What's your plan? Do you intend to fix the outstanding problems?

hongquan:
Yes, I will fix the build error, and hope that this PR will be accepted.

frankmorgner:
OK, please open a new PR when you've tested your changes (or if you need advice). Please do a clean rebase onto the current master.

PR #209, opened 2014-01-18
Dell reader cannot handle PIN without padding

jasp00:
The function export is needed by opensc-pkcs11.


frankmorgner:
As you said it @jasp00 , this hack is way to complicated to work around some strange reader. Its bad implementation is documented here http://pcsclite.alioth.debian.org/ccid/supported.html#0x413C0x2101. Users of OpenSC should set `enable_pinpad = false;` in `opensc.conf`.

However, I created #520 to at least take use the implementation of the new PC/SC command

PR #296, opened 2014-10-11
pkcs15-lib: prkey_bits - use private key field length

philipWendland:
sc_pkcs15init_keybits can behave erroneously with certain EC keys. It basically strips leading most significant zero-bits. It seems that EC keys (privateD), however, can have valid zeros at the beginning which legitimately count to the field length.
So, one can NOT calculate reliable key lengths only by looking at the privateD component of the private key. You have to look at the field length specified in the parameters.

---
I did some tests on the current (to be fixed) version using prime256v1. Different private keys of the same curve behave differently, depending on the privateD component. While some did produce correct lengths (the ones which had exactly one leading zero), others did not (calculated key lengths of 257, or even <256).
The fixed version works with all keys, also the ones that failed before.

Where does prkey_bits() get called from?
Only when trying to store private keys:
```
[*@* src] $ grep --include="*.c" -r -n "prkey_bits"
pkcs15init/pkcs15-lib.c:115:static int	prkey_bits(struct sc_pkcs15_card *, struct sc_pkcs15_prkey *);
pkcs15init/pkcs15-lib.c:1400:	keybits = prkey_bits(p15card, &key);
pkcs15init/pkcs15-lib.c:2138:prkey_bits(struct sc_pkcs15_card *p15card, struct sc_pkcs15_prkey *key)
```

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -2138,6 +2138,7 @@ static int
 prkey_bits(struct sc_pkcs15_card *p15card, struct sc_pkcs15_prkey *key)
 {
 	struct sc_context *ctx = p15card->card->ctx;
+	int rv;
 
 	switch (key->algorithm) {
 	case SC_ALGORITHM_RSA:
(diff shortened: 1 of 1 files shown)


philipWendland:
Resolved by #371 / 90860a850cca7879c3d4c9294d7fe0e55b2406ca

PR #304, opened 2014-10-19
EC key storing: make sure algorithm params are available

philipWendland:
When trying to import EC keys onto a card with pkcs15-init, the
params.ec of sc_pkcs15init_prkeyargs or pubkeyargs had not been
initialized. This commit fixes that by reading the curve name from the
EVP_KEY and initializing the parameters.

Also, in sc_pkcs15_encode_pubkey_as_spki, this commit adds code that
tries to construct the EC algorithm params from the key itself. This
prevents a NULL parameter in the SPKI/ASN1 encoding.


philipWendland:
Resolved by #371 / 90860a850cca7879c3d4c9294d7fe0e55b2406ca and #369 / 5e477e009b990f50601c3a430e602f6bbf758276, 967e3f51e4c835fb390dabd3243f6c4ab7be87f7

PR #305, opened 2014-10-19
Add support for the IsoApplet Java Card applet

philipWendland:
Please note: depends on ~~#294~~ ~~#322~~, ~~#296~~, ~~#304~~ and ~~#314~~ to be merged, or fixed otherwise.

The IsoApplet is an Open Source Java Card applet targeting newer (>= 2.2.2) Java Cards. It aims for Iso7816-comliance, but tries to find a balacance between standard-compliance ant not overcomplicating things.

The Applet has a Iso7816-filesystem that can hold PKCS#15 structures. The driver supports full read/wirte support, i.e. pkcs15-init --create-pkcs15 will create the PKCS#15 files on the smartcard.

It supports on-card-generation as well as private key import. RSA 2048 bit, and BrainpoolP192r1 and Prime256v1 for ECDSA are supported. 
The applet returns the API version upon selection, so if mayor things change, the driver can identify the API to use (or even not match). (Looking forward to see some Java Card >= 3.0.2 cards for the ECC improvements, IMHO a big problem of the version 2.2.2 is that the host can not do the hashing..)

I've tested the applet on 3 different Java Cards, all worked out of the box. Obviously, not all work with extended APDUs or ECC, but RSA 2048 bit with short apdus always worked.

Performed tests:

RSA: 
  - Encryption/Decryption using OpenSSL (and engine_pkcs11)
  - Signing/Verifying using OpenSSL
  - Authentication via OpenSSH
  - Sign/verify with GnuPG

ECC:
Application tests are difficult, PKCS11-support for ECC seems to be very scarce.
However, I can verify signatures generated by pkcs11-tool using OpenSSL. 
Only Prime256v1 though, the brainpool curve is not supported yet.

Others:
  - Recognized by Firefox

Currently, the Applet AID is random. Is this a blocker? 
If so, does anyone have experience with obtaining unique AIDs (costs, time)?
I suspect, as a student, I lack the resources for that. 
Does anyone have spare AIDs? Or don't you get "blocks" of AIDs?

Thank you.

URL: https://github.com/philipWendland/IsoApplet


frankmorgner:
You can find an external card driver for the German ID card here https://github.com/frankmorgner/vsmartcard/blob/master/npa/src/card-npa.c

To create an external driver you must implement `sc_module_init` and `sc_driver_version`, see https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/ctx.c#L360 and https://github.com/OpenSC/OpenSC/blob/master/src/pkcs15init/pkcs15-lib.c#L263. Then, you add your external driver to opensc.conf as done here https://github.com/frankmorgner/vsmartcard/blob/master/npa/opensc.conf.in

philipWendland:
Thank you, I will change the driver after the SC_CARDCTL_GET_SERIALNR is implemented.

philipWendland:
I had a look at the external driver mechanism and there are some open questions:

1. The E-Mail conversation linked by you indicates that new - external - card drivers are directly shipped with OpenSC. Is this correct?    
If not, I'd have to test different different versions of OpenSC against different versions of the card driver, which results in increased testing complexity. As I want to test different cards as well, the testing complexity would become unmanageable for me.
2. I understand external card drivers as "pure card drivers". How would a pkcs15-init driver be realized with this mechanism?
3. How are CARDCTL calls possible with external card drivers? Especially calls from the pkcs15init driver to the actual card driver, with operation codes solely used by that card.
4. Just a principal, practical observation: Given the above questions - why should a driver for an applet that was acutally designed to be used by OpenSC be realized as external card driver? Besides the card driver and pkcs15-init driver, no additions or modifications to OpenSC are needed.    
Don't get me wrong - I understand the reasons given for not integrating a card driver too quickly, especially with an applet without commercial interests behind it.    
Assuming that the external card driver *is* ment to be shipped with OpenSC, the practical consequence for the user would be the need to manually enable the specific card driver (and start thinking :-) ). In this case I find it more reasonable to introduce a "driver blacklisting feature" - new drivers would be blacklisted.    
The driver blacklisting could have other use cases as well: A user might have a Java Card with an applet pre-installed by the card manufactorer that is supported by OpenSC. The user might decide to try out / use the muscle or IsoApplet, or any other available applet for whatever reason. Blacklisting the driver for the proprietary applet might be more preferable for him than deleting the applet or forcing a specific card driver, especially when different smartcards are used on the computer.    
Instead of a seperate blacklisting feature, the existing card_drivers block in opensc.conf could be extended so that instead of loading every internal driver at once ("card_drivers = internal" in opensc.conf) every internal driver is seperately listed.

martinpaljak:
I have missed something but what do you mean by eternal driver and why
would this be necessary ?

Frank, the whole "external library" thing was a mishap that should have not
happened, it hardly helps against anything.



--
Martin
+372 515 6495

philipWendland:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

On 02/11/2015 06:57 PM, Martin Paljak wrote:
> I have missed something but what do you mean by eternal driver and
> why would this be necessary ?
> 

I was referring to the latest the comments here and:

[Opensc-devel] State of card drivers

from the mailing list [1].


[1] http://opensc-devel.narkive.com/xzXJ3yGj/state-of-card-drivers

> Frank, the whole "external library" thing was a mishap that should
> have not happened, it hardly helps against anything.
> 
> 
> 
> -- Martin +372 515 6495
> 
> On Wed, Feb 11, 2015 at 7:43 PM, philipWendland

frankmorgner:
@philipWendland  You can create a second external pkcs15 driver. I think this answers your second and third question. I have done this here https://github.com/frankmorgner/vsmartcard/blob/d35e8e8de1783b39d9a4418adeaf25e953838169/npa/src/pkcs15-npa.c and configured the driver here https://github.com/frankmorgner/vsmartcard/blob/d35e8e8de1783b39d9a4418adeaf25e953838169/npa/opensc.conf.in. You basically need a function that implements the version check and declare your function that implements the pkcs15 emulation.

Yes I'd like to ship the external card driver with OpenSC (your first question). (But even if it wasn't shipped with OpenSC, you would simply have to edit your configuration file for testing it with a new version of OpenSC.)

Loading an external card driver is the mechanism that works and is available for an "opt-in" for the user. Currently, a big chunk of OpenSC's code is in a very bad shape. It's so bad that even core maintainers don't understand it (see https://github.com/OpenSC/OpenSC/commit/08eb700b9749c17e8ff29ddb45b3132d8d38308e#commitcomment-9691474 for example). Unmaintained card drivers are part of the problems, see for example #362. And still, OpenSC is the software that gets exclusive access to your key. This means that although you might use an EAL5+ certified chip, it's key is easily compromised. I want safe defaults, which allow safe use of some supported cards. If someone wants to go leading edge, he may "opt-in" (and should "start thinking"). Your fourth point allows safe use of OpenSC by "opt-out", but it's most likely not a safe default.

@martinpaljak my suggestion has been out now for quite some time. There came hardly any answer. If needed, we may continue the discussion on the mailing list...

dengert:
On 2/11/2015 2:35 PM, Frank Morgner wrote:

>
> Loading an external card driver is the mechanism that works and is available for an "opt-in" for the user. Currently, a big chunk of OpenSC's code is in a very bad shape. It's so bad that even core
> maintainers don't understand it (see 08eb700#commitcomment-9691474 <https://github.com/OpenSC/OpenSC/commit/08eb700b9749c17e8ff29ddb45b3132d8d38308e#commitcomment-9691474> for example).


That specific problem was caused by your change to slot.c that was working.

   https://github.com/OpenSC/OpenSC/commit/3a92bf7af5aaa85fdbc0434d0a55393e68e0f98d

to add:
free_p11card = 1;

if (free_p11card)
    free(p11card);

base on the Coverity scan output.

I am working on cleaning it up.

-- 

philipWendland:
Frank,
thanks for the answer, but I think you got me wrong on two points:

I was not looking for PKCS#15 emulation (src/libopensc/pkcs15-*.c), but PKCS#15 initialization (src/pkcs15init/pkcs15-*.c). The IsoApplet does not need PKCS#15 emulation, infact it relies on `pkcs15-init --create-pkcs15` to fill up the empty filesystem with a PKCS#15 structure before use.

It depends on the default configuration "opensc.conf" whether the proposed driver blacklisting/whitelisting is "opt-in" or "opt-out". The blacklisting/whitelisting would also make it easily possible to deactivate some unwanted, old drivers without refactoring them. It was just an idea I wanted to throw in.

martinpaljak:
Frank,

First, I think the overall problem is the lack of agreement about what
OpenSC is supposed to be: is it a driver with simultaneous support for
multiple smart cards OR it is a toolkit for building card-specific
drivers with an annoying side-effect that by default the toolkit enables
support for "other" cards as well.

Originally it was the former: pkcs15-* tools and a pkcs11-module with
multiple card support. This,  multiple drivers x multiple cards exposed
via a single PKCS#11, also generates the very confusing "card" logic in
PKCS#11 module (especially because of apps with pkcs11 2.11 support with
static slot tables instead of 2.20 and dynamically changing slots)

Looking at src/tools/*-tool.c and the way Windows minidriver is
distribtued, this is no longer the case. Maybe changing the written down
target as something different is necessary. Things like p11-kit proxy or
caml-crush from ANSSI put this goal into a different perspective in 2015.


Second, I don't usually rfer to RMS but I think he would give very
convincing philosophical arguments why the support for (possibly
proprietary) loadable modules within OpenSC, *in the core competence* of
OpenSC (enabling support for a card in the common driver), with the
current architectural goal (single driver with support for many cards),
serves anyone well.

Technically it was introduced to allow a certain country to enable
support for a card without exposing stuff (that turned up to still be
license violation). It would mean that we need to have a techncially
sufficient and stable API within OpenSC. The OpenSC API (sc_) was closed
down because nobody used it and because we wanted to promote
interoperability by using "standard" interfaces (like PKCS#11).

You talk about refactoring OpenSC, which it indeed badly needs. Having
more externally defined/depended constraints like another fixed API/ABI
makes it even harder to accomplish. The fact that drivers have not been
included in OpenSC codebase is a sign that it needs more developer
attention in maintaining the codebase rather than "this is too
complicated, lets make a loadable plugin hole here, delegate the
solution and just pretend the problem doesn't exists".

Also, there is no clear understanding about "OpenSC support": is it
personalization via pkcs15-tool, ability to create objects via PKCS#11
or it just means read only driver? Having support for one but not the
others if the card is in fact writable is not correct.


Third, prior attempts at vitalizing OpenSC have not been very fruitful.
What you do is great. I could even say that the state of OpenSC is to
some extent similar to OpenSSL: people just want to throw in stuff that
makes their things work and don't care that much of what else is there
in the pot. Refactoring and overall maintainability of the code is very
important and nobody else has bothered to do it except you. But please
keep in mind usability in addition to technical feasibility: having
loadable modules around doesn't really excite anyone except developers.
Having a easily installable piece of code that JustWorks out of the box
at the same time is good for both users as well as developers. Ideally
there should not be a configuration file except for weird setups or
temporary workarounds. Exposing a configurable that *needs* to be tuned
to get something working is bad. Exposing a configurable that can be
tuned but nobody does it, because the default Just Works, is just wasted
code but better.

Best,
Martin




On 11/02/15 22:35, Frank Morgner wrote:
> @philipWendland  You can create a second external pkcs15 driver. I think this answers your second and third question. I have done this here https://github.com/frankmorgner/vsmartcard/blob/d35e8e8de1783b39d9a4418adeaf25e953838169/npa/src/pkcs15-npa.c and configured the driver here https://github.com/frankmorgner/vsmartcard/blob/d35e8e8de1783b39d9a4418adeaf25e953838169/npa/opensc.conf.in. You basically need a function that implements the version check and declare your function that implements the pkcs15 emulation.
> 
> Yes I'd like to ship the external card driver with OpenSC (your first question). (But even if it wasn't shipped with OpenSC, you would simply have to edit your configuration file for testing it with a new version of OpenSC.)
> 
> Loading an external card driver is the mechanism that works and is available for an "opt-in" for the user. Currently, a big chunk of OpenSC's code is in a very bad shape. It's so bad that even core maintainers don't understand it (see https://github.com/OpenSC/OpenSC/commit/08eb700b9749c17e8ff29ddb45b3132d8d38308e#commitcomment-9691474 for example). Unmaintained card drivers are part of the problems, see for example #362. And still, OpenSC is the software that gets exclusive access to your key. This means that although you might use an EAL5+ certified chip, it's key is easily compromised. I want safe defaults, which allow safe use of some supported cards. If someone wants to go leading edge, he may "opt-in" (and should "start thinking"). Your fourth point allows safe use of OpenSC by "opt-out", but it's most likely not a safe default.
> 
> @martinpaljak my suggestion has been out now for quite some time. There came hardly any answer. If needed, we may continue the discussion on the mailing list...


-- 
Martin
+372 5156 495

frankmorgner:
Your comments are a bit contradictory: On the one hand you are saying that today OpenSC is meant to enable support for a speciffic card, on the other hand you want everything to work out of the box for users. *If* only a specific card should be supported, it always needs a specific setup. Hence, at least a configuration setup which is too complex for the average user. Even if the former approach of a generic smart card middleware with support is what the average user wants. Then he most likely only wants safe defaults, which means that *not everything* just works, but at least something does.

Anyway, I, indeed, achieved separation from insecure code in disabling all internal drivers and only use my external driver. Technically, referencing an external driver is identical to referencing a library function in libopensc. Following your complexity-argument this would mean that libopensc should not exist as dll anymore, and instead should be merged with the pkcs#11 library. I am not pedantic on this, if you want to add "blacklisting" instead, go ahead. I'm fine if the default blacklists the unsupported cards. (Whatever "supported/unsupported" means, see https://github.com/OpenSC/OpenSC/commit/11881a61b8247bde7618cf0f842bf7fc0285e803#commitcomment-9609067)

martinpaljak:
My thoughts were not well phrased apparently.
It is not contradiction, in fact "maintainers" or overall developer consensus must decide if OpenSC is either an orange or an apple: "toolkit for making your card work" or a "driver with support for many cards out of the box". One target allows for compile time option, configuration file tunings and what not (and probably results in different binary distributions), the other requires that things would JustWork out of the box for 95%+ users with the OpenSC binary distribution. One approach has a sacred "card driver" that must not be touched because it might break, the other keeps end-user satisfaction a priority and ruthlessly refactors everything (including card drivers) as necessary.

Actually I don't care if it would be apples or oranges, as long it is *something*, not a Schrödinger's cat.  It really would be best if was written down as a manifest somewhere, to help current and future contributors to align their activities. Lack of focus is bad. One of the reasons of removing libopensc as a separate entity was to give OpenSC more focus: drivers (PKCS#11, Minidriver, Tokend) for using with "useful apps" (which probably counts for 99+% card transactions) and command line tools for either personalization/card managemnt or activities that are not possible from the drivers. No point in having a library "just in case" especially if providing for external stability of the interface does not give much added value to the core. That's also why original purpose of OpenSC (the way I see and remember it) - to provide a single driver that would work with any supported smart card - contradicts to the "card driver inside card driver" loadable ABI interface: it creates a constraint (interface stability) that does not contribute to the project's goal. Maybe this  was (is?) the reasons for quite many forks of OpenSC some time ago.

I gave up trying to live by the original OpenSC promise (a single codebase and binary distribution that can work with any card you throw at it) by the introduction of the "cardmod" pseudo-pcsc-driver in OpenSC codebase, which is in essence a dumb code copypaste that indeed "works" but is BS from maintenance point of view. If there is no common value or shared goal in the project, it really makes it difficult. Kitchen sink where you throw your stuff that "worked for me, don't touch" does not contribute to a maintainable codebase that requires constant refactoring to keep it up to date and relevant in modern world. You end up with a codebase that people are only willing to work with if they get paid.

This much for a grumpy rant inspired by the [original comment](#issuecomment-67638721). Unless there are obvious technical obstacles or code review points, I don't see why IsoApplet driver should not be included in OpenSC. I have only briefly looked at the JavaCard part of IsoApplet, I've never been using OpenSC in personalization context (creating PKCS#15 structures) for more than generating keys and loading certificates onto pre-profiled cards anyway, usually without a filesystem abstraction. I still *use* OpenSC, but only for accessing ready made cards, my main development focus has shifted to JavaCard-s. Anyway, first looks at IsoApplet and related OpenSC patch looks *way* better than MuscleApplet, so ...

mouse07410:
On Feb 17, 2015, at 10:36 , Martin Paljak <notifications@github.com<mailto:notifications@github.com>> wrote:

My thoughts were not well phrased apparently.
It is not contradiction, in fact "maintainers" or overall developer consensus must decide if OpenSC is either an orange or an apple: "toolkit for making your card work" or a "driver with support for many cards out of the box". One target allows for compile time option, configuration file tunings and what not (and probably results in different binary distributions), the other requires that things would JustWork out of the box for 95%+ users with the OpenSC binary distribution. One approach has a sacred "card driver" that must not be touched because it might break, the other keeps end-user satisfaction a priority and ruthlessly refactors everything (including card drivers) as necessary.

I don’t see why it cannot be both: a “set of drivers supporting many cards out of the box”, and a “toolkit for making your card work”, assuming your card is not one of the “many cards” that already work out of the box. There is no contradiction.

In fact, I’d much prefer for OpenSC to keep being both.
--
Uri Blumenthal

frankmorgner:
Did everyone already forget about the 2010 incident where a rouge smart card enabled access to the machine via OpenSC (see https://github.com/OpenSC/OpenSC/wiki/OpenSC-security-advisories)? This will happen again if all that gets added are new features (cards). I think before adding new features, we need to agree on safe defaults ("supported cards"). However, limiting OpenSC to only a hand full of drivers that are supported *and* fulfill 80% of the average user's needs is today only possible with external drivers. So, before adding yet an other card, either the old separation mechanism must be implemented or a new must be added.

CardContact:
I don't quite understand how external drivers (that run in the same process space) mitigate the risk with a rouge smart card (which I would consider a very sophisticated and rather rare attac).

IMHO the problem is much more the bloated code and intransparent layering (e.g. why is there a PKCS#11 framework if there is only one framework implementation). This is the the reason why I never try to touch the core code - I just can't oversee what will break.

We've reimplemented a PKCS#11 module for a customer that previously tried to audit the OpenSC code base. At the end it was cheaper to reimplement the PKCS#11 module than it was to review the code.

As I said before, it would be good for the project if we'd focus on cards that are relevant for the user base and for which we have active maintainers. Let's disable the stuff nobody cares about and we are unable to test.

frankmorgner:
@CardContact To load an external driver (i.e. the dll and all its code), you need to explicitly add this driver to opensc.conf, which (by default) would be missing. The user needs to "opt-in".

I agree, if you have physical access to the machine (i.e. you can hold a smartcard near it), there might be other attacks that do work better today. But that might change tomorrow. Here we can only fix OpenSC (and there are a lot of things to fix ;-) ).

dengert:
On 2/18/2015 3:37 AM, CardContact Software & System Consulting wrote:
> I don't quite understand how external drivers (that run in the same process space) mitigate the risk with a rouge smart card (which I would consider a very sophisticated and rather rare attac).
>
> IMHO the problem is much more the bloated code and intransparent layering (e.g. why is there a PKCS#11 framework if there is only one framework implementation). This is the the reason why I never try
> to touch the core code - I just can't oversee what will break.
>
> We've reimplemented a PKCS#11 module for a customer that previously tried to audit the OpenSC code base. At the end it was cheaper to reimplement the PKCS#11 module than it was to review the code.
>
> As I said before, it would be good for the project if we'd focus on cards that are relevant for the user base and for which we have active maintainers. Let's disable the stuff nobody cares about and
> we are unable to test.

How do you know what cards are still being used? Turning off some card might surprise end user when they
get their system updated...

One way would be if the card driver does not support RSA 2048, the users should be upgrading their cards...


>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/305#issuecomment-74836495>.
>

-- 

dengert:
On 2/18/2015 4:00 AM, Frank Morgner wrote:
> @CardContact <https://github.com/CardContact> To load an external driver (i.e. the dll and all its code), you need to explicitly add this driver to opensc.conf, which (by default) would be missing.
> The user needs to "opt-in".

Yes, opt-in, and if a user really needs a driver let us know, as it may be dropped in the future.
We need a "deprecated and will be removed unless we hear from you" strategy.

>
> I agree, if you have physical access to the machine (i.e. you can hold a smartcard near it), there might be other attacks that do work better today. But that might change tomorrow. Here we can only
> fix OpenSC (and there are a lot of things to fix ;-) ).
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/305#issuecomment-74839461>.
>

-- 

CardContact:
>How do you know what cards are still being used? Turning off some card might surprise end user
>when they get their system updated... 
I don't know, but I know which cards we can test with. What's the point of supporting cards that we are not able to test with ?

martinpaljak:
Maybe we should leave IsoApplet as a new contender into highly relevant on-card applets alone and move the "deprecation and refactoring" discussion into a separate ticket ?

dengert:
On 2/18/2015 8:09 AM, CardContact Software & System Consulting wrote:
>     How do you know what cards are still being used? Turning off some card might surprise end user
>     when they get their system updated...
>     I don't know, but I know which cards we can test with. What's the point of supporting cards that we are not able to test with ?

That is always an issue with open source. We let developers submit drivers where they may be the only one interested in the driver
They lose interest and move on. But users don't know that. So what does it mean to support a card?
(You know, I am retired. I used to get paid to do this. I do this for free now because it is interesting.)


>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/305#issuecomment-74868895>.
>

-- 

CardContact:
> So what does it mean to support a card?

It means that we (developers and contributors) have such a card, can do regression tests with it and can resolve issues raised by users. I don't see any other way to ensure that code changes and new releases are still work with the supported card base. And yes, maybe we get response from a user that has one of the older cards - how do we deal with that ? Simple: We ask him to become a maintainer for this card and do tests every once in while. That way we know a) if there is a strong demand for supporting that card and b) if there are still a couple of cards in circulation.

I have a desk full of smart cards from the past 25 years working in the industry. Very few of them would still work with current software.

viktorTarasov:
As for me I do not see objections to include IsoApplet driver into base code if it do not need hard changes in common part and there is someone who can test it and maintain.

I would not like to make a choice between apple and orange,
and continue to see OpenSC project as a driver with support of many card, together with common modules (PKCS#11, minidriver) and card tools, usability oriented, with default configuration sufficient to use it *out of the box*. 

I like idea from Philip to blacklist the unsupported cards in default configuration. It's a suitable *soft* solution for transition period. During this period the blacklisted driver can be allowed by changing configuration of installed SW.

philipWendland:
I rebased upon the current master because of #381 (I will still need to investigate how fed84c464980969d90770fc3953a9f5717774d8c can be beneficial). The Jenkins error seems to be due to:
"light.exe : error LGHT0001 : There is not enough space on the disk."
https://opensc.fr/jenkins/job/OpenSC-pr-win64/73/console

viktorTarasov:
@JenkinsForOpenSC : retest this please

viktorTarasov:
@JenkinsForOpenSC : retest this please

philipWendland:
Thank you very much.
I did not have time to adjust to 2abe135f971aca7a3ecc231d10f96b5fddee078a which changed the signature of sc_asn1_sig_value_sequence_to_rs(). This causes compiler warnings and most probably segfaults when doing ECDSA. I will provide a patch next week.

PR #308, opened 2014-10-25
Building on newer OSX versions without xcode command line tools

martinpaljak:
Somebody please review the autoconf change, it should be well isolated though.

This allows to build on OSX 10.10 as well (also applies to tokend, which still works, to some extent)


carlhoerberg:
Isn't working on 10.9 or 10.8 with Xcode command line tools. On my machine the libs are only available in `/System/Library/Frameworks/PCSC.framework/Headers`.

carlhoerberg:
Fails for homebrew's build server too: https://github.com/Homebrew/homebrew/pull/40737

Issue #335, opened 2014-12-03
Improve PKCS#11 compliance


tc-anssi:
As mentioned in 
http://comments.gmane.org/gmane.comp.encryption.opensc.devel/15647, OpenSC has
several issues related to the PKCS#11 standard.  For instance, the
standard specify the value of attributes such as `CKA_LOCAL` to reflect a specific state of objects. The
current implementation hard-code this value to `TRUE`. This is an issue in the case of imported objects (generated off-board) were it must be set to `FALSE`. Other important values such as `CKA_ALWAYS_SENSITIVE`
and `CKA_NEVER_EXTRACTABLE` are similarly set to a hard-coded value.

OpenSC should fetch real values from the underlying cards and expose them to PKCS#11 client applications.

viktorTarasov:
Also discussed in #336, #338;
applied in 4915eaa56b4e81346e1696b2cb38a0f2df14db5b, bbedd4ef00bf457fdf11be7899c64579248885e1, facaf5969bc80ac82395fa2789df32c48e373085.

Issue #336, opened 2014-12-03
IAS-ECC: improve PKCS#11 compliance


tc-anssi:
The current IAS-ECC code wrongly set the `CKA_ALWAYS_SENSITIVE` and 
`CKA_NEVER_EXTRACTABLE` attributes to `TRUE` when importing objects on the card.
Those attributes should only be set when the key-pair was generated on-board.

viktorTarasov:
Resolved in #348

PR #348, opened 2015-01-05
Fix issues #335 and #336, missing commits

tc-anssi:
Hi,

Regarding the two issues mentioned in the title, it seems two commits were missing from my previous pull requests.
The first missing commit enables the PKCS#11 compliance improvements otherwise `access_flags` is not read.
The second commit is related to `access_flags` for IAS-ECC cards.

Sorry I missed that!

Cheers.


viktorTarasov:
Agree with  <i>Improve PKCS#11 compliance</i>  (917653a).

Do not agree with <i>IAS-ECC: improve PKCS#11 compliance</i> ( 0d58595),
look my comment to #336.

tc-anssi:
Hi Viktor,

I re-read your comment on #336 and I have to disagree with your interpretation of CKA_ALWAYS_SENSITIVE and CKA_NEVER_EXTRACTABLE.

Let's take for instance the C_CreateObject() function, the standard states (>=v2.20):
> If C_CreateObject is used to create a key object, the key object will have its
> CKA_LOCAL attribute set to CK_FALSE. If that key object is a secret or private key
> then the new key will have the CKA_ALWAYS_SENSITIVE attribute set to
> CK_FALSE, and the CKA_NEVER_EXTRACTABLE attribute set to CK_FALSE.

Therefore, the value of those two attributes depends on the type of operation that was used to "instantiate" the object.

Now, the two attributes can safely be set to CK_TRUE when the key is generated on-board since OpenSC does not allow wrapping operations. This is what I did with (0d58595).

viktorTarasov:
> I re-read your comment on #336 and I have to disagree with your interpretation of
> CKA_ALWAYS_SENSITIVE and CKA_NEVER_EXTRACTABLE.


You have a reason.
Thank you.

viktorTarasov:
Applied in 4915eaa56b4e81346e1696b2cb38a0f2df14db5b and bbedd4ef00bf457fdf11be7899c64579248885e1,
thank you.

dengert:
On 1/5/2015 6:17 AM, Thomas Calderon wrote:
> Hi Viktor,
>
> I re-read your comment on #336 <https://github.com/OpenSC/OpenSC/issues/336> and I have to disagree with your interpretation of CKA_ALWAYS_SENSITIVE and CKA_NEVER_EXTRACTABLE.
>
> Let's take for instance the C_CreateObject() function, the standard states (>=v2.20):
>
>     If C_CreateObject is used to create a key object, the key object will have its
>     CKA_LOCAL attribute set to CK_FALSE. If that key object is a secret or private key
>     then the new key will have the CKA_ALWAYS_SENSITIVE attribute set to
>     CK_FALSE, and the CKA_NEVER_EXTRACTABLE attribute set to CK_FALSE.
>
> Therefore, the value of those two attributes depends on the type of operation that was used to "instantiate" the object.
>
> Now, the two attributes can safely be set to CK_TRUE when the key is generated on-boardsince OpenSC does not allow wrapping operations.

I agree with the conclusion, if the token enforces not reading a key that was generated on the card, but disagree with the logic:
"since OpenSC does not allow wrapping operations.". The attribute has to do with the token not with the middle-ware
currently accessing the token. If the IAS/ECC card do wrapped extraction, even if OpenSC does not support wrapping,
then the CKA_NEVER_EXTRACTABLE should be false.


If the OpenSC code has to determine what the attribute should be, other then reading it from the token,
then it maybe guessing at what the card issuer actually did, for example escrowing a key,
or generating the key off the card, and writing it to the card using some card vendor commands
that may not be documented. So some of these attributes may depend on the policies of the
card issuer.

So what can the IAS/ECC token do?
What can a card issuer set?
What attributes of access flags can be read from the token?



This is what I did with (0d58595
> <https://github.com/OpenSC/OpenSC/commit/0d585952541e6667c08ad4ae6accc1f5e17766fb>).
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/348#issuecomment-68700801>.
>

-- 

viktorTarasov:
Thomas cited the PKCS11 standard:
 >    If C_CreateObject is used to create a key object, the key object will have its
>     CKA_LOCAL attribute set to CK_FALSE. If that key object is a secret or private key
>     then the new key will have the CKA_ALWAYS_SENSITIVE attribute set to
>     CK_FALSE, and the CKA_NEVER_EXTRACTABLE attribute set to CK_FALSE.

My updated IMHO:
these three attributes are related to the hole key history;
in this logic they are set to FALSE when object is created with C_CreateObject 
(once upon a time the key material was outside the card).

Other two attributes
CKA_EXTRACTABLE and CKA_SENSITIVE are related to the current object state or token capacity.
For IAS/ECC card private key cannot be revealed off card; so these attributes are set to FALSE and TRUE.

PR #351, opened 2015-01-15
Add support for ACOS5-64 cards.

darconeous:
The ACOS5-64 cards have a different ATR than the original
ACOS5-32 cards. This change simply adds this ATR so that it
will be recognized properly.

### DIFF
--- a/src/libopensc/card-acos5.c
+++ b/src/libopensc/card-acos5.c
@@ -26,6 +26,8 @@
 #include "cardctl.h"
 
 static struct sc_atr_table acos5_atrs[] = {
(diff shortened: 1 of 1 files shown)


martinpaljak:
IIRC the ACOS support in OpenSC was only on filesystem level, even
read-only crypto operations were not supported? In addition to adding the
ATR, have you tried using the card for key based operations, with positive
results?

--
Martin
+372 515 6495

darconeous:
It seems I was a bit confused by the state of things with the ACOS5 card.

I'm guessing that no one is working on getting ACOS5 to be feature complete?

viktorTarasov:
@JenkinsForOpenSC : retest this please

viktorTarasov:
applied in 548c2780d3faf9419c09aea5d5909b5d82685515,
thank you.

PR #357, opened 2015-01-19
Added initial support for STARCOS 3.4 (German D-Trust cards).

fancycode:
Code inspired by experimental STARCOS 3 driver by Martin Vogt:
http://article.gmane.org/gmane.comp.encryption.opensc.devel/9846

Cards available from `https://www.bundesdruckerei.de/en/768-d-trust-signature-cards`.

Supports `opensc-explorer`, `pkcs15-tool` and `pkcs15-crypt` (signing with PKCS#1 padding).

Tested using
```
opensc-explorer
pkcs15-tool --dump
pkcs15-tool --list-applications
pkcs15-tool --list-keys
pkcs15-tool --list-certificates
pkcs15-tool --read-certificate <ID>
pkcs15-crypt -s -k <ID> --sha-<BITS> --pkcs1 <arguments>
```
Tested signing with SHA1, SHA224, SHA256, SHA384 and SHA512 (only signing with PKCS#1 padding supported for now).

Feedback welcome!

### DIFF
--- a/src/libopensc/card-starcos.c
+++ b/src/libopensc/card-starcos.c
@@ -32,6 +32,8 @@ static struct sc_atr_table starcos_atrs[] = {
 	{ "3B:B7:94:00:c0:24:31:fe:65:53:50:4b:32:33:90:00:b4", NULL, NULL, SC_CARD_TYPE_STARCOS_GENERIC, 0, NULL },
 	{ "3B:B7:94:00:81:31:fe:65:53:50:4b:32:33:90:00:d1", NULL, NULL, SC_CARD_TYPE_STARCOS_GENERIC, 0, NULL },
 	{ "3b:b7:18:00:c0:3e:31:fe:65:53:50:4b:32:34:90:00:25", NULL, NULL, SC_CARD_TYPE_STARCOS_GENERIC, 0, NULL },
+	/* STARCOS 3.4 */
+	{ "3b:d8:18:ff:81:b1:fe:45:1f:03:80:64:04:1a:b4:03:81:05:61", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_4, 0, NULL },
 	{ NULL, NULL, NULL, 0, 0, NULL }
 };
 
@@ -39,7 +41,7 @@ static struct sc_card_operations starcos_ops;
 static struct sc_card_operations *iso_ops = NULL;
 
 static struct sc_card_driver starcos_drv = {
-	"STARCOS SPK 2.3/2.4",
+	"STARCOS SPK 2.3/2.4/3.4",
 	"starcos",
 	&starcos_ops,
 	NULL, 0, NULL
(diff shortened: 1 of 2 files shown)


src/libopensc/card-starcos.c
```diff
@@ -215,6 +251,182 @@ static int process_fci(sc_context_t *ctx, sc_file_t *file,
...
+	if (buflen < 2)
+		return SC_ERROR_INTERNAL;
+	if (buf[0] != 0x62)
+		return SC_ERROR_INVALID_DATA;
+	len = (size_t)buf[1];
+	if (buflen - 2 < len)
+		return SC_ERROR_INVALID_DATA;
+	p = buf + 2;
+
+	tag = sc_asn1_find_tag(ctx, p, len, 0x80, &taglen);
```
frankmorgner: Please use defines instead of hard coded values

frankmorgner:
Could you tested in a pkcs#11 application or via minidriver?

frankmorgner:
Did you check backwards compatibility with starcos 2.4?

src/libopensc/card-starcos.c
fancycode: I would love to give these numbers some meaningful defines, but that's difficult without documentation of the APDUs. The code is based on the STARCOS 3 patch from Martin Vogt and some trial-and-error to determine the correct numbers for STARCOS 3.4.

Besides, the original STARCOS 2.3/2.4 code also has lots of magic numbers everywhere :-)

fancycode:
I didn't test with PKCS#11/minidriver yet but will try to do so. Any suggestions which app to test?

For testing backwards compatibility I would need a STARCOS 2.4 compatible card (which I don't have). However the new code paths should only be executed for 3.4 cards with the old behaviour for 2.3/2.4 cards.

frankmorgner:
there is `pkcs11-too --test`. You could also try setting up libreoffice or Adobe's acrobat reader for signing documents.

frankmorgner:
@fancycode did you try PKCS#11?

fancycode:
I played around a bit with the tools included in OpenSC but didn't test any third-party applications. We are only using the PKCS#15 part for signing/pin validation, so unfortunately I didn't have too much more time for additional tests after that was working...

frankmorgner:
what about `pkcs11-tool --test`?

fancycode:
```bash
$ ./src/tools/pkcs11-tool --module=src/pkcs11/.libs/opensc-pkcs11.so --test
Using slot 1 with a present token (0x1)
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
Signatures: not logged in, skipping signature tests
Verify: not logged in, skipping verify tests
Key unwrap: not a R/W session, skipping key unwrap tests
Decryption: not logged in, skipping decryption tests
No errors
```

I tried to run `--test` with `--login` but it always asks for my *User PIN* which can't be used for signing (fails with `error: PKCS11 function C_SignFinal failed: rv = CKR_GENERAL_ERROR (0x5)`). I couldn't find out how to make it ask for my *Signature PIN*. Signing with `pkcs15-crypt` and reading the certificate with `pkcs15-tool` works.

frankmorgner:
with `pkcs11-tool` you could have a look at `--list-slots` and `--list-objects`, which may show something. Also, you may want to check out `pkcs15-tool --dump` to see if the signature pin is associated with the signature key.

frankmorgner:
On the command line you can change to the signature PIN by using `--auth-id`. However, that's only possible with the CLI, but it won't be possible in an email client.

fancycode:
Thanks, found the problem. The signature cert is in the second slot.

I can successfully run the tests now:
```bash
$ ./src/tools/pkcs11-tool --module=src/pkcs11/.libs/opensc-pkcs11.so --test --login --slot 2 --pin MYPIN
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
Signatures (currently only RSA signatures)
  testing key 0 (SigG Signature Key) 
  testing signature mechanisms:
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
Verify (currently only for RSA):
  testing key 0 (SigG Signature Key)
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
Unwrap: not implemented
Decryption (RSA)
  testing key 0 (SigG Signature Key)  -- can't be used to decrypt, skipping
No errors
```

However for this to work I had to comment out the checks for `getALWAYS_AUTHENTICATE` in `pkcs11-tool.c` in `test_signature`, so it always logged in again after performing a signature. Is there a way to enforce this from the card driver?

frankmorgner:
Whether `getALWAYS_AUTHENTICATE` returns true or false depends on the PKCS15 object's flag `userConsent`. You could enable debugging to see if this flag is set for your key or not. Obviously it is not correctly set for your card... However, it should have nothing to do with your changes in the card driver.

unrzn0:
Does this driver also support STARCOS 3.0 and 3.2 cards? When looking at the code with my naïve thinking it seems like the code right now has no checks for STARCOS 3.0 and 3.2 an therefore fails to identify those.

Issue #364, opened 2015-01-26
ePass2003 concurrency/session issues


martin-css:
PKCS11 support for the Feitian ePass2003 token currently has major concurrency issues that prevent effective use of this token in real life application (at least in an Ubuntu environment). This relates to version 0.13 of OpenSC which is the latest version in the official Ubuntu repositories.

PKCS11 functionality will work fine when only a single application has initialised a session. When another application uses the library, this will also work as expected. However, the first applications session will now be invalid and any functions called will fail.

I've created a [Vagrant test image](https://github.com/martin-css/epass2003-test) that allows easy replication of this issue for diagnosis.

Please see the following PKCS11 spy logs that illustrate the test case in the above test environment.

[SSH Agent Initiailsation] (http://pastebin.com/WTXbfymC)
[Successful SSH signing](http://pastebin.com/taEVr6Wg)

The PKCS11 utility is then used as a simple way of causing another application to use the library. This is allowed to complete successfully. Subsequent calls by ssh-agent will now fail as per the example log below:

[Failed SSH signing](http://pastebin.com/Z6iwq1fr)

martin-css:
Please note this issue is still active as per latest master state at time of writing - commit b13bb3cbf53fede96e5d397329fd2ece18b98d0c.

martinpaljak:
PKCS#11 spy log is useful, but not enough to debug OpenSC. Please provide a (failing, at least) OpenSC debug log at sufficient level (try 9).

martin-css:
As far as I can see from the source code, the PKCS11 library does not have any option to specify a higher debug level. It's not a OpenSC tool that's failing, it's any application that uses the library, in my example ssh-agent.

I can provide a log of the pcscd daemon run in debug mode if this is helpful?

Alternatively, anyone who owns one of these tokens can use the test environment linked in the first post to replicate the issue.

frankmorgner:
use the environment variable OPENSC_CONF to specify a configuration file with a higher debug level

martin-css:
I've created level 9 debug log output for the initialisation, successful signing and failed signing so these can be compared. I've also re-included the spy log for this run in case this is useful.

## SSH Agent Initialisation

[OpenSC Debug Log](http://pastebin.com/wnTw3PQS)
(PKCS11 Spy Log)[http://pastebin.com/Hgab3eEV]

## Successful SSH signing

[OpenSC Debug Log](http://pastebin.com/NbMDzLz8)
[...]

PR #365, opened 2015-01-28
Fix for ePass2003 smart card token concurrency issues

martin-css:
Please see issue #364 for discussion of original problem and explanation of changes.

### DIFF
--- a/src/libopensc/card.c
+++ b/src/libopensc/card.c
@@ -334,6 +334,8 @@ int sc_lock(sc_card_t *card)
 				/* invalidate cache */
 				memset(&card->cache, 0, sizeof(card->cache));
 				card->cache.valid = 0;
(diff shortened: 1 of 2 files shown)


viktorTarasov:
applied in da0a88b7a040a55f6a91be3116e4ebc41b91f3be,
thank you

src/libopensc/card.c
```diff
@@ -334,6 +334,8 @@ int sc_lock(sc_card_t *card)
 				/* invalidate cache */
 				memset(&card->cache, 0, sizeof(card->cache));
 				card->cache.valid = 0;
+				if (card->sm_ctx.ops.open)
```
frankmorgner: shouldn't this be using `sc_card_sm_check` instead of *always* opening a SM channel?

PR #367, opened 2015-01-28
PKCS#11: Improve compliance for public keys.

tc-anssi:
* Key usage required when calling C_CreateObject for public key were not used.
    This adds the logic to convert from PKCS#11 attributes to PKCS#15 in the
    structure that is sent to underlying _store_pubkey functions.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -2274,18 +2274,18 @@ pkcs15_create_public_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
 			bn = &rsa->modulus; break;
 		case CKA_PUBLIC_EXPONENT:
 			bn = &rsa->exponent; break;
-        case CKA_VERIFY:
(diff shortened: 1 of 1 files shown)


tc-anssi:
Hi,

I did not open an issue which would be linked to this pull request, should I do so ?

src/pkcs11/framework-pkcs15.c
```diff
@@ -2274,6 +2274,18 @@ pkcs15_create_public_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
 			bn = &rsa->modulus; break;
 		case CKA_PUBLIC_EXPONENT:
 			bn = &rsa->exponent; break;
+        case CKA_VERIFY:
```
frankmorgner: please use tabs instead of spaces

frankmorgner:
This PR is enough, no issue required.

viktorTarasov:
applied in 22fb49b5545a09bf7b4be071ef189388e6a582b1,
thank you

PR #368, opened 2015-01-29
Prefer serial number from EF(TokenInfo) and cached serials over CARDCTL

philipWendland:
sc_pkcs15_get_object_guid() should prefer the serial number from
EF(TokenInfo) over card->serialnr because the user may override the
card's serial number with "pkcs15-init -C --serial". The card->serialnr
should be used before calling card CTL with SC_CARDCTL_GET_SERIALNR
because it might contain a cached value.

This change *should* allow cards to be used with the minidriver even if
SC_CARDCTL_GET_SERIALNR is not implemented. For example, a driver might
set card->serialnr at initialization or the user might initialize a
PKCS#15 card with its own "--serial".

See also the discussion at the mailing list:
[Opensc-devel] AT_SIGNATURE and AT_EXCHANGE Problem


viktorTarasov:
applied in faaa11e091001f063dac46f29f0aad28c2494a18,
thank you

PR #369, opened 2015-02-02
Encode EC domain parameters when public key encoded as SPKI

viktorTarasov:


philipWendland:
Hello,
FYI - this should make #304 obsolete, except for:
```
[philip@philip-pc OpenSC]$ git diff HEAD~1..HEAD
diff --git a/src/pkcs15init/pkcs15-lib.c b/src/pkcs15init/pkcs15-lib.c
index 1e7b7a4..a8569a2 100644
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -1505,7 +1505,8 @@ sc_pkcs15init_store_public_key(struct sc_pkcs15_card *p15card, struct sc_profile
        case SC_ALGORITHM_EC:
                type = SC_PKCS15_TYPE_PUBKEY_EC;
 
-               key.u.ec.params = keyargs->params.ec;
+               if (!key.u.ec.params.named_curve)
+                       key.u.ec.params = keyargs->params.ec;
                sc_pkcs15_fix_ec_parameters(ctx, &key.u.ec.params);
 
                keybits = key.u.ec.params.field_length;
diff --git a/src/tools/pkcs15-init.c b/src/tools/pkcs15-init.c
index aa3474a..78875ae 100644
--- a/src/tools/pkcs15-init.c
+++ b/src/tools/pkcs15-init.c
@@ -900,6 +900,11 @@ do_store_private_key(struct sc_profile *profile)
        r = sc_pkcs15_convert_prkey(&args.key, pkey);
        if (r < 0)
                return r;
+       if (args.key.algorithm == SC_ALGORITHM_EC) {
+               sc_pkcs15_fix_ec_parameters(ctx, &args.key.u.ec.params);
+               args.params.ec = args.key.u.ec.params;
+       }
+
        init_gost_params(&args.params.gost, pkey);
 
        if (ncerts) {
```
If I recall correctly, the first was necessary because of the duplication of the EC params in params.ec and key.u.ec.params, the key.u.ec.params would have been overwritten by keyargs->params.ec even if the information in keyargs is incomplete.

The second is needed to fill the parameters after reading the private key from the file. This could maybe also be done in pkcs15-prkey.c - sc_pkcs15_convert_prkey()

viktorTarasov:
Thanks, I will take into consideration .

philipWendland:
Thanks, I (successfully) did some tests with IsoApplet - This PR combined with #371 seems to cover everything in #296 and #304 (when calling sc_pkcs15_fix_ec_parameters() at the pkcs15init-driver when importing keys).

viktorTarasov:
applied in e374f88b82b047a55ce557f3e90a80b76eb0108e

PR #371, opened 2015-02-03
pkcs15init: superpluous 'ec-params' in init data

viktorTarasov:
Pkcs15init data, used to import/generate key objects with <I>pkcs15-init</i> tool,
include twice the same <i>EC parameters</i> data:
 - explicit <i>params</i> member in <i>pkcs15init-args</i> data
 - part of <i>sc_pkcs15_pubkey/sc_pkcs15_prkey</i>

Explicit <i>params</i> is removed.


viktorTarasov:
applied in 96265e6d08a0be51d7d83468d9e42bd54a7b409a

PR #372, opened 2015-02-05
Caching patches

akemnade:


src/libopensc/ctx.c
```diff
@@ -860,9 +860,18 @@ int sc_get_cache_dir(sc_context_t *ctx, char *buf, size_t bufsize)
 {
 	char *homedir;
 	const char *cache_dir;
+        scconf_block *conf_block = NULL;
 #ifdef _WIN32
 	char temp_path[PATH_MAX];
 #endif
+        conf_block = sc_get_conf_block(ctx, "framework", "pkcs15", 1);
```
frankmorgner: bad indenting

src/tools/pkcs15-tool.c
```diff
@@ -1515,11 +1515,47 @@ static int read_and_cache_file(const sc_path_t *path)
 		printf("out of memory!");
 		return -1;
 	}
-	r = sc_read_binary(card, 0, buf, size, 0);
-	if (r < 0) {
-		fprintf(stderr, "sc_read_binary() failed: %s\n", sc_strerror(r));
-		free(buf);
-		return -1;
+	if (tfile->ef_structure == SC_FILE_EF_LINEAR_VARIABLE_TLV) {
```
frankmorgner: not related to the commit message above. put it in a seperate commit.

frankmorgner: some cards don't return metadata about the files so there is no way of knowing whether its a `SC_FILE_EF_LINEAR_VARIABLE_TLV`. This change would render pkcs15-tool unusable for many cards.

src/libopensc/ctx.c
```diff
@@ -860,9 +860,18 @@ int sc_get_cache_dir(sc_context_t *ctx, char *buf, size_t bufsize)
...
+	cache_dir = scconf_get_str(conf_block, "file_cache_dir", NULL);
```
frankmorgner: You should provide a reasonable default here. Presumably the caching-directory we used to have...

src/tools/pkcs15-tool.c
akemnade: On Thu, 05 Feb 2015 18:10:16 -0800
Frank Morgner <notifications@github.com> wrote:

> > @@ -1515,11 +1515,47 @@ static int read_and_cache_file(const sc_path_t *path)
> >  		printf("out of memory!");
> >  		return -1;
> >  	}
> > -	r = sc_read_binary(card, 0, buf, size, 0);
> > -	if (r < 0) {
> > -		fprintf(stderr, "sc_read_binary() failed: %s\n", sc_strerror(r));
> > -		free(buf);
> > -		return -1;
> > +	if (tfile->ef_structure == SC_FILE_EF_LINEAR_VARIABLE_TLV) {
> 
> some cards don't return metadata about the files so there is no way of knowing whether its a `SC_FILE_EF_LINEAR_VARIABLE_TLV`. This change would render pkcs15-tool unusable for many cards.
> 

I do not understand that argument. The code I was changing is only hit during
pkcs15-tool -L  (hence the commit message)
What does it make worse?
We have three cases here:
1. we know we have a binary file (which can be read using sc_read_binary())
2. we know we have a record based file
3. we do not know what kind of file we have.

without the patch we handle 1. correctly and assume we have a binary file in 3.
With the patch we handle 1. and 2. correctly and assume we have a binary file in 3.

src/libopensc/ctx.c
akemnade: On Thu, 05 Feb 2015 18:15:26 -0800
Frank Morgner <notifications@github.com> wrote:

> >  #ifdef _WIN32
> >  	char temp_path[PATH_MAX];
> >  #endif
> > +        conf_block = sc_get_conf_block(ctx, "framework", "pkcs15", 1);
> > +	cache_dir = scconf_get_str(conf_block, "file_cache_dir", NULL);
> 
> You should provide a reasonable default here. Presumably the caching-directory we used to have...
> 

if the caching-directory is not found, the old one will be used
there is.
  if (cache_dir != NULL) {

The reason why I did not put it into the scconf_get_str() line was:
The default one is not a string constant. It is constructed. So why the
default path should be constructed if the result is not used.

frankmorgner: this part looks ok, sorry for the noise. I overlooked that the function is preemptively exited if the conf value was found

src/tools/pkcs15-tool.c
frankmorgner: sorry, again. the logic to handle record and binary files looks ok.

src/tools/pkcs15-tool.c
```diff
@@ -1515,11 +1515,47 @@ static int read_and_cache_file(const sc_path_t *path)
...
+				if (r < 4)
+					break;
+				memmove(head,head+4,r-4);
+				head += (r-4);
+			}
+		}
+		r = head - buf;
+		
+	} else {		
+
```
frankmorgner: bad whitespace

src/tools/pkcs15-tool.c
```diff
@@ -1515,11 +1515,47 @@ static int read_and_cache_file(const sc_path_t *path)
 		printf("out of memory!");
 		return -1;
 	}
-	r = sc_read_binary(card, 0, buf, size, 0);
-	if (r < 0) {
-		fprintf(stderr, "sc_read_binary() failed: %s\n", sc_strerror(r));
-		free(buf);
-		return -1;
+	if (tfile->ef_structure == SC_FILE_EF_LINEAR_VARIABLE_TLV) {
+		int i;
```
frankmorgner: please move declarations to the top of the function

martinpaljak:
Rather than making it configurable (configuration should exist so that people can modify it and there should be a realistic *need* to modify it) maybe it makes sense to define the behaviour of caching in case $HOME is missing (which would assume running in "system" configuration) and set sensible defaults, like caching under /var somewhere.

Sensible defaults is a must have, exposing a user or system administrator to a mini-language in the form of opensc.conf when there is no need for it to remain functioning as expected is not a good target.

Right now the problem seems to be that when $HOME is missing, caching doesn't work, rather than there exists a definite need to configure the caching folder. Also, there is no clear understanding if this should be a user preference problem ("I want to have my caches on ~/Desktop") or a system policy issue ("caches must go to folder X").

frankmorgner:
One general remark: Generating the cache uses copy/paste code from sc_pkcs15_read_file (which also includes support for variable TLV EFs). when generating the cache, please refactor pkcs15-tool so that it uses sc_pkcs15_read_file to avoid code duplication.

akemnade:
On Fri, 06 Feb 2015 01:30:26 -0800
Frank Morgner <notifications@github.com> wrote:

> One general remark: Generating the cache uses copy/paste code from sc_pkcs15_read_file (which also includes support for variable TLV EFs). when generating the cache, please refactor pkcs15-tool so that it uses sc_pkcs15_read_file to avoid code duplication.
> 
I cannot directly use sc_pkcs15_read_file() but I think a separate sc_pkcs15_get_file_content()
would help here.
the function in pkcs15-tool can then do:
select_file();
if (file_is_public) {
  sc_pkcs15_get_file_content();
}
store_file_in_cache();

sc_pkcs15_read_file() can then do a
if (use_cache) get_from_cache()
else {
  select_file();
  sc_pkcs15_get_file_content();
}

PR #374, opened 2015-02-09
Travis OSX integrations

martinpaljak:
So that all changes and pull requests would be build on Travis as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: C
+os:
+  - linux
+  - osx
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,3 +57,6 @@ script:
       make check;
       make dist;
     fi
+  - if [ $TRAVIS_OS_NAME == osx ]; then
+      ./MacOSX/build;
+    fi


martinpaljak:
As OSX only has clang and no mingw the build machinery should be adjusted slightly. For completeness the Tokend should also be built (maybe OSX "single build file" be split into two steps, building OpenSC + tokend + building the full .dmg?)

frankmorgner:
Could you instead use matrix include to build a speciffic configuration of osx to omit gcc/mingw on osx. Also, you should disable documentation on osx as the requirements are not met.

martinpaljak:
I need to look deeper into the build matrix setup of travis, no experience with that yet. If anyone feels like it, go ahead.

frankmorgner:
unforunately I had the same problem. Its try and error. However, you can always force an update on your branch which also updates this PR. Make sure to test your configuration with http://lint.travis-ci.org/

frankmorgner:
on github you could also search for `filename:.travis.yml ...` for similar solutions.

frankmorgner:
I'll try in #420

PR #375, opened 2015-02-09
Coverity

frankmorgner:
despite my expectations, some of the recent commits really got reviewed ;-) . So here is the next chunk of fixes.

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -89,9 +89,10 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
 			 * unsigned int */
 			return SC_ERROR_INVALID_ASN1_OBJECT;
 
+	} else {
+		if (left == 0)
+			return SC_ERROR_INVALID_ASN1_OBJECT;
 	}
-	if (left == 0)
-		return SC_ERROR_INVALID_ASN1_OBJECT;
 	/* parse length byte(s) */
 	len = *p & 0x7f;
 	if (*p++ & 0x80) {
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -1441,6 +1441,8 @@ static int asn1_decode(sc_context_t *ctx, struct sc_asn1_entry *asn1,
 				left, depth,
 				choice ? ", choice" : "");
 
+	if (!p)
+		return SC_ERROR_ASN1_OBJECT_NOT_FOUND;
 	if (left < 2) {
 		while (asn1->name && (asn1->flags & SC_ASN1_OPTIONAL))
 			asn1++;
@@ -1789,6 +1791,8 @@ _sc_asn1_decode(sc_context_t *ctx, struct sc_asn1_entry *asn1,
 int
 sc_der_copy(sc_pkcs15_der_t *dst, const sc_pkcs15_der_t *src)
 {
+	if (!dst)
+		return SC_ERROR_INVALID_ARGUMENTS;
 	memset(dst, 0, sizeof(*dst));
 	if (src->len) {
 		dst->value = malloc(src->len);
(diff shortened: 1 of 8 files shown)


src/libopensc/card-piv.c
```diff
@@ -2031,14 +2031,16 @@ static int piv_get_serial_nr_from_CHUI(sc_card_t* card, sc_serial_number_t* seri
...
 					gbits = 0; /* set to skip using guid below */
 				}
 			}
 			if (guid && gbits) {
-				serial->len = guidlen < SC_MAX_SERIALNR ? guidlen : SC_MAX_SERIALNR;
+				/* with gbits != 0 guidlen==16 < SC_MAX_SERIALNR */
+				serial->len = guidlen;
 				memcpy (serial->value, guid, serial->len);
 				r = SC_SUCCESS;
 			}
```
dengert: The original code was still correct. 

The change appears to make the assumption the fascnlen  (25) is always less then SC_MAX_SERIALNR (32) which is true as long as  the difinition of SC_MAX_SERIALNR does not change. 
If it did, there is a chance of over writing in line 2036.  

Same goes for guidlen (16) and SC_MAX_SERIALNR  and overwrite in line 2044. 

If SC_MAX_SERIALNR was to change, would  Conerity  catch the possible overwrite?
Does a good optimizing compiler optimize out the code?

src/libopensc/asn1.c
```diff
@@ -89,9 +89,10 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
...
 
+	} else {
+		if (left == 0)
+			return SC_ERROR_INVALID_ASN1_OBJECT;
 	}
-	if (left == 0)
-		return SC_ERROR_INVALID_ASN1_OBJECT;
 	/* parse length byte(s) */
 	len = *p & 0x7f;
 	if (*p++ & 0x80) {
```
dengert: The original looks like it would function correctly. The change  just makes the code look more complicated, and more pront to error if additional modifications are added in the future. The removed code in 93, would always check for left == 0
no mater what code was changed above.

src/libopensc/card-piv.c
frankmorgner: I don't expect to shrink `SC_MAX_SERIALNR` any time soon.

Yes indeed, coverity catches overwrites of fixed size buffers such as the serial number.

And, for what it's worth, I think that the code looks cleaner without a ternary if.

src/libopensc/asn1.c
frankmorgner: Both versions are semantically identical, see coverity scan CID 13628. It's hard to argue about the beauty of the code. I, personally, think that superfluous operations should be omitted.

dengert: On 2/10/2015 3:21 AM, Frank Morgner wrote:
> In src/libopensc/asn1.c <https://github.com/OpenSC/OpenSC/pull/375#discussion_r24398396>:
>
>>  	}
>> -	if (left == 0)
>> -		return SC_ERROR_INVALID_ASN1_OBJECT;
>>  	/* parse length byte(s) */
>>  	len = *p & 0x7f;
>>  	if (*p++ & 0x80) {
>
> Both versions are semantically identical, see coverity scan CID 13628. It's hard to argue about the beauty of the code. I, personally, think that superfluous operations should be omitted.
>

My personal option, some minor performance can be sacrificed for code readability, especially in code that
gets enhanced and modified by a committee of volunteers (current and future) none of them who have the hardware
to test all the code.

> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/375/files#r24398396>.
>

-- 

viktorTarasov: @frankmorgner 
"if (left == 0)" in your updated version will not work,
because of  " if (left < 2) return ...tt" few lines upper .

Better to remove it completely:
````
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -87,14 +87,13 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
                if (left == 0 || n == 0)
                        /* either an invalid tag or it doesn't fit in
                         * unsigned int */
 
        }
-       if (left == 0)
-               return SC_ERROR_INVALID_ASN1_OBJECT;
+
        /* parse length byte(s) */
        len = *p & 0x7f;
        if (*p++ & 0x80) {
                unsigned int a = 0;
                if (len > 4 || len > left)
````

viktorTarasov:
partially applied in da07e22c97effe27d8decf4037f314e4015da1eb .. 9f3dbaa39dbb2c6c4515bb2ea0ec78023d10bdb4

PR #380, opened 2015-02-15
Introduce EC curve OID into internal algorithm info data

viktorTarasov:
It's needed to store information about EC curve supported by card.
Primary usage is, when importing/generating key,  to get know if particular curve is supported by card.


frankmorgner:
See comments on #381

viktorTarasov:
applied in fa923831f8e8073f0b57ea492c9e2e93f661a373

PR #383, opened 2015-02-17
When deleting a private key object, overwrite its contents.

martelletto:
When a private key is deleted from a CardOS 4 smartcard, the PKCS15 file system object is removed, but the key remains configured on the card. The key can still be used just as before, provided the user bypasses the PKCS15 layer and references the key directly. This is the case of PKCS11 providers which have been opened before the key was removed.

The change implemented in this pull request triggers an overwrite of the private key object upon the removal of its corresponding PKCS15 object, rendering the deleted key unusable.

### DIFF
--- a/src/pkcs15init/pkcs15-cardos.c
+++ b/src/pkcs15init/pkcs15-cardos.c
@@ -403,6 +403,74 @@ out:
 }
 
 /*
+ * Object deletion.
+ */
+static int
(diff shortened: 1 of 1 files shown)


viktorTarasov:
Applied in ced77ae6c7360914e746178b369561ad7fac84dc,
thank you

PR #385, opened 2015-02-21
Allow PKCS15 cache to work with PIV cards (if enabled)

darconeous:
These changes simply allow the use of the file cache if it is enabled. More specifically, it allows the file cache to actually be created properly in the first place. Once the cache files are created properly then they are used.

If `use_file_cache` is not enabled, then the behavior continues as normal, fetching all certificates every time. The big win is that if it can fetch a certificate from a file instead of from the card, it will do so.


dengert:
committed

PR #389, opened 2015-03-08
isoApplet: prototype of asn1_sig_value_sequence_to_rs() has changed

viktorTarasov:
Sorry, I do not seen warnings in PR build.
Here is proposed fix.


philipWendland:
Thank you.
Line 1108 looks wrong. Maybe the EF(PrKDF) needs to be read to get the field_length. I will look at it tonight.

viktorTarasov:
No, for EC key the alg-ref IS the field length.
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-sec.c#L364

philipWendland:
Oh, I see. Thank you, I would not have seen this.

If you would review the following - I think the length check for len/outlen was the wrong way around:
```
[philip@philip-pc OpenSC]$ git diff
diff --git a/src/libopensc/card-isoApplet.c b/src/libopensc/card-isoApplet.c
index ac6d6ed..d3364cb 100644
--- a/src/libopensc/card-isoApplet.c
+++ b/src/libopensc/card-isoApplet.c
@@ -1202,17 +1202,14 @@ isoApplet_compute_signature(struct sc_card *card,
                LOG_FUNC_RETURN(ctx, r);
        }
 
-       /* If we used ECDSA for the signature op, OpenSC thinks it has to
-        * convert it to {sequence{integer,integer}} which is already done by the
-        * card actually.
-        * To fix this, I strip the {sequence, sequence} structual information,
-        * which will be added again later.*/
+       /* If ECDSA was used, the ASN.1 sequence of integers R,S returned by the
+        * card needs to be converted to the raw concatenation of R,S for PKCS#11. */
        if(drvdata->sec_env_alg_ref == ISOAPPLET_ALG_REF_ECDSA)
        {
                u8* p = NULL;
                size_t len = drvdata->sec_env_ec_field_length / 4;
 
-               if (len < outlen)
+               if (len > outlen)
                        LOG_FUNC_RETURN(ctx, SC_ERROR_BUFFER_TOO_SMALL);
 
                p = calloc(1,len);

```
With this change it works for me.

viktorTarasov:
Applied with patch in 877fa862d99ee6177973ac2d80178449836cb5c3 .

PR #390, opened 2015-03-10
Fixed issue with building dnie-tool bash_completion script

lbakman:
Fixed an issue where the dnie-tool was not generated correctly form bash_completion.

The XML formatting caused the Makefile to not generate the proper arguments for /etc/bash_completion.d/dnie-tool


frankmorgner:
Thanks, it looks very good! Could you squash the two commits into one?

lbakman:
The two commits should now be squashed.

frankmorgner:
thanks

PR #399, opened 2015-03-22
Revert broken a4c8d671108bca5ce93472f4e3555d59682aff55

CardContact:
The change introduced by the above commit breaks the generation of DKEK shares with a password not split in an n-of-m share.

This commit reverts the change, but adds the parameter checking that was intended by the original committer. It also removes compiler warning related to unchecked returns.

### DIFF
--- a/src/tools/sc-hsm-tool.c
+++ b/src/tools/sc-hsm-tool.c
@@ -420,7 +420,11 @@ static int cleanUpShares(secret_share_t *shares, unsigned char n)
 
 void clearScreen()
 {
-	if (system( "clear" )) system( "cls" );
+	if (system( "clear" )) {
+		if (system( "cls" )) {
+			fprintf(stderr, "Clearing the screen failed\n");
+		}
+	}
 }
 
 
@@ -591,6 +595,11 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
 	secret_share_t *shares = NULL;
 	secret_share_t *sp;
 
+	if (num_of_password_shares < 2) {
+		fprintf(stderr, "--pwd-shares-total must 2 or larger\n");
+		return -1;
+	}
+
 	/*
 	 * Initialize prime and secret
 	 */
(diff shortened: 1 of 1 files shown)


src/tools/sc-hsm-tool.c
```diff
@@ -803,10 +821,9 @@ static void ask_for_password(char **pwd, int *pwdlen)
 
 
 
-static int generate_pwd_shares(sc_card_t *card, char **pwd, int *pwdlen, int password_shares_threshold, unsigned int password_shares_total)
+static int generate_pwd_shares(sc_card_t *card, char **pwd, int *pwdlen, int password_shares_threshold, int password_shares_total)
```
frankmorgner: If `password_shares_threshold` and `password_shares_total` must both be bigger than 2 (as tested below) then why do you bother with negative values here at all? Why don't you catch `-1` (and any other negative value) in `create_dkek_share`, which leaves `password_shares_threshold` with accepting `unsigned int`s only.

CardContact: Because -1 is the default value of any option and I have the habit to do semantical checks close to were the actual processing code is.
Sure, one could define 0 as the default value meaning "Option not defined". I could as well say: "Hey 34563621 is the default value". Why is it important to make it an unsigned int rather than an int ? The range we expect is for these parameter is somewhere between 2 and 10, no real range issue here, nothing to save by converting to unsigned and no benefit in parameter checking - other than having to rewrite working code.

frankmorgner: That's OK. I'm fine with `-1` as default value. I am talking about the following different layers:
1. `create_dkek_share` which creates shares from a given user input (e.g. negative values)
2. `generate_pwd_shares` which creates shares from well defined `password_shares_threshold` and `password_shares_total` (e.g. at least positive/unsigned values)

There is no need for a rewrite when keeping these logical layers alive. But, what ever, it's way too unimportant to spend so much time on this discussions ;-)

viktorTarasov:
applied in b1bdfae20094de41c5d8435dc4d64b8896e04773,
thank you.

PR #398, opened 2015-03-22
harmonize supported ECC curves with pkcs11-tool

philipWendland:
pkcs15-pubkey.c holds a struct containing supported ECC curves. The contents of
this struct are being harmonized with pkcs11-tool supported curves.

PR #402, opened 2015-03-25
Isoapplet enhancements

philipWendland:
Hi,
this adds new EC curves and GET CHALLENGE to the IsoApplet driver and is backward compatible to 00.05 (The current version of the driver im OpenSC).
It would be nice to have this in 0.15.0. Thank you.


src/libopensc/card-isoApplet.c
```diff
@@ -67,6 +71,23 @@ static struct sc_card_driver isoApplet_drv =
 	NULL, 0, NULL
 };
 
+static struct isoapplet_supported_ec_curves {
+		struct sc_object_id oid;
+		size_t size;
+		unsigned int min_applet_version;
+} ec_curves[] = {
+	{{{1, 2, 840, 10045, 3, 1, 1, -1}},     192, 0x0000}, /* secp192r1, nistp192, prime192v1, ansiX9p192r1 */
```
frankmorgner: These OIDs are already present somewhere in OpenSC. Is it possible to summarize them in a header file?

philipWendland: The OID is a bijective identifier for a curve - the name is not. Generally, I find it (arguably) preferable to use the OID instead of the name to reference a curve.

`sc_pkcs15_fix_ec_parameters()`in pkcs15-publey.c can be used for a mapping of names to OIDs and field length. But there is information that is duplicated in pkcs11-tool, see #398. 

If we are talking about refactoring this, maybe this can be extended to provide some form of explicit parameters for pkcs15-init drivers?

frankmorgner: Drivers and frameworks should use a bigger common code base. Currently, OIDs/names/parameters can already be found in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-sc-hsm.c#L42, https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-pubkey.c#L1381, https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L65 and you are now introducing yet an other code duplication in `card-isoApplet.c`. Yes, I am suggesting to merge this code.

Also, supporting named curves and explicit parameters at the same time is not too hard...

dengert: The only duplication is the order of the curve or key_length  in bits, (refered to as size in the structure) The array and version number are there to say which curve the applet supports, so this file will still need a table of OIDs with flags.

philipWendland: @frankmorgner 
```
you are now introducing yet an other code duplication in card-isoApplet.c and pkcs15-isoApplet.c
```
I don't see where this PR introduces "yet another code duplication". Could you point me to it?
(What I infact see, is the removal of c&p code through the introduction of the table...)
This table is used to register EC curves supported by the applet. This, unfortunately, depends on the version of the applet. Thus, I need a table that maps OIDs to the version of the applet. How can this be realised with common code?
I included the length as t's not like a NIST curve would suddenly change it. Allocating seperate structs and calling sc_pkcs15_fix_ec_parameters() on each one to obtain the length would IMHO be just unnecessary and opaque bloat.

There is no common source for explicit parameters in OpenSC and that is why pkcs15-init drivers need to define them themselves. But which format is most suitable for all future cards?
For example, how are leading zero bytes handled..like in secp256k1 coefficient A - 32 bytes zero or one byte zero? The IsoApplet and Java Card API needs 32 bytes zero.

```
Also, supporting named curves and explicit parameters at the same time is not too hard...
```
I don't get your point here, sorry.

viktorTarasov:
applied in 090aed2..c8d206e,
thank you.

Issue #405, opened 2015-03-26
"New" belpic card not working anymore


bigon:
Hello,

A few weeks back I got a new beid card and I'm not not able to login on websites using firefox or connecting to a VPN using openconnect.

The governmental website seems to say that you need a new version of their middleware for these new cards: http://eid.belgium.be/en/

frankmorgner:
closed by #406

PR #406, opened 2015-03-26
Support 2K cards

yoe:
Recent BELPIC cards (issued since March 2014) have a validity of 10
years (rather than 5 as before), and therefore also increased the key
size from 1024 bits to 2048 bits.

Key size can be detected by checking the applet version, for which we
have to issue a "GetCardData" command. If the applet is version 0x17 or
higher, keys are 2048 bits.

### DIFF
--- a/src/libopensc/card-belpic.c
+++ b/src/libopensc/card-belpic.c
@@ -577,6 +577,41 @@ static int str2lang(sc_context_t *ctx, char *lang)
 	return -1;
 }
 
+static int get_applet_version(sc_card_t *card)
+{
+	sc_apdu_t apdu;
+	u8 carddata[28];
+	u8 carddata_cmd[] = { 0x80, 0xE4, 0x00, 0x00, 0x1C };
+	int r;
+
+	r = sc_bytes2apdu(card->ctx, carddata_cmd, sizeof(carddata_cmd), &apdu);
+	if(r) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "bytes to APDU conversion failed: %d\n", r);
+		return r;
+	}
+
+	apdu.resp = carddata;
+	apdu.resplen = sizeof(carddata);
+
+	r = sc_transmit_apdu(card, &apdu);
+	if(r) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData command failed: %d\n", r);
+		return r;
+	}
+
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if(r) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d\n", r);
+		return r;
+	}
+	if(apdu.resplen < sizeof(carddata)) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d bytes rather than expected %d\n", apdu.resplen, sizeof(carddata));
+		return SC_ERROR_WRONG_LENGTH;
+	}
+
+	return (int)(carddata[21]);
+}
+
 #ifdef GET_LANG_FROM_CARD
 
 /* str is in lower case, the case of buf can be both, and buf is large enough */
@@ -843,6 +878,8 @@ static int belpic_init(sc_card_t *card)
 {
 	struct belpic_priv_data *priv = NULL;
 	scconf_block *conf_block;
+	int applet_version;
+	int key_size = 1024;
 #ifdef BELPIC_PIN_PAD
 	int r;
 #endif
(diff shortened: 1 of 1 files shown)


frankmorgner:
does this fix #405?

src/libopensc/card-belpic.c
```diff
@@ -577,6 +577,41 @@ static int str2lang(sc_context_t *ctx, char *lang)
...
+	if(r) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d\n", r);
+		return r;
+	}
+	if(apdu.resplen < sizeof(carddata)) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d bytes rather than expected %d\n", apdu.resplen, sizeof(carddata));
+		return SC_ERROR_WRONG_LENGTH;
+	}
+
+	return (int)(carddata[21]);
```
frankmorgner: I'm not familiar with your card, but could it be that the version number is TLV encoded? Then parsing that data would actually be more robust and future proof...

yoe:
It should fix #405, yes.

src/libopensc/card-belpic.c
yoe: No, it's not TLV-encoded.

See <https://github.com/Fedict/eid-mw/blob/master/doc/sdk/documentation/Public_Belpic_Applet_v1%207_Ref_Manual%20-%20A01.pdf>, §6.9, for the full documentation of what the "GET CARD DATA" command returns.

bigon:
I can confirm that this fix my bug #405

frankmorgner:
You may want to also set `card->serialnr` which also comes with the GET CARD DATA Response.

src/libopensc/card-belpic.c
frankmorgner: I see. Would you mind using a define instead of a magic value to access this byte?

yoe:
On Fri, Mar 27, 2015 at 10:25:24AM -0700, Frank Morgner wrote:
> You may want to also set card->serialnr which also comes with the GET CARD DATA
> Response.

I had thought of that, but this was a quick hack during lunch break,
which didn't allow me too much time, and I didn't (immediately) find
where the serial number was stored. Thanks for pointing me to it ;-)

-- 
It is easy to love a country that is famous for chocolate and beer

  -- Barack Obama, speaking in Brussels, Belgium, 2014-03-26

src/libopensc/card-belpic.c
yoe: On Fri, Mar 27, 2015 at 10:26:05AM -0700, Frank Morgner wrote:
> In src/libopensc/card-belpic.c:
> 
> > +     if(r) {
> > +             sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData command failed: %d\n", r);
> > +             return r;
> > +     }
> > +
> > +     r = sc_check_sw(card, apdu.sw1, apdu.sw2);
> > +     if(r) {
> > +             sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d\n", r);
> > +             return r;
> > +     }
> > +     if(apdu.resplen < sizeof(carddata)) {
> > +             sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "GetCardData: card returned %d bytes rather than expected %d\n", apdu.resplen, sizeof(carddata));
> > +             return SC_ERROR_WRONG_LENGTH;
> > +     }
> > +
> > +     return (int)(carddata[21]);
> 
> I see. Would you mind using a define instead of a magic value to access this
> byte?

Sure. I'll change the code to that effect.

-- 
It is easy to love a country that is famous for chocolate and beer

  -- Barack Obama, speaking in Brussels, Belgium, 2014-03-26

yoe:
Patch pushed which refactors stuff so that the one GET CARD DATA command can be reused to also store the serial number.

However, I'm not sure if I understand the sc_card_t structure correctly; when I write something to card->serialnr.value and set card->serialnr.len correctly, `opensc-tool --serial` doesn't produce a serial number.

Help?

philipWendland:
Hi yoe,
AFAIK the card->serialnr is only used by card drivers for internal caching. 
If you want to see the serial number in opensc-tool --serial, I think you need to implement the SC_CARDCTL_GET_SERIALNR cardctl (just *grep* for this).
At the pkcs15 level there is also (sc_pkcs15_card) p15card->tokeninfo->serial_number (they might differ, a user may chose the pkcs15 serial: `pkcs15-init --create-pkcs15 --serial`). You might want to have a look at this too.

yoe:
Hrm, that seems way more complicated than just setting a variable. I suppose I could look at it, but please note that this isn't as critical a fix as getting basic RSA to work with the card (which the other bit of this patch is).

Thanks,

viktorTarasov:
applied in 5149dd3e62594eb2477f699d834584991ab54d5f,
thank you.

PR #408, opened 2015-03-27
Initial support for Morpho IAS Agent Card

tc-anssi:
* This commit adds initial support for Morpho French Agent card which is an
    IAS card. Signature operations are working. Since my test card was
    read-only, I was unable to test object management functions.


src/libopensc/card-iasecc.c
```diff
@@ -566,6 +572,53 @@ iasecc_init_amos(struct sc_card *card)
...
+	flags = IASECC_CARD_DEFAULT_FLAGS;
+
+	_sc_card_add_rsa_alg(card, 1024, flags, 0x10001);
+	_sc_card_add_rsa_alg(card, 2048, flags, 0x10001);
+
+	card->caps = SC_CARD_CAP_RNG;
+	card->caps |= SC_CARD_CAP_APDU_EXT;
+	card->caps |= SC_CARD_CAP_USE_FCI_AC;
+
+    resp_len = sizeof(resp);
```
frankmorgner: Could you unify the indenting according to the file's standard?

viktorTarasov:
applied in 435291f2166935e741e67539a55f191695874268,
thank you.

PR #410, opened 2015-03-28
added parameter checking for sc_pkcs11_get_mechanism_list

### LINKED ISSUE #232: Memory corruption with firefox
I have found that the opensc-pkcs11 plugin induces memory corruption in firefox 28 on a Feora 20 system.
This happens both with the distribution package AND with a recompiled version of the latest git checkout.

Authentication with the smartcard is succesful and the browser prompts correctly for the certificate therein stored; however, as soon as the certificate is shown it seems that the memory gets corrupted and I get either

 a) a segmentation fault on part of the browser
 b) first the error message 
  Received incorrect handshakes hash values from peer. (Error code: ssl_error_bad_handshake_hash_value)
 followed by
 security library: memory allocation failure. (Error code: sec_error_no_memory)

The reference for the bug on fedora bugzilla is
https://bugzilla.redhat.com/show_bug.cgi?id=1089476

### LINKED ISSUE #409: Memory corruption with firefox (reopen the bug #232)
Hi,

I was able to reproduce the problem outlined in the bug #232
I'm running Firefox 36.0.4 on Windows 7 x64 with OpenSC 0.14.0 with a card for which I'm beginning to write a driver.
Insert the card triggers a crash few seconds after the insertion.
Tested with the binary version available on the website.

After an analysis, the crash occured when calling sc_pkcs11_get_mechanism_list from C_GetMechanismList as outline in the following capture
bug

Because slot->card is null, the application crashes in the line:
for (n = 0; n < p11card->nmechanisms; n++) { of the function sc_pkcs11_get_mechanism_list
sans titre

I saw your comment about a probable defect in the NSS interface, but it should be possible to introduce a defense programming check and return an error if there is no card.

for example:
if (slot->card == NULL)
    return CKR_TOKEN_NOT_PRESENT;

The bug is still present in the source code

regards,
Vincent LE TOUX
--
http://www.mysmartlogon.com

frankmorgner:
fixes #409

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -83,6 +83,9 @@ sc_pkcs11_get_mechanism_list(struct sc_pkcs11_card *p11card,
 	unsigned int n, count = 0;
 	int rv;
 
+	if (!p11card)
+		return CKR_TOKEN_NOT_PRESENT;
+
 	for (n = 0; n < p11card->nmechanisms; n++) {
 		if (!(mt = p11card->mechanisms[n]))
 			continue;


vletoux:
It does fix it.
regards,
Vincent

frankmorgner:
great! thanks for reporting.

I'll wait for additional feedback from an other maintainer.

viktorTarasov:
applied in c019a6230900e0703c2b701332f9cb34166adc0e,
thank you.

Issue #404, opened 2015-03-25
Testing for new release 0.15.0 required


frankmorgner:
In preparation of a new release (currently in the branch `opensc-0.15.0`) the @OpenSC/opensc-maintainers are currently testing existing cards and applications. We plan to add a matrix of what is supported (and what is untested).

Could you please add your test cases and cards here if possible?

@viktorTarasov @frankmorgner @marschap @martinpaljak @LudovicRousseau @hongquan @philipWendland @dengert @szikora @mbrossard @tc-anssi @metsma @poupas @resoli @ttaylor249 @germanblanco @mjrider @tkil @hean01 @Flameeyes @hiviah @cbancroft @sjoblomt @frankthater @dirkx @Ecordonnier-theobroma @puccia @hhonkanen @eighthave @henryk

LudovicRousseau:
I do not use OpenSC these days.
I have no test cases to provide. Sorry.

dirkx:
> In preparation of a new release (currently in the branch opensc-0.15.0) we are currently testing existing cards and applications. We plan to add a matrix of what is supported (and what is untested).
> Could you please add your test cases and cards here if possible.

Those on Yosemite - you may need the not yet incorperated

		https://github.com/dirkx/OpenSC/commit/1cd8d02449b2c1aeb0383b9c583aa59265ab1460

in order to build it.

Dw.

dirkx:
> In preparation of a new release (currently in the branch opensc-0.15.0) we are currently testing existing cards and applications. We plan to add a matrix of what is supported (and what is untested).
> 
For Yosemite - had to massage in:

	https://github.com/dirkx/OpenSC/commit/1cd8d02449b2c1aeb0383b9c583aa59265ab1460

to make the OSX build work again. Not sure if that is specific to my setup ? I also noted that the version number in the build packages is still 0.14:

	config.log:PACKAGE_VERSION='0.14.0'

Besides that - works for the common test cases here (Yosemite, FreeBSD); except the ones that require the still pending 

	https://github.com/dirkx/OpenSC/commit/59d06ae43b1f8f57e9ca3cb7ad63ab0819982e7a

on readers with and without PIN pads (SPR532, HID, etc) against a range entersafe (feitian), aventra, aet, BE/PT/IT govt. cards, medical NL cards.

I do notice that opensc-tool/-explore now produce more sensible (albeit gargbled) output on PKCS15 cards or not yet initialized cards. E.g. stuff like

[...]

Issue #353, opened 2015-01-16
IAS-ECC: Cannot segregate key usage


tc-anssi:
I had previously reported (opensc-devel) that IAS-ECC cards with the current implementation did not allow to segregate key usage. I have located that the issue lies in `src/pkcs15init/pkcs15-iasecc.c` in the `iasecc_pkcs15_fix_private_key_attributes` function.
From what I understood, the code fetch the card ACLs and maps that to key usage regardless of the key usage that was specified. This leads to an overwrite of the key usage the user required (or an union with them).
For instance, if I require a private key to be restricted to signature operations, I end up with all key usage bits enabled which is wrong.

The quick workaround below comments the `key_usage` overwrite, with which the card is still fully functional and keeps the key usage that was required by the application.

Viktor, I did not find something in the IAS-ECC specification that mandates to do such mapping, what is your opinion ?
```
--- a/src/pkcs15init/pkcs15-iasecc.c
+++ b/src/pkcs15init/pkcs15-iasecc.c
@@ -937,9 +937,9 @@ iasecc_pkcs15_fix_private_key_attributes(struct sc_profile *profile, struct sc_p
                                        IASECC_ALGORITHM_RSA_PKCS | IASECC_ALGORITHM_SHA2);
                        LOG_TEST_RET(ctx, rv, "Cannot add RSA_PKCS SHA2 supported mechanism");
 
-                       key_info->usage |= SC_PKCS15_PRKEY_USAGE_SIGN;
+                       //key_info->usage |= SC_PKCS15_PRKEY_USAGE_SIGN;
                        if (sdo_prvkey->docp.non_repudiation.value && sdo_prvkey->docp.non_repudiation.value[0])   {
-                               key_info->usage |= SC_PKCS15_PRKEY_USAGE_NONREPUDIATION;
+                               //key_info->usage |= SC_PKCS15_PRKEY_USAGE_NONREPUDIATION;
                                object->user_consent = 1;
                        }
                }
@@ -947,14 +947,14 @@ iasecc_pkcs15_fix_private_key_attributes(struct sc_profile *profile, struct sc_p
                        rv = iasecc_pkcs15_add_algorithm_reference(p15card, key_info, IASECC_ALGORITHM_RSA_PKCS);
                        LOG_TEST_RET(ctx, rv, "Cannot add RSA_PKCS supported mechanism");
 
[...]

PR #417, opened 2015-03-31
iasecc: Fix key usage when provisioning card

tc-anssi:
* Avoids overriding key_usage when creating objects on the card.

Solves #353, discussed in #404.


viktorTarasov:
You have a reason,
in PKCS#15 the key usages are *intended usages* .

Applied in 117f3a74beb715f7d0fd43bd11a378ecf9db928e, thank you.

PR #418, opened 2015-03-31
iasecc: Fix log output is always displayed

tc-anssi:
* iasecc_read_public_key function uses SC_SUCCESS instead of log level
  value, hence the log output is always displayed. This uses
  SC_LOG_DEBUG_NORMAL instead.


viktorTarasov:
Applied short form of LOG macro (4a4d750e7383da4a4b71f55465efe3749df63735),
thank you.

PR #419, opened 2015-03-31
RSA: PKCS1 for signature, RAW for decryption

frankmorgner:
Hi!

I am using a card that supports RSA PKCS#1 for signature only and RAW input for decryption only. currently the card's capabilities flag is used for both operations which is also what is exorted from the PKCS#11 module. This PR changes this behavior, but I am not sure if this breaks some other card or application.

From the comments in the PKCS#11 module I can see that there are many deficits. What would be the correct(tm) way of implementing method (i.e. signature/decryption) dependant card capabilities? Is PKCS#1 even used for decryption (or RAW input for signature) in some application?

Thanks for comments.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -4701,11 +4701,12 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
 
 	/* Check if we support raw RSA */
 	if (flags & SC_ALGORITHM_RSA_RAW) {
+		mech_info.flags &= ~CKF_SIGN;
 		mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_X_509, &mech_info, CKK_RSA, NULL);
 		rc = sc_pkcs11_register_mechanism(p11card, mt);
 		if (rc != CKR_OK)
 			return rc;
-
+		mech_info.flags |= CKF_SIGN;
 	}
 
 	/* We support PKCS1 padding in software */
@@ -4729,6 +4730,7 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
 
 	/* No need to Check for PKCS1  We support it in software and turned it on above so always added it */
 	if (flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
+		mech_info.flags &= ~CKF_DECRYPT;
 		mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS, &mech_info, CKK_RSA, NULL);
 		rc = sc_pkcs11_register_mechanism(p11card, mt);
 		if (rc != CKR_OK)
@@ -4770,6 +4772,7 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
 				return rc;
 		}
 #endif /* ENABLE_OPENSSL */
+		mech_info.flags |= CKF_DECRYPT;
 	}
 
 	/* TODO support other padding mechanisms */


philipWendland:
Hi,
I think padding is really needed for most decryption use cases. In many cases, asymmetric cryptography is only used to encrypt a symmetric key, which is then used for the actual encryption/decryption. Now imagine, for RSA 2048 bit, you have a blocksize of 256 bytes. If you want to encrypt, say, a 256 bit AES key, it is only 32 bytes. The rest must be filled up with padding bytes.

For signatures I don't really see a use case for RAW RSA. Padding will need to be done on the host side to fill up the space that is not used by the hash.

In line 4704 you seem to disable signing if the card supports RAW RSA, but Padding can be done on the host-side and then be sent to the card as RAW block. Disabling the padding-decryption case (4733) is IMHO wrong for the above reasons.

Maybe the mechanism of registering card capabilities does not always cover all requirements. We might need a way to specifiy:
* for each key type (RSA/ECC/..)
* what precalculated hashes the card accepts, or at least whether it does enforce that the hash is done on-card
* what hashes the card can perform by itself
* in what operations (sign/decrypt)

dengert:
On 3/31/2015 1:39 PM, Frank Morgner wrote:
> Hi!
>
> I am using a card that supports RSA PKCS#1 for signature only and RAW input for decryption only.  currently the card's capabilities flag is used for both operations which is also what is exorted from
> the PKCS#11 module. This PR changes this behavior, but I am not sure if this breaks some other card or application.

It sure looks it will break other cards!

The PIV card for example only does RAW RSA on the card. RAW RSA makes no distinction between
a sign and decrypt operation because it it the same operation!

If OpenSC is compiled with OpenSSL, the OpenSC code can do the padding in software and the results
are the same size as the RSA key. This then processed by an RAW RSA operation.

Decryption takes a key size block of data and runs it against the RAW RSA.
The output is the key size. For a RSA-PKCS1 the returned data is checked to see if it
was padded. Does you card support doing the unpadding? Sounds like it does not,
so software needs to do the unpadding.

In both cases the same RAW RSA operation is preformed by the card.

The OpenSC code will try and let the card do the padding and unpadding if the card can do it.
If not OpenSC can do the padding and unpadding in software.

If you card needs special requirements, then ann code changes for the card should be in the card driver.

>
>  From the comments in the PKCS#11 module I can see that there are many deficits. What would be the correct(tm) way of implementing method (i.e. signature/decryption) dependant card capabilities? Is
> PKCS#1 even used for decryption (or RAW input for signature) in some application?



>
> Thanks for comments.
>
> --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/OpenSC/OpenSC/pull/419
>
>
>         Commit Summary
>
>   * RSA: PKCS1 for signature, RAW for decryption
>
>
>         File Changes
>
>   * *M* src/pkcs11/framework-pkcs15.c <https://github.com/OpenSC/OpenSC/pull/419/files#diff-0> (5)
>
>
>         Patch Links:
>
>   * https://github.com/OpenSC/OpenSC/pull/419.patch
>   * https://github.com/OpenSC/OpenSC/pull/419.diff
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419>.
>

-- 

dengert:
On 3/31/2015 2:21 PM, Philip Wendland wrote:
> Hi,
> I think padding is really needed for most decryption use cases. In many cases, asymmetric cryptography is only used to encrypt a symmetric key, which is then used for the actual encryption/decryption.
> Now imagine, for RSA 2048 bit, you have a blocksize of 256 bytes. If you want to encrypt, say, a 256 bit AES key, it is only 32 bytes. The rest must be filled up with padding bytes.
>
> For signatures I don't really see a use case for RAW RSA. Padding will need to be done on the host side to fill up the space that is not used by the hash.
>
> In line 4704 you seem to disable signing if the card supports RAW RSA, but Padding can be done on the host-side and then be sent to the card as RAW block. Disabling the padding-decryption case (4733)
> is IMHO wrong for the above reasons.
>
> Maybe the mechanism of registering card capabilities does not always cover all requirements. We might need a way to specifiy:
>
>   * for each key type (RSA/ECC/..)
>   * what precalculated hashes the card accepts, or at least whether it does enforce that the hash is done on-card
>   * what hashes the card can perform by itself
>   * in what operations (sign/decrypt)

I believe all of those have been done.
(Have to run, will  answer this later or tomorrow)


>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88217191>.
>

-- 

frankmorgner:
I believe that conversion doesn't work for me: I could disable PKCS1 padding by calling `_sc_card_add_rsa_alg` without `SC_ALGORITHM_RSA_PAD_PKCS1` which only leaves `SC_ALGORITHM_RSA_RAW` enabled. But I believe that the output of the signature (which now needs to be PKCS#1 padded in software) will not be compatible with the typical raw RSA output.

dengert:
On 3/31/2015 5:43 PM, Frank Morgner wrote:
> I believe that conversion doesn't work for me: I could disable PKCS1 padding by calling |_sc_card_add_rsa_alg| without |SC_ALGORITHM_RSA_PAD_PKCS1| which only leaves |SC_ALGORITHM_RSA_RAW| enabled.
> But I believe that the output of the signature (which now needs to be PKCS#1 padded in software) will not be compatible with the typical raw RSA output.

Do a grep for SC_ALGORITHM_RSA_RAW
It looks like your mod would break 19 cards.

The caller provides the hash of the data to be signed. The hash is then padded to the key size and the result is then sent to the card to be run through
the RAW RSA. The results are the signature.

sig = RSA_RAW(pad_PKCS1(hash(data)))

 From what I understand you were saying, the card could do one of

card_sig = card_RAW_RSA(card_pad_PKCS1(software_hash(data))

SC_ALGORITHM_RSA_PAD_PKCS1 the hash is sent to the card, and the card does the rest, returning the signature.

Or
card_sig = card_RAW_RSA(software_pad_PKCS1(software_hash(data))
SC_ALGORITHM_RSA_RAW where the padding is done in software.
This is how the PIV does it, as it only has SC_ALGORITHM_RSA_RAW.

The software_padding could be done either by the caller of PKCS#11 doing the padding of the hash then using the RSA-X-509
mechanisum.

Or by the PKCS#11 caller passing the hash and using RSA-PKCS.
In this case the OpenSC can do either:

ard_sig = card_RAW_RSA(card_pad_PKCS1(software_hash(data)) If the card supports it,
or
card_sig = card_RAW_RSA(sc_pkcs1_add_01_padding(software_hash(data))  OpenSC's padding routine.

The _sc_card_add_rsa_alg tells OpenSC what the card can do,
The framwork-pkcs15.c sets what the card plus software can do,


See framework-pkcs15.c:

4711         /* We support PKCS1 padding in software */
4712         /* either the card supports it or OpenSC does */
4713         flags |= SC_ALGORITHM_RSA_PAD_PKCS1;
4714

4730         /* No need to Check for PKCS1  We support it in software and turned it on above so always added it */
4731         if (flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
4732                 mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS, &mech_info, CKK_RSA, NULL);
4733                 rc = sc_pkcs11_register_mechanism(p11card, mt);
4734                 if (rc != CKR_OK)
4735                         return rc;
4736

So even if you turn off SC_ALGORITHM_RSA_PAD_PKCS, the CKM_RSA_PKCS is set to say the software or card can do it.

sc_pkcs15_compute_signature in pkcs15-sec.c does most of the work.

Do you have a OpenSC debug trace without your mod that show what problem you are seeing?

What card are you using?

An output of pkcs11-tool -M
would also be helpful.


The PIV card issues only:

2876          flags = SC_ALGORITHM_RSA_RAW;
2877
2878         _sc_card_add_rsa_alg(card, 1024, flags, 0); /* manditory */
2879         _sc_card_add_rsa_alg(card, 2048, flags, 0); /* optional */
2880         _sc_card_add_rsa_alg(card, 3072, flags, 0); /* optional */
2881
2882         flags = SC_ALGORITHM_ECDSA_RAW;
2883         ext_flags = SC_ALGORITHM_EXT_EC_NAMEDCURVE | SC_ALGORITHM_EXT_EC_UNCOMPRESES;
2884
2885         _sc_card_add_ec_alg(card, 256, flags, ext_flags, NULL);
2886         _sc_card_add_ec_alg(card, 384, flags, ext_flags, NULL);
2887
2888         card->caps |= SC_CARD_CAP_RNG;
2889



But pkcs11-tool -M shows what the card and OpenSC (and OpenSSL if OpenSC compiled with OpenSSL) the card can do:
Supported mechanisms:
   SHA-1, digest
   SHA256, digest
   SHA384, digest
   SHA512, digest
   MD5, digest
   RIPEMD160, digest
   GOSTR3411, digest
   ECDSA, keySize={256,384}, hw, sign, other flags=0x1800000
   ECDSA-SHA1, keySize={256,384}, hw, sign, other flags=0x1800000
   ECDH1-COFACTOR-DERIVE, keySize={256,384}, hw, derive, other flags=0x1800000
   ECDH1-DERIVE, keySize={256,384}, hw, derive, other flags=0x1800000
   RSA-X-509, keySize={1024,3072}, hw, decrypt, sign, verify
   RSA-PKCS, keySize={1024,3072}, hw, decrypt, sign, verify
   SHA1-RSA-PKCS, keySize={1024,3072}, sign, verify
   SHA256-RSA-PKCS, keySize={1024,3072}, sign, verify
   SHA384-RSA-PKCS, keySize={1024,3072}, sign, verify
   SHA512-RSA-PKCS, keySize={1024,3072}, sign, verify
   MD5-RSA-PKCS, keySize={1024,3072}, sign, verify
   RIPEMD160-RSA-PKCS, keySize={1024,3072}, sign, verify

>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88275385>.
>

-- 

dengert:
On 3/31/2015 2:21 PM, Philip Wendland wrote:
> Hi,
> I think padding is really needed for most decryption use cases. In many cases, asymmetric cryptography is only used to encrypt a symmetric key, which is then used for the actual encryption/decryption.
> Now imagine, for RSA 2048 bit, you have a blocksize of 256 bytes. If you want to encrypt, say, a 256 bit AES key, it is only 32 bytes. The rest must be filled up with padding bytes.
>

I would word it, padding needs to be removed, if the encryption used padding.

> For signatures I don't really see a use case for RAW RSA. Padding will need to be done on the host side to fill up the space that is not used by the hash.

Not sure what you mean by this. The card will be doing a RAW RSA operation internally in any case.  But the card might not expose the RAW operation, rather it
will always pad what ever data it receives then do the RAW RSA. In the case the card is saying it only supports SC_ALGORITHM_RSA_PAD_PKCS1

There is more then one type of padding, but if a card only supports SC_ALGORITHM_RSA_PAD_PKCS1 you are stuck with
only the PKCS1 padding.  So there is a need for exposing the SC_ALGORITHM_RSA_RAW.
Most modern cards only support SC_ALGORITHM_RSA_RAW, and let the host do the padding.

http://en.wikipedia.org/wiki/RSA_(cryptosystem)#Padding_schemes

>
> In line 4704 you seem to disable signing if the card supports RAW RSA, but Padding can be done on the host-side and then be sent to the card as RAW block. Disabling the padding-decryption case (4733)
> is IMHO wrong for the above reasons.

I total agree. It will break otrher cards.

>
> Maybe the mechanism of registering card capabilities does not always cover all requirements. We might need a way to specifiy:
>
>   * for each key type (RSA/ECC/..)

It already does.

>   * what precalculated hashes the card accepts, or at least whether it does enforce that the hash is done on-card

Once the data is hashed, the card does can not tell if is a hash, a symmetric key or some random data.
and the card does not care.

>   * what hashes the card can perform by itself

I don't think there are any cards that do hashes on the card. They are the output of a digest routine
that can take large amounts of data. Cards are slow, and if the card was to do the digest to produce the hash,
you would need to send the all the data to the card. Newer hashes are have been added since PKCS#11
standards have been written. Allowing the caller to do the hashing,

Hashing could be done on the card, which also implies padding needs to be done on the card as well.
But lets get past the padding issues first.

>   * in what operations (sign/decrypt)

Decrypt does not use hashes, RSA decrypt returns a small amount of data that may be the actual data that
was encrypted, but  usually return the symmetric key.

I am still trying to figure out what Frank's problem is.

running some combination of these commands:

grep -l "SC_ALGORITHM_RSA_PAD_PKCS1" card-*.c
grep -l "SC_ALGORITHM_RSA_RAW" card-*.c

might help to show how different cards support combinations of RAW and/or PKCS1 padding.

If I did my greps correctly this should show all the cards that support SC_ALGORITHM_RSA_PAD_PKCS1 but not SC_ALGORITHM_RSA_RAW:

$grep -L "SC_ALGORITHM_RSA_RAW" `grep -l "SC_ALGORITHM_RSA_PAD_PKCS1" card-*.c`
card-atrust-acos.c
card-authentic.c
card-belpic.c
card-gemsafeV1.c
card-gpk.c
card-ias.c
card-iasecc.c
card-isoApplet.c
card-jcop.c
card-oberthur.c
card-starcos.c

And these only support SC_ALGORITHM_RSA_RAW:
$ grep -L "SC_ALGORITHM_RSA_PAD_PKCS1" `grep -l "SC_ALGORITHM_RSA_RAW" card-*.c`
card-asepcos.c
card-cardos.c
card-dnie.c
card-entersafe.c
card-epass2003.c
card-flex.c
card-incrypto34.c
card-itacns.c
card-muscle.c
card-piv.c
card-rtecp.c


Frank, have a look at this piece of code in pkcs15-sec.c. Do you need to set SC_ALGORITHM_NEED_USAGE
to get around your problem?

389         /* if the card has SC_ALGORITHM_NEED_USAGE set, and the
390            key is for signing and decryption, we need to emulate signing */
391         /* TODO: -DEE assume only RSA keys will ever use _NEED_USAGE */
392
393         sc_log(ctx, "supported algorithm flags 0x%X, private key usage 0x%X", alg_info->flags, prkey->usage);
394         if ((alg_info->flags & SC_ALGORITHM_NEED_USAGE) &&
395                 ((prkey->usage & USAGE_ANY_SIGN) &&
396                 (prkey->usage & USAGE_ANY_DECIPHER)) ) {
397                 size_t tmplen = sizeof(buf);
398                 if (flags & SC_ALGORITHM_RSA_RAW) {
399                         r = sc_pkcs15_decipher(p15card, obj,flags, in, inlen, out, outlen);
400                         LOG_FUNC_RETURN(ctx, r);
401                 }



>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88217191>.
>

-- 

viktorTarasov:
@dengert 
> I don't think there are any cards that do hashes on the card

Some cards perform the final stage of hash calculation and sign the result in one COMPUTE_SIGNATURE operation.
That's how the *qualified signature* is done with IAS/ECC card .

philipWendland:
@dengert 
> Not sure what you mean by this.

Sorry for being unclear; I was trying to say, I don't see why a specific *application* would request OpenSC to use RAW RSA for signatures, except that the application itself does the padding beforehand.

> There is more then one type of padding, but if a card only supports SC_ALGORITHM_RSA_PAD_PKCS1 you are stuck with
> only the PKCS1 padding.  So there is a need for exposing the SC_ALGORITHM_RSA_RAW.
> Most modern cards only support SC_ALGORITHM_RSA_RAW, and let the host do the padding.

There are (security) reasons to enforce padding, too (Be it in the application or the card). 
However, what is important here is that there are cards that enforce to do the padding on-card and others don't.

>  Maybe the mechanism of registering card capabilities does not always cover all requirements. We might need a way to specifiy:
> 
>    * for each key type (RSA/ECC/..)
> It already does.

Not with different flags. See https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L4662 ff., the flags are or'ed for every key-type. So currently there is a problem if a card driver needs to set other flags / hashes for different algorithms.

> Once the data is hashed, the card does can not tell if is a hash, a symmetric key or some random data.
> and the card does not care.

Unless it enforces restrictions on the length, like 20 bytes for SHA-1 input (or others) only. 
But I agree, that does not make much sense to me either.

> But lets get past the padding issues first.

OK. 

> Decrypt does not use hashes, RSA decrypt returns a small amount of data that may be the actual data that
> was encrypted, but  usually return the symmetric key.

I know, but a card might accept different padding schemes for sign/decrypt, like Frank's card seems to do. Unless it really does, I'm not sure whether I really understood the problem yet.

@frankmorgner 
> But I believe that the output of the signature (which now needs to be PKCS#1 padded in software) will not be compatible with the typical raw RSA output.

I don't really understand this. What exactly do you refert to as "output of the signature"?
The padding needs to be done before encrypting with the private key, and I understand the Encrypt(Pad(Hash(data))) as "output of the signature".

frankmorgner:
@dengert I tried `SC_ALGORITHM_NEED_USAGE`, but I did not get further.

@philipWendland What I want to achieve is that the card driver's `set_security_env` is only called with one of the following configurations:

* `SC_ALGORITHM_RSA_PAD_PKCS1` + `SC_SEC_OPERATION_SIGN` or
* `SC_ALGORITHM_RSA_RAW` + `SC_SEC_OPERATION_DECIPHER`

But at the card driver's level I only found out that `_sc_card_add_rsa_alg` can be used to generally limit the algorithms to `SC_ALGORITHM_RSA_PAD_PKCS1|SC_ALGORITHM_RSA_RAW` which would also allow the following unwanted configurations for `set_security_env`:

* `SC_ALGORITHM_RSA_RAW` + `SC_SEC_OPERATION_SIGN` and
* `SC_ALGORITHM_RSA_PAD_PKCS1` + `SC_SEC_OPERATION_DECIPHER`

 Sorry for not being clearer, I am not too familiar with the upper layers of OpenSC.

dengert:
On 4/1/2015 12:07 PM, Philip Wendland wrote:
> @dengert <https://github.com/dengert>
>
>     Not sure what you mean by this.
>
> Sorry for being unclear; I was trying to say, I don't see why a specific /application/ would request OpenSC to use RAW RSA for signatures, except that the application itself does the padding beforehand.
>
>     There is more then one type of padding, but if a card only supports SC_ALGORITHM_RSA_PAD_PKCS1 you are stuck with
>     only the PKCS1 padding. So there is a need for exposing the SC_ALGORITHM_RSA_RAW.
>     Most modern cards only support SC_ALGORITHM_RSA_RAW, and let the host do the padding.
>
> There are (security) reasons to enforce padding, too (Be it in the application or the card).

If it going to be enforced, in must be on the card or by the verifier. A user could always modify
OpenSC or use some other middleware.


> However, what is important here is that there are cards that enforce to do the padding on-card and others don't.

Yes, I believe that the current OpenSC code supports this, and will use the card's padding if possible.
Some cards only support RSA-PKCS1, so The OpenSC code must be using the card has to do the padding.

As Viktor pointed out the IAS/ECC can go one setp further, and do the final hash round then padding then RSA on the card.

>
>     Maybe the mechanism of registering card capabilities does not always cover all requirements. We might need a way to specifiy:
>
>       * for each key type (RSA/ECC/..) It already does.
>
> Not with different flags. See https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L4662 ff., the flags are or'ed for every key-type. So currently there is a problem if a card
> driver needs to set other flags / hashes for different algorithms.

OK, That's my TODO comment at line #4656.

The only issue would have to do with which hashes mechanisms are supported,
and only for RSA does PKCS#11 try and spell these out. Table 34 in PKCS#11 V2.20.
For ECDSA only CKM_ECDSA_SHA1 is defined for DSA only CKM_DSA_SHA1.

Based on the or'ed flags, register_ec_mechanisms, register_gost_mechanisms,
or for RSA, sc_pkcs11_new_fw_mechanism(CKM_RSA_X_509, &mech_info, CKK_RSA, NULL) are called.

sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS, &mech_info, CKK_RSA, NULL) is always called, all cards are assumed
to do some type of RSA, and padding can be done in in the OpenSC padding.c

Only RSA does padding.

>
>     Once the data is hashed, the card does can not tell if is a hash, a symmetric key or some random data.
>     and the card does not care.
>
> Unless it enforces restrictions on the length, like 20 bytes for SHA-1 input (or others) only.
> But I agree, that does not make much sense to me either.
>
>     But lets get past the padding issues first.
>
> OK.
>
>     Decrypt does not use hashes, RSA decrypt returns a small amount of data that may be the actual data that
>     was encrypted, but usually return the symmetric key.
>
> I know, but a card might accept different padding schemes for sign/decrypt, like Frank's card seems to do. Unless it really does, I'm not sure whether I really understood the problem yet.
>
> @frankmorgner <https://github.com/frankmorgner>
>
>     But I believe that the output of the signature (which now needs to be PKCS#1 padded in software) will not be compatible with the typical raw RSA output.
>
> I don't really understand this. What exactly do you refert to as "output of the signature"?
> The padding needs to be done before encrypting with the private key, and I understand the Encrypt(Pad(Hash(data))) as "output of the signature".
>

I would still like to here from Frank too especially on the SC_ALGORITHM_NEED_USAGE 3 card drivers use this flag.
Is this the flag he needs to set to avoid this pull-request?



> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88552747>.
>

-- 

dengert:
On 4/1/2015 12:26 PM, Frank Morgner wrote:
> @dengert <https://github.com/dengert> I tried |SC_ALGORITHM_NEED_USAGE|, but I did not get further.
>
> @philipWendland <https://github.com/philipWendland> What I want to achieve is that the card driver's |set_security_env| is only called with one of the following configurations:
>
>   * |SC_ALGORITHM_RSA_PAD_PKCS1| + |SC_SEC_OPERATION_SIGN| or
>   * |SC_ALGORITHM_RSA_RAW| + |SC_SEC_OPERATION_DECIPHER|
>
> But at the card driver's level I only found out that |_sc_card_add_rsa_alg| can be used to generally limit the algorithms to |SC_ALGORITHM_RSA_PAD_PKCS1|SC_ALGORITHM_RSA_RAW| which would also allow
> the following unwanted configurations for |set_security_env|:

Have you seen these unwanted configurations get through the upper layers of OpenSC?

>
>   * |SC_ALGORITHM_RSA_RAW| + |SC_SEC_OPERATION_SIGN| and
>   *

So you want to enforce that the card must do the padding?
This would limit the card to only supporting PKCS1

Do you have OpenSC debug output showing how your card gets into one of these configurations?

If it can get into one of these, you must be passing
flags = SC_ALGORITHM_RSA_PAD_PKCS1|SC_ALGORITHM_RSA_RAW



>
>     |SC_ALGORITHM_RSA_PAD_PKCS1| + |SC_SEC_OPERATION_DECIPHER|
>
>     Sorry for not being clearer, I am not too familiar with the upper layers of OpenSC.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88565720>.
>

-- 

frankmorgner:
@dengert the unwanted configuration is triggered by `pkcs11-tool --test`, but I have not yet seen it in a real application...

dengert:
I have tried your mod with the PIV card, and indeed it causes problems. 

(In the following p11t is an alias for pkcs11-tool --module ...)

With the mod:
$p11t -M
  RSA-X-509, keySize={1024,3072}, hw, decrypt, verify
  RSA-PKCS, keySize={1024,3072}, hw, sign, verify

$ p11t -s -i sig -l -d 01 -m RSA-X-509 -o sig.RSA-X-X509.mod
Logging in to "PIV_II (PIV Card Holder pin)".
Please enter User PIN: 
Using signature algorithm RSA-X-509
Aborting.

This would simulate a PKCS#11 caller doing their own padding then using the 
RSA-X-509 mechanism to do the signature.

Without the mod:
$p11t -M
  RSA-X-509, keySize={1024,3072}, hw, decrypt, sign, verify
  RSA-PKCS, keySize={1024,3072}, hw, decrypt, sign, verify

$ p11t -s -i sig -l -d 01 -m RSA-X-509 -o sig.RSA-X-X509.nomod
Logging in to "PIV_II (PIV Card Holder pin)".
Please enter User PIN: 
Using signature algorithm RSA-X-509

It looks like the mod is modifying:
   PKCS#11 V2.20 
     Section 12 Mechanisms,
       Table 34, Mechanisms vs. Functions.

CKM_RSA_PKCS removes the Decrypt check
CKM_RSA_X_509 removs the Sign check. 

The paragraph above Table 34 does say:
 "There is also no guarantee that a token which supports one mechanism for some operation supports any other mechanism for any other operation (or even supports that same mechanism for any other operation)."
This sounds like the table is not fast and hard.

*BUT* the mod affects ALL cards, which it should not do. 

If there is some card that will enforce signatures to use the CKM_RSA_PKCS
i.e. not allow the use of CKM_RSA_X_509 with the same key and the card
can tell the difference between an attempt to use the RAW mode for signing vs decrypt
then we need a way for the card driver to pass in a flag to do something  like what 
the mod proposed,  but only for that card. 

The card needs to enfoce that restrictions by key. If the card can not enforce it, then a user/hacker could get around it by remving any code we adde to OpenSC, or use other code. 

Keep in mind, by the nature of RAW RSA, the following all produce the same outout:

$p11t -s -i sig -l -d 01 -m RSA-X-509 -o sig.RSA-X-X509.nomod
$pkcs15-crypt -c -i sig -k 01 -R -o pkcs15-crypt-c.nomod
$pkcs15-crypt -s -i sig -k 01 -R -o pkcs15-crypt-s.nomod

PIV card, RSA 2048 key, input and output are all 256 bytes.

dengert:
Modifing the CK_MECHANISM_INFO flags is a course grain control, as it applies to all keys 
that can use that mechanism, which may work for your card.

For fine grain control, you may want to look at Table 27, Common Key Attributes, CKA_ALLOWED_MECHANISMS .
But it looks like OpenSC does not support returing that attribute or setting it up.

The more restrictions you place on the keys, the harder it will be for calling applications
to use your card.

frankmorgner:
I know that the changes in the PR are suboptimal ;-) I primarily wanted to get some feedback on the problem, thanks so far.

I did some more digging through the code and it seems that in EF.TokenInfo there is a sequence supportedAlgorithms that splits into key references, mechanisms and operations. This looks like to be what I need to change for excluding a certain combination of mechanism+operation...

dengert:
Note that PKCS#11 2.10 (1998) does not have CKA_ALLOWED_MECHANISMS for a key. It shows up on PKCS#11 v 2.20 (2004)
The TokenInfo is from PKCS#15 1.1 (2000), and I don't see in PKCS#15 any references to padding or hashing
being restricted to a card. That may have been discovered as a missing piece between 2000 and 2004
then added to PKCS#11 2.20.

What you may be seeing in the OpenSC code is meant to cover mechanisms for the card, RSA, DSA, ECDSA, and
not the PKCS#11 combination mechanisms like RSA-PKCS1

Rather then looking at the OpenSC code, see what PKCS#15 says about the issues.

PKCS#15 may not have the information needed to restrict selected keys to force padding to be done
on the card.


On 4/2/2015 11:09 AM, Frank Morgner wrote:
> I know that the changes in the PR are suboptimal ;-) I primarily wanted to get some feedback on the problem, thanks so far.
>
> I did some more digging through the code and it seems that in EF.TokenInfo there is a sequence supportedAlgorithms that splits into key references, mechanisms and operations. This looks like to be
> what I need to change for excluding a certain combination of mechanism+operation...
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/419#issuecomment-88961342>.
>

-- 

frankmorgner:
@dengert allowedAlgorithms is part of ISO 7816-15 at least since 2004. Though sometimes only loosely coupled, I think ISO 7816-15 superseeded PKCS#15 a long time ago. It can even encode PACE-secrets soon...

dengert:
Looks like it did. 
 http://www.emc.com/emc-plus/rsa-labs/standards-initiatives/pkcs-15-cryptographic-token-information-format.htm
Buit OpenSC is based on PKCS#15 or card drivers that can emulate parts of it) with a PKCS#11 v2.20 API. If you want to upgrade OpenSC to newer versions of the standards. you will need to convince the comunity.

dengert:
ANd if you weant to start that conversionation, start with a new github issue or sendto the opensc-dev list.

PR #429, opened 2015-04-06
register mechanisms more precisely

philipWendland:
**This PR should be reviewed carefully, it might affect cards that I can not test with**

In a nutshell, this PR addresses the following:

  - support different flags for different key types
  - register ECDH mechanisms only if card driver indicates support with SC_ALGORITHM_ECDH_CDH_RAW

It touches card driver code for piv (@dengert), sc-hsm (@cardcontact) and myeid (who is the maintainer?), only to keep the existing behavior: It adds the SC_ALGORITHM_ECDH_CDH_RAW flag, because the cards/drivers seem to support ECDH.

How exactly is the SC_ALGORITHM_ECDH_CDH_RAW flag defined? Do we need a seperate flag for ECDH with cofactor multiplication (for curves with a cofactor=1 it should produce the same result)?

I don't have a way to test how the change (seperation of flags for different algorithms) affects GOSTR. The rtcep card seems to call _sc_card_add_algorithm for GOSTR with seperate flags. The change does not look like it will affect the existing behavior.

### DIFF
--- a/src/libopensc/card-isoApplet.c
+++ b/src/libopensc/card-isoApplet.c
@@ -233,9 +233,10 @@ isoApplet_init(sc_card_t *card)
 		 * driver. */
 		flags = 0;
 		flags |= SC_ALGORITHM_ECDSA_RAW;
+		flags |= SC_ALGORITHM_ECDSA_HASH_NONE;
 		flags |= SC_ALGORITHM_ONBOARD_KEY_GEN;
-		flags |= SC_ALGORITHM_EXT_EC_UNCOMPRESES;
-		ext_flags =  SC_ALGORITHM_EXT_EC_NAMEDCURVE;
+		ext_flags = SC_ALGORITHM_EXT_EC_UNCOMPRESES;
+		ext_flags |=  SC_ALGORITHM_EXT_EC_NAMEDCURVE;
 		ext_flags |= SC_ALGORITHM_EXT_EC_F_P;
 		for (i=0; ec_curves[i].oid.value[0] >= 0; i++)
 		{
--- a/src/libopensc/card-isoApplet.c
+++ b/src/libopensc/card-isoApplet.c
@@ -233,7 +233,7 @@ isoApplet_init(sc_card_t *card)
 		 * driver. */
 		flags = 0;
 		flags |= SC_ALGORITHM_ECDSA_RAW;
-		flags |= SC_ALGORITHM_ECDSA_HASH_NONE;
+		flags |= SC_ALGORITHM_ECDSA_HASH_SHA1;
 		flags |= SC_ALGORITHM_ONBOARD_KEY_GEN;
 		ext_flags = SC_ALGORITHM_EXT_EC_UNCOMPRESES;
 		ext_flags |=  SC_ALGORITHM_EXT_EC_NAMEDCURVE;
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -158,7 +158,7 @@ static int myeid_init(struct sc_card *card)
 			card->version.fw_major >= 4)   {
 		int i;
 
-	        flags |= SC_ALGORITHM_ECDSA_RAW;
+		flags |= SC_ALGORITHM_ECDSA_RAW | SC_ALGORITHM_ECDH_CDH_RAW;
 		ext_flags = SC_ALGORITHM_EXT_EC_NAMEDCURVE | SC_ALGORITHM_EXT_EC_UNCOMPRESES;
 
 		for (i=0; ec_curves[i].curve_name != NULL; i++)
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -158,7 +158,8 @@ static int myeid_init(struct sc_card *card)
 			card->version.fw_major >= 4)   {
 		int i;
 
-		flags |= SC_ALGORITHM_ECDSA_RAW | SC_ALGORITHM_ECDH_CDH_RAW;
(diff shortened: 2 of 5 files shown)


dengert:
https://www.nsa.gov/ia/_files/suiteb_implementer_g-113808.pdf
says:  "Note that for the Suite B curves, the cofactor h = 1, which reduces “cofactor ECDH” to “ordinary ECDH”." which is were I believe the the comment came from for the PIV card. This may not be a general statement, but only apply to the 2 curves supported by the PIV. 

NIST 800-73-3 Part 2 says:
A. 4.2.1 ECDH with the PIV KMK
The PIV Card supports ECDH key agreement by performing the Elliptic Curve Cryptography Cofactor Diffie-Hellman (ECC CDH) Primitive (see Section 5.7.1.2 of SP 800-56A [7]) using its ECC KMK private key and an ECC public key that is provided as input to the GENERAL AUTHENTICATE command. All other procedures required to complete the key agreement are performed by the cardholder’s client application and its associated cryptographic module.

philipWendland:
AFAICT, the Nist[1] and Brainpool[2] curves over prime fields all have cofactor=1. However, for example, NIST curves of binary fields have cofactor=2 or 4.

Does OpenSC, or any card, support ECDH with cofactor multiplication at all? I.E. should CKM_ECDH1_COFACTOR_DERIVE be registered at all at the current state?

[1] http://csrc.nist.gov/groups/ST/toolkit/documents/dss/NISTReCur.pdf
[2] http://www.ecc-brainpool.org/download/draft-lochter-pkix-brainpool-ecc-00.txt

dengert:
The PIV only supports 2 curves, P-256 and P-384, and does not need CKM_ECDH1_COFACTOR_DERIVE
The way I read it, "P" curves are prime with cofactor=1 and CKM_ECDH1_DERIVE give same results.

Not sure about other cards.

On 4/7/2015 3:47 PM, Philip Wendland wrote:
> AFAICT, the Nist[1] and Brainpool[2] curves over prime fields all have cofactor=1. However, for example, NIST curves of binary fields have cofactor=2 or 4.
>
> Does OpenSC, or any card, support ECDH with cofactor multiplication at all? I.E. should CKM_ECDH1_COFACTOR_DERIVE be registered at all at the current state?
>
> [1] http://csrc.nist.gov/groups/ST/toolkit/documents/dss/NISTReCur.pdf
> [2] http://www.ecc-brainpool.org/download/draft-lochter-pkix-brainpool-ecc-00.txt
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/429#issuecomment-90728383>.
>

-- 

philipWendland:
I think the card drivers currently don't have a way of differentiating between ECDH and codfactor ECDH anyway (They seem to detect ECDH by checking for (1) EC algorithm and (2) derive operation). So this change might be big and break things - I think it should be postponed until we know/have a card that produces different results for those algorithms.

A different point, I saw:
https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L3563
I think it might be beneficial to differentiate between CKM_ECDSA and CKM_ECDSA_SHA1 as well, depending on whether the card driver did set HASH_NONE and HASH_SHA1. I added this as a commit (bc747180037e88eab442f9a445a7238a56e82c66), do you agree with it? Should I set HASH_SHA1 for the PIV, too?
Thanks.

dengert:
Agree with postponing  cofactor ECDH.

The CKM_ECDSA is a single part operation. CKM_ECDSA_SHA1 is multi-part operation and does the hash,
then the CKM_ECDSA of the hash. This could get messy, as the HASH could be in software or hardware.

The use of SHA1 for signatures is deprecated and being phased out everywhere these days.
So I would not implement SHA-1 now. PKCS#11 does not specify any other hashing for ECDSA leaving it up to the caller to do the hashing.
So no need to add a ECDSA hashing in to the OpenSC pkcs11.

The PIV can not do the hash in hardware anyway. Even if someone else wants SHA1 for their card, don't
touch the PIV ECDSA.

On 4/8/2015 3:30 PM, Philip Wendland wrote:
> I think the card drivers currently don't have a way of differentiating between ECDH and codfactor ECDH anyway (They seem to detect ECDH by checking for (1) EC algorithm and (2) derive operation). So
> this change might be big and break things - I think it should be postponed until we know/have a card that produces different results for those algorithms.
>
> A different point, I saw:
> https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L3563
> I think it might be beneficial to differentiate between CKM_ECDSA and CKM_ECDSA_SHA1 as well, depending on whether the card driver did set HASH_NONE and HASH_SHA1. I added this as a commit (bc74718
> <https://github.com/OpenSC/OpenSC/commit/bc747180037e88eab442f9a445a7238a56e82c66>), do you agree with it? Should I set HASH_SHA1 for the PIV, too?
> Thanks.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/429#issuecomment-91028329>.
>

-- 

philipWendland:
> This could get messy, as the HASH could be in software or hardware. The use of SHA1 for signatures is deprecated and being phased out everywhere these days. So I would not implement SHA-1 now. PKCS#11 does not specify any other hashing for ECDSA leaving it up to the caller to do the hashing. So no need to add a ECDSA hashing in to the OpenSC pkcs11. 

I agree, I don't see a reason to spend time to extend CKM_ECDSA_SHA1 either. 

(The Java Card API <= 3.0.3 has a really weird limitation regarding ECDSA, basically limiting it to CKM_ECDSA_SHA1. Really looking forward to obtaining a Java Card 3.0.4...)

PR #431, opened 2015-04-07
Issue 395. Commenting a couple of needed free statements in order to avoid failures...

germanblanco:
This changes propose to skip to required free statements for now, in order to avoid failures since the memory is used later in the code. Rewriting will be required in order to free all allocated memory.
There is a tiny non-related change of adding dnie-tool to .gitignore.


germanblanco:
I will make a different proposal that seems to avoid this memory leak.

PR #432, opened 2015-04-08
Issue 395. Avoiding a couple of memory leaks.

germanblanco:
Would this way forward be acceptable?
The buffers are freed now in all cases, except when apdubuf is assigned to the data pointer in the apdu.
This will happen often, I think.
And without digging too much, I have seen a few other places where pointers seem to be overwritten or left out of scope without being properly freed. I would work on fixing this, but it will take some time and I would prefer to do it in several steps. Is that ok?
If someone is willing to help, I would be very happy.


viktorTarasov:
applied in 6caa85f238d437198f92a70644675df7e2128c20,
thank you.

PR #435, opened 2015-04-10
Honour key capabilities for signing and decryption

frankmorgner:
I solved the problem from #419 by creating a PKCS#15 emulator and assigning specific capabilities to the signature and decryption keys. Now only the supported configurations can be triggered by the application via `C_SignInit`.


viktorTarasov:
Applied in d7ab0df51d3df026522f811e5d86809112e294d2..6e84ee0ba779c4c527f5c979ef2751ee256da0ba,
thank you.

PR #437, opened 2015-04-14
check ATRs even for forced card driver

frankmorgner:
some card drivers depend on a card type which is initialized by matching the card's ATR


viktorTarasov:
Applied in a7a903fd811fbe95606fad4b8087f936d6f711cf,
thank you.

PR #438, opened 2015-04-14
pkcs11-tool: Add support for creating EC privkey

tc-anssi:
* This patch allows to create EC private keys. The feature re-use the GOST
  parsing function as instanciating an ECDSA key is the same as a GOST key.
  This way we can import existing ECDSA keys on a token using pkcs11-tool.


viktorTarasov:
Applied in 23ca1f101db0b4fca09bfb51b4fb4d67dbfc2435,
thank you.

Issue #433, opened 2015-04-09
MyEID: failed to import EC private key bignum with length less then key size.


viktorTarasov:
Following key cannot be imported:
```
   0   94: SEQUENCE {
   2    1:   INTEGER 1
   5   23:   OCTET STRING
         :     7B 0C DA D3 F8 AA 7B 9B 13 DD 2A 1C 36 73 E7 6B
         :     69 40 EF 1C 4E CC 0E
  30   10:   [0] {
  32    8:     OBJECT IDENTIFIER ansiX9p192r1 (1 2 840 10045 3 1 1)
         :     }
  42   52:   [1] {
  44   50:     BIT STRING
         :       04 6E 92 D8 B3 B6 46 26 EB 96 0C A6 10 79 C3 50
         :       10 C7 99 68 A1 FC D1 55 31 18 14 B7 56 F1 E4 60
         :       A4 4D C1 ED 05 EF EA DF 6A 22 D7 90 AB F1 56 C8
         :       10
         :     }
         :   }
```

dengert:
Sounds like it is related to:
http://openssl.6102.n7.nabble.com/openssl-dev-Reminder-OpenSSL-s-EC-private-key-encoding-is-broken-tt57165.html#none

where the length of a bignum is used to create the OCTETSTRING, and leading zeros are lost. Note the patch has been applied to OpenSSL. It might take a while  for the patch to make it into the field.

Something similar could be used in OpenSC to pad the short string where the length is gotten from the curve. For a 192 bit curve, the octes string should  be 24.

dengert:
The field length in bits may not be a multiple of 8. openssl ecapram -list_curve
show  curves with field lengths of  521, 571, 239, 233 and many more. 

dst->params.field_length = EC_GROUP_get_degree(grp);

buf_len= (dst->params.field_length +7) / 8;

would support these curves too,and dst->params.field.length set correctly.


buf_len = (EC_GROUP_get_degree(a->group) + 7) / 8;

viktorTarasov:
@dengert 
> The field length in bits may not be a multiple of 8 ...

In this line it's forced to be multiple of 8:
https://github.com/OpenSC/OpenSC/commit/96556dea7bff431e794641fcae6b732395e00854#diff-9640be6a2410e9f891deff03f1013247R709

dengert:
There is the field length in bits, and there is the length of an OCTETSTRING in bytes that can
[...]

PR #440, opened 2015-04-17
EC field_length changes for non-multiple of 8 bits curves

dengert:
THIS CODE NEEDS TO BE TESTED. 

While reviewing #438 questions about the OpenSC EC field_length came up. Most curve names have the number of bits in the name.  Early code in OpenSC made the assumption that tey were always a multiple of 8, but this is not the case. When getting memory to hold the X, Y, privateD or other curve parameters, the number of bytes need to hold number = (field_length + 7) / 8 to allow for an extra byte for the odd bits. 

In OpenSC the EC field_length is the number of bits in the field.
Most curves have a field_length which is a multiple of 8 bits
but there are many that are not.

The X and Y points and privateD are stored in octetstrings
so there may need to be an extra byte in the octetstring.

An OpenSSL BIGNUM will drop leading zero bytes, so its size can not be used
to determine the field_length.

### DIFF
--- a/src/libopensc/pkcs15-prkey.c
+++ b/src/libopensc/pkcs15-prkey.c
@@ -705,12 +705,16 @@ sc_pkcs15_convert_prkey(struct sc_pkcs15_prkey *pkcs15_key, void *evp_key)
 		memcpy(dst->ecpointQ.value, buf, buflen);
 		dst->ecpointQ.len = buflen;
 
-		/* calculate the field length */
-		dst->params.field_length = (buflen - 1) / 2 * 8;
+		/*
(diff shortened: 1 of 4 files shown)


dengert:
Also see #433

dengert:
@philipWendland  - Please check  this patch especially the isoApplet.c modification, and the second  commit .

@CardContact  - Please chech this patch, especiall the comment in pkcs15-sc-hsm.c.

philipWendland:
Will test late sunday / monday when I am home.

philipWendland:
Seems to work, thank you. One less stumbling block when those curve(s) are added to the pkcs15init driver.

PR #443, opened 2015-04-19
minidriver update

vletoux:
modification minidriver

merge CardUnblockPin and CardChangeAuthenticator into CardChangeAuthenticatorEx
add support for the msroot file
avoid to ask the PUK twice (especially on pinpad)
change the UI
add a translation function to change opensc error code to windows one


vletoux:
@viktorTarasov or someone else, it is possible to accept this pull request into the minidriver test branch ?
I committed my changes in my local master branch and doing a pull request for the card driver I'd like to propose is becoming a branching / merging hell. I'd like to restart from scratch from the OpenSC repository without loosing the changes referenced in this pull request.

There is 2 more things in my todo list:
- add the opensc icon to the pinpad dialog instead of the windows default icon
- when the card is removed while requesting the pinpad, the Windows AD logon screen isn't refreshed. I suspect that the correct error code isn't returned, forcing the display of the error dialog while the background thread request to delete the logon tile. This change will probably need a new error code to avoid that.

PR #446, opened 2015-04-22
Honour HAVE_CONFIG_H

frankmorgner:


LudovicRousseau:
It would be nice to explain in the commit message _why_ you made this change.

I have no idea of the problems your patch will solve.

frankmorgner:
Sorry, I thought this would be self-explanatory: It lets the code compile when no `config.h` is available. ;-)

This currently is no problem either because autoconf generates one on Mac OS X/Linux and the user is required to supply one on Windows. However, in an external card driver I am using some internal functions that are not exported from OpenSC (see https://github.com/frankmorgner/vsmartcard/blob/master/npa/src/scutil.c#L25-L37 and https://github.com/frankmorgner/vsmartcard/blob/master/npa/src/boxing.c#L958-L960). Here, the content of my project's `config.h` conflicts with what OpenSC expects. That's why I need to disable `config.h` when including OpenSC files

LudovicRousseau:
I can understand _what_ the patch does but not _why_ the patch is needed.
Thanks for the details.

More than once I had the problem that I re-read code I wrote 10 years ago and I have not idea _why_ I made a particular change.

PR #448, opened 2015-04-23
Create minidriver-italian-cns.reg

sschutte:
These are the required registry values to get the Italian CNS card working under Windows 7 32 bit and 64 bit.


frankmorgner:
I think the file should end with `.reg`, right?

Is there a proprietary middleware that could be used for the card (i.e. a potential conflict with a different minidriver)?

sschutte:
Yes my mistake, the file should end with ```.reg```. I fixed the error.

There is definitely proprietary middlware that can be used for the card. In fact there are two varients: The proprietary Siemens middleware for the older cards and the proprietary Actalis middleware for the newer cards. So far however, I have not noticed any potential conflicts when testing with the minidriver. Is there a maybe better way to check for potential confilcts? So far, having the proprietary middleware and OpenSC installed side by side, there have been no noticable errors.

frankmorgner:
@vletoux do you know how to test for conflicts?

vletoux:
@frankmorgner if you add registry file directly no. A custom action has to be created to check for existing registration. It can disable the installation of the component. A custom action is a dll. No problem to write it, but this requires a new directory.

vletoux:
@frankmorgner second option: make the registry creation / deletion at uninstall from the custom action dll.

frankmorgner:
OK, I see. One needs to have this in mind when creating a default installer for the minidriver, but if installed directly to the registry it should work. Thanks @vletoux .

@sschutte how does it come that the itacns card driver only knows the ATR "3b:f4:18:00:ff:81:31:80:55:00:31:80:00:c7" but the registry entries use (three) different ATRs?

frankmorgner:
I just saw that `itacns_match_cns_card` is a bit more complex than I thought. `itacns_atr_match` mimics the behavior of `_sc_match_atr` when provided with an `atrmask` (member of `struct sc_atr_table`. Actually I think that the ATR matching should be fixed to only use `_sc_match_atr` with a correct mask, because this very ATR mask can be reused in the registry entry, too!

dengert:
In regard to conflicts, one is the GUID used to register certificates in the cert store. OpenSC and the vendor code may provide different GUIDs for the same certificate. Thus if the user had been using the vendor code, then installed OpenSC, an attemp to use an existing certificate from the store will request the card based on the stored GUID. This will not match the GUID now reported by the new minidriver.
The may not be a problem fot the CNS cards, but may for others.

It is also not clear what Microsoft does if multiple registry entires can handle the ATR. Which one does it use?

vletoux:
What I'm planning to do is to create a custom action: if the card is
already registered => nothing
This will solve the problem.

2015-04-24 14:08 GMT+02:00 Doug Engert <notifications@github.com>:

> In regard to conflicts, one is the GUID used to register certificates in
> the cert store. OpenSC and the vendor code may provide different GUIDs for
> the same certificate. Thus if the user had been using the vendor code, then
> installed OpenSC, an attemp to use an existing certificate from the store
> will request the card based on the stored GUID. This will not match the
> GUID now reported by the new minidriver.
> The may not be a problem fot the CNS cards, but may for others.
>
> It is also not clear what Microsoft does if multiple registry entires can
> handle the ATR. Which one does it use?
>
> —
> <https://github.com/OpenSC/OpenSC/pull/448#issuecomment-95912217>.
>



-- 
---
Vincent

dengert:
OK.


On 4/24/2015 7:36 AM, vletoux wrote:
> What I'm planning to do is to create a custom action: if the card is
> already registered => nothing
> This will solve the problem.
>
> 2015-04-24 14:08 GMT+02:00 Doug Engert <notifications@github.com>:
>
>  > In regard to conflicts, one is the GUID used to register certificates in
>  > the cert store. OpenSC and the vendor code may provide different GUIDs for
>  > the same certificate. Thus if the user had been using the vendor code, then
>  > installed OpenSC, an attemp to use an existing certificate from the store
>  > will request the card based on the stored GUID. This will not match the
>  > GUID now reported by the new minidriver.
>  > The may not be a problem fot the CNS cards, but may for others.
>  >
>  > It is also not clear what Microsoft does if multiple registry entires can
>  > handle the ATR. Which one does it use?
>  >
>  > —
>  > Reply to this email directly or view it on GitHub
>  > <https://github.com/OpenSC/OpenSC/pull/448#issuecomment-95912217>.
>  >
>
>
>
> --
> ---
> Vincent
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/448#issuecomment-95919008>.
>

-- 

frankmorgner:
thanks

PR #449, opened 2015-04-27
Fixed broken C_Decrypt from 643080baf9ffd404c88cb99493668eacfff3ab5e

CardContact:
Add handling for CKR_FUNCTION_NOT_SUPPORTED returned for cards not supporting algorithm references with private keys.

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -593,7 +593,7 @@ done:
 }
 
 /*
- * Initialize a signature operation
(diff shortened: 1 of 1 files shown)


frankmorgner:
Could you please also check `sc_pkcs11_verify_init`? The check for `CKR_FUNCTION_NOT_SUPPORTED` is missing there, too.

CardContact:
Done, however I have no way to test this. Does anyone have a card that supports C_Verify ?

PR #453, opened 2015-05-02
Inclusion of the MaskTech card driver for MTCOS cards

vletoux:
As discussed on the opensc-devel mailing list, can you please include this new driver ?
The driver doesn't modify existing functions except the one needed to register it.

### DIFF
--- a/src/libopensc/card-masktech.c
+++ b/src/libopensc/card-masktech.c
@@ -20,7 +20,9 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
+#if HAVE_CONFIG_H
 #include "config.h"
+#endif
 
 #include <stdlib.h>
 #include <string.h>
--- a/src/libopensc/card-masktech.c
+++ b/src/libopensc/card-masktech.c
@@ -1,0 +1,481 @@
+/*
+ * card-masktech.c: Support for Masktech smart cards using the MTCOS operating system.
+ *
+ * Copyright (C) 2011-2015 MaskTech GmbH Fischerstrasse 19, 87435 Kempten, Germany
+ * Copyright (C) 2011 Andrey Uvarov (X-Infotech) <andrejs.uvarovs@x-infotech.com>
+ * Copyright (C) 2015 Vincent Le Toux (My Smart Logon) <vincent.letoux@mysmartlogon.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "config.h"
+
+#include <string.h>
+
+#include "internal.h"
+#include "cardctl.h"
+#include "iso7816.h"
+
+static struct sc_atr_table masktech_atrs[] = {
+	{"3B:89:80:01:4D:54:43:4F:53:70:02:02:05:3B", NULL, NULL,
+	 SC_CARD_TYPE_MASKTECH_GENERIC, 0, NULL},
+	{"3B:9D:13:81:31:60:37:80:31:C0:69:4D:54:43:4F:53:73:02:02:05:41", NULL, NULL,
+	 SC_CARD_TYPE_MASKTECH_GENERIC, 0, NULL},
+	{NULL, NULL, NULL, 0, 0, NULL}
+};
+
+static struct sc_card_operations *iso_ops;
+static struct sc_card_operations masktech_ops;
+static struct sc_card_driver masktech_drv = {
+	"MaskTech Smart Card",
+	"MaskTech",
+	&masktech_ops,
+	masktech_atrs, 0, NULL
+};
+
+struct masktech_private_data {
+	/* save the key reference set at set_masktech_set_security_env to recover it as the signature step */
+	int	rsa_key_ref;
+
+};
+
+/* function not used - remains just in case the inclusion of the application inside dir.c is denied */
+static int masktech_select_aid(struct sc_card *card)
+{
+	struct sc_apdu apdu;
+	unsigned char apdu_resp[SC_MAX_APDU_BUFFER_SIZE];
+	struct auth_private_data *data =  (struct auth_private_data *) card->drv_data;
+	int rv;
+	unsigned char cm[7] = {0xA0,0x00,0x00,0x00,0x03,0x00,0x00};
+	unsigned char aid[15] = {0xE8,0x28,0xBD,0x08,0x0F,0xA0,0x00,0x00,0x01,0x67,0x45,0x53,0x49,0x47,0x4E};
+	struct sc_path tmp_path;
+
+	/* Select Card Manager (to deselect previously selected application) */
+	card->cla = 0x00;
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x0C);
+	apdu.lc = sizeof(cm);
+	apdu.data = cm;
+	apdu.datalen = sizeof(cm);
+	apdu.resplen = sizeof(apdu_resp);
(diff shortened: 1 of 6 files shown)


src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,480 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "config.h"
```
frankmorgner: please use `#ifdef HAVE_CONFIG_H

src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,480 @@
...
+	sc_apdu_t apdu;
+	u8 rbuf[SC_MAX_EXT_APDU_BUFFER_SIZE];
+	u8 sbuf[SC_MAX_EXT_APDU_BUFFER_SIZE];
+	unsigned char use_ext;
+
+	assert(card != NULL && crgram != NULL && out != NULL);
+	sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "masktech_decipher()\n");
+	
+	use_ext = crgram_len > 255; 
+	cse = use_ext ? SC_APDU_CASE_4_EXT : SC_APDU_CASE_4_SHORT;
```
frankmorgner: whether the card supports extended length or not, should be decided in it's init function. This allows setting cse `SC_APDU_CASE_4` which lets the lower level decide about extended or short length.

src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,480 @@
...
+		size_t len = apdu.resplen > outlen ? outlen : apdu.resplen;
+
+		memcpy(out, apdu.resp, len);
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, len);
+	}
+	SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, sc_check_sw(card, apdu.sw1, apdu.sw2));
+}
+
+/* we built a special APDU to be compatible with pinpad readers */
+/* maybe patch opensc to strip the first pin in iso build APDU if SC_PIN_CMD_IMPLICIT_CHANGE is set */
```
frankmorgner: Could you elaborate on the shortcomings of the iso driver? I am more interested in a generic solution...

src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,480 @@
...
+	
+	use_ext = crgram_len > 255; 
+	cse = use_ext ? SC_APDU_CASE_4_EXT : SC_APDU_CASE_4_SHORT;
+	sc_format_apdu(card, &apdu, cse, 0x2A, 0x80, 0x86);
+	apdu.resp = rbuf;
+	apdu.resplen = use_ext ? sizeof(rbuf) : 256;
+	apdu.le = use_ext ? 65536 : 256;
+
+	memcpy(&sbuf, crgram, crgram_len);
+	apdu.data = sbuf;
```
frankmorgner: why copy everything to `sbuf` when you could directly use `crgram` here?

src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,480 @@
...
+	sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "masktech_card_ctl()\n");
+	switch (cmd) {
+		case SC_CARDCTL_GET_SERIALNR:
+			return masktech_get_serialnr(card, (sc_serial_number_t *) ptr);
+		default:
+			return SC_ERROR_NOT_SUPPORTED;
+	}
+}
+
+static struct sc_card_driver *sc_get_driver(void)
```
frankmorgner: function could be renamed to something more specific

vletoux: All other drivers implement sc_get_driver. It is a copy from all driver paste ...
Check also this example: https://github.com/OpenSC/OpenSC/wiki/New-card-driver:-EnterSafe-card-example

frankmorgner: The naming currently suggests that its a global call to OpenSC, no need to copy other peoples mistakes ;-) However, it's just a suggestion. You may leave it as is.

src/libopensc/card-masktech.c
vletoux: The card doesn't support SC_PIN_CMD_CHANGE with the old and new pin at the same time.

I need p1 = 1 in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/iso7816.c#L1009
pin1 is never set while pin2 can be set. But if a pinpad is used, the behavior change because in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/iso7816.c#L1002. It asks for the previous one.

Maybe the test
`if (data->pin1.len != 0 || use_pin_pad) {`
can be changed to : 
`if (data->pin1.len != 0 || (use_pin_pad && !( data->flags & SC_PIN_CMD_IMPLICIT_CHANGE))) {`

Same problem with SC_PIN_CMD_UNBLOCK

But I was advised to not change everything for this driver ;-)

src/libopensc/card-masktech.c
vletoux: fixed as requested

src/libopensc/card-masktech.c
vletoux: fixed as requested

src/libopensc/card-masktech.c
frankmorgner: True, your driver looks good so far. However, what you pointed out looks like a mistake in the iso driver. Could you open a separate PR for changing the iso driver? If accepted, you may rebase #453 on the code with the improved iso driver.

src/libopensc/card-masktech.c
```diff
@@ -0,0 +1,481 @@
...
+	if (crgram_len > sizeof(sbuf)) SC_ERROR_INVALID_ARGUMENTS;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_EXT, 0x2A, 0x80, 0x86);
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+	/* the card doesn't support anything else here (+1 / -1 is not working) */
+	apdu.le = 65536;
+
+	/* crgram is a const u8 while apdu.data is not a const => it cannot be used directly */
+	memcpy(&sbuf, crgram, crgram_len);
```
frankmorgner: Are you operating on an older base? apdu.data is of type `const unsigned char *` which should be equivalent to `const u8 *`

src/libopensc/card-masktech.c
vletoux: @frankmorgner, as suggested, pull request #454

viktorTarasov:
Applied rebased in 976db5cb0400147e9108d434796d68c27f7b4c5c.

src/libopensc/card-masktech.c
frankmorgner: As #454 was accepted, you may want to switch to the iso driver implementation

src/libopensc/card-masktech.c
frankmorgner: did you look into this?

vletoux: I Will asap
I'm sick today

Le mardi 5 mai 2015, Frank Morgner <notifications@github.com> a écrit :

> In src/libopensc/card-masktech.c
> <https://github.com/OpenSC/OpenSC/pull/453#discussion_r29689234>:
>
> > +	u8 sbuf[SC_MAX_EXT_APDU_BUFFER_SIZE];
> > +	u8 rbuf[SC_MAX_EXT_APDU_BUFFER_SIZE];
> > +
> > +	assert(card != NULL && crgram != NULL && out != NULL);
> > +	sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "masktech_decipher()\n");
> > +	
> > +	if (crgram_len > sizeof(sbuf)) SC_ERROR_INVALID_ARGUMENTS;
> > +
> > +	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_EXT, 0x2A, 0x80, 0x86);
> > +	apdu.resp = rbuf;
> > +	apdu.resplen = sizeof(rbuf);
> > +	/* the card doesn't support anything else here (+1 / -1 is not working) */
> > +	apdu.le = 65536;
> > +
> > +	/* crgram is a const u8 while apdu.data is not a const => it cannot be used directly */
> > +	memcpy(&sbuf, crgram, crgram_len);
>
> did you look into this?
>
> —
> <https://github.com/OpenSC/OpenSC/pull/453/files#r29689234>.
>


-- 
--
Vincent Le Toux

My Smart Logon
www.mysmartlogon.com

frankmorgner: Take your time, get well soon!

Issue #395, opened 2015-03-18
Broken Secure channel in dnie driver


miguel-cv:
Hi:

Testing HEAD code, i can't establish a secure session with card. Failure log:
<pre> <code>
0x7f604a43f740 13:48:04.787 [opensc-pkcs11] reader-pcsc.c:182:pcsc_internal_transmit: called
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] apdu.c:185:sc_apdu_log:
Incoming APDU data [    2 bytes] =====================================
90 00 ..
======================================================================
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] apdu.c:397:sc_single_transmit: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] apdu.c:550:sc_transmit: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] card.c:386:sc_unlock: called
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] cwa-dnie.c:787:dnie_transmit_apdu_internal: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] card-dnie.c:1542:dnie_check_sw: called
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] card-dnie.c:1554:dnie_check_sw: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] cwa-dnie.c:870:dnie_wrap_apdu: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] card-dnie.c:1542:dnie_check_sw: called
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] card-dnie.c:1554:dnie_check_sw: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] cwa14890.c:740:cwa_external_auth: returning with: 0 (Success)
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] cwa14890.c:1369:cwa_create_secure_channel: Step 8.4.2: Compute Session Keys
0x7f604a43f740 13:48:05.472 [opensc-pkcs11] cwa14890.c:769:cwa_compute_session_keys: called
[...]

Issue #362, opened 2015-01-22
Bad implementation of DNIe's UI


frankmorgner:
1. The configure option `--disable-openssl` doesn't work with `--enable-dnie-ui`

  The problem is that `sc_ask_user_pin` and `sc_ask_user_consent` depend on internal structures of the DNIe driver. However, this card driver is only compiled if OpenSSL is present.

  Possible solutions:
  - Merge the content of `user-interface.c` into `card-dnie.c`
  - Make the user interface implementation standalone

2. user_consent_app in opensc.conf should be documented in opensc.conf and *not* be hardcoded in card-dnie.c

3. Also, the code in `user-interface.c` is quite messy and needs some cleanup. For example, sc_ask_user_pin is useless because ui_ask_user_pin is not implemented.

germanblanco:
Hello Frank,
I thought that the entire dnie implementation depended on OpenSSL. If I am right with that, "enable-dnie-ui" doesn't make any sense together with "disable-openssl". In any case, the cleanup suggestions sound good to me too, so I will propose some fixes if nobody else volunteers.
Regards,
/German

frankmorgner:
If so, you may want to surround the UI implementation with `#ifdef ENABLE_OPENSSL` to disable compilation when OpenSSL is not available.

germanblanco:
Yes, thank you, it should have been done like that! In fact, the entire card-dnie.c is surrounded by those. I will make a Pull request for that.

frankmorgner:
These problems are still open:

2. user_consent_app in opensc.conf should be documented in opensc.conf and not be hardcoded in card-dnie.c
3. Also, the code in user-interface.c is quite messy and needs some cleanup. For example, sc_ask_user_pin is useless because ui_ask_user_pin is not implemented.

germanblanco:
I am working on it now

rickyepoderi:
@germanblanco I had a look to this file and it is currently used only for displaying the warning that DNIe displays before performing a signature. It is only used if --enable-dnie-ui is used at configure time. It uses OS methods in windows and mac and pinentry in linux. There are two options  that are placed in "card_driver dnie" section, user_consent_app (the pinentry command) and user_consent_enabled (a flag to disable the warning).

The other method sc_ask_user_pin is not used by DNIe (it is not implemented) and can be removed. 

[...]

PR #456, opened 2015-05-06
Issue #451: Newer DNIe not working with OpenSC.

hsanjuan:
This patch fixes 3 issues which consecutively have shown up when debugging the original problem:

1 - Newer DNIe report a byte count for public certificates which is the compressed size,
while older DNIe report the uncompressed size. This resulted in short-reading the x509 certificates,
and in an error parsing. Therefore, during initialization we proceed to set path->count for
public certificates to -1. This ensures that the lenght of the certificates for reading
will be set to file-> length, which has the correct size.

2 - pkcs11-tool -t was broken for DNIe (old and new)as it tried to strip pcks11 padding
from the data to sign and OpenSC tried signatures with non-padded data
(as the card had SC_ALGORITHM_RSA_RAW).
The new algoflags (SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_PAD_PKCS1) and the
removal of the strip-padding call fix the issue.

3 - The new cards won't allow setting the LE bytes when calculating the TLV, when LE equals
256. This caused an wrong SM object error response (0x69 0x88). Therefore,
we don't send the LE bytes anymore in this case.

The patch has been tested to work on the new problematic card and on another old one.

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -497,8 +497,8 @@ static inline void init_flags(struct sc_card *card)
 	card->max_send_size = (255 - 12);	/* manual says 255, but we need 12 extra bytes when encoding */
 	card->max_recv_size = 255;
 
-	algoflags = SC_ALGORITHM_RSA_RAW;	/* RSA support */
-	algoflags |= SC_ALGORITHM_RSA_HASH_NONE;
+	/* RSA Support with PKCS1.5 padding */
+	algoflags = SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_PAD_PKCS1;
 	_sc_card_add_rsa_alg(card, 1024, algoflags, 0);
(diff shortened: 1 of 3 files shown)


frankmorgner:
Your changes look OK.

Are you maybe interested in doing a more detailed review of the dnie driver? In the past we have identified some implementation issues that could pose a problem when used in real world applications...

hsanjuan:
I am maybe interested, but I'll probably be very slow and let's hope I even know how to address the issues. It would be a good starting point to have a list of the issues.

viktorTarasov:
Applied in fd3d07a88496dc0adfa2f6ff7eca9078ec087985.

frankmorgner:
@hsanjuan for problems that need to be fixed, have a look at #362 and the memory leaks from here #395. A good starting point would be to run something like `valgrind pkcs11-tool --test ...` to find bad memory behavior of the driver. Also,the manual TLV handling in the DNIE driver should be changed to use OpenSC's internal ASN.1 functions.

PR #457, opened 2015-05-07
card-masktech.c: code improvements

vletoux:
As requested in #453


frankmorgner:
nice! always good to see more deletions than additions

PR #459, opened 2015-05-10
iso7816.c: allow file length stored in more than 2 bytes

vletoux:
As indicated in chapter 7.4.3 of iso 7816-4, file length, encoded in tag 0x80 or 0x81 can be stored in more than 2 bytes.
As is, the code allows only 2 bytes and check only for the first tag. If there is 2 tags with 4 bytes then 2 bytes, the file size computation fails.
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/iso7816.c#L362-373

chapter 7 4 3

This patch push the limit to 4 bytes and moreover, if one tag has an incorrect value, it tries the next tag.

### DIFF
--- a/src/libopensc/iso7816.c
+++ b/src/libopensc/iso7816.c
@@ -350,6 +350,7 @@ iso7816_process_fci(struct sc_card *card, struct sc_file *file,
 {
 	struct sc_context *ctx = card->ctx;
 	size_t taglen, len = buflen;
+	int i;
 	const unsigned char *tag = NULL, *p = buf;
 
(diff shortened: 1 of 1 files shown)


frankmorgner:
Could you check if `sc_asn1_decode_integer` can be used for calculating `file->size`?

vletoux:
@frankmorgner sc_asn1_decode_integer seems to be better. I'll upgrade my patch.

frankmorgner:
`sc_asn1_decode_integer` looks for negative values, too. this may be a problem

vletoux:
@frankmorgner : done

PR #460, opened 2015-05-11
Card removal fixes

kemnade-uni:
OpenSC reacts sometimes with crashes or assertion failures when a card is removed during operation when the timing is bad.
This adds some error checks to avoid these crashes.


frankmorgner:
looks OK, thanks

PR #463, opened 2015-05-12
Add MUSCLE support for Tyfone mSD card

drew-tyfone:
Add Tyfone Connected Smart Card ATR to list for MUSCLE support.


src/libopensc/card-muscle.c
```diff
@@ -18,9 +18,7 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
-#if HAVE_CONFIG_H
 #include "config.h"
-#endif
```
frankmorgner: The deletion above is actually not desired.

Please rebase your changes on the current master and apply your changes. This would also remove the merge commits.

src/libopensc/card-muscle.c
```diff
@@ -50,7 +50,7 @@ static struct sc_atr_table muscle_atrs[] = {
 	/* JCOP31 v2.4.1 RF interface */
 	{ "3b:88:80:01:4a:43:4f:50:76:32:34:31:5e", NULL, NULL, SC_CARD_TYPE_MUSCLE_JCOP241, 0, NULL },
 	{ NULL, NULL, NULL, 0, 0, NULL }
-};
+};
```
frankmorgner: whitespace not needed/desired

src/libopensc/card-muscle.c
```diff
@@ -483,6 +483,9 @@ static int muscle_init(sc_card_t *card)
 	if(card->type == SC_CARD_TYPE_MUSCLE_JCOP241) {
 		card->caps |= SC_CARD_CAP_APDU_EXT;
 	}
+	if(card->type == SC_CARD_TYPE_MUSCLE_JCOP242_NO_APDU_EXT) {
+	        /* JCOP v2.4.2 card that doesn't support extended APDUs */
```
frankmorgner: I have a JCOP 2.4.2 that actually *has* support for extended length APDUs.

As far as I can see, the Tyfone card does not customize the ATR. Do you have an other possibility do differentiate between your card and other 2.4.2 cards? If so, use this proberty to distinguish extended/short length support.

frankmorgner:
@viktorTarasov 8b62221 still includes the removal of `HAVE_CONFIG_H`. And, as stated above, there actually are JCOP 2.4.2 cards that have support for extended length.

viktorTarasov:
@frankmorgner 
> 8b62221 still includes the removal of HAVE_CONFIG_H. 

What are you talking about?

> And, as stated above, there actually are JCOP 2.4.2 cards that have support for extended length.

@drew-tyfone  is first who introduced this ATR and, I guess, tested. By default I believe to his results.
If you have different experience -- expose it and do the PR.

frankmorgner:
The ATR @drew-tyfone introduced, is the ATR that's preconfigured from NXP, as far as I know. So it should not be used to identify a specific card. tyfone should rather configure an individual ATR.

Furthermore, I have a JCOP 2.4.1 card that's capable of doing extended length APDUs. Support for extended length started at least with JCOP 2.2.2. You can check with the Security Target for 2.4.1 https://www.commoncriteriaportal.org/files/epfiles/0675b_pdf.pdf

frankmorgner:
@viktorTarasov  For `HAVE_CONFIG_H` have a look at the commented code https://github.com/OpenSC/OpenSC/pull/463#discussion-diff-30208415

viktorTarasov:
Dear Frank, 

 *  what the commented code has to upper commit ?
 *  nothing to add to contribution of @drew-tyfone. Specially for you I repeat it once more:
  * @drew-tyfone was first to introduce this ATR -- his contribution reflects his experience
  * if you have other experience -- do the PR, discuss with the primary author, and so on...

frankmorgner:
I have no problem leaving the merged code as is. However, this will confuse some users, because *every* unpersonalized JCOP 2.4.1 card will show up as being supported by OpenSC (which is not true).

The ATR has obviously not been configured (forgotten, maybe?). @drew-tyfone Could you contact the vendor if this behavior is intended by the vendor?

philipWendland:
> I have no problem leaving the merged code as is. However, this will confuse some users, because every unpersonalized JCOP 2.4.1 card will show up as being supported by OpenSC (which is not true).

I don't think so. 
The muscle driver uses applet selection/AID, not the ATR to match, doesn't it?
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-muscle.c#L88

frankmorgner:
Yes, you're right. I was mistaken on that that part. So it does not pose a problem to OpenSC.

However, I still believe that the card is not fully personalized...

drew-tyfone:
All,

Sorry for the delayed response.  I'm new to this area and had to do some
research and consulting with colleagues on the correct approach for this
given your feedback in this thread.

A bit of background on the extended length aspect ... Tyfone's mSD smart
card solution uses some associated firmware that does not support extended
length APDUs, thus the distinction in the code that has been mentioned
where extended length support is not enabled.

My goal with the submitted change was to add support to the stock OpenSC
package for Tyfone's cards.  I did not intend for the change to affect all
cards using NXP chips with the default ATR response.  Given your comments I
believe the change should be modified to correct this aspect.

Although I don't have a full understand yet of all the ways the ATR string
can be customized, I'm looking at using the historical bytes to identify
the card as a Tyfone model and then restrict the APDU length to
non-extended format.

Does that approach seem reasonable?  Feedback is much appreciated.

Drew


On Thu, May 14, 2015 at 2:32 AM, Frank Morgner <notifications@github.com>
wrote:

> Yes, you're right. I was mistaken on that that part. So it does not pose a
> problem to OpenSC.
>
> However, I still believe that the card is not fully personalized...
>
> —
> <https://github.com/OpenSC/OpenSC/pull/463#issuecomment-101984011>.
>



-- 

*Drew Thomas*

*Tyfone, Inc.*
*Sr. Technical Program Manager*
*Portland | Bangalore | Taipei*
*www.tyfone.com <http://www.tyfone.com/>*
*Mobile: **+1.503.528.6214*
*Fax: **+1.503.228.5539*

frankmorgner:
For the basic structure of the ATR see ISO 7816-3, for some standardized values within the historical bytes (for example to indicate extended length support) see ISO 7816-4. You should conform to the basic structure, but you are relatively free in choosing the historical bytes.

frankmorgner:
Just to clearify: The committed changes are OK and don't break anything within OpenSC. However, you should still fix the ATR.

drew-tyfone:
Pull request submitted to update the ATR historical byte for Tyfone's
MUSCLE card.

On Thu, May 14, 2015 at 10:18 AM, Frank Morgner <notifications@github.com>
wrote:

> Just to clearify: The committed changes are OK and don't break anything
> within OpenSC. However, you should still fix the ATR.
>
> —
> <https://github.com/OpenSC/OpenSC/pull/463#issuecomment-102106776>.
>



-- 

*Drew Thomas*

*Tyfone, Inc.*
*Sr. Technical Program Manager*
*Portland | Bangalore | Taipei*
*www.tyfone.com <http://www.tyfone.com/>*
*Mobile: **+1.503.528.6214*
*Fax: **+1.503.228.5539*

PR #464, opened 2015-05-13
Add --test-fork option to pkcs11-tool

dwmw2:
The PKCS#11 Usage Guide, at least up to v2.40, says that calling
C_Initialize() in the child after fork is "considered to be good
Cryptoki programming practice, since it can prevent the existence of
dangling duplicate resources that were created at the time of the fork()
call."

(It neglects to mention that doing so in the child of a multi-threaded
process is a clear violation of POSIX, mind you. Not to mention being
utterly pointless if all you're going to do in the child is exec something
else anyway.)

Regardless of the sagacity of this recommendation, we need to cope when
it happens. Historically, we've been quite bad at that. Let's add a test
to pkcs11-tool in the hope it'll help...

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -24,6 +24,11 @@
 #include <stdio.h>
 #include <errno.h>
 #include <stdlib.h>
+#ifndef _WIN32
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#endif
 #ifdef ENABLE_OPENSSL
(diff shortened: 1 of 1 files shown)


frankmorgner:
do you know if using Windows' `CreateThread ` is a problem?

dwmw2:
It isn't. Threads are different. The odd semantics here are only for
fork(), which has no equivalent in Windows.

dwmw2:
Thanks. Probably worth having this in 0.15.0 too... along with the fix! :)

PR #471, opened 2015-05-23
sc_pkcs11_card improvements

dengert:
This is name change only fix.

    The variable name card was being used to refer to a struct sc_card or a struct sc_pkcs11_card
    in some of the routines in the pkcs11 directory, while other routines used the variable name
    p11card for struct sc_pkcs11_card. This creates confusing code, such as:
    slot->card->card.

    All definitations of sc_pkcs11_card *card now use p11card as the variable name.


viktorTarasov:
@JenkinsForOpenSC retest this please

Issue #472, opened 2015-05-24
Memory leaks in DNIe module


germanblanco:
Using 0.15.0, I get the following report from valgrind:
==2524== LEAK SUMMARY:
==2524==    definitely lost: 12,982 bytes in 35 blocks
==2524==    indirectly lost: 11,085 bytes in 272 blocks
==2524==      possibly lost: 0 bytes in 0 blocks
==2524==    still reachable: 32,532 bytes in 923 blocks
==2524==         suppressed: 0 bytes in 0 blocks
==2524== 
==2524== For counts of detected and suppressed errors, rerun with: -v
==2524== ERROR SUMMARY: 38 errors from 37 contexts (suppressed: 0 from 0)
This is the command line:
valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes pkcs11-tool --module /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so -l -O -pin APIN
I am already working on it.

frankmorgner:
closed with #474

PR #474, opened 2015-05-26
Fixing part of the memory allocation problems in DNIe module. Issue #472

germanblanco:
Addresses issue #472 in part.
New Valgrind report for the same command:
==9180== LEAK SUMMARY:
==9180==    definitely lost: 9,611 bytes in 51 blocks
==9180==    indirectly lost: 72 bytes in 3 blocks
==9180==      possibly lost: 0 bytes in 0 blocks
==9180==    still reachable: 34,533 bytes in 975 blocks
==9180==         suppressed: 0 bytes in 0 blocks

So it looks quite better, but there is still some way to go.
The patch also removes some unnecessary memory allocation activity.

### DIFF
--- a/src/libopensc/cwa-dnie.c
+++ b/src/libopensc/cwa-dnie.c
@@ -729,10 +729,15 @@ static int dnie_transmit_apdu_internal(sc_card_t * card, sc_apdu_t * apdu)
 		size_t e_txlen = 0;
 		size_t index = 0;
 		sc_apdu_t e_apdu;
-		u8 e_tx[2*SC_MAX_APDU_BUFFER_SIZE];
+		u8 * e_tx = NULL;
 
 		/* envelope needed */
 		sc_log(card->ctx, "envelope tx required: lc:%d", apdu->lc);
+
+		e_tx = calloc(7 + apdu->datalen, sizeof(u8));	/* enveloped data */
+
+		if (!e_tx)
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
 
 		/* copy apdu info into enveloped data */
 		*(e_tx + 0) = apdu->cla;	/* apdu header */
@@ -774,9 +779,19 @@ static int dnie_transmit_apdu_internal(sc_card_t * card, sc_apdu_t * apdu)
 			}
 			/* send data chunk bypassing apdu wrapping */
 			res = sc_transmit_apdu(card, &e_apdu);
-			LOG_TEST_RET(card->ctx, res,
-				     "Error in envelope() send apdu");
+			if (res != SC_SUCCESS) {
+				if (e_tx) {
+					free(e_tx);
+					e_tx = NULL;
+				}
+				sc_log(card->ctx, "Error in envelope() send apdu");
+				LOG_FUNC_RETURN(card->ctx, res);
+			}
 		}		/* for */
+		if (e_tx) {
+			free(e_tx);
+			e_tx = NULL;
+		}
 		/* last apdu sent contains response to enveloped cmd */
 		apdu->resp = calloc(1, MAX_RESP_BUFFER_SIZE);
 		if (apdu->resp == NULL)
--- a/src/libopensc/cwa-dnie.c
+++ b/src/libopensc/cwa-dnie.c
@@ -265,6 +265,8 @@ int dnie_read_file(sc_card_t * card,
 	res = SC_SUCCESS;
 	goto dnie_read_file_end;
  dnie_read_file_err:
+	if (data)
+		free(data);
 	if (*file) {
 		sc_file_free(*file);
 		*file = NULL;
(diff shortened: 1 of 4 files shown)


src/libopensc/cwa-dnie.c
```diff
@@ -711,8 +711,10 @@ static int dnie_transmit_apdu_internal(sc_card_t * card, sc_apdu_t * apdu)
 			if (tmp == SC_APDU_CASE_3_SHORT)
 				apdu->cse = SC_APDU_CASE_4_SHORT;
 			if (apdu->resplen == 0) {	/* no response buffer: create */
-				apdu->resp = buf;
-				apdu->resplen = 2048;
+				apdu->resp = calloc(1, 2*SC_MAX_APDU_BUFFER_SIZE);
```
frankmorgner: Are you sure that you want to reduce the length here? A raw RSA cryptogram won't fit anymore in a response with about 500 bytes.

frankmorgner:
Looks good, thank you.

You could use smaller commits next time, for example one commit per file or one commit per type of fix.

frankmorgner:
There are some warnings from in the travis-ci log that you might want to review:
```
cwa-dnie.c:243:6: warning: variable 'data' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]
        if (fsize <= 0) {
            ^~~~~~~~~~
cwa-dnie.c:268:6: note: uninitialized use occurs here
        if (data)
            ^~~~
cwa-dnie.c:243:2: note: remove the 'if' if its condition is always false
        if (fsize <= 0) {
        ^~~~~~~~~~~~~~~~~
cwa-dnie.c:226:6: warning: variable 'data' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]
        if (res != SC_SUCCESS) {
            ^~~~~~~~~~~~~~~~~
cwa-dnie.c:268:6: note: uninitialized use occurs here
        if (data)
            ^~~~
cwa-dnie.c:226:2: note: remove the 'if' if its condition is always false
        if (res != SC_SUCCESS) {
        ^~~~~~~~~~~~~~~~~~~~~~~~
cwa-dnie.c:212:10: note: initialize the variable 'data' to silence this warning
        u8 *data;
                ^
                 = NULL
2 warnings generated.
  CC       user-interface.lo
cwa-dnie.c:243:6: warning: variable 'data' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]
        if (fsize <= 0) {
            ^~~~~~~~~~
cwa-dnie.c:268:6: note: uninitialized use occurs here
        if (data)
            ^~~~
cwa-dnie.c:243:2: note: remove the 'if' if its condition is always false
        if (fsize <= 0) {
        ^~~~~~~~~~~~~~~~~
cwa-dnie.c:226:6: warning: variable 'data' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]
        if (res != SC_SUCCESS) {
            ^~~~~~~~~~~~~~~~~
cwa-dnie.c:268:6: note: uninitialized use occurs here
        if (data)
            ^~~~
cwa-dnie.c:226:2: note: remove the 'if' if its condition is always false
        if (res != SC_SUCCESS) {
        ^~~~~~~~~~~~~~~~~~~~~~~~
cwa-dnie.c:212:10: note: initialize the variable 'data' to silence this warning
        u8 *data;
                ^
                 = NULL
```

src/libopensc/cwa-dnie.c
germanblanco: No I am not sure, and I should not have changed any functionality with this pull request. I should have stick to memory allocation things and avoid mistakes. I will change it back to 2048 if that number is right. I would prefer to use a constant with a meaningful name here, but obviously the correct constant. Thanks a lot for your comment!

germanblanco:
Could you please help me with these two questions?
Should I correct things on the same branch with additional commits, or would it be better to create a new branch so that only one commit goes in the history?
And even though I have started already looking for information myself, any pointers (to a howto page or similar) on how I can invoke this "travis-ci" tool myself would be very appreciated.

src/libopensc/cwa-dnie.c
frankmorgner: You could use an internal define in `cwa-dnie.c`, for example `DNIE_MAX_INTERNAL_TRANSMIT_BYTES`...

frankmorgner:
Click "show all checks" in this pull request and you will get the link to the log of travis-ci. You will get the same compiler warnings with clang on OS X or Linux.

germanblanco:
OK, thank you! It seems clean now. Please tell me if anything else is required.

miguel-cv:
Working well and impressive result XD:

valgrind --leak-check=full --show-leak-kinds=all pkcs11-tool --module /usr/lib/opensc-pkcs11.so -O -l --pin XXX

=25206== LEAK SUMMARY:
==25206==    definitely lost: 757 bytes in 7 blocks
==25206==    indirectly lost: 72 bytes in 3 blocks
==25206==      possibly lost: 0 bytes in 0 blocks
==25206==    still reachable: 34,557 bytes in 976 blocks
==25206==         suppressed: 0 bytes in 0 blocks

germanblanco:
Dear committers,
I would appreciate action or feedback on this one, since we (at least Miguel and me) intend to keep working on the same code and it helps to have the baseline already in. Could you please take a look?

dengert:
All the changes are in the DNIe only code, and it looks it OK to
    me.  +1
    
    On 6/29/2015 3:25 AM, Germán Blanco
      wrote:
    
    
      Dear committers,
        I would appreciate action or feedback on this one, since we (at
        least Miguel and me) intend to keep working on the same code and
        it helps to have the baseline already in. Could you please take
        a look?
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

src/libopensc/cwa-dnie.c
```diff
@@ -724,17 +728,12 @@ static int dnie_transmit_apdu_internal(sc_card_t * card, sc_apdu_t * apdu)
...
-		sc_apdu_t *e_apdu = NULL;
-		u8 *e_tx = NULL;
+		sc_apdu_t e_apdu;
+		u8 e_tx[2*SC_MAX_APDU_BUFFER_SIZE];
 
 		/* envelope needed */
 		sc_log(card->ctx, "envelope tx required: lc:%d", apdu->lc);
 
-		e_apdu = calloc(1, sizeof(sc_apdu_t));	/* enveloped apdu */
-		e_tx = calloc(7 + apdu->datalen, sizeof(u8));	/* enveloped data */
```
frankmorgner: With your new changes the size of `e_tx`is now fixed to `2*SC_MAX_APDU_BUFFER_SIZE` instead of being calculated from `apdu->datalen`. This seems less flexible than before. Also, there is no local length check of the buffer which might lead to an out-of-bounds-write in the future if `dnie_transmit_apdu_internal`is called in a different situation (e.g. a new version of the card might have a different `max_send_size` which potentially grows the chunk size of the enveloped commands).

germanblanco: I will change it back. The intention was to avoid unnecessary dynamic memory allocations, but I clearly see the benefit now.

frankmorgner: You're reasoning is perfectly fine. I just wanted to let you know about what could be a problem in the future. Maybe you want to choose an internal define for fixing the maximum number of bytes?

germanblanco: The main goal of this fix is to avoid the memory leaks. I consider that changing from dynamic to static allocation is a minor performance improvement. Once that you have highlighted the potential problems of this change, I prefer to stick to dynamic allocation for e_tx and free the memory in all the correct places, which doesn't seem that complex in this case. It seems more future proof, and performance is a minor problem now. When there are no memory leaks at all and there are no functional problems, maybe we have time to make additional improvements. So, unless you object, I will make these changes. Thank you again for spotting this.

frankmorgner:
thanks

PR #476, opened 2015-06-01
Empty user-interface.c if there is no SSL support. Related with issue #362

germanblanco:
The module with a default user interface for DNIe should not be compiled if there is no SSL support (just as other modules of DNIe).


frankmorgner:
thanks

PR #483, opened 2015-06-15
Update ATR and ATR mask for MaskTech smartcards

adminmt:
added mask for ATR


frankmorgner:
Could you squash the commits into a single commit and give it a more meaningfull commit message, please?

adminmt:
Hi frankmorgner, I squashed the commits into a single one and changed the commit message, is there something else I should change? Thanks!

adminmt:
Thank you victor for the note on ATR Mask.

PR #485, opened 2015-06-17
Restore compability on os x

carlhoerberg:
Fallback to
/System/Library/Frameworks/PCSC.framework/Versions/Current/Headers when
there are no OS X SDKs installed. Fixes bugs introduced in #308


frankmorgner:
thanks

PR #494, opened 2015-07-22
After a fork do not release resources shared with parent

frankmorgner:
This PR implements the second option proposed in https://github.com/OpenSC/OpenSC/pull/493#issuecomment-121917436 . I hope my approach is a bit cleaner and more complete than what is proposed in #493. Everything works as expected with a CardOS card and `pkcs11-tool --test-fork --login --test`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,8 +39,8 @@ OPENSC_VS_FF_PRODUCT_NAME="VS_FF_PRODUCT_NAME"
 #   (Code changed:                      REVISION++)
 #   (Oldest interface removed:          OLDEST++)
 #   (Interfaces added:                  CURRENT++, REVISION=0)
-OPENSC_LT_CURRENT="3"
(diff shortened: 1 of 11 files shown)


frankmorgner:
@nmav what do you think?

nmav:
I haven't checked yet. I just returned from holidays. I'll try to find out some time and check it.

nmav:
That patch looks simpler and cleaner than 493. As of memory leaks these remain, but as I said in 493 they should be treated as a separate issue.

frankmorgner:
@nmav thanks for the feedback.

Could you open a separate issue with the missing memory leaks? thank you!

nmav:
On second view, this patch doesn't handle the case where after a fork, C_Finalize() is called (without C_Initialize()). If that happens the same error will occur. I'll open an issue for that.

Issue #496, opened 2015-07-26
src/test/regression broken - change "--erase" to "--erase-card" to fix it


tolonuga:
it seems pkcs15-init --erase was renamed to --erase-card (and --erase-application added), but noone updated src/tests/regressions/functions file to use the new option name. Please fix.

tolonuga:
Also --slot-label is now --token-label, please fix as well.

frankmorgner:
closed with #497

PR #497, opened 2015-07-26
Fix regression tests: some parameters got renamed (#496)

tolonuga:
This fixes two renamed parameters in the regression tests plus invalid key size in test crypt0007.


frankmorgner:
thanks

PR #500, opened 2015-07-28
honour PC/SC pt 10 dwMaxAPDUDataSize

### LINKED ISSUE #306: Recognize maximum send/receive size of Readers
... just a reminder. We stumbled upon this multiple time, for example when a card supports extended length APDUs, but the reader doesn't.

Solutions would be:
- White list of readers
- evaluation of the PC/SC property for send/receive size (use GET_TLV_PROPERTIES and parse PCSCv2_PART10_PROPERTY_dwMaxAPDUDataSize)

frankmorgner:
closes #306

### DIFF
--- a/src/libopensc/card.c
+++ b/src/libopensc/card.c
@@ -254,12 +254,12 @@ int sc_connect_card(sc_reader_t *reader, sc_card_t **card_out)
 	 *  Note that zero means no limitations at all.
 	 */
 	if ((card->max_recv_size == 0) ||
-			((reader->driver->max_recv_size != 0) && (reader->driver->max_recv_size < card->max_recv_size)))
-		card->max_recv_size = reader->driver->max_recv_size;
+			((reader->max_recv_size != 0) && (reader->max_recv_size < card->max_recv_size)))
+		card->max_recv_size = reader->max_recv_size;
 
 	if ((card->max_send_size == 0) ||
-			((reader->driver->max_send_size != 0) && (reader->driver->max_send_size < card->max_send_size)))
-		card->max_send_size = reader->driver->max_send_size;
+			((reader->max_send_size != 0) && (reader->max_send_size < card->max_send_size)))
+		card->max_send_size = reader->max_send_size;
 
 	sc_log(ctx, "card info name:'%s', type:%i, flags:0x%X, max_send/recv_size:%i/%i",
 		card->name, card->type, card->flags, card->max_send_size, card->max_recv_size);
--- a/src/libopensc/ctx.c
+++ b/src/libopensc/ctx.c
@@ -289,15 +289,28 @@ static void load_reader_driver_options(sc_context_t *ctx)
 {
 	struct sc_reader_driver *driver = ctx->reader_driver;
 	scconf_block *conf_block = NULL;
-
-	driver->max_send_size = 0;
-	driver->max_recv_size = 0;
+	sc_reader_t *reader;
+	int max_send_size;
+	int max_recv_size;
 
 	conf_block = sc_get_conf_block(ctx, "reader_driver", driver->short_name, 1);
 
 	if (conf_block != NULL) {
-		driver->max_send_size = scconf_get_int(conf_block, "max_send_size", driver->max_send_size);
-		driver->max_recv_size = scconf_get_int(conf_block, "max_recv_size", driver->max_recv_size);
+		max_send_size = scconf_get_int(conf_block, "max_send_size", -1);
+		max_recv_size = scconf_get_int(conf_block, "max_recv_size", -1);
+		if (max_send_size >= 0 || max_recv_size >= 0) {
+			if (list_iterator_start(&ctx->readers)) {
+				reader = list_iterator_next(&ctx->readers);
+				while (reader) {
+					if (max_send_size >= 0)
+						reader->max_send_size = max_send_size;
+					if (max_recv_size >= 0)
+						reader->max_recv_size = max_recv_size;
+					reader = list_iterator_next(&ctx->readers);
+				}
+				list_iterator_stop(&ctx->readers);
+			}
+		}
 	}
 }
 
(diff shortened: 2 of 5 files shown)


nmav:
This commit breaks my smartcard-hsm.

Before:
$ ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so -O 
Using slot 1 with a present token (0x1)
Public Key Object; RSA 2048 bits
  label:      SSH test key
  ID:         3951e13da94cb32209193598ab1ef3e2c5f6f5a7
  Usage:      none
Public Key Object; RSA 2048 bits
  label:      GnuTLS-Test
  ID:         2707b6d7ac8f0216c0c35bc17cf4b8e87f2f5cdf
  Usage:      none

After:
./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so -O 
error: PKCS11 function C_GetSlotInfo failed: rv = CKR_DATA_INVALID (0x20)

Aborting.

dengert:
opensc debug output would be helpful.

On 8/4/2015 4:10 AM, Nikos Mavrogiannopoulos wrote:
> This commit breaks my smartcard-hsm.
>
> Before:
> $ ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so -O
> Using slot 1 with a present token (0x1)
> Public Key Object; RSA 2048 bits
> label: SSH test key
> ID: 3951e13da94cb32209193598ab1ef3e2c5f6f5a7
> Usage: none
> Public Key Object; RSA 2048 bits
> label: GnuTLS-Test
> ID: 2707b6d7ac8f0216c0c35bc17cf4b8e87f2f5cdf
> Usage: none
>
> After:
> ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so -O
> error: PKCS11 function C_GetSlotInfo failed: rv = CKR_DATA_INVALID (0x20)
>
> Aborting.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/500#issuecomment-127534791>.
>

-- 

nmav:
http://paste.fedoraproject.org/251368/92097143/

frankmorgner:
@nmav @CardContact this is a bug in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-sc-hsm.c#L254 Accepting `count > 256` but fixing the APDU to `SHORT` is bad. Simply remove `SHORT` and let the lower layers decide the APDU format.

Could you make a PR to fix this behavior?

CardContact:
Hmmm, don't quite understand how this commit breaks the code.

Anyway I draft a PR for it.

frankmorgner:
This is what happens when `recv_size` is initialized with `65536` (with #500) instead of `0` (without #500):
```
0x7f9a73223700 14:41:33.641 [opensc-pkcs11] apdu.c:352:sc_check_apdu: Invalid Case 4 short APDU:
cse=04 cla=00 ins=b1 p1=00 p2=00 lc=4 le=512
resp=0x7ffcebf2d180 resplen=512 data=0x7ffcebf2d100 datalen=4
0x7f9a73223700 14:41:33.641 [opensc-pkcs11] card-sc-hsm.c:263:sc_hsm_read_binary: APDU transmit failed: -1300 (Invalid arguments)
```

CardContact:
The only explanation is, that Nikos reader does actually reports dwMaxAPDUDataSize and thus card->max_recv_size changes from 0 to 65535 causing sc_read_binary to changes to extended length APDUs.

All my readers don't report dwMaxAPDUDataSize, so the code work fine here.

However I can provoke the issue by setting card->max_recv_size to something other than 0.

Which gets me to the question of what card->max_recv_size = 0 actually means: Does it mean "the card supports arbitrary length in case 2 / case 4 APDUs ?

IMHO this PR will also break other card / reader combinations, as the code suddenly switches from short to extended APDUs, just because the reader reports dwMaxAPDUDataSize.

frankmorgner:
If there is a limitation for the card, its driver should set `card->max_recv_size` accordingly, see https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L254-L263. This may have not always been done by every developer, but I guess we'll find out soon...

CardContact:
We should have at least a line in the log, reporting the value of dwMaxAPDUDataSize received from the reader.

dengert:
Many cards don't have active developers, so "we'll find out soon" is really we will find out after the next release and some user's card fails.
We need to be careful when changing code is core modules that can affect all cards.

Are there cards that can do extended APUDs, but have a limit < 65K?
Should there be a card_max_send_extended_size and card_max_recv_extended_size?


On 8/4/2015 10:37 AM, Frank Morgner wrote:
> If there is a limitation for the card, its driver should set |card->max_recv_size| accordingly, see https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L254-L263. This may have not
> always been done by every developer, but I guess we'll find out soon...



>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/500#issuecomment-127652928>.
>

-- 

frankmorgner:
@dengert no, currently max_recv_size/max_send_size is used for cards that support short length APDUs only.

philipWendland:
@dengert
> Are there cards that can do extended APUDs, but have a limit < 65K? Should there be a card_max_send_extended_size and card_max_recv_extended_size?

Might be worth a look.

I do believe most of the cards have such a limitation. As far as I can tell, *all* Java Cards do:
" the applet may receive and send up to 32767 bytes of APDU data"
http://www.win.tue.nl/pinpasjc/docs/apis/jc222/

This is due to the Java limitation of only having signed `short` and Java Card being based on short. (Support for `int` is an optional extension that does not need to be supported)

dengert:
@frankmorgner  There appears to be a number of issues with commit.  Can you please look at this mod closer again?

frankmorgner:
@dengert it depends how you see it... The actual commit includes a feature that works fine or itself. However, it revealed two bugs (#513 and #515) that were always present but have never been triggered...

Issue #461, opened 2015-05-12
Unsuported Aladdin eToken PRO USB 72k Java


ieugen:
Hello, 

I have a smart card for digital signatures bought from one of our national providers and I wish to use it in Debian GNU/Linux 8. Unfortunately I haven't been able to do so. I've spent the last few days. I need it because I wish to submit my tax forms online, and I'm required to sign them with an approved device. 

The card is sold by DigiSign Romania. 
[Card technical description (Romanian) ](https://www.digisign.ro/documente/fise-tehnice/)
[Windows drivers for card](https://www.digisign.ro/download/)

In hopes of better support in the future, I'm giving some information bellow:

**lsusb**
```
Bus 002 Device 002: ID 8087:0024 Intel Corp. Integrated Rate Matching Hub
Bus 002 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 001 Device 009: ID 12d1:1569 Huawei Technologies Co., Ltd. 
Bus 001 Device 005: ID 1bcf:2c08 Sunplus Innovation Technology Inc. 
Bus 001 Device 004: ID 8087:07da Intel Corp. 
Bus 001 Device 003: ID 08ff:168f AuthenTec, Inc. AES1660 Fingerprint Sensor
Bus 001 Device 002: ID 8087:0024 Intel Corp. Integrated Rate Matching Hub
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 004 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 003 Device 022: ID 18d1:4ee3 Google Inc. Nexus 4 (tether)
Bus 003 Device 027: ID 0529:0620 Aladdin Knowledge Systems Token JC
Bus 003 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
```
**opensc-tool -l**
```
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Aladdin eToken PRO USB 72K Java [Main Interface] 00 00
```

**opensc-tool -a**
```
Using reader with a card: Aladdin eToken PRO USB 72K Java [Main Interface] 00 00
3b:d5:18:00:81:31:fe:7d:80:73:c8:21:10:f4
```
**opensc-tool -n**
```
Using reader with a card: Aladdin eToken PRO USB 72K Java [Main Interface] 00 00
Unsupported card
```
**opensc-tool --info**
```
[...]

PR #501, opened 2015-07-28
Aladdin work in progress

ieugen:
Details in #461


frankmorgner:
Before actually implementing stuff, you should read the docs. https://www.digisign.ro//uploads/Fisa-tehnica-eToken.pdf says something about
```
Modele (in functie de memorie)    32K, 64K (Siemens CardOS), 72K (Java)
```
Though I don't speak Romanian, this sounds like you should be able to reuse the CardOS driver, which is already part of OpenSC. You could try adding your ATR to the CardOS driver...

ieugen:
Thanks.
It means that it might be CardOS or Java. How can I make sure? The output (in #461 ) mentions 72k Java. And what are my options in that case?  

p.s. I've installed more packages and I can build OpenSC right now :D

frankmorgner:
sorry, your documents are not very detailed. you're on your own...

frankmorgner:
please re-open if you have some more progress

PR #502, opened 2015-07-29
openpgp: match application, not ATR

### LINKED ISSUE #391: Support emulated OpenPGP applet in Android HCE
I created a simple [Host Card Emulation (HCE) App for Android](https://frankmorgner.github.io/vsmartcard/ACardEmulator/README.html) which integrates the [OpenPGP applet](https://github.com/Yubico/ykneo-openpgp) of @Yubico.

Unfortunately I can't use the applet out of the box since the ``card-openpgp.c`` recognizes a card by its ATR rather than the AID. I am not familiar how to determine the OpenPGP version (1 or 2) so I can't really fix this at the moment. @hongquan do you know how this can be done in a more flexible way?

frankmorgner:
fixes #391

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -283,7 +283,8 @@ struct pgp_priv_data {
 	sc_security_env_t	sec_env;
 };
 
-/* ABI: check if card's ATR matches one of driver's */
+/* ABI: check if card's ATR matches one of driver's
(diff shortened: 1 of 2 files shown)


frankmorgner:
Though the problem in #391 gets fixed with this PR, Yuboco's OpenPGP applet is not read out correctly. There are still some other vender specific/hard coded things that need to be generalized...

PR #511, opened 2015-08-04
Handle properly C_Finalize() when called after a fork

### LINKED ISSUE #510: After a fork do not release resources shared with parent when C_Finalize is called
The issue of C_Login returning CKR_GENERAL_ERROR after a fork was resolved by setting a special flag when C_Initialize was called in #494. However, if after a fork, C_Finalize is called the same issue occurs.

For a reproducer download:
https://bugzilla.redhat.com/attachment.cgi?id=1022342

and remove the C_Initialize call after the fork.

nmav:
Resolves #510


frankmorgner:
closing this without any new input, see #510

PR #513, opened 2015-08-04
Fixed wrong APDU case declaration detected after PR #500

CardContact:

PR #515, opened 2015-08-04
don't always overwrite max_send_size/max_recv_size

frankmorgner:
If the reader announces extended length support, but the card driver leaves `max_send_size`/`max_recv_size` at `0`, `max_send_size`/`max_recv_size` previously would have been overwritten with the reader's size though the card might not have set `SC_CARD_CAP_APDU_EXT`. This commit fixes this behavior.

Additionally card->max_send_size/max_recv_size is always initialized to a value different from `0` after the card initialization. This removes the need to check for this special value in all subsequent calls.

Yes, this is the most intrusive way of fixing this issue. The alternative would be to simply change ´card.c´ to something like
```
	/*  Override card limitations with reader limitations.
	 *  Note that zero means no limitations at all.
	 */
	if ((card->max_recv_size == 0 && card->caps & SC_CARD_CAP_APDU_EXT && reader->max_recv_size < 0xffff+1)
			|| (card->max_recv_size == 0 && !card->caps & SC_CARD_CAP_APDU_EXT && reader->max_recv_size < 0xff+1)
			|| ((reader->max_recv_size != 0) && (reader->max_recv_size < card->max_recv_size)))
		card->max_recv_size = reader->max_recv_size;

	if ((card->max_send_size== 0 && card->caps & SC_CARD_CAP_APDU_EXT && reader->max_send_size < 0xffff)
			|| (card->max_recv_size == 0 && !card->caps & SC_CARD_CAP_APDU_EXT && reader->max_send_size < 0xff)
			|| ((reader->max_send_size != 0) && (reader->max_send_size < card->max_send_size)))
		card->max_send_size = reader->max_send_size;

```
while leaving everything else as is. However, I consider the first approach better because it reduces the overall complexity.

Feedback welcome!

### DIFF
--- a/src/libopensc/apdu.c
+++ b/src/libopensc/apdu.c
@@ -581,7 +581,7 @@ int sc_transmit_apdu(sc_card_t *card, sc_apdu_t *apdu)
 		 * bytes using command chaining */
 		size_t    len  = apdu->datalen;
 		const u8  *buf = apdu->data;
-		size_t    max_send_size = card->max_send_size > 0 ? card->max_send_size : 255;
+		size_t    max_send_size = card->max_send_size;
 
 		while (len != 0) {
 			size_t    plen;
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -165,6 +165,9 @@ static int cardos_have_2048bit_package(sc_card_t *card)
 static int cardos_init(sc_card_t *card)
 {
 	unsigned long	flags, rsa_2048 = 0;
+	size_t data_field_length;
+	sc_apdu_t apdu;
+	u8 rbuf[2];
 
 	card->name = "CardOS M4";
 	card->cla = 0x00;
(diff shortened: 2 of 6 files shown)


philipWendland:
This fix is necessary for the IsoApplet. Thank you.

frankmorgner:
Do you have an opinion about the two ways to solve it? @viktorTarasov @dengert @martinpaljak

dengert:
The OpenSC APDU code has always been very complicated and touchy. It also depends on the reader and card support of extended APDUs
and also depends on the if the caller requested SHORT or extended. It is possible that a card may support extended APDUs
for some but not all commands,(I don't know if there are any)  It is also complicated by command chaining, which might be needed with extended APDUs
if the reader size is to limited.

#513 appears to address the problem of sc-hsm  using a recvbuf that is two small.
It looks like #513 in the read binary, was always doing SHORT APDUs, but was changed to lets APDU code decide which to use.
Are there other drivers that may have similar issues?

Have you looked at all the drivers that may set SC_CARD_CAP_APDU_EXT? (grep shows about 12)

in #515, "if (card->caps & SC_CARD_CAP_APDU_EXT) {" is done during sc_card_connect.
Are there any cards that set SC_CARD_CAP_APDU_EXT after calling sc_card_connect,
If so card->max_send_size = 65535; and card->max_recv_size = 65536; may not be set.

For the PIV card application I am interested in, the NIST specifications do not talk about extended APDUs, but do require chaining.
Chaining allows for the use of short APDU in all cases. (And could in allow for objects larger then 65K in the future.)
But some of the test cards I have appear be able to do extended ADPUs as indicated in the ATR historical bytes.
I have not looked at the code to see if using extended APDUs would be break anything.


On 8/7/2015 3:38 AM, Frank Morgner wrote:
> Do you have an opinion about the two ways to solve it? @viktorTarasov <https://github.com/viktorTarasov> @dengert <https://github.com/dengert> @martinpaljak <https://github.com/martinpaljak>
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/515#issuecomment-128642027>.
>

-- 

dengert:
It looks like this patch is needed, and shoud address my main concerns abount old drivers that set the sizes less then <255, or don't set them at all, or don't set SC_CARD_CAP_APDU_EXT. 

Disregard my  comments abount  "Are there any cards that set SC_CARD_CAP_APDU_EXT after calling sc_card_connect" Looking at code, sc_connect_card calls the card's init function that should set sizes before the above code is executed.

PR #487, opened 2015-06-22
First fix for bug #480 (opensc sc_reset leaves unreleased locks with pcsc-lite)

rickyepoderi:
Finally I decided to follow with the first option (not adding the lock again if pcsc-lite is used, ifdef _WIN32). This is the first solution that was thought for bug #480, the benefits for this solution are the following:

* The lock is just maintained over the card without doing any other operation.
* The header file opensc.h does not contain sc_reconnect so just sc_reset is exported. This way the sc_reset behavior is exactly the same in windows or pcsc-lite.

PR #516, opened 2015-08-04
Only re-lock for Windows and Apple

frankmorgner:
See discussion in #487. Apple (beginning with OS X 10.10) behaves like Windows.

If someone tells me how to check for >= OS X 10.10, I'm happy to refine the commit.

### DIFF
--- a/MacOSX/build-package.in
+++ b/MacOSX/build-package.in
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Building the installer is only tested and supported on 10.9+ with Xcode 6.0.1
-# Built package targets 10.9+
+# Built package targets 10.10
 # Building should also work on older versions with older revisions or slight changes, YMMV
 
 # You need to have  the following from homebrew or macports or fink:
@@ -15,7 +15,7 @@ SDKS_PATH="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs"
 SDK_PATH="${SDK_PATH:-$SDKS_PATH/$(ls -1 ${SDKS_PATH} | sort -n -k2 -t. -r | head -1)}"
 
 # Set SDK path
-export CFLAGS="-isysroot $SDK_PATH -arch i386 -arch x86_64 -mmacosx-version-min=10.7"
+export CFLAGS="-isysroot $SDK_PATH -arch i386 -arch x86_64 -mmacosx-version-min=10.10"
 
 # OpenSSL is deprecated on OSX since 10.7 and that generates lots of
 # "false positive" warnings and there is no alternative option.
--- a/configure.ac
+++ b/configure.ac
@@ -576,6 +576,7 @@ if test "${enable_pcsc}" = "yes"; then
 	CFLAGS="${CFLAGS} ${PCSC_CFLAGS}"
 	# We must cope with mingw32 that does not have winscard.h mingw64 has it.
 	AC_CHECK_HEADERS([winscard.h],,[test "${WIN32}" != "yes" && AC_MSG_ERROR([winscard.h is required for pcsc])])
+	AC_CHECK_HEADERS([pcsclite.h])
 	CFLAGS="${saved_CFLAGS}"
 
 	if test "${with_pcsc_provider}" = "detect"; then
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -41,6 +41,12 @@
 
 #include "pace.h"
 
+#ifdef HAVE_PCSCLITE_H
+#if !defined (__MAC_OS_X_VERSION_MIN_REQUIRED) || __MAC_OS_X_VERSION_MIN_REQUIRED < 101000
+#define HAVE_PCSCLITE 1
+#endif
+#endif
+
 /* Logging */
 #define PCSC_TRACE(reader, desc, rv) do { sc_log(reader->ctx, "%s:" desc ": 0x%08lx\n", reader->name, rv); } while (0)
 #define PCSC_LOG(ctx, desc, rv) do { sc_log(ctx, desc ": 0x%08lx\n", rv); } while (0)
@@ -431,7 +437,7 @@ static int pcsc_reconnect(sc_reader_t * reader, DWORD action)
 	if (check_forced_protocol(reader->ctx, &reader->atr, &tmp))
 		protocol = tmp;
 
-#ifndef HAVE_PCSCLITE_H
+#ifndef HAVE_PCSCLITE
 	/* reconnect unlocks transaction everywhere but in PCSC-lite */
 	priv->locked = 0;
 #endif
@@ -591,7 +597,7 @@ static int pcsc_release(sc_reader_t *reader)
 static int pcsc_reset(sc_reader_t *reader, int do_cold_reset)
 {
 	int r;
-#ifndef HAVE_PCSCLITE_H
+#ifndef HAVE_PCSCLITE
 	struct pcsc_private_data *priv = GET_PRIV_DATA(reader);
 	int old_locked = priv->locked;
 #endif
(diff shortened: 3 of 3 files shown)


frankmorgner:
We wait until https://alioth.debian.org/tracker/index.php?func=detail&aid=315083&group_id=30105&atid=410088 is resolved

martinpaljak:
For OSX versions: https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/cross_development/Using/using.html

frankmorgner:
as we found out, the re-locking-problem only occures in PCSC-Lite (and won't be fixed any time soon, see #487), I adapted the fixes within OpenSC accordingly. @rickyepoderi could you check whether the problem is fixed with this PR?

rickyepoderi:
Hi @frankmorgner,

I have applied your patch and it works as expected in my debian machine.  First the resulting config.h has the define as expected:

    $ grep HAVE_PCSCLITE_H config.h
    #define HAVE_PCSCLITE_H 1

Using the little program I did the second execution starts exactly after the lock release of the first execution (and not after the disconnect as before):

    | start: Mon Oct  5 11:30:23 2015                     |                                                     |
    | context created: Mon Oct  5 11:30:23 2015           |                                                     |
    | connect and lock adquired: Mon Oct  5 11:30:23 2015 |                                                     |
    | reset done: Mon Oct  5 11:30:33 2015                |                                                     |
    |                                                     | start: Mon Oct  5 11:30:35 2015                     |
    |                                                     |                                                     |
    | lock released: Mon Oct  5 11:30:43 2015             |                                                     |
    |                                                     | context created: Mon Oct  5 11:30:43 2015           |
    |                                                     | connect and lock adquired: Mon Oct  5 11:30:43 2015 |
    |                                                     | reset done: Mon Oct  5 11:30:54 2015                |
    | disconnected: Mon Oct  5 11:31:04 2015              |                                                     |
    |                                                     | lock released: Mon Oct  5 11:31:04 2015             |
    |                                                     | disconnected: Mon Oct  5 11:31:14 2015              |

I have also tested with DNIe (it uses the reset method) and you can use different programs. For example you can log in the card using firefox and then execute dnie-tool command (previously the firefox login had locked the card forever until you close firefox).

Finally I have tested the DNIe card with the Spanish government test page:
https://av-dnie.cert.fnmt.es/compruebacert/compruebacert

It worked again, you can access the page, then start the applet (a different java program that previously got stuck because of this issue) and sign the test sample. There are a subsequent problem but it is not related to the lock issue.

If you need some more testing do not hesitate in contacting me.

frankmorgner:
I tested your program with this PR on OS X 10.10.5 and got the expected result: The card is locked and only when it is reset the concurrent process is able to lock the card:
```
                                                                  TIMESTAMP start: Wed Oct  7 13:04:16 2015                                                                       
                                                                  TIMESTAMP context created: Wed Oct  7 13:04:16 2015 
                                                                  Using reader with a card: SCL011 Contactless Reader 
                                                                  Connecting to card in reader SCL011 Contactless Reader... 
                                                                  Using card driver Default driver for unknown cards. 
                                                                  TIMESTAMP connect and lock adquired: Wed Oct  7 13:04:16 2015 
TIMESTAMP start: Wed Oct  7 13:04:23 2015 
                                                                  3b:89:80:01:4a:43:4f:50:32:34:32:52:33:4b 
                                                                  TIMESTAMP reset done: Wed Oct  7 13:04:26 2015 
TIMESTAMP context created: Wed Oct  7 13:04:26 2015 
Using reader with a card: SCL011 Contactless Reader 
Connecting to card in reader SCL011 Contactless Reader... 
Using card driver Default driver for unknown cards. 
TIMESTAMP connect and lock adquired: Wed Oct  7 13:04:26 2015 
3b:89:80:01:4a:43:4f:50:32:34:32:52:33:4b 
TIMESTAMP reset done: Wed Oct  7 13:04:36 2015 
                                                                  TIMESTAMP lock released: Wed Oct  7 13:04:36 2015 
                                                                  TIMESTAMP disconnected: Wed Oct  7 13:04:46 2015 
Error in sc_unlock: Unknown error (-1900)
```
The last error is due to a `SCARD_W_RESET_CARD`, which is OK (could be converted to `SC_SUCCESS` though).

frankmorgner:
By the way, MacOSX10.10.sdk ships with headers for PCSC-Lite.

frankmorgner:
I'll upgrade to El Capitan in the next days to test without PCSC-Lite

rickyepoderi:
@frankmorgner For what I understand you the HAVE_PCSCLITE_H defined in 10.10 (OSX), so you are applying the patch, don't you? The idea is that the patch should not be applied in 10.10 (and I think you're getting the error because the patch is incorrectly being used, it thinks the lock is in place when it is not). If you test in OS X without the patch you won't get that error.

Sorry if I misunderstand you (this issue is very confusing).

frankmorgner:
Yeah, I think you're right. I was confused that Apple still comes with `pcsclite.h`. I'll try again tomorrow...

frankmorgner:
with the patch above and `CFLAGS=-mmacosx-version-min=10.10` it works without an error on 10.10.5

rickyepoderi:
I've been thinking and maybe we can check the PCSCLITE_VERSION_NUMBER that has the pcsclite.h with an AC_TRY_RUN or similar. This way if pcsc-lite fixes the bug we could check the version of pcsc-lite. I have checked that OS X 10.8 and below has "1.4.0" in this define:

http://smartcardservices.macosforge.org/trac/browser/releases/Apple/OSX-10.8.0/SmartCardServices-55105/src/PCSC/pcsclite.h

@frankmorgner What about 10.10? Is there a PCSCLITE_VERSION_NUMBER in pcsclite.h of 10.10? What string is assigned?

This is just another idea, I think your last patch is enough for the moment.

rickyepoderi:
This is a simple AC_TRY_RUN for the configure.ac that works in linux (not very tested, just an example). I don't know what happens in OS X 10.10 because I don't know what the pcsclite.h contains...

```
*** configure.ac.orig	2015-05-16 22:30:24.000000000 +0200
--- configure.ac	2015-10-08 10:48:42.981200859 +0200
*************** if test "${enable_pcsc}" = "yes"; then
*** 575,580 ****
--- 575,605 ----
  	fi
  	AC_DEFINE_UNQUOTED([DEFAULT_PCSC_PROVIDER], ["${DEFAULT_PCSC_PROVIDER}"], [Default PC/SC provider])
  	AC_DEFINE([ENABLE_PCSC], [1], [Define if PC/SC is to be enabled])
+ 
+ 	saved_CFLAGS="${CFLAGS}"
+ 	CFLAGS="${CFLAGS} ${PCSC_CFLAGS}"
+ 	AC_TRY_RUN([
+ #include <pcsclite.h>
+ #include <stdio.h>
+ 
+ int
+ main () {
+   int major, minor, micro;
+   int rc = sscanf(PCSCLITE_VERSION_NUMBER, "%d.%d.%d", &major, &minor, &micro);
+   if (rc == 3) {
+     printf("pcsc-lite version: %d.%d.%d\n", major, minor, micro);
+     /* checks on version can be done when pcsc-lite fixes the issue */
+     return 0;
+   } else {
+     printf("Invalid version number: %s\n", PCSCLITE_VERSION_NUMBER);
+     return 1;
+   }
+ }
+ ],fix_pcsclite=yes,,)
+ 	CFLAGS="${saved_CFLAGS}"
+ 	if test "x$fix_pcsclite" = "xyes"; then
+ 		AC_DEFINE([FIX_PCSCLITE], [1], [Define if the fix for PCSLITE is needed])
+ 	fi
  fi
```

frankmorgner:
`1.4.0` is still present for OS X 10.10.0 sdk. So I think they are shipping with unmodified headers. I don't believe you can determine in which of both systems you are just by looking at `pcsclite.h`

rickyepoderi:
Yep! If "1.4.0" is still present my little program is useless. Just continue with your solution. Thanks!

frankmorgner:
tested OK with OS X 10.11

PR #519, opened 2015-08-11
Fix to CKA_PRIVATE handling pcks11-tool

NWilson:
There's a copy-and-paste bug in there, where the CKA_PRIVATE attribute is being set on the wrong variables! As well as fixing that, we should explicitly set CKA_PRIVATE to "false" for certificates and public keys, since the PKCS#11 spec doesn't specify a default. The intention of pkcs11-tool is to create certs and keys as public, but some PKCS#11 drivers use "private" as the default, so for those drivers pkcs11-tool needs to specify a value for CKA_PRIVATE or the protection level for the written object won't be right.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -1921,6 +1921,7 @@ static int parse_gost_private_key(EVP_PKEY *evp_key, struct gostkey_info *gost)
 static int write_object(CK_SESSION_HANDLE session)
 {
 	CK_BBOOL _true = TRUE;
+	CK_BBOOL _false = FALSE;
 	unsigned char contents[MAX_OBJECT_SIZE + 1];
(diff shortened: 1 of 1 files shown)


frankmorgner:
thanks! sounds and looks sane. I don't have a card to test it, though.

Next time, you could split up the commit.

NWilson:
Thanks, I think you're saying there's no action for me to take right now (but I could split the commit if you want).

I've tested it using SoftHSM, which is a convenient PKCS#11 library for development (but not with a physical card).

frankmorgner:
@NWilson correct! I'm waiting for someone else to confirm the fix.

viktorTarasov:
applied in 4df35b922c8eb7e0776a23260b65e570b33e4d42,
thanks.

PR #520, opened 2015-08-11
added call back for getting vendor/product id

frankmorgner:
implementation taken from
https://github.com/jasp00/OpenSC/commit/83142d4caeae5958850795757c7d6c5cf072a15c

### DIFF
--- a/src/libopensc/internal-winscard.h
+++ b/src/libopensc/internal-winscard.h
@@ -201,6 +201,8 @@ typedef LONG (PCSC_API *SCardGetAttrib_t)(SCARDHANDLE hCard, DWORD dwAttrId,\
 #define PCSCv2_PART10_PROPERTY_sFirmwareID 8
 #define PCSCv2_PART10_PROPERTY_bPPDUSupport 9
 #define PCSCv2_PART10_PROPERTY_dwMaxAPDUDataSize 10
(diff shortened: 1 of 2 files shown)


frankmorgner:
Maybe you're right...

frankmorgner:
looking through use cases of vid/pid I stumbled across https://www.opensc-project.org/opensc/ticket/340. Since the PIN recognition of this reader has been fixed in @LudovicRousseau's CCID driver and the MS CCID driver does not support PIN entry, I think the OpenSC code for handling this reader can be removed, right?

frankmorgner:
I updated the commit for internal usage only. currently the vendor/prduct id is only debugged. in the future it could be used for implementing workarounds of reader bugs

src/libopensc/reader-ctapi.c
```diff
@@ -561,7 +561,7 @@ struct sc_reader_driver * sc_get_ctapi_driver(void)
 	ctapi_ops.perform_verify = ctbcs_pin_cmd;
 	ctapi_ops.perform_pace = NULL;
 	ctapi_ops.use_reader = NULL;
-
+
```
martinpaljak: Aha!

frankmorgner:
I rebased the commits onto the current master.

I also removed handling the HP reader as it is solved already by libccid.

PR #522, opened 2015-08-12
really set the --login option when using --pin

HenryJacques:
Until now, if -p was used without -l, we didn't authenticate to the token (see man pkcs11-tool)


HenryJacques:
Hum... I'm not a MacOS user at all, and I've strictly no idea why the build job https://travis-ci.org/OpenSC/OpenSC/jobs/75264392 has failed...

martinpaljak:
IT succeeded now.

PR #530, opened 2015-08-21
Yubikey NEO pin functions support

NWilson:
Two changes here, in two commits.

One of them adds support for C_GetTokenInfo to ISO7816 cards. If you do a PIN "VERIFY" command without sending the PIN, the card is supposed to give you back the number of tries remaining without counting it as a failed attempt. This feature of the spec seems to work fine with the Yubikey, and is hopefully safe for other cards (I'm completely new to OpenSC's codebase, so do please check I've implemented this is the right way!)

The other thing is a hack for the Yubikey, which returns status codes for the PIN attempts in a format not described by the spec, but I think it's safe to add support for the Yubikey format. I sent in a technical support request to Yubico to ask if they knew why that behaviour was there, I haven't heard back yet.

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -170,7 +170,7 @@ static int myeid_init(struct sc_card *card)
 #endif
 
 	/* State that we have an RNG */
-	card->caps |= SC_CARD_CAP_RNG;
+	card->caps |= SC_CARD_CAP_RNG | SC_CARD_CAP_ISO7816_PIN_INFO;
 
 	card->max_recv_size = 255;
 	card->max_send_size = 255;
@@ -524,53 +524,12 @@ static int myeid_delete_file(struct sc_card *card, const struct sc_path *path)
 	LOG_FUNC_RETURN(card->ctx, sc_check_sw(card, apdu.sw1, apdu.sw2));
 }
 
-static int myeid_pin_info(sc_card_t *card, struct sc_pin_cmd_data *data,
-			   int *tries_left)
-{
-	sc_apdu_t apdu;
-	int r;
-
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_1, 0x20, 0x00, data->pin_reference);
-
-	r = sc_transmit_apdu(card, &apdu);
-	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
-
-	r =  sc_check_sw(card, apdu.sw1, apdu.sw2);
-
-	if (r == SC_ERROR_PIN_CODE_INCORRECT) {
-		data->pin1.tries_left = apdu.sw2 & 0xF;
-		r = SC_SUCCESS;
-	} else if (r == SC_ERROR_AUTH_METHOD_BLOCKED) {
-		data->pin1.tries_left = 0;
-		r = SC_SUCCESS;
-	}
-	LOG_TEST_RET(card->ctx, r, "Check SW error");
-
-	if (r == SC_SUCCESS)
-	{
-		data->pin1.pad_length = data->pin2.pad_length = 8;
-	        data->pin1.pad_char = data->pin2.pad_char = 0xFF;
-	}
-
-
-	if (tries_left != NULL) {
-		*tries_left = data->pin1.tries_left;
-	}
-
-	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
-}
-
 static int myeid_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 			 int *tries_left)
 {
 	myeid_private_data_t *priv = (myeid_private_data_t *) card->drv_data;
 
 	LOG_FUNC_CALLED(card->ctx);
-
-	if (data->cmd == SC_PIN_CMD_GET_INFO)
-	{
-	        return myeid_pin_info(card, data, tries_left);
-	}
 
 	sc_log(card->ctx, "ref (%d), pin1 len(%d), pin2 len (%d)\n",
 	              data->pin_reference, data->pin1.len, data->pin2.len);
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2895,7 +2895,7 @@ static int piv_init(sc_card_t *card)
 	_sc_card_add_ec_alg(card, 256, flags, ext_flags, NULL);
 	_sc_card_add_ec_alg(card, 384, flags, ext_flags, NULL);
 
-	card->caps |= SC_CARD_CAP_RNG;
+	card->caps |= SC_CARD_CAP_RNG | SC_CARD_CAP_ISO7816_PIN_INFO;
 
 	/*
 	 * 800-73-3 cards may have a history object and/or a discovery object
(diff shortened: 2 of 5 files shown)


frankmorgner:
https://github.com/NWilson/OpenSC/commit/93f9709db0ed00925bcc54947495fc032dd81a21 looks good. It's also compliant to ISO 7816-4:

> If INS = '20', the command data field is normally present for conveying verification data. The absence of command data field is used to check whether the verification is required (SW1-SW2 = '63CX' where 'X' encodes the number of further allowed retries), or not (SW1-SW2 = '9000').

However, as quoted above, the `0xC0` nibble is required which makes https://github.com/NWilson/OpenSC/commit/ab1ae0a2add7df259a71e31da561ffb55efdf561 break the standard. Hence, this *should* go into a card specific driver for the Yubikey NEO...

frankmorgner:
ISO defines `6340` and `6381` which would be misinterpreted as retry counters with NWilson@ab1ae0a. I think that only the first commit should be merged

NWilson:
Thanks for the review. Oops, well spotted, I broke other legal return codes! We could make the check more specific, to test for 0xCX or 0x0X (which I think is safe, no compliant card would return 0x0X).

I don't know what OpenSC's policy is on spec compliance, it seems relatively harmless in this instance.. is it worth adding a new driver for this tiny case?  If you do want a Yubikey NEO driver, you'd have to advise me how to do that. I wouldn't want to have to copy and paste any code, or it would become a source of bitrot and maintenace frustration.

NWilson:
OK, I'll come back to this next week then and make a `card-yubikey-neo.c` driver which uses all the functions from the driver in `card-piv.c`, but overrides the `piv_pin_cmd` with the hack here. As I do more testing with my Yubikey, we can  put any more workarounds in there too.

Thanks for the feedback.

dengert:
On 8/22/2015 5:34 AM, Nicholas Wilson wrote:
> OK, I'll come back to this next week then and make a |card-yubikey-neo.c| driver which uses all the functions from the driver in |card-piv.c|, but overrides the |piv_pin_cmd| with the hack here. As I
> do more testing with my Yubikey, we can put any more workarounds in there too.

Please don't make a whole new driver, this will only cause maintenance problems in the future.
We are talking about a command that is not often used, but is standard, its the Neo that has the problem.

Look at the other cards that test for  data->cmd == SC_PIN_CMD_GET_INFO Id this the right command to
use for the verify with no data?

Continue to work on the iso7816 code, and implement SC_PIN_CMD_GET_INFO (or what ever) with the "C" expected.


Add to opensc.h soething like:
#define SC_PIN_CMD_63_NO_C 0x00008

Then in iso7816.c if the command is SC_PIN_CMD_GET_INFO and SC_PIN_CMD_63_NO_C and sw1=63 and sw2 != 00, turn on the "C" and continue with the command.


The card-piv.c (or any other driver) could then set SC_PIN_CMD_63_NO_C, if the card has a problem.
The card-piv could set this of the ATR was for a Neo.


>
> Thanks for the feedback.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/530#issuecomment-133676325>.
>

-- 

frankmorgner:
I still believe that handling the missing nibble should be solved locally in the piv driver. It turns out the PIN command already has something in place to solve this problem...

@NWilson could you check whether https://github.com/OpenSC/OpenSC/commit/623aa5e9d78a7934cc5b523686c0817092f6dc4d behaves like https://github.com/NWilson/OpenSC/commit/ab1ae0a2add7df259a71e31da561ffb55efdf561?

frankmorgner:
Taking @dengert's comment into account (see https://github.com/NWilson/OpenSC/commit/ab1ae0a2add7df259a71e31da561ffb55efdf561#comments) 623aa5e could be extended to check if the card is actually the NEO since it even violates the PIV standard.

martinpaljak:
While not the result of current patch, a reminder: there are cards (no list at hand) out there (several which I recall from mailing list posts) that fail with empty VERIFY (do actual verification that fails), and *may* be currently delegating to plain ios7816.c PIN command. Which would be a really annoying regression for those users who are hit. So just don't make it a default somewhere inside the generic (PKCS#15-ish) layers. I know that some pinpad readers with PIN firewalling capabilities or more extensive APDU processing have tried to make it a usability helper and thus made hard to discover blocked cards...

martinpaljak:
And of course, +1 to @frankmorgner: even if the bug of a missing "C" is not unique to NEO (there's one in Estonia as well!) it should be handled in a case-by-case basis, inside the specific driver.

dengert:
Please look at:

https://github.com/OpenSC/OpenSC/pull/531

This is self contained to the PIV drivers and will work with PIV cards and Neo tokens. 

I have tried this with NIST test cards and the Yubikey Neo, with the pkcs11-tool -L -v

dengert:
See the https://github.com/OpenSC/OpenSC/pull/531,

It is self contained in the PIV card drivers and has been tested with NIST PIV test cards, and Yubico Neo.

P.S. A simple way to see if the card meets the specs is to do something like:

$ opensc-tool -s 00:20:00:80
Using reader with a card: SCM Microsystems Inc. SCR 355 [CCID Interface] 00 00
Sending: 00 20 00 80
Received (SW1=0x63, SW2=0xC3)

Just set P1 and P2 based on the card. The above is a PIV card.


On 8/23/2015 2:54 PM, Martin Paljak wrote:
> And of course, +1 to @frankmorgner <https://github.com/frankmorgner>: even if the bug of a missing "C" is not unique to NEO (there's one in Estonia as well!) it should be handled in a case-by-case
> basis, inside the specific driver.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/530#issuecomment-133910560>.
>

-- 

dengert:
The C_GetTokenInfo tries to set the CKF_USER_PIN_* flags, so there is code already in the "generic (PKCS#15-ish) layers",
but iso7816.c PIN command does not support SC_PIN_CMD_GET_INFO.

On 8/23/2015 2:52 PM, Martin Paljak wrote:
> While not the result of current patch, a reminder: there are cards (no list at hand) out there (several which I recall from mailing list posts) that fail with empty VERIFY (do actual verification that
> fails), and /may/ be currently delegating to plain ios7816.c PIN command. Which would be a really annoying regression for those users who are hit. So just don't make it a default somewhere inside the
> generic (PKCS#15-ish) layers. I know that some pinpad readers with PIN firewalling capabilities or more extensive APDU processing have tried to make it a usability helper and thus made hard to
> discover blocked cards...
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/530#issuecomment-133910057>.
>

-- 

NWilson:
Right, coming back to this after the weekend, there are a few conflicting comments here!

Frank suggested a card-specific driver for the NEO, Doug wants me to put it in iso7816.c, and Martin wants it at least not to be turned on by default in the common code.

I've updated this PR with new commits, which might make everyone happy! The VERIFY implementation of PIN_CMD_GET_PIN_INFO is part of the ISO7816 standard, so it shouldn't be duplicated in every card's driver that supports it. As it stands, we have identical copy-and-paste copies of it in card-myeid.c and card-sc-hsm.c, which is bad, and I'm not going to add a third copy of the same code to card-piv.c.

I've removed the two copies of the code from card-myeid.c and card-sc-hsm.c, and put it in in iso7816.c but guarded with a flag, so that it's only used for cards that indicate support for it.

Finally, the "0xCN" hack for the NEO has been put in card-piv.c by overriding the "check_sw" method.

As long as I haven't made any mistakes and broken myeid or sc-hsm cards, is this new approach better? I can't see many places in OpenSC where the "flags" field is actually used, so I might not be setting or initialising it in an idiomatic way.

NWilson:
Regarding the Yubikey NEO bug -- I reported this to Yubico on Friday and they created ticket number 00012364 for me. A Yubico engineer from Sweden has confirmed today that they intend to fix it in a future revision of the PIV applet.

In yet another thread (!) in this fragmented conversation (one of Frank's edits of my commit, 623aa5e), @klali from Yubico has asked for OpenSC to ship a workaround for the bug:
> I would be very happy if there was a workaround for it since we've shipped quite a few with that bug.

martinpaljak:
Workarounds for bugs is a thing that is natural to smart card middlwares, as the shelf-life of the cards is usually 3..5 years, at least. Is there a way to localize the bug to a specific version of NEO ?

NWilson:
Perhaps, I'm not great with smartcards, I've literally just bought one last week knowing nothing about them... I think we can probably assume it applies to all NEOs sold until the first one that has the fix in, in which case we won't have a full list of GUIDs or ATRs the bug applies to, until we see the first fixed one released?

frankmorgner:
Though I would have rather taken the blacklist approach for the PIN info command, I think your changes are good enough so that I won't offer yet another PR and that also #531 can be closed. Just one remark, could you use `card->caps` instead of `card->flags`? (there are already some T-COS specific flags; this can be adapted for your ISO-specific flag).

Your error handling of the Neo also superseds my suggestion from here https://github.com/OpenSC/OpenSC/commit/623aa5e9d78a7934cc5b523686c0817092f6dc4d. However, I agree, that your workaround should be limited to the Neo only at least by checking the ATR...

frankmorgner:
@CardContact should test the generic implementation in the iso driver if it still works with sc-hsm.

NWilson:
I've updated the PR again with those comments (using `card->caps`, and checking the ATR for the NEO). Tested fairly well on the NEO, it seems to work.

Needs a test of the sc-hsm myeid changes.

PR #531, opened 2015-08-23
Support the use of a VERIFY command to get the PIN retry count for PI…

dengert:
…V cards

ISO 7816-4 and NIST 800-73-3 support using the VERIFY command as a type 1 APDU
to get the current pin retry count via the stats of sw1=63, sw2=CX where X
is the retry count.

Current Neo tokens return sw1=63, sw2=0X, this is also accepted.

The max_retry count is set when the PIV cards are personilized, and is not
defined in the NIST standards. There is no way to read the max_retry count
from the card.  The pkcs15-piv.c assumes it is 8, which is then used in
C_GetTokenInfo to set some of the CKF_USER_PIN_ flags.

The iso7816.c was not modified for these reasons:
  (1) The Neo does not follow the standard.
  (2) Some non piv cards already have code to handle SC_PIN_CMD_GET_INFO
  (3) Some cards may fail if this code was added to iso7816.c

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2920,6 +2920,35 @@ static int piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 	 * FIPS 140-2 (6 character minimum) requirements.
 	 */
 	struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
+
+	/* do the query of tries left here as the iso7816.c does not implement this
+	 * and Neo does set return correctly 
+	 */
+	if (data->cmd == SC_PIN_CMD_GET_INFO) {
+		struct sc_apdu local_apdu, *apdu;
+		int r;
+
+		apdu = &local_apdu;
+		if (tries_left)
+		    *tries_left = -1;
+
+		sc_format_apdu(card, apdu, SC_APDU_CASE_1, 0x20, 0x00, data->pin_reference);
+		apdu->lc = 0;
+		apdu->datalen = 0;
+		apdu->data = NULL;
+
+		r = sc_transmit_apdu(card, apdu);
+		LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+		if (apdu->sw1 == 0x63) {
+			/* we should be testing for 63 CX, but Neo returns 63 0X */
+			if (tries_left != NULL)
+				*tries_left = apdu->sw2 & 0x0F;
+			data->pin1.tries_left = apdu->sw2 & 0x0F;
+			return SC_SUCCESS;
+		}
+		return sc_check_sw(card, apdu->sw1, apdu->sw2);
+	}
+
 	if (data->cmd == SC_PIN_CMD_CHANGE) {
 		int i = 0;
 		if (data->pin2.len < 6) {
--- a/src/libopensc/pkcs15-piv.c
+++ b/src/libopensc/pkcs15-piv.c
@@ -796,6 +796,7 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		pin_info.attrs.pin.pad_char      = pins[i].pad_char;
 		sc_format_path(pins[i].path, &pin_info.path);
 		pin_info.tries_left    = -1;
+		pin_info.max_tries = 8; /* The value is up to the agency and can not be read from the card */
 
 		label = pins[i].label;
 		if (i == 0 &&


frankmorgner:
Ignoring the `C0` nibble should also be done on normal PIN verification (with an actual PIN). This commit doesn't capture this notion while https://github.com/OpenSC/OpenSC/commit/623aa5e9d78a7934cc5b523686c0817092f6dc4d does.

Implementing `SC_PIN_CMD_GET_INFO` in `iso7816.c` is the right thing to do because it *is standardized by ISO*. It's the other card drivers that need to be edited to preserve backward compatibility! (Your argument of not modifying the iso driver because of the other cards that depend on it would prevent *any* modification of the iso driver)

frankmorgner:
Luckily, preserving backward compatibility is easy in this case: Simply modify all pin_cmd call backs that rely on the iso implementation to throw an error on `SC_PIN_CMD_GET_INFO`...

dengert:
In https://github.com/OpenSC/OpenSC/pull/530 @frankmorgner said:

"I still believe that handling the missing nibble should be solved locally in the piv driver. It turns out the PIN command already has something in place to solve this problem..."

And @martinpaljak said: "And of course, +1 to @frankmorgner: even if the bug of a missing "C" is not unique to NEO (there's one in Estonia as well!) it should be handled in a case-by-case basis, 
inside the specific driver."

I think we have changed positions and I agree with the above, the card drivers should handle it.

In regard to a return from a failed verify with an incorrect password, even though a 63 CX (or 63 0X) is returned the verify failed,
and the use has the wrong PIN, This has not been an issue with the NEO. Maybe they will fix the return code in future versions.

On 8/23/2015 5:44 PM, Frank Morgner wrote:
> Ignoring the |C0| nibble should also be done on normal PIN verification (with an actual PIN). This commit doesn't capture this notion while 623aa5e
> <https://github.com/OpenSC/OpenSC/commit/623aa5e9d78a7934cc5b523686c0817092f6dc4d> does.
>
> Implementing |SC_PIN_CMD_GET_INFO| in |iso7816.c| is the right thing to do because it /is standardized by ISO/. It's the other card drivers that need to be edited to preserve backward compatibility!
> (Your argument of not modifying the iso driver because of the other cards that depend on it would prevent /any/ modification of the iso driver)
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/531#issuecomment-133950457>.
>

-- 

dengert:
If you want to modify iso7816.c and all the card drivers, do it.
https://github.com/NWilson wanted this for the NEO.




On 8/23/2015 5:49 PM, Frank Morgner wrote:
> Luckily, preserving backward compatibility is easy in this case: Simply modify all pin_cmd call backs that rely on the iso implementation to throw an error on |SC_PIN_CMD_GET_INFO|...
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/531#issuecomment-133950708>.
>

-- 

NWilson:
Somehow we've split the discussion between two PRs and ended up with three competing branches from me, Doug, and Frank! I hadn't read this thread until today...

I think it's going to become a problem in the long run if we copy and paste this code into many card drivers. Already we have essentially identical code two drivers (card-myeid.c, card-sc-hsm.c), and this PR suggests copying into a third file.

I've updated my commits in the other PR. What I'm suggesting now is instead of duplicating this code into many card drivers, we can selectively implement it in iso7816.c so that it's only used on cards that support it.

martinpaljak:
Yes: code duplication is bad, but implicitly enabling (possibly hazardous) card communication without asking is not good. Explicitly using the code from iso7816.c would be the best option.

frankmorgner:
PRs are good to see what everyone means. I'll try to hammer a third PR this weeks that includes:
- a general implementation of `SC_PIN_CMD_GET_INFO` in the iso driver
- neo specific workaround in the piv driver
- backward compatibility code to not use the iso implementation of  `SC_PIN_CMD_GET_INFO` in all other drivers

I think this will take all comments so far into account...

NWilson:
OK, but that sounds a bit wasteful. I'm not proud, I don't mind if you fix my problem, but in most projects I've sent in patches to, there's been a chance to respond to feedback until the PR is accepted. I changed my PR to take the feedback into account, because I thought that was being requested, but I wouldn't have done that if I'd realised it wasn't actually moving things forward.

I think my PR currently satisfies your three bullet points. I think overriding "check_sw" in card-piv.c is a neat solution to the NEO's problems, maybe you could at least use that bit.

frankmorgner:
@CardContact could you have a look at the changes?

frankmorgner:
... I think the current PR of @NWilson takes all remarks into account.

dengert:
I withdraw my pull request, as you have a different solution that is more general.

PR #532, opened 2015-08-23
implemented sloppy initialization for C_GetSlotInfo

### LINKED ISSUE #523: CKR_SLOT_ID_INVALID Exception with java pkcs11
I have an application that accesses the data on an Italian ID card which I wrote using version 0.13.0 of OpenSC. Since version 0.14.0 the application no longer works giving the following when trying to access the card:

Exception in thread "main" java.security.ProviderException: Initialization failed
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:376)
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:103)
	at smartcardtest.SmartCardCIE.setProvider(SmartCardCIE.java:104)
	at smartcardtest.SmartCardTestCIE.main(SmartCardTestCIE.java:32)
Caused by: sun.security.pkcs11.wrapper.PKCS11Exception: CKR_SLOT_ID_INVALID
	at sun.security.pkcs11.wrapper.PKCS11.C_GetSlotInfo(Native Method)
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:365)
	... 3 more

This is happening in my code where I'm setting the security provider:

        Provider p = new sun.security.pkcs11.SunPKCS11(configName);

where my config file contains:

name = myConfig
library = /usr/lib/opensc-pkcs11.so
slot = 1

Has something changed between version 0.13.0 and 0.14.0 that would be causing this. I also get the same error with version 0.15.0. Anyone have any ideas?

frankmorgner:
Makes things work for Java and closes #523 

Honestly, I don't think a 'strict' mode is necessary as suggested in https://github.com/OpenSC/OpenSC/issues/523#issuecomment-133909209.

I did not test this in depth, @wtmann could you check if this solves your issue even without setting `slotListIndex = X`?


wtmann:
@frankmorgner  Sorry about the delay. I just finished applying the patch and checking things. Unfortunately this does not solve the problem. If I directly select a slot I get the error in question and not specifying a slot at all just gives me a pkcs11 keystore not available exception.

frankmorgner:
Hmm, @wtmann what would be an easy way of reproducing this behaviour?

wtmann:
@Frank Morgner  Having a security provider configuration file like this:

pkcs11.properties
==============
name = myConfig
library = /usr/lib/opensc-pkcs11.so
slot = 1

and a piece of Java code to set the security provider like so:

    public void setProvider(String config) {    // Name is the name of the
configuration file pkcs11.properties
        // Add the PKCS11 security provider
        this.configName = config;
        Provider p = new sun.security.pkcs11.SunPKCS11(configName);
        Security.addProvider(p);
    }

As soon as you execute the line to create the provider you get the error.


On Tue, Aug 25, 2015 at 8:08 PM, Frank Morgner <notifications@github.com>
wrote:

> Hmm, @wtmann <https://github.com/wtmann> what would be an easy way of
> reproducing this behaviour?
>
> —
> <https://github.com/OpenSC/OpenSC/pull/532#issuecomment-134688885>.
>



-- 
William Mann

frankmorgner:
Hmm, I need a test case that ideally involves copy/pasting only...

wtmann:
I'm attaching a link to a zip file containing: a small java test file, the
compiled jar, the properties file, and a run command (on my 64-bit machine
I have to specifiy the location of libpcsc). When run you should get a
printout of the ATR and the historical bytes. Then go into the properties
file and switch the comment between slot and slotindexlist and you should
get the error right away. Hope this helps.

https://drive.google.com/file/d/0B_lWKDdIYuQTcGtGeFBKc01FaTA/view?usp=sharing

On Wed, Aug 26, 2015 at 3:01 AM, Frank Morgner <notifications@github.com>
wrote:

> Hmm, I need a test case that ideally involves copy/pasting only...
>
> —
> <https://github.com/OpenSC/OpenSC/pull/532#issuecomment-134781711>.
>



-- 
William Mann

frankmorgner:
@wtmann it should work now, I overlooked the hotplug slot.

@martinpaljak I added a configuration setting, which is on by default.

frankmorgner:
@wtmann did you have time looking into the new fixes, again?

frankmorgner:
@martinpaljak do you think that this can be merged given the possibility it can be switched off via `opensc.conf`?

src/pkcs11/pkcs11-global.c
```diff
@@ -485,6 +486,19 @@ CK_RV C_GetSlotInfo(CK_SLOT_ID slotID, CK_SLOT_INFO_PTR pInfo)
 
 	sc_log(context, "C_GetSlotInfo(0x%lx)", slotID);
 
+	if (sc_pkcs11_conf.plug_and_play)
+		uninit_slotcount = 1;
```
martinpaljak: I personally prefer braces with all conditionals (as per https://www.securecoding.cert.org/confluence/display/c/EXP19-C.+Use+braces+for+the+body+of+an+if%2C+for%2C+or+while+statement) but as there is no consistent style in OpenSC, it doesn't really matter...

martinpaljak:
Sure, conceptually (if it fixes it for Java) it is OK.

martinpaljak:
I'm wondering if it would be possible to have a single "default" config file for all platforms, though...

frankmorgner:
OK. I've tested with the java example provided. But I'd like to get additional confirmation from @wtmann if the fixes work for him.

frankmorgner:
> I'm wondering if it would be possible to have a single "default" config file for all platforms, though...

It's possible, see the last commit ;-)

wtmann:
Hi. Sorry for the delay,  been away for a while... I'll test the code this
afternoon and let you know.

wtmann:
@frankmorgner  I applied and tested the patches this morning and everything worked! Your sloppy mode solution fixed everything. Thanks!

Issue #535, opened 2015-08-27
Add --raw option to pkcs15-tool


nioncode:
It would be great if there was a '--raw' option in pkcs15-tool to get the 8bit raw data when using "pkcs15-tool -R 'data' --pin PIN" on stdout. pkcs15-crypt has this option already.

Currently, we have to write the data to an output file first and then read it in again.
This is rather inconvenient and also less secure if the file gets written to persistent storage where it can't be easily and securely deleted.

frankmorgner:
I'd be happy to review your pull request... ;-)

nioncode:
I just implemented it and created the pull request.

nioncode:
Closed with #540.

PR #540, opened 2015-09-03
add '--raw' option to pkcs15-tools '--read-data-object'

nioncode:
This adds a '--raw' option to pkcs15-tool that affects how '--read-data-object' outputs its data.
If it is not given, the current behavior is not changed.
If present, pkcs15-tool outputs the data in raw binary mode, i.e. not in hex mode.

This allows an application to read a data object with pkcs15-tool directly from stdout instead of redirecting it to a file first and reading it back in afterwards, which is inconvenient and might be less secure (e.g. when the data is written to unsecure persistent storage).

This implements #535.

### DIFF
--- a/doc/tools/pkcs15-tool.1.xml
+++ b/doc/tools/pkcs15-tool.1.xml
@@ -156,6 +156,18 @@
 
 				<varlistentry>
 					<term>
+						<option>--raw</option>
+					</term>
+					<listitem><para>Changes how <option>--read-data-object</option> prints the content
+					to standard output. By default, when <option>--raw</option> is not given, it will
+					print the content in hex notation. If <option>--raw</option> is set, it will print
+					the binary data directly. This does not affect the output that is written to the
+					file specified by the <option>--output</option> option. Data written to a file will
+					always be in raw binary.</para></listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>
 						<option>--read-certificate</option> <replaceable>cert</replaceable>,
 						<option>-r</option> <replaceable>cert</replaceable>
 					</term>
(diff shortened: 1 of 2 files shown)


nioncode:
By the way: is there a reason why you don't use the convenience functions defined in utils.h (util_print_binary, util_hex_dump, etc.). I did not use them either to stay consistent with your current code.

Also: does it make sense to output the data to stdout when the user requests the output to be written to a file by providing the --output option? I would expect to not see any output on stdout then.

frankmorgner:
use the helpers where ever you can! Previous developers didn't always do so, unfortunately.

I'd prefer if `--raw` would have the same meaning for pkcs15-tool and pkcs15-crypt. With your implementation in pkcs15-tool the switch only decides how the data is printed on stdout while the raw output is always written into the output file. In pkcs15-crypt, however, the switch how the data is printed to the output file (may be stdout).

frankmorgner:
You should also update https://github.com/OpenSC/OpenSC/blob/master/doc/tools/pkcs11-tool.1.xml

nioncode:
I did not want to change existing behavior, that's why the `--raw` option only modifies the stdout behavior.
If it would also apply to the output file, existing applications would break because they would not use the `--raw` option, because the file output always uses binary mode.

Doesn't pkcs15-crypt also always output binary data to the file? If I'm not mistaken, the ternary `int output_binary = (opt_output == NULL && opt_raw == 0 ? 0 : 1)` evaluates to 1 whenever opt_output is != NULL. The ternary is difficult to read and requires knowledge about the precedence of `&&` vs the ternary operator.
I don't think there is any benefit of printing the hex values to a file anyway, is it?

I will update the documentation file later today.

frankmorgner:
Backwards compatibility... agreed! But please change the help message so that it influences 8 bit output of stdout only.

nioncode:
I just updated the help message and added documentation for `--raw`.

doc/tools/pkcs15-tool.1.xml
```diff
@@ -156,6 +156,18 @@
 
 				<varlistentry>
 					<term>
+						<option>--raw</option>
+					</term>
+					<listitem><para>Changes how <option>--read-data-object</option> prints the content
+					to standard output. By default, when <option>--raw</option> is not given, it will
+					print the content in hex notation. If <option>--raw</option> is set, it will print
```
frankmorgner: typo: prints

frankmorgner:
thanks

PR #550, opened 2015-09-11
adjust Make.rules.mak to work with AppVeyor

frankmorgner:
# Work in progress, do not yet merge

The old configuration doesn't work (anymore):

- The current builds of OpenSSL from https://slproweb.com/products/Win32OpenSSL.html include security measures https://blogs.microsoft.com/cybertrust/2012/01/26/enhancements-to-gs-in-visual-studio-11/ that previously yielded an error (`error LNK2001: unresolved external symbol ___report_rangecheckfailure`)
- I guess that the configuration for zlib 1.2.5 never ever has worked anyway, because the library names were totally wrong...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,9 @@ matrix:
     - compiler: gcc
       os: linux
       env: ENABLE_DOC=--enable-doc
-    - compiler: gcc
-      os: linux
+    - os: linux
+      env: HOST=x86_64-w64-mingw32
+    - os: linux
       env: HOST=i686-w64-mingw32
 
 before_install:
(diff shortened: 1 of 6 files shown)


frankmorgner:
I finally got this working! AppVeyor even allows everyone to download the build artifact, i.e. the OpenSC installer.

However, I am using the pre-built shared libraries of zlib and OpenSSL, which are currently not included in our Bundle. @vletoux do you, maybe, want to include that into the installer/appveyor.yml/nmake-files? We would have ready to use nightly builds for all Windows users...

frankmorgner:
It would also be possible to upload artifacts from travis (e.g. dmg/deb packages for OSX/Ubuntu) and to use a permanent link for the most recent nightly build for all three OSs. Should we upload this on opensc.fr @viktorTarasov or opensc-project.org @martinpaljak ?

LudovicRousseau:
Maybe we could continue to use https://sourceforge.net/projects/opensc/files/OpenSC/nightly/ ?

I see that regular uploads are made for tarball, win32 and win64. I don't know where the uploads come from. Adding a README.txt file would be fine :-)

frankmorgner:
@LudovicRousseau I just saw, that your AppVeyor builds wait for completion of a lot of obsolete PRs. There are a lot of builds queued for this PR (#550) with commits that got overwritten. For my remote (https://ci.appveyor.com/project/frankmorgner/opensc) those old jobs were cancled once I pushed a new (or modified) patch to the branch. This seems to be a bug in AppVeyor, but it's still possible for you to cancel all the pending obsolete jobs...

frankmorgner:
An issue has already been filed here https://github.com/appveyor/ci/issues/38

frankmorgner:
I added a workaround for the AppVeyor job queue

Issue #538, opened 2015-08-31
Correctly handle a card with multiple PKI applications (e.g. PIV *and* PGP)


dengert:
Smart card appications these days can be loaded on many different cards. Yet OpenSC still runs through a list of card drivers that look at the ATR of a card to determine how to handle the card. 
The PIV application, and OpenPGP are examples of two alpplications that can be on on many different cards. ID cards in Europe also appear to be applications on multiple cards. These applications in many cases are not PKCS#15 cards, but use emulation. 

Code changes proposed by #507 show the problems with the current code, where a card has one ATR but could have multiple applications on the card.  #507 just added the  ATR of the Yubico Neo to the card-openpgp.c and a match of the ATR in card-openpgp.c caused the search to stop with the card-driver, which later in the code determined the  OpenPGP application was not on the card. (The historical  bytes in an ATR can contain an AID, but not all cards do this. The Neo does not, and thus two card drivers could try and use the card. 

The card-piv.c on the other hand does a select for the AID during the match process, and if the AID was not found, other card drivers would be given a chance to match the card.  A proposed re-ordering of the card drivers in ctx.c was proposed in #507, but this only addressed the two applications. 

I would like to propose, that the list of card drivers include a psudo driver that would look for an AID
and then call the appropriate card-driver that supports the AID. This would also keep extra operations to a card down, as each driver does not have to read the the AID. 

A quick grep of the source code shows that many of the OpenSC card drivers already  look for an AID.
The Neo appears to be the first card where the ATR does not change, yet it can have two applications that OpenSC could support. 

Still to be addressed is how a user could select one or more applications on the same card.

martinpaljak:
I think that the policy should be, that *if* the ATR is known to be uniq and of meaningful value, it SHOULD be used as a shortcut. But ATR MUST NOT be used as the sole indicator of card identity for cases where it is known to be non-uniq for an application/driver.

frankmorgner:
No matter if it's mulitple pseudo drivers looking for mulitple AIDs or mulitple ATR-based driver that look for mulitple ATRs, the real problem stays the same: How do you deal with multiple drivers that match for the current card?

This actually is the problem you leave for future work in your last paragraph. If this problem would have been solved, there would be no problem with a NEO that matches for two card drivers by AID (piv) or ATR (pgp).

dengert:
Pull Request #551  adds the NEO USB ATR to card-openpgp.c. It looks like it will not check the AID
if the ATR is found. 
[...]

PR #551, opened 2015-09-12
Fix OpenPGP driver to work correctly with YubiKey NEO

frankmorgner:
I tried to sum up the discussion from #507 in this PR.

I don't have a NEO. @dengert, @rqou @mouse07410 could you check with your NEOs whether:

1. the NEO defaults to the PIV application
2. You can force the PGP application via opensc.conf as discussed in #507
3. the PGP application works as expected

 Thank you!

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -46,6 +46,7 @@ static struct sc_atr_table pgp_atrs[] = {
 	{ "3b:fa:13:00:ff:81:31:80:45:00:31:c1:73:c0:01:00:00:90:00:b1", NULL, "OpenPGP card v1.0/1.1", SC_CARD_TYPE_OPENPGP_V1, 0, NULL },
 	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:c5:73:c0:01:40:00:90:00:0c", NULL, "CryptoStick v1.2 (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ "3b:da:11:ff:81:b1:fe:55:1f:03:00:31:84:73:80:01:80:00:90:00:e4", NULL, "Gnuk v1.0.x (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_GNUK, 0, NULL },
+	{ "3b:fc:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:4e:45:4f:72:33:e1", NULL, "Yubikey NEO (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ NULL, NULL, NULL, 0, 0, NULL }
 };
 
@@ -439,6 +440,7 @@ pgp_get_card_features(sc_card_t *card)
 		/* get card capabilities from "historical bytes" DO */
 		if ((pgp_get_blob(card, priv->mf, 0x5f52, &blob) >= 0) &&
 		    (blob->data != NULL) && (blob->data[0] == 0x00)) {
+			i = 0;
 			while ((i < blob->len) && (blob->data[i] != 0x73))
 				i++;
 			/* IS07816-4 hist bytes 3rd function table */
@@ -449,7 +451,7 @@ pgp_get_card_features(sc_card_t *card)
 					priv->ext_caps |= EXT_CAP_APDU_EXT;
 				}
 				/* bit 0x80 in byte 3 of TL 0x73 means "Command chaining" */
-				if (hist_bytes[i+3] & 0x80)
+				if (blob->data[i+3] & 0x80)
 					priv->ext_caps |= EXT_CAP_CHAINING;
 			}
 
@@ -766,8 +768,7 @@ pgp_read_blob(sc_card_t *card, pgp_blob_t *blob)
 
 	if (blob->info->get_fn) {	/* readable, top-level DO */
 		u8 	buffer[2048];
-		size_t	buf_len = (card->caps & SC_CARD_CAP_APDU_EXT)
-				  ? sizeof(buffer) : 256;
+		size_t	buf_len = sizeof(buffer);
 		int r = SC_SUCCESS;
 
 		/* Buffer length for certificate */
@@ -835,6 +836,14 @@ pgp_enumerate_blob(sc_card_t *card, pgp_blob_t *blob)
 			cla <<= 8;
 		}
 		tag |= cla;
+
+		/* Awful hack for composite DOs that have
+		 * a TLV with the DO's id encompassing the
+		 * entire blob. Example: Yubikey Neo */
+		if (tag == blob->id) {
+			in = data;
+			continue;
+		}
 
 		/* create fake file system hierarchy by
 		 * using constructed DOs as DF */
@@ -1542,6 +1551,10 @@ pgp_compute_signature(sc_card_t *card, const u8 *data,
 			"invalid key reference");
 	}
 
+	/* if card/reader does not support extended APDUs, but chaining, then set it */
+	if (((card->caps & SC_CARD_CAP_APDU_EXT) == 0) && (priv->ext_caps & EXT_CAP_CHAINING))
+		apdu.flags |= SC_APDU_FLAGS_CHAINING;
+
 	apdu.lc = data_len;
 	apdu.data = (u8 *)data;
 	apdu.datalen = data_len;
@@ -1605,6 +1618,9 @@ pgp_decipher(sc_card_t *card, const u8 *in, size_t inlen,
 	if (card->type == SC_CARD_TYPE_OPENPGP_GNUK) {
 		apdu.flags |= SC_APDU_FLAGS_CHAINING;
 	}
+	/* if card/reader does not support extended APDUs, but chaining, then set it */
+	if (((card->caps & SC_CARD_CAP_APDU_EXT) == 0) && (priv->ext_caps & EXT_CAP_CHAINING))
+		apdu.flags |= SC_APDU_FLAGS_CHAINING;
 
 	apdu.lc = inlen;
 	apdu.data = (u8 *)in;
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -311,7 +311,8 @@ struct pgp_priv_data {
 	sc_security_env_t	sec_env;
 };
 
-/* ABI: check if card's ATR matches one of driver's */
+/* ABI: check if card's ATR matches one of driver's
+ * or if the OpenPGP application is present */
 static int
 pgp_match_card(sc_card_t *card)
 {
@@ -321,6 +322,24 @@ pgp_match_card(sc_card_t *card)
 	if (i >= 0) {
 		card->name = pgp_atrs[i].name;
 		return 1;
+	} else {
+		sc_path_t	partial_aid;
+		unsigned char aid[16];
+
+		/* select application "OpenPGP" */
+		sc_format_path("D276:0001:2401", &partial_aid);
+		partial_aid.type = SC_PATH_TYPE_DF_NAME;
+		if (SC_SUCCESS == iso_ops->select_file(card, &partial_aid, NULL)) {
+			/* read information from AID */
+			i = sc_get_data(card, 0x004F, aid, sizeof aid);
+			if (i == 16) {
+				if (((aid[6] << 8) | aid[7]) >= 0x0200)
+					card->type = SC_CARD_TYPE_OPENPGP_V2;
+				else
+					card->type = SC_CARD_TYPE_OPENPGP_V1;
+				return 1;
+			}
+		}
 	}
 	return 0;
 }
@@ -366,6 +385,16 @@ pgp_init(sc_card_t *card)
 	if (!file)   {
 		pgp_finish(card);
 		return SC_ERROR_OBJECT_NOT_FOUND;
+	}
+
+	if (file->namelen != 16) {
+		/* explicitly get the full aid */
+		r = sc_get_data(card, 0x004F, file->name, sizeof file->name);
+		if (r < 0) {
+			pgp_finish(card);
+			return r;
+		}
+		file->namelen = r;
 	}
 
 	/* read information from AID */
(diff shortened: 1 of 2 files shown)


mouse07410:
Frank,

I’ve done some testing of this PR, and found it working fine on my YubiKey NEO (PIV 0.1.3, OpenPGP 1.0.10). I haven’t tested OpenSC.tokend, as it is giving me problems in general.

$ yubico-piv-tool -a status
CHUID: 3019d4e739da739ced39ce……..
Slot 9a:
Algorithm: RSA2048
Subject DN: CN=uri
Issuer DN: C=US, ST=MA, O=The Burrow, OU=Gatherers, CN=Forest RSA CA 3
Fingerprint: e403cd9afb75328b16289b59e082f673736386f4a57…..
Not Before: Sep  1 03:06:27 2015 GMT
Not After: Sep  1 03:06:27 2018 GMT
…
$ piv-tool -n
PIV-II card
$ OPENSC_CONF=/Library/OpenSC/etc/neo-openpgp-opensc.conf openpgp-tool --pretty -U
Account:   uri
Name:      Blumenthal Uri
Language:  en
Gender:    male
$ gpg2 --card-status
Application ID ...: D276000…..
Version ..........: 2.0
Manufacturer .....: Yubico
Serial number ....: xxxxxxxx
Name of cardholder: Uri Blumenthal
Language prefs ...: en
Sex ..............: male
URL of public key : [not set]
Login data .......: uri
Signature PIN ....: forced
Key attributes ...: 2048R 2048R 2048R
Max. PIN lengths .: 127 127 127
PIN retry counter : 10 10 10
Signature counter : 11
Signature key ....: 0EB9 0C09 F70A 8D69 4338  AD9B A068 AD0D 7D7B DC2C
      created ....: 2015-09-01 02:23:54
Encryption key....: 5241 5DEF 8A73 6E73 E660  EE60 E18F 852E E8FC B143
      created ....: 2015-02-12 09:59:21
Authentication key: C888 C3CD D0E6 F8A7 ADA3  99FF F193 66B0 FA43 FF7D
      created ....: 2015-02-12 10:00:04
General key info..: pub  2048R/0xA068AD0D7D7BDC2C 2015-09-01 Mobile Mouse 4 (Yubi) <mouse07410@hotmail.com<mailto:mouse07410@hotmail.com>>
sec   2048R/0x684AC1FCEAA16B87  created: 2015-02-12  expires: 2018-02-11
ssb>  2048R/0xE18F852EE8FCB143  created: 2015-02-12  expires: 2018-02-11
                      card-no: 0006 03643907
ssb>  2048R/0xF19366B0FA43FF7D  created: 2015-02-12  expires: 2018-02-11
                      card-no: 0006 03643907
ssb>  2048R/0xA068AD0D7D7BDC2C  created: 2015-09-01  expires: 2018-08-31
                      card-no: 0006 03643907



On Sep 12, 2015, at 13:34 , Frank Morgner <notifications@github.com<mailto:notifications@github.com>> wrote:


I tried to sum up the discussion from #507<https://github.com/OpenSC/OpenSC/pull/507> in this PR.

I don't have a NEO. @dengert<https://github.com/dengert>, @rqou<https://github.com/rqou> @mouse07410<https://github.com/mouse07410> could you check with your NEOs whether:

  1.  the NEO defaults to the PIV application
  2.  You can force the PGP application via opensc.conf as discussed in #507<https://github.com/OpenSC/OpenSC/pull/507>
  3.  the PGP application works as expected

Thank you!

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/OpenSC/OpenSC/pull/551

Commit Summary

  *   Fix OpenPGP driver to work correctly with YubiKey NEO
  *   openpgp: match application, not ATR

File Changes

  *   M src/libopensc/card-openpgp.c<https://github.com/OpenSC/OpenSC/pull/551/files#diff-0> (54)
  *   M src/libopensc/ctx.c<https://github.com/OpenSC/OpenSC/pull/551/files#diff-1> (2)

Patch Links:

  *   https://github.com/OpenSC/OpenSC/pull/551.patch
  *   https://github.com/OpenSC/OpenSC/pull/551.diff

—
Reply to this email directly or view it on GitHub<https://github.com/OpenSC/OpenSC/pull/551>.

--
Uri Blumenthal

frankmorgner:
Great, thanks! I'll wait for some more comments, though.

dengert:
Working on it...

On 9/13/2015 1:59 AM, Frank Morgner wrote:
> Great, thanks! I'll wait for some more comments, though.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/551#issuecomment-139847987>.
>

-- 

dengert:
The NEO using the PIV applet continues to work as expected with these mods.

dengert:
While trying the NEO on a NFC reader, ATR is:
$ opensc-tool -a
Using reader with a card: ACS ACR122U PICC Interface 00 00
3b:8c:80:01:59:75:62:69:6b:65:79:4e:45:4f:72:33:58

While plugging NEO in as USB ATR is:
$ opensc-tool -a
Using reader with a card: Yubico Yubikey NEO OTP+CCID 00 00
3b:fc:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:4e:45:4f:72:33:e1

This may not be an issue, but these mods do add USB ATR but not NFC ATR to card-openpgp.c.

The NEO PIV supports verify and crypto over NFC, a violation of NIST specifications for PIV. Not clear if OpenPGP will work over NFC, or if there are security issues.

frankmorgner:
well, transmit of an unprotected PIN via NFC  is always a problem, for PIV and for PGP

frankmorgner:
Feel free, to add the contact-less ATR in a PR.

dengert:
I don't think the ATR is needed if the OpenPGP code because it now test for AIDs. The patch added the NEO ATR for some reason.
Someone with multiple OpenPGP cards could test if the ATRs are still needed, as older cards may not
have an AID on the card.

Microsoft does not use ATRs for their drivers such as the PIV. I think they look for AIDs.
For non windows drivers, I think the registry still needs ATRs and/or ATR masks to use the OpenSC minidriver.
There may be a way to get around this, but I have not looked for it.

On 9/13/2015 6:35 PM, Frank Morgner wrote:
> Feel free, to add the contact-less ATR in a PR.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/551#issuecomment-139928505>.
>

-- 

frankmorgner:
The ATR in the PGP driver is used as shortcut and to set the name of the card.

dengert:
It does more then set the name, pgp_match_card returns 1, without even checking if there is as AID.



  290 pgp_match_card(sc_card_t *card)
  291 {
  292         int i;
  293
  294         i = _sc_match_atr(card, pgp_atrs, &card->type);
  295         if (i >= 0) {
  296                 card->name = pgp_atrs[i].name;
  297                 return 1;

Following does not get executed...

  298         } else {
  299                 sc_path_t       partial_aid;
  300                 unsigned char aid[16];
  301
  302                 /* select application "OpenPGP" */


Only checking the ATR was the original problem. The real fix in this patch was the reordering of the drivers in ctx.c
so the openpgp did not grab the card before other drivers (i.e. piv) could see if the card had a AID for a different application.
The patch does will allow for new cards to have the openpgp applet without having to add the ATR.
The NEO should be treated as one of these new cards, and not have its ATR added to the code.

You can leave the patch as is, the point I am trying to make is this patch still uses the ATR for the NEO.


On 9/14/2015 2:18 AM, Frank Morgner wrote:
> The ATR in the PGP driver is used as shortcut and to set the name of the card.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/551#issuecomment-139982334>.
>

-- 

dengert:
Also see #391, as it should work with out adding the ATR, and the NEO should be treated the same way, anf the NEO ATR not added either.

frankmorgner:
the pgp application does get selected in the pgp's init function anyway, that's why the neo doesn't need it anymore in the match function.

dengert:
You are not getting my point. I am not saying the OpenPGP does not work, I am saying the openpgp driver interferes with
some other application being selected when the card does not have an OpenPGP application.

If we were to add some other driver that could use the NEO and added it in ctx.c
after the openpgp driver, the new driver would never get selected, even if the NEO did not have the openpgp applet.

card.c line 260 to 292  call ops->match_card(card) and r = ops->init(card) loking at all drivers.
But only  if match_card(card) != 1 and init(card) returns SC_ERROR_INVALID_CARD will another card driver be tested.
If init(card) fails the card is disconnected.

Looking at  card-openpgp.c line 350, pgp_init calls iso_ops->select_file and line 352 returns r, which
I do not believe is SC_ERROR_INVALID_CARD. The code in pgp_match_card that tries to the select the AID
if it fails, will have pgp_match_card return 0, so other drivers can look at the card.

So the best place to do the AID check is in the match_card which should be doing only what is needed to verify this driver
is the driver that should be using this card. The init function could fail for a number of reasons.

It also looks like the code on pgp_match_card is much more robust, setting the card->type from the AID.
and not requiring the ATR.

I would like to see some developers with OpenPGP cards to remove these lines, and let pgp_match_card query the AID:

  294         i = _sc_match_atr(card, pgp_atrs, &card->type);
  295         if (i >= 0) {
  296                 card->name = pgp_atrs[i].name;
  297                 return 1;
  298         } else

If card name is important, and ATRs are still needed, the above code could still do the _match_card, and even if it works,
do the AID matching.

OpenSC should look closet at ISO7816-4 section 9 "Application independent Card Services" to select an application.
Only for non compliant cards, with single applications should the ATR be used.

On 9/14/2015 10:48 AM, Frank Morgner wrote:
> the pgp application does get selected in the pgp's init function anyway, that's why the neo doesn't need it anymore in the match function.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/551#issuecomment-140121898>.
>

-- 

frankmorgner:
AFAIK, all neo's are produced with the PGP applet on board. AFAIK, the neo's ATR is unique and does not interfere with other cards. In consequence, it's reasonable to assume that a card with this ATR supports PGP. (Whether that's ISO compliant or not is on a different page. Most vendors don't even seem to be very committed to the ISO standard...) So, what the pgp driver does is not elegant, but it's OK.

Imagine the problem you described with a PGP driver that only recognizes the AID (a) and with a PGP driver that only recognizes the ATR (b). What would happen to the neo?

1. neo gets inserted
2. `sc_connect_card` iterates through card drivers `match_card`
3. `sc_connect_card` breaks when *the first* driver returns 1 for `match_card`
4. card is returned initialized with exactly this driver.

You may have noticed that it doesn't matter whether `match_card` returns 1 because of an ATR or because of an AID. The real problem here, is that only the first matching driver gets valued. All the others are ignored. This kind of "interference" with an other card driver as you say is *intrinsic* to OpenSC's card handling, it is *not* a bad design decision within the PGP driver.

Solving this intrinsic problem is what we need to do as future work. And this is also what I tried to point out in #538. In PKCS\#11, for example, one solution could be to *not ignore the rest of the card drivers* during the matching, but to instead create slots for all card drivers that are usable, while internally mapping each slot to the correct card/card driver.

dengert:
I am not talking two PGP drivers. I am talking about the card-openpgp.c driver that does not do a very good job, and some future driver
for some other applet not getting called because of the code in card-openpgp.c.

Calling the match_card of the piv driver before the match in the the openpgp to avoid the issues in card-openpgp was my main concern, and you fixed that part.
A user can also select the application by using opensc.conf.

#538 could work if the card can support two applications possibly being used at the same time.

I am not getting anywhere with this discussion. The problem will show up again.


On 9/14/2015 3:08 PM, Frank Morgner wrote:
> AFAIK, all neo's are produced with the PGP applet on board. AFAIK, the neo's ATR is unique and does not interfere with other cards. In consequence, it's reasonable to assume that a card with this ATR
> supports PGP. (Whether that's ISO compliant or not is on a different page. Most vendors don't even seem to be very committed to the ISO standard...) So, what the pgp driver does is not elegant, but
> it's OK.
>
> Imagine the problem you described with a PGP driver that only recognizes the AID (a) and with a PGP driver that only recognizes the ATR (b). What would happen to the neo?
>
>  1. neo gets inserted
>  2. |sc_connect_card| iterates through card drivers |match_card|
>  3. |sc_connect_card| breaks when /the first/ driver returns 1 for |match_card|
>  4. card is returned initialized with exactly this driver.
>
> You may have noticed that it doesn't matter whether |match_card| returns 1 because of an ATR or because of an AID. The real problem here, is that only the first matching driver gets valued. All the
> others are ignored. This kind of "interference" with an other card driver as you say is /intrinsic/ to OpenSC's card handling, it is /not/ a bad design decision within the PGP driver.
>
> Solving this intrinsic problem is what we need to do as future work. And this is also what I tried to point out in #538 <https://github.com/OpenSC/OpenSC/issues/538>. In PKCS#11, for example, one
> solution could be to /not ignore the rest of the card drivers/ during the matching, but to instead create slots for all card drivers that are usable, while internally mapping each slot to the correct
> card/card driver.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/551#issuecomment-140191922>.
>

-- 

frankmorgner:
I rebased to resolve the conflicts, if there are no objections I'll merge the PR.

philipWendland:
I did not follow this discussion with full concentration, but a simple
comment:

AFAIK the first neos had default ISD keys, or at least the possibility
to install/delete applets. So it could be possible that there are neos
out there with no PGP applet.

mouse07410:
Only development (and early release) NEOs had the ability to manage (install/delete) applets. Current/production versions do not allow the user that capability.

For practical purposes, I think we can safely ignore those early devices, and assume that all the relevant applets are present/installed. Plus, there is no reason (that I see) to delete an applet (rather than just letting it be there, unused) - it's not like it would occupy memory needed for something else. So in the worst case, a user who does not need a PGP applet, would have to install it. I could live with it, if it were me. :-)

Sent from my iPad

On Sep 16, 2015, at 05:49, Philip Wendland <notifications@github.com<mailto:notifications@github.com>> wrote:

I did not follow this discussion with full concentration, but a simple
comment:

AFAIK the first neos had default ISD keys, or at least the possibility
to install/delete applets. So it could be possible that there are neos
out there with no PGP applet.

Issue #557, opened 2015-09-18
The ECDSA signature format again


philipWendland:
Hello,

thanks in advance to everyone looking at this issue.

We had trouble with the format of ECDSA signatures earlier this year. I don't know what changed, but the way it is now, it is wrong.

When converting from R|S to the ASN1 sequence of R,S, we need to strip (or add) excess zeroes.
See the following test case:
```
Outgoing APDU data [  260 bytes] =====================================
10 2A 9E 9A FF 36 74 76 32 42 32 34 6A 69 4C 4C .*...6tv2B24jiLL
31 65 63 7A 47 77 6D 6C 59 37 47 30 74 51 48 67 1eczGwmlY7G0tQHg
4E 51 52 4B 43 38 66 79 66 4F 44 39 59 43 43 59 NQRKC8fyfOD9YCCY
59 42 62 43 76 4C 31 39 4C 67 44 6F 77 34 75 4B YBbCvL19LgDow4uK
53 4F 70 50 34 50 6E 57 43 39 6D 78 42 65 56 34 SOpP4PnWC9mxBeV4
38 6C 43 34 31 72 42 66 6F 45 45 5A 49 62 37 66 8lC41rBfoEEZIb7f
79 57 42 6C 76 4B 54 57 50 67 31 4D 6E 4D 34 6B yWBlvKTWPg1MnM4k
69 6C 51 54 59 34 6A 68 4B 42 4D 32 48 74 53 30 ilQTY4jhKBM2HtS0
68 69 46 4A 4E 4E 6D 54 56 64 62 44 58 6C 41 68 hiFJNNmTVdbDXlAh
67 77 4C 31 63 4F 4B 4A 48 79 61 37 64 30 64 78 gwL1cOKJHya7d0dx
53 32 36 30 74 66 79 6E 35 61 61 4C 6F 6E 58 79 S260tfyn5aaLonXy
69 69 58 30 73 71 54 35 51 6E 73 48 48 4E 39 30 iiX0sqT5QnsHHN90
6C 65 4E 32 79 76 37 4B 66 50 32 48 5A 38 4A 38 leN2yv7KfP2HZ8J8
57 67 5A 6A 7A 41 66 6A 77 33 78 43 5A 4E 31 43 WgZjzAfjw3xCZN1C
32 7A 54 41 53 6A 74 65 49 51 33 47 64 6A 4B 79 2zTASjteIQ3GdjKy
[...]

PR #559, opened 2015-09-20
ECDSA helper functions: strip zeroes when converting from R,S to sequence

philipWendland:
For ECDSA signatures, there are multiple ways to format the signature:
- R|S (R and S filled with zeroes at the most significant bytes)
- ASN1 sequence of R,S integers (e.g. used by OpenSSL).

It is rare that the filling with zeroes is needed.
But if it is, in the second case, the filling zeroes should not be there
or the verification of the signature by OpenSSL will fail.

See issue #557.


frankmorgner:
@dengert Could you review/verify this commit?

dengert:
Looks good, OK to commit.

frankmorgner:
thanks

PR #560, opened 2015-09-22
Removed error check to support card with PKCS#15 emulation but no mat…

CardContact:
This patch enables support for cards that use PKCS15 emulation but do not maintain an EF.DIR (notably the SmartCard-HSM). For those cards, the PKCS15 AID has no matching entry found by sc_find_app().

This patch removes the error exit, so that the default profile AID is used.

The problem can be observed in minidriver, which always passes the aid parameter taken from p15card.

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -727,10 +727,6 @@ sc_pkcs15init_finalize_profile(struct sc_card *card, struct sc_profile *profile,
 	if (aid)   {
 		sc_log(ctx, "finalize profile for AID %s", sc_dump_hex(aid->value, aid->len));
 		app = sc_find_app(card, aid);
-		if (!app)   {
-			sc_log(ctx, "Cannot find oncard application");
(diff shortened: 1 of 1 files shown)


frankmorgner:
Isn't this a problem for all applets that are *not* selected by default and hence cannot provide EF.DIR? @philipWendland did you stumble over this, too, with the IsoApplet?

frankmorgner:
Please use a shorter subject in the commit, see http://chris.beams.io/posts/git-commit/

frankmorgner:
... and add the description of this PR to the body of the commit...

philipWendland:
> Isn't this a problem for all applets that are not selected by default and hence cannot provide EF.DIR? @philipWendland did you stumble over this, too, with the IsoApplet?

Not yet. But IsoApplet does not have PKCS#15 emulation. And it has an EF(DIR) written to the file system of the applet, so this issue might not apply.

frankmorgner:
I see.

I think the PR is good to be merged if @CardContact fixes the commit message.

frankmorgner:
thanks

PR #563, opened 2015-09-23
Fix reselect applet for pin verification

CardContact:
This patch is required to ensure that the SmartCard-HSM applet is reselected after a reset and before the PIN is verified. This is required by minidriver.

The patch for pkcs15-pin.c fixes a bug that would skip sc_select_file() if only the AID is defined in auth_info->path.

### DIFF
--- a/src/libopensc/pkcs15-pin.c
+++ b/src/libopensc/pkcs15-pin.c
@@ -402,7 +402,7 @@ int sc_pkcs15_change_pin(struct sc_pkcs15_card *p15card,
 	r = sc_lock(card);
 	LOG_TEST_RET(ctx, r, "sc_lock() failed");
 	/* the path in the pin object is optional */
-	if (auth_info->path.len > 0) {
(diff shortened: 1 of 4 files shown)


src/libopensc/pkcs15-pin.c
```diff
@@ -358,7 +358,7 @@ int sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card,
 	LOG_TEST_RET(ctx, r, "sc_lock() failed");
 
 	/* the path in the pin object is optional */
-	if (auth_info->path.len > 0) {
+	if ((auth_info->path.len > 0) || ((auth_info->path.aid.len > 0))) {
 		r = sc_select_file(card, &auth_info->path, NULL);
```
frankmorgner: This should be changed in `sc_pkcs15_change_pin` and `sc_pkcs15_unblock_pin`, too

CardContact: Done

PR #565, opened 2015-09-25
Build a lightweight version of OpenSC

frankmorgner:
... without dependencies to external libraries. AppVeyor now also builds an installer without OpenSSL and zlib.


vletoux:
I've checked the Windows installation part (wix). OK for me.

frankmorgner:
thanks for the feedback

vletoux:
@frankmorgner : have you seen that AppVeyor is complaining in x64 built about missing symbols ?
compression.obj : error LNK2001: unresolved external symbol uncompress

see the ECC minidriver pull request & AppVeyor compilation about that

frankmorgner:
@vletoux, yes, I have. zlib's pre-built library is 32bit thats why it won't build on 64 bit, see https://github.com/OpenSC/OpenSC/blob/master/appveyor.yml#L19-L20. To improve the situation, we could be a bit more sophisticated on 64 bit by *not* adding zlib...

frankmorgner:
I'll push https://github.com/frankmorgner/OpenSC/commit/b7f794ae48c6be2e9e905b1bd1442a4f022f8d47 as soon as I get a positive feedback from AppVeyor.

vletoux:
Better, but "nmake" is not available for VS=10 in x64 (ok for vs=12)

see https://ci.appveyor.com/project/LudovicRousseau/opensc/build/0.15.0.227
(last build from PR 566)

vincent

2015-10-04 14:07 GMT+02:00 Frank Morgner <notifications@github.com>:

> I'll push frankmorgner@b7f794a
> <https://github.com/frankmorgner/OpenSC/commit/b7f794ae48c6be2e9e905b1bd1442a4f022f8d47>
> as soon as I get a positive feedback from AppVeyor.
>
> —
> <https://github.com/OpenSC/OpenSC/pull/565#issuecomment-145340036>.
>



-- 
---
Vincent

frankmorgner:
The best we can do here, is to allow this build to fail (which it already is). Unfortunately there is no possibility to exclude a certain build combination in AppVeyor.

PR #569, opened 2015-10-01
fix: change SC_TERMINATE (undefined) to SC_CTX_FLAG_TERMINATE

mdealencar:
This file was not compiling because SC_TERMINATE is not defined anywhere. It seems like the intended expression is what I propose.


frankmorgner:
thanks

PR #571, opened 2015-10-04
Fixes accessing fixed size arrays

frankmorgner:
The pkcs_15_object_t label is a fixed size array, which means that it is possibly not terminated with `\0`. This PR fixes potentially bad access to the memory.


frankmorgner:
i also fixed the same problem in sc-hsm's implementation of CVCs

frankmorgner:
i also fixed the same problem in the minidriver

PR #577, opened 2015-10-14
Fixes build and installation on OS X 10.11

frankmorgner:


metsma:
There is still one copy of headers available in original Xcode 7 install
/Applications/Xcode.app//Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdk/MacOSX.sdk/usr/include/openssl

PR #579, opened 2015-10-14
Do not expose access to PIV emulated files from utilities

dengert:
PIV cards uses get/put data not select file and read_binary.
To allow access via pkcs15 emulation card-piv.c emulates
select_file and read_binary but only when used with the path as
created by the piv emulation.

There are no MF.DIR or EF.DIR files.
opensc-tool and opensc-explorer will not work with this emulation.

Patch removes code that caused problems with opensc-tooland opensc-explorer.

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2440,15 +2440,16 @@ static int piv_select_file(sc_card_t *card, const sc_path_t *in_path,
 	pathlen = in_path->len;
 
 	/* only support single EF in current application */
+	/* 
+	 * PIV emulates files, and only does so becauses sc_pkcs15_* uses
+	 * select_file and read_binary. The emulation adds path emulated structures
+	 * so piv_select_file will find it.
+	 * there is no dir. Only direct access to emulated files
+	 * thus opensc-tool and opensc-explorer can not read the emulated files
+	 */
 
 	if (memcmp(path, "\x3F\x00", 2) == 0) {
-		if (pathlen == 2)   {
-			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, NULL, NULL);
-			SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "Cannot select PIV AID");
-
-			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
-		}
-		else if (pathlen > 2) {
+		if (pathlen > 2) {
 			path += 2;
 			pathlen -= 2;
 		}


src/libopensc/card-piv.c
```diff
@@ -2440,15 +2440,16 @@ static int piv_select_file(sc_card_t *card, const sc_path_t *in_path,
...
 
 	if (memcmp(path, "\x3F\x00", 2) == 0) {
-		if (pathlen == 2)   {
-			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, NULL, NULL);
-			SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "Cannot select PIV AID");
-
-			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
-		}
-		else if (pathlen > 2) {
+		if (pathlen > 2) {
```
frankmorgner: The code you remove *should* work. What was the problem?

The change you introduced will *always* remove the first FID. This will not work with a path relative to the current DF.

dengert: The problem was the SC_TEST_RET would return 0 without filling in *file which caused the segfault later in
opensc-tool. In the comments above the removed code:

         /* only support single EF in current application */
         /*
          * PIV emulates files, and only does so because sc_pkcs15_* uses
          * select_file and read_binary. The emulation adds path emulated structures
          * so piv_select_file will find it.
          * there is no dir. Only direct access to emulated files
          * thus opensc-tool and opensc-explorer can not read the emulated files
          */

Only selected paths are supported as defined in the code. piv_find_obj_by_containerid called below the code does this.

The NIST 800-73-3 part 2 defines a subset of ISO 7816-4 commands to be used. It makes no assumption
that any other commands need to be supported. It has "SELECT" command which happens to be 00 A4 04 00 Lc AID Le
All reading of data from the card is done using "GET DATA"

The PIV driver was written to stick to the standards as close as possible and not assume anything else
about the card and present a PKCS#11 interface. (Because of the way OpenSC uses PKCS#15 some of the sc_pkcs15
routines also work but not all.)



On 10/16/2015 4:44 AM, Frank Morgner wrote:
> In src/libopensc/card-piv.c <https://github.com/OpenSC/OpenSC/pull/579#discussion_r42225468>:
>
>>
>>  	if (memcmp(path, "\x3F\x00", 2) == 0) {
>> -		if (pathlen == 2)   {
>> -			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, NULL, NULL);
>> -			SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "Cannot select PIV AID");
>> -
>> -			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
>> -		}
>> -		else if (pathlen > 2) {
>> +		if (pathlen > 2) {
>
> The code you remove /should/ work. What was the problem?
>
> The change you introduced will /always/ remove the first FID. This will not work with a path relative to the current DF.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/579/files#r42225468>.
>

-- 

dengert: Previous reply said SC_TET_RET, its the SC_FUNC_RETURN  that returns 0 and without setting *file

On 10/16/2015 4:44 AM, Frank Morgner wrote:
> In src/libopensc/card-piv.c <https://github.com/OpenSC/OpenSC/pull/579#discussion_r42225468>:
>
>>
>>  	if (memcmp(path, "\x3F\x00", 2) == 0) {
>> -		if (pathlen == 2)   {
>> -			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, NULL, NULL);
>> -			SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "Cannot select PIV AID");
>> -
>> -			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
>> -		}
>> -		else if (pathlen > 2) {
>> +		if (pathlen > 2) {
>
> The code you remove /should/ work. What was the problem?
>
> The change you introduced will /always/ remove the first FID. This will not work with a path relative to the current DF.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/579/files#r42225468>.
>

-- 

frankmorgner: Sorry, I misread the diff. I thought `if (memcmp(path, "\x3F\x00", 2) == 0) {` was removed, too. However, this still leaves the open question what happens if the MF shall be selected. This should give an error because `piv_find_obj_by_containerid` won't find anything. Is that intended? Wouldn't it be better if you'd leave the select for the applet and just add initialization of the `file_out` object?

dengert: piv_find_obj_by_containerid returns -1 if the path is not knownand
piv_select_file returns SC_ERROR_FILE_NOT_FOUND. Does it need a different return code?
 
2458         i = piv_find_obj_by_containerid(card, path);
2459 
2460         if (i < 0)
2461                 SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_FILE_NOT_FOUND);

frankmorgner: *If* you want to return an error if MF is selected, then your changes are totally OK. I just remarked that the code previously tried to catch MF while *not* returning an error. This is a small change in behavior that may have unexpected results.

If you want to fix the NULL dereference but keep the current semantics, you simply need to handle the `file_out` object.

dengert: The previous code should not have tried to catch the MF. It should return an error. I do not want to keep the previous
semantics.

The card-piv.c emulates the select_file and read_binary and only works with paths defined in
card-piv.c as static const struct piv_object piv_objects[] to map a path to an object on the card
Get DATA is then used to read the object.

pkcs15-piv.c also has these paths. There are set when creating the emulated pkcs15 structures.
It may add CECE for certificates, as the certificate is contained  in a certificate object,
(along with other flags, and the certificate my be zipped to.) The pkcs15_pubkey is obtained
from the certificate. pkcs15_privkey type, size and parameters are obtained from the public key.

The values used for these paths are defined in NIST 800-73-3, as a hold over from previous federal smart cards,
but there is no requirement for the card to support access using select and read_binary. GET_DATA
is used instead.

The above tries to stick to the NIST 800-73-3 standards as close as possible.
There are also some nonstandard objects in card-piv.c to represent keys:
  295 /* following not standard , to be used by piv-tool only for testing */


*BUT* in the future if the standards change, or there is some compelling reason to extend the functionality
this code could be extended.



On 10/16/2015 12:36 PM, Frank Morgner wrote:
> In src/libopensc/card-piv.c <https://github.com/OpenSC/OpenSC/pull/579#discussion_r42268601>:
>
>>
>>  	if (memcmp(path, "\x3F\x00", 2) == 0) {
>> -		if (pathlen == 2)   {
>> -			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, NULL, NULL);
>> -			SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "Cannot select PIV AID");
>> -
>> -			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
>> -		}
>> -		else if (pathlen > 2) {
>> +		if (pathlen > 2) {
>
> /If/ you want to return an error if MF is selected, then your changes are totally OK. I just remarked that the code previously tried to catch MF while /not/ returning an error. This is a small change
> in behavior that may have unexpected results.
>
> If you want to fix the NULL dereference but keep the current semantics, you simply need to handle the |file_out| object.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/579/files#r42268601>.
>

-- 

frankmorgner:
OK, thanks for the explanation. I bet we will be talking about this again, when we try to correctly include a card with multiple applications...

dengert:
Could be. I would think each application has its own driver. And we already do, NEO has OpenPGP and PIV.


On 10/17/2015 10:34 AM, Frank Morgner wrote:
> OK, thanks for the explanation. I bet we will be talking about this again, when we try to correctly include a card with multiple applications...
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/579#issuecomment-148925184>.
>

-- 

dengert:
Thanks.

On 10/17/2015 10:34 AM, Frank Morgner wrote:
> Merged #579 <https://github.com/OpenSC/OpenSC/pull/579>.
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/579#event-438253826>.
>

-- 

PR #580, opened 2015-10-14
Fix issues related to selecting a file

frankmorgner:
The card operation `select_file` may return an output file on request, but `sc_select_file` is now guaranteed to return it with `SC_SUCCESS`. To achieve this, I did the following:

1. remove the file parameter from `sc_select_file` or `select_file` if it was not needed to simplify things
2. if the result was actually used after a `select_file` call then I added a check if a correct file object was returned
3. initialize the file object with `NULL` and check for it in `sc_select_file` so that all high-level functions using `sc_select_file` can rely on the initialized object.

Actually I'd prefer delegating the checking to the piece of code that actually accesses some memory, but this would have been way too much work for me...

### DIFF
--- a/src/libopensc/card-entersafe.c
+++ b/src/libopensc/card-entersafe.c
@@ -468,7 +468,7 @@ static int entersafe_select_fid(sc_card_t *card,
 	SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, r, "APDU transmit failed");
 
 	/* update cache */
-	if (file->type == SC_FILE_TYPE_DF) {
+	if (file && file->type == SC_FILE_TYPE_DF) {
 		 card->cache.current_path.type = SC_PATH_TYPE_PATH;
 		 card->cache.current_path.value[0] = 0x3f;
 		 card->cache.current_path.value[1] = 0x00;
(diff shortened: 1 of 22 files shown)


frankmorgner:
@mouse07410 did you find the time for testing if this fixes the crash with `opensc-tool -f`?

mouse07410:
Oh, sorry - yes I did. And yes it did cure the problem:

```
$ opensc-tool -f
SELECT FILE failed: File not found
$ opensc-tool -vf
Connecting to card in reader Yubico Yubikey NEO OTP+U2F+CCID...
Using card driver PIV-II  for multiple cards.
SELECT FILE failed: File not found
$ opensc-tool --list-algorithms
Algorithm: rsa
Key length: 1024
Flags: padding ( none ) hashes ( )

Algorithm: rsa
Key length: 2048
Flags: padding ( none ) hashes ( )

Algorithm: rsa
Key length: 3072
Flags: padding ( none ) hashes ( )

Algorithm: ec
Key length: 256
Flags:

Algorithm: ec
Key length: 384
Flags:

$
```

Thanks!

frankmorgner:
thanks for testing

PR #584, opened 2015-10-15
Card openpgp cleanup2

marschap:
New version of this morning's pull request addressing the comments.

Promised rewrite of getting historical bytes in pgp_get_card_features() not included: I need more tests.


frankmorgner:
Looks OK now. I still leave this open for review. Maybe @hongquan want's to have a look...

As a side note, I wonder if the special treatment of the Gnuk token is needed at all or if it can be derived from the 2.0 specification...

hongquan:
I appreciate this cleanup. I agree to merge.

Gnuk token has less feature than 2.0 spec:

- Always use short APDU
- Support only 1 key length (2048)
- 1 more thing that I forget.

So, we cannot derive from 2.0 spec, or we may mistakenly use long APDU for it or tell it to generate 1024-bit key...

frankmorgner:
Is this limitation not included in the gnuk metadata?

frankmorgner:
@hongquan did you consider to only disable announcing extended length support for gnuk here https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L462 and here https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L479 ? I think the rest of the code should respect the card's limitation.

hongquan:
@frankmorgner Sorry I lost the documentation about Gnuk's metadata. But I remember that, for Gnuk, we cannot base on its metadata for APDU length & command chaining.

In the code comment, I also mention some difference of Gnuk, such as: command length is always even, different way of deleting certificate DO...

Your idea to disable announcing extended length support in `pgp_get_card_features` is great. We can improve this (and always enable "command chaining").

frankmorgner:
OK, I hope you find the time for this clean up soon!

PR #591, opened 2015-10-23
isoApplet: fix card not recognized by minidriver

vletoux:


vletoux:
@philipWendland Here is the PR to fix the isoApplet profile

Needs to recreate every card to take profit of this fix

the minidriver initialization fails at this place:
https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L375-L379
Then
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15.c#L1726-L1727

Indeed the flag requested is SC_PKCS15_PIN_FLAG_INITIALIZED

philipWendland:
Thanks, I can see how this solves the problem.

But I am not quite sure what this PIN flag actually means and *when* it should be set.
PKCS#15 says about this flag: "is initialized, meaning that the PIN has been initialized;"
When the IsoApplet is freshly installed, the PIN is *not* initialized, it needs to be set first. So the flag in the PKCS#15 structure should be set when actually setting the PIN value (pkcs15-init --create-pkcs15). The question that remains for me is: Is this the right place to set this flag or might there be a general problem in OpenSC with this flag (i.e., should this flag be set automatically when the PIN is written, and not statically in a profile file)?

vletoux:
I've checked the differences with SC-HSM
Indeed SC-HSM has in its PIN info containing the applet AID while there is no auth_info->path.aid for the isoApplet
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-pin.c#L361-L367

As a consequence, when the Base CSP/KSP calls the minidriver CardAuthenticatePin for the second time after the card has been reseted, the applet is not selected again.
The problem is hidden in PKCS#11 because the applet is select at the beginning of all transactions.

I'm not an applet specialist but I think the problem is located in the PKCS15 formating. Right ?
Where the problem should be fixed ? In the code or in the profile definition ?

vletoux:
@dengert I see in iso7816-15 chap 8.9.2 that a PIN contains a password attributes and within it an optional path. This field is mapped into pin->auth_info->path
isoApplet doesn't resist to a reset of the card because the applet needs to be reselected after a reset and because pin->auth_info->path is not set (see my previous post).

How should this attribute be set ?
I looked into the profile (profile.c) and I didn't find which keyword can set the path.aid
I checked the pkcs15 functions (select_pin_reference / create_pin) but without any things related.

Where the aid related to the PIN should be set for the isoApplet ?
- in the profile ?
- in the pkcs15 functions (select_pin_reference / create_pin) ?
- in the card driver to overwrite the object after an initialization ?

philipWendland:
The IsoApplet provides an ISO-7816-like interface to OpenSC. OpenSC will 
write the PKCS#15 file structure to the applet.
The PIN in IsoApplet has a global scope and is not stored in any file. 
So there is really no path attached to the PIN.

There is a difference between AIDs/paths on the smart card (AIDs of 
applets, e.g. isoApplet, which needs to be selected at the begining of 
all transactions and obviously after a card reset) and in the applet 
(PKCS#15 structure written by OpenSC to the applet).
I am saying this, because the DF(PKCS#15) has also an AID attached to it 
in the applet as far as I know.

I was always under the impression that OpenSC was originally not 
designed for cards which have applets. It was designed for ISO-7816 
cards which are matched using the ATR of the card, not the AID of the 
applet. The problem is, nowadays, the functionality of most smart cards 
is determined by the applet present (AID) *and selected*, but not the 
actual smart card (ATR).

I think we see the limitation of the current approach to solve this 
(selecting the applet in *_match_card()). The problem seems to be that, 
when fixing the minidriver de-authentication bug by resetting the card, 
the currently selected applet might not be selected anymore.
@vletoux Do you know whether this was considered when resetting the card?
Is this what auth_info->path.aid should be used for? (Having in mind 
that OpenSC was *not* designed for applets at the time)
It might work, instead of the isoApplet the ISD might catch the APDU and 
select the applet, I just don't know whether this is really the way it 
is supposed to be (same goes with sc-hsm and other applets of course).

Does anybody else have the feeling that OpenSC might need some re-work 
in this section (complete and natural support of *applets*)?

kind regards,
Philip



Am 10/24/2015 um 1:43 PM schrieb vletoux:

dengert:
I am not sure why you are asking me. I don't have a copy of ISO7816-15.
But if path is to an AID, then that would cause the card to select the AID.

The main issue may be that the minidriver does the associate and disassociate. This was added
because the caller could change handles to a different card. Rather then associate and disassociate
maybe the minidriver should keep a list of the handles and related data that are still active,
similar to slots in PKCS#11. Associate and disassociate are over kill.

On a side note...

My expertise has been in PIV. PIV specs say the PIV AID must be the default application,
so a select of the AID has not been a problem for PIV cards.

But it appears some CAC cards have the PIV applet but the PIV is not the default. See #570 #574 ans #579.
I will look at the PIV emulation code again and see if and see if the AID should be added.

(PIV defines the APDUs and status that are requited. I have not seen any requirement to support other ISO 7816
requirements, although all the defined APDUs are ISO 7816.)


On 10/24/2015 6:43 AM, vletoux wrote:
> @dengert <https://github.com/dengert> I see in iso7816-15 chap 8.9.2 that a PIN contains a password attributes and within it an optional path. This field is mapped into pin->auth_info->path
> isoApplet doesn't resist to a reset of the card because the applet needs to be reselected after a reset and because pin->auth_info->path is not set (see my previous post).
>
> How should this attribute be set ?
> I looked into the profile (profile.c) and I didn't find which keyword can set the path.aid
> I checked the pkcs15 functions (select_pin_reference / create_pin) but without any things related.
>
> Where the aid related to the PIN should be set for the isoApplet ?
>
>   * in the profile ?
>   * in the pkcs15 functions (select_pin_reference / create_pin) ?
>   * in the card driver to overwrite the object after an initialization ?
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/591#issuecomment-150796813>.
>

-- 

vletoux:
the reset is done by the Base CSP / KSP. There is nothing I can do except providing a deauthentication function which will avoids a card being reseted.
The problem is that as today, there is no such deauthentication function.

(the associate / disassociate is mandatory because when the screen go to the Ctrl+alt+del screen, the smart card handles are broken)

I see to possibility to fix this:

Lead 1:
implement a deauthentication function in OpenSC
(select the applet or a do VERIFY with if P1 = FF as described in iso7816)
=> the applet will not be unselected anymore (preferred)

Lead 2:
implement a "select applet if deselected"
For example, get the serial number of the applet, and if an error, reselect the applet
=> the applet will be reselected after a card reset

dengert:
I totally agree. Applets/AID are the way most newer cards should be selected. Also see #538

Examples are:

NEO supports PIV and OpenPGP applet. OpenSC has two different drivers and no good way to select which one the user wants.

PIV and CAC cards. There is a transition from CAC to PIV, and some cards have both applets.
OpenSC PIV driver was assuming only PIV cards. See  #570 #574 ans #579. which I need to look at closer.

On Windows, Microsoft supports PIV, and does so be selecting the AID,
then caching the ATRs in the registry.


Over the years my machine has seen these PIV cards. Some test, some production:

  [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\PIV Device ATR Cache]
"843e991e-05dc-4536-82b0-d0745acd7639"=hex:3b,db,96,00,81,b1,fe,45,1f,03,80,f9,\
   a0,00,00,03,08,00,00,10,00,18
"ed421ee5-d0d3-4297-ba30-56c5db6e5be8"=hex:3b,fa,13,00,00,81,31,fe,15,59,75,62,\
   69,6b,65,79,4e,45,4f,a6
"1a839b32-f79c-4db9-bdd2-faaec79ae6cf"=hex:3b,df,96,00,81,b1,fe,45,1f,83,80,\
   73,cc,91,cb,f9,a0,00,00,03,08,00,00,10,00,79
"a013fc65-49e0-4dec-a41d-1d01eadb7c0c"=hex:3b,7d,96,00,00,80,31,80,65,b0,83,11,\
   17,d6,83,00,90,00
"d376f72a-56a2-4f9e-a410-eefc8bb0f689"=hex:3b,fc,13,00,00,81,31,fe,15,59,75,\
   62,69,6b,65,79,4e,45,4f,72,33,e1
"7ebc4efd-7423-4860-9fd5-38efac67aefc"=hex:3b,8f,80,01,80,91,e1,31,80,65,b0,83,\
   11,17,d6,83,00,90,00,da
"dec792e7-c33a-43b2-b5a2-4389d2de78b4"=hex:3b,8f,80,01,80,73,cc,91,cb,f9,a0,00,\
   00,03,08,00,00,10,00,29
"2215468f-cf4e-4a37-867c-4822561ce88f"=hex:3b,8c,80,01,59,75,62,69,6b,65,79,4e,\
   45,4f,72,33,58

There are many more PIV card vendors then on the above list.


On 10/24/2015 7:08 AM, Philip Wendland wrote:
> The IsoApplet provides an ISO-7816-like interface to OpenSC. OpenSC will
> write the PKCS#15 file structure to the applet.
> The PIN in IsoApplet has a global scope and is not stored in any file.
> So there is really no path attached to the PIN.
>
> There is a difference between AIDs/paths on the smart card (AIDs of
> applets, e.g. isoApplet, which needs to be selected at the begining of
> all transactions and obviously after a card reset) and in the applet
> (PKCS#15 structure written by OpenSC to the applet).
> I am saying this, because the DF(PKCS#15) has also an AID attached to it
> in the applet as far as I know.
>
> I was always under the impression that OpenSC was originally not
> designed for cards which have applets. It was designed for ISO-7816
> cards which are matched using the ATR of the card, not the AID of the
> applet. The problem is, nowadays, the functionality of most smart cards
> is determined by the applet present (AID) *and selected*, but not the
> actual smart card (ATR).
>
> I think we see the limitation of the current approach to solve this
> (selecting the applet in *_match_card()). The problem seems to be that,
> when fixing the minidriver de-authentication bug by resetting the card,
> the currently selected applet might not be selected anymore.
> @vletoux Do you know whether this was considered when resetting the card?
> Is this what auth_info->path.aid should be used for? (Having in mind
> that OpenSC was *not* designed for applets at the time)
> It might work, instead of the isoApplet the ISD might catch the APDU and
> select the applet, I just don't know whether this is really the way it
> is supposed to be (same goes with sc-hsm and other applets of course).
>
> Does anybody else have the feeling that OpenSC might need some re-work
> in this section (complete and natural support of *applets*)?
>
> kind regards,
> Philip
>
>
>
> Am 10/24/2015 um 1:43 PM schrieb vletoux:
>  >
>  > @dengert <https://github.com/dengert> I see in iso7816-15 chap 8.9.2
>  > that a PIN contains a password attributes and within it an optional
>  > path. This field is mapped into pin->auth_info->path
>  > isoApplet doesn't resist to a reset of the card because the applet
>  > needs to be reselected after a reset and because pin->auth_info->path
>  > is not set (see my previous post).
>  >
>  > How should this attribute be set ?
>  > I looked into the profile (profile.c) and I didn't find which keyword
>  > can set the path.aid
>  > I checked the pkcs15 functions (select_pin_reference / create_pin) but
>  > without any things related.
>  >
>  > Where the aid related to the PIN should be set for the isoApplet ?
>  >
>  > * in the profile ?
>  > * in the pkcs15 functions (select_pin_reference / create_pin) ?
>  > * in the card driver to overwrite the object after an initialization ?
>  >
>  > —
>  > Reply to this email directly or view it on GitHub
>  > <https://github.com/OpenSC/OpenSC/pull/591#issuecomment-150796813>.
>  >
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/591#issuecomment-150803240>.
>

-- 

CardContact:
For the SmartCard-HSM we had to solve two issues:

1. The applet is only selected in pkcs15bind()
2. The PIN authentication state is reset when the applet is re-selected

The first one was solved by adding the applet AID to pin->auth_info->path, which ensures that the pkcs15 layer properly issues a select before PIN operations (even after a card reset). However this only worked after applying 83a28a1b, because the code did not properly handle aids in the path.

This created the second issue, as the PIN authentication in the applet is cleared on deselect and the pkcs15 layer also send a select before each crypto operation. This issue was traced down to the minidriver code, that passed an AID to the pkcs15 layer for each operation, rather than using the currently selected application.

I guess the core problem is something different: In the pkcs15 layer, you would usually have a sequence pkcs15_bind(), pkcs15_op()* and finally a pkcs15_unbind(). Resetting the card in between is not supported, an application must restart with pkcs15_bind().

Unfortunately pkcs15_bind() is an expensive operation with all the ATR, command probing and initialization.

I would suggest a solution where pkcs15 supports a deauthenticate function that resets the authentication state but leaves other on-card states (e.g. selected applet) untouched.

dengert:
You said: "do VERIFY with if P1 = FF as described in iso7816" I can't find that.
All the references I have say: "Only P1='00' is valid (other values are RFU)"



-- 

vletoux:
quote from ISO 7816-4:2013
verify
last paragraph:
verify2

dengert:
NIST 800-73-4 Part 2 May 2015 does say:
"If P1='FF', and Lc and the command data field are absent, the command shall reset the security status of the key reference in P2."
Sounds like a "logout" command.

I don't think there are any cards available today that support 800-73-4.


-- 

dengert:
OK, that matches what NIST 800-73-4 says.
I must have misread: "the reset is done by the Base CSP / KSP. There is nothing I can do except providing a deauthentication function which will avoids a card being reseted."

If you provide a deauthentication function, the Base CSP/KSP will not do a reset.

But many cards may not support this.

A quick way to test is use opensc-tool to do a verify For PIV: applet PIN P2=80, or Global PIN P2=00

on Linux using a Oberthur ID - One PIV (type A)

$ opensc-tool -s "00:20:00:80:08:31:32:33:34:35:36:FF:FF" \
 >  -s "00:20:FF:80"
Using reader with a card: SCM Microsystems Inc. SCR 355 [CCID Interface] 00 00
Sending: 00 20 00 80 08 31 32 33 34 35 36 FF FF
Received (SW1=0x90, SW2=0x00)
Sending: 00 20 FF 80
Received (SW1=0x6A, SW2=0x86)

$ opensc-tool -s "00:20:00:00:08:31:32:33:34:35:36:37:38"  -s "00:20:FF:00"
Using reader with a card: SCM Microsystems Inc. SCR 355 [CCID Interface] 00 00
Sending: 00 20 00 00 08 31 32 33 34 35 36 37 38
Received (SW1=0x90, SW2=0x00)
Sending: 00 20 FF 00
Received (SW1=0x6A, SW2=0x86)


A GemALto PIV 1.5.5DLv1

$ opensc-tool -s "00:20:00:80:08:31:32:33:34:35:36:FF:FF"  -s "00:20:FF:80"
Using reader with a card: SCM Microsystems Inc. SCR 355 [CCID Interface] 00 00
Sending: 00 20 00 80 08 31 32 33 34 35 36 FF FF
Received (SW1=0x90, SW2=0x00)
Sending: 00 20 FF 80
Received (SW1=0x6A, SW2=0x86)

On all the above 6A 86 is Incorrect parameter in P1"

With a NEO PIV 0.1.2:

$ opensc-tool -s "00:20:00:80:08:31:32:33:34:35:36:FF:FF"  -s "00:20:FF:80"
Using reader with a card: Yubico Yubikey NEO OTP+CCID 00 00
Sending: 00 20 00 80 08 31 32 33 34 35 36 FF FF
Received (SW1=0x90, SW2=0x00)
Sending: 00 20 FF 80
Received (SW1=0x6A, SW2=0x88)

6A 88 is Key reference not found. (looks like it should be 6A 86. The key ref of 80 worked with the PIN!)


On 10/24/2015 11:34 AM, vletoux wrote:
> quote from ISO 7816-4:2013
> verify <https://cloud.githubusercontent.com/assets/10632326/10711496/6cae2126-7a7d-11e5-8ec7-85433039353a.png>
> last paragraph:
> verify2 <https://cloud.githubusercontent.com/assets/10632326/10711497/707e2062-7a7d-11e5-8fd8-0a588baad07e.png>
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/591#issuecomment-150831827>.
>

-- 

vletoux:
Yes, having an optional deauthentication function can solve the problem for future drivers (isoApplet may implement this new function and fallback to the old behavior for older applet - sc-hsm can do it by selecting the applet).

All other cards I know use the default applet / are native so the applet reloading case is not a problem for them.

As a workaround, we (with @CardContact) remarked that reloading the applet before doing any pin authentication solves the problem. That's why setting the aid of the pin info can be a workaround, especially for the isoApplet like we did with sc-hsm.

philipWendland:
Thank you for the explanations and opinoins. I would prefer such a deauthentication function as well. When such a function is defined, I am happy to implement it for the IsoApplet (it would be a card reset with older versions, for newer versions I should implement the 'FF' VERIFY addition in the new version of 7816).

philipWendland:
To get back on the original topic of this PR, the question still remains:

Should the 'initialized' flag really be set in the profile, or in the PKCS#15 structure (i.e. common pkcs15 init code) when actually setting the PIN value for the first time?
Grepping for SC_PKCS15_PIN_FLAG_INITIALIZED shows that OpenSC does not really use this flag. 
Additionally, as the PIN is set when creating the PKCS#15 file structure, it does not make much difference for the IsoApplet.
If nobody else can imagine that this flag (or the evaluation thereof by OpenSC) can be useful in future, I am OK with merging this PR. In the other case, this might have to be dealt with differently. Thanks.

vletoux:
Not exactly on my opinion.
See: https://github.com/OpenSC/OpenSC/blob/c50a95133711922be994ef75879b0f2f6e1daafd/src/libopensc/pkcs15.h#L70-L85
https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L400-L405
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15.c#L1726-L1727

In the code I highlighted, you do a AND with SC_PKCS15_PIN_TYPE_FLAGS_MASK and then check for equality for SC_PKCS15_PIN_TYPE_FLAGS_PIN_GLOBAL, ...
=> the flag SC_PKCS15_PIN_FLAG_INITIALIZED is invisible in the code but if it is missing, the test for equally fails.

philipWendland:
Yes I know. The flag has to be in the PKCS#15 structure on the applet.

What I meant was, setting the flag statically in the profile like in this PR should do no harm because it would be set when creating the PKCS#15 structure anyway. 
I.e. OpenSC does nothing with this flag between the stages (1) completely uninitialized applet/smartcard (and an *uninitialized* PIN) and (2) PKCS#15 structure written.

frankmorgner:
Let's look at this from the Specification-point-of-view: In the PKCS#15 structure this means (quoting from ISO/IEC 7816-15):
> *is initialized*, meaning that the password has been initialized;

OK, this is not particularly helpfull :wink:... Just one simple question: *Does* the PIN have some (transport value) after `pkcs15-init`? If so, we should merge this PR.

@philipWendland If `pkcs15-init` does not initialize the PIN there should be an other tool that

1. initializes the PIN
2. rewrites the PKCS#15 structure

And yes, this tool does not yet exist today, but if this is the way you are using your applet, it should be created.

philipWendland:
Am 10/30/2015 um 12:48 PM schrieb Frank Morgner:
>
> Let's look at this from the Specification-point-of-view: In the 
> PKCS#15 structure this means (quoting from ISO/IEC 7816-15):
>
>     /is initialized/, meaning that the password has been initialized;
>
> OK, this is not particularly helpfull :wink:... Just one simple 
> question: /Does/ the PIN have some (transport value) after 
> |pkcs15-init|? If so, we should merge this PR.
>

I don't understand 'transport value'. After pkcs15-init --create-pkcs15 
the PIN is initialized (pkcs15-init *asks* for the PIN value to set).

> @philipWendland <https://github.com/philipWendland> If |pkcs15-init| 
> does not initialize the PIN there should be an other tool that
>
>  1. initializes the PIN
>  2. rewrites the PKCS#15 structure
>
> And yes, this tool does not yet exist today, but if this is the way 
> you are using your applet, it should be created.
>

See above, pkcs15-init does initialize the PIN.
It seems that the usage of this flag in the profile was confusing to me:

What it means seems to mean:
The profile (PKCS#15 structure) written to the card with pkcs15-init 
--create-pkcs15 will have the PIN initialized.

What I thought it meant:
The card comes with a pre-initialized PIN value.

So yes, I think this should be merged. Thanks.

frankmorgner:
ok fine, thanks

PR #597, opened 2015-10-31
win32: make OpenSSL static again

vletoux:


frankmorgner:
The workaround in `winconfig.h.in` does not seem to work in VS2010 as reported by AppVeyor...

vletoux:
got stupid: defining a function in a .h generate assembly code for all .obj file
As a consequence the linker complains that a function is defined in ALL .obj file
=> conflict

vletoux:
@frankmorgner new compilation in progress. Handler moved to libopensc/errors.c
If it works, hearing for feedback for a better place or better way to do this

frankmorgner:
no problem, take your time!

vletoux:
The good news is that the problem is fixed on the newest version of visual studio/sdk
Maybe the simplest solution would be to drop the support for older versions (problem is here only for Windows XP)
PR built will be faster too ...

Your opinion?

vletoux:
@frankmorgner Seems to be working

src/common/compat_report_rangecheckfailure.c
```diff
@@ -0,0 +1,20 @@
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifndef HAVE_RANGECHECKFAILURE	/* empty file if __report_rangecheckfailure is available */
+
+// do not fail when linked with /GS dll and when /GS is not available
+#ifdef _WIN32
+#if _MSC_VER < 1700 // only for vs 2012 or later
```
frankmorgner: Could you use `#if defined(_MSC_VER) && (_MSC_VER <= 1700)` to make it compatible with mingw? Then we could add this file to `Makefile.am`, too.

src/libopensc/errors.c
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
frankmorgner: You could remove this modification on `errors.c`

frankmorgner:
thanks a lot! could you squash all commits into one single commit?

PR #602, opened 2015-11-09
DNIe user interface moved into card-dnie.c. Handles:  Bad implementation of DNIe's UI #362 

germanblanco:
This change removes unused code and removes user-interface module.


rickyepoderi:
Hi @germanblanco ,

I have tested your changes and they work (firefox, thunderbid, signing with pkcs11-tool) but I would add two things:

a) Add the comments for the two options that DNIe manages in opensc.conf (opensc.conf.in tamplate file). I have tested the two options and they are like this:

        card_driver dnie {
                # Disable / enable warning message when performing a
                # signature operation with the DNIe.
                # Only used if compiled with --enable-dnie-ui
                # user_consent_enabled = yes;

                # Specify the pinentry application to use if warning
                # is configured to be displayed using pinentry.
                # Default: /usr/bin/pinentry
                # Only used if compiled with --enable-dnie-ui
                # user_consent_app = "/usr/bin/pinentry";
        }

I think @frankmorgner recommended to add something like this for explaining the options. The options work in your changes smoothly.

b) The second thing is different, The current implementation for --enable-dnie-ui only supports windows, mac (using OS methods) and linux (using pinentry). All the rest OSes cannot set this option (BSD, Solaris,...). I would change just a bit your code to simply default to pinentry in all the rest of OSes. The changes are just a few lines:

    diff --git a/src/libopensc/card-dnie.c b/src/libopensc/card-dnie.c
    index 0c9f5c2..360a531 100644
    --- a/src/libopensc/card-dnie.c
    +++ b/src/libopensc/card-dnie.c
    @@ -180,7 +180,7 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
            CFStringRef header_ref; /* to store title */
            CFStringRef message_ref; /* to store message */
     #endif
    -#ifdef linux
    +#if !defined(__APPLE__) && !defined(_WIN32)
            pid_t pid;
            FILE *fin=NULL;
            FILE *fout=NULL;        /* to handle pipes as streams */
    @@ -247,10 +247,10 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
            if( result == kCFUserNotificationAlternateResponse )
            LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_ALLOWED);
    -#elif linux
    +#else
            /* check that user_consent_app exists. TODO: check if executable */
            res = stat(GET_DNIE_UI_CTX(card).user_consent_app, &st_file);
    -       if (res != 0) {
    +       if (res != 0 || !(st_file.st_mode & S_IXUSR)) {
                    sc_log(card->ctx, "Invalid pinentry application: %s\n",
                           GET_DNIE_UI_CTX(card).user_consent_app);
    @@ -336,8 +336,6 @@ do_error:
            /* close out channel to force client receive EOF and also die */
            if (fout != NULL) fclose(fout);
            if (fin != NULL) fclose(fin);
    -#else
    -#error "Don't know how to handle user consent in this (rare) Operating System"
     #endif
        if (msg != NULL)
                    sc_log(card->ctx, "%s", msg);

Maybe this second options should add some check for pinentry in the configure.ac. What do you think?

germanblanco:
Thank you!
I will create a new Pull request for the configuration option in opensc.conf.
The second proposal also sounds good. However, I also have on the table support for DNIe 3.0 and cleaning all the remaining memory leaks. I would suggest that you please write an issue for it as "Enhancement" and we put it in the TODO list for now.

frankmorgner:
you need to update Makefile.mak so that user-interface.obj is not build anymore

germanblanco:
Rebasing created a mess with the history, so I am going to create a new branch and Pull request.

PR #604, opened 2015-11-10
Adding configuration of pinentry application for DNIe.

germanblanco:
This change adds a configuration block in opensc.conf.in in order to change whether to show an additional warning when signing a document and the application to use in that case.


frankmorgner:
thanks!

PR #606, opened 2015-11-11
Fixing some memory leaks in DNIe.

germanblanco:
Please consider this change.
As far as I can test, it produces a clean output with Valgrind as far as DNIe is concerned.
Additional testing by other people will be very much appreciated.

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -333,11 +333,9 @@ static int dnie_get_info(sc_card_t * card, char *data[])
 		sc_file_free(file);
 		file = NULL;
 	}
-	if (buffer) {
-		free(buffer); 
-		buffer=NULL; 
-		bufferlen=0;
-	}
+	free(buffer); 
+	buffer=NULL; 
+	bufferlen=0;
 	res = dnie_read_file(card, &path, &file, &buffer, &bufferlen);
 	if (res != SC_SUCCESS) {
 		data[3]=NULL;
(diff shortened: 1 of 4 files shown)


frankmorgner:
`dnie_transmit_apdu` should not modify the original APDU. This way, a caller is not forced to clean up buffers that are created in the underlying functions. Maybe the implementation of the envelope command from `dnie_wrap_apdu` could be merged into `sc_transmit_apdu` as alternative to chaining...

germanblanco:
Thank you for the review and your comments, @frankmorgner 
I agree with what you say, that would be the way to do it. I would like to take that step as soon as possible.
However, this pull request fixes other memory leaks in the code, and it fixes all memory leaks. If it is ok with you, I would suggest to put this in and leave the code working without memory leaks in an ugly way. Then we can work on moving dnie_wrap_apdu into sc_transmit_apdu. The "free" that we is there now will have to be removed, but that will be a small detail.

frankmorgner:
When do you think, will you look at `dnie_transmit_apdu`? If we start fixing things, we should fix them at the right place. Otherwise we will create confusion and inconsistency.

germanblanco:
I should be able to do that before the end of the year. It will also require some regression test with other cards in order to check that nothing is broken.
Are there any goal dates for the next release already?
If you prefer, I could remove the "free" of apdu->data from this Pull Request and keep the rest.

frankmorgner:
Yes, I'd prefer that. When editing this commit, please also look at the merge conflict, thanks.

frankmorgner:
I added #610 as a reminder

Issue #596, opened 2015-10-30
DNIe looses login after the card is used by another program


rickyepoderi:
Hi again,

Once the another issue is fixed (issue #480 ) I've been trying to continue testing the DNIe. There is another problem but this time (I think) it is related to the card internals.

First I'm going to describe the problem:

1.- Go to the DNIe test page:  https://av-dnie.cert.fnmt.es/compruebacert/compruebacert
2.- The browser requests the password and the https process is completed successfully. 
3.- Inside the page the information of the certificate is shown. There is a second link that let you test a signature process.
4.- Click the signature link. The new page starts a Java applet (a separate program) that performs a signature sample (it again requests the password and so on).
5.- The signature process also completes successfully.
6.- Reload again the https page. 
7.- Here an error is displayed: SC_ERROR_SECURITY_STATUS_NOT_SATISFIED => CKR_USER_NOT_LOGGED_IN. This error continues until you unplug the card from the reader.

I've been looking the code for a long time, and I think the issue can be described like this:

1.- It seems that the DNIe can only maintain one secure channel at a time (I don't know if this is a typical behavior or not). The DNIe implementation take this fact into account and the secure channel is re-established if broken, this re-connection is done in the dnie_wrap_apdu function inside cwa-dnie.c:

                if ( res == SC_ERROR_SM ) {
                        sc_log(ctx,"Detected SM error/collision. Try %d",retries);
                        switch(provider->status.session.state) {
                                /* No SM or creating: collision with other process
                                   just retry as SM error reset ICC SM state */
                                case CWA_SM_NONE:
                                case CWA_SM_INPROGRESS:
                                        continue;
                                /* SM was active: force restart SM and retry */
                                case CWA_SM_ACTIVE:
                                        res=cwa_create_secure_channel(card, provider, CWA_SM_COLD);
                                        LOG_TEST_RET(ctx,res,"Cannot re-enable SM");
                                        continue;
                        }
                }

2.- But the current problem is that (although the channel is correctly re-established) the card has also lost the pin (the user is not logged in). Therefore the card return SC_ERROR_SECURITY_STATUS_NOT_SATISFIED. the card also needs to pin_verify again if the secure channel is broken. I think the behavior is intrinsic to DNIe (if you re-create the secure channel, you have to pin_verify also).
[...]

PR #608, opened 2015-11-14
Use PC/SC lock in PKCS#11 sessions

frankmorgner:
... a potential fix for #596 that's caused by concurrent applications accessing the same card.

This PR needs careful review, there might be unforeseen strings attached...

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -575,6 +575,9 @@ app opensc-pkcs11 {
 		#
 		# Default: false
 		# lock_login = true;
+		#
+		# Default: true
+		# lock_session = false;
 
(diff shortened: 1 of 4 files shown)


frankmorgner:
Previous versions of OpenSC can achieve a similar behaviour by setting `lock_login = true` in `opensc.conf`.

frankmorgner:
See also discussion from here http://thread.gmane.org/gmane.comp.encryption.opensc.devel/16363/focus=16376

frankmorgner:
should be solved as discussed in PKCS#11 session management, which is a little more complicated, see also https://github.com/OpenSC/OpenSC/issues/596#issuecomment-158672703

PR #609, opened 2015-11-17
Allow ASN.1 decoding if the file seems incomplete

yoe:
Some cards (e.g., BELPIC) have a hardcoded file length that does not
match the actual file length (e.g., 65535 bytes), and simply return the
data that is actually on the card when asked.

It is useful to still be able to do an ASN.1 decode in that case.

Signed-off-by: Wouter Verhelst <w@uter.be>

### DIFF
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -1720,12 +1720,17 @@ static int do_asn1(int argc, char **argv)
 		goto err;
 	}
 	if ((size_t)r != len) {
-		printf("expecting %lu, got only %d bytes.\n", (unsigned long) len, r);
-		goto err;
+		printf("WARNING: expecting %lu, got %d bytes.\n", (unsigned long) len, r);
+		/* some cards return a bogus value for file length. As
+		 * long as the actual length is not higher than the expected
+		 * length, continue */
+		if(r > len) {
+			goto err;
+		}
 	}
 
 	/* asn1 dump */
-	sc_asn1_print_tags(buf, len);
+	sc_asn1_print_tags(buf, r);
 
 	err = 0;
 err:


frankmorgner:
Thanks, that's better.

However, what if `file->size` is `0` because the card did not return this optional information? I used to read the file until there is no more data, see https://github.com/frankmorgner/vsmartcard/blob/master/npa/src/scutil.c#L177-L239. Is something like this available in OpenSC?

frankmorgner:
sorry, we can't apply this commit if `r > len` then the `buf` would have been written out of bounds (or `r` was miscalculated). Did you check your code with valgrind?

yoe:
Eh? Maybe that check is superfluous, but `sc_read_binary()` is called with len as an argument (of the buffer size). If it would be out of bounds, that would be because of a bug in `sc_read_binary()`, rather than in this code. I added the check because a useful return value of `sc_read_binary()` (and common, for similar functions) could be to return the number of bytes that *could* have been read had the buffer been large enough, not because I expect it would have read more than was available.

If you want me to remove it, I can do that; but your comment (about it maybe causing a buffer overflow) seems wrong.

frankmorgner:
Your change looks at something that *should* not happen. If you have encountered this problem, please have a look at the lower levels where the actual problem can be found, e.g. with valgrind.

yoe:
I haven't encountered that problem. I think you're misunderstanding what I'm trying to do here.

The current code produces an error if `r != len` (i.e., if `r < len` *or* `r > len`). My proposed change reduces the error condition to a *sub*case of that existing code; it makes the code produce a warning rather than an error if `r != len`, but *retains* the error (i.e., it does not introduce it) in the `r > len` case.

I can drop the error completely if you prefer that, but that's orthogonal to the fact that I just want it to *not* produce an error in the `r < len` case, as it does now. That is, I don't care whether it produces an error in the `r > len` case (I'm not encountering that problem), so I kept the current behaviour.

If you think the current behaviour in the `r > len` case is wrong, I'll update my patches so they drop the error completely. Otherwise, please reconsider.

frankmorgner:
can you fix this problem within the card driver?

yoe:
Only at a performance cost (but I can look into it).

Regardless of that, in accordance with the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), it makes sense to keep this. The fact that the data is shorter than expected does not imply that it is invalid; producing an error message in that case just seems wrong to me.

Perhaps that's just me, though.

frankmorgner:
maybe I was to harsh, could you squash your commit into a single one?

dengert:
Should line 1733  
    sc_asn1_print_tags(buf, len);
be
    sc_asn1_print_tags(buf,r);

To only scan the data actually read?

yoe:
@dengert -- yes, that's an oversight. Fixed that.

@frankmorgner -- did that.

frankmorgner:
@yoe thanks, have you tested your changes in valgrind with your card?

yoe:
Have now. Output:

    ==2176== 
    ==2176== HEAP SUMMARY:
    ==2176==     in use at exit: 135,492 bytes in 236 blocks
    ==2176==   total heap usage: 607 allocs, 371 frees, 514,843 bytes allocated
    ==2176== 
    ==2176== 72 bytes in 8 blocks are definitely lost in loss record 21 of 63
    ==2176==    at 0x4C28C4F: malloc (vg_replace_malloc.c:299)
    ==2176==    by 0x4E6C298: xmalloc (in /lib/x86_64-linux-gnu/libreadline.so.6.3)
    ==2176==    by 0x4E4B345: readline_internal_teardown (in /lib/x86_64-linux-gnu/libreadline.so.6.3)
    ==2176==    by 0x4E4C3D1: readline (in /lib/x86_64-linux-gnu/libreadline.so.6.3)
    ==2176==    by 0x4024C7: read_cmdline (opensc-explorer.c:1846)
    ==2176==    by 0x4024C7: main (opensc-explorer.c:1983)
    ==2176== 
    ==2176== 112 (40 direct, 72 indirect) bytes in 1 blocks are definitely lost in loss record 27 of 63
    ==2176==    at 0x4C28C4F: malloc (vg_replace_malloc.c:299)
    ==2176==    by 0x6897F15: ???
    ==2176==    by 0x6897687: ???
    ==2176==    by 0x50CA174: pcsc_detect_readers (reader-pcsc.c:1118)
    ==2176==    by 0x509BFAF: sc_ctx_detect_readers (ctx.c:649)
    ==2176==    by 0x509CA42: sc_context_create (ctx.c:773)
    ==2176==    by 0x402282: main (opensc-explorer.c:1909)
    ==2176== 
    ==2176== LEAK SUMMARY:
    ==2176==    definitely lost: 112 bytes in 9 blocks
    ==2176==    indirectly lost: 72 bytes in 3 blocks
    ==2176==      possibly lost: 0 bytes in 0 blocks
    ==2176==    still reachable: 135,308 bytes in 224 blocks
    ==2176==         suppressed: 0 bytes in 0 blocks
    ==2176== Reachable blocks (those to which a pointer was found) are not shown.
    ==2176== To see them, rerun with: --leak-check=full --show-leak-kinds=all
    ==2176== 
    ==2176== For counts of detected and suppressed errors, rerun with: -v
    ==2176== ERROR SUMMARY: 2 errors from 2 contexts (suppressed: 1 from 1)

Neither of those seem related to my changes.

frankmorgner:
thanks

PR #611, opened 2015-11-24
Add Support for electronic signature with German ID card

frankmorgner:
The German ID card has two applications, eID and eSign. Whereas only the eSign application has a key that is associated with an X.509 user certificate (for signature). This PR adds support for the eSign application.

Typically eSign can only be performed with a reader that encompasses a Signature Terminal Certificate for performing EAC with the card. If someone happens to have such a key/certificate on the disk, we can perform EAC in Software using OpenPACE. Note, that the OpenPACE is not required when using a Comfort Reader (which *is* a Signature Terminal).

The card driver for the German ID card is written as external card driver for two reasons. First, this is what was imported from https://github.com/frankmorgner/vsmartcard/tree/master/npa. Second, it can be used as example implementation, which has been requested earlier from OpenSC.

The convenience libraries for EAC and for SM according to ISO 7816-8 are written according to their standards and may be used for different cards in the future. Also the boxing commands to execute PIN verification on a reader's PIN pad are not strictly bound the the German ID card, but can be used by any reader/middleware.


frankmorgner:
Unfortunately I can't make much progress on this. Everything is fine and tested on Linux since last year. However, on Windows there seems to be an invalid free on card->applications on opensc.dll which I have no idea of how to fix (the memory address equals the allocated one and it has not been free'd before). I am suspecting a Windows specific dll-loading problem... Any help would be appreciated!

CardContact:
Does this patch work with a CAT-S reader or does it always need a Signature Terminal Authentication Key in software (which nobody has for productive German eID card) ?

frankmorgner:
All boxing commands are working for all kind of readers (i.e. CAT-S/K) and so will the already merged implementation of `FEATURE_EXECUTE_PACE` via PC/SC.

However, in order to read the certificates and/or use the private key of the German ID card within the eSign application, the reader must prove to be a signature terminal (ST) via Terminal Authentication. This means in order to be able to *use* the card you *must* have a CAT-K (which is available here https://www.chipkartenleser-shop.de/shop/rsct/article/4543 for example). If these requirements are met, then the above code will work. If they are not met, OpenSC as well as any other software for QES with the German ID card will not work.

martinpaljak:
Sorry for noise, just out of curiosity an offtopic question: given that the price of such dedicated reader is comparable to the price of cheap and simple android tablets with NFC - does anyone actually use the card daily, for business or otherwise? I mean home users?

frankmorgner:
The problem on Windows that I reported earlier was fixed with 40600f54aef540f91a890e38bcd1c5f231a3c038. If there are no objections, I'll merge this PR in a couple of days

PR #616, opened 2015-11-26
Heap-based buffer overflow fix in mutex_create() with PKCS11_THREAD_LOCKING and HAVE_PTHREAD defined

mtrojnar:
mutex_create() allocated sizeof(void *) instead of sizeof(pthread_mutex_t) bytes for the pthread mutex. Consequently, pthread_mutex_init() and subsequent thread operations caused heap corruption.

Security impact is limited by the following factors:
1. This function is **not** currently compiled by default, because PKCS11_THREAD_LOCKING is not defined by default.
2. The attacker has limited control over the mutex structure. Depending on the specific implementation, careful heap manipulation may still allow for local/remote code execution.


frankmorgner:
thanks

PR #617, opened 2015-11-26
Enable OS locking

mtrojnar:
libp11 depends on OS locking implementation in the PKCS#11 library (by requesting CKF_OS_LOCKING_OK and **not** providing application locking callbacks). On the other hand, OpenSC **only** implements OS locking if PKCS11_THREAD_LOCKING is defined.

Some applications (e.g. OpenVPN) implement relatively complex workarounds for this simple issue. At least libp11 (and consequently engine_pkcs11) does not implement application locking, causing concurrent requests to fail with the CKR_OPERATION_ACTIVE error.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,6 +143,13 @@ AC_ARG_ENABLE(
 )
 
 AC_ARG_ENABLE(
+	[thread_locking],
+	[AS_HELP_STRING([--disable-thread-locking],[disable OS thread locking @<:@enabled@:>@])],
+	,
+	[enable_thread_locking="yes"]
+)
+
+AC_ARG_ENABLE(
 	[zlib],
 	[AS_HELP_STRING([--enable-zlib],[enable zlib linkage @<:@detect@:>@])],
 	,
@@ -367,6 +374,10 @@ if test "${WIN32}" = "no"; then
 		[AC_MSG_ERROR([POSIX thread support required])]
 	)
 	CC="${PTHREAD_CC}"
+fi
+
+if test "${enable_thread_locking}" = "yes"; then
+	CPPFLAGS="${CPPFLAGS} -DPKCS11_THREAD_LOCKING"
 fi
 
 if test "${enable_minidriver}" = "yes"; then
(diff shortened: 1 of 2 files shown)


frankmorgner:
1. why are you changing the includes?
2. Why don't you pass `PKCS11_THREAD_LOCKING` with `CFLAGS` instead? This is the way it used to be handled.
3. `PKCS11_THREAD_LOCKING` is not documented, what does it mean and why should we enable it by default?

mtrojnar:
Ad 1. The following symbols require additional header file: CRITICAL_SECTION, InitializeCriticalSection, EnterCriticalSection, LeaveCriticalSection, DeleteCriticalSection.
Ad 2. OS locking needs to be enabled by default, because otherwise OpenSC ignores the CKF_OS_LOCKING_OK flag, which in turn violates the PKCS#11 standard and breaks several multi-threaded applications.
Ad 3. This question is already answered in the pull request description at the top of this page.  Please read it.

frankmorgner:
I agree, that this feature should be enabled. However, I don't understand why it was configurable and even disabled in the first place...

frankmorgner:
And by the way, if you always enable this feature you should remove this define completely!

mtrojnar:
Good point.  I have updated the pull request according to your recommendation.

frankmorgner:
You could remove `HAVE_OS_LOCKING` as well.

I'll leave this open for a while, maybe some other developer (@dengert @martinpaljak ?) has an idea why locking used to be disabled.

dengert:
I don't know why it used to be disabled. But there are a number of possibilities:

 It was buggy, and may still be, and still considered experimental so it was not enabled. 

 Some want to build a minimal package to run on other systems then Windows, linux or MacOS.

I would leave the \#ifdef in, and change configure to test if it should be on by default by allow it to be turned off.

martinpaljak:
Very good question that does not have a good answer. I *think* the reason is/was the combination of "what/how Mozilla or something comparable did at some point of time" and "have a minimalistic build by default". Also, because traditionally smart cards have been mostly sequential devices (with buggy or missing implementations of C_WaitForSlotEvent and comparable stacks).

I lack the knowledge currently to reply what is the state of threading in OpenSC, but would also opt for "default to on" with an ifdef/config option for a while (but remove them every second release or so)

mtrojnar:
I have added the --disable-thread-locking config option as requested by @martinpaljak.

frankmorgner:
could you squash your commits int a single one? all changes essentially allow a single thing: setting the define via `./configure`.

mtrojnar:
More or less...  Also, old code did not really compile on WIN32.

mtrojnar:
Okay, @frankmorgner.  I squashed my commits into a single one.

frankmorgner:
thanks!

PR #618, opened 2015-11-26
Implemented atomic transactions for PKCS#11

### LINKED ISSUE #596: DNIe looses login after the card is used by another program

frankmorgner:
See discussion here http://opensc.1086184.n5.nabble.com/State-of-stateless-support-td15513.html

fixes https://www.opensc-project.org/opensc/ticket/186
fixes #596

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -594,6 +594,27 @@ app opensc-pkcs11 {
 		# Default: false
 		# lock_login = true;
 
+		# By default, interacting with the OpenSC PKCS#11 module may change the
+		# state of the token, e.g. whether a user is logged in or not.
+		#
+		# Thus other users or other applications may change or use the state of
+		# the token unknowingly. Other applications may create signatures
+		# abusing an existing login or they may logout unnoticed.
+		#
+		# With this setting enabled the login state of the token is tracked and
+		# cached (including the PIN). Every transaction is preceeded by
+		# restoring the login state.  After every transaction a logout is
+		# performed. This setting by default also enables `lock_login` (see
+		# above) to disable access for other applications during the atomic
+		# transactions.
+		#
+		# Please note that any PIN-pad should be disabled (see `enable_pinpad`
+		# above), because the user would have to input his PIN for every
+		# transaction.
+		#
+		# Default: false
+		# atomic = true;
+
 		# With this setting disabled, the OpenSC PKCS#11 module will initialize
 		# the slots available when the application calls `C_GetSlotList`. With
 		# this setting enabled, the slots will also get initialized when
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -1221,8 +1221,13 @@ cardos_pin_cmd(struct sc_card *card, struct sc_pin_cmd_data *data,
 static int
 cardos_logout(sc_card_t *card)
 {
-	if (card->type == SC_CARD_TYPE_CARDOS_M4_01 ||
-	    card->type == SC_CARD_TYPE_CARDOS_M4_2) {
+	if (card->type == SC_CARD_TYPE_CARDOS_M4_01
+		   	|| card->type == SC_CARD_TYPE_CARDOS_M4_2
+		   	|| card->type == SC_CARD_TYPE_CARDOS_M4_2B
+		   	|| card->type == SC_CARD_TYPE_CARDOS_M4_2C
+		   	|| card->type == SC_CARD_TYPE_CARDOS_M4_3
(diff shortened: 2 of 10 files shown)


frankmorgner:
I additionally added some commits that do the following:
- correctly propagate the login_user via PKCS#11 if an error in the lower layers occur
- PIN-caching now also works while selecting the key and issuing MSE
- clean up some of the mess in `pkcs15-sec.c`

frankmorgner:
I tested with CardOS 4.3 and `pkcs11-tool --login --test`. It's reported to work with DNIe in #596 .

Issue #610, opened 2015-11-18
`dnie_transmit_apdu` should not modify the original APDU


frankmorgner:
This way, a caller is not forced to clean up buffers that are created in the underlying functions.

Maybe the implementation of the envelope command from `dnie_wrap_apdu` could also be merged into sc_transmit_apdu as alternative to chaining...

see also #606

frankmorgner:
closed with https://github.com/OpenSC/OpenSC/pull/694

PR #622, opened 2015-11-27
Removing all memory leaks in DNIe but one. Refactoring pending. Continued from -> Fixing some memory leaks in DNIe. #606 

germanblanco:
This Pull Requests resolves memory leaks in DNIe module.
It is a continuation of " Fixing some memory leaks in DNIe. #606 ", with a rebase and compressed to just one commit.
A memory leak is pending, because it will be solved together with the issue in " `dnie_transmit_apdu` should not modify the original APDU #610 "
Tests:
- Signing a document with AFirma application
- Checking the authentication certificate in the web https://av-dnie.cert.fnmt.es/compruebacert/compruebacert
- Running valgrind: valgrind --leak-check=yes pkcs11-tool --module /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so -l -t (all memory leaks are in the same point, modification of the apdu data by the wrapper)

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -634,17 +634,16 @@ static int dnie_get_serialnr(sc_card_t * card, sc_serial_number_t * serial)
 		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 	}
 	/* not cached, retrieve it by mean of an APDU */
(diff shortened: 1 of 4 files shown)


frankmorgner:
looks good, thanks. I'll wait a bit if someone else wants to peek into the changes.

Issue #629, opened 2015-12-08
Events


faezeh-R:
Hi;
The function of waitForSlotEvent with detaching the reader doesn't get the correct event.
It returns the NO_Event. I tested it with pkcs11-tools.exe
Also in using the pkcs#11.dll in Firefox , when the reader has been attached, the list of slots changes but when the reader has been detached the list doesn't change. 
The driver of reader is pcsc.
thanks

frankmorgner:
How can we reproduce this error?

faezeh-R:
If you use the Firefox, You see that after detaching the reader, the list of slot doesn't change.
If you want to test it with pkcs11-tools.exe, with running the "--test-hotplug"  options for testing the waitForSlotEvent ,you can see this error.
It shows that the event of detaching reader doesn't get the correct event.

faezeh-R:
I correct it.

dengert:
Can you run FireFox with SPY, to see what FireFox is actualing
    doing? 
    
    C_WaitForSlotEvent returns that aan event has happened. The
    application has to then test what happened. 
    
    I suspect that FireFox is either getting CKR_FUNCTION_NOT_SUPPORTED
    and will then use  a loop  calling  C_GetSlotInfo  to see shat
    happened. 
    
    https://github.com/OpenSC/OpenSC/pull/624 addresses the issue that 
    C_GetSlotInfo  does not returning the true state. 
    
    
    
    On 12/8/2015 5:47 AM, Faezeh wrote:
    
    
      If you use the Firefox, You see that after detaching the
        reader, the list of slot doesn't change.
        If you want to test it with pkcs11-tools.exe, with running the
        "--test-hotplug" options for testing the waitForSlotEvent ,you
        can see this error.
        It shows that the event of detaching reader doesn't get the
        correct event.
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

 Douglas E. Engert  <DEEngert@gmail.com>

dengert:
#624 addresses C_GetSessionInfo, not C_GetTokenInfo, but a similar issue may exist.

frankmorgner:
I think the real problem is that PKCS#11 defines the list of slots to be static until `C_GetSlotList` is called the next time with `pSlotList == NULL`. Quoted from PKCS#11 v2.30:
[...]

PR #630, opened 2015-12-08
fixing the issue #629

faezeh-R:


src/pkcs11/slot.c
```diff
@@ -454,9 +455,11 @@ CK_RV slot_token_removed(CK_SLOT_ID id)
 CK_RV slot_find_changed(CK_SLOT_ID_PTR idp, int mask)
 {
 	unsigned int i;
+	CK_RV rv;
 	LOG_FUNC_CALLED(context);
 
-	card_detect_all();
+	rv = card_detect_all();
+	if(rv == CKR_TOKEN_NOT_PRESENT ) LOG_FUNC_RETURN(context, CKR_TOKEN_NOT_PRESENT);
```
frankmorgner: this doesn't work. card_detect_all will always return `CKR_OK`!

src/pkcs11/pkcs11-global.c
```diff
@@ -656,9 +656,13 @@ CK_RV C_WaitForSlotEvent(CK_FLAGS flags,   /* blocking/nonblocking flag */
 		mask |= SC_EVENT_READER_EVENTS;
 	}
 
+
+	sc_ctx_detect_readers(context);
+
 	rv = slot_find_changed(&slot_id, mask);
-	if ((rv == CKR_OK) || (flags & CKF_DONT_BLOCK))
+	if (rv == CKR_OK   ||  (rv != CKR_TOKEN_NOT_PRESENT  && (flags & CKF_DONT_BLOCK) ))
```
frankmorgner: this doesn't work. card_detect_all will never return `CKR_TOKEN_NOT_PRESENT`!

faezeh-R: I changed the return of carddetectAll in slot.c

frankmorgner:
I think this is the wrong fix. This may work with firefox, but I it is not compliant with PKCS#11. The return code is not defined in this context, it does not set the slot that changed.

dengert:
I agree with @frankmorgner 

 An OpenSC SPY trace would be most helpful to see what FireFox is actually doing.

faezeh-R:
But this solution works in firefox and pkcs11-tools.exe , both of them not just in firefox.
I traced the code and saw that when the reader doesn't found, card_detect function returns the error of "READER_NOT_FOUND" but the function of card_detect_all always returns CKR_OK.
then In function of slot_find_changed for doesn't detecting the event of reader, It returns CKR_NO_EVENT.
finally in function of C_WaitForSlotEvent, the codes with label again never execute.

frankmorgner:
I was not able to reproduce this behavior with pkcs11-tool, only with FF.

If the reader was removed its slot *should* be marked with a removed token and a changed state. However, currently `slot_find_changed` does *not recognize* the changes state because the data structure "slot" has already been removed from the internal memory in `card_detect_all`.

The real fix would be to keep the slot alive though the reader has been removed and to propagate the changed state (token removed). You could use https://github.com/frankmorgner/OpenSC/commit/87327030b42641815a6bdc0f8491a1e7b6f32b20 as starting point, but beware it is neither functional (events are still not propagated correctly) nor correct (segfaults in FF)...

frankmorgner:
as pointed out, above a fix should take a different path

PR #633, opened 2015-12-09
minidriver: add logout function to avoid if supported a smart card reset

vletoux:


frankmorgner:
> the minidriver does not perform a deauthentication - set it to NULL according to the specification

how is this comment related to the new code?

vletoux:
Yes the specifications tell that but I know for sure that the card is reset if there is an error reported here.
You will just fail the cmck.exe test and you won't be able to deploy the opensc minidriver via Windows update ;-(

frankmorgner:
>4.2.5 CardDeauthenticate
Description:
CardDeauthenticate is an optional export that should be provided if it is possible within the card minidriver to efficiently reverse the effect of authenticating a user or administrator without resetting the card. If this function is not implemented, the card minidriver should place NULL in the CARD_DATA structure pointer for this function.
The CSP/KSP tests this pointer for NULL value before calling it. If it is found NULL, the CSP/KSP deauthenticates a user by resetting the card. Because a card reset is a time-consuming operation, the card minidriver should implement this function if it can be done:

I don't quite understand why MS allows implementing `CardDeauthenticate` by its spec, but disallows using it with `cmck.exe`. But if that's the case, could you clarify the comment in the OpenSC minidriver accordingly, please?

@CardContact , @vletoux could you both clean up the commit messages, please. Avoiding the additional merge conflict would be ideal, too.

vletoux:
No; cmck.exe (the minidriver qualification tool), does report an anomaly in the minidriver if the CardDeauthenticate is present here and does return anything else than 0 (success).

You can find it in the "minidrivriver certification requirements" published here:
https://msdn.microsoft.com/en-us/library/windows/hardware/dn653312%28v=vs.85%29.aspx
Then in CMR_119: "CardDeauthenticate should return 0 on success"

Ok I'll remove the comment in the source code

vletoux:
comment removed & tested using the binary version produced by AppVeyor (the one which is working ...)
=> the minidriver works

For the merge message, I do not know how to remove them. Tutorial link welcome.

philipWendland:
`git rebase -i HEAD~1` to rebase the latest commit (or `HEAD~3` for the latest 3), replace "pick" by "reword", save and exit. An editor will pop up and you can edit the commit message.

Commits can be squashed by using "squash" in the rebase selection.

`git push -f` if you are sure you are done; this will overwrite the commits shown here now.

See `git help rebase` for additional info. You might want to test this on a local branch first :-)

vletoux:
@CardContact Andreas, can you merge https://github.com/CardContact/OpenSC/commit/aac30661cc54f981a1966178e5790a2664855dcf and the comment line that Frank referenced that should be removed in https://github.com/CardContact/OpenSC/tree/fix_631 ?
I've broken everything in my repository and I want to rebuilt it (understand remove everything).

Thanks in advance

frankmorgner:
There is https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history, for example

frankmorgner:
@vletoux, @CardContact I tried to give you a hand in #635, if you're OK with it.

vletoux:
Ok for me
(Thanks frank)

PR #636, opened 2015-12-14
Certificate removal fixed (IB#1047758)

pboguslawski:
Hello,

We've hit a bug in opensc (Debian 8, 0.14.0-2) that blocked certificate update operation on our SC (Athena ASEPCOS):
```
$ pkcs15-init -U personal_cert.crt -a 01 --id 4fa066fd58b2e24da6618173bf1f654c57d96ad6 --format pem
Using reader with a card: Gemalto GemPC Express 00 00
Security officer PIN [Security Officer PIN] required.
Please enter Security officer PIN [Security Officer PIN]:
Failed to update certificate: Security status not satisfied
```
PIN was ok. Certificate deletion generated same error.

After debugging we've found that the following code is executed...
```
        sc_log(ctx, "Try to get the parent's 'DELETE' access");
        /*file_type = file->type;*/
        if (file_path->len >= 2) {
            /* Select the parent DF */
            path.len -= 2;
            rv = sc_select_file(p15card->card, &path, &parent);
            LOG_TEST_RET(ctx, rv, "Cannot select parent");

            rv = sc_pkcs15init_authenticate(profile, p15card, parent, SC_AC_OP_DELETE);
            sc_file_free(parent);
            LOG_TEST_RET(ctx, rv, "parent 'DELETE' authentication failed");
        }
```
...in src/pkcs15init/pkcs15-lib.c > sc_pkcs15init_delete_by_path() and leaves path variable pointing to dir not file in
```
    /* Reselect file to delete: current path could be changed by 'verify PIN' procedure */
    rv = sc_select_file(p15card->card, &path, &file);
    LOG_TEST_RET(ctx, rv, "cannot select file to delete");
```
When path variable is reset...
```
path = *file_path;
```
...before calling final sc_select_file, problem does not occur.

Please verify this PR and consider fixing it.

Regards,
Pawel

IB Development Team
https://dev.ib.pl/

PR #638, opened 2015-12-15
Fix smartcard certificate removal and update

pboguslawski:
This patch fixes bug found in opensc 0.14.0-2 (Debian 8) that blocks
certificate update and removal on Athena ASEPCOS smartcard with error

    Failed to update certificate: Security status not satisfied

OpenSC tries to get the parent's 'DELETE' access before cert removal
and leaves path variable pointing to dir not file to be removed.

Author-Change-Id: IB#1047758


pboguslawski:
This PR is continuation of https://github.com/OpenSC/OpenSC/pull/636
If this PR won't meet your requirements, please use it as a base for your own fix.
Thank you.

frankmorgner:
thanks!

Issue #632, opened 2015-12-09
get_response should be called recursively.


martinpaljak:
Here's a scenario: READ BINARY is called, card replies with <data>6100, GET RESPONSE is called, card replies with <data>6100 and after this, next call is issued by the OpenSC but the card responds with 6F00.

I read out from 7816-4 that GET RESPONSE should be called recursively until data all data is fetched?

dengert:
I read it :   only when  61XX is returned should it be called again.
    
    6F00 is a "checking error"  "No precise diagnosis"
    
    Sounds like hardware error, or card  has an internal error and does
    not know what else to return.
    
    
    
    On 12/9/2015 8:45 AM, Martin Paljak
      wrote:
    
    
      Here's a scenario: READ BINARY is called, card replies with
        6100, GET RESPONSE is called, card replies with 6100 and after
        this, next call is issued by the OpenSC but the card responds
        with 6F00.
      I read out from 7816-4 that GET RESPONSE should be called
        recursively until data all data is fetched?
      —
        Reply to this email directly or view it on
          GitHub.
      
        
          
          
        
        
      
    
    
    -- 

 Douglas E. Engert  <DEEngert@gmail.com>

frankmorgner:
Isn't this implemented?
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/apdu.c#L548-L549
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/apdu.c#L435-L515

dengert:
I think it is implemented, 

iso7816.c  checks for it:
 { 0x6F00, SC_ERROR_CARD_CMD_FAILED,     "No precise diagnosis" },

apdu.c line 476 - 484 should then return  SC_ERROR_CARD_CMD_FAILED

germanblanco:
I am currently having a problem with this.
You can see in 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/apdu.c#L435-L515
that get_response is called within a loop and the loop does not check for 6100, it only checks the returned length.
The DNIe is returning 6900, get_response is called again, and then the card returns "Invalid status".
Shoudn't this loop check the response code for 6100 before doing get_response?

dengert:
In Line 476, 477:

rv = card->ops->get_response(card, &resp_len, resp);
if (rv < 0)   {

rv could be the length, or an error code. 
So it depends on the card driver to issue the command, and check the response. 

[...]

PR #639, opened 2015-12-15
Adding sm decode function with complete response.

germanblanco:
This is what I think would be required to be able to decode secure messages longer than 255 bytes.
Please see the thread in:
get_response should be called recursively. #632
I believe that the decode method needs to be called when the response data is complete in order to do a correct parsing. At least this seems to be the case with DNIe.
I haven't been able to test this code yet, but I would like to start the discussion in order to know how to do the coding of DNIe card software using the SM module in OpenSC.


frankmorgner:
I think this approach is incorrect and does not solve a problem. SM decoding should have been done already within the while loop (thus, decoding afterwards is incorrect). And it's not clear whether any card works better with this patch applied.

PR #642, opened 2015-12-21
Fix public key access and attributes

mouse07410:
1. Make public key accessible without entering the PIN.
2. Stop setting attribute CKA_SENSITIVE (which means "stays in hardware only") for public keys.

Since currently certificates are accessible and retrievable without PIN authentication, and public key can be extracted from certificate, this change does not impact security model negatively. This change also makes the implementation PKCS#11-compliant.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -3879,8 +3879,9 @@ pkcs15_pubkey_get_attribute(struct sc_pkcs11_session *session, void *object, CK_
 		break;
 	case CKA_TOKEN:
 	case CKA_SENSITIVE:
+		/* By PKCS#11 v2.20 public key cannot have SENSITIVE attr TRUE */
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
-		*(CK_BBOOL*)attr->pValue = TRUE;
+		*(CK_BBOOL*)attr->pValue = FALSE;
 		break;
 	case CKA_LOCAL:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -3878,6 +3878,9 @@ pkcs15_pubkey_get_attribute(struct sc_pkcs11_session *session, void *object, CK_
 		*(CK_OBJECT_CLASS*)attr->pValue = CKO_PUBLIC_KEY;
 		break;
 	case CKA_TOKEN:
+		check_attribute_buffer(attr, sizeof(CK_BBOOL));
+		*(CK_BBOOL*)attr->pValue = TRUE;
+		break;
 	case CKA_SENSITIVE:
 		/* By PKCS#11 v2.20 public key cannot have SENSITIVE attr TRUE */
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
--- a/src/tools/pkcs15-tool.c
+++ b/src/tools/pkcs15-tool.c
@@ -688,9 +688,7 @@ static int read_public_key(void)
 	if (r >= 0) {
 		if (verbose)
 			printf("Reading public key with ID '%s'\n", opt_pubkey);
-		r = authenticate(obj);
-		if (r >= 0)
-			r = sc_pkcs15_read_pubkey(p15card, obj, &pubkey);
+		r = sc_pkcs15_read_pubkey(p15card, obj, &pubkey);
 	} else if (r == SC_ERROR_OBJECT_NOT_FOUND) {
 		/* No pubkey - try if there's a certificate */
 		r = sc_pkcs15_find_cert_by_id(p15card, &id, &obj);


src/pkcs11/framework-pkcs15.c
```diff
@@ -3879,8 +3879,9 @@ pkcs15_pubkey_get_attribute(struct sc_pkcs11_session *session, void *object, CK_
 		break;
 	case CKA_TOKEN:
 	case CKA_SENSITIVE:
```
dengert: Before make this change, the question that should be asked, is why is OpenSC returning the CKA_SENSITIVE attribute for a public key, as that is not a valid attribute for a public key? 


When was this added, and why?

mouse07410: I don't know why - but it is clearly against both the standard and the common sense (the whole idea of PKC is that you can freely share public keys).

And I definitely am not the one to know who and why wrote that code. I can only say that it looks like a bug.

frankmorgner: check `git blame` to get to know who did it. But it should more usefull to check the standard...

mouse07410: PKCS#11 standard says that CKA_SENSITIVE is the attribute of either a private or a secret key. Take a look at the references below.

For example, [http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cs01/pkcs11-curr-v2.40-cs01.pdf] clearly states that CKA_SENSITIVE is an attribute of private keys (and secret keys),  **and omits it** from the public key templates.

Other references may be helpful too:
www.cryptsoft.com/pkcs11doc/STANDARD/pkcs-11v2-20.pdf
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=3&cad=rja&uact=8&ved=0ahUKEwi3x9Xh7u7JAhVDbSYKHebtCLgQFggpMAI&url=https%3A%2F%2Fwww.oasis-open.org%2Fcommittees%2Fdownload.php%2F49502%2Fpkcs11-cka-public-key-info-r4.docx&usg=AFQjCNGks6-yexjogHCRZNLsUG0TbFapuw&sig2=us78EEjV4r5iqsTNhS4wCQ
https://books.google.com/books?id=tEO6BQAAQBAJ&pg=PA618&lpg=PA618&dq=pkcs11+meaning+of+key+attribute+CKA_SENSITIVE&source=bl&ots=THU1LvK1se&sig=ttcL8bHtaptxw2BC7TBx4Txn7Xg&hl=en&sa=X&ved=0ahUKEwi3x9Xh7u7JAhVDbSYKHebtCLgQ6AEIMDAD#v=onepage&q=pkcs11%20meaning%20of%20key%20attribute%20CKA_SENSITIVE&f=false

It might be interesting to learn who wrote the OpenSC code the way it is, and why - but as you pointed out yourself, the standard should take precedence.
```
@git blame src/pkcs11/framework-pkcs11.c
..........
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3142)  case CKA_TOKEN:
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3143)          check_attrib
ute_buffer(attr, sizeof(CK_BBOOL));
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3144)          *(CK_BBOOL*)attr->pValue = TRUE;
.............
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3386)  case CKA_CLASS:
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3387)          check_attrib
ute_buffer(attr, sizeof(CK_OBJECT_CLASS));
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3388)          *(CK_OBJECT_
CLASS*)attr->pValue = CKO_PRIVATE_KEY;
54f87942 (nils                 2005-02-03 22:44:19 +0000 3389)          break;
6e6d7177 (fabled               2002-01-16 22:49:03 +0000 3390)  case CKA_TOKEN:
4915eaa5 (Thomas Calderon      2014-12-03 17:59:04 +0100 3391)          check_attrib
ute_buffer(attr, sizeof(CK_BBOOL));
4915eaa5 (Thomas Calderon      2014-12-03 17:59:04 +0100 3392)          *(CK_BBOOL*)attr->pValue = TRUE;
4915eaa5 (Thomas Calderon      2014-12-03 17:59:04 +0100 3393)          break;
.................
252323e0 (okir                 2002-03-12 14:36:40 +0000 3880)  case CKA_TOKEN:
252323e0 (okir                 2002-03-12 14:36:40 +0000 3881)  case CKA_SENSITIVE:
252323e0 (okir                 2002-03-12 14:36:40 +0000 3882)          check_attrib
ute_buffer(attr, sizeof(CK_BBOOL));
00000000 (Not Committed Yet    2015-12-22 01:59:09 -0500 3883)          *(CK_BBOOL*)
attr->pValue = FALSE;
```

But according to this https://github.com/OpenSC/OpenSC/commit/4915eaa5, `case CKA_SENSITIVE:` should've been removed from that place by @viktorTarasov on 2014-12-03. His code makes it clear that this attribute makes sense **only** for private keys...

dengert:
@mouse07410  
Please check out this pull request. 
https://github.com/OpenSC/OpenSC/pull/643

The problem of requiring the PIN may have been in the pkcs15-piv.c code, which was setting an auth_id for pubkeys but not certificates. 

The sensitive issue may not be your problem at all.

mouse07410:
> Please check out this pull request #643

Thank you - will check soon.

> The problem of requiring the PIN may have been in the pkcs15-piv.c code,
> which was setting an auth_id for pubkeys but not certificates.

Great find! I'm sure it was **a** problem, but I am not certain yet it was **the** (only) problem. I still think that forcing CKA_SENSITIVE to TRUE not only violates PKCS#11, but also prevents the public key in question from being retrieved from the token by applications like OpenSSL and subsequently used.

> The sensitive issue may not be your problem at all.

With `pkcs15-tool --read-public-key` requesting PIN it's very likely that you're right.

I'm not sure it would be sufficient to allow openssl to access the public key though... *On the other hand, I'm not sure the change I'm proposing would be sufficient either.*

frankmorgner:
In pkcs#11 v2.30 `CKA_SENSITIVE` is *not defined* in Table 28, Common Public Key Attributes. This attribute only exists in Table 30, Common Private Key Attributes, and Table 31, Common Secret Key Attributes. So throwing an error would actually be the most appropriate thing to do according to the standard.  @mouse07410 could you check if this would work with engine_pkcs11?

mouse07410:
> In pkcs#11 v2.30 CKA_SENSITIVE is not defined in Table 28, Common Public Key Attributes.

I know. I kept telling that. :-)

The current version of the standard is 2.40, but it doesn't make much difference.

> So throwing an error would actually be the most appropriate thing to do according to the standard.

Please point me at the place in this standard that actually says you should throw an error in such a case.

> could you check if this would work with engine_pkcs11?

Yes I can, but strongly suggest to reconsider. 

First, I'm going to hold off until the exact reference is provided, justifying such a change. Second, this goes against all the principles that IETF tried to build the Internet upon (summarizing: **Be conservative in what you send, and liberal in what you receive**). Assuming there were no earlier standard versions that did allow this attribute for public keys, and no weird (but still used) devices that did set that attribute (regardless whether the standard condones it or not), the infraction still is extremely benign. There is no reason to blast an application for asking whether an attribute is set, even if that attribute should never be set for a given type of key. 
*Example: you ask "Is it prohibited to throw garbage here?", and get slapped on the face as a response. Normal response should be "Yes it is", or "No it isn't", and that's all.*
Also note, that there are no **values** of these attributes. Our code **maps** attribute **presence** to **boolean TRUE**, and **absence** to **boolean FALSE**. So to make an application explode if it asks whether an attribute of a key is present seems counterproductive. If that attribute cannot/shouldn't be present, well just answer "No" always (represented by returning FALSE).

dengert:
"Please point me at the place in this standard that actually says you should throw an error in such a case."

PKCS#11 v 2.20:
"C_GetAttributeValue"
"2. Otherwise, if the specified attribute for the object is invalid (the object does not
possess such an attribute), then the ulValueLen field in that triple is modified to hold
the value -1."
"If case 2 applies to any of the requested attributes,
then the call should return the value CKR_ATTRIBUTE_TYPE_INVALID."

CKR_ATTRIBUTE_TYPE_INVALID can also be returned by many other functions
if a template has an invalid attribute. 

pkcs15_pubkey_get_attribute in ./framework-pkcs15.c:
4018         default:
4019                 return CKR_ATTRIBUTE_TYPE_INVALID;

Also the current default action does not set
attr->ulValueLen = -1;

Note PKCS#11 2.20  was written in 2004. 
V1.10 was written in 2001 with the same text.
The OpenSC code dates from 2002 but may have been written based on an earlier draft.

dengert:
"Be conservative in what you send, and liberal in what you receive" 
Yes I agree with that principal.
I believe your problem has nothing to do with sensitive.

dengert:
"Our code maps attribute presence to boolean TRUE, and absence to boolean FALSE."

No, the code always returns TRUE  to a query for the CKA_SENSITIVE on a publey.
The caller has to ask for the attribute. 

./pkcs11-object.c has:
 232         for (i = 0; i < ulCount; i++) {
 233                 res = object->ops->get_attribute(session, object, &pTemplate[i]);
 234                 if (res != CKR_OK)
 235                         pTemplate[i].ulValueLen = (CK_ULONG) - 1;

So it does set ulValueLen = -1

mouse07410:
> "Otherwise, if the specified attribute for the object is invalid (the object does not
> possess such an attribute), then the ulValueLen field in that triple is modified to
> hold the value -1"

I don't think this clause applies. Because for the attributes we're discussing - they're either present or absent. So "the object does not possess such an attribute" would simply mean this attribute is not set.

> Note PKCS#11 2.20 was written in 2004.

Noted. I believe the current version is 2.40, circa 2011.

> "Our code maps attribute presence to boolean TRUE, and absence to boolean FALSE."
> No, the code always returns TRUE to a query for the CKA_SENSITIVE on a pulley.
> The caller has to ask for the attribute.

Well, the **original** OpenSC code always returns TRUE. **My** fork always returns FALSE (which is the only response that makes sense to me under the circumstances). The ideal solution would be to query the token, and reply with what the token says. But it seems simpler to just say "no CKA_SENSITIVE is not set for this key" and be done with it. In OpenSC language it means setting it to FALSE.

> I believe your problem has nothing to do with sensitive.

You may well be right here. On the other hand, having the software claim that public keys cannot leave the token does not seem right either, regardless of whether this causes other software components to force authentication, or not. So while I'm not arguing with you, I think this should be fixed the way I propose (returning FALSE). In the meanwhile we can look for the real cause of the problem, manifesting in OpenSSL/engine_pkcs11 inability to access/locate public keys unless it does token login first. I must add that libp11 doesn't even find the token unless one does the login before key enumeration.

PR #643, opened 2015-12-22
PIV pubkey auth_id fix

dengert:
pkcs15-piv.c was setting the auth_id of the public keys
which would cause some appications to require a login to access
a public key. The public keys are obtained from the certificates
which do not require the PIN to read.

Very early drafts of NIST 800-73 did require the PIN to access the
certificates, and the auth_id  was removed in the opensc code for
certificates many years ago, but not from the public keys.


mouse07410:
I confirm that this patch is OK.

frankmorgner:
thanks!

mouse07410:
@frankmorgner and @dengert, I've re-built and tested OpenSC and OpenSCtokend with the latest merges. I confirm that the final code works fine with RSA (tested with RSA2048) and ECDSA (tested with P-384, the latest YubiKey 4) with El Capitan 10.11.2 Apple Mail. My other tests that utilize OpenSC library and command tools (and OpenSSL-1.0.2e), successfully tested RSA signature (PKCS-1.5 and PSS), RSA encryption/decryption, ECDSA signature (over P-384), and ECDH key derivation.

My remaining problems/issues are: (a) OpenSSL can't work directly with public keys (by that I mean extracting the key and using it in software) unless the token is PIN-unlocked, and (b) ECDH-based S/MIME (which we don't know how to deal with anyway - don't even know if any tokend-using application knows how to use ECDH).

frankmorgner:
> My remaining problems/issues are: (a) OpenSSL can't work directly with public keys (by that I mean extracting the key and using it in software) unless the token is PIN-unlocked

Does this mean that #642 broke youre test case (a) with OpenSSL?

mouse07410:
> Does this mean that #642 broke youre test case (a) with OpenSSL?

Oh no - sorry for erroneously giving that impression. Everything's fine wrt. OpenSC. That was about engine_pkcs11 problem.

Also, with the latest patches to `OpenSC/engine_pkcs11` (PR there, not merged yet) and the `PKCS11_enumerate_public_keys` branch of `OpenSC/libp11` (waiting to be merged into the master), OpenSSL works just fine!

PR #645, opened 2015-12-26
add zlib compilation into the appveyor built process

vletoux:
zlib feature now available for x64
zlib is compiled statically and avoid the need to deploy zlib1.dll into system32

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,6 +30,7 @@ install:
   - set PATH=C:\cygwin\bin;%PATH%
   - set OPENSSL_VER=1_0_2e
   - set ZLIB_VER=128
+  - set ZLIB_VER_DOT=1.2.8
   - ps: >-
       If ($env:Platform -Match "x86") {
         $env:VCVARS_PLATFORM="x86"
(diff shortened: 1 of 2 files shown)


frankmorgner:
Thanks!

You could also enable caching on `C:\zlib`

vletoux:
zlib.lib is the same binary for both x86 and x64 => needs to rename the file or directory to cache it.
Then caching it, implies that the build code for zlib has to be moved from the build part of the script to the install part.
Is that what you want ?
Maybe the best would be to cache only the zlib.zip file.

frankmorgner:
Your right, caching zlib.zip should be best.

Am 27. Dezember 2015 10:07:49 MEZ, schrieb vletoux <notifications@github.com>:
>zlib.lib is the same binary for both x86 and x64 => needs to rename the
>file or directory to cache it.
>Then caching it, implies that the build code for zlib has to be moved
>from the build part of the script to the install part.
>Is that what you want ?
>Maybe the best would be to cache only the zlib.zip file.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/OpenSC/OpenSC/pull/645#issuecomment-167396002

-- 
Frank Morgner

vletoux:
done

frankmorgner:
nice!

PR #654, opened 2016-01-06
Changes needed for OpenSSL 1.1

dengert:
This change is the first cut at changes to OpenSC so it can be compiled against OpenSSL 1.1 as well as previous versions.

OpenSSL 1.1 will be released later this year.

	https://www.openssl.org/policies/releasestrat.html

	https://www.openssl.org/news/changelog.html#x0

17 OpenSC routines need changes ranging from adding headers for ossl_type.h, rsa.h and bn.h to rewriting code to use changed interfaces. 

5 routines that should be looked at closely are:

	src/libopensc/card-epass2003.c
	src/libopensc/pkcs15-itacns.c
	src/pkcs15init/pkcs15-oberthur-awp.c
	src/tools/pkcs15-init.c
	src/tools/sc-hsm-tool.c

src/tools/sc-hsm-tool.c needs so many changes that I did not attempt to change it. If compiled against OpenSSL 1.1 it will produce a dummy routine, and a warning message.
It uses the BIGNUM structure in line in many places, but BIGNUM is now just a typedef and the actual structure is now hidden.
@CardContact 
The sc-hsm developers need to make the changes.

I have compiles against OpenSSL 1.0.1g, 1.0.2d and 1.1.0-pre2-dev
from github.com Minimal testing has been done.

Next is engine and libp11 so ECDH can be used from engine.

### DIFF
--- a/src/common/simclist.h
+++ b/src/common/simclist.h
@@ -44,11 +44,20 @@ typedef INT64   int64_t;
 #include <errno.h>
 #include <sys/types.h>
 
+/* OpenSSL-1.1 e_os2.h does #if !defined(inline)
+ * the following causes problems. One of them needs to be fixed
+ * other files  may still want inline!
+ *
+ * only do the define inline when in the simclilst.c  code
+ */
+
 /* Be friend of both C90 and C99 compilers */
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
     /* "inline" and "restrict" are keywords */
 #else
+#ifdef SIMCLIST_C_INTERNAL
 #   define inline           /* inline */
+#endif /* SIMCLIST_C_INTERNAL */
 #   define restrict         /* restrict */
 #endif
 
--- a/src/common/simclist.h
+++ b/src/common/simclist.h
@@ -45,7 +45,7 @@ typedef INT64   int64_t;
 #include <sys/types.h>
 
 /* bases on OpenSSL's version in e_os2.h  */
-#if !defined(inline) && !defined(__cplusplus)
+#if !defined(inline)
 /* Be friend of both C90 and C99 compilers */
 # if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
    /* "inline" and "restrict" are keywords */
@@ -62,7 +62,8 @@ typedef INT64   int64_t;
 #endif
 
 /* bases on OpenSSL's version in e_os2.h  */
-#if !defined(restrict) && !defined(__cplusplus)
+/* On MacOS  C++ is used for tokend */
+#if !defined(restrict)
 /* Be friend of both C90 and C99 compilers */
 # if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
    /* "inline" and "restrict" are keywords */
--- a/src/common/simclist.h
+++ b/src/common/simclist.h
@@ -44,21 +44,38 @@ typedef INT64   int64_t;
 #include <errno.h>
 #include <sys/types.h>
 
-/* OpenSSL-1.1 e_os2.h does #if !defined(inline)
- * the following causes problems. One of them needs to be fixed
- * other files  may still want inline!
- *
- * only do the define inline when in the simclilst.c  code
- */
-
+/* bases on OpenSSL's version in e_os2.h  */
+#if !defined(inline) && !defined(__cplusplus)
 /* Be friend of both C90 and C99 compilers */
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
-    /* "inline" and "restrict" are keywords */
-#else
-#ifdef SIMCLIST_C_INTERNAL
-#   define inline           /* inline */
-#endif /* SIMCLIST_C_INTERNAL */
-#   define restrict         /* restrict */
+# if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+   /* "inline" and "restrict" are keywords */
+#  define simclist_inline  inline
+# elif defined(__GNUC__) && __GNUC__>=2
+#  define simclist_inline  __inline__
+# elif defined(_MSC_VER)
+#  define simclist_inline __inline
+# else
+#  define simclist_inline
+# endif
+#else    /* use what caller wants as inline  may be from config.h */
+#   define simclist_inline  inline           /* inline */
+#endif
+
+/* bases on OpenSSL's version in e_os2.h  */
+#if !defined(restrict) && !defined(__cplusplus)
+/* Be friend of both C90 and C99 compilers */
+# if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+   /* "inline" and "restrict" are keywords */
+#  define simclist_restrict restrict         /* restrict */
+# elif defined(__GNUC__) && __GNUC__>=2
+#  define simclist_restrict __restrict__
+# elif defined(_MSC_VER)
+#  define simclist_restrict __restrict
+# else
+#  define simclist_restrict
+# endif
+#else    /* use what caller wants as restrict may be from config.h */
+#   define simclist_restrict  restrict
 #endif
 
 
@@ -145,7 +162,7 @@ typedef list_hash_t (*element_hash_computer)(const void *el);
  * @param serialize_buffer  reference to fill with the length of the buffer
  * @return                  reference to the buffer with the serialized data
  */
-typedef void *(*element_serializer)(const void *restrict el, uint32_t *restrict serializ_len);
+typedef void *(*element_serializer)(const void *simclist_restrict el, uint32_t *simclist_restrict serializ_len);
 
 /**
  * a function for un-serializing an element.
@@ -162,7 +179,7 @@ typedef void *(*element_serializer)(const void *restrict el, uint32_t *restrict 
  * @param data_len          reference to the location where to store the length of the data in the buffer returned
  * @return                  reference to a buffer with the original, unserialized representation of the element
  */
-typedef void *(*element_unserializer)(const void *restrict data, uint32_t *restrict data_len);
+typedef void *(*element_unserializer)(const void *simclist_restrict data, uint32_t *simclist_restrict data_len);
 
 /* [private-use] list entry -- olds actual user datum */
 struct list_entry_s {
@@ -222,7 +239,7 @@ typedef struct {
  * @param l     must point to a user-provided memory location
  * @return      0 for success. -1 for failure
  */
-int list_init(list_t *restrict l);
+int list_init(list_t *simclist_restrict l);
 
 /**
  * completely remove the list from memory.
@@ -233,7 +250,7 @@ int list_init(list_t *restrict l);
  *
  * @param l     list to destroy
  */
-void list_destroy(list_t *restrict l);
+void list_destroy(list_t *simclist_restrict l);
 
 /**
  * set the comparator function for list elements.
@@ -247,7 +264,7 @@ void list_destroy(list_t *restrict l);
  *
  * @see element_comparator()
  */
-int list_attributes_comparator(list_t *restrict l, element_comparator comparator_fun);
+int list_attributes_comparator(list_t *simclist_restrict l, element_comparator comparator_fun);
 
 /**
  * set a seeker function for list elements.
@@ -261,7 +278,7 @@ int list_attributes_comparator(list_t *restrict l, element_comparator comparator
  *
  * @see element_seeker()
  */
-int list_attributes_seeker(list_t *restrict l, element_seeker seeker_fun);
+int list_attributes_seeker(list_t *simclist_restrict l, element_seeker seeker_fun);
 
 /**
  * require to free element data when list entry is removed (default: don't free).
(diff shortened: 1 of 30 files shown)


frankmorgner:
I suppose this changes the requirement for the minimum version of OpenSSL in `configure.ac`.

frankmorgner:
CI testing for the new version would also be nice.

dengert:
The intent was not to have to change the minimum requirement, which configure.ac says is 0.9.7. 
But the oldest OpenSSL I have compiled the code agasinst is 1.0.1g. 

        https://www.openssl.org/policies/releasestrat.html
Says 
      Support for version 0.9.8 will cease on 2015-12-31 
      Support for version 1.0.0 will cease on 2015-12-31
      Version 1.0.1 will be supported until 2016-12-31
      Version 1.0.2 will be supported until 2019-12-31

So only 1.0.1 and above are supported as of today. 
Do we want newer versions of OpenSC to use unsupported versions of OpenSSL?
What OS versions still provide older OpenSSL  versions? 

I would vote th have the OpenSC github version require 1.0.1.

dengert:
What versions of OpenSSL to we test against today? Do we want to test against the 1.1-pre2,
or wait till it is released?

OpenSSL says:
"Version 1.1.0 will (moderately) break source compatibility (for example we will make most structures opaque etc)."

"moderately" is an understatment...
Many of the changes I made were to use functions (or macros in older versions) that are in 1.0.0 and above (any maybe 0.9.7) to access information in opaque structures. 
For example:
     - 	n = i2d_X509_NAME(x->cert_info->subject, NULL);
    +	n = i2d_X509_NAME(X509_get_subject_name(x), NULL);

1.1 is still pre2 and not released yet. 

Most modules that were changes are just includes for  header files that are already included by 
OpenSSL header files. OpenSSL 1.1 does not automaticly include them, these should not be a problem.

But the 5 modules need to be looked at closer, even if run with older OpenSSL versions.

@CardContact 
please look at  sc-hsm-tool.c

frankmorgner:
I fully agree that we must support 1.0.1 and above. Older versions are optional, but I think your changes introduced functions that were not present in 0.9.7/8. Could you check what's the current minimum version and change https://github.com/OpenSC/OpenSC/blob/master/configure.ac#L483-L501 accordingly?

I think this PR should be left open until 1.1.0 is finally released.

mouse07410:
I concur: support of OpenSSL-1.0.1 and above is required, older versions don't have to be supported. And it would be nice to be compatible with 1.1.x.

dengert:
In response to what versions of OpenSSL is the minimum, I have compiled this PR against OpenSSL-0.9.8zh  which is the oldest veriosn available from OpenSSL. with out problems. 
It was released on 2015-Dec-03. 

 https://www.openssl.org/source/openssl-0.9.8zh.tar.gz

If someone has a verision or 0.9.7, I (or they) could also try it. Let me know where there is a 0.9.7 version available.

dengert:
For 1.1 @CardContact  to have a look at sc-hsm-tool.c. I could hahck at it, but would like them to look
at the use of BIGNUM as in 1.1 it is hidden and must be acccesed by pointers and functions only.

src/pkcs15init/pkcs15-oberthur-awp.c
```diff
@@ -996,6 +996,20 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
...
+	
+	/* TODO the der encoding of a ANS1_INTEGER is a TLV, the original code only as using the 
+	 * i2c_ASN1_INTEGER which is not in OpenSSL 1.1  
+	 * It was adding the tag V_ASN1_INTEGER and the one byte length back in in effect creating
+	 * a DER encoded ASN1_INTEGER
+	 * So we can simplifty the code and make compatable with OpenSSL 1.1. This needs to be tested
+	 */
+	ci->serial.value = NULL;
+	r = ci->serial.len = i2d_ASN1_INTEGER(X509_get_serialNumber(x), &ci->serial.value);
+	SC_TEST_RET(ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_INTERNAL, "AWP encode cert failed: cannot get serialNumber");
```
frankmorgner: You need to translate the OpenSSL error code to an OpenSC error code. Currently this will never work, and always returns `SC_ERROR_INTERNAL`.

src/tools/pkcs11-tool.c
```diff
@@ -1771,36 +1777,38 @@ static void	parse_certificate(struct x509cert_info *cert,
...
 		util_fatal("issuer name too long");
 	/* green light, actually do it */
 	p = cert->issuer;
-	n = i2d_X509_NAME(x->cert_info->issuer, &p);
+	n =i2d_X509_NAME(X509_get_issuer_name(x), &p);
 	cert->issuer_len = n;
 
 	/* check length first */
-	n = i2d_ASN1_INTEGER(x->cert_info->serialNumber, NULL);
+	/* TODO fix this */
```
frankmorgner: what needs to be fixed here?

src/pkcs15init/pkcs15-oberthur-awp.c
```diff
@@ -996,6 +996,20 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
 	/*
 	 * serial number
 	 */
+#if 1
+	
+	/* TODO the der encoding of a ANS1_INTEGER is a TLV, the original code only as using the 
+	 * i2c_ASN1_INTEGER which is not in OpenSSL 1.1
```
frankmorgner: Although I appreciate this simplification, it's not very helpful when it is untested. Wouldn't it be safe to simply replace `i2c_ASN1_INTEGER` with `i2d_ASN1_INTEGER`?

src/tools/pkcs11-tool.c
dengert: On this one, I forgot to remove 
    /* TODO fix this */

    So it is fixed.
    
    
    On 1/15/2016 3:06 AM, Frank Morgner
      wrote:
    
    
      In src/tools/pkcs11-tool.c:
      >  	cert->issuer_len = n;
>  
>  	/* check length first */
> -	n = i2d_ASN1_INTEGER(x->cert_info->serialNumber, NULL);
> +	/* TODO fix this */

      what needs to be fixed here?
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

src/pkcs15init/pkcs15-oberthur-awp.c
dengert: Sorry I have no way to test this and i2c_ASN1_INTEGER is  not
    supported in OpenSSL-1.1, so some change has to be made. 
    Someone who has the cards will have to test it. 
    
    In effect the original code was using i2c_ASN1_INTEGER to get the
    content then converting it to DER. 
    The patch uses i2d_ASN1_INTEGER to get the DER directly. But  the
    error checking is wrong. 
    
    I will fix the error checking.  
    
    For now the #if 1 could be changed to #if OPENSSL_VERSION_NUMBER
    >= 0x10100000L
    but then the code would not be tested until someone built with 1.1.
    The simplification should work with any version of OpenSSL so could
    introduced to the code base once I fix the error checking. 
     
    
    
    On 1/15/2016 3:12 AM, Frank Morgner
      wrote:
    
    
      In src/pkcs15init/pkcs15-oberthur-awp.c:
      > @@ -996,6 +996,20 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
>  	/*
>  	 * serial number
>  	 */
> +#if 1
> +	
> +	/* TODO the der encoding of a ANS1_INTEGER is a TLV, the original code only as using the 
> +	 * i2c_ASN1_INTEGER which is not in OpenSSL 1.1  

      Although I appreciate this simplification, it's not very
        helpful when it is untested. Wouldn't it be safe to simply
        replace i2c_ASN1_INTEGER with i2d_ASN1_INTEGER?
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

frankmorgner:
CI does not work on both, Unix and Windows.

dengert:
@frankmorgner Can you look at the appveyor logs and explain why some are failing and some not? 
Looks like they are not finishing. Server problem?

frankmorgner:
```
card-dnie.c(690) : error C2054: expected '(' to follow 'inline' 
card-dnie.c(691) : error C2085: 'init_flags' : not in formal parameter list 
card-dnie.c(691) : error C2143: syntax error : missing ';' before '{'
```
This looks like there is some `;` missing, but I can't see where that would be.

src/common/simclist.h
```diff
@@ -44,11 +44,20 @@ typedef INT64   int64_t;
...
+ *
+ * only do the define inline when in the simclilst.c  code
+ */
+
 /* Be friend of both C90 and C99 compilers */
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
     /* "inline" and "restrict" are keywords */
 #else
+#ifdef SIMCLIST_C_INTERNAL
 #   define inline           /* inline */
```
frankmorgner: *not having this* could cause the problem in `card-dnie.c`

vletoux:
Why not removing the inline keyword in card-dnie.c or to move the content of the function called only once ?
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-dnie.c#L690
Indeed, VS doesn't support "static inline" when compiling as pure C. Changing as C++ or replacing inline by __inline may be an alternative.

dengert:
OpenSSL 1.1 does a much better job of handling inline. in e_os2.h 
     They changed all their  use of inline to ossl_inline but have a
    bug.  
    
    https://rt.openssl.org/Ticket/Display.html?id=4245
    
    We could do something similar replace inline with osc_inline  or
    just remove the inline from card-dnie.c?
    
    Once the OpenSC bug is fixed, we could just leave the simclist.c and
    simclist.h alone.  But the problem of inline will show up again. 
    
    
    
    On 1/20/2016 6:37 AM, vletoux wrote:
    
    
      Why not removing the inline keyword in card-dnie.c or to move
        the content of the function called only once ?
        https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-dnie.c#L690
        Indeed, VS doesn't support "static inline" when compiling as
        pure C. Changing as C++ or replacing inline by __inline may be
        an alternative.
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

frankmorgner:
remove the inline use in card-dnie.c

viktorTarasov:
Do we have any sensible advantage in using _inline_ and _restrtict_ keywords in card project ?
I would propose to remove them both from _card-dnie_ and _simclist_.

@dengert 
> If someone has a verision or 0.9.7 ...

In _0.9.7_ _EVP_CIPHER_CTX_new_ is not defined,
and in _1.1.0-pre3_ the _EVP_CIPHER_CTX_init_ is still defined for compatibility.

Probably we should not try to keep _0.9.7_ supported;
or implement our own wrappers, to not use everywhere macros _#if OPENSSL..._

Also I would suggest to reorganize the PR into commits by changes in OpenSSL,
at least for the few ones with the biggest impact onto OpenSC sources, like:
- _EVP_CIPHER_CTX_new_ vs. _EVP_CIPHER_CTX_init_;
- _X509_get_xxx_ vs. _cert->cert_info->xxx;
...

dengert:
I will download openssl-0.9.7m and give it a try. 

Yes the mods need to be reorganized especially for EVP_CIPHER_CTX I hope to look at this later this week.

dengert:
We have way to many \#if OPENSSL_VERSION_NUMBER in the code. 
I would like to propose that we do the following:

Modify the OpenSC code to use the 1.1 API when possible. Some many structures are now hidden
and OpenSSL has provided for years macros, and now functions to access the hidden structures.
It would be easier to write to the 1.1 API, then trying to use some of the deprecated functions which will be removed from OpenSSL in following releases. Now is the time to update or use of OpenSSL.  

Write a opensc-openssl-compat.h  header file, that will contain \#define and or inline routines
to convert the 1.1 API to calls into older versions of OpenSSL. This would be included by each module
as needed. 

For example 0.9.7 does not have a  EVP_CIPHER_CTX_new or EVP_CIPHER_CTX_free.
These could defined in opensc-openssl-compat.h  something like ( I have yet to test this)

`#define EVP_CIPHER_CTX_new() ({ \
     EVP_CIPHER_CTX * tmp; \
     tmp = OPENSSL_malloc(sizeof(struct evp_cipher_ctx_st); \
     if (tmp) EVP_CIPHER_CTX_init(&tmp); \
     tmp; \
     })
`

@viktorTarasov 
I now have a OpenSSL-0.9.7m build to test with.

Comments?

dengert:
@viktorTarasov  @frankmorgner @CardContact 

Have a look at this latest addition. It has been tested by on 0.9.7 as wel as 0.9.8, 1.0.1, 1.0.2 and 1.1.0-pre4. 

@CardContact  Please have a look at sc-hsm-tool.c. It has a lot of changes because 
OpenSSL requires the application to to use BN_new.

dengert:
If need be I can squish these changes down to 1 commit.

dengert:
This PR is being replaced by #717. Please add comments to #717

PR #655, opened 2016-01-06
Removing all memory leaks in DNIe module and use of SM functions.

germanblanco:
This is related to
`dnie_transmit_apdu` should not modify the original APDU #610
And to
Fixing some memory leaks in DNIe. #606

As far as I can see, there are no more memory leaks in the DNIe module.
The wrapping and unwrapping functions of OpenSC SM module are used now, however I haven't been able to move the "tricks" in dnie_transmit_apdu to these functions ... yet.
I have done a lot of cleanup, note that the change involves around 380 deleted lines and less than 200 added (if I remember correctly) and these added lines include the OpenSC SM get and free which are not that useful at the moment.
I have tested this code with the Afirma application, the FNMT web and valgrind pkcs11-tool. It might be my imagination, but it even seems faster than the previous version to me.
I would like to have this changes in, and then move to work on support of DNIe 3.0, which seems the most important thing to do now. 
Please consider reviewing and merging these changes.

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -729,6 +729,62 @@ int dnie_match_card(struct sc_card *card)
 	LOG_FUNC_RETURN(card->ctx, result);
 }
 
+static int dnie_sm_free_wrapped_apdu(struct sc_card *card,
+		struct sc_apdu *plain, struct sc_apdu **sm_apdu)
+{
+	struct sc_context *ctx = card->ctx;
+	int rv = SC_SUCCESS;
+
+	LOG_FUNC_CALLED(ctx);
+	if (!sm_apdu)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+	if (!(*sm_apdu))
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+
+	plain->resp = (*sm_apdu)->resp;
+	plain->resplen = (*sm_apdu)->resplen;
+	plain->sw1 = (*sm_apdu)->sw1;
+	plain->sw2 = (*sm_apdu)->sw2;
+
+	if (((*sm_apdu)->data) != plain->data)
+		free((unsigned char *) (*sm_apdu)->data);
+	free(*sm_apdu);
+	*sm_apdu = NULL;
+
+	LOG_FUNC_RETURN(ctx, rv);
+}
+
+static int dnie_sm_get_wrapped_apdu(struct sc_card *card,
+		struct sc_apdu *plain, struct sc_apdu **sm_apdu)
+{
+	struct sc_context *ctx = card->ctx;
+	struct sc_apdu *apdu = NULL;
+	int rv = SC_SUCCESS;
+
+	LOG_FUNC_CALLED(ctx);
+	if (!plain || !sm_apdu)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+
+	*sm_apdu = NULL;
+	//construct new SM apdu from original apdu
+	apdu = calloc(1, sizeof(struct sc_apdu));
+	if (!apdu)
+		return SC_ERROR_OUT_OF_MEMORY;
+
+	memcpy(apdu, plain, sizeof(sc_apdu_t));
+
+	if (rv != SC_SUCCESS) {
+		dnie_sm_free_wrapped_apdu(card, NULL, &apdu);
+		goto err;
+	}
+
+	*sm_apdu = apdu;
+	apdu = NULL;
+err:
+	free(apdu);
+	LOG_FUNC_RETURN(ctx, rv);
+}
+
 /**
  * OpenDNIe card structures initialization.
  *
@@ -759,8 +815,9 @@ static int dnie_init(struct sc_card *card)
 #ifdef ENABLE_SM
 	/** Secure messaging initialization section **/
 	memset(&(card->sm_ctx), 0, sizeof(sm_context_t));
-	card->sm_ctx.ops.get_sm_apdu = NULL;
-	card->sm_ctx.ops.free_sm_apdu = NULL;
+	card->sm_ctx.ops.get_sm_apdu = dnie_sm_get_wrapped_apdu;
+	card->sm_ctx.ops.free_sm_apdu = dnie_sm_free_wrapped_apdu;
+	card->sm_ctx.sm_mode = SM_MODE_NONE;
 #endif
 
 	init_flags(card);
@@ -1293,6 +1350,10 @@ static int dnie_get_challenge(struct sc_card *card, u8 * rnd, size_t len)
 	*/
 	while (len > 0) {
 		size_t n = len > 8 ? 8 : len;
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, 0x84, 0x00, 0x00);
+		apdu.le = 8;
+		apdu.resp = buf;
+		apdu.resplen = 8;	/* include SW's */
 		result = dnie_transmit_apdu(card, &apdu);
 		if (result != SC_SUCCESS) {
 			dnie_free_apdu_buffers(&apdu, buf, 8);
--- a/src/libopensc/cwa-dnie.c
+++ b/src/libopensc/cwa-dnie.c
@@ -619,6 +619,13 @@ static int dnie_create_pre_ops(sc_card_t * card, cwa_provider_t * provider)
 	return sc_card_ctl(card, SC_CARDCTL_GET_SERIALNR, &serial);
 }
 
+static int dnie_create_post_ops(sc_card_t * card, cwa_provider_t * provider)
+{
+        card->sm_ctx.sm_mode = SM_MODE_TRANSMIT;
+
+        return SC_SUCCESS;
+}
+
 /**
  * Main entry point for DNIe CWA14890 SM data provider.
  *
@@ -638,7 +645,7 @@ cwa_provider_t *dnie_get_cwa_provider(sc_card_t * card)
 
 	/* pre and post operations */
 	res->cwa_create_pre_ops = dnie_create_pre_ops;
-	res->cwa_create_post_ops = NULL;
+	res->cwa_create_post_ops = dnie_create_post_ops;
 
 	/* Get ICC intermediate CA  path */
 	res->cwa_get_icc_intermediate_ca_cert =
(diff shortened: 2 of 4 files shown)


src/libopensc/card-dnie.c
```diff
@@ -729,6 +729,62 @@ int dnie_match_card(struct sc_card *card)
...
+	LOG_FUNC_CALLED(ctx);
+	if (!sm_apdu)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+	if (!(*sm_apdu))
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+
+	plain->resp = (*sm_apdu)->resp;
+	plain->resplen = (*sm_apdu)->resplen;
+	plain->sw1 = (*sm_apdu)->sw1;
+	plain->sw2 = (*sm_apdu)->sw2;
```
frankmorgner: Does the DNIe have authenticated status bytes? If so, they should be used instead of the status bytes of the encrypted APDU.

src/libopensc/card-dnie.c
```diff
@@ -729,6 +729,62 @@ int dnie_match_card(struct sc_card *card)
...
+	int rv = SC_SUCCESS;
+
+	LOG_FUNC_CALLED(ctx);
+	if (!sm_apdu)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+	if (!(*sm_apdu))
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+
+	plain->resp = (*sm_apdu)->resp;
+	plain->resplen = (*sm_apdu)->resplen;
```
frankmorgner: If some memory for `plain->resp` has already been initialized this introduces a memory leak. If I'm not mistaken then `get_sm_apdu`/`free_sm_apdu` should only do the memory handling for the encrypted APDU. Allocating memory of the unencrypted response data is managed by the application (i.e. something like opensc-pkcs11 or opensc-explorer).

src/libopensc/cwa-dnie.c
```diff
@@ -905,9 +814,20 @@ static int dnie_wrap_apdu(sc_card_t * card, sc_apdu_t * apdu)
...
+	sc_context_t *ctx;
+	cwa_provider_t *provider = NULL;
+	ctx=card->ctx;
+	provider = GET_DNIE_PRIV_DATA(card)->cwa_provider;
+	if ((provider->status.session.state == CWA_SM_ACTIVE) &&
+		(card->sm_ctx.sm_mode == SM_MODE_TRANSMIT)) {
+		res = dnie_wrap_apdu(card, apdu);
+		LOG_TEST_RET(ctx, res, "Error in dnie_wrap_apdu process");
+		res = cwa_decode_response(card, provider, apdu);
+		LOG_TEST_RET(ctx, res, "Error in cwa_decode_response process");
```
frankmorgner: what's the point in encrypting and immediately decrypting the APDU?

src/libopensc/card-dnie.c
germanblanco: Thanks a lot for reviewing this so quickly!!!

DNIe does have authenticated status bytes, however they are not used here. I know this is a bit twisted, but it is the only way I have managed to make it work so far. dnie_sm_free_wrapped_apdu doesn't actually decrypt the APDU, it just copies the encrypted response and the status bytes of the encrypted APDU to "plain" (which I have just noticed, it is an awful choice of a name). Decryption still takes place in the end of dnie_transmit_apdu, cwa-dnie.c:825.
My final goal here is to remove dnie_transmit_apdu entirely and put all wrapping and unwrapping in dnie_sm_free_wrapped_apdu and dnie_sm_get_wrapped_apdu in the card-dnie.c module. I use the same names for the static methods in cwa-dnie.c and card-dnie.c, for me it is not too bad since they perform the same task and the goal is to move code from those in cwa-dnie.c to those in card-dnie.c, but I should have clarified this.
As it is right now:
- dnie_sm_free_wrapped_apdu and dnie_sm_get_wrapped_apdu in card-dnie.c just copy the encrypted APDU back and forth. They are use from the OpenSC SM code.
- dnie_sm_get_wrapped_apdu in cwa-dnie.c encodes the APDU and modifies some of the fields. It is called from dnie_transmit_apdu.
- dnie_sm_free_wrapped_apdu in cwa-dnie.c copies the response to the plain APDU, which is decoded in dnie_transmit_apdu, although decoding could be moved within this method without problems. It is also called from dnie_transmit_apdu.
It is clearly not the prettiest pull request I have done, but it has taken me around a month to reach this far, and I would like to get some (working) partial result in before continuing. Your input on how to proceed is highly appreciated.

src/libopensc/cwa-dnie.c
germanblanco: This is again a bad choice of names. dnie_wrap_apdu does not only wrapping, it encodes and transmits the APDU. The response is then decoded after transmission.

src/libopensc/card-dnie.c
germanblanco: These two functions in card-dnie.c just use a different instance of APDU with exactly the same information. They could be replaced with this:
```C
static int dnie_sm_free_wrapped_apdu(struct sc_card *card,
		struct sc_apdu *plain, struct sc_apdu **sm_apdu)
{
	return SC_SUCCESS;
}

static int dnie_sm_get_wrapped_apdu(struct sc_card *card,
		struct sc_apdu *plain, struct sc_apdu **sm_apdu)
{
	*sm_apdu = plain;
	return SC_SUCCESS;
}
```

src/libopensc/card-dnie.c
frankmorgner: OK, I understand. But why do you continue carrying this historic burden? This still leaves the code a little bit confusing.

 You could rewrite the code so that `dnie_sm_get_wrapped_apdu()` encrypts the APDU and `dnie_sm_free_wrapped_apdu()` decrypts the response. This allows removing the special case in `dnie_transmit_apdu()`, which checks for `SM_MODE_TRANSMIT`. This is already done via `sc_transmit_apdu()` (https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/apdu.c#L391 to be exactly).

germanblanco: That is what I am trying to do.
I haven't been able to do it in one shot yet.
It might look easy to do, maybe it is and I just haven't found the way yet. If I move the encoding and the surrounding APDU modifications to dnie_sm_get_wrapped_apdu I get APDU validation errors that I haven't been able to sort out so far. If I move the decoding to dnie_sm_free_wrapped_apdu, I get decoding errors.
I need to work a bit on how to make the current transformations in a different place so that they have the same effect.
I certainly do want to remove all this historic burden as you call it.

frankmorgner: It's totally OK to take multiple steps in the process, but your comment on working on DNIe 3.0 made me think that this will introduce new problems that will lower the chances of disentangling the SM stuff some day.

germanblanco: I can spend a few more days in the disentangling and see how far I get.
It would be good if @miguel-cv and @rickyepoderi please take a look at this code in any case. The enveloping has been removed, we need to check if that will cause failures for bigger input data, and in that case, we need to think how to make this enveloping work with the wrapping.

frankmorgner:
> I can spend a few more days in the disentangling and see how far I get.
It would be good if @miguel-cv and @rickyepoderi please take a look at this code in any case. The enveloping has been removed, we need to check if that will cause failures for bigger input data, and in that case, we need to think how to make this enveloping work with the wrapping.

OK, please send a notification in case you encounter problems with the current state of OpenSC's SM framework.

rickyepoderi:
I've been on vacation and then quite busy, sorry for not answering. I have tested the branch for a while and it has problems when the SM is broken. @germanblanco you have to be sure to return SC_ERROR_SECURITY_STATUS_NOT_SATISFIED in order to re-login when the SM is not OK. With this little change it works again (but check it carefully please):

```
diff --git a/src/libopensc/cwa-dnie.c b/src/libopensc/cwa-dnie.c
index 4a7d58e..2f48599 100644
--- a/src/libopensc/cwa-dnie.c
+++ b/src/libopensc/cwa-dnie.c
@@ -702,6 +702,8 @@ static int dnie_sm_free_wrapped_apdu(struct sc_card *card,
        if (rv == SC_SUCCESS) {
                plain->resp = (*sm_apdu)->resp;
                plain->resplen = (*sm_apdu)->resplen;
+       } else if (rv == SC_ERROR_SM) {
+               rv = SC_ERROR_SECURITY_STATUS_NOT_SATISFIED;
        }
 
        if (((*sm_apdu)->data) != plain->data)
@@ -804,7 +806,7 @@ static int dnie_wrap_apdu(sc_card_t * card, sc_apdu_t * apdu)
                msg = "Error in dnie_transmit_apdu process";
        }
 
-       dnie_sm_free_wrapped_apdu(card, apdu, &wrapped);
+       res = dnie_sm_free_wrapped_apdu(card, apdu, &wrapped);
 
        if (msg)
                sc_log(ctx, msg);
```

Besides I don't understand the whole modification. I think that the idea is making opendnie in the opensc way. In my humble opinion that goal means removing completely dnie_transmit_apdu and using sc_transmit_apdu throughout all dnie code. I suppose that there are issues with that approach. @germanblanco can you explain why dnie_transmit_apdu is needed? (Sorry, but this is very complicated).

germanblanco:
Hola Ricardo!
Thank you for the change, I will take it in and test it.
The goal is to remove dnie_transmit_apdu completely, but this is an intermediate step. I am still working on this change and I hope to be able to propose a different Pull Request without dnie_transmit_apdu in a few weeks.
The reason for this intermediate step is that I hit two difficult points in the process:
1 - Is handling all the formats of APDU so that the APDU from the DNIe module to the SM module passes all check_sw checks and the correct data.
2 - Is putting together the complete response and decoding it in separate steps.

This is why in this Pull Request the decoding and encoding is done outside of the sm_wrap and unwrap functions. I am trying to move them in now, but I proposed the Pull Request with all the other changes that I had done to get feed back on them.

Regards,

/German

frankmorgner:
@germanblanco what's the current status, do you need any help? If I am not mistaken you wanted to
- integrate the fix proposed by @rickyepoderi 
- check if a deeper integration in OpenSC's SM framework is possible

germanblanco:
Hello Frank,

I believe the integration is possible and I already integrated the fix
proposed by @rickyepoderi in a local branch.
Unfortunately, I haven't had time to progress in the last weeks.
I have seen that you are preparing for the next release already, right?
It is going to be hard to make it before that, I guess.

Best regards,

/German

frankmorgner:
I made the pre-release only to appease the OS X users that were complaining about Apple's latest changes with El Capitan. It will still take some time for the final release.

If you need more time, we could start by integrating your code as is (with the fix) to at least get rid of the memory leaks.

germanblanco:
I have work a bit on it and it seems it is very close to working now.
So I will give it a push.
I hope to have something that works in two weeks.

germanblanco:
Hello,

Could you please help me solve what seems to be the last problem with this?
I am using now only the wrapping methods of SM, the encoding works fine.
However, for the decoding I think that I would need to be able to modify
the executing when the complete response is received. As far as I see, it
seems that I can only override get_response, and that gives me only a
partial result.
The current code calls sc_transmit_apdu, and then performs decoding, is
that the only way to do it?

Regards,

/German Blanco

germanblanco:
In case you want to take a look, my current code is here:
https://github.com/germanblanco/OpenSC/tree/dnie_use_sm

The decoding is done in cwa-dnie.c:699

Regards,

/German

dengert:
Could we define another SC_APDU_FLAGS_* and add code to apdu.c to test for it? The code could then call
your routine to do what you need.

On 2/25/2016 3:18 PM, Germán Blanco wrote:
> Hello,
>
> Could you please help me solve what seems to be the last problem with this?
> I am using now only the wrapping methods of SM, the encoding works fine.
> However, for the decoding I think that I would need to be able to modify
> the executing when the complete response is received. As far as I see, it
> seems that I can only override get_response, and that gives me only a
> partial result.
> The current code calls sc_transmit_apdu, and then performs decoding, is
> that the only way to do it?
>
> Regards,
>
> /German Blanco
>
> On Thu, Feb 25, 2016 at 7:34 AM, German Blanco <german.blanco@gmail.com>
> wrote:
>
>  > I have work a bit on it and it seems it is very close to working now.
>  > So I will give it a push.
>  > I hope to have something that works in two weeks.
>  >
>  > On Wed, Feb 24, 2016 at 8:56 AM, Frank Morgner <notifications@github.com>
>  > wrote:
>  >
>  >> I made the pre-release only to appease the OS X users that were
>  >> complaining about Apple's latest changes with El Capitan. It will still
>  >> take some time for the final release.
>  >>
>  >> If you need more time, we could start by integrating your code as is
>  >> (with the fix) to at least get rid of the memory leaks.
>  >>
>  >> —
>  >> Reply to this email directly or view it on GitHub
>  >> <https://github.com/OpenSC/OpenSC/pull/655#issuecomment-188130903>.
>  >>
>  >
>  >
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/655#issuecomment-188993112>.
>

-- 

germanblanco:
I have put a proposal in this branch:
https://github.com/germanblanco/OpenSC/tree/dnie_use_sm_and_custom_decode
Please check the changes in sm.h and apdu.c. I believe this is what you mean, only using a function pointer instead of a flag.
This finally removes the need for the "dnie_transmit_apdu" function, which would be great. I would need to test it more thoroughly, but at least it passes the tests in "pkcs11-tool --module /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so -l -t".
Your feedback will be very appreciated.

frankmorgner:
You're still using `dnie_transmit_apdu` (which performs wrapping/unwrapping) as replacement for `sc_transmit_apdu` (which performs wrapping/unwrapping). A naive question: Why don't you do the following?
1. in `card-dnie.c`: replace all occurences of `dnie_transmit_apdu` with `sc_transmit_apdu`
2. rewrite `dnie_sm_free_wrapped_apdu` so that it effectively maps to `cwa_decode_response` and then frees the allocate SM buffers.
I think this should solve your problem above. (The `free_sm_apdu` is actually a misnomer, because it does not only `free` the buffers, but it also decrypts the SM APDU...)

germanblanco:
@frankmorgner, if you look at the branch I sent in the last message:
https://github.com/germanblanco/OpenSC/tree/dnie_use_sm_and_custom_decode
I have replaced all occurences of dnie_transmit_apdu as you suggest. I guess you are talking about the changes in the previous message.
In order to do this, I need the change in apdu.c and sm.h. I have tried to "rewrite dnie_sm_free_wrapped_apdu so that it effectively maps to cwa_decode_response", but it doesn't work. In order to decode, I need the full response as in this latest branch.

frankmorgner:
Yes, my comment was two hours late and referred to your previous message.

I think your latest changes not only obsoletes `dnie_transmit_apdu` but also the call to `dnie_free_apdu_buffers` in error cases...

Looking at
https://github.com/germanblanco/OpenSC/commit/b289c9eb90dd1e28518cd948ab33858149bcad3c#diff-6b74859ea2d4e1aad1102d7825bd537bR514, the global change you make is the special treatment of the GET RESPONSE command when used with SM. Could you give more details on how this works with DNIe?

frankmorgner:
Just as a reminder, we already discussed this here https://github.com/OpenSC/OpenSC/issues/632 and here https://github.com/OpenSC/OpenSC/pull/639

germanblanco:
Thank you, I should have referenced that here, I just forgot it was
discussed before.
Regarding germanblanco@b289c9e#diff-6b74859ea2d4e1aad1102d7825bd537bR514
<https://github.com/germanblanco/OpenSC/commit/b289c9eb90dd1e28518cd948ab33858149bcad3c#diff-6b74859ea2d4e1aad1102d7825bd537bR514>,
this is an entry point to call an specific function for decoding the entire
response. The only module that uses that entry point at the moment would be
dnie, and it calls dnie_sm_decode_apdu in card-dnie.c.
As far as I can see memory allocation works fine. There are no memory leaks.

frankmorgner:
Could you explain how DNIe responds/treats GET RESPONSE commands (with and without SM)? Is there a technical document available?

germanblanco:
I will send what I have in a private message to you.
It is in Spanish, but I am willing to translate any part that you consider
interesting.
In any case, I didn't find a lot of relevant information for this case.

According to the code that is working against the card, the way that it
works seems to be the following:

1 - issue of a command that sends a response
2 - the card indicates that there are N bytes available for reading.
3 - The bytes are read in a loop until no more bytes are available.
4 - all the responses are put together.
5 - If the channel is encrypted, the entire response is decrypted.

It could be that the problem of decrypting each individual response is that
cwa_decode_response does both parsing and decoding. Information in the
encoded response seems to contain tag-length-value elements, and it seems
that a partial response could contain only part of one of this elements
making the parsing impossible. But this is only what I can gather from the
code (and the errors that I get when trying to decode each partial response
separately). If you (or some other reading this) can help me to understand
this better, I will be very happy.

frankmorgner:
In your documentation there are some examples of the usage of GET RESPONSE (unfortunately none of them uses GET RESPONSE multiple times). It seems that though an SM channel is established, GET RESPONSE is issued unencrypted. Hence, I suspect that the datapart of the GET RESPONSE RAPDU should be interpreted as if it came from the original command. So I guess to get some RESP for some command CMD it should be done something like this:

1. encrypt CMD to SM(CMD)
2. RAPDU: 61XX (plain)
3. APDU: GET RESPONSE (plain)
4. RAPDU: DATA+SW1SW2 (plain)
5. if SW1==61 goto 3.
6. Assemble all response data from GET RESPONSE as DATA1||...||DATAn == SM(RESP)
7. decrypt SM(RESP) to RESP

In my opinion it's very strange to allow unencrypted APDUs within an encrypted session. So I'd like to see this solved at the card driver level. Please try the following:

in `dnie_sm_get_wrapped_apdu` do the following:

1. use `SC_APDU_FLAGS_NO_GET_RESP` to avoid OpenSC's handling of GET RESPONSE

in `dnie_sm_free_wrapped_apdu` do the following:

1. If the return code was 61XX continue with 2., otherwise continue with 4.
2. Get all of the response by issuing GET RESPONSE (multiple times if needed). You should issue this command directly via the reader (`card->reader->ops->transmit(card->reader, get_resp_apdu`), because otherwise OpenSC's SM handling would kick in (which is not what we want because your GET RESPONSE is unencrypted)
3. Assamble all the data of the response data and use this as `sm_apdu->resp`.
4. decrypt sm_apdu->resp

frankmorgner:
A cleaner way of implementing the second step in `dnie_sm_free_wrapped_apdu` would be to temporarily disable SM (`SM_MODE_NONE`) and to issue `sc_get_response`. This should get you all the data even if multiple GET RESPONSE commands are needed.

germanblanco:
Is there a way to issue sc_get_response without making it public?
I have just seen it is an static method in apdu.c.

germanblanco:
I understand the interest in handing only common SC features in the core of
OpenSC, and I am very interested in removing this ugly dnie_transmit_apdu
from the DNIe module.
I have tried your proposal. It didn't work straight away, and I don't know
the reason yet. Something fails in some of the cases in the decryption, I
would need to debug it carefully and correct it. But at this stage, now
that we have seen that there seems to be a reason for this ugly
dnie_transmit_apdu after all (namely, a peculiar way to handle get response
under encryption in the DNIe card), I am willing to live with ugly for now.
I would like to prioritize support of DNIe 3.0, since there are people
already demanding it. So I will put forward a pull request removing all
memory leaks from the module and using wrapping and unwrapping within the
SM module, and leaving decryption handling in dnie_transmit_apdu. Once DNIe
3.0 is in, we can go back to removing the dnie_transmit_apdu issue. I don't
believe the addition of DNIe 3.0 support will affect this other issue in
any way. However having SM wrapping and unwrapping in before adding support
for DNIe 3.0 could make a difference.
Pull request is in https://github.com/OpenSC/OpenSC/pull/655

rickyepoderi:
In my little understanding DNIe, under secure channel condition, uses a SC_APDU_CASE_4_SHORT apdu (it always needs a get_response to retrieve the real encoded data). Then the get_response is called to retrieve the encoded data (as many times as needed) and the final data should be decoded to get the real / original response.

The problem is currently opensc calls get_sm_apdu / free_sm_apdu in every call (it doesn't differentiate the get_response itself).  That's why German needs this comparison in "dnie_sm_get_wrapped_apdu":

     if (((plain->cla & 0x0C) == 0) && (plain->ins != 0xC0)) { 

The 0xC0 comparison is to not encode the get_response itself. The get_response is not encoded cos it's just used to retrieve the real encoded response.

German was forced to add another call "decode_sm_apdu" because in the loop inside sc_get_response (apdu.c) the free_sm_apdu is called for each individual response but not for the complete buffer that should be decoded when all the data has been retrieved (maybe calling get_response several times). Now that final decoding is done in the "dnie_transmit_apdu" and that is why that function is needed.

I think the problem with DNIe SM is that it transforms any apdu to a SC_APDU_CASE_4_SHORT type and common/plain get_response is used to get the encoded response. I don't know if this is a typical behavior or a weird one.

frankmorgner:
@rickyepoderi Above, I depicted one way of avoiding SM for GET RESPONSE https://github.com/OpenSC/OpenSC/pull/655#issuecomment-189309318. I also tried to explain how that fits in the overall SM framework.

@germanblanco it's very unfortunate that the efforts are still without a satisfying result.

Just for the record, in `dnie_sm_free_wrapped_apdu` you could have issued a GET RESPONSE via `sc_transmit` with the `SM_MODE_NONE` flag set. This would have triggered the internal `sc_get_response` if even more data was available. This should have worked without exposing `sc_get_response`.

Issue #657, opened 2016-01-14
pkcs11-tool writes some message to stdout making pipes useless


celledge:
I was trying to sign a digest with pkcs11-tool using pipes so that it could be called externally without needing temporary files.
````
pkcs11-tool --module /usr/lib/opensc-pkcs11.so -s -f openssl --id 10 --pin 1234 < ~/test.png.sha256 > ~/test.png.sig
`````
I would get this as output in the signature files:
````
0E G»©«×£Qs¼89/BòK{ÞG(Æ,»-!¼÷t`ÒH;ÕÃÂ©±gkð4WD¥©RÂÃ¿®gâUsing signature algorithm ECDSA
````
This works perfectly if the "Using signature algorithm ECDSA" is removed.

It seems that the culprit is in pkcs11-tool.c line 1430. There are other similar lines scattered throughout the file that would probably be better suited using fprintf(stderr,...).

**Comment by mouse07410** [comment]
```
$ openssl rand -out t256.dat 32
$ pkcs11-tool -s -f openssl --id 02 -i t256.dat -o t256.sig
Using slot 1 with a present token (0x1)
Logging in to "PIV_II (PIV Card Holder pin)".
Please enter User PIN:
Using signature algorithm ECDSA
$ dumpasn1 t256.sig
  0  70: SEQUENCE {
  2  33:   INTEGER
       :     00 F1 B8 1A BD 18 F7 D0 61 A5 F1 BE 9F EA AC 11
       :     D8 22 B4 2B F7 AE 60 4F 2C 7F 95 81 F2 3B 32 46
       :     29
 37  33:   INTEGER
       :     00 CB 36 A6 42 B0 11 5F C6 A8 DA 89 C0 B3 AB 78
       :     97 C0 89 0D EE 65 6C 15 63 51 92 61 AA DE 61 97
       :     13
       :   }

0 warnings, 0 errors.
$ openssl pkeyutl -engine pkcs11 -keyform engine -verify -pubin -inkey id_02 -sigfile t256.sig -in t256.dat
engine "pkcs11" set.
Signature Verified Successfully
$
```

What seems to be the problem?

frankmorgner:
@celledge, your suggestion is reasonable. Please make a pull request while you have your hands on the code!

elad:
Reviving this old issue...

Would it be possible to do the same for *all* commands? For example `--derive` still prints stuff to stdout. :/

Jakuje:
[...]

PR #662, opened 2016-01-20
pkcs11-tool: Switched some printfs to fprintfs to allow pipes to work.

celledge:
As discussed in issue #657, I switched some of the printf statements to fprintf(stderr) that would prevent the use of stdout for sign, decrypt, and digest.


frankmorgner:
thanks

PR #663, opened 2016-01-20
pkcs15-crypt: Allow the use of stdin if no input file is specified.

celledge:
Allow pkcs15-crypt to accept input on stdin if no -i is specified for signing or deciphering. Removes need for temporary files when using pkcs15-crypt from external programs.


frankmorgner:
looks OK.

Could you please document the new default in `doc/tools/pkcs15-crypt.1.xml` and via `pkcs15-crypt -h` (and while you have your hands on it, you may also document the stdout as default for `--output`), thanks!

celledge:
I'm unclear on why one of the checks failed.

frankmorgner:
np, thanks

PR #666, opened 2016-01-25
'Disable' and 'ignore-usage' PKCS#15 application configuration options 

viktorTarasov:
Proposed are two configuration options related to the visibility and using of the on-card PKCS#15 applications:
 -- 'disable' -- to not expose application in pkcs15 framework (and so do not create dedicated slot in PKCS#11) ;
 -- 'ignore_usage' -- ignore 'usage' attribute (in IAS/ECC case it's the SDO's key-usage) when looking for a suitable key slot to import/generate key. Thus allowing the applications, that do not bother with key-usage in template for a new key object, to generate/import key in the slot with restricted usage.

### DIFF
--- a/src/libopensc/dir.c
+++ b/src/libopensc/dir.c
@@ -61,6 +61,7 @@ parse_dir_record(sc_card_t *card, u8 ** buf, size_t *buflen, int rec_nr)
 {
 	struct sc_context *ctx = card->ctx;
 	struct sc_asn1_entry asn1_dirrecord[5], asn1_dir[2];
+	scconf_block *conf_block = NULL;
 	sc_app_info_t *app = NULL;
 	struct sc_aid aid;
 	u8 label[128], path[128], ddo[128];
@@ -82,6 +83,25 @@ parse_dir_record(sc_card_t *card, u8 ** buf, size_t *buflen, int rec_nr)
 	if (r == SC_ERROR_ASN1_END_OF_CONTENTS)
 		LOG_FUNC_RETURN(ctx, r);
 	LOG_TEST_RET(ctx, r, "EF(DIR) parsing failed");
+
+	conf_block = sc_get_conf_block(ctx, "framework", "pkcs15", 1);
+	if (conf_block)   {
+		scconf_block **blocks = NULL;
+		char aid_str[SC_MAX_AID_STRING_SIZE];
+		int ignore_app = 0;
+
+		sc_bin_to_hex(aid.value, aid.len, aid_str, sizeof(aid_str), 0);
+		blocks = scconf_find_blocks(card->ctx->conf, conf_block, "application", aid_str);
+		if (blocks)   {
+			ignore_app = (blocks[0] && scconf_get_str(blocks[0], "disable", 0));
+                        free(blocks);
+		 }
+
+		if (ignore_app)   {
+			sc_log(ctx, "Application '%s' ignored", aid_str);
+			LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+		}
+	}
 
 	app = calloc(1, sizeof(struct sc_app_info));
 	if (app == NULL)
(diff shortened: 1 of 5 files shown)


viktorTarasov:
@JenkinsForOpenSC retest this please
retest this please

viktorTarasov:
@JenkinsForOpenSC retest this please
retest this please

frankmorgner:
Isn't disabling an application what you can achieve by marking it as `protected`?
```
application E828BD080FD2504543432D654944 {
			type = protected;
			model = "ECC eID";
		}
```

viktorTarasov:
@frankmorgner 
> Isn't disabling an application what you can achieve by marking it as protected?

Not. Protected application is (can be) also PKCS#15.
Application type is used to expose to the 'generic' application when only one application is expected, like in minidriver case.

frankmorgner:
You said it, it doesn't get exposed in a certain case. Why not add `disabled` to the possible values of `type`. And while you have your hands on the code, please document the different choices.

frankmorgner:
Please move `ignore-usage` frm the generic level to a iasecc driver conig section, which is similarly done for dnie here https://github.com/OpenSC/OpenSC/blob/master/etc/opensc.conf.in#L154-L165. unfortunately your implementation is currently not usable for other cards

viktorTarasov:
@frankmorgner 
> unfortunately your implementation is currently not usable for other cards

It's usable for all other cards that have more then one PKCS#15 application.

viktorTarasov:
It's more usable on per-application level, and can be used by all cards with more then one application.

viktorTarasov:
@frankmorgner 
> Why not add disabled to the possible values of type.

No,  'type' and 'status' are different meanings and destination in code.

> And while you have your hands on the code, please document the different choices.

You mean configuration options, or something else?

frankmorgner:
>>And while you have your hands on the code, please document the different choices.

>You mean configuration options, or something else?

Yes, (at least to me) it's unclear what the configuration option `type` actually does. If you know exactly what it does, please add some documentation. I can see in the code that all we do expose `generic` applications to PKCS#11, minidriver or tokend. As consequence, the applications of type `protected` are only visible for the OpenSC tools (if at all). From a user's point of view (anything outside the OpenSC world) those applications are effectively disabled. That's why I believe we don't need an other configuration option to achieve this (regardless where it's located in the code). Maybe we can get some more opinions on this.

>>unfortunately your implementation is currently not usable for other cards

>It's more usable on per-application level, and can be used by all cards with more then one application.

I agree that the option `ignore_usage` could have been useful for other cards *if implemented differently*. Unfortunately currently if setting this option for any other card than iasecc will have no effect at all. Leaving this option on a global level will only create confusion, that's why it belongs into the iasecc card driver section.

src/libopensc/dir.c
```diff
@@ -83,6 +84,25 @@ parse_dir_record(sc_card_t *card, u8 ** buf, size_t *buflen, int rec_nr)
...
+
+		sc_bin_to_hex(aid.value, aid.len, aid_str, sizeof(aid_str), 0);
+		blocks = scconf_find_blocks(card->ctx->conf, conf_block, "application", aid_str);
+		if (blocks)   {
+			ignore_app = (blocks[0] && scconf_get_str(blocks[0], "disable", 0));
+                        free(blocks);
+		 }
+
+		if (ignore_app)   {
+			sc_log(ctx, "Application '%s' ignored", aid_str);
```
frankmorgner: the length to be printed should be limited here, either by using `%.s` or by setting `aid_str[SC_MAX_AID_STRING_SIZE-1] = '\0'`

viktorTarasov:
@frankmorgner 
> the length to be printed should be limited here, either by using %.s or by setting aid_str[SC_MAX_AID_STRING_SIZE-1] = '\0'

It's already done [here](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/sc.c#L112)

viktorTarasov:
@frankmorgner 
>Unfortunately ... any other card than iasecc ... 

This option is defined on per-application level without any relation to ias/ecc. Any card that seek for this possibility will use it in it's own card-specific part.

viktorTarasov:
@frankmorgner 
> ... Maybe we can get some more opinions on this.

Agree.

viktorTarasov:
* finally, the _ignore-usage_ is not good idea -- PKCS#15 descriptors will not reflect the real key-slot usage
* _disable-application_ will be applied separately

PR #665, opened 2016-01-24
pkcs11-tool: support write of GOST Public Key object

sfff:
Example:
```bash
// Make keys
$ openssl genpkey -engine gost -out key.pem -algorithm gost2001 -pkeyopt paramset:A
$ openssl pkey -engine gost -in key.pem -outform DER > key.der
$ openssl pkey -engine gost -in key.pem -outform PEM -pubout > key.pub.pem
$ openssl pkey -engine gost -in key.pem -outform DER -pubout > key.pub.der

$ pkcs11-tool --module /tmp/librtpkcs11ecp.so --login --pin "12345678" -y privkey -w key.pem 
Using slot 0 with a present token (0x0)
Created private key:
Private Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022301
  label:      
  Usage:      sign
$ pkcs11-tool --module /tmp/librtpkcs11ecp.so --login --pin "12345678" -y privkey -w key.der 
Using slot 0 with a present token (0x0)
Created private key:
Private Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022301
  label:      
  Usage:      sign
$ pkcs11-tool --module /tmp/librtpkcs11ecp.so --login --pin "12345678" -y pubkey -w key.pub.pem 
Using slot 0 with a present token (0x0)
Created public key:
Public Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022301
  VALUE:      5c32a4cdbad50516c3dd3c92efba7578d711ba48de4a84a3669923f49663ee6d
              48b574aa950ca70e27378b88ad35bf94803adf96123d224f7e7f5f57d49ae640
  label:      
  Usage:      verify
$ pkcs11-tool --module /tmp/librtpkcs11ecp.so --login --pin "12345678" -y pubkey -w key.pub.der 
Using slot 0 with a present token (0x0)
Created public key:
Public Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022301
  VALUE:      5c32a4cdbad50516c3dd3c92efba7578d711ba48de4a84a3669923f49663ee6d
              48b574aa950ca70e27378b88ad35bf94803adf96123d224f7e7f5f57d49ae640
  label:      
  Usage:      verify
$ 
```

PR #669, opened 2016-01-30
Add PEM support when importing RSA private key using pkcs11-tool

vletoux:
1) Support the PEM header "-----BEGIN RSA PRIVATE KEY-----"
2) Add PEM support when importing RSA private key

Key import can be tested with:
openssl.exe genpkey -algorithm RSA -out private_key.pem -pkeyopt rsa_keygen_bits:1024
pkcs11-tool --login --pin <mypin> --write-object private_key.pem --type privkey


frankmorgner:
Isn't that part of https://github.com/OpenSC/OpenSC/pull/665?

vletoux:
nevermind, I didn't see #665

PR #675, opened 2016-02-17
Fix BCD encoding for PIN with hex digits

### LINKED ISSUE #674: Wrong SO-PIN in SmartCard-HSM after changing with pkcs11-tool --change-PIN
Steps to reproduce

pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --login --login-type so --so-pin 3537363231383830 --change-pin --new-pin 0123456789ABCDEF
sc-hsm-tool --initialize --so-pin 0123456789ABCDEF --pin 648219

fails with invalid SO-PIN.

CardContact:
If a provided PIN contains hexadecimal characters, then the characters in the range A to F are not converted correctly into BCD.

This patch adds support for the full range 0-9,A-F and a-f and validates that no other characters are in the PIN.

Fixes #674

### DIFF
--- a/src/libopensc/sec.c
+++ b/src/libopensc/sec.c
@@ -249,7 +249,15 @@ int sc_build_pin(u8 *buf, size_t buflen, struct sc_pin_cmd_pin *pin, int pad)
 			return SC_ERROR_BUFFER_TOO_SMALL;
 		for (i = j = 0; j < pin_len; j++) {
 			buf[i] <<= 4;
-			buf[i] |= pin->data[j] & 0xf;
+			if (pin->data[j] >= '0' && pin->data[j] <= '9') {
+				buf[i] |= pin->data[j] & 0xf;
+			} else if (pin->data[j] >= 'A' && pin->data[j] <= 'F') {
+				buf[i] |= pin->data[j] - 'A' + 10;
+			} else if (pin->data[j] >= 'a' && pin->data[j] <= 'f') {
+				buf[i] |= pin->data[j] - 'a' + 10;
+			} else {
+				return SC_ERROR_INVALID_ARGUMENTS;
+			}
 			if (j & 1)
 				i++;
 		}


dengert:
I agree that the code is not handling the PIN encoding correctly.
But BCD is Binary Coded Decimal, and only 0-9 are valid digits.
The BCD encoding can also have an odd number of digits, and the last byte is then padded with a pad nibble.
The pad nibble is usually "F"and the pin is usually padded with "FF"

So if a user uses a pin like 123F567 The F could be misinterpreted as the end of the BCD field.

Comments in sec.c at lines 210 - 219 give a reference to:
Reference: Global Platform - Card Specification - version 2.0.1' - April 7, 2000
which I don't have.

pcsc10_v2.02.08.pdf  has the bmFormatString, so allowing the pad nibble of "F" could cause problems with pin-pad readers too.

If you want to allow binary pins, maybe your change should be the calling module that does not set the
pin->encoding == SC_PIN_ENCODING_BCD



On 2/17/2016 2:44 PM, CardContact Systems GmbH wrote:
> If a provided PIN contains hexadecimal characters, then the characters in the range A to F are not converted correctly into BCD.
>
> This patch adds support for the full range 0-9,A-F and a-f and validates that no other characters are in the PIN.
>
> Fixes #674 <https://github.com/OpenSC/OpenSC/issues/674>
>
> --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/OpenSC/OpenSC/pull/675
>
>
>         Commit Summary
>
>   * Fix BCD encoding for PIN with hex digits
>
>
>         File Changes
>
>   * *M* src/libopensc/sec.c <https://github.com/OpenSC/OpenSC/pull/675/files#diff-0> (10)
>
>
>         Patch Links:
>
>   * https://github.com/OpenSC/OpenSC/pull/675.patch
>   * https://github.com/OpenSC/OpenSC/pull/675.diff
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/675>.
>

-- 

CardContact:
The SO-PIN is defined as fixed even length, so no padding is used in this case. And this PIN can never be entered on the PIN-PAD as it has 16 digits.

I don't quite understand what problem you'd expect ? Currently the code just masks the lower 4 bits and hasn't caused any issue. With the patch the checking is more restrictive, but still allows A-F if the user wants that. Before you could enter any character and would get unpredictable results.

frankmorgner:
Binary Coded *Decimal* is made for digits, not characters, see https://en.wikipedia.org/wiki/Binary-coded_decimal.

frankmorgner:
The current code should rather throw an error on the bad input...

dengert:
I agree with Frank, the code should check for decimal digits. 

If @CardContact  needs hex digits, in the PIN, maybe add the conversion to the driver of tool,
and specify SC_PIN_ENCODING_ASCII.

CardContact:
Specifying SC_PIN_ENCODING_ASCII doesn't help, as the 16 hexchar SO-PIN must be passed as 8 byte binary in the VERIFY command.

I could also add a new PIN format SC_PIN_ENCODING_HEX, but that would also require a new SC_PCKS15_PIN_TYPE_HEX and changes all over the code.

I guess fixing support for A-F is the better option.

And yes, BCD means digits only, Thanks @Frank ;-)

frankmorgner:
I'd prefer `SC_PIN_ENCODING_RAW` so that the input PIN will be directly piped as binary data from the application to the card driver. So that both ends must be aware what they are doing.

My real question is whether those modifications are really needed:
- No standard application (minidriver, PKCS#11, tokend) will ever use or capture this type of PIN.
- All you are achieving is adding A-F to the allowed character set, but for what reason. Entropy? Not really; you may simply use a longer PIN.
- As far as I know, a binary PIN has no representation in any of the standards involved (PC/SC, ISO 7816-4, PKCS#15, PKCS#11). Hence, it will be of limited use and justifies only modifications within a (proprietary) card driver and a (proprietary) application.

I guess, nobody will ever use this feature, hence forget about it!

CardContact:
> I guess, nobody will ever use this feature, hence forget about it!

Ahhh, no. Customers are actually using this, that's how the issue came up initially.

The SO-PIN at the PKCS#11 layer is actually the Initialization Code at the SmartCard-HSM APDU layer and that is a 64 bit key, aka a 16 digit hexadecimal string. So the story is about preserving the entropy of the key.

Commonly we recommend to generate the Initialization Code using an ECDH operation with a private key on a master device and the public device authentication key. That way the Initialization Code (which is quite powerful as it allows resetting the user PIN) is generated on a per card basis from a key own by the operator. It allows the operator to unblock/rewrite the user PIN or to re-initialize the device.

frankmorgner:
Well, you could change the 16 character limit. Or you switch your applet to accept ASCII encoded PINs instead.

Or, you could set the encoding to `SC_PIN_ENCODING_GLP+1` which is undefined in OpenSC. No magic will happen with the PIN string (padding/converting) will happen until it hits your card driver. Then you can handle the string as you like (i.e. convert it to binary data and add padding if needed).

I don't like introducing a hex encoded PIN as general mechanism for the reasons given above. This would also require new padding semantics, because we can't simply pad with say `0x00` anymore. The strings `"1100"` and `"11"` would both translate to `{0x11,0x00}` with padding. You could do bit-padding, but that requires an additional byte. And if making the PIN longer for padding, you are back at changing the 16 bit limit.

CardContact:
How does allowing A-F in a PIN that is going to be BCD encoded affect the padding rules ?

The way the PIN is interpreted depends on the code in the card. Padding does not at all look at the content of the PIN, it just ensures that the PIN is encoded in the format expected by the card.

And how would changing the 16 character limit or even changing the applet code solve the problem ?

PR #676, opened 2016-02-17
Fix some problems reported by Coverity

frankmorgner:


viktorTarasov:
@frankmorgner 

[Rebased and slightly updated version](https://github.com/viktorTarasov/OpenSC-SM/tree/master_and_pr676) of your patch.
Update mostly concerns the pkcs11/misc.c after yours frankmorgner@fb8ff0b and frankmorgner@7fbb76b. Check for _(ptr=Null && len!=0)_ is added.

frankmorgner:
regarding your rewrite of misc.c in https://github.com/viktorTarasov/OpenSC-SM/commit/f2ad63e6767f4e05d1286d9e491267ea5e160f84, I always have the impression that if we use mulitple exit points we finally forget to release some buffer. This is then only revealed by yet an other iteration of scanning/analyzing and fixing. A single exit point is generally a good programming pattern, it reduces code size, separate boilerplate from the actual logic and has a better readibility.

I'd like to keep misc.c as is.

viktorTarasov:
Ok, I see.
- the essential of proposed update is the missing checks;
- the coding style of proposed update do not differ from the current coding style of the updated file;
- excluding two upper mentioned commits, all other are merged;
- you can add the missing checks and re-submit the PR for _misc.c_, using the coding style that seems to you appropriated.

PR #678, opened 2016-02-18
sc-hsm: Fix BCD encoding bug with SO-PIN

### LINKED ISSUE #674: Wrong SO-PIN in SmartCard-HSM after changing with pkcs11-tool --change-PIN

CardContact:
Fixed #674 at the driver level. Obsoletes #675.


frankmorgner:
good solution!

frankmorgner:
The fix of bad usage of memset could have gone into a separate commit, though (something like https://github.com/frankmorgner/OpenSC/commit/98d5b364eb03e1bf0e223150a7142a61f3dbdf4a :wink:)

PR #679, opened 2016-02-19
GemSAFE ATR + format string fixes

l1k:
Add a GemSAFE ATR and fix a few format string bugs spotted by @dengert in https://github.com/OpenSC/OpenSC/issues/508#issuecomment-138103762

### DIFF
--- a/src/libopensc/card-gemsafeV1.c
+++ b/src/libopensc/card-gemsafeV1.c
@@ -63,6 +63,8 @@ static struct sc_atr_table gemsafe_atrs[] = {
     {"3B:7D:95:00:00:80:31:80:65:B0:83:11:00:C8:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_PTEID, 0, NULL},
     /* Swedish eID card */
     {"3B:7D:96:00:00:80:31:80:65:B0:83:11:00:C8:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_SEEID, 0, NULL},
+    /* European Patent Office epoline card*/
+    {"3b:7d:96:00:00:80:31:80:65:b0:83:02:01:f3:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_SEEID, 0, NULL},
     {NULL, NULL, NULL, 0, 0, NULL}
 };
 
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -415,7 +415,7 @@ sc_pkcs11_signature_update(sc_pkcs11_operation_t *operation,
 		LOG_FUNC_RETURN(context, CKR_DATA_LEN_RANGE);
 	memcpy(data->buffer + data->buffer_len, pPart, ulPartLen);
 	data->buffer_len += ulPartLen;
-	sc_log(context, "data length %li", data->buffer_len);
+	sc_log(context, "data length %u", data->buffer_len);
 	LOG_FUNC_RETURN(context, CKR_OK);
 }
 
@@ -428,7 +428,7 @@ sc_pkcs11_signature_final(sc_pkcs11_operation_t *operation,
 
 	LOG_FUNC_CALLED(context);
 	data = (struct signature_data *) operation->priv_data;
-	sc_log(context, "data length %li", data->buffer_len);
+	sc_log(context, "data length %u", data->buffer_len);
 	if (data->md) {
 		sc_pkcs11_operation_t	*md = data->md;
 		CK_ULONG len = sizeof(data->buffer);
@@ -441,7 +441,7 @@ sc_pkcs11_signature_final(sc_pkcs11_operation_t *operation,
 		data->buffer_len = len;
 	}
 
-	sc_log(context, "%li bytes to sign", data->buffer_len);
+	sc_log(context, "%u bytes to sign", data->buffer_len);
 	rv = data->key->ops->sign(operation->session, data->key, &operation->mechanism,
 			data->buffer, data->buffer_len, pSignature, pulSignatureLen);
 	LOG_FUNC_RETURN(context, rv);


src/libopensc/card-gemsafeV1.c
```diff
@@ -63,6 +63,7 @@ static struct sc_atr_table gemsafe_atrs[] = {
     {"3B:7D:95:00:00:80:31:80:65:B0:83:11:00:C8:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_PTEID, 0, NULL},
     /* Swedish eID card */
     {"3B:7D:96:00:00:80:31:80:65:B0:83:11:00:C8:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_SEEID, 0, NULL},
+    {"3B:7D:96:00:00:80:31:80:65:B0:83:02:01:F3:83:00:90:00", NULL, NULL, SC_CARD_TYPE_GEMSAFEV1_SEEID, 0, NULL},
```
frankmorgner: You could unify all the previous ATRs by adding an ATR mask

l1k:
@frankmorgner: Unifying all ATRs starting with `3B:7D:96` is hindered by the different card types (one has `SC_CARD_TYPE_GEMSAFEV1_GENERIC`, two have `SC_CARD_TYPE_GEMSAFEV1_SEEID`):

3B:7D:96:00:00:80:31:80:65:B0:83:**11:48:C8**:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_GENERIC
3B:7D:96:00:00:80:31:80:65:B0:83:**11:00:C8**:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_SEEID
3B:7D:96:00:00:80:31:80:65:B0:83:**02:01:F3**:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_SEEID

However what I *can* do is add a mask to all the existing ATRs (derived from the Gemalto driver, assuming they know what they're doing). This eliminates two ATRs. I've verified that my two cards (GemSAFE V1 and V3) still work with this commit added, and that there are no overlaps between the newly added masks and any other drivers. Is that what you had in mind?

l1k:
Forgot to mention: The ATR list was originally tab indented but unfortunately 9c7eb8122afb changed this to 4 blanks (which seems to contradict regular OpenSC coding style). I retained that and modified the one line that was still tab indented to use 4 blanks as well, thus avoiding having to change the *entire* ATR list, however I'll be happy to convert everything to tab indentation if preferred.

dengert:
There are PIV cards from GemALto. I have two NIST demo cards with ATR:
3b:7d:96:00:00:80:31:80:65:b0:83:*11:17:d6*:83:00:90:00

https://smartcard-atr.appspot.com/parse

calls the B0:83:11:17:d6 "pre-issuing data"

It also says the 31:80 is
   "Card service data byte:128"
   "Application selection: by full DF name"

Does this mean that all the GEMSAFE cards all have an application?
Do we know what the AID is for each of the GEMSAFE applications?

Could selection by application be a better choice then using the ATR?



On 2/20/2016 7:40 AM, l1k wrote:
> @frankmorgner <https://github.com/frankmorgner>: Unifying all ATRs starting with |3B:7D:96| is hindered by the different card types (one has |SC_CARD_TYPE_GEMSAFEV1_GENERIC|, two have
> |SC_CARD_TYPE_GEMSAFEV1_SEEID|):
>
> 3B:7D:96:00:00:80:31:80:65:B0:83:*11:48:C8*:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_GENERIC
> 3B:7D:96:00:00:80:31:80:65:B0:83:*11:00:C8*:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_SEEID
> 3B:7D:96:00:00:80:31:80:65:B0:83:*02:01:F3*:83:00:90:00 = SC_CARD_TYPE_GEMSAFEV1_SEEID
>
> However what I /can/ do is add a mask to all the existing ATRs (derived from the Gemalto driver, assuming they know what they're doing). This eliminates two ATRs. I've verified that my two cards
> (GemSAFE V1 and V3) still work with this commit added, and that there are no overlaps between the newly added masks and any other drivers. Is that what you had in mind?
>
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/679#issuecomment-186597850>.
>

-- 

l1k:
@dengert:

<blockquote>
Does this mean that all the GEMSAFE cards all have an application?
</blockquote>

No idea.

<blockquote>
Do we know what the AID is for each of the GEMSAFE applications?
</blockquote>

The Gemalto driver supports one AID for GemSAFE V1 cards and three AIDs for GemSAFE V2V3 cards:
````
GemSAFE V1   a0000000180a000001634200 = gemsafe_def_aid[]
GemSAFE V2V3 a0000000180c000001634200 = gemsafe_seeid_aid[]
GemSAFE V2V3 a0000000180e000001634200 = missing in card-gemsafeV1.c
GemSAFE V2V3 a0000000180f000001634200 = missing in card-gemsafeV1.c
````

The AID of the Portuguese eID card that OpenSC supports is nowhere to be found in the Gemalto driver.

<blockquote>
Could selection by application be a better choice then using the ATR?
</blockquote>

The Gemalto driver uses ATRs to recognize if it supports a specific card. However it seems it cannot recognize by itself if a card is a GemSAFE V1 versus V2V3. One has to install a different driver for either of them. The only difference is a minute change to a configuration file specifiying which dynamic library to load (libgck2014x for GemSAFE V1 versus libgck2015x for GemSAFE V2V3).

The Gemalto driver does not send an APDU to my card to select an applet. The function which is responsible for that merely returns "Applet already selected". However beforehand it sends an APDU
`00 CA 9F 7F 2D`
which the card responds to with
`9F 7F 2A 47 90 50 81 12 91 11 02 02 01 41 50 00 00 21 DC 7B 38 10 02 42 53 41 93 42 53 41 95 42 53 00 00 C1 19 00 00 00 00 00 00 00 00 90 00`.
Also, it seems the applet version can be read with
`00 CA DF 30 08`
to which my card responds
`DF 30 05 76 33 2E 30 33 90 00`
(the actual version string is `76 33 2E 30 33`).

I don't know how the two applets work whose AIDs we're currently missing. I also don't know how the card would respond if we tried to select all known AIDs in order to find out which are supported by the card. My fear is that we might regress users.

dengert:
The main issue I would have with any changes to any of the gemsafe drivers is that the changes do not  cause problems with other drivers that AIDs  and may be on a GemAlto cards.
(For example a GelAlto card with the PIV application.) 

ISO 7816-4  Section 8.1.1 Historical  Section 8.22 Application selection help show how to find the default AID and might help you to determined the difference in the applications on the cards.   All the ATRs in question have the Historical section with a 31:80  Section 8.1.1.2.3 Card service data 
says  80 is Application selection  by full DF name  and  EF.DIR and EF.ATR access services by the READ RECORD (S) command (record structure) 

I don't have any of the cards other then the PIV GemAlto card, but you might be able to use the information in 7816-4 to distinguish between versions of the card.

frankmorgner:
Hmm, the ATR recognition table did not get much simpler with the ATR mask applied. Maybe rolling back to your original proposal is the best idea. @dengert would this appease your concerns since we had only verified ATRs in the table?

dengert:
it would appease my concerns as long as driver modules that use AIDs are called before driver modules that only use ATRs.   But using the mask may make things worse, as other drivers or applications may map to the same masked ATR. 

Also the mask is covering up the vendor part of the Historical bytes that are the only difference between some of the ATRs.  

OpenSC will eventually need to stop using ATR matching. Windows gets away with it because only the ATR of cards that are going to be used on the local machine are in the registry. OpenSC's matching of ATRs is done for every driver OpenSC supports. 
AIDs are registered to may them unique, ATRs are not.

l1k:
@frankmorgner: Sure thing, I've rolled back the ATR masks commit.

frankmorgner:
thanks

PR #680, opened 2016-02-19
pkcs11/framework: no 'for-applications' mode of slot creation

viktorTarasov:
Simplify create tokens rules, no need to manipulate applications in
'pkcs11' configuration part,
applications can be enabled/disabled on the 'pkcs15' one.

Fix the possibility to expose only 'sign' PIN


frankmorgner:
Looks OK, but to be honest, I don't really know what the application mode was for in the first place.

PR #692, opened 2016-02-27
travis: Increase the compilation warning level

LudovicRousseau:
- Increase the compilation warning level
- Fail on warning to detect bogus code before any merge

WORK IN PROGRESS: DO NOT MERGE YET

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
 before_script:
   - ./bootstrap
   - if [ -z "$HOST" ]; then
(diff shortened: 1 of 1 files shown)


frankmorgner:
I fully support your mission. However, I won't put more time into it than I already have done https://github.com/OpenSC/OpenSC/pull/222 (without succeeding to get everything merged for whatever reason)...

You should do the same for the windows build by adding `/WX` to the compiler flags.

PR #694, opened 2016-02-28
DNIe. Removing all memory leaks and using SM wrapping and unwrapping.

germanblanco:
See discussion in https://github.com/OpenSC/OpenSC/pull/655
This is a very significant change in the DNIe module in order to align better with the OpenSC SM module.
It also solves memory leaks in the module.


germanblanco:
Covers (at least in part) https://github.com/OpenSC/OpenSC/issues/610

Issue #691, opened 2016-02-26
Compile in version number


hexum:
Please add info about version in executable files and add flag to easily get it.
It's really can help in bug reporting on systems like Gentoo in case of installing from git.
When you install from ebuild-9999 you can't be quite sure about last commit reference number.

Thanks!

(Now I'm adding opensc-9999 to portage tree to make master version testing easy for gentoo users)

hexum:
Gentoo ebuilds adding request.
https://bugs.gentoo.org/show_bug.cgi?id=575770

martinpaljak:
There is `opensc-tool --info`, do you mean something different?

hexum:
People first try to find out version number with options -v, -V or similar. Option --info is strange but OK.
I've just installed version 0.16.0_pre1 and --info shows 0.15.0. What will it show me when I install version from git? To make correct bug report I need to have git ref number...

frankmorgner:
@hexum please make a pull request with your suggestion

hexum:
OK, I will think about t at weekends.
I have little C experience, so it will take some time.

viktorTarasov:
Possibility to get revision number of installed package was missing for a long time.
I propose to return as result of _-v_ the _git describe_ and date of last commit.

Something like this: viktorTarasov@dd823ad07c7c94b37c7fc62a322925c35bacc7b3

@hexum 
> I've just installed version 0.16.0_pre1 and --info shows 0.15.0.

It will be fixed with the first RC of 0.16.0

viktorTarasov:
@hexum 
Sorry, I intercepted this issue with a PR #697.
It implies not only _C_, but also _autoconf_, _git_ and other more local issues.

@frankmorgner 
I propose to transfer discussion onto PR #697.

hexum:
Great!

viktorTarasov:
closed with 1c0ab2a2d213f38ae8fd016159669cd1fc6a134c

PR #697, opened 2016-03-02
tools: print package revision

viktorTarasov:
Follows issue #691

- Initialize the _PACKAGE_VERSION_REVISION_ with commits number from _'git describe'_.
- Introduces the OpenSC package revision printed by tools when tools are used with _--version_ cmd argument. For a while only for _opensc-tools_, but will also be applied to other main opensc-xx and pkcs15x-xx tools.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,7 +9,6 @@ define([PACKAGE_VERSION_MAJOR], [0])
 define([PACKAGE_VERSION_MINOR], [15])
 define([PACKAGE_VERSION_FIX], [0])
 define([PACKAGE_SUFFIX], [])
-define([PACKAGE_VERSION_REVISION], [0])
 
 define([VS_FF_LEGAL_COPYRIGHT], [OpenSC Project])
 define([VS_FF_LEGAL_COMPANY_NAME], [OpenSC Project])
@@ -31,7 +30,6 @@ AM_INIT_AUTOMAKE(foreign 1.10)
 OPENSC_VERSION_MAJOR="PACKAGE_VERSION_MAJOR"
 OPENSC_VERSION_MINOR="PACKAGE_VERSION_MINOR"
 OPENSC_VERSION_FIX="PACKAGE_VERSION_FIX"
-OPENSC_VERSION_REVISION="PACKAGE_VERSION_REVISION"
 
 OPENSC_VS_FF_LEGAL_COPYRIGHT="VS_FF_LEGAL_COPYRIGHT"
 OPENSC_VS_FF_COMPANY_NAME="VS_FF_LEGAL_COMPANY_NAME"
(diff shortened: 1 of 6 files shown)


frankmorgner:
Here is an alternative implementation that does not depend on `./bootstrap` https://github.com/frankmorgner/OpenSC/commit/fbb5c2513b7f3817a2e392aed1fc05072392a2b3. Also, I find the actual hash of the commit more usefull than counting the number of commits after the last tag...

viktorTarasov:
@frankmorgner 
Nice, thanks.
few remarks:
- _OPENSC_VERSION_REVISION_ is used in windows resource file as a part of FILEVERSION. Here short hash will not pass. That's for the "number of commits after release" is used.;
- As a version string I prefer more expanded info, tag, hash and the time of commit.

frankmorgner:
@viktorTarasov agreed.

However, I'd still like to see the environment built in one single place if possible. Simply put your shell scripts into `configure.ac` and your done.

viktorTarasov:
New version of PR, with the scripts in _configure.ac_, instead of _bootstrap_ (thanks to @frankmorgner).

Package version is slightly different for _travis_ and _appveyor_ builds.
The reason in the depth of git clone, limited by 50 in case of _travis_. (_Travis_ can be configured to have larger depth.)

Package version will be the same when a new annotated tags will be created during the coming release.

configure.ac
```diff
@@ -303,8 +301,36 @@ AC_MSG_RESULT([${xslstylesheetsdir}])
...
+		fi
+
+		if test "${REVISION_DESCRIPTION}" = "${NUMBER_OF_COMMITS}"; then
+			NUMBER_OF_COMMITS=0
+		fi
+	else
+		NUMBER_OF_COMMITS=0
+	fi
+
+	COMMIT_DATE="$(${GIT} log -1 --pretty=fuller --date=iso | grep CommitDate: | sed -E 's/^(CommitDate:\s.*)\s+.*/\1/')"
```
frankmorgner: maybe you want to use `git log -1 --pretty=format:%ci`

frankmorgner: or you even unify this with `REVISION_COMMIT` by using `git log -1 --pretty=format:"%h %ci"`

frankmorgner: For the number of commits I found `git rev-list  `git rev-list --tags --no-walk --max-count=1`..HEAD --count`, which worked well for me, see https://stackoverflow.com/questions/8595391/how-to-show-git-commit-using-number-of-commits-since-a-tag

viktorTarasov: Ok, thanks.

By default _'git describe'_ use the annotated tags and
for the current _master_ it composes name relative to _0.15.0_ tag.

For _master_ it's more appropriate then reference to _v0.16.0-pre1_ obtained with _'git rev-list --tags'_. This last tag had to be separated to 0.16.0 dedicated branch. It's not so important and will be fixed by the next release.

frankmorgner: that heuristic does not always work. `git describe 440289a09113d19d83b20cfecd2ba8dd98ba5826` shows `0.14.0rtm-2-g440289a`. Using a regex for the tag should work better.

viktorTarasov: You've took the commit inside the branch dedicated for release preparation, and naturally it took the latest annotated tag. This commit is not in _master_ branch, so, such situation do not concerns builds on _master_.

Once more,  the rc/prep/rtm tags have not be done directly on master branch, but on the branches dedicated to prepare release. In this terms the actual _v0.16.0-pre1_ is misleading,  but it will by fixed after next release.

viktorTarasov:
further refactoring.
More thanks for @frankmorgner.

frankmorgner:
looks good now

PR #702, opened 2016-03-04
pkcs11-tool: fix invalid parameter for 'find-mechanism'

### LINKED ISSUE #700: pkcs11-tool --genkeypair fails with Generate xxx mechanism not supported
Happens if -m is not given.

Introduced with dbca856. See inline comment.

viktorTarasov:
Fixes #700 

Invalid arguments in _find_mechanism()_ call.


frankmorgner:
looks OK. @CardContact do you confirm?

CardContact:
Fine with me.

PR #705, opened 2016-03-06
Issue/671: rtecp: fix delete private key

viktorTarasov:
when deleting private key, delete also it's internal public part


frankmorgner:
this does not compile

PR #707, opened 2016-03-07
ASN1: invalid parse of high-number-tags

viktorTarasov:
Trying to understand _coverity-scan_ issue, I found out that _high-value-tag_ is not correctly parsed.
Apparently it has never been used, and will never be used, 
but for the purity of code this PR is proposed.

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -73,14 +73,17 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
 		*tag_out = SC_ASN1_TAG_EOC;
 		return SC_SUCCESS;
 	}
-	/* parse tag byte(s) */
+	/* parse tag byte(s)
+	 * Resulted tag is presented by integer that has not to be
+	 * confused with the 'tag number' part of ASN.1 tag.
+	 */
 	cla = (*p & SC_ASN1_TAG_CLASS) | (*p & SC_ASN1_TAG_CONSTRUCTED);
 	tag = *p & SC_ASN1_TAG_PRIMITIVE;
 	p++;
 	left--;
 	if (tag == SC_ASN1_TAG_PRIMITIVE) {
 		/* high tag number */
-		size_t n = sizeof(int) - 1;
+		size_t n = SC_ASN1_TAGNUM_SIZE - 1;
 		/* search the last tag octet */
 		while (left-- != 0 && n != 0) {
 			tag <<= 8;
@@ -93,10 +96,8 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
 			/* either an invalid tag or it doesn't fit in
 			 * unsigned int */
 			return SC_ERROR_INVALID_ASN1_OBJECT;
-
 	}
-	if (left == 0)
-		return SC_ERROR_INVALID_ASN1_OBJECT;
+
 	/* parse length byte(s) */
 	len = *p & 0x7f;
 	if (*p++ & 0x80) {
--- a/src/libopensc/asn1.h
+++ b/src/libopensc/asn1.h
@@ -135,6 +135,7 @@ int sc_asn1_sig_value_sequence_to_rs(struct sc_context *ctx,
 #define SC_ASN1_CONS			0x01000000
 
 #define SC_ASN1_TAG_MASK		0x00FFFFFF
+#define SC_ASN1_TAGNUM_SIZE		3
 
 #define SC_ASN1_PRESENT			0x00000001
 #define SC_ASN1_OPTIONAL		0x00000002


viktorTarasov:
Just curious:
_coverity-scan_ issue in question is classified as _intentional_ - _ignore_.
Was it done by a human, or by _coverity-scan_ itself?

LudovicRousseau:
AFAIK the coverity-scan classification is done by a human.
I don't know if it is possible to know by who and when the classification was done.

frankmorgner:
@philipWendland didn't you introduce the extended ASN.1 parsing specifically to support high number tags?

viktorTarasov:
Looking on 7e7a44acff9d3aec082076b556a9363081df862a of @philipWendland:
- code of @philipWendland concerns encoding of extended tag;
- proposed PR breaks code of @philipWendland;
- code of @philipWendland expects the tag in it's raw, non-parsed form -- integer obtained by concatenation of the tags bytes.
- we can:
  - decide to consider this form as an internal OpenSC tag representation and leave all as it now;
  - adopt parsed form to use natural and expected tag value and change code of @philipWendland.

philipWendland:
I don't have access to the coverity issues, so I don't really know what you are talking about :-)

AFAIR, the ISO 7816 did specify the tags in raw (e.g. `0x7FFF`) format, so I did not really think about "natural encoding" back then.

I'm happy to adopt the IsoApplet code after we decided for a consistent format. I am not really biased towards any format either. I don't know whether there is another card driver that expects the current behavior.

philipWendland:
@viktorTarasov I think it is used in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-isoApplet.c#L634, the card returns tag 0x7F49 ("Interindustry template for nesting one set of public key data objects with the following tags", ISO-7818 part 8).

frankmorgner:
I don't have a strong opinion about the implementation, but @dengert showed some love to ASN.1 in the past :wink:

From the comments, I can see that this could need rewriting some other code. If so, this change should be postponed for now until 0.16.0 is released.

viktorTarasov:
@philipWendland 
As you I worked with such double-byte tags in card drivers without doubts and questions.
But one can insiste that in terms of ASN1 the 0x7F49  is in fact encoded _constructed application tag 73 (x49)_. 

@frankmorgner 
_sc_asn1_read_tag()_ is used by several card drivers and tools. Who knows, in a _future_, they could expect the canonical ASN.1 tag value .
Change to code is not complicate, and not-yet-started release is probably occasion to test it.

Personally, I do not insist.

viktorTarasov:
rebased on _master_

viktorTarasov:
After reflections:
- this PR do not breaks _put_tag_ of @philipWendland  -- with actual _sc_asn1_read_tag()_ value 0x7F49 for tag cannot be obtained, so common asn1 parse procedures seems to be not used.
- it's OK to use the internal form of ASN.1 tag inside the common ASN1 procedures, but it has not to be used outside.
- currently _sc_asn1_read_tag()_ is used by card drivers and tools and I think that it **should** return the expected, specified result.

dengert:
I will have a look at this later this week.
grep 7F49 
shows 5 files that look for this tag.

dengert:
Your patch looks correct to return the tag value as an integer. But there are a lot of code needs to be  f changed if the return value changes 

./libopensc/card-isoApplet.c pusts the tag back together with the class:
if((tag != 0x1F49) || (cla != 0x60))

This is testing the tag as returned by the current code. The new code would return 0x49

./libopensc/asn1.cm calls sc_asn1_read_tag in three places,  print_tags_recursive looks like it does not print a long tag correctly.  sc_ans1_find_tag also expects the tag to be in the current format.
sc_asn1_skip_tag does not use the tag, but just skips it. 

./libopensc/pkcs15-sc-hsm.c only looks for tag = 7. So its OK

./libopensc/card-openpgp.c tries to put the tag back together with the class.

./libopensc/pkcs15-pubkey.c does not use the tag_out. 

sc_asn1_find_tag takes a tag with the class, and tries to find it. Any use of sc_asn1_find_tag needs to be looked at too. 

./libopensc/card-piv.c calls it 8 places. I need to look closer at it if we do this change.  

./libopensc/card-mcrd.c tries to skip tag. It may be OK. 

We may want to leave it as is, and call it internal format.

viktorTarasov:
I've spent too much time on this PR, so:
- the list of changes seems not to be exorbitant;
- there is ambiguity in the name of functions _sc_asn1_read_tag()_ and _sc_asn1_put_tag()_ that seems to be complementary, but in fact they use different type of data
  - so, I propose to change the name of second for   _sc_asn1_put_encoded_tag();
- as I've told, it could be used as internal format inside the common _asn1_, but preferably is not to be used outside, especially in the sake of _"next generation of developers"_; 
- I will prepare new version of PR.

viktorTarasov:
Well, finally I have to agree with @dengert  and  withdraw the initial proposal -- in code the using of internal _tag number_ format is too wide. _asn1_find_ makes almost direct comparaison of encoded tag values (like 0x7F53) and tag number obtained by parse (with _sc_asn1_read_tag()_).

Updated PR contains limitation of tag number size in internal format and fix of coverity-scan issue.

PR #709, opened 2016-03-10
pkcs15: use file cache for path with AID

viktorTarasov:
When activated by configuration, the cache of the on-card files is applied only to paths of type _PATH_ -- full path starting from MF with or without _3F00_.
Proposed PR make it also possible for the application paths (path with AID) of type _FILE_ID_.

Initial motivation of PR was attempt to resolve the _TOCTOU_ coverity-scan issue.

### DIFF
--- a/src/libopensc/pkcs15-cache.c
+++ b/src/libopensc/pkcs15-cache.c
@@ -41,38 +41,53 @@ static int generate_cache_filename(struct sc_pkcs15_card *p15card,
 				   char *buf, size_t bufsize)
 {
 	char dir[PATH_MAX];
-        char pathname[SC_MAX_PATH_SIZE*2+1];
+	char *last_update = NULL;
 	int  r;
-        const u8 *pathptr;
-        size_t i, pathlen;
+	unsigned u;
 
-	if (path->type != SC_PATH_TYPE_PATH)
-                return SC_ERROR_INVALID_ARGUMENTS;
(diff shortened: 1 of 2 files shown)


dengert:
If the AID file is cached, will a select of the AID still be sent to the card? Selecting the AID can change the state of the card.

viktorTarasov:
Well seen, I missed it.
Certainly _select file_ has to be done for the path with _AID_,
But it's also true for the path without _AID_, isn't it? Here _select file_ has also be kept?

dengert:
Just want to make sure that the select of the AID is not optimized out by any caching. 
ISO 7816-4 Section 8.2.2 talks about how the Application is selected. 
Its not clear how to distinguish at a low level if a select file APDU is for a common file or the the AID to force card to switch applications.  We could have flag, or separate select AID routine.

viktorTarasov:
The path _DF_NAME_ is the application path, with a dedicated INS value in _select_file_ apdu.
Two  _select_ _APDUs_ is needed to select path of type FILE_ID with aid.

Yes, we could distinguish paths with AID,
but even for path without AID it seems to me reasonable to select the parent of cached file.

Well, for a while the parent selected only for the cached path with AID.

viktorTarasov:
PR updated:
- move 'select parent' to caller of _read-cached-file_;
- extract fix of 'last-update' read length into dedicated commit.

PR #714, opened 2016-03-20
Compile from a separate build directory

dengert:
doc/toots/Makefile.am is missing a $(srcdir)

 Changes to be committed:
	modified:   doc/tools/Makefile.am

### DIFF
--- a/doc/tools/Makefile.am
+++ b/doc/tools/Makefile.am
@@ -28,7 +28,7 @@ tools.html: $(srcdir)/tools.xml $(wildcard $(srcdir)/*.1.xml) $(wildcard $(srcdi
 
 %: $(srcdir)/%.1.xml
 	@echo $< $@
-	@cat completion-template \
+	@cat $(srcdir)/completion-template \
 		| sed "s,ALLOPTS,\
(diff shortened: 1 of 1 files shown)


frankmorgner:
Actually, `make dist-check` compiles with different source and build directories. However, it always uses the source directory as current directory...

Anyway, your change doesn't hurt.

dengert:
I have always use the build directory as the current directory, using
../OpenSC/configure ... parameters
make
make install

So the build directories are outside the source directory.
Since I don't have any true pkcs15 cards, the lack of the templates
was never an issue so I never paid attention to them not being built.

  https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Build-Directories.html

  "Make command lines should always refer to source files by prefixing them with ‘$(srcdir)/’."


On 3/24/2016 4:50 AM, Frank Morgner wrote:
> Actually, |make dist-check| compiles with different source and build directories. However, it always uses the source directory as current directory...
>
> Anyway, your change doesn't hurt.

>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/714#issuecomment-200761209>
>

-- 

PR #717, opened 2016-03-21
Prep openssl 1.1 v2 

dengert:
This is a replacement for #654  but based on v0.16.0-rc. The bulk of the commits have been sguashed into 1 large commit. The first (small one) could be committed today, the second could wait till OpenSSL-1.1.0 is released. 

I believe I have addressed all the comments in #654 

I am looking for additional comments, ans there are a lot of changes.

### DIFF
--- a/src/libopensc/card-entersafe.c
+++ b/src/libopensc/card-entersafe.c
@@ -194,7 +194,7 @@ static int entersafe_cipher_apdu(sc_card_t *card, sc_apdu_t *apdu,
 								 u8 *key, size_t keylen,
 								 u8 *buff, size_t buffsize)
 {
-	 EVP_CIPHER_CTX ctx;
+	 EVP_CIPHER_CTX * ctx = NULL;
 	 u8 iv[8]={0};
 	 int len;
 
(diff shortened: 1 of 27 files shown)


viktorTarasov:
> 8f28dcd 	Handle C and C++ "inline" and "restrict" keywords for various compilers

I still do not understand, what is the benefit from using _restrict_ and _inline_ modifiers in software to manage smart cards.

I propose to eliminate all occurrences of these modifiers, including _simclist_.
"Simclist -- advanced techniques for improving performance of lists handling". What for? To handle few short lists: slots, sessions, readers, objects ?

dengert:
As I said in https://github.com/OpenSC/OpenSC/commit/8f28dcd7b71754675e2ecbe6bc5a5548d2ef2849 
: "(Another option is to just remove all inline and restrict keywords from OpenSC code.)"

There is not much to be gained in performance.
I would still keep simclist just without the inline and restrict keywords and without the tests to set them.

0.16.0 should at least address the issue. Either by doing a better job of handling the keywords, or
not use the keywords at all.

If we chose to eliminate the keywords, we may in the future run it to the issue if new code is added
that uses them. That is why I added https://github.com/OpenSC/OpenSC/commit/8f28dcd7b71754675e2ecbe6bc5a5548d2ef2849

viktorTarasov:
@dengert sorry, have not seen.

8f28dcd applied to _0.16.0_.

dengert:
Now that OpenSSL-1.1.0-pre5 is released, I am going to close this PR, and open a new PR with these mods and additional mode that are required by OpenSSL-1.1.0-pre5

dengert:
Replaced by #861. OpenSSL-1.1.0 has been released.

PR #718, opened 2016-03-22
pkcs11-tool.1.xml: Add documentation for missing options in  pkcs11-tool

Jakuje:
I started working with `pkcs11-tool` recently and found out that most of the switches are not documented in the manual page.

I identified missing switches with simple script (at some point to some scale, it might be nice to have something similar in test cases):

    https://github.com/Jakuje/stuff/blob/master/opensc_manual.sh

and created simple "almost-copy-paste" of missing options from usage. I found commit trying to make the options sorted by name, but it was not the case because of other commits since that time. Let me know if the prefered way would be sorted alphabetically and I will rebase the patch.

Also in one option i found that the option is missing tag, so I added appropriate one.

### DIFF
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -56,6 +56,15 @@
 
 				<varlistentry>
 					<term>
+						<option>--unlock-pin</option>
+					</term>
+					<listitem><para>Unlock User PIN (without <option>--login</option>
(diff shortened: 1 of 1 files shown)


Jakuje:
Yes, and there might be other manual pages missing options. I didn't check yet.

viktorTarasov:
applied in 972a93a of _opensc-0.16.0_ branch,
thanks.

Issue #720, opened 2016-03-23
Multiple ePass2003 error


velter:
### Expected behaviour

Use serveral ePass2003 on the same machine sould be possible
### Actual behaviour

When two ePass2003 are plugged in a machine, only the second slot is usable. As soon as you try to login in the first slot a C_Login failed with error CKR_DATA_INVALID is raised.
### Steps to reproduce
1. Plug two ePass2003
2. pkcs15-init --reader 0 -E
3. pkcs15-init --reader 1 -E
4. pkcs15-init --reader 0 -C --use-default-transport-keys --profile pkcs15+onepin --pin 123456 --puk 123456
5. pkcs15-init --reader 1 -C --use-default-transport-keys --profile pkcs15+onepin --pin 123456 --puk 123456
6. pkcs11-tool --slot 5 -t -l -p 123456 return OK
7. pkcs11-tool --slot 1 -t -l -p 123456 return the following error :

"error: PKCS11 function C_Login failed: rv = CKR_DATA_INVALID (0x20)"
### Logs

[opensc-debug.txt](https://github.com/OpenSC/OpenSC/files/186022/opensc-debug.txt)

I tried to analyze this. What happen is while
1. OpenSc init first token
2. OpenSc init second token
3. OpenSc try to login to second token

Return OK
1. OpenSc init first token
2. OpenSc init second token
3. OpenSc try to login to first token

Fail

This seem to show that the second token init mess up something for the first one.

Looking at card-epass2003.c there is a lot of static variable (including encryption key) for example on line 97

static unsigned char g_sk_enc[16] = { 0 };  /\* encrypt session key _/
static unsigned char g_sk_mac[16] = { 0 };  /_ mac session key _/
static unsigned char g_icv_mac[16] = { 0 }; /_ instruction counter vector(for sm) */

which should be card dependant, not global. any ideas what's the best way to fix this ?

Thanks for your help

FeitianSmartcardReader:
Hi Velter

This is Ben, from FEITIAN tech team.
We will take a look this issue, should back to you tomorrow, thanks

velter:
Hi Ben,

Nice thanks.

it looks like card-epass2003.c should use card->sm_ctx.info.session.gp_keyset.enc/mac/kek ...instead of static global vars.

I found another sm related issue. This time with a single token but with two shells. Steps to reproduce are (with an already initialized token) :

Shell 1 : pkcs11-tool -t -l (Do NOT enter pin at this time)
[...]

PR #721, opened 2016-03-25
Proposed fix for #720

velter:
This fix consist of two part :
* Allow use of serveral epass2003 pluged in simultaneously. Replace static globals with card->sm_ctx.info.session members. As far as I understand the code, this structure is only used by smm-local which is not used by this card driver. This change have been tested and pkcs11-tool -t -l work as expected now.
* Allow use from several process. I added an epass2003_refresh(card) in epass2003_pin_cmd. This fix the #720 exposed testcase but I thinks that it's not the most efficient way to solve this problem. It would probably be better to do the refresh only on failure


frankmorgner:
@FeitianSmartcardReader could you have a look?

PR #725, opened 2016-03-30
A bug fix to the MyEID driver.

hhonkanen:
I found an old bug while testing the MyEID driver. card->name pointed to a buffer that fell out of scope after returning from myeid_get_info, resulting in garbage displayed when running for example opensc-tool -n. Fixed this by declaring the buffer in file level. 

Please could somebody review this and merge if ok?


viktorTarasov:
slightly modified and applied in fe2312dd1966f9fac6751345ab894b89b4f7df3c

hhonkanen:
Thank you Viktor!

PR #728, opened 2016-04-08
Auxiliary, non-pkcs15, data in pkcs15 data types

viktorTarasov:
Sorry for the long list of commits, but they forms relatively dependent chain and it's too laborious to split them onto multiple PRs.

At the origin there was RC of OpenSC-0.16.0 that failed with test of X509 Enrollment with minidriver for IAS/ECC card -- part of the non-pkcs15 data was removed from pkcs15 data types during the recent source history.

OpenSC driver for IAS/ECC is supposed to create and maintain the on-card file system completely compatible with the native MW (Gemalto, Oberthur, ...), including file-system part to support MD.
From here there is a more general need: possibility to transfer some information from the non-pkcs15 applications (minidriver) until the card specific level through the common framework.

So that, following the commit list:
- introduced 'auxiliary-data.[c,h]' to hold the non-pkcs15 data types (like minidriver), keep its types and related procedures, and finally expose general auxiliary type that contain the union of the non-pkcs15 data to be inlcuded into pkcs15 data types as a _auxiliary data pointer_. For a while it's going about _sc_pkcs15_prkey_info_;
- parameter _aid_ is added to _pkcs15-synthetic-bind_ procedures, in the same manner as it was done for the _internal-bind_: this will allows to PKCS#15 emulation to manage more then one on-card application, like _internal-bind_ do; 
- iasecc card is binded by pkcs#15 emulator -- it uses all procedures of internal pkcs15 bind, the only supplement is when parsing PrKEY DF it tries to get MD related info and to initialize the _auxiliary-data_ (MD) for the private key.
- IAS/ECC specific code is removed from MD and pkcs15init sources and replaces by _general_ procedures to  manage auxiliary data.
- pkcs15init manages auxiliary data (for private keys)
- do not use '{}' frame for GUID -- GUID generated on WIndows can be on the limit of size reserved in MD CMAP record.

### DIFF
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -11,7 +11,7 @@ noinst_HEADERS = cards.h ctbcs.h internal.h esteid.h muscle.h muscle-filesystem.
 	cardctl.h asn1.h log.h \
 	errors.h types.h compression.h itacns.h iso7816.h \
 	authentic.h iasecc.h iasecc-sdo.h sm.h card-sc-hsm.h \
-	pace.h cwa14890.h cwa-dnie.h card-gids.h
+	pace.h cwa14890.h cwa-dnie.h card-gids.h aux-data.h
 
 AM_CPPFLAGS = -DOPENSC_CONF_PATH=\"$(sysconfdir)/opensc.conf\" \
 	-I$(top_srcdir)/src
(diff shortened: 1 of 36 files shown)


frankmorgner:
looks OK on the first glance.

Why are you using the term auxiliary data if all it does is dealing with md data? Do you think this should be used for other use cases, too? If so, would you be needing new call backs for new applications? If so, wouldn't it be more useful to use a mechanism similar to the sc_cardctl stuff that is more flexible in adding more functionality?

viktorTarasov:
I see _aux-data.[c,h]_ as a place holder for all non-pkcs15 method and data types that can be needed some card specific or by some application (that's why it's not in sc-cardctl).

frankmorgner:
Same here, if you want comments and suggestions please leave this open for a longer time and don't merge it just a couple of hours after the first comment came in.

Just for the record, I suggested an implementation *similar* to `sc_card_ctx`, which could have looked something like `int sc_aux_data(int ctrl, u8 *in, size_t inlen, u8 *out, size_t *outlen)` and `#define SC_AUX_DATA_CTRL_SET_MD_GUID 1`. This way new functionality would have gotten a control code instead of a new function call.

PR #730, opened 2016-04-12
Solved multiple ePass2003 error #720

FeitianSmartcardReader:
Fixed bug when using multiple ePass2003, more information, please refer
to https://github.com/OpenSC/OpenSC/issues/720

The problem is the old code only can detect inserted key. the problem is when application running, will init all public data with all inserted key, but all the data is stored in last inserted key. So if at this time, do operation with first inserted key, it will have problem. user will get rv = CKR_DATA_INVALID (0x20) error.

Now, before verify PIN, we do re-init special key public data.

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -22,6 +22,7 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+
 #ifdef ENABLE_SM		/* empty file without SM enabled */
 #ifdef ENABLE_OPENSSL		/* empty file without openssl */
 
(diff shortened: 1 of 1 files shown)


frankmorgner:
This does not compile cleanly. I suggest you have a look at @velter's approach

frankmorgner:
@FeitianSmartcardReader did you check?

FeitianSmartcardReader:
Sorry for late reply, was busy for other projects. We do check Velter codes and approval these changes(https://github.com/OpenSC/OpenSC/pull/731).

Do I need submit another pull request with all changes? Thanks,

frankmorgner:
closed with https://github.com/OpenSC/OpenSC/commit/73b5d8421abedf2fafa54720b9624e18ef2630fd applied

please open an other PR if some changes were missing

PR #731, opened 2016-04-12
Fix for #720

velter:
This is a cleaner fix for #720 which take part of #721 and #730
A driver private struct is used which allow to remove all globals which might changed base on the token in use.


frankmorgner:
@FeitianSmartcardReader This is your card, right? Could you have a look, please?

viktorTarasov:
applied in 73b5d84 of _opensc-0.16.0_ branch,
thanks.

PR #732, opened 2016-04-13
Move PIN type label front of description

metsma:
With very long labels it is unable identify which PIN Firefox is asking.
PKCS11 label is 32 char's and it gets truncated.

Current format "LABEL (PIN1)"
New format "PIN1: LABEL"

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -972,7 +972,7 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 		}
 		else   {
 			if (auth->label[0])
-				snprintf(label, sizeof(label), "%s (%.*s)", p15card->tokeninfo->label, (int) sizeof auth->label, auth->label);
+				snprintf(label, sizeof(label), "%.*s: %s", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
 			else
 				snprintf(label, sizeof(label), "%s", p15card->tokeninfo->label);
 			slot->token_info.flags |= CKF_LOGIN_REQUIRED;


frankmorgner:
Your card issuer may want to think about a more useful name for the PIN instead...

The `auth->label` may be up to 255 bytes, is it always that short in case of a PIN?

martinpaljak:
The problem comes from not-so-great-mapping between PKCS#15 and PKCS#11 (tokens vs slots). The labels are exact and useful, but the way the presenting is constructed in OpenSC lends to bad interpretation. The fields in PKCS#11 are length-capped, which means that for sufficiently meaningful token labels (a long name in case of Estonian eID), the PKCS#11 label will not have the PIN label visible, so you won't know which PIN from the token is asked. Which makes Firefox ask for "something" you don't know what it is. So the slot label is usually only meaningful before login, the actual objects in the slot (like certificates) are usually displayed natively (common name from subject etc)

frankmorgner:
Could you give a screenshot before and after the change?

metsma:
Before

    /Library/OpenSC/bin/pkcs11-tool --module=/Library/OpenSC/lib/opensc-pkcs11.so  -L
    Available slots:
    Slot 1 (0x1): ACS ACR 38U-CCID
      token label        : MÄNNIK,MARI-LIIS,47101010033 (P

After

    /Library/OpenSC/bin/pkcs11-tool --module=/Library/OpenSC/lib/opensc-pkcs11.so  -L
    Available slots:
    Slot 1 (0x1): ACS ACR 38U-CCID
      token label        : PIN1: MÄNNIK,MARI-LIIS,47101010

frankmorgner:
Ah, OK. Could you change it to "PIN1 (LABEL)"?

PR #733, opened 2016-04-14
Define MD log level with key from windows registers

viktorTarasov:


frankmorgner:
- Using the environment should be explicitly activated by the user if he wants to. The environment may be hostile... In the commandline tools we can already use something like `--pin=env:VAR`
- I fear that having multiple methods of configuring opensc will confuse users. You should either make *all* config options available in the registry with the exact meaning of opensc.conf or leave everything in one place (i.e. opensc.conf). Also, as far as I can see, options can only be documented in opensc.conf, but not in the registry.

viktorTarasov:
@frankmorgner 
You have a reason while reasoning with the common considerations, and, certainly it has to be done when someone will have more time and possibilities.

But in this case it's going about the quite handicapping side of MD logs -- they are always activated, with the permanently growing MD debug file, even without card activity.
The MD debug option cannot be put into opensc.conf file, because MD logs starts without card, without sc context, ... ; I would not like reduce the MD logs and limit them only by the card activity.
That's for the dedicated registry is introduced.

As for multiple source of configuration options -- it always been like this, and, IMHO, even if confusing will have place, it will last long.

viktorTarasov:
applied in _opensc-0.16.0_ branch

frankmorgner:
If you want comments and suggestions please leave this open for a longer time and don't merge it just a couple of hours after the first comment came in.

viktorTarasov:
Sorry,
- this PR was here 4 days;
- this feature is not too intrusive or complicated;
- release branch is not for a long implementation and discussion, but to resolve and fix;
- Sunday evening was programmed to fix and close and to tag rc2.  

You still can comment and suggest, result could be applied afterwards.

PR #734, opened 2016-04-16
sc-hsm: Fix public key usage for keys extracted from CVC

CardContact:
The SmartCard-HSM PKCS#15 emulation does not set the public key usage flags for public key extracted from Card-Verifiable-Certificates.


viktorTarasov:
applied in 41f34fd of _opensc-0.16.0_ branch,
thanks.

Issue #735, opened 2016-04-17
CardOS init broken for readers without extended APDU supported


viktorTarasov:
After 2e21163273978475b31f664f80ad1ad3c02755c7
CardOS card cannot be initialized
when using card reader without Extended APDU supported.

I guess that a new OpenSC configuration option to control using of extended APDU has to be introduced. Disabled by default.

frankmorgner:
There is already an option to limit the transceive length of any reader. You should use it. Also, if the reader announces short/extended length support this *is* honored.

Though, it could be discussed, whether the reader's limit should assumed to be at short length if it is *completely silent* about its limits.

viktorTarasov:
> .. is already an option to limit the transceive length of any reader

this option exist only in comments of opensc.conf -- not supported by code

>  if the reader announces short/extended length support this is honored

yes, it's true, if the reader clearly declares its capacity or if a reader specific config section will be introduced 

> it could be discussed, whether the reader's limit should assumed to be at short length if it is completely silent about its limits.

it hardly could be discussed -- if reader is silent, then short length has to be assumed.

So, to resume,
- by default extended APDUs are disabled.
- it can be enabled by reader, that clearly declares it's capacities, or by reader specific section -- all that joined by card capacity.

frankmorgner:
set `max_send_size` and `max_recv_size`
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/ctx.c#L303-319

frankmorgner:
There may be a problem here https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L877, though

viktorTarasov:
> set max_send_size and max_recv_size

fine, thanks.

So,
- _max_send_size_ and _max_send_size_ has to be enabled by default;
- parse of _reader_driver_ config section moved from _ctx_ to _detect_readers_ handler of reader's _ops_, and placed before _detect_reader_features_;
- check if common card framework overwrites, if needed,  the card specific _max_XX_sizes_

viktorTarasov:
fixed in _release_ branch

PR #739, opened 2016-04-19
reader: by default only short APDU supported

viktorTarasov:
default values of reader's max send/receive sizes correspond to only short APDU supported;
these values can be overwritten by reader itself with the proper value of dwMaxAPDUDataSize from TLV properties,
or with the corresponding options in OpenSC configuration.

resolves issue #735

### DIFF
--- a/src/libopensc/ctx.c
+++ b/src/libopensc/ctx.c
@@ -356,34 +356,6 @@ load_parameters(sc_context_t *ctx, scconf_block *block, struct _sc_ctx_options *
 	return err;
 }
 
-static void load_reader_driver_options(sc_context_t *ctx)
-{
-	struct sc_reader_driver *driver = ctx->reader_driver;
-	scconf_block *conf_block = NULL;
-	sc_reader_t *reader;
-	int max_send_size;
-	int max_recv_size;
-
-	conf_block = sc_get_conf_block(ctx, "reader_driver", driver->short_name, 1);
-
-	if (conf_block != NULL) {
-		max_send_size = scconf_get_int(conf_block, "max_send_size", -1);
-		max_recv_size = scconf_get_int(conf_block, "max_recv_size", -1);
-		if (max_send_size >= 0 || max_recv_size >= 0) {
-			if (list_iterator_start(&ctx->readers)) {
-				reader = list_iterator_next(&ctx->readers);
-				while (reader) {
-					if (max_send_size >= 0)
-						reader->max_send_size = max_send_size;
-					if (max_recv_size >= 0)
-						reader->max_recv_size = max_recv_size;
-					reader = list_iterator_next(&ctx->readers);
-				}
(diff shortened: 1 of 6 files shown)


src/libopensc/reader-pcsc.c
```diff
@@ -870,31 +875,32 @@ part10_find_property_by_tag(unsigned char buffer[], int length,
...
 			goto err;
 		}
 
 		r = part10_find_property_by_tag(rbuf, rcount,
 				PCSCv2_PART10_PROPERTY_dwMaxAPDUDataSize);
-		if (r >= 0)
+		sc_log(reader->ctx, "get dwMaxAPDUDataSize property returned %i", r);
+
+		/* 256 < X <= 0x10000: short and extended APDU of up to X bytes of data */
+		if (r > 0x100 && r <= 0x10000)
```
frankmorgner: There are readers that only support up to 251 bytes of data, which would not be recognized here. This should check for `r > 0` instead of `r > 0x100`.

src/libopensc/ctx.c
```diff
@@ -356,34 +356,6 @@ load_parameters(sc_context_t *ctx, scconf_block *block, struct _sc_ctx_options *
 	return err;
 }
 
-static void load_reader_driver_options(sc_context_t *ctx)
-{
-	struct sc_reader_driver *driver = ctx->reader_driver;
```
frankmorgner: You have now broken the possibility of limiting the data to any of the other reader drivers! This should not be moved, I think.

src/libopensc/reader-pcsc.c
```diff
@@ -2318,6 +2343,9 @@ int cardmod_use_reader(sc_context_t *ctx, void * pcsc_context_handle, void * pcs
 		}
 		priv->gpriv = gpriv;
 
+		reader->max_send_size = PCSC_SHORT_APDU_MAX_SEND_SIZE;
```
frankmorgner: The value from `opensc.conf` or from the reader's tlv is not read on Windows? It should be done!

src/libopensc/reader-pcsc.c
viktorTarasov: http://ludovicrousseau.blogspot.fr/2011/11/extended-apdu-support-reported-by-pcsc.html

> There are readers that only support up to 251 bytes

I've seen this with cards, but not with readers. Even if so, for the most exotic cases there is opensc.conf.

src/libopensc/reader-pcsc.c
viktorTarasov: Agree, it should be done.

viktorTarasov:
Updated version: all reader drivers try to use the max send/recv options from config file.

frankmorgner:
looks OK

viktorTarasov:
applied in e95b5151631f71fcee60a24d8d9d79f1c9719d8c

PR #740, opened 2016-04-20
Prep openssl 1.1.0 pre5

dengert:
OpenSSL-1.1.0-pre5 has been released. This PR has additional changes required by OpenSSL-1.1.0-pre5 and include previous changes from #717. Pre5 now hides the RSA and DSA structures which require additional changes to OpenSC. 

These changes are experimental and need testing, testing that I do not have the hardware to test. 
 
OpenSSL-1.1.0 may have another RC before a final release which may require additional changes. 

As with #717 the changes are designed to allow OpenSC code to compile with OpenSSL versions 0.9.7,  0.9.7, 0.9.8, 1.0.1, 1.0.2 and 1.1.0-pre5

### DIFF
--- a/src/libopensc/card-entersafe.c
+++ b/src/libopensc/card-entersafe.c
@@ -194,7 +194,7 @@ static int entersafe_cipher_apdu(sc_card_t *card, sc_apdu_t *apdu,
 								 u8 *key, size_t keylen,
 								 u8 *buff, size_t buffsize)
 {
-	 EVP_CIPHER_CTX ctx;
+	 EVP_CIPHER_CTX * ctx = NULL;
 	 u8 iv[8]={0};
 	 int len;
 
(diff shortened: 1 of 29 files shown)


frankmorgner:
Is #717 obsolete then?

frankmorgner:
These changes are huge and they are still getting bigger... Looking at https://www.openssl.org/policies/releasestrat.html, I don't see a switch from OpenSSL 1.0.2 to a different version in OpenSC before 2019.

The only thing I could think of is creating a dedicated branch within OpenSC that is compatible with OpenSSL 1.1.0. However, this will be have to synced with master for years while hardly anybody seems to be interested...

dengert:
Yes, The changes for OpenSSL-1.1.0 in commit:
"OpenSSL multiple versions OpenSSL-0.9.7 to 1.1.0 for OpenSC-0.16.0-rc1"
https://github.com/OpenSC/OpenSC/pull/717/commits/60e0f11d09c0bef9a78f23ba0084c04155e176b5
where rebased with one minor fix on to opensc-0.16.0 as:
https://github.com/OpenSC/OpenSC/pull/740/commits/b3a01f942140946ed58406d70e0d069a7748ee2d
and is one of two commitsin #740

The other commit in #740 is "OpenSSL-1.1.0-pre5 mods for OpenSC-0.16.0"
https://github.com/OpenSC/OpenSC/pull/740/commits/0a75ab20a5206272809a22c0ad72a57b1849ef09
It has the additional changes needed because of changes in OpenSSL-1.1.0-pre5.

OpenSSL-1.1.0-pre5 is a really big change for everyone, because OpenSSL has hidden the RSA and DSA structures!

The approach I have taken is to get OpenSC code to work using mostly the 1.1.0 API, to avoid
so many \#if OPENSC_VERSION_NUMBER ... statements.
Please have a look at libopensc/sc-ossl-compat.h It has defines and inline functions to be used with
older versions of OpenSSL. The inline functions where basically copied from OpenSSL-1.1.0. If this is a problem,
we could rewrite them. The point being that the inline functions are only meant to be use for backward
compatibility and will eventually be dropped from the OpenSC code.

Based on how this roll out of 1.1.0 is gone, I expect OpenSSL to make more changes to the API,
and not just bug fixes.

#740 will need a lot of testing by others, as I can not test most of it. So any testing anyone can do would be nice,
especially for card specific tools like sc-hsm-tool and cryptoflex-tool that have a lot of untested changes.


On 4/22/2016 5:08 AM, Frank Morgner wrote:
> Is #717 <https://github.com/OpenSC/OpenSC/pull/717> obsolete then?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/740#issuecomment-213366281>
>

-- 

dengert:
In regard to your comment on "These changes are huge...", yes they are. 

But the approach is  start coding to the 1.1.0 API now (or when 1.1.0 is finally released) 
libopensc/sc-ossl-compat.h contains the macros, defines and inline functions need to use the
1.1.0 API with older versions of OpenSSL. The APIs for the older versions don't change.

dengert:
OpenSC will have to make the change, sometime, putting it off till the future will only make it harder.

In effect I have done the major code conversion, but keeping two branches of OpenSC, will be more
then doing it with a common code base.

dengert:
One more thing. I would *NOT* not merge this to 0.16.0. Its not tested and  Openssl-1.1.0 is not released.

The PR is against the opensc-0.16.0 branch because the many changes in 0.16.0 that are not yet in master. So the only way to test with AppVeyor  and Travis-ci is to base the PR on OpenSC-0.16.0.

It could be merged to master after OPenSC-0.16.0 is merged. Note it has passed all the checks.
This would get developers getting used to using the 1.1.0 API even with OpenSSL-1.0.1 or OpenSSL-1.0.2. Once OpenSSL-1.1.0 is released, AppVeyor and Travis-ci could test against OpenSSL-1.1.0.
This would catch developers using  older APIs while still wokring on the code. 

(P.S. Don't expect any response today, I will be playing golf.)

frankmorgner:
Agreed, if we are still backwards compatible with OpenSSL pre 1.1.0 *and* there is not currently an OpenSC release pending *and* OpenSSL 1.1.0 *is* stable, we should switch.

PR #742, opened 2016-04-22
Make sure that data for ECDSA signature is truncated to the size of the key

Jakuje:
Based on the paragraph from *PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures*:

	If the length of the hash value is larger than the bit length of n, only
	the leftmost bits of the hash up to the length of n will be used. Any
	truncation is done by the token.

This is PKCS#11 specification, so all the tokens behind this interface should reflect this condition. I don't have around different cards to test so this is implemented for in the `piv-card.c`, but it might also be case for different cards. General sign&verify tests against my PIV Test cards succeeded.

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2280,6 +2280,15 @@ static int piv_validate_general_authentication(sc_card_t *card,
 
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
+	/* Make sure that data for ECDSA signature is truncated to the size of the key:
+	 *   If the length of the hash value is larger than the bit length of n, only
+	 *   the leftmost bits of the hash up to the length of n will be used. Any
+	 *   truncation is done by the token.
+	 *     -- PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures */
+	if (priv->algorithm == SC_ALGORITHM_EC) {
+		datalen = MIN(datalen, priv->key_size/8);
+	}
+
 	/* should assume large send data */
 	p = sbuf;
 	put_tag_and_len(0x7c, (2 + put_tag_and_len(0, datalen, NULL)) , &p);


Jakuje:
Travis reports:

    card-piv.c:2289:13: error: comparison of integers of different signs: 'size_t'
      (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
                datalen = MIN(datalen, priv->key_size/8);
                          ^   ~~~~~~~  ~~~~~~~~~~~~~~~~
    ./internal.h:56:25: note: expanded from macro 'MIN'
    #define MIN(x, y) (((x) < (y)) ? (x) : (y))
                        ^
    1 error generated.

I am still not sure if we want to fix it this way. But if yes, I will add explicit type cast.

dengert:
This is not the place (i.e. a specific card driver) to do the truncation. There are 4 card drivers
that can do ECDSA: card-isoApplet.c, card-myeid.c, card-myeid.c and card-piv.c.

The truncation could be done in ./pkcs15-sec.c:sc_pkcs15_compute_signature
after line:
331                 case SC_PKCS15_TYPE_PRKEY_EC:
332                         modlen = ((prkey->field_length +7) / 8) * 2;  /* 2*nLen */

and change inlen.


On 4/22/2016 9:21 AM, Jakuje wrote:
> Travis reports:
>
> |card-piv.c:2289:13: error: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare] datalen = MIN(datalen, priv->key_size/8); ^ ~~~~~~~
> ~~~~~~~~~~~~~~~~ ./internal.h:56:25: note: expanded from macro 'MIN' #define MIN(x, y) (((x) < (y)) ? (x) : (y)) ^ 1 error generated. |
>
> I am still not sure if we want to fix it this way. But if yes, I will add explicit type cast.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub <https://github.com/OpenSC/OpenSC/pull/742#issuecomment-213446616>
>

-- 

Jakuje:
Thank you for the comment. I will take a look at it next week, if it will work with the cards (I don't have them around now) and submit another pull request.

I hope the idea behind it is correct and I interpret the specification correctly.

frankmorgner:
how does this relate to https://github.com/OpenSC/OpenSC/pull/743? are both PRs needed? if so, why don't you merge them?

frankmorgner:
closing this in favor of #743

PR #743, opened 2016-04-25
Truncate data for ECDSA signature to the size of the key

Jakuje:
Based on the paragraph from PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures:

	If the length of the hash value is larger than the bit length of n, only
	the leftmost bits of the hash up to the length of n will be used. Any
	truncation is done by the token.

Reworked pull request #742 based on the comment. Verified Sign&Verify functionality against NIST PIV Test Cards (ECC: Curve P-256) for all surrounding lengths.

I missed to add the **original problem** for this PR:

Signature of more than 32 B of data fails with opensc, if we have a card supporting only CKM_ECDSA without SHA1. It is reproducible simple using `pkcs11-tool`:

    echo 12345678901234567890123456789012|pkcs11-tool --module opensc-pkcs11.so --id 01 --sign --pin 123456

### DIFF
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -432,6 +432,16 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 		}
 		inlen = modlen;
 	}
+	/* PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures
+	 * If the length of the hash value is larger than the bit length of n, only
+	 * the leftmost bits of the hash up to the length of n will be used. Any
+	 * truncation is done by the token.
+	 */
+	else if (senv.algorithm == SC_ALGORITHM_EC &&
+			(flags & SC_ALGORITHM_ECDSA_HASH_NONE) != 0) {
+		inlen = MIN(inlen, (prkey->field_length+7)/8);
+	}
+
 
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);


frankmorgner:
could you please check the CI report why your code does not compile?

Jakuje:
I don't have a slightest idea. It looks like failures with Microsoft compilers or cygwin? completely unrelated to my changes in code.

Some (`VSVER=10`) fails on:

    'nmake' is not recognized as an internal or external command,

which does not look like my fault.

Others (`VSVER=12`) fail on:

    Command executed with exception: Cannot rename because item at 'c:\zlib-1.2.8' does not exist.

which looks like (also) broken environment.

The last type (`VSVER=14`) I have no idea how to interpret the errors. Rescheduling the job might probably help, if there is such possibility and if the other PR builds work.

frankmorgner:
It should be this one:
```
libeay32MT.lib(cryptlib.obj) : error LNK2019: unresolved external symbol __iob_func referenced in function OPENSSL_stderr
libeay32MT.lib(pem_lib.obj) : error LNK2001: unresolved external symbol __iob_func
libeay32MT.lib(ui_openssl.obj) : error LNK2001: unresolved external symbol __iob_func
libeay32MT.lib(cryptlib.obj) : error LNK2019: unresolved external symbol _vsnwprintf referenced in function OPENSSL_showfatal
opensc.dll : fatal error LNK1120: 2 unresolved externals
```

frankmorgner:
But doesn't look like it's related to your changes

frankmorgner:
this change effects all ec cards. how did you test the code?

frankmorgner:
this effects all ec cards. @philipWendland can you confirm the PR with your applet?

Jakuje:
Yes. It is explained in the first post. I tested using bunch of NIST PIV Test Cards. I would welcome testing with different cards.

No, only one is needed (this one is preferred).

philipWendland:
> this effects all ec cards. @philipWendland can you confirm the PR with your applet?

This PR breaks IsoApplet ECDSA. I suspect this is due to the IsoApplet performing the hash calculation on-card due to limitations of JC 2.2.2.
The plain input seems to get truncated: The line added in the PR assumes it is the hash value, whereas it can be plain text as well.

CardContact:
This PR does not work, as is doesn't handle CKM_ECDSA_SHA1 where hashing is done in the card.

Jakuje:
Thank you for testing. Yes, this should affect only the case with CKM_ECDSA method. If it affects also CKM_ECDSA_SHA1 it is wrong.
I will investigate possible solutions.

Jakuje:
I moved the check little bit further and added a check for the HASH algorithm. Also mentioned in the description what is the driving force for fixing this issue.

Tested again with the same cards and Yubikey NEO (both support only `CKM_ECDSA`), not `CKM_ECDSA_SHA1` and it solved the issue for me.

As I check now, there could be rather 

    inlen = MIN(inlen, modlen/2);

which would be more readable.

philipWendland:
Seems to not affect CKA_ECDSA_SHA1 anymore, works for IsoApplet

frankmorgner:
@Jakuje could you please squash your two commits into one?
@CardContact do you confirm that with the additional changes this is good to be merged?

Jakuje:
@frankmorgner The commits are squashed into one. I tested even with the device from CardContact and it worked fine.

frankmorgner:
thanks

PR #746, opened 2016-04-26
asn1: parse 'securityCondition'

viktorTarasov:
Currently the pkcs15-object data type, together with the existing level of parse of PKCS#15 xDF data,
assumes that protected object contains reference (AuthID) to protecting object in
CommonObjectAttributes.

The PKCS#15 standard and xDF data of existing PKCS#15 cards are more complicated then that:
AuthID of protecting object can be referenced per operation (and referenced in
'AuthID' identifier of related securityCondition),
it can also be present in logical operations 'and', 'or', 'not' combining the different
authentication methods (for ex. SM and/or PUK).

Full parse of such data is too complicated and needs considerable changes in pkcs15-object
data types.

Here is partial improvement:
pkcs15-object is protected if the reference to protecting object is found in:
- CommonObjectAttributes as 'AuthID' identifier (currently implemented);
- 'AuthID' identifier in one of SecurityCondtions (the first occurence);
- 'AuthID' from securityCondition operand of 'and' operation.

### DIFF
--- a/src/libopensc/pkcs15.h
+++ b/src/libopensc/pkcs15.h
@@ -360,9 +360,27 @@ struct sc_pkcs15_keyinfo_gostparams
 #define SC_PKCS15_ACCESS_RULE_MODE_INT_AUTH     0x200
 #define SC_PKCS15_ACCESS_RULE_MODE_EXT_AUTH     0x400
 
+struct sc_pkcs15_auth_reference {
+	unsigned auth_method;
+	int se_num;
+};
+
+struct sc_pkcs15_auth_id_and_ref {
+	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_auth_reference auth_ref;
+};
+
+struct sc_pkcs15_security_condition {
+	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_auth_reference auth_ref;
+	struct sc_pkcs15_auth_id_and_ref sc_not;
+	struct sc_pkcs15_auth_id_and_ref sc_and[2];
+	struct sc_pkcs15_auth_id_and_ref sc_or[2];
+};
+
 struct sc_pkcs15_accessrule {
 	unsigned access_mode;
-	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_security_condition sc;
 };
 typedef struct sc_pkcs15_accessrule sc_pkcs15_accessrule_t;
 
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -1113,11 +1113,38 @@ err:
 	return ret;
 }
 
-/* TODO: According to specification type of 'SecurityCondition' is 'CHOICE'.
- *       Do it at least for SC_ASN1_PKCS15_ID(authId), SC_ASN1_STRUCT(authReference) and NULL(always). */
+static const struct sc_asn1_entry c_asn1_auth_reference[3] = {
+	{ "authMethod", SC_ASN1_BIT_FIELD, SC_ASN1_TAG_BIT_STRING, SC_ASN1_OPTIONAL, NULL, NULL },
+	{ "seIdentifier", SC_ASN1_INTEGER, SC_ASN1_TAG_INTEGER, SC_ASN1_OPTIONAL, NULL, NULL },
+	{ NULL, 0, 0, 0, NULL, NULL }
+};
+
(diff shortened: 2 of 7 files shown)


src/libopensc/pkcs15.h
```diff
@@ -360,9 +360,27 @@ struct sc_pkcs15_keyinfo_gostparams
...
+	struct sc_pkcs15_auth_reference auth_ref;
+	struct sc_pkcs15_auth_id_and_ref sc_not;
+	struct sc_pkcs15_auth_id_and_ref sc_and[2];
+	struct sc_pkcs15_auth_id_and_ref sc_or[2];
+};
+
 struct sc_pkcs15_accessrule {
 	unsigned access_mode;
-	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_security_condition sc;
```
frankmorgner: could you rename the member to `condition`, sc is already heavily used in a different context.

src/libopensc/pkcs15.h
```diff
@@ -360,9 +360,27 @@ struct sc_pkcs15_keyinfo_gostparams
...
+	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_auth_reference auth_ref;
+};
+
+struct sc_pkcs15_security_condition {
+	struct sc_pkcs15_id auth_id;
+	struct sc_pkcs15_auth_reference auth_ref;
+	struct sc_pkcs15_auth_id_and_ref sc_not;
+	struct sc_pkcs15_auth_id_and_ref sc_and[2];
+	struct sc_pkcs15_auth_id_and_ref sc_or[2];
```
frankmorgner: this could also be `condition_or`, which would be more readable.

frankmorgner:
As far as I can see, this changes the parsing of the security condition. But the actual cards still assume that there is only a single object that needs to be verified, correct? In other words, there is currently no magic function call that evaluates all condtions, right? Also, there is no magic that checks the security condition with the current security environment, correct?

src/libopensc/pkcs15.h
viktorTarasov: agree

src/libopensc/pkcs15.h
viktorTarasov: agree

viktorTarasov:
yes,
in current pkcs15-object data type only one object can be referenced by AuthID. That's why the full implementation of specification is too complicated.

I some of actual cards the reference to protecting object is not present in Common attributes, but present somewhere deep inside the security conditions.
For such cases, and with the current pkcs15-object data type,  and if in security conditions there is only one AuthID reference,
we can treat this reference as usual AuthID referenced object.
 
No, there is no code to evaluate multiple conditions.
SM conditions in PKCS#15 are simply ignored -- SM is managed on the card level, after analyse of ACLs of operation to be done.

frankmorgner:
I agree that complicated things are complicated. However, we could at least evaluate simple "and" conditions automatically.

viktorTarasov:
@frankmorgner 
> However, we could at least evaluate simple "and"  conditions automatically.

_AND_ is evaluated -- if one of its operand references protecting object, and if there is no such reference in CommonAttribute, then the first one is used by OpenSC PKCS#15 framework as AuthID of protecting object.

What do you mean under _automatically_?

frankmorgner:
hmm, I can see that the security attributes are parsed and that the authid gets initialized. However, what I mean with automatic evaluation is that somewhere in `pkcs15_login` *all pins required by the AND condition* would be verified.

viktorTarasov:
It has to be done not (or not only)  on _login_, but on authorization of operation. This needs a considerable intervention into the common framework, even for partial implementation.

I'm thinking, for example, about 'OR' condition -- both operands cannot have identical preference for the user, should some dialog be implemented? configuration? cache?
I've already seen 'OR' with two PINs in some of the Oberthur's card.
With 'AND' it's more simple, afais, most frequently one of the operands is SM, which is simply ignored by OpenSC PKCS#15 framework.

frankmorgner:
I'm fine with not having this built in right now. I don't say it's easy to do. I just think it would be nice to have...

frankmorgner:
The `NEWS` file already contains a reference to this PR, will you integrate this into the new release?

viktorTarasov:
I will remove it from NEWS, thanks.

frankmorgner:
will you do the renaming?

PR #753, opened 2016-05-06
Use two hex digits with sc_pkcs15_format_id

dengert:
Avoid having problems with sc_hex_to_bin  when called from sc_pkcs15_format_id

 Changes to be committed:
	modified:   pkcs15-piv.c


dengert:
Will submit new version...

PR #754, opened 2016-05-08
Use two hex digits with sc_pkcs15_format_id (2) 

dengert:
Avoid having problems with sc_hex_to_bin  when called from sc_pkcs15_format_id
both for id and auth_id numbers

 Changes to be committed:
	modified:   pkcs15-piv.c


dengert:
This is replacement for #753 that was closed.

dengert:
@frankmorgner  I think this is ready to merge. The appveyor error is a timeout in the appveyor scripts.

frankmorgner:
I agree, but I don't know where to which branch merge this. @viktorTarasov ?

dengert:
@viktorTarasov 
towards-opensc-0.16.0  would be where this should be committed. 

This fix cleans up the code so it can work with  https://github.com/OpenSC/OpenSC/commit/c63ba858e32950b975bc6817c4dd9276ceea8715
which is already on the towards-opensc-0.16.0 branch.  Without this fix the 5 or so cards will not work
with that commit.

But if you have a version of sc_pkcs15_format_id (or sc_hex_to_bin  that accepts a single digit input
and the one call sc_pkcs15_format_id("NONE", ... this fix would not be needed. 
(But we have not settled on how to change sc_hex_to_bin. )

The fix cleans up the code to not use single digit input or "NONE" so would also work with an 
any changes to sc_hex_to_bin, and sc_hex_to_bin does not need to do special single digit handling.

viktorTarasov:
I prefer to keep the possibility of _one character hex string_ in _sc_hex_to_bin()_
and apply the fix to not use the _one character hex string_. To be in safety .

PR #756, opened 2016-05-11
Changes to ECDH implementation

hhonkanen:
After testing ECDH algorithm with pkcs11-tool and a new MyEID card, we noticed that some changes were needed. 

In derive_key in pkcs11-tool there were variables that were used via pointers after falling out of scope, and this worked only if you were lucky and the memory was not overwritten yet. 

In pkcs15-sec.c, it seems that  sc_pkcs15_derive was designed so that the derived key is expected to be stored on card. However in ECDH at least with MyEID the shared secret is returned in this operation. Reviewers please check that our changes don't break other functionality.

card-myeid.c required minor changes too. 

After these modifications successfully derived a shared secret using a 521 bit EC key.

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -193,7 +193,7 @@ static int myeid_init(struct sc_card *card)
 	card->caps |= SC_CARD_CAP_RNG | SC_CARD_CAP_ISO7816_PIN_INFO;
 
 	card->max_recv_size = 255;
-	card->max_send_size = 255;	
+	card->max_send_size = 255;
 
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
@@ -921,7 +921,7 @@ int myeid_ecdh_derive(struct sc_card *card, const u8* pubkey, size_t pubkey_len,
 
 	struct sc_apdu apdu;
 	u8 sbuf[SC_MAX_APDU_BUFFER_SIZE];
-	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];	
+	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
 
 	int r;
 	size_t ext_len_bytes;
@@ -933,6 +933,8 @@ int myeid_ecdh_derive(struct sc_card *card, const u8* pubkey, size_t pubkey_len,
 	
 	/* Fill in "Data objects in dynamic authentication template" (tag 0x7C) structure 
 	*
+	* TODO: encode the structure using OpenSC's ASN1-functions.
+	* 
 	*  Size of the structure depends on key length. With 521 bit keys two bytes are needed for defining length of a point.
 	*/		
 	
@@ -977,7 +979,7 @@ int myeid_ecdh_derive(struct sc_card *card, const u8* pubkey, size_t pubkey_len,
 
 	memcpy(out, rbuf, apdu.resplen);
 
-	LOG_FUNC_RETURN(card->ctx, r);
+	LOG_FUNC_RETURN(card->ctx, apdu.resplen);
 }
 
 
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -674,10 +674,10 @@ static int myeid_set_security_env_ec(sc_card_t *card, const sc_security_env_t *e
 		sc_log(card->ctx, "Decipher operation is not supported with EC keys.");
 		return SC_ERROR_NOT_SUPPORTED;
 		break;
-	case SC_SEC_OPERATION_SIGN:	
+	case SC_SEC_OPERATION_SIGN:
 		apdu.p1 = 0x41;
-		apdu.p2 = 0xB6;		
-		break;	
+		apdu.p2 = 0xB6;
+		break;
 	case SC_SEC_OPERATION_DERIVE:
 		apdu.p1 = 0x41;
 		apdu.p2 = 0xA4;
@@ -1354,6 +1354,7 @@ static int myeid_get_info(struct sc_card *card, u8 *rbuf, size_t buflen)
 	card->version.fw_minor = rbuf[7];
 	/* add version to name */
 	snprintf(card_name_buf, sizeof(card_name_buf),
+			"%s %d.%d.%d", card->name, rbuf[5], rbuf[6], rbuf[7]);
 	card->name = card_name_buf;
 
 	LOG_FUNC_RETURN(card->ctx, r);
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -134,7 +134,7 @@ static int myeid_init(struct sc_card *card)
 	u8 defatr[SC_MAX_ATR_SIZE];
 	size_t len = sizeof(defatr);
 	const char *atrp = myeid_atrs[MYEID_INFINEON_CHIP_ATR];
-
+	
 	LOG_FUNC_CALLED(card->ctx);
 
 	card->name = myeid_card_name;
@@ -193,7 +193,7 @@ static int myeid_init(struct sc_card *card)
 	card->caps |= SC_CARD_CAP_RNG | SC_CARD_CAP_ISO7816_PIN_INFO;
 
 	card->max_recv_size = 255;
-	card->max_send_size = 255;
+	card->max_send_size = 255;	
 
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
@@ -1327,7 +1327,6 @@ static int myeid_get_info(struct sc_card *card, u8 *rbuf, size_t buflen)
 	card->version.fw_minor = rbuf[7];
 	/* add version to name */
 	snprintf(card_name_buf, sizeof(card_name_buf),
-			"%s %d.%d.%d", card->name, rbuf[5], rbuf[6], rbuf[7]);
 	card->name = card_name_buf;
 
 	LOG_FUNC_RETURN(card->ctx, r);
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -674,9 +674,13 @@ static int myeid_set_security_env_ec(sc_card_t *card, const sc_security_env_t *e
 		sc_log(card->ctx, "Decipher operation is not supported with EC keys.");
 		return SC_ERROR_NOT_SUPPORTED;
 		break;
-	case SC_SEC_OPERATION_SIGN:
+	case SC_SEC_OPERATION_SIGN:	
 		apdu.p1 = 0x41;
-		apdu.p2 = 0xB6;
+		apdu.p2 = 0xB6;		
+		break;	
+	case SC_SEC_OPERATION_DERIVE:
+		apdu.p1 = 0x41;
+		apdu.p2 = 0xA4;
 		break;
 	default:
 		return SC_ERROR_INVALID_ARGUMENTS;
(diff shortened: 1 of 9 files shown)


frankmorgner:
CI can't build your changes

src/libopensc/card-myeid.c
```diff
@@ -131,9 +135,13 @@ static int myeid_init(struct sc_card *card)
 	u8 defatr[SC_MAX_ATR_SIZE];
 	size_t len = sizeof(defatr);
 	const char *atrp = myeid_atrs[MYEID_INFINEON_CHIP_ATR];
-
+
```
frankmorgner: please remove this whitespace

src/libopensc/card-myeid.c
```diff
@@ -186,7 +194,7 @@ static int myeid_init(struct sc_card *card)
 	card->caps |= SC_CARD_CAP_RNG | SC_CARD_CAP_ISO7816_PIN_INFO;
 
 	card->max_recv_size = 255;
-	card->max_send_size = 255;
+	card->max_send_size = 255;
```
frankmorgner: please remove this whitespace

src/libopensc/card-myeid.c
```diff
@@ -910,25 +922,46 @@ int myeid_ecdh_derive(struct sc_card *card, const u8* pubkey, size_t pubkey_len,
 
 	struct sc_apdu apdu;
 	u8 sbuf[SC_MAX_APDU_BUFFER_SIZE];
-	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
+	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
```
frankmorgner: please remove this whitespace

src/libopensc/card-myeid.c
```diff
@@ -47,6 +47,10 @@
 #define MYEID_STATE_ACTIVATED	0x07
 
 #define MYEID_INFINEON_CHIP_ATR		0x04
+#define MYEID_CARD_NAME_MAX_LEN		100
+
+static const char *myeid_card_name = "MyEID";
+static char card_name_buf[MYEID_CARD_NAME_MAX_LEN];
```
frankmorgner: use a local variable

src/libopensc/card-myeid.c
```diff
@@ -131,9 +135,13 @@ static int myeid_init(struct sc_card *card)
 	u8 defatr[SC_MAX_ATR_SIZE];
 	size_t len = sizeof(defatr);
 	const char *atrp = myeid_atrs[MYEID_INFINEON_CHIP_ATR];
-
+	
 	LOG_FUNC_CALLED(card->ctx);
+
+	card->name = myeid_card_name;
```
frankmorgner: as far as i can see the name did not fell out of scope, it was never initialized. and to be honest, you don't need it. All information you added with this commit was already present in the card driver's name

src/libopensc/card-myeid.c
```diff
@@ -918,25 +922,46 @@ int myeid_ecdh_derive(struct sc_card *card, const u8* pubkey, size_t pubkey_len,
...
+	
 	sbuf[0] = 0x7C;
-	sbuf[1] = pubkey_len + 4;
-	sbuf[2] = 0x85;
-	sbuf[3] = pubkey_len;
-	memcpy(&sbuf[4], pubkey, pubkey_len);
+	ext_len_bytes = 0;
 
-	apdu.lc = pubkey_len + 4;
+	if (pubkey_len > 127)
```
frankmorgner: could you please detail the actual problem you are solving in the commit message?

src/tools/pkcs11-tool.c
```diff
@@ -2695,7 +2698,7 @@ derive_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 		if (!ecpoint || !ecgroup)
 			util_fatal("Failed to parse other EC kry from %s", opt_input);
 
-		len = EC_POINT_point2oct(ecgroup, ecpoint, POINT_CONVERSION_UNCOMPRESSED, buf, sizeof(buf),NULL);
+		len = EC_POINT_point2oct(ecgroup, ecpoint, POINT_CONVERSION_UNCOMPRESSED, buf, buf_size, NULL);
```
frankmorgner: could you please detail the actual problem you are solving in the commit message?

src/libopensc/card-myeid.c
hhonkanen: See myeid_init. card->name is set to point to this buffer, and it cannot be unallocated when returning, so a local variable won't do. I changed this to a global earlier, because when it was implemented as a local, card->name pointed to garbage later. Do you have another solution to this?

src/tools/pkcs11-tool.c
```diff
@@ -2683,7 +2683,11 @@ derive_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 		const EC_GROUP *ecgroup = NULL;
 		const EC_POINT * ecpoint = NULL;
 		
-		buf = malloc(buf_size);	/* TODO: caculate some sensible value from key length */	
+		buf_size = 512; /* TODO: calculate buffer size from key length */	
+		buf = malloc(buf_size);	
+		
+		if (!buf)
+		    util_fatal("Failed to allocate memory for other party's public point.")
```
frankmorgner: please update your pull request by squashing the clean up commit onto the actual commit that contain the real logic

src/libopensc/card-myeid.c
hhonkanen: Take a look at myeid_get_info, it is used there again. This change fixed an actual problem that resulted in garbage being displayed with command opensc-tool -n

src/libopensc/card-myeid.c
hhonkanen: The problem solved here is that 521 bit ECC points require two bytes in DER encoding to indicate length, whereas with smaller keys only one length byte is needed.

src/tools/pkcs11-tool.c
hhonkanen: Moved buf to function level because it went out of scope too early, and allocate dynamically only when needed. In future implementations of other algorithms, required buffer size can be calculated differently. Sorry I am a beginner wtih GIT and don't know how to modify the commit message later, should I create a new commit?

src/libopensc/card-myeid.c
hhonkanen: BTW, this change has been merged to master already. I don't know why this is included again in this PR. I come from Windows background have minimal knowledge of GIT, so my PR's might be a bit messy, sorry about extra work caused by this.

hhonkanen: To be more clear, myeid_get_info adds version to card->name. There was a bug in a previous version, please see https://github.com/OpenSC/OpenSC/commit/fe2312dd1966f9fac6751345ab894b89b4f7df3c
Maybe you are right that we could do without global myeid_card_name (is card->name set already when myeid_init is called?), but card_name_buf cannot be local for this to work. This has to be tested if we modify it.

src/libopensc/card-myeid.c
frankmorgner: please use OpenSC's asn1 wrappers then

src/libopensc/card-myeid.c
frankmorgner: OK, then nevermind the card name. However, you still need to rebase onto the current master. git shell on windows is no different than on linux. I'm afraid you will need to get familiar with it. here are some links that may be interesting http://chris.beams.io/posts/git-commit/ https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

hhonkanen: Yes I should get more familiar with git. I did "git merge upstream/master" before making my changes and believed that this syncs my code with the master, but apparently that is not enough.Please understand that with this PR I was trying to quickly fix something that was not working at all, before moving on to another project. I am willing to correct all of the details but currently have very limited time to allocate for OpenSC dev so it might take time.

src/libopensc/card-myeid.c
hhonkanen: This would be better, but it might take a while before I will have the next possibility to get familiar with the asn1 functions and refactor this. Would it be possible to make this a TODO and merge this implementation to fix non working code?

src/libopensc/pkcs15-sec.c
```diff
@@ -287,7 +287,15 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
-	*poutlen = r;
+	
+	switch (obj->type) {
+		case SC_PKCS15_TYPE_PRKEY_EC:
+	/* But in case the card returned the shared secret, don't overwrite *poutlen */
+			break;
+		default:
+			*poutlen = r;
+	}
+
```
dengert: This sections of the code is not needed. 
The use_key in line 277 should be returning the length of the derived data the card returned, 
or 0 if the card did not return any data. if r<0 indicates an error and returned at 278.
If card returned rsa padded data  the padding is removed and r would be the new length. 

So in the original code:
*poutlen = r;

This will be setting *poutlen to the size of the data returned - any padding removed. 
if the the card as in the EC case, there is no padding , and if the card returned data, *poutlen would already be the same as r. 

Please remove this section as it makes it more confusing.

hhonkanen: Thanks for your comment. This code was added due to a misunderstanding. Actually the problem was in the myeid driver where functions myeid_decipher and myeid_ecdh_derive returned SC_SUCCESS instead of the data length, so we got 0 from use_key and thought that the problem is here. I will remove this code and fix the myeid driver.

hhonkanen:
First of all thanks to Frank and Doug for reviewing this PR. 

I have made the modifications to the code as requested, except for building the DER structure with OpenSC's asn1-functions in card_myeid.c:myeid_ecdh_derive, which I marked as a TODO. 
Diffing with upstream/master now shows the changes as they should be and build succeeds. 

However, due to my miserably lousy git skills I was unable to edit/merge/squash the commit messages as Frank requested. I tried to do this using interactive rebase but ran into some conflicts and only managed to mess up my local repo more. So I gave up for now and decided that I have to take my time to learn to do these things correctly. 

I kindly ask if anyone with more experience with git could modify the comments as necessary, or give me instructions on how to do that? 

What I would like to do is to combine all the differences between OpenSC master and the current state of my fork into one commit with a message like this: 

pkcs11:tool: Fixes to derive_key in pkcs11-tool. Some variables used in ECDH operation were falling out of scope when they were still needed.

card-myeid.c: myeid_ecdh_derive: 
		 Handle various keys sizes correctly when encapsulating public key to a DER structure.
		 Return length of shared secret as return value.
   	      set_security_env_ec: 
		 Added setting security environment for derive operations.

frankmorgner:
Try the following:
```
git remote add 1 https://github.com/OpenSC/OpenSC
git fetch 1
git checkout 1/master -b new
git merge master
git reset 1/master
```
Now your commits should be identical with OpenSC/master and you changes should be what is in your master (i.e. this PR). Now you can individually pick and commit changes of your files.

frankmorgner:
@hhonkanen did you make any progress?

src/tools/pkcs11-tool.c
```diff
@@ -2651,7 +2651,10 @@ derive_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 	CK_BBOOL false = FALSE;
 	CK_OBJECT_HANDLE newkey = 0;
 	CK_RV rv;
+	CK_ECDH1_DERIVE_PARAMS ecdh_parms;
 	int fd, r;
+	unsigned char* buf = NULL;  /* buffer for handling input data */
```
frankmorgner: what's the reason for using an allocated buffer instead of a static buffer?

what's the reason for making the buffer global in `derive_key` instead of leaving it in the `CKM_ECDH1_DERIVE` case?

hhonkanen: The "buf" buffer is indirectly referred after the CKM_ECDH1_DERIVE case falls out of scope. That's why it has to be declared outside of the case. Look at the code just before p11->C_DeriveKey.
I made it dynamic thinking that if more mechanisms are added later, each can calculate and allocate the space they need depending on the algorithm and key length. 

See commit 2ba87640d23bcff3f1a130ea28bf02f4f8afbb3e By mouse07410, it fixes the same thing.

viktorTarasov:
Will you agree with a550045237ed5e1fab4b3fb19fdda3577a08b74a ?
It's re-based version of your PR, that was slightly modified in _pkcs11-tool_ part.

hhonkanen:
Viktor, thank you for rebasing and fixing this PR. 

Your changes in a550045 are ok on my part.

viktorTarasov:
applied to master in 6cd28cfc7c5d9aa2ace94d60549f58f5cd3c4a49

PR #759, opened 2016-05-16
Unbreak tests without arguments (automatic selection of the reader)

Jakuje:
The logic is obvious, but probably copy-paste error. Using negative `opt_reader` in `sc_ctx_get_reader()` made the test suite segfault.

There is one more segfault in `p15dump`.

### DIFF
--- a/src/tests/p15dump.c
+++ b/src/tests/p15dump.c
@@ -69,8 +69,10 @@ static int dump_unusedspace(void)
 	else if (p15card->file_app != NULL) {
 		path = p15card->file_app->path;
 		sc_append_path_id(&path, (const u8 *) "\x50\x33", 2);
-	} else
+	} else {
+		printf("\nCan't find unused space file.\n");
 		return -1;
+	}
 	path.count = -1;
 
 	r = sc_pkcs15_read_file(p15card, &path, &buf, &buf_len);
--- a/src/tests/p15dump.c
+++ b/src/tests/p15dump.c
@@ -66,10 +66,11 @@ static int dump_unusedspace(void)
 
 	if (p15card->file_unusedspace != NULL)
 		path = p15card->file_unusedspace->path;
-	else {
+	else if (p15card->file_app != NULL) {
 		path = p15card->file_app->path;
 		sc_append_path_id(&path, (const u8 *) "\x50\x33", 2);
-	}
+	} else
+		return -1;
 	path.count = -1;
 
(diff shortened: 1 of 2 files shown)


src/tests/p15dump.c
```diff
@@ -66,10 +66,11 @@ static int dump_unusedspace(void)
 
 	if (p15card->file_unusedspace != NULL)
 		path = p15card->file_unusedspace->path;
-	else {
+	else if (p15card->file_app != NULL) {
 		path = p15card->file_app->path;
 		sc_append_path_id(&path, (const u8 *) "\x50\x33", 2);
-	}
+	} else
+		return -1;
```
frankmorgner: Could you use a relative path to `"\x50\x33"` instead of returning an error? This would also print a nice error message if the file could not be found.

Jakuje: (sorry for the late reply. I missed this comment somehow)
@frankmorgner I am not sure what you mean by the relative path in this context.
Writing some error would make sense. Thanks for the idea.

frankmorgner: I thought of selecting `0x5033` within the current directory.

viktorTarasov:
this PR resolves the _segfault_ case and is merged to _towards-0.16.0_ as it is, 

but will stay open for a while to decide the usefulness of selecting _5033_ in the current DF.
This supplement can be realized in a dedicated PR.

Jakuje:
If I read [pkcs#15](ftp://ftp.cert.dfn.de/pub/pca/docs/PKCS/ftp.rsa.com/pkcs-15/pkcs-15v1_1.doc) correctly:

    unusedPath 	[1] Path OPTIONAL,

The `unusedPath` is optional and file does not have to be there (and for read-only cards it does not make much sense). Also the `file_app` does not have to be there. I didn't find any card where would this test work.

I added the error message and I am open to test if there would be some idea how to find a file in current directory. All the ideas I tried failed for me.

frankmorgner:
nevermind

Jakuje:
It was not so easy. `sc_detect_card_presence()` returns not only `1`, but also different values and therefore it is segfaulting again with different configurations. I will check if there will be some more problems and open another PR, since I can't open this one.

frankmorgner:
With "nevermind", I meant my comments regarding automatically finding the file. This never seemed to be a problem so far, so no need to add this now. Thanks for the fix.

frankmorgner:
all your changes are merged now

Issue #737, opened 2016-04-18
sc_hex_to_bin() allows odd number of hexchars


CardContact:
sc_hex_to_bin() allows decoding hexadecimal strings with an odd number of characters, e.g. 12345 is decoded as 0x123405.

While this seems reasonable if the hexadecimal string contains delimiters, it poses a problem if a character was accidentally left out or added.

We could change the code to check that two nibbles are required to form a byte, but this would break existing code that allows or uses 1 digit hex values.

DDvO:
Here is a better solution: allow for odd numbers of hexchars, while putting the leading zero required in this case in the first (most-significant) byte, not in the last one. For the given examle, 12345 should be encoded as 0x01, 0x23, 0x45. Here's a variant of `sc_hex_to_bin()` with the respective fix:
```
int sc_hex_to_bin(const char *in, u8 *out, size_t *outlen)
{
	int err = SC_SUCCESS;
	size_t left, count = 0;
	int nybbles = 2;

	assert(in != NULL && out != NULL && outlen != NULL);
        left = *outlen;

	if (strlen(in) % 2)
		nybbles = 1; // any leading zero should be in most-significant byte, not last one!
	while (*in != '\0') {
		int byte = 0;

		while (nybbles-- && *in && *in != ':' && *in != ' ') {
			char c;
			byte <<= 4;
			c = *in++;
			if ('0' <= c && c <= '9')
				c -= '0';
			else
			if ('a' <= c && c <= 'f')
				c = c - 'a' + 10;
			else
			if ('A' <= c && c <= 'F')
				c = c - 'A' + 10;
			else {
				err = SC_ERROR_INVALID_ARGUMENTS;
				goto out;
			}
			byte |= c;
		}
		if (*in == ':' || *in == ' ')
			in++;
		if (left <= 0) {
                        err = SC_ERROR_BUFFER_TOO_SMALL;
			break;
		}
		out[count++] = (u8) byte;
		left--;
		nybbles = 2;
	}

out:
	*outlen = count;
	return err;
}
```
BTW, the same holds for `hex_to_bin()` in `pkcs11-tools.c`.

DDvO:
Here is one more related (minor) fix: add
```
while (*in == '0' && *(++in) != '\0') ; // strip leading zeros in input
```
just before 
```
	if (strlen(in) % 2)
		nybbles = 1; // any leading zero in output should be in most-significant byte, not last one!
```

frankmorgner:
[...]

PR #761, opened 2016-05-17
Corrected handling of hex input with uneven number of digits in in sc_hex_to_bin()

DDvO:
This is the pull request for the solution proposal I gave for #737


viktorTarasov:
Sorry, but this has been already discussed.
Decision is to let the hex length be odd only for the length '1' (and to clean up the code from using of static strings with length '1').
The other lengths has to be even.

PR #763, opened 2016-05-17
Update reference URLs in ReadMe.html.in

scootersmk:
RSA.com link broken due to move to EMC site.

http://www.opensc-project.org/opensc appears to be a parked domain.


frankmorgner:
hmm, this now makes me think longer than I had expected. Actually we should link to https://www.oasis-open.org/committees/pkcs11/ and to ISO 7816-15 instead... However, there is no clear URL for the latter...

frankmorgner:
Also, people may want to enjoy reading PKCS#15 for free at http://www.emc.com/emc-plus/rsa-labs/standards-initiatives/pkcs-15-cryptographic-token-information-format.htm

viktorTarasov:
merged to _towards-0.16.0_

PR #770, opened 2016-05-27
Fixed multi-epass2003 problem and improve code

FeitianSmartcardReader:
Fixed bug #731 #730

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -67,7 +67,6 @@ static struct sc_card_driver epass2003_drv = {
 
 #define KEY_TYPE_AES	0x01	/* FIPS mode */
 #define KEY_TYPE_DES	0x02	/* Non-FIPS mode */
-static unsigned char g_smtype;	/* sm cryption algorithm type */
 
 #define KEY_LEN_AES	16
 #define KEY_LEN_DES	8
@@ -79,7 +78,6 @@ static unsigned char PIN_ID[2] = { ENTERSAFE_USER_PIN_ID, ENTERSAFE_SO_PIN_ID };
 /*0x00:plain; 0x01:scp01 sm*/
 #define SM_PLAIN				0x00
 #define SM_SCP01				0x01
-static unsigned char g_sm;	/* if perform sm or not */
 
 static unsigned char g_init_key_enc[16] = {
 	0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C,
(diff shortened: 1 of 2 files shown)


src/libopensc/card-epass2003.c
```diff
@@ -324,10 +328,10 @@ gen_init_key(struct sc_card *card, unsigned char *key_enc, unsigned char *key_ma
 	apdu.le = apdu.resplen = 28;
 	apdu.resp = result;	/* card random is result[12~19] */
 
-	tmp_sm = g_sm;
-	g_sm = SM_PLAIN;
+	tmp_sm = exdata->sm;
```
frankmorgner: You could check if `exdata != NULL`

frankmorgner: This should always be done before dereferencing exdata

frankmorgner: (I think there should not be any problems. Its just a precautionary measure.)

frankmorgner:
@velter could you have a look?

velter:
@frankmorgner This look like to be the same patch as #731 but applied to master instead of towards-opensc-0.16.0

I agree with the missing dereferencing test. What do you think about adding

LOG_TEST_RET(card->ctx, exdata?0:SC_ERROR_INVALID_ARGUMENTS, "unable to get extended driver data");

in each function trying to access card->drv_data ?

I can provide a patch if you are ok with this.

frankmorgner:
If you could provide a PR, that would be great. I'd use a simple
```
if (!card->drv_data) return SC_ERROR_INVALID_ARGUMENTS
```

FeitianSmartcardReader:
@frankmorgner and @velter  modify the code, have a look, thanks

velter:
@frankmorgner PR #780 add check for invalid driver datas. The PR is against the towards-opensc-0.16.0 branch

FeitianSmartcardReader:
@frankmorgner just modify card-entersafe.c file, add new ATR support and also double check the source code, add code to avoid memory leak, thanks

frankmorgner:
before adding a commit, please at least try to compile it:
```
card-epass2003.c(324): error C2146: syntax error: missing ';' before identifier 'epass2003_exdata'
```

frankmorgner:
@FeitianSmartcardReader I am expecting you to maintain the code of your cards.

FeitianSmartcardReader:
@frankmorgner  Sorry for that. Did fix it and compile it to make sure.
Have a check, thanks https://github.com/OpenSC/OpenSC/pull/770/commits/7502c2727e55b841648f9df84297f83ca476af0d

frankmorgner:
Could you please squash 7502c2727e55b841648f9df84297f83ca476af0d onto a3dcc0cd820f411c6f5ae8de0d5988545f8825ec so that it appears as one single commit? Thanks!

FeitianSmartcardReader:
Thanks @frankmorgner , Will submit another pull request to make clearly report in OpenSC.

PR #748, opened 2016-04-27
Document gids tool

Jakuje:
I put together manual page for `gids-tool`, only based on the `./gids-tool -?`. Related to pull request #651. Can you review if it makes sense or add any needed information.

Also adding related `.gitignore` entries.

PR #772, opened 2016-05-29
gids bux fixing

vletoux:
fix the problem if the serial number is entered in the command line
add gids-tool to gitignore


Jakuje:
Gitignore is already merged to `towards-0.16.0` from pull request #748.

vletoux:
I removed to commit in conflict with #748

frankmorgner:
why don't you check `strlen(serial)`? With the current implementation you are allowing strings bigger than 16 bytes input. The input gets truncated to `sizeof(param.cardid)` in `sc_hex_to_bin`

PR #776, opened 2016-05-30
Fixed bug in pkcs11_tool.c:derive_key()

mouse07410:
Correct execution depended on undefined compiler behavior. Now it doesn't, as the two offending variables are moved to the correct scope.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2659,7 +2659,11 @@ derive_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 		{CKA_ENCRYPT, &true, sizeof(true)},
 		{CKA_DECRYPT, &true, sizeof(true)}
 	};
-
+#if defined(ENABLE_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x00908000L && !defined(OPENSSL_NO_EC) && !defined(OPENSSL_NO_ECDSA)
+	CK_ECDH1_DERIVE_PARAMS ecdh_parms;
+	unsigned char buf[512];
+#endif /* ENABLE_OPENSSL etc */
+	
 	if (!opt_mechanism_used)
 		if (!find_mechanism(slot, CKF_DERIVE|CKF_HW, NULL, 0, &opt_mechanism))
 			util_fatal("Derive mechanism not supported\n");
@@ -2674,8 +2678,6 @@ derive_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 	case CKM_ECDH1_DERIVE:
 		/*  Use OpenSSL to read the other public key, and get the raw verion */
 		{
-		CK_ECDH1_DERIVE_PARAMS ecdh_parms;
-		unsigned char buf[512];
 		int len;
 		BIO     *bio_in = NULL;
 		const EC_KEY  *eckey = NULL;


frankmorgner:
applied in towards-opensc-0.16.0

mouse07410:
Thanks! 

@frankmorgner is there any difference/reason why it was "applied" rather than "merged"?

frankmorgner:
It's the same, but without the "merge commit". Yes, you may delete your branch.

viktorTarasov:
During this release there has been questions, discussion about git workflow.

So, my question, will it be more comprehensive if we return back to WF
with two permanent branch _master_ and _devel_, and temporary branches dedicated to _feature_, _fix_ and _release_ ?

If so, to finalize the current release, I will
- abandon current _master_ -- _towards-0.16.0_ contain all commits that have been applied to _master_ after release started;
- create new _master_,  where only existing official releases will be gathered (merged with _no-ff_);
- merge current _towards-0.16.0_ into this _master (always with _no-ff_) with amend of the update of _version.m4_ (to remove RC package suffix) and tag this commit as official _0.16.0_.
- fork from resulting _master_ a new _devel_ branch and set it as a default branch in github project.

From this moment all PR, development, fixes, etc.  has-to-be/will-be applied to _devel_,
until the next official release, when release dedicated branch will be created, worked on, and merged into _master_.

dengert:
On 6/1/2016 11:43 AM, viktorTarasov
      wrote:
    
    
      During this release there has been questions, discussion about
        git workflow.
      So, my question, will it be more comprehensive if we return
        back to WF
        with two permanent branch master and devel,
        and temporary branches dedicated to feature, fix
        and release ?
      If so, to finalize the current release, I will
      
        abandon current master -- towards-0.16.0
          contain all commits that have been applied to master
          after release started;
        create new master, where only existing official
          releases will be gathered (merged with no-ff);
      
    
    Do you need to abandon current master? It looks like if you just
    merged towards-0.16.0 into master would do the same thing and
    preserve any history.  towards-0.16.0 is in effect the devel branch
    described below. 
    
    
      
        merge current towards-0.16.0 into this master
            (always with _no-ff) with amend of the update of version.m4
          (to remove RC package suffix) and tag this commit as official
          0.16.0.
        fork from resulting master a new devel
          branch and set it as a default branch in github project.
      
    
    
    I still don't understand why master is not the default branch, and
    it only contains changes we expect to be in the next release. Many
    new developers  base their changes on master expecting master will
    what is in the next release. The recent use of the towards-0.16.0
    branch was not obvious to new developers.  Does setting the default
    branch in github to devel make things better?
    
    
      
      
      From this moment all PR, development, fixes, etc.
        has-to-be/will-be applied to devel,
        until the next official release, when release dedicated branch
        will be created, worked on, and merged into master.
    
    Sounds like you can start right now, and merge towards-0.16.0 into
    master. 
    
    
      —
        thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

frankmorgner:
I am still confused, too. @viktorTarasov are the following links describing your idea: https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows (or in more depth http://nvie.com/posts/a-successful-git-branching-model/)?

If so, I think it looks good and it has worked for others, so :+1: However, I agree with @dengert that master, i.e. the current stable branch with hotfixes applied should be the default branch, because this is what people see first. Also, I agree with him that master should not be re-written or abandoned; simply use it as is with the new policy applied.

viktorTarasov:
Ok. To resume, we stay with what we have now:
- keep for _master_ its current _development_ role -- default branch, placeholder for all activity between releases;
- with such _master_ there is no _devel_ branch;
- for future I'm ready to abandon my obstinacy to have  release dedicated branch and do all release work directly on _master_;
- dedicated temporary branches for the more or less considerable features.

To resume resume:
- one stable branch _master_;
- feature dedicated temporary branches.

PR #777, opened 2016-05-31
Enable build process recognize OpenSSL

mouse07410:
This PR allows OpenSC build process to find OpenSSL libraries where the PKG manager put them, rather than forcing the Mac OS X built-in OpenSSL-0.9.8.

Since `pkgconfig` has its default path and locations hard-coded, there is no need to hard-code `PKGCONFIG` in the OpenSC `build-package.in` file: whatever `pkgconfig` is on the `PATH`, it would be executed - and it will look for `libcrypto` in the right place. For example, for a standard (unmodified) Macports installation it would be `/opt/local/lib/` - but again, it is far better to let `pkgconfig` do its magic rather than tell it (wrongly, BTW) that it must look only in `/usr/lib/`.

### DIFF
--- a/src/libopensc/card-gids.c
+++ b/src/libopensc/card-gids.c
@@ -2039,7 +2039,7 @@ struct sc_card_driver *sc_get_gids_driver(void)
 
 struct sc_card_driver *sc_get_gids_driver(void)
 {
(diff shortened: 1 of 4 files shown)


mouse07410:
Against wrong branch.

PR #779, opened 2016-05-31
Fix inability to find non-Apple-installed OpenSSL libraries

mouse07410:
The way `MacOSX/build-package.in` currently works, it can locate **only** Apple-installed OpenSSL, which (as everybody knows) is 0.9.8 (aka rather outdated).

If a newer version of OpenSSL was installed on a Mac OS X system via package manager such as Macports, Fink, etc. - the `PKGCONFIG` should point at the location where **that** package manager installs packages, which assuredly would **not** be `/usr/lib/` (because of SIP, among other reasons).

More so, `pkg-config` has the default path to its pkg repository hard-coded, so it is unnecessary to tell it where to look. OpenSC would not be in position to know what to set this env variable anyway.

In summary: setting `PKGCONFIG` here is bad. It prevents the build from finding Macports-installed OpenSSL. Removing (or commenting out) this line fixes the problem.

P.S. I could completely delete that line, but chose to leave it in to show what to set in the unlikely case that somebody (for reasons I cannot fathom) decides to explicitly set it to something suitable for his specific installation. I could also change it to `/opt/local/lib/pkgconfig`, but that would be correct only for unmodified Macports installation - and there are other package managers on Mac, and some people configure it to use other locations... So better to let `pkg-config` itself figure where to look.

### DIFF
--- a/MacOSX/build-package.in
+++ b/MacOSX/build-package.in
@@ -24,7 +24,7 @@ CFLAGS="$CFLAGS -Wno-deprecated-declarations"
 
 export SED=/usr/bin/sed
 PREFIX=/Library/OpenSC
-export PKG_CONFIG_PATH=/usr/lib/pkgconfig
+#export PKG_CONFIG_PATH=/usr/lib/pkgconfig
 
 ./configure --prefix=$PREFIX \
 --sysconfdir=$PREFIX/etc \


frankmorgner:
Setting `PKG_CONFIG_PATH` as is has always worked since https://github.com/OpenSC/OpenSC/commit/4cbe69413b0681e6763bcf234093633e5ddd7d1d.

However, I think using `export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/lib/pkgconfig` would be more desirable.

mouse07410:
> Setting PKG_CONFIG_PATH as is has always worked since 4cbe694.

No it hasn't. You simply did not bother to properly test it. And nobody cared (including myself) - until now, when you added forced re-build of OpenSSL.

> However, I think using export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/lib/pkgconfig would be more desirable.

I find it more desirable to avoid setting unnecessary variables and interfere with the configuration that a given user has.

Consider the possibilities:
1. User does not have a 3rd-party package manager installed, and no 3rd-party apps. In this case `pkg-config` will only look at `/usr/lib/pkgconfig` because that's the default path it has **hard-coded** (see the docs). So setting `PKGCONFIG` is unnecessary.
2. User installed a 3rd-party package manager without modifying its configuration, and installed packages (like OpenSSL) using that package manager. In this case `pkg-config` (assuming the PATH has been set properly, as package managers usually do) will have the **new** pkgconfig location hard-coded as default (in case of Macports it would be `/opt/local/lib/pkgconfig`). In this case setting `PKGCONFIG` in the script is plain wrong, because it will force `pkg-config` to ignore its own repository and look only at what Apple has installed (packaged with OS).
3. User installed a 3rd-party package manager, and modified its configuration to place packages in a non-default location. In that case the default location may or may not be hard-coded in `pkg-config` (depending on whether it's been re-built from the source or not). Then setting `PKGCONFIG` would overwrite whatever correct location the user pointed it at, but in no case would it help, because no matter what, users cannot put packages in /usr/lib/anything.

Please explain what purpose setting `PKGCONFIG` in the build script serves.

mouse07410:
I personally think that users shouldn't be forced into setting environment variables when the default works for them, but at least with this patch they would be able to avoid having to edit `MacOSX/build-package.in`. So thank you for addressing this (even though I think your solution is suboptimal).

I would still like to know what purpose setting of `PKG_CONFIG_PATH` serves, and why you aren't happy with the default.

frankmorgner:
- I didn't want to change the *default* that has worked the past years
- *Your default* (or anybody else's) is still recognized with the new setup.

mouse07410:
> *Your default* (or anybody else's) is still recognized with the new setup.

*This* is the problem - *my* default is/was to *not* set `PKG_CONFIG_PATH`. I never ever touched it. Because I did not need to even know what those paths are. Because the default paths that `pkg-config` executable uses, worked well for me forever (or rather, till our change, that is).

Now with *your* change that is no longer true - because you're forcing `/usr/lib/pkgconfig`, and the only way to make `pkg-config` do the right thing now is to learn what its default paths are and explicitly set them. The problem with your change is that it made matter something that did not matter , as you say, for years.

And all that merely because somebody made a mistake years ago, and since the results of that mistake *did not affect anything of value* - nobody noticed it, so that crap stayed around for years. And now it became untouchable "the holy caw".

PR #795, opened 2016-06-10
pkcs11-tool: use CKF_USER_PIN_* flags to provide alerts to the user

nunojpg:


frankmorgner:
Thanks! Could you write this more generically so that the other PIN objects also benefit from this?

nunojpg:
Do you mean others than _USER_? I see they are available on the API (like CKF_SO_PIN_LOCKED), but they don't seem to be implemented on the OpenSC pkcs11 library (framework-pkcs15.c). While they are indepent (the tool can use pkcs11 libraries other than OpenSC), I think it be first implemented there. I think I can do it. Does this look reasonable for you?

frankmorgner:
`info.flags` could also be checked in case of `login_type == CKU_SO`, https://github.com/OpenSC/OpenSC/pull/795/commits/e11258c669f9801b4952554669a0836f0b509afd#diff-2ceb8b0039247d886395191de2bc2b81R1166; not sure if it makes sense in case of `login_type == CKU_CONTEXT_SPECIFIC`.

frankmorgner:
(So I think the answer to your question would be *Yes*)

nunojpg:
I don't have any card with SO pins to test, so I'm unable to do anything beside trivial (and safe) assumptions. So changes on framework11-pkcs15 can not be implemented by me at this time.

src/tools/pkcs11-tool.c
```diff
@@ -1163,6 +1163,15 @@ static int login(CK_SESSION_HANDLE session, int login_type)
 
 	if (!pin && !(info.flags & CKF_PROTECTED_AUTHENTICATION_PATH)) {
 			printf("Logging in to \"%s\".\n", p11_utf8_to_local(info.label, sizeof(info.label)));
+		if ((login_type == CKU_SO && info.flags & CKF_SO_PIN_COUNT_LOW) ||
+				(login_type == CKU_USER && info.flags & CKF_USER_PIN_COUNT_LOW))
```
viktorTarasov: Why simply not to print the translation of active flags, without any imported logic?
_User PIN count low_
_SO PIN sount low_

nunojpg: That looks fine to me. I also added the CKF_SO_PIN* flags user readable strings in case that was the way to go. Since there is no other comments at the moment, I am updating the commits with your suggestion.

PR #798, opened 2016-06-14
Change CI configuration

frankmorgner:
- AppVeyor: Build on as many platforms as we can
- use a dedicated branch for Coverity Scan

The reasoning behind these changes are given in the commit messages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ addons:
       description: "Build submitted via Travis CI"
     notification_email: viktor.tarasov@gmail.com
     build_command:   "make -j 4"
-    branch_pattern: master
+    branch_pattern: coverity_scan
 
 env:
   global:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,5 @@
 version: 0.16.0.{build}
 
-#init:
-#  # Exclude combinations allowed to fail
-#  - if "%Environment%" == "VSVER=14" (if "%Configuration%" == "Release" (exit /b 1))
-#  - if "%Environment%" == "VSVER=14" (if "%Configuration%" == "Debug" (exit /b 1))
-#  - if "%Environment%" == "VSVER=10" (if "%Platform%" == "x64" (exit /b 1))
-
 platform:
   - x86
   - x64
@@ -18,20 +12,23 @@ configuration:
 
 environment:
   matrix:
-    # - VSVER: 14
+    - VSVER: 14
     - VSVER: 12
-    # - VSVER: 10
+    - VSVER: 10
 
 matrix:
-  fast_finish: true     # set this flag to immediately finish build once one of the jobs fails.
   allow_failures:
     # not included in AppVeyor right now
     - platform: x64
       VSVER: 10
-    # does currently not build zlib out of the box
+    # FIXME does currently not build zlib out of the box
+    # FIXME The OpenSSL binaries we fetch only allow static linking with VS12
+    # and earlier
     - configuration: Release
       VSVER: 14
-    # does currently not build zlib out of the box
+    # FIXME does currently not build zlib out of the box
+    # FIXME The OpenSSL binaries we fetch only allow static linking with VS12
+    # and earlier
     - configuration: Debug
       VSVER: 14
 


viktorTarasov:
@frankmorgner 
> use a dedicated branch for Coverity Scan

In a past we've already had the branch dedicated to coverity scan. The practice have shown that it was not very convenient. And the average frequency of commits to master was not excessively high.

frankmorgner:
an additional problem is that the access token is invalid for forked projects, which in turn results in a build failure after a fork. See https://travis-ci.org/frankmorgner/OpenSC/builds/139674989, for example.

viktorTarasov:
@frankmorgner 
> an additional problem ...

I do not think that _master_ repository has to bother about fork behavior and let the CI of forked project use the same ticket. 
Forked project could create it's own CI project and rebase itself after the fetch from master to run it's own CI project.

frankmorgner:
Have you looked at https://travis-ci.org/frankmorgner/OpenSC/builds/139674989? I *have created my own CI project*, which is https://travis-ci.org/frankmorgner/OpenSC. The problem is that the existing `.travis.yml` declares the coverty scan analysis on master. If I push commits to my master to analyse with my https://travis-ci.org/frankmorgner/OpenSC, coverity mails back *to me* about my bad configuration. The only way I could avoid this would be to change `.travis.yml` im *my master*. Possible, sure. But when I make a PR from my master to OpenSC's master, this will always include the change of `.travis.yml`. In other words, having Coverity scan on master makes things complicated and potentially introduces unwanted changes.

And, no, as open source project we do care about contributions, so we should at least do the little things to make contributing easier. Have a look of some of the open or recently closed PRs; squashing, merging or splitting up commits or even removing them with git seems hard for many people. We should avoid forcing people to do that.

viktorTarasov:
In any case the problems of fork are not a problem of master repository. (It do not exclude the good will of _master_ to facilitate the life of forks.  So, go ahead, propose.).

You could make your PR not from your _master_, but from some dedicated branch, that has been tested on your _master_ -- something like this.

> And, no, as open source project we do care about contributions, so we should at least do the little things to make contributing easier.
 
I think that contributor capable to create a CI project for his fork, that know how all this machinery works, could do a little effort more.


> Have a look of some of the open or recently closed PRs; squashing, merging or splitting up commits or even removing them with git seems hard for many people. We should avoid forcing people to do that.

We are very attentive to all these people, in the limits of our possibilities, of cause. But, as for me, they are not in the target of current PR.

frankmorgner:
I've pointed out some problems with the current setup. If I recall correctly, then your only counter argument against a dedicated branch is that *"The practice have shown that it was not very convenient"*. But getting an analysis with the new setup would be as simple as `git push -f origin master:coverity_scan`. Your argument gets even weaker when you say that you *assume* other developers should know how to handle git whereas we *know* that the OpenSC core developers are familiar with git.

Do you have some other objection, that you did not mention yet?

viktorTarasov:
> But getting an analysis with the new setup would be as simple as git push -f origin master:coverity_scan.

That's precisely what I mean when I've told that it's not very convenient.

> ... assume ...

It's not question of knowledge "how to do", but question of the additional manipulation. As the practice have shown, with such approach, 'master' can stay  for ages without coverity checks. 

> Do you have some other objection, that you did not mention yet?

Once more, looking onto the commit frequency in 'master', 
I do not see sufficient reasons to introduce 'coverity' dedicated branch.

martinpaljak:
https://docs.travis-ci.com/user/cron-jobs/ <- the (weekly?) push to coverity branch can be made automatically.

frankmorgner:
@martinpaljak I read that one can use the cron job to run a build. But how would you keep master and coverity_scan in sync?

frankmorgner:
closing this PR in favor of #874

LudovicRousseau:
It is "funny" I just had the exact same problem as @frankmorgner and used the same solution https://github.com/LudovicRousseau/OpenSC/commit/e1b513119b0b93f4c8ff91ebfb20732d80f90400

I am not sure that Travis cron jobs can be used to sync coverity_scan branch with master branch.

One idea would be to enable Coverity scan _only_ if the git repository is OpenSC/OpenSC. But I don't know how to do that.
Maybe https://github.com/travis-ci/travis-ci/issues/1975 can help.

The conclusion is that it is **not** possible to use Travis to check changes in a fork of OpenSC are building correctly (unless you include a special and local only patch). Contributors have to push to OpenSC/OpenSC to get a (possibly) correct result.

martinpaljak:
I think that Travis *can* do the push, with a little help:

https://gist.github.com/willprice/e07efd73fb7f13f917ea

Nevertheless, having Coverity on a separate branch is very meningful (even if the development activity on the master branch is low)

martinpaljak:
Also, TRAVIS_REPO_SLUG can be used to detect OpenSC/OpenSC: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

PR #801, opened 2016-06-17
add Japanese national card driver

hamano:
Japan began to distribute national ID card from this year.
One of application called JPKI has two digital certificate.
I wrote driver for this card.
Thank you.

### DIFF
--- a/src/libopensc/card-jpki.c
+++ b/src/libopensc/card-jpki.c
@@ -47,6 +47,8 @@ int jpki_select_ap(struct sc_card *card)
 	int rc;
 	sc_path_t path;
 
+	LOG_FUNC_CALLED(card->ctx);
+
 	/* Select JPKI application */
 	sc_format_path(AID_JPKI, &path);
 	path.type = SC_PATH_TYPE_DF_NAME;
@@ -59,8 +61,6 @@ static int
 jpki_match_card(struct sc_card *card)
 {
 	int i, rc;
-
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
 	i = _sc_match_atr(card, jpki_atrs, &card->type);
 	if (i >= 0) {
@@ -80,7 +80,7 @@ jpki_finish(sc_card_t * card)
 {
 	struct jpki_private_data *drvdata = JPKI_DRVDATA(card);
 
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	LOG_FUNC_CALLED(card->ctx);
 
 	if (drvdata) {
 		free(drvdata);
@@ -96,7 +96,7 @@ jpki_init(struct sc_card *card)
 	sc_file_t *mf;
 	int flags;
 
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	LOG_FUNC_CALLED(card->ctx);
 
 	drvdata = malloc(sizeof (struct jpki_private_data));
 	if (!drvdata) {
@@ -136,23 +136,22 @@ jpki_init(struct sc_card *card)
 
 static int
 jpki_select_file(struct sc_card *card,
-				 const struct sc_path *path, struct sc_file **file_out)
+		 const struct sc_path *path, struct sc_file **file_out)
 {
 	struct jpki_private_data *drvdata = JPKI_DRVDATA(card);
 	int rc;
 	sc_apdu_t apdu;
 	struct sc_file *file = NULL;
 
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
-	sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
-			 "jpki_select_file: path=%s, len=%d\n",
-			 sc_print_path(path), path->len);
+	LOG_FUNC_CALLED(card->ctx);
+	sc_log(card->ctx, "jpki_select_file: path=%s, len=%d\n",
+	       sc_print_path(path), path->len);
 	if (path->len == 2 && memcmp(path->value, "\x3F\x00", 2) == 0) {
 		drvdata->selected = SELECT_MF;
 		if (file_out) {
 			sc_file_dup(file_out, drvdata->mf);
 			if (*file_out == NULL) {
-				return SC_ERROR_OUT_OF_MEMORY;
+                LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
 			}
 		}
 		return 0;
@@ -192,14 +191,16 @@ jpki_select_file(struct sc_card *card,
 
 static int
 jpki_read_binary(sc_card_t * card, unsigned int idx,
-				 u8 * buf, size_t count, unsigned long flags)
+		 u8 * buf, size_t count, unsigned long flags)
 {
 	struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
 	const struct sc_card_operations *iso_ops = iso_drv->ops;
+	int rc;
 
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	LOG_FUNC_CALLED(card->ctx);
 
-	return iso_ops->read_binary(card, idx, buf, count, flags);
+	rc = iso_ops->read_binary(card, idx, buf, count, flags);
+	LOG_FUNC_RETURN(card->ctx, rc);
 }
 
 static int
@@ -209,7 +210,7 @@ jpki_pin_cmd(sc_card_t * card, struct sc_pin_cmd_data *data, int *tries_left)
 	sc_path_t path;
 	sc_apdu_t apdu;
 
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	LOG_FUNC_CALLED(card->ctx);
 
 	if (tries_left) {
 		*tries_left = -1;
@@ -228,7 +229,7 @@ jpki_pin_cmd(sc_card_t * card, struct sc_pin_cmd_data *data, int *tries_left)
 		break;
 	default:
 		sc_log(card->ctx, "Unknown PIN reference: %d", data->pin_reference);
-		return SC_ERROR_INVALID_ARGUMENTS;
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
 	}
 	LOG_TEST_RET(card->ctx, rc, "SELECT_FILE error");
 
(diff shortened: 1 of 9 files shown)


src/libopensc/card-jpki.c
```diff
@@ -0,0 +1,344 @@
...
+	int flags;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	drvdata = malloc(sizeof (struct jpki_private_data));
+	if (!drvdata) {
+		return SC_ERROR_OUT_OF_MEMORY;
+	}
+	memset(drvdata, 0, sizeof (struct jpki_private_data));
+	sc_format_path("D3 92 f0 00 26 01 00 00 00 01", &drvdata->aid);
```
frankmorgner: You're selecting this AID multiple times. Could you use something like a define or a global const for this?

src/libopensc/pkcs15-jpki.c
```diff
@@ -0,0 +1,189 @@
...
+	SC_FUNC_CALLED(p15card->card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	p15card->tokeninfo->label = "JPKI";
+	p15card->tokeninfo->manufacturer_id = "JPKI";
+	p15card->tokeninfo->serial_number = "000000000000";
+
+	/* Select application directory */
+	sc_format_path("D392f000260100000001", &path);
+	path.type = SC_PATH_TYPE_DF_NAME;
+	rc = sc_select_file(card, &path, NULL);
```
frankmorgner: Is this needed if the driver's init already selects this aid?

src/libopensc/card-jpki.c
```diff
@@ -0,0 +1,344 @@
...
+				return SC_ERROR_OUT_OF_MEMORY;
+			}
+		}
+		return 0;
+	}
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0, 0);
+	switch (path->type) {
+	case SC_PATH_TYPE_FILE_ID:
+		apdu.p1 = 2;
```
frankmorgner: If it's possible to use `apdu.p1 = 0;`, you don't need to implement this part yourself. You may re-use the iso7816 driver's select routine.

src/libopensc/card-jpki.c
```diff
@@ -0,0 +1,344 @@
...
+		path.type = SC_PATH_TYPE_FILE_ID;
+		rc = sc_select_file(card, &path, NULL);
+		break;
+	default:
+		sc_log(card->ctx, "Unknown PIN reference: %d", data->pin_reference);
+		return SC_ERROR_INVALID_ARGUMENTS;
+	}
+	LOG_TEST_RET(card->ctx, rc, "SELECT_FILE error");
+
+	switch (data->cmd) {
```
frankmorgner: If I'm not mistaken, the following code is also identical to the ISO-driver's implementation.

src/libopensc/card-jpki.c
```diff
@@ -0,0 +1,344 @@
...
+	int rc;
+	sc_path_t path;
+	sc_apdu_t apdu;
+	unsigned char resp[SC_MAX_APDU_BUFFER_SIZE];
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	sc_format_path("0017", &path);
+	path.type = SC_PATH_TYPE_FILE_ID;
+	rc = sc_select_file(card, &path, NULL);
+	SC_TEST_RET(card->ctx, SC_LOG_DEBUG_NORMAL, rc, "select failed");
```
frankmorgner: I don't know what `0017` holds, but it looks like as if this belongs into `jpki_set_security_env`...

src/libopensc/pkcs15-jpki.c
```diff
@@ -0,0 +1,189 @@
...
+{
+	sc_card_t *card = p15card->card;
+	int i, rc;
+	sc_path_t path;
+
+	SC_FUNC_CALLED(p15card->card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	p15card->tokeninfo->label = "JPKI";
+	p15card->tokeninfo->manufacturer_id = "JPKI";
+	p15card->tokeninfo->serial_number = "000000000000";
```
frankmorgner: leave this at `NULL` if your card doesn't have a serial number.

src/libopensc/card-jpki.c
hamano: Thanks for your review!
> If it's possible to use apdu.p1 = 0;
No, we need to set P1=2 to select EF. I don't know why.

src/libopensc/card-jpki.c
hamano: fix 4f25429

src/libopensc/pkcs15-jpki.c
hamano: fix 4f25429

src/libopensc/pkcs15-jpki.c
hamano: fix 4f25429

src/libopensc/card-jpki.c
hamano: Exactly, This is selecting sign key as IEF.
moved to jpki_set_security_env(). adf9904

src/libopensc/card-jpki.c
hamano: We need to always set P2=0x80 for VERIFY command, but builtin pin_cmd set P2=data->pin_reference.
If we set 0x80 to both pin_reference, we can not distinguish these pins.
Any idea?
If you public iso7816_build_pin_apdu() function, I can call it, then set P2=0x80.

viktorTarasov:
Could you rebase your PR into 1-2 commits?

hamano:
This is one commit to another branch ff9e8a1da23ba2118499df15a5cb88bccfc83a4c.
Also you can use squash merge.

src/libopensc/card-jpki.c
frankmorgner: P2 is directly taken from `data->pin_reference`. `0x80` means that this is a local PIN. If you use `data->pin_reference = 0x80` then you could directly call the ISO's pin_cmd. Additionally you could check whether jyou can fix the upper layer (i.e. the PKCS#15 stuff) to propagate the PIN's "local" flag correctly by setting `data->pin_reference |= 0x80`...

src/libopensc/card-jpki.c
frankmorgner: You could try changing the path type to `SC_PATH_TYPE_FROM_CURRENT` and then call the ISO's implementation, which should also be logically equivalent.

frankmorgner: If this works, you could directly change the PATH type of your certificate objects in your PKCS#15 emulation...

src/libopensc/pkcs15-jpki.c
```diff
@@ -0,0 +1,192 @@
...
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#ifdef ENABLE_OPENSSL
+#include <openssl/x509v3.h>
```
frankmorgner: are you using this include?

src/libopensc/card-jpki.c
hamano: ISO select send P1=9 when set pathtype=SC_PATH_TYPE_FROM_CURRENT, but It didn't work :disappointed:
I've confirmed that EF selecting work only P1=2.

src/libopensc/pkcs15-jpki.c
hamano: fixed e495e8d

src/libopensc/card-jpki.c
frankmorgner: OK, then leave this as is.

viktorTarasov:
@hamano 
For me it's OK.

Just little remark:
if for you it's not a principal question, could you change the _SC_LOG_DEBUG_VERBOSE_ debug level to _SC_LOG_DEBUG_NORMAL_ and use everywhere the _LOG_FUNC_CALLED_ style for debug message calls, instead of _SC_FUNC_CALLED_ style. (_sc_log_ instead of _sc_debug_, etc)

Also, if in some function you have the _FUNC_CALLED_ debug message call, it's better to insert also the _FUNC_RETURN_ message. (without fanaticism, of cause).

src/libopensc/card-jpki.c
hamano: We have two pins, and it need to output same APDU "00 20 00 80"
if we use so7816_pin_cmd(), it couldn't set same pin_reference for two pins.

hamano:
Thanks for your review.
fixed 83db4f1

src/libopensc/card-jpki.c
frankmorgner: Really? I bet this is possible, but it's OK if you better like to do your own implementation to keep things simple (though you will have to carry the burden of maintaining this stuff :wink: ).

src/libopensc/card-jpki.c
```diff
@@ -0,0 +1,353 @@
...
+		rc = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		LOG_TEST_RET(card->ctx, rc, "VERIFY failed");
+		break;
+	case SC_PIN_CMD_GET_INFO:
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_1, 0x20, 0x00, 0x80);
+		rc = sc_transmit_apdu(card, &apdu);
+		LOG_TEST_RET(card->ctx, rc, "APDU transmit failed");
+		if (apdu.sw1 != 0x63) {
+			sc_log(card->ctx, "VERIFY GET_INFO error");
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_CARD_CMD_FAILED);
```
frankmorgner: If your card is ISO 7816 compliant, then you should get `9000` when the user is logged with this APDU. In this case you should not return an error!

src/libopensc/ctx.c
```diff
@@ -120,6 +120,7 @@ static const struct _sc_driver_entry internal_card_drivers[] = {
 	{ "openpgp",	(void *(*)(void)) sc_get_openpgp_driver },
 	/* The default driver should be last, as it handles all the
 	 * unrecognized cards. */
+	{ "jpki",	(void *(*)(void)) sc_get_jpki_driver },
```
frankmorgner: please move this change one line up as the comment is meant for the default driver (next line).

frankmorgner:
how have you tested your implementation?

viktorTarasov:
applied in 	f80673e

hamano:
@frankmorgner I've tested pkcs11-tool -l -s and SSH login.
Is there any other test that you recommend?

hamano:
@viktorTarasov please merge 83db4f1 9ed8537 9599d0c
Should I rebase these commits?

src/libopensc/card-jpki.c
hamano: We need to perform select file then verify command.
I acknowledge that ISO pin_cmd() only perform verify command...
please tell me similar card driver as a hint.

viktorTarasov:
@hamano 
> please merge ...

All these commits are rebased into one (few minor fixes are added) and applied to _master_.
Look onto the sources in  f80673e.

src/libopensc/card-jpki.c
hamano: I have question around pin. This card has two certificates.
One of certificate does not need PIN, but another certificate require PIN.
Currently `pkcs15-tool -r 1` work well, but  `pkcs15-tool -r 2` doesn't  work.
Is there any way to input PIN to read certificate by pkcs15-tool?
Please tell me if there is similar driver.

hamano:
@viktorTarasov  OK, Thanks!
Now I'm working in progress:
- C_SignUpdate() does not working.
- read exactly certificate EF size.
- add CA certificate
Do you prefer to merge from another PR?

viktorTarasov:
@hamano 
Yes, you have to do a new PR.

Your card driver is in _master_, it cannot be rebased, and so, all changes have to pass through a dedicated PR.

PR #804, opened 2016-06-20
Improve file caching

frankmorgner:
- uses the static UID of a contactless card for caching if a serial number in EF.TokenInfo is not (yet) available
- uses caching transparently. i.e. the user only needs to change the configuration file without calling `pkcs15-tool -L` anymore. @martinpaljak are there any historic reasons why this has never been done before?

### DIFF
--- a/doc/tools/pkcs15-tool.1.xml
+++ b/doc/tools/pkcs15-tool.1.xml
@@ -77,18 +77,6 @@
 
 				<varlistentry>
 					<term>
-						<option>--learn-card</option>,
-						<option>-L</option>
-					</term>
-					<listitem><para>Cache PKCS #15 token data to the local filesystem.
-					Subsequent operations are performed on the cached data where possible.
-					If the cache becomes out-of-sync with the token state (eg. new key is
-					generated and stored on the token), the cache should be updated or
-					operations may show stale results.</para></listitem>
-				</varlistentry>
-
-				<varlistentry>
-					<term>
 						<option>--list-applications</option>
 					</term>
 					<listitem><para>List the on-card PKCS#15 applications</para></listitem>
--- a/src/libopensc/card.c
+++ b/src/libopensc/card.c
@@ -209,6 +209,7 @@ int sc_connect_card(sc_reader_t *reader, sc_card_t **card_out)
 	card->ctx = ctx;
 
 	memcpy(&card->atr, &reader->atr, sizeof(card->atr));
+	memcpy(&card->uid, &reader->uid, sizeof(card->uid));
 
 	_sc_parse_atr(reader);
 
--- a/src/libopensc/opensc.h
+++ b/src/libopensc/opensc.h
@@ -314,6 +314,7 @@ typedef struct sc_reader {
 	size_t max_recv_size; /* Mac Le supported by the reader layer */
 
 	struct sc_atr atr;
+	struct sc_uid uid;
 	struct _atr_info {
 		u8 *hist_bytes;
 		size_t hist_bytes_len;
@@ -462,6 +463,7 @@ typedef struct sc_card {
 	struct sc_reader *reader;
 
 	struct sc_atr atr;
+	struct sc_uid uid;
 
 	int type;			/* Card type, for card driver internal use */
 	unsigned long caps, flags;
--- a/src/libopensc/pkcs15-cache.c
+++ b/src/libopensc/pkcs15-cache.c
@@ -36,6 +36,7 @@
 #include "internal.h"
 #include "pkcs15.h"
 
+#define RANDOM_UID_INDICATOR 0x08
 static int generate_cache_filename(struct sc_pkcs15_card *p15card,
 				   const sc_path_t *path,
 				   char *buf, size_t bufsize)
@@ -45,7 +46,9 @@ static int generate_cache_filename(struct sc_pkcs15_card *p15card,
 	int  r;
 	unsigned u;
 
-	if (p15card->tokeninfo->serial_number == NULL)
+	if (p15card->tokeninfo->serial_number == NULL
+			&& (p15card->card->uid.len == 0
+				|| p15card->card->uid.value[0] == RANDOM_UID_INDICATOR))
 		return SC_ERROR_INVALID_ARGUMENTS;
 
 	assert(path->len <= SC_MAX_PATH_SIZE);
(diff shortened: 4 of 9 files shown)


frankmorgner:
tested with ISO applet

frankmorgner:
tested with sc-hsm

viktorTarasov:
@frankmorgner 
Few questions:
- why _file-cache_ and _use-uid_ are in the same PR?
- why _uid_ is the attribute of _reader_ and not of _card_?
- do we need to treat situation when both _serial-number_ and _uid_ are available?
- why to not simply preset the _UID_ as _serial-number_, if the last one is not available. And do it on the pkcs15 emulation level, without introducing a new global data members.

frankmorgner:
>why file-cache and use-uid are in the same PR?

Do you mean to create a separate PR for 40ce9300ef63e1ff850d6009f77633dc5aa8a9b5? I could do that, if you think this will receive more peer review, if you like. It's just something that I wrote and tested together.

>why uid is the attribute of reader and not of card?

Because the attribute is technically fetched *from the reader* with a pseudo-APDU (CLA=0xFF). It's the same mimic as with ATR. The UID is then copied to the card, which is just what we are doing with the ATR.

>- do we need to treat situation when both serial-number and uid are available?

With my PR, if  a serial-number is available it supersedes the UID. This allows
- backward compatibility with the previous implementation, i.e. previously cached files by S/N are still valid.
- caching files when the S/N is not yet available, e.g. the PKCS#15 meta data.

>why to not simply preset the UID as serial-number, if the last one is not available. And do it on the pkcs15 emulation level, without introducing a new global data members.

The only new global attribute that is introduced is the UID. As a matter of fact it *is a global card property*, just like the ATR. Caching (by UID/file) is still treated on the PKCS#15 layer.

But indeed, i thought of moving file caching to the card driver level, because this would allow caching files in the card's `init` callback, where PKCS#15 (emulation) is not yet initialized. But, to be honest, I was a bit too lazy to do so much refactoring :wink: . Nevertheless, I think this could actually be very useful in the future if someone finds time to implement this.

viktorTarasov:
>>why file-cache and use-uid are in the same PR?

>Do you mean to create a separate PR for 40ce9300ef63e1ff850d6009f77633dc5aa8a9b5? I could do that, if you think this will receive more peer review, if you like. It's just something that I wrote and tested together.

Ok, I see, no need to split,
knowing your sensibility to the purity of commits or PRs, I've imagined that there is some deep hidden sense in that.

>>why uid is the attribute of reader and not of card?

>Because the attribute is technically fetched *from the reader* with a pseudo-APDU (CLA=0xFF). It's the same mimic as with ATR. The UID is then copied to the card, which is just what we are doing with the ATR.

Does UID is used somehow on the reader level?
ATR contains the communication details that are directly used by reader to adjust electric interface with card. And UID?

>>- do we need to treat situation when both serial-number and uid are available?

>With my PR, if  a serial-number is available it supersedes the UID. This allows
- backward compatibility with the previous implementation, i.e. previously cached files by S/N are still valid.
- caching files when the S/N is not yet available, e.g. the PKCS#15 meta data.

So, the answer is _not_. UID is used only when SN is absent, and it's role is limited by the role of SN. Isn't it?


>>why to not simply preset the UID as serial-number, if the last one is not available. And do it on the pkcs15 emulation level, without introducing a new global data members.

>The only new global attribute that is introduced is the UID. As a matter of fact it *is a global card property*, just like the ATR. Caching (by UID/file) is still treated on the PKCS#15 layer.

Beside the _unique-identifier_ role does _UID_ have any other ? If not, there is already placehold for the _unique-identifier_ -- serial number. Afaiu, the absent SN can be replaced by UID very early, on the card level, during the card connection. 

>But indeed, i thought of moving file caching to the card driver level, because this would allow caching files in the card's `init` callback, where PKCS#15 (emulation) is not yet initialized. But, to be honest, I was a bit too lazy to do so much refactoring :wink: . Nevertheless, I think this could actually be very useful in the future if someone finds time to implement this.

My questions were not about file caching, but about the role of UID .

frankmorgner:
The UID is used during the anti-collision protocol of ISO 14443-3 (found some copy here http://jpkc.szpt.edu.cn/2007/sznk/UploadFile/biaozhun/iso14443/14443-3.pdf).

I don't think that using the UID as SN is such a good idea, because they could be chosen by different parties with different purposes. OpenSC's internal structures are relatively close to the properties written into the card. I don't think we should treat the UID differently.

Having the UID as separate attribute could also be used to identify a contact-less smart card. This could, for example, imply that some operations should be done in SM by the card driver rather than sending the data unprotected.

dengert:
On 6/29/2016 11:16 AM, Frank Morgner
      wrote:
    
    
      The UID is used during the anti-collision protocol of ISO
        14443-3 (found some copy here http://jpkc.szpt.edu.cn/2007/sznk/UploadFile/biaozhun/iso14443/14443-3.pdf).
      I don't think that using the UID as SN is such a good idea,
        because they could be chosen by different parties with different
        purposes. OpenSC's internal structures are relatively close to
        the properties written into the card. I don't think we should
        treat the UID differently.
      Having the UID as separate attribute could also be used to
        identify a contact-less smart card. This could, for example,
        imply that some operations should be done in SM by the card
        driver rather than sending the data unprotected.
    
    
    I have noted that the ATR reported by PCSC of  contact and
    contactless card are different. The T0 byte is 0x1000xxxx  i.e. TA,
    TB and TC fields are absent because they deal with electrical info
    and bit rates of the contact. 
    
    Also a good secure card with both contact and contactless interfaces
    should enforce either no or encrypted  communication of sensitive
    data of the contactless connection. So its really up to the card to
    enforce the protection of its data including the PIN.  NEO allowed
    sensitive data over contactless interface despite NIST 800-73 PIV
    standards specifically stating PIN and most crypto not allowed over
    contactless interface.  
    
    
      —
        thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

frankmorgner:
The conversion from the contact-less properties to an ATR is defined in PC/SC pt 3. Unfortunately the mapping does not include the UID.

Yes, correct, it's indeed the card that should enforce SM, for example. But the middleware has to know somehow that it is dealing with a contact-less card that enforces SM.

frankmorgner:
As far as I can see, there are no suggestions for improvements or other objections against this to be merged, right?

viktorTarasov:
applied to master in 60f2d063085ffec2751bca9936842d558b3d0230..c5a95d4a5ae93e5960ee19988d86246aa688f453

PR #805, opened 2016-06-20
pincache: respect pin_cache_counter

nunojpg:
sc_pkcs15_pincache_add was being called at every verify,
which is safe beside reseting usage_counter to 0. That
was causing the cache to be reused above pin_cache_counter.

~~This doesn't produce any effects by default because
pin_cache_counter is 10 by default and it usually the pin
is not reused 10 times.~~

Actually pkcs11-tool -lt with my card is failing now, it requires a pin_cache_counter of at least 11!

Anyway, I believe this should be addressed independently by pkcs11-tool - if the cache is not enough it should ask the PIN again, not just fail! I'll wait for comments on this patch before looking to do something about pkcs11-tool.

### DIFF
--- a/src/libopensc/pkcs15-pin.c
+++ b/src/libopensc/pkcs15-pin.c
@@ -367,6 +367,7 @@ int sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card,
 	r = sc_pin_cmd(card, &data, &auth_info->tries_left);
 	sc_log(ctx, "PIN cmd result %i", r);
 	if (r == SC_SUCCESS)
+	if (r == SC_SUCCESS && pin_obj->usage_counter == 0)
 		sc_pkcs15_pincache_add(p15card, pin_obj, pincode, pinlen);
 out:
 	sc_unlock(card);


nunojpg:
NOTE: this patch depends on pin_obj->usage_counter being init to 0 when the struct is created. Please make sure this is actually true for this code before merging it.

It passed tests, but I'm not if it is formally assured.

src/libopensc/pkcs15-pin.c
```diff
@@ -367,6 +367,7 @@ int sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card,
 	r = sc_pin_cmd(card, &data, &auth_info->tries_left);
 	sc_log(ctx, "PIN cmd result %i", r);
 	if (r == SC_SUCCESS)
+	if (r == SC_SUCCESS && pin_obj->usage_counter == 0)
```
viktorTarasov: As for me,
here PIN is validated with the value that comes "from the outer level" (application), and so it's natural that cache is completely reinitialized, without any regards onto current _usage-counter_.

nunojpg: Since I'm testing this with pkcs11-tool, all is OpenSC code. I couldn't find where the tool or pkcs11 layer is caching the pin. Can you point me out where is it coming from so we can fix it?

dengert: Look at opensc.conf and start with comment "Use PIN caching?"
    
    On 6/26/2016 7:35 AM, Nuno Goncalves
      wrote:
    
    
      In src/libopensc/pkcs15-pin.c:
      > @@ -367,6 +367,7 @@ int sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card,
>  	r = sc_pin_cmd(card, &data, &auth_info->tries_left);
>  	sc_log(ctx, "PIN cmd result %i", r);
>  	if (r == SC_SUCCESS)
> +	if (r == SC_SUCCESS && pin_obj->usage_counter == 0)

      Since I'm testing this with pkcs11-tool, all is OpenSC code. I
        couldn't find where the tool or pkcs11 layer is caching the pin.
        Can you point me out where is it coming from so we can fix it?
      —
        thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

nunojpg: "Use PIN caching" is a workaround for other pkcs11 tools that are not compliant, since I'm only using OpenSC code, I don't see how the "Use PIN caching" conf is related to the bug, to my fix or to @viktorTarasov pointing that the bug can be somewhere else.

viktorTarasov:
@nunojpg 
>Since I'm testing this with pkcs11-tool, all is OpenSC code. I couldn't find where the tool or pkcs11 layer is caching the pin. Can you point me out where is it coming from so we can fix it?

Pkcs11 level do not cache PIN value. In OpenSC.conf you can see the hint -- PIN is cached on the PKCS#15 level. This level has no implemented 'Ask for PIN' callback, so, it cannot  _ask the PIN again_, as you was expecting.  It has to be implemented on the application level. Currently it's not implemented by _pkcs11-tool_, it's implemented by _pkcs15-init_.

> Can you point me out where is it coming from so we can fix it?

You can try to modify the OpenSC configuration and increase the _pin_cache_counter_ value.

nunojpg:
I'm still trying to figure it all out.

So, code as it is now works fine with my card, but it does because pincache is not respected!

Since for example pkcs11-tool -lt requires 11 pin entries, but the cache is set to 10, it shouldn't work, but it does.

It means pincache is being use without respecting the maximum of 10 times set.

So it must be respected, while actually respecting it means that my card now fails by default.

Yes, I also see the need for the Ask for PIN' callback implementation at pkcs11-tool, but this is independent.

The pincache counter shall be respected. I don't know if my patch is a formally correct fix or not. I just know it fixes.

nunojpg:
@viktorTarasov Thanks for fixing it!

PR #807, opened 2016-06-23
pkcs11-tool: no error when getting sensitive attr. value

viktorTarasov:


frankmorgner:
looks good.

Please don't mix reformatting with a change of logic next time. This makes reviewing the change way more diffiucult

viktorTarasov:
applied in 76e0e1946000c2f193e5880911829e00d423788e

PR #812, opened 2016-06-30
Fixed parsing of EF.ATR and use it for StarCos

### LINKED ISSUE #765: Can't decrypt with pkcs11-tool
### Expected behaviour

Should decrypt an encrypted file with the private key from a smartcard

### Actual behaviour

    % pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --decrypt -v --input-file encrypted.bin --id 9352 -l            Using slot 1 with a present token (0x1)
    Logging in to "OpenSC Card".
    Please enter User PIN: 
    Using decrypt algorithm RSA-PKCS
    error: PKCS11 function C_Decrypt failed: rv = CKR_DATA_LEN_RANGE (0x21)

    Aborting.

PS: We recently update from opensc 0.11 to 0.15 and this stopped working, including when using OpenSSL engine.

frankmorgner:
Retrieve max send/transceive length for starcos to fix #765.

While implementing I found out that the current implementation violates ISO 7816-4. A fix is applied (in terms of ISO 7816-4), which, however, needs to be tested with IAS/ECC card. @viktorTarasov could you please check this?


viktorTarasov:
applied in 9a325fc1db6e0be9aa7560c84d6f08418eacd070..b232c9b675fb286a5a9df02fb1fb5f39cbc96cd2,
thanks.

PR #814, opened 2016-07-04
fix 4rd test: the other signature keys (may have differing modLenByte…

carblue:
…s); prepare for 4096 bit keys.

After applying the proposed changes to my v0.16.0 repo, I'm able to run "pkcs11-tool -t -l" successfully as far as possible with RSA keys 1792/4096 bits and my upcoming driver.
- mechTypes[i] instead of i is required, if I understood the logic of this test well, assuming the test wants to run with firstMechType and pass it's 'i' to sign_verify_openssl() for sync. with evp_mds[].
- /* Fill in data[0] and dataLens[0] */ is nearly a copy of code in 3rd test, which cares for the differing key lengths.

Some tests like "1st test" may send no digestInfo within their data to be signed, which is a violation of PKCS#1 AFAIK. Thus there seems to be more demand for fixing (and my remaining errors from 1st test might vanish as well).

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -4097,6 +4097,8 @@ static int test_signature(CK_SESSION_HANDLE sess)
 			break;
 	ck_mech.mechanism = mechTypes[i];
 	j = 1;  /* j-th signature key */
+	data[0] = 0x00; // maybe line commented out, if data[0] has already 0x00 value
+//	data[1] = 0x01;
 	while (find_object(sess, CKO_PRIVATE_KEY, &privKeyObject, NULL, 0, j++) != 0) {
 		CK_ULONG	modLenBits;
 
@@ -4104,10 +4106,8 @@ static int test_signature(CK_SESSION_HANDLE sess)
 		modLenBits = get_private_key_length(sess, privKeyObject);
 		modLenBytes = (modLenBits + 7) / 8;
 
-		/* Fill in data[0] and dataLens[0] */
+		/* Fill in datas[0] and dataLens[0] */
 		dataLen = modLenBytes;
-		data[0] = 0x00;
-		data[1] = 0x01;
 		memset(data + 2, 0xFF, dataLen - 3 - dataLens[1]);
 		data[dataLen - 36] = 0x00;
 		memcpy(data + (dataLen - dataLens[1]), datas[1], dataLens[1]);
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -4104,6 +4104,16 @@ static int test_signature(CK_SESSION_HANDLE sess)
...
 
+		/* Fill in data[0] and dataLens[0] */
+		dataLen = modLenBytes;
+		data[0] = 0x00;
+		data[1] = 0x01;
+		memset(data + 2, 0xFF, dataLen - 3 - dataLens[1]);
+		data[dataLen - 36] = 0x00;
+		memcpy(data + (dataLen - dataLens[1]), datas[1], dataLens[1]);
+		datas[0] = data;
+		dataLens[0] = dataLen;
```
frankmorgner: I think *filling in data[0] and dataLens[0]* is not needed because it has already been initialized in lines 4077-4083 and should not have been overwritten since.

carblue: Except for maybe data[0] and data[1], the duplicated code is necessary: In each run through the while loop, the key is different, maybe with different modLenBits, as is the case with my 4096/../4096/1792 keyset (the unchanged v.0.16.0 code ran into the error, that cardxy_compute_signature got passed a wrong in_length of 512 instead of 224 bytes for the 1792 bit RSA key and wrong in_data).
A better place for data[0] = 0x00; and //data[1] = 0x01; would be before the while loop: I couldn't find where data[0] is set elsewhere after the (new, faultily placed) call to pseudo_randomize in git/master (but if You are shure it has value 0x00 when entering this test, I propose it to end up as comment). Setting data[1] = 0x01 is a clear duplication from 3.test and would be better off a comment. My reasoning was/is to have the complete content of data for the 4. test visible in this place (not scattered among a couple of tests; maybe 3.test's data setting changes sometime?). Thus my new proposal resumes to:

	j = 1;  /* j-th signature key */
	data[0] = 0x00; // maybe line commented out, if data[0] has already 0x00 value
//	data[1] = 0x01;
	while (find_object(sess, CKO_PRIVATE_KEY, &privKeyObject, NULL, 0, j++) != 0) {
		CK_ULONG	modLenBits;

		label = getLABEL(sess, privKeyObject, NULL);
		modLenBits = get_private_key_length(sess, privKeyObject);
		modLenBytes = (modLenBits + 7) / 8;

		/* Fill in datas[0] and dataLens[0] */
		dataLen = modLenBytes;
		memset(data + 2, 0xFF, dataLen - 3 - dataLens[1]);
		data[dataLen - 36] = 0x00;
		memcpy(data + (dataLen - dataLens[1]), datas[1], dataLens[1]);
		datas[0] = data;
		dataLens[0] = dataLen;

frankmorgner: OK, I understand your reasoning. Please leave it with your original suggestion.

PR #797, opened 2016-06-13
Add check_state functions for PKCS#11 C_GetSessionInfo

frankmorgner:
OpenSC tracks the user that's logged in in its PKCS#11 layer. However, due to some other process that's interfering, this state might get corrupted unnoticed. With this PR the login state is actively requested from the card (if supported by the card). See https://github.com/OpenSC/OpenSC/pull/624 for a more detailed problem description.

This PR is intended to replace https://github.com/OpenSC/OpenSC/pull/624 with the following thoughts in mind:
- reuse SC_PIN_CMD_GET_INFO at the card driver level to get the login state
- allow reusing `sc_pkcs15_get_pin_info` in OpenSC's tokend

This PR is tested with sc-hsm.

PR #624, opened 2015-11-28
Add check_state functions for PKCS#11 C_GetSessionInfo

dengert:
Also sees #596 and #618

Firefox and other applications rely on C_GetSessionInfo to return if a login is needed.
They do this before doing other operations. OpenSC would return stale information because
it does not check with PCSC or the card to determine the state of the card.

This commit adds a stack of functions so C_GetSessionInfo gets the current state of the card.
PCSC only returns if the card has been reset or removed. A card driver can provide a
a check_state function that can issue commands to a card to get its true state.

pkcs11-session.c: C_GetSessionInfo
framework-pkcs15.c: pkcs15_check_state
pkcs15.c: sc_pkcs15_check_state
card.c: sc_check_state
card-<driver>.c: <driver>_check_start

sc_pkcs15_check_state could also be used from minidriver or tokend if needed.

The <driver>_check_state function would be different for each card, and is optional.
Only a piv_check_state is included. It sets the AID back to the PIV AID then tests if
the cards logged state agrees with the the expected pkcs11 session state.

Cards without a <driver>_check_state function can benefit from this modification,
as sc_lock is now called from sc_pkcs15_check_state. if the lock is not held,
pcsc will get involved and could return card removed, card reset or other error.
sc_lock will also do some SM checks.

These routines try to match the calls as uses for sign, decrypt and derive.

A call to sc_pkcs15_pincache_revalidate could be added to sc_pkcs15_check_state.

In reader-pcsc.c: pcsc_lock if SCARD_W_RESET was returned, a sleep(1) was added.
Testing showed pcsc might allow access to the reader while the reader is still busy
with the card reset. Further testing may be needed.

This is experimental, and changes may be needed for other cards, or situations.

PR #816, opened 2016-07-04
PIV and PIV-Want-To-Be Issues

dengert:
Fix a number of PIV NEO card issues with using pin_cmd  This is designed to get pin_cmd to return verification state even when the care does not follow the ISO and NIST specs. 
It also addresses other isses with compliance with NIST standards.

It is designed to be usable with #797 and does much of what #624 would have done, but only for PIV cards. 

Many of the fixes are are not related to #797. 

This can be committed before #797  and any tokend changes to need pin_cmd to work.

@mouse07410 Can you see if this solves many of you issues when using #797?

Not all PIV cards follow the  NIST 800-73-3 standard. This commit is designed to address some
of the issues.  OpenSC developers don't have access to all the different versions of devices
or access to release notes for the devices to see when a bug was introduced and when it is fixed.

To make OpenSC code changes easier,  the code is divided into four sections:

(1) Identify the card/token as best possible by looking at the "Historical bytes" in the ATR.
For the Yubico devices read their version number and log it via sc_debug.

(2) Define the card_issues  CI_* defines in card-piv.c. There are 8 of them at the moment.
See below.

(3) based on the card->type and possibly Yubico version set the priv->card_issues flags that
apply to  current card or device.

(4) Implement in the code changes needed for each issue.

Other issues can be added. As more info is obtained (3) can be updated using the version
number as needed.

The card issues are:

CI_VERIFY_630X - VERIFY "tries left" returns 630X rather then 63CX

CI_VERIFY_LC0_FAIL - VERIFY Lc=0 never returns 90 00 if PIN not needed. Will also test after
first PIN verify if protected object can be used instead

CI_CANT_USE_GETDATA_FOR_STATE - No object to test verification in place of VERIFY Lc=0

CI_LEAKS_FILE_NOT_FOUND - GET DATA of empty object returns 6A 82 even if PIN not verified

CI_OTHER_AID_LOSE_STATE - Other drivers match routines may reset our security state and lose AID

CI_NFC_EXPOSE_TOO_MUCH - PIN, crypto and objects exposed over NFS in violation of 800-73-3

CI_NO_RSA2048 - does not have RSA 2048

CI_NO_EC384 - does not have EC 384

The piv_card_match and piv_init interactions were cleaned up.

 Changes to be committed:
	modified:   card-piv.c
	modified:   cards.h


mouse07410:
Thank you! Will try with #797 and report here.

dengert:
In regards to your question on #624 about CAC. It they still fail, try changing line 3042: 
from priv->card_issues = 0; 
to priv->card_issues = CI_OTHER_AID_LOSE_STATE;
and see if that helps. 

I would like to see a debug log if CAC fails without the above change.

mouse07410:
Hmm... Patch to `card-piv.c` failed. Here's what I did:

1. Cloned OpenSC/OpenSC.git master into a new branch (named 797).
2. Pulled PR #797 as a patch, and applied it (everything successful).
3. Pulled PR #816.
4. Tried to apply it - and it failed:

```
$ wget https://github.com/OpenSC/OpenSC/pull/816.patch
--2016-07-06 12:36:29--  https://github.com/OpenSC/OpenSC/pull/816.patch
Resolving github.com (github.com)... 192.30.253.112
Connecting to github.com (github.com)|192.30.253.112|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://patch-diff.githubusercontent.com/raw/OpenSC/OpenSC/pull/816.patch [following]
--2016-07-06 12:36:31--  https://patch-diff.githubusercontent.com/raw/OpenSC/OpenSC/pull/816.patch
Resolving patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)... 192.30.253.112
Connecting to patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)|192.30.253.112|:443... connected.
HTTP request sent, awaiting response... 200 OK
Cookie coming from patch-diff.githubusercontent.com attempted to set domain to github.com
Length: unspecified [text/plain]
Saving to: ‘816.patch’

816.patch                    [ <=>                              ]  20.58K  --.-KB/s    in 0.03s

Last-modified header missing -- time-stamps turned off.
2016-07-06 12:36:32 (660 KB/s) - ‘816.patch’ saved [21075]

$ patch -p1 < 816.patch
patching file src/libopensc/card-piv.c
Hunk #2 FAILED at 159.
Hunk #3 succeeded at 189 (offset 1 line).
Hunk #4 succeeded at 475 (offset 1 line).
Hunk #5 succeeded at 550 (offset 1 line).
Hunk #6 succeeded at 941 (offset 1 line).
Hunk #7 FAILED at 2870.
Hunk #8 succeeded at 2966 (offset 2 lines).
Hunk #9 succeeded at 3068 (offset 2 lines).
Hunk #10 FAILED at 3093.
Hunk #11 FAILED at 3215.
Hunk #12 succeeded at 3295 (offset 27 lines).
4 out of 12 hunks FAILED -- saving rejects to file src/libopensc/card-piv.c.rej
patching file src/libopensc/cards.h
```

dengert:
#797 is making too many changes to card-piv.c that are not needed. 
    
    With #816 card-piv.c  does not try and keep track of the login
    state,  but to allow upper levels to request the state be
    determined  using pin_cmd. This gets the state from the card  and
    not from what the opensc in the current process thinks the state
    should be. Interference from other processes may have reset the card
    or changed the verification state by reselecting a different AID 
    for example.  
    
    #816 was meant to be commited on top of master without #797. 
    Can you try it without #797?
    
    If #797 is going to be commited, then its changes for card-piv.c
    need to be revised. 
    
    
    On 7/6/2016 11:47 AM, Mouse wrote:
    
    
      Hmm... Patch to card-piv.c failed. Here's what I
        did:
      
        Cloned OpenSC/OpenSC.git master into a new branch (named
          797).
        Pulled PR #797
          as a patch, and applied it (everything successful).
        Pulled PR #816.
        Tried to apply it - and it failed:
      
      $ wget https://github.com/OpenSC/OpenSC/pull/816.patch
--2016-07-06 12:36:29--  https://github.com/OpenSC/OpenSC/pull/816.patch
Resolving github.com (github.com)... 192.30.253.112
Connecting to github.com (github.com)|192.30.253.112|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://patch-diff.githubusercontent.com/raw/OpenSC/OpenSC/pull/816.patch [following]
--2016-07-06 12:36:31--  https://patch-diff.githubusercontent.com/raw/OpenSC/OpenSC/pull/816.patch
Resolving patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)... 192.30.253.112
Connecting to patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)|192.30.253.112|:443... connected.
HTTP request sent, awaiting response... 200 OK
Cookie coming from patch-diff.githubusercontent.com attempted to set domain to github.com
Length: unspecified [text/plain]
Saving to: ‘816.patch’

816.patch                    [ <=>                              ]  20.58K  --.-KB/s    in 0.03s

Last-modified header missing -- time-stamps turned off.
2016-07-06 12:36:32 (660 KB/s) - ‘816.patch’ saved [21075]

$ patch -p1 < 816.patch
patching file src/libopensc/card-piv.c
Hunk #2 FAILED at 159.
Hunk #3 succeeded at 189 (offset 1 line).
Hunk #4 succeeded at 475 (offset 1 line).
Hunk #5 succeeded at 550 (offset 1 line).
Hunk #6 succeeded at 941 (offset 1 line).
Hunk #7 FAILED at 2870.
Hunk #8 succeeded at 2966 (offset 2 lines).
Hunk #9 succeeded at 3068 (offset 2 lines).
Hunk #10 FAILED at 3093.
Hunk #11 FAILED at 3215.
Hunk #12 succeeded at 3295 (offset 27 lines).
4 out of 12 hunks FAILED -- saving rejects to file src/libopensc/card-piv.c.rej
patching file src/libopensc/cards.h

      —
        You are receiving this because you authored the thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
>  #816 was meant to be committed on top of master without #797. Can you try it without #797?

I sure can. Will report.

dengert:
If it solves some of your problems, we can look at what it takes
      to get #797 on top of it.
    
    The src/libopensc/card-piv.c.rej file  shows
      which part of the patch did not 
      work. 
      
      If you used git and had two branches, say b816 and b797 then you
      could use git rebase b816 b797 to have git try and merge
      797 on top of 816.
      Or git rebase b797 b816 to have git merge 816 on top of
      797 
      Git rebase should take care of some of some of the failed chunks.
      
    
    But there will still be conflicts, as both are trying to address the problem of state in different ways. 

    
       
      
      
    
    On 7/6/2016 3:46 PM, Mouse wrote:
    
    
      
        #816
          was meant to be committed on top of master without #797.
          Can you try it without #797?
      
      I sure can. Will report. 
      —
        You are receiving this because you authored the thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

dengert:
Even better... #797 has 5 commits. The PIV is only in the last one 
https://github.com/OpenSC/OpenSC/pull/797/commits/71c595eb046d88776311d4f3479fde0293e9d542
    
    I am going to try doing the above. 
    
    
    
    On 7/6/2016 3:46 PM, Mouse wrote:
    
    
      
        #816
          was meant to be committed on top of master without #797.
          Can you try it without #797?
      
      I sure can. Will report. 
      —
        You are receiving this because you authored the thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

src/libopensc/card-piv.c
```diff
@@ -2965,7 +3213,47 @@ static int piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
...
+static int piv_logout(sc_card_t *card)
+{
+	int r = 0;
+	/* piv_private_data_t * priv = PIV_DATA(card); */
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	/* TODO 800-73-3 does not define a logout, 800-73-4 does */
+
+	SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, r);
```
frankmorgner: If the card doesn't implement a logout function, you should return `SC_ERROR_NOT_SUPPORTED` instead of ignoring the explicit logout. If a logout according to 800-73-3 (whatever this may be) is not possible you could use `sc_reset` instead.

src/libopensc/card-piv.c
```diff
@@ -2965,7 +3213,47 @@ static int piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
...
+
+	priv->pin_cmd_verify_sw1 = 0x00U;
+
+	priv->pin_cmd_verify = 1; /* tell piv_check_sw its a verify to save sw1, sw2 */
+	r = iso_drv->ops->pin_cmd(card, data, tries_left);
+	priv->pin_cmd_verify = 0;
+
+	/* if access to applet is know to be reset by other driver  we select_aid and try again */
+	if ( priv->card_issues & CI_OTHER_AID_LOSE_STATE && priv->pin_cmd_verify_sw1 == 0x6DU) {
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "AID may be lost doing piv_find_aid and retry pin_cmd");
```
frankmorgner: `priv->pin_cmd_verify_sw1 == 0x6DU` will always be false because you're setting `priv->pin_cmd_verify_sw1 = 0x00U;` a couple of lines before and `r = iso_drv->ops->pin_cmd(card, data, tries_left)` won't change this value

mouse07410:
@dengert how should I interface this new code from Tokend? With #624 I was calling `sc_pkcs15_check_state()`. What function do I call now?

src/libopensc/card-piv.c
dengert: iso_drv->ops->pin_cmd issues the command, then calls sc_check_sw which calls piv_check_sw. 
piv_check_sw  lines 3100-3103 will save priv->pin_cmd_verify_sw1 and sw2  then in line 3120  calls iso_check_sw
and returns to iso_drv->ops->pin_cmd. 
The sw1 and sw2 are saved because pin_cmd does return what may have really happened, and saving the sw1 and sw2 in priv is a way to determine in the piv_pin_cmd may have failed because the PIV applet may have been de-selected, and in line 3226 the opin AID is selected and the pin_cmd is retried in line 3229.

dengert:
@mouse07410,
@frankmorgner  has removed the card-piv.c changes from #797. Thus #816 and #797 can be added to your test branch in either order. You would then call whatever function you and @frankmorgner have for tokend and #797 

mkdir test
cd test
git clone git@github.com:OpenSC/OpenSC.git
cd OpenSC
git fetch origin pull/816/head:pr816
git fetch origin pull/797/head:pr797
git branch -a
git rebase master pr797
git rebase master pr816
git checkout master
git checkout -b 797_816
git merge pr816
git merge pr797

dengert:
@frankmorgner,
here is an opens-debug using a Yubikey 4 which has the issue of the PIV aid being lost by a selection of a different AID and a reset being done by another process. 
[yubikey4-reset-by-other-process.797_816-20160707.opensc-debug.log.txt](https://github.com/OpenSC/OpenSC/files/352978/yubikey4-reset-by-other-process.797_816-20160707.opensc-debug.log.txt)

Tat is the output of:
/opt/smartcard/bin/pkcs11-tool --module /opt/smartcard/lib/opensc-pkcs11.so -l -O

it was started and then when the prompt for the pin is displayed, the following  command is run in another window to cause interference. 

/usr/bin/opensc-tool -s "00 A4 04 00 0F D2 33 00 00 00 45 73 74 45 49 44 20 76 33 35"
(/etc/opensc/opensc.conf  has the default disconnect_action = reset;)

This does a select file (select AID) as issued by the card-mcrd.c from mcrd_match_card. 
This was chosen to show that  Yubikey 4 loses the selection of the PIV AID when the NIST specs say it shall not lose the card state for an invalid AID. 

Then the PIN is entered. 

In the opensc-debuglog only the pkcs11-tool output is shown. The other command was run  and when it completed, the PIN was entered. This in the time period between 14:42:14.112 and 14:42:23.293

C_Login is called, pcsc responds that the card has been reset by the other process, and sc_pin_cmd is called 14:42:23.293 line 1708. 
Line 1709 is piv_pin_cmd.
Line 1718 the verify command is sent and it retuens 6D 00 
Line 1726:  iso7816.c:121:iso7816_check_sw: Instruction code not supported or invalid.
Line 1727: card-piv.c:3225:piv_pin_cmd: AID may be lost doing piv_find_aid and retry pin_cmd
Line 1728: card-piv.c:754:piv_find_aid: called
Line 1739: PIV AID is sent. 
Line 1759: PI command is resent returning 90 00 

Line 1727 is in the report of the issue in the code you asked about testing for 6D 00. 

When run the interference program is run with disconnect_action=leave; the same 6D 00 is returned,
because the AID is not selected.

src/libopensc/card-piv.c
frankmorgner: OK, I see. Nevermind.

frankmorgner:
First of all, it's great that you're solving the PIV (wanna be) quirks inside the card driver. Also, I think it's best to solve the PIN related stuff inside `piv_pin_cmd`.

A simple workaround for the problem with the reset could be to re-initialize the card when the reset was reported (i.e. somewhere around https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L405). Without any big changes we could do the following here:

1. call `card->finish`
2. call `card->init`

Thought this possibly introduces some superfluous communication, it should definitely set the card into a usable state after the reset.

frankmorgner:
Re-initializing the card after a reset should even work *for all cards*, and doesn't rely on a special call back implemented in the card driver.

dengert:
I disagree. Interference of our process by some other process comes in many forms. A reset by another process is just one type of interference which PCSC can detect and report. (I can send you an opensc-debug log were there was no reset done, but the AID was reset.) Other types of interference include a different AID was selected then the one our process was using. A Secure message session was interrupted and our process need to reestablish it. (This may already be handled.)  A logout was done and a verify needs to be done again.  Other cards may have more issues. 

Re-initializing the card can be very time consuming and aggravating for the user, as many cards read every cert on re-initialization and any pin caching would also be lost. 

Re-initializing can loose session objects. For example PKCS#11 ECDH  key derivation is a two step operation. PKCS#11 key derivation returns a secret key object. The PKCS#11 application does not have to use it right away, giving time for interference from some other process. If the secret key is stored on the card, interference may cause it to be lost, if it is stored in memory interference will not cause a problem, but a card re-initialization will lose it even if in memory or on the card (PIV key derivation returns the secret key in one operation from the card and the key gets stored in memory until PKCS#11 requests it.)

So I still think the card driver should get involved, it has routines for every other type of operation, in the sc_card_operations structure. If you want some default routine in iso7816.c that is OK, but the card driver should be able to override it.

Interference from other processes cannot occur while our process holds the PCSC lock, i.e. from SCardBeginTransaction to SCardEndTransaction. Or SCardConnect to SCardDisconnect if SCARD_SHARE_EXCLUSIVE or SCARD_SHARE_DIRECT are used. 

So a good place for a card driver routine to check and fix if possible it is when the PCSC lock is obtained. the <card>_check_state routines never went that far, but they cold have.

frankmorgner:
I agree that there are many other ways an other process may cause trouble. However, session objects are most likely no problem since they are typically lost by a card after a reset.

It is true that in theory it is possible to put a lot of thought into the process of how to recover interference. Doing this at the card driver level will most likely not be enough, because for some things need a global view to protect against all possible loss of information. In the end, I think, protecting against everything is not very practical. This would introduce a lot of unused code which will most likely not be of much use.

I suggested to protect against a detectable reset, because it's easily done and it's likely. If you are paranoid, you could extend the `atomic` configuration. Its purpose is to rebuild the session for every meaningful transaction and to destroy it afterwards. This currently tracks the login state, but could do more...

mouse07410:
Actually, while it's fun making alternatives for working code, I'm asking myself how much effort it is really worth. A simplified #624 did the job for me satisfactory enough.

My criteria: 
1. OpenSC CLI, library, and apps like OpenSSL & libp11 should work reliably
2. Tokend should work reliably (and therefore, the native Mac OS X apps)
3. These two should minimally interfere with each other, ideally not requiring removal and re-insertion of the token; for sure they shouldn't require re-login or reboot

Anyway...  #797 + #816 (applied an hour ago to the OpenSC master) failed to work:
```
$ pkcs11-rsa-pss-sign-demo2
Generating ephemeral file /tmp/derive.15851.text to test RSA-PSS signature...

openssl rand -hex -out /tmp/derive.15851.text 5120

Signing file /tmp/derive.15851.text...
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.15851.text.sig /tmp/derive.15851.text
engine "pkcs11" set.
PKCS#11 token PIN:
Signature is stored in /tmp/derive.15851.text.sig

Verifying signature:
openssl dgst -engine pkcs11 -keyform engine -sha384 -verify id_02 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature /tmp/derive.15851.text.sig  /tmp/derive.15851.text
engine "pkcs11" set.
Verified OK

$ pkcs11-rsa-pss-sign-demo stamp-h1
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out stamp-h1.sig stamp-h1
engine "pkcs11" set.
PKCS#11 token PIN:
Login failed
login to token failed, returning NULL...
PKCS11_get_private_key returned NULL
cannot load key file from engine
140735227502672:error:80005005:Vendor defined:PKCS11_login:General Error:p11_slot.c:216:
140735227502672:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:eng_pkey.c:124:
unable to load key file
. . . . .
```

Here's the log:
[opensc-797-2-pss2.txt](https://github.com/OpenSC/OpenSC/files/353386/opensc-797-2-pss2.txt)

And here's the log of failed sequence of Keychain Access unlock->lock->unlock etc. At the end of this log the token cannot be unlocked by Keychain Access any more - it doesn't even prompt for the PIN.
[opensc-797-2-keychain1.txt](https://github.com/OpenSC/OpenSC/files/353389/opensc-797-2-keychain1.txt)

dengert:
Looking at opesc-797-2-pss2.txt ...

Your NEO has version 1.0.4 Mine have 0.1.2
For both : (line 141 in your log)
card-piv.c:3050:piv_init: PIV card-type=14003 card_issues=0x00020213
all looks OK
Line 1726 sends the verify Lc=0
status returned gets 6D 00  which should be handled by CI_OTHER_AID_LOSE_STATE. I had not see this in my NEO version 0.1.2, but have see it in the Yubikey4.  

Can you add this so lines 3027-2032 look like:

                        priv->card_issues = CI_NO_EC384
                                | CI_VERIFY_630X
                                | CI_VERIFY_LC0_FAIL
                                | CI_OTHER_AID_LOSE_STATE
                                | CI_LEAKS_FILE_NOT_FOUND
                                | CI_NFC_EXPOSE_TOO_MUCH;

Also change  line 3182 Or 3183 if the above line was added):
r = piv_get_data(card, protected_objects[priv->object_test_verify], &rbuf, &buf_len);
r = piv_get_data(card, priv->object_test_verify, &rbuf, &buf_len);

mouse07410:
> Can you add this ...

Adding, and will report.

>  `| CI_NFC_EXPOSE_TOO_MUCH;`

Around line 2032 - what does that flag mean? What's its effect?

dengert:
It is a statement that the NEO does not follow PIV standards and
    presents a security risk.
    But it is not used anywhere in the code. 
    
    On 7/8/2016 11:45 AM, Mouse wrote:
    
    
      
        Can you add this ...
      
      Adding, and will report.
      
        | CI_NFC_EXPOSE_TOO_MUCH;
      
      Around line 2032 - what does that flag mean? What's its effect?
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
> It is a statement that the NEO does not follow PIV standards and presents a security risk. But it is not used anywhere in the code.

OK, thanks!

dengert:
I have an updated card-piv.c that that is built upon #797 and #816.  #797 changes how the pin_cmd returned the login state and the tries_left. There are some other changes in the commit:
https://github.com/OpenSC/OpenSC/commit/df62b359cc1b76a59417550550d24c8e890f45a1
is applied upon a merge of #797 and #816.  There are some other changes in the commit which @mouse07410 had pointed out. 

@frankmorgner  it is not clear how to make a PR of all of this. 
https://github.com/dengert/OpenSC/tree/not-all-pivs-are-pivs-797

This list of commits can be found here: 
https://github.com/OpenSC/OpenSC/compare/master...dengert:not-all-pivs-are-pivs-797?expand=1

@mouse07410 can you try this? https://github.com/OpenSC/OpenSC/commit/df62b359cc1b76a59417550550d24c8e890f45a1

frankmorgner:
@dengert to create a branch `new-pr` for a PR you could use the following:
```
git fetch origin pull/797/head:new-pr
git checkout new-pr
git cherry-pick bb0ae9bdee0cf0a94d1cd5f174d4a2404662a1c3
git cherry-pick df62b359cc1b76a59417550550d24c8e890f45a1
```
I am not sure which other changes you mean, but they may be cherry picked as well.

mouse07410:
Right now I seem unable to download or pull a single commit (unlike a PR), and it's a bit too large to just go in and do it manually.

Perhaps you want to do what @frankmorgner suggested and create a new branch? Or just push it as a part of #816?

dengert:
@frankmorgner  That is close to what I did. 
pulled #797 and #816 the rebased #816 on top of #797 then added the additional commit with the new changes. 

The question is: do I create a new PR that also includes your #797?  
Or do you commit #797 first?

mouse07410:
@dengert and @frankmorgner I have been able to try this commit df62b35 only with NEO. But at this time the results are **very promising**. All of my tests passed successfully (so far :).

I still need to try it with CAC (and see if Gemalto Dual reader with its extra performance toll and NFC throws any monkey wrench) - but so far the results seem to be even better than with the original #624. 

**Good work, if I may say so.**

dengert:
Good to hear. On thing that is not handled very well, is the card
    driver does not know when another process reset the card. Additional
    command like a select AID may be needed. 
    
    On 7/9/2016 9:32 PM, Mouse wrote:
    
    
      @dengert and @frankmorgner I have been able to try
        this commit df62b35 only with NEO. But at
        this time the results are very promising. All
        of my tests passed successfully (so far :).
      I still need to try it with CAC (and see if Gemalto Dual reader
        with its extra performance toll and NFC throws any monkey
        wrench) - but so far the results seem to be even better than
        with the original #624.
      
      Good work, if I may say so.
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

frankmorgner:
https://github.com/OpenSC/OpenSC/pull/797 is merged now.

mouse07410:
> Good to hear. 

Indeed! :-)

> On thing that is not handled very well, is the card driver does not know when another process reset the card. Additional command like a select AID may be needed.

You can probably add it...?

P.S. Checked with Firefox and Thunderbird - A.O.K. Good!

mouse07410:
@frankmorgner I am concerned - right now only #797 is merged, which means it broke the *master* for Tokend. It requires both #816 *and* df62b35 to actually work.

dengert:
I was going to submit an updated #816 with df62b35 , but since I wasted 6
    hours Sunday trying to figure out why it master would not compile, I
    did not get back to #816. I will get a new PR for 816 this afternoon
    for you to try. 
    
    
    
    On 7/11/2016 8:01 AM, Mouse wrote:
    
    
      @frankmorgner
        I am concerned - right now only #797
        is merged, which means it broke the master for Tokend.
        It requires both #816
        and df62b35 to actually work.
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
Did a partial test with CAC. First thought it cannot recognize CAC at all. Then noticed that it took **minutes** to recognize it. Here's the log:

[opensc-797-2-cac1.txt](https://github.com/OpenSC/OpenSC/files/357967/opensc-797-2-cac1.txt)

dengert:
@mouse07410  The CAC card looks good. Verify Lc=0 is working as expected. 

15:52:48 Log starts 
              (Other opensc drivers were looking at card  dropping unwanted drivers could help some) 
15:52:58 piv_match_card looks for the PIV AID 
              (PIV gets certs from card. looks like a slow card...) 
15:53:38 All the certs have been read  
               (Tokend populates its structures with the certs and keys) 
15:53:38 Tokend starts polling for the state
               (It polled 106 times every 3 seconds or so)  (What else is going on?)
15:55:02 Pin is sent to card. 


Can you test this updated PR?

If you have problems, can you test with using the card_disconnect = "leave" option in opensc.conf. 
This will cause interference from other opensc programs to be minimal until we get a card_reset test.

mouse07410:
> (Other opensc drivers were looking at card dropping unwanted drivers could help some) 

Somehow I don't seem to have luck with that. This is what I put:
```
    card_drivers = isoApplet, PIV-II, openpgp;
```
and  the CAC is not recognized, with no record in `/tmp/opensc-debug.log` file.

> Can you test this updated PR?

It seems to work, once the initial CAC recognition succeeds (which takes forever). 

Surprisingly, the previous versions (including #624) were recognizing CAC almost instantaneously.

*Need to run now, can't wait for it to finally/eventually recognize CAC. Actually, something got broken - it doesn't recognize NEO either. Need to reset something...*

mouse07410:
Two issues:

1. In 0.16.0 and #624 CAC was detected and recognized in seconds. Now with #797 it takes *minutes* for the recognition to complete, and frequently it fails to recognize it at all. Keychain would show no token then.
2. If the token is left in the reader and sits idly - it "falls asleep". OpenSC can access it when invoked - but as far as tokend and therefore all the native Apple apps are concerned, the token just isn't there any more. So it cannot be unlocked, or awakened. Re-insertion is the only way to reanimate it. This could be a problem in a configuration where token removal means automatic screenlock or logoff.

Jakuje:
Not sure how much off-topic I will be, but since I see the notes about CAC cards, which is a thing we worked on recent months. We have a patch for CAC support in OpenSC already, but I am waiting to finish some tests, before I will open a pull request. If anyone has a CAC cards and wants to give it a try, there is my branch with the patches:
https://github.com/Jakuje/OpenSC/tree/jjelen-cac

mouse07410:
@Jakuje indeed off-topic but very interesting. We access CAC as PIV - and it seems to work fine, so far so good. Except for the current problem that started showing up since #797. Perhaps your patch would make things even better, or remedy the above problem. 

I might be able to give it a spin - but please don't count on it.

mouse07410:
@dengert thank you - this makes it easy to pull and test. Can I ask to you to do the same for df62b35 (i.e., add it to #816)? So I don't have to chase it down separately (and my previous tests showed that #816 does not work without it)?

dengert:
I did that.  I rebased #816 when  #797 was committed. The  commit
    you are interested in is now:
https://github.com/OpenSC/OpenSC/pull/816/commits/ace2775a0444b279deb23e5dabe0c3e1781de7fe
    
    With the new commits being added to master, (some that causes
    problems for me) I try and rebase any PRs I have to keep them up to
    date with master.
    
    On 7/15/2016 7:57 AM, Mouse wrote:
    
    
      @dengert thank you - this makes it
        easy to pull and test. Can I ask to you to do the same for df62b35 (i.e., add it to #816)?
        So I don't have to chase it down separately (and my previous
        tests showed that #816
        does not work without it)?
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
@dengert I've installed the updated OpenSC. With it somehow the delay CAC introduces is not as outrageous - between 10 and 25 seconds against 3-4 seconds of Yubikey, but still acceptable.

Here's the log:
[opensc-797-3-cac2.txt](https://github.com/OpenSC/OpenSC/files/366403/opensc-797-3-cac2.txt)

I tend to think that we reached the point when #816 can (and should) be merged, at the very least to remedy *master*'s inability to work with PIV. Whatever other rough edges #816 might have, we can find and remedy along the way (like with the rest of OpenSC code).

mouse07410:
Problem report:
```
$ pkcs11-rsa-pss-sign-demo stamp-h1
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out stamp-h1.sig stamp-h1
engine "pkcs11" set.
Specified object not found
PKCS11_get_private_key returned NULL
cannot load key file from engine
140735112011856:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:eng_pkey.c:124:
unable to load key file
. . . . .
```

Log: 
[opensc-797-3-cac3.txt](https://github.com/OpenSC/OpenSC/files/366508/opensc-797-3-cac3.txt)

mouse07410:
At this time I'm very happy with how #816 works with Yubikey in PIV mode.

CAC seems to still have problems - though I don't think it should stop us from merging #816. After all, #797 was merged despite the fact that it completely broke all the PIV. #816 remedies that for most of the PIV tokens.

Now, the specific problem seems to be that the CAC starts rejecting `00 20 00 80` (with PIN and without PIN), returning `6A 86` ("Incorrect P1 or P2 parameters"). This happens after a bunch of OpenSC and/or OpenSSL CLI commands successfully worked with the CAC:

```
$ pkcs11-rsa-pss-sign-demo stamp-h1
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out stamp-h1.sig stamp-h1
engine "pkcs11" set.
PKCS#11 token PIN: 
Signature for stamp-h1 is stored in stamp-h1.sig

pkcs11-tool -r -y pubkey -d 02 -o /tmp/derive.52807.pubkey.02
Using slot 2 with a present token (0x8)
openssl dgst -verify /tmp/derive.52807.pubkey.02 -keyform DER -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature stamp-h1.sig < stamp-h1
Verified OK

openssl dgst -engine pkcs11 -keyform engine -sha384 -verify id_02 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature stamp-h1.sig  stamp-h1
engine "pkcs11" set.
Verified OK

$ pkcs11-rsa-pss-sign-demo stamp-h1
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out stamp-h1.sig stamp-h1
engine "pkcs11" set.
PKCS#11 token PIN: 
Signature for stamp-h1 is stored in stamp-h1.sig

pkcs11-tool -r -y pubkey -d 02 -o /tmp/derive.52820.pubkey.02
Using slot 2 with a present token (0x8)
openssl dgst -verify /tmp/derive.52820.pubkey.02 -keyform DER -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature stamp-h1.sig < stamp-h1
Verified OK

openssl dgst -engine pkcs11 -keyform engine -sha384 -verify id_02 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature stamp-h1.sig  stamp-h1
engine "pkcs11" set.
Verified OK

$ pkcs11-rsa-pss-sign-demo2
Generating ephemeral file /tmp/derive.52833.text to test RSA-PSS signature...

openssl rand -hex -out /tmp/derive.52833.text 5120

Signing file /tmp/derive.52833.text...
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:object=SIGN%20key;object-type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.52833.text.sig /tmp/derive.52833.text
engine "pkcs11" set.
PKCS#11 token PIN: 
Signature for /tmp/derive.52833.text is stored in /tmp/derive.52833.text.sig

Verifying signature:
openssl dgst -engine pkcs11 -keyform engine -verify "pkcs11:object=SIGN%20pubkey;object-type=public" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature /tmp/derive.52833.text.sig  /tmp/derive.52833.text
engine "pkcs11" set.
Verified OK

$ pkcs11-rsa-encr-demo2
Generate a random data file (1000 bytes), Base64-encoded...
openssl rand -base64 -out /tmp/derive.52845.text 1000

Generated
  key: c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f
  iv:  99e5e5049aa1daf43213100bb5bc5a72

Encrypt file /tmp/derive.52845.text with this key and IV...
openssl enc -aes-256-cfb -a -e -in /tmp/derive.52845.text -out /tmp/derive.52845.text.enc -K c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f -iv 99e5e5049aa1daf43213100bb5bc5a72

Encrypting this random symmetric key to token RSA KEY MAN key...
echo c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f | xxd -r -p -c 200 | openssl rsautl -engine pkcs11 -keyform engine -encrypt -pubin -inkey "pkcs11:object=KEY%20MAN%20pubkey;object-type=public" -pkcs -out /tmp/derive.52845.key.enc
engine "pkcs11" set.

Decrypting the symmetric key on the token...
openssl rsautl -engine pkcs11 -keyform engine -decrypt -inkey "pkcs11:object=KEY%20MAN%20key;object-type=private" -pkcs -in /tmp/derive.52845.key.enc | xxd -p -c 200
engine "pkcs11" set.
PKCS#11 token PIN: 

Decrypt file /tmp/derive.52845.text.enc with this key and IV...
openssl enc -aes-256-cfb -a -d -in /tmp/derive.52845.text.enc -out /tmp/derive.52845.text.dec -K c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f -iv 99e5e5049aa1daf43213100bb5bc5a72

KEY1="c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f"
KEY2="c30ddc8d2cd7147cafb104b2b3af6d5b43710342314fe37f5fe049a1cf678a9f"
Original and decrypted keys  match
Original and decrypted files match
$
```

At this point I tried unlocking the CAC with Keychain Access. It correctly prompted for the PIN, according to the log it seems that the PIN was passed along, but the card refused it:
[opensc-797-3-cac4.txt](https://github.com/OpenSC/OpenSC/files/366627/opensc-797-3-cac4.txt)

dengert:
@LudovicRousseau,  
@mouse07410 is getting this error from PCSC on MacOS using a  Gemalto Prox Dual USB PC Link Reader. Tokend is running, and then a OPenSC based utility is run. The command that failed it trying to do a getdata of an object on the card. 

Any idea why MacOS pcsc would return this? 
Is it a problem with the Gemalto Prox Dual USB PC Link Reader that has a chip reader and a NFC reader in one device?

In the trace:
Line 14942: 0x7fff725be000 13:09:14.229 [opensc-pkcs11] reader-pcsc.c:228:pcsc_internal_transmit: Gemalto Prox Dual USB PC Link Reader(1):SCardTransmit/Control failed: 0x80100016

https://pcsclite.alioth.debian.org/api/group__ErrorCodes.html#gab02a33c2ef61f12a851dfe85c575d7cc
https://msdn.microsoft.com/en-us/library/windows/desktop/aa374738(v=vs.85).aspx#smart_card_return_values

SCARD_E_NOT_TRANSACTED 0x80100016 An attempt was made to end a non-existent transaction.

Google for:  PCSC  80100016
This is not an OpenSC only issue.

OpenSC in reader-pcsc does not handle this very well, but try some error handling at 
reader-pcsc.c:397 calling  pcsc_detect_card_presence twice. (looks like lines 234 and 235 call it twice.)
They return 5 == SC_READER_CARD_PRESENT | SC_READER_CARD_INUSE 
Line 237  return SC_ERROR_TRANSMIT_FAILED; (-1107) 

So no real error processing is done to retry the apdu again and the cmd line command fails.


So why did pcsc return 0x80100016 and what can be done to retry the operation? 


refresh_attributes

dengert:
https://github.com/OpenSC/OpenSC/files/366627/opensc-797-3-cac4.txt

This is a great example of SCARD_W_RESET_CARD  interference of the tokend session by the opensc-pkcs11 process resetting the card on disconnect. The card driver is not notified of this, and #797 or #624 did handle this in all cases. 

The PIV specs say the default application is to be the PIV. But I suspect the default application on the CAC card you are using is not the PIV application. card-piv.c would test if the card has a PIV application by selecting the PIV AID.   (All the cards I have have PIV as default, so am not able to test this.)

line 456843 and 456844:
0x7fff725be000 14:15:36.083 [tokend] reader-pcsc.c:561:pcsc_lock: Gemalto Prox Dual USB PC Link Reader(1):SCardBeginTransaction returned: 0x80100068
0x7fff725be000 14:15:36.083 [tokend] reader-pcsc.c:575:pcsc_lock: Gemalto Prox Dual USB PC Link Reader(1):SCardBeginTransaction calling pcsc_reconnect: 0x80100068

card-piv.c is saying your card is: 
line 122 in trace piv_init: PIV card-type=14001 card_issues=0x00000102 (CI_VERIFY_LC0_FAIL | CI_OTHER_AID_LOSE_STATE)
Thus the piv_select_aid is never being sent.  (Some card would lose the pin state if piv_select_aid was called. Yours appears to need it after a card reset.)

Further complicating this whole process is upon connecting to a card, libopensc tries to match the card to a driver. Some of the other drivers tried before the PIV will try and select their AID. The PIV specs says if the PIV application is selected and an unknown AID is received, the card should ignore the other AID. But some cards don't ignore this and reset the state. CI_OTHER_AID_LOSE_STATE flags such a card You card may be one of these or not. Still not clear. 
 
What I need to do is change the card-piv.c to  detect this is a CAC card. and set the card_issues as needed. All CAC cards may not be the same, so this is a learning process. 

But all the above does not solve the card reset issue, because a card reset could be received with any APDU, not just the pin_cmd commands and a card reset needs to trigger the use the  card_issues to determine if a piv_select_aid needs to be sent after a card reset. 

@mouse07410, what you can do for now is set in opensc.conf section "reader_driver pcsc"
disconnect_action = leave;  OpenSC based software will not reset the card.  (I know you don't like this but that is the circumvention for now.)

It looks like you have already limited in opensc the list of drivers, so the issue of some other driver trying to select its AID may not an issue for you. But you have the isoApplet and PIV in that order, 
and line 54 has:
0x7fff725be000 14:12:20.606 [tokend] card-isoApplet.c:114:isoApplet_select_applet: called
Line 64 sends: 
00 A4 04 00 0C F2 76 A2 88 BC FB A6 9D 34 F3 10 10
that could be interpreted as a select AID. 

You may want to reverse the order of piv and isoApplet.


(These last two opensc-debug logs show a lot of information that I could never produce in my environment. Thanks for doing the testing.)

mouse07410:
> This is a great example of SCARD_W_RESET_CARD interference of the tokend session by the opensc-pkcs11 process resetting the card on disconnect.

Excellent! I'm very happy that you're diagnosing this problem - because it means you'll probably find a solution.

> You may want to reverse the order of piv and isoApplet.

Done.

I seem to observe something (somewhat) similar with NEO (this is before `disconnect_action = leave;`). Here's the sequence:

1. Insert NEO and unlock it with Keychain (succeeds, it is unlocked).
2. Lock NEO with Keychain (succeeds).
3. Run OpenSSL CLI (succeeds).
4. Try to unlock NEO with Keychain - fails and does not even prompt for the PIN.

Log: 
[opensc-797-3-neo2.txt](https://github.com/OpenSC/OpenSC/files/366948/opensc-797-3-neo2.txt)

After setting `disconnect_action = leave;` the above sequence seems to work fine - here's the log:
[opensc-797-3-neo3.txt](https://github.com/OpenSC/OpenSC/files/366949/opensc-797-3-neo3.txt)

> I know you don't like this but that is the circumvention for now.

 :-)  I can live with that. Especially knowing that you're thinking about a better solution.

> These last two opensc-debug logs show a lot of information that I could never produce in my environment. Thanks for doing the testing.

You're very welcome. I'm helping myself, and the world too. What could possibly be better? ;-)

P.S. I still think #816 should be merged now, and the improvement work should continue.

dengert:
@frankmorgner, Can you merge #816. See @mouse07410 comment above.

dengert:
@LudovicRousseau, 
Is the following code listed below  still correct for the MacOS version @mouse07410  is running?
Is the comment in line 454 still valid?

@mouse07410,
What MacOS version are you running?
Does you build define HAVE_PCSCLITE_H in config.h?
Are you building with  pcsc-lite  header files but running with the MacOS pcsc?

OpenSC/src/libopensc/reader-pcsc.c:
 44 #ifdef HAVE_PCSCLITE_H
  45 #if !defined (__MAC_OS_X_VERSION_MIN_REQUIRED) || __MAC_OS_X_VERSION_MIN_REQUIRED < 101000
  46 #define HAVE_PCSCLITE 1
  47 #endif
  48 #endif

 453 #ifndef HAVE_PCSCLITE
 454         /* reconnect unlocks transaction everywhere but in PCSC-lite */
 455         priv->locked = 0;
 456 #endif

mouse07410:
> What MacOS version are you running?

Mac OS X El Capitan 10.11.5.

> Does your build define `HAVE_PCSCLITE_H` in config.h?

```
. . . . .
/* Default PC/SC provider */
#define DEFAULT_PCSC_PROVIDER "/System/Library/Frameworks/PCSC.framework/PCSC"
. . . . .
/* Have OpenSSL libraries and header files */
#define ENABLE_OPENSSL 1

/* Define if PC/SC is to be enabled */
#define ENABLE_PCSC 1
. . . . .
/* Define to 1 if you have the <openssl/crypto.h> header file. */
#define HAVE_OPENSSL_CRYPTO_H 1

/* Define to 1 if you have the <pcsclite.h> header file. */
#define HAVE_PCSCLITE_H 1

/* Define if you have POSIX threads libraries and header files. */
#define HAVE_PTHREAD 1

/* Have PTHREAD_PRIO_INHERIT. */
#define HAVE_PTHREAD_PRIO_INHERIT 1
. . . . .
/* Enabled OpenSC features */
#define OPENSC_FEATURES " locking zlib readline openssl pcsc(/System/Library/Frameworks/PCSC.framework/PCSC)"

/* OpenSC version Git describe revision */
#define OPENSC_SCM_REVISION "OpenSC-0.16.0-69-g9c1714c, rev: 9c1714c, commit-time: 2016-07-15 12:14:52 -0400"

/* OpenSC version fix component */
#define OPENSC_VERSION_FIX 0

/* OpenSC version major component */
#define OPENSC_VERSION_MAJOR 0

/* OpenSC version minor component */
#define OPENSC_VERSION_MINOR 16

/* OpenSC file version revision */
#define OPENSC_VERSION_REVISION 69

/* OpenSC version-info Comments */
#define OPENSC_VS_FF_COMMENTS "Provided under the terms of the GNU Lesser General Public License (LGPLv2.1+)."

/* OpenSC version-info CompanyName value */
#define OPENSC_VS_FF_COMPANY_NAME "OpenSC Project"
. . . . .
```

> Are you building with pcsc-lite header files but running with the MacOS pcsc?

It is my understanding that `pcsc-lite` header files and libraries are those that came with Mac OS X pcsc. 

How can I check that (and do I need to)?

dengert:
Rebased on master see #823.
@mouse07410  See if this get you back to were you were a few days ago. card reset and CAC have not been addressed.

mouse07410:
@dengert it looks like the current #816 rebased against the current master (see #823) works fine with NEO (I kept `disconnect_action = leave;` for now). I hope to run more tests, especially with CAC - but so far so good.

I want to remind again @frankmorgner and @viktorTarasov that the current master broke PIV cards (because of the merged #797), and merging #816 is necessary to restore PIV functionality.

dengert:
@frankmorgner. @viktorTarasov, can you merge #816? ( I could do it, but prefer not to get in you way.)

mouse07410:
@dengert, I've just done a few gentle tests with CAC (current master with your current #816), and it appears to work:

1. Insert CAC, unlock it and lock again using Keychain (both succeeded).
2. Run OpenSSL CLI (succeeded).
3. Unlock CAC using Keychain (succeeded).
4. Run OpenSSL CLI (succeeded).
5. Open Keychain Access app (succeeded, showed CAC correctly unlocked).
6. Lock and unlock CAC using Keychain (both succeeded).

Here's the log: 
[opensc-816-2-cac1.txt](https://github.com/OpenSC/OpenSC/files/369342/opensc-816-2-cac1.txt)

Maybe you should "get in their way" and merge #816 yourself. After all, *master* remains broken (for PIV) - and it doesn't look like people care too much, obviously being busy with other important things.

dengert:
@viktorTarasov, @frankmorgner, 
When are you going to merge #816? The PIV does not work from master without #816 since #797 was merged. If either of you does not respond or take action on #816, I will merge it myself tomorrow at this time.

frankmorgner:
@dengert @viktorTarasov A security issue was just merged; my comment above regarding the logout function is still relevant:

>If the card doesn't implement a logout function, you should return `SC_ERROR_NOT_SUPPORTED` instead of ignoring the explicit logout. If a logout according to 800-73-3 (whatever this may be) is not possible you could use `sc_reset` instead.

For example, the minidriver will automatically do a reset if the logout is not implemented (return code `SC_ERROR_NOT_SUPPORTED`) so that the authentication status can not be misused by an other process. With the current implementation the authentication status is still maintained (no reset) leaving the card authenticated for misuse by any other process. I have not checked, but I guess it's similar with PKCS#11

dengert:
@viktorTarasov, you squished 2 of the 4 commits for improved logging and merged it.  The other two commits still need to be merged to get the piv cards working with #797.

dengert:
@viktorTarasov, 
The other two commits that were not merged address @frankmorgner  concerns.

dengert:
@viktorTarasov @frankmorgner, I take back what I said. It looks like you did merge #816. 

Thanks.

dengert:
Will look at this as a separate issue.  
    
    On 7/19/2016 9:14 AM, Frank Morgner
      wrote:
    
    
      @dengert @viktorTarasov
        A security issue was just merged; my comment above regarding the
        logout function is still relevant:
      
        If the card doesn't implement a logout function, you should
          return SC_ERROR_NOT_SUPPORTED instead of
          ignoring the explicit logout. If a logout according to
          800-73-3 (whatever this may be) is not possible you could use
          sc_reset instead.
      
      For example, the minidriver will automatically do a reset if
        the logout is not implemented (return code SC_ERROR_NOT_SUPPORTED)
        so that the authentication status can not be misused by an other
        process. With the current implementation the authentication
        status is still maintained (no reset) leaving the card
        authenticated for misuse by any other process. I have not
        checked, but I guess it's similar with PKCS#11
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
> ... automatically do a reset if the logout is not implemented (return code SC_ERROR_NOT_SUPPORTED) so that the authentication status can not be misused by an other process. With the current implementation the authentication status is still maintained (no reset) leaving the card authenticated for misuse by any other process. I have not checked, but I guess it's similar with PKCS#11

What's the relation between this "auto-reset" and `disconnect_action = leave;` in the config file?

dengert:
Its complicated. 
    
    For the PIV, NIST 800-73-4 setion 3.2.1 VERIFY Card Command defines
    a logout APDU 00 20 FF 80 lc=0 (of if global pin is used, 00 20 ff
    00)  ISO 7816-4 says: "P1 00 (any other value is reserved for future
    use)"  So the NIST command may not be standard. And AFAIK there are
    no 800-73-4 cards in the field.  I also don't know if CAC has this
    command. 
    
    Frank is right, the logout function should either do the logout or
    return not supported I will submit a PR. 
    
    sc_logout is called by the minidriver, and the pkcs11 framework for
    C_Logout and a few utility programs.
    
    The "auto-reset" is more like disconnect_action
      = reset rather then leave. Leave is specifically allowing "for
      misuse by any other process" But that may be your intention, so
      tokend and keychain can both use the token without having to give
      the PIN. 
      
      Also note the in some of your traces, if a process ended without
      ever doing a disconnect, in that case it is equivalent leave.
      Maybe this could be fixed with an atexit routine? Maybe there is a
      PCSC parameter? 
      
      This is further complicated if connect_exclusive = true; or false
      is set in opensc.conf
    
    connect_exclusive = false allows shared access from multiple
      processes, any one of which could do a verify of the PIN and thus
      let any other process access the card. Tis may be the intent of
      the user. 
      
      PCSC EstablishContext has a Scope parameter:
      
      "The Scope parameter is intended to allow a caller to designate
      the “security” context that this Resource Manager Context operates
      within. Normally, this is associated with the user on whose behalf
      the calling process is running, and operations will be restricted
      to devices that the user is allowed to access. Alternately, one
      could request the security context appropriate for a specific
      terminal or the system as a whole."
      
      But AFAIK PCSC-lite only implements a system wide access. Thus
      processes from other users 
      could access the card once verify PIN has been done. (Thus only
      usable on single user systems.)
      
      So to really restrict the access to a card so other processes can
      not use it would require connect_exclusive = true;
      and a logout or reset must be done before the process terminates.
      This could be via disconnect_action = reset or
        sc_logout and the loop-hole of not calling the disconnect needs
        to be closed. 
        
      
    
    On 7/19/2016 4:13 PM, Mouse wrote:
    
    
      
        ... automatically do a reset if the logout is not implemented
          (return code SC_ERROR_NOT_SUPPORTED) so that the
          authentication status can not be misused by an other process.
          With the current implementation the authentication status is
          still maintained (no reset) leaving the card authenticated for
          misuse by any other process. I have not checked, but I guess
          it's similar with PKCS#11
      
      What's the relation between this "auto-reset" and disconnect_action
          = leave; in the config file?
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
There's a strange problem with Gemalto Prox Dual card reader and MacBook Air. I was practically unable to use the NEO token (PIV) in that reader. Smartcard logon did not work. Token recognition was screwed up. And it again displayed token name as "PIV-II" instead of the CN of the first available certificate.  NEO took 25-35 **seconds** to be recognized - against 2-4 seconds it used to take before this merge. 

Here's the log of the NEO insertion - we didn't even get to the unlock yet.
[opensc-master-f-up1.txt](https://github.com/OpenSC/OpenSC/files/372735/opensc-master-f-up1.txt)

Here's the same as above but with a successful (albeit taking 3-4 seconds!) unlock:
[opensc-master-f-up2.txt](https://github.com/OpenSC/OpenSC/files/372745/opensc-master-f-up2.txt)

This is also of interest:
```
$ pkcs15-tool -C
Using reader with a card: Gemalto Prox Dual USB PC Link Reader(2)
Data object 'Card Capability Container'
	applicationName: Card Capability Container
	applicationOID:  2.16.840.1.101.3.7.1.219.0
	Path:            db00
Data object read failed: File not found
Data object 'Card Holder Unique Identifier'
	applicationName: Card Holder Unique Identifier
	applicationOID:  2.16.840.1.101.3.7.2.48.0
	Path:            3000
	Data (61 bytes): 533B3019D4E739DA739CED39C.....
Data object 'Unsigned Card Holder Unique Identifier'
	applicationName: Unsigned Card Holder Unique Identifier
	applicationOID:  2.16.840.1.101.3.7.2.48.2
	Path:            3010
Data object read failed: File not found
Data object 'X.509 Certificate for PIV Authentication'
	applicationName: X.509 Certificate for PIV Authentication
	applicationOID:  2.16.840.1.101.3.7.2.1.1
	Path:            0101
Data object read failed: File not found
Data object 'Cardholder Fingerprints'
	applicationName: Cardholder Fingerprints
	applicationOID:  2.16.840.1.101.3.7.2.96.16
	Path:            6010
	Auth ID:         01
Data object 'Printed Information'
	applicationName: Printed Information
	applicationOID:  2.16.840.1.101.3.7.2.48.1
	Path:            3001
	Auth ID:         01
Data object 'Cardholder Facial Image'
	applicationName: Cardholder Facial Image
	applicationOID:  2.16.840.1.101.3.7.2.96.48
	Path:            6030
	Auth ID:         01
Data object 'X.509 Certificate for Digital Signature'
	applicationName: X.509 Certificate for Digital Signature
	applicationOID:  2.16.840.1.101.3.7.2.1.0
	Path:            0100
Data object read failed: File not found
Data object 'X.509 Certificate for Key Management'
	applicationName: X.509 Certificate for Key Management
	applicationOID:  2.16.840.1.101.3.7.2.1.2
	Path:            0102
Data object read failed: File not found
Data object 'X.509 Certificate for Card Authentication'
	applicationName: X.509 Certificate for Card Authentication
	applicationOID:  2.16.840.1.101.3.7.2.5.0
	Path:            0500
Data object read failed: File not found
Data object 'Security Object'
	applicationName: Security Object
	applicationOID:  2.16.840.1.101.3.7.2.144.0
	Path:            9000
Data object read failed: File not found
Data object 'Discovery Object'
	applicationName: Discovery Object
	applicationOID:  2.16.840.1.101.3.7.2.96.80
	Path:            6050
	Data (20 bytes): 7E124F0BA0000003080000100001005F2F024000
Data object 'Cardholder Iris Image'
	applicationName: Cardholder Iris Image
	applicationOID:  2.16.840.1.101.3.7.2.16.21
	Path:            1015
Data object read failed: File not found
```
with the corresponding log:
[opensc-master-f-up3.txt](https://github.com/OpenSC/OpenSC/files/372747/opensc-master-f-up3.txt)

Here's the same thing when I unplugged Gemalto reader and inserted NEO directly into USB slot (and now everything appears kosher):
```
pkcs15-tool -C
Data object 'Card Capability Container'
	applicationName: Card Capability Container
	applicationOID:  2.16.840.1.101.3.7.1.219.0
	Path:            db00
	Data (53 bytes): 5333F015A000000116FF.....
Data object 'Card Holder Unique Identifier'
	applicationName: Card Holder Unique Identifier
	applicationOID:  2.16.840.1.101.3.7.2.48.0
	Path:            3000
	Data (61 bytes): 533B3019D4E739.....
Data object 'Unsigned Card Holder Unique Identifier'
	applicationName: Unsigned Card Holder Unique Identifier
	applicationOID:  2.16.840.1.101.3.7.2.48.2
	Path:            3010
Data object read failed: File not found
Data object 'X.509 Certificate for PIV Authentication'
	applicationName: X.509 Certificate for PIV Authentication
	applicationOID:  2.16.840.1.101.3.7.2.1.1
	Path:            0101
	Data (999 bytes): 538203E3708203DA308203D6.....
Data object 'Cardholder Fingerprints'
	applicationName: Cardholder Fingerprints
	applicationOID:  2.16.840.1.101.3.7.2.96.16
	Path:            6010
	Auth ID:         01
Data object 'Printed Information'
	applicationName: Printed Information
	applicationOID:  2.16.840.1.101.3.7.2.48.1
	Path:            3001
	Auth ID:         01
Data object 'Cardholder Facial Image'
	applicationName: Cardholder Facial Image
	applicationOID:  2.16.840.1.101.3.7.2.96.48
	Path:            6030
	Auth ID:         01
Data object 'X.509 Certificate for Digital Signature'
	applicationName: X.509 Certificate for Digital Signature
	applicationOID:  2.16.840.1.101.3.7.2.1.0
	Path:            0100
	Data (1007 bytes): 538203EB708203E23082.....
Data object 'X.509 Certificate for Key Management'
	applicationName: X.509 Certificate for Key Management
	applicationOID:  2.16.840.1.101.3.7.2.1.2
	Path:            0102
	Data (1013 bytes): 538203F1708203E830.....
Data object 'X.509 Certificate for Card Authentication'
	applicationName: X.509 Certificate for Card Authentication
	applicationOID:  2.16.840.1.101.3.7.2.5.0
	Path:            0500
	Data (916 bytes): 5382039070820387308203.....
Data object 'Security Object'
	applicationName: Security Object
	applicationOID:  2.16.840.1.101.3.7.2.144.0
	Path:            9000
Data object read failed: File not found
Data object 'Discovery Object'
	applicationName: Discovery Object
	applicationOID:  2.16.840.1.101.3.7.2.96.80
	Path:            6050
	Data (20 bytes): 7E124F0BA0000.....
Data object 'Cardholder Iris Image'
	applicationName: Cardholder Iris Image
	applicationOID:  2.16.840.1.101.3.7.2.16.21
	Path:            1015
Data object read failed: File not found
```

@dengert I know my tokens work over NFC, and I use this type of reader for that (though usually on MacBook Pro). Yesterday I updated my OS from 10.11.5 to 10.11.6. And now I moved from my master to OpenSC master (probably I shouldn't've done that, but I was stupid). Do you see any of these as factors capable of causing the above?

dengert:
This is not a PIV issue. It is a PCSC, reader-pcsc.c,  reader or
    MacOS  issue. Please open a new issue. This is NOT related to #816.
    Try a different NFC reader. 
    
    The problem is:  
     Gemalto Prox Dual USB PC Link Reader(2):SCardTransmit/Control
    failed: 0x80100016
    and that is the NFC part of the reader. 
    
    OpenSC does not know how to handle the error other then to give up. 
    
    
    The first dump has 4 of these errors.  The second has 8 of these.
    (One of your earlier dumps had one of these errors.) 
    
    Google for:  0x80100016
    
    Som insight can be gained from:
http://ludovicrousseau.blogspot.com/2010/05/how-to-know-pin-sizes-supported-by.html
    
    On 7/19/2016 8:59 PM, Mouse wrote:
    
    
      There's a strange problem with Gemalto Prox Dual card reader
        and MacBook Air. I was practically unable to use the NEO token
        (PIV) in that reader. Smartcard logon did not work. Token
        recognition was screwed up. And it again displayed token name as
        "PIV-II" instead of the CN of the first available certificate.
        NEO took 25-35 seconds to be recognized -
        against 2-4 seconds it used to take before this merge. 
      Here's the log of the NEO insertion - we didn't even get to the
        unlock yet.
        opensc-master-f-up1.txt
      Here's the same as above but with a successful (albeit taking
        3-4 seconds!) unlock:
        opensc-master-f-up2.txt
      This is also of interest:
      $ pkcs15-tool -C
Using reader with a card: Gemalto Prox Dual USB PC Link Reader(2)
Data object 'Card Capability Container'
    applicationName: Card Capability Container
    applicationOID:  2.16.840.1.101.3.7.1.219.0
    Path:            db00
Data object read failed: File not found
Data object 'Card Holder Unique Identifier'
    applicationName: Card Holder Unique Identifier
    applicationOID:  2.16.840.1.101.3.7.2.48.0
    Path:            3000
    Data (61 bytes): 533B3019D4E739DA739CED39C.....
Data object 'Unsigned Card Holder Unique Identifier'
    applicationName: Unsigned Card Holder Unique Identifier
    applicationOID:  2.16.840.1.101.3.7.2.48.2
    Path:            3010
Data object read failed: File not found
Data object 'X.509 Certificate for PIV Authentication'
    applicationName: X.509 Certificate for PIV Authentication
    applicationOID:  2.16.840.1.101.3.7.2.1.1
    Path:            0101
Data object read failed: File not found
Data object 'Cardholder Fingerprints'
    applicationName: Cardholder Fingerprints
    applicationOID:  2.16.840.1.101.3.7.2.96.16
    Path:            6010
    Auth ID:         01
Data object 'Printed Information'
    applicationName: Printed Information
    applicationOID:  2.16.840.1.101.3.7.2.48.1
    Path:            3001
    Auth ID:         01
Data object 'Cardholder Facial Image'
    applicationName: Cardholder Facial Image
    applicationOID:  2.16.840.1.101.3.7.2.96.48
    Path:            6030
    Auth ID:         01
Data object 'X.509 Certificate for Digital Signature'
    applicationName: X.509 Certificate for Digital Signature
    applicationOID:  2.16.840.1.101.3.7.2.1.0
    Path:            0100
Data object read failed: File not found
Data object 'X.509 Certificate for Key Management'
    applicationName: X.509 Certificate for Key Management
    applicationOID:  2.16.840.1.101.3.7.2.1.2
    Path:            0102
Data object read failed: File not found
Data object 'X.509 Certificate for Card Authentication'
    applicationName: X.509 Certificate for Card Authentication
    applicationOID:  2.16.840.1.101.3.7.2.5.0
    Path:            0500
Data object read failed: File not found
Data object 'Security Object'
    applicationName: Security Object
    applicationOID:  2.16.840.1.101.3.7.2.144.0
    Path:            9000
Data object read failed: File not found
Data object 'Discovery Object'
    applicationName: Discovery Object
    applicationOID:  2.16.840.1.101.3.7.2.96.80
    Path:            6050
    Data (20 bytes): 7E124F0BA0000003080000100001005F2F024000
Data object 'Cardholder Iris Image'
    applicationName: Cardholder Iris Image
    applicationOID:  2.16.840.1.101.3.7.2.16.21
    Path:            1015
Data object read failed: File not found

      with the corresponding log:
        opensc-master-f-up3.txt
      Here's the same thing when I unplugged Gemalto reader and
        inserted NEO directly into USB slot (and now everything appears
        kosher):
      pkcs15-tool -C
Data object 'Card Capability Container'
    applicationName: Card Capability Container
    applicationOID:  2.16.840.1.101.3.7.1.219.0
    Path:            db00
    Data (53 bytes): 5333F015A000000116FF.....
Data object 'Card Holder Unique Identifier'
    applicationName: Card Holder Unique Identifier
    applicationOID:  2.16.840.1.101.3.7.2.48.0
    Path:            3000
    Data (61 bytes): 533B3019D4E739.....
Data object 'Unsigned Card Holder Unique Identifier'
    applicationName: Unsigned Card Holder Unique Identifier
    applicationOID:  2.16.840.1.101.3.7.2.48.2
    Path:            3010
Data object read failed: File not found
Data object 'X.509 Certificate for PIV Authentication'
    applicationName: X.509 Certificate for PIV Authentication
    applicationOID:  2.16.840.1.101.3.7.2.1.1
    Path:            0101
    Data (999 bytes): 538203E3708203DA308203D6.....
Data object 'Cardholder Fingerprints'
    applicationName: Cardholder Fingerprints
    applicationOID:  2.16.840.1.101.3.7.2.96.16
    Path:            6010
    Auth ID:         01
Data object 'Printed Information'
    applicationName: Printed Information
    applicationOID:  2.16.840.1.101.3.7.2.48.1
    Path:            3001
    Auth ID:         01
Data object 'Cardholder Facial Image'
    applicationName: Cardholder Facial Image
    applicationOID:  2.16.840.1.101.3.7.2.96.48
    Path:            6030
    Auth ID:         01
Data object 'X.509 Certificate for Digital Signature'
    applicationName: X.509 Certificate for Digital Signature
    applicationOID:  2.16.840.1.101.3.7.2.1.0
    Path:            0100
    Data (1007 bytes): 538203EB708203E23082.....
Data object 'X.509 Certificate for Key Management'
    applicationName: X.509 Certificate for Key Management
    applicationOID:  2.16.840.1.101.3.7.2.1.2
    Path:            0102
    Data (1013 bytes): 538203F1708203E830.....
Data object 'X.509 Certificate for Card Authentication'
    applicationName: X.509 Certificate for Card Authentication
    applicationOID:  2.16.840.1.101.3.7.2.5.0
    Path:            0500
    Data (916 bytes): 5382039070820387308203.....
Data object 'Security Object'
    applicationName: Security Object
    applicationOID:  2.16.840.1.101.3.7.2.144.0
    Path:            9000
Data object read failed: File not found
Data object 'Discovery Object'
    applicationName: Discovery Object
    applicationOID:  2.16.840.1.101.3.7.2.96.80
    Path:            6050
    Data (20 bytes): 7E124F0BA0000.....
Data object 'Cardholder Iris Image'
    applicationName: Cardholder Iris Image
    applicationOID:  2.16.840.1.101.3.7.2.16.21
    Path:            1015
Data object read failed: File not found

      @dengert I know my tokens work over
        NFC, and I use this type of reader for that (though usually on
        MacBook Pro). Yesterday I updated my OS from 10.11.5 to 10.11.6.
        And now I moved from my master to OpenSC master (probably I
        shouldn't've done that, but I was stupid). Do you see any of
        these as factors capable of causing the above?
      —
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
> This is not a PIV issue. It is a PCSC, reader-pcsc.c,  reader or MacOS  issue. Please open a new issue. This is NOT related to #816. Try a different NFC reader.

Thank you - submitted https://github.com/OpenSC/OpenSC/issues/833. I'm sure you're right, and this has to do with PCSC or reader-pcsc.c. MacOS - maybe, but I doubt it because it works on MacBook Pro with the same software (including OS and reader). Maybe performance requirements of the reader can't be satisfied by the I/O of MacBook Air...? I don't know... But I'd like it to work...

PR #303, opened 2014-10-18
Gnuk support

hongquan:
Hi

This is the redo of Pull Requet #153, with some update based on Viktor and Frank's comments.

This is the log to show that it works http://kopy.io/6wqKd

PR #817, opened 2016-07-05
TrueCrypt/VeraCrypt support in OpenPGP

hongquan:
Hello

This is a redo work of #150. However, most of its commits, by my mistake, is already included in #303 and gets merged (It is tested and works well, though).

Here are some minor changes for the feature.

### DIFF
--- a/src/libopensc/pkcs15-openpgp.c
+++ b/src/libopensc/pkcs15-openpgp.c
@@ -398,7 +398,7 @@ sc_pkcs15emu_openpgp_add_data(sc_pkcs15_card_t *p15card)
 		 */
 		r = read_file(p15card->card, path, content, sizeof(content));
 		if (r <= 0 ) {
-			sc_log(ctx, "No data get from DO 010%d", i);
+			sc_log(ctx, "Cannot read DO 010%d or there is no data in it", i);
 			/* Skip */
 			continue;
 		}
--- a/src/libopensc/pkcs15-openpgp.c
+++ b/src/libopensc/pkcs15-openpgp.c
@@ -390,8 +390,8 @@ sc_pkcs15emu_openpgp_add_data(sc_pkcs15_card_t *p15card)
 		memset(&dat_info, 0, sizeof(dat_info));
 		memset(&dat_obj, 0, sizeof(dat_obj));
 
-		sprintf(name, "PrivDO%d", i);
-		sprintf(path, "3F00010%d", i);
+		snprintf(name, 8, "PrivDO%d", i);
+		snprintf(path, 9, "3F00010%d", i);
 
 		/* Check if the DO can be read.
 		 * We won't expose pkcs15 DATA object if DO is empty.
--- a/src/pkcs15init/pkcs15-openpgp.c
+++ b/src/pkcs15init/pkcs15-openpgp.c
@@ -306,7 +306,7 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile 
 		}
 		if (r > 0) {
 			sc_log(ctx, "DO 0101 is full.");
-			r = SC_ERROR_TOO_MANY_OBJECTS;
+			r = SC_ERROR_NOT_ENOUGH_MEMORY;
 			break;
 		}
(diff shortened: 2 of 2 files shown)


frankmorgner:
your changes are not related to Truecrypt, as far as I can see... But they don't do any harm either. What's the current state of OpenPGP/OpenSC support in Truecrypt/Veracrypt?

hongquan:
> your changes are not related to Truecrypt

The changes are related to OpenPGP's private DO, which is used for storing TrueCrypt's keyfile. Actually, this PR should say about OpenPGP "data" DO, but TrueCrypt is mentioned because I want to notify the OpenSC team that the TrueCrypt support has already gone to OpenSC and already work. This PR adds some minor improvements.

> What's the current state of OpenPGP/OpenSC support in Truecrypt/Veracrypt?

I can answer only for OpenPGP case in OpenSC (don't know about other cards). It basically works, but with some limitations due to OpenPGP's limit, as mentioned in https://github.com/OpenSC/OpenSC/pull/150#issue-12506015

mouse07410:
I say again: making TrueCrypt/VeraCrypt use smart cards to unlock the volume is a very good idea. But the current design is broken: many smart cards (particularly those issued by organizations) do **not** allow write access by users.

On the other hand, **all** of those smart cards allow RSA or ECC key wrapping (a bit more difficult with ECC as in this case you cannot pre-generate the key, but still quite usable).

To fix this design flaw, you need to move away from trying to **write** your DO **to** the card, and instead **use** the card as it's supposed to be used, i.e., to encrypt the volume master key with the KEY MANagement key (for PIV) or Encryption subway (OpenPGP).

hongquan:
> you need to move away from trying to write your DO to the card, and instead use the card as it's supposed to be used

I'm aware of this, but it is not the way TrueCrypt does. TrueCrypt uses PKCS#11 card to store keyfile (!), not to encrypt its key/data.

mouse07410:
>> you need to move away from trying to write your DO to the card, and instead use the card as it's supposed to be used

> I'm aware of this, but it is not the way TrueCrypt does. TrueCrypt uses PKCS#11 card to store keyfile (!), not to encrypt its key/data.

In turn, I'm aware of what TrueCrypt does. But I'm trying to tell you that it's a silly way to use a smart card because (a) many smart cards will *not* allow write access, and (b) *all* the smart cards support key encryption in one way or the other.

"...it's not the way TrueCrypt does..." is not an acceptable excuse.

hongquan:
> I'm trying to tell you that...
I'm totally agree with you about how smartcard should be used.
But I make this feature, not because I encourage to use smartcard for storing TrueCrypt keyfile. It is because some TrueCrypt user, who owns a OpenPGP card, wants to use his smartcard with TrueCrypt, and TrueCrypt also offers smartcard support. The problem is that he doesn't know that TrueCrypt doesn't support in correct way, and actually user doesn't need to care it.

mouse07410:
Sorry for not making myself clear enough. The way "some TrueCrypt user" wants it to work is broken, and (more importantly) will *not* work for many users and many OpenPGP (and other) cards. If you decided to enable OpenPGP card support, the least you could do was to make it right - rather than following *literally* that one ignorant user's request and keeping this thing broken.

After all, TrueCrypt is open source, isn't it?

frankmorgner:
@hongquan please look at https://github.com/OpenSC/OpenSC/pull/817/files, your PR does not contain any functional or logical change. Since you keep saying that those changes are related to Truecrypt, this makes me believe that you forgot to add some change...

hongquan:
@frankmorgner I think you miss one my [comment](#issuecomment-230668019), or I couldn't make clear my idea (with my poor English).
> you forgot to add some change
Those changes, by my mistake, was included in #303 and got merged. But I still mention TrueCrypt in this PR, because at least I want to inform OpenSC team that the OpenSC-OpenPGP code already includes support for TrueCrypt. I don't want the code to have something hidden, unknown to the team.

hongquan:
> After all, TrueCrypt is open source
Yes. Actually, TrueCrypt is discontinued. Its living fork is Veracrypt.

frankmorgner:
@hongquan got it, thanks for the clearification.

PR #821, opened 2016-07-10
Add dkek share hex print to sc-hsm-tool

RaymiiOrg:
The SmartCard-HSM SDK allows a decryption of a keyblob if all DKEK shares are provided. This addition adds the ability to print the HEX value of a DKEK share as the `.js` file requires.

### DIFF
--- a/src/tools/sc-hsm-tool.c
+++ b/src/tools/sc-hsm-tool.c
@@ -77,6 +77,7 @@ static const struct option options[] = {
 	{ "initialize",				0, NULL,		'X' },
 	{ "create-dkek-share",		1, NULL,		'C' },
 	{ "import-dkek-share",		1, NULL,		'I' },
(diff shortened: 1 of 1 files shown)


RaymiiOrg:
Used here: https://raymii.org/s/articles/Decrypt_NitroKey_HSM_or_SmartCard-HSM_private_keys.html

frankmorgner:
@CardContact ?

CardContact:
I'm not sure if the code should promote unsafe practises. The features makes sense in reverse engineering (as in the linked article), but in real-world-use it might lead users into using the wrong options (hey cool, why don't we dump the plain DKEK share to paper and circumvent the password protection).

Hmmmm.....

RaymiiOrg:
Would it help to turn it into a non-default compile time option so that it is available but only if you explicitly compiled it in?

CardContact:
Yes, definitely.

PR #825, opened 2016-07-11
Fixed name inconsistency for PIV driver by making it "piv" in ctx.c

mouse07410:
Making PIV-II driver consistently appear as "piv" in `opensc-tool` and `opensc.conf`


mouse07410:
As usually, messed up indentations and merges back-and-forth

Issue #823, opened 2016-07-10
Building with seperate SOURCE and BUILD directories npa-tool.ggo issues


dengert:
### Expected behaviour

npa-tool.ggo needs to be written to either SOURCE or BUILD so it can be read from that directory. 
Problem started today after git pull of master. 

### Actual behaviour

SOURCE = /afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/OpenSC
BUILD = /afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1
../src/configure  \
    --disable-silent-rules\
   --prefix=$PREFIX  (plus other options not part of this problem)

make then has these errors:

make[3]: Entering directory `/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools'

/bin/sed -e 's,[@]CVCDIR[@],,g' -e 's,[@]PACKAGE[@],opensc,g' -e 's,[@]PACKAGE_BUGREPORT[@],opensc-devel@lists.sourceforge.net,g' -e   's,[@]PACKAGE_NAME[@],Open    SC,g' -e 's,[@]PACKAGE_TARNAME[@],opensc,g' -e 's,[@]PACKAGE_URL[@],,g' -e 's,[@]PACKAGE_SUMMARY[@],,g' -e 's,[@]PACKAGE_VERSION[@],"0.16.0",g' -e 's,[@]X509DI    R[@],,g' < ../../../src/src/tools/npa-tool.ggo.in > ../../../src/src/tools/npa-tool.ggo

/usr/bin/gengetopt --output-dir=../../../src/src/tools < npa-tool.ggo

/bin/bash: npa-tool.ggo: No such file or directory

make[3]: *** [cmdline.h] Error 1

make[3]: Leaving directory `/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools'

The problem is:
/bin/sed writes ../../../src/src/tools/npa-tool.ggo  into SOURCE
/usr/bin/gengetopt  tried to read npa-tool.ggo from the current BUILD

Where should it be written?

What needs to be changed here. The help2man looks like it will also have a problem.
 

./src/tools/Makefile.in
1171 $(BUILT_SOURCES): npa-tool.ggo
1172         $(GENGETOPT) --output-dir=$(srcdir) < $<
1173
[...]

Issue #822, opened 2016-07-10
build of 'master' fails without gengetopt installed


viktorTarasov:
```
not found --output-dir=. < npa-tool.ggo
/bin/bash: not: command not found
Makefile:1172: recipe for target 'cmdline.h' failed
make[3]: *** [cmdline.h] Error 127
```

frankmorgner:
I will have a look later this week

PR #827, opened 2016-07-11
build with seperate SOURCE and BUILD directories

frankmorgner:
Fixes https://github.com/OpenSC/OpenSC/issues/823


dengert:
make -v gives:

/usr/bin/help2man \
                --output=/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools/npa-tool.1 \
                --no-info \
                --source='OpenSC 0.16.0' \
                ./npa-tool
help2man: can't get `--help' info from ./npa-tool
Try `--no-discard-stderr' if option outputs to stderr
make[3]: *** [/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools/npa-tool.1] Error 2

help2man is expecting npa-tool --help to produce output.
config.h has:
/* #undef ENABLE_OPENPACE */

npa-tool.c has:
20 #ifdef HAVE_CONFIG_H
 21 #include "config.h"
 22 #endif
 23 
 24 #ifdef ENABLE_OPENPACE
...
893 #else
894 int
895 main (int argc, char **argv)
896 {
897     return 1;
898 }
899 #endif

Thus npa-tool without ENABLE_OPENPACE will not produce any output. 

Please consider something like this to not build npa-tool at all, and to skip trying to build its man page. 

diff --git a/configure.ac b/configure.ac
index 92720a0..69f5d27 100644
--- a/configure.ac
+++ b/configure.ac
@@ -760,6 +760,8 @@ if test "${enable_man}" = "yes" -o "${enable_doc}" = "yes"; then
        AC_MSG_RESULT([ok])
 fi
 
+if test "${enable_openpace}" = "yes"
+then
 ENABLE_NPATOOL="yes"
 AC_ARG_VAR([HELP2MAN],
            [absolute path to help2man used for man page generation of npa-tool])
@@ -791,6 +793,9 @@ else
     touch "${srcdir}/src/tools/npa-tool.ggo"
     touch "${srcdir}/src/tools/cmdline.h" "${srcdir}/src/tools/cmdline.h"
 fi
+else
+    ENABLE_NPATOOL="no"
+fi

frankmorgner:
Thanks, should be fixed with 578b352

dengert:
The make completes  *BUT*  configure is touching  src/tools/npa-tool.1 which did not exist before. 

In  a separate BUILD and SOURCE environment the intention is SOURCE  is only modified by git and ./bootstrap. Configure, make and make install never write into SOURCE, they only modify BUILD. 
 
In a separate BUILD and SOURCE environment you never have to do make clean in SOURCE and you can start from an empty BUILD directory. 

Please consider writing cmdline.c, cmdline.h, npa-tool.ggo, and npa-tool.1 only into BUILD. They are not source files. 

I should not have to do a git clean -f.  I have not had to do git clean it in the past, why should this PR be any different? 

GENGETOPT, HELP2MAN are only used for npa-tool at the moment. But if they are needed in the future, the whole section in configure.am will need to be rewritten. That cleanup could be done now.
Or at least  move the last "fi" in your commit down to just before the line with  OPENSC_FEATURES=""

frankmorgner:
Leaving the generated files in the SOURCE directory allows distributing them in a distribution tarball (`make dist`). So that one can build them even if no help2man or gengetopt is available. This is also what `configure.ac` tests for and this is why this file gets `touch`ed.

dengert:
If you set SOURCE and BUILD to the same directory(The default if you run configure in  source diorectory) , you get the same effect and can do a make dist. 

So for the those of us who want to use separate SOURCE and BUILD   Do what I ask.

frankmorgner:
I am now putting everything into ${builddir}, could you check again if it fulfills your needs?

Could you add a test in the CI for this kind of stuff?

dengert:
Yes, it fixes **my** problem, I can live with it **but** the whole process of creating the nap-tool is going to come back to haunt the project if it is not fixed now.

The more I look at how this was done: writing out a C source file, using the extra packages, and the way configure.ac test for them, and the touch commands in configure are going to make distro builders refuses to port OpenSC to their systems. 

I agree with @viktorTarasov that the code should be removed from master, and the configure processing of the npa-tool needs a **major** rewrite before being resubmitted. 

With the current fix I get this warning using automake 1.14.1 on Ubuntu:

./bootstrap
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `.'.
libtoolize: copying file `./ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIR, `m4'.
libtoolize: copying file `m4/libtool.m4'
libtoolize: copying file `m4/ltoptions.m4'
libtoolize: copying file `m4/ltsugar.m4'
libtoolize: copying file `m4/ltversion.m4'
libtoolize: copying file `m4/lt~obsolete.m4'
autoreconf: running: /usr/bin/autoconf --force
autoreconf: running: /usr/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
src/tools/Makefile.am:14: warning: source file '$(abs_builddir)/cmdline.c' is in a subdirectory,
src/tools/Makefile.am:14: but option 'subdir-objects' is disabled
automake: warning: possible forward-incompatibility.
automake: At least a source file is in a subdirectory, but the 'subdir-objects'
automake: automake option hasn't been enabled.  For now, the corresponding output
automake: object file(s) will be placed in the top-level directory.  However,
automake: this behaviour will change in future Automake versions: they will
automake: unconditionally cause object files to be placed in the same subdirectory
automake: of the corresponding sources.
automake: You are advised to start using 'subdir-objects' option throughout your
automake: project, to avoid future incompatibilities.
autoreconf: Leaving directory `.'

I can live with that. Searching the web, others have gotten this warning, and suggested adding subdir-objects to configure.am:
-AM_INIT_AUTOMAKE(foreign 1.10)
+AM_INIT_AUTOMAKE(foreign subdir-objects 1.10)

Automake 2.0 will force subdir-objects. If not fixed now, it will be a problem in the future.

Trying to add AM_INIT_AUTOMAKE(foreign subdir-objects 1.10) in configure.ac 
 leads to an error in Make that I can not live with:

make[3]: Entering directory `/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools'
Makefile:810: .deps/npa_tool-cmdline.Po: No such file or directory
make[3]: *** No rule to make target `.deps/npa_tool-cmdline.Po'.  Stop.
make[3]: Leaving directory `/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-1.0.1/src/tools'

This then leads to another set of issues on the Web, as many projects have run into these non standard automake configurations, where you are generates source "c" files. 

## Suggestions for the resubmission of the whole NPA commits:

cmdline.h and cmdline.c are are too generic of names, when they are only used in npa-tool.
Rename these at least to npa-tool-cmdline.h and npa-tool-cmdline.c

Since I am not using openpace and not building npa-tool or the cmdline.c,  I still can not figure out what should be in cmdline.c!!

Better still: rather the creating a cmdline.c, create a npa-tool-cmdline-source.h then have a npa-tool.c source file with a few includes for npa-tool-cmdline.h and npa-tool-cmdline-source.h.

dengert:
@frankmorgner  also asked: "Could you add a test in the CI for this kind of stuff?"

I am not getting involved with the CI effort, I leave that up to others. But it is very easy to test. Say you have your git source in: ~/mysource/OpenSC  After running ./bootstrap in OpenSC You could do:
cd ~/mysource
mkdir OpenSC-test-build
cd OpenSC-test-build
../OpenSC/configure ... (same options you would use if in the source directory.)

When done:
cd ~/mysource/OpenSC-test-build
rm -fr *

For testing with OpenSSL versions, I have these 5 build directories and git source:
amd64_linux26-0.9.7  amd64_linux26-0.9.8  amd64_linux26-1.0.1  amd64_linux26-1.0.2  amd64_linux26-1.1  OpenSC

I also have  in /opt  ossl-0.9.7  ossl-1.0.2 ossl-0.9.8  ossl-1.1 and ossl-1.0.1 These install directories
for OpenSSL, libp11 and OpenSC, and any othr package I may want to test with multiple versions of OpenSSL. 

I also have config.dee script that looks at what build directory it is being run from then sets 
OPENSSL_CFLAGS= OPENSSL_LIBS= and --prefix= to point into /opt/ossl-* and runs ../OpenSC/configure with the options to match.

frankmorgner:
Again, I am happy for all the suggestions and I wished I had them earlier. I'll follow your suggestion to add the generated files to git, which eliminates the need for gengetopt and help2man at compile time (in most cases). I, as maintainer of this stuff, will handle the extra bits for re-generation if needed.

dengert:
Good, that will simplify things a lot.
    
    On 7/14/2016 2:12 PM, Frank Morgner
      wrote:
    
    
      Again, I am happy for all the suggestions and I wished I had
        them earlier. I'll follow your suggestion to add the generated
        files to git, which eliminates the need for gengetopt and
        help2man at compile time (in most cases). I, as maintainer of
        this stuff, will handle the extra bits for re-generation if
        needed.
      —
        You are receiving this because you commented.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

dengert:
This looks good to me it solves my #823 issue 
@viktorTarasov  you may want to try this as solution to #822

frankmorgner:
Use https://github.com/OpenSC/OpenSC/pull/831 instead

PR #829, opened 2016-07-13
Improve JPKI driver

hamano:
successfully passed pkcs11-tool --test

### DIFF
--- a/src/libopensc/pkcs15-jpki.c
+++ b/src/libopensc/pkcs15-jpki.c
@@ -138,10 +138,8 @@ sc_pkcs15emu_jpki_init(sc_pkcs15_card_t * p15card)
 	for (i = 0; i < 2; i++) {
 		static int prkey_pin[2] = { 1, 2 };
 		static int prkey_usage[2] = {
-			SC_PKCS15_PRKEY_USAGE_ENCRYPT |
-			SC_PKCS15_PRKEY_USAGE_DECRYPT |
 			SC_PKCS15_PRKEY_USAGE_SIGN,
-			SC_PKCS15_PRKEY_USAGE_NONREPUDIATION
+			SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_NONREPUDIATION
 		};
 		static const char *prkey_name[2] = {
 			"User Authentication Key",


src/libopensc/pkcs15-jpki.c
```diff
@@ -45,8 +45,8 @@ sc_pkcs15emu_jpki_init(sc_pkcs15_card_t * p15card)
 
 	LOG_FUNC_CALLED(p15card->card->ctx);
 
-	p15card->tokeninfo->label = "JPKI";
-	p15card->tokeninfo->manufacturer_id = "JPKI";
+	p15card->tokeninfo->label = strdup("JPKI");
```
frankmorgner: needs to be `free`d somewhere

frankmorgner: nevermind, i just saw that it gets freed in pkcs15.c

viktorTarasov:
@hamano 
For the reasons not related with support of jpki the _master_ was rebased.

Using this occasion I rebased the initial commit that introduced support of jpki card
have included the first three commits of thie PR.

There is a question to the last commit -- private key usage flags seems to be invalids.
You will have to change or confirm your implementation and it will be applied.

src/libopensc/pkcs15-jpki.c
```diff
@@ -139,7 +139,6 @@ sc_pkcs15emu_jpki_init(sc_pkcs15_card_t * p15card)
 		static int prkey_pin[2] = { 1, 2 };
 		static int prkey_usage[2] = {
 			SC_PKCS15_PRKEY_USAGE_ENCRYPT |
-			SC_PKCS15_PRKEY_USAGE_DECRYPT |
 			SC_PKCS15_PRKEY_USAGE_SIGN,
 			SC_PKCS15_PRKEY_USAGE_NONREPUDIATION
```
viktorTarasov: Usually _non-repudiation_ usage has to be combined with _sign_ usage. It's a common practice for the "Digital Signature Key". 

The pure "Authentication" key has to be with _sign_ usage. 

'Encrypt' usage is usually used with the public key,
private key is usually used with 'Decrypt' flag.

Looking onto the context of _prkey_usage[2]_ data, I suggest that it's value has to be:
``` 
static int prkey_usage[2] = {
    SC_PKCS15_PRKEY_USAGE_SIGN,
    SC_PKCS15_PRKEY_USAGE_SIGN  | SC_PKCS15_PRKEY_USAGE_NONREPUDIATION
};
```

Please, confirm or object this suggestion and we will close you PR.

hamano:
@viktorTarasov OK, I've fixed key usage as you recommended.

viktorTarasov:
applied in cb1da562f466ea50e14dd949e6121196e912c67e

Issue #820, opened 2016-07-10
About src/libcardnpa


viktorTarasov:
Recent commits have introduced the card specific sources on the level where it has no place.

My questions:
- why ?
- have you looked for the other ways to compile your own 'purified' module? Like, for ex. a new compile object in 'libopensc'?

Question of codding style of a newly introduced sources I leave for a moment aside.

frankmorgner:
libcardnpa.dll is an external card driver, that's why it is compiled and created separately. Other users have asked about how to use the external card driver since a reference was missing and OpenSC only comes with internal card drivers. The new card driver can be used as one example how to do that.

viktorTarasov:
For me it's not the reason to put card specific staff on this level.

You can do dedicated make target in libopensc, create folder dedicated to external modules (probably inside libopensc).

frankmorgner:
So you're suggesting to move the folder libcardnpa into libopensc?

viktorTarasov:
_master_ rebased without _npa_ merge.

Let's start new discussion to include the _npa_ support.

Issue #810, opened 2016-06-26
Support DNIe 3.0


Yajo:
### Expected behaviour

I should be able to use DNIe 3.0 with Firefox after configuring it.
### Actual behaviour

Dialog asks for the PIN again and again endlessly, no matter if you enter the right PIN or you press Cance.
### Steps to reproduce
1. Insert your DNIe.
2. Configure OpenSC's PKCS#11 module in Firefox.
3. Visit a DNIe-enabled page, such as https://www1.agenciatributaria.gob.es/wlpl/DASR-CORE/AccesoCO2015RVlt
### Logs

Not sure how to get those.

CC @miguel-cv @germanblanco

frankmorgner:
Indeed @germanblanco wanted to look into this, see https://github.com/OpenSC/OpenSC/pull/655

cameta:
With the old  DNIe (personal data have been edited)

tux cameta # dnie-tool -a
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
DNIe Number:   00000000A
SurName:       ESPAÑOL
Name:    ESPAÑOL
IDESP:   AGR148953
DNIe Version:  DNIe 01.13 B11 H 4C34 EXP 1-((4.2-5))
Serial number: 06CFC2227A398A

with a DNIe 3.0


tux cameta # dnie-tool -a
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
Error: Get info failed: Unsupported CLA byte in APDU	

With the old  DNIe

cameta@tux ~ $ opensc-tool -a
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
3b:7f:38:00:00:00:6a:44:4e:49:65:20:02:4c:34:01:13:03:90:00

cameta@tux ~ $ opensc-tool -n
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
dnie

cameta@tux ~ $ opensc-tool --serial
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
06 CF C2 22 7A 39 8A ..."z9.	


With a DNIe 3.0

cameta@tux ~ $ opensc-tool -a
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
3b:7f:96:00:00:00:6a:44:4e:49:65:20:01:01:55:04:10:03:90:00

cameta@tux ~ $ opensc-tool -n
Using reader with a card: C3PO LTC31 v2 (00509883) 00 00
dnie

[...]

PR #831, opened 2016-07-18
Add Support for electronic signature with German ID card

frankmorgner:
For background please see https://github.com/OpenSC/OpenSC/pull/611, https://github.com/OpenSC/OpenSC/pull/83, https://github.com/OpenSC/OpenSC/pull/827, https://github.com/OpenSC/OpenSC/issues/820

I did all the refactoring suggested in https://github.com/OpenSC/OpenSC/issues/823

> - join the _libisosm_ with _libsm_ ;

The folder _libsm_ is renamed to _sm_. It now contains the SM encoding according to ISO-7816-4 (_libsmiso_) as well as SM establishment according to Extended Access Control which is also referenced in the new ISO-7816-4 standard (_libsmeac_). Please note that the latter is historically associated to travel documents (MRTDs), but it's currently used in millions of PKI cards as well, e.g. SC-HSM which is already supported in OpenSC.

> - move eac, npa cards/stuff into libopensc, (probably create dedicated sub-folder inside libopensc);

_libcardnpa_ was moved to folder _libopensc_.

> - think about some common denominator for the name of _pace_, _boxing_, _npa_, _sceac_ source files (something like _xxx-mrtd-yyy_,  not sure)
- _boxing_ files are named _reader-boxing_. Please note that executing a PIN verification via boxing commands with pseudo APDUs have been standardized to overcome the need for a special smart card driver. More specifically, it allows using reader's PIN pad with Windows' CCID driver (which has no PIN pad support). Thus, its use is not limited to MRTDs nor is it limited to PACE. My implemenatation also adds support for VERIFY/MODIFY PIN commands.
- _sceac_ files are named _sm-eac_
- _npa_ files are (and have always be) named _card-npa_
- _pace.h_ was created from _opensc.h_ in https://github.com/OpenSC/OpenSC/pull/88. Since it defines an internal type to access a reader's feature (PC/SC standard pt 10 amd1) and now lets you perform a creation of a SM channel, I'd move it to _opensc.h_ or _types.h_. I am leaving the choice up to @viktorTarasov.

Additionally the complaint about not being able to build with seperate SOURCE and BUILD directories has been fixed. No additional compile time dependencies are needed to build the core of OpenSC.

Some commits can still be squashed if needed.

@viktorTarasov and @dengert, please review the changes as soon as possible. I'd like to merge the EAC support of SC-HSM from @CardContact as well, which builds up on this SM implementation.


viktorTarasov:
What I've seen:
- you have to rebase PR and to forge 'added ...' with 'moved ...'
- _reader-boxing_:
  - is not a new reader driver -- choose name of source file more precisely;
  - still coding style
- use  short form of log calls;
- why _sfid read/write_ are in SM library ?

frankmorgner:
>reader-boxing:

I tried twice to be as precise as I can. The implementation overwrites a reader's PIN-pad callbacks and it can be applied to any `sc_reader_t`. What do you think about `foobar.h/c`? Honestly, please choose a name that you're most comfortable with.

Please be more precise in what you mean by *coding style* in reader-boxing.*. Github conveniently offers inline code comments!

>use short form of log calls;

I don't see an advantage with it. The long form gives more control when the log is shown (debug vs. verbose). You should not see most of the logs with your cards anyway. However, I'm not pedantic with this, you may rewrite this to your taste (just like you did with https://github.com/OpenSC/OpenSC/commit/eb21c41e533d218589580ca999dbb6d9eb4c6371).

>why sfid read/write are in SM library ?

Short answer: because `read_binary_sfid` is needed here and `write_binary_sfid` is almost identical.
Long answer: there is no better place to put it. The cards in libopensc don't offer such a call back and the concept of a short ID only applies to ISO 7816 cards, it should only be used if we are operating with a compliant card. Unsurprisingly, compliance with ISO 7816-4 comes with ISO 7816-8 which comes with compliance to EAC.

viktorTarasov:
>>reader-boxing:

> I tried twice to be as precise as ...

Afais, _reader-boxing_ contains PACE specific version for to PIN command reader's handle.
In pcsc-reader this handle is implemented using PCSC-10 . 

So, for me, the name of this file has to contain reference to PACE,
unless you put a generic content, that is not related to PCSC-10 and to PACE into _reader-boxing_ and create source files with PCSC-10 and PACE implementation of PIN cmd.


>> coding style

> ...

https://github.com/OpenSC/OpenSC/issues/823#issuecomment-231779343


> > use short form of log calls;

> I don't see an advantage ... 

In your PR you are using different log call forms.
From there:
- my suggestion is to use unique log call style in your non-common source files;
- my firm recommendation to use the short form in the common ones.

>> why sfid read/write are in SM library ?

> ..  there is no better place to put it.  ..

In your long justification I do not seen any reference to SM.
So, still, do not understand why SM library.

Possibility of read/write with Short EFID is specified in ISO7816-4 pp. 7.2.2. 
For me, it has to be in ISO7816 part of libopensc.

frankmorgner:
>Afais, reader-boxing contains PACE specific version for to PIN command reader's handle.
>In pcsc-reader this handle is implemented using PCSC-10 .
>
>So, for me, the name of this file has to contain reference to PACE,
>unless you put a generic content, that is not related to PCSC-10 and to PACE into reader-boxing and create source files with PCSC-10 and PACE implementation of PIN cmd.

Did you read the content of the files? The boxing commands *are implementing all PIN commands via a transmit of an APDU*.

1. It allows you to perform all PIN verification commands from PC/SC pt. 10 (VERIFY/MODIFY) and all commands from PC/SC pt. 10 amd1 (PACE). Hence, it is so generic that it does not only concerns PIN verification with PACE but *all PIN commands*.
2. Since boxing commands are using a standard transmit, they can be used on all other reader implementation, for example on an OpenCT reader. Hence, it is so generic that not only the PC/SC reader is concerned, but *all reader implementations*.

>In your long justification I do not seen any reference to SM.
>So, still, do not understand why SM library.

Honestly, I don't understand your nitpicking. If you read the code than you can see that EAC needs to read a file. To speed this up, I'm using a read by SFID because that doesn't need a previous select command. OpenSC uses `sc_transmit` everywhere bypassing the abstraction offered by libopensc. I don't see a point why doing the same during EAC should not be OK. I guess there would be no discussion if I would have not created a separate file and would have created a `static` function. Would that satisfy your pedantry?

>Possibility of read/write with Short EFID is specified in ISO7816-4 pp. 7.2.2.
>For me, it has to be in ISO7816 part of libopensc

libopensc offers an abstraction from the card specific details. It doesn't export any function that's specific to ISO 7816. There is only a `sc_read_binary` which does not offer an interface for immediate selection of the file via short FID. But, as said before, I'm open to suggestion if you are more specific.

viktorTarasov:
>> ... for me, the name of this file has to contain reference to PACE ...

> Did you read the content of the files?

Yes, I've read, and I will help you to understand your code.

What we see [here](https://github.com/frankmorgner/OpenSC/blob/npa/src/libopensc/reader-boxing.c#L937) ?
The static PINPAD reader's handle is dynamically replaced with the one implemented as PC_to_RDR_Secure command from CCID v1.1 standard.
Implementation of static PINPAD handle is based on PC/SC-10.

From the reader's point of view, the both implementations are on the same level --
both implement the Secure PIN entry, both support verify and modify PIN operations.

From there I expect that the name of sources with CCID PC_to_RDR_Secure handle
clearly contain the reference to the standard.


Second part of _reader-boxing_, seems to be completely unrelated to its CCID part . ([here](https://github.com/frankmorgner/OpenSC/blob/npa/src/libopensc/reader-boxing.c#L945)) 
It implements the PACE-perform reader's handle according to the TR-03110 PACE standard.
This handle is introduced into PC/SC-10 amd1 as a reverence to the Germain eID card.

So, my suggestion is to split the content of _reader-boxing_ onto two parts according to the used standards,
with the names of sources that clearly contains the reference to these two standards.



>  ... I don't understand your nitpicking... 
 
You are introducing a lot of sources with the names and places that pretend to be generic or common.
You can do anything that you want in xx-npa.x sources, I will not say a word, unless I have to poke my nose there to fix some outstanding anomaly.


> libopensc offers an abstraction from the card specific details. It doesn't export any function that's specific to ISO 7816. 

Libopensc contains default implementation of card operations based on ISO-7816 standards.
Many cards are using these default implementation as they are, in their default version.

Using of SFID in read/write operations is a part of ISO-7816, so its implementation has to find its place in an adequate place.
  
My rapid suggestion is to introduce the SFID card operations and place their default implementations into iso7816.c

frankmorgner:
>>> ... for me, the name of this file has to contain reference to PACE ...

>> Did you read the content of the files?

>Yes, I've read, and I will help you to understand your code.

Well, you should at least pretend to not have a god complex. In regard to the boxing commands, you may want to re-read the code, because you're misinterpreting it! I tried to explain the logic in the commit message of 1b9c5bd91f3ead70d618c622e572c1a61ee35283 and in this PR's description. You can get all details in the [technical guideline (english)](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03119/BSI-TR-03119_V1_pdf.pdf).

viktorTarasov:
> Well, you should at least pretend to not have a god complex.

Why should I hide my uttermost quality?


> In regard to the boxing commands, you may want to re-read the code, because you're misinterpreting it! I tried to explain the logic in the commit message of 1b9c5bd and in this PR's description. You can get all details in the technical guideline (english).

In a most honest manner I've re-read your source, your commit and referenced document, and do not understood what relation all this have to our subject -- _naming of source files_.

Unbelievable: what can be unclear here?
There are two implementations of secure PIN entry -- both dispute the same reader's handle, both have the same operations, almost the same data types, but different standards.
What is of supra-natural in the wish to see these standards referenced in the name of source files?


By the way -- why _boxing_? What is _boxing_?
Point me, please, the standard where this term is explained.

frankmorgner:
Usage of the Pseudo APDUs for PIN verification is defined in [BSI TR-03119 D.3 Mapping on CCID](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03119/BSI-TR-03119_V1_pdf.pdf). This specific interface to the reader is lent from PC/SC part 9.

Think of a post office (reader driver) that knows how to handle boxes (TRANSMIT), but doesn't know how to handle letters (VERIFY). You may put your letter into a box, send that via the post office, and let the receiver (reader) handle the letter after unboxing. Piggybacking is the most prominent name for this, but I made the experience that reader vendors use the term boxing. Anyway, you may call it (i.e. rename the files) as you like.

viktorTarasov:
This discussion have exhausted itself,
I have seen neither answers on my questions ("words, words, words") nor constructive attitude  (multiple suggestions to do all the necessary myself are really funny).

frankmorgner:
>I have seen neither answers on my questions ("words, words, words")

Well, words should let me understand you, in theory. Sometimes I wish we would just talk in French, which would make some things easier, I guess...

Anyway, with the input I got so far I think I could start refactoring... again.

> nor constructive attitude

Did you actually realize that calling something "dumb", as you did, is the opposite of constructive or encouraging?

>multiple suggestions to do all the necessary myself are really funny

You are talking about renaming some file as if it was equivalent to the work it took for writing 11k loc (not counting the actual crypto in openpace), that's ridiculous!

frankmorgner:
@viktorTarasov sorry to have not triggered you before. I think that all your remarks are now taken into account with the current patches.

frankmorgner:
I reworded "boxing commands" to escape commands since this is the terminology used in PC/SC pt. 3. Consequently, the `enable_escape` switch is also used for getting the UID (which is an escaped command from PC/SC pt. 3, see discussion in #810).

frankmorgner:
Rebased to master.

Since all problems mentioned above were fixed, I'll merge this PR in a couple of days it there are no more objections.

PR #839, opened 2016-07-24
pkcs11-tool: re-login if pincache is disabled or expired

nunojpg:


nunojpg:
Also, the current pin_cache_counter default is 10. My test card requires 12 logins for a "pkcs11-tool -lt". 

With this patch I am requested the pin 3 times. The first covers 10 logins, and the following 2 are not cached because they are done as CKU_CONTEXT_SPECIFIC.

So my question is, what is the rational for the pin_cache_counter default? Is it a security/convenience tradeoff? What is the standard use case of multiple logins required? Why not 3 or 20?

PR #828, opened 2016-07-11
Setting PIV-II driver short_name to "piv" everywhere

mouse07410:

Issue #818, opened 2016-07-10
Inconsistent PIV driver name


mouse07410:
### Expected behaviour

Based on the output of `opensc-tool --list-drivers`, I should be able to use the name `piv` when configuring OpenSC via `opensc.conf` file, editing/adding commands like `card_drivers = openpgp, piv, internal;`, etc.

```
$ opensc-tool --list-drivers
Configured card drivers:
  piv              PIV-II  for multiple cards
  openpgp          OpenPGP card
  cardos           Siemens CardOS
  flex             Schlumberger Multiflex/Cryptoflex
  cyberflex        Schlumberger Cyberflex
  gpk              Gemplus GPK
  gemsafeV1        driver for the Gemplus GemSAFE V1 applet
  miocos           MioCOS 1.1
  asepcos          Athena ASEPCOS
  starcos          STARCOS SPK 2.3/2.4/3.4
  tcos             TCOS 3.0
  jcop             JCOP cards with BlueZ PKCS#15 applet
  oberthur         Oberthur AuthentIC.v2/CosmopolIC.v4
  authentic        Oberthur AuthentIC v3.1
  iasecc           IAS-ECC
  belpic           Belpic cards
  ias              IAS
  incrypto34       Incard Incripto34
  acos5            ACS ACOS5 card
  akis             TUBITAK UEKAE AKIS
  entersafe        entersafe
  epass2003        epass2003
  rutoken          Rutoken driver
  rutoken_ecp      Rutoken ECP driver
  westcos          WESTCOS compatible cards
  myeid            MyEID cards with PKCS#15 applet
  sc-hsm           SmartCard-HSM
  dnie             DNIe: Spanish eID card
  MaskTech         MaskTech Smart Card
  mcrd             MICARDO 2.1 / EstEID 1.0 - 3.0
  setcos           Setec cards
  muscle           MuscleApplet
  atrust-acos      A-Trust ACOS cards
  itacns           Italian CNS
  isoApplet        Javacard with IsoApplet
  gids             GIDS Smart Card
  default          Default driver for unknown cards
```

### Actual behaviour

Instead, it does not seem to work unless I use the name `PIV-II` instead of `piv`, so the above command in `opensc.conf` has to look like `card_drivers = openpgp, PIV-II, internal;`

### Steps to reproduce

1. Build and install the current master of OpenSC
[...]

PR #841, opened 2016-07-25
adding a CAC support into OpenSC

Jakuje:
During last months we worked on the CAC support in OpenSC. Some of the code is based on the PIV driver, the rest is completely new to work flawlessly with CAC.

We tested the code with both CAC emulator ([libcacard](https://github.com/SPICE/libcacard) in qemu) and real CAC1 and CAC2 cards with positive results (used [testsuite](https://github.com/Jakuje/OpenSC/tree/jjelen-testsuite/src/tests/p11test)).

Notes about related changes in the shared code:
- `pkcs15-cert.c`: changed `parse_x509_cert` to handle extensions correctly. The code attempted to handle extensions assuming extensions were ordered. The only extension it handled was crl's, but the handling was wrong and I didn't find any actual use of the crl code. I've changed it to cache all the extensions and then provided accessors functions to read a specific extension. I needed this to read the key Usage, but the extension fetching code can work with any extension (though the caller will need to parse the result. I also added code that parses DN and returns a specifically requested DN component. I needed this to get the Common Name for the certificate Subject. This gives the token a 'unique' name rather than some generic name (like CAC-I or CAC-II). Both of these can be used to enhance the piv support as well. 

The patch is rebased on the current master and should apply. I noticed that there was some change in coding style (or finally applied?) in eb21c41e. Let me know if I should also reflect these changes or not. Any other feedback is welcomed.


Jakuje:
For the record, I fixed the complains from Travis CI, but I don't understand what is the AppVeyor complaining about.

mouse07410:
Nice! What tokend are you using, and did you take a look at https://github.com/mouse07410/OpenSC.tokend.git?

Jakuje:
I don't have OSX so I don't use tokend. We tested in RHEL7 and Fedora 23 and with direct PKCS#11 API and NSS (also PKCS#11 with some abstractions).

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1580 @@
...
+
+struct cac_object_list_entry {
+	cac_object_list_entry_t *next;
+	cac_object_t obj;
+};
+
+typedef struct cac_object_list {
+	cac_object_list_entry_t *head;
+	cac_object_list_entry_t *tail;
+} cac_object_list_t;
```
frankmorgner: There is already a list type in `simclist.h`. Please reuse this and avoid duplication!

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1580 @@
...
+}
+
+
+/*
+ * Set up the normal CAC paths
+ */
+#define CAC_STRING_AND_LENGTH(x)  x, sizeof(x)-1
+#define CAC_TO_AID(x) CAC_STRING_AND_LENGTH(x)
+#define CAC_TO_PATH_VALUE(x) CAC_STRING_AND_LENGTH(x)
+#define CAC_NULL_PATH { 0 }, 0
```
frankmorgner: These `defines` are a readability nightmare. And they don't even save a lot of time writing this out!

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1580 @@
...
+			return SC_ERROR_OUT_OF_MEMORY;
+		}
+		memcpy(*out_buf, buf, r);
+	}
+	*out_len = r;
+	return r;
+}
+
+/* create a fake tag/length file for cac1 cards based on the val_len */
+static int cac_cac1_get_cert_tag(sc_card_t *card, size_t val_len, u8 **tlp, size_t *tl_len_p)
```
frankmorgner: Please use the builtin asn1 parser in `asn1.h` to avoid duplication.

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1580 @@
...
+		tl[tl_len-2] = (u8)val_len & 0xff;
+		tl[tl_len-1] = (u8)(val_len >> 8) & 0xff;
+	}
+	*tlp = tl;
+	*tl_len_p = tl_len;
+	return SC_SUCCESS;
+}
+		
+/* get the tlv length return that len and return the length of the tl header. Don't overrun the buffer */
+static size_t cac_get_len(u8 *tl_ptr, size_t tl_len, size_t *tl_head_len)
```
frankmorgner: Again, please reuse the functionality of `asn1.h`

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1580 @@
...
+		outplen -= n;
+	}
+	free(rbuf);
+	rbuf = NULL;
+	r = outlen-outplen;
+
+err:
+	sc_unlock(card);
+	if (r < 0) {
+		bzero(out, outlen);
```
frankmorgner: you should call `sc_mem_clear` instead

src/libopensc/cards.h
```diff
@@ -231,6 +231,12 @@ enum {
 
 	/* JPKI cards */
 	SC_CARD_TYPE_JPKI_BASE = 31000,
+
+	/* CAC cards */
+	SC_CARD_TYPE_CAC_BASE = 31000,
```
frankmorgner: must be `32000`

src/libopensc/pkcs15-cac.c
```diff
@@ -0,0 +1,429 @@
...
+		return SC_ERROR_INVALID_CARD;
+	return SC_SUCCESS;
+}
+
+#define CAC_NUM_CERTS_AND_KEYS 10
+
+const char * cac_get_name(int type)
+{
+    switch (type) {
+    case SC_CARD_TYPE_CAC_I: return ("CAC I");
```
frankmorgner: If I'm not mistaken, you should return a pointer to `static` data!

src/libopensc/pkcs15-cac.c
```diff
@@ -0,0 +1,429 @@
...
+ * These could move to a helper file for other cards that wish to use usage as a way of getting flags
+ */
+
+/* Only certain usages are valid for a given algorithm, return all the usages that the algorithm supports so we
+ * can use it as a filter for all the public and private key usages */
+static unsigned int
+cac_alg_flags_from_algorithm(int algorithm)
+{
+	switch (algorithm) {
+#ifdef SC_ALGORITHM_RSA
```
frankmorgner: why are you wrapping this (and the following algorithims) into an `ifdef`? It is useless and makes the code less readable.

src/libopensc/pkcs15-cac.c
```diff
@@ -0,0 +1,429 @@
...
+	return 0;
+}
+
+
+/* These are the cert key usage bits that map to various PKCS #11 (and thus PKCS #15) flags */
+#define SC_CERT_USAGE_SIGNATURE 0x63ULL /* Signature, Non repudication, Cert Sign, Crl sign */
+#define SC_CERT_USAGE_DERIVE    0x10ULL /* Key agreement */
+#define SC_CERT_USAGE_UNWRAP	0x14ULL /* Key encipherment, Key agreement */
+#define SC_CERT_USAGE_DECRYPT	0x88ULL /* Data encipherment, encipher */
+#define SC_CERT_USAGE_NONREPUDIATION 0x02ULL
```
frankmorgner: Are you re-defining these usage values? If so, why don't you use the original ones?

src/libopensc/pkcs15-cac.c
```diff
@@ -0,0 +1,429 @@
...
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	memset(&serial, 0, sizeof(serial));
+
+	/* could read this off card if needed */
+
+	p15card->tokeninfo->label = strdup(cac_get_name(card->type)); 
+	p15card->tokeninfo->manufacturer_id = strdup(MANU_ID);
+	p15card->opts.use_file_cache = 1;
```
frankmorgner: `use_file_cache` is a user setting and should not be written here.

frankmorgner:
Quickly scrolling through the code, I saw some problems and some more duplicated code that is already present in OpenSC in other places (see comments above).

What I understood from PIV and CAC is that both are almost identical from a technical point of view. So in general, I wonder if we need to have the two implementations in OpenSC instead of improving the one implementation for PIV that we already have.

frankmorgner:
Your code doesn't build on Windows:
```
ctx.obj : error LNK2001: unresolved external symbol _sc_get_cac_driver
pkcs15-syn.obj : error LNK2001: unresolved external symbol _sc_pkcs15emu_cac_init_ex
```

dengert:
The CAC is the older proprietary US DOD smart card. PIV is the
    replacement for the CAC, and to be used by other federal agencies.
    As a transition CAC cards may include a PIV applet. 
    
https://gcn.com/articles/2005/12/05/defense-to-test-piv-iicompliant-access-cards.aspx
    
    But that was in 2005, and here we are in 11 years later.  Many CAC
    cards are dual CAC and PIV, for example
    https://github.com/mouse07410 has some.
    
    Most of the issues of https://github.com/mouse07410 are using a CAC
    card with tokend, not necessarily with the PIV driver. But it also
    appears that some dual CAC-PIV cards have issues with the default
    AID not being the PIV AID. These show up in a multi process
    environment where long running processes such as tokend, keychain, 
    Mozilla NSS are interfered with by command line utilities  resetting
    the card and the recovery processing in OpenSC. See #837, #836,
    #816, #624, #574 and #797  This is still an ongoing activity to
    detect what may have changed if some other process has changed state
    or reset the card.  The use of cards with multiple applets
    complicates this. 
    
    This might be helpful in understanding some of the issues. Its from
    2008:
http://www.cac.mil/docs/ref1.c.i-CAC_End-Point_Implementation_Guide_v1-22.pdf
    
    I do not have an CAC cards. I do not know if there are still CAC
    only cards in the field. So there may still be a need for a CAC
    driver. 
    The current PIV driver will claim the card if the PIV AID can be
    selected. So if a CAC driver is added, the order of the match_card
    to  drivers may be important. And even they there may be
    interference issues if one process wants to use the PIV driver and
    another the CAC driver. 
    
    
    On 7/25/2016 4:44 PM, Frank Morgner
      wrote:
    
    
      Quickly scrolling through the code, I saw some problems and
        some more duplicated code that is already present in OpenSC in
        other places (see comments above).
      What I understood from PIV and CAC is that both are almost
        identical from a technical point of view. So in general, I
        wonder if we need to have the two implementations in OpenSC
        instead of improving the one implementation for PIV that we
        already have.
      —
        thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

src/libopensc/card-cac.c
Jakuje: Fixed in 768912e

mouse07410:
> So in general, I wonder if we need to have the two implementations in OpenSC instead of improving the one implementation for PIV that we already have.

Probably yes - see below.

> I do not know if there are still CAC only cards in the field. So there may still be a need for a CAC     driver.

Exactly.

> The current PIV driver will claim the card if the PIV AID can be selected. So if a CAC driver is added, the order of the match_card to  drivers may be important. 

Yes!

> And even they there may be interference issues if one process wants to use the PIV driver and another the CAC driver.

I shudder, thinking of these possibilities. :-(

I can only hope that on Mac tokend would decide that the card is PIV or CAC, and stick to it. And that OpenSC CLI would follow the same logic to choose the same driver.

dengert:
On 7/26/2016 4:01 PM, Mouse wrote:
    
    
      
        So in general, I wonder if we need to have the two
          implementations in OpenSC instead of improving the one
          implementation for PIV that we already have.
      
      Probably yes - see below.
      
        I do not know if there are still CAC only cards in the field.
          So there may still be a need for a CAC driver.
      
      Exactly.
      
        The current PIV driver will claim the card if the PIV AID can
          be selected. So if a CAC driver is added, the order of the
          match_card to drivers may be important. 
      
      Yes!
      
        And even they there may be interference issues if one process
          wants to use the PIV driver and another the CAC driver.
      
      I shudder, thinking of these possibilities. :-(
      I can only hope that on Mac tokend would decide that the card
        is PIV or CAC, and stick to it. And that OpenSC CLI would follow
        the same logic to choose the same driver.
    
    
    Other non-OpenSC based programs might access the card too.  Doesn't
    the next MacOS have  replacement for tokend with a their own PIV
    middleware? 
    
    
      —
        You are receiving this because you commented.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

Jakuje:
The Windows build is passing now.

Notes to the recent comments:
 * The implementations are somehow similar on low level, but the operations and content is different (PIV = files x CAC - containers). Merging both drivers would result in a lot of conditions and affect the readability.
 * As far as we know, there are CAC-only cards (at least we have some).
 * The order of drivers can be changed to prioritize PIV over CAC (to preserve current behavior for dual cards), if needed.

I will try to address soon the other comments on the code.

mouse07410:
> Other non-OpenSC based programs might access the card too.  Doesn't the next MacOS have  replacement for tokend with a their own PIV middleware?

It is my understanding that yes, the next MacOS "somehow" accesses tokens and makes them available as keychains. Therefore all the applications that could access keys stored in keychains (such as Keychain Access, Apple Mail, etc. etc., and probably including Firefox and such!) would be able to use keys/certs on the token without any extra software.

AFAICT, this would leave only OpenSSL that does not use keychains at all, and either accesses key files directly, or relies upon `libp11` to access tokens. I wonder if this wouldn't require a significant re-design - replacing `libp11` with `libkeychain` for Mac.

> The order of drivers can be changed to prioritize PIV over CAC (to preserve current behavior for dual cards), if needed.

I would like to insist on this.

> Merging both drivers would result in a lot of conditions and affect the readability.

I agree.

dengert:
On 7/27/2016 2:26 AM, Jakuje wrote:
    
    
      The Windows build is passing now.
      Notes to the recent comments:
      
        The implementations are somehow similar on low level, but
          the operations and content is different (PIV = files x CAC -
          containers). Merging both drivers would result in a lot of
          conditions and affect the readability.
      
    
    
    The PIV does not have files as ISO7816 would define them. 
    
    NIST 800-73 lists a set of APDUs that appear to be a subset of
    ISO7816 It defines command chaining,  get response and short APDUs. 
    The GET DATA command  00:CB:3F:FF command is used to read objects.
    This is an ISO7816 even number get data command with P1-P2= 3F:FF.
    The only use of a "file" is with the SELECT command used to select
    the AID. It is defined as  00:A4:04:00 which is a ISO7816  SELECT by
    P1=04(DF name) P2=00(First occurrence) 
    The pkcs15-piv.c , emulates the ISO7816 select and read commands so
    it can run in the OpenSC PKCS#15 environment, and uses the
    card-piv.c to use the GET DATA command to retrieve the objects. A
    certificate object contains the certificate and flags that indicate
    if the certificate is stored as DER or GZIPED DER. 
    
    NIST 800-73 only requires the card vendor to support the commands 
    defined in 800-73. The PIV driver was written to only support the
    commands defines in 800-73.  (There are a few commands supported for
    creating test cards, but 800-73 leaves up to the card vendors any
    card provisioning commands needed by a card management system. The
    PIV driver was written for user access to the already provisioned
    cards.)
    
    NIST 800-73 also define the key usage of each certificate (and key).
    It is the responsibility of the card issuer to have the attributes
    in the certificate match what is defined in NIST 800-73.  From your
    comments it sounds like the CAC expects to obtain the key usage info
    from the certificates. 
    
    
      
        As far as we know, there are CAC-only cards (at least we
          have some).
        The order of drivers can be changed to prioritize PIV over
          CAC (to preserve current behavior for dual cards), if needed.
      
      I will try to address soon the other comments on the code.
      —
        You are receiving this because you commented.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

dengert:
On 7/27/2016 7:24 AM, Mouse wrote:
    
    
      
        Other non-OpenSC based programs might access the card too.
          Doesn't the next MacOS have replacement for tokend with a
          their own PIV middleware?
      
      It is my understanding that yes, the next MacOS "somehow"
        accesses tokens and makes them available as keychains. Therefore
        all the applications that could access keys stored in keychains
        (such as Keychain Access, Apple Mail, etc. etc., and probably
        including Firefox and such!) would be able to use keys/certs on
        the token without any extra software.
      AFAICT, this would leave only OpenSSL that does not use
        keychains at all, and either accesses key files directly, or
        relies upon libp11 to access tokens. I wonder if
        this wouldn't require a significant re-design - replacing libp11
        with libkeychain for Mac.
    
    
    That is sort of out of scope for OpenSC. One would need an OpenSSL
    engine using keychain. It looks like there is an  openssl engine
    based on CAPI for Windows.  So why not one on MacOS using keychain
    or whatever the replacement for tokend is?
    
    
    
      
        The order of drivers can be changed to prioritize PIV over
          CAC (to preserve current behavior for dual cards), if needed.
      
      I would like to insist on this.
      
        Merging both drivers would result in a lot of conditions and
          affect the readability.
      
      I agree.
      —
        You are receiving this because you commented.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
    
    
    -- 

mouse07410:
>> AFAICT, this would leave only OpenSSL that does not use keychains at all, and either accesses key files directly, or relies upon libp11 to access tokens. I wonder if this wouldn't require a significant re-design - replacing `libp11` with `libkeychain` for Mac.
> 
> That is sort of out of scope for OpenSC. 

Probably not more out of scope than `libp11` is...

> One would need an OpenSSL engine using keychain. 

Yes. And for backwards compatibility sake it should probably be called "pkcs11"...

> It looks like there is an  openssl engine based on CAPI for Windows.  So why not one on MacOS using keychain or whatever the replacement for tokend is?

Yeah... But who'd write it, and when?

src/libopensc/card-cac.c
Jakuje: I rewrote the lists in f03bff0181 to use the `simclist.h`. The code is a bit simplified:
>  3 files changed, 57 insertions(+), 76 deletions(-)

Comments welcomed.

src/libopensc/card-cac.c
Jakuje: AFAIK, we can't get rid of this one simply. The CAC have LT (Tags and Lengths) and V (Values) stored separately and the [`cac_read_binary()` ](https://github.com/OpenSC/OpenSC/pull/841/files#diff-848b13147a344ba2c6361d91ca77feb1R569) function does the interleaving.

I didn't find in your `asn1.h` library function to read TL separately from the Values.

src/libopensc/cards.h
Jakuje: Thanks. I missed this during the rebase to the master (with already a new card). Fixed in 56c164986

src/libopensc/pkcs15-cac.c
Jakuje: Fixed in 9eeda4da

src/libopensc/pkcs15-cac.c
Jakuje: The `RSA`, `DSA`, `GOSTR3410` are defined always in `opensc.h` now, if I see right. But not the `SC_ALGORITHM_DH`. The usage is probably stemming from similar case in `pkcs15-algo.c` for possible future use cases. I will clean up this.

src/libopensc/pkcs15-cac.c
Jakuje: which original ones? You mean the values in the `pkcs15.h`? They are PKCS15 usage values. These are usage values of a certificate and I don't see them defined  anywhere else.

src/libopensc/pkcs15-cac.c
Jakuje: It looks like that. I removed this line.

I am wondering, if there should not be some manual page for the `opensc.conf`. Its syntax is explained in the installed `/etc/opensc-x86_64.conf`, but having an formal overview would be plus for new users.

FYI, also `pkcs15-actalis.c` is setting this option for some reason.

src/libopensc/pkcs15-cac.c
dengert: There are 3 sets of usage flags. X509, PKCS#11 and PKCS#15 You could look
at how the card-piv.c and pkcs15-piv.c handle the usage flags. NIST 800-73
defines what the are required to be for PIV certs. I think the SC_CERT_
look like X509 keyUsage  better check this.

I would also not expect a CAC card to support GOST,

src/libopensc/card-cac.c
Jakuje: The above encoding is `simple TLV`, unlike the BER/DER encoding used in `asn1.h` and therefore I am not able to simplify the encoding using the above library. Though it might be possible to use it in the decoding.

Example of difference between BER TLV and Simple TLV can be found in an Appendix C of [CAC End-Point Implementation Guide](http://www.cac.mil/docs/ref1.c.i-CAC_End-Point_Implementation_Guide_v1-22.pdf) (linked from previous comments).

src/libopensc/pkcs15-cac.c
Jakuje: @dengert The PIV driver have the usage flags set based on the ID and on a hard-coded list in `pkcs15-piv.c` (knows all its possible objects that can be on the card and their attributes). There is no such table for CAC and the usage flags are derived from the X509 key usage flags.

This is what is one of the differences between PIV and CAC. It can be theoretically used in other cards, if we will move these constants and helper functions out of `pkcs15-cac.c`, but it would need somebody with a bit better overview about the whole project.

src/libopensc/card-cac.c
Jakuje: Actually the Simple TLV is used also here, which is a reason why the `asn1.h` is not useful (the length parsing is a bit different).

src/libopensc/card-cac.c
Jakuje: Yes, having them written three times is probably not the most understandable thing. Even though they are used in different context, they do the same. I can squash them into single define, if you would insist.

src/libopensc/pkcs15-cac.c
Jakuje: > These could move to a helper file for other cards that wish to use usage as a way of getting flags ...

The point of having there also GOST, DSA is if it would get used by other drivers. If not, we can remove that.

src/libopensc/pkcs15-cac.c
nmav: @frankmorgner why return a pointer to static data instead of constant data? The latter is certainly better as multiple or consequent calls will not overwrite any previously returned data (given the way the function is used today that wouldn't matter, but as a function return value constant pointers are certainly better than static buffers)

Jakuje:
@frankmorgner Frank, I addressed most of your remarks and comment on the ones that need some consideration. Could you please review them?

To sum them up here:
 * Usage of `asn1` parser is not possible, because the data we handle are in Simple TLV. This is now explained in the comments not to confuse.
 * I can squash the `#defines`, but I am not sure how much readability it will add.
 * The flags `SC_CERT_USAGE_` in question are certificate flags mapping to the PKCS#11 flags, possibly useful in other cards. If you have them somewhere else, please point me in that direction. I was not able to find a definition of them. This is related to the `#ifdefs` and GOST lines. If it will not be considered useful, we can drop everything except RSA
 * various build issues in the CI were fixed.

The commits are separate. I will squash them before the merge.

frankmorgner:
I found a nice summary for BER/DER vs Simple TLV http://stackoverflow.com/questions/18853800/simlpe-tlv-vs-ber-tlv. As Simple TLV is defined by ISO 7816-4, I guess your CAC implementation could be reused in other places. Maybe we get some more feedback from others (in general and) whether the simple tlv parser would be used for the common part in libopensc.

src/libopensc/pkcs15-cac.c
frankmorgner: @nmav it's undefined when going out of scope

nmav: I think you misread the code. The returned value is not a variable on the stack that goes out of scope, but a pointer to a constant string (stored in the global data section).

frankmorgner: you're right.

mouse07410:
> I guess your CAC implementation could be reused in other places...

I concur.

dengert:
I would be interested in the pkcs15-cert.c and pkcs15.h  being in a separate PR. I would like to get the keyUsage flags to use with #847 so the Yubico tokens could use the retired key management keys for any purpose as defined in the matching certificate.

Jakuje:
@dengert I created PR #852  with the changes in `pkcs15-cert.c` and `pkcs15.h` to make it more digestible. Let me know if you are interested in the part `cac_map_usage()` function, currently in the `pkcs15-cac.c`, which is a bit related to this one.
Once it will be merged, I will drop it from this PR. 

@frankmorgner @mouse07410 Do you have any examples how would you like to use them? I can extract existing function into new, lets say `/src/libopensc/simpletlv.{h,c}` and call them `simpletlv_get_len()` and some simple tag composer `simpletlv_create_tag_length()`.
There still might be a difference between CAC (using separate files TL and V) and other cards, if I see right.

mouse07410:
> Do you have any examples how would you like to use them?

Not offhand. I like the idea of reusing your functions in general, but nothing specific in mind.

Jakuje:
I moved the function `lebytes2ushort()` to `internal.h` header file to be used by other drivers (there are already conversion functions for the other ways).
I created `simpletlv.c` with the above discussed functions used by CAC so far and made them a bit more generic than before.

src/libopensc/pkcs15-cac.c
rjrelyea: re: GOST, yeah, I've really coded the function so that it could be pulled out and used for other cards. There's nothing really cac specific about the choice of PKCS15 flags from the actual cert usage. For any card that doesn't carry actual usage, getting the key usage from the associated certificate seems reasonable.

The usages are X509 usages, but the bit decoder doesn't reverse the bits coming out of it, so the actual flags are reversed from the flags defined in pkcs15-init.h. It might be worth fixing the decoder and using the flags defined there. The other thing about the defines is they are or'ed values from the usage, so that if either key encipherment or key agreement is set, then UNWRAP is set.

dengert: Can you change the names of the SC_CERT_USAGE_* defines or remove them? There are at least 3 different definitions of key usage
https://tools.ietf.org/html/rfc5280#section-4.2.1.3

KeyUsage ::= BIT STRING {
           digitalSignature        (0),
           nonRepudiation          (1), -- recent editions of X.509 have
                                -- renamed this bit to contentCommitment
           keyEncipherment         (2),
           dataEncipherment        (3),
           keyAgreement            (4),
           keyCertSign             (5),
           cRLSign                 (6),
           encipherOnly            (7),
           decipherOnly            (8) }

pkcs#15 v1.1 
   KeyUsageFlags ::= BIT STRING {
   encrypt (0),
   decrypt (1),
   sign (2),
   signRecover (3),
   wrap (4),
   unwrap (5),
   verify (6),
   verifyRecover (7),
   derive (8),
   nonRepudiation (9)

and PKCS#11 usage returns different attributes for pubkey and privkey. (But the pkcs15 tp pkcs11 framework takes care of these mappling from pkcs15 flags.)

I am also surprised that the CAC card would have support SC_ALGORITHM_GOSTR3410.

It also looks like your set every possible key use bits based on the algorithm, rather then based on policy usage.  (PIV mandates specific key usage for all its keys) or key usage based on the certificate. (PIV expects the CMS to follow the policies set in NIST 800-73 when creating certificates.

Defines usually have some indication if they refer to PKCS15, PKCS11 or for a specific driver.

dengert:
During these conversations some questions have been asked about do we need separate PIV and CAC drivers. Or could they be combined. Keep in mind that this has been asked before by others, and there are dual CAC cards that support both the PIV and the CAC API (i.e. AID application) which can use the same or different keys and certificates from the card. 

- But one is not a subset of the other. CAC has some objects on the card not accessible from the PIV API for example.

- PIV uses GET_DATA to access its objects. CAC uses more standard iso7816 commands. 

- PIV objects use ASN1 tag and length it does not use simpleTLV as CAC does.  

- PIV has specific key usage requirements as set by policies in the specs. I assume CAC has its own policies.   

I would vote for two separate drivers. There are enough technical and policy differences, that the drivers should be different. 

In the case of a dual PIV CAC  cards which driver should be used?  The choice is really up to the user (or software application) which driver should be used.  OpenSC has not addressed this very well. This is further complicated by how OpenSC picks a driver for any card. OpenSC can match on the ATR or let a driver try and read a file or select an AID.  But it turns out with some cards, an attempt to select an invalid AID for that card will reset the security state of the card. Some older PIV, CAC and Yubico cards have this problem.  PIV specs say that should not happen. I don't know what the CAC specs say about this. 

Having separate drivers would allow selection of a driver by the user/admin via a custom opensc.conf file.  

I do know that the PIV specs were defined to simplify and standardize smart cards for the U.S. government and replace CAC cards. But it has been over 10 years now and there still appears to be a need for a CAC driver. There are opens source and commercial PIV and CAC drivers other then OpenSC. Microsoft has  builtin PIV driver. Apple has (or had) PIV and CAC drivers. 

NIST 800-73-4 also defines additional new features, such as optional Secure Messaging, which can be used over contactless connection. It also discusses on card fingerprint matching  in place of a PIN. Oberthur has such a card and reader.  I did not have a chance to look at it closely before I retired and don't have any test cards that support either SM or on card fingerprint matching.  (Don't know about future CAC standards.)

So I believe there should be separate CAC and PIV driver, as long as CAC only cards or applications  are still around.

src/libopensc/pkcs15-cac.c
Jakuje: > It also looks like your set every possible key use bits based on the algorithm, rather then based on policy usage.

You mean the function `cac_alg_flags_from_algorithm()`? This is just a default mask, that is later filtered by X.509 Key Usage Extension (the PR #852) and final results are different for public and private key in `cac_map_usage()`.

> Can you change the names of the SC_CERT_USAGE_* defines or remove them?

I would probably be for the name `CAC_X509_USAGE_*`and leave it in CAC driver, or to  `SC_X509_USAGE_*` and moving these constants to the PR #852 too, because they are related to the work done there. Problem of reusing the `pkcs15-init.h` constants is that we have the bits reversed, as mentioned by @rjrelyea in previous comment.
Reusing the bits from there (and adding missing bits) would add readability. Though I am not sure how to handle the bit order correctly in this case (if it should take place in here or in #852 and in one way or another). I will take with Bob about that.

> I am also surprised that the CAC card would have support SC_ALGORITHM_GOSTR3410.

There was also idea about reuse in the other drivers in future as confirmed by @rjrelyea comment above.

Jakuje:
I rebased the current set of patches against the current master (after applying #852, thanks @viktorTarasov). It applies and build fine now again.

Jakuje:
The topics in question seems to be resolved, the functionality verified and we agreed on having two separate drivers is the way to go.
Is there anything blocking this pull request? Anything I can do (squashing commits, clean up something, modify logging functions, whatever, ..)? Can we proceed to apply this PR?

dengert:
I would try and squash into 2 or 3 commits or PRs, that could be committed one at a time thus helping to narrow down and any problems. 

Cosmetic: I see one of the files changed has only a spelling mistake. 

Support routines in base code: Like the simpleTLV changes, and any base routines that were modified.

Code to add: the CAC driver, including code and configure changes to compile and use it. 

But its really up to @frankmorgner and @viktorTarasov

src/libopensc/card-cac.c
frankmorgner: Actually what I meant was that for readability it is better to remove the four defines completely and to instead duplicate the code fragment every time.

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1540 @@
...
+/* NOTE: returns a bool, 1 card matches, 0 it does not */
+static int cac_match_card(sc_card_t *card)
+{
+	int r;
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	/* Since we send an APDU, the card's logout function may be called...
+	 * however it may be in dirty memory */
+	card->ops->logout = NULL;
+
+	r = cac_find_and_initialize(card);
```
frankmorgner: `cac_find_and_initialize` returns success if the card driver has already been initialized:
```
+	/* already found? */
+	if (card->drv_data) {
+		return SC_SUCCESS;
+	}
``
This, however, doesn't mean that the current card actually *is a cac*. I suggest you initialize the data structures in `cac_init` only.

File: src/libopensc/card-cac.c [hunk 12]
```diff
@@ -0,0 +1,1540 @@
...
+	cac_ops.compute_signature = cac_compute_signature;
+	cac_ops.decipher =  cac_decipher;
+	cac_ops.card_ctl = cac_card_ctl;
+	cac_ops.pin_cmd = cac_pin_cmd;
+
+	return &cac_drv;
+}
+
+
+#if 1
```
frankmorgner: should be removed

src/libopensc/simpletlv.h
```diff
@@ -0,0 +1,50 @@
...
+int simpletlv_create_tag_length(u8 tag, size_t val_len, u8 *tlp, size_t *tl_len_p);
+
+/* get the Simple TLV length return that len and return the length of the TL
+ * header. Don't overrun the buffer.
+ * @param  tl_ptr      Pointer to the TL file
+ * @param  tl_len      The length of TL file
+ * @param  tl_head_len The length of parsed header
+ * @return             The length from TL file
+ */
+size_t simpletlv_get_len(u8 *tl_ptr, size_t tl_len, size_t *tl_head_len);
```
frankmorgner: Your simpletlv implementation has some limitations, which is OK. But could you please match the API to those of `sc_asn1_put_tag` and `sc_asn1_read_tag`?

src/libopensc/card-cac.c
```diff
@@ -0,0 +1,1540 @@
...
+	}
+
+	return  iso_drv->ops->pin_cmd(card, data, tries_left);
+}
+
+static struct sc_card_operations cac_ops;
+
+static struct sc_card_driver cac_drv = {
+	"CAC",
+	"cac",
```
frankmorgner: I think this should be `"CAC"`. please check if `opensc-tool --list-drivers` shows the correct short name for forcing the card driver in `opensc.conf`.

frankmorgner:
Thanks for fixing the previous remarks, @Jakuje. If you look into the four new ones, I think this PR is good to be merged.

src/libopensc/card-cac.c
Jakuje: We can simply get rid of `CAC_TO_PATH_VALUE` which is used only once and `CAC_NULL_PATH`, which is just static replacement (committed already).

But I am still not completely convinced about the `CAC_TO_AID` and how to make it "better". I can rewrite

    { CAC_TO_AID(CAC_1_RID "\x01\x00") }

to

    #define CAC_AID_LEN 7
    { CAC_1_RID "\x01\x00", CAC_AID_LEN }

with hard-coded length would be more error-prone in case we would do some modification in these paths. Frank, would this work for you or could we agree on leaving this one in the code in current state?

src/libopensc/card-cac.c
Jakuje: This is present in most of the other card drivers (for some historic reasons?) and therefore it was left here. If it is not needed anymore, I will remove that.

src/libopensc/card-cac.c
Jakuje: It is tricky to match with other cards. The case differs among cards:

    ./opensc-tool --list-drivers
    Configured card drivers:
    [...]
      atrust-acos      A-Trust ACOS cards
      cac              CAC
      PIV-II           PIV-II  for multiple cards
    [...]

but the long name should probably mention "Common Access Card (CAC)" on the first line (committed) and the second should be left lowercase as it is -- all the cards except PIV are lowercase.

dengert:
See
      https://github.com/OpenSC/OpenSC/issues/818
     https://github.com/OpenSC/OpenSC/pull/828 
    for why PIV-II is upper case. 
     
    
    
    On 9/16/2016 2:23 AM, Jakuje wrote:
    
    
      @Jakuje commented on this pull request.
      
      In src/libopensc/card-cac.c:
      > +		}
+		for(i=0; i < data->pin2.len; ++i) {
+			if (!isdigit(data->pin2.data[i])) {
+				return SC_ERROR_INVALID_DATA;
+			}
+		}
+	}
+
+	return  iso_drv->ops->pin_cmd(card, data, tries_left);
+}
+
+static struct sc_card_operations cac_ops;
+
+static struct sc_card_driver cac_drv = {
+	"CAC",
+	"cac",

      It is tricky to match with other cards. The case differs among
        cards:
      ./opensc-tool --list-drivers
Configured card drivers:
[...]
  atrust-acos      A-Trust ACOS cards
  cac              CAC
  PIV-II           PIV-II  for multiple cards
[...]

      but the long name should probably mention "Common Access Card
        (CAC)" on the first line (committed) and the second should be
        left lowercase as it is -- all the cards except PIV are
        lowercase.
      —
        Reply to this email directly, view it on
          GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/OpenSC/OpenSC","title":"OpenSC/OpenSC","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/OpenSC/OpenSC"}},"updates":{"snippets":[{"icon":"PERSON","message":"@Jakuje reviewed #841"}],"action":{"name":"View Pull Request","url":"https://github.com/OpenSC/OpenSC/pull/841"}}}
    
    
    -- 

Jakuje:
@dengert Thank you for the clarifying comment. Yes, there was some inconsistency in the naming across the changes which is fixed now.
If I understand it correctly, there is not a requirement to have it uppercase, but to have it the same way in `card-piv.c`, `ctx.c` and `pkcs15-syn.c`. At the moment, all of them is uppercase.

src/libopensc/simpletlv.h
Jakuje: I reimplemented the function into the `simpletlv_read_tag(*tl_ptr, tl_len, &tag, &len);` and `simpletlv_put_tag(u8 tag, size_t datalen, u8 *out, size_t outlen, u8 **ptr);` (similar interface as the ASN) and prefixed them with `sc_` to match with the asn1.
It worked fine both with emulator and the cards.

src/libopensc/card-cac.c
Jakuje: Good catch. Is there any reason why the `cac_match_card()` would get called twice with the same context or the `cac_init()` before `cac_match_card()`? From my testing, I always saw them called in the "correct" order.
The problem is that we detect whether it is a  CAC or a CACII card in this function and we need this knowledge in the initialization phase.

src/libopensc/card-cac.c
```diff
@@ -1499,7 +1499,7 @@ static struct sc_card_operations cac_ops;
 
 static struct sc_card_driver cac_drv = {
 	"Common Access Card (CAC)",
-	"cac",
+	"CAC",
```
frankmorgner: I'm sorry to be pedantic about this. Would it be possible to change the shortname to lowercase? The PIV's short name is only uppercase, because we need to cope with the mistakes of the past.

Jakuje: Sure. That was the point of my comment to make sure what is the expected case (lower one) and where else I should align the name. If the above list of occurrences is complete (`card-piv.c`, `ctx.c` and `pkcs15-syn.c`), then it will be fixed in the commit b1779838.

frankmorgner: thanks!

src/libopensc/card-cac.c
frankmorgner: It may be true now that match_card() is only called once. But who knows what will happen in the future. Better keep the implementation similar to what a naive developer would expect ;-)

But you can be sure that `match_card` has been called prior to `init`, see a7a903fd. This means that you could set `card->type` in `match_card` and use this value in `init`.

src/libopensc/simpletlv.c
```diff
@@ -0,0 +1,91 @@
...
+	if (outlen < 2 || (outlen < 4 && datalen >= 0xff))
+		return SC_ERROR_INVALID_ARGUMENTS;
+
+	*p++ = tag; /* tag is single byte */
+	if (datalen < 0xff) {
+		/* short value up to 255 */
+		*p++ = (u8)datalen; /* is in the second byte */
+	} else if (datalen < 0xffff) {
+		/* longer values up to 65535 */
+		*p++ = (u8)0xff; /* fisrt byte is 0xff */
```
frankmorgner: typo

src/libopensc/simpletlv.c
```diff
@@ -0,0 +1,91 @@
...
+ * content length.
+ */
+int
+sc_simpletlv_put_tag(u8 tag, size_t datalen, u8 *out, size_t outlen, u8 **ptr)
+{
+	u8 *p = out;
+
+	if (outlen < 2 || (outlen < 4 && datalen >= 0xff))
+		return SC_ERROR_INVALID_ARGUMENTS;
+
```
frankmorgner: Could you check for `tag != 0xFF && tag != 0x00`, which are reserved values?

src/libopensc/simpletlv.c
```diff
@@ -0,0 +1,91 @@
...
+	*tag_out = *p++;
+	len = *p++;
+	if (len == 0xff) {
+		len = 0;
+		/* don't crash on bad data */
+		if (buflen >= 4) {
+			len = lebytes2ushort(p);
+			p++;
+		}
```
frankmorgner: shouldn't we throw an error if it is badly encoded?

src/libopensc/simpletlv.c
Jakuje: That makes sense, according to the [specification](http://geek-mag.com/posts/251450/).

src/libopensc/simpletlv.c
Jakuje: I missed this one. Thanks.

src/libopensc/simpletlv.c
Jakuje: We return `taglen == 0`. Error would be more appropriate I guess in the library-like code. Modified now also with the checks.

src/libopensc/card-cac.c
Jakuje: That makes sense. Thanks for patience :)
Fixed now and the argument (we need a bit more context (slot id) so the detection is done twice in this case. But seems cleaner than copying the code back and forth.

frankmorgner:
Looks good now, thanks. Could you squash all CAC related commits into a single one?

Jakuje:
The bunch of changes are squashed into few thematic commits (first to preserve original author, the later addressing the modification based on this pull request, the last one for the OIDs in the commit for #852).
The travis failure does not look like my fault.

frankmorgner:
@viktorTarasov could you please review the changes?

nmav:
@frankmorgner @viktorTarasov is there some way I can help to get that included?

Jakuje:
The PR was rebased on the current master to resolve conflicts.

Jakuje:
@frankmorgner @viktorTarasov Is there anything more needed to do in this way?

frankmorgner:
@viktorTarasov insists on having a final look, that's why I'm not merging this PR. I approved these changes on 21 Sep 2016, but I saw you rewrote the commits since then. I'll check the changes later again.

Jakuje:
Thank you for an answer. The only changes were rebases against current master, resolving conflicts and problems found because of different work (example same as in Coolkey: 3a9dc2e or what we agreed on 987179d -- drivers order).

I understand this decision (though it was not clearly stated before). @viktorTarasov do you have some estimation when it can be?

frankmorgner:
@Jakuje looks good for me. Still waiting for review from @viktorTarasov

frankmorgner:
@viktorTarasov if you have no objections, I'll merge the code in a couple of days.

PR #843, opened 2016-08-02
install opensc.pc

nmav:
This mainly allows applications to detect the installed opensc version, as well as link against opensc's pkcs11 library if needed. This also deletes the unused since long time libopensc.pc.in.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -795,11 +795,11 @@ AC_CONFIG_FILES([
 	src/Makefile
 	src/common/Makefile
 	src/libopensc/Makefile
-	src/libopensc/libopensc.pc
 	src/libsm/Makefile
 	src/pkcs11/Makefile
 	src/pkcs11/versioninfo-pkcs11.rc
 	src/pkcs11/versioninfo-pkcs11-spy.rc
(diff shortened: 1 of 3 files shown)


frankmorgner:
1. Checking some specific detail of `opensc-pkcs11.so` defeats the purpose of PKCS#11, which is to offer an abstraction from the implementation. It does not make sense to install this. (Deleting libopensc.pc, however is OK.)
2. If included nevertheless, the file should be named `opensc-pkcs11.pc`

nmav:
> It does not make sense to install this. 

It does make sense to install opensc for applications to be able to check opensc's installed version. I run into this for libp11.

> Checking some specific detail of opensc-pkcs11.so defeats the purpose of PKCS#11

It is perfectly legitimate to link with a pkcs11 library (caml-crush for example links with pkcs11 libraries). I can remove the linking information if you want to be strict about it.

frankmorgner:
applied with some fixes in 2059f79793f0ed9a89a29f400f6e02b5b7c49f59

nmav:
Thank you

PR #844, opened 2016-08-03
Fixing too small buffer

llogar:
Fixes too small buffer introduced with commit c5a95d4a5ae93e5960ee19988d86246aa688f453


frankmorgner:
fixed in 94bedaab9358a1ff66840168f737c4568548c00d

Issue #847, opened 2016-08-06
OpenSC PIV doesn't expose all available Yubikey 4 slots


titanous:
### Expected behaviour

When using the OpenSC PIV module, it should provide access to the 20 "retired" slots on the Yubikey 4. These slots don't appear to have any limitations and can be used in the same fashion as the other four slots on the device.
### Actual behaviour

The 20 slots are not displayed when using the PIV module provided by OpenSC, but do appear when using the YKCS11 module from Yubico.
### Logs

https://gist.github.com/titanous/57a8993894d457cadf8d2057f036d112

dengert:
NIST 800-73-3 defines the Key History Object which has a
keysWithOnCardCerts and keysWithOffCardCerts These say how many of the keys
and certs are on the crd and how many have keys on the card with the
certificate off the card and a  URL to get to the offline certificates.

The OpenSC PIV driver is following the NIST specs and tries to read the Key
History Object to find out how many of the retired keys and certs are on
the card.

(The OpenSC PIV driver will read from the user's home directory the offline
certs
if the file from the URL has been downloaded. It has to do this because the
only way to find the public key from the cert so it can get the key type
and key size needed to use the key.)

Note the key usage for these keys is for key management not signatures, The
PIV driver sets the key usage accordingly.

The proper way to deal with is is add a KeyHistoryObject to the card.
To see Key History Object  in:

http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf






On Sat, Aug 6, 2016 at 3:17 PM, Jonathan Rudenberg <notifications@github.com

titanous:
The Yubikey 4 allows using all 24 keypairs/certificates for any purpose, and does not expose a Key History Object. Since I don't work at Yubico and the devices are not flashable, there is no way to expose the "Key History Object". Thus in order for the additional 20 keys to be accessible to OpenSC's PIV driver it would need to be modified to support the way the Yubikey does things.

dengert:
The Yubico-piv-tool can be used to write objects. One just has to create a
binary file with teh contents of the object. I am not able to give you all
the details,  I am on vacation. But look at
https://github.com/OpenSC/OpenSC/issues/479
This has example of writing some other objects,

NIST 800-73 defines the usage, and the keys must be in order. So this is
another example of Yubico not following the PIV standard!




On Sat, Aug 6, 2016 at 4:56 PM, Jonathan Rudenberg <notifications@github.com

[...]

PR #852, opened 2016-08-15
change  parse_x509_cert()  to handle extensions correctly.

Jakuje:
The code attempted to handle extensions assuming extensions were ordered. The
only extension it handled was crl's, but the handling was wrong and I didn't
find any actual use of the crl code. I've changed it to cache all the extensions
and then provided accessors functions to read a specific extension. I needed this
to read the key Usage, but the extension fetching code can work with any extension
(though the caller will need to parse the result. I also added code that parses DN
and returns a specifically requested DN component. I needed this to get the Common
Name for the certificate Subject. This gives the token a 'unique' name rather than
some generic name (like CAC-I or CAC-II). Both of these can be used to enhance the
piv support as well.

(Chunk of #841 providing this functionality to other card)

### DIFF
--- a/src/libopensc/pkcs15-cert.c
+++ b/src/libopensc/pkcs15-cert.c
@@ -136,12 +136,12 @@ parse_x509_cert(sc_context_t *ctx, struct sc_pkcs15_der *der, struct sc_pkcs15_c
 /* get a component of a Distinguished Name (e.i. subject or issuer) from an oid tag. dn can be either
  * cert->subject or cert->issuer. dn_len would be cert->subject_len or cert->issuer_len.
  * common types:
- *   CN: type[] = {0x55, 0x04, 0x03} type_len = 3;
- *   Country: type[] = {0x55, 0x04, 0x06} type_len = 3;
- *   L: type[] = {0x55, 0x04, 0x07} type_len = 3;
- *   S: type[] = {0x55, 0x04, 0x08} type_len = 3;
- *   O: type[] = {0x55, 0x04, 0x0a} type_len = 3;
- *   OU: type[] = {0x55, 0x04, 0x0b} type_len = 3;
+ *   CN:      struct sc_object_id type = {{85, 4, 3, -1}};
+ *   Country: struct sc_object_id type = {{85, 4, 6, -1}};
+ *   L:       struct sc_object_id type = {{85, 4, 7, -1}};
+ *   S:       struct sc_object_id type = {{85, 4, 8, -1}};
+ *   O:       struct sc_object_id type = {{85, 4, 10, -1}};
+ *   OU:      struct sc_object_id type = {{85, 4, 11, -1}};
  * if *name is NULL, sc_pkcs15_get_name_from_dn will allocate space for name.
  */
 int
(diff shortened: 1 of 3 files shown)


dengert:
OpenSC uses "struct sc_object_id" to store and compare OIDs. Can you change your extension code to  use "const struct sc_object_id *oid" rather "const u8 *type" This is more inline with the rest of the OpenSC handling of OIDs. 

This would require lines like these  to be changed:

u8 *val, *oid;
{ "extensionType", SC_ASN1_OCTET_STRING, SC_ASN1_TAG_OBJECT, SC_ASN1_ALLOC, &oid, &oid_len },

to something like:
struct sc_object_id  oid;
{ "x509v3 entry OID",       SC_ASN1_OBJECT,  SC_ASN1_TAG_OBJECT , 0, &oid, 0 },

and then use the sc_compare_oid to compare. 


Can you change the sc_pkcs15_get_extension routine to accept a "struct sc_object_id" rather then the type?

Jakuje:
@dengert I rewrote the arguments to the `struct sc_object_id` and also rewrote the #852 to use this new interface. I hope I got the idea you explained in the previous comment right.
So far tested with cac emulator and it worked fine.

src/libopensc/pkcs15-cert.c
```diff
@@ -141,6 +133,195 @@ parse_x509_cert(sc_context_t *ctx, struct sc_pkcs15_der *der, struct sc_pkcs15_c
...
 
+/* get a component of a Distinguished Name (e.i. subject or issuer) from an oid tag. dn can be either
+ * cert->subject or cert->issuer. dn_len would be cert->subject_len or cert->issuer_len.
+ * common types:
+ *   CN: type[] = {0x55, 0x04, 0x03} type_len = 3;
+ *   Country: type[] = {0x55, 0x04, 0x06} type_len = 3;
+ *   L: type[] = {0x55, 0x04, 0x07} type_len = 3;
+ *   S: type[] = {0x55, 0x04, 0x08} type_len = 3;
+ *   O: type[] = {0x55, 0x04, 0x0a} type_len = 3;
+ *   OU: type[] = {0x55, 0x04, 0x0b} type_len = 3;
```
dengert: The comments should be changed to use sc_object_id way to list a OID.  
For example:
  struct sc_object_id keyUsage_oid={{2,5,29,15,-1}};

src/libopensc/pkcs15-cert.c
```diff
@@ -141,6 +133,195 @@ parse_x509_cert(sc_context_t *ctx, struct sc_pkcs15_der *der, struct sc_pkcs15_c
...
+	LOG_FUNC_RETURN(ctx, SC_ERROR_ASN1_OBJECT_NOT_FOUND);
+}
+
+/* get a specific extension from the cert. The extension is identified by it's oid value. NOTE:
+ * extensions can occur in any number or any order, which is why we can't parse them with a single
+ * pass of the asn1 decoder. If is_critical is supplied, then it is set to 1 if the extention is critical and 0
+ * if it is not. The data in the extension is extension specific. The following are
+ * common extension values:
+ *   Subject Key ID: type[] = {0x55, 0x1d, 0x0e} type_len = 3;
+ *   Key Usage: type[] = {0x55, 0x1d, 0x0f} type_len = 3;
```
dengert: struct sc_object_id keyUsage_oid={{2,5,29,15,-1}};

src/libopensc/pkcs15-cert.c
```diff
@@ -141,6 +133,195 @@ parse_x509_cert(sc_context_t *ctx, struct sc_pkcs15_der *der, struct sc_pkcs15_c
...
+		 * to make a final copy of the extension value before we return */
+		critical = 0;
+		r = sc_asn1_decode(ctx, asn1_cert_ext, ext, ext_len, NULL, NULL);
+		if (r < 0) {
+			LOG_FUNC_RETURN(ctx, r);
+		}
+
+
+		/* is it the RN we are looking for */
+		if(sc_compare_oid(&oid, type) == 0) {
```
dengert: sc_comapre_id returns 1 if a match and 0 if not. 
Line 274 should be: 
if(sc_compare_oid(&oid, type) != 0) {


I have not checked any other uses of sc_compare_id in this PR.

Jakuje: Thanks. You are right. I read the docs too fast and the wording in comments confused me:

>  * @return 1 if the oids are equal and a non-zero value otherwise

It should probably say

>  * @return 1 if the oids are equal and a zero value otherwise

Will fix in the next commit

src/libopensc/pkcs15-cert.c
Jakuje: Makes sense, same as above. Updated now.

Jakuje:
@dengert  Thanks for comments. Should be fixed now.
Though some unit tests would be probably something nice-to-have, just to make sure it does what it is supposed to do. Any plans on this?

dengert:
I see you have already committed the compare bug. See #847. I want to get @titanous involved, as he requested the change that depends on the #852.

As for testing, #852 I have been using https://github.com/dengert/OpenSC/commit/740a41b  the pkcs15-cert.c part that already parses all the certificates, to look at the keyUsage flags retrieved. I have been using NIST demo cards. Really need @titanous to test out if #852 and https://github.com/dengert/OpenSC/commit/740a41b  do what he is looking for.

PR #855, opened 2016-08-17
fixing of myeid padding problem.

larssilven:
There is a problem with pinpad readers and the myeid smart card.
I believe there will be similar problems with other cards.

C_Login works fine when done with a NULL_PTR PIN; the pinpad prompts the user for PIN and then a CHV is done when entered.
But then if more than one C_xxx is done requiring the application to be logged on; this second C_xxx will fail with p11 error code CKR_PIN_LEN_RANGE (0xA2).

This pull request fixes the p11 so that the user will be prompted for PIN on the pinpad every time a CHV is needed.
Maybe there is a better way to fix this bogus long PIN code (256 chars) than done in this request.  Someone with good opensc knowledge ought to review it and change it if needed.

Also the myeid card should format the PIN objects so that no extra CHV is needed after C_Logon until C_Logout or application end.


larssilven:
Title was bad. Creating a new request.

PR #861, opened 2016-08-26
Use OpenSSL versions OpenSSL-0.9.7 to 1.1.0 for OpenSC

dengert:
OpenSSL-1.1.0 was released today, 8/25/2016

  https://www.openssl.org/news/openssl-1.1.0-notes.html

Changes to allow the OpenSC code base to work with OpenSSL versions from
0.9.7 to 1.1.0 with few changes.

This is an update  and rebased version of my prep-openssl-1.1.0-pre6 branch.

No attempt was made to back port any OpenSSL features. These changes
just allow an updated OpenSC code base to use what is in the various OpenSSL
releases.

A new header libopensc/sc-ossl-compat.h contains extra defines
to reduce the need for so many #if OPENSSL_VERSION_NUMBER statements
in the source code.

The OpenSC source can now use the OpenSSL 1.1 API. The libopensc/sc-ossl-compat.h
has defines for the new API for use with older versions of OpenSSL.

sc-ossl-compat.h is included by libopensc/internal.h so all OpenSC
library routines can take advantage of it. For the tools, which do not use
libopensc/internal.h, libopensc/sc-ossl-compat.h is included by the tools.

The OpenSC source has been modified to use OpenSSL functions to access
hidden structures, such X509, BIGNUM, EVP_CIPHER_CTX, and use XXX_new
functions to allocate structures which most use pointer such as
BIGNUM and EVP_CIPHER_CTX.

For backward compatability sc-ossl-compat.h now defines inline routines
to emulate the RSA and DSA  access routines in OpenSSL-1.1.0. Thus
the same OpenSC source code can be used with openSSL versions from
0.9.7 to 1.1.0.

Inline routines were chosen, because using macros does not work on all platforms.
Having OpenSC versions of these routines in libopensc would be a posibility,
but they are only used for older version of OpenSSL, and could be removed in
the future.

These mods need much testing. Testing code that I can not test.
Input is welcome.

 Changes to be committed:
	modified:   src/libopensc/card-entersafe.c
	modified:   src/libopensc/card-gids.c
	modified:   src/libopensc/card-gpk.c
	modified:   src/libopensc/card-oberthur.c
	modified:   src/libopensc/card-westcos.c
	modified:   src/libopensc/cwa-dnie.c
	modified:   src/libopensc/cwa14890.c
	modified:   src/libopensc/internal.h
	modified:   src/libopensc/p15card-helper.c
	modified:   src/libopensc/pkcs15-itacns.c
	modified:   src/libopensc/pkcs15-prkey.c
	modified:   src/libopensc/pkcs15-pubkey.c
	new file:   src/libopensc/sc-ossl-compat.h
	modified:   src/pkcs15init/pkcs15-lib.c
	modified:   src/pkcs15init/pkcs15-oberthur-awp.c
	modified:   src/pkcs15init/pkcs15-oberthur.c
	modified:   src/pkcs15init/pkcs15-oberthur.h
	modified:   src/pkcs15init/pkcs15-westcos.c
	modified:   src/tools/cryptoflex-tool.c
	modified:   src/tools/gids-tool.c
	modified:   src/tools/netkey-tool.c
	modified:   src/tools/piv-tool.c
	modified:   src/tools/sc-hsm-tool.c
	modified:   src/tools/westcos-tool.c


dengert:
Before this PR is committed, ALL developers will need to look at their code. The idea is to write all code to the OpenSSL 1.1.0 AP. The sc-ossl-compat.h has defines and inline routines to map the 1.1 API to the APIs used in previous versions of OpenSSL.

The code that really need testing is code for devices I do not have and code I do not use. This includes:

sc-hsm-tool
westcos-tool
netkey-tool
gids-tool
cryptoflex-tool
any card that uses SM cwa-dnie or cwa14890
any uses of the pkcs15init to read or write keys
any use of pkcs11-tool that reads or writes keys
The major changes with OpenSSL 1.1.0 are in accessing variables in structures that are now hidden such as RSA, DSA, and BIGNUM.

EVP_CIPHER_CTX and any EVP_encrypt* or EVP_decrypt* routines have changed.

Many hidden structures can no longer be allocated on the stack, but must be allocated by OpenSS via *_new functions: BIGNUM and EVP_CIPHER_CTX

The testing needs to be done with this PR installed, against OpenSSL-1.1.0 AND against current the versions of OpenSSL. The PR provides macros and inline routines that are used with the previous versions of OpenSSL. Look at the sc-ossl-compat.h to see these defines and macros.

frankmorgner:
Not having looked at the patch, I just found some interesting bits in http://blog.fefe.de/?ts=a93f56bb. Points are only translated, not verified:

- LibreSSL also uses OPENSSL_VERSION_NUMBER, but with a different meaning. One could use something like
```
#if (OPENSSL_VERSION_NUMBER > 0x10100000L) || !defined(LIBRESSL_VERSION_NUMBER)
```

- There were some changes to the API that change the behaviour from some functions that were also present in pre 1.1.0, which will introduce errors though the code compiles. One example seems to be `HMAC_Init_ex`, but there are said to be more...

dengert:
Thanks for starting to look at this. 
    
    On 8/27/2016 4:06 PM, Frank Morgner
      wrote:
    
    
      Not having looked at the patch, I just found some interesting
        bits in http://blog.fefe.de/?ts=a93f56bb.
        Points are only translated, not verified:
      
        LibreSSL also uses OPENSSL_VERSION_NUMBER, but with a
          different meaning. One could use something like
      
      #if (OPENSSL_VERSION_NUMBER > 0x10100000L) || !defined(LIBRESSL_VERSION_NUMBER)
    
    
    OpenSC does not support LIBRESSL  today. If you think that is going
    to be a problem in the future, we can address that as a separate
    issue. 
    http://www.libressl.org/
    
      
        There were some changes to the API that change the behaviour
          from some functions that were also present in pre 1.1.0, which
          will introduce errors though the code compiles. One example
          seems to be HMAC_Init_ex, but there are said to
          be more...
      
    
    Note: the PR only address the routine used by OpenSC.  HMAC_Init_ex
    is not used by OpenSC today.
    
    Yes, there may be more. Everyone is ticked off at OpenSSL making all
    these changes in one release. 
    One routine they dropped (used by OpenSC) is i2c_ANS1_INTEGER. It
    gets a hex string of the certificate serialNumber. I think I got
    this converted correctly. It needs to be tested by someone that
    actually uses this. 
    
    What I would hope would happen, is OpenSC developers would try the
    PR with whatever OpenSSL they use today. Then eventually try it
    with OpenSSL-1.1.0. The intent was to start coding to the
    OpenSSL-1.1.0 API while still using older versions of OpenSSL.
    sc-ossl-compat.h contains  #defines and inline routines  to use the
    new  API but then access the structures exposed in older version of
    OpenSSL.  with older versions of OpenSSL.
    
    Many of the OpenSC code changes are to use functions available in
    older versions of OpenSSL to access structures. No one used them
    because direct access to the structures continued to work. But will
    not work with OpenSSL-1.1.0.   
    
    If some OpenSC developer adds new code to OpenSC that is using the
    old API, we can catch it by compiling against 1.1.0 and get them to
    use the new API. 
    Maybe a CI-travis test?  If they use the new API to access some
    feature of an older OpenSSL that is not current handled in
    sc-ossl-compat.h, we can add it to sc_ossl-compat.h.
    .
    The idea was to not have so many #ifdefs in the code that it became
    unreadable. 
    
    Most of the code changes are to use the new API. It is imperative
    that the OpenSC developers, including vendors like CardContact and
    Yubico 
    test their code and utilities.  There are a lot of code changes and
    could only test a small part. 
    
    I have not heard of anyone trying the PR. I am looking for
    testers!!!
    
    (I also submitted a PR for pam_pkcs11 using a version of
    pam-pkcs11-ossl-compat.h but still need to fix the i2c_ANS1_INTEGER
    issue.)
    
    
    
    
      
      
      —
        You are receiving this because you authored the thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/OpenSC/OpenSC","title":"OpenSC/OpenSC","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/OpenSC/OpenSC"}},"updates":{"snippets":[{"icon":"PERSON","message":"@frankmorgner in #861: Not having looked at the patch, I just found some interesting bits in http://blog.fefe.de/?ts=a93f56bb. Points are only translated, not verified:\r\n\r\n- LibreSSL also uses OPENSSL_VERSION_NUMBER, but with a different meaning. One could use something like\r\n```\r\n#if (OPENSSL_VERSION_NUMBER \u003e 0x10100000L) || !defined(LIBRESSL_VERSION_NUMBER)\r\n```\r\n\r\n- There were some changes to the API that change the behaviour from some functions that were also present in pre 1.1.0, which will introduce errors though the code compiles. One example seems to be `HMAC_Init_ex`, but there are said to be more..."}],"action":{"name":"View Pull Request","url":"https://github.com/OpenSC/OpenSC/pull/861#issuecomment-242940872"}}}
    
    
    -- 

dengert:
@frankmorgner, @martinpaljak, @viktorTarasov, @CardContact,  and "Remy" 

OpenSSL-1.1.0a was released yesterday. 

https://github.com/OpenSC/OpenSC/commit/a6dff51c6bd1d5eb6528dff7f6e7b86f29430839 was commited 9/20/2016, and changed sc-hsm-tool.c and added code that is not compatible with OpenSSL-1.1.0.

This PR 861 now addresses both.

src/tools/pkcs15-init.c
```diff
@@ -1136,19 +1150,30 @@ do_read_check_certificate(sc_pkcs15_cert_t *sc_oldcert,
...
 	r = SC_ERROR_INVALID_ARGUMENTS;
-	if (oldpk->type == newpk->type)
+	if (oldpk_type == newpk_type)
 	{
-		if ((oldpk->type == EVP_PKEY_DSA) &&
-			!BN_cmp(oldpk->pkey.dsa->p, newpk->pkey.dsa->p) &&
-			!BN_cmp(oldpk->pkey.dsa->q, newpk->pkey.dsa->q) &&
-			!BN_cmp(oldpk->pkey.dsa->g, newpk->pkey.dsa->g))
+#if  OPENSSL_VERSION_NUMBER >= 0x10002000L
+		if (EVP_PKEY_cmp(oldpk, newpk) == 1)
```
frankmorgner: You could duplicate the implementation of `EVP_PKEY_cmp` in `sc-ossl-compat.h/c` if it's missing just as you did with the other implementations to make this code more readable.

src/tools/pkcs15-init.c
```diff
@@ -2004,7 +2029,8 @@ do_read_pkcs12_private_key(const char *filename, const char *passphrase,
 		return SC_ERROR_CANNOT_LOAD_KEY;
 	}
 
-	CRYPTO_add(&user_key->references, 1, CRYPTO_LOCK_EVP_PKEY);
+	EVP_PKEY_up_ref(user_key);
```
frankmorgner: Is a lock identical to an incremented reference counter?

src/tools/sc-hsm-tool.c
```diff
@@ -157,7 +160,11 @@ static int generatePrime(BIGNUM *prime, const BIGNUM *s, const int bits, unsigne
 
 	do {
 		// Generate random prime
+#if OPENSSL_VERSION_NUMBER  >= 0x00908000L /* last parm is BN_GENCB which is null in our case */
+		BN_generate_prime_ex(prime, bits, 1, NULL, NULL, NULL);
```
frankmorgner: You could duplicate the implementation of `BN_generate_prime_ex` in `sc-ossl-compat.h/c` if it's missing just as you did with the other implementations to make this code more readable.

src/libopensc/card-gpk.c
```diff
@@ -1539,16 +1551,15 @@ gpk_pkfile_load(sc_card_t *card, struct sc_cardctl_gpk_pkload *args)
...
-		if (!EVP_EncryptUpdate(&ctx, temp+n, &outl, args->data + n, 8)) {
+		if (!EVP_EncryptUpdate(ctx, temp+n, &outl, args->data + n, 8)) {
 			r = SC_ERROR_INTERNAL;
 			break;
 		}
 	}
-	if (!EVP_CIPHER_CTX_cleanup(&ctx) || r != SC_SUCCESS)
-		return SC_ERROR_INTERNAL;
+	if (ctx)
+		EVP_CIPHER_CTX_free(ctx);
```
dengert: OK will fix.

src/libopensc/p15card-helper.c
```diff
@@ -169,13 +171,15 @@ CERT_HANDLE_FUNCTION(default_cert_handle) {
...
-	if(pkey->pkey.rsa->n == NULL) {
+	rsa = EVP_PKEY_get0_RSA(pkey);
+	if( rsa == NULL) {
 		sc_debug(p15card->card->ctx, SC_LOG_DEBUG_NORMAL, "Error: no modulus associated with the certificate");
 		r = SC_ERROR_INTERNAL;
 		goto err;
 	}
 	
-	modulus_len = 8 * BN_num_bytes(pkey->pkey.rsa->n); /* converting to bits */
+	modulus_len =  RSA_bits(rsa); /* converting to bits */
```
dengert: OK will fix

src/libopensc/pkcs15-itacns.c
```diff
@@ -30,17 +30,25 @@
...
+#ifdef ENABLE_OPENSSL
+#include <openssl/x509.h>
+#include <openssl/x509v3.h>
+#endif
+
+#include "libopensc/internal.h"
+#include "libopensc/pkcs15.h"
+#include "libopensc/log.h"
+#include "libopensc/cards.h"
+#include "libopensc/itacns.h"
```
dengert: They were move to be after libopensc/internal.h

libopensc/internal.h includes the libopensc/sc-ossl-compat.c 

libopensc/sc-ossl-compat.c  includes openssl/opensslv.h and openssl/opensslconf.h
which should be the first openssl header files included. 

the names now include "libopensc/" as is done in other source files. 

This makes the order of the includes more inline with other source files.


and "libopensc/" was add to each to match what most other source files do.

src/libopensc/pkcs15-prkey.c
```diff
@@ -27,23 +27,28 @@
...
 	#include <openssl/ec.h>
 	#endif
 #endif
 #endif
 
+#include "libopensc/internal.h"
+#include "libopensc/asn1.h"
+#include "libopensc/pkcs15.h"
+#include "common/compat_strlcpy.h"
+#include "libopensc/aux-data.h"
```
dengert: They were moved  and "libopensc/" was added For the same reasons as pkcs15-iatcns.c

src/libopensc/sc-ossl-compat.h
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
```
dengert: I will leave that upto the thye maintainers if this is an issue.

src/libopensc/sc-ossl-compat.h
```diff
@@ -0,0 +1,240 @@
...
+#   define sc_ossl_inline __inline__
+#  elif defined(_MSC_VER)
+#   define sc_ossl_inline __inline
+#  else
+#   define sc_ossl_inline
+#  endif
+# else
+#  define sc_ossl_inline inline
+# endif
+#endif
```
dengert: I did what OpenSSL does with inline. They have found that there is no way to write the code and let all compilers figure it out.  I also used  sc_ossl_inline so any other macros that try and do something with inline don't cause problems in OPenSC.

src/pkcs15init/pkcs15-oberthur-awp.c
```diff
@@ -996,6 +997,30 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
...
+			ci->serial.len = 0;
+			r = SC_ERROR_OUT_OF_MEMORY;
+			goto done;
+		}
+		ci->serial.len = i2d_ASN1_INTEGER(X509_get_serialNumber(x), &ci->serial.value);
+	}
+	/* if len == 0, and value == NULL, then the cert did not have a serial number.*/
+	sc_debug(ctx, SC_LOG_DEBUG_NORMAL, "cert. serial encoded length %i", ci->serial.len);
+
+#else
```
dengert: OK after the testing.

src/pkcs15init/pkcs15-oberthur.c
```diff
@@ -20,20 +20,20 @@
...
-#include <errno.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+//#include "config.h"
+#include "pkcs15-oberthur.h"
+//#include <errno.h>
+//#include <stdio.h>
+//#include <stdlib.h>
+//#include <string.h>
```
dengert: OK will fix.

src/tools/pkcs11-tool.c
```diff
@@ -456,12 +460,20 @@ int main(int argc, char * argv[])
...
+		| OPENSSL_INIT_ADD_ALL_DIGESTS
+		| OPENSSL_INIT_LOAD_CONFIG,
+		NULL);
+#else
 	/* OpenSSL magic */
-	SSLeay_add_all_algorithms();
-	CRYPTO_malloc_init();
+	OpenSSL_add_all_algorithms();
+	OPENSSL_malloc_init();
+#endif
```
dengert: Will look at it later.

src/tools/pkcs11-tool.c
```diff
@@ -2023,16 +2036,24 @@ do_read_key(unsigned char *data, size_t data_len, int private, EVP_PKEY **key)
 
 #define RSA_GET_BN(RSA, LOCALNAME, BNVALUE) \
 	do { \
-		RSA->LOCALNAME = malloc(BN_num_bytes(BNVALUE)); \
-		if (!RSA->LOCALNAME) \
-			util_fatal("malloc() failure"); \
-		RSA->LOCALNAME##_len = BN_bn2bin(BNVALUE, RSA->LOCALNAME); \
+		if (BNVALUE) { \
+			RSA->LOCALNAME = malloc(BN_num_bytes(BNVALUE)); \
```
dengert: In parse_rsa_pkey one of the parameters is struct rsakey_info *rsa. 
The macro RSA_GET_BN(RSA, LOCALNAME, BNVALUE) is only used in parse_pkey. 
So the macro is operating on a struct rsakey_info not an OpenSSL RSA object. 

The new code in parse_rsa_pkey uses the RSA_get0_* routines to get pointers to:
 +	const BIGNUM *r_n, *r_e, *r_d;
+	const BIGNUM *r_p, *r_q;
+	const BIGNUM *r_dmp1, *r_dmq1, *r_iqmp;

then the RSA_GET_BN macro is use to extract from teh BIGNUMs what it needs.

dengert: But I did duplicate a few lines, and is being fixed.

src/tools/pkcs15-init.c
dengert: This is the only place in OpenSC that tries to compare public keys, and 1.0.2 and above have the EVP_PKEY_cmp. So I did not try and add it to sc-ossl-compat.h. 
I would be open to doing this after this big PR gets committed.

src/tools/pkcs15-init.c
dengert: Crypto_lock is not supported in 1.1.0 See:
https://github.com/openssl/openssl/issues/1260

If we need this, we may have to rewrite other code too.

src/tools/sc-hsm-tool.c
dengert: It could be done, but this is the only place in OpenSC it is used. 
I would like @CardContact to look at all the changes in this module first.

dengert:
See above comments.

dengert:
The PR has been rebased, and some change as requested by @frankmorgner  have been added. See the comments in the review above. 

Issues of CRYPTO_lock not being supported in 1.1.0 need to be looked at. 

Can the sc-ossl-compat.h have a copyright? 

Still needed to be tested on many cards, and the tools especially sc-hsm-tool.c needs to be reviewed by @CardContact

frankmorgner:
compatibility with LibreSSL has been added in #878, so the version check needs be changed to something like
```
#if (OPENSSL_VERSION_NUMBER > 0x10100000L) || !defined(LIBRESSL_VERSION_NUMBER)
```
*OR* you could instead check for the missing function in configure.ac by using
```
AC_CHECK_FUNCS(DSA_get0_key ...)
```
and check in the code for
```
#ifndef HAVE_DSA_GET0_KEY
static sc_ossl_inline void DSA_get0_key(const DSA *d, const BIGNUM **pub_key, const BIGNUM **priv_key)
{
    ...
}
#endif
```

src/libopensc/sc-ossl-compat.h
frankmorgner: please move the implementation of compatiblity functions to `ssl-compat.c`. The current setup bloats all object files that are using `ssl-compat.h`. This also removes the requirement to do this as` inline`.

src/libopensc/pkcs15-itacns.c
frankmorgner: I think it's better to not introduce a dependency to the includes that are listed in `internal.h`. That might change in the future...

dengert:
#861 was rebased and the conflict introduced from #878 was resolved in openssl.c 

Are you now saying that #861 now has to address differences between OpenSSL versions and LibreSSL because LibreSSL differs from OpenSSL with the DSA_get0_key? What other issues will LibreSSL have?

Personally, I thing it is more important OpenSC work with OpenSSL then LibreSSL. I would like to see  #878 temporarily reverted, #861 merged The authors of #878 can see other changes they need and resubmit #878.  The longer we wait to get OpenSSL-1.1.0 changes in to OpenSC the harder it will be. There needs to be a lot of testing by every developer including developers who want to use LibreSSL.

When #861 is finally merged, I would also like to see travis-ci or appveyor modified to compile against both OpenSSL-1.0.2 and OpenSSL-1.1.0 in at least one test. 

If LibreSSL is important, travis-ci or appveyor should also have at least one test using LibreSSL.

src/libopensc/pkcs15-itacns.c
dengert: In master All  card-\<card\>.c (38 files)  and pkcs15-\<card\>.c (26 files) used #include "internal.h"
pkcs15-iatcns.c was the only pkcs15-\<card\>.c  that did NOT include internal.h. 

I will change  card-gpk.c, pkcs15-prkey.c, pkcs15-pubkey.c and pkcs15-itacns.c back to use 
#include "internal.h"
rather then 
#include "libopensc/internal.h"
And remove all the other "libopensc/" in these files when I moved the includes.  

Note that in master, there is many other inconsistencies on how includes are done for example try this:
cd src/libopensc
grep "include.*log.h" *
A subject for another PR to clean this up.

src/libopensc/pkcs15-prkey.c
dengert: will fix these too.

src/libopensc/sc-ossl-compat.h
dengert: The intent of using inline, macros and defines  was to NOT have another file (ssl-compat.c) with routines. This would avoid of any conflicts with other packages that might also provide their own versions of these routines.   Programs that load modules for example PAM, might have conflicts.

The inline functions are not compiled when using OpenSSL-1.1.0. The inline functions are all small, and in my option don't add much bloat.

frankmorgner:
What I'm asking in regard to libressl is easy, I think: Since it doesn't support OpenSSL's new API, please treat like it was openssl 1.0.1.

Additionally, I proposed an alternative approach, which is to actually check for the functionality you need in `configure.ac`, e.g. by looking for a certain function (`AC_CHECK_FUNCS(DSA_get0_key`). This approach works with both, OpenSSL and LibreSSL.

If you have trouble verifying your PR, we can ask @lbschenkel for help, or I'll look at it myself.

frankmorgner:
By the way, do we still need to support 1.0.0 or earlier, which are not supported by the OpenSSL team?

lbschenkel:
I have made changes to this patch to make it buildable with LibreSSL (2.4.2, but I believe that it would work for earlier versions as well):
https://github.com/dengert/OpenSC/compare/openssl-1.1.0-compat...lbschenkel:libressl-compat?expand=1

They do not need to be taken as they are (honestly I believe the logic for version detection could be streamlined quite a bit and duplication eliminated), but my intent was just to show the specific bits that are missing from LibreSSL.

frankmorgner:
thanks, @lbschenkel!

dengert:
@lbschenkel 
Your changes look OK to me.

But in the future is LibreSSL going to support the OpenSSL-1.1.0 API? 
How would you "streamline" the version detection? 

I see the LiberSSL has [openssl/opensslv.h](https://github.com/libressl--ortable/openbsd/blob/master/src/lib/libcrypto/opensslv.h) and sets both  
#define LIBRESSL_VERSION_NUMBER	0x2050000fL
#define OPENSSL_VERSION_NUMBER	0x20000000L

You said you got it to compile with LibreSSL 2.4.2. Do we need to worry about version 2.5?

If and when it does support OpenSSL-1.1.0 API what changes would we have to make at that time?
Just checking if defined(LIBRESSL_VERSION_NUMBER) would not catch any future versions.  

@frankmorgner 
Are you implying that LibreSSL does not have DSA_get0_key? Or was that just an example?

lbschenkel:
Honestly I didn't spend a lot of time analysing this PR, just wanted to help identifying what bits LibreSSL doesn't support.

Regarding the "streamlining", I feel that a lot of `#ifdef` keep repeating the same tests which to me is a violation of the DRY principle. If this was my PR, I would probably try an approach in which I categorize the missing functionality in groups (let's say 2 or 3), and based on a single place in which you test OPENSSL/LIBRESSL_VERSION you `#define SUPPORTS_FUNCTIONALITY_1` and `#define SUPPORTS_FUNCTIONALITY_2` and only have `#ifdef`s for these in the code.

This is mostly a rephrasing of @frankmorgner already said. Most likely autotools already have all the functionality to do that and set `#define`s based on the available functionality (but I don't have much knowledge in autotools).

You're right that the logic for handling version numbers is a mess and very hard to maintain. That's why to me that is a strong sign that the right approach uses feature detection, not version detection — and if you must use version detection, put the logic in one single place and define the features there.

lbschenkel:
Just a clarification of the autotools approach since it maybe wasn't clear: if I understand autotools correctly, you can simply use 1.1.x features in the code and declare them in `configure.ac` as @frankmorgner explained. Then if the `configure` script detects that those are not available at build time, the configure script will automatically add `missing_function.c` in the Makefiles for all sources that use that function — all you need to do is implement all functions that may be missing in their own .c files and they will be automatically used if needed without needing to pollute the code with `#ifdefs` and alternative implementations inline, resulting in a much more readable codebase (in my opinion, at least).

dengert:
The approach to this BIG pr was to change as little code as possible keeping \#ifdefs to a minimum.
Don't add any new modules, don't try and add new features, start using the OpenSSL API and use inline or macros for older version compatibility. And have it work with OpenSSL and accepted as a PR. 

Other changes to make it more readable  were left for future PRs. 
I am OK with your changes, but they need to be tested as well as most of the PR needs to be tested. If anyone feels LibreSSL 2.5 is needed, we need a tester.

lbschenkel:
Fair enough. I'm not really reviewing this PR, I was just mentioned and wanted to give my two cents. Regarding LibreSSL, I think as long as you're falling back like you're doing here it does not really matter if 2.5+ implement the missing features — it will still build. If later versions of LibreSSL break the API by removing some features that are being used, well, then it's not really a problem of your PR and somebody will need to fix it (same applies to OpenSSL in any case). For this PR, supporting LibreSSL is really the same as supporting OpenSSL 1.0.x.

lbschenkel:
I volunteer to test LibreSSL 2.3.x/2.4.x/2.5.x. Just ping me when it's ready to test and what should be tested.

dengert:
OK, I will look closer at your changes. Give me a day or two.

dengert:
I have the changes as suggest by https://github.com/lbschenkel and have compiled OpenSC against LibreSSL-2.5.0 from github.com as well as several versions of OpenSSL.  pkcs11-tool -t -l works.  

This still needs a lot of testing.

dengert:
I would say no. But others may still want it. 
    
    On 10/2/2016 9:06 AM, Frank Morgner
      wrote:
    
    
      By the way, do we still need to support 1.0.0 or earlier, which
        are not supported by the OpenSSL team?
      —
        You are receiving this because you authored the thread.
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/OpenSC/OpenSC","title":"OpenSC/OpenSC","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/OpenSC/OpenSC"}},"updates":{"snippets":[{"icon":"PERSON","message":"@frankmorgner in #861: By the way, do we still need to support 1.0.0 or earlier, which are not supported by the OpenSSL team?"}],"action":{"name":"View Pull Request","url":"https://github.com/OpenSC/OpenSC/pull/861#issuecomment-250972939"}}}
    
    
    -- 

viktorTarasov:
Tested with IAS/ECC card, with SM (cwa14890) using the pkcs15-init and pkcs15-tool,
with OpenSSL versions 1.1.0b, 0.9.8zh and 0.9.7k.

Not related to current PR,
cwa14890 is not to be used with OpenSS-0.9.7k without FIPS -- check for OpenSSL version will be added.

For me it's ready to be applied,
probably rebased into one/two commit ("OpenSSL-0.9.7 to 1.1.0" and "support for LibreSSL").

Great work, Doug!

CardContact:
Tested with SmartCard-HSM and OpenSSL 1.0.2g.

Found no issues.

Thanks for the all the effort, Doug !

dengert:
@viktorTarasov 
Because #861 modifies so many files, every time OpenSC is updated, #861 needs to be rebased.  I would prefer to have it merged as is. 

The fact that the major developers that needed to look at the code have done so in the last few days, leads me to say now is the time to do the merge before other PRs are merged then will require additional testing. 

#861 has 5 commits, which has has been tested by you and @CardContact and reviewed by @frankmorgner and @lbschenkel. One is the base code, that keeps getting rebased as other changes to OpenSC are merged. One is some changes in sc-hsm-tool.c that I missed. One is from comments by @frankmorgner. One is from comments from @lbschenkel. 

If I do squash them all together, do you and other want a chance to test the squashed version before it is finally merged? 

The last commit addressed LibreSSL. It was only added because #878 was merged on 9/26/2016 and now needed to be addressed. @lbschenkel is the primary developers interests in LibreSSL, and has provided the changes. 

@viktorTarasov, @frankmorgner So what is the next step? Merge now or squash with more testing?

src/tools/pkcs15-init.c
frankmorgner: I'd like to have all OpenSSL version specific stuff in `sc-ssl-compat.c`. This would be easier to track.

frankmorgner:
- Github recently added "Squash and merge", so there's no need to do this manually.
- I agree to leave the copyright as is.
- Compatibility with LibreSSL is fixed (enough), thanks to @lbschenkel.

However, I'd like to see the comments regarding the code quality to be resolved since you also agreed to most of them. Especially moving the static implementation of the OpenSSL wrapper functions from `sc-ssl-compat.h` to non-static implementations in `sc-ssl-compat.c` still needs to be done.

There was some uncertainty whether oberthur cards still work correctly (see https://github.com/OpenSC/OpenSC/pull/861#discussion_r81024000). @viktorTarasov, could you check with your cards?

viktorTarasov:
@dengert 
> Because #861 modifies so many files, every time OpenSC is updated, #861 needs to be rebased. I would prefer to have it merged as is.

Many files are modified only in first commit, the following ones bring the minor or cosmetic changes, without big functional impact -- _rebase_ is not dangerous. I mean _rebase_ into two commits, as proposed above.
If there are no remarks to the PR itself, I will merge it. As for me, the clarity of git history (venture to say the _beauty_) is also a precious project's quality,
but I do not insist.

@frankmorgner 
> "Squash and merge"

Ok, nice to know, but this PR I would like to see rebased into two commits, as proposed above.

I will test with Oberthur.

dengert:
@frankmorgner
I agreed to most of your suggestions, and addressed them. I did not agree to removing and/or moving the inline code and creating a sc-ssl-compat.c

There is little or no bloat. With gcc -O2, at least, the static inline routines are only expanded in the modules that actually use the routines. Try this to see which object files include the RSA_get0_ and DSA_get0_ routines in your build environment:

find . -type f -name \*.o -exec bash -c 'nm {} | grep SA_get0_' \; -ls

I only see only 9 object files that contain them, all of which call these routines. Do you have some environment where the compiler does add these to every object file?

I don't want to create  a sc-ossl-compat.c. I don't want to loose the code to inline the routines by removing it before a merge, as I feel this is the better way to go.

The PR is working now, but as more developers are FORCED to test it, I am sure they will find bugs.  That is why I feel it should be committed as is. (I have been asking for months to get others to look at and test the PR. Finally This week 3 developers did. Thanks @viktorTarasov @CardContact and @lbschenkel)  

If after the merge  you still disagree and feel that creating a sc-ossl-compat.c is a better, I suggest that you write a PR.

As I said before:
"The intent of using inline, macros and defines was to NOT have another file (ssl-compat.c) with routines. This would avoid of any conflicts with other packages that might also provide their own versions of these routines. Programs that load modules for example PAM, might have conflicts.

The inline functions are not compiled when using OpenSSL-1.1.0. The inline functions are all small, and in my option don't add much bloat."

viktorTarasov:
For me it's OK with oberthur.

@dengert
Once more about rebase, understand your exasperation, but I'll venture to try:

Commits to squash:

##### Use OpenSSL versions OpenSSL-0.9.7 to 1.1.0 for OpenSC
initial commit
##### Use OpenSSL versions OpenSSL-0.9.7 to 1.1.0a
Only touched the _sc-hsm-tool_, for which the _0.9.7->1.1.0_ change is applied in EVP_Decrypt part.

Previous commit, for _sc-hsm-tool_, contains the mirror change in EVP_Encrypt part ,
and the same change in EVP_Decrypt part for the other tools.
No reason to keep this commit  non squashed. 

##### Minor issues  from a review
minor, cosmetic changes --  coverity warning, useless code line, comments removed.
No reason to keep it non squashed. 

##### revert  changes tha added libopensc/ to some #include statments
Message itself appeal to squash this commit. 
The changes concern compile, install;  without impact onto functional side.

##### Add support for LibreSSL compatability as well as OpenSSL-1.1.0
Will not be squashed

dengert:
OK, I will squash and rebase. 

The question "Does OpenSC need to support 0.9.7" was asked. Do we drop it?
sc-ossl-compat.h has defines to support EVP_CIPHER_CTX_new and EVP_CIPHER_CTX_free that were not in 0.9.7.
I also see unrelated to this PR, card-oberthur.c has:
 49 /* keep OpenSSL 0.9.6 users happy ;-) */
 50 #if OPENSSL_VERSION_NUMBER < 0x00907000L

I would suggest that the PR would include the 0.9.7 and a future PR would address the minimum version supported.

src/libopensc/cwa-dnie.c
```diff
@@ -353,11 +357,18 @@ static int dnie_get_root_ca_pubkey(sc_card_t * card, EVP_PKEY ** root_ca_key)
...
+
+	root_ca_rsa_n = BN_bin2bn(icc_root_ca_modulus, sizeof(icc_root_ca_modulus), NULL);
+	root_ca_rsa_e = BN_bin2bn(icc_root_ca_public_exponent, sizeof(icc_root_ca_public_exponent), NULL);
+	if (RSA_set0_key(root_ca_rsa, root_ca_rsa_n, root_ca_rsa_e, NULL) != 1) {
+		if (*root_ca_key)
+			EVP_PKEY_free(*root_ca_key);
+		if (root_ca_rsa)
+			RSA_free(root_ca_rsa);
+		sc_log(card->ctx, "Cannot set RSA valuses for CA public key");
+		return SC_ERROR_INTERNAL;
```
frankmorgner: `root_ca_rsa_n` and `root_ca_rsa_e` need to be freed in case of an error.

src/libopensc/cwa-dnie.c
```diff
@@ -439,13 +451,16 @@ static int dnie_get_ifd_privkey(sc_card_t * card, EVP_PKEY ** ifd_privkey)
...
-	    BN_bin2bn(ifd_private_exponent, sizeof(ifd_private_exponent),
-		      ifd_rsa->d);
+	ifd_rsa_n = BN_bin2bn(ifd_modulus, sizeof(ifd_modulus), NULL);
+	ifd_rsa_e = BN_bin2bn(ifd_public_exponent, sizeof(ifd_public_exponent), NULL);
+	ifd_rsa_d = BN_bin2bn(ifd_private_exponent, sizeof(ifd_private_exponent), NULL);
+	if (RSA_set0_key(ifd_rsa, ifd_rsa_n, ifd_rsa_e, ifd_rsa_d) != 1) {
+		RSA_free(ifd_rsa);
+		EVP_PKEY_free(*ifd_privkey);
+		sc_log(card->ctx, "Cannot set RSA values for IFD private key");
+		return SC_ERROR_INTERNAL;
```
frankmorgner: `ifd_rsa_n`/`e`/`d` need to be freed in case of an error.

src/libopensc/cwa14890.c
```diff
@@ -1340,8 +1346,8 @@ int cwa_create_secure_channel(sc_card_t * card,
 
 	/* compose signature data for external auth */
 	res = cwa_prepare_external_auth(card,
-					icc_pubkey->pkey.rsa,
-					ifd_privkey->pkey.rsa, sn_icc, sm);
+					EVP_PKEY_get0_RSA(icc_pubkey),
+					EVP_PKEY_get0_RSA(ifd_privkey), sn_icc, sm);
```
frankmorgner: This is an error, you need to extract the private and public key first.

src/libopensc/sc-ossl-compat.h
frankmorgner: collisions of function names are not a problem since the compatibility functions are not exported.

src/tools/pkcs11-tool.c
frankmorgner: did you have a look?

frankmorgner:
I found some more problems, that need to be fixed (see above review).

@CardContact you should check with OpenSSL 1.1.0 to be sure

src/libopensc/cwa-dnie.c
dengert: OK, will fix. RSA_set0_key checks up front. So we can free if these is an error.

src/libopensc/cwa-dnie.c
dengert: Will fix this too.

src/libopensc/cwa14890.c
dengert: I don't see the error. The 1.1.0 function returns a pointer to the RSA:

RSA *EVP_PKEY_get0_RSA(EVP_PKEY *pkey);

sc-ossl-compat.h has:
#define EVP_PKEY_get0_RSA(x)            (x->pkey.rsa)

dengert:
@frankmorgner about your comments of:
src/libopensc/sc-ossl-compat.h
src/tools/pkcs11-tool.c
(I don't see a place on the web page to add a reply after your questions, so am adding them here.)

I don' t understand what you want to change in either of these.

dengert:
Memory leaks in cwa-dnie.c are fixed, and squashed into https://github.com/OpenSC/OpenSC/pull/861/commits/c6dba96f559311576532578c4b274d50d710c225  The PR was also rebased. 

I regards to question on src/tools/pkcs11-tool.c and OPENSSL_malloc_init:
All the tools include sc-ossl-compat.h  which has \#defines for functions like these where the name has changed.

viktorTarasov:
Thanks for  your work.

mouse07410:
@viktorTarasov and @dengert current master breaks compilation with OpenSSL-1.0.2j installed in /opt/local. It correctly identified the library files, but appears to have missed the include files location.

```
$ MacOSX/build
+ test -x ./configure
+ ./configure
checking for a BSD-compatible install... /opt/local/bin/ginstall -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /opt/local/bin/gmkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking build system type... x86_64-apple-darwin15.6.0
checking host system type... x86_64-apple-darwin15.6.0
checking for gcc... clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether clang accepts -g... yes
checking for clang option to accept ISO C89... none needed
checking whether clang understands -c and -o together... yes
checking for style of include used by make... GNU
checking dependency style of clang... gcc3
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ accepts -g... yes
checking dependency style of clang++... gcc3
checking for pkg-config... /opt/local/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking how to run the C preprocessor... clang -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking whether byte ordering is bigendian... no
checking how to run the C preprocessor... clang -E
checking whether ln -s works... yes
checking for a sed that does not truncate output... /usr/bin/sed
checking whether make sets $(MAKE)... (cached) yes
checking how to print strings... printf
checking for a sed that does not truncate output... (cached) /usr/bin/sed
checking for fgrep... /usr/bin/grep -F
checking for ld used by clang... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
checking if the linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) is GNU ld... no
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
checking the name lister (/usr/bin/nm -B) interface... BSD nm
checking the maximum length of command line arguments... 196608
checking how to convert x86_64-apple-darwin15.6.0 file names to x86_64-apple-darwin15.6.0 format... func_convert_file_noop
checking how to convert x86_64-apple-darwin15.6.0 file names to toolchain format... func_convert_file_noop
checking for /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld option to reload object files... -r
checking for objdump... no
checking how to recognize dependent libraries... pass_all
checking for dlltool... no
checking how to associate runtime and link libraries... printf %s\n
checking for ar... ar
checking for archiver @FILE support... no
checking for strip... strip
checking for ranlib... ranlib
checking command to parse /usr/bin/nm -B output from clang object... ok
checking for sysroot... no
checking for a working dd... /bin/dd
checking how to truncate binary pipes... /bin/dd bs=4096 count=1
checking for mt... no
checking if : is a manifest tool... no
checking for dsymutil... dsymutil
checking for nmedit... nmedit
checking for lipo... lipo
checking for otool... otool
checking for otool64... no
checking for -single_module linker flag... yes
checking for -exported_symbols_list linker flag... yes
checking for -force_load linker flag... yes
checking for dlfcn.h... yes
checking for objdir... .libs
checking if clang supports -fno-rtti -fno-exceptions... yes
checking for clang option to produce PIC... -fno-common -DPIC
checking if clang PIC flag -fno-common -DPIC works... yes
checking if clang static flag -static works... no
checking if clang supports -c -o file.o... yes
checking if clang supports -c -o file.o... (cached) yes
checking whether the clang linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking dynamic linker characteristics... darwin15.6.0 dyld
checking how to hardcode library paths into programs... immediate
checking whether stripping libraries is possible... yes
checking if libtool supports shared libraries... yes
checking whether to build shared libraries... yes
checking whether to build static libraries... yes
checking how to run the C++ preprocessor... clang++ -E
checking for ld used by clang++... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
checking if the linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) is GNU ld... no
checking whether the clang++ linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking for clang++ option to produce PIC... -fno-common -DPIC
checking if clang++ PIC flag -fno-common -DPIC works... yes
checking if clang++ static flag -static works... no
checking if clang++ supports -c -o file.o... yes
checking if clang++ supports -c -o file.o... (cached) yes
checking whether the clang++ linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking dynamic linker characteristics... darwin15.6.0 dyld
checking how to hardcode library paths into programs... immediate
checking for windres... no
checking for xsltproc... xsltproc
checking for git... git
checking xsl-stylesheets... /opt/local/share/xsl/docbook-xsl
checking git checkout... yes
checking for inline... inline
checking for ANSI C header files... (cached) yes
checking for sys/wait.h that is POSIX.1 compatible... yes
checking whether to enable assertions... yes
checking errno.h usability... yes
checking errno.h presence... yes
checking for errno.h... yes
checking fcntl.h usability... yes
checking fcntl.h presence... yes
checking for fcntl.h... yes
checking for stdlib.h... (cached) yes
checking for inttypes.h... (cached) yes
checking for string.h... (cached) yes
checking for strings.h... (cached) yes
checking sys/time.h usability... yes
checking sys/time.h presence... yes
checking for sys/time.h... yes
checking for unistd.h... (cached) yes
checking getopt.h usability... yes
checking getopt.h presence... yes
checking for getopt.h... yes
checking sys/mman.h usability... yes
checking sys/mman.h presence... yes
checking for sys/mman.h... yes
checking for an ANSI C-conforming const... yes
checking for uid_t in sys/types.h... yes
checking for size_t... yes
checking whether time.h and sys/time.h may both be included... yes
checking for error_at_line... no
checking whether lstat correctly handles trailing slash... no
checking whether stat accepts an empty string... no
checking for vprintf... yes
checking for _doprnt... no
checking for getpass... yes
checking for gettimeofday... yes
checking for memset... yes
checking for mkdir... yes
checking for strdup... yes
checking for strerror... yes
checking for getopt_long... yes
checking for getopt_long_only... yes
checking for strlcpy... yes
checking for strlcat... yes
checking for strnlen... yes
checking size of void *... 8
checking for socket in -lsocket... no
checking for library containing dlopen... none required
checking whether pthreads work with -pthread... yes
checking for joinable pthread attribute... PTHREAD_CREATE_JOINABLE
checking if more special flags are required for pthreads... -D_THREAD_SAFE
checking for PTHREAD_PRIO_INHERIT... yes
checking for inflate in -lz... yes
checking zlib.h usability... yes
checking zlib.h presence... yes
checking for zlib.h... yes
checking for readline in -lreadline... yes
checking readline/readline.h usability... yes
checking readline/readline.h presence... yes
checking for readline/readline.h... yes
checking for OPENSSL... yes
checking openssl/crypto.h usability... yes
checking openssl/crypto.h presence... no
configure: WARNING: openssl/crypto.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: openssl/crypto.h: proceeding with the compiler's result
checking for openssl/crypto.h... yes
checking winscard.h usability... yes
checking winscard.h presence... no
configure: WARNING: winscard.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: winscard.h: proceeding with the compiler's result
checking for winscard.h... yes
checking pcsclite.h usability... yes
checking pcsclite.h presence... no
configure: WARNING: pcsclite.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: pcsclite.h: proceeding with the compiler's result
checking for pcsclite.h... yes
checking XSLTPROC requirement... ok
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating doc/Makefile
config.status: creating doc/tools/Makefile
config.status: creating etc/Makefile
config.status: creating src/Makefile
config.status: creating src/common/Makefile
config.status: creating src/libopensc/Makefile
config.status: creating src/libsm/Makefile
config.status: creating src/pkcs11/Makefile
config.status: creating src/pkcs11/versioninfo-pkcs11.rc
config.status: creating src/pkcs11/versioninfo-pkcs11-spy.rc
config.status: creating src/pkcs11/opensc-pkcs11.pc
config.status: creating src/pkcs15init/Makefile
config.status: creating src/scconf/Makefile
config.status: creating src/tests/Makefile
config.status: creating src/tests/regression/Makefile
config.status: creating src/tools/Makefile
config.status: creating src/tools/versioninfo-tools.rc
config.status: creating src/smm/Makefile
config.status: creating src/minidriver/Makefile
config.status: creating src/minidriver/versioninfo-minidriver.rc
config.status: creating src/minidriver/opensc-minidriver.inf
config.status: creating win32/Makefile
config.status: creating win32/versioninfo.rc
config.status: creating win32/versioninfo-customactions.rc
config.status: creating win32/winconfig.h
config.status: creating win32/OpenSC.iss
config.status: creating win32/OpenSC.wxs
config.status: creating MacOSX/Makefile
config.status: creating MacOSX/build-package
config.status: creating MacOSX/Distribution.xml
config.status: creating MacOSX/resources/ReadMe.html
config.status: creating MacOSX/resources/Welcome.html
config.status: creating config.h
config.status: executing depfiles commands
config.status: executing libtool commands

OpenSC has been configured with the following options:


Version:                 0.16.0
Version fix:             0
Version revision:        4
Git revision:            OpenSC-OpenSC_working_snapshot-4-g8702141, rev: 8702141, commit-time: 2016-10-10 13:13:14 -0400

Copyright:               OpenSC Project
Company:                 OpenSC Project
Company URL:             https://github.com/OpenSC
Comments:                Provided under the terms of the GNU Lesser General Public License (LGPLv2.1+).
Product name:            OpenSC smartcard framework
Product updates:         https://github.com/OpenSC/OpenSC/releases
Product URL:             https://github.com/OpenSC/OpenSC

User binaries:           /usr/local/bin
Configuration files:     /usr/local/etc
XSL stylesheets:         /opt/local/share/xsl/docbook-xsl

man support:             yes
doc support:             no
thread locking support:  yes
zlib support:            yes
readline support:        yes
OpenSSL support:         yes
PC/SC support:           yes
OpenCT support:          no
CT-API support:          no
minidriver support:      no
SM support:              yes
SM default module:       libsmm-local.so
DNIe UI support:         no
Debug file:              /tmp/opensc-debug.log

PC/SC default provider:  /System/Library/Frameworks/PCSC.framework/PCSC
PKCS11 default provider: opensc-pkcs11.so

Host:                    x86_64-apple-darwin15.6.0
Compiler:                clang
Preprocessor flags:       -DPKCS11_THREAD_LOCKING
Compiler flags:          -fno-strict-aliasing -maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -Wall -Wextra -Wno-unused-parameter -Werror=declaration-after-statement
Linker flags:            
Libraries:               

READLINE_CFLAGS:         
READLINE_LIBS:           -lreadline 
ZLIB_CFLAGS:             
ZLIB_LIBS:               -lz
OPENSSL_CFLAGS:          -maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -I/opt/local/include
OPENSSL_LIBS:            -L/opt/local/lib -lssl -lcrypto
OPENCT_CFLAGS:           
OPENCT_LIBS:             
PCSC_CFLAGS:             -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/PCSC.framework/Versions/Current/Headers

+ bash ./MacOSX/build-package
+ test -x ./configure
+ BUILDPATH=/Users/ur20980/src/OpenSC
++ xcode-select -p
+ SDKS_PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs
++ ls -1 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs
++ sort -n -k2 -t. -r
++ head -1
+ SDK_PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk
+ export 'CFLAGS=-maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -arch x86_64 -mmacosx-version-min=10.10'
+ CFLAGS='-maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -arch x86_64 -mmacosx-version-min=10.10'
+ export SED=/usr/bin/sed
+ SED=/usr/bin/sed
+ PREFIX=/Library/OpenSC
+ export PKG_CONFIG_PATH=/opt/local/lib/pkgconfig:/opt/local/share/pkgconfig:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig:/usr/lib/pkgconfig
+ PKG_CONFIG_PATH=/opt/local/lib/pkgconfig:/opt/local/share/pkgconfig:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig:/usr/lib/pkgconfig
+ pkg-config libcrypto --atleast-version=1.0.1
+ ./configure --prefix=/Library/OpenSC --sysconfdir=/Library/OpenSC/etc --disable-dependency-tracking --enable-shared --disable-static --enable-strict --disable-assert --enable-sm
checking for a BSD-compatible install... /opt/local/bin/ginstall -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /opt/local/bin/gmkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking build system type... x86_64-apple-darwin15.6.0
checking host system type... x86_64-apple-darwin15.6.0
checking for gcc... clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether clang accepts -g... yes
checking for clang option to accept ISO C89... none needed
checking whether clang understands -c and -o together... yes
checking for style of include used by make... GNU
checking dependency style of clang... none
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ accepts -g... yes
checking dependency style of clang++... none
checking for pkg-config... /opt/local/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking how to run the C preprocessor... clang -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking whether byte ordering is bigendian... no
checking how to run the C preprocessor... clang -E
checking whether ln -s works... yes
checking for a sed that does not truncate output... /usr/bin/sed
checking whether make sets $(MAKE)... (cached) yes
checking how to print strings... printf
checking for a sed that does not truncate output... (cached) /usr/bin/sed
checking for fgrep... /usr/bin/grep -F
checking for ld used by clang... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
checking if the linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) is GNU ld... no
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
checking the name lister (/usr/bin/nm -B) interface... BSD nm
checking the maximum length of command line arguments... 196608
checking how to convert x86_64-apple-darwin15.6.0 file names to x86_64-apple-darwin15.6.0 format... func_convert_file_noop
checking how to convert x86_64-apple-darwin15.6.0 file names to toolchain format... func_convert_file_noop
checking for /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld option to reload object files... -r
checking for objdump... no
checking how to recognize dependent libraries... pass_all
checking for dlltool... no
checking how to associate runtime and link libraries... printf %s\n
checking for ar... ar
checking for archiver @FILE support... no
checking for strip... strip
checking for ranlib... ranlib
checking command to parse /usr/bin/nm -B output from clang object... ok
checking for sysroot... no
checking for a working dd... /bin/dd
checking how to truncate binary pipes... /bin/dd bs=4096 count=1
checking for mt... no
checking if : is a manifest tool... no
checking for dsymutil... dsymutil
checking for nmedit... nmedit
checking for lipo... lipo
checking for otool... otool
checking for otool64... no
checking for -single_module linker flag... yes
checking for -exported_symbols_list linker flag... yes
checking for -force_load linker flag... yes
checking for dlfcn.h... yes
checking for objdir... .libs
checking if clang supports -fno-rtti -fno-exceptions... yes
checking for clang option to produce PIC... -fno-common -DPIC
checking if clang PIC flag -fno-common -DPIC works... yes
checking if clang static flag -static works... no
checking if clang supports -c -o file.o... yes
checking if clang supports -c -o file.o... (cached) yes
checking whether the clang linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking dynamic linker characteristics... darwin15.6.0 dyld
checking how to hardcode library paths into programs... immediate
checking whether stripping libraries is possible... yes
checking if libtool supports shared libraries... yes
checking whether to build shared libraries... yes
checking whether to build static libraries... no
checking how to run the C++ preprocessor... clang++ -E
checking for ld used by clang++... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
checking if the linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) is GNU ld... no
checking whether the clang++ linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking for clang++ option to produce PIC... -fno-common -DPIC
checking if clang++ PIC flag -fno-common -DPIC works... yes
checking if clang++ static flag -static works... no
checking if clang++ supports -c -o file.o... yes
checking if clang++ supports -c -o file.o... (cached) yes
checking whether the clang++ linker (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld) supports shared libraries... yes
checking dynamic linker characteristics... darwin15.6.0 dyld
checking how to hardcode library paths into programs... immediate
checking for windres... no
checking for xsltproc... xsltproc
checking for git... git
checking xsl-stylesheets... /opt/local/share/xsl/docbook-xsl
checking git checkout... yes
checking for inline... inline
checking for ANSI C header files... (cached) yes
checking for sys/wait.h that is POSIX.1 compatible... yes
checking whether to enable assertions... no
checking errno.h usability... yes
checking errno.h presence... yes
checking for errno.h... yes
checking fcntl.h usability... yes
checking fcntl.h presence... yes
checking for fcntl.h... yes
checking for stdlib.h... (cached) yes
checking for inttypes.h... (cached) yes
checking for string.h... (cached) yes
checking for strings.h... (cached) yes
checking sys/time.h usability... yes
checking sys/time.h presence... yes
checking for sys/time.h... yes
checking for unistd.h... (cached) yes
checking getopt.h usability... yes
checking getopt.h presence... yes
checking for getopt.h... yes
checking sys/mman.h usability... yes
checking sys/mman.h presence... yes
checking for sys/mman.h... yes
checking for an ANSI C-conforming const... yes
checking for uid_t in sys/types.h... yes
checking for size_t... yes
checking whether time.h and sys/time.h may both be included... yes
checking for error_at_line... no
checking whether lstat correctly handles trailing slash... no
checking whether stat accepts an empty string... no
checking for vprintf... yes
checking for _doprnt... no
checking for getpass... yes
checking for gettimeofday... yes
checking for memset... yes
checking for mkdir... yes
checking for strdup... yes
checking for strerror... yes
checking for getopt_long... yes
checking for getopt_long_only... yes
checking for strlcpy... yes
checking for strlcat... yes
checking for strnlen... yes
checking size of void *... 8
checking for socket in -lsocket... no
checking for library containing dlopen... none required
checking whether pthreads work with -pthread... yes
checking for joinable pthread attribute... PTHREAD_CREATE_JOINABLE
checking if more special flags are required for pthreads... -D_THREAD_SAFE
checking for PTHREAD_PRIO_INHERIT... yes
checking for inflate in -lz... yes
checking zlib.h usability... yes
checking zlib.h presence... yes
checking for zlib.h... yes
checking for readline in -lreadline... yes
checking readline/readline.h usability... yes
checking readline/readline.h presence... yes
checking for readline/readline.h... yes
checking for OPENSSL... yes
checking openssl/crypto.h usability... yes
checking openssl/crypto.h presence... no
configure: WARNING: openssl/crypto.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: openssl/crypto.h: proceeding with the compiler's result
checking for openssl/crypto.h... yes
checking winscard.h usability... yes
checking winscard.h presence... no
configure: WARNING: winscard.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: winscard.h: proceeding with the compiler's result
checking for winscard.h... yes
checking pcsclite.h usability... yes
checking pcsclite.h presence... no
configure: WARNING: pcsclite.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: pcsclite.h: proceeding with the compiler's result
checking for pcsclite.h... yes
checking XSLTPROC requirement... ok
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating doc/Makefile
config.status: creating doc/tools/Makefile
config.status: creating etc/Makefile
config.status: creating src/Makefile
config.status: creating src/common/Makefile
config.status: creating src/libopensc/Makefile
config.status: creating src/libsm/Makefile
config.status: creating src/pkcs11/Makefile
config.status: creating src/pkcs11/versioninfo-pkcs11.rc
config.status: creating src/pkcs11/versioninfo-pkcs11-spy.rc
config.status: creating src/pkcs11/opensc-pkcs11.pc
config.status: creating src/pkcs15init/Makefile
config.status: creating src/scconf/Makefile
config.status: creating src/tests/Makefile
config.status: creating src/tests/regression/Makefile
config.status: creating src/tools/Makefile
config.status: creating src/tools/versioninfo-tools.rc
config.status: creating src/smm/Makefile
config.status: creating src/minidriver/Makefile
config.status: creating src/minidriver/versioninfo-minidriver.rc
config.status: creating src/minidriver/opensc-minidriver.inf
config.status: creating win32/Makefile
config.status: creating win32/versioninfo.rc
config.status: creating win32/versioninfo-customactions.rc
config.status: creating win32/winconfig.h
config.status: creating win32/OpenSC.iss
config.status: creating win32/OpenSC.wxs
config.status: creating MacOSX/Makefile
config.status: creating MacOSX/build-package
config.status: creating MacOSX/Distribution.xml
config.status: creating MacOSX/resources/ReadMe.html
config.status: creating MacOSX/resources/Welcome.html
config.status: creating config.h
config.status: executing depfiles commands
config.status: executing libtool commands

OpenSC has been configured with the following options:


Version:                 0.16.0
Version fix:             0
Version revision:        4
Git revision:            OpenSC-OpenSC_working_snapshot-4-g8702141, rev: 8702141, commit-time: 2016-10-10 13:13:14 -0400

Copyright:               OpenSC Project
Company:                 OpenSC Project
Company URL:             https://github.com/OpenSC
Comments:                Provided under the terms of the GNU Lesser General Public License (LGPLv2.1+).
Product name:            OpenSC smartcard framework
Product updates:         https://github.com/OpenSC/OpenSC/releases
Product URL:             https://github.com/OpenSC/OpenSC

User binaries:           /Library/OpenSC/bin
Configuration files:     /Library/OpenSC/etc
XSL stylesheets:         /opt/local/share/xsl/docbook-xsl

man support:             yes
doc support:             no
thread locking support:  yes
zlib support:            yes
readline support:        yes
OpenSSL support:         yes
PC/SC support:           yes
OpenCT support:          no
CT-API support:          no
minidriver support:      no
SM support:              yes
SM default module:       libsmm-local.so
DNIe UI support:         no
Debug file:              /tmp/opensc-debug.log

PC/SC default provider:  /System/Library/Frameworks/PCSC.framework/PCSC
PKCS11 default provider: opensc-pkcs11.so

Host:                    x86_64-apple-darwin15.6.0
Compiler:                clang
Preprocessor flags:       -DPKCS11_THREAD_LOCKING
Compiler flags:          -fno-strict-aliasing -maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -arch x86_64 -mmacosx-version-min=10.10 -Wall -Wextra -Wno-unused-parameter -Werror=declaration-after-statement
Linker flags:            
Libraries:               

READLINE_CFLAGS:         
READLINE_LIBS:           -lreadline 
ZLIB_CFLAGS:             
ZLIB_LIBS:               -lz
OPENSSL_CFLAGS:          -maes -mpclmul -msse2 -mssse3 -msse4 -msse4.2 -mrdrnd -mrdseed -Os -Ofast -I/opt/local/include
OPENSSL_LIBS:            -L/opt/local/lib -lssl -lcrypto
OPENCT_CFLAGS:           
OPENCT_LIBS:             
PCSC_CFLAGS:             -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/PCSC.framework/Versions/Current/Headers

. . . . . .
  CC       sc-test.o
In file included from base64.c:6:
In file included from ../../src/libopensc/asn1.h:29:
In file included from ../../src/libopensc/pkcs15.h:29:
In file included from ../../src/libopensc/aux-data.h:31:
In file included from ../../src/libopensc/internal.h:44:
../../src/libopensc/sc-ossl-compat.h:30:10: fatal error: 'openssl/opensslv.h' file not found
#include <openssl/opensslv.h>
         ^
1 error generated.
In file included from sc-test.c:15:
In file included from ./sc-test.h:4:
In file included from ../../src/libopensc/pkcs15.h:29:
In file included from ../../src/libopensc/aux-data.h:31:
In file included from ../../src/libopensc/internal.h:44:
../../src/libopensc/sc-ossl-compat.h:30:10: fatal error: 'openssl/opensslv.h' file not found
#include <openssl/opensslv.h>
         ^
make[4]: *** [base64.o] Error 1
. . . . .
```

dengert:
The OPENSSL_CFLAGS in your output has:  -I/opt/local/include
Can you verify if: /opt/local/include/openssl/opensslv.h is present?
Also look for the /opt/local/lib/pkgconfig/openssl.pc

Every version of OpenSSL has always had a openssl/opensslv.h 
It looks like your build environment is not finding it. 

These lines in your output also look strange:

checking for OPENSSL... yes
checking openssl/crypto.h usability... yes
checking openssl/crypto.h presence... no
configure: WARNING: openssl/crypto.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: openssl/crypto.h: proceeding with the compiler's result
checking for openssl/crypto.h... yes

The Warnings look like the preprocessor can't find the headers, but the compiler can.

Did you do a ./bootstrap in the OpenSC source when you updated the source?

dengert:
Was the first error from the sc-test.c compilation?  sc-test is one of the last things compiled,

I see in the tests/Makefile it has a OPENSSL_CFLAGS but it never get used. 
On my Ubuntu-16.4 system sc_test.c gets compiled to .sc-test.o,  It could be in my environment, an openssl/opensslv.h is found, but maybe not the one intended. But in your environment it does not fine any. 

There may have been some unexpected dependency on openssl that was not present before.

frankmorgner:
you included `sc-ssl-compat.h` in `internal.h` which is also used outside of libopensc, namely in the tests... Strange, this should have been caught by CI...

dengert:
@mouse07410 
Can you try adding AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) to src/tests/Makefile.am
and doing  ./bootstrap and the then a compile?



diff --git a/src/tests/Makefile.am b/src/tests/Makefile.am
index 2fcc686..271697a 100644
--- a/src/tests/Makefile.am
+++ b/src/tests/Makefile.am
@@ -6,6 +6,7 @@ EXTRA_DIST = Makefile.mak
 SUBDIRS = regression
 noinst_PROGRAMS = base64 lottery p15dump pintest prngtest
 
+AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS)
 AM_CPPFLAGS = -I$(top_srcdir)/src
 LIBS = \
        $(top_builddir)/src/libopensc/libopensc.la \

frankmorgner:
the better option would be to include openssl header (i.e. sc-compat-ssl.h) only where it's needed.

dengert:
@frankmorgner  Yes it is strange. My system has a /usr/include/openssl which is on the default include paths but on his Mac system the openssl include may not be on one of the default include search paths. 

looking at all the OpenSC Makefile.am files, many have have AM_CFLAGS  ./src/tests did not. 

$ find . -name Makefile.am -exec grep "AM_CFLAGS" {} \; -print
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(PTHREAD_CFLAGS)
onepin_opensc_pkcs11_la_CFLAGS = -DMODULE_APP_NAME=\"onepin-opensc-pkcs11\" $(AM_CFLAGS)
./src/pkcs11/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(OPTIONAL_READLINE_CFLAGS)
./src/smm/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(OPTIONAL_READLINE_CFLAGS)
./src/libsm/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS)
./src/pkcs15init/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(OPTIONAL_READLINE_CFLAGS) $(PTHREAD_CFLAGS)
./src/tools/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(OPTIONAL_OPENCT_CFLAGS) \
./src/libopensc/Makefile.am
AM_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS)
./src/tests/Makefile.am

mouse07410:
> The OPENSSL_CFLAGS in your output has: `-I/opt/local/include`

Correct.

> Can you verify if: /opt/local/include/openssl/opensslv.h is present?
> Also look for the /opt/local/lib/pkgconfig/openssl.pc

```
$ ll /opt/local/include/openssl/opensslv.h
-rw-r--r--  1 root  admin  3941 Sep 26 12:03 /opt/local/include/openssl/opensslv.h
$ cat /opt/local/lib/pkgconfig/openssl.pc
prefix=/opt/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: OpenSSL
Description: Secure Sockets Layer and cryptography libraries and tools
Version: 1.0.2j
Requires: libssl libcrypto
$
```

> Did you do a ./bootstrap in the OpenSC source when you updated the source?

Yes.

> Was the first error from the sc-test.c compilation? sc-test is one of the last things compiled

Yes, it appears so. The compilation stops on the first error.

Unrelated issue (found as I'm scrubbing the build output): problem with `pkcs15-tool.c`, which I will also report separately:
```
. . . . .
Making all in tools
  CC       opensc-tool.o
  CC       util.o
  CC       opensc-explorer.o
  CC       pkcs15-tool.o
pkcs15-tool.c:345:2: warning: implicitly declaring library function 'snprintf' with type 'int (char
      *, unsigned long, const char *, ...)' [-Wimplicit-function-declaration]
        snprintf(title, sizeof(title), "%s (%lu bytes): ", kind, (unsigned long) data_len);
        ^
pkcs15-tool.c:345:2: note: include the header <stdio.h> or explicitly provide a declaration for
      'snprintf'
pkcs15-tool.c:1149:17: warning: implicitly declaring library function 'strdup' with type
      'char *(const char *)' [-Wimplicit-function-declaration]
                return (u8 *) strdup(pincode);
                              ^
pkcs15-tool.c:1149:17: note: include the header <string.h> or explicitly provide a declaration for
      'strdup'
. . . . .
```

Hmm... Weird. I've rerun the compilation to get more info - and this time it completed successfully.

Looks like I should apologize... In three days I will repeat this upgrade on three other machines, and if there are any problems - report here. Otherwise, this problem seems to have disappeared.

dengert:
sc-compat-ssl.h first tests if ENABLE_OPENSSL is defined.  The tests (and tools) include config.h and internal.h and they should expect that openssl may be used internally or in some future modidfication.

dengert:
@mouse07410  Please see #888 for the fix for sc-test.c

mouse07410:
#888 is good.

PR #864, opened 2016-08-30
OpenPGP card small fixes

maciejsszmigiero:
This pull request contains small fixes related to OpenPGP driver, picked from bigger pull request #850,
as @frankmorgner has suggested.

Specifically it contains the following changes:
1) Remove raw RSA algo flag from OpenPGP card since it doesn't support it.

2) Make OpenPGP card user/signature PIN order match expected by _get_auth_object_by_name() function in OpenSC PKCS#11 framework.

3) Strip execute permission bits from two .c files that have them set,

4) Improve handling of OpenPGP card PIN change and unblock commands,
provide more debug messages for user in unsupported cases and use correct
reference to PW1-mode 2 (82) in these commands.


frankmorgner:
Looks good. However, I can't tell whether the internal magic of changing the pin length and pin reference is breaking something. @hongquan could you please verify the changes?

hongquan:
Looks good to me.

frankmorgner:
Is there any negative impact by changing the pin length and pin reference in your opinion/tests.

maciejsszmigiero:
> Is there any negative impact by changing the pin length and pin reference in your opinion/tests?

PIN length is zeroed for version 1.x cards since they don't support explicit PIN in change or unblock
commands and would fail if you try to supply it with command (they rely on implicit result of last
verify command instead).

With regard to changing PIN reference:
Existing code has already changed PIN reference by or'ing it with 0x80 (unconditionally).

If you try to do PIN change or unblock with PIN reference of 0x82 on version 2.x card it will reject
such command, so this use case wasn't working anyway.

At the same time, encryption and authentication keys are protected by this PIN and its reference 
needs to be specified as 0x82 during verify command (if you verify it as 0x81 these keys won't get
unlocked) so you can't simply remove this PIN reference and use only 0x81.

src/libopensc/pkcs15-openpgp.c
```diff
@@ -216,7 +216,7 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
 
 		pin_info.auth_type = SC_PKCS15_PIN_AUTH_TYPE_PIN;
 		pin_info.auth_id.len      = 1;
-		pin_info.auth_id.value[0] = i + 1;
+		pin_info.auth_id.value[0] = pin_cfg[i].reference;
```
hongquan: Could you please explain me more about this change?

maciejsszmigiero: Hi Nguyễn,

> Could you please explain me more about this change?

Did you read commit message for this change?
It's commit 0a6c1c4fb3fee9b27cd71fd97e7e8b40a880286d .

Issue #750, opened 2016-05-03
Don't always lock a token with the tools


frankmorgner:
### Expected behaviour

OpenSC tools should only lock a card (`sc_lock`) when needed, i.e. when there is a transaction that must not be interrupted by a different smart card application.
### Actual behaviour

`util_connect_card` always locks the card when connecting to it https://github.com/OpenSC/OpenSC/blob/master/src/tools/util.c#L159. This utility function is used by all tools.

On Windows the maximum time for locking is on 5 seconds, by default. This leads to unwanted interrupts in some cases, see https://github.com/OpenSC/OpenSC/issues/703, for example.
### Steps to reproduce

See https://github.com/OpenSC/OpenSC/issues/703

frankmorgner:
All general tools are using the new locking mechanism now.

Issue #703, opened 2016-03-04
"stale" locks lead to card reset on windows


philipWendland:
Hi all,

I report this on behalf of a different user, i.e. I can hardly test this. Maybe he will report back here.

TL;DR: A long-term sc_lock in https://github.com/OpenSC/OpenSC/blob/master/src/tools/util.c#L159 leads to a card reset after 5 seconds when entering the user PIN in Windows 8.1 (According to the user, it is fine in W7).

https://github.com/OpenSC/OpenSC/blob/master/src/tools/util.c#L159 calls sc_lock which is maintained over a long time. The duration of the lock includes the time at which the user PIN is entered when using `pkcs15-init --create-pkcs15`.

sc_lock calls SCardBeginTransaction. See the following documentation of this function:
```
Remarks

If a transaction is held on the card for more than five seconds with no operations happening on that card, then the card is reset. 
```
https://msdn.microsoft.com/de-de/library/windows/desktop/aa379469%28v=vs.85%29.aspx
I.e. the reset after 5 seconds is documented.

Subsequently, the following will happen when OpenSC tries to send the next APDU:
```
2016-02-04 19:25:23.081 Dell Smart Card Reader Keyboard 0:SCardTransmit/Control failed: 0x80100068
```

This error code means SCARD_W_RESET_CARD.

I wonder why the call to sc_lock in https://github.com/OpenSC/OpenSC/blob/master/src/tools/util.c#L159 is present. Does anyone see a reason for this?

martinpaljak:
This is a real annoyance. Either one needs to re-do the code to support the MS concept of only short-lived transactions or if uninterrupted access is expected during a longer period (for whatever reasons) then exclusive access can be used. Check `connect_exclusive` in opensc.conf for the last option.

frankmorgner:
Though MS's decision on this can be debated http://permalink.gmane.org/gmane.comp.encryption.opensc.devel/16416, I think that `sc_lock()` should not be used in `util_connect_card()`. Locking was moved from the application level (e.g. opensc-explorer) into `utils.c` in beb86f351b05bdf36668887301654de86d19af35 without explanation without ever calling `sc_unlock()` again.

However, we still need `sc_lock()` somewhere to avoid accessing a card within a bad state. For example in `pkcs15-init` you don't want your card to be read out while the PKCS#15 metadata is not yet fully written.

philipWendland:
[...]

PR #868, opened 2016-09-06
Do not lock card early (in function from utils.c) #750 #703

szszszsz:
An attempt to solve #750 and #703.
Windows 8.0+ specific issue.
Please advise would additional changes be required.


szszszsz:
Compiled stable `0.16` release with this change is available under:
https://ci.appveyor.com/project/szszszsz/opensc/build/0.16.0.3

frankmorgner:
Disabling sc_lock/sc_unlock only for the command line tools (i.e. in util.c) may be OK as workaround, but disabling it completely on windows is unnecessary and generally not a good idea.

The real fix would be to disable locking in util.c and do this within each command line tool for each coherent transaction separately. So limiting your PR to util.c would at least go one step down this path.

szszszsz:
Hi! Thank you for review.
I have not disabled the locking completely. It is just not done at the beginning of using the tool and still done for every volatile transaction. I believe the lock in `util.c` is redundant - please check both logs linked below for execution of `sc-hsm-tool.exe`:
`Lock` log: http://pastebin.com/zJ7ErQiF
Full log: http://pastebin.com/TCNPs5j7
Use of mentioned `util_connect_card` function: http://pastebin.com/EyE1MP7e

PS I will add a summary of changes

PR #875, opened 2016-09-13
Allow tools to choose if lock is required

CardContact:
This pull request aims at fixing #750. It's an alternative to #868.

The API remains backwards compatible and tools can now choose in util_connect_card_ex() if they want to lock cards immediately or not.

### DIFF
--- a/src/tools/sc-hsm-tool.c
+++ b/src/tools/sc-hsm-tool.c
@@ -1646,7 +1646,7 @@ int main(int argc, char * const argv[])
 		sc_ctx_log_to_file(ctx, "stderr");
 	}
 
-	r = util_connect_card(ctx, &card, opt_reader, opt_wait, verbose);
+	r = util_connect_card_ex(ctx, &card, opt_reader, opt_wait, 0, verbose);
 	if (r != SC_SUCCESS) {
 		if (r < 0) {
 			fprintf(stderr, "Failed to connect to card: %s\n", sc_strerror(err));
@@ -1687,7 +1687,6 @@ fail:
 	err = 1;
 end:
 	if (card) {
-		sc_unlock(card);
 		sc_disconnect_card(card);
 	}
 	if (ctx)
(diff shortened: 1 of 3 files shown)


frankmorgner:
Looks better than #868 since no special handling of `card->lock_count` is required and because it allows incrementallly fixing #750 for all tools.

frankmorgner:
@szszszsz do you agree?

szszszsz:
Hi! Indeed it is better in the long run to fix and check each tool separately. Would not all of them require such change (which was done in pull #868)?

CardContact:
Long term all tools need to be reviewed if they require locking and if yes, how the lock is released during user interaction. Not all tools do user interaction.

We can only speak for the sc-hsm-tool, which does not require locking. Other tools have other maintainers and require specific cards for testing. We'd better leave it to them to make the appropriate changes and do regression testing.

frankmorgner:
@CardContact can you do some testing with `pkcs15-tool`, which is also working with your card? Looking into the code, I can see that locking is mostly done in the underlying pkcs15 layer (i.e. it is not required directly after connecting to the card) with one exception: Locking needs to be added to `--test-update` and `--update`.

dengert:
As a side note: After an update of the card including a PIN change should a card_reset be done? This would force any other application that is accessing the card to invalidate anything it may have cached, including cached pins.

frankmorgner:
@dengert, I'm not sure if I completely understand the logic behind the update. If someone uses this and is able to verify this with a reset, I'd be happy to merge this (in a different PR)

dengert:
Multiple long running applications, (firefox, ssh, rdesktop or some other application) may be using the card that is being updated. They have no way to know if an update was done. They may continue to use old  cached certificates or old cached PINS.  If enough of these applications try and use an old PIN, they will each lower the tries_left count and could even endup locking the PIN.  Reset (or card removal) is the only way they get notified update may have been done.

PR #876, opened 2016-09-20
Coverity remarks

Jakuje:
If interested, I can provide specific messages from Coverity for each change.

It also found one flaw in the change #852 (already accepted) when `free()`ing data, which was caused by my modifications. It is also covered in this pull request.

### DIFF
--- a/src/libopensc/pkcs15-cert.c
+++ b/src/libopensc/pkcs15-cert.c
@@ -273,26 +273,27 @@ sc_pkcs15_get_extension(struct sc_context *ctx, struct sc_pkcs15_cert *cert,
 			LOG_FUNC_RETURN(ctx, r);
 
 		/* is it the RN we are looking for */
-		if(sc_compare_oid(&oid, type) != 0) {
+		if (sc_compare_oid(&oid, type) != 0) {
 			if (*ext_val == NULL) {
-				*ext_val= val;
+				*ext_val = val;
 				val = NULL;
 				*ext_val_len = val_len;
+				/* do not free here -- return the allocated value to caller */
 			}
 			else {
 				*ext_val_len = MIN(*ext_val_len, val_len);
 				memcpy(*ext_val, val, *ext_val_len);
+				free(val);
 			}
 
 			if (is_critical)
 				*is_critical = critical;
 
 			r = val_len;
-			free(val);
 			LOG_FUNC_RETURN(ctx, r);
 		}
-		free(val);
 	}
+	free(val);
 
 	LOG_FUNC_RETURN(ctx, SC_ERROR_ASN1_OBJECT_NOT_FOUND);
 }
--- a/src/pkcs11/misc.c
+++ b/src/pkcs11/misc.c
@@ -178,11 +178,10 @@ CK_RV push_login_state(struct sc_pkcs11_slot *slot,
 	struct sc_pkcs11_login *login = NULL;
 
 	if (!sc_pkcs11_conf.atomic || !slot) {
-		r = CKR_OK;
-		goto err;
+		return CKR_OK;
 	}
 
-	login = (struct sc_pkcs11_login *) malloc(sizeof *login);
+	login = (struct sc_pkcs11_login *) calloc(1, sizeof *login);
 	if (login == NULL) {
 		goto err;
 	}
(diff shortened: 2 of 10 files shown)


frankmorgner:
Thanks! No need to this into individual commits.

src/libopensc/pkcs15-cert.c
```diff
@@ -273,26 +273,27 @@ sc_pkcs15_get_extension(struct sc_context *ctx, struct sc_pkcs15_cert *cert,
 			LOG_FUNC_RETURN(ctx, r);
 
 		/* is it the RN we are looking for */
-		if(sc_compare_oid(&oid, type) != 0) {
+		if (sc_compare_oid(&oid, type) != 0) {
 			if (*ext_val == NULL) {
 				*ext_val= val;
 				val = NULL;
 				*ext_val_len = val_len;
+				/* do not free here -- leaking the allocated value */
```
frankmorgner: what's the reason for intentionally leaking `val`?

src/pkcs11/misc.c
```diff
@@ -202,9 +201,11 @@ CK_RV push_login_state(struct sc_pkcs11_slot *slot,
 	r = CKR_OK;
 
 err:
-	if (r != CKR_OK && login) {
-		sc_mem_clear(login->pPin, login->ulPinLen);
-		free(login->pPin);
+	if (login) {
```
frankmorgner: `login` can be inside the `slot->logins` queue here. You could to the following to avoid this:
```
if (0 > list_append(&slot->logins, login)) {
		goto err;
	}
login = NULL;
```

src/libopensc/pkcs15-cert.c
Jakuje: This is intention as described in the header of the function:

    if *ext_val is NULL, sc_pkcs15_get_extension will allocate space for ext_val.

which means we are leaking the memory allocated by the call of `sc_asn1_decode()`.

src/pkcs11/misc.c
Jakuje: There were two separate problems reported by Coverity in context of this chunk of code:

    Error: CLANG_WARNING: [#def125]
    opensc-0.16.0/src/pkcs11/misc.c:211:9: warning: Potential leak of memory pointed to by 'login'

and

    Error: CLANG_WARNING: [#def124]
    opensc-0.16.0/src/pkcs11/misc.c:206:3: warning: Function call argument is an uninitialized value

You are probably right about the first one. Coverity missed the case the `login` got stored in the `slot->logins`. On success we should not free the variable. Thanks for catching that (added in the new commit).

The second problem was when the `sc_mem_alloc_secure()` failed, we were trying to call `sc_mem_clear()` with undefined `login->ulPinLen` (therefore the second check).

src/libopensc/pkcs15-cert.c
frankmorgner: I see. Maybe rephrasing this to something like `do not free here -- return the allocated value to caller` would be more appropriate.

src/pkcs11/misc.c
```diff
@@ -178,8 +178,7 @@ CK_RV push_login_state(struct sc_pkcs11_slot *slot,
 	struct sc_pkcs11_login *login = NULL;
 
 	if (!sc_pkcs11_conf.atomic || !slot) {
-		r = CKR_OK;
-		goto err;
+		return CKR_OK;
 	}
 
 	login = (struct sc_pkcs11_login *) malloc(sizeof *login);
```
frankmorgner: actually this should be `calloc` instead of `malloc` to avoid a bad free below in the error handling.

src/pkcs11/misc.c
```diff
@@ -199,12 +198,15 @@ CK_RV push_login_state(struct sc_pkcs11_slot *slot,
 		goto err;
 	}
 
+	login = NULL
 	r = CKR_OK;
 
 err:
 	if (r != CKR_OK && login) {
```
frankmorgner: with `login = NULL` the check for `r != CKR_OK` could be removed here.

src/pkcs11/misc.c
Jakuje: That would be even more elegant. Thanks for comments. I will fix that.

src/libopensc/pkcs15-cert.c
Jakuje: Thanks for comment. Reworded the comment to be more clear.

PR #894, opened 2016-10-26
prkey_fixup_rsa changes for OpenSSL-1.1.0

dengert:
As reported in https://github.com/OpenSC/OpenSC/issues/810#issuecomment-255855491 and https://github.com/OpenSC/OpenSC/issues/810#issuecomment-256098736
the prkey_fixup_rsa() in pkcs15init/pkcs15-lib.c was reporting errors in the GETBN macro with some compilers. The code was computing missing RSA private key idmp1, idmq1 and/or iqmp numbers.
It was then saving them to static 256 byte u8 arrays. A point to these was then added to the sc_pkcs15_prkey_st. Point a pointer to a static in an allocated structure is not a good idea.

This patch compiles on OpenSSL versions 0.9.7. 0.9.8, 1.0.1, 1.0.2 and 1.1.0a and on libressl-2.5.0
I am unable to test the code. It is used when writing a privare RSA key to a card, and from teh comments GPK and cryptoflex cards need idmp1, idmq1 and/or iqmp written to the card.

This patch compiles on OpenSSL versions 0.9.7. 0.9.8, 1.0.1, 1.0.2 and 1.1.0a and on libressl-2.5.0

Remove restrictions in prkey_fixup_rsa:
  /\* Not thread safe, but much better than a memory leak _/
  /_ TODO put on stack, or allocate and clear and then free */
Compute dmp1, dmp1 and/or iqmp if not in sc_pkcs15_prkey_rsa

Remove the GETBN macro that was causing problems.

 Changes to be committed:
    modified:   src/pkcs15init/pkcs15-lib.c

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -46,6 +46,7 @@
 #endif
 #include <assert.h>
 #ifdef ENABLE_OPENSSL
+#include <openssl/opensslv.h>
 #include <openssl/bn.h>
 #include <openssl/evp.h>
 #include <openssl/pem.h>
@@ -55,6 +56,7 @@
 #include <openssl/pkcs12.h>
 #endif
 
+#include "libopensc/sc-ossl-compat.h"
 #include "common/compat_strlcpy.h"
 #include "common/libscdl.h"
 #include "libopensc/pkcs15.h"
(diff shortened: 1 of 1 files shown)


emunicio:
Great, I will test it ASAP

viktorTarasov:
No problem when importing PKCS#12 into IAS/ECC card.

viktorTarasov:
No objections for PR -- afais, the _dmp1_, _dmq1_, _iqmp_ are calculated in the same manner as openssl do.

PR #896, opened 2016-11-01
adding a Coolkey support

Jakuje:
This pull request is adding a support for [Coolkey](http://pki.fedoraproject.org/wiki/CoolKey) cards (applets) into OpenSC. It conflicts with the #841 (the same numbers in `SC_CARD_TYPE_*` -- either will get earlier in, the other will need to be rebased).

I tried to address the respective issues pointed out in the previous pull request, but still I might have missed something. For example I found very limiting SimCList with only single iterator (therefore the ugly workaround).

### DIFF
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -35,7 +35,7 @@ libopensc_la_SOURCES = \
 	card-cardos.c card-tcos.c card-default.c \
 	card-mcrd.c card-starcos.c card-openpgp.c card-jcop.c \
 	card-oberthur.c card-belpic.c card-atrust-acos.c \
-	card-entersafe.c card-epass2003.c \
+	card-entersafe.c card-epass2003.c card-coolkey.c \
 	card-incrypto34.c card-piv.c card-muscle.c card-acos5.c \
 	card-asepcos.c card-akis.c card-gemsafeV1.c card-rutoken.c \
 	card-rtecp.c card-westcos.c card-myeid.c card-ias.c \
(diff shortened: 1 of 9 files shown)


frankmorgner:
Hmm, one of the weirder cards I have seen.

The Changes look good at first glance. Just out of curiosity, the nonce that's returned in exchange for the PIN verification. does it need to be added to every subsequent APDU or only to re-authenticate the user (aka session PIN)?

Jakuje:
To the CI failures:

```
Users/travis/build/OpenSC/OpenSC/OpenSC.tokend/build/opensc-src/libopensc/cardctl.h:1071:34: error: expected member name or ';' after declaration specifiers

        sc_cardctl_coolkey_attribute_t *template; /* in paramter */

        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ^
```

seems like related to the Tokend and I am not sure what wrong here nor how is tokend build from the current opensc.

    card-coolkey.c(41): fatal error C1083: Cannot open include file: 'types.h': No such file or directory

This should say `<sys/types.h>`

    pkcs15-coolkey.c(255) : error C2054: expected '(' to follow 'inline'

should not use `inline` to build on Windows.

    card-coolkey.c:753:2: error: unknown type name ‘ushort’

should say `unsigned short` for mingw

    card-coolkey.c:1996:121: error: ‘COMPRESSION_AUTO’ undeclared (first use in this function)

should be conditionality executed using `#ifdef ENABLE_ZLIB`

Jakuje:
@frankmorgner Yes, the coolkey is designed, that the nonce is used instead of cached pin to make it more secure. The nonce is provided on every APDU that requires authentication. From this you can notice [the line](https://github.com/OpenSC/OpenSC/pull/896/files#diff-da302d201d1743dccf53ebb27c6ad95fR492)

	p15card->opts.use_pin_cache = 0; 

which is preventing the pin caching for this card (because it is useless here) and reduces an attack surface. As we discussed in previous pull request, it is not ideal solution to write user settings, but I don't know about any other way to achieve that.

viktorTarasov:
Your PR has to be rebased to one commit and
also has to be changed the name of _template_ member from _sc_cardctl_coolkey_find_object_ data type -- it seems that travis job for _OSX_ fails for this reason.

Look the last commit in [this branch](https://github.com/viktorTarasov/OpenSC-SM/commits/pr-896-jjelen-coolkey). This commit is your PR transformed as suggested above -- no travis errors for this commit. If you are agree with, I'll merge it to _master_.

Jakuje:
@viktorTarasov thank you. It did the job and Travis is happy. I also squashed the changes together.

PR #899, opened 2016-11-10
Restore blocking wait-for-slot functionality

Jakuje:
* Add configure-time dependency on pcsclite (already from comments in reader-pcsc)
 * The functionality is already supported in pcsc-lite

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -583,7 +583,10 @@ if test "${enable_pcsc}" = "yes"; then
 	if test "${WIN32}" != "yes"; then
 		PKG_CHECK_EXISTS(
 			[libpcsclite],
-			[PKG_CHECK_MODULES([PCSC], [libpcsclite])]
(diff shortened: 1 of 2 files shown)


frankmorgner:
Apple's sources indicate that it uses [pcsc-lite version 1.4.0 on OS X 10.11](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.11.sdk/System/Library/Frameworks/PCSC.framework/Versions/A/Headers/pcsclite.h#L246), which is well behind 1.6.5. Additionally, Apple's own implementation starting on Mac OS 10.12 and Windows should to be tested.

Jakuje:
Thank you for the comment. Well, then we should create a backup for older versions. I modified the pull request with possible fallback for older PCSC-Lite versions.

Also linking the initial message on the mailing list reasoning for this change:
https://sourceforge.net/p/opensc/mailman/message/35478574/

Issue #898, opened 2016-11-05
Derive secret with ECC


Ovomaltine85:
I generated a new keypair on HSM (softhsm2) with the following code:

`pkcs11-tool --module /usr/local/lib/softhsm/libsofthsm2.so -l --pin 1234 --slot 1 --keypairgen --key-type EC:secp384r1 --id 1111 --label testtoken`

With this private key stored on the hsm I want to derive a shared secret with another pubkey. Therefor I converted the public key from a x509 with openssl ec to DER format.

When I use the created pubkey.der to derive a shared secret I get the following error.
`pkcs11-tool --module /usr/local/lib/softhsm/libsofthsm2.so --slot 1 -l --pin 1234 --id 1111 --derive --input-file pubkey.der -m ECDH1-DERIVE --output-file shared`

Error:
`Using derive algorithm 0x00001050 ECDH1-DERIVE
error: PKCS11 function C_DeriveKey failed: rv = CKR_TEMPLATE_INCOMPLETE (0xd0)
Aborting.`

ECDH1-DERIVE is a valid supported mechanism by the HSM.

Why is the CKR Template incomplete and how can I solve this problem to derive a shared secret?

Regards

dengert:
pkcs11-tool.c in routine derive_ec_key builds the newkey_template with:
CKA_TOKEN=FALSE  (Session only)
CKA_CLASS=CKO_SECRET_KEY
CKA_KEY_TYPE=CKK_GENERIC_SECRET
CKA_ENCRYPT=TRUE
CKA_DECRYPT=TRUE

Examples from PKCS\#11 2.20 in section 11.14  and PKCS\#11 2.40 under C_Derive would indicate that above has all the required attributes in the template. 

The problem might be coming  from SoftHSM.cpp:

 8866                 case CKK_GENERIC_SECRET:
 8867                         if (byteLen == 0)
 8868                         {
 8869                                 INFO_MSG("CKA_VALUE_LEN must be set");
 8870                                 return CKR_TEMPLATE_INCONSISTENT;
 8871                         }

I don't see any PKCS\#11 requirement that during a derive for a  CKK_GENERIC_SECRET the length must be supplied. In particular, the size if the  CKK_GENERIC_SECRET is based on the key size use to derive it.

After the key is derived, the calling application can then use PKCS#11 calls to use the key, or extract it, if its is extractable.

 I think this is a SoftHSM issue, not a OpenSC pkcs11-tool issue. 

[...]

PR #901, opened 2016-11-14
Pkcs11-tool.c changes to accommodate ECDH operations using SoftHSM.

dengert:
PKCS#11 v2.20 in not clear on the format of the public key of the other party
pased during ECDH key derivation. Some implementations (OpenSC) pass just the value
of the public key (RAW), while others (SoftHSM) pass an ASN.1 DER encoded OCTET_STRING.

PKCS$11 v2.40 points out this problem and says implementations must support the
RAW format and may also support the DER format.

To allow pkcs11-tool.c to work with ECDH derivation and using the current libsofthsm2.so
a new parameter was added to pkcs11-tool, --derive-pass-der.

Also added to the template for the new key were:

CKA_SENSITIVE = false
CKA_EXTRACTABLE = true
CKA_VALUE_LEN = size of key to be derived.

OpenSC currently only support derivation of ECDH session keys, (CKA_TOKEN = false)
The derived key must be CK_KEY_TYPE = CKK_GENERIC_SECRET
Additional changes could be made to support AES or DES3 keys.

It is not clear if there is a need to support CKA_TOKEN =  true which says the
derived key must be on the hardware token. For ECDH, these keys are short lived.

 On branch pkcs11-tool-simple-ecdh
 Changes to be committed:

This PR is response to #898 

Also see:
https://github.com/opendnssec/SoftHSMv2/pull/266#issuecomment-260441348
but the SoftHSM changes are on hold.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -140,6 +140,7 @@ enum {
 	OPT_LOGIN_TYPE,
 	OPT_TEST_EC,
 	OPT_DERIVE,
+	OPT_DERIVE_PASS_DER,
 	OPT_DECRYPT,
 	OPT_TEST_FORK,
 	OPT_GENERATE_KEY,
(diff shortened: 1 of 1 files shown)


Ovomaltine85:
I tested it with SoftHSM 2.2.0rc1 and ECC 384 bit keys. Works fine!

PR #903, opened 2016-11-16
Dnie30

rickyepoderi:
Add support for DNIe 3.0. See bug #810 for more information.


frankmorgner:
Could you please look at the CI build error?

rickyepoderi:
There are only two types of errors:

* [This one](https://travis-ci.org/OpenSC/OpenSC/jobs/176459944) that exited with 0 but (I don't know why) finally timed out. So I think something went wrong and the job didn't notice it was finished.

* All the other errors (4) are in the other type of jobs and they are associated to the same machine (Environment: VSVER=10; Configuration: Debug; Platform: x64). The error (for example [this job](https://ci.appveyor.com/project/LudovicRousseau/opensc/build/0.16.0.966/job/993jney6v3lt4lsa) is alway the same: "'nmake' is not recognized as an internal or external command, operable program or batch file". So the program does not compile after the configure.

@frankmorgner  I don't know, the two types seem not related to my changes. Take in mind I just changed pure C code not even adding any new include. 

Thanks!

Issue #824, opened 2016-07-11
Crash attempting to query AIS card


programatix:
### Expected behaviour

What should happen?
I'm not expecting a crash
### Actual behaviour

What happens instead?
The application crashed (opensc-explorer.exe, opensc-tool.exe and etc)
### Steps to reproduce

I'm using Windows 10. I tried OpenSC 32-bit and 64-bit with same result.
1. added custom ATR using "Gemalto MultiApp IAS/ECC v1.0.1" config in opensc.conf
2. execute "opensc-tool --atr -v -v -v"
3. crashed with message "OpenSC command line tool has stopped working."

Additional info
If I set "debug_file" in opensc.conf, it doesn't crash.
### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

https://gist.github.com/programatix/96396074e23815971af14b7f44e7ec2d

frankmorgner:
You could try running this on Unix in gdb or similar. This would tell the exact place where the program crashes...

@viktorTarasov ?

programatix:
I traced the source code and I think I can safely say that it crashed when calling module_init from dll-tool.dll.

programatix:
I meant smm-local.dll

viktorTarasov:
On Win7-64x with the latest release I do not have such problems.
Will try to find the Win10.

viktorTarasov:
I do not have problems on Win10-64 with the [_latest release_](https://github.com/OpenSC/OpenSC/releases/download/0.16.0/OpenSC-0.16.0-win64_vs12-Release.msi).

Have you used the same release?

viktorTarasov:
> If I set "debug_file" in opensc.conf, it doesn't crash.

Well, taking into consideration this remark, I also have segfault.

viktorTarasov:
Dedicated PR #910 created.

PR #910, opened 2016-11-23
#824: tools: fix segfault with verbose log into 'stderr'

viktorTarasov:
Issue #824

In Windows, file handles (including 'stderr', 'stdout') can not be shared
between DLL-s, and so, the log handle (File *), defined in one module, cannot
be reused in another.

That is the situation when, for example, the SM is processed
in external, dynamically loadable module as it currently implemented for
IAS/ECC card.

That's for the configuration option 're-open log file on every log message' was
introduced.

This 're-open' logic has not been tested in the particular case of opensc-*
tools used with verbose log into 'stderr' -- in dynamically loaded module the
'stderr' handle, defined in the 'main' module, was not recognized as 'stderr'
and there was an attempt to close it.

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -21,14 +21,14 @@ app default {
 	#
 	# debug_file = @DEBUG_FILE@
 
-	# Re-open debug file  (used in WIN32)
+	# Reopen debug file at the every debug message.
 	#
-	# In Windows, file handles can not be shared between DLL-s,
-	#  each DLL has a separate file handle table.
-	# For that reason reopen debug file before every debug message.
+	# For Windows it's forced to 'true'. The reason is that
+	#   in Windows file handles can not be shared between DLL-s,
+	#   each DLL has a separate file handle table.
 	#
-	# Default: true
-	# reopen_debug_file = false;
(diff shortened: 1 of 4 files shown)


frankmorgner:
Forcing this option to be enabled by default on Windows adds a considerable overhead. Just look at Windows debug logs, where printing many lines is already quite slow.

- How about *not* dynamically loading the external module and instead use static linking?
- What about re-opening the file handle in the SM module only instead of every single time someone writes a line to the log?

frankmorgner:
I just realized that re-opening the file for every line in the log is the current default, which could explain the performance issues mentioned above. This default behavior (including your suggestion to force it to be enabled on Windows) is a blunder, I think.

- The default should be `false` and configuratble by the user.
- Within OpenSC we need to find a solution that does not require the file to be opened every time (i.e. by using static linking or by reopening only within the dll in question; see [above comment](https://github.com/OpenSC/OpenSC/pull/910#issuecomment-263696036))

frankmorgner:
I implemented the alternative approach in https://github.com/OpenSC/OpenSC/pull/913. Please have a look and test it with your card, thanks!

Issue #906, opened 2016-11-19
OpenSC continuous-integration/travis-ci/pr for mingw  are timing out after 10 minutes 


dengert:
### Expected behaviour

PR builds for the last few days are timing out with the message below. 

### Actual behaviour

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated

viktorTarasov:
Do we really need the cygwin cross compilation?

At least, looking onto the current situation, we can make the results of cross-compilation as the ones without impact onto the general build status.

frankmorgner:
Yes, there are people that use this.

dengert:
How do we fix the build/tracix-ci system so it does not time out?

frankmorgner:
Don't know for sure. Maybe file a bug at Travis, maybe reproduce the Travis build locally?

frankmorgner:
Maybe it's related to https://www.traviscistatus.com/incidents/nq0sf0srvx8s

frankmorgner:
Found the problem.

By the way, cross compilation with mingw does **not** require cygwin.

Issue #791, opened 2016-06-07
Recent changes to the CI environment


frankmorgner:
I just noticed some changes that I'm not perfectly comfortable with. However, they are in some regard a matter of taste, so I hope to get some input whether you agree or don't agree...
### Appveyor Configuration

bd1e7aa03297bcd63f2e88b6cc49e481ae418a65 changes the configuration to only use Visual Studio 2013. I think we should test all configurations that are known to work, i.e. we should re-enable all other configurations. @viktorTarasov what was the intent to remove VS2010/VS2015? Also, I'd like to go back to APPVEYOR_BUILD_VERSION in the naming convention as it generates a unique name of the installer per build.
### Coverity Scan

With 88f85a2ef1bc6009809f8edc815aee9bf63a7fa3 analysis was bound to the branch `towards-opensc-0.16.0` which doesn't exist anymore. I suggest to use a dedicated branch `coverity_scan` instead of constantly changing the configuration. When an analysis is needed (e.g. for a new release) I think we should push there even allowing to overwrite the git history. So nobody should rely on a consistent `coverity_scan` branch...
### Github Deployment

94e43b023d57fbf1f7f4d2f2d43a50d8ed329db1 lets AppVeyor export its binaries as Github releases. I am not sure if that's the right step. I think most people are expecting an official binary to come from the OpenSC team, instead of being auto-generated from some third party. I, personally, like to trust people instead of systems. At the very least the current setup has to be made explicit.

martinpaljak:
1) I'd prefer to see the output of compilation from as many compilers as possible. There might be dependant components that only function with a specific version. Yet the latest version of a platform compiler should always be tested, even if not used for the binaries.
2) yes, push -f to a dedicated branch keeps everything simple, and keeps it a "configure once and forget" service (combine with a timed "twice a week" push)
3) Ideally the releases could be signed by someone from the team, if they are not, it doesn't really matter from where the binaries come from, a cloud buildhost is *almost* as good as anything else.

viktorTarasov:
@frankmorgner 
> what was the intent to remove ...

As far as I remember it do not worked properly for the other configurations, also it was very long. So, for the release-preparation branch I reduced it to one configuration. After merge his configuration has been transfered to _master_.

We can restore it.

@frankmorgner 
> coverity scan

According to travis configuration coverity-scan has to be done for _master_.

@frankmorgner 
>  I, personally, like to trust people instead of systems ...

[...]

PR #911, opened 2016-11-23
allow failed cross-builds in travis

viktorTarasov:
Results of the cross-build jobs have no impact onto the build status.

Proposal for issue #906

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
     - secure: "UkHn7wy4im8V1nebCWbAetnDSOLRUbOlF6++ovk/7Bnso1/lnhXHelyzgRxfD/oI68wm9nnRV+RQEZ9+72Ug1CyvHxyyxxkwal/tPeHH4B/L+aGdPi0id+5OZSKIm77VP3m5s102sJMJgH7DFd03+nUd0K26p0tk8ad4j1geV4c="
 
 matrix:
+  fast_finish: true
   include:
     - compiler: clang
       os: osx
@@ -39,6 +40,11 @@ matrix:
     - compiler: gcc
       os: linux
       env: ENABLE_DOC=--enable-doc
+    - os: linux
+      env: HOST=x86_64-w64-mingw32
+    - os: linux
+      env: HOST=i686-w64-mingw32
+  allow_failures:
     - os: linux
       env: HOST=x86_64-w64-mingw32
     - os: linux


frankmorgner:
Putting our heads into the sand doesn't bring us any further, the actual error in MinGW-w64 builds needs to be investigated.

Also, `fast_finish` should be `false`

frankmorgner:
Please note that current build status on the main home page is just pathetic. Since the coverity scan limit is constantly exceeded, CI doesn't build any of our code in `master` and the build icon only shows whether we can successfully run `./configure`. See https://github.com/OpenSC/OpenSC/issues/791#issuecomment-258037760

viktorTarasov:
@frankmorgner 
As far as I see, the actual error do not related to OpenSC itself -- it's the problem of build system.

We'll wait a little bit to see if it'll be resolved.

viktorTarasov:
@frankmorgner 
> ... just pathetic ... 

There are at least two reasons:
- _"We're continuing to experience severe build delays on our OS X infrastructure."_ (Travis-CI broadcast) . That's why this PR;
- It seems that there are multiple triggers of coverit-scan per build, for the same source version. It exhausts very rapidly the allowed limit. 

By the way,
could you be more exhaustive with your assertion _"fast_finish should be false"_ ?

frankmorgner:
Opinions from me and other here: https://github.com/OpenSC/OpenSC/issues/791

viktorTarasov:
@frankmorgner 
> See opinions here: #791

Sometimes so enigmatic...
What have I see there? Something more then the same observation?
Or the details about your _"fast_finish should be false"_ assertion?

frankmorgner:
Quoting from #791:

> I'd prefer to see the output of compilation from as many compilers as possible

There are many reasons why a build may fail. In the recent past we have seen problems with download mirrors, the build infrastructure and, quite recently, seemingly arbitrary failures for no reason. I don't want any of these reasons to stop other build jobs.

Also, the warnings (and potentially errors) on Windows/macOS/Linux differ. I want to get all of them, even if a single build fails.

And most important, the review cycles for a PR span from at least one week (simple merge/review done by contributor) to more than several months (more complex merge that you want to have your eyes on). There is no reason to restrict ourself to quick builds.

viktorTarasov:
You still will be able to see the output of compilation for as many compilers as possible,
the only difference is that the failed builds for the, sorry, _second-order-configurations_, which fails for the reasons that are external to the project, do not impact the general status of the build,

viktorTarasov:
The problems with the cross-builds on the travis platform are still lasting.

PR #913, opened 2016-11-30
Avoid re-opening the log file for every debug message on Windows (fixes #824, superseedes #910)

### LINKED ISSUE #824: Crash attempting to query AIS card

frankmorgner:
#910 fixes #824 by extending the logic for reopening the debug file for every log message. This approach, however, is very problematic in regard to performance.

Instead of building on sandy ground, this PR tries to solve #824 by reopening the debug file once, when needed. This improves overall performance for logging on Windows, simplifies the user configuration options and, of course, fixes the segmentation fault in question.

This PR is compile tested only and needs to be verified on Windows with a IASECC card.

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -727,9 +727,9 @@ app onepin-opensc-pkcs11 {
 # Used by OpenSC.tokend on Mac OS X only.
 app tokend {
 	# The file to which debug log will be written
-	# Default: /tmp/opensc-tokend.log
+	# Default: stderr
 	#
-	# debug_file = /Library/Logs/OpenSC.tokend.log
+	# debug_file = /tmp/opensc-tokend.log
 
 	framework tokend {
 		# Score for OpenSC.tokend
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -21,15 +21,6 @@ app default {
 	#
 	# debug_file = @DEBUG_FILE@
 
-	# Re-open debug file  (used in WIN32)
-	#
-	# In Windows, file handles can not be shared between DLL-s,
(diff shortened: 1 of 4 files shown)


viktorTarasov:
@frankmorgner 
- _"problematic in regard to performance"_
  Do you mean the performance of **smart card** application in **debug mode** ?
  I do not think that it's a valid argument.

- _"This PR is compile tested only and needs to be verified on Windows with a IASECC card"_
  I will not even try to test it.
  It will segfault in the _main_ module at the first log message after the loaded SM DLL used.

- This PR contains the commits that are weakly linked one with other.

Question  for @frankmorgner:
How can you put in the same message both these statements:
 > ... of course, fixes the segmentation fault in question...
 > ... This PR is compile tested only and needs to be verified ...

frankmorgner:
This PR was meant to show a different kind of direction to illustrate my comments on #910. Since I don't have any way of testing, I'm closing this PR.

PR #917, opened 2016-12-14
Fix segfault in pkcs11-tool

pnaaijkens:
Due to a missing comma, pkcs11-tool segfaulted when displaying the help text, because the number of options and descriptions did not match.


frankmorgner:
thanks!

PR #918, opened 2016-12-15
opensc-explorer: show tag 0x82 for unknown files

popovec:
For reserved/unknown file type print value of 0x82 tag, instead of "???" string.

### DIFF
--- a/src/libopensc/iso7816.c
+++ b/src/libopensc/iso7816.c
@@ -391,6 +391,8 @@ iso7816_process_fci(struct sc_card *card, struct sc_file *file,
 			}
 			sc_log(ctx, "  type: %s", type);
 			sc_log(ctx, "  EF structure: %d", byte & 0x07);
+			sc_log(ctx, "  tag 0x82: 0x%02x", byte);
+			sc_file_set_type_attr(file, &byte, 1);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/libopensc/iso7816.c
```diff
@@ -391,6 +391,8 @@ iso7816_process_fci(struct sc_card *card, struct sc_file *file,
 			}
 			sc_log(ctx, "  type: %s", type);
 			sc_log(ctx, "  EF structure: %d", byte & 0x07);
+			sc_log(ctx, "  tag 0x82: 0x%02x", byte);
+			sc_file_set_type_attr(file, &byte, 1);
```
frankmorgner: tag 82 may be followed by a variable size of information between one and 6 bytes. You should instead use
```
sc_file_set_type_attr(file, tag, taglen);
```

src/tools/opensc-explorer.c
```diff
@@ -845,6 +851,9 @@ static int do_info(int argc, char **argv)
 	} else
 		return usage(do_info);
 
+	if(!file->type_attr_len)
+		st = "Unknown File";
```
frankmorgner: Should better be `Unknown File Type`...

frankmorgner:
@viktorTarasov you missed the changes requested [above](https://github.com/OpenSC/OpenSC/pull/918#pullrequestreview-13409155). The commit that got merges misses some bytes of the file attributes in the iso driver.

PR #926, opened 2016-12-27
Added card capabilities check to correctly determine which algorithms…

hhonkanen:
… and key sizes are supported.

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -223,7 +223,7 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
 
 	if (GET_DNIE_UI_CTX(card).user_consent_enabled == 0
-			|| card->ctx & SC_CTX_FLAG_ENABLE_POPUPS) {
+			|| card->ctx->flags & SC_CTX_FLAG_DISABLE_POPUPS) {
 		sc_log(card->ctx,
 		       "User Consent or popups are disabled in configuration file");
 		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -46,8 +46,16 @@
 #define MYEID_STATE_CREATION	0x01
 #define MYEID_STATE_ACTIVATED	0x07
 
-#define MYEID_INFINEON_CHIP_ATR		0x04
 #define MYEID_CARD_NAME_MAX_LEN		100
+
+/* The following flags define the features supported by the card currently in use.
+  They are used in 'card_supported_features' field in myeid_card_caps struct */
+#define MYEID_CARD_CAP_RSA		0x01
+#define MYEID_CARD_CAP_3DES		0x02
+#define MYEID_CARD_CAP_AES		0x04
+#define MYEID_CARD_CAP_ECC		0x08
+#define MYEID_CARD_CAP_GRIDPIN		0x10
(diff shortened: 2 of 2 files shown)


dengert:
The  MYEID_CARD_CAP_PIV_EMU	flag looks interesting. Looking at their web site, I found:

https://webservices.aventra.fi/wordpress/wp-content/downloads/MyEID_PKI_JavaCard_Applet_Reference_Manual_2-1-4.pdf

Very informative document with details on all the MYEID commands.  

On page 6: "PIV / CIV emulation"
talks about mapping PKCS#15 to PIV objects and setting the PIV AID.  But it does not have any detail on thw mapping or what subset of PIV does is support.

Do you have any more information or know where it can be found?

hhonkanen:
PIV emulation is a new feature since MyEID 4.0 and not yet included in the reference guide. When ordering MyEID 4 cards one should mention that the PIV feature is needed, to ensure that this feature is activated before shipping. PIV emulation is read only. You have to personalize the card as a MyEID card, using OpenSC or some other tools.  Then you send PIV INITIALIZE command and define the mapping to files in PKCS#15 structure in its data. After that the card appears as a PIV card. 

We have successfully tested Windows smart card logon with PIV emu. Here is an example of the PIV initialization command. Copy it to an editor that has fixed character width.

PIV init command (activate and deactivate)

#Init PIV functionality
#         P1
#         | Initialise PIV emulation
#         | |   State + flags (0x80 = PIV functionality activated, 0x7F = RFU)
#         | |   | ACL
#         | |   | |     PIV auth key FID
#         | |   | |     |   PIV auth cert FID
#         | |   | |     |   |   Card auth key FID
#         | |   | |     |   |   |   Card auth cert FID
#         | |   | |     |   |   |   |   Signature key FID
#         | |   | |     |   |   |   |   |   Signature cert FID
#         | |   | |     |   |   |   |   |   |   Management key FID
#         | |   | |     |   |   |   |   |   |   |   Management cert FID
#         | |   | |     |   |   |   |   |   |   |   |
/send 00DA01501480111FFF4B014304000000000000000000000000 9000

Better documentation is coming at some point.

hhonkanen:
Travis CI build failed at card-dnie.c, but this PR contains changes only to card-myeid.c. Is there something I can do to resolve this, and does the failed build prevent from merging?

dengert:
@rickyepoderi   can you address the card-dnie.c failure In Travi-Ci failure, this look like a card-dnie.c problem, and this PR #926 is not the cause.

rickyepoderi:
This was introduced in the commit: https://github.com/OpenSC/OpenSC/commit/e7f118d797efc56fc8caa9392951e8332fbb2208

I suppose the change was done blindly because you have no dnie. It should be like this:

    diff --git a/src/libopensc/card-dnie.c b/src/libopensc/card-dnie.c
    index dce15977..78c48072 100644
    --- a/src/libopensc/card-dnie.c
    +++ b/src/libopensc/card-dnie.c
    @@ -223,7 +223,7 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
     
            if (GET_DNIE_UI_CTX(card).user_consent_enabled == 0
    -                       || card->ctx & SC_CTX_FLAG_ENABLE_POPUPS) {
    +                       || card->ctx->flags & SC_CTX_FLAG_DISABLE_POPUPS) {
                    sc_log(card->ctx,
                           "User Consent or popups are disabled in configuration file");

At least, this way it compiles and then it works.

Do you want me to submit a pull request?

dengert:
@hhonkanen 
I am going to continue the discussion of MyEID and PIV on opensc-devel as it in not part of this PR.

hhonkanen:
Thanks rickyepoderi.  I edited card-dnie.c according to your comment and committed it, so you don't need to submit another pull request.  The build goes ok now.

Issue #934, opened 2017-01-03
pkcs15-init --create-pkcs15 --no-so-pin -> segfault (EXC_BAD_ACCESS)


minfrin:
### Expected behaviour

What should happen?

Following command runs successfully, or reports a sensible error message:

pkcs15-init --create-pkcs15 --no-so-pin

### Actual behaviour

What happens instead?

Little-Net:OpenSC minfrin$ lldb -- pkcs15-init --create-pkcs15 --no-so-pin
(lldb) target create "pkcs15-init"
Current executable set to 'pkcs15-init' (x86_64).
(lldb) settings set -- target.run-args  "--create-pkcs15" "--no-so-pin"
(lldb) run
Process 62920 launched: '/usr/local/bin/pkcs15-init' (x86_64)
Using reader with a card: ACS ACR 38U-CCID
Process 62920 stopped
* thread #1: tid = 0x2393a, 0x0000000100001e34 pkcs15-init`do_init_app + 980, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000100001e34 pkcs15-init`do_init_app + 980
pkcs15-init`do_init_app:
->  0x100001e34 <+980>: movq   (%rax), %rdi
    0x100001e37 <+983>: callq  0x100128a6e               ; symbol stub for: sc_lock
    0x100001e3c <+988>: movl   %eax, -0x30c(%rbp)
    0x100001e42 <+994>: cmpl   $0x0, -0x30c(%rbp)

### Steps to reproduce

1. Build latest git master on MacOSX
2. Run "pkcs15-init --create-pkcs15 --no-so-pin"
3. Crash

### Logs

Little-Net:OpenSC minfrin$ opensc-tool -n
Using reader with a card: ACS ACR 38U-CCID
entersafe

minfrin:
Dropping the --no-so-pin option also causes a crash:

Little-Net:OpenSC minfrin$ lldb -- pkcs15-init --create-pkcs15
(lldb) target create "pkcs15-init"
Current executable set to 'pkcs15-init' (x86_64).
(lldb) settings set -- target.run-args  "--create-pkcs15"
(lldb) run
Process 62947 launched: '/usr/local/bin/pkcs15-init' (x86_64)
Using reader with a card: ACS ACR 38U-CCID
New User PIN.
Please enter User PIN: 
Please type again to verify: 
[...]

PR #935, opened 2017-01-03
Fix segfault: p15card is NULL, we cannot lock.

minfrin:
Fix for https://github.com/OpenSC/OpenSC/issues/934.


frankmorgner:
This change is wrong and incomplete, please use https://github.com/OpenSC/OpenSC/pull/937 instead

Issue #943, opened 2017-01-16
OpenSC fails to detect card when inserted after the library initialization (C_WaitForSlotEvent() does not return)


Jakuje:
### Expected behaviour

Initializing pkcs11 library and then waiting for slot events (card insertion) should work.

### Actual behaviour

Trying interactive tests depending on the `C_WaitForSlotEvent()` fails to detect the Coolkey card (or CAC -- from #841), because the card is detected in the `card-mcrd` driver and its initialization fails.

### Steps to reproduce

0. Remove the card from the reader
1. Start the interactive testsuite (init library and call C_WaitForSlotEvent), for example using `pkcs11-tool --test-hotplug` or [`./p11test -p pin -i -s 0`](https://github.com/Jakuje/OpenSC/tree/jjelen-testsuite/src/tests)
2. Insert the card
3. Watch nothing happen.

### Logs

With `OPENSC_DEBUG=999` we can observe these logs behavior (with comments):

start detection in `mcrd`:
```
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:266:sc_connect_card: trying driver 'mcrd'
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:993:match_atr_table: ATR     : 3b:ff:14:00:ff:81:31:fe:45:80:25:a0:00:00:00:56:57:53:43:36:35:30:03:03:38
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:1004:match_atr_table: ATR try : 3B:FF:94:00:FF:80:B1:FE:45:1F:03:00:68:D2:76:00:00:28:FF:05:1E:31:80:00:90:00:23
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:1007:match_atr_table: ignored - wrong length
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:1004:match_atr_table: ATR try : 3b:6f:00:ff:00:68:d2:76:00:00:28:ff:05:1e:31:80:00:90:00
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:1007:match_atr_table: ignored - wrong length
0x7f336f056980 13:03:02.486 [opensc-pkcs11] card.c:1004:match_atr_table: ATR try : 3b:ff:11:00:ff:80:b1:fe:45:1f:03:00:68:d2:76:00:00:28:ff:05:1e:31:80:00:90:00:a6
[...]

PR #945, opened 2017-01-23
Fixes for issues found during investigation of #943

Jakuje:
The attached changes make the detection working well, though it might not resolve the original problem (probably underlying in pcscd).


frankmorgner:
thanks!

PR #947, opened 2017-01-26
Initial CardOS 5.3 support

Jakuje:
CardOS 5.3 cards have different ATR so they are not picked up by OpenSC at this moment. Unless there will be more differences in the future, using some ATR mask would probably make sense.

These cards also present ODF with non-strict encoded ASN1:
```
6F 81 25 83 02 50 31 82 01 01 80 02 00 3C 88 00 o.%..P1......<..
85 01 00 8A 01 05 AB 10 80 01 01 90 00 80 01 06 ................
A4 06 83 01 11 95 01 80 90 00                   ..........
```
Note the length part `81 25` combining into the length of 37, which is not handled by current code (expecting the length of the "length" to be only single byte).

Further changes to incorporate the support of the CardOS 5.3 might be needed, but so far submitting what we have so far.


Jakuje:
Thanks. It looks like we are not completely done here. The cards do not respond correctly to our sign/decode commands correctly. Hopefully next week I will be able to uncover this other part to make it working, which will need 5.3-specific changes.

PR #949, opened 2017-01-26
pteid: order objects by address (code cosmetics only)

nunojpg:

### DIFF
--- a/src/libopensc/pkcs15-pteid.c
+++ b/src/libopensc/pkcs15-pteid.c
@@ -290,19 +290,19 @@ static int sc_pkcs15emu_pteid_init(sc_pkcs15_card_t * p15card)
 	for (i = 0; i < 5; i++) {
 		static const char *object_ids[5] = {"1", "2", "3", "4", "5"};
 		static const char *object_labels[5] = {
+			"TRACE",
 			"Citizen Data",
 			"Citizen Address Data",
-			"Citizen Notepad",
 			"SOD",
-			"TRACE",
+			"Citizen Notepad",
 		};
-		static const char *object_authids[5] = {NULL, "3", NULL, NULL, NULL};
+		static const char *object_authids[5] = {NULL, NULL, "3", NULL, NULL};
 		static const char *object_paths[5] = {
+			"3f000003",
 			"3f005f00ef02",
 			"3f005f00ef05",
-			"3f005f00ef07",
 			"3f005f00ef06",
-			"3F000003",
+			"3f005f00ef07",
 		};
 		static const int object_flags[5] = {
 			0,


frankmorgner:
I can see that this doesn't change any of the existing logic. But I don't understand why *your* way of formatting the output should be more consistent than the existing way of doing it.

frankmorgner:
Just realized one change: Your commit changes the object IDs, which would break existing applications relying on the original order... However, since I don't know this card, I can't tell which impact this has in the end.

frankmorgner:
The object flags are changed too. Please review your changes!

nunojpg:
@frankmorgner  You are right and thanks for spoting the flags issue. It is now fixed.

I'm ordering objects by address because:

* there is no other reasonable order, as otherwise it is random.
* the object ID will change but current object IDs have also been changed since last stable release and there is no reason to assure stability over something published on master a few months ago.

dengert:
If the current order is in any previous release, like 0.16.0, I vote for "stability over something published on master a few months ago" is a reason not to change the order.

If I read the change correctly, the user currently see when listing objects: 
      1 "Citizen Data",
      2 "Citizen Address Data",
      3 "Citizen Notepad",
      4 "SOD",
      5 "TRACE",

but with the change they will see:
      1 "TRACE",
      2 "Citizen Data",
      3 "Citizen Address Data",
      4  "SOD",
      5 "Citizen Notepad",

What is SOD, and why when would a user want to use TRACE?    

I would agree with @frankmorgner that making changes like this could  break existing applications relying on the original order.  You are calling it  "cosmetics"  but the use's or applications never see the addresses or file paths but may use the object IDs to refer to the object  (SLOT_1-ID_1)  and the  "cosmetics" aspect of this change is only for the code developer who looks at the things like this. But like @frankmorgner I don't use the card, or have anyway to test it. 

Is it "random"? I would say the original order is more users then the developer.

nunojpg:
SOD is the Document Security Object. The same as in a MRTD.

It is a pkcs7 object, and contains hashes of the other data on the card, a certificate, and a signature.

It is used to authenticate the data with my tool, for example:

https://github.com/nunojpg/cartaodecidadao-explorer

TRACE have some version stuff, that I'm not using, but might be eventually relevant...and it doesn't hurt to be available.

This new objects were committed only in November, so they are unreleased in 0.16.

IDs are just arbitrarily assigned and should not be expected to stay the same with new versions if they are not defined by the card specification. I mean, anyone reading "Address" by ID=2, is asking for trouble.

My reason to change was that during the card dump I was seeing ID sequence and address sequence misaligned, and I would always assume that to mean something, as if the ID had any reason to be on that specific order.

So I would by default prefer to keep IDs aligned to memory address. To make clear my wording was wrong, this was not code cosmetics but ID realignment. My new pull request removed that wording.

dengert:
OK. I was mistaking looking at:
```
static const char *object_ids[5] = {"1", "2", "3", "4", "5"};
```
and assuming these to be the source for a PKCS\#11 CKA_ID. 

PKCS\#15 and PKCS\#11 do not define IDs for data objects. In PKCS\#11, CKA_ID is only for keys and certificates so they can be matched. And in pkcs11-framework.c pkcs15_dobj_get_attribute, has line 4139  `#if  0` around code that could return a CKA_ID for an object. 

But PKCS\#11 defines  CKA_OBJECT_ID  and PKCS\#15 defines "applicationOID OBJECT IDENTIFIER OPTIONAL". So if the specs define an OID for this type of object then that OID should be defined. 

The OpenSC tools can search on the OIDs. 
```
static const char *object_ids[5] = {"1", "2", "3", "4", "5"};
```
is wrong, as these are not real OIDs. If the specs don't define the OIDs, then 
```
sc_pkcs15_format_id(object_ids[i], &obj_info.id);
```
should not be called for that object. 

In that case you can put them in any order you want.

nunojpg:
@dengert Feedback incorporated!

Any discussion should now continue on #954.

PR #951, opened 2017-01-26
react better to zero-length pin entering

nunojpg:
Considering a PIN which have a min length of 4 (eg), in a card with iso7816 support, the following happens:

| Entered PIN length      | pkcs11-tool                                             | pkcs15-tool                                                                   |
|-------------------------|---------------------------------------------------------|-------------------------------------------------------------------------------|
| 0 (just return pressed) | dies with "No PIN entered" (after this pull request)    | attempts a zero-length pin login at the card and corresponding Verification failure. |
| 1-3                     | dies with "CKR_PIN_LEN_RANGE" (after this pull request) | prompts with "PIN code too short, try again."                                 |
| 4+                      | OK                                                      | OK    

So in the pkcs15-tool with zero-length case, the behavior is not good.

Unfortunately fixing this requires a little thinking. Starting by, is the pin length max and min to be respected? Shall opensc refuse to try a authentication with a pin length that is not within the expected valid range to avoid at (almost) all cost verification failures and eventual PIN 0-tries left lock?

Can this verification be taken down until iso7816.c (if supported) or pkcs15-pin.c?

Depending on this a common strategy for all tools might be better defined.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1486,8 +1486,8 @@ pkcs15_login(struct sc_pkcs11_slot *slot, CK_USER_TYPE userType,
 		if (ulPinLen == 0)
 			pPin = NULL;
 	}
-	else if (ulPinLen > pin_info->attrs.pin.max_length)   {
-		return CKR_ARGUMENTS_BAD;
+	else if (ulPinLen > pin_info->attrs.pin.max_length || ulPinLen < pin_info->attrs.pin.min_length)   {
+		return CKR_PIN_LEN_RANGE; //The specified PIN is too long or too short.
 	}
 
 
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -1094,6 +1094,10 @@ static int do_verify(int argc, char **argv)
 			printf("Please enter PIN: ");
 			r = util_getpass(&pin, &len, stdin);
 			if (r < 0) {
+				printf("util_getpass error.\n");
+				return -1;
+			}
+			if (r < 1) {
 				printf("No PIN entered - aborting VERIFY.\n");
 				return -1;
 			}
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -1245,6 +1245,8 @@ static int login(CK_SESSION_HANDLE session, int login_type)
 
 		r = util_getpass(&pin, &len, stdin);
 		if (r < 0)
+			util_fatal("util_getpass error");
+		if (r < 1)
 			util_fatal("No PIN entered");
 		pin_allocated = 1;
 	}


dengert:
"...is the pin length max and min to be respected?  "I vote yes.  

pkcs15-tool.c list_pins() should also list the `auth_info->attrs.pin.min_length`  and `auth_info->attrs.pin.max_length`. 

Note: pkcs15-pin.c sc_pkcs15_verify_pin says:
```
285  * If the code given to us has zero length, this means we
286  * should ask the card reader to obtain the PIN from the
287  * reader's PIN pad
```
But _sc_pkcs15_verify_pin does not check pin provided against the min and max.
Maybe it should? That would address all the utilities, and pkcs\#11 too.

nunojpg:
sc_pkcs15_verify_pin is clearly not doing it.

If we pass pincode=NULL, and pinlen=0, then it will use a pinpad if present, or otherwise try a zero lenght pin.

I would say than when pincode==NULL, then if pinpad is not present it should abort immediatly.
Only if pincode is a empty string (pincode!=NULL && pinlen==0), a zero-length pin should be attempted, if the min_length allows it.

frankmorgner:
The PIN length is checked on the PKCS#15 layer in `src/pkcs11/framework-pkcs15.c:_validate_pin()`. We should leave that here. You'll get `SC_ERROR_INVALID_PIN_LENGTH` if it's not within the range. `pkcs11-tool` should listen to this error and report `CKR_PIN_LEN_RANGE` instead of implementing its own length checking. So I would not accept this change in its current form.

Comments should be updated if they are not accurate anymore.

nunojpg:
@frankmorgner could you please clarify what you mean?

From my perspective pkcs11-tool is a pkcs11-only aware client that should, ideally, even work with other non-opensc pkcs11 library implementations.

For a pkcs11 client "SC_ERROR_INVALID_PIN_LENGTH", means nothing. 

I would agree that pkcs11 emulation should listen to SC_ERROR_INVALID_PIN_LENGTH and report CKR_PIN_LEN_RANGE, without any specific checks, if this is what you mean.

This means that first the problem must be fixed at SC_ERROR_INVALID_PIN_LENGTH, as it is not working properly now.

I didn't dare to touch that code yet because I was waiting for comments, but I sure agree to move that logic down as advised by @dengert.

frankmorgner:
Sorry, I meant that the PKCS#11 layer (i.e. `src/pkcs11/framework-pkcs15.c`) should rely on the pin length checking of the PKCS#15 layer (i.e. `_validate_pin()`). Indeed, `pkcs11-tool` should not do this.

Actually, `sc_to_cryptoki_error_common()` is already translating the return codes correctly. You should check where this information gets lost or why this translation is not used...

nunojpg:
@frankmorgner Yes, sc_to_cryptoki_error_common() had it implemented and the issue was exclusively on the pkcs15 layer. _validate_pin() was not being called at all.

So a general fix becomes trivial. Thanks for the comments.

I'm merging that fix at my pull request #948 and closing this.

PR #954, opened 2017-01-28
pteid: order objects by address (supersede #949)

nunojpg:


frankmorgner:
What do you mean by ordering the objects "by address"? I still don't get why you want this change so hard if you admitted in #949 that any order would be fine.

nunojpg:
I would't say I am trying that hard, I just want it committed if it is "the right thing to do". According to comments that the ID should be removed, and that I also noted some objects had wrong casing, I've updated it again.

And now it is a real breaking change, for example for some of my code that was already depending on the name "SOD".

PR #957, opened 2017-01-31
pkcs15-init,pkcs15-tool: reword --no-prompt to --use-pinpad (close #944)

### LINKED ISSUE #944: pkcs15-tool "--no-prompt" option rewording to "--use-pinpad"
Would like comments to reword the "--no-prompt" to "--use-pinpad".

--no-prompt looks a terrible wording for this, and it hides the fact that a pinpad is not used by default.

Will provide a pull request according to feedback.

nunojpg:
Wording was confusing for a novice user. Old option is mantained as an alias,
but will print to stderr a deprecation warning.

Deprecation related code is all marked with deprecated word to easy future removal.

### DIFF
--- a/src/tools/pkcs15-init.c
+++ b/src/tools/pkcs15-init.c
@@ -144,13 +144,14 @@ enum {
 	OPT_MD_CONTAINER_GUID,
 	OPT_VERSION,
 
-	OPT_PIN1     = 0x10000,	/* don't touch these values */
-	OPT_PUK1     = 0x10001,
-	OPT_PIN2     = 0x10002,
-	OPT_PUK2     = 0x10003,
-	OPT_SERIAL   = 0x10004,
-	OPT_NO_SOPIN = 0x10005,
-	OPT_NO_PROMPT= 0x10006
+	OPT_PIN1      = 0x10000,	/* don't touch these values */
+	OPT_PUK1      = 0x10001,
+	OPT_PIN2      = 0x10002,
+	OPT_PUK2      = 0x10003,
+	OPT_SERIAL    = 0x10004,
+	OPT_NO_SOPIN  = 0x10005,
+	OPT_USE_PINPAD= 0x10006,
+	OPT_USE_PINPAD_DEPRECATED
 };
 
 const struct option	options[] = {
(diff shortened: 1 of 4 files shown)


src/tools/pkcs15-init.c
```diff
@@ -201,7 +202,8 @@ const struct option	options[] = {
 	{ "insecure",		no_argument, NULL,		OPT_INSECURE },
 	{ "use-default-transport-keys",
 				no_argument, NULL,		'T' },
-	{ "no-prompt",		no_argument, NULL,		OPT_NO_PROMPT },
+	{ "use-pinpad",		no_argument, NULL,		OPT_USE_PINPAD },
+	{ "no-prompt",		no_argument, NULL,		OPT_USE_PINPAD_DEPRECATED },
```
frankmorgner: Please make it a silent option or simply remove the old option.

nunojpg:
AFAIK this option was not documented at all under doc/tools (I checked it.)

src/tools/pkcs15-init.c
nunojpg: Yes, also thought it was the better. Just didn't find any example of a silent option, in this code. Can you point me how to do it?

nunojpg: owh, I found, if the help is string (const char*) NULL it becomes silent. Will fix it. Thanks.

nunojpg:
@frankmorgner updated as advised.

src/tools/pkcs15-init.c
```diff
@@ -2627,8 +2630,10 @@ handle_option(const struct option *opt)
 	case OPT_NO_SOPIN:
 		opt_no_sopin = 1;
 		break;
-	case OPT_NO_PROMPT:
-		opt_no_prompt = 1;
+	case OPT_USE_PINPAD_DEPRECATED:
+		fprintf(stderr, "--no-prompt have been deprecated and will be removed. Use instead '--use-pinpad'.\n");
```
frankmorgner: Please change the messages to "'--no-prompt' is deprecated , use '--use-pinpad' instead.\n

nunojpg:
@frankmorgner updated.

frankmorgner:
good work, thanks!

PR #958, opened 2017-02-03
card-rtecp: Fix rtecp_change_reference_data (resolves #931)

### LINKED ISSUE #931: Can't store key on Rutoken ECP 2.0
Can't store SSH key on Rutoken ECP.
ART: 3b:8b:01:52:75:74:6f:6b:65:6e:20:44:53:20:c1

### Steps to reproduce

Opensc-0.16.0

```
rm ssh.key
echo | ssh-keygen -b 2048 -t RSA -f ssh.key
pkcs15-init -E
pkcs15-init -C --so-pin '12345678' --so-puk ''
pkcs15-init -P --id 2 -l "User" --so-pin '12345678'  --pin '12345678' --puk ''
pkcs15-init -a 2 -S ssh.key -l ssh.key --pin '12345678'
```
Error: "Failed to store private key: Invalid arguments"
Subsequent repeating of the last operation produces error: "Failed to store private key: File already extsts"

### Note

I do not want to generate key on the token. Key should be validated and backuped.

konstantinpersidskiy:
Hello

This branch resolves the issue #931 . I inspected some code and probably found out the cause of  the trouble. The problem seemed to appear in OpenSC of version 0.14.0 and higher and it was caused by rtecp-dependent code relying on previous implementation of function asn1_put_tag. So when asn1_put_tag was changed to treat long data more accurate, some of rtecp-dependent code became invalid in some cases. I rewrote that rtecp code and now it seems to work properly when it comes to importing long pieces of data (like RSA keys) to rtecp (tested on Rutoken ECP 2.0).

Regards, 
Konstantin Persidskiy

### DIFF
--- a/src/libopensc/card-rtecp.c
+++ b/src/libopensc/card-rtecp.c
@@ -446,7 +446,10 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
 	
 	max_transmit_length = sc_get_max_send_size(card);
 	assert(max_transmit_length > 2);
-	/* (2 + sizeof(rsf_length) + newlen) - total length of data we need to transfer, (max_transmit_length - 2) - amount of useful data we can transfer in one transmit (2 bytes for 0xA5 tag) */
+	/*
+	 * (2 + sizeof(rsf_length) + newlen) - total length of data we need to transfer,
+	 * (max_transmit_length - 2) - amount of useful data we can transfer in one transmit (2 bytes for 0xA5 tag)
+	 */
 	transmits_num = (2 + sizeof(rsf_length) + newlen) / (max_transmit_length - 2) + 1;
 	/* buffer length = size of 0x80 TLV + size of RSF-file + (size of Tag and Length)*(number of APDUs) */
 	buf_length = (2 + sizeof(rsf_length)) + newlen + 2*(transmits_num); 
--- a/src/libopensc/card-rtecp.c
+++ b/src/libopensc/card-rtecp.c
@@ -430,52 +430,69 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
 		const u8 *newref, size_t newlen, int *tries_left)
 {
 	sc_apdu_t apdu;
-	u8 tmp[2], buf[0x1000], *p = buf;
-	size_t taglen;
-	int r;
+	u8 rsf_length[2], *buf, *buf_end, *p; 
+	/* update used_buf_length after putting TLV */
+	size_t val_length, buf_length, max_transmit_length;
+	int apdu_num, r;
 
 	assert(card && card->ctx && newref);
 	sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "newlen = %u\n", newlen);
-	if (newlen > sizeof(buf) - 2 - sizeof(tmp) - 2 * (sizeof(buf) / 0xFF + 1))
+	if (newlen > 0xFFFF)
 		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_INVALID_ARGUMENTS);
-
 	if (type == SC_AC_CHV && old && oldlen != 0)
 	{
 		r = sc_verify(card, type, ref_qualifier, old, oldlen, tries_left);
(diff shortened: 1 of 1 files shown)


dengert:
The comment in line 450 to 453 is hard to read, and not clear if this is code or not. Can you use:
```
/*
 * comment
 * comment
 */
```
You comment that the card in one case does not support sc_asn1_put_tag properly, but you also use sc_asn1_put_tag in another place. The card may be expecting SIMPLE-TLV not BER-TLV  a form of ASN.1. 
In SIMPLE-TLV  the length is either 1 or 3 bytes, the single length byte can be between 0-254. 
ISO 7816-4  "5.4.4 Data field bytes"

https://github.com/OpenSC/OpenSC/commit/abf8f36 has a version of simpletlv.c and simpletlv.c but its is not in master. 

Does the card really support chaining  (SC_APDU_FLAGS_CHAINING)? 

If it did, I would not expect each each segment to have to start with the A5 Len. 
The point in chaining is is to send large blocks in smaller segments, without having to add
extra fields for each segment.

What you might be seeing is the OpenSC code sees the chaining bit set, and sends multiple APDUs
to the card each with the A5 Len as you reconstructed the buffer to have the A5 len in the right places. But the card is ignoring the chaining bit. So in effect you are using the OpenSC chaining code to loop for you.

konstantinpersidskiy:
Comment  at 450-453 should really be a bit more neat - thanks for noticing

TLV-format used in rtecp is not SIMPLE-TLV and not BER-TLV - that's why we have to put A5 tags in a special loop, not with any of implemented in OpenSC functions. But when TLV length is less than 128, the output of asn1_put_tag() matches the expected format, and for putting TLV with small length asn1_put_tag() is used - to fit into general OpenSC codestyle

Rtecp supports chaining, but it's a bit 'special' either. The card doesn't ignore the chaining bit - it is needed to ensure the consistency of chaining APDU, but card also expects each transmit not to cut TLVs in parts, that's why it may look a little confusing.

src/libopensc/card-rtecp.c
```diff
@@ -430,52 +430,69 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
...
-	if (r == SC_ERROR_INVALID_ARGUMENTS)
+	
+	max_transmit_length = sc_get_max_send_size(card);
+	assert(max_transmit_length > 2);
+	/*
+	 * from inequality:
+	 * buf_length <= apdu_num * max_transmit_length;
+	 * 2 + sizeof(rsf_length) + newlen + 2 * apdu_num <= apdu_num * max_transmit_length;
+	 * apdu_num >= (2 + sizeof(rsf_length) + newlen) / (max_transmit_length -2) 
+	 */
```
frankmorgner: Please clearify the comments. If nobody knows what you're talking about you might as well remove the comments...

src/libopensc/card-rtecp.c
```diff
@@ -430,52 +430,69 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
...
-			newlen -= taglen;
-			if (newlen)
-				apdu.flags |= SC_APDU_FLAGS_CHAINING;
-		}
+		assert(buf_end - p >= (int)(newlen + 2));
+		if ((p - buf) % max_transmit_length + newlen + 2 > max_transmit_length)
+			val_length = max_transmit_length - (p - buf) % max_transmit_length - 2;
+		else
+			val_length = newlen;
+		assert(val_length <= newlen);
```
frankmorgner: please replace `assert` with proper error handling.

konstantinpersidskiy:
@frankmorgner , thanks for reviewing the code!

The assert you mentioned won't be triggered in any situation in this version of file - that is guaranteed by the if-else block (lines 471-474), but it'll help future developers to control violating of input data borders (if he change code between if-else and assert). Also asserts are used evwerywhere in this particular file, so I decided to follow the style and use them either.

I also changed the comment to be more informative and clear.

> If you change the comments, I'd merge this commit.
I'm looking forward to it!

frankmorgner:
Indeed, we could do much better with the error handling, but that shouldn't keep us from starting somewhere.

PR #961, opened 2017-02-07
set dummy serial number

hamano:
tokeninfo->serial_number = NULL; doesn't work with OSX tokend.
see https://github.com/OpenSC/OpenSC.tokend/blob/96b89d1bedfd00d095fdac7e3a9628127e227a78/OpenSC/OpenSCToken.cpp#L304
Also PKCS#11 spec doesn't allow this.
http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html

> serialNumber       character-string serial number of the device.  MUST be padded with the blank character (‘ ).  MUST NOT be null-terminated.

Thank you.

### DIFF
--- a/src/libopensc/pkcs15-jpki.c
+++ b/src/libopensc/pkcs15-jpki.c
@@ -47,8 +47,8 @@ sc_pkcs15emu_jpki_init(sc_pkcs15_card_t * p15card)
 
 	p15card->tokeninfo->label = strdup("JPKI");
 	p15card->tokeninfo->manufacturer_id = strdup("JPKI");
-	/* set NULL until we found serial number */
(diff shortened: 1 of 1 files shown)


frankmorgner:
didn't you say that this is a contactless card? If the UID is non-random, you could use the UID as serial number...

hamano:
Thanks for your comment.
I was able to get IDm by sending "FF CA 00 00 00", but the JPKI card also has contact interface.
https://www.kojinbango-card.go.jp/en/kojinbango/index.html
It is not able to get IDm when using contact reader.
I'm plannning to use certificate serial, but could you merge this at the moment?

hamano:
duplicate #964

PR #963, opened 2017-02-07
macOS: Use a better Installer and Uninstaller

frankmorgner:
An apple script converted to an App calls the local
/usr/local/bin/opensc-uninstall


frankmorgner:
Maybe someone else could add some nice icons...

metsma:
You can compile the app build time to avoid apple binaries in source.
osacompile -o MyScript.app MyScript.scpt

frankmorgner:
thanks

frankmorgner:
- fixed compiling the app from source via `build-package`
- fixed `make dist`

- added simplified user interactions during installation

PR #969, opened 2017-02-14
Improve hash handling for PtEID and other cards

dengert:
The Pull request has two commits: https://github.com/OpenSC/OpenSC/commit/dbbe26c1866052c993b51fba65c84ddf692209d6
are changes to the base OpenSC that introduce two new unsigned long variables, that are initialized to zero by existing code. This allows for the introduction of these flags to be used by selected drivers.
This appears to work with any card, as the code was designed to not cause problems with other drivers.  

https://github.com/OpenSC/OpenSC/commit/4086aa51e9c26c0d7680b8c98da678e7a468ec25
are change to card-gemsafeV1.c for the PtEID card to use the above flags. 

Over the last few weeks @nunojpg, @l1k and myself have been working on issue of the
PtEid card with handing SHA256 and with handling 2048bit RSA keys. @nunojpg has been testing changes with his PtEid card and @l1k was the last to make changes for this card.

@nunojpg had posted on OpenSSH problems with using SHA512. It turns out the card can not
support SHA512, as the digest header + hash  is 81 bytes, and the hash by itself is 64 bytes. 
The card has a limit of 35 bytes. It also turned OpenSSH was using SHA512 and could use SHA256 which the PtEid card can use if the digest header is striped and only the 32 byte hash sent to the card.
The card would then add the SHA256 digest header then pad and sign.   

So these card-gemsafev1.c part of the pull request still needs more testing, but the Improved hash handling could be useful for other cards too. 

card-gemsafeV1.c is used by other cards, including the SEEID card. So these changes may effect or improve the usefullness of these cards.

### DIFF
--- a/src/libopensc/card-gemsafeV1.c
+++ b/src/libopensc/card-gemsafeV1.c
@@ -84,6 +84,7 @@ static const u8 gemsafe_def_aid[] = {0xA0, 0x00, 0x00, 0x00, 0x63, 0x50,
 typedef struct gemsafe_exdata_st {
 	u8	aid[16];
 	size_t	aid_len;
+	u8 alg_ref;	/* from last set_securuty_env */
 } gemsafe_exdata;
 
 static int get_conf_aid(sc_card_t *card, u8 *aid, size_t *len)
@@ -207,26 +208,17 @@ static int gemsafe_init(struct sc_card *card)
 		flags |= SC_ALGORITHM_RSA_HASH_NONE;
 
 		/* GemSAFE V3 cards support SHA256 */
-		if (card->type == SC_CARD_TYPE_GEMSAFEV1_PTEID ||
-		    card->type == SC_CARD_TYPE_GEMSAFEV1_SEEID)
-			flags |= SC_ALGORITHM_RSA_HASH_SHA256;
+		/* but they do this by accepting just a SHA256 hash without digest */
+
+		/* do not create PKCS#11 mechanisms for these digest+hash. They are too long*/
+		card->negate_hashes = SC_ALGORITHM_RSA_HASH_SHA224 |
+			SC_ALGORITHM_RSA_HASH_SHA384  |
+			SC_ALGORITHM_RSA_HASH_SHA512;
 
 		_sc_card_add_rsa_alg(card,  512, flags, 0);
 		_sc_card_add_rsa_alg(card,  768, flags, 0);
 		_sc_card_add_rsa_alg(card, 1024, flags, 0);
 		_sc_card_add_rsa_alg(card, 2048, flags, 0);
-
-		/* fake algorithm to persuade register_mechanisms()
-		 * to register these hashes */
-		if (card->type == SC_CARD_TYPE_GEMSAFEV1_PTEID ||
-		    card->type == SC_CARD_TYPE_GEMSAFEV1_SEEID) {
-			flags  = SC_ALGORITHM_RSA_HASH_SHA1;
-			flags |= SC_ALGORITHM_RSA_HASH_MD5;
-			flags |= SC_ALGORITHM_RSA_HASH_MD5_SHA1;
-			flags |= SC_ALGORITHM_RSA_HASH_RIPEMD160;
-
-			_sc_card_add_rsa_alg(card,  512, flags, 0);
-		}
 	}
 
 	card->caps |= SC_CARD_CAP_ISO7816_PIN_INFO;
@@ -388,7 +380,7 @@ static u8 gemsafe_flags2algref(struct sc_card *card, const struct sc_security_en
 	u8 ret = 0;
 
 	if (env->operation == SC_SEC_OPERATION_SIGN) {
-		if (env->algorithm_flags & SC_ALGORITHM_RSA_HASH_SHA256)
+		if (env->algorithm_hash & SC_ALGORITHM_RSA_HASH_SHA256)
 			ret = GEMSAFEV3_ALG_REF_SHA256;
 		else if (env->algorithm_flags & SC_ALGORITHM_RSA_PAD_PKCS1)
 			ret = (card->type == SC_CARD_TYPE_GEMSAFEV1_PTEID ||
@@ -428,9 +420,11 @@ static int gemsafe_set_security_env(struct sc_card *card,
 				    const struct sc_security_env *env,
 				    int se_num)
 {
+	gemsafe_exdata *exdata = (gemsafe_exdata *)card->drv_data;
 	u8 alg_ref;
 	struct sc_security_env se_env = *env;
 	struct sc_context *ctx = card->ctx;
+	exdata->alg_ref =  0;
 
 	SC_FUNC_CALLED(ctx, SC_LOG_DEBUG_VERBOSE);
 
@@ -440,6 +434,7 @@ static int gemsafe_set_security_env(struct sc_card *card,
 		if (alg_ref) {
 			se_env.algorithm_ref = alg_ref;
 			se_env.flags |= SC_SEC_ENV_ALG_REF_PRESENT;
+			exdata->alg_ref = alg_ref;
 		}
 	}
 	if (!(se_env.flags & SC_SEC_ENV_ALG_REF_PRESENT))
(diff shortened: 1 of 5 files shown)


nunojpg:
I can confirm this works with my Portuguese Citizen Card. I will further test it with other newer PTEID cards, but I need over 1 month for that.

I really appreciate all @dengert work.

frankmorgner:
I think so far the approach was to whitelist the available algorithms. Wouldn't it be possible to whitelist digest+hash and hash separately?

I guess you're using a black listing approach for easy backwards compatibility. But I think it's less readable and less future proof (think of a new hash function, for example).

dengert:
Yes I chose a blacklist approach for backwards compatability. I could look at a whitelist approach.
If all bits are off, it can handle all the well known hashes. If not it can handle only the hashes listed.

You asked: "digest+hash and hash separately?"  I would say no. The card can handle CKM_RSA_PKCS as long as the data is less the 36 bytes. A SHA1 digest header and hash fits. For SHA256 the card has a feature to allow send a flag saying the data being sent is a 32 byte SHA256 hash (without digest) and tells the card to add the SHA256 digest header before padding.  

The size restriction is independent of where the hash is created, i.e. internal to OpenSC  because CKM_<hash>_RSA_PKCS mechanism are advertised or by the calling PKCS\#11 application using its own hashing methods or the PKCS\#11 hash only mechanism followed by using the CKM_RSA_PKCS mechanism. But if the card can not sign, for example SHA512, why should the OpenSC PKCS#11 advertise hash mechanisms that it can not handle?

frankmorgner:
Currently, card capabilities are made explicit (whitelisted) via `_sc_card_add_rsa_alg` by specifying the algorithm flags accordingly. I'd like to keep this paradigm. That's why asked for the whitelisting (possibly by creating algorithm flags separately for each padding method).

The problem I see with the newly introduced flags in `sc_card_t` and `sc_security_env_t` is that it's yet an other implementation detail a (new) developer has to keep in mind. The same goes for the blacklisting approach, which is even a new paradigm. I don't think this change is good in terms of maintainability.

Also, this change solves a very specific problem for one single card. I don't think its worth changing OpenSC core library given the fact that there's already a card driver specific workaround for this problem.

dengert:
My design philosophy for code changes to OpenSC has always been based on:

1. Keep card specific changes in the card driver if at all possible. 

2. If code to the base is need for a card or cards, design it so it can be used by new or existing cards.

3. Changes to base code should not impact existing drivers without providing ways for existing drivers to continue to function.  

@frankmorgner  _sc_card_add_rsa_alg is not the only place where mechanisms are added. As stated in the commit message: 
> "pkcs11/framework-pkcs15.c and openssl.c are overly aggressive in trying to create PKCS#11 mechanisms for every hash OpenSSL supports even if the card could never sign that hash."

These new mechanisms are added  **after** the card driver has done all its  _sc_card_add_rsa_alg  and the card driver does not regain control until **after** a calling PKCS\#11 application has selected a mechanism and done a C_Sign* operation which in this case will fail. And as stated in the commit message: 

> Thus a calling application will not be mislead into using a mechanism the card cannot
    support.

 I can change the patch so zero means add all the possible hashes as is currently done and non zero flag means only add the hashes listed in the flag. Thus new developer don't have to set or use the flag unless their card has the same type of issues the gemsafeV1 card has.

(The old commits that added the original changes to framework-pkcs15.c and openssl.c in effect violated my design philosophy 3 by adding the extra mechanisms be default. It was not obvious at the time that there are cards that could have issues. My commit to add the flag gives control back to the  card drivers l to only add mechanisms the card and card driver can support. I suspect there are other cards drivers in OpenSC that need this too. Its just that many of these older card drivers are not used anymore as best we can tell.)

As for the other flag, this is introduced to give the card driver additional control because upper levels of OpenSC have made assumptions. The commit message states:

>  sc_pkcs15_compute_signature looks at its flags and at the data to be signed to set
    senv.algorithm_hash The sc_pkcs1_strip_digest_info_prefix is used to check if the
    data looks like a known hash with a digest created externally by the calling
    application or a hash and digest created internally by OpenSC.
    The card drivers set_security_env can then use this information.

When sc_pkcs15_compute_signature is called it has all the information it needs including the data to be signed. Use_key is called which calls  sc_set_security_env, which is the first time the card driver gains control. But for the gensafe_set_security_env it does not have the information it needs.  It does not have access to the data that is to be signed to determine if it can strip off the SHA256 digest header and just send the hash to the card. 

Again this is information that any card driver could use or ignore.  Thus new developer don't have to set or use the flag unless their card has the same type of issues the gemsafeV1 card has.

I have tried to be very careful in the design of this patch to not require any changes to existing drivers but allow existing drivers to take advantage of the code in case they have similar issues. (Design philosophy points 2 and 3)

As for changes...

With the changes to pkcs11/framework-pkcs15.c and openssl.c I don't see any other way around the problem, as the existing code was making an invalid assumption that existing cards could handle any size hash or any new hash that OpenSSL might support. 
 
With the sc_set_security_env the other ways around this are:

- passing the pointer to the data and its size in the senv that could easily be done. Following design philosophy 2, I made the change to look at the hash in the upper levels so other card drivers could take advantage of the change. It would be easy to pass in the pointer and size and let the gemsafe_set_security_env (or any card's set_security_env)  look at the data. 

- Following design philosophy 1 (i.e. no changes in base code) , tell the PtEID developers to make their gemsafe_set_security_env routine a noop when first called, and wait for the call to the gemsafe_compute_signature to do the real gemsafe_set_security_env. It is much more complicated, and other card may have similar issues. (Talk about making life difficult for new developers?)

frankmorgner:
Technically, Your changes are well crafted and I absolutely agree with your design principles, but if I understood correctly, @l1k has already found an elegant solution to this problem in 47347211805578c638a0271e77e2734f23a66605 (it even uses much less code to solve it!). I don't see a reason to let this very specific problem of one single card seep through the other software layers. There seems to be no other card that has these types of problems; it adds complexity without adding functionality and, at least for me, the usage `negate_hash` is not very intuitive.

What I totally agree with is that the card driver could have more control over announcing the card capabilities. That's why I'm preferring whitelisting and that's why I'm referencing `_sc_card_add_rsa_alg`. True, `_sc_card_add_rsa_alg` is not the only source of mechanisms listed in PKCS#11, but it's the only source of mechanisms that are coming from the card.

Apart from the whitelisting mentioned above, I could also think about a card capability named `max_sign_size` which gives the maximum number of bytes that the card can sign. In the PKCS#11 layer this could be translated to the appropriate mechanisms (e.g. without aggressively adding OpenSSL mechanisms)...

Regarding the change with setting the security environment, I need to take a deeper look. For example, I wonder whether the information what data is going to be signed is already present somewhere else. I even wonder why at all we need to have two card driver callbacks for setting the security environment and for issuing a signature when we are always using them together. (Read: I wonder why your noop approach isn't the default one)

dengert:
https://github.com/OpenSC/OpenSC/commit/47347211805578c638a0271e77e2734f23a66605 did not solve the problem. It only solve part of the problem. 

As stated above, @nunojpg  @l1k and myself had been working on this for some time. 

The main problem is a PKCS\#11 application in @nunojpg case OpenSSH could generate a SHA256 hash and digest, and call C_Sign  with CKM_RSA_PKCS passing in the hash and digest.  https://github.com/OpenSC/OpenSC/commit/47347211805578c638a0271e77e2734f23a66605 would not know what this was or its size when calling set_security_env because the mechanism was CKM_RSA_PKCS which can run the padding and RSA operation on any data. Accept the card had a limitation id 36 bytes and the operation would fail. 

https://github.com/OpenSC/OpenSC/commit/47347211805578c638a0271e77e2734f23a66605 would only work if OpenSC internally did the hash. The PKCS\#11 application would in that case use C_Sign with CKM_SHA256_RSA_PKCS, and  set_security_env would tell the card via  GEMSAFEV3_ALG_REF_SHA256
to expect a SHA256 hash without the digest header. 

In card-gemsafeV1.c  see:
 31 #define GEMSAFEV1_ALG_REF_FREEFORM      0x12
 32 #define GEMSAFEV3_ALG_REF_FREEFORM      0x02
 33 #define GEMSAFEV3_ALG_REF_SHA1          0x12
 34 #define GEMSAFEV3_ALG_REF_SHA256        0x42


Max_sign_size for this card would not help. The card can handle SHA1 and SHA256 but not SHA224.

dengert:
This PR is being replaced by PR #980. Please continue conversations on this subject on #980.

dengert:
Please continue conversations on this subject on #980.

Issue #970, opened 2017-02-18
Avoid dnie_transmit_apdu in the dnie driver


rickyepoderi:
Today I have started to look the code in order to delete the _dnie_transmit_apdu_ in _cwa-dnie.c_ as it was commented in pull #955. As it is known, inside the secure channel the dnie card transforms all the apdus in a CASE4 apdu. This way all the apdus through the SM need to call to the _sc_get_response_ method. The final response (after all the needed get_responses fill the data into the initial one) is the encoded apdu that should be decoded into the real apdu.

Looking to the code a typical _sc_transmit_apdu_ follows this path:

    --> sc_transmit_apdu(card, apdu): apdu.c
        |--> sc_transmit(card, pdu): apdu.c
             |--> sc_single_transmit(card, apdu): apdu.c
             |    |--> if SM_MODE_TRANSMIT sc_sm_single_transmit(card, apdu): sm.c
             |                             |--> get_sm_apdu(card, apdu, sm_apdu) (encode apdu)
             |                             |--> transmit(card, sm_apdu)
             |                             |--> free_sm_apdu(card, apdu, sm_apdu) (decode apdu)
             |--> if 0x61 sc_get_response(card, apdu, olen): apdu.c
                          |--> while data
                               |--> get_response(card, resp_len, resp) (implemented in iso7816.c)
                               |    |--> sc_transmit_apdu(card, apdu) (back to the beginning but with the new get_response apdu)
                               |--> accumulate the data to the first apdu

So the sm methods to get and free the encoded apdu are called in _sc_sm_single_transmit_. But in dnie we don't have the real encoded data until all the get_responses are transmitted and accumulated into the first apdu. That's why the _dnie_transmit_apdu_ exists, the _cwa_decode_response_ is called after all the get_responses are processed. It's impossible to emulate this behavior calling _free_sm_apdu_, just because it is never called when the initial apdu is fully filled by the get_response calls.

For DNIe we need another point of calling the _free_sm_apdu_ or another method that would be called after all the apdu is filled.

Please, correct me if I'm messing something.

dengert:
I don't use SM, so I may be all wrong on this...

But it looks like free_sm_apdu does not decode anything, it just calls 
`card->sm_ctx.ops.free_sm_apdu` 

But any decoding of the response could be done after all the get_resposnes are done. It looks like that is what  cwa_decode_response should do. But only the cwa-dnie.c calls this routine. Could it have errors, or is being called in the wrong place?

[...]

PR #975, opened 2017-02-25
Avoid dnie_transmit_apdu in the dnie driver (#970)

rickyepoderi:
This pull request is motivated by the issue #970. During that conversation some doubts arisen about how get responses should be processed inside SM communication. In the particular case of the DNIe all SM apdus work the following way:

1. If in SM the apdu is encoded and converted in a CASE4 apdu.
2. The apdu always returns 61XX (get response is needed).
3. One or more plain (no encoded) get response apdus are needed to recover the data.
4. The data is the real response apdu but encoded, it should be decoded and returned back to the caller.

So in order to avoid the use of a specific wrapper _dnie_transmit_apdu_ it is needed to call to _free_sm_apdu_ after the get responses loop. That means changes in _apdu.c_ and _sm.c_. The main idea is calling recursively to _sc_transmit_apdu_ inside _sc_sm_single_transmit_ but marking the apdu as NO_SM. It's the same idea provided by @dengert but minimally modified to be more similar to the existing flags.

Please test this pull with other SM cards. The changes are at the OpenSC core and could affect other SM implementations. I personally have only checked it with DNIe.

For more information see #970.


src/libopensc/types.h
```diff
@@ -279,6 +279,8 @@ typedef struct sc_file {
  * returns 0x6Cxx (wrong length)
  */
 #define SC_APDU_FLAGS_NO_RETRY_WL	0x00000004UL
+/* APDU is from Secure Messaging  */
```
frankmorgner: explanation could be better.

dengert:
Looks good. As @frankmorgner  pointed out [other piece of code](https://github.com/OpenSC/OpenSC/blob/678f2bb1a65e5848dffc995f63e81d1f8092352f/src/libopensc/apdu.c#L466-L469) needs to be looked at.  Only called if `card->ops->get_response` returns error and  resp_len != 0. 

@viktorTarasov should have a look.

rickyepoderi:
I have realized that the _apdu.c_ was modified by @germanblanco not very long ago in the previous try to integrate dnie into sm (https://github.com/OpenSC/OpenSC/commit/e9f94d781879f7a5fb71dc89db2b9a2a80fe19fc).

IMHO I think that not a lot of cards use the _sc_get_response_ function and, even less, in conjunction with SM. Nevertheless that sm related code in the function should be managed (removed or protected by another if that check SM condition). But I really think that part is never executed.

Two more things:

* If there are cards that need SM with the get_response apdu we will need to change a bit the present pull. (For example add the flag inside _get_sm_apdu_, each card can do SM or not inside get response,  and execute _sc_transmit_apdu_ or _card->reader->ops->transmit_ depending the same flag. The name of the flag should be changed to something more meaningful.) But I prefer to wait until current SM cards will be checked.
* @frankmorgner if the change goes on change the flag comment to whatever you like or ask me to do it.

Thanks!

dengert:
See discussion in https://github.com/OpenSC/OpenSC/issues/632 as to why https://github.com/OpenSC/OpenSC/commit/e9f94d781879f7a5fb71dc89db2b9a2a80fe19fc has the code as;
 `} while (rv != 0 && minlen != 0);` 

I don't think that needs to be changed. 

It is not a question of: "IMHO I think that not a lot of cards use the sc_get_response function",
if we have any cards that use it, we have to keep it working. And my favorite card, PIV, uses it a lot.

I don't see how you could have: "that assumes GET RESPONSE commands are SM protected"
because the get_response can not be issued on its own, it is always a type 2 APDU and is returning data that could not be transmitted  for the previous apdu, i.e. either the original apdu or a previous get_response. 
 It even has its own section in ISO 7816-4 "7. Transmission-oriented Interindustry Commands"

The only thing about a get response that might be different when using SM, is the  CLA B4 and B3 i.e. the SM bits in the class byte. The card may be expecting them because it is trying to return the results of the previous SM protected command. That may be card dependent.

Still need @viktorTarasov  to comment.

rickyepoderi:
I agree with you. My point was that the existence of that issue (it complicated the loop of get_responses) is an indication of not many cards using that code or not very intensively. I still believe the same.

The first point of my previous comment is just an idea in case this pull don't get through. A way of making the changes needed by dnie without affecting current paths of execution.

We really need testing the pull with current SM cards (authentic and epass). I suppose that Viktor is our guy here, so better wait and see.

viktorTarasov:
Sorry for the delay,
will try to test it during the week.

PR #978, opened 2017-02-27
macOS: added support for access via CryptoTokenKit

frankmorgner:
Binaries still need to be signed with the com.apple.security.smartcard
entitlement. The command should look something like this:
```
codesign --force --entitlements MacOSX/pcsc.entitlements --sign "Mac Developer" target/Library/OpenSC/bin/*
codesign --force --entitlements MacOSX/pcsc.entitlements --sign "Mac Developer" target/Library/OpenSC/lib/*.dylib
codesign --force --entitlements MacOSX/pcsc.entitlements --sign "Mac Developer" --deep target/Library/OpenSC/lib/opensc-pkcs11.bundle
codesign --force --entitlements MacOSX/pcsc.entitlements --sign "Mac Developer" --deep target/Library/Security/tokend/OpenSC.tokend
```
Note that this is my first experience with Objective-C; there may be some memory or synchronization issues left.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,6 +58,7 @@ AC_CANONICAL_HOST
 AC_PROG_CC
 # AC_PROG_CXX is needed to built the win32 custom action. Indeed dutil.h use [extern "C"] definition which fails on pure c compiler
 AC_PROG_CXX
+AC_PROG_OBJC
 PKG_PROG_PKG_CONFIG
 AC_C_BIGENDIAN
 
@@ -196,6 +197,13 @@ AC_ARG_ENABLE(
 )
 
 AC_ARG_ENABLE(
+	[cryptotokenkit],
+	[AS_HELP_STRING([--disable-cryptotokenkit],[disable CryptoTokenKit support @<:@enabled@:>@])],
+	,
+	[enable_cryptotokenkit="no"]
+)
+
+AC_ARG_ENABLE(
 	[ctapi],
 	[AS_HELP_STRING([--enable-ctapi],[enable CT-API support @<:@disabled@:>@])],
 	,
@@ -267,11 +275,11 @@ AC_ARG_WITH(
 
 dnl ./configure check
 reader_count=""
-for rdriver in "${enable_pcsc}" "${enable_openct}" "${enable_ctapi}"; do
+for rdriver in "${enable_pcsc}" "${enable_cryptotokenkit}" "${enable_openct}" "${enable_ctapi}"; do
 	test "${rdriver}" = "yes" && reader_count="${reader_count}x"
 done
 if test "${reader_count}" != "x"; then
-	AC_MSG_ERROR([Only one of --enable-pcsc, --enable-openct, --enable-ctapi can be specified!])
+	AC_MSG_ERROR([Only one of --enable-pcsc, --enable-cryptotokenkit, --enable-openct, --enable-ctapi can be specified!])
 fi
 
 dnl Checks for programs.
@@ -711,6 +719,22 @@ if test "${enable_pcsc}" = "yes"; then
 	AC_DEFINE([ENABLE_PCSC], [1], [Define if PC/SC is to be enabled])
 fi
 
+if test "${enable_cryptotokenkit}" = "yes"; then
+	if test -z "${CRYPTOTOKENKIT_CFLAGS}"; then
+		case "${host}" in
+			*-apple-*)
+				CRYPTOTOKENKIT_CFLAGS="-framework CryptoTokenKit -framework Foundation"
+				LDFLAGS="${LDFLAGS} -framework CryptoTokenKit -framework Foundation"
+			;;
+			*)
+				AC_MSG_ERROR([CryptoTokenKit only supported on Darwin])
+			;;
+		esac
+	fi
+	AC_DEFINE([ENABLE_CRYPTOTOKENKIT], [1], [Define if CryptoTokenKit is to be enabled])
+fi
+
+
 AC_SUBST(DYN_LIB_EXT)
 AC_SUBST(LIBDIR)
 AC_SUBST(LIB_PRE)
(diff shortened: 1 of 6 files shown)


configure.ac
```diff
@@ -633,6 +641,29 @@ if test "${enable_pcsc}" = "yes"; then
...
 fi
 
+if test "${enable_cryptotokenkit}" = "yes"; then
+	if test -z "${CRYPTOTOKENKIT_CFLAGS}"; then
+		case "${host}" in
+			*-apple-*)
+				# Locate the latest SDK.
+				SDKS_PATH="$(xcode-select --print-path)/Platforms/MacOSX.platform/Developer/SDKs"
+				if test -d $SDKS_PATH; then
+					SDK_PATH="${SDK_PATH:-$SDKS_PATH/$(ls -1 ${SDKS_PATH} | sort -n -t. -k2 -r | head -1)}"
```
metsma: SDK_PATH=$(xcrun -sdk macosx --show-sdk-path)

configure.ac
```diff
@@ -633,6 +641,29 @@ if test "${enable_pcsc}" = "yes"; then
...
+					CRYPTOTOKENKIT_CFLAGS="-I$SDK_PATH/System/Library/Frameworks/CryptoTokenKit.framework/Versions/Current/Headers"
+				else
+					CRYPTOTOKENKIT_CFLAGS="-I/System/Library/Frameworks/CryptoTokenKit.framework/Headers"
```
metsma: CRYPTOTOKENKIT_CFLAGS="-framework Foundation" is enough also for include headers

configure.ac
frankmorgner: The current setup seems to work. my changes are copied from e5ae77cae; @carlhoerberg do you have an opinion on that?

configure.ac
frankmorgner: My approach is identical to what we are currently using as  `PCSC_CFLAGS`. @LudovicRousseau do you have an opinion about correctly setting the flags.

frankmorgner:
By the way, big thanks to @LudovicRousseau and @dirkx for providing the only two examples dealing with CryptoTokenKit I could find on the web.

configure.ac
metsma: same goes to PCSC_CFLAGS="-framework PCSC"

metsma: And if needed we can use -isysroot ${SDK_PATH}.
OSX compilator can figure out all the paths with frameworks for include and link parameters

src/libopensc/reader-cryptotokenkit.m
```diff
@@ -0,0 +1,426 @@
...
+
+	r = sc_apdu_get_octets(reader->ctx, apdu, &sbuf, &ssize, reader->active_protocol);
+	if (r != SC_SUCCESS)
+		goto err;
+
+	if (reader->name)
+		sc_log(reader->ctx, "reader '%s'", reader->name);
+	sc_apdu_log(reader->ctx, SC_LOG_DEBUG_NORMAL, sbuf, ssize, 1);
+
+	[priv->tksmartcard transmitRequest:
```
LudovicRousseau: Maybe you could use the *synchronous* version of the `sendIns` method instead of using an asynchronous call + semaphore.
https://developer.apple.com/reference/cryptotokenkit/tksmartcard/1778266-sendins?language=objc

frankmorgner: I need low level access to send raw bytes; sendIns is not enough.

src/libopensc/reader-cryptotokenkit.m
```diff
@@ -0,0 +1,426 @@
...
+	ctk_set_proto(reader);
+
+	r = _sc_add_reader(ctx, reader);
+
+err:
+	if (r != SC_SUCCESS) {
+		free(priv);
+		if (reader != NULL) {
+			if (reader->name)
+				free(reader->name);
```
metsma: Its perfectly safe to call free on NULL pointer
C standard, 7.20.3.2/2 from ISO-IEC 9899
If ptr is a null pointer, no action occurs.

src/libopensc/reader-cryptotokenkit.m
```diff
@@ -0,0 +1,426 @@
...
+	ctk_set_proto(reader);
+
+	return SC_SUCCESS;
+}
+
+static int cryptotokenkit_disconnect(sc_reader_t * reader)
+{
+	struct cryptotokenkit_private_data *priv = reader->drv_data;
+
+	priv->tksmartcard = NULL;
```
metsma: I wonder if autorelease is here required

frankmorgner: autorelease is specified when we're getting the handle in [cryptotokenkit_use_reader](https://github.com/OpenSC/OpenSC/pull/978/files#diff-ca55e123350cb370207cb5465e4dafc3R302)

metsma: Maybe it is safer to enable ARC and let compiler do the work?

src/libopensc/reader-cryptotokenkit.m
frankmorgner: fixed

configure.ac
frankmorgner: done

src/libopensc/reader-cryptotokenkit.m
frankmorgner: This gave warnings due to a "Cast of Objective-C pointer to C Pointer type". Maybe I can get around this by better specifying ownership of the C Pointer...

frankmorgner: Well, I already spent way too long to figure out this kind of detail for some language I don't really want to use. If you make a suggestion, I'll try it out, otherwise I'll leave the code with what I've tested.

PR #979, opened 2017-02-27
#850 accepted and rebased

frankmorgner:
This PR covers most patches from https://github.com/OpenSC/OpenSC/pull/850 that have been discussed and accepted. These commits were deliberately excluded:

* b83a806a8386abe9c7c1398a9602f2d254d94a59
* db5e03cd6e74bf0f601f402c940697c4d3d2431d
* f2df771049af1a7261ed2ba92a19e5d7b3c2552a

I omitted some other Patches from #850, because they resulted in conflicts due to changes that were not accepted. If you could help resolve the conflicts, @maciejsszmigiero, we can include those as well.

Currently, the unblocking per 2a9e17514311f2fec470ea3206a5b31f7d5d4e82 is undocumented. That should be written down somewhere...

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -406,8 +406,12 @@ if test "${enable_minidriver}" = "yes"; then
 	AC_CHECK_HEADER(
 		[cardmod.h],
 		,
-		[AC_MSG_ERROR([cardmod.h is not found and required for minidriver])]
-	)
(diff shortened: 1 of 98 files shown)


frankmorgner:
Actually, the cherry-picked changes don't compile. @maciejsszmigiero could you create a PR without those three commits resolving the conflicts? This could then be quickly merged.

PR #980, opened 2017-02-27
Improve hash handling for PtEID and other cards - version 2

dengert:
This is a complete replacement for pull request for #969. To addresses @frankmorgner concerns.   

It replaces the  `negate_hashes` with `restrict_to_these_hashes`  which lists all the hashes  the card driver and card can support. If the flag is all zero, it means there are no restrictions. Thus the code has no impact on existing cards. 

It adds less code to the OpenSC base then #969 by passing  to the card driver's set_security_env the address of the data to be signed and its length and the original algorithm_flags.  The card driver can then look at the data to see if it is a hash with digest if it chooses to do so.   

The changes to card-gemsafeV1.c  for the PTEID card was rewritten to use the above. 

Since I do not have a PTEID card, this code is untested. @nunojpg and @l1k need to approve these changes. 

@frankmorgner commented that _sc_card_add_rsa_alg could be used to provide the additional information, i.e. the `restrict_to_these_hashes`. That is true, but would require changes to the _sc_card_add_rsa_alg parameters
 and major changes to framework-pkcs15.c and openssl.c. I chose to pass it by `card->restrict_to_these_hashes` to framework-pkcs15.c and openssl.c.

`restrict_to_these_hashes` could have a better name. 

#969 can be closed when this PR is accepted.

### DIFF
--- a/src/libopensc/card-gemsafeV1.c
+++ b/src/libopensc/card-gemsafeV1.c
@@ -84,6 +84,7 @@ static const u8 gemsafe_def_aid[] = {0xA0, 0x00, 0x00, 0x00, 0x63, 0x50,
 typedef struct gemsafe_exdata_st {
 	u8	aid[16];
 	size_t	aid_len;
+	u8	alg_ref; /* from last set_securuty_env */
 } gemsafe_exdata;
 
 static int get_conf_aid(sc_card_t *card, u8 *aid, size_t *len)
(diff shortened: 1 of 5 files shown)


frankmorgner:
As discussed via private mail, I believe that with modifying [`card->algorithms`](https://github.com/OpenSC/OpenSC/blob/3635dbe78a687dba032c7cccb760b6d243f08255/src/libopensc/opensc.h#L488) there is enough flexibility available to work around the limitations of the card. The combination of mechanisms can be even further restricted by specifying a [private key's algorithms](https://github.com/OpenSC/OpenSC/blob/3635dbe78a687dba032c7cccb760b6d243f08255/src/pkcs11/framework-pkcs15.c#L3747).

I don't think we need to add a new mechanism to restrict this. We'll wait for @nunojpg to check whether specifying the algorithms by hand with specific combinations works for this card.

nunojpg:
Hi, I have now tested the patch and it works fine to all my test cards.

Actually I don't use in any way the mechanisms information, so I have no comment or a prefered fix.

Thanks

dengert:
No interest from card developer.

PR #982, opened 2017-03-01
Coverity fixes from last run (though still not all of them)

Jakuje:
card-cac.c
 * CLANG_WARNING: The left operand of '<' is a garbage value

card-coolkey.c
 * CLANG_WARNING: overwriting variable
 * CPPCHECK_WARNING: memory leak / overwrite variable
 * CLANG_WARNING: null pointer dereference
 * UNUSED_VALUE: unused return value

card-gids.c
 * CLANG_WARNING: Branch condition evaluates to a garbage value
 * SIZEOF_MISMATCH: suspicious_sizeof
card-myeid.c

 * RESOURCE_LEAK: Variable "buf" going out of scope leaks the storage it points to.
 * CLANG_WARNING: overwriting variable
 * (rewrite not to confuse coverity)

pkcs15-cac.c
 * RESOURCE_LEAK: Variable "cert_out" going out of scope leaks the storage it points to.

pkcs15-coolkey.c
 * UNUSED_VALUE: unused return value

pkcs15-piv.c
 * RESOURCE_LEAK: Variable "cert_out" going out of scope leaks the storage it points to.

pkcs15-sc-hsm.c
 * DEADCODE

pkcs11/framework-pkcs15.c
 * RESOURCE_LEAK: Variable "p15_cert" going out of scope leaks the storage it points to.

pkcs15init/pkcs15-lib.c
 * CLANG_WARNING: Assigned value is garbage or undefined

pkcs15init/pkcs15-myeid.c
 * UNREACHABLE: Probably wrong placement of code block

tests/p15dump.c
 * IDENTICAL_BRANCHES

pkcs15-init.c
 * CLANG_WARNING: Potential leak of memory pointed to by 'args.der_encoded.value'

pkcs15-tool.c
 * RESOURCE_LEAK: Variable "cert" going out of scope leaks the storage it points to.
 * MISSING_BREAK: The above case falls through to this one.

sc-hsm-tool.c
 * CLANG_WARNING: Potential leak of memory pointed to by 'sp'

westcos-tool.c
 * FORWARD_NULL: Passing null pointer "pin" to "unlock_pin", which dereferences it.
 * (rewrite not to confuse coverity)

### DIFF
--- a/src/libopensc/pkcs15-cac.c
+++ b/src/libopensc/pkcs15-cac.c
@@ -352,7 +352,8 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 		r =  sc_pkcs15_read_certificate(p15card, &cert_info, &cert_out);
 		if (r < 0 || cert_out->key == NULL) {
 			sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "Failed to read/parse the certificate r=%d",r);
-			sc_pkcs15_free_certificate(cert_out);
+			if (cert_out != NULL)
+				sc_pkcs15_free_certificate(cert_out);
 			continue;
 		}
 
--- a/src/libopensc/pkcs15-cac.c
+++ b/src/libopensc/pkcs15-cac.c
@@ -292,7 +292,7 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 		struct sc_pkcs15_object pubkey_obj;
 		struct sc_pkcs15_object prkey_obj;
 		sc_pkcs15_der_t   cert_der;
-		sc_pkcs15_cert_t *cert_out;
+		sc_pkcs15_cert_t *cert_out = NULL;
 
 		r = (card->ops->card_ctl)(card, SC_CARDCTL_CAC_GET_NEXT_CERT_OBJECT, &obj_info);
 
@@ -352,12 +352,14 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 		r =  sc_pkcs15_read_certificate(p15card, &cert_info, &cert_out);
 		if (r < 0 || cert_out->key == NULL) {
 			sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "Failed to read/parse the certificate r=%d",r);
+			sc_pkcs15_free_certificate(cert_out);
 			continue;
 		}
 
 		r = sc_pkcs15emu_add_x509_cert(p15card, &cert_obj, &cert_info);
 		if (r < 0) {
 			sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, " Failed to add cert obj r=%d",r);
+			sc_pkcs15_free_certificate(cert_out);
 			continue;
 		}
 		/* set the token name to the name of the CN of the first certificate */
@@ -393,6 +395,7 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 				 usage, pubkey_info.usage, prkey_info.usage);
 		if (cert_out->key->algorithm != SC_ALGORITHM_RSA) {
 			sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL,"unsupported key.algorithm %d", cert_out->key->algorithm);
+			sc_pkcs15_free_certificate(cert_out);
 			continue;
 		} else {
 			pubkey_info.modulus_length = cert_out->key->u.rsa.modulus.len * 8;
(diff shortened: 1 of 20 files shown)


src/libopensc/pkcs15-cac.c
```diff
@@ -352,12 +352,14 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 		r =  sc_pkcs15_read_certificate(p15card, &cert_info, &cert_out);
 		if (r < 0 || cert_out->key == NULL) {
 			sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "Failed to read/parse the certificate r=%d",r);
+			sc_pkcs15_free_certificate(cert_out);
```
frankmorgner: if `cert_out` is `NULL`, then we will hit an `assert` in `sc_pkcs15_free_certificate`. So you should check for NULL here or remove the assert.

src/pkcs15init/pkcs15-myeid.c
```diff
@@ -510,10 +510,10 @@ myeid_store_key(struct sc_profile *profile, struct sc_pkcs15_card *p15card,
 				keybits = key_info->field_length;
 			else
 				key_info->field_length = keybits;
-			break;
-			
+
 			if (sc_card_find_ec_alg(p15card->card, keybits, &prkey->u.ec.params.id) == NULL)
 				LOG_TEST_RET(ctx, SC_ERROR_INVALID_ARGUMENTS, "Unsupported algorithm or key size");			
+			break;
```
frankmorgner: This change should be confirmed with an actual card. @mtrojnar or @hhonkanen could you have a look?

src/libopensc/pkcs15-cac.c
Jakuje: Yes, I will add the null check here to make it sane. Thanks for pointing it out (already fixed in the other case).

src/pkcs15init/pkcs15-myeid.c
Jakuje: Thank you for pointing to the appropriate people to discuss this change with.

mtrojnar: The `LOG_TEST_RET(..., SC_ERROR_INVALID_ARGUMENTS, ...)` macro always returns `SC_ERROR_INVALID_ARGUMENTS`, so `break;` is redundant (does not do anything) here.  The code would be more clear if `SC_FUNC_RETURN()` was used instead of `LOG_TEST_RET()`.

Jakuje: @mtrojnar have a look on the full diff of the file. Originally, the `break` was before the `sc_card_find_ec_alg()` so it was not even called.

Also if `sc_card_find_ec_alg()` succeed, it would fall through to the `default` of the switch, which is certainly not intended.

dengert:
The comment from @frankmorgner : 
When calling sc_pkcs15_free_certificate with NULL, we will hit an assert. So you should check for NULL before or remove the assert in the function.:

also applies to the  change to card-piv.c and needs  a change like https://github.com/OpenSC/OpenSC/pull/982/commits/e73b2ad2e01cbcc3fdee471ce9692ab95a83b8a0

Jakuje:
*banging my head against the table*. Obviously. Thanks for patience :)

src/pkcs15init/pkcs15-myeid.c
hhonkanen: The change looks sensible and the break was clearly in wrong place. In the previous function "myeid_create_key", there is the same check implemented correctly and I am 95% sure that this is ok, because I have tested this code. Please look at lines 440-447. I can test with an actual card but right now I am under high workload so it might take a while before I have the chance.

src/scconf/scconf.c
```diff
@@ -411,6 +411,8 @@ char *scconf_list_strdup(const scconf_list * list, const char *filler)
 	if (filler) {
 		len += scconf_list_array_length(list) * (strlen(filler) + 1);
 	}
+	if (len == 0)
+		return NULL;
 	buf = malloc(len);
```
frankmorgner: according to the Linux manual, calling `malloc` with `0` should be fine, but that may be different on other platforms.

src/pkcs15init/pkcs15-myeid.c
frankmorgner: @hhonkanen did you have time to test the change?

hhonkanen: Not yet, but I am trying to find some time for OpenSC dev this and next week. I will confirm this change after I have tested.

frankmorgner: OK, thanks

hhonkanen: I tested myeid_store_key as follows: Using pkcs15-init tool I imported a 521 bit ECC key into a card that supports 521 bit keys. Traced through this function with debugger and it goes fine, the key was imported into the card. 

Tested the same command with a card that supports ECC only up to 256 bit keys. We don't even get into this function, OpenSC notices earlier that the key size is not listed as supported and returns with error.

frankmorgner: OK, thanks. Sounds good enough for me.

Issue #892, opened 2016-10-19
Merge cardmod reader driver with pcsc reader driver


frankmorgner:
### Expected behaviour

PC/SC reader driver should be identical to the reader driver that's used in the minidriver ("cardmod").
### Actual behaviour
- cardmod reader driver searches for the almost undocumented "cardmod" block in `opensc.conf`.
- Since e95b5151631f71fcee60a24d8d9d79f1c9719d8c the user needs to configure `max_send_size` and `max_recv_size` for both, the pcsc reader and the cardmod reader (while the latter is undocumented, as said above) for a reader that doesn't announce `dwMaxAPDUDataSize`.
- cardmod reader configuration and implementation misses features of what is already implemented in PC/SC (e.g. PACE)
- there is a lot of code duplication in `reader-pcsc.c`
- `reader-pcsc.c` contains comments about removing cardmod driver somewhere in the future.
### How to proceed?

Merge cardmod and PC/SC driver! Remove some code and use the same configuration blocks.

Since the cardmod driver has been there for years, I'd like to know if the early maintainers are fine with this... @martinpaljak @dengert @viktorTarasov

martinpaljak:
Absolutely. I practically gave up on OpenSC after the merge of this one.

dengert:
Maybe I am missing something here. The cardmod driver is in ./reader-pcsc.c. 

The main difference is reader-pcsc.c looks at all readers PC/SC provides and presents all the readers to the calling application so the application can select a reader. 

Cardmod driver is given a pcsc_context_handle and pcsc_card_handle Microsoft code has already selected the reader and opened a PC/SC connection and passed these to the minidriver.  And while the minidriver is still running the Microsoft code can presents a new pcsc_context_handle and pcsc_card_handle.  The Microsoft code also handles the PCSC locking as best I can tell.

The main difference is thus:
`cardmod_ops.detect_readers = NULL;`
`cardmod_ops.use_reader = cardmod_use_reader;`

"Merge cardmod and PC/SC driver! Remove some code and use the same configuration blocks." is a start.

frankmorgner:
OK, great! Thanks for the feedback.

PR #983, opened 2017-03-01
reader-pcsc: removed cardmod driver

frankmorgner:
- pcsc driver takes over all the functionality
- no dedicated reader driver config values for cardmod, use application specific blocks to define a different behavior for the pcsc reader if needed
- removes legacy code; requiring at least libpcsclite 1.6.5

Fixes https://github.com/OpenSC/OpenSC/issues/892

Tested with:
- `certutil -scinfo` (Windows 7)
- `pkcs11-tool --test --login` (pcsc-lite version 1.8.20)
- Safari Client authentication (macoOS Sierra)

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -769,7 +769,7 @@ app onepin-opensc-pkcs11 {
 	}
 }
 
-# Used by OpenSC.tokend on Mac OS X only.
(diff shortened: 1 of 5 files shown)


frankmorgner:
@LudovicRousseau could you have a look if I accidentally missed something?

PR #985, opened 2017-03-03
Add the final NUL to the CN used for token label

martinpaljak:
Arbitrary strings were shown in PKCS#11 apps (even though the field is cut to 32 characters)

### DIFF
--- a/src/libopensc/pkcs15-esteid.c
+++ b/src/libopensc/pkcs15-esteid.c
@@ -117,7 +117,10 @@ sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
 				sc_pkcs15_get_name_from_dn(card->ctx, cert->subject,
 					cert->subject_len, &cn_oid, &cn_name, &cn_len);
 				if (cn_len > 0) {
-					set_string(&p15card->tokeninfo->label, (const char*)cn_name);
+					char *token_name = malloc(cn_len+1);
+					memcpy(token_name, cn_name, cn_len);
+					token_name[cn_len] = 0;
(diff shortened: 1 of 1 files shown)


src/libopensc/pkcs15-esteid.c
```diff
@@ -117,7 +117,10 @@ sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
 				sc_pkcs15_get_name_from_dn(card->ctx, cert->subject,
 					cert->subject_len, &cn_oid, &cn_name, &cn_len);
 				if (cn_len > 0) {
-					set_string(&p15card->tokeninfo->label, (const char*)cn_name);
+					char *token_name = malloc(cn_len+1);
```
frankmorgner: please add error checking

src/libopensc/pkcs15-esteid.c
```diff
@@ -117,7 +117,10 @@ sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
...
+					memcpy(token_name, cn_name, cn_len);
+					token_name[cn_len] = 0;
```
frankmorgner: should use `'\0'`

src/libopensc/pkcs15-esteid.c
```diff
@@ -117,7 +117,10 @@ sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
...
+					set_string(&p15card->tokeninfo->label, (const char*)token_name);
```
frankmorgner: token_name needs to be freed

PR #900, opened 2016-11-10
pkcs15-tool --list-info and tree (supersede #885)

nunojpg:
Hi,

I've implemented:

```
./pkcs15-tool --list-info
PKCS#15 Card [CARTAO DE CIDADAO]:
	Version        : 1
	Serial number  : 000004093842343
	Manufacturer ID: GEMALTO
	Flags          : Read-only, PRN generation, EID compliant
```

and

```
./pkcs15-tool -D --tree  
PKCS#15 Card [CARTAO DE CIDADAO]:
	Version        : 1
	Serial number  : 000004093842343
	Manufacturer ID: GEMALTO
	Flags          : Read-only, PRN generation, EID compliant
Authentication object(s) [3]
  PIN | 01 | Ref:129 (0x81) | Tries left:3 | Auth PIN
  PIN | 02 | Ref:130 (0x82) | Tries left:3 | Sign PIN
  PIN | 03 | Ref:131 (0x83) | Tries left:3 | Address PIN
Private key(s) [2]
  RSA[1024] | 45 | Ref:2 (0x2) | Auth ID:01 | CITIZEN AUTHENTICATION KEY [0x4, sign]
  RSA[1024] | 46 | Ref:1 (0x1) | Auth ID:02 | CITIZEN SIGNATURE KEY [0x200, nonRepudiation]
Public key(s) [0]
Certificate(s) [5]
  Path:3f005f00ef09 | 45
  Path:3f005f00ef08 | 46
  Path:3f005f00ef0f | 51
  Path:3f005f00ef10 | 52
  Path:3f005f00ef11 | 50
Data object(s) [5]
  Path:3f005f00ef02 | Size: 15500 | Citizen Data        
  Path:3f005f00ef05 | Auth ID: 03 | Citizen Address Data
  Path:3f005f00ef07 | Size:  1000 | Citizen Notepad     
  Path:3f005f00ef06 | Size:  4000 | SOD                 
  Path:3f000003     | Size:     6 | TRACE
```

The tree format is highly dependent on the specific objects present on the card, and I tried to make it as generic as possible, but it must be checked with other cards to make sure the output continues to be reasonable. Please give it a try.

Thanks!

PR #989, opened 2017-03-07
pkcs11-tool: login if pin_cache disabled or exhausted

nunojpg:
This fixes the remaining case mentioned on #948 according to the suggestion of reacting to a CKR_USER_NOT_LOGGED_IN by trying to login.

| CKA_ALWAYS_AUTHENTICATE | FALSE                             | TRUE |
|----------------------------------------------|-----------------------------------|---------|
| With pinpad                                   | OK (with merged #948)    | OK   |
| Without pinpad                              | OK after this pull request | OK   |

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2945,29 +2945,10 @@ VARATTR_METHOD(EC_POINT, unsigned char);
 VARATTR_METHOD(EC_PARAMS, unsigned char);
 
 static void  authenticate_if_required(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE privKeyObject){
-	CK_SESSION_INFO sessionInfo;
 	CK_TOKEN_INFO	info;
-	CK_RV rv;
-
-	rv = p11->C_GetSessionInfo(session, &sessionInfo);
-	if (rv != CKR_OK)
-		p11_fatal("C_OpenSession", rv);
-
-	switch(sessionInfo.state){
-		case CKS_RW_USER_FUNCTIONS:
-		   	//logged in, not need to continue.
-			return;
-		case CKS_RW_PUBLIC_SESSION:
-			break;
-		default:
-			util_fatal("unexpected state");
-	}
-
 	get_token_info(opt_slot, &info);
-	if (!(info.flags & CKF_PROTECTED_AUTHENTICATION_PATH) && !getALWAYS_AUTHENTICATE(session, privKeyObject))
-		return;
-
-	login(session,CKU_CONTEXT_SPECIFIC);
+	if ((info.flags & CKF_PROTECTED_AUTHENTICATION_PATH) || getALWAYS_AUTHENTICATE(session, privKeyObject))
+		login(session,CKU_CONTEXT_SPECIFIC);
 }
 
 static void list_objects(CK_SESSION_HANDLE sess, CK_OBJECT_CLASS  object_class)
@@ -3934,17 +3915,17 @@ static int test_digest(CK_SESSION_HANDLE session)
 		for (j = 0; j < 10; j++)
 			data[10 * i + j] = (unsigned char) (0x30 + j);
 
-
 	for (i = 0; mechTypes[i] != 0xffffff; i++) {
 		ck_mech.mechanism = mechTypes[i];
 
+		printf("  %s: ", p11_mechanism_to_name(mechTypes[i]));
 		rv = p11->C_DigestInit(session, &ck_mech);
-		if (rv == CKR_MECHANISM_INVALID)
+		if (rv == CKR_MECHANISM_INVALID){
+			fprintf(stderr, "Mechanism not supported\n");
 			continue;	/* mechanism not implemented, don't test */
+		}
 		if (rv != CKR_OK)
 			p11_fatal("C_DigestInit", rv);
-
-		printf("  %s: ", p11_mechanism_to_name(mechTypes[i]));
 
 		hashLen1 = sizeof(hash1);
 		rv = p11->C_Digest(session, data, sizeof(data), hash1,
@@ -4081,7 +4062,7 @@ static int sign_verify_openssl(CK_SESSION_HANDLE session,
 		unsigned char *verifyData, CK_ULONG verifyDataLen,
 		CK_ULONG modLenBytes, int evp_md_index)
 {
-	int 		errors = 0;
+	int 		errors = 0, first_login;
 	CK_RV           rv;
 	unsigned char   sig1[1024];
 	CK_ULONG        sigLen1;
@@ -4102,23 +4083,31 @@ static int sign_verify_openssl(CK_SESSION_HANDLE session,
 #endif
 	};
 #endif
+	for(first_login = 1; ; first_login = 0){
+		rv = p11->C_SignInit(session, ck_mech, privKeyObject);
+		/* mechanism not implemented, don't test */
+		if (rv == CKR_MECHANISM_INVALID)
+			return errors;
+		if (rv != CKR_OK)
+			p11_fatal("C_SignInit", rv);
 
-	rv = p11->C_SignInit(session, ck_mech, privKeyObject);
-	/* mechanism not implemented, don't test */
-	if (rv == CKR_MECHANISM_INVALID)
-		return errors;
-	if (rv != CKR_OK)
-		p11_fatal("C_SignInit", rv);
-
-	authenticate_if_required(session, privKeyObject);
-	printf("    %s: ", p11_mechanism_to_name(ck_mech->mechanism));
-
-	sigLen1 = sizeof(sig1);
-	rv = p11->C_Sign(session, data, dataLen, sig1,
-		&sigLen1);
-	if (rv != CKR_OK)
-		p11_fatal("C_Sign", rv);
-
+		if(first_login){
+			printf("    %s: ", p11_mechanism_to_name(ck_mech->mechanism));
+			authenticate_if_required(session, privKeyObject);
+		}
+		
+		sigLen1 = sizeof(sig1);
+		rv = p11->C_Sign(session, data, dataLen, sig1,
+			&sigLen1);
+		if (rv == CKR_USER_NOT_LOGGED_IN){
+			login(session,opt_login_type);
+			continue;
+		}
+		if (rv != CKR_OK)
+			p11_fatal("C_Sign", rv);
+		break;
+	}
+	
 	if (sigLen1 != modLenBytes) {
 		errors++;
 		printf("  ERR: wrong signature length: %u instead of %u\n",
@@ -4308,15 +4297,23 @@ static int test_signature(CK_SESSION_HANDLE sess)
 		if (rv != CKR_OK)
 			p11_fatal("C_SignFinal", rv);
 
-		rv = p11->C_SignInit(sess, &ck_mech, privKeyObject);
-		if (rv != CKR_OK)
-			p11_fatal("C_SignInit", rv);
 		authenticate_if_required(sess, privKeyObject);
 
-		sigLen2 = sizeof(sig2);
-		rv = p11->C_Sign(sess, data, dataLen, sig2, &sigLen2);
-		if (rv != CKR_OK)
-			p11_fatal("C_Sign", rv);
+		for(;;){
+			rv = p11->C_SignInit(sess, &ck_mech, privKeyObject);
+			if (rv != CKR_OK)
+				p11_fatal("C_SignInit", rv);
+
+			sigLen2 = sizeof(sig2);
+			rv = p11->C_Sign(sess, data, dataLen, sig2, &sigLen2);
+			if (rv == CKR_USER_NOT_LOGGED_IN){
+				login(sess,opt_login_type);
+				continue;
+			}
+			if (rv != CKR_OK)
+				p11_fatal("C_Sign", rv);
+			break;
+		}
 
 		if (sigLen1 != sigLen2) {
 			errors++;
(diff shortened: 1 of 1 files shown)


dengert:
Using this patch with a PIN PAD reader, (CM 3821),   the PIN is requested for every operation, because 
`info.flags & CKF_PROTECTED_AUTHENTICATION_PATH` is always true.

Either:
1. The deleted code to test for `case CKS_RW_USER_FUNCTIONS:` needs to be added back, and the line 
`//logged in, not needed to continue.`
should be replaced by: 
`if (getALWAYS_AUTHENTICATE(session, privKeyObject))`
                 `login(session,CKU_CONTEXT_SPECIFIC);`


2. the test that the session for a RW session  when requesting  CKU_CONTEXT_SPECIFIC should be added to the login() routine.

nunojpg:
Hi,

That code had a bug, in which the test was being made for the whole variable and just not testing the bit.

It should have been: "if(sessionInfo.state & CKS_RW_USER_FUNCTIONS)".

And I removed it because all the paths for authenticate_if_required already include this test in the call path.

Could you please let me know in which case you are seeing a wrong behavior?

frankmorgner:
Please use `goto` only for simplifying error handling. If possible, don't create new spaghetti code (we already have enough of that)...

nunojpg:
@frankmorgner clearly it could be improved.

I also introduced now a small format change for pkcs --test:

* remove "testing signature mechanisms:"
* print "Mechanism not supported" when is the case.
* indent "all 4 signature functions seem to work" at key level.

Master:

```
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (CITIZEN AUTHENTICATION KEY)
  testing signature mechanisms:
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
  testing key 0 (CITIZEN AUTHENTICATION KEY)
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
Unwrap: not implemented
  testing key 0 (CITIZEN AUTHENTICATION KEY)  -- can't be used to decrypt, skipping
No errors
```

Becomes:

```
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (CITIZEN AUTHENTICATION KEY)
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
  testing key 0 (CITIZEN AUTHENTICATION KEY)
    RSA-X-509: Mechanism not supported
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
Unwrap: not implemented
  testing key 0 (CITIZEN AUTHENTICATION KEY)  -- can't be used to decrypt, skipping
No errors
```

frankmorgner:
please also have a look at the build errors from the CI platforms

nunojpg:
Since there was no comment againts https://github.com/OpenSC/OpenSC/pull/900#issuecomment-261192029, I had the idea OpenSC was now relying on C99.... Fixed.

dengert:
You stated above: "That code had a bug, in which the test was being made for the whole variable and just not testing the bit. It should have been: "if(sessionInfo.state & CKS_RW_USER_FUNCTIONS)".

No, sessionInfo.state is of type CKS_STATE that has 5 possible values, and can be used in a case statement.
You may have been confusing sessionInfo.state with sessionInfo.flags that is of type CKF_FLAGS, 
and  CKS_RW_USER_FUNCTIONS with CKF_RW_SESSION. 

I would like to propose an alternative patch https://github.com/dengert/OpenSC/commits/pkcs11-tool-fix that fixes the issues in authenticate_if_needed of calling login for every crypto operation if a PIN PAD reader is used.

nunojpg:
It looks I am once again confused about the CKS/CKF stuff :(

For me all worked fine when I was using a PINPAD, but it failed when I did a pkcs11-tool --login --test that required 12 logins, and my pin cache allowed for less than that.

With this pull request all the 4 combinations of PINPAD and CKA_ALWAYS_AUTHENTICATE I tested to work.

With your branch, without PINPAD and regardless of pincache enabled or disabled:

```
$ pkcs11-tool -lt
Logging in to "Auth PIN (CARTAO DE CIDADAO)".
Please enter User PIN:
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (CITIZEN AUTHENTICATION KEY)
Logging in to "Auth PIN (CARTAO DE CIDADAO)".
Please enter User PIN:
error: PKCS11 function C_Login failed: rv = CKR_USER_ALREADY_LOGGED_IN (0x100)
Aborting.
```

So for me is totally unclear:

* What was the problem you were having with master when using a Pinpad.
* What is the problem you are detecting on my branch.
* How does your branch solve the reason for this pull request: "pkcs11-tool: login if pin_cache disabled or exhausted".

dengert:
Both our changes added a call in the decrypt tests because it had a missing call to authenticate_if_needed.  That being said...

On master I just lived with what pkcs11-tool did and did would not test pinpad with pkcs11-tool. Other programs could test  the pinpad. I do not use the pin cache, and try and setup the environment so a single login will keep the card in the login state, and as needed by CKA_ALWAYS_AUTHENTICATE to enter the pin again. 

I do not put the PIN on the pkcs11-tool command line, I want to be prompted for the PIN when it is needed and never use the pin_cache. 

 PKCS#11 2.40 says:

> The CKA_ALWAYS_AUTHENTICATE attribute can be used to force re-authentication (i.e. force the user to provide a PIN) for each use of a private key. “Use” in this case means a cryptographic operation such as sign or decrypt. This attribute may only be set to CK_TRUE when CKA_PRIVATE is also CK_TRUE.

> Re-authentication occurs by calling C_Login with userType set to CKU_CONTEXT_SPECIFIC immediately after a cryptographic operation using the key has been initiated (e.g. after C_SignInit). In this call, the actual user type is implicitly given by the usage requirements of the active key. If C_Login returns CKR_OK the user was successfully authenticated and this sets the active key in an authenticated state that lasts until the cryptographic operation has successfully or unsuccessfully been completed (e.g. by C_Sign, C_SignFinal,..). A return value CKR_PIN_INCORRECT from C_Login means that the user was denied permission to use the key and continuing the cryptographic operation will result in a behavior as if C_Login had not been called. In both of these cases the session state will remain the same...

(There may be other context specific cases to use it, but it is not a substitute for a normal C_Login.)

The problem with your patch as I stated above when using a PIN PAD: " the PIN is requested for every operation, because info.flags & CKF_PROTECTED_AUTHENTICATION_PATH is always true."
The card has 4 keys. Only one has  CKA_ALWAYS_AUTHENTICATE. 
Thus it was doing 4 times the pin requests that it should have done. 

You change to authenticate_if_needed also removed any test to the state of the session, i.e. 
CKS_*_FUNCTIONS or not. Most failures that return CKR_USER_NOT_LOGGED_IN are for the wrong type
of session. 

I thing the main problem is authenticate_if_needed was designed to handle CKA_ALWAYS_AUTHENTICATE and do C_Login(CKU_CONTEXT_SPECIFIC). 

Note:
authenticate_if_needed is called **before** each crypto operation. It is not called to test if the pin_cache has run out. It could, but its not clear how to tell if the pin_cache_limit has been reached without an operation failing with CKR_USER_NOT_LOGGED_IN.  

What we need is another way to handle CKR_USER_NOT_LOGGED_IN and does another C_LOGIN
that is not  CKU_CONTEXT_SPECIFIC. PKCS#11 2.40 says:

> If the user type is CKU_CONTEXT_SPECIFIC , the behavior of C_Login depends on the context in which it is called. Improper use of this user type will result in a return value CKR_OPERATION_NOT_INITIALIZED..


We could have a authenticate_because_ckr_user_not_logged_in (or some other name) to handle retries if 
ckr_user_not_logged_in is returned. 

What we need is a PKCS11SPY trace when the pin_cache limit is reached, to see what is the state of the session.  **If the state is  CKS_*_FUNCTIONS and function got a  CKR_USER_NOT_LOGGED_IN,
that would imply to try and login again.** 

As far as I know, how a pkcs11 module can do pin caching and tell the application the pin_cache_limit has been reached is not defined in PKCS#11 specs. What is defined is a session can lose it login state, for many reasons, and CKR_USER_NOT_LOGGED_IN can be returned by may functions. 

pkcs11-tool needs to work with more then just opensc-pkcs11.so.

nunojpg:
I'm not understanding what is going in many levels.

When CKF_PROTECTED_AUTHENTICATION_PATH is true, doesn't it mean that the user must insert the pin for every operation? How could the pinpad cache it?

And are you reporting it as a general bug or as a regression created by this patch?

dengert:
Here is another though...

The pin_cache_code sets the pin_cache_counter. But in my option, can get in the way if an application such as pkcs11-tool  can not handle CKR_USER_NOT_LOGGED_IN. 

Using my patch, 
pkcs11-tool does one C_Login to authenticate to the card, then does all its tests.

So the "pin_cache_counter" from the opensc.conf  that maps to 
and is only called from `sc_pkcs15_pincache_revalidate` from `use_key` in `pkcs15-sec.c` 
when a command fails. So  a single C_login should work for all keys that do not have CKA_ALWAYS_AUTHENTICATE or user_consent. 

`sc_pkcs15_pincache_add` will set a `pin_obj->usage_counter = 0;`
and `sc_pkcs15_pincache_revalidate` does `pin_obj->usage_counter++;`
and this is the only place where `pin_obj->usage_counter`. 

So I don't see why a default pin_cache_counter would not work in pkcs11-tool.

Wint modified opensc.conf, running `/opt/ossl-1.0.2/bin/pkcs11-tool --module /opt/ossl-1.0.2/lib/pkcs11-spy.so -l -t`
shows `use_pin_cache=1 pin_cache_counter=3 pin_cache_ignore_user_consent=0`
and so calls no calls to  `sc_pkcs15_pincache_revalidate`
running under gdb confirms it is not called.

Spy shows CKR_USER_NOT_LOGGED_IN is never returned. 
 
Does you card or card require the pin for every crypto operation or for one of the keys?
Does any of its PKCS#15 keys set user_consent?
user_consent can be an integer, not just one. 

pkcs#11 only has CKA_ALWAYS_AUTHENTICATE which in effect is implied if user_consent==1
But it is assumed that CKA_ALWAYS_AUTHENTICATE is fixed for an object, and does not change based on how may operations have been.

Is you card drive doing its own pin_cache  code doing its own usage counter?

dengert:
CKF_PROTECTED_AUTHENTICATION_PATH  PKCS#11 says:
> means that there is some way for a user to be authenticated to the token without having the application send a PIN through the Cryptoki library. One such possibility is that the user enters a PIN on a PINpad

In C_Login it says:
> To log into a token with a protected authentication path, the pPin parameter to C_Login should be NULL_PTR.

Note that PKCS#11 leaves it up to the application to look at the flag and set the NULL_PTR. But if its a good pin pad reader, the reader will not allow a normal verify command to pass, thus forcing the use of the pin pad reader. 

The pkcs11-tool.c:login()   will set the NULL_PTR if CKF_PROTECTED_AUTHENTICATION_PATH is set. 
So authenticate_if_needed does not need to do anything with it as it calls login().

No Pin Pad reader I know of can cache pins in the reader.  OpenSC sends a template of the verify command that the reader then fills in with teh PIN typed on the pin pad reader. So in theory it could
be cached on the reader. 
Normally the user has to enter it each time. The whole  point of a pin pad reader is to enforce a policy to to make the user send in the pin via the pin pad reader each time, thus never exposing the pin to the host where it could be stolen or cached thus avoiding the policy the card provider wants enforced. 
(Not perfect but close.)

OpenSC has a pin_cache capability, that if a crypto operation fails, it will call `sc_pkcs15_pincache_revalidate` and use a cached pin it it is not at its limit.
` if (pin_obj->usage_counter >= p15card->opts.pin_cache_counter) {`

Maybe the problem is  pin_obj->usage_counter is not initialized to zero?

You asked: "How does your branch solve the reason for this pull request: "pkcs11-tool: login if pin_cache disabled or exhausted".
Even when I am using the pin cache, and setting the limit to 3, I never run into the limit.
CKA_ALWAYS_AUTHENTICATE is handled differently then a normal pin. and does not count
toward the usage_count. 



So I am asking you if sc_pkcs15_pincache_revalidate is ever called. 




and  up the usege count for the pin
but only if the usage count is not at its limit.

dengert:
You also asked: "When CKF_PROTECTED_AUTHENTICATION_PATH is true, doesn't it mean that the user must insert the pin for every operation? How could the pinpad cache it?"

Not every operation, only for any login.

CKA_ALWAYS_AUTHENTICATE for a key, means that pin must be entered before every crypto operation 
using that key. If the token also has CKF_PROTECTED_AUTHENTICATION_PATH then the pin needs to be entered. 

(Some PKCS#11 applications can not handle CKA_ALWAYS_AUTHENTICATE/user_consent see opensc.conf  pin_cache_ignore_user_consent which allow an opensc cached pin to be used.)

nunojpg:
I require 12 logins for a pkcs11-tool --test with my card, either with a slot marked with user_consent or not.

So in my case the default pincache_count=10 triggers a failure, because I would need at least 11.

> Using this patch with a PIN PAD reader, (CM 3821), the PIN is requested for every operation, because
> info.flags & CKF_PROTECTED_AUTHENTICATION_PATH is always true.

So was this a regression report or just a confirmation that all works as expected?

dengert:
Can you send a opensc-debug log and a spy log?

(I have to run... will reply more later.)

nunojpg:
I require 12 logins for a pkcs11-tool --test with my card, either with a slot marked with user_consent or not.

So in my case the default pincache_count=10 triggers a failure, because I would need at least 11.

I think I now understand the problem: Login is attempted for operations which shouldn't require it.

I can't experience that for my card, so I will forward you a log using a pinpad for:

* Master - slot user_consent=false
* Master - slot user_consent=true
* This PR - slot user_consent=false
* This PR - slot user_consent=true

Thanks!

nunojpg:
I've discussed with @dengert offline and will continue the discussion here by beginning of April.

nunojpg:
After reviewing #1039 and @dengert comments I don't think I can do any contribution to solve this, so I'm abandoning it.

dengert:
Also see #1064

PR #990, opened 2017-03-09
MD: add msroots file

hamano:
MSDN spec say "mscp\msroots" not "mscp\msroot".
https://msdn.microsoft.com/windows/hardware/drivers/smartcard/file-system-requirements
I'm testing now.

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -1159,9 +1159,9 @@ md_set_cardid(PCARD_DATA pCardData, struct md_file *file)
 	return SCARD_S_SUCCESS;
 }
 
-/* fill the msroot file from root certificates */
+/* fill the msroots file from root certificates */
 static void
-md_fs_read_msroot_file(PCARD_DATA pCardData, char *parent, struct md_file *file)
+md_fs_read_msroots_file(PCARD_DATA pCardData, char *parent, struct md_file *file)
 {
 	CERT_BLOB dbStore = {0};
 	HCERTSTORE hCertStore = NULL;
@@ -1298,8 +1298,8 @@ md_fs_read_content(PCARD_DATA pCardData, char *parent, struct md_file *file)
 			CopyMemory(file->blob, cert->data.value, cert->data.len);
 			sc_pkcs15_free_certificate(cert);
 		}
-		if (!strcmp(file->name, "msroot")) {
-			md_fs_read_msroot_file(pCardData, parent, file);
+		if (!strcmp(file->name, "msroots")) {
+			md_fs_read_msroots_file(pCardData, parent, file);
 		}
 	}
 	else   {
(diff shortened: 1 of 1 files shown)


src/minidriver/minidriver.c
```diff
@@ -1490,16 +1490,16 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
...
+			else if ((prkey_info->usage & USAGE_ANY_SIGN) == USAGE_ANY_SIGN)
 				cont->size_sign = prkey_info->modulus_length;
 			else
 				cont->size_key_exchange = prkey_info->modulus_length;
 		} else if (key_obj->type == SC_PKCS15_TYPE_PRKEY_EC) {
-			if (prkey_info->usage & USAGE_ANY_AGREEMENT)
+			if ((prkey_info->usage & USAGE_ANY_AGREEMENT) == USAGE_ANY_AGREEMENT)
 				cont->size_key_exchange = prkey_info->field_length;
-			else if (prkey_info->usage & USAGE_ANY_SIGN)
+			else if ((prkey_info->usage & USAGE_ANY_SIGN) == USAGE_ANY_SIGN)
```
frankmorgner: Your change forces both flags to be set: `SC_PKCS15_PRKEY_USAGE_SIGN ` **and** `SC_PKCS15_PRKEY_USAGE_NONREPUDIATION`. The current logic only enforced at least one flag to be set. I think this is not intended and your changes about checking the key usage should be removed also in the other cases.

src/minidriver/minidriver.c
```diff
@@ -1402,7 +1402,7 @@ md_fs_add_msroot(PCARD_DATA pCardData, struct md_file **head)
 	for(ii = 0; ii < cert_num; ii++)   {
 		struct sc_pkcs15_cert_info *cert_info = (struct sc_pkcs15_cert_info *) prkey_objs[ii]->data;
 		if (cert_info->authority) {
-			dwret = md_fs_add_file(pCardData, head, "msroot", EveryoneReadUserWriteAc, NULL, 0, NULL);
+			dwret = md_fs_add_file(pCardData, head, "msroots", EveryoneReadUserWriteAc, NULL, 0, NULL);
```
frankmorgner: Although you're certainly right about the documentation, I didn't experience any problems with the current setup, did you? @vletoux  do you have an opinion on that?

vletoux: @hamano is right: in all my projects, I'm using either szROOT_STORE_FILE or directly the string "msroots".  I may have done the typo in my test program and then reported it in the source code. Sorry.

src/minidriver/minidriver.c
vletoux: agreed: this is a breaking change. I think you meant != 0 instead.

src/minidriver/minidriver.c
frankmorgner: ok, thanks for your feedback

src/minidriver/minidriver.c
hamano: Our JPKI card has two keys:

~~~
% pkcs15-tool -D
Private RSA Key [User Authentication Key]
        Object Flags   : [0x1], private
        Usage          : [0x4], sign

Private RSA Key [Digital Signature Key]
        Object Flags   : [0x1], private
        Usage          : [0x204], sign, nonRepudiation
~~~

The first key for authentication such as TLS client auth, and second one for signature.
Also other national cards are roughly  similar.
But minidriver would handle both key as signature(set size_sign and not set size_key_exchange)
Is this expected behavior?
Thanks for your comment.

vletoux: Yes, exactly. Key exchange = sign + decryption. If it does not decrypt, it is a sign key.

hamano: @vletoux Thanks for your help.
I'll drop the patch.

hamano:
Now I confirmed that ca certificate are avalable in "Smart Card Trusted Roots"

frankmorgner:
OK, thanks

PR #991, opened 2017-03-11
Verify pubkey as spki

dengert:
Fix pkcs11/openssl.c to allow for the CKA_VALUE of a public key to be in spki format. 

Note that PKCS#11 does not define the CKA_VALUE for a public key, but OpenSC does, and it can be in either "direct" or "spki" format as defined in PKCS#15. "spki" is the same as on X.509 subjectPublicKeyInfo  and d2i_PUBKEY can be used to create a EVP_KEY. 

So either format is now supported by sc_pkcs11_verify_data.

The problem was found while testing pkcs11-tool -t -l  where the  verify tests would fail with a CKR_GENERAL_ERROR because  the card driver stored the public key as a spki.

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -391,7 +391,8 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
 {
 	int res;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	EVP_PKEY *pkey;
+	EVP_PKEY *pkey = NULL;
+	const unsigned char *pubkey_tmp;
 
 	if (mech == CKM_GOSTR3410)
 	{
@@ -405,7 +406,20 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
 #endif
 	}
 
-	pkey = d2i_PublicKey(EVP_PKEY_RSA, NULL, &pubkey, pubkey_len);
+	/*
+	 * PKCS#11 does not define CKA_VALUE for public keys, and different cards
+	 * return either the direct or spki versions as defined in PKCS#15
+	 * And we need to support more then just RSA.
+	 * We can use d2i_PUBKEY which works for SPKI and any key type. 
+	 */
+	pubkey_tmp = pubkey; /* pass in so pubkey pointer is not modified */
+	pkey = d2i_PUBKEY(NULL, &pubkey_tmp, pubkey_len);
+
+	/* if failed, try old way that only works for RSA */
+	if  (pkey == NULL) {
+		pubkey_tmp = pubkey;
+		pkey = d2i_PublicKey(EVP_PKEY_RSA, NULL, &pubkey_tmp, pubkey_len);
+	}
 	if (pkey == NULL)
 		return CKR_GENERAL_ERROR;
 
@@ -435,6 +449,7 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
 		 case CKM_RSA_X_509:
 		 	pad = RSA_NO_PADDING;
 		 	break;
+		/* TODO support more then RSA */
 		 default:
 			EVP_PKEY_free(pkey);
 		 	return CKR_ARGUMENTS_BAD;
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2884,30 +2884,44 @@ VARATTR_METHOD(GOSTR3410_PARAMS, unsigned char);
 VARATTR_METHOD(EC_POINT, unsigned char);
 VARATTR_METHOD(EC_PARAMS, unsigned char);
 
+
+/* called to check for CKA_ALWAYS_AUTHENTICATE and  CKU_CONTEXT_SPECIFIC login */
 static void  authenticate_if_required(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE privKeyObject){
 	CK_SESSION_INFO sessionInfo;
 	CK_TOKEN_INFO	info;
 	CK_RV rv;
 
+	get_token_info(opt_slot, &info);
 	rv = p11->C_GetSessionInfo(session, &sessionInfo);
 	if (rv != CKR_OK)
-		p11_fatal("C_OpenSession", rv);
+		p11_fatal("C_GetSessionInfo", rv);
 
 	switch(sessionInfo.state){
 		case CKS_RW_USER_FUNCTIONS:
-		   	//logged in, not need to continue.
-			return;
+		case CKS_RO_USER_FUNCTIONS:
+		case CKS_RW_SO_FUNCTIONS:
+			if (getALWAYS_AUTHENTICATE(session, privKeyObject)) {
+				login(session,CKU_CONTEXT_SPECIFIC);
+			}
+			break;
 		case CKS_RW_PUBLIC_SESSION:
+			/*
+			 * May be called if key does not require login
+			 * in which case it does not require CKU_CONTEXT_SPECIFIC
+			 * May be called if cached pin count has expired 
+			 * or session was set to CKS_RO_PUBLIC_SESSION
+			 * will assume a login at this point will
+			 * satisfy CKA_ALWAYS_AUTHENTICATE CKU_CONTEXT_SPECIFIC 
+			 */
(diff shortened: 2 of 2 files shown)


dengert:
commit https://github.com/OpenSC/OpenSC/pull/991/commits/cb748421ebc08a856c34790334acdd8e45103409 is independent of the other commits in this pull request, and should have been the only commit in the pull request.

src/pkcs11/openssl.c
```diff
@@ -405,7 +406,20 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
...
+	 * And we need to support more then just RSA.
+	 * We can use d2i_PUBKEY which works for SPKI and any key type. 
+	 */
+	pubkey_tmp = pubkey; /* pass in so pubkey pointer is not modified */
+	pkey = d2i_PUBKEY(NULL, &pubkey_tmp, pubkey_len);
+
+	/* if failed, try old way that only works for RSA */
+	if  (pkey == NULL) {
+		pubkey_tmp = pubkey;
+		pkey = d2i_PublicKey(EVP_PKEY_RSA, NULL, &pubkey_tmp, pubkey_len);
```
frankmorgner: Could you check in OpenSSL whether using the "old method" is needed? I would expect that OpenSSL does this internally already...

dengert:
(I used the wrong term in the commit text. "direct" should have been "raw")

It is not an OpenSSL issue, it is an PKCS#15 - OpenSC issue.  

PKCS#15 has:

> RSAPublicKeyChoice  ::= CHOICE {
       raw RSAPublicKey -- See PKCS \#1,
      spki [1] SubjectPublicKeyInfo, -- See X.509. Must contain a public RSA key
      ...
     }
     ECPublicKeyChoice ::= CHOICE {
         raw ECPoint,
       spki SubjectPublicKeyInfo, -- See X.509. Must contain a public EC key
       ...
    }

and others like DH too.  These are stored in:

> pkcs15.h: sc_pkcs15_pubkey_info
      struct {
           struct sc_pkcs15_der raw;
          struct sc_pkcs15_der spki;
     } direct;

./libopensc/pkcs15-pubkey.c does a lot of checking to decide if it can return the raw or spki versions as a CKA_VALUE for a public key. 

 I tested on OpenSSL-1.1, 1.0.2 and compiled on 1.0.1 and libressl-2.5.0
The have the d2i_PUBKEY . It for use with SPKI. I would not expect to do anything related to PKCS#15 raw.

frankmorgner:
My read is that the functionality of both functions is to convert an ASN.1 structure into a `EVP_PKEY` object. According to a mailing list post, [only d2i_PublicKey should be used in new code](http://openssl.6102.n7.nabble.com/difference-between-i2d-PUBKEY-and-i2d-PublicKey-td43869.html); `i2d_PUBKEY` mainly exists for backward compatibility.

Do you see some limitation that forces you to still call `i2d_PUBKEY`? If not, you can remove the additional call to `i2d_PUBKEY` in the error case.

dengert:
That article is talking about i2d_PublicKey and i2d_PUBLICKEY. It also says: "The companion decoder d2i_PublicKey can't determine the key type from the structure so needs to be explicitly told." 

d2i_PublicKey needs to be told what the algorithm is and its data depends on the algorithm.  

d2i_PUBLICKEY is expecting an SPKI, that contains algorithm, parameters and public key string. 
It works for RSA, EC, or any other algorithm OpenSSL supports, and the SPKI contains all 
the information needed to create an EVP_PEY for any supported algorithm.  

PKCS#11 specs do NOT define a CKA_VALUE for a CKO_PUBLIC_KEY. For the reasons of what would it return? Instead each keytype can return algorithm specific attributes. 

OpenSC made the mistake of defining a CKA_VALUE for a CKO_PUBLIC_KEY.
When it was RSA only this was not a problem. It is today, with multiple algorithms. 
 
Without the patch only this is supported. 
`pkey =d2i_PublicKey(EVP_PKEY_RSA ,...)`
(So without some change to sc_pkcs11_verify_data it will never support EC.)

When EC was added to OpenSC, if the PKC#15 had an SPKI, or an emulator added it,
OpenSC would save this DER string as direct-spki. The code to create the CKA_VALUE 
for a CKO_PULIC_KEY could return either an SPKI if available or a public key without
the algorithm or parameters. 

Since sc_pkcs11_verify_data is a consumer of the CKA_VALUE, it has to try both
d2i_PublicKey(EVP_PKEY_RSA and d2i_PUBLICKEY to create an EVP_KEY.
This is the what the patch is doing.

The real fix in sc_pkcs11_verify_data openssl.c would be to use 
sc_pkcs15_encode_pubkey_as_spki followed d2i_PUBLICKEY. But to call it would require 
adding framework code to avoid using the CKA_VALUE attribute. 

And then remove the CKA_VALUE for a CKO_PUBLIC_KEY. But there may be backward
compatibility issues if this is removed. Some utilities may also expect it and 
expect it to return what was in the original PKCS#15, which is what it does.
Either the direct.spki, the direct.raw or some other form.

frankmorgner:
Your comment says:

>We can use d2i_PUBKEY which works for SPKI and any key type

This is also how I read the conversation on the OpenSSL mailing list. And it makes me think that `d2i_PUBLICKEY` also works for a RSA keys. Hence, I believe that we don't need to call `d2i_PublicKey` in case of an error.

Could you please squash the fix from https://github.com/OpenSC/OpenSC/pull/991/commits/23278cd42dc9ed71432ef0a098548a9ae8dc2883 into the erroneous changes from https://github.com/OpenSC/OpenSC/pull/991/commits/9b10df8a270ca0d419bf6a6b76554934478b070d?

dengert:
I will look at late next week.

frankmorgner:
thank you!

frankmorgner:
Looking at the OpenSSL implementation, `d2i_PUBKEY` works as folllows (`x_pubkey.c`):

1. parse `X509_PUBKEY` ASN.1 structure
2. construct `EVP_PKEY` object (`X509_PUBKEY_get`) by
2.1 Setting the key type according to the ASN.1's OID
2.2 Calling the key type's `pub_decode` (which maps to `rsa_pub_decode`/`d2i_RSAPublicKey` in case of RSA)

`d2i_PublicKey` does the following (`d2i_pu.c`):

1. set the key type according to the input parameter
2. call the key type's decoding routine directly (`d2i_RSAPublicKey` in case of RSA)

Hence, my assumption above is correct: `d2i_PUBKEY` supersedes the functionality of `d2i_PublicKey` and that additional call in your code should be removed.

dengert:
Disagree. The problem is what ASN1 is presented. is it just the RSA modulus
and exponent, which d2i_PublicKey is expecting with the RSA key type, or is
the input a SPKI that contains the algorithm OID, parameters and pub key.

Problem is opensc will create a CKA_VALUE and this code uses it even though
PKCS#11 says a public key does not have  CKA_VALUE.

The CKA_VALUE is based on  PKCS#15 direct or spki.

we may still have drivers that do not use the or store a spki.

I am still on vacation, will look closer next week.

pkcs11-tool -O should show the "public key" based on the CKA_VALUE, I can
not test till next week.

frankmorgner:
OK, I think I finally understood the actual problem though your quotation of PKCS#15 makes it quite clear(sorry for the delay). Though I think you changes are good to be merged, I'll wait for your feedback from next week.

frankmorgner:
If I'm reading the final changes of `authenticate_if_required()` correctly, then you are not changing the actual logic, right? If so, please remove your changes from this PR.

dengert:
Yes, I stuck the commit on the wrong branch. I will fix it next week and
look at a better solution I described above:

The real fix in sc_pkcs11_verify_data openssl.c would be to use
sc_pkcs15_encode_pubkey_as_spki
followed d2i_PUBLICKEY. But to call it would require adding framework code
to avoid using the CKA_VALUE attribute.

Or use some other opensc routine (don't remember the name)  that returns a
EVP_KEY. That wold need minor routine in the the framework too.

frankmorgner:
@dengert did you have time to look at this, again?

dengert:
Not yet. Was gone longer the expected, and came back to a computer zapped by lighting. Hope to get to it in the next few days.

frankmorgner:
I understand, take your time. I hope nothing critical got lost...

dengert:
I am closing this pull request in favor of #1018

PR #1017, opened 2017-04-07
Improved key file creation with MyEID card to allow creating private keys under specific PINs. A fix to sc_pkcs15_verify_pin when called with empty pin data.

hhonkanen:
The changes have been tested with a MyEID card. Myeid_create_key has been tested by creating a card with four pins. A key was generated under pin 1 and another key under pin 4. After creation, verified that the correct security attributes were set into both keys.

The change in pkcs15-pin.c was made to fix a problem, which appears when creating objects using C_CreateObject, when pin caching is disabed. With a normal smart card reader, C_CreateObject would fail, because sc_pkcs15_verify_pin was called with empty PIN data, even if C_Login had been called and the required pin had been verified. With a pinpad reader, the user was prompted for pin multiple times. The code has been updated to check the pin verification state if this is possible, and return success if the access condition is already open.

### DIFF
--- a/src/libopensc/pkcs15-pin.c
+++ b/src/libopensc/pkcs15-pin.c
@@ -293,15 +293,31 @@ sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *pi
 		const unsigned char *pincode, size_t pinlen)
 {
 	struct sc_context *ctx = p15card->card->ctx;
+	struct sc_pkcs15_auth_info *auth_info = (struct sc_pkcs15_auth_info *)pin_obj->data;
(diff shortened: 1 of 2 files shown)


frankmorgner:
The code looks OK. Can you get confirmation of some other users/maintainers of this card?

hhonkanen:
@viktorTarasov: Please could you take a look at this PR as you have experience with MyEID?
@larssilven: You tested my changes into pkcs15-pin.c couple of months ago with a pinpad reader. Please could you confirm in this thread that it worked as expected?

larssilven:
Yes it worked as expected.

PR #1018, opened 2017-04-07
PKCS#11 does not define a CKA_VALUE for public keys and is missused

dengert:
This is a replacement for #991. It tests if the CKA_VALUE is a spki before calling d2i_PUBKEY. The previous pull request referred to "**direct**" and "spki", but should have used "**raw**" and "spki"  

OpenSC openssl.c in sc_pkcs11_verify_data assumes that it can
retieve the CKA_VALUE for a public key object, and expect it to
be usable as RSA.

But internally sc_pkcs15_pubkey can have a "raw" or "spki"
version of the public key as defined by PKCS#15.  Card drivers
or pkcs15-<card> routines may store either the "raw" or "spki"
versions. A get attribute request for CKA_VALUE for a public key
will return either the raw, spki or will derived rsa verison of the
pubkey.

This commit will test if the CKA_VALUE is a spki and use d2i_PUBKEY
which takes a spki version and returns an EVP_KEY. If it not an spki
the current method, d21_PublicKey(EVP_PKEY_RSA,...) is used which
only works for RSA.

The problem was found while testing pkcs11-tool -t -l  where
the  verify tests would fail with a CKR_GENERAL_ERROR because
the card driver stored the public key as a spki.

On branch verify-pubkey-as-spki-2
 Changes to be committed:

Date:      Fri Apr 07 07:50:00 2017 -0600

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -405,7 +405,9 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
 {
 	int res;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	EVP_PKEY *pkey;
+	EVP_PKEY *pkey = NULL;
+	const unsigned char *pubkey_tmp;
+	int is_spki = 0;
 
 	if (mech == CKM_GOSTR3410)
 	{
@@ -419,7 +421,38 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
 #endif
 	}
 
-	pkey = d2i_PublicKey(EVP_PKEY_RSA, NULL, &pubkey, pubkey_len);
+	/*
+	 * PKCS#11 does not define CKA_VALUE for public keys, and different cards
+	 * return either the raw or spki versions as defined in PKCS#15
+	 * And we need to support more then just RSA.
+	 * We can use d2i_PUBKEY which works for SPKI and any key type. 
+	 */
+	pubkey_tmp = pubkey; /* pass in so pubkey pointer is not modified */
+	/* SPKI ASN.1 starts with SEQUENCE, SEQUENCE, OBJECT IDENTIFIER */
+	if (*pubkey_tmp++ == 0x30) {
+		if (*pubkey_tmp & 0x80)
+			pubkey_tmp += *pubkey_tmp & 0x7F;
+
+		pubkey_tmp++;
+		if (*pubkey_tmp++ == 0x30) {
+			if (*pubkey_tmp & 0x80)
+				pubkey_tmp += *pubkey_tmp & 0x7F;
+
+			pubkey_tmp += 1;
+			if (*pubkey_tmp == 0x06)
+				is_spki = 1;
+		}
+	}
+
+	pubkey_tmp = pubkey; /* pass in so pubkey pointer is not modified */
+
+	if (is_spki) {
+		pkey = d2i_PUBKEY(NULL, &pubkey_tmp, pubkey_len);
+	}else {
+		/* Assume it is RSA */
+		/* TODO support others  but GOST is done above, and EC always uses SPKI */
+		pkey = d2i_PublicKey(EVP_PKEY_RSA, NULL, &pubkey_tmp, pubkey_len);
+	}
 	if (pkey == NULL)
 		return CKR_GENERAL_ERROR;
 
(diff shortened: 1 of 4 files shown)


src/pkcs11/openssl.c
```diff
@@ -419,7 +421,38 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
...
+		pubkey_tmp++;
+		if (*pubkey_tmp++ == 0x30) {
+			if (*pubkey_tmp & 0x80)
+				pubkey_tmp += *pubkey_tmp & 0x7F;
+
+			pubkey_tmp += 1;
+			if (*pubkey_tmp == 0x06)
+				is_spki = 1;
+		}
+	}
```
frankmorgner: I like the Idea of determining the type beforehand. However, you should at least use the ASN.1 parsing routines of OpenSC. Even better, you should check whether you can reuse the [existing code that parses this kind of data](https://github.com/OpenSC/OpenSC/blob/e7915ec198a4f2861141157b8fcd6e4e1322eaff/src/libopensc/pkcs15-cert.c#L62)

src/pkcs11/openssl.c
```diff
@@ -419,7 +421,38 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, int pubkey_len,
...
+
+	pubkey_tmp = pubkey; /* pass in so pubkey pointer is not modified */
+
+	if (is_spki) {
+		pkey = d2i_PUBKEY(NULL, &pubkey_tmp, pubkey_len);
+	}else {
+		/* Assume it is RSA */
+		/* TODO support others  but GOST is done above, and EC always uses SPKI */
```
frankmorgner: PKCS#15 forces RSA if it's not spki, so the above comment can be removed.

dengert:
OK, I will withdraw this PR too.

The real problem is sc_pkcs11_verify_final in pkcs11./mechanism.c is trying to use the CKA_VALUE of a public key. **PKCS#11 does **NOT** define a CKA_VALUE for a CKO_PUBLIC_KEY.** Each of the CKK_* key types have their own set of key specific attributes. OpenSC tries to define a CKA_VALUE, which can end up as either the "raw" or "spki" as from PKCS#15, but there is no way to tell them apart other then to pars them. 

The code in pkcs15_pubkey_get_attribute in framework-pkcs15.c that tries to return something for a CKA_VALUE. It should return only a spki or should be removed or, at least, not used internally in  sc_pkcs11_verify_final.  Are there any requirements to support a CKA_VALUE of a CKO_PUBLIC_KEY?

git blame framwork-pkcs15.c shows in lines 3978 - 4020 that @viktorTarasov and @frankmorgner  in 2014 introduced the ambiguous code that would return the CKA_VALUE as  a raw or spki for any CKO_PUBLIC_KEY. 

I also don't agree with your statement:
"PKCS#15 forces RSA if it's not spki, so the above comment can be removed." 
Each of the PKCS#15 ASN.1 define what "raw" can be based on the key type:

RSAPublicKeyChoice ::= CHOICE {
raw RSAPublicKey -- See PKCS #1,
spki [1] SubjectPublicKeyInfo, -- See X.509. Must contain a public RSA key
...
}

ECPublicKeyChoice ::= CHOICE {
raw ECPoint,
spki SubjectPublicKeyInfo, -- See X.509. Must contain a public EC key
...
}

DHPublicKeyChoice ::= CHOICE {
raw DiffieHellmanPublicNumber,
spki SubjectPublicKeyInfo, -- See X.509. Must contain a public D-H key
...
}

DSAPublicKeyChoice ::= CHOICE {
raw INTEGER,
spki SubjectPublicKeyInfo, -- See X.509. Must contain a public DSA key.
...
}

KEAPublicKeyChoice ::= CHOICE {
raw INTEGER,
spki SubjectPublicKeyInfo, -- See X.509. Must contain a public KEA key
...
}

frankmorgner:
No need to withdraw the PR, simply push more changes on top. We'll do the squashing later.

Sorry, I only looked at `RSAPublicKeyChoice`. You're right that RSA is not being forced if the type is not spki.

Regarding the commits, I think you meant dd5115bc0 followed by 808fff224, which I was not concerned with. Anyway, for whatever reason you need a workaround, I'd like to make sure that it's not only a quick hack. If possible, I'd like to see a *real fix*, but currently I don't have the insights how that may look like.

dengert:
https://github.com/OpenSC/OpenSC/pull/1018/commits/5c5a5405d8a67ef2ad9909b683979b74c1a839de and https://github.com/OpenSC/OpenSC/pull/1018/commits/fcef1ba07157c2dc9c15df6866bc4d2096766312
 now adds a CKA_SPKI to be used internally in place of the CKA_VALUE. CKA_SPKI returns an SPKI. 

I have tested this with pkcs11-tool -t -l.  It needs to be tested with older cards that use the "raw" format of the PKCS#15 CHOICE. It also needs to be tested with GOST. 

This should also allow for ECDSA signatures to be verified. 

The GOST code was treated special in previous code and that was not touched.

frankmorgner:
I'll leave this open for an other week to see if someone with a *raw* public key encoding can test this...

PR #1019, opened 2017-04-11
Secret key preparation work

fabled:
The first two commits are already pending merge in pull request #919.

Please look at the the final commit which migrates sc_pkcs15_skey_info to be sc_pkcs15_prkey_info instead. This is preparatory work for better secret key support in OpenSC. Much of the skey operations are similar to prkey ones (generate, upload, etc.) and the further works are greatly simplified if they share the same backing data type. Alternatively, the common parts could be merged to 'struct pkcs15_common_key_info' or similar. However, I would prefer this approach.


fabled:
@viktorTarasov Please review.

I also sent earlier heads up on this approach at https://sourceforge.net/p/opensc/mailman/opensc-devel/thread/20170405150701.515204ec%40vostro.util.wtbts.net/#msg35770186

dengert:
Keep in mind while making these changes that  C_DeriveKey and C_UnwrapKey produce a key object.  The OpenSC ECDH code returns a generic secret key today and sets in `sc_pkcs15_ skey_info`:
`unsigned long key_type;`

It attempts to provide support for the derived key to be on the token:
 `sc_pkcs15_skey_info`: `struct sc_path path; /* if on card */`

By  `/* convert to union if other types are supported */` in sc_pkcs15_prkey, should an indicator as to what is in the union be added?  For example add `unsigned long key_type;` I have not looked at this closely, but there may be code that is testing the modulus_length or field_length to determine if this is RSA or EC. This would eliminate (or allow for testing)  mistaking a secret key for a private key.

In regard to the CKA_ID. PKCS#11 says "In the case of secret keys, the meaning of the CKA_ID attribute is up to the application." 

I would nice if  you used a  `struct pkcs15_common_key_info` as `sc_pkcs15_pubkey_info` also has the  `/* convert to union if other types are supported */` But that would require a lot more code changes. 

You may also need to look closer at all the GOST code, that these changes may impact.

fabled:
@dengert Thanks for the comments.

Yes, I am aware of the PKCS#11 usage. This commit was just mechanical merge of the structs. And primary motivation for this pull request is to get direction on how to do this. If using some common_key_info struct usage is preferred, I would like to get clear direction on this, because much of the code will depend on this core change.

Each _info is linked to from sc_pkcs15_object. The union's usage is based on the sc_pkcs15_object's type attribute (SC_PKCS15_TYPE_* enum). To my understanding all existing usage of field_length and modulus_size is already based on this. But will double check this, if this proposed solution is acceptable.

dengert:
I will leave the choice of `struct pkcs15_common_key_info` up to others. 

In `sc_pkcs15_prkey_info` in the union you have: `unsigned long key_type; /* PKCS#11 CKK_* */`  I would suggest that the `unsigned long key_type; /* PKCS#11 CKK_* */` be moved outside the union to identify what is in the union.  Because this is a PKCS15 structure and used by the minidriver or other application that do not use PKCS11,  it should not use the PKCS11 flags, but rather the `SC_PKCS15_TYPE_*` (or subset) which is the same as in the `sc_pkcs15_object`:`unsigned int type`. The `sc_pkcs15_free_prkey_info` is one routine that may need to know (maybe in the future)  what is in the union as it is only passed the `sc_pkcs15_prkey_info`.

fabled:
@dengert The simplest way to add `sc_pkcs15_key_info` would be to add `void *common_info` to `sc_pkcs15_object` or union with different common info struct pointers in it. This might allow similar harmonization with some other object types too. However, the mechanical diff would be rather large as many of the fields from currently separate structs would be moved to the common struct.

Additionally the `sc_pkcs15_prkey_info.key_type; /*PKCS#11 CKK_* */` is intentional. While it is currently used in the PKCS#11 code only, it will be useful for the generic PKCS#15 code too. ISO/IEC 7816-15 states that the CIO-ALGORITHM.&id should be the CKK_* value from PKCS#11. It would allow to deduce the key type for `SC_PKCS15_TYPE_SKEY_GENERIC` type. There is similar generic key types specified for public and private key objects too in ISO 7816-15. So in fact, it makes sense to move the key_type outside of the union.

dengert:
sc_pkcs15_prkey_info.key_type; /*PKCS#11 CKK_* */ was added in: @9c5dbea8838904cfa700d502e4ecbf8d834418e8  This was used internally in the framework-pkcs11.c for ECDH derived secret keys, and only the generic secret key is supported. 
 
But I think if it is stored in a PKCS15 structure, it should not require the used of PKCS11. 
The purpose of the framework-pkcs15.c is to isolate PKCS11 from PKCS15. 

pkcs15_prkey_get_attribute already does mapping for prkeys to get CKA_KEY_TYPE. Line 3434.
pkcs15_skey_get_attribute should be changed to  use the SC_PKCS15_TYPE_* in line 4303. 
When I wrote that code, it was only used within PKCS11. Now that you are going to add proper support for secret keys, line 4303 should be changed and use the SC_PKCS15_TYPE_*

fabled:
This concerns the other relevant question in my email to the opensc-devel list. In ISO/IEC 7816-15 the encoding for e.g. AES key happens using the algIndependentKey encoding (PKCS#15's genericSecretKey). The actual key type is indicated via CommonKeyAttributes.algReference to CIAInfo.supportedAlgorithms that describes the algorithm. That is also the structure where PKCS#15 requests a PKCS#11 algorithm id enumeration to be stored. Granted, we can internally map that and use the `struct sc_algorithm_id` instead. Perhaps the PKCS#11 module can be updated to use that, or move the key_type to be member of internal PKCS#11 structures.

Currently I'm thinking that the SC_PKCS15_TYPE_* should map directly to the ASN.1 type used, and thus SC_PKCS15_TYPE_SKEY_GENERIC (similar ASN1 choice exists for prkey and pubkey) could use an additional key type field. When loading PKCS#15 data structures it would be initialized from the supportedAlgorithm's AlgorithmInfo structure, and serve as cached value to identify the key type. This is another area where public key / private key / secret key code can be shared. Thus merging the structures, or defining "common key attributes" is useful -- and should probably be extended to public keys too. Looking at the spec, it seems similar shared attributes amongs different major key types exists only for public/private/secret keys. Would make more sense to have a `struct sc_pkcs15_common_key_info *` in `sc_pkcs15_object` OR to merge all three key info structures to be one and make the differing parts an union? Thoughts?

dengert:
So far no one else has commented on this  PR. 

I don't deal with PKCS#15 cards, but use OpenSC via the emulation capability.
That said...

OpenSC as far as I can tell is based on PKCS#15 v 1.1, although it does  ISO/IEC 7816-15 parse CommonKeyAttributes.algReference. Not clear if it uses this or not.  This is not to say that it should not be updated to 7816-15, but that sounds like a much bigger project then to implement  SecretKey support. 

SC_PKCS15_TYPE_* are to be used in the `struct sc_pkcs15_object`  `unsigned int type`
secret keys have these defined:
#define SC_PKCS15_TYPE_SKEY                     0x300
#define SC_PKCS15_TYPE_SKEY_GENERIC             0x301
#define SC_PKCS15_TYPE_SKEY_DES                 0x302
#define SC_PKCS15_TYPE_SKEY_2DES                0x303
#define SC_PKCS15_TYPE_SKEY_3DES                0x304

So if you want this to match PKCS#11 CKK_* that will not work.

Changing the SC_PKCS15_TYPE_* are a **big** change and much bigger then just adding secret key support.
You could define some equivalent defines to the CKK_*. My concern is dragging PKCS11 in to code that  does not use PKCS11. That is the function of the framework-pkcs15. 

Be careful to not try and rewrite all of OpenSC and do it under a title that sounds very limited. People have tried to do this in the pass and it leads to no review of the code until they have extended much effort, just to find out that the PR will not be accepted.  

Others developers need to comment on the PR.

fabled:
Yes, this issue exactly to get feedback from as many developers as possible.

No, there is no intention to match SC_PKCS15_TYPE_* with CKK_*; the point is that SC_PKCS15_TYPE_ may not fully define the key type but be a generic one, so it would make sense to have another field for the exact key type. SC_PKCS15_TYPE_* matches the ASN.1 presentation. But PKCS#15 does refer to PKCS#11 on spec level, mostly to use some of the enumerations on the PKCS#15 structures.

ISO/IEC 7816-15 basically supercedes PKCS#15. And OpenSC already supports some of the extensions made there as you noted. For the secret key part some of the information relies on that. The original PKCS#15 wanted to use the CHOICE type to define the key algorithm, and define per-algorithm data. However, in ISO/IEC this was abstracted out by means of generic CHOICE types and having a OID binding, or algorithm reference to the card info structure to define the key. The ISO way is more generic and extensible on ASN1 level...

While I do have a desire to rewrite parts of the code, I'm trying to minimize it. The motivation is to make the codebase be more object-oriented and have more code shared between the different key object types. This will mean less code and less bugs. I also want to do this incrementally in as small commits as possible. Which is exactly why I posted this pull request early before progressing to the secret key implementation bits.

Thanks.

fabled:
If make changing to the abstraction, or data structures is deemed to take more discussion. The other option is to just duplicate the code from prkey and copy it with relevant parts as skey one. Would this be the preferable approach to do the initial implementation? And then leave the abstraction fixup and merging of data structures as potential later work. Would this be less controversial approach?

frankmorgner:
In tha past, the paradigm was to always add new structures, also if that implied duplicating structures or data.

I see the benefit of re-using data structures. But if changing the paradigm to apply unions, it should be applied to all such structures. *But* as @dengert stated above, such a big change needs additional consent from @viktorTarasov (or some other project owner than me).

fabled:
I will do the initial implementation by duplicating code where necessary. Let's see how to proceed from that. I'll close this for now then.

PR #1021, opened 2017-04-12
GetTokenInfo does not release the acquired lock and its return value is not properly checked in pkcs11-tool

Jakuje:
`C_GetTokenInfo()` does not release the acquired lock and its return value is not properly checked in `pkcs11-tool`. This makes NSS `modutil` choke because it is using a global lock for PKCS#11 operations.

This is result of testing with Athena ASEPCOS cards, which behaves somehow somehow weird with uninitialized card:
```
5: C_GetSlotInfo
2017-04-12 11:19:41.825
[in] slotID = 0x0
[out] pInfo: 
      slotDescription:        'OMNIKEY AG CardMan 3021 00 00   '
                              '                                '
      manufacturerID:         'OMNIKEY AG                      '
      hardwareVersion:         3.2
      firmwareVersion:         0.0
      flags:                   7
        CKF_TOKEN_PRESENT                
        CKF_REMOVABLE_DEVICE             
        CKF_HW_SLOT                      
Returned:  0 CKR_OK

6: C_GetTokenInfo
2017-04-12 11:19:41.920
[in] slotID = 0x0
Returned:  224 CKR_TOKEN_NOT_PRESENT
```

Longer write up in the following bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1376090

The following investigation would be to find out why the `CKF_TOKEN_PRESENT` flag is present when the card detection failed. There is the whole log:

https://gist.github.com/Jakuje/64dc7df1d8e11d74c8d756aa20a6b5fd

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -487,12 +487,16 @@ CK_RV C_GetTokenInfo(CK_SLOT_ID slotID, CK_TOKEN_INFO_PTR pInfo)
 		goto out;
 	}
 
-	if (slot->p11card == NULL)
-		return CKR_TOKEN_NOT_PRESENT;
+	if (slot->p11card == NULL) {
+		rv = CKR_TOKEN_NOT_PRESENT;
+		goto out;
+	}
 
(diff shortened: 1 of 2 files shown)


frankmorgner:
thanks

PR #1022, opened 2017-04-13
pkcs11-tool: Set CKA_PRIVATE=false for new public keys

Jakuje:
Keygen should write public keys with explicit CKA_PRIVATE=false by default (possibility to modify by --private switch)

Related to 4df35b92 discussing writing separate objects years ago.


frankmorgner:
thanks

PR #1023, opened 2017-04-13
fixed more coverity issues

frankmorgner:


src/libopensc/iso7816.c
```diff
@@ -397,7 +397,8 @@ iso7816_process_fci(struct sc_card *card, struct sc_file *file,
 			sc_log(ctx, "  EF structure: %d", byte & 0x07);
 			sc_log(ctx, "  tag 0x82: 0x%02x", byte);
 			/* FIXME: check return value? */
```
Jakuje: the comment is probably no longer needed

frankmorgner: done, thanks.

PR #1025, opened 2017-04-14
Secret key support

fabled:
This implements better Secret key support in OpenSC, and so far allows uploading secret keys on MyEID cards. Many of the first commits are ready for review and merging, but some comments have FIXME notes in the commit message. Consider this as requesting feedback on how to fix those issues properly.


fabled:
@frankmorgner @viktorTarasov @dengert Care to take a look at this?

The CI fails due to the overloaded use store_key() in the last three commits, and it's marked on them with a FIXME. I would be interested to know if it'd be more appropriate to make a separate store_skey() or just change the current store_key()'s last argument to be void* or union that can contain either of the key types presentation.

dengert:
Should _sc_card_add_generic be called something like _sc_card_add_secret? 
"generic" is a type of secret key, but you are adding multiple types of secret keys. 

You use the term secrkey in a few places and parameters. Would secretkey be a better name?

fabled:
@dengert We could rename it to _sc_card_add_secret but it also accepts non-secret key algorithms. The point was to add an utility function that does not require any algorithm specific extra data.

For `secrkey` the commit explains: that user visible string is exposed in pkcs11-tool, and I used the same string in pkcs15-init to keep it harmonious. I have no strong opinion on what it should be.

frankmorgner:
looks OK at first glance. Is there someone who can verify this functionality with myeid?

If you want to use `_sc_card_add_generic`, please do so only within `card.c`, but expose algorith specific functions (e.g. `_sc_card_add_aes_alg`) in `card.h`.

Regarding `store_key`, I'd suggest to use `void *` as parameter and to use the function for both, secret and private key import.

fabled:
I modified sc_pkcs15_prkey to be able to contain an sc_pkcs15_skey. This allows existing store_key abstraction to be used nicely and type safely. If casting to void* is preferred, I can do that instead, but the cruft for changing all pkcs15init card drivers to accommodate that is rather large.

I'd rather not add separate `_add_aes_alg` and `_add_des_alg` functions that do exactly the same job. That's why I called it generic instead, because it can and will be used with multiple different symmetric cipher algorithms.

dengert:
There is one part of OpenSC that depends on secret keys today, ECDH derive. With all the proposed changes to support secret key, do you have a card with EC keys to test ECDH?

It can be tested using pkcs11-tool if you have two cards each with a different EC key. Giving each card the public key of the other, both cards will derive the same secret key. Today the derived secret key is generic, and stored in memory. Part of your changes could also be to allow this key to be stored on the card. 

I would hope you would test to make sure any of the proposed changes do not break the ECDH derive code.

fabled:
> With all the proposed changes to support secret key, do you have a card with EC keys to test ECDH?
Yes.

> I would hope you would test to make sure any of the proposed changes do not break the ECDH derive code.
That code is isolated currently in pkcs11 module. It is not touched at all (except one mechanical variable type change), and it handles the ECDH derive in isolated way - the pkcs15 code is not involved in it. If and when I get to adding support for supporting on-board secret keys via pkcs11, this needs more careful attention. But I am aware of how it works, and have hardware to test it with.

mouse07410:
I've tested this PR with Yubikey NEO (EC keys over P256). It does break ECDH key derivation:
```
$ openssl pkeyutl -engine pkcs11 -keyform engine -derive -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20key;type=private" -peerform PEM -peerkey ec256-pub.pem -hexdump
engine "pkcs11" set.
Enter PKCS#11 token PIN for PIV Card Holder pin (PIV_II):
Public Key operation error
140736553579528:error:8002C063:Vendor defined:PKCS11_ecdh_derive:bad key parameters format:p11_ec.c:517:
$ 
```
Here's the log:
[1025-bad.log.txt](https://github.com/OpenSC/OpenSC/files/926015/1025-bad.log.txt)

Here's the screen output of OpenSC without this PR:
```
$ openssl pkeyutl -engine pkcs11 -keyform engine -derive -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20key;type=private" -peerform PEM -peerkey ec256-pub.pem -hexdump
engine "pkcs11" set.
Enter PKCS#11 token PIN for PIV Card Holder pin (PIV_II):
0000 - e8 f7 a9 7d fe c9 8f a8-f2 c0 e1 a9 e0 06 f2 7b   ...}...........{
0010 - b5 d6 50 bc 4a 80 ca 03-30 f2 e0 38 5c f5 55 cd   ..P.J...0..8\.U.
$ openssl pkeyutl -engine pkcs11 -derive -keyform PEM -inkey ec256-priv.pem -peerform engine -peerkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20pubkey;type=public" -hexdump
engine "pkcs11" set.
0000 - e8 f7 a9 7d fe c9 8f a8-f2 c0 e1 a9 e0 06 f2 7b   ...}...........{
0010 - b5 d6 50 bc 4a 80 ca 03-30 f2 e0 38 5c f5 55 cd   ..P.J...0..8\.U.
$
```
And the good log:
[pre-1025-good.log.txt](https://github.com/OpenSC/OpenSC/files/926025/pre-1025-good.log.txt)

dengert:
In https://github.com/OpenSC/OpenSC/files/926015/1025-bad.log.txt  Problem appears to be in after Line:11591 the session is then closed.
 0x7fffc84873c0 12:07:30.450 [opensc-pkcs11] framework-pkcs15.c:3336:pkcs15_prkey_get_attribute: pkcs15_prkey_get_attribute() called
0x7fffc84873c0 12:07:30.450 [opensc-pkcs11] pkcs11-session.c:158:C_CloseSession: C_CloseSession(0x7fb1a65125b0)

In good line:11259

0x7fffc84873c0 12:12:47.231 [opensc-pkcs11] framework-pkcs15.c:3316:pkcs15_prkey_get_attribute: pkcs15_prkey_get_attribute() called
0x7fffc84873c0 12:12:47.231 [opensc-pkcs11] framework-pkcs15.c:3316:pkcs15_prkey_get_attribute: pkcs15_prkey_get_attribute() called

Then when the secret key in memory is being created:
0x7fffc84873c0 12:12:47.231 [opensc-pkcs11] pkcs11-object.c:99:sc_create_object_int: called

It looks like  in framework-pkcs15.c, this is failing:


2063 /* TODO Only Session secret key objects are supported for now
2064  * Sesison objects have CKA_TOKEN=false
2065  * This is used by the C_DeriveKey with ECDH to hold the
2066  * key, and the calling application can then retrieve tha attributes as needed.
2067  * TODO If a card can support  secret key objects on the card, this
2068  * code will need to be expanded.
2069  */
2070 static CK_RV
2071 pkcs15_create_secret_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile,
2072                 CK_ATTRIBUTE_PTR pTemplate, CK_ULONG ulCount,
2073                 CK_OBJECT_HANDLE_PTR phObject)
2074 {

@fabled  Have you looked at  pkcs15_create_secret_key?

mouse07410:
> ...Many of the first commits are ready for review and merging...

I don't think so.

> This implements better Secret key support in OpenSC...

It broke ECDH, which I consider pretty fundamental/important for all the EC-capable cards, not just MyEID.

fabled:
@mouse07410
> Here's the log: 1025-bad.log.txt
> And the good log: pre-1025-good.log.txt

The logs indicate that there was other changes in addition to this change PR (many of the logs line numbers that should match, do not match). Could describe the exact versions you tested? Did you do a full install of opensc? If the OpenSC pkcs11 module and the pkcs15 library in system do not match, things will break.

If you could bisect the offending commit, it would be even more helpful. I am suspicious because a) I am not touching PKCS#11 code that would break this, and b) I tested the PKCS#11 module with MyEID and EC Derive works (see below).

@dengert
> Have you looked at pkcs15_create_secret_key [of PKCS#11]?

Yes. I am not touching it. This PR implements only PKCS#15 side improvements to maintain SKDF and to  upload secret keys. It is potential future work to improve PKCS#11 module to support onboard secret key objects. The only change to PKCS#11 module was a mechanical change in commit e57b581ee567bdabdbabce9169da38b99221d1bc. Additionally PKCS#11 module does not call to PCKS#15 code for the stub secret key object.

@mouse07410 
> ...Many of the first commits are ready for review and merging...
> I don't think so.

Why? I think they are ready for review. Yes, if something is broken, it needs to be merged before the PR in entirety is merged. However, if one late commit has a problem does not mean the first ones are bad.

> This implements better Secret key support in OpenSC...
> It broke ECDH, which I consider pretty fundamental/important for all the EC-capable cards, not just MyEID.

I am not fully convinced yet that this breaks it. You compare is not clean, but includes other changes too. When I tried to reproduce the issue with pkcs11-tool, MyEID EC derive works just fine even with the PR applied:
```
$ ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so -vvv  --derive --login --id "fa151a1502df69549579f5bc3ca753786eabaaf1" --input-file ec-pub.der
Logging in to "MyEID".
Please enter User PIN: 
Using derive algorithm 0x00001051 ECDH1-COFACTOR-DERIVE
�s��oೲVE��8g${Ã�p��Ֆ�C�po��kg�H(�cr@x�cA~��ޘ��R��
```

> P.S. On a separate topic, it looks like the only thing CI tests is whether the submitted code compiles. While it is important - we're missing a crucial part: does this code actually work? I propose addition of self-tests that would use SoftHSMv2 as a software-based token emulator, to actually test whether the command-line tools do what they're supposed to. Without that we're risking catching serious problems too late down the pipe. I realize the risks of relying on somebody else's package for testing - but the current state is hardly better.

Yes, agreed. Test suite could be improved. That is worth a separate issue / PR.

dengert:
I have tested this pull request on Ubuntu using NIST demo card 4 and 15 with commands like this where the /tmp/derive.2580.other.pubkey.der is the pubkey of the other card:

`pkcs11-tool --slot 0 --module /opt/ossl-1.0.2/lib/opensc-pkcs11.so -l --derive
 -m ECDH1-COFACTOR-DERIVE -O -d 03 -i /tmp/derive.2580.other.pubkey.der
` 

The derive operations work and the two cards produce the same generic secret (Anyone else using the NIST demo cards 4 aND 15 should get the same result.):
Secret Key Object; Generic secret length 32
  VALUE:      4a13df485f9c4c132c94cd0f1ef3a37b0316729e0b288cfe55a7d691eae754d1
  label:      �
  Usage:      encrypt, decrypt, wrap, unwrap

@mouse07410     I don't see where the code has broken ECDH. There may be something else going on in your environment. As I said above, in you opensc-debug logs it looks like pkcs15_create_secret_key is failing, but there is not enough information in the logs.

You are using OpenSSL commands via libp11 and the engine. I am using pkcs11-tool.

You are using a Yubico,  I am using NIST PIV demo cards. (With the keyUsage modes that allows Yubico devices to bypass NIST PIV policies, make sure the certificate for the key MAN  key has keyUsage = keyAgreement when the key is ECC. Or has no keyUsage that would then default ti the NIST policies.)

SPY output might be helpful too. 

The only other thing I can suggest is to use a debugger and step through pkcs15_create_secret_key. 
I build  OpenSSL, OpenSC, libp11  with the CFLAGS=-g and LDFLAGS=-g to get the debugging symbols. I use gcc and gdb on Ubuntu.  I don't use MacOS, but would expect there is some compiler, loader and a debugger you could use.

mouse07410:
>  I think they are ready for review.

That I agree with.

> I don't see where the code has broken ECDH. There may be something else going on in your environment. 

I'm running on MacOS 10.12.4 with Xcode-8.3.1 (Apple-modified clang). Both differ from Ubuntu. My version of OpenSC is practically the main master in its entirety, with modified `opensc.conf` and one function logging the normal return (plus comments added to ifdef's to keep track of what endif matches what condition). You can see your yourself:
[diff.txt](https://github.com/OpenSC/OpenSC/files/929760/diff.txt)
On the other hand, for the sake of the cleanliness of the test I'm re-doing it with the "stock" OpenSC master.

Of course I use my fork of tokend (because the original one doesn't work properly, as you know - but that's a different story). And MacOS Sierra also has `pivtoken` - they're returning to providing their own tokend, but with a twist.

Regardless, without this PR all of the above works correctly. Therefore, it's reasonable to assume that this PR does something that breaks it.

> As I said above, in you opensc-debug logs it looks like pkcs15_create_secret_key is failing, but there is not enough information in the logs.

I will be happy to provide more.

> You are using OpenSSL commands via libp11 and the engine. I am using pkcs11-tool.

Well, for me the main point of OpenSC is to enable other applications to use tokens. The biggest (and IMHO the most important) examples of such apps/components are OpenSSL and Java (as lots of applications access crypto features through one or the other). Yes, I care very much that the command line tools (`pkcs11-tool`, `pkcs15-tool`, etc) keep working - but if they're the only way the ECC cards remain usable - thanks but no thanks.

My OpenSSL is the standard 1.0.2k:
```
$ openssl version -a
OpenSSL 1.0.2k  26 Jan 2017
built on: reproducible build, date unspecified
platform: darwin64-x86_64-cc
options:  bn(64,64) rc4(ptr,int) des(idx,cisc,16,int) idea(int) blowfish(idx) 
compiler: /usr/bin/clang -I. -I.. -I../include  -fPIC -fno-common -DOPENSSL_PIC -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -arch x86_64 -O3 -DL_ENDIAN -Wall -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DMD5_ASM -DAES_ASM -DVPAES_ASM -DBSAES_ASM -DWHIRLPOOL_ASM -DGHASH_ASM -DECP_NISTZ256_ASM
OPENSSLDIR: "/opt/local/etc/openssl"
$
```

My libp11 is the current GitHub master. My build uses shared libraries of OpenSSL.

> ...make sure the certificate for the key MAN key has keyUsage = keyAgreement when the key is ECC

Judge for yourself:
screen shot 2017-04-18 at 09 33 42
```
$ pkcs11-tool -O -l
. . . . .
Private Key Object; EC
  label:      KEY MAN key
  ID:         03
  Usage:      derive
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxxxxxxxx...xxx
  label:      KEY MAN pubkey
  ID:         03
  Usage:      none
. . . . .
```
(it is weird that the pubkey has usage "None" instead of "derive", but hopefully the "new" OpenSC takes what's in the certificate in this case)

> SPY output might be helpful too.

Yes I will do that. Could you please remind how to use SPY? I always keep forgetting what env variable I need to point at `/Library/OpenSC/lib/pkcs11-spy.so` (but I remember well `export  PKCS11SPY=/Library/OpenSC/lib/opensc-pkcs11.so` :).

**Will provide the SPY log as soon as I figure out how what env variable to use.**

> The only other thing I can suggest is to use a debugger and step through pkcs15_create_secret_key.
I build OpenSSL, OpenSC, libp11 with the `CFLAGS=-g` and `LDFLAGS=-g` to get the debugging symbols. I use gcc and gdb on Ubuntu. I don't use MacOS, but would expect there is some compiler, loader and a debugger you could use.

In general I love MacOS. But one thing I badly miss from Linux/Unix is a usable debugger interface. So yes, I have all of your list, but unfortunately find the debugger hardly usable. :-(

Perhaps you could suggest a way to instrument `pkcs15_create_secret_key()`, so that even without the debugger the log would contain sufficient info?

mouse07410:
More stuff:
```
$ PKCS11SPY_OUTPUT=~/1025-spy.txt pkcs11-tool --module /Library/OpenSC/lib/pkcs11-spy.so -vvv --derive --login --id 03 --input-file ec256-pub.der 
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter User PIN: 
Using derive algorithm 0x00001051 ECDH1-COFACTOR-DERIVE
error: PKCS11 function C_DeriveKey failed: rv = CKR_KEY_TYPE_INCONSISTENT (0x63)
Aborting.
$ pkcs11-tool -O -l
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter User PIN: 
Private Key Object; EC
  label:      PIV AUTH key
  ID:         01
  Usage:      none
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxxx
  EC_PARAMS:  0608xxxxxxxxxx
  label:      PIV AUTH pubkey
  ID:         01
  Usage:      none
  label:      Certificate for PIV Authentication
  ID:         01
Private Key Object; EC
  label:      SIGN key
  ID:         02
  Usage:      sign
  Access:     always authenticate
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxxxxxxx
  EC_PARAMS:  0608xxxxxxx
  label:      SIGN pubkey
  ID:         02
  Usage:      verify
  label:      Certificate for Digital Signature
  ID:         02
Private Key Object; EC
  label:      KEY MAN key
  ID:         03
  Usage:      none
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxxxxxxx
  EC_PARAMS:  0608xxxxxxxxxxx
  label:      KEY MAN pubkey
  ID:         03
  Usage:      none
  label:      Certificate for Key Management
  ID:         03
Data object 224431904
  label:          'Cardholder Fingerprints'
  application:    'Cardholder Fingerprints'
  app_id:         2.16.840.1.101.3.7.2.96.16
  flags:           private
. . . . .
```

Here's the SPY (with `pkcs11-tool` apparently one just points `--module `):
[1025-spy.txt](https://github.com/OpenSC/OpenSC/files/929989/1025-spy.txt)

And here's the "good" SPY:
```
$ PKCS11SPY_OUTPUT=~/good-spy.txt pkcs11-tool --module /Library/OpenSC/lib/pkcs11-spy.so -vvv --derive --login --id 03 --input-file ec256-pub.der 
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter User PIN: 
Using derive algorithm 0x00001051 ECDH1-COFACTOR-DERIVE
???}?ɏ??????{??P?J??0??8\?U?
```
[good-spy.txt](https://github.com/OpenSC/OpenSC/files/930002/good-spy.txt)

fabled:
@mouse07410 
> (it is weird that the pubkey has usage "None" instead of "derive", but hopefully the "new" OpenSC takes what's in the certificate in this case)

If the key has 'none' usage, it will cause derive operation to fail. The cause is probably of the same root cause. Please ensure that pkcs11 and pkcs15 libraries match -- that is that opensc is fully installed to system after applying the PR, and not just the other library.

I am genuinely interested on not breaking things. pkcs11-tool is used as it's the simplest way to test things, and triggers same code paths as openssl engine. But as we have two hardware items working, it suggest something specific to your environment or hardware. I will look at the logs you provided a bit later in detail. Thanks.

mouse07410:
> If the key has 'none' usage, it will cause derive operation to fail.

Yes I understand that. But with this PR the usage is 'none', and without it it is 'derive'.

> Please ensure that pkcs11 and pkcs15 libraries match...

They *cannot* "not match" because I always install the entire thing - on MacOS it builds the installer (aka `OpenSC.pkg`).

> I am genuinely interested on not breaking things.

:-)

> I will look at the logs you provided a bit later in detail...

Thank you! If more details are needed, I'd be happy to provide, and/or instrument the code as requested.

dengert:
You said earlier: "(it is weird that the pubkey has usage "None" instead of "derive", but hopefully the "new" OpenSC takes what's in the certificate in this case)"

Today you said:
"For some reason with this PR, Usage is seen as "none" for both prkey and pubkey. WIthout it - only pubkey has Usage: none, and the prkey shows the correct Usage: derive."

In pkcs15-piv.c, the keyUsage is checked at lines 768 to 892, and for EC and keyAgreement lines 867 and 868. the derive chould have been set for both pub and private, and it does not make sense that private key has derive and public key has none. 

The ckis[i].cert_keyUsage is an unsigned long long, and this could be a compiler problem. 
`if(ckis[i].cert_keyUsage & 0x10u) { /* keyAgreement */`
but looks more like something else. 

@mouse07410  I need to have **you** do some debugging.

mouse07410:
> You said earlier: "(it is weird that the pubkey has usage "None" instead of "derive", but hopefully the "new" OpenSC takes what's in the certificate in this case)"

Yep. That's what I observe, so what else can I say? Here's the output of `pkcs11-tool -O -l` (the only way I know of to view those flags):
```
$ pkcs11-tool -O -l
Private Key Object; EC
  label:      SIGN key
  ID:         02
  Usage:      sign, non-repudiation
  Access:     always authenticate
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxxxxx
  EC_PARAMS:  0608xxxxxxx
  label:      SIGN pubkey
  ID:         02
  Usage:      verify
Certificate Object, type = X.509 cert
  label:      Certificate for Digital Signature
  ID:         02
Private Key Object; EC
  label:      KEY MAN key
  ID:         03
  Usage:      derive
Public Key Object; EC  EC_POINT 256 bits
  EC_POINT:   xxxxxx
  EC_PARAMS:  0608xxxxxx
  label:      KEY MAN pubkey
  ID:         03
  Usage:      none
. . . . .
```

> Today you said: "For some reason with this PR, Usage is seen as "none" for both prkey and pubkey...

And I stand by it. :-)

With this PR somehow the output of `pkcs11-tool -O -l` changes to this.

> In pkcs15-piv.c, the keyUsage is checked at lines 768 to 892, and for EC and keyAgreement lines 867 and 868. the derive chould have been set for both pub and private, and it does not make sense that private key has derive and public key has none.

I fully agree. And in general I suspect that this new code (keyUsage stuff) is too fragile, and poorly traced. So I will try to instrument that code and collect more info. 

> @mouse07410 I need to have **you** do some debugging.

Sure, would love to. But I need guidance wrt. what/where to look, and what kind of info would be helpful for you to pinpoint this problem.

Also, the code would benefit from much better and far more extensive logging/tracing. Right now in one aspect it traces too much (dumps the PIN at the lowest usable debugging level), and too little in everything else.

mouse07410:
@dengert I'm going through that `pkcs15-piv.c` code, and it's rather frustrating. My first question - what does the message (line 739) mean?! "Cert can not be present"? "Can not" in what sense???

fabled:
@mouse07410 
> I suspect that 67ea96d is broken, at least on Mac.

That commit has nothing to do with this pull request. I would like to get clarification on which builds you are comparing? What is the commit id of the build that works? Can you confirm if the current git master works for you or not? If current opensc git master does not work for you, all this should go to a new / separate bug.

Please, use [git bisect](https://git-scm.com/docs/git-bisect) to determine which change breaks the usage for you.

frankmorgner:
@mouse07410 for your feedback to be helpful, please make sure to

- checkout [exactly this pull request](https://help.github.com/articles/checking-out-pull-requests-locally/)
- install OpenSC locally instead of using the pkg (which drops the debug symbols), for example by using `./configure --prefix=/tmp/ && make install && env DYLD_LIBRARY_PATH=${DYLD_LIBRARY_PATH}:/tmp/lib /tmp/bin/pkcs11-tool`
- Use `lldb` to find out where the problem is exactly (`lldb` has a ncurses based `gui`)
- Point to a specific change (in this PR) that you think *needs work*

If I read the above thread correctly, ECDH key derivation still works as expected for NIST's PIV test cards and the myeid card.

Yubikey's PIV ECDH key derivation seems to be broken due to a different (prior) commit. If so, you should indeed analyse this problem with [git bisect](https://git-scm.com/docs/git-bisect) in a seperate issue...

mouse07410:
Situation worse than it seemed (https://github.com/OpenSC/OpenSC/issues/1029 revealed more problems than reported there). I will need to `git bisect`.

dengert:
IMHO, @mouse07410 has a git problem, Mac problem configuration problem, compiler problem  or changes he may have in his version of OpenSC are causing strange problems.  Until https://github.com/OpenSC/OpenSC/issues/1029 is resolved, I am not looking at https://github.com/OpenSC/OpenSC/commit/67ea96d18b32690b74f873e21c9c4f570acdab5f

dengert:
Its not the "New English" See this comment in card-piv.c has:

2139 /*
2140  * If the object can not be present on the card, because the History
2141  * object is not present or the History object says its not present,
2142  * return 1. If object may be present return 0.
2143  * Cuts down on overhead, by not showing non existent objects to pkcs11
2144  * The path for the object is passed in and the first 2 bytes are used.
2145  * Note: If the History or Discovery object is not found the
2146  * PIV_OBJ_CACHE_NOT_PRESENT is set, as older cards do not have these.
2147  * pkcs15-piv.c calls this via cardctl.
2148  */

dengert:
@mouse07410  

You says things like: " I've added a ton of debug-logging there, and none of it appears in the log."
But you don't show what you added. 
This sounds like you are linking/loading against the wrong libraries which don't have your changes. You could look the trace and see if the line number of all the entries in the trace match the line numbers in your modified code. 

in https://github.com/OpenSC/OpenSC/files/931707/1025-bad.txt two of the highest pkcs15-piv.c trace entries are is:
pkcs15-piv.c:1167:sc_pkcs15emu_piv_init: USAGE: cert_keyUsage_present:1 usage:0x00000022
[opensc-pkcs11] pkcs15-piv.c:1174:sc_pkcs15emu_piv_init: returning with: 0 (Success)
Do these line number match? 
They match the git  version I have and my HEAD is at https://github.com/OpenSC/OpenSC/pull/1025/commits/c4d4c42ceeaa8fe931a79319caa486a01fe12537

We are talking about ECDH, but in the last output of pkcs11-tool -O -l all your certs are RSA!

You cant find out how to use OpenSC SPY on your own because you forgot?  Google is your friend
Google for: OpenSC SPY

You have local mods in your git and can not build a clean copy. This is a requirement to do any testing.
Does the MacOS require a make depend?

You need to lean how to use a debugger and other tools like @frankmorgner pointed out.  
Add gitk to that list too.

mouse07410:
> Until #1029 is resolved, I am not looking at 67ea96d

That's fair. 

> IMHO, @mouse07410 has a git problem, Mac problem configuration problem, compiler problem or changes he may have in his version of OpenSC are causing strange problems.

Let's see. Everything else works. Botan, Crypto++, all the Macports-installed packages (e.g., Boost) all work fine. I'm doing a ton of things on Github, all of them are OK. OpenSC doesn't work - so the problem must be with my git, Mac config, compiler, etc.? For the #1029 report I used the stock repo https://github.com/OpenSC/OpenSC.git instead of my fork, to make sure the problem lies there. And indeed, it does. 

To be more specific, the problem is that (a) the build process on Mac does not properly track dependencies, and (b) does not properly clean from the past builds.

#1029 manifests itself if before the new build I run `OpenSC Uninstaller.app`, which wipes the installation properly/thoroughly.

> You says things like: " I've added a ton of debug-logging there, and none of it appears in the log."
But you don't show what you added.

I can post the code here, no problem.

> This sounds like you are linking/loading against the wrong libraries which don't have your changes.

This is a part of #1029 - it appears that the old libraries aren't cleaned, and worse - the new ones are not built to replace the old ones. That's why I say that the build process on Mac is broken.

> You could look the trace and see if the line number of all the entries in the trace match the line numbers in your modified code.

Will do and report.

> We are talking about ECDH, but in the last output of pkcs11-tool -O -l all your certs are RSA!

No, we **were** talking about ECDH - but then I found that the problem was not with ECDH itself. It's a more fundamental problem that affects **both** RSA and EC tokens alike. And it is related to the current OpenSC's inability to determine/set the proper KeyUsage flags. And that in turn seems to be related to the current OpenSC's inability to fetch certificates from the token.

> You cant find out how to use OpenSC SPY on your own because you forgot? Google is your friend
Google for: OpenSC SPY

Great advice, unfortunately it falls short. I went to https://github.com/OpenSC/OpenSC/wiki/Using-OpenSC several times. I thought I made it clear in my report: what I forgot was how to point the app-under-test to the SPY module, not the `PKCS11SPY` variable setting (which is documented perfectly). BTW, it refers to an *EnvironmentVariables* page, but I was unable to locate one.

> You have local mods in your git and can not build a clean copy. This is a requirement to do any testing.

Oh no! My report is based on the clean copy of OpenSC, straight from the Github. Exactly to avoid a discussion along the above lines. I *started* with my fork, but then switched to the stock repo.

> Does the MacOS require a make depend?

MacOS by itself - no it doesn't. Some projects do, some don't. I've no idea whether OpenSC requires it, as I'm not the one who put together the OpenSC-on-MacOS build process. All I can say is that it seemed to work (though till now I haven't looked closely at whether it *really* replaced the old libraries - since the behavior seemed to be changing accordingly with the new installs, I just assumed it did the right thing).

> checkout exactly this pull request

Turns out this is not necessary - the problem seems to be with the master itself. :-(

> If I read the above thread correctly, ECDH key derivation still works as expected for NIST's PIV test cards and the myeid card.

I think ECDH works correctly *on Linux* (for NIST PIV and MyEID cards).

> Can you confirm if the current git master works for you or not? If current opensc git master does not work for you, all this should go to a new / separate bug.

You are correct. This is the git master problem, not one with this particular PR. 

I'm moving this discussion to a new bug report here: https://github.com/OpenSC/OpenSC/issues/1030

P.S. @frankmorgner thanks for the help. Full debugging output is at the #1030 issue.

mouse07410:
@fabled as you can see in #1030, the offending commit has been identified. With it reverted, OpenSC returned to functioning properly (with two glitches - one small and one moderate :).

However, I was unable to install your patch over OpenSC master with 777e2a3 reverted. I don't know if it's fair to ask you to take a look at that. In case you are willing - here's the patch against the current master that removes the CAC-adding commit: 
[opensc-remove-cac.zip](https://github.com/OpenSC/OpenSC/files/938508/opensc-remove-cac.zip)
I clone the master, then apply that patch, then build (as @frankmorgner suggested - using `./configure ...` rather than blindly relying upon `MacOSX/build`).

mouse07410:
@fabled After #1030 was taken care of, I've tested your PR for ECDH key derivation, and it worked fine:
```
$ /tmp/OpenSC/bin/pkcs11-tool --module /tmp/OpenSC/lib/opensc-pkcs11.so -l --derive --id 03 -i ec256pub.der -o sec-derived.dat
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter User PIN: 
Using derive algorithm 0x00001051 ECDH1-COFACTOR-DERIVE
$ openssl pkeyutl -engine pkcs11 -peerform engine -derive -inkey ec256priv.pem -peerkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20pubkey;type=public" -out sec-derived2.dat
engine "pkcs11" set.
$ cmp sec-derived.dat sec-derived2.dat
$ 
```

fabled:
Rebased on top of current git master.
@hhonkanen Care to look at this for the MyEID parts as you seem to be working with it?

frankmorgner:
please be more concise and move the problem about keyusage to an other thread.

mouse07410:
> please be more concise and move the problem about keyusage to an other thread.

Done. It's #1030

hhonkanen:
I looked at the changes in card-myeid.c and pkcs15-myeid.c. Basically they look good and I have no objection against any of them, but I haven't tested the code yet. myeid_init is cleaner after your refactoring. What kind of testing have you already done? I guess the basic RSA and ECC functionality should be tested. 

You removed key size validation from myeid_create_key, for example: 
-			if (sc_card_find_ec_alg(p15card->card, keybits, NULL) == NULL) 
-				LOG_TEST_RET(ctx, SC_ERROR_INVALID_ARGUMENTS, "Unsupported EC key size"); 
Is it guaranteed that the key size has been checked elsewhere before calling this function?

fabled:
> You removed key size validation from myeid_create_key, for example:
> 
> Is it guaranteed that the key size has been checked elsewhere before calling this function?

Only pkcs15-lib.c calls ops->create_key(). In all use cases it uses check_keygen_params_consistency() and/or check_key_compatibility() to verify that the request is sane.

hhonkanen:
Did some testing with a MyEID card

I first initialized a card:
pkcs15-init -P -a 1 -l "Basic PIN"

pkcs15-init -P -a 2 -l "Sign PIN"
**** this command fails, it outputs "Failed to store PIN: Not allowed" *** 
**** Tested running master with the same card in the same state and it worked. ****

pkcs15-init -F

Then ran the following commands:

* Generate a 1024 bit RSA key
pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --login --pin 1111 --keypairgen --key-type rsa:1024 --id 10 -a "rsa1024"

* test signing with the key 
pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -lt

* Generate a 2048 bit RSA key
./pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --login --pin 1111 --keypairgen --key-type rsa:2048 --id 11 -a "rsa2048"

* Generate a 256 bit EC key
pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -l -k --key-type EC:prime256v1 -d 12 -a "ec256"

* Sign data with the EC key
pkcs15-crypt -s -k 12 -i sha256.bin -o myeid-sha256.bin --pin "1111"

* 521 bit ECC test
pkcs15-init -G ec:secp521r1 -l "EC 521 Generated" --id 46 -a 1 --pin 1111 --so-pin "12345678"
pkcs15-crypt -s -k <previous key guid> -i sha512.bin -o myeid-sha512.bin --pin "1111"

* RSA key import
openssl genrsa -out rsa2048.pem 2048
pkcs15-init -a 1 --label "rsa2048-I" -S ./rsa2048.pem

All commands succeeded except for initializing PIN 2. Do you have an idea what has changed that affects PIN initialization?

fabled:
> pkcs15-init -P -a 2 -l "Sign PIN"
> **** this command fails, it outputs "Failed to store PIN: Not allowed" ***
> **** Tested running master with the same card in the same state and it worked. ****

This fails in git master too. PR #1034 fixes it.

mouse07410:
Did some testing on the current master with this PR using Yubikey NEO (EC keys over NIST P256). Success.

fabled:
It seems that TokenInfo is not written after creating a secret key that adds the supportedAlgorithms. Is it expected that TokenInfo is fully initialized in `pkcs15-init --create-pkcs15` ? Or what would be the way to signal `sc_pkcs15init_unbind` to update TokenInfo regardless of `profile->pkcs15.do_last_update` value when the TokenInfo contents are "dirty"?

mouse07410:
Naive question re. 53605a7: what's the purpose of this secret key id? Who is going to need/use that id, and how it would be discovered/referred-to?

dengert:
PKCS\#11 2.40 says:

"In the case of secret keys, the meaning of the CKA_ID attribute is up to the application."

A application could do a C_FindObjects for CKO_SECRET_KEY  and get the CKA_ID for each.

mouse07410:
> PKCS#11 2.40 says:
>
> "In the case of secret keys, the meaning of the CKA_ID attribute is up to the application."

Is there a "good practice" suggesting what that value should be? Do we know of any (reasonably well-known) application that creates/uses that value?

fabled:
The purpose of id is to uniquely identify a key object. Additionally, in public/private key pairs (+certificates) the id is to connect objects together. This is the reason for the other key types normally an intrinsic key id is calculated (hash of the public key component) which is a deterministic way to generate id from these objects. However, secret keys are different, there is only the secret key value. We could theoretically calculate PBKDF2 with very high iteration count over the value (in upload case, but not for onboard generation). But still that can be considered to leak information about the key and might not be wise. As the only purpose of the id is to allow users identify the key (there's no connected other key objects like with private and public keys), a GUID or any random generated value is sufficient. This is superior to the default selection method of using key reference which is a one byte id (it is guaranteed unique on card, but not globally). You can google on id usage, using GUID is also suggested in some related websites for generating secret key id.

frankmorgner:
I am not a user of `pkcs15-init --create-pkcs15`, so I can't suggest how `EF. TokenInfo` should be treated. But if I followed the discussion correctly, then there are no more issues with the suggested changes, right?

src/libopensc/card.c
```diff
@@ -952,6 +952,19 @@ int _sc_card_add_algorithm(sc_card_t *card, const sc_algorithm_info_t *info)
 	return SC_SUCCESS;
 }
 
+int _sc_card_add_generic(sc_card_t *card, unsigned int algorithm,
+			 unsigned int key_length, unsigned long flags)
```
frankmorgner: Oh, I remember that you should change this. The function is simply not *generic*. Either rename and export this to `_sc_card_add_symmetric_alg` as suggested by @dengert or export four dedicated functions for each algorithm (`_sc_card_add_des_alg`, _`sc_card_add_3des_alg`, `_sc_card_add_aes_alg`) possibly mapping to a single implementation of `_sc_card_add_symmetric_alg`.

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1274,6 +1277,115 @@ sc_pkcs15init_init_prkdf(struct sc_pkcs15_card *p15card, struct sc_profile *prof
...
+	r = select_id(p15card, SC_PKCS15_TYPE_SKEY, &keyargs->id);
+	LOG_TEST_GOTO_ERR(ctx, r, "Cannot select ID for SKey object");
+
+	key_info->id = keyargs->id;
+
+	r = select_object_path(p15card, profile, object, &key_info->path);
+	LOG_TEST_GOTO_ERR(ctx, r, "Failed to select secret key object path");
+
+#if 0
+	/* See if we need to select a key reference for this object */
```
frankmorgner: What's the status of this piece of code? If it's not of use, it should be removed.

src/pkcs15init/pkcs15-myeid.c
```diff
@@ -349,19 +352,33 @@ myeid_new_file(sc_profile_t *profile, sc_card_t *card,
...
 	char name[64];
-	const char *tag;
+	const char *tag = NULL;
 	int r;
 
 	LOG_FUNC_CALLED(card->ctx);
-	if (type == SC_PKCS15_TYPE_PRKEY_RSA || type == SC_PKCS15_TYPE_PRKEY_EC)
+	switch (type) {
+	case SC_PKCS15_TYPE_PRKEY_RSA:
+	case SC_PKCS15_TYPE_PRKEY_EC:
```
frankmorgner: please add `/* fall through */` comments

src/tools/pkcs15-init.c
```diff
@@ -1598,68 +1726,73 @@ static int
...
+}
 
-	if (*spec)   {
-		if (isalpha(*spec) && keygen_args.prkey_args.key.algorithm == SC_ALGORITHM_EC)   {
-			keygen_args.prkey_args.key.u.ec.params.named_curve = strdup(spec);
-			keybits = 0;
-		}
-		else {
-			char	*end;
+#if 0
```
frankmorgner: What about this piece of code?

src/libopensc/card.c
fabled: Renamed to `_sc_card_add_symmetric_alg`.

src/pkcs15init/pkcs15-lib.c
fabled: Removed unsupported code, made it return error for now if card requires key reference selection.

src/tools/pkcs15-init.c
fabled: Implemented the high level logic of requesting symmetric key generation. No card driver supports it yet, though.

src/pkcs15init/pkcs15-myeid.c
fabled: The only problem left is what I described earlier:

> It seems that TokenInfo is not written after creating a secret key that adds the supportedAlgorithms.
> what would be the way to signal sc_pkcs15init_unbind to update TokenInfo regardless of profile->pkcs15.do_last_update value when the TokenInfo contents are "dirty"?

After the TokenInfo on card is updated properly, this pull request should be ready for merge unless additional issues are found.

fabled:
The only problem left is what I described earlier:

>  It seems that TokenInfo is not written after creating a secret key that adds the supportedAlgorithms.
>  what would be the way to signal sc_pkcs15init_unbind to update TokenInfo regardless of profile->pkcs15.do_last_update value when the TokenInfo contents are "dirty"?

Any suggestions how to do this?

After the TokenInfo on card is updated properly, this pull request should be ready for merge unless additional issues are found.

frankmorgner:
Unfortunately, I cannot help with this point. Maybe @CardContact can look a this, since SC-HSM could also benefit from secret key support (think of sc-hsm's "device key encryption key").

fabled:
Added the code to update TokenInfo structure when needed.

frankmorgner:
Only the `/* fall through */` comments are missing. Have you used `valgrind` to check for memory issues?

fabled:
@frankmorgner Where would fallthrough comment be required?

To me the only place my patch adds is myeid_new_file() and myeid_store_key() but in there it's just grouping same action for multiple cases. It is standard across codebase to not add this comment in this case. The only placse current code base has fallthrough comment is in src/libopensc/aux-data.c and src/tools/pkcs15-tool.c where the first case does things and then falls through to second case's code.

src/pkcs15init/pkcs15-myeid.c
frankmorgner: coverity should not report these (http://security.coverity.com/blog/2013/Sep/gimme-a-break.html), no need to add comments.

frankmorgner:
If you want to have the changes in the upcoming release, please hurry up adding the documentation, thanks!

fabled:
I am busy this weekend, but looking at this on Monday. Hope it's good enough.

frankmorgner:
OK, thanks!

fabled:
Added documentation.

frankmorgner:
thanks

PR #1032, opened 2017-04-20
Fix CAC-introduced bug that broke access to certificates

mouse07410:
Thanks to @dengert for finding this problem.

Please review and (hopefully) merge soon.

PR #1033, opened 2017-04-21
Fix CKA_DERIVE bug in pkcs11-tool.c 

mouse07410:
This bug prevented it from displaying CKA_DERIV key usage for EC public keys. Thanks to @dengert for finding the bug and proposing this fix.

I've tested it on MacOS El Capitan and Sierra. Output of `pkcs11-tool -O -l` is correct now (assuming #1032 is applied, as without #1032 OpenSC simply cannot retrieve certificates, at least on Mac).

Request review and merge. Since it's a one-liner, it shouldn't take too long.


mouse07410:
Thank you!

PR #1038, opened 2017-04-25
Restore the functionality of deriving usage flags from X509 certificates in CAC

Jakuje:
Use the correct constants (correct bit order).

The `SC_X509_*` constants will most probably have some better place in the code. `libopensc/pkcs15.h`? Ideas welcome.


Jakuje:
We should probably use the same constants in the `pkcs15-piv.c` after placing them in the common header file to prevent hardcoding of magic numbers in PIV code (see the second commit).

dengert:
the pkcs15-piv.c looks good.

PR #1045, opened 2017-05-09
If the underlying PKCS#15 structure does not provide label for a cert…

Jakuje:
If the underlying PKCS#15 structure does not provide label for a certificate, try to use DN from the certificate.
This is helpful to simplify the orientation in the CA chains stored on the smart card, which currently happens to us with CardOS 5 cards, but I believe other cards can benefit from this too.

I am not entirely sure when the call from `check_cert_data_read()` will get invoked (the tests with `pkcs11-tool` I used invoke the function directly from `__pkcs15_create_cert_object()`, but I believe it is the right place to do this also with the invocation from `pkcs15_cert_get_attribute()`. But if somebody with better overview of the code can come up with better place, I am fine.

The idea comes from coolkey and from the following Red Hat bugzilla:

https://bugzilla.redhat.com/show_bug.cgi?id=1448555

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -553,6 +553,27 @@ public_key_created(struct pkcs15_fw_data *fw_data, const struct sc_pkcs15_id *id
 	return SC_ERROR_OBJECT_NOT_FOUND;
 }
 
+static void
+pkcs15_cert_extract_label(struct pkcs15_cert_object *cert)
+{
+	sc_log(context, "pkcs15_cert_extract_label() called. Current label: %s", cert->cert_p15obj->label);
+
+	/* if we didn't get a label, set one based on the CN */
+	if (*cert->cert_p15obj->label == '\0') { /* can't be NULL -- static array */
+		static const struct sc_object_id cn_oid = {{ 2, 5, 4, 3, -1 }};
+		u8 *cn_name = NULL;
+		size_t cn_len = 0;
+		int rv = sc_pkcs15_get_name_from_dn(context,
+			cert->cert_data->subject, cert->cert_data->subject_len,
+			&cn_oid, &cn_name, &cn_len);
+		sc_log(context, "pkcs15_cert_extract_label(): Name from DN is %s", cn_name);
+		if (rv == SC_SUCCESS) {
+			memcpy(cert->cert_p15obj->label, cn_name, cn_len);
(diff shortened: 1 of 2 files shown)


frankmorgner:
can this code be removed from the coolkey driver then?

Jakuje:
Maybe it was a bit confusing but, this code is/was NOT in Coolkey driver in OpenSC, but it is in the Coolkey library (outside of OpenSC) we were using before (problems with terminology).

frankmorgner:
i see.

src/pkcs11/framework-pkcs15.c
```diff
@@ -553,6 +553,27 @@ public_key_created(struct pkcs15_fw_data *fw_data, const struct sc_pkcs15_id *id
...
+	if (*cert->cert_p15obj->label == '\0') { /* can't be NULL -- static array */
+		static const struct sc_object_id cn_oid = {{ 2, 5, 4, 3, -1 }};
+		u8 *cn_name = NULL;
+		size_t cn_len = 0;
+		int rv = sc_pkcs15_get_name_from_dn(context,
+			cert->cert_data->subject, cert->cert_data->subject_len,
+			&cn_oid, &cn_name, &cn_len);
+		sc_log(context, "pkcs15_cert_extract_label(): Name from DN is %s", cn_name);
+		if (rv == SC_SUCCESS) {
+			memcpy(cert->cert_p15obj->label, cn_name, cn_len);
```
frankmorgner: I think there is a length check for `sizeof cert->cert_p15obj->label` missing.

Jakuje: Thanks for comment. Fixed by truncating the length to `SC_PKCS15_MAX_LABEL_SIZE-1` in case it is longer.

frankmorgner:
```
error LNK2019: unresolved external symbol _MIN referenced in function
```

Jakuje:
I missed that `internal.h`, where the `MIN` is defined, is not included in this file.
I also hope including it will not cause some more trouble, otherwise I would have to redefine that somewhere else (or if you can point me to some other general purpose header file where it could land -- I see the `define MIN(` in three places so far around the code base).

Jakuje:
I don't see the readon why the AppVeyor failed this time.

frankmorgner:
could be related to the WiX update recently https://www.appveyor.com/updates/2017/05/18/

PR #1046, opened 2017-05-09
Openssl 1 1 additional changes

dengert:
With newer versions of OpenSSL-1.1, two Opensc tools: gids-tool and piv-tool were found to get in a loop within OpenSSL during OpenSSL initialization. Changes were added to each to avoid the situation.  No other tools were found to get in this loop. 

Both tools were found to fail with OpenSC-1.1.0c and OpenSSL-1.1.0e.  They worked before and after the changes with OpenSSL-1.0.2. 

The only testing of gids-tool was to make sure it did not hang in the loop, and at least get to the "card not present" message.

### DIFF
--- a/src/tools/gids-tool.c
+++ b/src/tools/gids-tool.c
@@ -33,10 +33,19 @@
 #include <ctype.h>
 #include <sys/stat.h>
 
+#include <openssl/opensslv.h>
+#include "libopensc/sc-ossl-compat.h"
+#if OPENSSL_VERSION_NUMBER >= 0x10000000L
+#include <openssl/opensslconf.h>
+#include <openssl/crypto.h>
+#endif
+#if OPENSSL_VERSION_NUMBER >= 0x00907000L
+#include <openssl/conf.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


frankmorgner:
why don't you create a wrapper for `OPENSSL_init_crypto`? Ideally, new code only uses API calls of one specific OpenSSL version (1.1?), which may get mapped to lower versions internally.

dengert:
I found that the piv-tool would loop, when using OpenSSL-1.1.0e, so once I fixed it, I tested all  
18 OpenSC tools. The test was done with no card in the reader.  Either it would loop or quit.  Only the piv-tool and gids-tool would loop. So I only fixed these two. 

Each tool loads different sets of OpenSSL header files and in different orders. They call different OpenSSL routines  so I was reluctant to touch them.

But if you want, I can go through all of them and clean them up.

dengert:
@frankmorgner  can you at least commit this PR. I can then look at cleaning up all the tools and libopensc routines that try and initialize OpenSSL.

PR #1058, opened 2017-06-05
Some more Coverity Scan fixes

frankmorgner:
Closes https://github.com/OpenSC/OpenSC/issues/1057

### DIFF
--- a/src/libopensc/card-entersafe.c
+++ b/src/libopensc/card-entersafe.c
@@ -503,7 +503,7 @@ static int entersafe_select_fid(sc_card_t *card,
 	path.len=2;
 
 	r = iso_ops->select_file(card,&path,&file);
-	if (r)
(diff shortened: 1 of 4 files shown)


dengert:
@frankmorgner 
In iso7816.c, if you are going to change `read` to `MAX_SM_APDU_RESP_SIZE`, why only in only place?
This looks like coverity got a little carried away. The code is now more confusing.

frankmorgner:
Thanks for the review!

This specific size of chunks to be read is only used in `iso7816_read_binary_sfid` so there's no need to change anything else.

However, the actual problem which this piece of code solves should eventually be applied to `sc_get_max_recv_size`. If SM is enabled, `sc_get_max_recv_size` should return the amount of plaintext data I can receive before encryption. This, however, requires more thought than the suggested quick fix.

PR #1064, opened 2017-06-10
Fix pkcs11-tool handling CKR_USER_NOT_LOGGED_IN and CKA_ALWAYS_AUTHETICATE

dengert:
A macro C_X_INIT was added to be used vor any of the C_*Init routines that might
fail with CKR_USR_NOT_LOGGED_IN . A login is triggered, and the C_*Init routine is tried again.

The authenticate_if_required routin was replaced with two routines:
login_if_needed, and context_specific_login_if_needed to be used for the two
different types of logins that may be needed.

 Changes to be committed:
	modified:   pkcs11-tool.c

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -77,13 +77,13 @@
  * relogin then try again. 
  */
 
-#define C_X_INIT(R, F, S, M, K)  ({ \
+#define C_X_INIT(R, F, S, M, K)  { \
 	R = F(S, M, K); \
 		if (R == CKR_USER_NOT_LOGGED_IN) { \
 			login_if_required(S); \
 			R = F(S, M, K); \
 		} \
-	})
+	}
 
 
 			
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -70,6 +70,23 @@
 #endif
 #endif
 
+
+/*
+ * Macro to call crypto init routines like C_SignInit
+ * if it fails with CK_USER_NOT_LOGGED_IN
+ * relogin then try again. 
+ */
+
+#define C_X_INIT(R, F, S, M, K)  ({ \
+	R = F(S, M, K); \
+		if (R == CKR_USER_NOT_LOGGED_IN) { \
+			login_if_required(S); \
+			R = F(S, M, K); \
+		} \
+	})
+
+
+			
 extern void *C_LoadModule(const char *name, CK_FUNCTION_LIST_PTR_PTR);
 extern CK_RV C_UnloadModule(void *module);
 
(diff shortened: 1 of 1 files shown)


frankmorgner:
Thanks for looking into this issue again. However, I'm not convinced that the patch you're proposing is ideal for solving this issue.

1. The macro is a bit cryptic (though your comment certainly helps for understanding)
2. You should use `do {...} while (0)` with for these kind of macros
3. I don't think issuing a cryptographic operation a second time on `CKR_USER_NOT_LOGGED_IN` is the best we can do. As far as I know, Firefox requests the session state before using a key. This would be a cleaner approach...
4. Why is returning an error on `CKR_USER_NOT_LOGGED_IN` no option? Unless `getALWAYS_AUTHENTICATE()` is true, I don't see a good reason to log into the token more than once. (Yes, we could magically recover from things like card resets, but we are in `pkcs11-tool` which claim completeness!)

dengert:
Thanks for reverting https://github.com/OpenSC/OpenSC/commit/423375c6f85853d359d502a28e676a09e33a0ac1 it should allow 0.17.0 to proceed.

I will close #1064 

But I would like to address your questions.  

(1) and (2) The macro can easily be changed. 

(3) I agree this may not be the best we can do. I was trying to keep it as simple as possible. See below on FireFox comments. 

(4) The reason for the macro, was to handle a `CKR_USER_NOT_LOGGED_IN`  by doing a C_Login then  retry the crypto operation and see if it works. If not fail with `CKR_USER_NOT_LOGGED_IN` The idea was pkcs11-tool  should use the pin from the --pin option or ask the user for the pin a second time and not fail. 

There are number of reasons for `CKR_USER_NOT_LOGGED_IN` being returned and by multiple routines. Looking at the crypto routines: they have 3 types of routines (using Sign as the example). C_SignInit, C_SignUpdate and (C_Sign or C_SignFinal)

Any of these could fail because of the login state was lost. FireFox checks the login state in order to catch this before the start of a crypto sequence. In FireFox case, the session may have been inactive for  seconds or minutes where the user may have used the card from some other application. Once a crypto sequence is started they are done without delay, and little chance of interference. 

The patch checks the login state in both the authenticate_if_needed and context_specific_login_if_needed routines much like FireFox does.  

As discussed in  https://github.com/OpenSC/libp11/issues/160#issuecomment-304931271 SafeNet can return  `CKR_USER_NOT_LOGGED_IN` for C_SignInit as it does not support CKA_ALWAYS_AUTHENTICATE but wants a C_Login to be done. 

Pin caching can cause a crypto operation to fail, and a new C_Login is needed. The patch may not address this very well, but pin caching is an issue for @nunojpg  

@frankmorgner 
Which one of the C_SignInit, C_SignUpdate, C_Sign or C_SignFinal would fail with a pin_cache limit?

Can the failure from a pin_cache be prevented by checking the login state before doing one of the above?  i.e. does the pin_cache code change the login state?

Could  OpenSC use one of the VENDOR_SUPPLIED routines or attributes to pass or set pin_cache parameters?  

It is interesting to note that PKCS\#15 supports `user_consent=N` but PKCS\#11 only supports CKA_ALWAYS_AUTHENTICATE which appears to be equivalent to  user_consent=1.  Why did PKCS\#11 not address `user_consent>1`? Do we need to look at how OpenSC supports `user_consent=N` via PKCS\#11?

PKCS\#11 CKA_ALWAYS_AUTHENTICATE expects the application to call C_Login(CKU_CONTECT_SPECIFIC) immediately before the C_Sign or C_SignFinal. Can we do something with this for user_consent>1?

dengert:
Closing for now.

PR #1067, opened 2017-06-14
pkcs11-tool: respect CKA_ALWAYS_AUTHENTICATE

frankmorgner:
Quoting from PKCS#11:

The CKA_ALWAYS_AUTHENTICATE attribute can be used to force re-authentication (i.e. force the user to provide a PIN) for each use of a private key. “Use” in this case means a cryptographic operation such as sign or decrypt. This attribute may only be set to CK_TRUE when CKA_PRIVATE is also CK_TRUE.

Re-authentication occurs by calling C_Login with userType set to CKU_CONTEXT_SPECIFIC immediately after a cryptographic operation using the key has been initiated (e.g. after C_SignInit).

Closes https://github.com/OpenSC/OpenSC/pull/1066

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -426,6 +426,80 @@ static CK_ULONG		get_private_key_length(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE
 # define O_BINARY 0
 #endif
 
+#define ATTR_METHOD(ATTR, TYPE) \
+static TYPE \
+get##ATTR(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj) \
+{ \
+	TYPE		type = 0; \
+	CK_ATTRIBUTE	attr = { CKA_##ATTR, &type, sizeof(type) }; \
+	CK_RV		rv; \
+ \
+	rv = p11->C_GetAttributeValue(sess, obj, &attr, 1); \
+	if (rv != CKR_OK) \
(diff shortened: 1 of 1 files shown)


dengert:
Note in pkcs11-tool. c session is sometimes named sess. 

The key may be called key, priv_key or  privKeyObject. 

As I noted in #1066, not all crypto operations in pkcs11-tool.c test for CKA_ALWAYS_AUTHENTICATE.
The first C_SignInit is sign_data at line 1547.  There are 11 calls to C_SignInit, but only 6 callst to getALWAYS_AUTHENTICATE.

The getALWAYS_AUTHENTICATE macro is defined around line  2981: `ATTR_METHOD(ALWAYS_AUTHENTICATE, CK_BBOOL);`

To do a complete mod, would require moving all the ATTR_* stuff in lines 2936-3007 up up in the source file before sign_data, and adding additional calls to getALWAYS_AUTHENTICATE. 

To keep #1066 simple, (2 lines) I did not address all the other places that should also be changed, and was wiling to wait til after 0.17.0.  

#1064 did address all the calls and used a context_specific_login_if_needed  routine which did not require moving the macro code around. 

You did not like #1064 so I closed it and submitted #1066 as a simple short term fix. 

Do we fix pkcs11-tool.c correctly before releasing 0.17.0 or release it some minimal patches that does not cover all the cases?

frankmorgner:
I see, sorry for the inconvenience. The patch is fixed and tested now.

frankmorgner:
The changes are fairly simple and obvious, so I'd like to include them in 0.17.0

PR #1069, opened 2017-06-17
sc_pkcs15_read_der_file fails when used with piv to read pubkey from file

dengert:
I would like this to make it in 0.17.0. As stated below, it looks like only the piv code would be effected by this change. 

sc_pkcs15_read_der_file reads 16 bytes to get the tag. It then calls
  sc_asn1_read_tag(&body, len, &cla_out, &tag_out, &bodylen);
with len = 16 passed into sc_asn1_read_tag as buflen.

sc_asn1_read_tag parses the tag then checks to make sure the data
does overflow the size left.

At this point len = the size of the data  for the tag, which in most cases
is greater the 16.
  if (len > left)

This causes sc_pkcs15_read_der_file to fail.

The change replaces the call with:
  sc_asn1_read_tag(&body, 0xffff, &cla_out, &tag_out, &bodylen);
Thus the if above does not fail.

The PIV code  used sc_asn1_read_tag  with 0xffff to get the size of
an object so to can allocate a buffer then read the full object.

As far as I can tell the only place this ie ever used (and used to work)
in from pkcs15-piv.c  when there is key on the card but no certificate.
and is used when trying to sign a certificate request to get the public key.

 Changes to be committed:
	modified:   pkcs15-pubkey.c

### DIFF
--- a/src/libopensc/pkcs15-pubkey.c
+++ b/src/libopensc/pkcs15-pubkey.c
@@ -1263,7 +1263,11 @@ sc_pkcs15_read_der_file(sc_context_t *ctx, char * filename,
 	len = r;
 
 	body = tagbuf;
-	r = sc_asn1_read_tag(&body, len, &cla_out, &tag_out, &bodylen);
+	/*
+	 * We are only getting the tag, without the rest of the data
+	 * don't test that the tag and data is lenn theg the len 
+	 */
+	r = sc_asn1_read_tag(&body, 0xffff, &cla_out, &tag_out, &bodylen);
 	if (r != SC_SUCCESS)
 		goto out;
 


frankmorgner:
I think this hard coded length could lead to problems in the future. I'm wondering why `sc_asn1_read_tag` is looking at the completeness of the object at all since the name implies that it only looks at the tag.

There is already `SC_ERROR_ASN1_END_OF_CONTENTS`, which is used in some parts of the code to gracefully handle incomplete ASN.1 data without emitting an error. I think `sc_asn1_read_tag` should be changed to something like this:
```
diff --git a/src/libopensc/asn1.c b/src/libopensc/asn1.c
index 9b931433..5122ebc8 100644
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -112,12 +112,15 @@ int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
                }
                len = a;
        }
-       if (len > left)
-               return SC_ERROR_INVALID_ASN1_OBJECT;
+
        *cla_out = cla;
        *tag_out = tag;
        *taglen = len;
        *buf = p;
+
+       if (len > left)
+               return SC_ERROR_ASN1_END_OF_CONTENTS;
+
        return SC_SUCCESS;
 }
 
diff --git a/src/libopensc/pkcs15-pubkey.c b/src/libopensc/pkcs15-pubkey.c
index 57bbf01d..a9431aa6 100644
--- a/src/libopensc/pkcs15-pubkey.c
+++ b/src/libopensc/pkcs15-pubkey.c
@@ -1256,7 +1256,7 @@ sc_pkcs15_read_der_file(sc_context_t *ctx, char * filename,
 
        body = tagbuf;
        r = sc_asn1_read_tag(&body, len, &cla_out, &tag_out, &bodylen);
-       if (r != SC_SUCCESS)
+       if (r != SC_SUCCESS && r != SC_ERROR_ASN1_END_OF_CONTENTS)
                goto out;
 
        if (tag_out == SC_ASN1_TAG_EOC || body == NULL)   {
```
Unfortunately this could have side effects so that I'd like to postpone it...

frankmorgner:
@dengert have you looked at the suggestion above?

dengert:
Yes it looks good to me. But you said: "so that I'd like to postpone it..." and it is only used  for PIV to get the public key from a file when creating a certificate request because the there is never a public key on the card only in the certificate. So I was willing to let that wait till after 0.17.0. It would be nice if it was fixed in 0.17.0.

sc_asn1_read_tag is called in card-piv.c  with a  length of 0xffff that could also be benefit from a similar  change to pass in the tag length and test for SC_ERROR_ASN1_END_OF_CONTENTS.

frankmorgner:
OK, then I'll apply my suggestion as soon as 0.17.0 is done

dengert:
@frankmorgner since your proposed change https://github.com/OpenSC/OpenSC/pull/1069#issuecomment-309375635 will not be implemented in 0.17.0.  I would still like to see in 0.17.0 either my 0xffff change or the alternative patch below.

I would like to see something  in 0.17.0 to fix this regression. Otherwise it will not be fixed  until 0.18.0. in a year or two.
 
Both changes are in  pkcs15-pubkey.c in sc_pkcs15_read_der_file.  sc_pkcs15_read_der_file is only called from sc_pkcs15_pubkey_from_spki_file that is only called from pkcs15-piv.c, and only called  when there is no certificate on the card, and only when signing a certificate request. So either of these two patches have no effect any other code or cards. 

Proposed alternate patch:


```
diff --git a/src/libopensc/pkcs15-pubkey.c b/src/libopensc/pkcs15-pubkey.c
index e1935a1..000849f 100644
--- a/src/libopensc/pkcs15-pubkey.c
+++ b/src/libopensc/pkcs15-pubkey.c
@@ -1234,7 +1234,7 @@ sc_pkcs15_read_der_file(sc_context_t *ctx, char * filename,
 	int r;
 	int f = -1;
 	size_t len, offs;
-	u8 tagbuf[16]; /* enough to read in the tag and length */
+	u8 tagbuf[1024]; /* 16 enough to read in the tag and length. 1024 is hack till PR 1069 */
 	u8 * rbuf = NULL;
 	size_t rbuflen = 0;
 	const u8 * body = NULL;

```

This is a one line patch that increases the read buffer so the full pubkey  is read in one operation. All pubkey in the foreseeable future are less then 1024 bytes. The issue with  sc_asn1_read_tag will not be triggered. This has been tested against 0.17.0-rc2 using gdb and stepping through the  code. 

When your improved  sc_asn1_read_tag is added, the  tagbuf can be set back to 16.

frankmorgner:
OK

PR #1070, opened 2017-06-17
Minidriver improvements from #850

maciejsszmigiero:
This pull requests contains minidriver improvements from #850, without card reset handling.

Specifically it contains two commits:
* Multiple PINs support in minidriver (practically the old commit from #850 ported on top of the current master),

* Minidriver card reinitialization cleanup, making sure that functions in card reinitialization path leave minidriver context in a consistent state if some error happens during their execution.
It is mostly a cleanup part of "Keep track of card resets by other contexts in minidriver"
(without the actual reset handling code introduced by that commit), simplified.

There is also a third, one-line commit fixing a compiler warning in sc-hsm.
This commit is not related to minidriver, I have added it here to not create a separate pull request for such a tiny change.

There is more information about each change in an individual commit message.

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -32,6 +32,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <time.h>
 
 #include <windows.h>
@@ -179,6 +180,8 @@ struct md_guid_conversion md_static_conversions[MD_MAX_CONVERSIONS] = {0};
 
 typedef struct _VENDOR_SPECIFIC
 {
+	BOOL initialized;
+
 	struct sc_pkcs15_object *pin_objs[MD_MAX_PINS];
 
 	struct sc_context *ctx;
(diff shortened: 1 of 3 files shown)


frankmorgner:
Thanks, I'll look into this when the upcoming release is done.

maciejsszmigiero:
Thanks.

src/minidriver/minidriver.c
```diff
@@ -394,43 +476,65 @@ md_get_pin_by_role(PCARD_DATA pCardData, PIN_ID role, struct sc_pkcs15_object **
...
-	*ret_obj = NULL;
-
-	if (role == ROLE_USER)   {
-		if (!vs->obj_user_pin)   {
-			/* Get 'global' User PIN; if no, get the 'local' one */
-			rv = sc_pkcs15_find_pin_by_flags(vs->p15card, SC_PKCS15_PIN_TYPE_FLAGS_PIN_GLOBAL,
-					SC_PKCS15_PIN_TYPE_FLAGS_MASK, NULL, &vs->obj_user_pin);
-			if (rv)
+	/* please keep me in sync with _get_auth_object_by_name() in pkcs11/framework-pkcs15.c */
+	if (role == ROLE_USER) {
```
maciejsszmigiero: This code was copied from _get_auth_object_by_name() in pkcs11/framework-pkcs15.c.
It was left mostly 'as-is' to make it easier to spot inconsistencies.

src/minidriver/minidriver.c
```diff
@@ -1384,12 +1523,35 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
...
 
 	logprintf(pCardData, 2, "set 'cmapfile'\n");
 	vs = pCardData->pvVendorSpecific;
+	if (!vs)
+		return SCARD_E_INVALID_PARAMETER;
+
+	dwret = md_get_pin_by_role(pCardData, ROLE_USER, &vs->pin_objs[ROLE_USER]);
+	if (dwret != SCARD_S_SUCCESS)   {
+		logprintf(pCardData, 2, "Cannot get User PIN object");
+		return dwret;
```
maciejsszmigiero: The old code (before these commits) also required presence of an user PIN - otherwise associate_card() would fail and so would then CardAcquireContext().

From a quick glance at the code it seems to me that it would be possible to make minidriver work even for cards that don't have an user PIN, however that would be a separate change from these
implementing multiple PINs support.

src/minidriver/minidriver.c
frankmorgner: OK, I see

src/minidriver/minidriver.c
frankmorgner: OK, I see

frankmorgner:
Turns out, I was not able to sign with a second pin due to a missing signature mechanism. The new support of the signature PIN itself is fine.

PR #1075, opened 2017-06-23
Support for new MinInt agent card

af-anssi:
This card uses the same ATR as the existing card, but the applet installed
does not have the same AID. This card actually works exactly as the
IASECC_SAGEM.


frankmorgner:
thanks. I'll look at this when the upcoming release is done.

af-anssi:
Thank you. The patch is written in a way there is no regression with any existing card, so it can be safely integrated in the upcoming release.

PR #1079, opened 2017-06-25
CardOS 5.0 support improvement

debrouxl:
I saw that OpenSC 0.17-rc1 became available in Debian sid, so I attempted to test it against my CardOS 5.0 card. I noticed that `pkcs11-tool -t -l` failed the same way it did in April before I applied the relevant part of an additional change by @Jakuje on top of #1003 .

If this gets into the 0.17 release, I can quit using my own patched builds of OpenSC :)
The fact that the 0.17 release is close is why I made a minimal change to use P1 = 0x41 selectively on CardOS 5.0, instead of the more invasive code simplification (unifying CardOS 5.0 and 5.3 support) from https://github.com/Jakuje/OpenSC/commit/7356ee75a848456c871ab99267b1e49616fe2100 .

### DIFF
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -789,7 +789,7 @@ cardos_set_security_env(sc_card_t *card,
 	if (card->type == SC_CARD_TYPE_CARDOS_CIE_V1) {
 		cardos_restore_security_env(card, 0x30);
 		apdu.p1 = 0xF1;
-	} else if (card->type == SC_CARD_TYPE_CARDOS_V5_3) {
+	} else if (card->type == SC_CARD_TYPE_CARDOS_V5_0 || card->type == SC_CARD_TYPE_CARDOS_V5_3) {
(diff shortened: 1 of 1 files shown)


frankmorgner:
@Jakuje does this patch fix your problems you had with testing the release?

Jakuje:
Frank, I don't think so.
I have only 5.3 cards (which do not work now) and this PR changes the behavior of 5.0 cards. Though I am wondering why that referenced branch was not brought in (probably I forgot to finish the PR).
I would be for merging rather the referenced branch, since it is removing the 5.3 complexity which I introduced before knowing the 5.0 and 5.3 are the same for the supported use cases so far.

Jakuje:
@debrouxl  Testing #1080 with CardOS 5.0 to make sure it solves the problem and does not break anything else there would be appreciated. I put together simple commands to do that since here we can not use `pkcs11-tool`:

Issue #1073, opened 2017-06-19
ePass2003 EC not supported


majkrzak:
It seems that OpenSC (**0.16.0** Arch Linux build) don't support elliptic curves on Feitian Technologies, Inc. ePass2003 USB token (**096e:0807**).

    ~> opensc-tool --list-algorithms
    Using reader with a card: Feitian ePass2003 00 00
    Algorithm: rsa
    Key length: 512
    Flags: onboard key generation padding ( none ) hashes ( )

    Algorithm: rsa
    Key length: 768
    Flags: onboard key generation padding ( none ) hashes ( )

    Algorithm: rsa
    Key length: 1024
    Flags: onboard key generation padding ( none ) hashes ( )

    Algorithm: rsa
    Key length: 2048
    Flags: onboard key generation padding ( none ) hashes ( )

    ~> pkcs11-tool -k --key-type EC:prime256v1
    Using slot 0 with a present token (0x0)
    error: Generate EC key mechanism not supported

    ~> pkcs15-init --generate-key EC:prime256v1 --key-usage sign --auth-id 1
    Using reader with a card: Feitian ePass2003 00 00
    Failed to generate key: Not supported

Bay the way: why is epass2003.profile described as `pkcs15 profile for entersafe`

frankmorgner:
The initial implementation was provided by @viktorTarasov, most recent changes are from @FeitianSmartcardReader. Maybe they can help.

FeitianSmartcardReader:
@majkrzak  We will check the issue and back to you soon, thanks

FeitianSmartcardReader:
@majkrzak The issue already solved, I am working submit pull request of this issue, will keep update, thanks

FeitianSmartcardReader:
@majkrzak please using latest version
https://github.com/FeitianSmartcardReader/OpenSC

**test log**
Hongbins-MacBook-Pro:OpenSC Ben$ pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -l -p 1234 -k --key-type EC:prime256v1 -a myaeskey
Hongbins-MacBook-Pro:OpenSC Ben$ pkcs15-init --generate-key EC:prime256v1 --key-usage sign --auth-id 1
Using reader with a card: FT  ePass2003Auto
User PIN [User PIN] required.
Please enter User PIN [User PIN]: 
Hongbins-MacBook-Pro:OpenSC Ben$ 
Hongbins-MacBook-Pro:OpenSC Ben$ 
[...]

PR #1085, opened 2017-07-03
Solve 2 issues

FeitianSmartcardReader:
1. Latest code has failure when generate RSA key, fix it in sm.c file
2. Solve issue->
https://github.com/OpenSC/OpenSC/issues/1073#issuecomment-309582992, add
ECC support

These code already passed test
https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Testing

### DIFF
--- a/src/libopensc/pkcs15-sc-hsm.c
+++ b/src/libopensc/pkcs15-sc-hsm.c
@@ -956,7 +956,7 @@ static int sc_pkcs15emu_sc_hsm_init (sc_pkcs15_card_t * p15card)
 		r = sc_pin_cmd(card, &pindata, NULL);
 	}
 
-	if (r != SC_ERROR_DATA_OBJECT_NOT_FOUND)
+	if ((r != SC_ERROR_DATA_OBJECT_NOT_FOUND) && (r != SC_ERROR_INCORRECT_PARAMETERS))
 		card->caps |= SC_CARD_CAP_PROTECTED_AUTHENTICATION_PATH;
 
 


Jakuje:
@FeitianSmartcardReader Any possibility to resolve the issue "Failed to erase card: Security status not satisfied" described in the [wiki](https://github.com/OpenSC/OpenSC/wiki/Feitian-ePass2003), without running binary blobs from internet? I have ePass token, but I am unable to us that.

FeitianSmartcardReader:
@Jakuje The issue need using our tool can download from below and to re-format token, if you cannot using this tool, please show me your error or log, I will help check and solve it, thanks http://download.ftsafe.com/files/ePass/Fix_Tool.tar.gz

src/libopensc/sm.c
```diff
@@ -158,7 +158,7 @@ sc_sm_single_transmit(struct sc_card *card, struct sc_apdu *apdu)
 
 	/* send APDU flagged as NO_SM */
 	sm_apdu->flags |= SC_APDU_FLAGS_NO_SM;
-	rv = sc_transmit_apdu(card, sm_apdu);
+	rv = card->reader->ops->transmit(card->reader, sm_apdu);
```
frankmorgner: This change will most likely lead to problems with other cards.

See discussion in https://github.com/OpenSC/OpenSC/issues/970

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1238,7 +1238,7 @@ sc_pkcs15init_init_prkdf(struct sc_pkcs15_card *p15card, struct sc_profile *prof
 		key_info->params.data = &keyargs->key.u.ec.params;
 		key_info->params.free_params = sc_pkcs15init_empty_callback;
 		key_info->field_length = ecparams->field_length;
-		key_info->modulus_length = 0;
+//		key_info->modulus_length = 0;
```
frankmorgner: This change will most likely break other cards

FeitianSmartcardReader: I will talk internal and make the change, thanks

FeitianSmartcardReader:
@frankmorgner Do modify, check changes https://github.com/OpenSC/OpenSC/pull/1085/commits/8021c472762d07b903f26e15bae755d42deeda2e

dengert:
To avoid a strcmp, for every card, The test should be 
`switch(card->type)  {`
    `case:   SC_CARD_TYPE_ENTERSAFE_FTCOS_EPASS2003:`
        `key_info->modulus_length = 0;;`
`}`

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1238,7 +1238,13 @@ sc_pkcs15init_init_prkdf(struct sc_pkcs15_card *p15card, struct sc_profile *prof
 		key_info->params.data = &keyargs->key.u.ec.params;
 		key_info->params.free_params = sc_pkcs15init_empty_callback;
 		key_info->field_length = ecparams->field_length;
-		key_info->modulus_length = 0;
```
frankmorgner: According to ISO 7816-15 (successor of PKCS#15), an `PublicECKeyAttributes` doesn't have a `modulusLength` field (while a `
PublicRSAKeyAttributes` *does* have one):
image
image

What's the reason for this change? As far as I can see, you're not using the `modulus_length` somewhere else.

src/libopensc/sm.c
```diff
@@ -158,7 +158,16 @@ sc_sm_single_transmit(struct sc_card *card, struct sc_apdu *apdu)
 
 	/* send APDU flagged as NO_SM */
 	sm_apdu->flags |= SC_APDU_FLAGS_NO_SM;
-	rv = sc_transmit_apdu(card, sm_apdu);
+	/* if token is epass2003, using below code*/
+	switch(card->type){
+		case SC_CARD_TYPE_ENTERSAFE_FTCOS_EPASS2003:
```
frankmorgner: I'd like to avoid card specific code on this layer. What's the reason for this change?

src/pkcs15init/pkcs15-lib.c
FeitianSmartcardReader: Right, no need do modify code here, we will do it in pkcs15-epass2003.c file, will submit another pull request

src/libopensc/sm.c
FeitianSmartcardReader: We double check the code, no need do modify in this layer, we will modify the file apdu.c.
The reason is our epass2003 series support plain APDU and cipher APDU, while in cipher APDU, after receive 6C command, then we will need get data back and re-package APDU with cipher mode, at this time, the set_le API seems only modify the LE, and nothing change of APDU, so we will modify apdu.c to solve it.

sandys:
@FeitianSmartcardReader im unable to download your fix-tool from http://download.ftsafe.com/files/ePass/Fix_Tool.tar.gz. Can you post the link again please

FeitianSmartcardReader:
check below:
[Fix_Tool.tar.gz](https://github.com/OpenSC/OpenSC/files/1192802/Fix_Tool.tar.gz)

PR #1088, opened 2017-07-04
[coolkey] Copy labels from certificate objects to the keys

Jakuje:
After some testing by users in Fedora who are using Coolkey applets (or their modifications), we noticed missing labels on the keys. In the underlying structures, they are available only in the certificate objects and for usability reasons it would be good to see them also on the keys, as it was in the old Coolkey module.

This PR goes through the emulated PKCS15 objects and if one is missing a label, it tries to search for the matching certificate and copy the label to that key too.

I was trying to reuse the provided functions by the `pkcs15.h`, but comments and improvements are welcomed.


frankmorgner:
Is the label shown in some meaningful application other than `pkcs11-tool`?

Jakuje:
I do not have other specific use case where these labels would be visible. But other cards (PIV, ...) have meaningful labels for all the objects so I don't think coolkey should come with empty labels.

PR #1091, opened 2017-07-07
MyEID support for RAW RSA signature for 2048 bit keys

popovec:
MyEID does not support RAW RSA signature for 2048 bit key.
(Source:  MyEID reference manual 2.1.4)
This hack uses decipher operation for calculating
RAW 2048 bit signature.


frankmorgner:
Looks good. I'll include this when 0.17.0 is done

PR #1097, opened 2017-07-13
jpki: add disable_sign_cert parameter

hamano:
JPKI card has two certificates I've described here: #959
- Certificate for authentication: it's able to read any time.
- Certificate for signature: it's protected by PIN

I just realized that PIN protected certificate will cause harm on sevral browsers, so I added a parameter to disable this.
@frankmorgner  Could you merge the patch before 0.17 release? It's only effect to JPKI card driver.

Setting example:
~~~
    card_driver jpki {
        disable_sign_cert = true
    }
~~~
Thank you.

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -783,6 +783,11 @@ app tokend {
 		#
 		# score = 10;
 	}
+
+	# workaround for PIN protected certificate issue #1097
+	card_driver jpki {
+		disable_sign_cert = true;
+	}
 }
 
 # Used by OpenSC minidriver on Windows only
--- a/src/libopensc/card-jpki.c
+++ b/src/libopensc/card-jpki.c
@@ -82,7 +82,10 @@ jpki_finish(sc_card_t * card)
 	struct jpki_private_data *drvdata = JPKI_DRVDATA(card);
 
 	LOG_FUNC_CALLED(card->ctx);
-
+	if (drvdata->mf) {
+		free(drvdata->mf);
+		drvdata->mf = NULL;
+	}
 	if (drvdata) {
 		free(drvdata);
 		card->drv_data = NULL;
--- a/src/libopensc/card-jpki.c
+++ b/src/libopensc/card-jpki.c
@@ -97,6 +97,9 @@ jpki_init(struct sc_card *card)
 	sc_file_t *mf;
 	int flags;
 
+	scconf_block **blocks, *blk;
+	int i;
+
 	LOG_FUNC_CALLED(card->ctx);
 
 	drvdata = malloc(sizeof (struct jpki_private_data));
@@ -132,6 +135,23 @@ jpki_init(struct sc_card *card)
 	flags = SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_PAD_PKCS1;
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
 
+	for (i = 0; card->ctx->conf_blocks[i]; i++) {
+		blocks = scconf_find_blocks(card->ctx->conf,
+					    card->ctx->conf_blocks[i],
+					    "card_driver", "jpki");
+		if (!blocks) {
+			continue;
+		}
+		blk = blocks[0];
+		free(blocks);
+		if (blk == NULL) {
+			continue;
+		}
+		drvdata->disable_sign_cert = scconf_get_bool(blk, "disable_sign_cert", 0);
+		if (drvdata->disable_sign_cert) {
+			sc_log(card->ctx, "disabled sign cert");
+		}
+	}
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
 
--- a/src/libopensc/jpki.h
+++ b/src/libopensc/jpki.h
@@ -39,6 +39,7 @@ struct jpki_private_data {
 	sc_file_t *mf;
 	int selected;
 	int logged_in;
+	int disable_sign_cert;
 };
 
 int jpki_select_ap(struct sc_card *card);
--- a/src/libopensc/pkcs15-jpki.c
+++ b/src/libopensc/pkcs15-jpki.c
@@ -207,6 +207,43 @@ sc_pkcs15emu_jpki_init(sc_pkcs15_card_t * p15card)
 		if (rc < 0)
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
 	}
+
+	/* add public keys */
+	for (i = 0; i < 2; i++) {
+		static int pubkey_id[2] = { 1, 2 };
+		static const char *jpki_pubkey_names[2] = {
+			"User Authentication Public Key",
+			"Digital Signature Public Key"
+		};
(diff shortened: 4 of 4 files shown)


frankmorgner:
Could you be more specific about the *harm caused on several browsers*?

Does using `onepin-opensc-pkcs11.dll` solve your problem?

hamano:
The Firefox 53 or earlier popup PIN prompt twice before selectiong certificate.
But it seems improved after Firefox 54, only shown PIN prompt of the certificate presented by the server even if not using onepin-.dll.
We have tokend issue raither than browser.
I'll describe this tomorrow...

dengert:
Does the card require the PIN before the certificate can be read? 
If so ignoore this comment. 

pkcs15-jpki.c :77 has `SC_PKCS15_CO_FLAG_PRIVATE` for the certificate. Is this really needed?

Normally only the private keys are marked with SC_PKCS15_CO_FLAG_PRIVATE. Unless the certificate has sensitive information that should not be exposed without entering a PIN in case card is lost or stolen.  T

If the intent is to request the PIN before every sign operation with the sign private key, then 
prkey_obj.user_consent = 1 is the way to tell PKCS\#11 to set the CKA_ALWAYS_AUTHENTICATE attribute for the key

frankmorgner:
@dengert yes indeed, it seems that reading the certificate requires entry of the PIN (https://github.com/OpenSC/OpenSC/pull/959#issuecomment-277866141).

What I'm a bit confused about is that in https://github.com/OpenSC/OpenSC/pull/959#issuecomment-277954362 you are writing that the card has two PINs (Auth/Sign), but in the PKCS#15 emulator, you're only declaring a single pin object. Could this cause confusion in the upper layer? Are you sure that tokend supports multiple PINs?

Even if we cannot better solve this problem than by disabling the signature certificate, I'm not sure whether a driver specific configuration is the best approach for this kind of problem. In PKCS#11 we already have the option to only propagate specific PIN objects (`create_slots_for_pins` aka `onepin-opensc-pkcs11.dll`) ignoring the other PINs (and keys/certificates). I wonder if a similar configuration option in tokend/minidriver would be more useful. So, again, my question @hamano: Does the onepin module fix your problem at least on the PKCS#11 level?

To be honest, I'd like to not (yet) include this change in 0.17.0, because I think that there is either a more generic solution (or even a fix if it's some functionality missing in tokend).

hamano:
This PR solves the problem around OSX tokend.
tokend read certificate in the following order:
1. Authentication Certificate
2. Signature Certificate(PIN protected)
3. Authentication Certificate CA
4. Signature Certificate CA

This is exactly as the code I wrote:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-jpki.c#L62

When an error occurs in reading the "2. Signature Certificate" with Tokend, then aborts reading subsequent certificate.
As a result, since the third certificate is not read, we can not trust the "1. Authentication certificate.
The most effect of this patch is that we can now appare the CA certificate in the keychain.
Actually, CA certificate is not trusted by default, so I need to trust them manually for now.
I will try to trust it automatically like a minidriver.(MD: add msroots file #990)

Also, Google Chrome on OSX seems abort TLS Authentication process when an error occurs in reading PIN protected certificate that is not used TLS Authentication.
If I set "disable_sign_cert = true" I confirmed that error code is changed ERR_BAD_SSL_CLIENT_AUTH_CERT to ERR_SSL_CLIENT_AUTH_SIGNATURE_FAILED.
I've not succeeded yet, I am sure I have approached the success of TLS client authentication on OSX.

For reference, I'm logging browser test here:
https://github.com/jpki/OpenSC/wiki/JPKI-Browser-Test

hamano:
@dengert
> Does the card require the PIN before the certificate can be read?

Yes.

> If the intent is to request the PIN before every sign operation with the sign private key, then
prkey_obj.user_consent = 1 is the way to tell PKCS#11 to set the CKA_ALWAYS_AUTHENTICATE attribute for the key

I agree with the need for prkey_obj.user_consent = 1.
Thanks for your advice.

hamano:
@frankmorgner 
> What I'm a bit confused about is that in #959 (comment) you are writing that the card has two PINs (Auth/Sign), but in the PKCS#15 emulator, you're only declaring a single pin object. Could this cause confusion in the upper layer? Are you sure that tokend supports multiple PINs?

I've declare two pin as PKCS#15 object.
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-jpki.c#L102
Tokend seems support multiple PINs, But it doesn't assume PIN protected certificate.
If our certificate does not protected PIN, it would have no problem at all.

> I wonder if a similar configuration option in tokend/minidriver would be more useful. So, again, my question @hamano: Does the onepin module fix your problem at least on the PKCS#11 level?

I've not found any problems in PKCS#11 level at the moment.
Firefox 54 works good with opensc-pkcs11.dll and onepin-opensc-pkcs11.dll, also pkcs11-tool works.
Even so, I welcome your suggestion that make similar configuration option to tokend/minidriver.

> To be honest, I'd like to not (yet) include this change in 0.17.0, because I think that there is either a more generic solution (or even a fix if it's some functionality missing in tokend).

I leave the decision to you, but this patch doesn't effect by default and It's useful as a temporary workaround.
Thanks.

hamano:
@frankmorgner @dengert

I just succseed TLS client authentication with Safari and Chrome on OS X Sierra!
It need to apply this(OpenSC/OpenSC.tokend#31) and the this PR the set "disable_sign_cert = true"
disable_sign_cert = false(deflaut) is not work.

hamano:
@frankmorgner @dengert
Hi, I tried to resolve the issue by fixing tokend side. And I've tesing to OSX smartcard logon with authentication certificate.
The issue is that when an error occurs in handling pin protected signature certificate, we can not use authentication certificate with  smartcard logon.
First, I skipped handling of the signature certificate that is protected by pin, but this was insufficient(see https://github.com/hamano/OpenSC.tokend/commit/ac0f1d893f9aeb3eb2e03eeea41a9dc109d983c0).
In addition, it was necessary to skip handling private key.(https://github.com/hamano/OpenSC.tokend/commit/8f9cfccb4572faaa7450ba5e3d8f60bfe8b71287)
It seems tokend does not assume private key without a corresponding certificate.
Please merge the tokend branch(https://github.com/hamano/OpenSC.tokend/commits/fix_for_pin_protected_certificate) or this PR. I raher prefer this PR.
Furthermore I would like to set "disable_sign_cert = true" by default in case of via tokend.

frankmorgner:
I'd like to merge https://github.com/hamano/OpenSC.tokend/commits/fix_for_pin_protected_certificate rather than this PR, because I think it tackels the root of the problem, which is Tokend's shortcoming of handling a PIN protected certificate.

In https://github.com/hamano/OpenSC.tokend/commits/fix_for_pin_protected_certificate please additionally add a configuration value of tokend to skip the PIN protected certificate. With default:yes this should fix your problem, right?

hamano:
OK, I just added  ignore_private_certificate parameter for tokend.
Please merge the tokend PR OpenSC/OpenSC.tokend#32
Also I'm requesting OpenSC PR #1114, There is no problem that closing this PR.

Issue #1050, opened 2017-05-19
SC-HSM's PKCS#15 emulator leaks memory


frankmorgner:
The problem can be detected with valgrind:
```
valgrind --leak-check=full pkcs11-tool --test --login
```
...
```
==8571== 880 bytes in 2 blocks are definitely lost in loss record 112 of 117
==8571==    at 0x4C2BBAF: malloc (vg_replace_malloc.c:299)
==8571==    by 0x5274371: sc_pkcs15_decode_prkdf_entry (pkcs15-prkey.c:341)
==8571==    by 0x53864A1: sc_pkcs15emu_sc_hsm_add_prkd (pkcs15-sc-hsm.c:613)
==8571==    by 0x5387DDB: sc_pkcs15emu_sc_hsm_init (pkcs15-sc-hsm.c:968)
==8571==    by 0x5387F92: sc_pkcs15emu_sc_hsm_init_ex (pkcs15-sc-hsm.c:998)
==8571==    by 0x5281E7A: sc_pkcs15_bind_synthetic (pkcs15-syn.c:150)
==8571==    by 0x5268EB6: sc_pkcs15_bind (pkcs15.c:1228)
==8571==    by 0x6686032: ???
==8571==    by 0x6681070: ???
==8571==    by 0x66807CC: ???
==8571==    by 0x667A231: ???
==8571==    by 0x10EAA8: main (pkcs11-tool.c:765)
```
The results above were done with the changes from https://github.com/OpenSC/OpenSC/pull/1004, but the problem should also be present in `master`, since the PR doesn't change the handling of PRKD. @CardContact , please have a look at this because I didn't find a good solution for this problem.

PR #1100, opened 2017-07-17
sc-hsm: Fix memory leak #1050

CardContact:


frankmorgner:
If I remember correctly, I already tried calling `sc_pkcs15_free_prkey_info(key_info)` at the end of this function resulting in a segmentation fault somewhere. How did you verify your changes?

CardContact:
Tested with pkcs11-tool --test, however tests with our PKCS#11 suite failed with SEGV.

The issue is caused by sc_pkcs15emu_object_add() not making a deep but shallow copy of the object. So in sc_pkcs15emu_sc_hsm_add_prkd() it would be sufficient to free the key_info without releasing referenced objects. I'll prepare a new pull request.

PR #1103, opened 2017-07-18
sc-hsm: build without sm/openssl

alonbl:


frankmorgner:
This file builds fine in the Window's *light* flavor, i.e. without OpenSSL. What's the problem with the current implementation?

alonbl:
I am unsure how... physically impossible...
Please notice ENABLE_OPENPACE requires openssl/ includes.
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-sc-hsm.c#L405

try to build with --disbale-sm and notice additional failures.

alonbl:
Maybe you can refine the secure messaging to enable partial functionality in this module. In any case waiting for this to be resolved to be able to bump version in gentoo.

frankmorgner:
please try https://github.com/OpenSC/OpenSC/pull/1104

gentoo is using OpenSC without SM?

alonbl:
#1104 working.
however, the ENABLE_OPENPACE still requires ENABLE_OPENSSL, maybe in different patch but should be applied.

gentoo provides most of what the configure is capable of. if SM is not to be disabled, please remove the option from configure.

Issue #1102, opened 2017-07-18
Reduce memory footprint of configuration parser


frankmorgner:
### Expected behaviour

Reasonable memory consumption.

### Actual behaviour

`pkcs11-tool --test --login` takes about 400 KB of memory, of which most memory is consumed by `scconf_parse()` and friends (via `strdup()`).

### Steps to reproduce

1. `valgrind --tool=massif pkcs11-tool --test --login`
2. `ms_print massif.out.*`

### Logs

```
Command:            pkcs11-tool --test --login
Massif arguments:   (none)
ms_print arguments: massif.out.11963
--------------------------------------------------------------------------------


    KB
382.1^                                                   #                    
     |                                                   #                    
     |                                                   #                    
     |                                                   #                    
     |                                                   #        :           
     |                                                   #        :           
     |                                                   #        :           
     |           ::::::::::::::@@@@@@::::::::::::@:::::::#::::@:::::@:::::@:::
     |         :@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |         :@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |        @:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |       @@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |       @@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |      @@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |     @@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |    @@@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |   @@@@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |  :@@@@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     | ::@@@@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
     |:::@@@@@@:@::     ::    :@     ::    ::    @:    ::#::::@:::::@:::::@:::
   0 +----------------------------------------------------------------------->Mi
     0                                                                   81.00

Number of snapshots: 88
[...]

PR #1123, opened 2017-08-03
Do not call list_init() twice (amend 6bfb39454)

Jakuje:
During rework of error handling of memory allocation (#1020), there was left `list_init()` twice. It was jumping on me in valgrind for awful long time, but it took me some time to understand what was going on.

This commit removes the double-init and should make memory footprint clean.


frankmorgner:
Nice find. However, I don't see any impact on #1102.

frankmorgner:
It's strange that valgrind didn't bring this issue up earlier...

PR #1131, opened 2017-08-16
YubiHSM 2 Beta algorithms listed in human readable form

Jakuje:
The listing of mechanisms in this HSM shows only IDs of algorithms instead of PKCS#11 names.

The PSS mechanisms were defined without the PSS suffix (typo in some of the early commits), the HMACs were missing completely. This is just the issue in `pkcs11-tool`.


mouse07410:
@Jakuje could you tell me how you access YubiHSM2 with OpenSC tools? I'd love to do that.

Jakuje:
It has its own PKCS#11 module/library. But to do some operation, I was able to use `pkcs11-tool`.

mouse07410:
Ah, so you pointed `pkcs11-tool` to `yubihsm_pkcs11.so`?

Jakuje:
> mechtype-0x80000004, keySize={128,128}, hw, wrap

It is vendor specific (Starts with 0x8). Well ... we can say it is vendor specific mechanism (from pkcs11 specification) in `pkcs11-tool`, but I don't think we should assume more.

I have got some older version, which does not seem to get this far (to list this key over PKCS#11).

dengert:
Not only is it vendor specific, it is not not very specific and when used from something like p11-kit it can cause confusion.

For example Mozilla NSS  defines a number of vendor specific PKCS\#11 
./mozilla/security/lib/util/pkcs11n.h
```
#define NSSCK_VENDOR_NSS 0x4E534350 /* NSCP */
#define CKO_NSS (CKO_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKK_NSS (CKK_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKC_NSS (CKC_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKA_NSS (CKA_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKM_NSS (CKM_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKR_NSS (CKM_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
#define CKT_NSS (CKT_VENDOR_DEFINED|NSSCK_VENDOR_NSS)
```
4E534350 is NSCP for Netscape. 

OpenSC  pkcs11-display.c can print some of these (grep for _NETSCAPE). But without some uniqueness added to the define, it can't do much better then just printing the value.  

OpenSC has two vendor_defined attributes it uses internally. But unfortunately it does not add any uniqueness.
./pkcs11-opensc.h
```
#define CKA_OPENSC_NON_REPUDIATION      (CKA_VENDOR_DEFINED | 1UL)
#define CKA_SPKI			(CKA_VENDOR_DEFINED | 2UL)
```

We should add have something like:
```
#define SC_VENDOR_DEFINED 0x4F534300  /* OSC */
#define CKA_OPENSC_NON_REPUDIATION      (CKA_VENDOR_DEFINED | SC_VENDOR_DEFINED | 1UL)
#define CKA_SPKI			(CKA_VENDOR_DEFINED | SC_VENDOR_DEFINED | 2UL)
```

PR #1133, opened 2017-08-18
Use cardholder name in the PIN prompt (PIV)

Jakuje:
The current PIN prompt for PIV cards is not very informative. It says either of the following:
```
token label        : PIV Card Holder pin (PIV_II)
token label        : Global PIN (PIV_II)
```
It answers the simple question if the user should use Global or Application PIN, but it does not say what card is that. Based on the report in RH bugzilla [1] and further discussion on mailing list [2] I decided to make two changes, hopefully to better:

 * Use shorter name for the `PIV Card Holder pin`, such as `App PIN` to make some space for the actual cardholder name (I was considering `Application PIN`, but it would not fit most of the names next to that). The name is abbreviation from `PIV Card Application PIN` referenced in the NIST documentation [3]
 * Replace the `PIV_II` string with the cardholder name (it is usually not important for users to hear that it is PIV card), if there is CN in the first (assuming AUTH) certificate read from the card.

The current prompt for the NIST test cards look like this:
```
token label        : App PIN (Test Cardholder X)
token label        : Global PIN (Test Cardholder VII)
```
This is also simple way for the end-application to fast recognize that the same/different card was inserted (as used in GNOME).

The code is mostly copied from CAC driver, which already does the same.

Yubikey PIV applets work the same way and report what is written in the certificate. In my case, I have some `bar` certificate:
```
token label        : App PIN (bar)
```

Suggestions for better naming or other improvements are welcomed.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1449740
[2] https://sourceforge.net/p/opensc/mailman/message/35898117/
[3] http://csrc.nist.gov/groups/SNS/piv/documents/test-piv-card-data-specifications.pdf

##### Checklist
- [X] Tested with the following card: PIV-II card (NIST test cards, Yubikey NEO)
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/pkcs15-piv.c
+++ b/src/libopensc/pkcs15-piv.c
@@ -613,7 +613,7 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 	char buf[SC_MAX_SERIALNR * 2 + 1];
 	common_key_info ckis[PIV_NUM_CERTS_AND_KEYS];
 	int follows_nist_fascn = 0;
-
+	char *token_name = NULL;
 
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
@@ -765,6 +765,28 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 				sc_pkcs15_free_certificate(cert_out);
 			continue;
 		}
+
+		/* set the token name to the name of the CN of the first certificate */
+		if (!token_name) {
+			u8 * cn_name = NULL;
+			size_t cn_len = 0;
+			static const struct sc_object_id cn_oid = {{ 2, 5, 4, 3, -1 }};
+			r = sc_pkcs15_get_name_from_dn(card->ctx, cert_out->subject,
+				cert_out->subject_len, &cn_oid, &cn_name, &cn_len);
+			if (r == SC_SUCCESS) {
+				token_name = malloc (cn_len+1);
+				if (!token_name) {
+					SC_FUNC_RETURN(card->ctx,
+						SC_ERROR_OUT_OF_MEMORY, r);
+				}
+				memcpy(token_name, cn_name, cn_len);
+				free(cn_name);
+				token_name[cn_len] = 0;
+				free(p15card->tokeninfo->label);
+				p15card->tokeninfo->label = token_name;
+			}
+		}
+
 		/* 
 		 * get keyUsage if present save in ckis[i]
 		 * Will only use it if this in a non FED issued card
--- a/src/libopensc/pkcs15-piv.c
+++ b/src/libopensc/pkcs15-piv.c
@@ -359,7 +359,7 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 	};
 
 	static const pindata pins[] = {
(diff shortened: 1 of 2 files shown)


src/libopensc/pkcs15-piv.c
```diff
@@ -359,7 +359,7 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 	};
 
 	static const pindata pins[] = {
-		{ "01", "PIV Card Holder pin", "", 0x80,
+		{ "01", "App PIN", "", 0x80,
```
mouse07410: Frankly, I don't see any benefit in shortening the prompt. And "App" is far more confusing/ambiguous than "PIV Card Holder".

IMHO, just make the word "pin" all capital ("PIN") and be done here.

mouse07410:
And I'm certainly eager to see a14e5b7 reviewed and merged. Users would appreciate it.

src/libopensc/pkcs15-piv.c
frankmorgner: I'm not familiar with the PIV's terms, but *App* is ambiguous indeed.

dengert:
I have some concerns:

 - PIV CN names tend to be long. Based on the DOE issued certificates I used to work with the CN was "Firstname Lastname (Affiliate)"  The lab was considered a DOE contractor, so  (Affiliate) was added. 
With the few DOD certs I have seen, they used "lastname.Fristname.inital.nnnnnnnnnn.
framework-pkcs11.c then only takes 32 bytes. 
`  strcpy_bp(slot->token_info.label, label, 32);` 

- The token_info.label  may be used in a URI (p11-kit?) or script so some existing scripts may fail.

- The "App PIN " could be could be  NULL, as most cards (only have) and users (only know) one PIN for the card and the Discovery object defines how to present it to the card. 

So if these are not a concern to you, the code looks good.

src/libopensc/pkcs15-piv.c
Jakuje: This part of prompt is shortened to fit the cardholder name into it. It is passed through the Label of the `CK_TOKEN_INFO` structure, which is limited to 32 characters.
http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cs01/pkcs11-curr-v2.40-cs01.html

Jakuje:
@dengert thank you for comments.

 * Yes, I know that the names can be long. From 32 chars we have already 13 gone for `Global PIN` so we are left with ~19. There should most of (at least) first names fit to give some idea whose card is that.

 * The usage in PKCS11 URI was my initial concern and it was why I posted the mail in the mailing list two months ago. So far I didn't get any answer on that so I progressed here. If it is going to change, it should change for better and just once. I would like to avoid some more changes in future so I would not hurry with merging this yet.

 * Using NULL instead of "App PIN" is actually a clever idea, since the General PIN should not be too commonly used. If this will be considered as a good idea by others too, I will try that.

Jakuje:
On the second look, setting "App PIN" to `NULL`, would cause the "PKCS#15 layer" would see unnamed PIN, which does not look like an improvement.

The other solution to avoid ambiguity, would be to name the PIN just "PIN" (see the first modified commit). Possibly, we can also drop this "PIN" from token label, if it is the only thing in there, because it does not bring any useful information, that could be used by the user authenticating to this token.

PR #1141, opened 2017-08-31
Fix sc_pkcs15_get_bitstring_extension endianness

nwf:
Like asn1's decode_bit_field.  However, because we're decoding an
`unsigned long long` here, not an `unsigned int`, we can't just
change the decoder to SC_ASN1_BIT_FIELD.

##### Checklist
- [x] Tested with the following card: YubiKey 4
	- [x] tested PKCS#11

Tested on ar71xx MIPS box.

### DIFF
--- a/src/libopensc/pkcs15-cac.c
+++ b/src/libopensc/pkcs15-cac.c
@@ -139,7 +139,7 @@ cac_alg_flags_from_algorithm(int algorithm)
 
 /* map a cert usage and algorithm to public and private key usages */
 static int
-cac_map_usage(unsigned long long cert_usage, int algorithm, unsigned int *pub_usage_ptr, unsigned int *pr_usage_ptr, int allow_nonrepudiation)
+cac_map_usage(unsigned int cert_usage, int algorithm, unsigned int *pub_usage_ptr, unsigned int *pr_usage_ptr, int allow_nonrepudiation)
 {
 	unsigned int pub_usage = 0, pr_usage = 0;
 	unsigned int alg_flags = cac_alg_flags_from_algorithm(algorithm);
@@ -184,7 +184,7 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 	};
 	/* oid for key usage */
 	static const struct sc_object_id usage_type = {{ 2, 5, 29, 15, -1 }};
-	unsigned long long usage;
+	unsigned int usage;
 
 
 	/*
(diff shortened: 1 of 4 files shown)


frankmorgner:
There are several uses of `SC_ASN1_TAG_BIT_STRING`, did you check the other locations?

Should we check the length of the bit string to match the length of `unsigned long long`?

frankmorgner:
Please fix CI's build error.

nwf:
Auditing the uses of `SC_ASN1_TAG_BIT_STRING` is challenging due to the pervasive use of `sc_format_asn1_entry`, which lexically separates the identifier `SC_ASN1_TAG_BIT_STRING` from the corresponding `.parm` and `.arg` fields of the `struct sc_asn1_entry`.  However, all of them that are paired with `SC_ASN1_BIT_FIELD` are endian-correct because `decode_bit_field` does the right thing (assuming the callers are writing into a machine integer and not a `u8[]`!).  That leaves these uses.

```
src/libopensc/card-gids.c:              { "keyType", SC_ASN1_INTEGER, SC_ASN1_TAG_BIT_STRING | SC_ASN1_CTX, 0, NULL, NULL },
src/libopensc/pkcs15-cert.c:            { "signatureValue",     SC_ASN1_BIT_STRING, SC_ASN1_TAG_BIT_STRING, 0, NULL, NULL },
src/libopensc/pkcs15-cert.c:            { "bitString", SC_ASN1_BIT_STRING, SC_ASN1_TAG_BIT_STRING, 0, value, &val_len },
src/libopensc/pkcs15-pubkey.c:          { "subjectPublicKey", SC_ASN1_BIT_STRING_NI, SC_ASN1_TAG_BIT_STRING, SC_ASN1_ALLOC, NULL, NULL},
src/libopensc/pkcs15-pubkey.c:          { "key",        SC_ASN1_BIT_STRING_NI, SC_ASN1_TAG_BIT_STRING, 0, NULL, NULL },
```

The second `src/libopensc/pkcs15-cert.c` match is the one in question.  Let me look at the rest quickly.
* `"keyType"` is given a `.parm` of a `int *` when it is set later by `sc_format_asn1_entry`, but `sc_asn1_decode_integer` is endian-correct.  (Though its use of a machine-specific `int` type rather than something like `uint32_t` is highly suspect.)
* `"signatureValue`" truly has a `NULL` `.parm` by the time we get to its `sc_asn1_decode` use, so no decode is done.
* `"subjectPublicKey"` decodes into a `struct sc_pkcs15_der`'s `.value` field, a `u8 *`, and so is fine.
* `"key"` decodes into a `struct sc_pkcs15_u8`'s `.value` field, also a `u8 *`, and so is fine.

However, back to the topic at hand: the only uses of `sc_pkcs15_get_bitstring_extension` are to decode OID `2.5.29.15` in `pkcs15-cac.c` and `pkcs15-piv.c`.  This OID is defined by https://tools.ietf.org/html/rfc5280#section-4.2.1.3 to use only bits 0 through 8, and so does not justify an entire `unsigned long long`.  I have kicked it over to using an `unsigned int` and the `SC_ASN1_BIT_FIELD` decoder instead of the fix I originally proposed.  While a more sweeping change, it keeps the endianness glue entirely inside the asn1 decoder.  If there's truly need for a 64-bit integer type decode later, it can be added as a `SC_ASN1_BIT_FIELD64` or somesuch.

nwf:
One more revision to try to appease the build bot.

nwf:
Got a moment to test this patch on my MIPS target and it seemingly works fine; `pkcs11-tool -O` shows the correct key usage bits.

frankmorgner:
Thanks for the detailed analysis!

@vletoux could you look at `card-gids.c`?

vletoux:
I do not have any big endian computer to test on it.
the keyType just put one byte on it so I do not think that the endianness matters here.

You're free to adapt the code and change the asn1 encoding flag accordingly if you like
(when designing I copy paste existing code)

frankmorgner:
OK, thanks for the feedback!

PR #1143, opened 2017-08-31
Use |, not ||, when decoding CAC key_reference

nwf:
This quiets two compiler warnings and is almost surely what was actually meant.  However, I do not have a CAC card to test.


frankmorgner:
Looks good. @Jakuje?

PR #1150, opened 2017-09-13
pkcs11-tool testsuite improvements

Jakuje:
The randomization should not happen for the `RSA_PKCS` mechanism, since the data is already filled with ASN1 encoded DigestInfo + SHA1 hash.

The second commit improves handling of decryption testcases to get reasonable reports in single stream (not writing first part of line to stdout and the second to stderr) and avoids hard-failures with OAEP mechanisms (`CKR_MECHANISM_PARAM_INVALID`).

##### Checklist
- [X] Tested with the following card: YubiHSM2 with its own PKCS#11 module
	- [X] tested PKCS#11

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -4321,13 +4321,13 @@ static int test_signature(CK_SESSION_HANDLE sess)
 		break;
 	case CKM_RSA_X_509:
 		dataLen = modLenBytes;
+		pseudo_randomize(data, dataLen);
 		break;
 	default:
 		dataLen = sizeof(data);	/* let's hope it's OK */
+		pseudo_randomize(data, dataLen);
 		break;
 	}
-
-	pseudo_randomize(data, dataLen);
 
 	if (firstMechType == CKM_RSA_X_509) {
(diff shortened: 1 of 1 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -4847,7 +4847,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		return 0;
 
 	if (EVP_PKEY_size(pkey) > (int)sizeof(encrypted)) {
-		fprintf(stderr, "Ciphertext buffer too small\n");
+		printf("Ciphertext buffer too small\n");
```
frankmorgner: why print errors on stdout instead of stderr?

Jakuje: It is not an error, but part of the "testsuite report", which starts on line 4843, does not end with a newline and expects some result to be added (which is in this case "error").

    printf("    %s: ", p11_mechanism_to_name(mech_type));

We can go the other way round and print everything on the stderr too. But if part is printed to the first stream and the second part is printed to the second stream, the formatting is undefined.

frankmorgner:
OK, thanks for the clearification

Issue #1159, opened 2017-09-28
problems with minidriver and newer feitian cards


VelinPavlov:
### Problem Description
Hello. I have several generations of Feitian-manufactured smart cards, all supported by opensc. Types reported by opensc are 19002, 19003 (older ones, supported by opensc-pre-0.17) and 19004 (recent, supported only by opensc-0.17). I use them for a local PKI, and they all work fine with Mozilla and LibreOffice, on both Linux and Windows platforms.
My problem is that I need to be able to use them with Microsoft office, e.g. with minidriver.dll.

On Windows, certutil reports types 19002 and 19003 as ok (there are problems with trust-chain, OIDs, but this is not related to opensc). 19004 is not working with minidriver.dll - error message is "file not found".

### Proposed Resolution
I believe the problems are related to minidriver.dll not being built with support for the newer feitian cards. IMHO, changes are needed in 2 places - src/minidriver/minidriver-feitian.reg and win32/customactions.cpp to add the new ATR/ATRMask values for the newer 19004 == SC_CARD_TYPE_ENTERSAFE_EJAVA_PK_01C, as defined in src/opensc/card-entersafe.c. I can do the changes, but need some advice/approval by more knowledgeable people and a platform to build (nightly builds??). I can use local platforms / cards to test the builds.

### Steps to reproduce
standard installation of opensc-0.17 .MSI and testing as described in https://github.com/OpenSC/OpenSC/wiki/Windows-Quick-Start. Testing needs cards. I can do it locally.

### Logs
will suppliy if needed

VelinPavlov:
Sorry, my report is against .MSI for the released version. I now see that minidriver has been through a lot of changes. I'd expect that nightly-build minidriver would not work with any of my cards, but have not tested it. Is minidriver support for feitian cards being removed?

VelinPavlov:
Just for info: the nightly build works with the old cards and not with the newer ones. IMHO this is because feitian-support-code is not removed from customactions.cpp, just the registry-adding-files are removed. As I installed the nightly build _over_ the released-build,  the registry entries are from there. I then deleted manually the registry entries - all cards stopped working. I then 'repaired' the installation, and the registry entries did not re-appear again, so none of the cards works, as expected.

VelinPavlov:
Using 0.17-released we were able to manually add registry entries and MS-word is now willing to sign using feitian cards, both older and newer. After entry of the correct PIN, MS-word says it cannot sign. 

[...]

PR #1168, opened 2017-10-11
add newer feitian cards to windows-registry auto-update-on-install

VelinPavlov:
issue #1159 has several aspects:

a. registry entries for newer-feitian-cards have to be added manually. This commit should fix that. (verify (after build) that
0. this build is not a regression from released (in my specific environment);
1. M$-word is able to see certificates on both older and newer cards;
2. most probably M$-word will refuse to sign, see b. below).

b. In order for M$-word to be able to sign, some bug(s) have to be fixed in card-caching code. A next commit will 'fix' that by simply avoiding data / code that uses card-caching.

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend


frankmorgner:
Though I doubt the usefullness of the internal path caching, the problem needs further analysis by @FeitianSmartcardReader in #1159 . The problem should then be fixed locally, instead of globally disabling the cache.

PR #1169, opened 2017-10-12
Add support in pkcs11-tool and pkcs11-spy for RSA-OAEP

mouse07410:
And adds support for SHA224 to RSA-PSS

Addresses #1170 

##### Checklist

- [X] Tested with the following card: YubiKey NEO, YubiHSM2
	- [X] tested PKCS#11
	- [X] tested macOS Tokend

@frankmorgner @Jakuje asking for your attention.


dengert:
Suggest changing the name of PR to "Add support in pkcs11-tool and pkcs11-spy for RSA-OAEP"
It does not (yet) add any support to libopensc. 

Also note card-gids.c and authentic.h imply the card can support OAEP.

mouse07410:
> Suggest changing the name of PR to "Add support in pkcs11-tool and pkcs11-spy for RSA-OAEP"

Done, thanks.

> It does not (yet) add any support to libopensc.

The more I think about this, the more I tend to believe that it can't be added to `libopensc`. Because AFAIK, `libopensc` is a set of card *drivers* that "enable" or "comprise" PKCS#11 interface.

In this case, `yubihsm_pkcs11.so` takes place of `opensc-pkcs11.so` that uses `libopensc.so` to provide PKCS#11 interface to a card. So, one either has to *re-implement* `yubihsm_pkcs11.so` and add that code to `libopensc`, or just use `yubihsm_pkcs11.so` (or similar, for other such devices). I don't see a generic solution (add OAEP and/or PSS support to `libopensc` in general), and don't have enough info (or even desire) to re-implement `yubihsm_pkcs11`...

> Also note card-gids.c and authentic.h imply the card can support OAEP.

I'm not sure I follow. And I'm not sure at all whether it implies the card supports OAEP "fully" (aka does all the OAEP padding steps internally), or splits the implementation with the driver.

doc/tools/pkcs11-tool.1.xml
```diff
@@ -75,8 +75,16 @@
...
+					<listitem><para>
+	  Specify hash algorithm used with RSA-PKCS-PSS signature or RSA-OAEP decryption.
+          Allowed values are "SHA-1", "SHA256", "SHA384", "SHA512". Default is SHA-1.
+	  Note that the input to RSA-PKCS-PSS has to be of
+          the size equal to the specified hash algorithm. E.g., for SHA256 the signature input must be
+          exactly 32 bytes long (for mechanisms SHA256-RSA-PKCS-PSS there is no such restriction).
+          For RSA-OAEP, the plaintext input size mLen must be at most
+          keyLen - 2 - 2*hashLen. For example, for RSA 3072-bit key and SHA384, the longest plaintext
+          to encrypt with RSA-OAEP is (with all sizes in bytes): 384 - 2*48 -2 = 286, aka 286 bytes.
+          </para></listitem>
```
Jakuje: All the other lines are indented under the opening tag, which makes it look like waste of space on the screen, but makes it possible to orient inside of the file for future editors. I would recommend to indent also this block with 5 tabs.
Possibly split the paragraphs into two separate paragprahs using `<para>` tag should do the job.
And just a nitpick, there should come space in front of the two in term `-2`.

src/pkcs11/pkcs11-spy.c
```diff
@@ -844,6 +844,26 @@ C_DecryptInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechanism, CK_OBJECT
 	enter("C_DecryptInit");
 	spy_dump_ulong_in("hSession", hSession);
 	fprintf(spy_output, "pMechanism->type=%s\n", lookup_enum(MEC_T, pMechanism->mechanism));
+	switch (pMechanism->mechanism) {
+	case CKM_RSA_PKCS_OAEP:
+		if (pMechanism->pParameter != NULL) { /* XXX assuming RSA-OAEP parameter */
```
Jakuje: This comment is not needed anymore, since the mechanism is already checked. I would remove it to avoid confusing further readers.

src/tools/pkcs11-tool.c
```diff
@@ -1870,6 +1889,89 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+		default: 
+			oaep_params.hashAlg = CKM_SHA_1;
+			/* fallthrough to SHA-1 default */
+		case CKM_SHA_1:
+			oaep_params.mgf = CKG_MGF1_SHA1;
+			break;
+		}
+		break;
+
+#if 0 /* we do not have these definitions yet! */
```
Jakuje: Are we ever going to have them? I do not see its specification in current  PKCS#11 nor in the future work-in-progress version. I would be for simple removing of this unused code for now.

src/tools/pkcs11-tool.c
```diff
@@ -1870,6 +1889,89 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+		oaep_params.hashAlg = CKM_SHA512;
+		oaep_params.mgf = CKG_MGF1_SHA512;
+		break;
+#endif
+	case CKM_RSA_PKCS:
+		mech.pParameter = NULL;
+		mech.ulParameterLen = 0;
+		break;
+	default:
+		util_fatal("Illegal mechanism %s for RSA-OAEP\n", p11_mechanism_to_name(opt_mechanism));
```
Jakuje: isn't this `fatal` mismatched to wrong `switch` or just the message is wrong? The switch lists decryption mechanisms, RSA_OAEP and RSA_PKCS are already "caught", so this should be something like "unknown decryption mechanism", or something similar?

src/tools/pkcs11-tool.c
```diff
@@ -1870,6 +1889,89 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+	}
+
+
+	/* One of RSA-OAEP mechanisms above: They need parameters */
```
Jakuje: There is only one RSA-OAEP at this moment.

dengert:
@mouse07410  In response to the second half of your https://github.com/OpenSC/OpenSC/pull/1169#issuecomment-336709994   OpenSC is designed to support many cards were some of them can do complex operations on the card, where others can not. Libopensc and its drivers work together to implement these complex operations in hardware **or** in software where needed. RSA-PKCS1 is one example. There is no reasons libopensc could not also implement RSA-OAEP. This would require the parameters be passed to from the PKCS\#11 code to the libopensc and to card drivers that could support it.  If the card  does not support OAEP but does support RSA RAW, libopensc could do the padding in software. It already does this in software for other padding methods. 

But with other RSA padding methods it can also remove the padding to get the digest or even the hash so it can let the card **redo** the padding in hardware. But this will not work for RSA-PSS and RSA-OAEP One needs the parameters to be able to recover the hash.  

RSA-PSS and RSA-OAEP are different in that they are the only to RSA mechanisms that take a parameters. 

So to implement RSA-PSS and RSA-OAEP in libopensc would take some major changes in that the parameters have to be passed in from PKCS\#11 to the PKCS\#15 layer in libopensc then to the drivers if needed. 

In you case there is no OpenSC driver for the HSM, so it would require a new card driver. But  card-gids.c hints that it has some support for RSA-OAEP. Likewise card-sc-hsm.c hints at RSA-PSS support on the card. 

@mouse07410  I am not suggesting that you try and implement a card driver for your HSM if you can use the vendor's PKCS\#11 module. What I was suggesting is that it is possible to do that or to add support for two cards that hint they support  RSA-PSS and/or RSA-OAEP. 

What I was suggesting is that others developers of the other cards may want to to add support for 
RSA-PSS and RSA-OAEP.

doc/tools/pkcs11-tool.1.xml
mouse07410: Addressed, thanks.

src/pkcs11/pkcs11-spy.c
mouse07410: Addressed, thanks!

src/tools/pkcs11-tool.c
mouse07410: I do not know. But this code doesn't charge us for either space or compilation time - so I'd rather keep it there. It's never too late to remove it, IMHO.

src/tools/pkcs11-tool.c
mouse07410: Yep, thanks. Addressed.

src/tools/pkcs11-tool.c
mouse07410: Yeah...? ;-)

mouse07410:
> There is no reasons libopensc could not also implement RSA-OAEP. This would require the parameters be passed to from the PKCS#11 code to the libopensc and to card drivers that could support it. If the card does not support OAEP but does support RSA RAW, libopensc could do the padding in software. It already does this in software for other padding methods.

You are correct. However, as you yourself pointed out:

> ...to implement RSA-PSS and RSA-OAEP in libopensc would take some major changes in that the parameters have to be passed in from PKCS#11 to the PKCS#15 layer in libopensc then to the drivers if needed.

I am not keen to undertake such a major effort (no time, no deep enough understanding of OpenSC for that, no real need at this point). And, as you pointed out above, it would probably need cooperation of a card driver.

> What I was suggesting is that it is possible to do that or to add support for two cards that hint they support RSA-PSS and/or RSA-OAEP.

I'm sure it is. But my ambitions for this PR are far more modest. All I want is for the main master to support OAEP in `pkcs11-tool`, the same way it now supports PSS. Others who come after me, may add capabilities you described. ;-)

> What I was suggesting is that others developers of the other cards may want to to add support for
RSA-PSS and RSA-OAEP.

I completely agree. But that (probably distant) future shouldn't stand in the way of adding the feasible real capability to `pkcs11-tool` now.

src/pkcs11/pkcs11.h
```diff
@@ -764,6 +764,16 @@ typedef struct CK_ECDH1_DERIVE_PARAMS {
 	unsigned char *  pPublicData;
 } CK_ECDH1_DERIVE_PARAMS;
 
+typedef unsigned long CK_RSA_PKCS_OAEP_SOURCE_TYPE;
+
+typedef struct CK_RSA_PKCS_OAEP_PARAMS {
+  CK_MECHANISM_TYPE hashAlg;
+  CK_RSA_PKCS_MGF_TYPE mgf;
+  CK_RSA_PKCS_OAEP_SOURCE_TYPE source;
+  void *pSourceData;
```
frankmorgner: please use tabs for indenting

src/tools/pkcs11-tool.c
frankmorgner: I agree with @Jakuje.

mouse07410:
@frankmorgner change requests addressed.

mouse07410:
@frankmorgner...? ;-)

frankmorgner:
Thanks!

@Jakuje did you find time to test this with your HSM?

mouse07410:
> @Jakuje did you find time to test this with your HSM?

@Jakuje 's HSM is the same as mine. So a difference is unlikely (unless he's running on a different OS) ;-)

Jakuje:
Just checked the [minimal use case](https://gist.github.com/Jakuje/5a993d2b2d8a9cac35203599e49e6831#rsa-pkcs-oaep) with default parameters and it seems to be working as expected:

```
[jjelen@t470s tools ((a8d3f351...))]$ openssl rsautl -encrypt -inkey $ENC_KEY.cert -keyform DER -in data -pubin -out data.crypt -oaep
[jjelen@t470s tools ((a8d3f351...))]$ cat data.crypt | ./pkcs11-tool --id $ENC_KEY --decrypt -p $PIN -m RSA-PKCS-OAEP --module /home/jjelen/Downloads/yubihsm-0.2.0-sdk/linux-amd64/fedora25/lib/yubihsm_pkcs11.so
Using decrypt algorithm RSA-PKCS-OAEP
OAEP parameters: hashAlg=SHA-1, mgf=MGF1-SHA1, source_type=0, source_ptr=(nil), source_len=0
Data to encrypt

[jjelen@t470s tools ((a8d3f351...))]$ openssl pkeyutl -encrypt -inkey $ENC_KEY.cert -keyform DER -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha256 -pkeyopt rsa_mgf1_md:sha256 -in data -pubin -out data.crypt
[jjelen@t470s tools ((a8d3f351...))]$ cat data.crypt | ./pkcs11-tool --id $ENC_KEY --decrypt -p $PIN -m RSA-PKCS-OAEP --hash-algorithm=sha256 --module /home/jjelen/Downloads/yubihsm-0.2.0-sdk/linux-amd64/fedora25/lib/yubihsm_pkcs11.so
Using decrypt algorithm RSA-PKCS-OAEP
OAEP parameters: hashAlg=SHA256, mgf=MGF1-SHA256, source_type=0, source_ptr=(nil), source_len=0
Data to encrypt
```
Only thing I see missing is the the possibility to use the `source` type different than the default `0`, but I didn't get so far to check if it is possible to encrypt with OpenSSL with these arguments nor that YubiHSM2 supports them.

I believe we are good.

mouse07410:
> Only thing I see missing is the the possibility to use the source type different than the default 0,

Correct. 

> but I didn't get so far to check if it is possible to encrypt with OpenSSL with these arguments nor that YubiHSM2 supports them.

I've no idea whether YubiHSM2 supports `source`. Probably yes, though maybe not. ;-)

OpenSSL internals do support `source`, but the CLI interface doesn't. And I don't recall seeing programmatic API to set it to anything besides the default 0. In any case, I do not plan to add support for `source` at this time. It would have to come later, after this PR is done with.

mouse07410:
Thank you!

PR #1171, opened 2017-10-13
another attempt at #1159

VelinPavlov:
Previous build, disabling caching in card.c, had side effects (older cards stopped working).
I patched / compiled debian source (0.17-release); disabling in card.c has the same effect.
This patch should disable only path-caching, directly in card-entersafe.c. I tested it on my debian/testing platform as much as I could with all types of cards I have.
Need to test on Windows + MS-word signatures.

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend


VelinPavlov:
Sorry, I was too tired and have misunderstood my local test procedure. This patch should NOT be integrated.

PR #1172, opened 2017-10-15
card-entersafe.c code cleanup; there seems to be an excessive else

VelinPavlov:
rationale: default action is select-path-regardless-of-caching; implementation will not do that action in some specific cases (inner-outer-if-then-else-logic seems wrong).

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend


VelinPavlov:
At a second glance, the case is handled. I'll check further if it correct.

PR #1173, opened 2017-10-16
split entersafe_select_path into ..._old and ..._new

VelinPavlov:
rationale: split entersafe_select_path into ..._old and ..._new to have finer control over code;
also in this changeset: cleanup of card-entersafe.c atrs and related card types in cards.h (this may be a regression, but I still have no response from @FeitianSmartcardReader why they were present).

##### Checklist

- [ ] Documentation is added or updated (not yet, will do if necessary)
- [ ] New files have a LGPL 2.1 license statement (no new files; patches only)
- [x] Tested with the following card: entersafe (several generations, with different ATRs, behaving differently)
	- [x] tested PKCS#11
	- [x] tested Windows Minidriver, but see attached text
	- [ ] tested macOS Tokend


VelinPavlov:
[issue#1159.txt](https://github.com/OpenSC/OpenSC/files/1387260/issue.1159.txt)

VelinPavlov:
Tests on WIndows were done by my colleague. He told me build 1456 works at pkcs#11 level with both old and new cards, and MS-word can sign with new ones. I've put some logging code in a block that manipulates the call parameters with some local overrides (e.g. forcing 0x3F, 0x00 at the beginning of the path). I did not see that block being executed in my local tests, will check that tomorrow on windows - log might provide some clue.

So some little changes are to be done - the auto-registry-update on-install issue plus a small fix for the failing build on travis (it's a type mismatch I could not catch locally). As I don't need the artifacts of the build, I will submit these changes together with more important tests.

After that, I intend to look deeper into why MS-word cannot sign with the older cards - IMHO it should be possible. 

thanks

PR #1178, opened 2017-10-29
Treat Warnings as Errors

frankmorgner:
This PR fixes many more compiler warnings (gcc, clang, msvc) and some warnings from the clang's static code analyizer. In fact, all compiler warnings are gone now (though some are hidden on msvc with `_CRT_SECURE_NO_DEPRECATE` and `_CRT_NONSTDC_NO_WARNINGS`).

Anyway, we can now force **warning free compilation** on all platforms by default!

##### Checklist

- [x] Tested with the following card: CardOS 4.3, GoID (sc-hsm) 
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend


Jakuje:
Thanks for fixing all these bugs!

But with the new compiler options, I am unable to build OpenSC with Fedora 26, with GCC 7.2.1 on false positive errors, such as:
```
In file included from pkcs15-iasecc.c:48:0:
pkcs15-iasecc.c: In function ‘iasecc_file_convert_acls’:
../libopensc/log.h:126:45: error: this statement may fall through [-Werror=implicit-fallthrough=]
 #define SC_TEST_RET(ctx, level, r, text) do { \
                                             ^
../libopensc/log.h:134:36: note: in expansion of macro ‘SC_TEST_RET’
 #define LOG_TEST_RET(ctx, r, text) SC_TEST_RET((ctx), SC_LOG_DEBUG_NORMAL, (r), (text))
                                    ^~~~~~~~~~~
pkcs15-iasecc.c:338:5: note: in expansion of macro ‘LOG_TEST_RET’
     LOG_TEST_RET(ctx, SC_ERROR_NOT_SUPPORTED, "'IDA' not actually supported");
     ^~~~~~~~~~~~
pkcs15-iasecc.c:339:4: note: here
    case SC_AC_SCB:
    ^~~~
```
I would be for postponing this change until GCC will have better algorithm to recognize false-positives. Or use in the above example some other construction.

frankmorgner:
I don't think we should rely on some gcc timeline. Rewriting the code that you've pointed to above would be cleaner (to be honest, it looks a bit hacky). You can also use `./configure --disable-strict`.

Jakuje:
Thank you for the pointer. I didn't know about this configure switch. I will have a look if there is more of these issues and submit a PR. Thanks.

PR #1183, opened 2017-11-03
Enable CAC ALT token card operations

Jakuje:
The CAC ALT tokens that we encountered recently were not working as expected. For the record, these were the cards identified as the following in the Windows ActivClient:
```
Manufacturer            : Giesecke & Devrient
Model                   : SmartCafe Expert v7.0 144K DI
```
These cards do not have CCC object, which is partially against the GSC-IS specification and the ACA object is on different address than all the specifications suggest. The first step to make them working at least as a read-only was the PR #1120, but after that I was still not able to perform key operations. Though I was able to do so after selecting the ACA applet before calling the VERIFY APDU, which is the fix here, though I am not sure if it is the cleanest way to do that -- I hooked the applet selection after the object listing. If there is a way to prepend it before the PIN VERIFY command somehow, it would be cleaner. Improvements welcomed.

There could be a cleaner way of parsing all the structures stored in the ACA object, but it would be a lot of code and work for a card without any bright future (CAC is being replaced by PIV, but the process is very slow).

Red Hat bugzilla for more information:
https://bugzilla.redhat.com/show_bug.cgi?id=1473418

##### Checklist

- [X] Tested with the following card: 
	- [X] tested PKCS#11


frankmorgner:
Have you tried adding a path (with your AID) to the PKCS#15 PIN object? The AID [should then be selected before PIN verification](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-pin.c#L441-L446)...

Jakuje:
Thanks for pointer. I probably missed this part.
That looks like exactly the thing I will need. I do not have access to the card at this moment, but I will try it latest next week and update the PR if it will work.

dengert:
"'(CAC is being replaced by PIV, but the process is very slow)."

Yes very slow. HSPD-12 was from 2004, 13 years ago...
https://www.dhs.gov/homeland-security-presidential-directive-12

With NIST  SP 800-73 - "April 2005 Initial Release"

OpenSC had PIV driver in 2005.

Jakuje:
I tested the above changes with both the CAC ALT token and other already working CAC cards successfully. Only rough edge is that we can not select this path for all CAC cards, since some of them do not have it. Therefore there is a new communication from PKCS#15 layer to find out if this path should be used (and it is used only if the card was recognized by this path).

Also our coverity instance threw a warning about other code that I was building so I attached it also.

Also I am testing this card on a server without display, remotely and by using a master, I see two errors from the the notification code.
```
Error connecting: Cannot autolaunch D-Bus without X11 $DISPLAY
```
I believe there should be some condition that if we do not have `$DISPLAY`, we should not attempt to show the notification on Linux, but it is unrelated to this PR. I will either fill a new PR or just an issue to keep a track of it.

PR #1187, opened 2017-11-07
opensc-notify: register a desktop menu entry

### LINKED ISSUE #1186: Running without $DISPLAY shows error from notification code
### Problem Description

Running `pkcs11-tool` on machine without `$DISPLAY`/without X11 server running throws the following bogus errors, while attempting to show the smartcard notifications:
```
Error connecting: Cannot autolaunch D-Bus without X11 $DISPLAY
```
### Proposed Resolution
On Linux, the notification code should not attempt to show notification if there is no `$DISLPAY` environment variable set.

### Steps to reproduce
Connect remotely to machine with smart card and without display and list the PKCS#11 objects (without [PCSC access control](https://access.redhat.com/blogs/766093/posts/1976313)):
```
./pkcs11-tool -O
```
Unfortunately it is not possible to reproduce the issue by unsetting `$DISPLAY`.

frankmorgner:
- fixes showing notifications in gnome-shell via gio2
- removes gdbus interface for notifications
- fixes #1186

##### Checklist

- [x] Tested with the following card: sc-hsm
	- [x] tested PKCS#11

### DIFF
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -7,6 +7,7 @@ VDFORMAT=GZIP
 endif
 
 do_subst = $(SED) \
+	   -e 's,[@]bindir[@],$(bindir),g' \
 	   -e 's,[@]CVCDIR[@],$(CVCDIR),g' \
 	   -e 's,[@]PACKAGE[@],$(PACKAGE),g' \
 	   -e 's,[@]PACKAGE_BUGREPORT[@],$(PACKAGE_BUGREPORT),g' \
@@ -24,7 +25,7 @@ OPENSC_NOTIFY_BUILT_SOURCES = opensc-notify-cmdline.h opensc-notify-cmdline.c
 OPENSC_ASN1_BUILT_SOURCES = opensc-asn1-cmdline.h opensc-asn1-cmdline.c
 
 MAINTAINERCLEANFILES = $(srcdir)/Makefile.in $(srcdir)/versioninfo-tools.rc $(srcdir)/versioninfo-opensc-notify.rc
-EXTRA_DIST = Makefile.mak versioninfo-tools.rc.in versioninfo-opensc-notify.rc.in npa-tool.ggo.in npa-tool.1 opensc-notify.ggo.in egk-tool.ggo.in egk-tool.1 opensc-asn1.ggo.in opensc-asn1.1
+EXTRA_DIST = Makefile.mak versioninfo-tools.rc.in versioninfo-opensc-notify.rc.in npa-tool.ggo.in npa-tool.1 opensc-notify.ggo.in egk-tool.ggo.in egk-tool.1 opensc-asn1.ggo.in opensc-asn1.1 org.opensc-project.opensc-notify.desktop.in
 
 noinst_HEADERS = util.h fread_to_eof.h
 noinst_PROGRAMS = sceac-example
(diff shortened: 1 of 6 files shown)


frankmorgner:
@Jakuje could you check whether this fixes your problem?

Jakuje:
No. With this patch, it fails the show the notification altogether with the following error even if I have a display:
```
(process:18872): GLib-GIO-CRITICAL **: g_application_send_notification: assertion 'G_IS_APPLICATION (application)' failed
```
But I was testing just from the local build without installation (though the current master work with this setup properly).

frankmorgner:
I forgot to commit a change that was present locally. Sorry, for the inconvenience.

Jakuje:
This does not throw error messages, but does not show the notification either in GNOME as it was showing before. Does it require to install the desktop entry somewhere?

frankmorgner:
Yes, gnome shell requires you to put it into one of `$XDG_DATA_DIRS`'s `applications` directories (e.g. putting `org.opensc-project.opensc-notify.desktop` into `~/.local/share/applications/` will do)

src/tools/Makefile.am
```diff
@@ -178,5 +179,11 @@ gids_tool_SOURCES += versioninfo-tools.rc
 opensc_notify_SOURCES += versioninfo-opensc-notify.rc
 endif
 
+applicationsdir = $(datadir)/applications
+applications_DATA = org.opensc-project.opensc-notify.desktop
+
+org.opensc-project.opensc-notify.desktop: org.opensc-project.opensc-notify.desktop
```
Jakuje: This should say (adding `.in` to the end):
```
org.opensc-project.opensc-notify.desktop: org.opensc-project.opensc-notify.desktop.in
```

Jakuje:
For some reason, I don't have `~/.local/share` in `$XDG_DATA_DIRS`. But even after putting it in into `/usr/share/applications/` and after running `desktop-file-validate` with this desktop file I do not see any notification nor any error. Hard to say what could go wrong.

Looking into the configure log, it looks like everything is defined as expected. Not sure what else I could check or if the debug log would be useful.

frankmorgner:
Strange... have you tried a different desktop environment? AFAIK, gnome shell is the only one that requires you to have a desktop entry.

Jakuje:
I don't have any other DE installed and set up to try at this moment.

I tried to intercept the notification using DBus, but I did not see them either according to the following post but nothing happens (for example using `pkcs11-tool`):

https://askubuntu.com/questions/105566/is-there-a-way-to-view-notification-history/352125#352125

On the other hand, I can see the same failures
for other tools using `libopensc`, for example if I try  `tests/pintest`, which is not improvement either (before applying this patch, it was showing the notification properly even from here).
```
(process:20090): GLib-GIO-CRITICAL **: g_application_send_notification: assertion 'G_IS_APPLICATION (application)' failed
```
There should be some check, that the notify is initialized, before trying to show notification, should not it?

I even tried to run the example notification program using GIO2 from archlinux, but still no notification:

https://wiki.archlinux.org/index.php/Desktop_notifications

Though I am able to get the notifications using `libnofity` example program and I am getting the notifications from other programs on my desktop. I am probably out of ideas where to search or what to try so at least these ideas so far.

src/tools/Makefile.am
frankmorgner: done, thanks

frankmorgner:
I had the same problem for quite some time, that's why I've implemented the `gdbus` workaround. But as explained above, I solved the problem with a desktop file for gnome-shell. I hope that the default install location I've used will work on other distributions, too...

I've added the check on whether the notifications are initialized.

frankmorgner:
I've noticed that I've got the following in my shell startup configuration:
```
export XDG_DATA_DIRS=$XDG_DATA_DIRS:$HOME/.local/share
```
Setting this may let xdg find your desktop file.

PR #1189, opened 2017-11-08
Fix crash when certificate read failed

metsma:
##### Checklist
- [x] Tested with the following card: EstEID 3.5 cold
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend


frankmorgner:
thanks

PR #1196, opened 2017-11-10
Avoid GCC 7 warnings with -Werror

Jakuje:
The PR #1178 from @frankmorgner introduced treating all warnings as errors. But my GCC7 on current Fedora 26 has some more checks that needed to be addressed to make it build for me. I am not completely convinced that all the changes I made are correct/better so I would be glad for review/comments/improvements if some of the parts can be done better or if I misunderstood some of the unclear cases.

```
-Werror=implicit-fallthrough=
	libopensc/card-incrypto34.c
		not sure if this is a bug or intention
	libopensc/card-rutoken.c
		most probably intention
	libopensc/card-westcos.c
		remove bogus if so the compile is not confused
		I will fill a separate bug to gcc probably
	pkcs15init/pkcs15-iasecc.c
		Simplify the log and avoid compiler confusion
	sm/sm-common.c
		explicit fallthrough
	tools/pkcs11-tool.c
		use explicit fallthrough comment
	tools/pkcs15-init.c
		The fallthrough is obvious here

-Werror=format-truncation=
	libopensc/pkcs15-itacns.c
		use explicit string lengths
	pkcs11/framework-pkcs15.c
		calculate the truncation
	tests/pintest.c
		avoid sprintf
	tools/pkcs15-crypt.c
		avoid sprintf
	tools/pkcs15-init.c
		calculate the truncation
```

##### Checklist

- [ ] Tested with the following card: PIV, CAC
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-incrypto34.c
+++ b/src/libopensc/card-incrypto34.c
@@ -357,7 +357,7 @@ static int incrypto34_create_file(sc_card_t *card, sc_file_t *file)
 				 * generation. */
 			case SC_FILE_EF_LINEAR_VARIABLE_TLV:
 				type[1] = 0xff;
-				/* FALLTHROUGH */
+				/* fall through */
 			default:
 				type[0] |= file->ef_structure & 7;
 				break;
--- a/src/libopensc/card-incrypto34.c
+++ b/src/libopensc/card-incrypto34.c
@@ -357,6 +357,7 @@ static int incrypto34_create_file(sc_card_t *card, sc_file_t *file)
 				 * generation. */
 			case SC_FILE_EF_LINEAR_VARIABLE_TLV:
 				type[1] = 0xff;
+				/* FALLTHROUGH */
 			default:
 				type[0] |= file->ef_structure & 7;
 				break;
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1072,7 +1072,7 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 		else   {
 			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
 				int ti_len = MIN(strlen(p15card->tokeninfo->label),
-					sizeof(label) - strlen(auth->label) - 3);
+					sizeof(label) - strlen(auth->label) - 4);
 				snprintf(label, sizeof(label), "%.*s (%.*s)",
 					(int) strlen(auth->label), auth->label,
 					ti_len, p15card->tokeninfo->label);
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1043,7 +1043,7 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 {
 	struct pkcs15_slot_data *fw_data;
 	struct sc_pkcs15_auth_info *pin_info = NULL;
-	char label[64];
+	char label[(sizeof auth->label) + 10];
 
 	sc_log(context, "Called");
 	pkcs15_init_token_info(p15card, &slot->token_info);
@@ -1070,13 +1070,11 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 			pin_info = NULL;
 		}
 		else   {
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
-				int ti_len = MIN(strlen(p15card->tokeninfo->label),
-					sizeof(label) - strlen(auth->label) - 4);
-				snprintf(label, sizeof(label), "%.*s (%.*s)",
-					(int) strlen(auth->label), auth->label,
-					ti_len, p15card->tokeninfo->label);
-			} else
+			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
+				snprintf(label, sizeof(label), "%.*s (%s)",
+					(int) sizeof(auth->label), auth->label,
+					p15card->tokeninfo->label);
+			else
 				/* The PIN label is empty or says just non-useful "PIN" */
 				snprintf(label, sizeof(label), "%s", p15card->tokeninfo->label);
 			slot->token_info.flags |= CKF_LOGIN_REQUIRED;
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1070,9 +1070,13 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 			pin_info = NULL;
 		}
 		else   {
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
-				snprintf(label, sizeof(label), "%.*s (%s)", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
-			else
+			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
+				int ti_len = MIN(strlen(p15card->tokeninfo->label),
+					sizeof(label) - strlen(auth->label) - 3);
+				snprintf(label, sizeof(label), "%.*s (%.*s)",
+					(int) strlen(auth->label), auth->label,
+					ti_len, p15card->tokeninfo->label);
+			} else
 				/* The PIN label is empty or says just non-useful "PIN" */
 				snprintf(label, sizeof(label), "%s", p15card->tokeninfo->label);
 			slot->token_info.flags |= CKF_LOGIN_REQUIRED;
--- a/src/tests/pintest.c
+++ b/src/tests/pintest.c
@@ -66,7 +66,7 @@ static int ask_and_verify_pin(struct sc_pkcs15_object *pin_obj)
 	}
 
 	snprintf(prompt, sizeof(prompt), "Please enter PIN code [%.*s]: ",
-		(int) MIN(sizeof(prompt) - 26, strlen(pin_obj->label)),
+		(int) MIN(sizeof(prompt) - 27, strlen(pin_obj->label)),
 		pin_obj->label);
 	pass = (u8 *) getpass(prompt);
 
--- a/src/tests/pintest.c
+++ b/src/tests/pintest.c
@@ -13,6 +13,7 @@
 #include <unistd.h>
 #endif
 
+#include "libopensc/internal.h"
 #include "libopensc/opensc.h"
 #include "libopensc/pkcs15.h"
 #include "common/compat_getpass.h"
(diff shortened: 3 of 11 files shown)


src/libopensc/card-incrypto34.c
```diff
@@ -357,6 +357,7 @@ static int incrypto34_create_file(sc_card_t *card, sc_file_t *file)
 				 * generation. */
 			case SC_FILE_EF_LINEAR_VARIABLE_TLV:
 				type[1] = 0xff;
+				/* FALLTHROUGH */
```
frankmorgner: please avoid the capital letters. according to https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html, `/* fall through */` as used elsewhere in OpenSC should work, too, for `-Wimplicit-fallthrough=3` which gets enabled by `-Wextra` but not for `-Wall`...

src/pkcs11/framework-pkcs15.c
```diff
@@ -1070,9 +1070,13 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
...
 		}
 		else   {
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
-				snprintf(label, sizeof(label), "%.*s (%s)", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
-			else
+			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
+				int ti_len = MIN(strlen(p15card->tokeninfo->label),
+					sizeof(label) - strlen(auth->label) - 3);
+				snprintf(label, sizeof(label), "%.*s (%.*s)",
+					(int) strlen(auth->label), auth->label,
```
frankmorgner: this looks stupid. shouldn't it be `(int) MIN(sizeof auth->label, strlen(auth->label)`? `ti_len` should be calculated accordingly.

src/tests/pintest.c
```diff
@@ -64,7 +65,9 @@ static int ask_and_verify_pin(struct sc_pkcs15_object *pin_obj)
 		return 0;
 	}
 
-	sprintf(prompt, "Please enter PIN code [%.*s]: ", (int) sizeof pin_obj->label, pin_obj->label);
+	snprintf(prompt, sizeof(prompt), "Please enter PIN code [%.*s]: ",
+		(int) MIN(sizeof(prompt) - 26, strlen(pin_obj->label)),
```
frankmorgner: shouldn't we use `(int) MIN(sizeof(pin_obj->label), strlen(pin_obj->label)`?

src/tools/pkcs15-crypt.c
```diff
@@ -143,7 +144,8 @@ static char * get_pin(struct sc_pkcs15_object *obj)
 			return strdup(opt_pincode);
 	}
 
-	sprintf(buf, "Enter PIN [%.*s]: ", (int) sizeof obj->label, obj->label);
+	snprintf(buf, sizeof(buf), "Enter PIN [%.*s]: ",
+		(int) MIN(strlen(obj->label), sizeof(buf)-14), obj->label);
```
frankmorgner: same here, should we check for sizeof obj->label?

src/tools/pkcs15-init.c
```diff
@@ -3142,7 +3145,9 @@ static int verify_pin(struct sc_pkcs15_card *p15card, char *auth_id_str)
 			return SC_ERROR_OBJECT_NOT_FOUND;
 
 		if (pin_obj->label[0])
-			snprintf(pin_label, sizeof(pin_label), "User PIN [%s]", pin_obj->label);
+			snprintf(pin_label, sizeof(pin_label), "User PIN [%.*s]",
+				(int) MIN(sizeof(pin_label)-11, strlen(pin_obj->label)),
```
frankmorgner: same here.

src/libopensc/card-incrypto34.c
Jakuje: This sounds reasonably and it works as documented for me. Will fix.

src/pkcs11/framework-pkcs15.c
Jakuje: This is using assumption, that `auth->label` as a first argument should not be too long and therefore fit (though it is not good assumption). I am not sure what your `MIN` should give us, since the `label` is static buffer `char label[SC_PKCS15_MAX_LABEL_SIZE];`, therefore `sizeof()` is always 255, isn't it?
Fitting two flexible-length characters into fixed buffer is this way complicated, but the attempt here is to avoid truncation in the end of character, but keep the second buffer contained neatly in the brackets.

src/tests/pintest.c
Jakuje: the label is here again static buffer, so I am not sure how the `sizeof()` could help us here.

src/tools/pkcs15-init.c
Jakuje: same as above.

I am not sure if this is the right way of doing that or even improvement. But makes it the text contained in the formatting string if we happen to have longer ones.

From my testing, it looks like I still need to add -1 to accommodate null byte, but otherwise it looks like working well:

```
$ cat test.c
#include <stdio.h>
#include <string.h>
#define MIN(x, y) (((x) < (y)) ? (x) : (y))
int main() {
	char label[255] = "some long label that we do not want to truncate";
	char pin_label[40];
	printf("sizeof(label[255]) = %d\n", sizeof(label));

	snprintf(pin_label, sizeof(pin_label), "User PIN [%s]", label); 
	printf("%s\n", pin_label);

	snprintf(pin_label, sizeof(pin_label), "User PIN [%.*s]",
		(int) MIN(sizeof(pin_label)-12, strlen(label)),
		label);
	printf("%s\n", pin_label);
}
$ gcc test.c
$ ./a.out 
sizeof(label[255]) = 255
User PIN [some long label that we do no
User PIN [some long label that we do n]
```

Jakuje:
The travis build failures do not look related to these changes.

src/pkcs11/framework-pkcs15.c
```diff
@@ -1070,9 +1070,13 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
...
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
-				snprintf(label, sizeof(label), "%.*s (%s)", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
-			else
+			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
+				int ti_len = MIN(strlen(p15card->tokeninfo->label),
+					sizeof(label) - strlen(auth->label) - 4);
+				snprintf(label, sizeof(label), "%.*s (%.*s)",
+					(int) strlen(auth->label), auth->label,
+					ti_len, p15card->tokeninfo->label);
+			} else
```
frankmorgner: This still looks more complicated than it should be. What was the exact warning here?

src/tests/pintest.c
```diff
@@ -64,7 +65,9 @@ static int ask_and_verify_pin(struct sc_pkcs15_object *pin_obj)
 		return 0;
 	}
 
-	sprintf(prompt, "Please enter PIN code [%.*s]: ", (int) sizeof pin_obj->label, pin_obj->label);
+	snprintf(prompt, sizeof(prompt), "Please enter PIN code [%.*s]: ",
```
frankmorgner: Why don't you simply change the above code to `char prompt[(sizeof pin_obj->label) + 30];`

src/tools/pkcs15-crypt.c
```diff
@@ -143,7 +144,8 @@ static char * get_pin(struct sc_pkcs15_object *obj)
 			return strdup(opt_pincode);
 	}
 
-	sprintf(buf, "Enter PIN [%.*s]: ", (int) sizeof obj->label, obj->label);
+	snprintf(buf, sizeof(buf), "Enter PIN [%.*s]: ",
+		(int) MIN(strlen(obj->label), sizeof(buf) - 15), obj->label);
```
frankmorgner: same here, you could also change the size of `buf`

src/tools/pkcs15-init.c
```diff
@@ -3142,7 +3145,9 @@ static int verify_pin(struct sc_pkcs15_card *p15card, char *auth_id_str)
 			return SC_ERROR_OBJECT_NOT_FOUND;
 
 		if (pin_obj->label[0])
-			snprintf(pin_label, sizeof(pin_label), "User PIN [%s]", pin_obj->label);
+			snprintf(pin_label, sizeof(pin_label), "User PIN [%.*s]",
+				(int) MIN(sizeof(pin_label) - 12, strlen(pin_obj->label)),
+				pin_obj->label);
```
frankmorgner: same here.

src/pkcs11/framework-pkcs15.c
frankmorgner: can you change the size of `label`?

Jakuje: I have to agree with you. It would be nice if we could achieve the same somehow simpler.
We can change the size of labels, but it would
 * require dynamic allocation
 * could expand the prompt to unfriendly long line exposed to the users

The error looked this way:
```
  CC       opensc_pkcs11_la-framework-pkcs15.lo
framework-pkcs15.c: In function ‘pkcs15_init_slot’:
framework-pkcs15.c:1074:37: error: ‘%.*s’ directive output may be truncated writing up to 254 bytes into a region of size 64 [-Werror=format-truncation=]
     snprintf(label, sizeof(label), "%.*s (%s)", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
                                     ^~~~
framework-pkcs15.c:1074:5: note: ‘snprintf’ output 4 or more bytes (assuming 258) into a destination of size 64
     snprintf(label, sizeof(label), "%.*s (%s)", (int) sizeof auth->label, auth->label, p15card->tokeninfo->label);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

frankmorgner: my suggestion is to make the local buffer `label` bigger so that it can take everything you want to print into it in line 1074. Later, we still only copy 32 bytes from the local label to `token_info.label`.

Jakuje: The `sc_pkcs15_tokeninfo.label` in `p15card` is dynamically allocated so I believe there is no guarantee that it will have some size. Especially no guarantee that would satisfy gcc. Though in the other cases, it should work this way.

src/tests/pintest.c
Jakuje: Here it would make sense and it would make it simpler.  Thanks for the suggestion.

In the worst case, potentially malicious card/driver can have more space to display in prompt something we would not like (`\r` and overwriting our actual prompt).

src/tools/pkcs15-crypt.c
Jakuje: True. It will make it simpler, but the `snprintf()` should be used anyway.

src/pkcs11/framework-pkcs15.c
frankmorgner: I think the warning is due to the fact that the receiving field of `auth->label` (`%.*s`...`(int) sizeof auth->label`) is truncated. I think there is no problem with truncating `p15card->tokeninfo->label`, because the receiving field (`%s`) doesn't specify a length.

src/tests/pintest.c
frankmorgner: Appending bad stuff at the end is irrelevant if only the first bytes get copied over. However, generally speaking, you are right. The solution would be to validate all the card's inputs, which is kind of labor intensive with the given code base...

src/pkcs11/framework-pkcs15.c
Jakuje: Yes. You are right. Extending buffer helps here too and the dynamic part as `%s` does not matter here. See the latest change.

PR #1193, opened 2017-11-09
Fix reading EstEID certificates with T=0

metsma:
Fixes #1190

Signed-off-by: Raul Metsma <raul@metsma.ee>

##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [X] Tested with the following card: EstEID
	- [X] tested PKCS#11
	- [ ] tested Windows Minidriver
	- [ ] tested macOS Tokend

PR #1205, opened 2017-11-19
[AuthentIC v3] Correctly handle APDUs with more than 256 bytes

### LINKED ISSUE #1147: Can't use DNIe from Perú
### Problem Description
I'm trying to integrate a card reader, but I can not get this firm, in theory I should support sha256 but throw me CKR_MECHANISM_INVALID
My smardcard reader is a bit4id minilector

### Steps to reproduce
pkcs11-tool  --module /Library/OpenSC/lib/pkcs11/opensc-pkcs11.so --login --pin 666666  --slot=02 -s -m SHA256-RSA-PKCS -i mytest -o sha256.out --verbose
Using signature algorithm SHA256-RSA-PKCS
### Logs
/Library/OpenSC/bin/pkcs11-tool --module /Library/OpenSC/lib/pkcs11/opensc-pkcs11.so -M
RESULT:
Supported mechanisms:
  SHA-1, digest
  SHA256, digest
  SHA384, digest
  SHA512, digest
  MD5, digest
  RIPEMD160, digest
  RSA-PKCS, keySize={1024,2048}, hw, decrypt, sign, verify
  SHA1-RSA-PKCS, keySize={1024,2048}, sign, verify
  SHA256-RSA-PKCS, keySize={1024,2048}, sign, verify
  RSA-PKCS-KEY-PAIR-GEN, keySize={1024,2048}, generate_key_pair

can someone help me ?, what am I doing wrong?

sergioag:
Multiple sc_transmit_apdu calls are made when more than 256 bytes (read) or 255 bytes (write and update) are requested.
This fixes #1147.

##### Checklist

- [x] Tested with the following card: Oberthur AuthentIC 3.2.2
	- [x] tested PKCS#11
	- [x] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-authentic.c
+++ b/src/libopensc/card-authentic.c
@@ -807,46 +807,6 @@ authentic_select_file(struct sc_card *card, const struct sc_path *path,
 	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
 }
 
-
-static int
-authentic_apdus_allocate(struct sc_apdu **head, struct sc_apdu **new)
-{
-	struct sc_apdu *allocated_apdu = NULL, *tmp_apdu = NULL;
-
-	if (!head)
-		return SC_ERROR_INVALID_ARGUMENTS;
-
-	allocated_apdu = calloc(1, sizeof(struct sc_apdu));
-	if (!allocated_apdu)
-		return SC_ERROR_OUT_OF_MEMORY;
-
-	if (*head == NULL)
-		*head = allocated_apdu;
-
-	if (new)
-		*new = allocated_apdu;
-
-	tmp_apdu = *head;
-	while(tmp_apdu->next)
-		tmp_apdu = tmp_apdu->next;
-
-	tmp_apdu->next = allocated_apdu;
-
-	return 0;
-}
-
-
-static void
-authentic_apdus_free(struct sc_apdu *apdu)
-{
-	while(apdu)   {
-		struct sc_apdu *tmp_apdu = apdu->next;
-		free(apdu);
-		apdu = tmp_apdu;
-	}
-}
-
-
 static int
 authentic_read_binary(struct sc_card *card, unsigned int idx,
 		unsigned char *buf, size_t count, unsigned long flags)
@@ -888,8 +848,7 @@ authentic_read_binary(struct sc_card *card, unsigned int idx,
 		LOG_FUNC_RETURN(ctx, count);
 	}
 
-	if (!rv)
-		rv = sc_check_sw(card, apdu->sw1, apdu->sw2);
+	rv = sc_check_sw(card, apdu->sw1, apdu->sw2);
 	if (!rv)
 		count = ret_count;
 
@@ -905,7 +864,7 @@ authentic_write_binary(struct sc_card *card, unsigned int idx,
 		const unsigned char *buf, size_t count, unsigned long flags)
 {
 	struct sc_context *ctx = card->ctx;
-	struct sc_apdu *apdus = NULL, *cur_apdu = NULL;
+	struct sc_apdu *apdu = NULL;
 	size_t sz, rest;
 	int rv;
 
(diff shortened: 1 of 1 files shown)


frankmorgner:
The special APDU handling in the authentic driver has been introduced by @viktorTarasov, it includes code specifically designed for a remote reader (not part of OpenSC) that sets the otherwise unused flag `SC_READER_HAS_WAITING_AREA`. We need his comment on how his use cases are reacting upon this change.

Note that the logic that you're removing with this PR is present in some more places in `card-authentic.c`. Those should be changed, too, if we decide to integrate this change.

dengert:
@viktorTarasov, If the real problem is the card does not handle a READ_BINARY with Le=0, in T=0 mode,
would a solution like https://github.com/OpenSC/OpenSC/pull/1193 (i.e. max read is 255) also solve this problem?

What other cards/readers have this same issue with READ_BINARY? 

(It almost looks like a ambiguous issue in ISO7816-4 with with READ_BINARY. Does it return 9000 or 61xx)

sergioag:
@frankmorgner I'm already testing the changes for authentic_update_binary and authentic_write_binary. I will update this PR with those changes as soon as I'm satisfied with them.

sergioag:
@frankmorgner I have added the write and update changes. Should be ok now.

frankmorgner:
@sergioag please update the description on the top accordingly.

Did you try to get in touch with @viktorTarasov?

sergioag:
@frankmorgner I have updated the description. Hopefully it's better now. I also just sent out an email to Viktor.

viktorTarasov:
@sergioag If it 'works for you', I do not have objections.

src/libopensc/card-authentic.c
```diff
@@ -807,97 +807,53 @@ authentic_select_file(struct sc_card *card, const struct sc_path *path,
...
+		else
+			break;
 
 		idx += sz;
 		rest -= sz;
 	}
 
-	if (!apdus)   {
+	if (rv)   {
 		LOG_TEST_RET(ctx, SC_ERROR_INTERNAL, "authentic_read_binary() failed");
```
frankmorgner: You are not `free()`ing `apdu` on error. What about using a local variable instead of allocating some memory?

src/libopensc/card-authentic.c
```diff
@@ -917,37 +873,35 @@ authentic_write_binary(struct sc_card *card, unsigned int idx,
 	       "offs:%i,count:%"SC_FORMAT_LEN_SIZE_T"u,max_send_size:%"SC_FORMAT_LEN_SIZE_T"u",
 	       idx, count, card->max_send_size);
 
-	/* see comments for authentic_read_binary() */
-	sc_log(ctx, "reader flags 0x%lX", card->reader->flags);
-	if (count > 255 && !(card->reader->flags & SC_READER_HAS_WAITING_AREA))
-		LOG_TEST_RET(ctx, SC_ERROR_INVALID_DATA, "Invalid size of the data to read");
+	apdu = calloc(1, sizeof(struct sc_apdu));
+	if(!apdu)
+		LOG_TEST_RET(ctx, SC_ERROR_OUT_OF_MEMORY, "cannot allocate APDU");
```
frankmorgner: same here, if you use a local variable, you're saving the hassle of cleaning up the memory.

frankmorgner:
Thanks for the feedback, @viktorTarasov!

frankmorgner:
@sergioag the implementation in `card-authentic.c` looks like the same as in `iso7816.c` now (minus some additional error checks that the ISO driver performs). Please check if you can remove the implementation for authentic by directly calling the implementation of the ISO driver.

src/libopensc/card-authentic.c
sergioag: Done. Much simpler now.

sergioag:
@frankmorgner the iso7816.c implementation appears to use extended apdus, which don't appear to work on my card. It always returns invalid size.

src/libopensc/card-authentic.c
sergioag: Also done.

frankmorgner:
you could try something like:
```
size_t max_recv_size = card->max_recv_size;
rv = iso_ops->read_binary(card, idx, buf, count, flags);
card->max_recv_size = max_recv_size;
```

frankmorgner:
you could try something like:
```
size_t max_recv_size = card->max_recv_size;
card->max_recv_size = AUTHENTIC_READ_BINARY_LENGTH_MAX;
rv = iso_ops->read_binary(card, idx, buf, count, flags);
card->max_recv_size = max_recv_size;
```

frankmorgner:
@sergioag did you have a look at the suggestion above?

frankmorgner:
@sergioag, preparing a new release, could you look at the suggestion above?

sergioag:
@frankmorgner Will do it tonight or tomorrow.

frankmorgner:
thanks

sergioag:
Still working on it. Having some problems.

frankmorgner:
@sergioag do you have some update?

frankmorgner:
OK, thanks for trying.

PR #1231, opened 2018-01-02
Skip correctly two bytes after reading 2b size

Jakuje:
The SimpleTLV function used in CAC driver did not properly parse multi-byte lengths and therefore failing with CAC cards having some bigger field (currently ignored) before certificate data in TL file.

This might not be complete fix yet so do not hurry with merging, but it is clearly a bug that needs to get fixed. 

##### Checklist
- [X] Tested with the following card: CAC
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/simpletlv.c
+++ b/src/libopensc/simpletlv.c
@@ -90,8 +90,9 @@ sc_simpletlv_read_tag(u8 **buf, size_t buflen, u8 *tag_out, size_t *taglen)
 			*taglen = 0;
 			return SC_ERROR_INVALID_ARGUMENTS;
 		}
+		/* skip two bytes (the size) */
 		len = lebytes2ushort(p);
-		p++;
+		p+=2;
 	}
 	*taglen = len;
 	*buf = p;


Jakuje:
It would be nice to have some unit tests for such cases. Once there will be cmocka from #1224 in round, it should be pretty simple to add something for Simple TLV, but also for ASN parser, generator and other library functions.

frankmorgner:
@Jakuje did you think about adding `SC_ERROR_ASN1_END_OF_CONTENTS`?

Jakuje:
Sorry, I completely forgot about this change.

The bound check is already there (`if (buflen < 4)` -- the first byte is TAG, second byte is 0xFF and then we have 2B of size), but in this case we return `SC_ERROR_INVALID_ARGUMENTS`, which is quite the only error (bounds) that can happen during parsing the TL file, because the SimpleTLV is ... very simple.

From my point of view, it would not add any value.

frankmorgner:
OK, thanks

Issue #803, opened 2016-06-18
SmartCard-HSM EC keys with OpenSSH


RaymiiOrg:
I want to use an EC key with OpenSSH. First I generate a key:

    pkcs11-tool --module opensc-pkcs11.so --login --pin 648219 --keypairgen --key-type EC:secp256r1 --id 2 --label "SSH Key"

Then I get the pubkey:

    pkcs15-tool --read-public-key 2 > eckey.pub
    #cat eckey.pub:
    -----BEGIN PUBLIC KEY-----
    MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEPDJy6vEmcsjUoRovCW8ZNifKUYev
    XIEiSORGp1MLXXsY1OcaQohZfMYMZ//z7rVvfOBP6HXciEEo0PxoLoWtKw==
    -----END PUBLIC KEY-----

Convert that to OpenSSH format:

    ssh-keygen -i -m PKCS8 -f eckey.pub

Output:

    ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBDwycurxJnLI1KEaLwlvGTYnylGHr1yBIkjkRqdTC117GNTnGkKIWXzGDGf/8+61b3zgT+h13IhBKND8aC6FrSs=

Place that in the authorized_keys file on a modern (Ubuntu 16.04) machine. Try to SSH with the HSM and it fails:

    $ ssh -o "PKCS11Provider opensc-pkcs11.so" root@server
    C_GetAttributeValue failed: 18
    no keys
    root@server's password:

The regular way of getting an SSH key also doesnt work with an EC key:

    $ pkcs15-tool --verbose --read-ssh-key 2 
    Using reader with a card: Nitrokey Nitrokey HSM (010000000000000000000000) 00 00
    Connecting to card in reader Nitrokey Nitrokey HSM (010000000000000000000000) 00 00...
    Using card driver SmartCard-HSM.
    Trying to find a PKCS#15 compatible card...
    Found SmartCard-HSM!
    Reading ssh key with ID '2'

How do I use an EC key with OpenSSH?

mouse07410:
I tried the same with Yubikey NEO provisioned with ECC keys (on P256 curve), and faced the same problem, trying to extract public key in SSH format. Here's the log:
[...]

PR #1233, opened 2018-01-05
pkcs15-tool: added support for reading NIST ssh keys

popovec:
'pkcs15-tool --read-ssh-key' is now able to read NIST ECC keys from card.
Only 256, 384 and 521 field lengths are supported (same as allowed in
ssh-keygen -t ecdsa). Issue #803 is partialy fixed by this patch.
Openssh PKCS11 interface patches for ECC are now available, please check
https://bugzilla.mindrot.org/show_bug.cgi?id=2474

##### Checklist

- [x] Tested with the following card:  MyEID
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend


src/tools/pkcs15-tool.c
```diff
@@ -982,6 +982,67 @@ static int read_ssh_key(void)
...
+			struct sc_object_id curve_oid;
+			size_t size;
+		} ec_curves[] = {
+			{"secp256r1", {{1, 2, 840, 10045, 3, 1, 7, -1}},256},
+			{"secp384r1", {{1, 3, 132, 0, 34, -1}},         384},
+			{"secp521r1", {{1, 3, 132, 0, 35, -1}},         521},
+		        {NULL, {{-1}}, 0},
+		};
+		char alg[20];
+		unsigned char buf[4+30+4+9+4+256];
```
Jakuje: I would write there either "big enough" number as for the RSA section or add there some comment why specially these numbers. Otherwise the code becomes it is very hard to maintain.

popovec: Ok, what about:

#define ITEM_SIZE 4
#define ALG_NAME 30
#define CURVE_NAME 9
#define  KEY_DATA 256
..
unsigned char buf[3*ITEM_SIZE+ALG_NAME+CURVE_NAME+KEY_DATA];

???

Jakuje: Well ... you already propose that 30 chars is enough to fit `ecdsa-sha2-nistp521` (19 chars), so I would rather put there something like this:
```
/* Large enough to fit the following:
 * 3 x 4B item length headers
 * max 20B algorithm name,  9B curve name, max 256B key data */
unsigned char buf[300];
```
Using constants makes it clearer, but significantly longer.

popovec: Thanks for hint, pull request is already updated as your suggested.

frankmorgner:
@popovec could you keep us updated, when ECDSA support is merged into OpenSSH? We'll keep the PR hanging till then.

popovec:
Ok, I will let you know when openssh support for ECSDA is integrated into official openssh release.

frankmorgner:
thanks

Jakuje:
Fedora 28 shipped with ECDSA keys support in OpenSSH and it would be good to have also this in next OpenSC release.

dengert:
In response to https://github.com/OpenSC/OpenSC/pull/1233#issuecomment-385996015

The OpenSC pkcs11 can read ECC keys and do ECDSA and ECDH operations on some cards. 
Is there something missing other then pkcs15-tool? Can you do what is needed using pkcs11-tool?

Jakuje:
>  Is there something missing other then pkcs15-tool? Can you do what is needed using pkcs11-tool?

I don't think so. The `pkcs11-tool` does not allow listing the keys in SSH format and it is for some reason implemented in `pkcs15-tool`.

Regardless `pkcs11-tool`, the same can be achieved by `ssh-keygen -D /path/to/pkcs11.so`, with the OpenSSH ECDSA patch, but for consistency, it would be good to have this functionality also in OpenSC.

I just verified again with my yubikey, that this builds fine and the output matches the one from ssh-keygen for all the keys I have.

mouse07410:
Status update, please?

martinpaljak:
To be honest, the task of converting a pem public key (readable from pkcs11) to the SSH format should not necessarily be part of any of the OpenSC tools. Convenient - yes. Blocking or must have feature? Not really.

popovec:
Even when  openssh never gets support for ECC keys over pkcs#11 interface, pkcs15-tool should have support for listing _any_ keys in ssh  format. Or, to get consistent behavior, there is another option -  support for listing only RSA keys in ssh format should be removed from pkcs15-tool.

martinpaljak:
Tru. A "pem2ssh" probably exists?

frankmorgner:
It is possible to do
```
ssh-keygen -D /usr/lib/opensc-pkcs11.so >> ~/.ssh/authorized_keys
```

frankmorgner:
Sorry, I just noticed it was already suggested above. Anyway, if it is still wanted, we can include this feature after release of 0.18.0.

Jakuje:
Why after release? I am not saying this is a blocking issue, but it is something good to have for consistency as already mentioned by @popovec. It is always good to have more ways and tools of doing the same thing.

The OpenSSH upstream does not ship the ecdsa in pkcs11 yet, but there are several people using it already. If OpenSSH will release in a three months with it, the next OpenSC release will be until next year? It really has to take a year to get such simple and isolated change into OpenSC release?

Jakuje:
@frankmorgner I see it is already after the release. Is anything else blocking merging this PR?

PR #1232, opened 2018-01-05
Fix for OpenPGP applet selection

frankmorgner:
- fixes https://github.com/OpenSC/OpenSC/issues/1230
- partially implements https://github.com/OpenSC/OpenSC/issues/1215

##### Checklist

- [ ] Tested with the following card: OpenPGP v3.3 
	- [ ] tested PKCS#11
	- [ ] tested Windows Minidriver
	- [ ] tested macOS Tokend
	- [x] tested `opensc-tool -n`

Issue #962, opened 2017-02-07
Concurrent access to multiple applications on card - Yubico, MyEID and others  


dengert:
Over the last few years, card vendors  are offering cards with multiple applets on the same card. Yubico and MyEID in particular and there will by others.  But users have problems using the applets concurrently and especially from different applications.   (Some examples: OpenPGP, CAC and  PIV) 

We have talked about how to support these concurrently in OpenSC, but any concurrent use comes down to: Does the card support this and can the card keep the security state active and separate. 

ISO 7816 -4 "5.4.2 Logical channels" states:
"This clause specifies a mechanism whereby in the interindustry class, C-RPs can refer to logical channels.
Each logical channel has its own security status (see 9.3) and validity area (see 7.2)."
4 channels have bee defined for years. 19 channels are defined in later versions of ISO 7816-4:2013.

So one way to support multiple concurrent applets is to use Logical channels. But it is not clear if any of these cards can support this internally. I have seen some Oberthur cards where the ATR Historical bytes say the card supports logical channels. But the Yubico and MyEID both put their company name in the historical bytes and there is no indication. 

I am asking  Yubico and MyEID  if thety support logical channels, or would they be willing to add support for logical channels and work with OpenSC to support this. 

I am also asking if anyone knows of any other cards with multiple applets, that do support chaining.
Open Platform for example? 
 
Issues: 

- Card must support logical channels. - (Vendor issue) We need at least one type of card for testing. 

- PCSC locks on the card and not on a logical channel, so middleware can not hold locks for very long. 

- Allocation of channels. ISO 7816 MANAGE CHANNEL command (70) can be used, and for any of the cards OpenSC could always use a channel other then zero, as other middleware may be using only channel zero. 

- A card-reset resets the whole card, so must be avoided.  MANAGE CHANNEL. to reset the channel might be a replacement for card reset, and could be used for a "logout". 

I bring this up for discussion because  it  needs vendor input as will as OpenSC input.

frankmorgner:
Unfortunately, there doesn't seem to be much interest in solving concurrent access to multiple applications on a card. And quite frankly, I don't think it this is currently needed given the typical smart card use today.

Please re-open if there is new input!

dengert:
@frankmorgner In https://github.com/OpenSC/OpenSC/pull/1232#issuecomment-356605500
[...]

PR #1239, opened 2018-01-13
Re-select OpenPGP AID as first command after sCardBeginTransaction

dengert:
Partial fix for https://github.com/OpenSC/OpenSC/issues/962
This change depends on  https://github.com/OpenSC/OpenSC/pull/1232/commits/59963e051605c325a5dad83a3897705c90a3faeb which is part of https://github.com/OpenSC/OpenSC/pull/1232 

https://github.com/OpenSC/OpenSC/pull/1232/commits/59963e051605c325a5dad83a3897705c90a3faeb makes sure the OpenPGP AID is initially selected when doing a match based on AID and is required for this PR to function.  Once  https://github.com/OpenSC/OpenSC/pull/1232/commits/59963e051605c325a5dad83a3897705c90a3faeb is commited  this PR can be rebased on to master. 

The OpenPGP application may be on cards with other application such
as PIV. Doing a SELECT AID will make sure correct application is selected
as other processes may have selected a different application on the card.
   
Use the existing *_card_reader_lock_obtained hook called from sc_lock
 to do this.

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend
        - [X] tested with opensc-tool


Jakuje:
Thank you for the work. I tested the following scenario with my locally built OpenSC from your branch:

 * In first terminal:
   * I configured OpenSC to use OpenPGP driver in `/usr/local/etc/opensc.conf` [1]
 * In second terminal:
   * Started `ssh-agent`:  `OPENSC_DEBUG=9 eval $(ssh-agent -d)`
 * In third terminal:
   * Initialized the connection to the `ssh-agent`:
     * `SSH_AUTH_SOCK=/tmp/ssh-k6EwKJgeQAbX/agent.11719; export SSH_AUTH_SOCK;`
   * added a PKCS#11 provider to the `ssh-agent`:
     * `ssh-add -s /home/jjelen/devel/OpenSC/src/pkcs11/.libs/opensc-pkcs11.so`
   * Listed the public keys using `ssh-add -L`, added them to `~/.ssh/authorized_keys`
   * Attempted to connect several times using OpenPGP key from agent
     * `ssh localhost`

 * I configured OpenSC to use PIV-II driver in `/usr/local/etc/opensc.conf` [2]
 * I added the PKCS#11 driver to Firefox Security Devices
 * I started [`./gnutls-http-serv`](https://gitlab.com/gnutls/gnutls/blob/master/doc/credentials/gnutls-http-serv) with CA as the PIV certificate
 * I connected and authenticated from Firefox using the PIV certificate to the `https://localhost:5556/`
 * Then again, I tried the signature from the SSH in the third terminal

This test went successfully for me. If you have some other ideas how it can be simply tested or you want to test other cards, feel free to use above procedure or ask for clarifications.

[1] I can't use PIV in ssh-agent, because it does not respect always authenticate flags and fails with Signature key. [OpenSSH bug](https://bugzilla.mindrot.org/show_bug.cgi?id=2638)
[2] I can not use OpenPGP applet here, since out of the box it does not have X.509 certificates to be used in TLS authentication.

frankmorgner:
@Jakuje could you add the above description to https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Testing#pkcs11? I think it's useful to test ssh for every release...

frankmorgner:
I think this approach is not ideal on solving the issue with concurrency.

1. It introduces many card commands, which significantly slow down the card's usage. Card selection on JCOP is one of the more expensive commands, which is why I put considerable effort into the sc-hsm's driver to [avoid those commands](74ec7b04ffda539819eb8bfd6a0e766e69d8c946). In a multi process environment, this significantly reduced the overall-time spent for card transactions (I tested with Windows logon, where a hand full of processes are using the card). This consideration should be rethought for current implementation of the PIV driver as well.

2. The [OpenPGP Applet resets the login status on re-selection of the applet](https://github.com/Yubico/ykneo-openpgp/blob/master/applet/src/openpgpcard/OpenPGPApplet.java#L239-L240), hence, OpenSC must not re-select the applet when the user is logged in. This means the same tracking as in the PIV driver needs to be implemented. However, when the user is logged in, and a different process deselects the applet, OpenSC is not recovering anymore from this, because it thinks it needs to preserve the login status rather than re-selecting the applet. So we're back to where we started. This consideration should also be rethought for current implementation of the PIV driver as well.

frankmorgner:
Unfortunately, I cannot do any testing, since I don't have such a token.

The link on the sc-hsm optimization should have gone to the commit https://github.com/OpenSC/OpenSC/commit/74ec7b04ffda539819eb8bfd6a0e766e69d8c946

> You say "Card selection on JCOP is one of the more expensive commands" Do you mean AID selection?

Yes, if I remember correctly, this is about 100-150 Milliseconds. Multiplied by the number of processes accessing the card and the number of transactions this will amount to multiple seconds of delay...

I'll illustrate my second point with an example that's currently *not catched* by the PIV driver despite the special handling of the Yubikey (`disconnect` configuration is irrelevant):
1. Use PKCS#11+PIV
2. OpenSC selects PIV applet
3. OpenSC loggs into PIV and sets `priv->logged_in = SC_PIN_STATE_LOGGED_IN`
**Result**: `piv_card_reader_lock_obtained` does nothing, unless the card gets reset; `piv_pin_cmd` (`SC_PIN_CMD_GET_INFO`) will always return `SC_SUCCESS`
4. Use `gpg --card-status` (without resetting the card, with shared access; e.g. on Windows/macOS port of GnuPG).
5. gpg selects PGP applet
**Result**: OpenSC is trapped in thinking that the PIV applet is selected and the user is logged in.

Re-thinking about resets, I think you're right that we should catch all of them when checking the locking. What's missing is its inclusion in the minidriver...

Please don't get me wrong, I appreciate your effort on making OpenSC more robust. Selecting the applet in `lock_obtained()` is far from complete. Unfortunately, I don't have a silver bullet, and all I can offer is in-depth analysis of the specific card driver.

Jakuje:
@frankmorgner I edited the Wiki with the OpenSSH section and I also added a note (hopefully not only for future self) about the TLS client authentication server I am using. But the OpenSSH use case is generally very basic and works quite much with everything (except ECDSA and ALWAYS AUTHENTICATE, which is OpenSSH's fault).

@dengert If you have Yubikey with OpenPGP applet, there is nothing simpler than to initialize it:
https://www.yubico.com/support/knowledge-base/categories/articles/use-yubikey-openpgp/

At least what I remember, It was not too hard and it worked with out-of-the box gpg2 (except that exclusive locks from scdaemon prevention OpenSC working with it through PKCS#11, but it is different story unrelated here and now).

@frankmorgner The easiest cure for avoiding bogus SELECT AID commands (before each other command if not needed) would probably be to conditionalize them under either the `disconnect=leave` (it usually does not make sense in any other state, if we don't take into account different middleware that can touch the card) or under the potential resolution of #962 (we would have information about all the applets that were detected on the card and if there would be only one, we can probably assume we don't need it). Or just with different configuration option. But each of that would probably bring some more complexity, though it should be possible to solve it outside of the drivers, isn't it?

dengert:
@frankmorgner 
You are correct in the assessment of the current `piv_card_reader_lock_obtained`. It really only handles reset. To handle the case where a different AID was selected it would need  to select the PIV applet each time. much like the proposed `pgp_card_reader_lock_obtained`

To avoid during extra `SELECT AID` commands requires knowing what application is active. 
 
- I do not see any card command to query what AID is currently active. Is there one? Maybe on Java cards?

- Using the fact that a command to the wrong applet fails i.e. trapping  '6D00' also has problems. Both applets may respond to the same command.   For example (if I am reading the openpgp docs correctly) both PIV and OpenPGP  support  VERIFY P1=00 P2=81  Lc=8 So if a verify for a PIN is sent to the wrong applet it may respond with wrong PIN, not with 6D00, and will also decrement the tries left. Sending the wrong pin could end up locking the PIN .   Sending VERIFY without Lc to check login state will respond with  the login state of the active AID which does not say what AID is active. There may be other commands that might appear to work but have unexpected consequences. Both PIV and OpenPGP use GET DATA.  In the future when we have to deal with more applications, the problem will get worse. 

- OpenSC could keep track of the active AID in a single process  But other processes may change the  active AID. These processes are out of our control and may not be using OpenSC. And that is the main situation we are trying to address. 

- What if PCSC could keep track of the active AID, and could be queried. This would avoid the extra Select AID. PCSC  gets to see the commands, and could look for Select AID and save it. This would also catch other software changing the AID.  Not sure how to get Microsoft to add such a feature to Windows It is not clear how this would work with Remote Desktop.

So the only solution I see is send `SELECT AID` for each transaction. If the driver does a sc_lock for a collection of commands, say when reading a certificate, or doing a `SELECT FILE` followed by a number of `READ BINARY` commands, there would be only one Select AID per certificate.  I believe that internally the  OpenSC compute signature and other crypto operations already call sc_lock to get  all the commands into the same transaction.

I am busy today with other things, but will look at modifying `piv_card_reader_lock_obtained` to do the  `SELECT AID` and will look at setting up my Yubikey for OpenPGP in the next few days.

frankmorgner:
I tried using `gnutls-serv --priority NORMAL --x509cafile $HOME/auth.cer --x509keyfile $HOME/key.pem --x509certfile $HOME/cert.pem --require-client-cert` to start a server, where `auth.cer` is the certificate from the card and `key.pem`/`cert.pem` are the ssl certificates. Unfortunately, the connection is rejected. @Jakuje can you be more specific on how to setup the TLS server with client authentication?

Jakuje:
@frankmorgner Sorry, I was not clear enough and after some more tests I figured out that it might not always work as I suggested. The only change to the server you need to do is the `--x509cafile` (not the `--x509certfile`/`--x509keyfile`!).
You can use the argument with the certificate from the card, but it might not work, if it does not have appropriate flags (CA), but it is fine for self-signed certificates.
In general case, you should put there the issuer of your certificate (CA), which is able to verify your certificate is valid. For example with the PIV test cards, I downloaded the certificate from NIST:
https://csrc.nist.gov/Projects/PIV/NIST-Personal-Identity-Verification-Test-Cards

Hope it helps.

Issue #1240, opened 2018-01-15
PIN verification fails on IAS-SEC card


qsantos:
My initial bug report was erroneous, so here is a better description of the problem.

### Problem Description

OpenSC fails to verify the PIN on several IAS-SEC smartcards.

### Proposed Resolution

Per @redsectors investigation, it seems like the card is expecting a PIN padded to a length of 24 bytes using zeros. The following patch is sufficient to fix the verify command for this card:

```diff
diff --git a/src/libopensc/card-iasecc.c b/src/libopensc/card-iasecc.c
index a364a394..07754e64 100644
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1847,6 +1847,13 @@ iasecc_chv_verify(struct sc_card *card, struct sc_pin_cmd_data *pin_cmd,
                apdu.data = pin_cmd->pin1.data;
                apdu.datalen = pin_cmd->pin1.len;
                apdu.lc = pin_cmd->pin1.len;
+
+               unsigned char buf[24];
+               memset(buf, 0, 24);
+               memcpy(buf, pin_cmd->pin1.data, pin_cmd->pin1.len);
+               apdu.data = buf;
+               apdu.datalen = 24;
+               apdu.lc = 24;
        }
        else if ((card->reader->capabilities & SC_READER_CAP_PIN_PAD) && !pin_cmd->pin1.data && !pin_cmd->pin1.len)   {
                rv = iasecc_chv_verify_pinpad(card, pin_cmd, tries_left);
```

However, this patch may break functionality for other cards, and does not address the issue in other commands (i.e. changing the PIN). It seems to me that the best course of actions would be to introduce a workaround in `iasecc_pin_cmd()`, before dispatching, which would detect the affected cards and pad both `pin1` and `pin2` accordingly.

### Steps to reproduce

Introduce the card in reader, run `pkcs15-tool --verify`, and enter correct PIN.

### Logs

```$ OPENSC_DEBUG=2 pkcs15-tool --verify
Using reader with a card: Alcor Micro AU9560 00 00
0x7ff21a328340 13:07:53.480 [pkcs15-tool] card.c:200:sc_connect_card: called
Please enter PIN [PIN_Global]: 
0x7ff21a328340 13:07:55.972 [pkcs15-tool] sec.c:217:sc_pin_cmd: returning with: -1214 (PIN code or key incorrect)
Operation failed: PIN code or key incorrect
[...]

PR #1241, opened 2018-01-15
IAS/ECC: fixed applet selection

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1240

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -319,7 +319,7 @@ iasecc_select_aid(struct sc_card *card, struct sc_aid *aid, unsigned char *out, 
 	int rv;
 
 	/* Select application (deselect previously selected application) */
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x00);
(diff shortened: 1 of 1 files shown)


qsantos:
I think applet selection was already working for my card with the latest version (I mistakenly used 0.16.0-3 from Debian stretch repositories in my initial bug report). However, this does not seem to fix PIN verification. Should I maybe open a new, cleaner issue for this? Thanks anyway for your dedication to the project.

frankmorgner:
You can leave it in https://github.com/OpenSC/OpenSC/issues/1240, but you could write some update in the initial problem description.

I'd like to still merge the proposed change, because it fixes an obvious problem, that might lead to problems later.

Note, that I don't have the possibility to investigate further into the PIN issue.

qsantos:
As per your recommendations, I have updated the initial problem description in #1240. I am willing to help write the patch, but the only kind of smart card I have at my disposal is affected by the same bug, so I cannot test how the patch would behave with other cards.

PR #1248, opened 2018-01-22
muscle applet fix

luizurias1:
After some version of opensc, muscle applet was not supported anymore. This pull request modifies only files related with the muscle applet. Now, it can be recongnized and inicialized with opensc again.

### DIFF
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -83,6 +83,9 @@ static int muscle_match_card(sc_card_t *card)
 				if (r == SC_SUCCESS && response[0] == 0x01) {
 						card->type = SC_CARD_TYPE_MUSCLE_V1;
 						return 1;
+				} else {
+						card->type = SC_CARD_TYPE_MUSCLE_GENERIC;
+						return 1;
 				}
 			}
 	return 0;
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -355,7 +355,6 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
 {
 	mscfs_t *fs = MUSCLE_FS(card);
 	mscfs_file_t *file_data = NULL;
-	const u8 *path = path_in->value;
 	int pathlen = path_in->len;
 	int r = 0;
 	int objectIndex;
@@ -363,11 +362,11 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
 
 	mscfs_check_cache(fs);
 	r = mscfs_loadFileInfo(fs, path_in->value, path_in->len, &file_data, &objectIndex);
-	if(r < 0) SC_FUNC_RETURN(card->ctx, 2,r);
+	if(r < 0) SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE,r);
 
 	/* Check if its the right type */
 	if(requiredType >= 0 && requiredType != file_data->ef) {
-		SC_FUNC_RETURN(card->ctx, 0, SC_ERROR_INVALID_ARGUMENTS);
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_INVALID_ARGUMENTS);
 	}
 	oid = file_data->objectId.id;
 	/* Is it a file or directory */
@@ -390,8 +389,6 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
 		file->path = *path_in;
 		file->size = file_data->size;
 		file->id = (oid[2] << 8) | oid[3];
-		memcpy(file->name, path, pathlen);
-		file->namelen = pathlen;
 		if(!file_data->ef) {
 			file->type = SC_FILE_TYPE_DF;
 		} else {
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -42,6 +42,7 @@ static struct sc_card_driver muscle_drv = {
 	NULL, 0, NULL
 };
 
+
 #define MUSCLE_DATA(card) ( (muscle_private_t*)card->drv_data )
 #define MUSCLE_FS(card) ( ((muscle_private_t*)card->drv_data)->fs )
 typedef struct muscle_private {
@@ -464,6 +465,8 @@ static int muscle_init(sc_card_t *card)
 
 	card->flags |= SC_CARD_FLAG_RNG;
 	card->caps |= SC_CARD_CAP_RNG;
+	card->max_recv_size = 255;
+	card->max_send_size = 255;
 
 	if(card->type == SC_CARD_TYPE_MUSCLE_ETOKEN_72K) {
 		card->caps |= SC_CARD_CAP_APDU_EXT;
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -77,28 +77,13 @@ static u8 muscleAppletId[] = { 0xA0, 0x00,0x00,0x00, 0x01, 0x01 };
 
 static int muscle_match_card(sc_card_t *card)
 {
-	sc_apdu_t apdu;
-	u8 response[64];
 	int r;
 
-	/* Since we send an APDU, the card's logout function may be called...
-	 * however it's not always properly nulled out... */
 	card->ops->logout = NULL;
 
-	if (msc_select_applet(card, muscleAppletId, sizeof muscleAppletId) == 1) {
-		/* Muscle applet is present, check the protocol version to be sure */
-		sc_format_apdu(card, &apdu, SC_APDU_CASE_2, 0x3C, 0x00, 0x00);
-		apdu.cla = 0xB0;
-		apdu.le = 64;
-		apdu.resplen = 64;
-		apdu.resp = response;
-		r = sc_transmit_apdu(card, &apdu);
-		if (r == SC_SUCCESS && response[0] == 0x01) {
-				card->type = SC_CARD_TYPE_MUSCLE_V1;
-				return 1;
-		}
-	}
-	return 0;
+	r = msc_select_applet(card, muscleAppletId, 5);
+	card->drv_data = (void*)0xFFFFFFFF;
+	return r;
 }
 
 /* Since Musclecard has a different ACL system then PKCS15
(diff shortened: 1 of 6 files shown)


src/pkcs15init/pkcs15-muscle.c
```diff
@@ -69,7 +69,7 @@ muscle_create_dir(sc_profile_t *profile, sc_pkcs15_card_t *p15card, sc_file_t *d
 	struct sc_file *file;
 	struct sc_path path;
 	memset(&file, 0, sizeof(file));
-	sc_format_path("3F00", &path);
+	sc_format_path("3F005015", &path);
```
frankmorgner: Is there a specific reason for changing the directory (or is it just a matter of taste)?

src/pkcs15init/muscle.profile
```diff
@@ -8,7 +8,7 @@ cardinfo {
 
     max-pin-length	= 8;
     min-pin-length	= 4;
-    pin-encoding	= ascii-numeric;
+	pin-pad-char   = 0x00;
```
frankmorgner: Is there something not working with the current setup?

src/libopensc/card-muscle.c
```diff
@@ -365,18 +355,19 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
...
+	const u8 *path = path_in->value;
 	int pathlen = path_in->len;
 	int r = 0;
 	int objectIndex;
 	u8* oid;
 
 	mscfs_check_cache(fs);
 	r = mscfs_loadFileInfo(fs, path_in->value, path_in->len, &file_data, &objectIndex);
-	if(r < 0) SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE,r);
+	if(r < 0) SC_FUNC_RETURN(card->ctx, 2,r);
```
frankmorgner: There's no need to change this

src/libopensc/card-muscle.c
```diff
@@ -365,18 +355,19 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
...
 
 	/* Check if its the right type */
 	if(requiredType >= 0 && requiredType != file_data->ef) {
-		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_INVALID_ARGUMENTS);
+		SC_FUNC_RETURN(card->ctx, 0, SC_ERROR_INVALID_ARGUMENTS);
```
frankmorgner: see above

src/libopensc/card-muscle.c
```diff
@@ -399,6 +390,8 @@ static int select_item(sc_card_t *card, const sc_path_t *path_in, sc_file_t ** f
 		file->path = *path_in;
 		file->size = file_data->size;
 		file->id = (oid[2] << 8) | oid[3];
+		memcpy(file->name, path, pathlen);
+		file->namelen = pathlen;
```
frankmorgner: why do you need this change?

src/libopensc/card-muscle.c
luizurias1: Yes, thats true.

src/libopensc/card-muscle.c
luizurias1: I was testing and forget to delete this, it works without it.

src/pkcs15init/muscle.profile
luizurias1: see above

src/pkcs15init/pkcs15-muscle.c
luizurias1: Yes, at some point opensc try to access this directory (3F005015). I tried without this change and does not work :(

src/libopensc/muscle.h
```diff
@@ -32,8 +32,8 @@
 #define MSC_MAX_PIN_COMMAND_LENGTH ((1 + MSC_MAX_PIN_LENGTH) * 2)
 
 /* Currently max size handled by muscle driver is 255 ... */
-#define MSC_MAX_READ (card->max_recv_size > 0 ? card->max_recv_size : 255)
-#define MSC_MAX_SEND (card->max_send_size > 0 ? card->max_send_size : 255)
+#define MSC_MAX_READ (MIN(card->max_recv_size,255))
+#define MSC_MAX_SEND (MIN(card->max_send_size,255))
```
frankmorgner: This is wrong, too. if card->max_recv_size/card->max_send_size is 0, we want it to be 255.

src/pkcs15init/pkcs15-muscle.c
frankmorgner: Sorry, you need to dig deeper! The codes selects 3F00 and then creates 5015. This should perfectly fine...

src/libopensc/card-muscle.c
```diff
@@ -475,13 +465,6 @@ static int muscle_init(sc_card_t *card)
 	card->flags |= SC_CARD_FLAG_RNG;
 	card->caps |= SC_CARD_CAP_RNG;
 
-	/* Card type detection */
-	if (_sc_match_atr(card, muscle_atrs, &card->type) < 0)   {
```
frankmorgner: You may move the the ATR detection to `muscle_match_card()`, but keep muscle limited to these ATRs by default. The user may still force the muscle driver via `opensc.conf`.

src/libopensc/muscle.h
luizurias1: Ok, but the card->max_send_size was setting to 255 and card->recv_size was setting to 256, and muscle max is 255..so in ```
muscle_init()``` i initialized both with 255, is that ok?

src/libopensc/card-muscle.c
luizurias1: But if the card does not match with ``` muscle_atrs ``` it should return not supported? Because i forced the driver in ``` opensc.conf ``` but in the ATR detection does not match.

frankmorgner: If the ATR detection is in `muscle_match_card()` instead of `muscle_init()` you should be able to force the driver like this:
```
card_atr 11:22:33:44:55:66:77:88:99:00 {
	driver = "muscle";
}
```
Alternatively, you may add your ATR to the list...

src/libopensc/muscle-filesystem.c
```diff
@@ -224,7 +224,7 @@ int mscfs_loadFileInfo(mscfs_t* fs, const u8 *path, int pathlen, mscfs_file_t **
 		}
 		*file_data = NULL;
 	}
-	if(*file_data == NULL && (0 == memcmp("\x3F\x00\x00\x00", fullPath.id, 4) || 0 == memcmp("\x3F\x00\x3F\x00", fullPath.id, 4 ))) {
+	if(*file_data == NULL && (0 == memcmp("\x3F\x00\x00\x00", fullPath.id, 4) || 0 == memcmp("\x3F\x00\x50\x15", fullPath.id, 4 ) || 0 == memcmp("\x3F\x00\x3F\x00", fullPath.id, 4))) {
```
frankmorgner: I think this change can also be removed.

src/pkcs15init/muscle.profile
```diff
@@ -8,7 +8,7 @@ cardinfo {
 
     max-pin-length	= 8;
     min-pin-length	= 4;
-    pin-encoding	= ascii-numeric;
+	pin-encoding	= ascii-numeric;
```
Jakuje: This change is not needed either

src/libopensc/muscle-filesystem.c
luizurias1: As you said, the code select 3F00 and then creates 5015..at some point the directory 3F005015 is accessed and without this change in ```if```, file not found is returned as error. So, i think this is needed

frankmorgner:
Creating and selecting the DF `5015` seemed to work for everyone so far... Where's the source code of the muscle applet you're using? Which version of OpenSC were you using before?

Regarding the ATR detection, did you try my suggestion, what was your exact set of patches, what was your `opensc.conf`?

To make long story short, if you roll your own card and you're migrating anyway. Then you may consider rolling your own version of OpenSC without the need to merge this into the project.

luizurias1:
I get the muscle cap file from this: https://github.com/CardContact/scsh-scripts/tree/master/musclecard that has compiled from www.musclecard.com. The last version of OpenSC that supports it is 0.11.4 (that's the version i am using)

About the ```opensc.conf```, i am using ID-One Cosmo v7.0.1 and i added this:
 ```
card_atr 3b:db:96:00:80:b1:fe:45:1f:83:00:31:c0:64:1a:18:01:00:0f:90:00:52 {
		driver = "muscle";
	}
```

But still doest not match with any ATR (see above) at ```muscle_atr``` and then return not supported. 

```
 if (_sc_match_atr(card, muscle_atrs, &card->type) < 0)   {
 	SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_NOT_SUPPORTED);
 }
``` 


Hmm.. since muscle applet is still referenced and a well-know applet, my intention was to help the community that still uses it for some reason..

**Comment by dengert** [comment]
The first thing the card-muscle.c does it to verify that there is a Muscle  application on the card by 
selecting the Muscle AID. 

You can simulate this by using the following command:
`opensc-tool -c default -s "00 A4 04 00 06 A0 00 00 00 01 01"`
It should return  with a status of `9000`

An opensc-debug.log would also be helpful. 

(I do not have any Muscle cards for testing.)

File: src/libopensc/card-muscle.c [hunk 10]
```diff
@@ -474,13 +465,8 @@ static int muscle_init(sc_card_t *card)
...
 	card->caps |= SC_CARD_CAP_RNG;
-
-	/* Card type detection */
-	if (_sc_match_atr(card, muscle_atrs, &card->type) < 0)   {
-		free(priv->fs);
-		free(card->drv_data);
-		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_NOT_SUPPORTED);
-	}
+	card->max_recv_size = 255;
+	card->max_send_size = 255;
```
frankmorgner: please remove this change. Some muscle card indirectly set this data by specifying extended length capabilities

frankmorgner:
move `_sc_match_atr` from `muscle_init` to `muscle_match_card` as shortcut to the applet selection. We need this shortcut, otherwise the card's type is not properly initialized (See for example how we've done that [in an other applet-based driver](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-sc-hsm.c#L233-L241)). I've just realized that modifying opensc.conf should then not be needed.

src/libopensc/card-muscle.c
luizurias1: See [this comment](https://github.com/OpenSC/OpenSC/pull/1248#discussion_r163543010). I noticed that when card->max_recv_size is setted to 256, PIN structure is not created..

luizurias1:
Ok, but when does not match with any atr from ```muscle_atr``` list (my case), i should set card->type as what?  ```SC_CARD_TYPE_MUSCLE_GENERIC```?

src/libopensc/card-muscle.c
```diff
@@ -363,20 +348,76 @@ static void muscle_load_dir_acls(sc_file_t* file, mscfs_file_t *file_data)
...
+//			muscle_load_file_acls(file, file_data);
+//		} else {
+//			muscle_load_dir_acls(file, file_data);
+//			/* Setup directory acls... */
+//		}
+//
+//		file->magic = SC_FILE_MAGIC;
+//		*file_out = file;
+//	}
+//	return 0;
```
martinpaljak: And please do not commit commented code. Git history is for that.

src/libopensc/card-muscle.c
martinpaljak: As a general rule, publicly available JavaCard applet based drivers should contain NO hardcoded ATR-s. At least no generic JavaCard ATR-s. ATR-s could be considered IFF the applet sets historical bytes to some pre-defined value (and the given card does not have different cold/warm ATR-s).

src/libopensc/card-muscle.c
luizurias1: Sorry. I already removed

frankmorgner:
I've implemented my comments, could you check if https://github.com/OpenSC/OpenSC/pull/1316 equally solves the problems you had?

PR #1249, opened 2018-01-23
Refactored sc_get_challenge

frankmorgner:
Let sc_get_challenge() do sc_lock() and loop through the card driver's get_challenge() until enough bytes were collected. The card driver's get_challenge() now returns the number of bytes collected (less or equal than requested) or an error code.

- Allow more code re-use.
- PIV driver now uses ASN.1 parser for reading the random bytes
- SC-HSM now has a maximum for the challenge

##### Checklist

- [x] Tested with the following card: OpenPGP 3.3, SC-HSM 2.4, CardOS 4.3B
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-authentic.c
+++ b/src/libopensc/card-authentic.c
@@ -1661,40 +1661,27 @@ authentic_get_serialnr(struct sc_card *card, struct sc_serial_number *serial)
 }
 
 
-/* 'GET CHALLENGE' returns always 24 bytes */
 static int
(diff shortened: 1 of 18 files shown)


dengert:
The card-piv.c change has an error in the memcpy line. It is using rbuf rather the p as the source, so it is copying the tag and length and only part of the random data.

Issue #1244, opened 2018-01-18
Segmentation Fault using OpenSC with pam_pkcs11 (pkcs11_inspect)


thukk:
### Problem Description

Last week, I was able to use a token that OpenSC identifies as SmartCard-HSM.  I would at least get to the PIN request.  This week, I rebuilt a machine and reinstalled OpenSC 0.17 using the default options.  Now, I get a segmentation fault in pam_pkcs11 using the exact same token.

### Steps to reproduce

Install OpenSC and pam_pkcs11.  Token identifies as SmartCard-HSM by default.  

pam_pkcs11.conf is set up to use opensc.  
module is set to 'module = "/usr/lib/opensc-pkcs11.so";' (if left without the absolute path, the module is not found)

Run pkcs11_inspect.  Segmentation Fault.

thukk:
here is the GDB output:

`Starting program: /usr/bin/pkcs11_inspect 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff1221e71 in sc_pkcs15emu_sc_hsm_init (p15card=0x669e30) at pkcs15-sc-hsm.c:826
826		p15card->card->version.fw_major = file->prop_attr[file->prop_attr_len - 2];
(gdb) bt full
#0  0x00007ffff1221e71 in sc_pkcs15emu_sc_hsm_init (p15card=0x669e30) at pkcs15-sc-hsm.c:826
        card = 0x668d50
        priv = 0x667ee0
        file = 0x66a310
        path = {value = "\350+\006\001\004\001\201\303\037\002\001\000\000\000\000", len = 11, index = 0, 
          count = 0, type = 1, aid = {value = '\000' <repeats 15 times>, len = 0}}
        filelist = '\000' <repeats 1013 times>
        filelistlength = <optimized out>
        r = 0
        i = <optimized out>
        devcert = {cpi = 0, car = '\000' <repeats 16 times>, pukoid = {value = {0 <repeats 16 times>}}, 
          primeOrModulus = 0x0, primeOrModuluslen = 0, coefficientAorExponent = 0x0, 
          coefficientAorExponentlen = 0, coefficientB = 0x0, coefficientBlen = 0, basePointG = 0x0, 
          basePointGlen = 0, order = 0x0, orderlen = 0, publicPoint = 0x0, publicPointlen = 0, 
[...]

PR #1252, opened 2018-01-25
sc-hsm: fixed accessing version info

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1244

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend


Jakuje:
I had similar patch prepared to submit, since I was running into this issue from time to time.

I am not sure about the changes of the numbers, but if it comes in some specification, that these bits should be on these places (not on the last and one before last), I believe this should go in.

CardContact:
By definition, the major and minor number are always encoded in the last two bytes. Older sc-hsm version have  a shorter prop_attr, so this will break older cards.

frankmorgner:
Hmm, I didn't get this definition by reading the specification. Anyway, i changed the parsing so that it hopefully is backward compatible.

PR #1257, opened 2018-01-30
Removed unused configuration options

frankmorgner:
This cleans up some code and removes the following options:
- paranoid-memory (has never been used anyway)
- reopen_debug_file (now, always use this setting on Windows, never on other systems)
- force_card_driver (not needed, we have card_drivers)
- zero_ckaid_for_ca_certs (proprietary use, if needed pass the define CKAID_FOR_CA_CERTS)
- generalized ignored_readers

More details can be found in the commit messages.

##### Checklist

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -21,15 +21,6 @@ app default {
 	#
 	# debug_file = @DEBUG_FILE@
 
-	# Reopen debug file at the every debug message.
-	#
-	# For Windows it's forced to 'true'. The reason is that
-	#   in Windows file handles can not be shared between DLL-s,
-	#   each DLL has a separate file handle table.
-	#
-	# Default: false
-	# reopen_debug_file = true;
-
 	# PKCS#15 initialization / personalization
 	# profiles directory for pkcs15-init.
 	# Default: @PROFILE_DIR_DEFAULT@
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -27,7 +27,7 @@ app default {
 	#
 	# profile_dir = @PROFILE_DIR@;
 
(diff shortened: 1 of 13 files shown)


dengert:
Please don't  do https://github.com/OpenSC/OpenSC/pull/1257/commits/66ca9e52fd4442811fafa759443d56105bdae572 not only does it remove the `force_card_driver` from opensc.conf, it also removed the ability to use the environement variavble "OPENSC_DRIVER=" 

This is a feature that is used. It is especially needed with multiple applications on the same card right now. 

"OPENSC_DRIVER=" can be used with individual commands as you can see in some of the recent testing. Users can also use it for individual commands. Without it the user would have to have multiple copies of opensc.conf, a much more difficult process to manage.

dengert:
Never mind the above comment. I see you did add back in "OPENSC_DRIVER="

frankmorgner:
As the commit message states:

> Use `card_drivers` or `OPENSC_DRIVER` instead

i.e. `OPENSC_DRIVER` **is still** checked and using `card_drivers = openpgp;`, for example, allows you to force a card driver, too.

mouse07410:
> using `card_drivers = openpgp;`, for example, allows you to force a card driver, too.

True, but far less convenient when you're working with multiple applets from the command line, especially when testing or debugging.

`OPENSC_DRIVER=xxx` is the best for this kind of operations, so please keep it.

Jakuje:
Can this be merged (so I can extend it with #1266) or extended with the modification proposed in #1266 ?

frankmorgner:
I'd like to first make the release and then merge this PR.

Jakuje:
Ok. So I will have a look at the #1266 separately. Thank you for clarification.

PR #1259, opened 2018-02-01
pkcs15init: Fix rutokenS FCP parsing

Mardalemer:
RutokenS returns data with little endian byte order, due to this
fact token wouldn't work with standard function. So function for
parsing fcp from little endian data was inplemented.

### DIFF
--- a/src/libopensc/card-rutoken.c
+++ b/src/libopensc/card-rutoken.c
@@ -445,103 +445,28 @@ static int rutoken_select_file(sc_card_t *card,
 static int rutoken_process_fci(struct sc_card *card, sc_file_t *file,
 			const unsigned char *buf, size_t buflen)
 {
-	const unsigned char *p, *end;
-	unsigned int cla = 0, tag = 0;
-	size_t length;
+	size_t taglen;
+	int ret;
+	const unsigned char *tag;
 
-	/* Set default file properties. */
-	file->prop_attr_len = 0;
-	file->sec_attr_len = 0;
-	file->shareable = 0;
-	
-	for (p = buf, length = buflen, end = buf + buflen; p < end; p += length, length = end - p)
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	ret = iso_ops->process_fci(card, file, buf, buflen);
+	if (ret == SC_SUCCESS)
 	{
-		if (SC_SUCCESS != sc_asn1_read_tag(&p, length, &cla, &tag, &length))
-			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_CORRUPTED_DATA);
-
-		switch (cla | tag)
+		/* Rutoken S returns buffers in little-endian. */
+		/* Set correct file id. */
+		file->id = ((file->id & 0xFF) << 8) | ((file->id >> 8) & 0xFF);
+		sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "  file identifier: 0x%04X", file->id);
+		/* Determine file size. */
+		tag = sc_asn1_find_tag(card->ctx, buf, buflen, 0x80, &taglen);
+		/* Rutoken S always returns 2 bytes. */
+		if (tag != NULL && taglen == 2)
 		{
-		case 0x80:
-			/* Determine file size. */
-			if (length == 2)
-			{
-				/* Rutoken S always returns 2 bytes. */
-				/* Rutoken S returns buffers in little-endian. */
-				file->size = (p[1] << 8) | p[0];
-				sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "bytes in file: 0x%04zX", file->size);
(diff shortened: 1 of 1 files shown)


frankmorgner:
looks good. Instead of `sc_asn1_find_tag`, you could also use `sc_asn1_read_tag`, which would be more efficient...

Mardalemer:
@frankmorgner, replace `sc_asn1_find_tag` with `sc_asn1_read_tag`, looks more efficient `=)`

src/libopensc/card-rutoken.c
```diff
@@ -452,6 +442,108 @@ static int rutoken_select_file(sc_card_t *card,
...
+					file->status = SC_FILE_STATUS_CREATION;
+					break;
+				case 0x05:
+					file->status = SC_FILE_STATUS_ACTIVATED;
+					break;
+				}
+			}
+			break;
+		default:
+			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_CORRUPTED_DATA);
```
frankmorgner: Do you want to be more relaxed by ignoring an unknown tag?

frankmorgner:
... more efficient in the sense that the buffer doesn't need to be parsed multiple times. Thanks!

Your implementation looks almost identical to `iso7816_process_fci()`. Have you tried the existing implementation if it parses the FCI correctly?

Mardalemer:
There is some problem with our rutokenS, it returns little endian ordered data, so this check fails
`			case 0x80:
				/* determine the file size */
				if (sc_asn1_decode_integer(p, length, &size) == 0 && size >= 0) {
					file->size = size;
					sc_log(ctx, "  bytes in file: %"SC_FORMAT_LEN_SIZE_T"u",
							file->size);
				}
				break;
`

src/libopensc/card-rutoken.c
Mardalemer: We just want to guarantee that there are no unhandled tags

frankmorgner:
Why don't you use something like this:
```
static int rutoken_process_fci(struct sc_card *card, sc_file_t *file,
			const unsigned char *buf, size_t buflen)
{
    int r == iso_ops->process_fci(card, file, buf, buflen)
    if (r == SC_SUCCESS) {
        size_t length;
        unsigned char *p = sc_asn1_find_tag(card->ctx, buf, buflen, 0x80, &length);
        if (NULL != p && 2 == length) {
            /* Rutoken S always returns 2 bytes. */
            /* Rutoken S returns buffers in little-endian, which is different from ISO 7816-4 */
            file->size = (p[1] << 8) | p[0];
            sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL, "bytes in file: 0x%04"SC_FORMAT_LEN_SIZE_T"X", file->size);
        }
    }
    return r;
}
```
This allows code reuse, you're safe from future extensions that add currently unknown tags and you're making explicit what differs from ISO-7816-4.

Also note the use of `SC_FORMAT_LEN_SIZE_T`

Mardalemer:
The core of problem is that there is not the only tag that should be handled 
this way you've just offered. So that would lead to code duplicate. Additionally 
logging FCP tags would be incorrect what wont help in the future.

Moreover, as you can see we were using the iso7816_process_fci() function as 
well but someday this function was changed and our tokens had started to fail. 
So, to prevent this from happening again, we consider that in this case the best 
practice is to implement our specific features this way.

By the way, there are no references to byte order in ISO7816-4, so it wouldn't 
be right to say in this way /* Rutoken S returns buffers in little-endian, 
which is different from ISO 7816-4 */

frankmorgner:
Though many changes can render your card dysfunctional, I didn't see any testing or usage of this card recently. Even if you would duplicate everything from `iso7816.c`, this doesn't guarantee that your card will work without explicit testing. You should do this before every release.

`card-rutoken.c` has not received any meaningful updates since 7 years (93baf137c30bc509827c40aef6c4c28c9d06f887), so keeping the code maintainable for people without such a card is definitely needed. As consequence, we need to keep the complexity low (e.g. lines of code).

> By the way, there are no references to byte order in ISO7816-4

Well, `iso7816.c` assumes the file size is an ASN.1 Integer, which has a well defined byte order. But strictly speaking you may be right, since the version of ISO 7816-4 that I have doesn't define the object with tag `0x80` to be an Integer.

> Moreover, as you can see we were using the iso7816_process_fci() function as
well but someday this function was changed and our tokens had started to fail.

Parsing the file size in `iso7816_process_fci()` hasn't changed in quite some time, as far as I know. So I wonder what change caused the problem...

> The core of problem is that there is not the only tag that should be handled
this way you've just offered.

What else is parsed differently? From the source I cannot see anything else.

Long story short, I still hold my point: please use `iso7816_process_fci()` as good as you can.

Mardalemer:
IMHO our solution would be more stable for future changes, but we see no problem to use your variant.

frankmorgner:
thank you! please make sure to test your card when the upcoming release is about to be released.

PR #1264, opened 2018-02-07
Workaround for OpenPGP problems introduced with #1243

frankmorgner:
##### Checklist

- [x] Tested with the following card: OpenPGP 3.3
	- [x] tested PKCS#11

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2890,6 +2890,7 @@ static int pgp_card_reader_lock_obtained(sc_card_t *card, int was_reset)
 			/* reselect is required */
 			was_reset = 1;
 		}
+		r = SC_SUCCESS;
 	}
 
 	if (was_reset > 0) {
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -575,9 +575,12 @@ pgp_get_card_features(sc_card_t *card)
 				priv->max_challenge_size = bebytes2ushort(blob->data + 2);
 				/* max. cert size it at bytes 5-6 */
 				priv->max_cert_size = bebytes2ushort(blob->data + 4);
-				/* max. send/receive sizes are at bytes 7-8 resp. 9-10 */
-				card->max_send_size = bebytes2ushort(blob->data + 6);
-				card->max_recv_size = bebytes2ushort(blob->data + 8);
+				if (priv->bcd_version < OPENPGP_CARD_3_0) {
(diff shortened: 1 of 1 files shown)


mouse07410:
This PR seems to break PIV.

Without it:
```
$ opensc-tool -nal
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Gemalto Prox Dual USB PC Link Reader(2)
1    No              Gemalto Prox Dual USB PC Link Reader(1)
Using reader with a card: Gemalto Prox Dual USB PC Link Reader(2)
3b:7a:18:00:00:73:66:74:65:20:63:64:31:34:34
$ 
```

With it:
```
$ opensc-tool -nal
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Gemalto Prox Dual USB PC Link Reader(2)
1    No              Gemalto Prox Dual USB PC Link Reader(1)
Using reader with a card: Gemalto Prox Dual USB PC Link Reader(2)
Failed to connect to card: Incorrect parameters in APDU
$
```

mouse07410:
As I'm also playing with #1256, I need more testing.

dengert:
@mouse07410  where is the opensc-debug.log for the `Failed to connect to card: Incorrect parameters in APDU` case?

dengert:
I am not sure if is the correct solution
 https://github.com/OpenSC/OpenSC/pull/1264/commits/f3b27fc874594bdda9115cbe64607ff36de6d11f

If the  `r = card->ops->card_reader_lock_obtained(card, was_reset);`
is < 0, then the card_reader_lock_obtained may have some serious problems with the card, and it was not able to select the correct AID.  Ignoring that type of error could cause serious problems if the two applets on the card can execute the subsequent command but in the wrong applet.  (With OpenPGP and PIV, the commands of two applets do not overlap as far as I can tell.) 

The problem is that a `card_reader_lock_obtained` needs to return 0 or  and error less then zero,
and the openpgp_card_reader_lock_obtained was returning 16  the length of the get data of the  "4F" 

the sc_lock  code will also need some changes if the `r = card->ops->card_reader_lock_obtained(card, was_reset);` returns less then zero, as the sc_lock has already obtained the pcsc_lock, and it need to release lock before returning that there was an error in sc_lock caused by  `r = card->ops->card_reader_lock_obtained(card, was_reset);`

A *_card_reader_lock_obtained should be were the choice to ignore error from commands it issues,
and return 0 on its own.

frankmorgner:
@dengert You're right: I hopefully fixed the problem on the OpenPGP side now (hopefully I've got time for testing next week)

Issue #1266, opened 2018-02-08
OPENSC_DRIVER environment variable prevents parsing configuration file


Jakuje:
### Problem Description
From discussion in the #1243 it appears very confusing for many people that once the `OPENSC_DRIVER` environment variable is used, NO configuration file is parsed and therefore the ATR matches with flags implemented in there are not effective.

### Proposed Resolution
The configuration files should be parsed regardless this variable. It should only overwrite the drivers option, defined in the configuration file.

If we agree that this should be the way to go, I can have a look into possible implementation.

Jakuje:
Reading through the source code, one thing I do not completely understand is that there are two configuration options:

 * `card_drivers`, which affect which drivers are loaded and tried
 * `force_card_driver`, which selects only one driver, that is always selected regardless whatever is in the rest of configuration file (such as the ATRs we have problem with in the OpenPGP x PIV-II issues).

The first option looks like it was here for some time (cd4e3650a, Mar 2002), and the second option was authored by jey in Apr 2002 (9c39ca7e6).

The environment variable that is connected to the second option (`force_card_driver`) was added quite recently (#882) in 2016 by @lbschenkel, probably unaware of these side effects.

### Proposal
Long story short, my proposal would be to get rid of the `force_card_driver` configuration option (It can not do anything that could not be done with `card_drivers`, isn't it?) and hook the environment variable to the `card_drivers` option instead, overwriting the content of the configuration file. I didn't do that, but I believe I can give it a shot in coming days.

Are you against this approach for some reason? Do I miss some use case for this configuration option? @frankmorgner @dengert @mouse07410 @viktorTarasov ?

I also see that there is no Windows alternative as it was discussed in the #882. This is probably also a room for improvement, even though I have no way how to test the final product.

Also I see that this is nowhere documented. After we will resolve this, we should really update at least [Environment-variables](https://github.com/OpenSC/OpenSC/wiki/Environment-variables) wiki page. Additionally, it should be described someplace in configuration file, what does it do.

mouse07410:
I believe these two options aren't supposed to be used simultaneously. But it's clear that in some cases  I may want to "force" exactly one driver, while in other cases I want to restrict the set of drivers to a few. 

Thus, IMHO both should be there, maybe just better documented.

Jakuje:
Am I missing something, or restricting the drivers to the single driver is something else than forcing single driver?

mouse07410:
Ah, OK. Yes you're right.

[...]

PR #1277, opened 2018-02-26
OPENSC_DRIVER: restrict driver list instead of forcing one

Jakuje:
Using the forced-driver prevents parsing of additional constructions
in configuration files (for example flags based on ATRs). This
implementation replaces transparently the existing list defined in
card_drivers.

Resolves: #1266

##### Checklist

- [x] Tested with the following card: PIV-II, openpgp (Yubikey)
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend


frankmorgner:
@Jakuje I needed to commit a merge, could you check if that's OK for you?

Jakuje:
The merge commit looks sane and basic check confirms it behaves as I would expect:
```
$ ./src/tools/pkcs11-tool -L --module src/pkcs11/.libs/opensc-pkcs11.so 
Available slots:
Slot 0 (0x0): OMNIKEY AG CardMan 3121 00 00
  token manufacturer : Common Access Card
$ OPENSC_DRIVER=coolkey ./src/tools/pkcs11-tool -L --module src/pkcs11/.libs/opensc-pkcs11.so 
Available slots:
Slot 0 (0x0): OMNIKEY AG CardMan 3121 00 00
  (empty)
$ OPENSC_DRIVER=cac ./src/tools/pkcs11-tool -L --module src/pkcs11/.libs/opensc-pkcs11.so 
Available slots:
Slot 0 (0x0): OMNIKEY AG CardMan 3121 00 00
  token manufacturer : Common Access Card
```

frankmorgner:
thanks

PR #1280, opened 2018-03-01
Fix DNIe integration

igorcferreira:
This PR aims to fix the compilation of the project when using --enable-dnie-ui on a macOS system

##### Checklist

- [x] Tested with the following card: 
	- [x] tested PKCS#11
	- [x] tested macOS Tokend

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -450,14 +450,12 @@ if test "${enable_dnie_ui}" = "yes"; then
 			LDFLAGS="${LDFLAGS} -framework Carbon"
 		;;
 	esac
+	case "${host}" in
+		*-apple-*)
+			LDFLAGS="${LDFLAGS} -framework CoreFoundation"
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -446,14 +446,19 @@ if test "${enable_dnie_ui}" = "yes"; then
...
 		;;
 	esac
 fi
 
+
+case "${host}" in
+	*-apple-*)
+		LDFLAGS="${LDFLAGS} -framework CoreFoundation"
+	;;
+esac
```
frankmorgner: why don't you put this where the test for the UI is?

igorcferreira:
@frankmorgner Refactored. I'd put it out of the ui test because I tough that the CoreFoundation framework could relevant for any implementation to be applied on Mac. But, moved it to its proper place.

PR #1284, opened 2018-03-13
epass2003: Fix buffer underrun in epass2003_decipher(), etc.

vvp2014:
1. Buffer underrun in epass2003_decipher().
2. The parameter `data' in update_secret_key() must be constant.

(Discovered by Clang 4.0.0 on OpenBSD 6.2.)

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -72,6 +72,9 @@ static struct sc_card_driver epass2003_drv = {
 #define KEY_LEN_DES3	24
 #define HASH_LEN	24
 
+#define SHA1_DIGEST_LENGTH 20
+#define SHA256_DIGEST_LENGTH 32
+
 static unsigned char PIN_ID[2] = { ENTERSAFE_USER_PIN_ID, ENTERSAFE_SO_PIN_ID };
 
 /*0x00:plain; 0x01:scp01 sm*/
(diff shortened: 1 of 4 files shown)


src/libopensc/card-epass2003.c
```diff
@@ -72,6 +72,9 @@ static struct sc_card_driver epass2003_drv = {
 #define KEY_LEN_DES3	24
 #define HASH_LEN	24
 
+#define SHA1_DIGEST_LENGTH 20
+#define SHA256_DIGEST_LENGTH 32
```
Jakuje: `SHA_DIGEST_LENGTH` and `SHA256_DIGEST_LENGTH` are already defined in the `openssl/sha.h` included above.

vvp2014: Well, then please feel free to delete lines 75--76.

I forgot to note that (while I do not have the card and I did not follow up the stack) bugs of this kind sometimes have security implications.  Perhaps it should be noted in the change log.

frankmorgner:
I've simplified and merged the commits as 1ac08f96e762d98ef74f8599c71ee5eaf588e82a and e1d6f0869bdac873878d72044bfc44d6ca5373b4

PR #1287, opened 2018-03-18
pkcs11-tool: send DER for EC public keys with default compilation flags

### LINKED ISSUE #1286: RFE: make EC_POINT_NO_ASN1_OCTET_STRING a runtime option in pkcs11-tool
### Problem Description

Version: RHEL packaging opensc-0.16.0-5.20170227git777e2a3.el7.x86_64

For pkcs11-tool --write of EC public keys SafeNet HSM requires the full DER (including TAG, LENGTH). Depending on packagers option we send full DER or just content bytes. 

### Proposed Resolution
Command line option for pkcs11-tool to enable or disable DER TAG/LENGTH on writing EC public keys

### Steps to reproduce
1. opensc compiled without EC_POINT_NO_ASN1_OCTET_STRING; e.g. CentOS/RHEL packaging
2. pkcs11-tool --write ec_public_key.der will fail with SafeNet HSM with invalid EC_POINT
3. rebuild package with EC_POINT_NO_ASN1_OCTET_STRING defined, then the --write works

### See also
https://github.com/OpenSC/OpenSC/issues/1285

aalba6675:
[UPDATE] PR has been modified according to @DDvO's comment.
By default EC_POINT_NO_ASN1_OCTET_STRING is undefined and you will get standards behaviour.
Define this to get the write to send plain bytes.

-----------------------
This PR specifically fixes the default behaviour of `pkcs11-tool`. Makefile defaults to an undefined EC_POINT_NO_ASN1_OCTET_STRING and therefore EC public keys are written as plain bytes instead of with DER wrapping. 
1. https://github.com/OpenSC/libp11/issues/79
2. https://bugzilla.mozilla.org/show_bug.cgi?id=480280
 This will make pkcs11-tool consistent with the standard.

- Fixes #1286: "Does it even make sense to write EC keys not in DER?"
- the naming of the define was confusing, anyway

##### Checklist

- [x] Tested with the following card: 
	- [x] tested PKCS#11 (SafeNet HSM)
	- [ ] tested macOS Tokend

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2632,7 +2632,7 @@ parse_ec_pkey(EVP_PKEY *pkey, int private, struct gostkey_info *gost)
 		header_len = point-gost->public.value;
 		memcpy(point, buf, point_len);
 		gost->public.len = header_len+point_len;
-#ifndef EC_POINT_NO_ASN1_OCTET_STRING // workaround for non-compliant cards not expecting DER encoding
+#ifdef EC_POINT_NO_ASN1_OCTET_STRING // workaround for non-compliant cards not expecting DER encoding
 		gost->public.len   -= header_len;
 		gost->public.value += header_len;
 #endif


dengert:
PKCS#11 early version were not clear . Later versions do what OpenSC dies.
Talk ro SafeNET

aalba6675:
@dengert  - can I confirm that you mean the default behaviour of OpenSC (EC_POINT_NO_ASN1_OCTET_STRING undefined in Makefile) means that ECPoint should be written with content bytes only (no ASN.1 TAG/LENGTH).

CKA_EC_POINT is defined as v2.40: "DER-encoding of ANSI X9.62 ECPoint value Q" and in RFC5480 ECPoint ::= OCTET STRING which seems to imply the TAG/LENGTH as well.  Is there a good reference that says write this OCTET STRING without the TL-part?

DDvO:
Looking again at the condition I implemented in June 2016:

`#ifndef EC_POINT_NO_ASN1_OCTET_STRING`

I now think that I did a mistake and I should have written

`#ifdef EC_POINT_NO_ASN1_OCTET_STRING`

instead and the default behavior would already have been the standard-compliant one.
So instead of removing the conditional workaround as proposed in this PR,
it could also be kept and just corrected to

```
#ifdef EC_POINT_NO_ASN1_OCTET_STRING // workaround for non-compliant cards not expecting DER encoding
                gost->public.len   -= header_len;
                gost->public.value += header_len;
#endif

```

frankmorgner:
@DDvO thanks for the clearification.

dengert:
I would vote  remove it. PKCS#11 says use DER. And that is what OpenSC
does. If a card does not use DER, then the driver and PKCS#11 module are
responsible to to add or strip DER for the card.  The calling application
(pkcs11-tool) should not have to deal with this.

It should not be an OS issue either.

From the discussion it sounds like there was some PKCS#11 module at onetime
that was not compliant. Is it still being used? Has it been updated by the
vendor?

(I can't do much to help with this till late next week.)

DDvO:
Good point that the driver should be responsible for dealing with any DER encoding incompatibility.
Yet the driver itself could be faulty in this respect.

Back in Aug 2016 I was testing a beta CardOS 5.3 middleware with a beta CardOS 5.0 card.
Meanwhile I'm pretty far away from this topic and cannot say if this still applies and what might have changed in newer driver/module versions.

dengert:
So are you saying the original patch was for beta code for CardOS?

Is the code still in beta?

If the CardOS was fixed, and the OpenSC PR is no longer needed we should
revert the original OpenSC PR.

?

DDvO:
I've just tested writing an EC public key with the latest CardOS PKCS#11 module I have (of Marc 2017):
```
  V5.4 (Build 07) Beta Version
  CardOS-API-V5-4-7-512-20170222112927
```
 to the latest card I have:
```
  token model        : CardOS V5.3, 201
  hardware version   : 120.0
  firmware version   : 201.3
```
Unfortunately, without the workaround I still get:
```
error: PKCS11 function C_CreateObject failed: rv = CKR_DEVICE_ERROR (0x30)
```

So thanks @aalba6675 for adapting your PR such that the workaround can be enabled easily when needed.

dengert:
I object to the way this is implemented, It is a compile time change to an OpenSC utility to support a non standard PKCS#11 module that is still Beta. It means pkcs11-tool will not work with other complaint modules including the OpenSC modules. I realize that it is for EC only. 

The PR should be rewritten to make it an run time option or to base it on the  specific version of the module.

DDvO:
@dengert, it looks like you misunderstood the nature of the proposed change.
So far, the non-compliant variant has been the default, and by the change proposed here this workaround will not any more be the (compile-time) default.

BTW, the non-compliant behavior of CardOS (and presumably some other cards/drivers, too) has been around for long time in various (also non-beta) versions and their use is apparently wide-spread.

After the change, users who need the workaround will need to compile with `EC_POINT_NO_ASN1_OCTET_STRING` defined manually. 
I don't know if it would be worth adding a run-time option for enabling the workaround.

dengert:
What I am worried  about is some distro will compile with EC_POINT_NO_ASN1_OCTET_STRING  
That is why I would like to make sure this is not a compile time option.

aalba6675:
Indeed this has happened — in the opposite sense. RHEL did not define this in its packaging and as a result the pkcs11-tool in RHEL is sending plain bytes instead of DER.

DDvO:
Oops. Luckily this only affects sending EC public keys to a token - a pretty rare scenario.
BTW, receiving exportable private keys (of any type) is still not supported at all by the pkcs11-tool.

After the change proposed in this PR, I'd consider the risk that someone accidentally enables the encoding workaround by defining `EC_POINT_NO_ASN1_OCTET_STRING` pretty low. In particular since this flag is hardly known to anyone.

DDvO:
After some other point made above, I just suggest improving the comment 
`
// workaround for non-compliant cards not expecting DER encoding
`
to
`
// optional workaround for non-compliant tokens/drivers/modules expecting DER encoding
`

PR #1288, opened 2018-03-18
Add support for fixed-length pinpad readers

### LINKED ISSUE #1221: Gemalto PC Pinpad not working with OpenPGP card
### Problem Description

The Gemalto PC Pinpad does not supported variable length pins. This has been solved for some card drivers but not for openpgp cards. The gnupg solution is to store the PIN lengths in a DO object on the card. More information can be found here: https://wiki.gnupg.org/CardReader/GemaltoPC

### Proposed Resolution
card-openpgp.c should check for the login-data DO and the specific format if a Gemalto card reader is present.

### Steps to reproduce

1. Perform any operation with pkcs11-tool
2. Error returned is:

```
Aborting.

```

Thanks!

amaccuish:
Fixes #1221 

Some pinpad readers have to be told the length of the PIN to accept. OpenPGP has a DO for storing the length. This patch reads the and parses the DO if it is properly configued, and supplies it for use for APDUs.
- Tested with OpenSC and OpenPGP 2.1 card on MacOS and Linux (Ubuntu and Fedora)

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -1635,7 +1635,7 @@ pgp_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 						if (blob->len) {
 							if (digitp (blob->data)) {
 								char *q;
-								size_t n, m;
+								unsigned int n, m;
 		
 								n = strtol ((char*)blob->data, &q, 10);
 								if (q >= (char *)blob->data + blob->len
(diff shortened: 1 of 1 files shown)


frankmorgner:
How did you test this? Did you test this also without PIN pad?

frankmorgner:
Your code doesn't compile

amaccuish:
Sorry committed wrong version, will up date

amaccuish:
Sorry didn't check x86, now compiles :)

Issue #1273, opened 2018-02-22
SEGV in ASN1 code with GIDS card


fishsoupisgood:
### Problem Description

with current master [1d4f59ea51dd1713a3b99186498cb8fe074abc4c](https://github.com/OpenSC/OpenSC/commit/1d4f59ea51dd1713a3b99186498cb8fe074abc4c) and current master of the GidsApplet [7ba0d7b8ad5368cbdbe988164c90fc26744c5e0c](https://github.com/vletoux/GidsApplet/commit/7ba0d7b8ad5368cbdbe988164c90fc26744c5e0c) on a J3A081 I encounter a SEGV when running pkcs11-tool.

[gids_get_pin_status](https://github.com/OpenSC/OpenSC/blob/1d4f59ea51dd1713a3b99186498cb8fe074abc4c/src/libopensc/card-gids.c#L533) calls (reading upto 100 bytes into buffer)

       r = gids_get_DO(card, GIDS_APPLET_EFID, dataObjectIdentifier, buffer, &buffersize);

APDU: 00 CB 3F FF 04 5C 02 7F 71 00
SW: 7F 71 06 97 01 03 93 01 03 90 00

after which buffer is: 97 01 03 93 01 03 00
and buffersize is 6

it then calls 

        p = sc_asn1_find_tag(card->ctx, buffer, sizeof(buffer), GIDS_TRY_COUNTER_OLD_TAG, &datasize);

giving sizeof(buffer) rather than 6 as the length of data to parse.

[sc_asn1_find_tag](https://github.com/OpenSC/OpenSC/blob/1d4f59ea51dd1713a3b99186498cb8fe074abc4c/src/libopensc/asn1.c#L473) which parses the ASN1 in a loop:

* iteration 0 finds: 97 01 03  (class 0x80, tag= 0x17, length=1, value=3)
* iteration 1 finds: 93 01 03 (class 0x80, tag=0x13, length=1, value=3)
* iteration 2: *p is now 00 (by luck, we've read past the end of initialized data), and the call to [sc_asn1_read_tag](https://github.com/OpenSC/OpenSC/blob/1d4f59ea51dd1713a3b99186498cb8fe074abc4c/src/libopensc/asn1.c#L71)  sets p to be NULL and returns. left now gets set to a ludicrous value
* iteration 3: p is now NULL and the call to sc_asn1_read_tag SEGVs.

<pre>
(gdb) run -L
Starting program: /usr/bin/pkcs11-tool -L
[...]

PR #1289, opened 2018-03-19
fixed handling SC_ASN1_TAG_EOC from sc_asn1_read_tag

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1273

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -369,7 +369,7 @@ static void print_tags_recursive(const u8 * buf0, const u8 * buf,
 			return;
 		}
 		hlen = tagp - p;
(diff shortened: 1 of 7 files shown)


frankmorgner:
The problem is that `SC_ASN1_TAG_EOC` is defined to `0`, so testing for `tag == SC_ASN1_TAG_EOC` will also be true for the valid constructed tag `0x80`. So we need to switch to @fishsoupisgood['s fix](https://gist.github.com/fishsoupisgood/f4575d7cc8db2e48dae1bac356e9e0b6)

PR #1290, opened 2018-03-19
Fix two bugs 

FeitianSmartcardReader:
1. Change the APDU of signature
2. Fix bug when get pin status


frankmorgner:
The first two commits are merged already!

Please be more verbose on what your fixing with e11598d1291cfc35a3a78c9398875952e851fbb2 and ideally, split this into two separate commits, one for each issue.

FeitianSmartcardReader:
will do it today, thanks

Issue #1296, opened 2018-03-22
Cleanup belpic driver


bigon:
Hello,

The Belgian government is developing its own driver (complete rewrite) for the belpic/BeID cards.

So I guess all the legacy cruft (HAVE_GUI and BELPIC_PIN_PAD) could be removed as well as all the "language" related functions

frankmorgner:
OR we remove/disable the belpic driver entirely since it is unmaintained and there is a better alternative available...

frankmorgner:
Packages for any platform are available from the [officiel homepage](https://eid.belgium.be/en). The convienience that you demand needs to be requested from @Fedict, not from @OpenSC by shipping unmaintained code.

frankmorgner:
@yoe what do you think? Are there any plans to get eid-mw into the distributions?

bigon:
"any platform" == Windows, MacOS and GNU/Linux (i386 and amd64 only). Debian compiles on 11+ architectures so it's far from covering everything.

Anyway, the cleaned up code is at https://github.com/bigon/OpenSC/commit/f8290e5c449f7af01a92a9d3af3f6171ee2a5be5

yoe:
@frankmorgner I have no such plans, since there are repositories and that keeps people a bit more up to date; but that doesn't mean there will not ever be such plans by other people.

frankmorgner:
@bigon please make a pull request.

Could you check if there is a "user friendly" way to say that non repudiation signature is not supported anymore? Or - my favourite - can you change the code to simply ignore the consent via GUI and to just perform the signature?

PR #1298, opened 2018-03-23
Clean up the belpic driver

bigon:
Remove all the code related to the old GUI and PIN PAD.

This code was initially developed by Zetes and had the ability to
display a dialog to request the PIN to the user. It was also able to
manage some specific proprietary pin pads.

As the Belgian government/fedict has now its own implementation, all
these old crufts can be removed.

https://github.com/OpenSC/OpenSC/issues/1296

##### Checklist

- [x] Tested with the following card: Belpic cards
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend


frankmorgner:
Looks ok, do you think that we can handle the non-repudiation key more gracefully?

Issue #1267, opened 2018-02-13
Manual pages should document all commandline switches


Jakuje:
### Problem Description

Some time ago, I was adding missing command-line switches into manual page for pkcs11-tool and since then, there are already new ones without respective documentation added. In addition, there are few more tools with missing options in manual pages.

I updated my script to generate the missing configuration options list to go through all the tools and I propose to add it as a part of CI process to make sure the new tools are documented properly:

https://github.com/Jakuje/stuff/blob/master/check_opensc_manual.sh

I don't have a lot experience with github CI integration, but initially, it can be simple stage of `make check` in travis or whatever. I am open to proposals. The above bash script can be very simply adjusted to work there without absolute paths. I believe that would be more effective then thinking about it while reviewing every PR.

Second part, that probably needs to be done before is fixing the existing manual pages. The list of missing options is below (except some of the `pkcs11-tool` ones that I have already committed and I will propose in some future PR. Additionally, this can be easy-to-fix task for new contributors.

This would certainly be good to have before 0.18.0 release.

```
opensc-tool: missing switch --get-conf-entry
opensc-tool: missing switch -G
opensc-tool: missing switch --set-conf-entry
opensc-tool: missing switch -S
opensc-tool: missing switch --list-algorithms
pkcs15-init: missing switch -P
pkcs15-init: missing switch --store-data
pkcs15-init: missing switch -W
pkcs15-init: missing switch --delete-objects
pkcs15-init: missing switch -D
pkcs15-init: missing switch --change-attributes
pkcs15-init: missing switch -A
pkcs15-init: missing switch --sanity-check
pkcs15-init: missing switch --erase-application
pkcs15-init: missing switch --reader
pkcs15-init: missing switch -r
pkcs15-init: missing switch --no-so-pin
pkcs15-init: missing switch --serial
pkcs15-init: missing switch --puk-id
pkcs15-init: missing switch --puk-label
pkcs15-init: missing switch --public-key-label
pkcs15-init: missing switch --cert-label
pkcs15-init: missing switch --application-name
pkcs15-init: missing switch --application-id
pkcs15-init: missing switch --aid
pkcs15-init: missing switch --output-file
pkcs15-init: missing switch --passphrase
pkcs15-init: missing switch --authority
pkcs15-init: missing switch --key-usage
pkcs15-init: missing switch --finalize
pkcs15-init: missing switch -F
pkcs15-init: missing switch --update-last-update
[...]

PR #1310, opened 2018-04-03
Make the build pass after merging #1205 and other minor issues

Jakuje:
This is another bunch of changes that accumulated, that were to small to PR separately, so I am sorry for not doing separate PRs in advance. I can describe what they do separately, but commit messages should be self-explanatory.

##### Checklist
- [X] Tested with the following card: PIV-II
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend


Jakuje:
I see the Travis CI test fail, but it does not prevent the build itself from passing, because it is run in the script as

    make check; make dist

Should not this rather say the following?

    make check && make dist

In that case, we would need to adjust the tests to temporarily pass or to fix #1267.

PR #1315, opened 2018-04-06
reader-pcsc: allow fixing the length of a PIN

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1221
closes https://github.com/OpenSC/OpenSC/pull/1288

##### Checklist

- [x] Tested with the following card: OpenPGP
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -103,6 +103,12 @@ app default {
 		# Default: true
 		# enable_pinpad = false;
 		#
+		# Some pinpad readers can only handle one exact length of the PIN.
+		# fixed_pinlength sets this value so that OpenSC expands the padding to
+		# this length.
+		# Default: 0 (i.e. not fixed)
+		# fixed_pinlength = 6;
+		#
 		# Detect reader capabilities with escape commands (wrapped APDUs with
 		# CLA=0xFF as defined by PC/SC pt. 3 and BSI TR-03119, e.g. for getting
 		# the UID, escaped PIN commands and the reader's firmware version)
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -82,6 +82,7 @@ struct pcsc_global_private_data {
 	SCARDCONTEXT pcsc_ctx;
 	SCARDCONTEXT pcsc_wait_ctx;
 	int enable_pinpad;
+	int fixed_pinlength;
 	int enable_pace;
 	size_t force_max_recv_size;
 	size_t force_max_send_size;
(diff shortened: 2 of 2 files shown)


Jakuje:
This looks like a good solution for broken readers with pinpads. I was recently struggling with something similar with various Dell and HP keyboards with smart card readers and the only thing I came up with was disabling the pinpad detection.

If the user/administrator will know how long the pin is, this seems like a better solution security-wise. 

I personally don't have these keyboards so I can not verify the functionality of the change, but it looks good to me.

amaccuish:
Will compile and check later today against my reader that has this problem

amaccuish:
Tested on Linux using pkcs11, worked great!

amaccuish:
Just got a hold of a Dell keyboard and sadly this doesn't work with it :/ not for openpgp or estonian cards. Error returned is:
Aborting.

Weirdly gnupg works.

dengert:
Google for: dell keyboard smart card reader driver linux

The reader is non standard. Maybe one of the articles can help.

Jakuje:
I would be surprised if scdaemon from gnupg would support pinpad readers.

I will be ordering one of these readers soon too to try to figure out what is the matter with them. APDU traces might be helpful, but I do not think there is anything we could fix more inside of OpenSC, since there is no sensible way how to figure out how long the pin is going to be.

From what I remember last time reading the discussion about this reader, it required to pass zeroes of the length that is expected length of the pin. But the attached patch here uses `scconf_get_bool()`, which can really return only 0 or 1. Should not there be used `scconf_get_int()` function instead (sorry for not noticing earlier)?

amaccuish:
@Jakuje thanks for getting back. Yes gnupg works fine, it pops up a message saying please use the pinpad to enter the pin, I do so on my keyboard on the numeric keypad and hit enter, and everything proceeds normally. I'll take a trace when I'm back tonight. I do agree with you it's odd that gnupg appears to have few problems with pinpads.

Possibly this is a specific issue with OpenPGP cards? I know gnupg will read hints from the card to determine the PIN length, I suggested adding this to OpenSC but I think it was rejected.

Jakuje:
Ok, then we should be able to do this separately in the opengpg driver. The problem with #1288 was that it was quite cumbersome way of doing this, but if you would reworked it a bit after the @frankmorgner comments, it might be acceptable.

Recently @marschap is also working on improvements of the opengpg driver so he might also have some comments or interest in having this use case working.

amaccuish:
@Jakuje the problem is that this really has to be a card specific thing since for example, with OpenPGP cards, the pin can be any length, with a minimum of 6 for user pins and 8 for SO pins, and hints about this can be stored in a DO. But for example my estonian card has 4 for the auth pin and 6 for signing. This of course can be ignored if we know the reader supports variable length pins, but if not then we need card specific help no?

amaccuish:
So I try tried my OpenPGP specific patch again and it worked. I'll work on rebasing and tidying it up.

PR #342, opened 2014-12-18
iso7816: Interpret successful (but empty) response to SELECT as success.

darconeous:
Some apps, like the OpenPGP app on the Yubikey NEO, respond
only indicating success. Allowing such responses allows OpenSC to
work with such apps.

PR #1316, opened 2018-04-06
card-muscle: be more relaxed in card initialization

frankmorgner:
closes https://github.com/OpenSC/OpenSC/pull/1248

##### Checklist

- [x] Tested with the following card: muscle
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -94,9 +94,11 @@ static int muscle_match_card(sc_card_t *card)
 		apdu.resp = response;
 		r = sc_transmit_apdu(card, &apdu);
 		if (r == SC_SUCCESS && response[0] == 0x01) {
-				card->type = SC_CARD_TYPE_MUSCLE_V1;
-				return 1;
+			card->type = SC_CARD_TYPE_MUSCLE_V1;
+		} else {
+			card->type = SC_CARD_TYPE_MUSCLE_GENERIC;
 		}
+		return 1;
 	}
 	return 0;
 }
@@ -476,17 +478,16 @@ static int muscle_init(sc_card_t *card)
 	card->caps |= SC_CARD_CAP_RNG;
 
 	/* Card type detection */
-	if (_sc_match_atr(card, muscle_atrs, &card->type) < 0)   {
-		free(priv->fs);
-		free(card->drv_data);
-		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, SC_ERROR_NOT_SUPPORTED);
-	}
-
+	_sc_match_atr(card, muscle_atrs, &card->type);
 	if(card->type == SC_CARD_TYPE_MUSCLE_ETOKEN_72K) {
 		card->caps |= SC_CARD_CAP_APDU_EXT;
 	}
 	if(card->type == SC_CARD_TYPE_MUSCLE_JCOP241) {
 		card->caps |= SC_CARD_CAP_APDU_EXT;
+	}
+	if (!(card->caps & SC_CARD_CAP_APDU_EXT)) {
+		card->max_recv_size = 255;
+		card->max_send_size = 255;
 	}
 	if(card->type == SC_CARD_TYPE_MUSCLE_JCOP242R2_NO_EXT_APDU) {
 	        /* Tyfone JCOP v242R2 card that doesn't support extended APDUs */
--- a/src/libopensc/ctx.c
+++ b/src/libopensc/ctx.c
@@ -96,20 +96,23 @@ static const struct _sc_driver_entry internal_card_drivers[] = {
 #endif
 	{ "rutoken",	(void *(*)(void)) sc_get_rutoken_driver },
 	{ "rutoken_ecp",(void *(*)(void)) sc_get_rtecp_driver },
-	{ "westcos",	(void *(*)(void)) sc_get_westcos_driver },
 	{ "myeid",      (void *(*)(void)) sc_get_myeid_driver },
-	{ "sc-hsm",		(void *(*)(void)) sc_get_sc_hsm_driver },
 #if defined(ENABLE_OPENSSL) && defined(ENABLE_SM)
 	{ "dnie",       (void *(*)(void)) sc_get_dnie_driver },
 #endif
 	{ "masktech",	(void *(*)(void)) sc_get_masktech_driver },
+	{ "atrust-acos",(void *(*)(void)) sc_get_atrust_acos_driver },
+	{ "westcos",	(void *(*)(void)) sc_get_westcos_driver },
 
-/* Here should be placed drivers that need some APDU transactions to
- * recognise its cards. */
+/* Here should be placed drivers that need some APDU transactions in the
+ * driver's `match_card()` function. */
+	/* MUSCLE card applet returns 9000 on whatever AID is selected, see
+	 * https://github.com/JavaCardOS/MuscleCard-Applet/blob/master/musclecard/src/com/musclecard/CardEdge/CardEdge.java#L326
+	 * put the muscle driver first to cope with this bug. */
+	{ "muscle",	(void *(*)(void)) sc_get_muscle_driver },
+	{ "sc-hsm",	(void *(*)(void)) sc_get_sc_hsm_driver },
 	{ "mcrd",	(void *(*)(void)) sc_get_mcrd_driver },
 	{ "setcos",	(void *(*)(void)) sc_get_setcos_driver },
-	{ "muscle",	(void *(*)(void)) sc_get_muscle_driver },
-	{ "atrust-acos",(void *(*)(void)) sc_get_atrust_acos_driver },
 	{ "PIV-II",	(void *(*)(void)) sc_get_piv_driver },
 	{ "cac",	(void *(*)(void)) sc_get_cac_driver },
 	{ "itacns",	(void *(*)(void)) sc_get_itacns_driver },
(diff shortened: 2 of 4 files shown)


lucasmpalma:
The fix really worked, i tested with a couple of java cards running the last version of Muscle Applet. I am using a java card with the following atr 3b:db:96:00:80:b1:fe:45:1f:83:00:31:c0:64:1a:18:01:00:0f:90:00:52, i had to set the driver as "muscle" at the opensc.conf file.

frankmorgner:
the applet should be enough to detect your card, changing the configuration should not be needed...

dengert:
If a driver is forced, the results of muscle_match_card are ignored,  card.c:253:
`sc_log(ctx, "driver '%s' match_card() failed: %s (will continue anyway)", card->driver->name, sc_strerror(r)); 

If the driver is not forced,  the results of  muscle_match_card  is tested. 

Without the mode muscle_match_card only returns 1 only  if `if (r == SC_SUCCESS && response[0] == 0x01) {` is true.  With the card in question, the ADPU could have failed, or the response is on 0x01. 

With  https://github.com/OpenSC/OpenSC/pull/1316/commits/efc27fc450dff4192824cbfece0570cb001ae328 `
`if (r == SC_SUCCESS && response[0] == 0x01) {` is tested, but even if it fails, `return 1` is executed.
Not clear if card->type was changed or not. 

muscle_init is called in both  cases, (forced or if muscle_match_card returned 1).

I suspect the the ADPU failed, or the response[0] is not 0x01. 

`opensc_debug.log`'s from the failing and working configurations with or without the mod would help.

dengert:
And the mod overwrites the forced `card->type.`

frankmorgner:
Selection of the applet is mandatory, detecting the version is optional, that's what is implemented in `muscle_match_card()`, which looks fine.

@lucasmpalma could you send a log of the output of `opensc-tool -n -vvv` *without forcing the muscle driver*? Maybe this reveals the problem.

dengert:
Looks like card-sc-hsm.c is using an AID of `0xE8,0x2B,0x06,0x01,0x04,0x01,0x81,0xC3,0x1F,0x02,0x01`

According to ISO-7816 12.2.3 Application identifier , Table 121 — Categories of application identifiers this is a "'E' Standard Identification of a standard by an object identifier according to ISO/IEC 8825-1"
then later says:
"'E8' Object identifier (see Annex A) Application-specific application identifier extension"
Annex A only talks about "28" and not "2B" 

I can not quite decode it, but it looks like this may not be unique and the card in question either supports this AID or is treating it as a file or ignoring it by returning 90 00. 

Maybe the SmartCardHSM people can shed some light on this AID.

dengert:
Further searching the code, it looks like card-sm-hsm.c  will parse the returned FCI and expects to receive a valid FCI. 
**But** https://github.com/OpenSC/OpenSC/commit/3c1d8ad69557adf6aab08bc549188425c8b33a92 from 2015-02-07 has this comment ` /* For some cards 'SELECT' MF or DF_NAME do not return FCI. */` It goes on to create an FCI based on the requested AID to make it look like the SELECT AID actually returned the expected result.  Thus as best as I can tell, this is a false positive, because the questionable card actually returns a  zero length response with `90 00`.  (It is not clear why the questionable card did this.)

(I have no good way to test this, but if 

For any SELECT AID especially in a `match_card` or `init` routine, we should be very careful to check the expected results to avoid a false positive identification of a card. 

Other card drivers may have this same problem, and maybe we should consider adding an `sc_select_aid` command that does not invoke https://github.com/OpenSC/OpenSC/commit/3c1d8ad69557adf6aab08bc549188425c8b33a92 or any other code that tries to fill in missing information. The card driver knows what is a valid response from its own cards.

frankmorgner:
I think that sc-hsm just happens to be the first card driver that uses detection via AID selection. We've seen that before on some cards that just return 9000 to whatever AID was used. It's the card's problem.

Maybe the javacards used are misconfigured...

dengert:
Frank said: "Maybe the javacards used are misconfigured..." Maybe. But if OpenSC did not do 3c1d8ad69557adf6aab08bc549188425c8b33a92 on a SELECT AID I believe the `card-sc-hsm.c` would have caught the misconfiguration because it also check for a FCI being returned. 

You also said: "I think that sc-hsm just happens to be the first card driver that uses detection via AID selection"   `card-muscle.c` did it from the initial commit 50490acca0c32b37c71b8e86748648b7a2811ade  in 2006.  `muscle.c`  `msc_select_applet` builds its own SELECT AID APDU, with `apdu.le = 0`  So it looks like the Muscle applet does not expect to return anything and may be the reason for 3c1d8ad69557adf6aab08bc549188425c8b33a92. (But if the applet did not recognize the AID it should not return `90 00`.)

`card-piv.c` was doing detection via AID from 2005,  with initial commit b051b085846967795299e4bb92e19e1028b1048c in 2006, (but not correctly). Fixed in c6954a9c333087034c94d3ef460487eecb90160c 2007.  `piv_find_aid` also built its own ADPU and did not depend on iso7816.c select file code. So it does not have the issue caused by 3c1d8ad69557adf6aab08bc549188425c8b33a92

This year `card-piv.c` will look at an ATR to see if the historical bytes contain the default AID  or to deterning the card vendor or class of card (not the applet) before doing the SELECT AID. Recent changes will first try and read the Discover object which if it works means a select AID does not need to be done, because the PIV applet is already active. It does this because some cards will lose the login state when an SELECT AID is done even for the same AID.

card-sc-hsm.c was not added till 2012 with commit 1a7ca328659fdcd11a4755dcab5645f9a2b6dff8

dengert:
If https://github.com/JavaCardOS/MuscleCard-Applet is the muscle applet being used, it looks like a bug at [Cardedge.java lines 326-327](https://github.com/JavaCardOS/MuscleCard-Applet/blob/master/musclecard/src/com/musclecard/CardEdge/CardEdge.java#L326)

The code has already called `selectingApplet()`, so the select command if not for the Applet AID. 

On any other select APDU the applet  just returns. (I assume that results in `90 00`)  The applet does not support any so it should something like `ISOException.throw(SW_INVALID_PARAMETER);` at L327

If anyone agrees with this we can submit a bug to Muscle. 

@lucasmpalma  could you modify your opensc.conf to have `card_drivers = myeid,PIV-II,muscle,sc-hsm;`
(i.e. put sc_hsm after muscle.)  

myeid is last driver before sc-hsm, and does not select be AID. I am interested to see when  PIV-II  driver  send SELECT PIV AID, does it receive `90 00`, In any case it should not select the card because no FCI will be returned. Then muscle driver will send its AID and will be selected, and the sc-hsm will not be called. 
This order can be verified with a opensc-debug.log

dengert:
@CardContact 
It appears we have a situation where a card responds to a SELECT AID issued by `card-sc-hsm.c` 
But it is not a CardContact card.  The card only responds with `90 00` without a FCI. 

So we have a series of 5 events that  appear to cause a false positive selection:

- Questionable card applet returned `90 00` for a select file even though it does not use select file.

- `card-sc-hsm.c` uses the AID of the form "'E' Standard Identification of a standard by an object identifier according to ISO/IEC 8825-1"  i.e. starts with "E8" 

- The  card OS should not pass a SELECT AID command to the applet. It may not have recognized "E8" or 'E8 2B" as a AID. 

 - PR https://github.com/OpenSC/OpenSC/pull/342  added commit https://github.com/OpenSC/OpenSC/commit/3c1d8ad69557adf6aab08bc549188425c8b33a92  which builds an FCI if one is not returned. 

- `card-sc-hsm.c` should have checked the FCI to make sure it was valid.  

I would suggest that `card-sc-hsm.c` be modified to do its own select AID APDU and not rely on `(*iso_ops->select_file)` for doing AID selection. This would lower the risk of a false positive identification. 

OpenSC should check any other card drivers that do SELECT AID to make sure they do not have a similar problem.

frankmorgner:
The bug is in the applet, there's nothing to be fixed in `card-sc-hsm.c`. I've reordered the card detection to work around this card's bug.

dengert:
Short term fix, yes, Long term fix no. We need to address false positive identification of applets better. Rearranging the order of the driver list is a short term fix. 

@CardContact needs to look at their code to see if they could avoid this issue by better checking.

mouse07410:
> The bug is in the applet, there's nothing to be fixed in...

This is not about who to assign the blame to. It's about making user experience smooth, and the code as robust as possible. That includes assuming presence of bugs in other components and doing best to stay resilient.

frankmorgner:
Guys, we're talking about the MUSCLE applet here, an old and unmaintained piece of software. If vendors of security tokens still use it, users shouldn't buy their products; there are other free alternatives available. If it should get any more complicated than changing the detection order, we'll drop support of the MUSCLE applet. Users may still use older versions of OpenSC.

lucasmpalma:
Well first sorry for the late response, but @dengert  asked witch version o Muscle Applet i am using, it is the last one available at this project https://github.com/martinpaljak/AppletPlayground#applet-playground with these initialization instructions https://github.com/OpenSC/OpenSC/wiki/Muscle-applet
I tested the workaround b565c16 and it works just fine. I agree with @frankmorgner the real problem here is the applet it self, the applet project at https://github.com/martinpaljak/MuscleApplet is very clear saying "Do not use Muscle Applet", but at the same time i think this pull request should be accepted for the last users of the applet.

dengert:
@martinpaljak can you fix your copy of:
 https://github.com/martinpaljak/MuscleApplet/blob/d005f36209bdd7020bac0d783b228243126fd2f8/src/com/musclecard/CardEdge/CardEdge.java#L340 to return someting other then 90 00?

PR #1317, opened 2018-04-06
Provide more space for card holder name in CAC token label

dpward:
The PKCS#15 emulation layer for the CAC uses a single PIN. Set its label to "PIN" (rather than the card type "CAC I" or "CAC II"), so that the PIN label will be omitted from the token label, providing more space for the card holder name instead.

This is intended to match the behavior used for PIV cards, which was changed in #1133.

##### Checklist

- [x] tested PKCS#11
- [ ] tested macOS Tokend


Jakuje:
This can be a good idea to omit the CAC string the same way as we do it in PIV since recent time. On the other hand, it might bring some confusion with dual cards, providing both CAC and PIV endpoints.

@frankmorgner can we get this merged?

@jakuje approved this PR.

PR #1319, opened 2018-04-09
Some cards may return RSA signatures without leading zero bytes Fixes #1283

### LINKED ISSUE #1283: pkcs15-crypt creates trucated RSA signature file in case the resulting number happens to have a leading 0 byte
### Problem Description

I use pkcs15-crypt with an OpenPGP 2.1 card with a 4096 bit RSA key as follows:

openssl dgst -sha512 -binary $1  > $1.dgst
pkcs15-crypt --sign --input $1.dgst --output $1.sig --signature-format openssl --sha-512 --pkcs1 --key 1

where $1 points to some file. The problem is that this occasionally results in signature files which have only 511 bytes, not 512 as it should be for a 4096 bit RSA key. I guess this happens if the resulting number computed by RSA happens to have a leading 0 byte.

openssl dgst -sha512 -verify key1.asc -signature $1.sig $1

Fails to verify such signature files. If I prepend (not append!) a 0 byte to the signature, openssl can verify the signature.

### Proposed Resolution

pad signatures to the proper length by prepending 0s.

I use this workaround:

SIZE=$( stat --printf="%s" $1.sig )
PAD=$(( 512 - SIZE ))
head -c $PAD /dev/zero | cat - $1.sig > $1.sig.tmp
rm $1.sig
mv $1.sig.tmp $1.sig

### Steps to reproduce

See above, but this depends on the specific key, which is in the card and I can't get it out of the card. I guess that this should happen randomly in at least one out of 256 cases (if the modulus doesn't fill the 4096 bit range well it should happens more frequently). With the same private key I usually get 512 byte signatures, only sometimes I get 511 byte signatures. But I guess the problem should be easy to see in the code.

## Version:

I use the version comming with Debian 9, 0.16.0-3

dengert:
Add leading zero bytes in a signature so it is the size of modulus.
Fixes #1283 

 On branch short-signatures
 Changes to be committed:
	modified:   libopensc/pkcs15-sec.c

##### Checklist

	- [ ] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -452,6 +452,15 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
 	LOG_TEST_RET(ctx, r, "use_key() failed");
+
+	/* Some cards may return RSA signature as integer without leading zero bytes */
+	/* Already know outlen >= modlen and r >= 0 */
+	if (obj->type == SC_PKCS15_TYPE_PRKEY_RSA && (unsigned)r < modlen) {
+		memmove(out + modlen - r, out, r);
+		memset(out, 0, modlen - r);
+		r = modlen;
+	}
+
 	sc_mem_clear(buf, sizeof(buf));
 
 	LOG_FUNC_RETURN(ctx, r);


mouse07410:
Very preliminary test shows that this PR works with OpenPGP 2.0 (YubiKey NEO) and 2.1 (YubiKey 4).

dengert:
The fix for #1283 did not return the new length.  https://github.com/OpenSC/OpenSC/pull/1319/commits/0674dfef8f3035d9fa30a10039b24010f004e8c2 adds `r = modlen;`.  This needs to be tested with a card that fails, which I do not have.  

I can squash after https://github.com/MSoegtrop tests with failing card.

src/libopensc/pkcs15-sec.c
```diff
@@ -452,6 +452,16 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
 	LOG_TEST_RET(ctx, r, "use_key() failed");
+
+	/* Some cards may return RSA signature as integer without leading zero bytes */
+	/* Already know outlen >= modlen and r >= 0 */
+	tmpoutlen = r;
+		if (obj->type == SC_PKCS15_TYPE_PRKEY_RSA && tmpoutlen < modlen) {
```
frankmorgner: please fix the indenting and rewrite the code so that `tmpoutlen` is not needed.

frankmorgner:
What's the testing status, @MSoegtrop?

dengert:
Removed `temoutlen`  as requested and rebased into one commit. https://github.com/OpenSC/OpenSC/pull/1319/commits/dde41bbc063c056187a77c6b843ee34ec5902401
Tested with pkcs11-tool with card that does not have this problem. 

@MSoegtrop This still needs to be tested using a card that has this problem.

frankmorgner:
thanks

PR #1322, opened 2018-04-10
Allow specifying the enable_InitToken without ATR block

### LINKED ISSUE #1265: Allow card-insert events of empty cards for provisioning
### Problem Description
The ESC needs to receive card-insert events about inserted empty cards for provisioning. So far this functionality was provided by coolkey. We would be interested in something similar for OpenSC and this issue is mostly to discuss possibilities and preferred solution.

### Proposed Resolution
So far, the Coolkey PKCS#11 library was used and was using the [`pReserved` pointer in the end of the `CK_C_INITIALIZE_ARGS`](https://github.com/elmarco/coolkey/blob/master/src/coolkey/coolkey.cpp#L223) structure to enable this feature. The first possibility is to mimic this behavior and implement something similar in OpenSC PKCS#11. This argument [SHOULD be NULL in PKCS#11 2.40](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959692) and therefore is correctly checked in OpenSC, but not used to anything else yet. I agree that this might be non-standard and therefore not wanted.

The other possibility that we discussed was the implementation of bogus `blank` card driver, that would be disabled by default and would "match" also for empty cards. Turning this on would require passing environment variable `OPENSC_CONF` which would include also this driver in the end, but it would require ensuring that the second configuration file is in sync with the current.

Other possibility could be some configuration option, that would turn it on and off. That might be the cleanest and simplest way.

Or do you thing about some other way how the similar thing can be achieved, that would be acceptable by OpenSC?

Jakuje:
Resolves #1265.

As described in the referenced issue, the problem of the current solution is that we need to know an ATR of the card when creating the configuration file. This pull request is updating the code to allow specifying the option `enable_InitToken` outside of ATR blocks (at this moment in `framework pkcs15` -- not sure if there is a better place for it or better way how to handle it altogether). Comments welcomed.

Given the nature of the issue, it is certainly nothing, that should come by default, but a tools requiring it should present their own `opensc.conf` with these options set and in this case, requiring ATR is burden.

Edit: Other possibility would be to allow wildcard ATRs matches (I know there are masks, but it does not have the flexibility to match different lengths.

##### Checklist

- [X] Tested with the following card: coolkey
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/etc/opensc.conf.in
+++ b/etc/opensc.conf.in
@@ -263,10 +263,6 @@ app default {
 		#
 		# flags = "rng", "keep_alive", "0x80000000";
 
-		# Enable pkcs11 initialization.
-		# Default: no
-		# pkcs11_enable_InitToken = yes;
-
 		#
 		# Context: PKCS#15 emulation layer
 		#
@@ -951,6 +947,10 @@ app default {
 			# The location of the driver library
 			# module = @LIBDIR@@LIB_PRE@p15emu_custom@DYN_LIB_EXT@;
 		# }
+
+		# Enable initialization and card recognition in PKCS#11 layer.
+		# Default: no
+		# pkcs11_enable_InitToken = yes;
 
 		# some additional application parameters:
 		# - type (generic, protected) used to distinguish the common access application
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1806,14 +1806,13 @@ pkcs15_initialize(struct sc_pkcs11_slot *slot, void *ptr,
 {
 	struct sc_pkcs11_card *p11card = slot->p11card;
 	struct sc_cardctl_pkcs11_init_token args;
-	scconf_block *atrblock = NULL;
+	scconf_block *conf_block = NULL;
 	int rc, enable_InitToken = 0;
 	CK_RV rv;
 
 	sc_log(context, "Get 'enable-InitToken' card configuration option");
-	atrblock = sc_match_atr_block(p11card->card->ctx, NULL, &p11card->reader->atr);
-	if (atrblock)
-		enable_InitToken = scconf_get_bool(atrblock, "pkcs11_enable_InitToken", 0);
+	conf_block = sc_get_conf_block(p11card->card->ctx, "framework", "pkcs15", 1);
+	enable_InitToken = scconf_get_bool(conf_block, "pkcs11_enable_InitToken", 0);
 
 	memset(&args, 0, sizeof(args));
 	args.so_pin = pPin;
--- a/src/pkcs11/slot.c
+++ b/src/pkcs11/slot.c
@@ -302,12 +302,17 @@ again:
 		sc_log(context, "%s: Detected framework %d. Creating tokens.", reader->name, i);
 		/* Bind 'generic' application or (emulated?) card without applications */
 		if (app_generic || !p11card->card->app_count)   {
-			scconf_block *atrblock = NULL;
+			scconf_block *conf_block = NULL;
 			int enable_InitToken = 0;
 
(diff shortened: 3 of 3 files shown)


frankmorgner:
I just realized that the lack of documentation was already present before your change. Anyway, why do we need this configuration setting at all? I.e. is there a problem if we always allow token initialization?

Jakuje:
@frankmorgner Thank you for the comment.  I expanded a bit the comment in the configuration file to make it more clear what does it do.

The usage of the wildcard matches would probably be cleaner from the code and configuration point of view, but it would probably require some more changes in the ATR matching, which might in the end be more intrusive than what I did in this PR and might affect also other use cases. But I will try to elaborate on that.

We should  not implement something too complicated that will never be used in the atr match -- we already have exact match, we have masks and we would like some wildcard, but does it make sense to mix them together, or should the wildcard just as other option, such as "all"?

Example of mask:
```
card_atr 3b:8c:80:01:59:75:62:69:6b:65:79:4e:45:4f:72:33:58 {
    atrmask = "FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:00:00";
```
Example of wildcard that I can think of (not sure if the keyword "all" would not be better to prevent the expectation that the mixing with masks or partial ATRs could work):
```
card_atr * {
```
I don't think it makes sense to mix them together and allowing something like this, since it can be done to some extent using the masks already:
```
card_atr 3b:8c:80:01:59:75:62:69:6b:* {
```

> Anyway, why do we need this configuration setting at all? I.e. is there a problem if we always allow token initialization?

Well ... there was probably a reason why this was not by default out of the box. Applications using pkcs11 interface usually expect the card can do something sensible once is detected. It also yields events of card insertion, that says just "uninitialized", which is good to see that there is something going on, but does not help applications much.
On the other hand, this would make any card showing in the reader, but only if the default driver is enabled

Jakuje:
The alternative with wildcard ATR matches can look somehow like this:

https://github.com/Jakuje/OpenSC/commit/8a92ddd269e54ed998f278475d16871a484dc14a

If enabling this by default would be a solution, neither of these changes would be needed.

frankmorgner:
Wildcards would be useful already [here](https://github.com/OpenSC/OpenSC/blob/master/etc/opensc.conf.in#L472-L798). But as you can see in this example, it is not so difficult to just list all possible ATRs, because the historical bytes can contain at most 15 bytes.

I think specifying all ATRs would also be a viable option in your case, because if you're anticipating an initialization via PKCS#11, you very probably know what cards you want to initialize. In that sense, it could be possible to simply combine multiple ATR blocks with something like:
```
card_atr 3b:8c:80:01:59:4e:33:58, 3b:8c:80:01:59:75:62:69:6b:65:79:4e:45:4f:72:33:58 {
    atrmask = "FF:FF:FF:FF:FF:FF:00:00",  "FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:00:00";
```

Jakuje:
Unfortunately, specifying all the possible known ATRs is not possible, because there are various java cards from various vendors being provisioned. Though it is used as a workaround for now.

Using a list of all possible lengths with masks sounds for me like a hack and I would rather see some cleaner solution.

In the linked configuration, my proposed solution would not help, but we would really need some better pattern matching and I am not sure if I wand to go there so the most viable solution here sounds like enabling token initialization by default. If it would be acceptable, I will submit a new PR.

Jakuje:
So you mean to allow this configuration option to be accepted ONLY globally and not in the match blocks? That is also possibility. I will update PR soon.

frankmorgner:
thanks for thinking of the backward compatibility

Issue #1263, opened 2018-02-07
Experimental: PKCS#11 Support for Multiple Applications on One Card


dengert:
The Yubikey is the first card where two different applications are supported
by different card-*.c and pkcs15-*.c drivers. OpenSC in the past has assumed
that there would be only one card driver per card. pkcs11/slot.c is modified 
to allow for each slot to have a different card driver. 

Commit https://github.com/dengert/OpenSC/commit/dff1973ea238ad97848bcf4a900c40ab8952739d is built upon PRs #1243 and #1256 

@mouse07410 @Jakuje  and @frankmorgner I would be interested in your comments on the design and any testing you may want to do. Testing would be much easier when #1243 and #1256 are committed to master. 

 It is a proof of concept:

- It only selects the Yubikey and assumes the only applications that could
be on the card are PIV-II and OpenPGP. This needs to be generalized or added
to the opensc.conf. 

- There may be locking issues. Each card driver has its only card_lock, but share
the reader-pcsc.c lock. A multi-threaded application might have sessions
open to both drivers and there could be a dead lock situation. (not easy to test.) 

- Each application uses one or more slots, and the default is 4 slots per card. 

- It is not well tested. 

A Yubikey 4 with both PIV and OpenPGP and another reader with a PIV card where used in testing.

```
 pkcs11-tool --module /opt/ossl-1.1/lib/opensc-pkcs11.so -L
Available slots:
Slot 0 (0x0): Yubico Yubikey 4 CCID 00 00
  token label        : deengert
  token manufacturer : piv_II
  token model        : PKCS#15 emulated
  token flags        : login required, rng, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : dd03ee39c98b1e1d
  pin min/max        : 4/8
Slot 1 (0x1): Yubico Yubikey 4 CCID 00 00
  token label        : User PIN (OpenPGP card)
  token manufacturer : Yubico
  token model        : PKCS#15 emulated
  token flags        : login required, token initialized, PIN initialized
  hardware version   : 2.1
  firmware version   : 2.1
  serial num         : 000604035448
  pin min/max        : 6/127
Slot 2 (0x2): Yubico Yubikey 4 CCID 00 00
  token label        : User PIN (sig) (OpenPGP card)
  token manufacturer : Yubico
  token model        : PKCS#15 emulated
  token flags        : login required, token initialized, PIN initialized
  hardware version   : 2.1
[...]

PR #1333, opened 2018-04-16
coolkey: Make uninitialized cards working as expected with ESC

Jakuje:
These changes are needed for coolkey driver to recognize cards, that are just loaded the coolkey applet, but do not have any certificates and keys in. Also the card manager applet needs to be selected to get correctly the life-cycle information from the card.

Original patch from John Magne <jmagne@redhat.com>

##### Checklist

- [X] Tested with the following card: coolkey
	- [X] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/gp.c
+++ b/src/libopensc/gp.c
@@ -42,14 +42,23 @@ static int
 gp_select_aid(struct sc_card *card, const struct sc_aid *aid)
 {
 	struct sc_apdu apdu;
+	int rv;
 
 	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x0C);
 	apdu.lc = aid->len;
 	apdu.data = aid->value;
 	apdu.datalen = aid->len;
 
-	return sc_transmit_apdu(card, &apdu);
+	rv = sc_transmit_apdu(card, &apdu);
 
+	if (rv < 0)
+		return rv;
+
+	rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if (rv < 0)
+		return rv;
+
+	return apdu.resplen;
 }
 
 /* Select the Open Platform Card Manager */
@@ -59,7 +68,7 @@ gp_select_card_manager(struct sc_card *card)
 	int rv;
 
 	LOG_FUNC_CALLED(card->ctx);
-	rv =  gp_select_aid(card, &gp_card_manager);
+	rv = gp_select_aid(card, &gp_card_manager);
 	LOG_FUNC_RETURN(card->ctx, rv);
 }
 
--- a/src/libopensc/gp.c
+++ b/src/libopensc/gp.c
@@ -1,0 +1,75 @@
+/*
+ * gp.c: Global Platform Related functions
+ *
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Author: Jakub Jelen <jjelen@redhat.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "internal.h"
+
+/* The AID of the Card Manager defined by Open Platform 2.0.1 specification */
+static const struct sc_aid gp_card_manager = {
+	{0xA0, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00}, 7
+};
+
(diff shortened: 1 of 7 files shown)


Jakuje:
I am no expert in this life-cycle of smart cards so if you have a proposal how to do that better, I would be glad to verify that it still works for our use case.

@frankmorgner do you propose that we should have the Card Mgr selection as an unified function outside of drivers in this GP SM or in some of the other common files?

Jakuje:
The code in card-authentic seems to be just bogus comment, since the function is for selecting authentic applet, not the CM:

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-authentic.c#L387

I drafted something with the new `gp.c` file based on your comments, which basically unifies the select AID APDUs from the three mentioned places into one. The IASECC card was also using the "new" AID as a fallback so I added it as a separate function for now, or it can be used as automatic fallback also for the other cases.

The changes I pushed here should not introduce functional changes. If we say we should transparently try to select the "new" AID, if the "old" fails, it should be pretty simple, but I did not want to do that yet, because I scanned through these specs today for the first time.

src/libopensc/gp.c
```diff
@@ -0,0 +1,71 @@
...
+
+
+/* Select AID */
+static int
+gp_select_aid(struct sc_card *card, const unsigned char *aid)
+{
+	struct sc_apdu apdu;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x0C);
+	apdu.lc = sizeof(aid);
```
frankmorgner: this won't work. you'll have to pass a length or use `struct sc_aid`

Jakuje: True. This got lost in the transition. Thank you for the review. This is fixed now.

dengert:
Keep in mind that there are other cards that may also be Global Platform cards. The user may not have any Coolkey cards, but OpenSC is still trying to detect what applets and drivers to use with the card.  So be careful when selecting the GP AID and only do it after detecting that this is a Cool key card. 

I am not saying that this is a problem with this PR, but a card driver must make every effort to detect cards it supports without causing the state of the card to change in case it is not one of its cards and the card is in use by other processes using other drivers.

Jakuje:
@dengert If I understand your comment right, the GP applet selection is happening only when the coolkey applet is detected, but the combined object is not found (not provisioned card), during initialization. From this point on, the execution should never fall through to any other card driver (at least until the #1263 will get polished and fixed), but in any case, are the drivers assuming the initial default selection of applets?

@frankmorgner Could you have a look into the changes? Is it something that you had in your mind?

@martinpaljak I was able to find two references to GP AID throughout the OpenSC code. They might not be the only one, but to my understanding the most common ones. I don't have the cards in this configuration to try if using empty data would do the same thing. Can you point to some part of documentation that would clam or confirm that?

dengert:
What I was saying, and Frank pointed out, there are many GlobalPlatform based cards. The "GP Card Manager" traps all SELECT AID commands. GP 2.1.1 "3.2.1 GlobalPlatform Environment (OPEN)":
"Another important function provided by the OPEN is APDU command dispatching and Application selection. When a SELECT command is received, the OPEN sets the Application referenced in the SELECT command to be the selected Application and subsequent Application commands shall be dispatched to the selected Application."

It also sounds like the card is not responding to a SELECT AID for the Coolkey applet. 

The GP 2.1.1 has a big section on life cycles and issuer Security Domains. Are these any keys that need to be loaded for the issuer Security Domain?

There is also  "5.3 Application and Security Domain Life Cycle" It sounds like the card is in the "INSTALLED" state and not in the "SELECTABLE" state.  "5.3.1.2 Application Life Cycle State SELECTABLE" "The Application shall be properly installed and functional before it may be set to the state SELECTABLE."

"6.3.1.1.2 Explicit Selection on Basic Logical Channel"  "The OPEN shall determine if the requested
AID matches or partially matches an entry within the GlobalPlatform Registry **and whether this entry is
selectable.**"

So it looks like any card driver for a GP card could look at the GP Registry, to see what applications are on the card and if they are selectable using "9.4 GET STATUS Command" 

So what is the  state of your problem cards?

Is this a vendor problem, of is this a card issuer problem, in that they need to do one more operation before the applet is selectable? 

Since OpenSC tries to determine the card type and driver to use for any card inserted based on ATR or SELECT AID (or current selected AID) and it is not finding one, we may not know the user's intention of inserting the card. The card may have one or more Applets on the card that **are or are not selectable**, so how do we determine the user's actions? 

I could see a card that comes with Coolkey, OpenPGP, PIV, and Muscle applets loaded but not selectable. **The user may want one or more of these but not the others.** 

One way via PKCS\#11 (I believe that is you intent) is only if a driver is forced by env variable or opensc.conf would a driver try and make its applet selectable by first checking if the applet is loadable. 

Another way is some GP utility to make it selectable. 

(As a side note, one type of PIV card I worked with in the past were sent from vendor with a Security Domain SECURED and locked. GP 2.1.1 was needed to unlock cards in a batch. The key was sent to us separately from the batch of cards. This was to guarantee there was no tampering with the cards between vendor and card issuer.)

src/libopensc/gp.c
```diff
@@ -0,0 +1,75 @@
...
+gp_select_aid(struct sc_card *card, const struct sc_aid *aid)
+{
+	struct sc_apdu apdu;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x0C);
+	apdu.lc = aid->len;
+	apdu.data = aid->value;
+	apdu.datalen = aid->len;
+
+	return sc_transmit_apdu(card, &apdu);
```
frankmorgner: I think you also need to call `sw_check_sw()`, otherwise you will run into similar problems as in #1360...

Jakuje: Fixed in 0570ffb1
Thanks for the review!

PR #1334, opened 2018-04-16
PIV History Object Related Changes - Fixes #1330

### LINKED ISSUE #1330: Expression is always false
Hi!
Please take a look at [this expression:](https://github.com/OpenSC/OpenSC/blob/b4bc6de81bb70101eed716a67b893e5a90c3ac29/src/libopensc/card-piv.c#L2875)

`(*keyref < 0x82 && *keyref > 0x95)`
it is always false.
Probably  it should be:
`(*keyref < 0x82 || *keyref > 0x95)`

This possible defect found by [AppChecker](https://npo-echelon.ru/en/solutions/appchecker.php)

dengert:
&& is replaced by || in the test of valid key references
for retired keys found in the Historic object.

For retired keys, the user_consent flag was being set by default.
Thus a C_Login(CKU_CONTEXT_SPECIFIC) would be required.
NIST 800-73 only requires PIN_Always on the Sign Key.

To extend the usefullnes of "retired keys" on non government
issued PIV-like cards, code had already been added
to use the certificate keyUsage flags to override the NIST
defined key usage flags. The NONREPUDATION  flag is now used
to set the user_consent flag.

So rather then always requiring C_Login(CKU_CONTEXT_SPECIFIC)
for any retured key, the code only requires it for non government
cards where teh certificate has NONREPUDATION.

 Changes to be committed:
	modified:   card-piv.c
	modified:   pkcs15-piv.c

##### Checklist

- [x] Tested with the following card: NIST and Oberthur demo cards with History Objects and Retired Keys
	- [x] tested PKCS#11
	- [ ] tested macOS Tokend

### DIFF
--- a/src/libopensc/pkcs15-pin.c
+++ b/src/libopensc/pkcs15-pin.c
@@ -700,7 +700,7 @@ int sc_pkcs15_get_pin_info(struct sc_pkcs15_card *p15card,
 	/* Try to update PIN info from card */
 	memset(&data, 0, sizeof(data));
 	data.cmd = SC_PIN_CMD_GET_INFO;
-	data.pin_type = SC_AC_CHV;
+	data.pin_type = pin_info->auth_method;
 	data.pin_reference = pin_info->attrs.pin.reference;
 
 	r = sc_pin_cmd(card, &data, NULL);
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -3583,19 +3583,12 @@ static int piv_card_reader_lock_obtained(sc_card_t *card, int was_reset)
 		goto err;
 	}
 
-	/* can we detect and then select the PIV AID without losing the login state? */
-	if ((priv->card_issues & CI_DISCOVERY_USELESS)
-			&& (priv->card_issues & CI_PIV_AID_LOSE_STATE)) {
-		r = 0; /* do nothing, hope card was not interfered with */
-		goto err;
-	}
-
 	/* make sure our application is active */
 
 	/* first see if AID is active AID by reading discovery object '7E' */
 	/* If not try selecting AID */
 
-	/* but if x card does not support DISCOVERY object we can not use it */
+	/* but if card does not support DISCOVERY object we can not use it */
 	if (priv->card_issues & CI_DISCOVERY_USELESS) {
 	    r =  SC_ERROR_NO_CARD_SUPPORT;
 	} else {
@@ -3603,7 +3596,7 @@ static int piv_card_reader_lock_obtained(sc_card_t *card, int was_reset)
 	}
 
 	if (r < 0) {
-		if (!(priv->card_issues & CI_PIV_AID_LOSE_STATE)) {
+		if (was_reset > 0 || !(priv->card_issues & CI_PIV_AID_LOSE_STATE)) {
 			r = piv_select_aid(card, piv_aids[0].value, piv_aids[0].len_short, temp, &templen);
 		} else {
 			r = 0; /* cant do anything with this card, hope there was no interference */
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2871,7 +2871,7 @@ piv_process_history(sc_card_t *card)
 			}
 			keyref = sc_asn1_find_tag(card->ctx, seq, seqlen, 0x04, &keyreflen);
 			if (!keyref || keyreflen != 1 ||
-					(*keyref < 0x82 && *keyref > 0x95)) {
+					(*keyref < 0x82 || *keyref > 0x95)) {
 				sc_log(card->ctx, "DER problem");
 				r = SC_ERROR_INVALID_ASN1_OBJECT;
 				goto err;
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -3454,6 +3454,16 @@ piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 		data->pin1.tries_left = priv->tries_left;
 		if (tries_left)
 			*tries_left = priv->tries_left;
+
+		/*
+		 * If called to check on the login state for a context specific login
+		 * return not logged in. Needed because of logic in e6f7373ef066  
+		 */
+		if (data->pin_type == SC_AC_CONTEXT_SPECIFIC) {
+			data->pin1.logged_in = 0;
+			 LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+		}
+
 		if (priv->logged_in == SC_PIN_STATE_LOGGED_IN) {
 			/* Avoid status requests when the user is logged in to handle NIST
 			 * 800-73-4 Part 2:
--- a/src/libopensc/pkcs15-piv.c
+++ b/src/libopensc/pkcs15-piv.c
@@ -528,83 +528,83 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		{ "05", "Retired KEY MAN 1",
 				/*RSA*/SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_DERIVE,
-			"", 0x82, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1},
+			"", 0x82, "01", SC_PKCS15_CO_FLAG_PRIVATE, 0},
 		{ "06", "Retired KEY MAN 2",
 				/*RSA*/SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_DERIVE,
-			"", 0x83, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1},
+			"", 0x83, "01", SC_PKCS15_CO_FLAG_PRIVATE, 0},
 		{ "07", "Retired KEY MAN 3",
 				/*RSA*/SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_DERIVE,
-			"", 0x84, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1},
+			"", 0x84, "01", SC_PKCS15_CO_FLAG_PRIVATE, 0},
 		{ "08", "Retired KEY MAN 4",
 				/*RSA*/SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_DERIVE,
-			"", 0x85, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1},
+			"", 0x85, "01", SC_PKCS15_CO_FLAG_PRIVATE, 0},
 		{ "09", "Retired KEY MAN 5",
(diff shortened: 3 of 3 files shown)


dengert:
While testing a fix for #1330 it was discovered that CKA_ALWAYS_AUTHENTICATE was being set for any retired key. NIST only required "PIN Always" for the Sign Key. Retired keys are defined as Key Management keys, not Sign Keys. So this also being fixed. 

This will also help to get people who provision their own cards (usually Yubico) to stop trying to use the Sign Key as a second Auth key by allowing them to use the Retired keys for any purpose without having to enter the PIN every time.  Wiki documentation is still need to say how to do this, but all the code has been in place for sometime.

frankmorgner:
Looks good.

@mouse07410 could you test with your tokens?

dengert:
Just for  reference see #847 that introduced the code to use the keyUsage from a certificate for the retired keys. 

@apollo @gebi and @titanous  You may want to review this change too.

dengert:
I added https://github.com/OpenSC/OpenSC/pull/1334/commits/dac9634d87e38ec899713d36a389816b0435b767 to fix a problem when using a PIN PAD reader when using a CKU_CONTEXT_SPECIFIC login. for a key with CKA_ALWAYS_AUTHENTICATE (user_consent).   

The previous code would assume that if the login state was loged_in, it could skip the verify. This might for for some cards but not all. The intent of a CKU_CONTEXT_SPECIFIC login is to do a verify of the pin after the C_SignInit and  just before the C_Sign. 

The code should test for ignore_user_consent and  any decision on skipping the command should be upto the card driver.
 
Comments welcome.

This was found while trying to get OpenSSH 7.7 to handle CKA_ALWAYS_AUTHENTICATE. `pkcs11-tool --test --login` also failed to use PIN PAD reader for the CKU_CONTEXT_SPECIFIC login.

dengert:
Rebased onto 0.18.0-rc2 and ready for commit. This PR should be added before release of 0.18.0

src/libopensc/pkcs15-pin.c
```diff
@@ -306,8 +306,11 @@ sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *pi
 	 * in this case, to avoid error or unnecessary pin prompting on pinpad,
 	 * check if the PIN has been already verified and the access condition
 	 * is still open on card.
+	 *
+	 * But if context specific login we need to use pin pad reader
+	 * and not cache the pin 
 	 */
-	if (pinlen == 0) {
+	if (pinlen == 0 && auth_info->auth_method != SC_AC_CONTEXT_SPECIFIC) {
```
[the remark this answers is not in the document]
dengert: Just the opposite. It fixes a problem with using a PIN PAD reader with  CKA_ALWAYS_AUTHENTICATE. Without the change the code skips doing the PIN PAD verify. Read the comments in the code above the patch  lines 304-308, and the added comments 310-311: 
```
304         /*
305          * if pin cache is disabled, we can get here with no PIN data.
306          * in this case, to avoid error or unnecessary pin prompting on pinpad,
307          * check if the PIN has been already verified and the access condition
308          * is still open on card.
309          *
310          * But if context specific login we need to use pin pad reader
311          * and not cache the pin 
312          */
```
The previous code assumed if the login state of the card was logged_in, verify pin is not needed. That's true for normal verify but not a for a  CKA_ALWAYS_AUTHENTICATE/CONTEXT_SPECIFIC  verify, where the card is expecting a verify pin command before the next crypto command using the key with user_consent/CKA_ALWAYS_AUTHENTICATE.

The PIV requires the verify to be immediately before the crypto command.  OpenPGP only allows one crypto command per verify.  This might explain your  https://github.com/OpenSC/OpenSC/issues/1263#issuecomment-386853641 : " only one OpenPGP email - on the attempt to sign another OpenPGP email it fails." 

Without this patch 0.18.0 release candidates fail to work with a PIN PAD reader and a  CKA_ALWAYS_AUTHENTICATE key.

frankmorgner:
I didn't plan to include this PR, because I didn't notice the problem with OpenSSH earlier. However, if you try to limit your changes to `card.piv.c`, and update the [page with release tests](https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Release-Testing), we could still include this change.

I agree with @mouse07410 that it is strange that you're using `SC_AC_CONTEXT_SPECIFIC` as `CKA_ALWAYS_AUTHENTICATE`. My read of PKCS#11 is that a login for a key with `CKA_ALWAYS_AUTHENTICATE` should always be done as `Login(CKA_CONTEXT_SPECIFIC)`, but the opposite is *not true* (i.e. not every context specific login is meant for a key that always requires PIN authentication. PKCS#11 says it depends on the card what happens.)

If I understand the last commit correctly, then you want to prevent commands to be sent after PIN verification and before signature creation. This requirement is PIV specific and should be handled by `card-piv.c` (i.e. `piv_pin_cmd()`).

dengert:
I already updated https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Release-Testing  2 days ago. 5/5/2018. Look at page history and the reference to testing with NIST Demo card that has History object and retired keys.

The 3 files changes by this PR are card-piv.c, pkcs15-piv.c ( both are PIV specific) and pkcs15-pin.c. 

f3fad2ca02b022c7180522c9ac2342d067e7c784  fixes the problem in pkcs15-pin.c because the problem is pkcs15-pin.c makes a false assumption, and does not not pass on the verify to the card driver. There is no way to fix the problem in the card driver if it never sees the verify command. 

In reply to:"I agree with @mouse07410 that it is strange that you're using SC_AC_CONTEXT_SPECIFIC as CKA_ALWAYS_AUTHENTICATE." No, the application has to spot that the key has CKA_ALWAYS_AUTHENTICATE and the application has to do C_Login(CKU_CONTEXT_SPECIFIC) 
The only place in PKCS\#11 I have ever seen the use of a CONTEXT_SPECIFIC login is when a key has CKA_ALWAYS_AUTHENTICATE. In any case pkcs15-pin.c can not just fail to pass on the verify command to the driver.  a CONTEXT_SPECIFIC login is not the same as a normal CKU_USER login. 

In reply to: "If I understand the last commit correctly, then you want to prevent commands to be sent after PIN verification and before signature creation. This requirement is PIV specific and should be handled by card-piv.c (i.e. piv_pin_cmd()).he PIV driver handles the case"

It does handle this, but only if pkcs15-pin.c will pass the verify command to the driver f3fad2ca02b022c7180522c9ac2342d067e7c784 fixes that so pkcs15-pin.c does not just return if the card says it is in the loggged_in state.

frankmorgner:
Thanks for updating the wiki!

PKCS#11 is just vague with regard to `CKU_CONTEXT_SPECIFIC`. We shouldn't judge it in `pkcs15-pin.c`; just move the interpretation to `card-piv.c`! `sc_pkcs15_get_pin_info()` will call card->pin_cmd. I think the real problem is that it currently does the following:
```
data.pin_type = SC_AC_CHV;
```
rather than correctly passing `SC_AC_CONTEXT_SPECIFIC`, i.e.
```
data.pin_type = auth_info->auth_method;
```
Reading the code twice, it looks like `sc_pkcs15_change_pin()` has the same problem.

Hmm, even if you implemented and tested the requested changes, that's a lot which may have side effects for other cards. I think it's better to postpone this one change for the next release.

frankmorgner:
By the way, `sc_pkcs15_pincache_add()` should do nothing [if some key associated with that PIN always requires user consent](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-pin.c#L765).

mouse07410:
Nothing? How about "ignore_user_consent" condition?

frankmorgner:
Well, I ignored obvious corner cases. If the user specifies `ignore_user_consent`, then he wants caching of a `SC_AC_CONTEXT_SPECIFIC`-PIN.

dengert:
I disagree with holding up https://github.com/OpenSC/OpenSC/commit/f3fad2ca02b022c7180522c9ac2342d067e7c784 till the next release.  

The problem is in e6f7373e 2017-04-07 introduced in  0.17.0 that made the wrong assumption, that all verifies are the same.   The problem only shows up when a PIN PAD is used and would only have a problem with a context specific login.  

c24a91e4  2018-02-01 introduced the SC_AC_CONTEXT_SPECIFIC which meant that we could address the problem in this PR. We have done a lot of work with handing of CONTEXT_SPECIFIC logins in both pkcs11-tool.c and in drivers since 0.17.0. It would be a shame to not fix this problem with PIN PAD readers in 0.18.0. 

When would the next release be when you would fix this problem? Next year?

mouse07410:
What I described is **not** a corner case - it's a necessity for those who use tokens with Java apps such as 'keytool'.

frankmorgner:
@dengert could you change https://github.com/OpenSC/OpenSC/commit/f3fad2ca02b022c7180522c9ac2342d067e7c784 according to https://github.com/OpenSC/OpenSC/pull/1334#issuecomment-387130643, please?

dengert:
In response to https://github.com/OpenSC/OpenSC/pull/1334#issuecomment-387130643 Do you think that `sc_pkcs15_get_pin_info` 
line 703: 
`data.pin_type = SC_AC_CHV;`
should be 
`data.pin_type = auth_info->auth_method;` ?
Are you willing to make that change?

I have looked at `SC_AC_CONTEXT_SPECIFIC` to be a special case of the `SC_AC_CHV`. i.e. it is independent of the login state of the card as set by the `SC_AC_CHV` PIN as  checked by  
`sc_pkcs15_get_pin_info`.  So checking the login state does not make any sense for a `SC_AC_CONTEXT_SPECIFIC` verify. 
`iso7816.c` `iso7816_build_pin_apdu` treats `SC_AC_CHV`, `SC_AC_SESSION` and `SC_AC_CONTEXT_SPECIFIC` the same. 

I should point out that  `pkcs15-pin.c`  does a ` && auth_info->auth_method != SC_AC_SESSION) {` to check for similar situation where `auth_info->auth_method` has a special meaning much like https://github.com/OpenSC/OpenSC/commit/f3fad2ca02b022c7180522c9ac2342d067e7c784

I contend that the actual problem is in `sc_pkcs15_verify_pin` which works fine if there is no pin pad reader, but if there is a pin pad reader and `SC_AC_CONTEXT_SPECIFIC` is set, `sc_pkcs15_get_pin_info` gets called when it should not be called. 

Without https://github.com/OpenSC/OpenSC/commit/f3fad2ca02b022c7180522c9ac2342d067e7c784 or changing line 703, there is no way a for a card driver's `*_pin_cmd` to tell that it is being asked to do a `SC_PIN_CMD_GET_INFO` command when the verify is for a `SC_AC_CONTEXT_SPECIFIC` verify.

So what do you want to do? 
Use https://github.com/OpenSC/OpenSC/commit/f3fad2ca02b022c7180522c9ac2342d067e7c784  ( which I think is the straigh forward approach) or modify line 703 plus driver changes, or are we passing `SC_AC_CONTEXT_SPECIFIC` the wrong way?

dengert:
I have rebased and pushed a new version with the `data.pin_type = pin_info->auth_method;` change and changes `card-piv.c` Is this what you are looking for?
It works with PIV card, and PIN PAD reader, but not sure if the  `data.pin_type = pin_info->auth_method;` will cause problems with other drivers. 

If this is what you want I will squash the last two commits.

src/libopensc/pkcs15-pin.c
```diff
@@ -306,6 +306,8 @@ sc_pkcs15_verify_pin(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *pi
 	 * in this case, to avoid error or unnecessary pin prompting on pinpad,
 	 * check if the PIN has been already verified and the access condition
 	 * is still open on card.
+	 *
+	 * But if context specific login we need to use pin pad reader
```
frankmorgner: this comment can be removed

dengert: I left it in there on purpose. I can remove it.

dengert: Comment has been removed.

frankmorgner:
thank you

PR #1335, opened 2018-04-17
Do not temporarily set SC_READER_REMOVED on all readers

### LINKED ISSUE #1324: SIGSEGV during simultaneous C_WaitForSlotEvent and C_GetSlotList
### Context

I am running VMware Horizon Client 4.7.0 under Red Hat Enterprise Linux 7.4. The client uses a selected PKCS#11 library (in this case, OpenSC master) both to perform client certificate authentication to a server, and later to perform smart card redirection to a remote virtual machine (which has a driver installed for a "virtual smart card reader").

When launching the client and connecting to a server, I am prompted to insert my smart card for authentication. Inserting my PIV card will cause a SIGSEGV in the OpenSC library, and the client will terminate. (For some reason, if I launch the client a second time, it usually works.)

### Problem Description

The client calls `C_WaitForSlotEvent` once per second to determine if a smart card has been inserted or removed. When a card is inserted, then (whenever this issue happens) one thread calls `C_WaitForSlotEvent` again, while a second thread simultaneously calls `C_GetSlotList`. The second thread causes the flag `SC_READER_REMOVED` to be [set on all readers "temporarily"](https://github.com/OpenSC/OpenSC/blob/54097c0fc024005d37d5a90896bcc28cc603d3dd/src/libopensc/reader-pcsc.c#L1295-L1299); meanwhile the other thread sees this flag and actually removes the reader. The reader and card will be re-added, but the second thread will eventually find itself dereferencing an invalid pointer to the OpenSC context (or something similar). Exactly at what point that occurs changes each time this is run; one example is shown below.

### Proposed Resolution

There may be another underlying issue here, but this at least seems to be worked around by not temporarily setting the `SC_READER_REMOVED` flag on all readers — and instead setting the flag after an existing reader is actually not returned in the list from `SCardListReaders`.

[This patch](https://gist.github.com/dpward/8e4dd38ccfd47991413d0dcb655e0fbd) to the reader driver for PC/SC seems to work for me (I tried it with 4 readers connected). It looks like a similar change would be applicable to the [reader driver for CryptoTokenKit](https://github.com/OpenSC/OpenSC/blob/54097c0fc024005d37d5a90896bcc28cc603d3dd/src/libopensc/reader-cryptotokenkit.m), but I don't have a Mac and have never encountered Objective-C before now...

### Logs

See [opensc-debug.log](https://github.com/OpenSC/OpenSC/files/1896998/opensc-debug.log) before SIGSEGV is raised. (Notice that the card's actual ATR suddenly becomes empty near the end of the log.) The backtrace is:

```
Program terminated with signal 11, Segmentation fault.
#0  0x00007fc8ad850b73 in _sc_match_atr_block (ctx=0x200000000046, driver=driver@entry=0x0, atr=atr@entry=0x1a0bbd0) at card.c:1129
1129			for (i = 0; ctx->card_drivers[i] != NULL; i++) {
#1  0x00007fc8ad876d90 in check_forced_protocol (reader=reader@entry=0x1a0bb70, protocol=protocol@entry=0x7fc89bffe8b0) at reader-pcsc.c:424
#2  0x00007fc8ad8785e7 in pcsc_connect (reader=0x1a0bb70) at reader-pcsc.c:568
#3  0x00007fc8ad851303 in sc_connect_card (reader=reader@entry=0x1a0bb70, card_out=card_out@entry=0x7fc8940010e8) at card.c:203
#4  0x00007fc8adc01923 in card_detect (reader=0x1a0bb70) at slot.c:279
#5  0x00007fc8adc022e2 in card_detect_all () at slot.c:403
#6  0x00007fc8adbfc6e0 in C_GetSlotList (tokenPresent=1 '\001', pSlotList=0x0, pulCount=0x7fc89bffea48) at pkcs11-global.c:407
```

dpward:
Fixes #1324. Tested on Linux and on macOS 10.13.4, with multiple readers attached, inserting and removing different PIV cards in different readers at random.

##### Checklist
- [x] macOS CryptoTokenKit token driver is tested

### DIFF
--- a/src/libopensc/reader-cryptotokenkit.m
+++ b/src/libopensc/reader-cryptotokenkit.m
@@ -543,8 +543,10 @@ err:
 static int cryptotokenkit_detect_readers(sc_context_t *ctx)
 {
 	size_t i;
+	NSUInteger j;
 	int r;
 	TKSmartCardSlotManager *mngr = [TKSmartCardSlotManager defaultManager];
+	NSMutableArray *slotNames;
 
 	LOG_FUNC_CALLED(ctx);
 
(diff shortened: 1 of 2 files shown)


dpward:
I would like to include a [corresponding change](https://gist.github.com/dpward/371c27079513adfc905d8035449a81b4#file-opensc-cryptotokenkit-detect-readers-patch) to the CryptoTokenKit reader driver (this change is for the PC/SC reader driver).

That change builds correctly under Xcode 9.3, but unfortunately I haven't yet been able to get my hands on a macOS system with a smart card reader for quick sanity testing. I should be able to soon, but if anyone who uses CryptoTokenKit can test it easily, that would be appreciated!

PR #1343, opened 2018-04-25
C set attribute value

konstantinpersidskiy:
##### Checklist

- [ ] macOS tokend is tested

Hello, 
PKCS11 standard  v. 2.40 (http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.pdf) states that default value for CKO_DATA object's CKA_VALUE is empty (4.5.2, page 42). That points to the possibility of modifying this attribute for these objects via C_SetAttributeValue (same can be seen in the current code for object's id). 
These commits allow creating data objects with empty CKA_VALUE and modifying it later.

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -3303,8 +3303,42 @@ sc_pkcs15init_change_attrib(struct sc_pkcs15_card *p15card, struct sc_profile *p
 			LOG_TEST_RET(ctx, SC_ERROR_NOT_SUPPORTED, "Cannot change ID attribute");
 		}
 		break;
+	case P15_ATTR_TYPE_VALUE:
+		switch(df_type) {
+		case SC_PKCS15_DODF: {
+			u8 *nv;
+			struct sc_pkcs15_data_info *info = (struct sc_pkcs15_data_info *) object->data;
+			struct sc_path old_data_path = info->path;
+			struct sc_path new_data_path;
+			struct sc_pkcs15_der new_data;
+			new_data.len = new_len;
+			new_data.value = (u8 *) new_value;
+			
+			/* save new data as a new data file on token */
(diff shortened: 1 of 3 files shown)


frankmorgner:
The change looks OK in general, but I don't know how to test this. What tests did you run?

konstantinpersidskiy:
I wrote a small test using opensc-pkcs11 that creates a data object with empty CKA_VALUE, then finds this object and modifies it's CKA_VALUE to non-empty. I ran it on RutokenS and RutokenECP tokens and checked tokens' file structure for data object EF and changes I expected in DODF.
I can upload my test if you like

konstantinpersidskiy:
https://gist.github.com/konstantinpersidskiy/bc6824101a85fb2e8ce11743183d04a9  
That's my test, to compile change path to opensc-pkcs11.so in dlopen.  
@frankmorgner , is that enough?

src/pkcs15init/pkcs15-lib.c
```diff
@@ -3303,8 +3303,42 @@ sc_pkcs15init_change_attrib(struct sc_pkcs15_card *p15card, struct sc_profile *p
...
+			struct sc_pkcs15_der new_data;
+			new_data.len = new_len;
+			new_data.value = (u8 *) new_value;
+			
+			/* save new data as a new data file on token */
+			r = sc_pkcs15init_store_data(p15card, profile, object, &new_data, &new_data_path);
+			profile->dirty = 1;
+			LOG_TEST_RET(ctx, r, "Failed to store new data");
+
+			nv = (u8 *) malloc (new_len * sizeof(u8));
```
frankmorgner: missing error check. you could also use `realloc()`

frankmorgner:
@konstantinpersidskiy for testing, the example is more than I've expected. Actually, I was referring to the checklist above and wanted to know what card you tested with what functionality.

Anyway, please see the above comment about the missing error check.

Issue #1174, opened 2017-10-18
macOS: engine_pkcs11 fails to prompt & receive PIN


mouse07410:
### Problem Description

MacOS 10.12.6 Sierra, Xcode-9.0.1. 

```
$ echo bc09ceb7b67e28ffb7c7ef146091d1c5eeb1d66d6fc5810e4173478ec044a371 | xxd -r -p -c 200 | openssl pkeyutl -engine pkcs11 -keyform engine -encrypt -pubin -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20pubkey;type=public" -pkeyopt rsa_padding_mode:oaep -out /tmp/derive.58072.key.enc
engine "pkcs11" set.
$ openssl pkeyutl -engine pkcs11 -keyform engine -decrypt -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20key;type=private" -pkeyopt rsa_padding_mode:oaep -in /tmp/derive.58072.key.enc | xxd -p -c 200
engine "pkcs11" set.
Enter PKCS#11 token PIN for Uri the Great:
UI_process failed
No PIN code was entered
Login to token failed, returning NULL...
PKCS11_get_private_key returned NULL
cannot load Private Key from engine
140736170533896:error:80067065:pkcs11 engine:ctx_load_privkey:object not found:eng_back.c:886:
140736170533896:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:eng_pkey.c:124:
unable to load Private Key
Error initializing context
```

UI process is failing.

### Proposed Resolution

Downgrade to before the UI PR merge. Probably limit your UI improvements to Windows.

### Steps to reproduce

See above.

This is what the code outputs when I revert your 391b32c and 6a4e120.
```
$ echo bc09ceb7b67e28ffb7c7ef146091d1c5eeb1d66d6fc5810e4173478ec044a371 | xxd -r -p -c 200 | openssl pkeyutl -engine pkcs11 -keyform engine -encrypt -pubin -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20pubkey;type=public" -pkeyopt rsa_padding_mode:oaep -out /tmp/derive.58072.key.enc
engine "pkcs11" set.
[...]

PR #1346, opened 2018-04-27
macOS: disable notifications only in PKCS#11 module

frankmorgner:
I was unsatisfied having no notifications at all, so I basically reverted https://github.com/OpenSC/OpenSC/commit/c35eb1c9bc74e284723ffd726478720b69aed970 by applying a more selective fix for https://github.com/OpenSC/OpenSC/issues/1174

@mouse07410 could you please check if your PIN prompt in engine_pkcs11 is still working with this change?

##### Checklist

- [x] macOS tokend is tested

### DIFF
--- a/MacOSX/build-package.in
+++ b/MacOSX/build-package.in
@@ -61,7 +61,6 @@ if ! test -e ${BUILDPATH}/target/$PREFIX/lib/pkgconfig; then
 		--sysconfdir=$PREFIX/etc \
 		--enable-cvcdir=$PREFIX/etc/cvc \
 		--enable-x509dir=$PREFIX/etc/x509 \
-		--disable-notify \
 		--disable-dependency-tracking \
 		--enable-shared \
 		--enable-static \
--- a/src/ui/notify.c
+++ b/src/ui/notify.c
@@ -278,6 +278,17 @@ static void notify_proxy(struct sc_context *ctx,
 	 * we're including NotificationProxy which has similar features */
 	const char notificationproxy[] = "/Library/Security/tokend/OpenSC.tokend/Contents/Resources/Applications/NotificationProxy.app/Contents/MacOS/NotificationProxy";
 
+	if (ctx && ctx->app_name
+			&& (0 == strcmp(ctx->app_name, "opensc-pkcs11")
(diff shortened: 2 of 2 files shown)


mouse07410:
@frankmorgner it does not seem to interfere with pkcs11 engine or with my stuff. 

I re-enabled notifications in my build-package.in, and this PR did not cause any problem that I could see, even if I start `/Applications/OpenSC Notify.app` (except that it invoked Keychain Access without me asking for that, but I can live with it).

I guess this PR can be merged.

frankmorgner:
It opens Keychain when clicked on the notification, because its most likely the app you want to look at when dealing with a token (in the pre-CTK era).

Thanks for testing!

mouse07410:
Is it possible to configure what app to invoke on a notification click?

mouse07410:
Also, it looks like notification works on 10.12, but doesn't work on 10.13.

mouse07410:
@frankmorgner After a couple of days I noticed that `opensc-notify` seems to have broken OpenSSH that used PIV key. Here's the symptom:
```
$ git pull
Enter PIN for 'Uri the Great': 
sign_and_send_pubkey: signing failed: error in libcrypto
Already up to date.
$
```

I've been using this on three machines:

a.  macOS Sierra 10.12.6 - no apparent/visible problems (so far);
b. macOS High Sierra 10.13.4 - does not display notifications, and does not seem to interfere with anything;
c. macOS High Sierra 10.13.4 - does display notifications, but causes the above problem for OpenSSH that uses smart card (normal libp11+OpenSSL tests appear to pass fine).

frankmorgner:
[NotificationProxy](https://github.com/frankmorgner/NotificationProxy) doesn't have any configuration options, but you can change [the source code](https://github.com/frankmorgner/NotificationProxy/blob/master/NotificationProxy/main.m#L43) if you like.

It's unlikely that opensc-notify causes any trouble with ssh, because it basically only listens on [SCardGetStatusChange](https://github.com/OpenSC/OpenSC/blob/master/src/tools/opensc-notify.c#L112) without actually using the card.

I think you somehow mexed up the macOS version number, but I'll try to check again on my machine soon...

PR #1347, opened 2018-04-28
Starcos 3.2

jpecholt:
Added support for Starcos 3.2. cards.
Encryption, signature and decryption functionalities were implemented.
GetSerialnumber and PIN-info are supported. 
Superfluous "User Pin" removed from card label. 

Starcos 3.2. cards were used during testing.

- [x] Documentation is added and updated
- [x] PKCS#11 module is tested. Note: Starcos 3.2. only generates random numbers of 8-bytes length, which causes the generateRandom test case (./pkcs11-tool --test) to fail, since it excepts a different length.


frankmorgner:
Well, the current (and past) approach seems very hacky. Though we could continue with it until we hit a starcos version that's used in multiple configurations, it is also possible to try and be more generic, at least with Starcos 3.X. For example, which operation is allowed with what key, is specified in EF.KEYD. If parsed, it indicates which exact operation and mechanism to use for signature and encryption...

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,137 @@ typedef struct starcos_ex_data_st {
...
+/*Finds appropriate alg based on sec environment and writes these into the ptr
+* returns number of bytes which were added, 0 if no alg specified. -1 if not supported
+* The smartcard will use some standard algs (see "standard" comment) if no alg is defined */
+int starcos_find_algorithm_flags_3_2(sc_card_t *card, const sc_security_env_t *env, u8 *p){
+	sc_context_t *ctx = card->ctx;
+	if (env->flags & SC_SEC_ENV_ALG_REF_PRESENT) {
+		sc_debug(ctx, SC_LOG_DEBUG_NORMAL, "Found alg ref id%02x\n",env->algorithm_ref & 0xFF );
+		*p++ = 0x80;
+		*p++ = 0x01;
+		*p++ = env->algorithm_ref & 0xFF;
```
frankmorgner: missing length check

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,137 @@ typedef struct starcos_ex_data_st {
...
+				if(env->algorithm == SC_ALGORITHM_RSA){
+					*p++= 0x89;
+					*p++= 0x02;
+					*p++= 0x11;//encipher
+					*p++= 0x30;//rsa (standard)
+					return 4;
+				}
+				if(env->algorithm == SC_ALGORITHM_DES){
+					return -1; //for now, not supported
+					/**p++= 0x89;
```
frankmorgner: please remove dead code

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,137 @@ typedef struct starcos_ex_data_st {
...
+				}
+				return 0;
+			case SC_SEC_OPERATION_SIGN:
+				if (env->algorithm_flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
+					*p++= 0x89;
+					*p++= 0x02;
+					*p++= 0x13;//signature
+					*p++= 0x23;//PKCS RSA (standard)
+					if (env->algorithm_flags & SC_ALGORITHM_RSA_HASH_SHA1){
+						//TODO: not tested yet
```
frankmorgner: please remove untested code

src/libopensc/card-starcos.c
```diff
@@ -268,7 +406,7 @@ static int process_fci(sc_context_t *ctx, sc_file_t *file,
 	return SC_SUCCESS;
 }
 
-static int process_fci_v3_4(sc_context_t *ctx, sc_file_t *file,
+static int process_fci_v3_2_v3_4_v3_5(sc_context_t *ctx, sc_file_t *file,
```
frankmorgner: maybe renaming this to `process_fci_v3` would be useful

src/libopensc/card-starcos.c
```diff
@@ -304,7 +442,7 @@ static int process_fci_v3_4(sc_context_t *ctx, sc_file_t *file,
 	return SC_SUCCESS;
 }
 
-static int process_fcp_v3_4(sc_context_t *ctx, sc_file_t *file,
+static int process_fcp_v3_2_v3_4_v3_5(sc_context_t *ctx, sc_file_t *file,
```
frankmorgner: see above comment on the naming

src/libopensc/card-starcos.c
```diff
@@ -1653,17 +1908,37 @@ static int starcos_decipher(struct sc_card *card,
...
+			//TODO: should be set in init()? -> capabilities
 			card->caps |= SC_CARD_CAP_APDU_EXT;
 		}
-
-		r = sc_transmit_apdu(card, &apdu);
-		sc_mem_clear(sbuf, crgram_len + 1);
+		//starcos_v_3_2 specific changes
+		if(card->type == SC_CARD_TYPE_STARCOS_V3_2){
+			u8 *tempbuf = NULL;
+			tempbuf = malloc(SC_MAX_APDU_BUFFER_SIZE);
```
frankmorgner: missing error handling; why aren't you using `out`/`outlen` directly?

src/libopensc/card-starcos.c
```diff
@@ -1728,6 +2003,36 @@ static int starcos_get_serialnr(sc_card_t *card, sc_serial_number_t *serial)
...
+				return SC_ERROR_INTERNAL;
+
+			//Serial nr contained in data obj with following tag
+			tag = sc_asn1_find_tag(card->ctx, apdu.resp, apdu.resplen, 0x9F6A, &taglen);
+			if (tag  == NULL) {
+				return SC_ERROR_INTERNAL;
+			}
+			//only consider last 8 bytes (apparently common practice)
+			offs = taglen > 8 ? taglen - 8 : 0;
+			memcpy(&card->serialnr, tag + offs, MIN(8, taglen));
```
frankmorgner: out of bounds read. should be taglen-offs

src/libopensc/card-starcos.c
```diff
@@ -1826,14 +2131,17 @@ static int starcos_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
...
+			break;
 		case SC_CARD_TYPE_STARCOS_V3_4:
 		case SC_CARD_TYPE_STARCOS_V3_5:
 			data->flags |= SC_PIN_CMD_NEED_PADDING;
 			data->pin1.encoding = SC_PIN_ENCODING_GLP;
-			/* fall through */
-		default:
-			r = iso_ops->pin_cmd(card, data, tries_left);
 	}
+	r = iso_ops->pin_cmd(card, data, tries_left);
```
frankmorgner: please add error handling for cards of version below 3

src/libopensc/card-starcos.c
```diff
@@ -1857,7 +2165,8 @@ static struct sc_card_driver * sc_get_driver(void)
 	starcos_ops.card_ctl    = starcos_card_ctl;
 	starcos_ops.logout      = starcos_logout;
 	starcos_ops.pin_cmd     = starcos_pin_cmd;
-  
+	starcos_ops.list_files	= NULL;
```
frankmorgner: this is not needed

src/pkcs11/framework-pkcs15.c
```diff
@@ -1071,7 +1071,7 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 			pin_info = NULL;
 		}
 		else   {
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
+			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0 && strncmp(auth->label, "User Pin", 9) != 0)
```
frankmorgner: what do you need this change for?

frankmorgner: The term *User PIN* is normally used to differentiate from *SO PIN* or *Signature PIN*.

src/libopensc/card-starcos.c
jpecholt: Starcos 3.1 and 3.6 are not included, that's the reason for this naming scheme. Maybe process_fci_v3_2-5 ? Or solely put this information in a comment above the function and name it process_fci_v3?

src/pkcs11/framework-pkcs15.c
jpecholt: Without this string compare, the label is copied into the tokeninfo label, which causes the inserted card to be displayed as "User Pin()" instead of "". At the same time, the closing bracket is often cut off. With the additional clause, the string is not copied and hence not shown.

src/libopensc/card-starcos.c
jpecholt: Here, tag points to the beginning of a tag which is potentially longer than 8 bytes. If this is the case, the offset is specified so that only the last 8 bytes are read out and used for the serial number. Hence, offs = taglength - 8 is added so that tag + offs points to the start of the last 8 bytes.

src/libopensc/card-starcos.c
jpecholt: For this command, Starcos 3.2 cards will throw an error if they don't receive a buffer with the maximal possible size. Thus, we can only use the buffer in case it has the maximal size. Therefore, a temporary max-sized buffer is provided and later cropped and copied into the output buffer.

src/libopensc/card-starcos.c
frankmorgner: there is no 3.1 or 3.6 at the moment, hence there is only 3.x and everything else. just change the naming.

src/libopensc/card-starcos.c
frankmorgner: nevermind, that should work as intended.

However, `card->serialnr.len = 8;` needs to be changed to `card->serialnr.len = MIN(8, taglen);`

src/libopensc/card-starcos.c
frankmorgner: nevermind, you're right. thanks for the clearification

src/pkcs11/framework-pkcs15.c
frankmorgner: It's intended to be shown as *PIN LABEL (CARD LABEL)*. It seems your card doesn't have set `p15card->tokeninfo->label`. A better approach would be to only append *(CARD LABEL)* if `p15card->tokeninfo->label` is not empty.

src/libopensc/card-starcos.c
frankmorgner: i see. please add the error handling then

src/pkcs11/framework-pkcs15.c
jpecholt: Both auth->label and p15card->tokeninfo->label are not empty, Github just removed the strings surrounded by angle brackets from my example. What the output shows is "User Pin(Example card" -where the closing bracket is oftentimes missing- instead of the desired "Example card".

frankmorgner: As explained above, I don't think this should be done for all cards. Either do this specifically for your card in something like `pkcs15-starcos.c` or create a configuration option that allows setting the pkcs#11 slot labels.

jpecholt: I am not as well-versed in the code structure and order of execution of certain functions: Is either one of the `pkcs15-starcos.c::starcos_init_card( )` or `pkcs15-starcos.c::starcos_finalize_card( )` functions definitely going to be called After the above function `pkcs15_init_slot( )` ? In that case, one could change the card label in `pkcs15-starcos.c`, otherwise it might be overwritten again.

frankmorgner: please remove this change. We'll think about configuring the slot naming seperately.

frankmorgner:
If you look at the inline comments, we can merge this for the next release...

jpecholt:
I undid changes concerning the renaming of the card lable in `framework-pkcs15.c:1089` and integrated all other requests. I hope this satisfies all requirements.

frankmorgner:
There's some error handling missing with `malloc`

jpecholt:
As far as I could see, after each malloc, a pointer check is performed (Though this is not shown here, it is visible in the file)?

src/libopensc/card-starcos.c
```diff
@@ -1388,28 +1499,50 @@ static int starcos_set_security_env(sc_card_t *card,
...
-		if (env->algorithm_flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
-			*p++ = 0x80;
-			*p++ = 0x01;
-			*p++ = 0x02;
-		} else
-			return SC_ERROR_INVALID_ARGUMENTS;
-		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x22, 0x81,
-		               0xb8);
+		// adapt last byte : FIXME DIRTY HACK (somehow works?)
+		p += env->key_ref_len -1;
```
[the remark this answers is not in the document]
jpecholt: There is a discrepancy between provided key reference number by the card and reference number expected by it. This was tested on multiple cards and compared to other drivers which also receive the same numbers by the card. Even though e.g. a 0x84 is given as a key reference number when the card is read out, it produces an error when later specifying this exact key reference number for an operation. For an unknown reason, it always expects a number which is smaller by three, 0x81 in the example. The reason for this could not be found after thorough in depth search, so we left it at that. I see that there needs to be an additional check concerning a zero length key reference number, though.

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,87 @@ typedef struct starcos_ex_data_st {
...
 	} while (0);
 
+	#define CHECK_NOT_SUPPORTED_V3_2(card) \
+		do { \
+			if( card->type == SC_CARD_TYPE_STARCOS_V3_2) { \
+				sc_debug((card)->ctx, SC_LOG_DEBUG_NORMAL, \
+					"not supported for STARCOS 3.2 cards"); \
+				return SC_ERROR_NOT_SUPPORTED; \
+			} \
+		} while (0);
```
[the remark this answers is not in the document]
jpecholt: There is one additional operation which was not tested on Starcos 3.4. Apart from this, one could merge both checks and implement a manual error check at that point.

src/libopensc/card-starcos.c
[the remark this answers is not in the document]
frankmorgner: What is the exact key reference and the exact value you need to code in the MSE? What's the content of EF_KEYD?

I think understanding EF_KEYD is key in avoiding hacks like this. Please read the documentation and try to find some hints.

src/libopensc/card-starcos.c
```diff
@@ -1653,17 +1824,38 @@ static int starcos_decipher(struct sc_card *card,
...
 		apdu.datalen = crgram_len + 1;
 
+		//extended length APDU
 		if (sc_get_max_send_size(card) < crgram_len + 1) {
 			/* Starcos doesn't support chaining for PSO:DEC, so we just _hope_
 			 * that both, the reader and the card are able to send enough data.
 			 * (data is prefixed with 1 byte padding content indicator) */
 			card->max_send_size = crgram_len + 1;
 			card->reader->max_send_size = crgram_len + 1;
+			//TODO: should be set in init()? -> capabilities
```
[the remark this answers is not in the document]
jpecholt: It is in line 1863.

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,84 @@ typedef struct starcos_ex_data_st {
...
+		do { \
+			if( card->type == SC_CARD_TYPE_STARCOS_V3_2) { \
+				sc_debug((card)->ctx, SC_LOG_DEBUG_NORMAL, \
+					"not supported for STARCOS 3.2 cards"); \
+				return SC_ERROR_NOT_SUPPORTED; \
+			} \
+		} while (0);
+
+/*Finds appropriate alg based on sec environment and writes these into the ptr
+* returns number of bytes which were added, 0 if no alg specified. -1 if not supported
```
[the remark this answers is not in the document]
jpecholt: Error code, thank you.

src/libopensc/card-starcos.c
```diff
@@ -89,6 +90,84 @@ typedef struct starcos_ex_data_st {
...
+				if (env->algorithm_flags & SC_ALGORITHM_RSA_PAD_ISO9796) {
+					return SC_ERROR_NOT_SUPPORTED;
+				}
+				break;
+			case SC_SEC_OPERATION_AUTHENTICATE:
+				//symmetric
+				if(env->flags & SC_SEC_ENV_KEY_REF_SYMMETRIC){
+					/* For now, signature sometimes runs into this case -> return 0, not -1 to continue signature procedures
+					* Not supported yet. */
+					return 0;
```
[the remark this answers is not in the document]
jpecholt: I'll update the comment, thanks.

src/libopensc/card-starcos.c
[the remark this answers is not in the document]
jpecholt: I am fairly sure I tried the following, but I will look this up to confirm it:   
I was not able to access any EF_KEYD file as an additional source of information. Neither was this file accessed by any other driver, implementation or program which could be used as a reference. The documentation did not provide any further helpful information. Additionally, OpenSC chooses a key with the appropriate usage flags (see e.g. `pkcs15-framework.c:pkcs15_prkey_sign`, line 3773), whose reference number is written into the key-ref-pointer. This number is the exact same as the one which was initially read out from the Starcos Smart-Card. It is later not accepted as a valid ref-number anymore, the reported error code deeming the key not suitable or the number invalid. No information in the documentation suggests a change in or an adaption of the key reference number.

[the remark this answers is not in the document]
frankmorgner: Did you read the starcos reference manual? Did you search for *EF_KEYD*?

Sorry for being picky, but this hack is a little too problematic. Unfortunately, I currently don't have more time to implement this myself...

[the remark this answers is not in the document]
jpecholt: I completely understand. I spent an extensive amount of time myself trying to find the reason for this. This also lead to the fact that I compared the complete communication process between driver and card for different implementations, none was making use of this EF_KEYD file or gaining any further relevant information. I am certain that the information found in the manual was not sufficient, but I am going to check that again tomorrow and come back to you.

frankmorgner:

PR #1353, opened 2018-05-11
Fixed gcc 8 compilation errors

fbezdeka:
The following errors occured during a compilation using gcc 8 on Fedora 28:

In function »gids_create_file.constprop«,
    inserted by »gids_save_certificate.isra.8« beicard-gids.c:1548:7:
card-gids.c:465:2: Error: »strncpy« output may be truncated copying 8 bytes from a string of length 8 [-Werror=stringop-truncation]
  strncpy(record->filename, filename, 8);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

pkcs15-oberthur.c: In function »sc_pkcs15emu_oberthur_add_prvkey«:
pkcs15-oberthur.c:741:5: Error: »strncpy« output may be truncated copying 254 bytes from a string of length 254 [-Werror=stringop-truncation]
     strncpy(kobj.label, objs[ii]->label, sizeof(kobj.label) - 1);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

##### Checklist

- [ ] macOS tokend is tested

I was able to authenticate using a "Atos CardOS" card on a Fedora/Linux based system afterwards.  
I'm not able to test the Windows or macOS part due to missing infrastructure.

### DIFF
--- a/src/libopensc/card-gids.c
+++ b/src/libopensc/card-gids.c
@@ -33,6 +33,7 @@ Some features are undocumented like the format used to store certificates. They 
 
 #include <stdlib.h>
 #include <string.h>
+#include "../common/compat_strlcpy.h"
 
 #ifdef ENABLE_OPENSSL
 /* openssl only needed for card administration */
@@ -462,7 +463,7 @@ static int gids_create_file(sc_card_t *card, char* directory, char* filename) {
 	memset(masterfilebuffer + offset, 0, sizeof(gids_mf_record_t));
 	record = (gids_mf_record_t*) (masterfilebuffer + offset);
 	strncpy(record->directory, directory, 8);
-	strcpy(record->filename, filename);
+	strlcpy(record->filename, filename, sizeof(record->filename));
 	record->fileIdentifier = fileIdentifier;
 	record->dataObjectIdentifier = dataObjectIdentifier;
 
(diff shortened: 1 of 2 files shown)


fbezdeka:
I forgot to mention that I just wanted to point out where the problems for gcc 8 are.

The provided "fixes" were done to make the compiler happy.
The first change (inside card-gids.c) should be fine.
The caller cares about the termination / overflow that may happen.

I'm unsure about the second change.

Yes, you are right, both changes could be solved better using a `strlcpy`.
To minimize code-depth I would do it this way:

```c
size_t strlcpy(char *dst, const char *src, const size_t dstsize)
{
  if (dstsize == 0)
    return 0;

  size_t len = strlen(src);
  size_t bl = (len < dstsize-1 ? len : dstsize-1);
  
  ((char*)memcpy(dst, src, bl))[bl] = 0;

  return len;
}
```

I'm not very familiar with the OpenSC code.
So where to place the implementation and prototype to make it usable in both cases?
Maybe `src/libopensc/internal.h` should be used for the definition and `src/libopensc/sc.c` for implementation?

Btw: `src/common/compat_strlcpy.{c,h}` already provides a strlcpy function.

mouse07410:
I probably can live with these changes...

frankmorgner:
thanks to both of you for reviewing and updating the changes

PR #1356, opened 2018-05-13
Build fixes

marschap:
This pull request contains slight fixes for the build environment.
It comprises the following commits
* configure.ac: stop including version.m4
  As far as I understand version.m4 is the soruce for version.m4.ci created by bootstrap.ci
  In my opinion it should not be included directly into configure.ac as it overwrites
  already defined macros with outdated values.

* version.m4: overhaul
  Limit version.m4 only to those macros mangled by bootstrap.ci to avoid overriding
  "innocent" macros not used by bootstrap.ci

* flexibilize option -R
  Accept -rc or ~rc version string extensions in git descibe's output

* configure.ac: extend AC_INIT
  Define PRODUCT_URL and use it in AC_INIT()

It would be cool if some or all of them made their way into OpenSC


frankmorgner:
Looks good! Also, `PACKAGE_VERSION_REVISION` isn't used anywhere and can thus be deleted.

marschap:
Amended the pull request to get rid of unused PACKAGE_VERSION_REVISION and clean up bootstrap.ci a little.

marschap:
wrongly closed - reopen

marschap:
@mouse07410 You can also drown in useless information.
The script still writes to stdout when it changes the suffix, and I think this is sufficient, instead of having a full trace of all lines executed. We don't have that with configure either.
In addition:
* the script is short & trivial - no need for a log
* in case of errors around this script, it is still possible to re-enable the tracing on a case-by-case basis

Anyway: I leave the decision to @frankmorgner.
I do not want to make this last commit the reason for the rest of the pull request not being accepted.

PR #1364, opened 2018-05-20
openpgp driver cleanup

marschap:
Hi,

the commits in this pull requests implement correct parsing of the hist_bytes in ATR and the hist_bytes DO as compact-TLV instead of blindly searching for the key-byte.
In addition they prepare for parsing the hist_byte DO for features becoming available with OpenPGP card version 3.0 and higher.

Tested with:
* ZeitControl OpenPGP card v2.0
* Zeitcontrol OpenPGP card v3.3
* Yubikey NEO

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -85,8 +85,11 @@ enum _version {		/* 2-byte BCD-alike encoded version number */
 	OPENPGP_CARD_1_1 = 0x0101,
 	OPENPGP_CARD_2_0 = 0x0200,
 	OPENPGP_CARD_2_1 = 0x0201,
+	OPENPGP_CARD_2_2 = 0x0202,
 	OPENPGP_CARD_3_0 = 0x0300,
 	OPENPGP_CARD_3_1 = 0x0301,
+	OPENPGP_CARD_3_2 = 0x0302,
+	OPENPGP_CARD_3_3 = 0x0303,
 };
 
 enum _access {		/* access flags for the respective DO/file */
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -505,24 +505,26 @@ static int
 pgp_get_card_features(sc_card_t *card)
 {
 	struct pgp_priv_data *priv = DRVDATA(card);
-	unsigned char *hist_bytes = card->atr.value;
-	size_t atr_len = card->atr.len;
+	u8 *hist_bytes = card->reader->atr_info.hist_bytes;
+	size_t hist_bytes_len = card->reader->atr_info.hist_bytes_len;
 	size_t i;
 	pgp_blob_t *blob, *blob6e, *blob73;
 
-	/* parse card capabilities from historical bytes */
-	for (i = 0; (i < atr_len) && (hist_bytes[i] != 0x73); i++)
-		;
-	/* IS07816-4 hist bytes 3rd function table */
-	if ((hist_bytes[i] == 0x73) && (atr_len > i+3)) {
-		/* bit 0x40 in byte 3 of TL 0x73 means "extended Le/Lc" */
-		if (hist_bytes[i+3] & 0x40) {
-			card->caps |= SC_CARD_CAP_APDU_EXT;
-			priv->ext_caps |= EXT_CAP_APDU_EXT;
+	if (hist_bytes_len > 0) {
+		/* parse card capabilities from historical bytes */
+		for (i = 0; (i < hist_bytes_len) && (hist_bytes[i] != 0x73); i++)
+			;
+		/* IS07816-4 hist bytes 3rd function table */
+		if ((hist_bytes[i] == 0x73) && (hist_bytes_len > i+3)) {
+			/* bit 0x40 in byte 3 of TL 0x73 means "extended Le/Lc" */
+			if (hist_bytes[i+3] & 0x40) {
+				card->caps |= SC_CARD_CAP_APDU_EXT;
+				priv->ext_caps |= EXT_CAP_APDU_EXT;
+			}
+			/* bit 0x80 in byte 3 of TL 0x73 means "Command chaining" */
+			if (hist_bytes[i+3] & 0x80)
+				priv->ext_caps |= EXT_CAP_CHAINING;
 		}
-		/* bit 0x80 in byte 3 of TL 0x73 means "Command chaining" */
-		if (hist_bytes[i+3] & 0x80)
-			priv->ext_caps |= EXT_CAP_CHAINING;
 	}
 
 	if (priv->bcd_version >= OPENPGP_CARD_2_0) {
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -497,6 +497,41 @@ pgp_init(sc_card_t *card)
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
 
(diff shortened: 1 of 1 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -497,6 +500,41 @@ pgp_init(sc_card_t *card)
...
+/**
+ * Internal: get features of the card: capabilities, ...
+ */
+static void
+pgp_parse_hist_bytes(sc_card_t *card, u8 *ctlv, size_t ctlv_len)
+{
+	struct pgp_priv_data *priv = DRVDATA(card);
+	size_t offs;
+
+	for (offs = 0; offs < ctlv_len; offs++) {
```
frankmorgner: I think you should better use `sc_asn1_read_tag()` to cope with longer tags/values.

It's also possible to make this functionality available to all drivers, in `iso7816.c` via something like `int iso7816_process_hist_bytes(struct sc_card *card, unsigned long *caps, const unsigned char *buf, size_t buflen);`

src/libopensc/card-openpgp.c
```diff
@@ -510,20 +545,19 @@ pgp_get_card_features(sc_card_t *card)
 	size_t i;
 	pgp_blob_t *blob, *blob6e, *blob73;
 
+	/* parse card capabilities from historical bytes in ATR */
 	if (hist_bytes_len > 0) {
```
marschap: This part deals with the hist_bytes in the ATR.
They are optional, but if present, they serve as good fallback values.

src/libopensc/card-openpgp.c
```diff
@@ -505,45 +543,33 @@ static int
...
-				/* bit 0x40 in byte 3 of TL 0x73 means "extended Le/Lc" */
-				if (blob->data[i+3] & 0x40) {
-					card->caps |= SC_CARD_CAP_APDU_EXT;
-					priv->ext_caps |= EXT_CAP_APDU_EXT;
-				}
-				/* bit 0x80 in byte 3 of TL 0x73 means "Command chaining" */
-				if (blob->data[i+3] & 0x80)
-					priv->ext_caps |= EXT_CAP_CHAINING;
-			}
+			pgp_parse_hist_bytes(card, hist_bytes+1, hist_bytes_len-4);
```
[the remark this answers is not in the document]
marschap: Correct, and this is still done for all cards with version 2.0 or higher (per the spec).
See the unchanged logic 5 lines above.

pgp_parse_hist_bytes() simply parses the hist_bytes given to it, no matter where they come from: ATR or DO.
That part of the logic has not changed.

I only re-factored the parsing of the hist_bytes:
I replaced the clumsy and error-prone linear search, which appeared duplicated in the code by a new function that correctly parses hist_bytes as compact TLVs.

src/libopensc/card-openpgp.c
marschap: Using sc_asn1_read_tag() is not possible. 
hist_bytes use compact TLV where tag & length are encoded in one byte: 
- tag in the high nibble, 
- length of the value in the low nibble.

 This is different what sc_asn1_read_tag() deals with, and there is no need to deal with longer tags/values

I am not sure it is possible to have a generic hist_byte parsing function.
- According to what I read in the specs, the format of the hist_bytes starting with certain prefixes is defined as compact-TLV, but the individual tags and their lengths/values are not defined.
E.g. the tag 0x73 (to be painfully correct: tag 0x70 with length 3) and its meaning are specific to OpenPGP cards only.
=> If we wanted to implement a generic hist_byte parser, this would require card-specifc call-back functions.
- In addition: the only other card driver that seems to deal with hist_bytes as compact TLVs is the PIV driver.
=> very limited use today

Please let this request not block the safety improvements in hist_file parsing for OpenPGP.

dengert:
Look at `sc.c` `_sc_parse_atr` which parses the ATR for you and sets 
`reader->atr_info.hist_bytes_len = n_hist;` and  `reader->atr_info.hist_bytes = p;`

It does not parse the historical bytes. But an example of a driver that does  is `card-piv.c` It looks for `Yubikey4` or `Yubikey`, or if  it compact TLV, then looks for a default AID in the historical bytes and if it is the PIV AID.  

`_sc_parse_atr` could be expanded to include some of these, or support for compact TLV tags and lengths could be added as functions.

src/libopensc/card-openpgp.c
frankmorgner: You're right regarding the use of `sc_asn1_read_tag()`.

However, I still believe that generalizing the ISO-defined compact DOs is useful. For example, you could use it in the `init()` function of the default driver (`iso7816.c`) to detect those features. (This is currently `NULL` and typically overwritten when some card driver uses `iso_ops`.) And indeed, the default selected applet in the historical bytes (as used by the PIV driver) is also ISO-defined.

frankmorgner:
There are a lot of cards out there, that have historical bytes that don't conform with the ISO definitions, so it's better to keep the parsing within the scope of the card driver!

src/libopensc/card-openpgp.c
marschap: What do you mean by " ISO-defined compact DOs"?

ISO 7816-4 only talks about historical bytes in the ATR, some ISO supported formats (COMPACT-TLV), and some defined tags in those formats.
see: http://cardwerk.com/smart-card-standard-iso7816-4-section-8-historical-bytes/

ISO does not say anything about a hist_byte DO, as used in the OpenPGP card 2+ spec.

What about accepting this pull request and see whether some of it can be generalized later.

In any case, the parsing of hist_bytes in this pull request is formally correct (it parses compact-TLV), while the old one is plain wrong (it does linear search and triggers on the first occurence of a byte that looks like a tag value, but may be anything)

Please pull, and let's not make " best" the enemy of "better"!

dengert:
Those cards do conform, in that ISO 7816-4 (2016) Table 114 - Category indicator byte says:
```
Table 114 — Category indicator byte
Value           Meaning
'00'            A status indicator shall be present as the last three historical bytes (see 12.1.1.11)
'80'            A status indicator may be present in a COMPACT-TLV data object (one, two or three bytes, see 12.1.1.11)
'81' to '8F'    RFU
-----           Any other value indicates a proprietary format.
```

Most vendors I have seen, put their name  (char)  which ends up being the last case i.e. "Any other value indicates a proprietary format". In which case any general parse of the Historical bytes would not try and parse this, leaving the card driver to parse it if needed. 

Someone could look at the @LudovicRousseau 's smartcard_list.txt to see if any known cards do not follow the above table.  

In regard to a ATR as a file, ISO 7816-4 Section 12.2.2 EF.ATR/INFO says:
"NOTE EF.ATR/INFO was called EF.ATR in the previous edition of this document...."
"If EF.ATR/INFO supports DO handling, or if its contents are not stored in a file, a card may provide its
contents by the GET DATA command (see 11.4.3)."
I believe the OpenPGP is following this and the DO can be retrieve with all the other DOs. 

Whatever we do, we should make it easier to parse compact TLV.

src/libopensc/card-openpgp.c
marschap: What about a little tit-for-tat?

I promise that I will implement something like sc_compacttlv_read_tag() and adapt the openpgp driver to it, and in return you merge this pull request.

You can use https://github.com/OpenSC/OpenSC/pull/1372 as an indication that I stick to the promises I gave.

frankmorgner: Agreed, thanks

PR #1367, opened 2018-05-23
HID CAC Alt tokens improvements

Jakuje:
Recently, there showed up s new CAC Alt tokens from HID, which do not completely adhere to the old GSC-IS 2.1 specification (not sure if there is a newer) and that were trouble already some time ago (#1183). Once they started be used, some of them did not work, because they are enrolling all the certificates as separate objects inside the first PKI applet. These were ignored with the current CAC implementation, because no other card before used this technique to do anything useful.

This pull request tries to inspect the Alt token without CCC from other sources (Service Applet Table, Applet Properties) and locate all the PKI objects, rather than guessing AIDs and trying if they respond.

Additionally, there are few more changes, that made its way to this PR and that are inside of CAC driver, mostly logging of SimpleTLV buffer contents from CCC and other places, proper re-encoding the T and V buffers to PKCS#15 layer including the last element, ignoring data that is not in SimpleTLV (because of lack of other ideas what to do them), bug fixes, typos and comments for future improvements.

Let me know if some of the changes need some more clarification.

The code was tested with standard CAC cards and HID Alt tokens (uninitialized). I am still waiting for customer to test with they provisioned cards to confirm the complete functionality.

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/card-cac.c
```diff
@@ -421,6 +454,50 @@ static int cac_apdu_io(sc_card_t *card, int ins, int p1, int p2,
...
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	/* for simplicity we support only ACR without arguments now */
+	if (acr_type != 0x00 && acr_type != 0x10
+	    && acr_type != 0x20 && acr_type != 0x21) {
+		return SC_ERROR_INVALID_ARGUMENTS;
+	}
+
+	out = out_ptr = malloc(len);
```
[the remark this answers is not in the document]
Jakuje: Thanks. It will be cleaner. I will fix it in future commit.

src/libopensc/card-cac.c
```diff
@@ -421,6 +454,50 @@ static int cac_apdu_io(sc_card_t *card, int ins, int p1, int p2,
...
+	    "got %"SC_FORMAT_LEN_SIZE_T"u bytes out_ptr=%p", len, out_ptr);
+
+	*out_len = len;
+	*out_buf = out;
+	return SC_SUCCESS;
+
+fail:
+	if (out)
+		free(out);
+	*out_len = 0;
```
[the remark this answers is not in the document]
Jakuje: True. There was a bit mess in the pointers caused by the allocation above. I reworked this to make it more clear and make sure the results is as expected.

src/libopensc/card-cac.c
```diff
@@ -1008,6 +1094,129 @@ static int cac_decipher(sc_card_t *card,
...
+			/* ignore tags we don't understand */
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Unknown (0x%02x)",tag );
+			break;
+		}
+	}
+	if (parsed < 2)
+		return SC_ERROR_INVALID_DATA;
+
+	(*i)++;
```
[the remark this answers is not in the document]
Jakuje: The intention here was to avoid too deep nesting (two for cycles, two switches) and I was not yet sure if I will call the function from one place or from two (since it works with two tags that are handled by this function).

I can refactor the structure so I can pass the whole object that I am filling and increment & check only in one place, while checking for correct bounds.

src/libopensc/card-cac.c
```diff
@@ -651,7 +728,7 @@ static int cac_read_binary(sc_card_t *card, unsigned int idx,
 		priv->cache_buf_len = tlv_len;
 
 		for (tl_ptr = tl, val_ptr=val, tlv_ptr = priv->cache_buf;
-				tl_len > 2 && val_len > 0 && tlv_len > 0;
+				tl_len >= 2 /*&& val_len >= 0*/ && tlv_len > 0;
```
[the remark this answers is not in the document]
Jakuje: The idea here was to make sure this condition is stated, even though it is always true (unsigned type). Technically, it is checked for sanity inside the cycle.
Obviously, it was not clear so I will just remove it. I can not think of a good comment that would explain this in few words.

src/libopensc/card-cac.c
```diff
@@ -1390,6 +1690,56 @@ static int cac_process_CCC(sc_card_t *card, cac_private_data_t *priv)
...
+			if (len != 5)
+				break;
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Applet Information: Family: 0x%02x", val[0]);
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "     Applet Version: 0x%02x 0x%02x 0x%02x 0x%02x",
+			    val[1], val[2], val[3], val[4]);
+			break;
+		case CAC_TAG_NUMBER_APPLETS:
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
```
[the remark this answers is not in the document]
Jakuje: thanks. Added.

src/libopensc/card-cac.c
```diff
@@ -1390,6 +1690,56 @@ static int cac_process_CCC(sc_card_t *card, cac_private_data_t *priv)
...
+			    "TAG: Num applets = %hhd", *val);
+			break;
+		case CAC_TAG_APPLET_ENTRY:
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Applet Entry");
+			/* This is SimpleTLV prefixed with applet ID (1B) */
+			r = cac_parse_aid(card, priv, &val[3], val[2]);
```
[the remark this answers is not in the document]
Jakuje: Thanks. Added.

src/libopensc/card-cac.c
```diff
@@ -1008,6 +1096,130 @@ static int cac_decipher(sc_card_t *card,
...
+			object->simpletlv = val[0];
+			parsed++;
+			break;
+
+		case CAC_TAG_PKI_PROPERTIES:
+			/* 4th byte is "Private Key Initialized" */
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: PKI Properties: Private Key Initialized = 0x%02x",
+			    val[2]);
+			object->privatekey = val[2];
```
frankmorgner: This is still not checking the length, right? Here, you should check whether `len >= 3`. For the other cases above it's similar...

src/libopensc/card-cac.c
```diff
@@ -1008,6 +1096,130 @@ static int cac_decipher(sc_card_t *card,
...
+			    "TAG: Applet Information: Family: 0x%0x", val[0]);
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "     Applet Version: 0x%02x 0x%02x 0x%02x 0x%02x",
+			    val[1], val[2], val[3], val[4]);
+			break;
+		case CAC_TAG_NUMBER_OF_OBJECTS:
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Num objects = %hhd", *val);
+			/* make sure we do not overrun buffer */
+			prop->num_objects = MIN(val[0], CAC_MAX_OBJECTS);
```
frankmorgner: check for `len >= 1`

src/libopensc/card-cac.c
```diff
@@ -1148,7 +1362,37 @@ static int cac_select_file_by_type(sc_card_t *card, const sc_path_t *in_path, sc
 	if (r)
 		LOG_FUNC_RETURN(ctx, r);
 
-		/* CAC cards never return FCI, fake one */
+	/* This needs to come after the applet selection */
+	if (priv && in_path->len) {
```
frankmorgner: `in_path->len >= 2`

src/libopensc/card-cac.c
```diff
@@ -1313,9 +1610,14 @@ static int cac_parse_CCC(sc_card_t *card, cac_private_data_t *priv, u8 *tl,
 				return r;
 			break;
 		case CAC_TAG_CC_VERSION_NUMBER:
+			/* ignore the version numbers for now */
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+				"TAG:CC Version = 0x%02x", *val);
```
frankmorgner: need to check for `len >= 1`

src/libopensc/card-cac.c
```diff
@@ -1313,9 +1610,14 @@ static int cac_parse_CCC(sc_card_t *card, cac_private_data_t *priv, u8 *tl,
...
+			break;
 		case CAC_TAG_GRAMMAR_VERION_NUMBER:
 			/* ignore the version numbers for now */
-			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,"TAG:Version");
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+				"TAG:Grammar Version = 0x%02x", *val);
```
frankmorgner: need to check for `len >= 1`

src/libopensc/card-cac.c
```diff
@@ -1327,21 +1629,22 @@ static int cac_parse_CCC(sc_card_t *card, cac_private_data_t *priv, u8 *tl,
 		 * The following are really for file systems cards. This code only cares about CAC VM cards
 		 */
 		case CAC_TAG_PKCS15:
-			/* should verify that this is '0'. If it's not zero, we should drop out of here and
-			 * 	let the PKCS 15 code handle this card */
-			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,"TAG:PKCS5");
+			/* TODO should verify that this is '0'. If it's not
+			 * zero, we should drop out of here and let the PKCS 15
+			 * code handle this card */
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,"TAG:PKCS15 = 0x%02x", *val);
```
frankmorgner: need to check for `len >= 1`

src/libopensc/card-cac.c
Jakuje: True. The specification clearly says that this tag should have 4 bytes so explicit check for 4B length makes sense here. Thanks.

Jakuje:
Thank you for the comments. The above issues should be addressed by explicit checking of the buffer sizes.

src/libopensc/card-cac.c
```diff
@@ -1390,6 +1757,68 @@ static int cac_process_CCC(sc_card_t *card, cac_private_data_t *priv)
...
+				break;
+			}
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Num applets = %hhd", *val);
+			break;
+		case CAC_TAG_APPLET_ENTRY:
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "TAG: Applet Entry");
+			/* Make sure we match the outer length */
+			if (val[2] != len - 3) {
```
frankmorgner: need to check the length

Jakuje: Thank you. This is fixed in  c7efb82

frankmorgner:
The code looks good now.

Please add a switch to `configure` for setting `WITH_CAC1` or document the optional define via `AC_ARG_VAR`. This would avoid conflicts in the standard card detection but still allow using CAC1 with a little more effort.

Jakuje:
Oh ... I pushed to this branch also the CAC1 removal, which was not intentional. The commenting out the code out was more like proof of concept and I rather incline to the complete removal.

For the rest of the changes about HID Alt tokens, lets wait few more days before merging. I am still waiting for more feedback on the functionality.

frankmorgner:
OK

Jakuje:
I removed the CAC1 removal (it was not intentionally added and requires more discussions), added few more minor issues in the last two commits and rebased on current master.

There were no other issues reported yet, so I guess we are fine to merge it. If there will be further issues, I will fill a separate PR.

frankmorgner:
thanks

PR #1373, opened 2018-06-01
Fix #1371 introduced by 45ad44e

### LINKED ISSUE #1371: SmartCardHSM - Cannot Unwrap Key, "Invalid wrapped key format (Key binary)."
### Problem Description
Following the instructions in the [SmartCardHSM Wiki - Key Backup and Restore](https://github.com/OpenSC/OpenSC/wiki/SmartCardHSM#using-key-backup-and-restore), I'm seeing a ```Invalid wrapped key format (Key binary).``` error message when attempting to unwrap a private key. The ```sc-hsm-tool.exe``` wraps the key without error but then errors out when attempting to unwrap. I've attempted unwrapping to a secondary card (same DKEK) or the same card; the same error message is displayed in both cases.

I'm not sure if this would affect the wrap/unwrap procedures but I deviated from that section of the wiki in the length of the user PIN, i.e., 8 digits instead of 6.

**Environment Information**
- **Platform**: Windows 10 - Version 1803
- **OpenSC Version**: OpenSC 0.18.0 [Microsoft 1800] (32-Bit) - Enabled features:pcsc openssl zlib
- **Reader**: Broadcom Corp Contacted SmartCard
- **Card** : SmartCard-HSM version 2.5

### Proposed Resolution
Unknown

### Steps to reproduce
Per the Wiki entry mentioned above, I'm using the following commands when I encounter the error.
```
sc-hsm-tool --wrap-key wrap-key.bin --key-reference 1
sc-hsm-tool --unwrap-key wrap-key.bin --key-reference 10
```

### Logs
Debug log from the command line, i.e., using the ```-vvv``` parameter.
https://gist.github.com/CampinCarl/d26f92dd7dde8d06252754e98aac2189#file-gistfile1-txt-L93

CardContact:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/sc-hsm-tool.c
+++ b/src/tools/sc-hsm-tool.c
@@ -1524,7 +1524,7 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
 	}
 
 	if ((sc_asn1_read_tag(&ptr, len, &cla, &tag, &olen) != SC_SUCCESS)
-		   	|| ((cla & SC_ASN1_TAG_CONSTRUCTED) != SC_ASN1_TAG_CONSTRUCTED)
+		   	|| ((cla & SC_ASN1_TAG_CONSTRUCTED) == SC_ASN1_TAG_CONSTRUCTED)
 		   	|| (tag != SC_ASN1_TAG_OCTET_STRING) ){
 		fprintf(stderr, "Invalid wrapped key format (Key binary).\n");
 		return -1;


frankmorgner:
The ASN.1 data must *not* be constructed? If so, wouldn't it be better to check for `cla == SC_ASN1_TAG_PRIMITIVE` instead?

CardContact:
No, because SC_ASN1_TAG_PRIMITIVE is not the opposite of SC_ASN1_TAG_CONSTRUCTED, but a mask to extract the lower 5 bit of the tag value.

PR #1374, opened 2018-06-02
compact TLV - add generic sc_compacttlv_find_tag() function

marschap:
Hi,

as promised, here's the pull request that replaces the "proprietary" implementation to parse the compact TLV in the hist_bytes of the openpgp card driver with a generic function in sc.c, and used it in the openpgp card driver.

It has been tested with
* OpenPGP v2.0 card
* OpenPGP v3.3 card
* Yubikey Neo

Please include it into opensc master

##### Checklist

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -507,32 +507,25 @@ static void
 pgp_parse_hist_bytes(sc_card_t *card, u8 *ctlv, size_t ctlv_len)
 {
 	struct pgp_priv_data *priv = DRVDATA(card);
-	size_t offs;
+	const u8 *ptr;
 
-	for (offs = 0; offs < ctlv_len; offs++) {
-		switch (ctlv[offs]) {
-			case 0x73: /* IS07816-4 hist bytes 3rd function table */
-				if (offs+3 < ctlv_len) {
-					/* bit 0x40 in byte 3 of TL 0x73 means "extended Le/Lc" */
-					if (ctlv[offs+3] & 0x40) {
-						card->caps |= SC_CARD_CAP_APDU_EXT;
-						priv->ext_caps |= EXT_CAP_APDU_EXT;
-					}
-					/* bit 0x80 in byte 3 of TL 0x73 means "Command chaining" */
-					if ((ctlv[offs+3] & 0x40) &&
-					    (priv->bcd_version >= OPENPGP_CARD_3_0)) {
-						priv->ext_caps |= EXT_CAP_CHAINING;
-					}
-				}
-				break;
-			case 0x31:
-				if ((offs + 1 < ctlv_len) &&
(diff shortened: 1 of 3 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -507,32 +507,25 @@ static void
...
-				    (priv->bcd_version >= OPENPGP_CARD_3_0)) {
-					// ToDo ...
-				}
-				break;
+	const u8 *ptr;
+
+	/* IS07816-4 hist bytes: 3rd function table */
+	if ((ptr = sc_compacttlv_find_tag(ctlv, ctlv_len, 0x73)) != NULL) {
+		/* bit 0x40 in byte 3 of TL 0x73 means "extended Le/Lc" */
+		if (ptr[2] & 0x40) {
```
frankmorgner: The length needs to be checked.

Ideally, you could change the signature of `sc_compacttlv_find_tag()` to match `sc_asn1_find_tag()` for consistency (i.e. with a seperate input/output for lenght and value). For a (new) programmer it's easier if all parsing routines allow the same usage pattern.

marschap: Re length: 
No, it doesn't. At least not in the callers of sc_compacttlv_find_tag().
That's exactly the "trick" with compact TLV: the "tag" is a combination of "pure" tag in the upper nibble, and the length in the lower nibble, e.g. 0x73 = "pure" tag 0x70 + length 0x03.
I.e. sc_compacttlv_find_tag() is passed both: "pure" tag + length in one argument.
The length check is done in sc_compacttlv_find_tag() itself; no need to check

Re function signature:
In my opinion that's over-engineering for no gain. The peculiarity that you have with compact TLV is that you get the tag & length in one byte. Artificially splitting it makes everything less understandable. 
OpenPGP documentation and even ISO7816-4 talk refer to the combination as e.g. 0x73, 0x31, ...
Splitting these two values - just for the sake of getting a similar signature as sc_asn1_find_tag() makes it harder for every user to grasp.

frankmorgner: Sorry, I just now realized that your `tag` argument consists of a tag and a length. You're pre-defining a length which should be variable; that's how ISO 7816-4 defines compact TLV (although some specific combinations with fixed length are also defined). Please implement `sc_compacttlv_find_tag()` as generic parser.

marschap:
Requested changes should be addressed with the lattest commits added.
Please review & pull.

frankmorgner:
@dengert do you think you `sc_compacttlv_find_tag()` is useful for PIV as well?

dengert:
Yes it could be useful. card-piv.c could use it in one place to  parse ATR historical bytes for tag 0xF0 to get default AID. Some PIV cards set this to the PIV AID.  This could save 5 lines of code.

PR #1375, opened 2018-06-02
openpgp-tool cleanups

marschap:
This pull-request sixes some slight annoyances and inconsisencies in openpgp-tool

Please consider including it in OpenSC master.

### DIFF
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -316,6 +316,7 @@ static int decode_options(int argc, char **argv)
 				key_id = optarg[0] - '0';
 			else                              /* Arg string is 'all' */
 				key_id = 'a';
+			actions++;
 			break;
 		case 'd':
 			do_dump_idx = optarg[0] - '0';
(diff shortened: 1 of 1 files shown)


src/tools/openpgp-tool.c
```diff
@@ -419,7 +435,7 @@ static int do_dump_do(sc_card_t *card, unsigned int tag)
 		if (sizeof(buffer) != r)
 			return EXIT_FAILURE;
 	} else {
-		util_hex_dump_asc(stdout, buffer, sizeof(buffer), -1);
+		util_hex_dump_asc(stdout, buffer, r, -1);
```
frankmorgner: I think `r` also needs to be respected in line 410:
```
		if (fp) {
			r = (int)fwrite(buffer, sizeof(char), sizeof(buffer), fp);
		}
```
... and line 419-420 needs to be removed:
```
		if (sizeof(buffer) != r)
			return EXIT_FAILURE;
```
or am I getting this wrong?

marschap:
added a commit that re-factored do_dump_do() and addressed your issues.

PR #1376, opened 2018-06-03
card-dnie: use size_t-specific format specifier in printf

marschap:
Instead of blindly using "%lu", use "%"SC_FORMAT_LEN_SIZE_T"u" to cope with the various implementations.

This fixes a bug introduced in commit 20b1d829, which prevents building on 32-bit or Windows systems.


frankmorgner:
thanks

PR #1378, opened 2018-06-05
opensc-notify: make arguments to options -t & -m required

marschap:
There is no real point in having the arguments of an option be optional.
If you really do not want the argument to be set, simply do not use the option.

This is especially true, if the option argument is mandatory in the program
logic. In opensc-notify's case, this applies to title which is mandatory for
the GIO case.

Instead of everyone forcing to use the counterintuitive
	opensc-notify -t="title" -m="message text"
syntax, use the well-known
	opensc-notify -t "title" -m "message text"
and have those who whant to have a message with an empty title go the
extra mile of having to write
	opensc-notify -t ""

What's the use case of a notification with an empty title anyway?

Please note: this pull reuest requires postprocessing: 
- gengetopt has to be called to update opensc-notify-cmdline.*

### DIFF
--- a/src/tools/opensc-notify.ggo.in
+++ b/src/tools/opensc-notify.ggo.in
@@ -13,13 +13,11 @@ modeoption "title"          t
     "Title of the notification"
     string
     mode="customized"
-    argoptional
     optional
 modeoption "message"        m
     "Main text of the notification"
     string
     mode="customized"
-    argoptional
     optional
 
 modeoption "notify-card-inserted"    I


frankmorgner:
Whether the title or the message text are required or not, is platform dependent. On macOS, if I remember correctly, a notification with either of the two looks good, but with GIO at least a title is required. Since we're talking about a low level tool here, I would not like to predefine how to use the given functionality in too much detail and leave the code as is.

Issue #1379, opened 2018-06-05
Please hide opensc-notify from application list


bigon:
Hi,

opensc-notify has a .desktop file, but it makes no sense to display the icon in the applications list/launcher of the user as it doesn't provide an persistent interface and cannot really be "started"

PR #1380, opened 2018-06-05
Do not display opensc-notify in the application list

bigon:
opensc-notify doesn't propose a GUI that can be displayed to the users,
so it doesn't make sense to display it in the application list/launcher

Fixes: #1379

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
thanks

PR #1386, opened 2018-06-07
Filter certificates other than CKC_X_509

CardContact:
This patch adds a filter to ignore certificates other than CKC_X_509 (e.g. CVC certificates as used in [1])

[1] https://github.com/CardContact/sc-hsm-embedded


Jakuje:
This looks good to me. I did not have any non-x509 certificates on the cards I tested.

PR #1387, opened 2018-06-09
refactor PIV driver to use sc_compacttlv_find_tag()

marschap:
This pull request adapts the PIV driver to use sc_compacttlv_find_tag() to parse the hist bytes.
In addition it limits the scope of some variables.

It is only compile-tested, as I do not own a PIV card.
It is strongly recommended to test with a real PIV card whether it is correct.

@dengert Could you have a look at it.

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2985,8 +2985,7 @@ static int piv_match_card(sc_card_t *card)
 
 static int piv_match_card_continued(sc_card_t *card)
 {
-	int i, i7e;
-	sc_file_t aidfile;
+	int i;
 	int type  = -1;
 	piv_private_data_t *priv = NULL;
 	int saved_type = card->type;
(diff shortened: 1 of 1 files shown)


dengert:
I had also written a similar modification and tested on a PIV card that has the AID in the historical bytes.
But was waiting for the sc_compacttlv_find_tag to be committed to master.

It looks like yours should work. 
Why did you need to limit the scope of some variables?

dengert:
Actually it was not tested yet, I was waiting for the commit of the sc_compacttlv_find_tag.

dengert:
Tested this PR using GDB with NIST Demo card 2 with ATR:  `3b:df:96:00:81:b1:fe:45:1f:83:80:73:cc:91:cb:f9:a0:00:00:03:08:00:00:10:00:79` 
PR parsed the Historical bytes to find tag F9 and PIV AID of `a0:00:00:03:08:00:00:10:00`

OK to merge.

Issue #1377, opened 2018-06-04
Remove the CAC1 driver (another issues with card detection)


Jakuje:
### Problem Description

TL;DR: Remove CAC1 driver. It breaks detection of misbehaving cards (soon after reset?).

I run the [CI on master](https://gitlab.com/redhat-crypto/OpenSC/). We have a machine with two-slot reader with CAC and coolkey cards inserver. It worked fine until last week. The cards got swapped last week and now the coolkey card is no longer detected from my testsuite, while it works just fine if:

 * The `OPENSC_DRIVER` is set to `coolkey`
 * Configuration sets `card_drivers = coolkey, cac`
 * From `pkcs11-tool` (which is opening only one session, not more as a testsuite)

The `PKCS11SPY` trace looks like this (while before there was one more iteration of `C_Initialize()`, `C_Finalize()` before, whicih could have messed up something):
```
15: C_GetSlotInfo
2018-06-04 05:58:05.473
[in] slotID = 0x0
[out] pInfo: 
      slotDescription:        'OMNIKEY CardMan (076B:3022) 3021'
                              ' 00 00                          '
      manufacturerID:         '                                '
      hardwareVersion:         0.0
      firmwareVersion:         0.0
      flags:                   7
        CKF_TOKEN_PRESENT                
        CKF_REMOVABLE_DEVICE             
        CKF_HW_SLOT                      
Returned:  0 CKR_OK

16: C_OpenSession
2018-06-04 05:58:06.196
[in] slotID = 0x0
[in] flags = 0x6
pApplication=(nil)
Notify=(nil)
[out] *phSession = 0x16785f0
Returned:  0 CKR_OK

17: C_GetMechanismList
2018-06-04 05:58:06.196
[in] slotID = 0x0
Returned:  224 CKR_TOKEN_NOT_PRESENT
```

From the logs, there are two separate issues:

 * Misdetection of coolkey cards as CAC (again)
 * Breaking some data structures while issuing several cycles of Initialize/Finalize (?)

I can isolate the first issue to the configuration, where CAC driver is in front of coolkey driver, where I start seeing detection issues (the `pkcs11-tool` still detects the card, but in testsuite, it shows without any objects), but not yet this complete brokeness.

[...]

PR #1389, opened 2018-06-11
Add CI status of CAC and coolkey cards + few more accumulated issues

Jakuje:
Based on the discussion in #1258, I tweaked the table in the readme a bit and reworked the CI to report separately for both cards.

Also there were few more minor issues in the master that did not make it to any other PR yet:
 * The sc-hsm initialization code based on the code by @CardContact from #1377
 * Some of the variables in the testsuite build was not correctly defined, which prevented autodetection
 * The cleanup of the muscle applet seemed fine, until I opened the `free_*` function

The CI with valgrind passes as attached badges show.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,10 @@ Wiki is [available online](https://github.com/OpenSC/OpenSC/wiki)
 Please take a look at the documentation before trying to use OpenSC.
 
 [![Travis CI Build Status](https://travis-ci.org/OpenSC/OpenSC.svg)](https://travis-ci.org/OpenSC/OpenSC/branches) [![AppVeyor CI Build Status](https://ci.appveyor.com/api/projects/status/github/OpenSC/OpenSC?branch=master&svg=true)](https://ci.appveyor.com/project/LudovicRousseau/OpenSC/branch/master) [![Coverity Scan Status](https://scan.coverity.com/projects/4026/badge.svg)](https://scan.coverity.com/projects/4026)
+
+Build and test status of specific cards:
+
+| Cards                 | Status |
+|-----------------------|--------|
+| CAC                   | [![CAC](https://gitlab.com/redhat-crypto/OpenSC/badges/cac/build.svg)](https://gitlab.com/redhat-crypto/OpenSC/pipelines) |
+| Coolkey               | [![Coolkey](https://gitlab.com/redhat-crypto/OpenSC/badges/coolkey/build.svg)](https://gitlab.com/redhat-crypto/OpenSC/pipelines) |
--- a/configure.ac
+++ b/configure.ac
@@ -518,9 +518,14 @@ if test "${enable_notify}" = "yes"; then
 	fi
 fi
 
-PKG_CHECK_MODULES( [CMOCKA], [cmocka],
-		  [ have_cmocka="yes" ],
-		  [ have_cmocka="no" ])
+have_cmocka="yes"
+PKG_CHECK_MODULES([CMOCKA], [cmocka >= 1.1.0],,[have_cmocka="no"])
+AC_CHECK_HEADER([setjmp.h])
+AC_CHECK_HEADER([cmocka.h],, [have_cmocka="no"],
+[#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+])
 
 if test "${enable_tests}" = "detect"; then
 	if test "${have_cmocka}" = "yes"; then
--- a/configure.ac
+++ b/configure.ac
@@ -531,9 +531,7 @@ if test "${enable_tests}" = "detect"; then
 fi
 
 if test "${enable_tests}" = "yes"; then
-	if test "${have_cmocka}" = "yes"; then
-		AC_DEFINE([ENABLE_TESTS], [1], [Build tests in tests/ subdirectory])
-	else
+	if test "${have_cmocka}" != "yes"; then
 		AC_MSG_ERROR([Tests required, but cmocka is not available])
 	fi
 fi
@@ -1021,7 +1019,7 @@ AM_CONDITIONAL([ENABLE_MINIDRIVER_SETUP_CUSTOMACTION], [test "${enable_minidrive
 AM_CONDITIONAL([ENABLE_SM], [test "${enable_sm}" = "yes"])
 AM_CONDITIONAL([ENABLE_DNIE_UI], [test "${enable_dnie_ui}" = "yes"])
 AM_CONDITIONAL([ENABLE_NPATOOL], [test "${ENABLE_NPATOOL}" = "yes"])
-AM_CONDITIONAL([ENABLE_TESTS], [test "${ENABLE_TESTS}" = "yes"])
+AM_CONDITIONAL([ENABLE_TESTS], [test "${enable_tests}" = "yes"])
 AM_CONDITIONAL([GIT_CHECKOUT], [test "${GIT_CHECKOUT}" = "yes"])
 
 if test "${enable_pedantic}" = "yes"; then
--- a/src/libopensc/card-cac.c
+++ b/src/libopensc/card-cac.c
@@ -1224,8 +1224,10 @@ static int cac_get_properties(sc_card_t *card, cac_properties_t *prop)
 			}
 			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
 			    "TAG: TV Object nr. %"SC_FORMAT_LEN_SIZE_T"u", i);
-			if (i >= CAC_MAX_OBJECTS)
+			if (i >= CAC_MAX_OBJECTS) {
+				free(rbuf);
 				return SC_SUCCESS;
+			}
 
 			if (cac_parse_properties_object(card, tag, val, len,
 			    &prop->objects[i]) == SC_SUCCESS)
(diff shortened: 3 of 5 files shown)


Jakuje:
(dropped the sc-hsm change, since it is already in #1390)
I will investigate the build failures since it looks like it behaves somehow differently on ubuntu and mac, than here.

Jakuje:
The tests were not built before the 843fc77 in CI because variables were not set correctly, if I am right. The Ubuntu fails, because it has really outdated version of CMocka:
> Setting up libcmocka0:amd64 (0.3.2-1) ...
really? We have 1.1.1 ... Lets see if better configure check can gate that out.

The cmocka in OSX probably needs some other include path? Does anyone here know how that works?

That was probably the reason why I initially got down to build the cmocka during the CI run.

Jakuje:
Yes, this helped to build on Ubuntu without tests (with correctly reporting from configure). Some hints for OSX would be appreciated, rather than me throwing random ideas on Travis. But I assume there will be different include path so something like
```
diff --git a/src/tests/p11test/Makefile.am b/src/tests/p11test/Makefile.am
index 15de3174..1b380543 100644
--- a/src/tests/p11test/Makefile.am
+++ b/src/tests/p11test/Makefile.am
@@ -24,6 +24,7 @@ p11test_SOURCES = p11test.c p11test_loader.c \
 	p11test_helpers.c
 p11test_CFLAGS = -DNDEBUG
 p11test_LDADD = -lssl -lcrypto -lcmocka
+p11test_INCFLAGS = `pkg-config cmocka --cflags-only-I`
 
 if WIN32
 p11test_SOURCES += $(top_builddir)/win32/versioninfo.rc
```
might solve the issue, but I am not sure how elegant is that.

frankmorgner:
If I'm not mistaken then `brew` installs the headers into `/usr/local/include` which may not be in the default path of the compiler...

In general, it would be great if you could add header/library detection to `configure.ac`

Jakuje:
Thank you for the pointer.
The library detection using pkg-config is [already there](https://github.com/OpenSC/OpenSC/blob/master/configure.ac#L521), but the header check is missing. I will try to put something together tomorrow.

Jakuje:
Header check is in place results in some errors already during configure:
```
checking for cmocka >= 1.1.0... yes
checking cmocka.h usability... no
checking cmocka.h presence... yes
configure: WARNING: cmocka.h: present but cannot be compiled
configure: WARNING: cmocka.h:     check for missing prerequisite headers?
configure: WARNING: cmocka.h: see the Autoconf documentation
configure: WARNING: cmocka.h:     section "Present But Cannot Be Compiled"
configure: WARNING: cmocka.h: proceeding with the compiler's result
configure: WARNING:     ## ------------------------------------------------------ ##
configure: WARNING:     ## Report this to https://github.com/OpenSC/OpenSC/issues ##
configure: WARNING:     ## ------------------------------------------------------ ##
checking for cmocka.h... no
```
so with this state, it builds, but again without tests. Is there any other way to get the build log than letting the configure fail or printing the file from the `.travis.yml`?

frankmorgner:
In Travis CI, there is [`after_failure`](https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle) for debugging the problem. Pasting `config.log` could be used to get some pointers to the problem.

README.md
```diff
@@ -5,3 +5,10 @@ Wiki is [available online](https://github.com/OpenSC/OpenSC/wiki)
 Please take a look at the documentation before trying to use OpenSC.
 
 [![Travis CI Build Status](https://travis-ci.org/OpenSC/OpenSC.svg)](https://travis-ci.org/OpenSC/OpenSC/branches) [![AppVeyor CI Build Status](https://ci.appveyor.com/api/projects/status/github/OpenSC/OpenSC?branch=master&svg=true)](https://ci.appveyor.com/project/LudovicRousseau/OpenSC/branch/master) [![Coverity Scan Status](https://scan.coverity.com/projects/4026/badge.svg)](https://scan.coverity.com/projects/4026)
+
+Build and test status of specific cards:
+
+| Cards                 | Status |
+|-----------------------|--------|
+| CAC                   | [![CAC](https://gitlab.com/redhat-crypto/OpenSC/badges/cac/build.svg)](https://gitlab.com/redhat-crypto/OpenSC/pipelines) |
+| Coolkey               | [![Coolkey](https://gitlab.com/redhat-crypto/OpenSC/badges/coolkey/build.svg)](https://gitlab.com/redhat-crypto/OpenSC/pipelines) |
```
frankmorgner: These URLs show a "500 Server Error"

Jakuje: It was probably momentary error when you tried. Now they seem to work fine.

Jakuje:
> In Travis CI, there is after_failure for debugging the problem. Pasting config.log could be used to get some pointers to the problem.

The log is already printed in case of configure fails, but it is not the case here -- it succeeds and decides that the tests are not built because cmocka is not usable. I fired one build to get the config.log with the recent commit and we will see.

Jakuje:
From what I understood from the [last log](https://api.travis-ci.org/v3/job/391747248/log.txt), the configure IS able to detect and use cmocka correctly during the first two invocations, but not during the last one, which is invoked this way from `MacOSX/build-package:60`:
```
+./configure --prefix=/Library/OpenSC --sysconfdir=/Library/OpenSC/etc --enable-cvcdir=/Library/OpenSC/etc/cvc --enable-x509dir=/Library/OpenSC/etc/x509 --disable-dependency-tracking --enable-shared --enable-static --enable-strict --disable-assert --enable-sm
```
In this case, the configure indeed just does not find the header file:
```
configure:19935: checking for cmocka.h
configure:19935: gcc -c  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -arch x86_64  conftest.c >&5
conftest.c:64:10: fatal error: 'cmocka.h' file not found
#include <cmocka.h>
         ^
1 error generated.
configure:19935: $? = 1
```
but the `config.log` is no more helpful here. If I try to run the same configure on my laptop with Linux, the cmocka is correctly detected, but I assume it will be some constellation of configure flags. Further ideas welcomed.

dengert:
One thing to try: add -v to the CPPFLAGS and CFLAGS  to see if the compiler will tell you the include paths

Jakuje:
First, I wonder, why the clang is used in the `gcc` target on osx. But it is probably the way how it works when the sdk and xcode is used.

The verbose log looks like this:
```
configure:19935: gcc -c  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -arch x86_64 -v  conftest.c >&5

Apple LLVM version 8.1.0 (clang-802.0.42)

Target: x86_64-apple-darwin16.7.0

Thread model: posix

InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" -cc1 -triple x86_64-apple-macosx10.10.0 -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name conftest.c -mrelocation-model pic -pic-level 2 -mthread-model posix -mdisable-fp-elim -masm-verbose -munwind-tables -target-cpu core2 -target-linker-version 278.4 -v -dwarf-column-info -debugger-tuning=lldb -coverage-file /Users/travis/build/OpenSC/OpenSC/conftest.c -resource-dir /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/8.1.0 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -fdebug-compilation-dir /Users/travis/build/OpenSC/OpenSC -ferror-limit 19 -fmessage-length 0 -stack-protector 1 -fblocks -fobjc-runtime=macosx-10.10.0 -fencode-extended-block-signature -fmax-type-align=16 -fdiagnostics-show-option -o conftest.o -x c conftest.c

clang -cc1 version 8.1.0 (clang-802.0.42) default target x86_64-apple-darwin16.7.0
```
Indeed, this is the problem. The `-isysroot` prevents from searching the following paths in system root and the compiler searches only under the sdk directory and there is nothing:
```
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/local/include"

ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/Library/Frameworks"

#include "..." search starts here:

#include <...> search starts here:

 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/8.1.0/include

 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include

 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include

 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks (framework directory)

End of search list.

conftest.c:64:10: fatal error: 'cmocka.h' file not found
```
There are several other failed checks later. I really do not know why the osx build looks like this, but it looks like common from fast search in google. So I propose to ignore this and fill separate issue, that can be handled by somebody who knows what the osx build does. If you agree, I will remove the temporary commits and fill a new issue.

frankmorgner:
Yes, in terms of CI we can leave this as is and ignore the error for now. Please clean up the commits and we can merge this PR.

Jakuje:
The patch set is cleaned, I landed some more issues that I noticed in the recent days that are related to the memory management and and that is preventing the get_challenge() from working in CAC cards.

I will open a separate issue for the mac build in CI, since it is failing to detect also other dependencies and hopefully somebody will be able to have a look into that.

PR #1391, opened 2018-06-12
gids: fix gids_delete_cert

ytoku:
It is failed to delete a certificate from GidsApplet by pkcs11-tool:
```
> pkcs11-tool -b --id 00 --type cert --pin 0000
error: PKCS11 function C_DestroyObject() failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
```

This is caused by a bug that gids_delete_cert reads MF with wrong offset.

This patch fixes the bug by correcting the offset.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
@vletoux?

vletoux:
Indeed I confirm this fix - there is one byte before a record array.
The pr does fix this bug

frankmorgner:
thanks

PR #1392, opened 2018-06-12
gids: use file id instead of path in gids_delete_key_file

ytoku:
It is failed to delete a private key from GidsApplet by pkcs11-tool:
```
> pkcs11-tool -b --id 00 --type privkey --pin 0000
error: PKCS11 function C_DestroyObject() failed: rv = CKR_DATA_INVALID (0x20)
Aborting.
```

This is caused by a protocol violation that in deleting a private key libopensc select a file by path
which GIDS does not support.

This patch fixes the bug by using file id instead of path.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-gids.c
+++ b/src/libopensc/card-gids.c
@@ -1477,9 +1477,9 @@ static int gids_delete_key_file(sc_card_t *card, int containernum) {
 	int r;
 	char ch_tmp[10];
 	sc_path_t cpath;
-	snprintf(ch_tmp, sizeof(ch_tmp), "IB0%02X",containernum + GIDS_FIRST_KEY_IDENTIFIER);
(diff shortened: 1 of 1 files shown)


frankmorgner:
@vletoux?

vletoux:
Indeed, I wrote this code more than 2 years ago.
Should the fix be more like https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-gids.c#L1136 ?

there is a workaround in gids_select_file to deal with the path selection which forward the call to the iso implementation for this case.

Fix is ok for me. I didn't know you can change the type of selection by appending a I

frankmorgner:
The suggested modification sounds like it is more consistent with the rest of the implementation. @ytoku could you check if that works as well?

ytoku:
Thanks.

How is this? I have rewritten the patch with gids_select_file.
Both versions work well in my enviroment.

frankmorgner:
thanks

vletoux:
Thanks

PR #1393, opened 2018-06-12
C_WrapKey and C_UnwrapKey implementations

hhonkanen:
Hi,

This PR contains implementation of C_UnwrapKey and C_WrapKey added into OpenSC. I started coding this in last summer, maybe some of you remember the discussion on the mailing list. Now, after a few cycles of testing and bug fixing, I consider it ready for review.

I will explain the code a little. 

C_UnwrapKey first creates an empty key object (a key file and the PKCS#15 directory entry) on the card using pkcs15init. Then it performs a crypto-operation on the card, where the incoming wrapped key is decrypted on card and placed into the previously created key file. The latter step should happen inside the card, so that plain text key data is never held in host's memory.

C_WrapKey first locates the wrapping key (A) and the key to be wrapped (B), and then does an encryption operation on the card. It outputs a cryptogram, which contains key B encrypted usign key A.

We have been testing the code using a MyEID card, which in the upcoming new version contains the following features:

-	Unwrap a secret key with an RSA key
-	Unwrap a secret key with a secret key 
-	Wrap a secret key with a secret key

These use cases have been fully implemented and tested. The code could be used and extended to perform different use cases and key types, for example unwrapping an RSA key using RSA key, or unwrapping an RSA key with a secret key. We have been testing symmetric operations only with AES keys.

The crypto part goes in the PKCS#11 based way, where the mechanism and its parameters define the details of the operation. We can now wrap and unwrap keys using mechanisms CKM_AES_ECB, CKM_AES_CBC and CKM_AES_CBC_PAD, and unwrap with CKM_RSA_PKCS. 

There is also a new feature: On card session keys. Earlier, OpenSC has considered all keys with CKA_TOKEN = FALSE to be keys that are handled only in hosts memory. However, PKCS#11 doesn't define where such keys should be stored, it only defines that they are temporary and must be removed when the PKCS#11 session ends. HSM devices I am familiar with, can handle temporary (session) keys in hardware this way. Now we had a commercial need for a similar "on card session key". So I have added a card capapility flag SC_CARD_CAP_ONCARD_SESSION_OBJECTS, with which a card can announce it supports them. With MyEID, we have implemented it so that a CKA_TOKEN=FALSE key is automatically cleaned from the card during the next reset. When creating an on card session key, no PKCS#15 SecretKeyDir or PrivateKeyDir entry is created, the key metadata is only held as an in memory structure.

To handle IV in symmetric crypto, I added type sc_sec_env_param to opensc.h. I added an array of these into sc_sec_env, so optional parameters like IV or target key file can be dynamically added when needed. 

While having to make changes and additions to core files like opensc.h, framework-pkcs15.c and pkcs15-sec.c, I have taken care to not break existing functionality, and to implement the changes as additions when possible. I have also made some changes to the symmetric key code in pkcs15-lib written by @fabled in 2017. I have discussed about these changes with the original author. 

After the key wrapping code is merged, it would be quite easy to add support for symmetric crypto, at least using AES. Most of the new code is reusable to implement C_Encrypt and to add symmetric keys to C_Decrypt.

We have tested the code with new MyEID cards, which have the wrapping features. If someone is interested in testing it with these cards, please contact me privately (first name.last name@aventra.fi). There is a little test program at my Github account: https://github.com/hhonkanen/WrapTest
The code of the test program is not clean and it needs playing around with comments and setting the test data in code, to run different tests. One thing to note about the parameters: While PKCS#11 spec doesn't require this, in this implementation caller must supply CKA_VALUE_LEN in C_UnwrapKey to define length of the new key to be created. This is because of the two phase implementation where the file is created by pkcs15init at first. Handling unwrapping without supplying key length would require some more work.

My patches are rebased on top of OpenSC master. I am leaving the office for a vacation on Wed 20th June and unfortunately have only limited time to answer questions and do requested changes before that. If some issues are left open and the code cannot be merged, I will continue the work on my part in mid-August, when I return. 

Any comments and development suggestions are welcome!

- Hannu

##### Checklist

- [ ] Documentation is added or updated (C_WrapKey and C_UnwrapKey is documented in PKCS#11 spec. Please tell me if more documentation is needed, and I will provide it.)
- [ ] Windows minidriver is tested (new features callable only from PKCS#11, should not affect minidriver)
- [ ] macOS tokend is tested (I don't have a test environment for this)


hhonkanen:
There was a problem with light-release CI builds where USE_PKCS15_INIT is not defined. Added commit d8a0611 to fix this.

Because development of this branch was spread over a long period of time, I ended up having quite many separate commits in this PR. Is the commit history too messy like this? I could combine this all in one single commit, however I didn't do it yet, because some of the commits, for example d7a1fde and 10fd8b0 are not only related to key wrapping and I thought that with specific commit messages they might be easier to understand.

frankmorgner:
Thanks for so much work. Your basic approach seems OK, but since I don't have so many use cases for this functionality, I'm hesitant on a review. Maybe @fabled or @dengert, who have been active in #1019, can comment on this. I think this feature may also be useful for @CardContact's key import for SC-HSM...

mouse07410:
This capability would be very useful. I'm not sure if I'm qualified to review this work, but I'd love to see it merged.

Would try to test it, but can't promise.

src/libopensc/pkcs15-sec.c
```diff
@@ -448,7 +448,7 @@ int sc_pkcs15_wrap(struct sc_pkcs15_card *p15card,
 		const struct sc_pkcs15_object *key,
 		struct sc_pkcs15_object *target_key,
 		unsigned long flags,
-		u8 * cryptogram, size_t* crgram_len,
+		u8 * cryptogram, unsigned long* crgram_len,
```
martinpaljak: This seems bogus change, size_t should be "semantically correct" here?

dengert:
In a multiple application situation, more than one application, for example FireFox, Thunderbird, some OS keyRing etc. may be accessing the card.  For example `disconnect=leave` in opensc-conf. 

The card does not have a good way to know which application is accessing the card at any one time. 
 
Can you go into more detail on what the card is expected to support?

Is SM required or recommended in order to limit access to on-card session keys to the creating 
application? 

You said in the comments that upon reset the card would clean up on card session keys. But does the card cleaned up upon power up? i.e. is there scenario where an attacker might gain access to on-card keys (deleted or not)  if the reset can be avoided?
 
On a multi-applet cards, the login state is usually lost when a different applet is selected.  On some cards even a SELECT AID of an AID not even supported by the card can lead to the lost of the login state. 
What happens to on-card session keys in these situations? 

If the on-card session keys are only cleaned up upon reset, are there any card memory management issues?  
    
Have you considered what changes to PKCS\#11 specs are need to support the concept of on-card-session key?  Would adding a vendor supported attribute help a calling application to make the choice of which to use?  `grep _VENDOR_ src/pkcs11/*` This shows `#define SC_VENDOR_DEFINED 0x4F534300  /* OSC */`,  `SC CKA_OPENSC_NON_REPUDIATION` and `CKA_SPKI`  You could add others if needed.

martinpaljak:
@dengert Assuming it is a JavaCard, a card reset and powerup are functionally the same (on powerup a reset happens). Also, clear on deselect is "lower than" clear on reset (or powerup). Depending on applet design, the login state can be kept either until deselect or until reset.

src/libopensc/pkcs15-sec.c
hhonkanen: If I remember correctly, the reason for this is that size of size_t varies on different architectures but where this function is called (in framework-pkcs15) this argument is casted from PKCS#11 type CK_ULONG* and CK_ULONG is defined to be at least 32 bits. So if for example size_t is 16 bits (could be on some embedded architectures) and CK_ULONG is 32 bits, this causes a compilation error. While I also thought that size_t* looks better I ended up putting it this way to avoid type conversions in framework-pkcs15.c and also to be consistent with sc_pkcs15_derive where *poutlen is an unsigned long pointer too.

hhonkanen:
Thank you all for your comments. 

@dengert: My view on multi-application scenarios and on limiting access to session keys, is that PKCS#11 specification leaves room for different kinds of implementations. From OpenSC's point of view, an on card session key is created inside a session created by one application, and if another application calls C_OpenSession to access the card, it doesn't see it as it isn't stored in the directory file and therefore it isn't parsed. MyEID currently doesn't have any mechanism to prevent another application from accessing it in APDU level. However on PC/SC level an application could always use SCARD_SHARE_EXCLUSIVE to prevent other applications accessing the card when it creates on card session keys.

I don't know how HSM's handle cleaning up the session keys. They should be cleaned up during C_CloseSession, but compared to a HSM, a card can be ejected from a reader at any time and the card will power off. MyEID card is a JavaCard and the key objects remain on the card after power off, so cleaning them up during the next reset seems like the only solution. But on some other card architectures session keys could be stored only in RAM in the first place and they would in this case vanish automatically when the card is powered off. To sum it up, session keys should be erased during C_CloseSession, but if the device is removed or powered off before this function is called, it's up to the device to handle this in its own way.

PKCS#11 provides attributes CKA_WRAP_WITH_TRUSTED and CKA_TRUSTED to limit which keys can be used to wrap other keys. This is to prevent someone from importing a new key to the card, wrapping another key and then unwrapping the key by software. How keys can be marked CKA_TRUSTED seems to be implementation specific. We haven't yet implemented these attributes in MyEID card/driver. Currently the main motive to implement key wrapping has been to prevent listening to an USB wire of a smart card reader and reading a secret key from there. A kind of alternative to SM.

> Have you considered what changes to PKCS#11 specs are need to support the concept of on-card-session key? Would adding a vendor supported attribute help a calling application to make the choice of which to use? grep _VENDOR_ src/pkcs11/* This shows #define SC_VENDOR_DEFINED 0x4F534300 /* OSC */, SC CKA_OPENSC_NON_REPUDIATION and CKA_SPKI You could add others if needed.

Haven't really thought about this yet. I understand it so that PKCS#11 doesn't define where CKA_TOKEN=FALSE keys should be stored. In HSM's it's normal that they are stored in the device, in other PKCS#11 implementations they can be stored in software. Do you mean that PKCS#11 users should be able to ask OpenSC is the connected device (card) able to store session keys in the device?

dengert:
Thanks for the responses. Good to see your are addressing some of these issues. 

The use of  SCARD_SHARE_EXCLUSIVE in effect defeats  multi-application scenarios if it is held too long. But the main problem with the iso7816 is there is no concept of a multi-user token, other than  maybe multiple PINS or using SM and tying access to session objects to a session. I believe Windows has session based pins which might be similar to session objects tied to SM sessions. 

One thing I was concerned about was could a hacker gain access (remove then misuse) a card to gain access to a session key. They might gain from looking at the code how the session keys are accessed or find them by trial and error, even if they are not addressed via pkcs15. If lost of power causes them to be erased that should work.   

I understand why you are assuming CKA_TOKEN=FALSE could allow for a session object to be stored on the token. Have you discussed  (or joined)  Oasis PKCS11  TC about this issue?

https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=pkcs11

I see at least one OpenSC active developer, @Jakuje, is on the committee, and recognize a few other names.  Maybe he has some insite.

Jakuje:
I am still mostly following on the things going on in the PKCS#11 WG, but I would be glad to help if there would be some specific queries. They can be also sent directly to the comments mailing list:

https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=pkcs11#feedback

hhonkanen:
Hi @dengert and @Jakuje 

I'm answering with some delay, because I'm still on vacation.

I haven't followed Oasis PKCS11 TC

The explanation of session objects and token objects in the PKCS#11 specification concentrates on defining the life cycle and protection/visibility of objects in both categories, but there is no mention about where the objects actually should be stored. 

Here is some discussion about this subject: https://security.stackexchange.com/questions/117624/pkcs11-session-object-security
This is how I understand it. A PKCS#11 module could be software only or control any kind of device (HSM, Smart card). If you create a session object on a HSM using PKCS#11, I think you could expect that the object will be stored on the HSM device.

PKCS#11 states that session objects should be only visible to the application that created them. On hardware level, I think we don't currently fully comply with this with MyEID. On PKCS#11 level it works like this, but one could access the session objects with direct apdus from another application, when the card is in the reader. But when you know this, you can restrict other applications with SCARD_SHARE_EXCLUSIVE, which can be perfectly fine in specific usage scenarios. I think that this kind of security issues, and the simultaneous use issues, are quite driver/device specific and should be described in the documentation of each device. But currently I have no ideas on how we could address these issues in generic OpenSC core. We will have to put some effort into this when we publish documentation of these features on MyEID.

One thing I'm thinking is that is should we do something to make session objects be deleted when C_CloseSession is called? Can we configure OpenSC so that it uses SCARD_RESET_CARD when closing the session and calling SCardDisconnect? I think this would be ok for the initial version, but maybe this could be improved by walking through each session object and deleting the explicitly on C_CloseSession.
This would, in MyEID case, ensure that the objects are deleted even if the card remains in the reader after the session.

mouse07410:
> But the main problem with the iso7816 is there is no concept of a multi-user token, other than maybe multiple PINS or using SM and tying access to session objects to a session.

Well, a "multi-app token" doesn't mean a "multi-**user**" token. Everything's still owned by one principal, even if he's running more than one application, and/or accessing more than one applet on the card.

> One thing I was concerned about was could a hacker gain access (remove then misuse) a card to gain access to a session key. 

@dengert I'm not sure I understand the use case (and the threat). Usually Wrap/Unwrap keys is used either for key provisioning, or for key backup/restore. Regardless, the key is protected the same way other keys stored on the card are protected: whether a user or an app can access a secret key does not depend on whether this key was "unwrapped" to this card, or provisioned by some other means.

Also, even if it's a "session" key - still, sharing among multiple apps means the user probably wouldn't like a session to die underneath him when he did something with another app: e.g., if I'm browsing a token-protected Web site, and in the middle of it decrypt just-received email or sign an outgoing email - I wouldn't like at all to have to re-start my browser session (possibly having lost the work done there, like filling a form or whatever)... 

Sure, a hacker could steal your session key if it's exportable from the card. Or he could steal your PIN and make the card work for him while it's inserted...

Am I missing something here?

> If loss of power causes them to be erased that should work.

If it's an "unwrapped" key, it's lifespan probably should be "until explicitly deleted", and it's exportability from the card in the clear should be "never". If it's a "session" key - then why should it stay on the card at all? The card decrypts ("unwraps" it) and passes to the caller, much like it would pass any decrypted value. I'd expect the card in this case to immediately forget this unwrapped key, and the protection of this key from hackers etc. now is the app's responsibility.


Another question: cards' main job usually is asymmetric cryptography. Sure, they can derive symmetric (secret) keys using ECDH, they can wrap/unwrap using symmetric keys, etc. But their "main" workhorse is a set of *asymmetric* keys. What surprises me is that you didn't mention capability to wrap/unwrap *asymmetric* keys (e.g., RSA private keys). Does your card support this?


Also, NIST standardized AES-CCM based key wrapping. Your code seems to support AES-ECB (which IMHO shouldn't be there), and AES-CBC only? Doesn't the card support CCM?


Also, what cards support this functionality? Does PIV support key wrap/unwrap?

hhonkanen:
> If it's an "unwrapped" key, it's lifespan probably should be "until explicitly deleted", and it's exportability from the card in the clear should be "never". If it's a "session" key - then why should it stay on the card at all? The card decrypts ("unwraps" it) and passes to the caller, much like it would pass any decrypted value. I'd expect the card in this case to immediately forget this unwrapped key, and the protection of this key from hackers etc. now is the app's responsibility.

In this case (on card session key), an unwrapped session key would not be passed to the caller. Such key will be kept on the card, and it can perform crypto operations ON CARD for the duration of the session. When the session ends, the key will be deleted. In a MyEID card it is deleted during the next reset, other implementations are possible

> Another question: cards' main job usually is asymmetric cryptography. Sure, they can derive symmetric (secret) keys using ECDH, they can wrap/unwrap using symmetric keys, etc. But their "main" workhorse is a set of asymmetric keys. What surprises me is that you didn't mention capability to wrap/unwrap asymmetric keys (e.g., RSA private keys). Does your card support this?
Also, NIST standardized AES-CCM based key wrapping. Your code seems to support AES-ECB (which IMHO shouldn't be there), and AES-CBC only? Doesn't the card support CCM?

Currently, MyEID does not support unwrapping or wrapping asymmetric keys. We are planning to implement this later. We need the ability to unwrap and wrap secret keys in a specific project, this was the reason to support secret keys first. 
However, an RSA key can be used to unwrap symmetric keys.

I agree that AES-ECB is probably useless and could as well be left out. We don't support ACS-CCM yet but jk

mouse07410:
> In this case (on card session key), an unwrapped session key would not be passed to the caller. Such key will be kept on the card, and it can perform crypto operations ON CARD for the duration of the session. When the session ends, the key will be deleted. ... We need the ability to unwrap and wrap secret keys in a specific project, this was the reason to support only secret keys first.

A naive question: what's the use case for wrapping/unwrapping *symmetric* keys *and keeping them on the card*? I could imagine unwrapping a symmetric (e.g., AES) key blob on the card, and releasing it to the main CPU to run bulk encryption (or keeping the symmetric session key inside a high-performance HSM that does handle bulk encryption and can terminate IPsec and/or TLS). I cannot imagine using the card itself for heavy encryption (for example, it cannot hold the TLS endpoint with anything like reasonable performance) - so what is this going to be good for?

> Currently, MyEID does not support unwrapping or wrapping asymmetric keys. We are planning to implement this later. 

I think you'd find this capability quite useful (once it is implemented, of course).

> However, an RSA key can be used to unwrap symmetric keys.

I think it's a very useful capability.

src/libopensc/pkcs15-sec.c
```diff
@@ -297,6 +352,208 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
+	senv.operation = SC_SEC_OPERATION_WRAP;
+
+	memset(&path, 0, sizeof (sc_path_t));
+	memset(&target_file_id, 0, sizeof (sc_path_t));
+
+	switch (target_key->type) {
+		case SC_PKCS15_TYPE_PRKEY_RSA:
+		tkey_path = target_prkey->path;
+		break;
+	default: /* we already know it is a secret key */
```
frankmorgner: bad indent. either here or above

src/libopensc/pkcs15-skey.c
```diff
@@ -140,13 +143,36 @@ sc_pkcs15_decode_skdf_entry(struct sc_pkcs15_card *p15card, struct sc_pkcs15_obj
...
+
+		for (i = 0; i < SC_MAX_SUPPORTED_ALGORITHMS && info.algo_refs[i] != 0; i++) {
+			for (ii = 0; ii < SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo != 0; ii++) {
+				if (info.algo_refs[i] == p15card->tokeninfo->supported_algos[ii].reference) {
+				    temp_oid = p15card->tokeninfo->supported_algos[ii].algo_id;
+				    temp_oid.value[8] = -1; /* strip off AES subtype octet*/
+
+				    if (sc_compare_oid(&id_aes, &temp_oid))
+					if (info.key_type == 0)	{
+					    info.key_type = CKK_AES;
```
frankmorgner: Is PKCS#15/ISO 7816-15 really mixing some context from PKCS#11 into this?

src/pkcs11/framework-pkcs15.c
```diff
@@ -131,6 +131,8 @@ struct pkcs15_skey_object {
 	struct sc_pkcs15_skey *valueXXXX;
 };
 
+#define is_secret_key(obj) ((__p15_type(obj) & SC_PKCS15_TYPE_CLASS_MASK) == SC_PKCS15_TYPE_SKEY)
```
frankmorgner: use better names to differentiate from `is_skey`

src/pkcs11/framework-pkcs15.c
```diff
@@ -3861,6 +3904,66 @@ pkcs15_prkey_sign(struct sc_pkcs11_session *session, void *obj,
...
+	struct	pkcs15_any_object *targetKeyObj = (struct pkcs15_any_object *) targetKey;
+	int	rv, flags = 0;	
+
+	sc_log(context, "Initiating unwrapping with private key.");
+
+	fw_data = (struct pkcs15_fw_data *) p11card->fws_data[session->slot->fw_data_idx];
+	if (!fw_data)
+		return sc_to_cryptoki_error(SC_ERROR_INTERNAL, "C_UnwrapKey");
+
+	sc_log(context, "unwrapping %p %p %p %p %lu %p", session, obj,
```
frankmorgner: how usefull are the pure pointers?

src/pkcs11/framework-pkcs15.c
```diff
@@ -4064,10 +4167,11 @@ struct sc_pkcs11_object_ops pkcs15_prkey_ops = {
 	pkcs15_any_destroy,
 	NULL,	/* get_size */
 	pkcs15_prkey_sign,
-	NULL,	/* unwrap */
+	pkcs15_prkey_unwrap,
 	pkcs15_prkey_decrypt,
         pkcs15_prkey_derive,
-        pkcs15_prkey_can_do
+        pkcs15_prkey_can_do,
```
frankmorgner: please fix the indenting here and one line up.

src/pkcs11/framework-pkcs15.c
```diff
@@ -4581,6 +4692,144 @@ pkcs15_skey_get_attribute(struct sc_pkcs11_session *session,
...
+	struct	pkcs15_skey_object *targetKeyObj = (struct pkcs15_skey_object *) targetKey;
+	int	rv, flags = 0;
+
+	sc_log(context, "Initiating unwrapping with a secret key.");
+
+	fw_data = (struct pkcs15_fw_data *) p11card->fws_data[session->slot->fw_data_idx];
+	if (!fw_data)
+		return sc_to_cryptoki_error(SC_ERROR_INTERNAL, "C_UnwrapKey");
+
+	sc_log(context, "unwrapping %p %p %p %p %lu %p", session, obj,
```
frankmorgner: how usefull are these pointers?

src/pkcs11/pkcs11-object.c
```diff
@@ -1026,7 +1028,71 @@ CK_RV C_WrapKey(CK_SESSION_HANDLE hSession,	/* the session's handle */
...
+		return rv;
+
+	/* Check if the wrapping key is OK to do wrapping */
+	rv = get_object_from_session(hSession, hWrappingKey, &session, &wrapping_object);
+	if (rv != CKR_OK) {
+		if (rv == CKR_OBJECT_HANDLE_INVALID)
+			rv = CKR_KEY_HANDLE_INVALID;
+		goto out;
+	}
+	if (wrapping_object->ops->wrap_key == NULL_PTR) {
```
frankmorgner: shouldn't this check moved to `sc_pkcs11_wrap_operation()`?

src/pkcs11/pkcs11-object.c
```diff
@@ -1038,7 +1104,67 @@ CK_RV C_UnwrapKey(CK_SESSION_HANDLE hSession,	/* the session's handle */
...
+		return rv;
+
+	rv = get_object_from_session(hSession, hUnwrappingKey, &session, &object);
+	if (rv != CKR_OK) {
+		if (rv == CKR_OBJECT_HANDLE_INVALID)
+			rv = CKR_KEY_HANDLE_INVALID;
+		goto out;
+	}
+	if (object->ops->unwrap_key == NULL_PTR) {
+		rv = CKR_KEY_TYPE_INCONSISTENT;
```
frankmorgner: shouldn't this check moved to sc_pkcs11_unwrap_operation()?

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1294,7 +1295,7 @@ sc_pkcs15init_init_skdf(struct sc_pkcs15_card *p15card, struct sc_profile *profi
 	struct sc_pkcs15_object *object = NULL;
 	const char	*label;
 	unsigned int	usage;
-	unsigned int	keybits = keyargs->value_len;
+	unsigned int	keybits = keyargs->value_len * 8;
```
frankmorgner: has this been incorrect before?

hhonkanen:
@frankmorgner thank you for review. I will comment each issue and make changes where needed. Adding the tests to p11test is a good idea. I need to work with the test tool anyway to make it possible to test all use cases easily in one run. So maybe I'll start the work by moving the code to p11test. This will make it much easier to maintain and test new OpenSC releases. There's some work to do, because now I have a separate script to init the card with keys used in wrapping. This should be implemented within p11test to allow running the tests along with all the existing ones.

Regarding harmonizing session pins with session objects, I am not familiar enough with the current implementation so I cannot say anything about this before taking a look at how session pins are implemented. I'll get back into this.

src/libopensc/card-myeid.c
```diff
@@ -650,12 +668,39 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
...
+			return SC_ERROR_WRONG_LENGTH;
+		}
+		*p++ = 0x83;
+		*p++ = 2;
+		memcpy(p, target_file->value, 2);
+		p+= 2;
+		break;
+	    }
+
+	if (env->operation ==  SC_SEC_OPERATION_UNWRAP || env->operation == SC_SEC_OPERATION_WRAP) /* & SC_SEC_ENV_IV_PRESENT*/
```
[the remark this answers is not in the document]
hhonkanen: It's useless now, I removed it. I implemented handling of IV in a bit different way at first, and forgot this comment here after changing it.

src/libopensc/opensc.h
```diff
@@ -178,8 +201,11 @@ typedef struct sc_security_env {
 	struct sc_path file_ref;
 	unsigned char key_ref[8];
 	size_t key_ref_len;
+	struct sc_path target_file_ref; /* target key file in unwrap operation */
```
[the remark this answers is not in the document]
hhonkanen: No, because we need file_ref to identify the file where the unwrapping key (which performs the crypto-operation) resides. File pointed by target_file_ref receives the result of the operation (the unwrapped key data).

src/pkcs11/framework-pkcs15.c
hhonkanen: ok.

src/pkcs11/framework-pkcs15.c
hhonkanen: is_skey() macro is not used anywhere in the whole code base and it wouldn't even compile, because SC_PKCS15_TYPE_SKEY_OBJECT is not defined. Seems like this is why I added a new macro in the first place, didn't dare to change/remove existing code without understanding it. Both macros are not needed. Should we remove is_skey or fix it to compare to SC_PKCS15_TYPE_SKEY instead of SC_PKCS15_TYPE_SKEY_OBJECT, which naming would you prefer?

src/pkcs11/pkcs11-object.c
hhonkanen: Please explain why would you prefer that? 

I think there's a reason to do this check already at this point. It's a simple way to notice that this key cannot be used for unwrapping at all and there is no point in going further and creating the target key object. In line 1146 sc_create_object_int(hSession, pTemplate, ulAttributeCount, phKey, 0); leads to a complex operation which creates the object to receive the unwrapped key material. The comment is a bit misleading, if the target object is CKA_TOKEN=TRUE it actually creates the object on card. This happens before sc_pcks11_unwrap_operation(). I think it's safer and more efficient to get out earlier.

To explain the logic a little more: 
if (object->ops->unwrap_key == NULL_PTR) checks whether this card and this key type support unwrapping at all, and in the next check, looking for unwrap_attribute, we check if we have permission to do that with the specific key object we are handling.

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1254,144 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+		* P2:  0x84  Cmd: AES/3DES Cryptogram (plain value encoded in BER-TLV DO, but not including SM DOs) */
+	sc_format_apdu(card, &apdu,
+		(crgram_len < 256) ? SC_APDU_CASE_4_SHORT : SC_APDU_CASE_3_SHORT,
+		0x2A, 0x00, p2);
+
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+	apdu.le = crgram_len;
+
+	if (crgram_len == 256 && p2 == 0x86)
```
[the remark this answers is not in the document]
hhonkanen: I'll do some refactoring for this function. The code has originally been copied from myeid_decipher, where 256 byte (2048 bit RSA) input data is split into two APDUs, and we set LE only for the second APDU, because we don't expect return data from the first one. 

When we are doing unwrapping, actually we don't return any data from the card, as the decrypted data is placed into a file on card, so we need only SC_APDU_CASE_3_SHORT, it seems. 

I removed some other useless code from this function too. I'll have to test this first, and commit the changes then.

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1254,144 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+			LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+
+			if (apdu.sw1 == 0x90 && apdu.sw2 == 0x00)
+			{
+				LOG_FUNC_RETURN(card->ctx, r);
+			}
+		}
+		else
+		{
+			int len = apdu.resplen > outlen ? outlen : apdu.resplen;
```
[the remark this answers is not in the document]
hhonkanen: After all we don't need to copy data to out buffer in this function. But the idea has been to return the required buffer length to caller, so that caller can allocate long enough buffer. This convention seems to be used at least in decipher. I looked at other drivers too and they work like this. On the other hand, compute_signature returns SC_ERROR_BUFFER_TOO_SMALL if the buffer cannot hold the data returned from card. Do you know why there are different conventions in these two functions? 

I am returning required buffer size also in myeid_wrap_key. 

PKCS#11 defines returning buffer length in section 11.2: Conventions for functions returning output in a variable-length buffer
How does it work with OpenSC if you call C_Sign with pSignature = NULL?

frankmorgner:
It should be possible to use the `sc_pkcs15init_generate_secret_key()` in the minidriver to generate a session PIN (Windows allows binary data). However, I wonder how you're signalling the key from the card back to the PKCS#11 application...

src/libopensc/pkcs15-sec.c
```diff
@@ -297,6 +352,208 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
+		path = tkey->path;
+		memcpy(target_file_id.value, tkey->path.value + tkey->path.len - 2, 2);
+		target_file_id.len = 2;
+		target_file_id.type = SC_PATH_TYPE_FILE_ID;
+	}
+	else {
+		LOG_TEST_RET(ctx, SC_ERROR_INVALID_ARGUMENTS, "invalid unwrapping target key path");
+	}
+
+	senv_param = (sc_sec_env_param_t) { SC_SEC_ENV_PARAM_TARGET_FILE, (u8*) &target_file_id, sizeof(target_file_id)};
```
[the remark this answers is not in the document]
hhonkanen: If I remove it, I get a compilation error: 

pkcs15-sec.c:425:15: error: expected expression before ‘{’ token
  senv_param = { SC_SEC_ENV_PARAM_TARGET_FILE, (u8*) &target_file_id, sizeof(tar

or do you mean the second cast to u8*
?

src/libopensc/pkcs15-skey.c
hhonkanen: No. PKCS#15 has OIDs of supported algorithms listed in tokenInfo. Here we check, if the algorithm referenced in the PKCS#15 SecretKeyObject matches with AES algorithm's OID and set the key type to CKK_AES for OpenSC's internal use, out of PKCS#15's context.

I did ponder for a while, should I include the pkcs#11 header here or define new algorithm types in the PKCS#15 headers. I ended up doing it this way, because CKK_ key type definitions are already used in skey_pkcs15_skey_info before my pull request. Please take a look at line 2315 of framework-pkcs15.c in current master, function pkcs15_create_secret_key: 
	    skey_info->key_type = key_type; /* PKCS#11 CKK_* */

I decided to use the CKK_AES defintion in pkcs15-skey to be consistent with this. At least we have to use the same set of definitions and I don't know how much code it would affect to change it now.

src/libopensc/card-myeid.c
```diff
@@ -650,12 +668,39 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
...
+	    }
+
+	if (env->operation ==  SC_SEC_OPERATION_UNWRAP || env->operation == SC_SEC_OPERATION_WRAP) /* & SC_SEC_ENV_IV_PRESENT*/
+	{
+	    /* add IV if present */
+		for (i = 0; i < SC_SEC_ENV_MAX_PARAMS; i++)
+			if (env->params[i].param_type == SC_SEC_ENV_PARAM_IV) {
+				*p++ = 0x87;
+				*p++ = (unsigned char) env->params[i].value_len;
+				memcpy(p, env->params[i].value, env->params[i].value_len);
```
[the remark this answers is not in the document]
hhonkanen: added length check in 1105da2b4663b36dd6674ca62e6c245a5c418d2f

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1254,144 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+	struct sc_apdu apdu;
+	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
+	u8 sbuf[SC_MAX_APDU_BUFFER_SIZE];
+	u8 p2 = 0x86;
+	myeid_private_data_t* priv;
+
+	LOG_FUNC_CALLED(card->ctx);
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_NORMAL);
+
+	assert(card != NULL && crgram != NULL && out != NULL);
```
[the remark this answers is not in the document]
hhonkanen: Changed to error check in 1105da2b4663b36dd6674ca62e6c245a5c418d2f. Returns SC_ERROR_INVALID_ARGUMENTS if null args passed.

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1254,144 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+						p2 = 0x84;		/* Set correct P2 for symmetric crypto */
+		}
+	}
+
+	/* INS: 0x2A  PERFORM SECURITY OPERATION
+	    * P1:  0x00  Do not expect response - the deciphered data will be placed into the target key EF.
+	    * P2:  0x86  Cmd: Padding indicator byte followed by cryptogram
+		* P2:  0x84  Cmd: AES/3DES Cryptogram (plain value encoded in BER-TLV DO, but not including SM DOs) */
+	sc_format_apdu(card, &apdu,
+		(crgram_len < 256) ? SC_APDU_CASE_4_SHORT : SC_APDU_CASE_3_SHORT,
```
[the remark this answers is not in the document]
hhonkanen: Rewrote some of the code to make it cleaner. Removed pointless out buffer, because the data is unwrapped into an object on card and no data is returned.

src/libopensc/pkcs15-sec.c
```diff
@@ -32,39 +32,80 @@
...
+
+		return SC_SUCCESS;
+	    }
+	}
+
+	return SC_ERROR_TOO_MANY_OBJECTS;
+}
+
+
+static sc_path_t get_file_path(const struct sc_pkcs15_object* obj)
```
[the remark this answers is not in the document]
hhonkanen: I agree, it's better to return an error code. Updated in 1105da2b4663b36dd6674ca62e6c245a5c418d2f

src/pkcs11/framework-pkcs15.c
hhonkanen: Removed the line, but added some logging in case the function fails because of a NULL argument.

src/pkcs15init/pkcs15-lib.c
hhonkanen: The issue is that CKA_VALUE_LEN that comes in a PKCS#11 template means number of bytes, but in a PKCS#15 SecretKeyObject, the key length should be set in bits.

After looking at pkcs15-init tool and framework-pkcs15, I changed it so that the multiplication to bits is done in framework-pkcs15, before passing keyargs to pkcs15-lib, and removed the multiplication from pkcs15-lib.

pkcs15-init tool passes length in bits, and the functionality to store secret keys with the tool is already in the release and may be used by somebody. So I think it's best to be consistent with that.

I also changed line 2356 of framework-pkcs15.c to multiply key length by 8, where software-only secret key objects are created. At least I couldn't find any side effects this could cause.  This caused the confusion in the first place: framework-pkcs15 was setting sc_pkcs15_skey_info.value_len in bytes and pkcs15-lib was setting the same member of the same struct in bits.

mouse07410:
@please check the PKCS#11 spec: usually key wrapping and unwrapping is communicated to *outside* the card - because very few cards have the capacity to run bulk crypto with the unwrapped key, but it makes perfect sense to wrap key securely using hardware tokens. The only context for in-card keeping of the unwrapped key is backup-restore, and again, the only tokens I know that do it are HSM devices.

I would like to see this extended to returning the unwrapped key in the session context to the PKCS#11 caller, if the reference is provided.

hhonkanen:
@mouse07410 Please point out where in PKCS#11 specification it is stated that key wrapping and unwrapping is usually meant to be communicated outside of the hardware token? I have understood it so, that PKCS#11 in itself doens't define what happens in software and what happens in a separate physical device. There are pure software PKCS#11 implementations and there are HSM's. But if you have a physical device, to my underastanding if you call C_UnwrapKey the primary intention would be to have the key unwrapped into the device. At least that's the way HSM's I am familiar with work. You can for example define CKA_EXTRACTABLE=FALSE to the unwrapped key's template and isn't it meant to ensure that the key cannot be extracted from the token again? 

If you wanted to get the unwrapped result out of the card, or wrap a key that you have in software, why wouldn't you just use C_Encrypt and C_Decrypt to achieve the same thing? 

The motivation to implement wrap and unwrap commands into MyEID card has been to be able to guarantee the users of the card, that when using these commands the key material is never handled as plaintext in a workstation's memory. The unwrapped keys can be used to do crypto inside the card. One use case where this could be used is to for example wrap an encryption key in a HSM, unwrap it into a card and use the card for encryption. 

I considered defining a return buffer to card_unwrap_key, to be able to return the unwrapped data if a card doesn't support unwrapping to a file inside the card. I decided to not do this, to make a clear distinction to operations which return plain text data. If we would allow returning plain data, how can users of OpenSC PKCS#11 be sure, are their keys handled in plaintext or not?

src/pkcs11/framework-pkcs15.c
frankmorgner: I'd use `is_skey` and `is_skey_object`, i.e. I'd clearify the existing define.

src/pkcs11/mechanism.c
```diff
@@ -830,6 +830,106 @@ sc_pkcs11_decr(struct sc_pkcs11_session *session,
...
+	CK_BYTE_PTR wrappedData,
+	CK_ULONG_PTR wrappedDataLen)
+{
+	struct sc_pkcs11_card *p11card;
+	sc_pkcs11_operation_t *operation;
+	sc_pkcs11_mechanism_type_t *mt;
+	CK_RV rv;
+
+	if (!session || !session->slot
+	 || !(p11card = session->slot->p11card))
```
frankmorgner: bad indent

src/pkcs11/mechanism.c
```diff
@@ -830,6 +830,106 @@ sc_pkcs11_decr(struct sc_pkcs11_session *session,
...
+	struct sc_pkcs11_object *targetKey)
+{
+	struct sc_pkcs11_card *p11card;
+	sc_pkcs11_operation_t *operation;
+	sc_pkcs11_mechanism_type_t *mt;
+
+	CK_RV rv;
+
+	if (!session || !session->slot
+	 || !(p11card = session->slot->p11card))
```
frankmorgner: bad indent

src/pkcs11/pkcs11-object.c
```diff
@@ -1037,7 +1039,71 @@ CK_RV C_WrapKey(CK_SESSION_HANDLE hSession,	/* the session's handle */
 		CK_BYTE_PTR pWrappedKey,	/* receives the wrapped key */
 		CK_ULONG_PTR pulWrappedKeyLen)
 {				/* receives byte size of wrapped key */
-	return CKR_FUNCTION_NOT_SUPPORTED;
+	CK_RV rv;
+	CK_BBOOL can_wrap,
+		 can_be_wrapped;
```
frankmorgner: bad indent?

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1344,9 +1348,15 @@ sc_pkcs15init_init_skdf(struct sc_pkcs15_card *p15card, struct sc_profile *profi
 
 	if (keyargs->access_flags & SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE) {
 		key_info->access_flags &= ~SC_PKCS15_PRKEY_ACCESS_NEVEREXTRACTABLE;
-		key_info->native = 0;
+		/* The following commented out. I interpret PKCS#15 so that native only means that a key
+		 * can be used in on card crypto. Such key can be extractable (can be wrapped), so IMO this
+		 logic is not correct. */
+		/* key_info->native = 0; */
```
frankmorgner: @fabled added this code initially, maybe he can comment on this.

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1880,8 +1890,9 @@ sc_pkcs15init_store_secret_key(struct sc_pkcs15_card *p15card, struct sc_profile
 	if (check_key_compatibility(p15card, keyargs->algorithm, NULL, 0, keyargs->key.data_len * 8, 0)) {
 		/* Make sure the caller explicitly tells us to store
 		 * the key as extractable. */
-		if (!(keyargs->access_flags & SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE))
-			LOG_TEST_RET(ctx, SC_ERROR_INCOMPATIBLE_KEY, "Card does not support this key.");
+		/* Commented out. I don't understand why one couldn't store a key as non extractable. -HH */
+/*		if (!(keyargs->access_flags & SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE))
+			LOG_TEST_RET(ctx, SC_ERROR_INCOMPATIBLE_KEY, "Card does not support this key.");*/
```
frankmorgner: @fabled, could you comment?

fabled: This part is copied from sc_pkcs15init_store_private_key(). I think the purpose is that if someone wants to store a non-supported key on the token it becomes unusable unless it can be extracted. Since wrapping support is added, it would be better to check if the key can be extracted or wrapped. If changed, same change should be done in the function where this is copied from.

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1329,13 +1330,16 @@ sc_pkcs15init_init_skdf(struct sc_pkcs15_card *p15card, struct sc_profile *profi
 	key_info->key_reference = 0;
 	switch (keyargs->algorithm) {
 	case SC_ALGORITHM_DES:
-		key_info->key_type = CKM_DES_ECB;
+		key_info->key_type = CKK_DES;
```
fabled: Why is this changed? This value goes directly to PKCS#15 CardInfo.supportedAlgorithms. The specification I have says it should be chosen from and interpreted as **mechanism numbers in PKCS #11**. Please revert this change or give reference to standard where this changed.

hhonkanen: I changed it because it seemed a bit strange that the mechanism is set to ECB here when the keys can be used with CKM_AES_CBC mechanism as well. I though it would be better to use the generic CKK definitions to define the key type instead of the mechanism. I have taken care of this also in function myeid_fixup_supported_algos in commit f722646. There might have been some other places as well, where it seemed cleaner to use key type defintions rather than mechanisms. 

Do you think it's ok with f722646 or would it be better to go back to CKM_AES_ECB defs everywhere? Am I breaking up something else with this change?

src/pkcs15init/pkcs15-lib.c
hhonkanen: I may have misunderstood return value of check_key_compatibility, thinking zero means that the key is not compatible and nonzero would mean TRUE. But it actually returns SC_SUCCESS (0) if the card can handle the key. Would it be better to change this to 
if (check_key_compatibility(...) == SC_SUCCESS) to avoid misunderstandings?

I understand the logic better now. I have to think about this a little. One thing to consider is that a card could store a CKK_GENERIC_SECRET key, but they are not used to do crypto. So there's not point in storing such key with both CKA_SENSITIVE=CK_TRUE and CKA_EXTRACTABLE=CK_FALSE.

src/pkcs15init/pkcs15-lib.c
fabled: While inside OpenSC this change would work. The point is that this value gets written to token, and PKCS15 defines which type to use. So this breaks interop and standard compliance.

The structure is supported to announce which algorithms (mechanisms) can be used with the key. If we want to do things better, we should list all mechanisms supported with the key, and fill-in cross-reference to each of these supportedAlgorithms entries.

For simplicity and other implementation reasons I added only the ECB mechanism. But if improvement is needed, there should way to add all mechanisms the card supports and that can be used with the key in question.

hhonkanen: Please can you show me where this value actually gets written to token, because I don't find it? I understand that we have to use the CKM_ values in tokenInfo as defined in PKCS#15, 
and my intention has been to do so.

My idea is that it would be better to use the CKK_AES and CKK_DES3 definitons for key type internally, because when we add functionality to do crypto operations 
with the keys, there are many situations where we just want to check whether the key is AES or DES3, and it looks wrong to compare it to a mechanism type. I would then separately list the mechanisms associated with the key type into tokenInfo->supported_algos, using the CKM_ constants.

But isn't this just what we are already doing in myeid_fixup_supported_algos?

case SC_PKCS15_TYPE_SKEY_GENERIC:
	switch (skey_info->key_type | (skey_info->value_len << 16)) {
		case CKK_AES | (128 << 16):
			_add_supported_algo(profile, p15card, object, SC_PKCS15_ALGO_OP_DECIPHER|SC_PKCS15_ALGO_OP_ENCIPHER, CKM_AES_ECB, &id_aes128_ecb);
			_add_supported_algo(profile, p15card, object, SC_PKCS15_ALGO_OP_DECIPHER|SC_PKCS15_ALGO_OP_ENCIPHER, CKM_AES_CBC, &id_aes128_cbc);
			break;
		case CKK_AES | (256 << 16):
			_add_supported_algo(profile, p15card, object, SC_PKCS15_ALGO_OP_DECIPHER|SC_PKCS15_ALGO_OP_ENCIPHER, CKM_AES_ECB, &id_aes256_ecb);
			_add_supported_algo(profile, p15card, object, SC_PKCS15_ALGO_OP_DECIPHER|SC_PKCS15_ALGO_OP_ENCIPHER, CKM_AES_CBC, &id_aes256_cbc);
			break;
		}
		break;

We do use the CKM definitions here and they go into tokenInfo in sc_pkcs15_encode_tokeninfo(). This is actually your code. Shall we let each driver handle this in their own way, because cards might not support all mechanisms that are associated with specific key type? Or should we have some more generic way that would be used by default?

I verified by testing that this also works as intended. The algoritmInfo.algorithm values written to tokenInfo are the PKCS#11 mechanisms (4225 and 4226), not what we are setting in skey_info.key_type

I would prefer to not go back, but keep using the CKK_ values for key_type internally, and ensure that they are correctly "converted" to a list of CKM_ values when writing tokenInfo (and other way when reading tokenInfo). Do you agree with this? 

In skey_info we already have unsigned int algo_refs[SC_MAX_SUPPORTED_ALGORITHMS]. The value in this array reference the AlgorithmInfo entries in tokenInfo. Isn't this the cross-reference you are talking about?

fabled: Sorry. I remembered that part incorrectly. You are right.

My first impression would have been then to just suggest putting some non-pkcs11 value there. But further digging shows that pkcs11/framework-pkcs15.c also stashes CKK_* values in the same field (though, it's supported only for non-session objects for the time being). So yeah, it actually is good to do what you are doing.

Sorry for the noise. If you ever get to rebasing things, please add the above comment in the commit message.

src/libopensc/pkcs15-skey.c
```diff
@@ -140,13 +143,36 @@ sc_pkcs15_decode_skdf_entry(struct sc_pkcs15_card *p15card, struct sc_pkcs15_obj
...
 	if (r == SC_ERROR_ASN1_END_OF_CONTENTS)
 		return r;
 	LOG_TEST_RET(ctx, r, "ASN.1 decoding failed");
-	if (asn1_skey_choice[0].flags & SC_ASN1_PRESENT)
+	if (asn1_skey_choice[0].flags & SC_ASN1_PRESENT) {
 		obj->type = SC_PKCS15_TYPE_SKEY_GENERIC;
+
+		/* info.key_type = CKK_AES;*/	/* defaults to CKK_AES. TODO: implement choice 15 and GenericKeyAttributes like defined in ISO7816-15 and set key type accordingly. */
+
+	/* Check key type. framework-pkcs15 recognizes one type per key, and AES is the only algorithm supported for
```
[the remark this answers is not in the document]
hhonkanen: The upper comments seems pointless now, so I removed it.

src/pkcs15init/pkcs15-lib.c
hhonkanen: No problem and thank you for checking this.

I would like to add AlgorthmInfos for AES algorithm into tokenInfo already when the card is initialized. SO-PIN is required to update tokenInfo and this is not very convenient when calling C_WrapKey and C_UnwrapKey in PKCS#11. This probably should be a card specific choice, handled in pkcs15-init overrides. But this is out of the context of this PR.

src/pkcs11/framework-pkcs15.c
```diff
@@ -4659,9 +4767,8 @@ pkcs15_skey_get_attribute(struct sc_pkcs11_session *session,
 		*(CK_OBJECT_CLASS*)attr->pValue = CKO_SECRET_KEY;
 		break;
 	case CKA_TOKEN:
-		/*TODO DEE change if on card skeys are supported */
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
-		*(CK_BBOOL*)attr->pValue = FALSE;
+		*(CK_BBOOL*)attr->pValue = skey->base.p15_object->session_object == 0;
```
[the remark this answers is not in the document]
hhonkanen: I changed this one and the other cases of the same switch clause in pkcs15_skey_get_attribute (not previously touched by me) like you suggest. It's better to use CK_TRUE than TRUE I think, as we're treating *(attr->pValue) as CK_BBOOL. 

This same pattern is used also in other functions of framework-pkcs15, but let's not mix that with this PR, ok? The code is not consistent now.

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1261,149 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+			apdu.datalen = apdu.lc;
+			apdu.data = sbuf;
+
+			r = sc_transmit_apdu(card, &apdu);
+
+			LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+		}
+	}
+	LOG_FUNC_RETURN(card->ctx, sc_check_sw(card, apdu.sw1, apdu.sw2));
+}
```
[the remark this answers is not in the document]
hhonkanen: In line 1312 we validate crgram_len with this check, which is supposed to prevent buffer overrun in the later memcpys :

	if (crgram_len > MYEID_MAX_RSA_KEY_LEN / 8)

MYEID_MAX_RSA_KEY_LEN / 8 = 256.

For symmetric keys we have a separate check.

This already prevents buffer overrun in the later memcpys. When we increase MYEID_MAX_RSA_KEY_LEN, we have to work with all functions that do RSA crypto anyway, to support longer key sizes. So I think there's not much risk that we would change this constant and forget to update the argument validation at that point.

Well, actually I can find one situation which could go wrong… If crgram_len would be 255, there's a risk of buffer overrun here:

		sbuf[0] = 0; /* padding indicator byte, 0x00 = No further indication */
			memcpy(sbuf + 1, crgram, crgram_len);

By changing:

if (crgram_len > MYEID_MAX_APDU_DATA_LEN)
to
if (crgram_len >= MYEID_MAX_APDU_DATA_LEN)

are we on the safe side?

Can you find some other specific situation which could cause buffer overrun?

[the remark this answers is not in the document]
hhonkanen: After all even the issue above wouldn't cause a buffer overrun. sbuf is SC_MAX_APDU_BUFFER_SIZE long which is 261 bytes so there is a margin already. But is there something else?

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1261,149 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+	apdu.le = 0;
+	apdu.lc = 0;
+
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(ctx, r, "APDU transmit failed");
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	LOG_TEST_RET(ctx, r, "wrap key failed");
+
+	if (apdu.resplen <= outlen && out != NULL)
+		memcpy(out, apdu.resp, apdu.resplen);
```
frankmorgner: Should we return an error if the buffer is too small?

src/libopensc/card-myeid.c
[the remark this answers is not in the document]
frankmorgner: You're right it wouldn't cause any problems, currently. But it would be more safe and more obvious if you're checking the length when it's actually used. Hence, I would not compare against `MYEID_MAX_APDU_DATA_LEN` and instead compare `crgram_len` with `sizeof sbuf`...

src/libopensc/pkcs15-sec.c
```diff
@@ -297,6 +355,208 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
+	senv_param = (sc_sec_env_param_t) { SC_SEC_ENV_PARAM_TARGET_FILE, &target_file_id, sizeof(target_file_id)};
+	LOG_TEST_RET(ctx, sec_env_add_param(&senv, &senv_param), "failed to add target file path to security environment");
+
+	r = sc_get_encoding_flags(ctx, flags, alg_info->flags, &pad_flags, &sec_flags);
+	LOG_TEST_RET(ctx, r, "cannot encode security operation flags");
+	senv.algorithm_flags = sec_flags;
+
+	if ((sec_flags & (SC_ALGORITHM_AES_CBC | SC_ALGORITHM_AES_CBC_PAD)) > 0) {
+	    senv_param = (sc_sec_env_param_t) { SC_SEC_ENV_PARAM_IV, (void*) param, paramlen };
+	    LOG_TEST_RET(ctx, sec_env_add_param(&senv, &senv_param), "failed to add IV to security environment");
```
frankmorgner: bad indent

src/libopensc/pkcs15-sec.c
```diff
@@ -297,6 +355,208 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
+			*poutlen);
+
+	if (r > -1) {
+            if (*crgram_len < (unsigned) r) {
+					*poutlen = r;
+					if (out != NULL) /* if NULL, return success and required buffer length by PKCS#11 convention */
+						LOG_TEST_RET(ctx, SC_ERROR_BUFFER_TOO_SMALL, "Buffer too small to hold the wrapped key.");
+			}
+
+            *poutlen = r;
```
frankmorgner: bad indent

src/libopensc/pkcs15-sec.c
```diff
@@ -297,6 +355,208 @@ int sc_pkcs15_derive(struct sc_pkcs15_card *p15card,
...
+	    LOG_TEST_RET(ctx, sec_env_add_param(&senv, &senv_param), "failed to add IV to security environment");
+	}
+
+	out = cryptogram;
+	poutlen = crgram_len;
+	r = use_key(p15card, key, &senv, sc_wrap, in, inlen, out,
+			*poutlen);
+
+	if (r > -1) {
+            if (*crgram_len < (unsigned) r) {
```
frankmorgner: bad indent

src/libopensc/pkcs15-skey.c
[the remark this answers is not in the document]
frankmorgner: The indent of the comment is wrong.

src/pkcs11/framework-pkcs15.c
```diff
@@ -2296,41 +2333,40 @@ pkcs15_create_secret_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
...
 	    skey_info->usage = args.usage;
 	    skey_info->native = 0; /* card can not use this */
 	    skey_info->access_flags = 0; /* looks like not needed */
 	    skey_info->key_type = key_type; /* PKCS#11 CKK_* */
 	    skey_info->data.value = args.key.data;
 	    skey_info->data.len = args.key.data_len;
-	    skey_info->value_len = args.value_len; /* callers preferred length */
+	    skey_info->value_len = args.value_len * 8; /* key length comes in number of bytes, use length in bits in PKCS#15. */
 	    args.key.data = NULL;
+		key_obj->session_object = 1;
```
frankmorgner: The whole block has a bad indent. feel free to fix this.

src/libopensc/card-myeid.c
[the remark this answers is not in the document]
dengert: many card drivers overuse the SC_MAX_APDU_BUFFER_SIZE and get away with it. 
it is defined in types.h as:
```
#define SC_MAX_APDU_BUFFER_SIZE   261 /* takes account of: CLA INS P1 P2 Lc [255 byte of data] Le */
```
It defines the max size of a short APDU with data to be sent to the reader in a single buffer.

The code is also being written with the max RSA key size of 2048. 
May I suggest that you make the code more general in order to handle large RSA keys when they come out with a new card that supports a 4096 byte RSA key?

In other words base you buffer sizes of data to be sent or received on MYEID_MAX_RSA_KEY_LEN Then only deal with having to send it in chucks with the extra leading flag byte within a loop rather then the if at 1361. 

You divide the data in half to send in two equal size chunks. Is that a card restriction? Most drivers send as much as possible in an APDU (card->max_send_size), then the rest in the final APDU. This then can work with readers that may support max_send_size < 255.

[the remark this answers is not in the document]
hhonkanen: Currently we only need to split the APDU in two if the key size is 2048 bits so the logic is fixed as it is. Our implementation is to split it in half. We are going to release a new MyEID version with 4096 bit keys hopefully in the near future. We are already working with it. 

When we proceed into implementing support for 4096 bit keys into OpenSC we have to rewrite these functions (unwrap, deciper, compute_signature) anyway. It's better to use a loop then and make it more generic. As long as we can agree that the function is safe as it is, I'd prefer to not put effort in refactoring it now to avoid double work on the same issue. When we have the 4K keys, I'd rather improve the code in all of these functions at the same time and create a separate PR for this. 

I hope we could complete merging this PR asap because some of the code is already quite old and rebasing it is time consuming if same parts of code will be changed again in the main branch. I'll provide some more test results soon.

src/pkcs11/framework-pkcs15.c
frankmorgner: OK, then please make a seperate PR that replaces 4 spaces with 1 tab

src/pkcs15init/pkcs15-lib.c
hhonkanen: After all I think it is best to keep this check, so I uncommented it. It doesn't make sense to store a key which cannot be used for crypto and is not extractable. I made the check more explicit like this: if (check_key_compatibility(...) != SC_SUCCESS). I did the same modification to other functions that call this function too.

src/libopensc/card-myeid.c
hhonkanen: I copy and modify my answer from an earlier conversation:

The idea has been, that in case where all goes well but only the buffer is too small, to return the required buffer length to caller, so that caller can allocate a long enough buffer. This convention seems to be used at least in decipher. I looked at other drivers too and they work like this. On the other hand, compute_signature returns SC_ERROR_BUFFER_TOO_SMALL if the buffer cannot hold the data returned from card. Do you know why there are different conventions in these two functions?

PKCS#11 defines returning buffer length in section 11.2: Conventions for functions returning output in a variable-length buffer
How does it work with OpenSC if you call C_Sign with pSignature = NULL?

In the wrapkey case, insufficient buffer is handled in upper level and accurate required buffer size is returned to PKCS#11 level according to the PKCS#11 convention. If the return buffer is not NULL, caller finally gets CKR_BUFFER_TOO_SMALL.

hhonkanen:
I have updated this PR according to the latest review and discussion. I have fixed indentation in most places, but would prefer to handle the larger chunks of code which have mixed tabs and spaces in a separate PR. A new conflict with padding.c has appeared due to 8c535c1 and should be resolved. 

I did more complete testing today. Here are the results which include both the old functionality and wrapping. Here are the results:

1. Initialize card
pkcs15-init -P -a 1 -l "Basic PIN" --pin 1111 --puk 12345678 --so-pin 12345678
pkcs15-init -P -a 2 -l "Sign PIN" --pin 2222 --puk 12345678 --so-pin 12345678
pkcs15-init -F

Result: OK

2. Generate 1024 bit RSA key
./pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --login --pin 1111 --keypairgen --key-type rsa:1024 --id 10 -a "rsa1024"
Result: OK
3. Generate 2048 bit RSA key
./pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so --login --pin 1111 --keypairgen --key-type rsa:2048 --id 11 -a "rsa2048"
Result: OK

4. ** Test signing with the RSA keys
pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -lt
Result: Signing with 1024 bit and 2048 bit keys OK. C_UnwrapKey fails, but the new unwrapping functionality is not included in pkcs11-tool yet so it is not handled correctly.

New test with MyEID 4.0.1 which doesn't support wrap/unwrap
Result: RSA sign and verify tests ok, decrypt with RSA-X-509 and RSA-PKCS: OK, Error at AES-ECB
Test output:
--------------------------------------------------------------------------------
hannu@ubuntu:~/opensc_tests$ pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -lt
Logging in to "Basic PIN (MyEID)".
Please enter User PIN: 
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (rsa1024) 
  testing signature mechanisms:
    RSA-X-509: OK
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
  testing key 1 (2048 bits, label=rsa2048) with 1 signature mechanism
    RSA-X-509: OK
  testing key 0 (rsa1024)
    RSA-X-509: OK
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
  testing key 1 (rsa2048) with 1 mechanism
    RSA-X-509: OK
Unwrap: not implemented
  testing key 0 (rsa1024) 
    RSA-X-509: OK
    RSA-PKCS: OK
error: PKCS11 function C_DecryptInit failed: rv = CKR_KEY_TYPE_INCONSISTENT (0x63)
Aborting.
    AES-ECB: 
--------------------------------------------------------------------------------
We should check if this happens also with MyEID with OpenSC 0.19.0. I suppose this happens, because MyEID lists AES-ECB mechanism as supported, but the test in pkcs11-tool cannot handle it.

4.1 Test 2048 bit keys with separate commands
/usr/local/bin/pkcs15-crypt -s -k 10 --pkcs1 --sha-256 -i /home/demo/opensc_tests/sha256.bin -o /home/demo/opensc_tests/myeid-sha256-signed.bin --pin "1111"
Result: OK

5. Generate 256 bit EC key

pkcs11-tool --module /usr/local/lib/opensc-pkcs11.so -l -k --key-type EC:prime256v1 -d 12 -a "ec256"
Result: OK 

6. sign data with the EC key

	* Prereq: create a file named sha256.bin containing a sha256 hash as binary data
/usr/local/bin/pkcs15-crypt -s -k 12 -i /home/hannu/opensc_tests/sha256.bin -o /home/hannu/opensc_tests/myeid-sha256.bin --pin "1111"
Result: OK

// RSA key import test

9. openssl genrsa -out rsa2048.pem 2048
10. pkcs15-init -a 1 --label "rsa2048" -S ./rsa2048.pem

Result: OK


10. Wrap/unwrap tests

Test output:
RSA unwrap test
Unwrapping generic secret using RSA succeeded
Unwrapping a CKA_TOKEN=FALSE AES key using RSA
Unwrapping AES key using RSA succeeded
Symmetric unwrap test
Unwrapping Generic Secret key using AES succeeded

C_WrapKey test:
Wrapped data: 0F8A806BA022C48FC82CC12F4F1B383040D055D3CE0BC9DA3D18799B9E160F3F

Result: OK

src/libopensc/card-myeid.c
frankmorgner: OpenSC should be haved as defined by PKCS#11:

>The signing operation must have been initialized with C_SignInit. A call to C_Sign always terminates the active signing operation unless it returns CKR_BUFFER_TOO_SMALL or is a successful call (i.e., one which returns CKR_OK) to determine the length of the buffer needed to hold the signature.

The standard doesn't give such details about `C_WrapKey`, so I think your implementation should be fine

src/libopensc/card-myeid.c
```diff
@@ -1185,6 +1261,149 @@ static int myeid_decipher(struct sc_card *card, const u8 * crgram,
...
+		0x2A, 0x00, p2);
+
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+	apdu.le = 0;
+
+	if (symmetric_operation)
+	{
+		/* symmetric crypto, no padding indicator byte */
+		{
```
frankmorgner: why wrapping this into curly brackets, again?

frankmorgner:
(or you resolve the conflict instead of rebasing).

src/libopensc/card-myeid.c
hhonkanen: I guess one pair of curly brackets is a leftover from older version and used to contain some code. Removed them as pointless.

hhonkanen:
@frankmorgner: 
Great!

Thank you for all your work and cooperation for getting this code merged.

I rebased the code once again to master, so everything should be resolved now.

We will have some things to improve later, but now there are no known issues which would break something or considered as bugs, so it will be easier to create new branches and PRs for the upcoming improvements.

For example we should be able to include AES algorithms in the SupportedAlgorithms structure of tokenInfo already during card initialization, to avoid having to update it (and query SO-PIN) during unwrap-operations. Adding the tests to pkcs11-tool is in the roadmap as well, and we could implement symmetric crypto with C_Encrypt and C_Decrypt at some point.

hhonkanen:
Merged again from master and resolved a new conflict which was introduced in 54cb10.

frankmorgner:
thanks, finally merged into master.

hhonkanen:
Great! Thanks @frankmorgner and everyone who participated in the review and discussion.

PR #1404, opened 2018-06-22
Add support for RSA PSS with SmartCard-HSM

CardContact:
Tested with SmartCard-HSM

##### Checklist

- [ ] macOS tokend is tested


src/pkcs11/framework-pkcs15.c
```diff
@@ -3680,6 +3680,58 @@ pkcs15_prkey_get_attribute(struct sc_pkcs11_session *session,
...
+	case CKM_SHA256:
+		if (hlen != 32)
+			return CKR_MECHANISM_PARAM_INVALID;
+		break;
+	default:
+		return CKR_MECHANISM_PARAM_INVALID;
+	}
+
+	// SmartCards typically only support MGFs based on the same hash as the
+	// message digest
```
Jakuje: This limitation is not on PKCS#11 level so the errors might be confusing. Even though, this might be a good idea to implement now opensc-wise (since the sc-hsm is the only supporting PSS), future drivers might struggle with it so I would propose to do this limitation in the sc-hsm code, if you know your cards do not support these cases.

src/pkcs11/framework-pkcs15.c
```diff
@@ -3680,6 +3680,58 @@ pkcs15_prkey_get_attribute(struct sc_pkcs11_session *session,
...
+		break;
+	case CKG_MGF1_SHA256:
+		if (hlen != 32)
+			return CKR_MECHANISM_PARAM_INVALID;
+		break;
+	default:
+		return CKR_MECHANISM_PARAM_INVALID;
+	}
+
+	// SmartCards typically support only a salt length equal to the hash length
```
Jakuje: The same for this condition -- PKCS#11 happily allows wider range of combinations.

src/pkcs11/framework-pkcs15.c
CardContact: Unfortunately this would require substantial changes in the lower layer to pass PSS parameter down into the card driver. It is something we could do in the long run, however I don't expect cards to support arbitrary salt length and MGF anytime soon. PKCS#11 allows the flexibility, but most implementation use saltlen = hashlen and MGF hash = message hash as all other cases would need to be explicitly agreed between sender and receiver.

Jakuje: Thank you for clarification. Yes, it makes sense to support at least something for now.

src/libopensc/pkcs15-sec.c
```diff
@@ -420,12 +420,16 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 
 	/* add the padding bytes (if necessary) */
 	if (pad_flags != 0) {
-		size_t tmplen = sizeof(buf);
+		if (flags & SC_ALGORITHM_RSA_PAD_PSS) {
+			// TODO PSS padding
```
Jakuje: This will be needed to implement the PSS off the card. Do you plan to have a look into that, or should I have a look into the possibilities here?

frankmorgner: We could maybe use this: https://github.com/NWilson/OpenSC/commit/42f3199e66796db6f003668e77315203d07f1038

frankmorgner:
@CardContact the code looks good, thanks! I think, I'll have time to test this next week.

Could you please also check which the necessary bits are to get this functionality into the minidriver?

frankmorgner:
currently PSS is also announced for decryption. You need to apply this patch:
```diff
diff --git a/src/pkcs11/framework-pkcs15.c b/src/pkcs11/framework-pkcs15.c
index 3040df63..e8611591 100644
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -5072,6 +5072,9 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
        /* TODO support other padding mechanisms */
 
        if (rsa_flags & SC_ALGORITHM_RSA_PAD_PSS) {
+               CK_FLAGS old_flags = mech_info.flags;
+               mech_info.flags &= ~CKF_DECRYPT;
+
                mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS_PSS, &mech_info, CKK_RSA, NULL, NULL);
                rc = sc_pkcs11_register_mechanism(p11card, mt);
                if (rc != CKR_OK)
@@ -5087,6 +5090,8 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
                        if (rc != CKR_OK)
                                return rc;
                }
+
+               mech_info.flags = old_flags;
        }
 
        if (rsa_flags & SC_ALGORITHM_ONBOARD_KEY_GEN) {
```

frankmorgner:
thanks!

@CardContact please also have a look into the minidriver

@Jakuje looking forward for your port of https://github.com/NWilson/OpenSC/commit/42f3199e66796db6f003668e77315203d07f1038

src/pkcs11/framework-pkcs15.c
mouse07410: And some cards/tokens (like YubiHSM2) do support other SHA2 family members.

PR #1407, opened 2018-06-24
OpenPGP cleanup & extensions towards better OpenPGP card v3+ support

marschap:
Hi,,

this pull request cleans up some FIXMEs/ToDos and starts the way towards v3+ support.

In particular, it
* adds DOs introduced with OpenPGP card  specs v2.1, v3.0 and v3.3 resp.
* updates and extends comments & references to the specs
* improves parsing of the "extended capabilities" DO, including v3.x support
* starts parsing the "extended length" DO introduced with v3.0
* improves debugging capabilities: "symmetric" calls of LOG_FUNC_...
* increases resilience by more/better error checks
* recognizes the card in question in more detail [e.g. with version number] on init.

Please merge it into OpenSC master, as it paves the ground for better v3+ support.
Of course: reviews welcome.

Best
PEter

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -669,7 +669,6 @@ pgp_get_card_features(sc_card_t *card)
 					/* v3.0+: max. size of special DOs is at bytes 7-8 */
 					priv->max_specialDO_size = bebytes2ushort(blob->data + 6);
 				}
-				/* TODO read Extended length information from DO 7F66 in OpenPGP 3.0 and later */
 			}
 		}
 
@@ -698,6 +697,22 @@ pgp_get_card_features(sc_card_t *card)
 
 					_sc_card_add_rsa_alg(card, keylen, flags, 0);
 				}
+			}
+		}
+
+		if (priv->bcd_version >= OPENPGP_CARD_3_0) {
+			/* v3.0+: get length info from "extended length information" DO */
+			if ((pgp_get_blob(card, blob6e, 0x7f66, &blob) >= 0) &&
+				(blob->data != NULL) && (blob->len >= 8)) {
+				/* kludge: treat as SIMPLE DO and use appropriate offsets */
+				card->max_send_size = bebytes2ushort(blob->data + 2);
+				card->max_recv_size = bebytes2ushort(blob->data + 6);
+			}
+
+			/* v3.0+: get card features from "general feature management" DO */
+			if ((pgp_get_blob(card, blob6e, 0x7f74, &blob) >= 0) &&
+				(blob->data != NULL) && (blob->len >= 5)) {
+				// ToDo: parse tag 81 in this DO ...
 			}
 		}
 	}
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -315,13 +315,12 @@ static struct do_info		pgp33_objects[] = {	/* OpenPGP card spec 3.3 */
 	{ 0x5f48, CONSTRUCTED, READ_NEVER  | WRITE_PIN3,  NULL,               sc_put_data },
 	{ 0x5f50, SIMPLE,      READ_ALWAYS | WRITE_PIN3,  sc_get_data,        sc_put_data },
 	{ 0x5f52, SIMPLE,      READ_ALWAYS | WRITE_NEVER, sc_get_data,        NULL        },
-	/* The 7F21 is constructed DO in spec, but in practice, its content can be retrieved
-	 * as simple DO (no need to parse TLV). */
+	/* DO 7F21 is CONSTRUCTED in spec; we treat it as SIMPLE: no need to parse TLV */
 	{ DO_CERT, SIMPLE,      READ_ALWAYS | WRITE_PIN3,  sc_get_data,        sc_put_data },
 	{ 0x7f48, CONSTRUCTED, READ_NEVER  | WRITE_NEVER, NULL,               NULL        },
 	{ 0x7f49, CONSTRUCTED, READ_ALWAYS | WRITE_NEVER, NULL,               NULL        },
 	{ DO_AUTH,     CONSTRUCTED, READ_ALWAYS | WRITE_NEVER, pgp_get_pubkey,     NULL   },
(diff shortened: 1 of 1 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -489,23 +489,34 @@ pgp_init(sc_card_t *card)
...
 		if (card->name == default_cardname_v1 ||
 		    card->name == default_cardname_v2 ||
 		    card->name == default_cardname_v3) {
-			static char card_name[SC_MAX_APDU_BUFFER_SIZE] = "OpenPGP card";
-
-			snprintf(card_name, sizeof(card_name), "OpenPGP card v%u.%u",
-				 card->version.hw_major, card->version.hw_minor);
-			card->name = card_name;
+			snprintf(card_name, sizeof(card_name),
+				 "OpenPGP card v%u.%u (%04X %08lX)",
```
frankmorgner: Why include the serial number in the card name? All other cards keep this seperately; as it is available via `opensc-tool --serial`. Also, I doubt that it will have some meaning for the average user...

src/libopensc/card-openpgp.c
```diff
@@ -44,13 +51,18 @@
...
+static const char default_cardname[]    = "OpenPGP card";
+static const char default_cardname_v1[] = "OpenPGP card v1.x";
+static const char default_cardname_v2[] = "OpenPGP card v2.x";
+static const char default_cardname_v3[] = "OpenPGP card v3.x";
+
 static struct sc_atr_table pgp_atrs[] = {
-	{ "3b:fa:13:00:ff:81:31:80:45:00:31:c1:73:c0:01:00:00:90:00:b1", NULL, "OpenPGP card v1.0/1.1", SC_CARD_TYPE_OPENPGP_V1, 0, NULL },
-	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:c5:73:c0:01:40:00:90:00:0c", NULL, "CryptoStick v1.2 (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
+	{ "3b:fa:13:00:ff:81:31:80:45:00:31:c1:73:c0:01:00:00:90:00:b1", NULL, default_cardname_v1, SC_CARD_TYPE_OPENPGP_V1, 0, NULL },
+	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:c5:73:c0:01:40:00:90:00:0c", NULL, default_cardname_v2, SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
```
frankmorgner: This replaces the specific name `CryptoStick` with an unspecific name. Is this intended?

src/libopensc/card-openpgp.c
```diff
@@ -44,13 +51,18 @@
 #include <openssl/sha.h>
 #endif /* ENABLE_OPENSSL */
 
+static const char default_cardname[]    = "OpenPGP card";
+static const char default_cardname_v1[] = "OpenPGP card v1.x";
+static const char default_cardname_v2[] = "OpenPGP card v2.x";
+static const char default_cardname_v3[] = "OpenPGP card v3.x";
```
frankmorgner: Don't use too much information: please remove the `".x"` if you don't have any information about it.

src/libopensc/card-openpgp.c
```diff
@@ -629,19 +731,36 @@ pgp_get_card_features(sc_card_t *card)
...
+			/* v3.0+: get length info from "extended length information" DO */
+			if ((pgp_get_blob(card, blob6e, 0x7f66, &blob) >= 0) &&
+				(blob->data != NULL) && (blob->len >= 8)) {
+				/* kludge: treat as SIMPLE DO and use appropriate offsets */
+				card->max_send_size = bebytes2ushort(blob->data + 2);
+				card->max_recv_size = bebytes2ushort(blob->data + 6);
+			}
+
+			/* v3.0+: get card features from "general feature management" DO */
+			if ((pgp_get_blob(card, blob6e, 0x7f74, &blob) >= 0) &&
```
frankmorgner: Why read that data if you don't need it?

The same is true for other parsed values or data structures, e.g. `max_specialDO_size`.

I'd better like to see that included when we're actually doing something useful with it. Currently, that's dead weight!

src/libopensc/card-openpgp.c
martinpaljak: It _might_ be useful to differentiate between multiple tokens.

src/libopensc/card-openpgp.c
marschap: Yes, this was deliberate.

As mentioned in the comment to commit 20cd0c08, this ATR is the _generic_ ATR for OpenPGP v2 cards, which is not only used in the legacy [not available on the market anymore] CryptoStick, but also in ZeitControl's OpenPGP cards, in the NitroKey devices, and most probably others too.

The Name CryptoStick for this ATR is too narrow and misleading: whenever I put my NitroKey into the USB port, or my OpenPGP card into the reader, OpenSC wrongly announced it as CryptoStick.
The mentioned commit fixes the issue.

src/libopensc/card-openpgp.c
marschap: Hmmm, I tend to slightly disagree.
In my interpretation "v3" is equivalent "v3.0"
Hence the ".x", to indicate that we not know exactly whether it is "v3.0", "v3.1", ...

If you really insist, I can remove it. But please give it a 2nd thought.

src/libopensc/card-openpgp.c
marschap: OK, understandable.

Instead of amending this PR and reverting the chanes in additional commits, I lean to calceling this one, and putting in a new one with the offending passages removed.

What's your preference?

frankmorgner: choose as you like, thanks

src/libopensc/card-openpgp.c
frankmorgner: ok, whatever

src/libopensc/card-openpgp.c
frankmorgner: thanks for the clearification

marschap:
I will address the chanes in a separate pull request.
Hereby closing this one.

PR #1408, opened 2018-06-27
Fix #1403 and add GenericName

### LINKED ISSUE #1403: [PATCH] Fix completiondir
### Problem Description
Recent versions of bash is leaving /etc/bash_completion.d. The correct directory is specified by `pkg-config --variable completionsdir bash-completion`.

### Proposed Resolution
If possible, use pkg-config to determine the correct directory.

### Steps to reproduce
Install opensc with the latest bash.

### Logs
N/A

stanislav-brabec:
Fix https://github.com/OpenSC/OpenSC/issues/1403

Add GenericName. it is not mandatory by the spec, but in many desktop environments it is used for menu item visual rendering:
- Name: Title of the icon
- GenericName: Subtitle of the icon
- Comment: Pop-up message when user hovers over icon

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -839,6 +839,10 @@ if test "${enable_cryptotokenkit}" = "yes"; then
 	fi
 	AC_DEFINE([ENABLE_CRYPTOTOKENKIT], [1], [Define if CryptoTokenKit is to be enabled])
 fi
+PKG_CHECK_MODULES([BASH_COMPLETION], [bash-completion >= 2.0],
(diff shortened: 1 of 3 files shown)


frankmorgner:
On which windows manager can I see GenericName displayed?

stanislav-brabec:
At least GNOME, KDE and XFCE core libraries already support GenericName.

However it does not mean that it is actually used. For example the default GNOME menu theme uses only Name and nothing else.

It causes no harm to have all three defined correctly.

PR #1409, opened 2018-06-28
Few more improvements for HID tokens and make the testsuite build

### LINKED ISSUE #1405: p11test does not build on Ubuntu 16.04
### Problem Description

    ./configure --enable-tests
    make

fails with unresolved links to cmocka.

### Proposed Resolution

Replace CMOCKA_LDFLAGS in p11test/Makefile.am with CMOCKA_LIBS
Allow version of CMOCKA to be 1.0.1 in configure.ac

Jakuje:
The HID tokens present additional certificates in undocumented AID. Additionally, the logging is made more consistent for better debugging.

The make now references the correct autotools variable.
Resolves #1405

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tests/p11test/Makefile.am
+++ b/src/tests/p11test/Makefile.am
@@ -23,7 +23,7 @@ p11test_SOURCES = p11test.c p11test_loader.c \
 	p11test_case_pss_oaep.c \
 	p11test_helpers.c
 p11test_CFLAGS = -DNDEBUG $(CMOCKA_CFLAGS)
-p11test_LDADD = -lssl -lcrypto $(CMOCKA_LDLAGS)
+p11test_LDADD = -lssl -lcrypto $(CMOCKA_LIBS)
 
 if WIN32
 p11test_SOURCES += $(top_builddir)/win32/versioninfo.rc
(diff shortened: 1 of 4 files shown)


src/tests/p11test/Makefile.am
```diff
@@ -23,7 +23,7 @@ p11test_SOURCES = p11test.c p11test_loader.c \
 	p11test_case_pss_oaep.c \
 	p11test_helpers.c
 p11test_CFLAGS = -DNDEBUG $(CMOCKA_CFLAGS)
-p11test_LDADD = -lssl -lcrypto $(CMOCKA_LDLAGS)
+p11test_LDADD = -lssl -lcrypto $(CMOCKA_LIBS)
```
frankmorgner: Would it make sense to use `$(OPTIONAL_OPENSSL_LIBS)` and/or to additionally check for libssl in `configure.ac`?

Jakuje: Once we are in rewriting that, yes, we should use `OPTIONAL_OPENSSL_LIBS`. Looking thrugh it, libssl itself is not needed at all if I am right. Adjusting configure to build tests also only if we have openssl is also good idea. Will commit changes soon.

Jakuje:
I also pushed a change to pkcs15-tool, which was needed to build the OpenSC with current GCC in Fedora:
```
pkcs15-tool.c: In function ‘read_ssh_key’:
pkcs15-tool.c:1023:3: error: ‘strncpy’ specified bound 20 equals destination size [-Werror=stringop-truncation]
   strncpy(alg, (char *) buf+4, 20);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

frankmorgner:
thanks

PR #1422, opened 2018-07-06
no OPTSWITHARGS produces invalid completion file

### LINKED ISSUE #1421: completions for opensc-asn1 have invalid syntax (head/git)
### Problem Description

When bash completions are being sourced getting an error (bash parsing error) when using the head/git version of opensc (I believe this started after the change to use pkg-config for completions, but uncertain, haven't been using my laptop the last few days but these are maybe generated there?). This archlinux using the AUR opensc-git

### Proposed Resolution

N/A

### Steps to reproduce

Install head/git version of opensc and source the completion scripts

### Logs

```
bash: /etc/bash_completion.d/opensc-asn1: line 26: syntax error near unexpected token `)'
bash: /etc/bash_completion.d/opensc-asn1: line 26: `        		)'
bash: /etc/bash_completion.d/opensc-asn1: line 26: syntax error near unexpected token `)'
bash: /etc/bash_completion.d/opensc-asn1: line 26: ` 
```

this is line 26, the hanging `)`
```
            _filedir
            return 0
            ;;
                        )
            return 0
            ;;
```

enckse:
Fixes #1421, when a tool has no OPTSWITHARGS it will produce invalid bash (completion) output syntax.

I did a diff on the the bash completions produced before this patch and the output of this change and the only change is to opensc-asn1 which is now valid.

##### Checklist

### DIFF
--- a/doc/tools/Makefile.am
+++ b/doc/tools/Makefile.am
@@ -36,7 +36,7 @@ tools.html: $(srcdir)/tools.xml $(wildcard $(srcdir)/*.1.xml) $(wildcard $(srcdi
 				| sort -u | grep -- '^\-' | tr '\n' ' ')," \
 		| sed "s,OPTSWITHARGS,\
 			$(shell sed -n 's,.*<option>\([^<]*\)</option>.*<replaceable>.*,\1,pg' $< \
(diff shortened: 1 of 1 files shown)


enckse:
unrelated(ish): always irritating when travis errors (not a fail) for some reason :(

frankmorgner:
thanks

PR #1423, opened 2018-07-06
Bash_completionsdir should use PREFIX as set by configure

dengert:
When building OpenSC to be installed in alternate locations,
The bash completions location should also be based on the alternate
location.

In previous code or bash-completion version  less then 2.0:
completiondir="${sysconfdir}/bash_completion.d"
this would work, as ${sysconfdir} was based on ${prefix}

Current master uses the systems location for the completionsdir
completiondir="`pkg-config --variable=completionsdir bash-completion`"

This commit uses the pkg_config ability to overwrite the prefix
to restore the ability to install the completions in an alternate
location:
completiondir="`pkg-config --variable=completionsdir --define-variable=prefix=${prefix} bash-completion`"

 Changes to be committed:
	modified:   ../../configure.ac

For example Ubuntu-16.4  /usr/share/pkgconfig/bash-completion.pc:
```
prefix=/usr
compatdir=/etc/bash_completion.d
completionsdir=${prefix}/share/bash-completion/completions
helpersdir=${prefix}/share/bash-completion/helpers

Name: bash-completion
Description: programmable completion for the bash shell
URL: http://bash-completion.alioth.debian.org/
Version: 2.1
```

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -840,7 +840,7 @@ if test "${enable_cryptotokenkit}" = "yes"; then
 	AC_DEFINE([ENABLE_CRYPTOTOKENKIT], [1], [Define if CryptoTokenKit is to be enabled])
 fi
 PKG_CHECK_MODULES([BASH_COMPLETION], [bash-completion >= 2.0],
-	[completiondir="`pkg-config --variable=completionsdir bash-completion`"],
+	[completiondir="`pkg-config --variable=completionsdir --define-variable=prefix=${prefix} bash-completion`"],
 	[completiondir="${sysconfdir}/bash_completion.d"])
 AC_SUBST([completiondir])
 


frankmorgner:
I think the global completion directory should be used as default without our prefix, because this would mean we don't install to the default completion directory. Thus, I don't support this change.

However, we should allow overwriting the location by the user. Is it possible to do `./configure --prefix=/tmp/ completiondir="$(pkg-config --variable=completionsdir --define-variable=prefix=/tmp/ bash-completion)"` with current master?

dengert:
I would like to hear from @stanislav-brabec, @lbakman and @eighthave on this subject. 

My complaint is:  By default OpenSC configure installs OpenSC into `/usr/local`  but then attempt to install the completions into the running system at `completiondir = /usr/share/bash-completion/completions`
I consider that a bug.  The default should never be to overwrite system versions of anything. 

If I run  `configure --prefix=/mytest`  or `configure --prefix=$HOME/mytest` everything else will be installed using the prefix (libs, programs, man pages, includes, opensc.conf, etc.)  Except `bash completions` will still  be attempted to be installed in `/usr/share/bash-completion/completions.` 

This is not what a user would expect, overwriting the OS version of any file when they specifically told the configure where things should be installed.  (To overwrite the OS versions of the completion files the user would still have to use `sudo install`  but they would do that to install into `/usr/local` )
 
(I found this as I install several version of OpenSC using `--prefix=/xxxx` where xxx is based on openssl versions. I have write permission as a user, so don't need to use sudo to do an install. This prevents accidentally overwriting system files like in   `/usr/share/bash-completion/completions.` In addition the install fails making it impossible to get the rest of OpenSC to install.)

The change was introduced in https://github.com/OpenSC/OpenSC/commit/b0a20fa7ca30dc6c6198954fadc5a7fe812834dc 

Prior to this (or if bash_completion less then version 2)  `completiondir="${sysconfdir}/bash_completion.d"` was used  which is based on `${sysconfdir}`  which is based in `${prefix}`

If there is some other solution, OK.  But the default should never be to overwrite system versions of anything.

stanislav-brabec:
Your change will break bash completions of opensc. If you use bash in /usr or / and opensc in /usr/local completions will stop to work.

For bash completions, you have to use directory based on `PREFIX` that was used by configure script of bash, not configure script of opensc. bash does not read from other directories (except support for deprecated `/etc/bash_completion.d`). bash configured with /usr PREFIX does not read files in `/usr/local/etc/bash_completion.d` nor `/usr/local/share/bash-completion/completions` (I just verified that).

It is a general problem: Use of package specific prefix does not work for system integration files. You have to use prefix specific to the package that uses the integration files. Only few packages have a specific support for `/usr/local`.

It is possible to add `--disable-completions` (or something similar that already exists in other packages), but installation of completions (as well as e. g. systemd services) anywhere in `/usr/local` makes no sense. It is also possible to add a specific check for `/usr/local` and issue a warning.

dengert:
OK as you say: "It is a general problem: Use of package specific prefix does not work for system integration files." 
`--disable-completions`  would work to accidentally overwrite existing system system files.  

Right now I will settle for doing make -k install.

frankmorgner:
closed with 376de5bd4c223a6dd423b657892ccbe73145e4e3, use `--with-completiondir=PATH`

PR #1425, opened 2018-07-08
OpenPGP: extend manufacturer list in pkcs15-openpgp.c

marschap:
Hi,

this PR is a small extension to pkcs15-openpgp.c that extends the list of OpenPGP card manufacturers,
and brings them in line with those in GnuPG.

Please add this PR to OpenSC master.

Best
Peter


frankmorgner:
thanks

PR #1426, opened 2018-07-08
fixes & extenstions to opensc-explorer

marschap:
Hi 

this multi-commit PR fixes & extends opensc-explorer including the manual page.
It
* adds a new command 'pin_info' to get information a PIN or key
* extends the 'do_put' command to accept more than 2 parameters.
  All parameters starting with the 2nd are joined the same way as fopr 'apdu'
* extends the 'random' command to optionally accept a file name as second
  argument as the target to wrtie the generated random bytes to.
* overhaules the manual page, 
  - bringing a more consistent XML style
  - giving more & longer explanations
  - fxing errors where the man page differed from the implementation
* fixes an error that could have caused a buffer overflow.
* ...

Please include it in OpenSC master.

Best
Peter

##### Checklist


doc/tools/opensc-explorer.1.xml
```diff
@@ -104,13 +113,32 @@
 			interactive prompt or in script files passed via the command line parameter
 			<replaceable class="parameter">SCRIPT</replaceable>.
 			<variablelist>
+				<varlistentry>
+					<term>
+						<command>#</command>
+						<arg choice="plain" rep="repeat"><replaceable></replaceable></arg>
```
frankmorgner: wouldn't this make more sense with `<arg choice="plain" rep="repeat"><replaceable>comment</replaceable></arg>`?

src/tools/opensc-explorer.c
```diff
@@ -1670,28 +1763,31 @@ static int do_get_data(int argc, char **argv)
...
 	/* Extract the new content */
-	/* buflen is the max length of reception buffer */
-	r = parse_string_or_hexdata(argv[1], buf, &buflen);
-	if (r < 0) {
-		printf("unable to parse data\n");
-		return -1;
+	for (i = 1, len = 0; i < (unsigned) argc; i++)   {
+		size_t len0 = sizeof(buf) - len;
+
+		if ((r = parse_string_or_hexdata(argv[i], buf + len, &len0)) < 0) {
```
frankmorgner: Isn't it counter intuitive to have muliple arguments? You're effectively ommiting the white space between two arguments when writing to the card.

src/tools/opensc-explorer.c
```diff
@@ -1670,28 +1763,31 @@ static int do_get_data(int argc, char **argv)
...
 	tag = strtoul(argv[0], NULL, 16);
 
 	/* Extract the new content */
-	/* buflen is the max length of reception buffer */
-	r = parse_string_or_hexdata(argv[1], buf, &buflen);
-	if (r < 0) {
-		printf("unable to parse data\n");
-		return -1;
+	for (i = 1, len = 0; i < (unsigned) argc; i++)   {
+		size_t len0 = sizeof(buf) - len;
```
frankmorgner: Length checking is missing completely. I think it's possible to have an integer underflow here.

doc/tools/opensc-explorer.1.xml
marschap: This piece of DocBook yields   `# ...` in the manual page.
I do not really care whether it stays as it is or becomes `# comment...`
If you explicitly want `comment` included, please tell and I amend the PR.

src/tools/opensc-explorer.c
marschap: Maybe. But ...
If you call it with only one argument, there is **no** change in behaviour to the previous version.

This is because the command line is split into white-space (i.e. `" \t\n"`) separated tokens before the the functions are parsed and the handler for the individual function is called. This is how opensc-explorer works.
With this setup, it is (and this is unchanged!)
* impossible after the split to know what combination of white-space existed between the arguments before the split
* only possible to include white-space as a part of a hex-string (e.g. `20:08`) or a double-quoted string (e.g. `" "`).

My change only treats `do_put` the same as `apdu` by accepting multiple arguments.

Perhaps a clarifying sentence in the doc could help.
Let me see ...

src/tools/opensc-explorer.c
marschap: Oops, correct finding.

In current opensc-explorer it cannot happen because `sizeof(buf)` is larger than the whole command line (and thus also larger than the sum of the parsed command line arguments).
But we should not rely on that.

Let me see what I can do.

marschap:
the added commit d9709f64eac1ffea9ad3fdccc507c32e447d8082 should fix the missing length check - not only for `do_put`, but also for `apdu`.

marschap:
@frankmorgner The latest commit adds clarification to the manual, hopefully addressing your worries about multiple arguments for 'do_put'.

marschap:
A few more fixes, among them a potential buffer overrun.

frankmorgner:
Yes, you have clarified the behavior in the manual, but people are ignoring manuals all the time. I think the added value of concatenating strings in opensc-explorer rather than concatenating them externally, is not worth the trouble thinking about where to ignore spaces and maintaining this code. The benefit is simply too limited to break with basically all other operations that are expecting data (e.g. `update_binary`,  `update_record`, `verify`, ...) which all take a single argument of data.

marschap:
@frankmorgner I reverted the change allowing multiple `data` arguments for `do_put`.
I hope this addresses your issues and allows you to merge this PR.

marschap:
@frankmorgner Please tell what you are missing in this PR to get it pulled into master.

frankmorgner:
Sorry for my few words. For now, I don't have more objections, but I'd like to postpone this for the next release.

marschap:
@frankmorgner,

Change requests addressed with additional commits

BTW: Thanks for taking care of the gids compilation issue. I did not get round to react to your chanes in time, as I usually (i.e. outside vacations) only have time over the weekends to work on opensc.

frankmorgner:
Could you squash the fixups, or do you want me to to squash and merge the changes into a single commit?

marschap:
Withdrawn in favor of PR #1487

Issue #201, opened 2013-12-31
Broken reader, C3PO LTC31


ghost:
This reader doesn't have a pinpad. Could it be added to the broken list?

```
--- reader-pcsc.c.orig  2013-12-31 19:26:55.000000000 +0100
+++ reader-pcsc.c   2013-12-31 19:30:37.000000000 +0100
@@ -819,7 +819,10 @@
    PCSC_TLV_STRUCTURE *pcsc_tlv;
    LONG rv;
    const char *log_disabled = "but it's disabled in configuration file";
-   const char *broken_readers[] = {"HP USB Smart Card Keyboard"};
+   const char *broken_readers[] = {
+       "C3PO LTC31 v2",
+       "HP USB Smart Card Keyboard"
+   };

    SC_FUNC_CALLED(ctx, SC_LOG_DEBUG_NORMAL);

```

PR #1438, opened 2018-07-25
OMNIKEY 3x21 and 6121 Smart Card Reader are not pinpad readers

metsma:
macOS 10.13 ships with ccid driver 1.4.27 (fixed in 1.4.29) and this version identifies these readers wrongly as pinpad readers.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -1063,6 +1063,7 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 	PCSC_TLV_STRUCTURE *pcsc_tlv;
 	LONG rv;
 	const char *log_disabled = "but it's disabled in configuration file";
+	int id_vendor = 0, id_product = 0;
 
 	LOG_FUNC_CALLED(ctx);
 
@@ -1129,6 +1130,17 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 			reader->capabilities |= SC_READER_CAP_PIN_PAD;
 		} else {
 			sc_log(ctx, "%s %s", log_text, log_disabled);
+		}
+	}
+
+	/* Some readers claim to have PinPAD support even if they have not */
+	if ((reader->capabilities & SC_READER_CAP_PIN_PAD) &&
+		part10_get_vendor_product(reader, card_handle, &id_vendor, &id_product) == SC_SUCCESS) {
+		/* HID Global OMNIKEY 3x21/6121 Smart Card Reader, fixed in libccid 1.4.29 (remove when last supported OS is using 1.4.29) */
+		if ((id_vendor == 0x076B && id_product == 0x3031) ||
+			(id_vendor == 0x076B && id_product == 0x6632)) {
+			sc_log(ctx, "%s is not pinpad reader, ignoring", reader->name);
+			reader->capabilities &= ~SC_READER_CAP_PIN_PAD;
 		}
 	}
 


frankmorgner:
that should be fixed in libccid instead of opensc, right @LudovicRousseau ?

martinpaljak:
That probably is fixed and yes, it is "technically correct" to be a thing to be fixed there. But for now this is a meaningful UX fix.

frankmorgner:
Oh, reading your comment this is already the case. Then, we'll definately not modify OpenSC over this. Please use the `enable_pinpad` option of `opensc.conf` instead.

metsma:
I cannot distribute opensc with enable_pinpad=no option. Some are using real PinPAD-s there.
And the 3x21 is replacing 1021 Omnikey reader and is getting common reader these days.
We are seeing increased support calls and macOS 10.13 support is not going anywhere soon.

src/libopensc/reader-pcsc.c
```diff
@@ -1132,6 +1136,13 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 		}
 	}
 
+	for (i = 0; i < sizeof(broken_readers)/sizeof(broken_readers[0]); ++i) {
```
martinpaljak: Maybe add a "XXX" kind of header here to remind that this should at some point, be removed ?

frankmorgner:
Once libccid 1.4.27 is gone, this would effectively be dead code in OpenSC. Additionally, it is not standardized how the reader name has to look like, i.e. the detection mechanism you implemented may not be accurate.

Some other options you could think about:
- distribute a shell script that detects the bogus readers and configures `opensc.conf` accordingly
- distribute libccid 1.4.29 yourself
- add a new option to `opensc.conf`, which allows setting the reader driver options based on the reader's name:
```
reader_driver pcsc {
  reader_name "HID Global OMNIKEY 3x21 Smart Card Reader" {
    enable_pinpad = false;
  }
}
```

metsma:
Right now macOS 10.14 beta uses still 1.4.27 (created bug report to apple) and 10.13 will be still supported by opensc +3-4 years? This issue will not go anywhere anytime soon.
I can distribute script or libccid from my channels, but what happens with other opensc users? It will not help them.
I will look opensc.conf solution when I get time.

martinpaljak:
@frankmorgner this is basic UX (or UF - User-Friendlyness): if simple code can be added that makes things JustWork(tm) for many users - without them requiring to change something or add something - it should be done. With a meaningful comment that this code should be removed when the circumstances have changed. This is a generic fixup for a set of readers (not really related to .ee, even though there is probably a lot more of these readers in circulation and use here than elsewhere) so IMO it makes a lot of sense. Compare:  https://github.com/LudovicRousseau/CCID/blob/master/src/ccid.c#L54

martinpaljak:
Oh: these readers should be fixed on the *firmware* level but that is even less likely to happen, so such overrides make sense (with as good as possible localisation checks)

mouse07410:
Not to mention that with the SIP on MacOS an ordinary user cannot replace system-provided libccid. And even the brave souls like yours truly, would hesitate, not knowing all the consequences of replacing one of the system libraries. Does it have any "unlisted" modifications? Are there any dependencies on this maybe-modified 1.4.27 in any other system module?

It is not beautiful, but I think OpenSC should step in and provide the fix that the OS should provide in theory - but as we all know is extremely unlikely to provide in practice.

dengert:
I have a Omnikey 3821 and it is a PIN PAD reader. So why do you say "OMNIKEY 3x21" is not a PIN PAD reader? 

Windows says it is: `USB\VID_076B&PID_3821&REV_0600` and it works from Windows and when attached to VirtualBox VM running Ubuntu with libccid=1.4.22

Does HID or Omnikey have any firmware updates to fix the problem? 


@metsma  Do you or any of your users use any PIN_PAD readers? 
Since you say these readers are not pinpad reader I would assume the answer is no.

So is the fix for you just set in `opensc.conf` `enable_pinpad = false;`

For those uses who do use a pin-pad reader then don't set this.

Are there any uses who uses both a PIN PAD reader and one of the readers in question? (I would expect none or only some developers.) In that case buy a new reader.

martinpaljak:
@dengert yes we use pinpad readers, several of them.

The issue is, that these specific readers falsely advertise pinpad. And this obviously fails with nice messages from a browser or similar. 

https://ccid.apdu.fr/ccid/shouldwork.html#0x076B0x3031 is not the same as https://ccid.apdu.fr/ccid/shouldwork.html#0x076B0x3821 (do not take their name too seriously) and the matching is done on the name reported by the driver, not the "model of the reader".

The key here is not "change config file" as those affected (they are the cheapest, most common readers sold in supermarkets) probably don't even know they are using OpenSC somewhere....

metsma:
Seems like Omnikey is adding 3x21 chip in 1021 reader case.
https://ccid.apdu.fr/ccid/shouldwork.html#0x076B0x1021
You can identify the difference by blue sticker has changed "OK" to "HID"

LudovicRousseau:
A better way to identify a reader is to use `PCSCv2_PART10_PROPERTY_wIdVendor` and `PCSCv2_PART10_PROPERTY_wIdProduct`.
See https://ludovicrousseau.blogspot.com/2012/04/identifying-reader-model-part-2.html

These 2 readers have a new firmware that fixes the problem (update of `bPINSupport` field).
- https://salsa.debian.org/rousseau/CCID/commit/a9bab4d43756f3a4ecca490667897b8a7cc1e444
- https://salsa.debian.org/rousseau/CCID/commit/76cc52bc4a1de391d87b840f8927e16aa2eec513

I don't know if it is possible to upgrade the firmware on a reader.

frankmorgner:
Hmm, maybe it's better to include this code... The same issue has been reported multiple times in the past, e.g. in https://github.com/OpenSC/OpenSC/issues/201. There was also https://github.com/OpenSC/OpenSC/issues/1221, with a similar problem, that may be hard coded the same way.

@metsma could you check the possibility of identifying the reader as [described above](https://github.com/OpenSC/OpenSC/pull/1438#issuecomment-407823058)? Please look at https://github.com/OpenSC/OpenSC/pull/209, which had this method already implemented for some other readers.

@LudovicRousseau, do you know if Windows supports the request for `PCSCv2_PART10_PROPERTY_wIdVendor` and `PCSCv2_PART10_PROPERTY_wIdProduct` (if so, which version?)

LudovicRousseau:
I think the CCID driver provided by Microsoft does not support these properties.
The Microsoft CCID driver does not even support Secure Pin Entry (pinpad) AFAIK. So do not expect much from Microsoft.

dengert:
Windows 10 PRO with an Omnikey 3821 PIN PAD reader works with `pkcs11-tool --login --test` with  opensc 0.17.0. (one of the pre release I had installed), It displays the pin prompt on the reader and accepts the pin sometimes. It may be a timeout issue or the enter check key on the reader has a problem. 

```
2018-07-27 08:25:13.297 OMNIKEY CardMan 3821 0:SCardConnect(DIRECT): 0x00000000
2018-07-27 08:25:13.297 [opensc-pkcs11] reader-pcsc.c:1063:detect_reader_features: called
2018-07-27 08:25:13.297 Requesting reader features ...
2018-07-27 08:25:13.298 Reader feature 06 found
2018-07-27 08:25:13.298 Reader feature 07 found
2018-07-27 08:25:13.298 Reader feature 0f found
2018-07-27 08:25:13.298 Reader feature 0f is not supported
2018-07-27 08:25:13.298 Reader feature 11 found
2018-07-27 08:25:13.298 Reader feature 11 is not supported
2018-07-27 08:25:13.298 Reader feature 0a found
2018-07-27 08:25:13.298 Reader feature 10 found
2018-07-27 08:25:13.298 Reader feature 10 is not supported
2018-07-27 08:25:13.298 Reader supports pinpad PIN verification
2018-07-27 08:25:13.298 Reader supports pinpad PIN modification
2018-07-27 08:25:13.298 Reader has a display: 0210
```

But using the Microsoft built in PIV driver (i.e. no OpenSC code)  and `certutil -v -scinfo` it does not use the PIN PAD reader, but accepts PIN via a Message Box. 

So it looks like the windows CCID driver does retrieve the properity10 info, but the Microsoft utilities  don't use the feature. 

So a bogus card that reported the feature would not work with the opensc minidriver.

metsma:
You have probably Omnikey CCID driver installed from Windows Update.
If you take clean windows and try with offline machine, then it uses windows build in CCID driver
Microsoft Usbccid Smartcard Reader (WUDF).

dengert:
OK, Device manager says "HID Global", 3/20/2014 Version "1.2.24.27" 
After Driver update, it says "HID Global",  5/25/2017 Version 1.2.29.156. 
But `certutil -v -scinfo` still does not use the PIN PAD feature from Microsoft utilities. 

So this does not show  anything about MacOS. If problem is also seen on Windows try doing Update Driver and see if there is a newer version from HID for Windows.

martinpaljak:
I have a fair amount of confidence that the amount of test platforms and data is sufficient in @metsma team to focus on best code for OpenSC.

metsma:
Anyway I repeat 3x21 is not the same as 3821 and 3621.
It is NOT pinpad reader and making things worse chip is packaged into 1021 case.

Issue #1436, opened 2018-07-23
Multiple PIN prompts when using card authentication in Firefox


thw0rted:
### Problem Description

I'm getting prompted for a PIN multiple times when using card authentication in Firefox.

### Proposed Resolution

Using the Windows 10 default drivers (for Identiv SCR3310), I'm able to open the same page in Edge or Chrome (via MSCAPI, I assume) with a single PIN prompt.

### Logs

I logged in to the site where I'm seeing the issue with PKCS Spy enabled.  [This is the result](https://gist.github.com/thw0rted/81de0beb3319f7f6f6556bfcf6d68d49).  I also went through the procedure again with debug logging turned on, but a single login action generated more than 600kb of logs (at level=3) and I'm not sure I'm sanitizing it correctly, so please let me know if it's needed and I can share it individually.

thw0rted:
I swear I tried that before, but doing it now brings the prompts from 4 or 5 down to 2 or 3.  I can live with that, but do you know why even the onepin module might ask more than once?

enckse:
I saw this issue and have always wondered if there was a reliable way to get less prompts - I changed over to onepin as well and also get 2-3 prompts (which is what I was getting before).

dengert:
You card is a PIV or PIV like card. There are 4 certificates, one for User Authentication, one for digital Signatures, one foe key management (encryption) and one for the card to authenticate itself. 

The SPY trace shows that FireFox chose to to use the  "Certificate for Digital Signature" which is meant to be used for signing such things as e-mails, but not for web or login authentication. The NIST policy for the  "Certificate for Digital Signature" is to require the PIN for every use, and the card enforces it. 
Look at the trace for `CKA_ALWAYS_AUTHENTICATE  True` This tells the application (FireFox) to request the PIN for each use. And that is what it is doing. 

Why FireFox chose the "Certificate for Digital Signature" rather than the "Certificate for PIV Authentication" is not clear. It depends on the content of the certificates and the website. 

You choices are:
Look at the certificates and what the website is looking for in a certificate so it picks the Authentication certificate. 

You can also look at the opensc.conf file and make sure `use_pin_caching = true;` and 
`pin_cache_ignore_user_consent = true;` But Firefox is still enforcing the it, so this may not help.

The onepin module should give the same results.

mouse07410:
[...]

PR #1441, opened 2018-07-30
Unbreak coolkey getRandom

Jakuje:
It showed up that coolkey was using bad instruction code. Fixing that makes the get random working as expected. The new instruction code is taken from the coolkey applet source code:

https://github.com/dogtagpki/coolkey/blob/master/applet/src/com/redhat/ckey/applet/CardEdge.java#L216

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-coolkey.c
+++ b/src/libopensc/card-coolkey.c
@@ -1196,7 +1196,7 @@ static int coolkey_read_binary(sc_card_t *card, unsigned int idx,
 {
 	coolkey_private_data_t * priv = COOLKEY_DATA(card);
 	int r = 0, len;
-	u8 *data = NULL;;
+	u8 *data = NULL;
 
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 	if (idx > priv->obj->length) {
--- a/src/libopensc/card-coolkey.c
+++ b/src/libopensc/card-coolkey.c
@@ -89,7 +89,7 @@
 #define COOLKEY_INS_COMPUTE_CRYPT              0x36
 #define COOLKEY_INS_COMPUTE_ECC_KEY_AGREEMENT  0x37
 #define COOLKEY_INS_COMPUTE_ECC_SIGNATURE      0x38
-#define COOLKEY_INS_GET_RANDOM                 0x73
+#define COOLKEY_INS_GET_RANDOM                 0x72
 #define COOLKEY_INS_READ_OBJECT                0x56
 #define COOLKEY_INS_WRITE_OBJECT               0x54
 #define COOLKEY_INS_LOGOUT                     0x61


Jakuje:
Additionally, this card seems to support the SeedRandom, but this interface is not exposed in the opensc drivers (pkcs11 module returns not supported). Are there any other cards that might support that? Would it make sense to implement that?

frankmorgner:
I don't think this even makes sense to be supported by coolkey at all, because it should only affect PRNG, while coolkey only seems to use the true RNG

mouse07410:
@Jakuje and @frankmorgner I wonder if the CAC/PIV TRNG failure can be addressed in a similar way (by a similar patch)? Currently, TRNG on Yubikey tokens works with PIV driver, but on CAC it doesn't (and I'm not sure if I want to - or even if I can - force CAC driver for the CAC cards).

Jakuje:
@mouse07410 Which failure you have on your mind?

mouse07410:
This works:
```
$ OPENSC_DRIVER=cac pkcs11-tool --module /Library/OpenSC/lib/opensc-pkcs11.so --slot 0 --test --login
Logging in to "B..........".
Please enter User PIN: 
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (CAC ID Certificate) 
  testing signature mechanisms:
. . . . .
    RSA-PKCS: OK
  testing key 3 (CAC Cert 4)  -- can't be used to decrypt, skipping
No errors
```

This doesn't:
```
$ pkcs11-tool --module /Library/OpenSC/lib/opensc-pkcs11.so --slot 0 --test --login
Logging in to "B..........".
Please enter User PIN: 
  ERR: C_GenerateRandom failed: CKR_DATA_INVALID (0x20)
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
. . . . .
  testing key 2 (KEY MAN key) 
    RSA-X-509: OK
    RSA-PKCS: OK
1 errors
```

PIV driver fails to provide the correct instruction to the CAC card for `--generate-random` command.

It works OK for Yubikey:
```
$ pkcs11-tool --module /Library/OpenSC/lib/opensc-pkcs11.so --slot 0x4 --generate-random 4
+??t
$
```

dengert:
NIST 800-73-3 PART 2 does not define directly a random number generate function. But it does define in Section " 3.2.4 GENERAL AUTHENTICATE Card Command" following  Table 6 "+
'81 00' Returns '81 TL <random>' (as per external auth example)" a way to get random data from a card. 

Appendix A in Table 11  uses the above  ADPU '00 87 00 9B 04 7C 02 81 00' to "Client application requests a challenge from the PIV Card Application"  Note that this uses key "9B". 

The "9B" key (usually 3DES or AES) is for administering the card and vendors are not required to provide a "9B" key, as they may use other methods to administer the card. 

So your statement: "PIV driver fails to provide the correct instruction to the CAC card for --generate-random command."  could be rewritten to say a CAC card with a PIV applet does not support the use of a "9B" key for "GENERAL AUTHENTICATE".

There is no clear cut way for the PIV driver to determine if the card will or will not support this without trying it, which it does not.   

The Yubikey  has a "9B" key so it works. 

There was some discussion in a previous issue about how much random data could be returned and how many times this APDU could be issued.  Some cards appear to have no limitation, some appear to only allow it to be called only once and some not at all if they don't have a "9B" key. 

So using this feature as a true random number generator is not clear. Maybe should be dropped from the PIV driver.

mouse07410:
@dengert, I hear you. And I agree with how you rewrite the issue.

Since this feature works at least for Yubikey (and for YubiHSM), I strongly suggest that we keep it.

Since CAC dues support this feature, but via different command - perhaps it would be possible to accommodate did it in the PIV driver? Like we do for buggy hardware? Try the correct (9B) way, if it didn't work - try the CAC way, if that fails too - return error...?

mouse07410:
What's a good way of forcing tokens to use the CAC driver for a dual card?

Jakuje:
As you already pointed out, using `PENSC_DRIVER=cac` or configuration option in `opensc.conf`

Jakuje:
ok, I am not sure about the tokend, but it should also follow the opensc.conf, isn't it?

mouse07410:
>  I am not sure about the tokend...

And tokend is my primary concern, as it has to serve both PIV applet of Yubikey tokens and HID cards, and CAC/PIV applets of CAC cards.

`OPENSC_DRIVER=cac` or similar can only work for a manual (or script-based) invocation of a tool like `pkcs11-tool`.

And without a reasonable solution here, I'd much prefer a hack to the PIV driver, making `GENERAL AUTHENTICATE` try one more thing **if the "normal" 9B fails**.

dengert:
I agree with https://github.com/OpenSC/OpenSC/pull/1441#issuecomment-410976381 In addition to Yubico and HID PIV cards, there are other open source PIV applets and other card vendors such as the Taglio https://pivkey.com/  which says it is supported by OpenSC and other middleware. 

RNG is one of the differences. PIV has a Auth Key and cert.  As best I can tell, CAC does not have this.  So there are times a user may want to use the PIV applet  to use the AUTH key and may want to use the CAC applet for the RNG. 

I still have on the back burner: https://github.com/dengert/OpenSC/commit/97a3144a43c737b2703095649d65191fcaa7dcf1 which tries to address access to multiple applets on the same card, from a single PKCS\#11 application.  So each applet has its own driver. "For example OpenPGP and PIV or PIV and CAC."

But I never got any feedback on this from people who I thought were interested in doing this. (@mouse07410 and @Jakuje ) I don't have any CAC cards, so I can not do much testing.

mouse07410:
> PIV has a Auth Key and cert. As best I can tell, CAC does not have this

It seems to me that the two latest CACs I had both contained AUTH cert (or perhaps the older one has ID cert - wouldn't that be the same thing?).

> So there are times a user may want to use the PIV applet to use the AUTH key and may want to use the CAC applet for the RNG

That would be nice - unless it requires some torture-coding of the app...

> So each applet has its own driver. "For example OpenPGP and PIV or PIV and CAC."

What should the application know about multiple applets? The only example of such an app that I can think of would be Apple Mail with GPGTools installed (so it can process both S/MIME and PGP/MIME emails).

dengert:
Look at #1436 where the user says: "I think some users might not have PIV Auth, only DS and KE, so maybe that's why they request DS for authentication." This makes it sound like a CAC card. But I don't know the version.  NIST 800-73-3 Part 4 Table 1 says  "X.509 Certificate for PIV Authentication" is Mandatory So it makes it look less like a PIV card, but a CAC card which does not provide a PIV Auth key.

The Table 1 also lists Container ID that appear to also be listed in [Nice Picture](http://www.cac.mil/Portals/53/Documents/ref5-Technical_Bulletin_CAC_2_6_2b_Applet_Structure_June_2009.pdf)  

Container ID 0x0101  in PIV is "X.509 Certificate for PIV Authentication" and in CACv2 "PKI Sign" 
Container ID 0x0100 in PIV is "X.509 Certificate for Digital Signature" and in CACv2 "PKI ID"
Container ID 0x0102 in PIV is "X.509 Certificate for Key Management" and in CACv2 "PKI Enc"
Container ID 0x0500 is only in PIV as "X.509 Certificate for Card Authentication"
Container ID 0xA001 is only in the CAC card on the right as "PIV Auth"

Thus there is a lot of confusion as to "X.509 Certificate for PIV Authentication" is not the same as  "PIV Auth"?

I don't have any CAC cards and coming from a DOE lab, we did not use CAC only PIV. So what DOD did to implement a PIV applet (multiple versions it looks like)  for CAC has not be an interest of mine. The OpenSC PIV driver is intended to implement what is documented in NIST 800-73 -x.

mouse07410:
> Look at #1436 where the user says: "I think some users might not have PIV Auth, only DS and KE, so maybe that's why they request DS for authentication." This makes it sound like a CAC card.

Yeah, but... It also depends on what the actual web site is. For example, DEE Mail *requires* that authentication is done by the DS cert/key. It rejects the PIV Auth if that one is chosen.

> The OpenSC PIV driver is intended to implement what is documented in NIST 800-73 -x.

Yeah, I understand and respect this decision. Though if the quirk is small, perhaps it makes sense to extend the driver capability a bit to accommodate faulty/incomplete/etc card implementations.

PR #1442, opened 2018-07-30
Add supported algorithms for OpenPGP Card (Fixes #1432)

### LINKED ISSUE #1432: pkcs15-init: key generation and import for OpenPGP Card with different algorithms
### Problem Description

It is currently not possible to use `pkcs15-init` for generating and/or importing keys with attributes which are not i) the default (rsa/2048) or ii) previously set with other tools.

Thus, if one wants to generate e.g. a rsa/4096 key on a factory-reset OpenPGP Card, pkcs15-init will fail with:
```
$ pkcs15-init --delete-objects privkey,pubkey --id 3 --generate-key rsa/4096 --auth-id 3 --verify
Using reader with a card: Nitrokey Nitrokey Pro (0000320E0000000000000000) 00 00
User PIN required.
Please enter User PIN [Admin PIN]: 
NOTE: couldn't find privkey 03 to delete
NOTE: couldn't find pubkey 03 to delete
Deleted 0 objects
Failed to generate key: Not supported
```
As a workaround users need to generate a similiar key with `openpgp-tool` first, to make sure that the correct algorithm is set. (e.g. `openpgp-tool --verify CHV3 --pin 12345678 --gen-key 3 --key-len 4096`. The import/creation of keys with pkcs15-init is possible afterwards.

The algorithm at stake is checked [here](https://github.com/OpenSC/OpenSC/blob/4de0d06a93d384b91753458e7e63b5eb6a854fbb/src/pkcs15init/pkcs15-lib.c#L2325) and [here](https://github.com/OpenSC/OpenSC/blob/4de0d06a93d384b91753458e7e63b5eb6a854fbb/src/pkcs15init/pkcs15-lib.c#L2363) for a list of compatible algorithm for this card. The problem is, that this list is not correct. Instead it seems to be a known default (rsa/2048) and the currently set value as extracted through [this code](https://github.com/OpenSC/OpenSC/blob/4de0d06a93d384b91753458e7e63b5eb6a854fbb/src/libopensc/card-openpgp.c#L722). Other than stated in the comment, it does not extract supported available algorithm, but the currently set.

Please read 4.4.3.7 of [the specification](https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.3.pdf) which states

> The content of the DO is optionally changeable (announced in Extended Capabilities) with
PUT DATA. This is useful if the card supports several algorithms or different key length.
The attributes can be changed independent for each key, so it is possible for example to
use different key length for signing and decrypting. A card should reject unsupported val-
ues in the DO. The supported values are manufacturer specific, please ask your card
vendor for the correct parameters if the card supports changing of the attributes.

### Proposed Resolution
I think pkcs15-init either needs a correct list of all compatible algorithm for the cards or should just try to set the requested algorithm.

The second option appears to be already done by `openpgp-tool`. It works just fine because of the fact that a card should (and does) reject incompatible algorithms anyway. For example, a OpenPGP Card v3 does not accept rsa/1024 (other than v2), thus the output is:
```
openpgp-tool --verify CHV3 --pin 12345678 --gen-key 3 --key-len 1024
Using reader with a card: Nitrokey Nitrokey Pro (00005F120000000000000000) 00 00
Failed to generate key. Error Incorrect parameters in APDU.
```
I'd love to create a PR, but I had problems understanding the code to realize the first approach and the second approach is related to checks in pkcs15-libs.c that used by too many cards. I don't dare to touch this and I'd be happy about tips how I could fix this limitation :-)

PS: This issue is in some ways related to #1011

alex-nitrokey:
Tested with Nitrokey Pro (OpenPGP Card v2), Nitrokey Pro 2 (OpenPGP Card v3) and Nitrokey Start (Gnuk based).

Currently only used algorithms are list in card capabilities of OpenPGP Card based devices, instead of listing all supported algorithms. Now all possible RSA algorithm types are stored.

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -59,7 +59,7 @@ static const char default_cardname_v3[] = "OpenPGP card v3.x";
 static struct sc_atr_table pgp_atrs[] = {
 	{ "3b:fa:13:00:ff:81:31:80:45:00:31:c1:73:c0:01:00:00:90:00:b1", NULL, default_cardname_v1, SC_CARD_TYPE_OPENPGP_V1, 0, NULL },
 	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:c5:73:c0:01:40:00:90:00:0c", NULL, default_cardname_v2, SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
-	{ "3b:da:11:ff:81:b1:fe:55:1f:03:00:31:84:73:80:01:80:00:90:00:e4", NULL, "Gnuk v1.0.x (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_GNUK, 0, NULL },
+	{ "3b:da:11:ff:81:b1:fe:55:1f:03:00:31:84:73:80:01:80:00:90:00:e4", NULL, "Gnuk v1.x.x (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_GNUK, 0, NULL },
 	{ "3b:fc:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:4e:45:4f:72:33:e1", NULL, "Yubikey NEO (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ "3b:f8:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:34:d4", NULL, "Yubikey 4 (OpenPGP v2.1)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:f5:73:c0:01:60:00:90:00:1c", NULL, default_cardname_v3, SC_CARD_TYPE_OPENPGP_V3, 0, NULL },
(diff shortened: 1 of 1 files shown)


hongquan:
Look good to me.

frankmorgner:
would it be possible to modify `src/pkcs15init/pkcs15-openpgp.c` to add the missing algorithms?

alex-nitrokey:
> Sorry, but from the specification I don't understand your notion of current.

Well, it just does not detect the algorithms which _can_ be used on the card, but only which _are_ configured _right now_. Therefore it is not really useful to look for these only. In fact, using `opensc-tool --list-algorithms` will output rsa2048 three times on a factory-reset device, just because rsa2048 is configured for all three slots of the card. Every other possible algorithm will not be listed.

Please see 4.4.3.7 of the specification which states

> The content of the DO is optionally changeable (announced in Extended Capabilities) with PUT DATA. This is useful if the card supports several algorithms or different key length. The attributes can be changed independent for each key, so it is possible for example to use different key length for signing and decrypting. A card should reject unsupported values in the DO. The supported values are manufacturer specific, please ask your card vendor for the correct parameters if the card supports changing of the attributes.

You are right that it is up to the manufacturer what algorithms are actually implemented. But as a) we don't have a proper way to find out the supported algorithm without just _testing_ them, b) the card is supposed to refuse a algorithm it does not support anyway and c) these still seem to be the regularly supported algorithms, I would prefer to have these implemented than leave the user alone with a workaround to do the exact thing manually.

In fact, openpgp-tool already does this: it just _assumes_ that an algorithm is working, sends the information to the card and if it refuses to use this one, the user gets a proper error message. pkcs15-init on the other checks the card->algorithms first (which only includes these already configured, as stated above) and therefore fails.

> would it be possible to modify src/pkcs15init/pkcs15-openpgp.c to add the missing algorithms?

I don't see how this would be different to this approach? This would be an assumption about the supported algorithms, too. And it would not apply to other tools which may access the same information (card->algorithms). Or did I misunderstood?

Please let me know, if anything is unclear or you have another idea. This is the best I could come up with. As far as I can see there just is not extensive list on card...

marschap:
@frankmorgner alex-nitrokey is right: according to my reading of the specs there is no on-card information about the algorithms supported by the card.
The PR looks good to me too. 
Please be so kind to pull it into master.

frankmorgner:
The problem I see is that in PKCS#11 all of the registered (but possibly unsupported) algorithms will be populated to the application:
https://github.com/OpenSC/OpenSC/blob/336b282324a9990f94da7ab18a7dbd6957bd9e58/src/pkcs11/framework-pkcs15.c#L4955-L4982

Please just try to make a simple comparison for your hack whether you're actually running inside `pkcs15-init`. It should look similar to this:
https://github.com/OpenSC/OpenSC/blob/4de0d06a93d384b91753458e7e63b5eb6a854fbb/src/pkcs11/misc.c#L450-L454

alex-nitrokey:
> The problem I see is that in PKCS#11 all of the registered (but possibly unsupported) algorithms will be populated to the application:

I don't want to be too insisting here (thus, I added the requested limitation), but can you please elaborate why this could be a problem? 

Having the restriction to pkcs15-init implemented, the user can not use this code for e.g. listing all supported algorithms with ```opensc-tool --list-algorithms``` and may will get problems with other use cases as well. Honestly, I think this way of assuming algorithms is in most cases more accurate than just listing currently used ones.

I don't like the situation either. I just feel like this makes it a bit better. Of course I will totally accept your decision ;-)

alex-nitrokey:
Okay, thanks.

alex-nitrokey:
I just realized, that I used the wrong indentation, shall I fix this minor problem before this get merged?

frankmorgner:
Yes, please fix this and we're good to merge.

Issue #1449, opened 2018-08-15
Install replaces opensc.conf with a minimal opensc.conf #1102


dengert:
### Problem Description
 
Install in previous versions would install `opensc.conf.new`. Now with  c003f3825e5217cd3ef8272f4e8ebe47270a7847 it installs a minimal `opensc.conf` overwriting the existing `opensc.conf` with any changes the user may have made. (I spent all afternoon trying to figure out what happened.)

See #1102 
 
### Proposed Resolution

At a minimum, fix the code in  c003f3825e5217cd3ef8272f4e8ebe47270a7847 that tested for an existing opensc.conf in `etc/Makefile.am` and install the minimal `opensc.conf` as opensc.conf.new` Look at the removed test in:
https://github.com/OpenSC/OpenSC/commit/c003f3825e5217cd3ef8272f4e8ebe47270a7847#diff-3392c95f52e00a39c33e6f4ce45b54a3L35 
`if [ -f "$(DESTDIR)$(sysconfdir)/opensc.conf" ]; then \`

Another solution is to revert  c003f3825e5217cd3ef8272f4e8ebe47270a7847 and  provide a script to remove all the comments and empty stanza formatting to produce a minimal `opensc.conf` from a user edited version of a full `opensc.conf` or `opensc.conf.example` with user changes.

This may also require many of the lines in `opensc.conf.example` that are not comments  such as `card_atr` or `secure_messaging` stanzas. Users of these cards can uncomment them as needed.  

Then an additional way to address #1102 is to change `scconf_parse` to not allocate memory for stanzas that contain only comments.

Jakuje:
What platform install you refer to? For example on Fedora/RHEL with RPMs, this is not going to happen. If you update opensc with not modified .conf file, you will get a new one, but if you modified this configuration before, the new file will have suffix `.rpmnew`. Nothing will be lost and one can merge things manually (clean up in this case). If I remember well, there is similar mechanisms in dpkg. I am not sure if there is still anyone out there installing with `make install` only in production systems.

dengert:
I am referring to make install which used to install  a new opensc.conf as opensc.conf.new if  opensc.conf existed.  Luckily most distributions do a much better job of handling configuration files.

dengert:
[...]

PR #1453, opened 2018-08-20
Opensc.conf processing

dengert:
This is response to comment https://github.com/OpenSC/OpenSC/issues/1449#issuecomment-414085563
for issues #1449 and #1102

Two commits:
One makes sure an user's opensc.conf is not over written in the sysconfdir.
The second causes scconf_parse to not save comments in memory and to remove empty blocks which can address the memory issues in #1102 

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
I don't think changing some bits here and there will make scconf more maintainable. Since nobody really maintains scconf (and there are coverity issues about it), it should be replaced with something different.

Using a small default configuration is better from a user's perspective and it solves the memory issues. So unless there's a real alternative, I'd like to leave scconf as is.

PR #1454, opened 2018-08-20
Fix card-piv.c use of sc_asn1_read_tag

dengert:
Revert  the change to card-piv.c in 8fe377e93b4b56060e5bbfb6f3142ceaeca744fa

card-piv.c uses sc_asn1_read_tag to get the length of the expected
data to be read from the card so it can allocate buffer to read the data.

It does this by seting the length to 0xffff.
8fe377e93b4b56060e5bbfb6f3142ceaeca744fa casused  the length to be set to zero
which would always fail.

Please enter the commit message for your changes. Lines starting
 Changes to be committed:
	modified:   card-piv.c

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -479,6 +479,7 @@ static int piv_general_io(sc_card_t *card, int ins, int p1, int p2,
 	size_t * recvbuflen)
 {
 	int r;
+	int r_tag ;
 	sc_apdu_t apdu;
 	u8 rbufinitbuf[4096];
 	u8 *rbuf;
@@ -572,16 +573,12 @@ static int piv_general_io(sc_card_t *card, int ins, int p1, int p2,
 		 * the buffer is bigger, so it will not produce "ASN1.tag too long!" */
 
 		body = rbuf;
-		if (sc_asn1_read_tag(&body, rbuflen, &cla_out, &tag_out, &bodylen) !=  SC_SUCCESS
+		r_tag = sc_asn1_read_tag(&body, apdu.resplen, &cla_out, &tag_out, &bodylen);
+		sc_log(card->ctx, "r_tag:%d body:%p", r_tag, body);
+		if ( (r_tag != SC_SUCCESS && r_tag != SC_ERROR_ASN1_END_OF_CONTENTS)
 				|| body == NULL)  {
-			/* only early beta cards had this problem */
-			sc_log(card->ctx, "***** received buffer tag MISSING ");
 			body = rbuf;
-			/* some readers/cards might return 6c 00 */
-			if (apdu.sw1 == 0x61  || apdu.sw2 == 0x6c )
-				bodylen = 12000;
-			else
-				bodylen = apdu.resplen;
+			bodylen = apdu.resplen;
 		}
 
 		rbuflen = body - rbuf + bodylen;
@@ -856,6 +853,7 @@ static int piv_read_obj_from_file(sc_card_t * card, char * filename,
 	u8 **buf, size_t *buf_len)
 {
 	int r;
+	int r_tag;
 	int f = -1;
 	size_t len;
 	u8 tagbuf[16];
@@ -881,10 +879,11 @@ static int piv_read_obj_from_file(sc_card_t * card, char * filename,
 		goto err;
 	}
 	body = tagbuf;
-	if (sc_asn1_read_tag(&body, 0xfffff, &cla_out, &tag_out, &bodylen) != SC_SUCCESS
+	r_tag = sc_asn1_read_tag(&body, len, &cla_out, &tag_out, &bodylen);
+	if ((r_tag != SC_SUCCESS && r_tag != SC_ERROR_ASN1_END_OF_CONTENTS)
 			|| body == NULL) {
 		sc_log(card->ctx, "DER problem");
-		r = SC_ERROR_INVALID_ASN1_OBJECT;
+		r = SC_ERROR_FILE_NOT_FOUND;
 		goto err;
 	}
 	rbuflen = body - tagbuf + bodylen;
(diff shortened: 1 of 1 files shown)


dengert:
Your mod of replacing the 0xffff with rbuflen makes things worse. `rbuflen= 0;`  from a few lines above. So this PR reverts that to what has been working for years. 

`sc_asn1_read_tag` is assuming that it has the whole TLV, but in the PIV code it may only have the first 8 bytes of the TLV . 

In order to determine the size of an object, the code does a GET_DATA for the first 8 bytes. It then determines the size and will get a buffer then use GET_DATA to read the full object. 

So I will replace `sc_asn1_read_tag` with some code in card-piv.c which can parse just the tag and length.

But 0.19.0 needs either this PR or the replacement I will submit.  If   https://github.com/OpenSC/OpenSC/commit/8fe377e93b4b56060e5bbfb6f3142ceaeca744fa is included the way it stands some PIV cards will not work.

dengert:
I think I can use sc_asn1_read_tag and am working on replacement code for 3 places that use 0xffff. I should have  updated PR today or tomorrow.

dengert:
See the new commit that address the use of sc_asn1_read_tag when only the beginning of the block has been read.

src/libopensc/card-piv.c
```diff
@@ -880,11 +878,14 @@ static int piv_read_obj_from_file(sc_card_t * card, char * filename,
...
-	if (sc_asn1_read_tag(&body, 0xfffff, &cla_out, &tag_out, &bodylen) != SC_SUCCESS
+	r_tag = sc_asn1_read_tag(&body, len, &cla_out, &tag_out, &bodylen);
+	if ((r_tag != SC_SUCCESS && r_tag != SC_ERROR_ASN1_END_OF_CONTENTS)
 			|| body == NULL) {
 		sc_log(card->ctx, "DER problem");
-		r = SC_ERROR_INVALID_ASN1_OBJECT;
+		r = SC_ERROR_FILE_NOT_FOUND;
 		goto err;
 	}
 	rbuflen = body - tagbuf + bodylen;
```
frankmorgner: If I'm not mistaken then `bodylen` may not correctly initialized if `r_tag == SC_ERROR_ASN1_END_OF_CONTENTS`. The simplest would be to **not** try to cope with `SC_ERROR_ASN1_END_OF_CONTENTS`.

src/libopensc/card-piv.c
```diff
@@ -567,21 +567,19 @@ static int piv_general_io(sc_card_t *card, int ins, int p1, int p2,
...
 			body = rbuf;
-			/* some readers/cards might return 6c 00 */
-			if (apdu.sw1 == 0x61  || apdu.sw2 == 0x6c )
-				bodylen = 12000;
-			else
-				bodylen = apdu.resplen;
+			bodylen = apdu.resplen;
 		}
 
 		rbuflen = body - rbuf + bodylen;
```
frankmorgner: See problem with `SC_ERROR_ASN1_END_OF_CONTENTS` in the other comment.

frankmorgner:
@dengert I'd like to make the new release as soon as possible. How fast, do you think can you look at the comments for this PR?

src/libopensc/card-piv.c
dengert: At this point card-piv.c has only read the first few bytes, enough to read the tag and length. 
The point is to get the size of the full object into `rbuflen` so a buffer can be allocated then do a secondGET_DATA command to read the full object.

src/libopensc/card-piv.c
```diff
@@ -880,11 +878,14 @@ static int piv_read_obj_from_file(sc_card_t * card, char * filename,
 		r =  SC_ERROR_DATA_OBJECT_NOT_FOUND;
 		goto err;
 	}
+	int r_tag;
```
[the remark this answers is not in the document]
dengert: Minor, can wait till next release.

dengert:
The way I read `sc_asn1_read_tag`  the second parameter `buflen` is used for two purposes.
(1) tag and length must be found within `buflen` bytes and be valid. If not, `*buf == NULL` and  return `SC_ERROR_INVALID_ASN1_OBJECT`
(2) if the tag and  length are valid, it test to see if the value could also fit within the `buflen` if not 
return `SC_ERROR_ASN1_END_OF_CONTENTS` in either case `*buf` is set to where the value would start. 

Previous versions of `sc_asn1_read_tag` did not do (2). So to avoid failure, card-piv.c was setting 0xffff, as 
in this case all it wanted was the size of the object.  

**The only way to determine the size of a PIV object is to read it using GET_DATA.** So the card-piv.c would read the first 8 bytes which would have the tag and length It would then use `sc_asn1_read_tag` to get the expected size of the object, allocate buffers then do another GET_DATA` command to read the whole object.   

**I think this PR is ready now it addressed your issue of using `0xffff` for buflen.** 

The way I would redesign this is to add an optional  callback to the card driver in the apdu routines that could look at the tag and length and reallocate the apdu.resp and apdu.resplen before any additional `GET_RESPONSE` commands are done. This would be optional and only if some SC_APDU_FLAGS_* was set to do this.  This would eliminate the double reads, and a log of the code in card-piv.c and move it to the callback in piv-callback routine. 

But this is to complicated to do in the next release. But I won't start looking at it unless there is a chance it will be accepted. 

As some PIV objects like a Iris scan or picture can be quite large, I do not want to use fixed size buffers, but allocate buffers of the correct size.

src/libopensc/card-piv.c
frankmorgner: I see, OK

src/libopensc/card-piv.c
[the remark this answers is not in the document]
frankmorgner: No. Please provide plean patches! There is also a white space error at https://github.com/OpenSC/OpenSC/blob/17f74fee0f17a00b30e4ba0b7175214d96911c19/src/libopensc/card-piv.c#L960

frankmorgner:
Yes, redisign is not quickly done, and should be postponed.

dengert:
Did a fixup for two  `int rtag;` lines and the blank at end of like 960, and pushed with -f to my directory.

PR #1457, opened 2018-08-22
Clean up Openssl deprecations

plaes:


frankmorgner:
Considering that everything older than 1.0.2 is not longer supported by the OpenSSL team, this goes into the right direction. However, I'm not sure whether some distributions still use (and support) older versions of OpenSSL. What are other other people from @OpenSC/maintainers think about this?

Note, that there is a merge conflict...

plaes:
OK, merge conflicts fixed, and also improved some of the openssl checks (the two final commits).
And ff 1.0.2 is too "new", I can rework this branch. Just let me know.

dengert:
I agree with Frank, we don't know who may still be using older versions. We could do this it 2 steps,
saying the use of any version less than 1.0.2 is deprecated with 591e4914ee7517fef7be071b690efe13d5a15d4d  in next release which would allow a users to override the openssl_minimum_version if they really had to. 

Then in the release after the next release add the other commits. 

Have you tried this code against libressl? There are some people who use it.

plaes:
Ok, I pushed another branch with cleanups up to 0.9.8 and I will rebase this branch on top of that eventually.

plaes:
Rebased on top of latest master.  @OpenSC/maintainers - Feel free to cherry-pick something.

@dengert Nope, I haven't tried against libressl, if someone uses it and something is broken, I guess we can expect them to provide patches.

martinpaljak:
IMHO a good thing, who is affected will probably scream and also provide a localized (possibly downstream) fix.

Jakuje:
I guess it is time to get rid of the old cruft. Soon, there will be no other OpenSSL version supported than 1.1.1, since also the 1.0.2 is going away:

> Our previous LTS version (1.0.2 series) will continue to be supported until 31st December 2019

https://www.openssl.org/source/

Can you rebase the changes on top of current master so they apply?

plaes:
> I guess it is time to get rid of the old cruft. Soon, there will be no other OpenSSL version supported than 1.1.1, since also the 1.0.2 is going away:
> ...
> Can you rebase the changes on top of current master so they apply?

Rebased :)

plaes:
@metsma: I have patches up to 1.0.2 ready, but for some reason, travis's GCC job fails, so I'm going to push patches in again one-by-one...

metsma:
You can remove also ssl requirement and link only with crypto and remove the m4 define openssl_minimum_version
https://github.com/OpenSC/OpenSC/blob/08af8a23f5483c666441f6656b034e6dc6e9312d/configure.ac#L630

plaes:
OK, travis/gcc starts to fail after commit 2d81b3b - `openssl: Bump openssl requirements to 1.0.2`.

plaes:
Ubuntu 14.04 (in travis) comes with `openssl-1.0.1f`, so all the patches up to 1.0.1 are ok to use with current travis setup.

plaes:
I moved the clean patches that do not require travis updates to pull #1658 .

frankmorgner:
We require 1.0.1 for now. At some point we will switch to 1.1.1

PR #1463, opened 2018-08-30
Add support for SmartCard-HSM 4K (V3.0)

CardContact:
The upcoming SmartCard-HSM 4K (V3.0) product release will support RSA keys up to 4096 bits and ECC keys with up to 521 bits.

This patch adds the new ATR and fixes various issues with too-short buffers.

Tested with our own PKCS#11 test suite and p11tests.


src/libopensc/card-sc-hsm.c
```diff
@@ -1158,11 +1157,11 @@ static int sc_hsm_decipher(sc_card_t *card, const u8 * crgram, size_t crgram_len
 	LOG_FUNC_CALLED(card->ctx);
 	priv = (sc_hsm_private_data_t *) card->drv_data;
 
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_4, 0x62, priv->env->key_ref[0], priv->algorithm);
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_EXT, 0x62, priv->env->key_ref[0], priv->algorithm);
 	apdu.cla = 0x80;
 	apdu.resp = rbuf;
 	apdu.resplen = sizeof(rbuf);
-	apdu.le = 256;
+	apdu.le = 512;
```
frankmorgner: what happens if the reader doesn't support this length?

CardContact: The SmartCard-HSM requires extended length support for various other use cases as well. Readers without extended length are not supported.

Extended length was defined 20 years ago, so plenty of time for hardware vendors to adapt.

I would appreciate to have support for the new version in 0.19 as shipment will start shortly.

CardContact: I disagree. Key sizes beyond 2048 always need extended length. If you picked a reader that doesn't support a 20 year old protocol and still want to use keys larger than 2048, then go and get a new reader ;-)

The change is related to the SmartCard-HSM code. I don't remember a single customer issue reported that was based on a reader that only supported short APDUs. So this a rather a theoretical problem.

I know there are issues with mobile devices that use NFC chips and stacks that only support short APDUs. But again thats a vendors decision to implement an incomplete software stack.

Pferdebockwurst:
When will this version finally be available?

CardContact:
Software development is completed and we have a small quantity of devices for integration and testing. As soon as the next order of chips arrive, we will make the next version available at cardomatic - this will hopefully happen in December.

Pferdebockwurst:
Thank you.

PR #1466, opened 2018-08-31
Fix SM algorithm in extended capabilities

alex-nitrokey:
The index for getting the SM algorithm is wrong imho. We may can fix this quick before next release.
@marschap could you please have a look at it?

I did not add a lenght check here, as it is already done in line 730.


marschap:
@alex-nitrokey You're right: the PR fixes a coding glitch.
@frankmorgner Please pull, as this PR fixes a bug in SM algorithm detection for OpenPGP.

frankmorgner:
thanks

PR #1470, opened 2018-09-06
Remove long expired EstEID 1.0/1.1 card support

metsma:

##### Checklist

- [ ] macOS tokend is tested


martinpaljak:
The cards can still be used to decrypt data. IMHO the code reduction is not that huge to justify removal of support, but I would not mind removing this either. Better yet would be to split the esteid parts from mcrd parts, as with esteid3 + nothing relates to micardo platform. This would also allow to remove thoe whole micardo driver, maybe.

metsma:
1.0 last card was issued end of 2010? + 5 years. 2015
1.1 last card was issued 2014 Okt, +3 years. 2017 okt (they where revoked earlier because 1024 RSA key)

frankmorgner:
Currently, we are disabling old card driver by configuration. The plan is to disable those card drivers at compile time in some later release, see https://github.com/OpenSC/OpenSC/wiki/Removal-of-Old-Card-Drivers#rationale-for-removing-a-card-driver. Maybe you can use a compile time switch to disable compilation of the old estid cards.

metsma:
The whole idea was to cleanup some codepaths and conditions not make more additional conditions

PR #1474, opened 2018-09-08
card-gids: un-break compilation on 32bit Debian testing

marschap:
A test of the latest master fails to compile card-gids,c with Debian testing in 32-bit
to to a potential overflow of a snprint target string

This PR fixes the issue by
* increasing the target string size
* limiting the value of the argument passed to the snprintf() call to a sensible upper limit

[x] compilation tested on Debian testing in 32- and 64-bit.
I cannot test more, because I do not have a GIDS card.

Fun fact: same configuration on 64-bit compiled without any problems.

### DIFF
--- a/src/libopensc/card-gids.c
+++ b/src/libopensc/card-gids.c
@@ -1092,7 +1092,7 @@ gids_select_key_reference(sc_card_t *card, sc_pkcs15_prkey_info_t* key_info) {
 	PCONTAINER_MAP_RECORD records = (PCONTAINER_MAP_RECORD) data->cmapfile;
 	size_t recordsnum;
 	int r;
-	char ch_tmp[10];
+	char ch_tmp[16];
 
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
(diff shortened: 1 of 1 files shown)


frankmorgner:
I don't understand why this would give a compiler warning...

dengert:
I would bet that the compiler looks at `"3FFFB0%02X"` and assuming a 4 byte integer and the "X" format so it assumes it needs 8 characters to display the integer plus the "3FFFB0" or 6+8 = 14. i.e. compiler does not look at the 02 in the format. 

In the original cases  both `key_info->key_reference` `containernum` are `int`.
In the fixed up case both  use `0xFF & (u8) (...)`  this may still result in `int` or `unsigned int` 
maybe if  they used `(u8)(0xFF & containernum)` and `(u8) (0xff & key_info->key_reference) `  it could fit in 10 bytes.

frankmorgner:
@marschap please try rewriting as suggested by Doug instead of increasing the buffer.

sander85:
> maybe if they used `(u8)(0xFF & containernum)` and `(u8) (0xff & key_info->key_reference)` it could fit in 10 bytes.

This made it compile on 32-bit Mageia (cauldron / gcc 8).

PR #1480, opened 2018-09-14
Missing header release tarball

Jakuje:
The 0.19.0 release tarball is missing `p11test_common.h` file. The workaround is to configure without tests (`--disable-tests`).


frankmorgner:
cmocka should be missing in most cases, so tests would often not be built, I guess

PR #1484, opened 2018-09-21
pkcs15-tool: Add --list-public-key <id> argument

D4ryus:
Add option to print info about a single public key identified by its id.
Will print same Information as --list-public-keys does, while
restricting the output to the specified object.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
I think this doesn't make sense.

PR #1489, opened 2018-09-27
Another round of coverity review

Jakuje:
There was another run of coverity, that I shoveled through today. Attached are minor fixes that either make the code more straightforward for the static analyzers or fix some real (but minor) problem.

##### Checklist

- [X] PKCS#11 module is tested (PIV, CAC, coolkey, cardos)
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -1846,11 +1846,6 @@ epass2003_process_fci(struct sc_card *card, sc_file_t * file, const u8 * buf, si
 				case 0x04:
 					file->ef_structure = SC_FILE_EF_LINEAR_FIXED;
 					break;
-				case 0x03:
-				case 0x05:
-				case 0x06:
-				case 0x07:
-					break;
 				default:
 					break;
 				}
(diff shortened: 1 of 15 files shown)


src/tools/npa-tool-cmdline.c
```diff
@@ -1695,6 +1702,11 @@ void update_multiple_arg(void *field, char ***orig_field,
     default:
       break;
     };
+    if (*((void **)field) == NULL) {
+      free(old_field);
+      fprintf(stderr, "Failed to allocate memory: aborting");
+      exit(1);
+    }
```
frankmorgner: This is a generated file. The fix needs to applied to [gengetopt](https://www.gnu.org/software/gengetopt/gengetopt.html), not here.

Jakuje: Removed and rebased on current master. I will figure out how to properly fix it in gengetopt. Thanks.

Jakuje:
The AppVeyor build failure does not look related to the changes.

src/tools/npa-tool-cmdline.c
Jakuje: Landed here:
http://savannah.gnu.org/bugs/index.php?54761
Hope it is the right place.

frankmorgner:
thanks

PR #1490, opened 2018-09-27
Fix pkcs15-init for STARCOS SPK 2.3

gabrielmuller:
Upon structure creation, STARCOS 2.3 cards were throwing errors (security status not satisfied). This is because the program failed to verify the SO PIN before certain CREATE FILE operations. This fix changes two things:

* Use "STARCOS" as the default card name for 2.3 instead of outdated "STARCOS SPK 2.3" that was still used in some `strcmp` calls throughout the code;

* Changed `isf_acl` on the starcos profile in order to also require SO PIN for CREATE in addition to WRITE.

This seems to fix the aforementioned bug and I'm able to succesfully create a fresh pkcs15 structure on my card by specifying so-pin and so-puk args. However, I haven't tested STARCOS 2.3 functionality thoroughly.

### DIFF
--- a/src/libopensc/pkcs15-infocamere.c
+++ b/src/libopensc/pkcs15-infocamere.c
@@ -787,7 +787,7 @@ static int infocamere_detect_card(sc_pkcs15_card_t * p15card)
 	sc_card_t *card = p15card->card;
 
 	/* check if we have the correct card OS */
-	if (strcmp(card->name, "STARCOS SPK 2.3")
+	if (strcmp(card->name, "STARCOS")
 			&& strcmp(card->name, "CardOS M4"))
 		return SC_ERROR_WRONG_CARD;
 	return SC_SUCCESS;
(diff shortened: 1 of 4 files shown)


frankmorgner:
please keep the comparison with `"STARCOS SPK 2.3"` in the code. The rest of the code looks OK.

src/libopensc/card-starcos.c
```diff
@@ -146,6 +146,8 @@ static int starcos_init(sc_card_t *card)
 		_sc_card_add_rsa_alg(card,1976, flags, 0x10001);
 		_sc_card_add_rsa_alg(card,2048, flags, 0x10001);
 	} else {
+		card->name = "STARCOS SPK 2.3";
```
frankmorgner: Sorry, I didn't notice that your previous version was actually a fix for a regression introduced with d29bdf9da. So just using fcbcfaa63ed17a01b93070686d435d1c434a08d3 would be preferable

gabrielmuller: I have reset to https://github.com/OpenSC/OpenSC/commit/fcbcfaa63ed17a01b93070686d435d1c434a08d3.

frankmorgner:
thanks

PR #1471, opened 2018-09-07
Build cleanups

alonbl:
Some build fixups.

1. update .gitignore for recent automake test suites
2. remove generated sources out of tree.

PR #1492, opened 2018-09-29
remove source-less doc/tools/files.html

marschap:
The current version, which gets re-generated from XML source, is in doc/files/.


frankmorgner:
See https://github.com/OpenSC/OpenSC/pull/1471

marschap:
@frankmorgner This PR is not about deleting `doc/files/files.htm`, for which there is a source `doc/files/files.xml`) and a rule to re-generate it.
This PR is about `doc/tools/files.html` which 
* has no source, 
* no way to get re-generated 
* is outdated compared to `doc/files/files.html`.

Is **that** file really necessary?

frankmorgner:
thanks for the clearification!

PR #1498, opened 2018-10-07
OpenPGP start towards EC

marschap:
Hi, 

this PR intends to set the first steps towards EC support with OpenPGP cards.
It
* performs minor cleanups
* exposes additional attributes beyond those mentioned in the algorithm attributes DO C1 - C3 only if the extended capabilities indicate that these DOs can be changed
* fixes some minor glitches in the data structures for key generation & import
* extends those data structures to allow for non-RSA algorithms
* adds security checks to some functions to bail out in case of being called with non-RSA algorithms
* introduces a new function introduce `gpg_parse_algo_attr_blob()`
* refactors `pgp_get_card_features()` to make use of `gpg_parse_algo_attr_blob()`

None of these should change anything in the existing logic, but they pave the ground for further PRs that may be more intrusive.

Please include them in OpenSC master.

Thanks in advance
Peter

### DIFF
--- a/src/libopensc/cardctl.h
+++ b/src/libopensc/cardctl.h
@@ -952,24 +952,41 @@ typedef struct sc_cardctl_piv_genkey_info_st {
 typedef struct sc_cardctl_openpgp_keygen_info {
 	u8 key_id;		/* SC_OPENPGP_KEY_... */
 	u8 algorithm;		/* SC_OPENPGP_KEYALGO_... */
-	u8 *modulus;		/* New-generated pubkey info responded from the card */
-	size_t modulus_len;	/* Length of modulus in bit */
-	u8 *exponent;
-	size_t exponent_len;
+	union {			/* anonymous union */
+		struct {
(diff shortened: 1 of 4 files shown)


src/libopensc/cardctl.h
```diff
@@ -942,30 +942,51 @@ typedef struct sc_cardctl_piv_genkey_info_st {
...
+			u8 *q;
+			size_t q_len;
+			u8 *n;
+			size_t n_len;
+		} rsa;
+		struct {
+			u8 dummy;	/* placeholder */
+			// TODO: replace placeholder with real attributes
+		} ec;
+	};
```
alex-nitrokey: I would have called the union 'u' like it was done for [sc_pkcs15_pubkey](https://github.com/OpenSC/OpenSC/blob/6bf67f7917f8d1eda999320aad374ae0ed907835/src/libopensc/pkcs15.h#L240)/prkey to be compatible

PR #1502, opened 2018-10-10
Restore minimal CAC1 driver for legacy cards

Jakuje:
It showed up that the CAC1 protocol is not completely unused yet and there are still some quite new CAC Alt tokens that do not talk any other protocol. I took this opportunity to create a separate cac1 driver which is using some of the cac driver methods and the pkcs15 backend, but can be moved on the bottom of the list or later deprecated in the "old" drivers, while still leaving the CAC2 (which is significantly modified since that time) working.

I tried to share as much as possible from the existing code from CAC2. I also cleaned up some leftovers from the code (needless includes, needless function arguments).

Tested with both cards talking CAC1 and CAC2 protocol.

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
Rebased on top of the current master

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,558 @@
...
+	size_t len, next_len;
+	sc_apdu_t apdu;
+	int r = SC_SUCCESS;
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	/* get the size */
+	size = left = *out_buf ? *out_len : sizeof(buf);
+	out_ptr = *out_buf ? *out_buf : buf;
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, CAC_INS_GET_CERTIFICATE, 0, 0 );
+	next_len = MIN(left, 100);
+	for (; left > 0; left -= len, out_ptr += len) {
```
frankmorgner: I think this needs error checking so that the response isn't written beyond `buf`/`*out_buf`

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,558 @@
...
+		}
+		len = MIN(count, priv->cache_buf_len-idx);
+		memcpy(buf, &priv->cache_buf[idx], len);
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, len);
+	}
+
+	sc_debug(card->ctx, SC_LOG_DEBUG_NORMAL,
+		 "clearing cache idx=%d count=%"SC_FORMAT_LEN_SIZE_T"u",
+		 idx, count);
+	if (priv->cache_buf) {
```
frankmorgner: This check is technically not needed.

src/libopensc/card-cac-common.h
```diff
@@ -0,0 +1,89 @@
...
+} cac_object_t;
+
+/*
+ * CAC private data per card state
+ */
+typedef struct cac_private_data {
+	int object_type;		/* select set this so we know how to read the file */
+	int cert_next;			/* index number for the next certificate found in the list */
+	u8 *cache_buf;			/* cached version of the currently selected file */
+	size_t cache_buf_len;		/* length of the cached selected file */
```
frankmorgner: Caching is a somewhat difficult topic. We already have `struct sc_card_cache cache` in `sc_card_t`, which has led to problems (e.g. #1159) in the past. Also, there is `sc_pkcs15_cache_file()`. Would it be possible to use the global cache for real applications to avoid this complexity here?

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,558 @@
...
+	pathtype = in_path->type;
+
+	sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+		 "path->aid=%x %x %x %x %x %x %x  len=%"SC_FORMAT_LEN_SIZE_T"u, path->value = %x %x %x %x len=%"SC_FORMAT_LEN_SIZE_T"u path->type=%d (%x)",
+		 in_path->aid.value[0], in_path->aid.value[1],
+		 in_path->aid.value[2], in_path->aid.value[3],
+		 in_path->aid.value[4], in_path->aid.value[5],
+		 in_path->aid.value[6], in_path->aid.len, in_path->value[0],
+		 in_path->value[1], in_path->value[2], in_path->value[3],
+		 in_path->len, in_path->type, in_path->type);
```
frankmorgner: use `sc_dump_hex()` and `sc_print_path()`

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,558 @@
...
+	sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "file_out=%p index=%d count=%d\n",
+		 file_out, in_path->index, in_path->count);
+
+	/* Sigh, sc_key_select expects paths to keys to have specific formats. There is no override.
+	 * we have to add some bytes to the path to make it happy. A better fix would be to give sc_key_file
+	 * a flag that says 'no, really this path is fine'.  We only need to do this for private keys */
```
frankmorgner: Feel free to suggest a fix to `sc_key_select` in a separate PR.

src/libopensc/card-cac-common.c
```diff
@@ -0,0 +1,116 @@
...
+	priv = calloc(1, sizeof(cac_private_data_t));
+	if (!priv)
+		return NULL;
+	list_init(&priv->pki_list);
+	list_attributes_comparator(&priv->pki_list, cac_list_compare_path);
+	list_attributes_copy(&priv->pki_list, cac_list_meter, 1);
+	list_init(&priv->general_list);
+	list_attributes_comparator(&priv->general_list, cac_list_compare_path);
+	list_attributes_copy(&priv->general_list, cac_list_meter, 1);
+	/* set other fields as appropriate */
```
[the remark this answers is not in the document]
Jakuje: I will fix this. Thanks.

src/libopensc/card-cac1.c
Jakuje: I do not see how this could overflow. Is the apdu code able to write behind the `LE` boundary? The `apdu.le` is set to `len`, which is set from `next_len`, which properly checks against the maximum bound defined in the `left` variable.

src/libopensc/card-cac1.c
Jakuje: True. I will remove it.

src/libopensc/card-cac-common.h
Jakuje: I do not think that this is used for the same thing as `sc_pkcs15_cache_file()` -- this is just used in case a caller of `cac_read_binary()` would not pick up the certificate in one piece (while we still need it as a whole, because in case it is compressed, we need to decompress it first). We do not use the cache for any subsequent OpenSC invocation to store something in files.
I do not see the use case where this would cause problems. Can you clarify?

src/libopensc/card-cac1.c
Jakuje: That will simplify things. The same is in the cac2 driver so I fill change it also there.

src/libopensc/card-cac1.c
Jakuje: If I am right, the function in question here is `iso7816_select_file()`, which is something that should not take flags to behave differently than the ISO says for various partly incompatible cards.

The above function is mostly simplified and adjusted to the limitation that CAC1 has. But I updated the comment to reflect this.

src/libopensc/card-cac-common.h
frankmorgner: Your implementation looks OK, but introduces complexity that may not be needed. (However, there's still the risk that your internal cache points to the wrong file in concurrent access.)

`sc_pkcs15_read_file` derives the file size from `sc_select_file` and then calls `sc_read_binary`. Hence, the length should always fit the certificate. Is your cache used somewhere?

src/libopensc/card-cac1.c
frankmorgner: I think you're right. But at least there's the possibility that this loops forever, if the card returns 6300, which sets next_len to 0

Jakuje: You are right. Fixed in the next commit.

src/libopensc/card-cac-common.h
Jakuje: Yes, it could happen during threaded access to this function, but I do not think that it is something that opensc supports -- there would be many other places that would be broken by this.

The `sc_pkcs15_read_file()` is different, because it is storing the data in the files among pkcs11 module invocations, if enabled in the configuration. The `cache_buf` is living just among subsequent invocations of the `cac_read_binary()`.

frankmorgner: Without the cache, logic to read out a single file could be as follows:

1. read (a chunk of) a file 0x12 with sc_read_binary
2. read the whole file 0x12 with cac_read_binary
3. decompress the file's content if needed
4. return (a chunk of) the decompressed data to the caller of sc_read_binary

In understand the technical difference and agree that your cache should make the program faster if (and only if) the requested file is read in multiple chunks by the same program. However, I doubt that a thread will actually be spawned while we're reading out a file. So in effect, your cache won't be used.

Also, if a user cares for execution speed, he should turn on the file caching, which would also leave your cache unused.

Just to be clear: I think the current implementation is OK, and could stay as it is. I'm only suggesting a change that I think would make the program a little simpler and easier to maintain.

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,555 @@
...
+	pathlen = in_path->len;
+	pathtype = in_path->type;
+
+	sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+	    "path=%s, path->value=%s path->type=%d (%x)",
+	    sc_print_path(in_path),
+	    sc_dump_hex(in_path->value, in_path->len),
+	    in_path->type, in_path->type);
+	sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "file_out=%p index=%d count=%d\n",
+	    file_out, in_path->index, in_path->count);
```
frankmorgner: use tabs for indenting

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,555 @@
...
+	u8 *val;
+	size_t val_len;
+
+	/* populate PKI objects */
+	for (i = index; i < MAX_CAC_SLOTS; i++) {
+		r = cac_select_pki_applet(card, i);
+		if (r == SC_SUCCESS) {
+			pki_obj.name = get_cac_label(i);
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE,
+			    "CAC: pki_object found, cert_next=%d (%s),", i, pki_obj.name);
```
frankmorgner: use tabs for indenting

src/libopensc/card-cac1.c
```diff
@@ -0,0 +1,555 @@
...
+	if (r >= 0) {
+		priv->cac_id = malloc(20);
+		if (priv->cac_id == NULL) {
+			return SC_ERROR_OUT_OF_MEMORY;
+		}
+#ifdef ENABLE_OPENSSL
+		SHA1(val, val_len, priv->cac_id);
+		priv->cac_id_len = 20;
+		sc_debug_hex(card->ctx, SC_LOG_DEBUG_VERBOSE,
+		    "cuid", priv->cac_id, priv->cac_id_len);
```
frankmorgner: use tabs for indenting

src/libopensc/card-cac-common.h
Jakuje: We have actually a debug log to see what is called when with a real card I have here:
```
$ OPENSC_DEBUG=9 OPENSC_DRIVER=cac1 ./src/tools/pkcs11-tool -L --module ./src/pkcs11/.libs/opensc-pkcs11.so 2>&1 | grep -A 5 "cac_read_binary"
0x7fc524a54880 16:34:10.430 [opensc-pkcs11] card-cac1.c:135:cac_read_binary: called
0x7fc524a54880 16:34:10.430 [opensc-pkcs11] card-cac1.c:152:cac_read_binary: clearing cache idx=0 count=256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:195:cac_read_binary: returning with: 256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:135:cac_read_binary: called
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:141:cac_read_binary: returning cached value idx=256 count=256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:147:cac_read_binary: returning with: 256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:135:cac_read_binary: called
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:141:cac_read_binary: returning cached value idx=512 count=256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:147:cac_read_binary: returning with: 256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:135:cac_read_binary: called
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:141:cac_read_binary: returning cached value idx=768 count=256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:147:cac_read_binary: returning with: 251
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:135:cac_read_binary: called
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:141:cac_read_binary: returning cached value idx=1019 count=256
0x7fc524a54880 16:34:11.198 [opensc-pkcs11] card-cac1.c:147:cac_read_binary: returning with: 0 (Success)
[...]
```
For every file on the CAC card (CAC1 driver in this case), we got 4 separate calls to our function, where the first is server from the actual request to the card and the rest from our cache. This looks like 4x speedup for quite much nothing, which I believe is worth slightly more complex code.

src/libopensc/card-cac1.c
Jakuje: The indentation was changed in the last commit. I can squash them if the that is everything.

PR #1503, opened 2018-10-11
Fix minidriver padding

llogar:
Commit e5707b545e5a2dc33b0ca52a8bf63f36f71b3d85 broke signing using minidriver on Windows.

More specifically changing `#define SC_ALGORITHM_RSA_PAD_NONE` from `0x00000000` to `0x00000001` caused a call to _sc_pkcs1_encode()_ to fail as the padding algorithm was not specified anywhere in the _CardSignData()_ implementation. It kind of worked as long as `SC_ALGORITHM_RSA_PAD_NONE` was `0x00000000`, but the above mentioned commit broke this.

Now padding algorithm has to be explicitly specified, otherwise a call to `sc_pkcs1_encode()` will fail.


frankmorgner:
Good find, thanks :+1:

PR #1506, opened 2018-10-12
OpenPGP Card v3 ECC support

### LINKED ISSUE #1351: OpenPGP Card 3.3: pkcs11-tool does not work with ECC keys
### Problem Description
When using ECC key the `pkcs11-tool -L`  command fails to detect Card.

### Steps to reproduce
Set key usage to brainpoolP256r1 with gpg-connect-agent like this:
```
gpg-connect-agent "SCD SETATTR KEY-ATTR --force 1 19 brainpoolP256r1" /bye
gpg-connect-agent "SCD SETATTR KEY-ATTR --force 2 18 brainpoolP256r1" /bye
```
Restart pcscd, unplug and insert smartcard (in this case: Nitrokey Storage 2 with OpenPGP Card v3.3) and type
```
pkcs-tool -L
```

This results reproducible in:
```
$ pkcs11-tool -L -vvv
Available slots:
Slot 0 (0x0): Nitrokey Nitrokey Storage (0000000000000) 00 00
  manufacturer:  Nitrokey
  hardware ver:  0.0
  firmware ver:  0.0
  flags:         token present, removable device, hardware slot
C_GetTokenInfo() failed: rv = CKR_TOKEN_NOT_PRESENT
```

Any ideas what's happening here and how to fix?

PS: This does not happen if rsa-2048 or alike is used.

alex-nitrokey:
This enables basic ECC support for OpenPGP Card v3 and hereby fixes #1351.

While key generation works without problems, the import is still not working. This seems to be based on missing support in OpenSC (the key itself it rejected) and will be further tested/analysed). This code includes all changes needed for the OpenPGP Card specific parts of OpenSC though.

As everything expected to use RSA keys only in the past, some major changes needed to be made.

The code is tested with OpenPGP Card v2 and v3 and will be further tested by me the next weeks. Thus, this is a first main PR, some minor ones may follow.


src/libopensc/card-openpgp.c
```diff
@@ -2089,8 +2180,10 @@ pgp_decipher(sc_card_t *card, const u8 *in, size_t inlen,
 	/* padding according to OpenPGP card spec 1.1 & 2.x section 7.2.9 / 3.x section 7.2.11 */
 	if (!(temp = malloc(inlen + 1)))
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
-	/* padding byte: 0x00 = RSA; 0x02 = AES [v2.1+ only] */
-	temp[0] = 0x00;
+	if (env->algorithm == SC_ALGORITHM_EC)
```
marschap: maybe better understandable with ternary operator `?:` ?

src/libopensc/card-openpgp.c
```diff
@@ -2161,61 +2254,63 @@ pgp_update_new_algo_attr(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t *key_
...
 
 	LOG_FUNC_CALLED(card->ctx);
 
-	/* temporary workaround: protect v3 cards against non-RSA */
-	if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA)
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
-
-	/* get old algorithm attributes */
-	r = pgp_seek_blob(card, priv->mf, tag, &algo_blob);
+	r = pgp_seek_blob(card, priv->mf, (0x00C0 | key_info->key_id), &algo_blob);
```
marschap: Reason for undoing the consolidation: `tag` is defined exactly as needed.

src/libopensc/card-openpgp.c
```diff
@@ -2283,46 +2373,125 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
...
+		 * and https://tools.ietf.org/html/rfc6637 section 9 (page 8 and 9) */
+		pk_packet_len =   1   /* version number */
+						+ 4   /* creation time */
+						+ 1   /* algorithm */
+						+ 1   /* oid len */
+						+ (key_info->u.ec.oid_len) /* oid */
+						+ (key_info->u.ec.ecpoint_len); /* ecpoint */
+
+		/* KDF parameters for ECDH */
+		if (key_info->algorithm == 0x12) {
```
marschap: SC_OPENPGP_KEYALGO_ECDH ?

src/libopensc/card-openpgp.c
```diff
@@ -2451,23 +2630,34 @@ pgp_parse_and_set_pubkey_output(sc_card_t *card, u8* data, size_t data_len,
 
 		if (tag == 0x0081) {
 			/* set the output data */
-			if (key_info->rsa.modulus) {
-				memcpy(key_info->rsa.modulus, part, len);
+			if (key_info->u.rsa.modulus) {
+				memcpy(key_info->u.rsa.modulus, part, len);
+			}
+			else {
+				sc_log(card->ctx, "Error: key_info->u.rsa.modulus is NULL");
```
marschap: I thought long about that function.
If you do not pass the `modulus` and `exponent` as explicit arguments to other functions, may be it's best to malloc() the necessary memory and copy those elements in any case.
@hongquan What was your intention here?

src/libopensc/card-openpgp.c
```diff
@@ -2451,23 +2630,34 @@ pgp_parse_and_set_pubkey_output(sc_card_t *card, u8* data, size_t data_len,
 
 		if (tag == 0x0081) {
 			/* set the output data */
-			if (key_info->rsa.modulus) {
-				memcpy(key_info->rsa.modulus, part, len);
+			if (key_info->u.rsa.modulus) {
+				memcpy(key_info->u.rsa.modulus, part, len);
```
marschap: The memory allocated might be too small for the modulus generated with the current key.
I am aware that it is an "inherited" issue.

src/libopensc/card-openpgp.c
```diff
@@ -2735,54 +2934,73 @@ pgp_build_extended_header_list(sc_card_t *card, sc_cardctl_openpgp_keystore_info
 		"modulus"
 	};
 	size_t comp_to_add = 3;
-	size_t req_e_len = 0;     /* The exponent length specified in Algorithm Attributes */
-	pgp_blob_t *alat_blob;
+	/* The maximum exponent length is 32 bit, as set on card
+	 * we use this variable to check against actual exponent_len */
+	size_t max_e_len = 0x20>>3;
```
marschap: Magic number; maybe better add a symbolic constant in cardctl.h

alex-nitrokey:
Hi,

thanks a lot for your valuable review. I try to look at all your hints today.

I never meant to be rude and therefore I am truly sorry if my PR looked impolite to you in any way! Please let me shortly explain, why I submitted this PR with the named union (although I knew, that you decided to use a unnamed one).

* I used the named union in the first place, because it was used this way in different locations in the code. Honestly, I have no strong feelings about named/unnamed unions, other than prefering to have it consistent in the code.
* I asked for a change [here](https://github.com/OpenSC/OpenSC/pull/1498#pullrequestreview-162788471), but got no response yet. Instead of already changing all variables in my PR immediately, I decided to propose it here again. For me it is okay to change it, if the project prefers it this way.
* We worked on the same code base at the same time (in fact, I started it long before I knew you would work on it). I wasn't sure, how to address this the best way. Obviously my way wasn't good. But I felt, that my changes went further than yours and therefore proposed it here for discussion.

I would like to either keep the named union as is in this PR or to change it in the whole project. So if the project is okay with me changing the named union in the whole code base (e.g. for sc_pkcs15_pubkey as well), I'd do it.
I even would change it to a unnamed union in this PR only, if you prefer, but I would like this option the least.

Again, I would prefer if you see my actions rather as unprofessional than rude. It's never my intention to upset anyone.

src/libopensc/card-openpgp.c
alex-nitrokey: I am not sure. We only need this value in this function and in [pgp_upgate_new_algo_attr](https://github.com/OpenSC/OpenSC/blob/fc7dba0a40c69e32cb8a0a199a1773f236f44f95/src/libopensc/card-openpgp.c#L2295-L2299). I'd prefer to have a clear comment in these parts, instead of having a constant in cardctl.h.

src/libopensc/pkcs15-openpgp.c
```diff
@@ -296,14 +292,20 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
-			prkey_info.modulus_length = bebytes2ushort(cxdata + 1);
 
 			strlcpy(prkey_obj.label, key_cfg[i].label, sizeof(prkey_obj.label));
 			prkey_obj.flags = SC_PKCS15_CO_FLAG_PRIVATE | SC_PKCS15_CO_FLAG_MODIFIABLE;
 			prkey_obj.auth_id.len      = 1;
 			prkey_obj.auth_id.value[0] = key_cfg[i].prkey_pin;
 
-			r = sc_pkcs15emu_add_rsa_prkey(p15card, &prkey_obj, &prkey_info);
+			if (cxdata[0] == 0x01) { /* RSA */
+				prkey_info.modulus_length = bebytes2ushort(cxdata + 1);
```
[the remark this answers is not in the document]
alex-nitrokey: Just to make sure: You mean something like
```
if (cxdata[0] == SC_OPENPGP_KEYALGO_RSA && r > 1) {
```
right?

src/libopensc/card-openpgp.c
alex-nitrokey: Honestly, I am not sure, if I understand the issue here.

Do you mean in case the card response is different than requested? Like, the modulus length differs from what was inserted in the private key template? In this case I would prefer an error message instead.
Would it be sufficient to add
```
if (key_info->rsa.modulus && len == key_info->rsa.modulus_len){
```
as the value stored in key_info->rsa.modulus_len should be the size of the allocated memory (like it is done for the ec_point)?

src/libopensc/card-openpgp.c
alex-nitrokey: see above

frankmorgner:
I agree that we should keep the code base consistent. OpenSC uses named unions and therefor this style should also be used for new code.

Don't take changes of the code personally. If you want to avoid changes to *your code*, then please don't suggest them for merging into master. To get feedback, make a pull request marked as *work in progress* and we'll only merge it when all of the code is ready.

src/libopensc/card-openpgp.c
hongquan: Sorry, @marschap, I'm not understand what you mean here. Could you please give an example?

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
marschap: Almost.

You removed the length check after the call to read_file() a few lines earlier, and only keep the check whether the read failed. But the length read still may be too small (e.g. 0).

Looking at the whole section, I think there are more issues
1) the check for `cxdata[0] == 0` in line 289.
    If the length returned by read_file() is 0, then cxdata[]'s contents are undefined.
    (AFAIK only `static` variables get auto-initialized per the C standard)
    So for that check to be meaningful even in that case, I'd add a 
   ```
   memset(cxdata, 0, sizeof(cxdata));
   ```
   at around line 275.
2) the length check in line 301
    I'd use 
   ```
   if (cxdata[0] == SC_OPENPGP_KEYALGO_RSA && r >= 3) {
   ```
   because cxdata[]'s contents, that are used by the `bebytes2ushort()` call, are valid only if  3 bytes or more have been read.

The same applies to the logic for the public keys a few lines later.

Maybe a silly question:
The loops for private and public keys look very similar; is there a reason they are kept as two separate loops?

src/libopensc/card-openpgp.c
marschap: @hongquan 
I was referring to the whole code section starting in line 2627
```
                if (tag == 0x0081) {
                        /* set the output data */
                        if (key_info->u.rsa.modulus) {
                                memcpy(key_info->u.rsa.modulus, part, len);
                        }
                        else {
                                sc_log(card->ctx, "Error: key_info->u.rsa.modulus is NULL");
                        }
                        /* always set output for modulus_len */
                        key_info->u.rsa.modulus_len = len*8;
                }
```
I am wondering why the modulus parsed (in variable `part`) is only copied to the key_info structure if the `key_info->u.rsa.modulus` pointer is not NULL.
* Shouldn't it be copied in any case?
  I.e. shouldn't the memory be allocated it it is NULL?
  Especially with the change now that the explicit `modulus` variable & parameter to `pgp_calculate_and_store_fingerprint()` got (IMHO correctly) removed by @alex-nitrokey 's patch.
* Furthermore, shouldn't it be checked that the size allocated for `key_info->u.rsa.modulus` is large enough to keep the contents of the parsed modulus.
 E.g. when going from a 2k modulus to a 4k modulus, `key_info->u.rsa.modulus` may perhaps be allocated but too small.

@alex-nitrokey I think your code catches at least one potential error case and reports error on it.
The issues above can be left for a later commit after @honquan clarified.

src/libopensc/card-openpgp.c
marschap: I'd still go for the symbolic constant.
The magic 32 for the exponent length is also used in line 2299 of card-openpgp.c:
```
                /* OpenPGP Card only accepts 32bit as exponent length field,
                 * although you can import keys with smaller exponent;
                 * thus we don't change rsa.exponent_len, but ignore it here */
                data[3] = 0x00;
                data[4] = 0x20;
                data[5] = 0x00; /* Import-Format of private key (e,p,q) */
```
And there is something related around line 3119
```
                /* we only support exponent of maximum 32 bits */
                if (key_info->u.rsa.e_len > 4) {
                        sc_log(card->ctx,
                                   "Exponent %"SC_FORMAT_LEN_SIZE_T"u-bit (>32) is not supported.",
                                   key_info->u.rsa.e_len * 8);
                }
```
BTW, is it correct that `sc_cardctl_openpgp_keygen_info_t`'s `exponent_len` is measured in bits, while `sc_cardctl_openpgp_keystore_info_t`'s `e_len` is in bytes?

marschap:
@frankmorgner
It was not about my code being changed; it was more
* ignoring my requests to review changes in my repo
* doing a massive update with an "but I was working on it" comment without prior information
  (e.g. as a WIP PR)
* the time I unnecessarily spent on solving the same issue as @alex-nitrokey 
  (with him being ahead(
that hurt my pride ;-)
Please consider the issue settled.

Despite some nitpicking on details, I think this PR is very valuable, and I look forward to having it in master.

Additional question: Would you find a patch set acceptable that changes the code style to anonymous unions for all of OpenSC?

src/libopensc/card-openpgp.c
alex-nitrokey: Okay, I'll change it to a constant.

> BTW, is it correct that sc_cardctl_openpgp_keygen_info_t's exponent_len is measured in bits, while sc_cardctl_openpgp_keystore_info_t's e_len is in bytes?

Good point. As far as I can see, this was done, because sc_pkcs15_prkey_t does use bytes. This value is copied to e_len in openpgp_store_key function of pkcs15-openpgp.c. It does make much more sense to transform it to bits there and thus use e_len consistent to exponent_len of sc_cardctl_openpgp_keygen_info_t. This saves us all the shifting operations in card-openpgp.c.

I'll change this these days.

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
alex-nitrokey: Mh, it seems that my comment is somehow lost. Anyway. I think I fixed the issues.

> Maybe a silly question:
The loops for private and public keys look very similar; is there a reason they are kept as two separate loops?

Actually, I needed to think about this as well. Then I saw the comment above the first loop:

```
/* XXX: check if "halfkeys" can be stored with gpg2. If not, add keypairs in one loop */
```

I'd like to keep it like it is for now.

src/libopensc/card-openpgp.c
hongquan: I didn't think that `key_info->u.rsa.modulus` can be less than enough, because it was `malloc` with the size of expected key length.

But I'm happy that you add more checking that I might miss.

frankmorgner: I agree with @marschap , this whole section needs length checking! The length is obtained from `sc_asn1_read_tag`, which may initialize `len` to something beyond the length of the buffer, because you're not erroring on `SC_ERROR_ASN1_END_OF_CONTENTS`.

src/libopensc/card-openpgp.c
```diff
@@ -664,13 +697,35 @@ pgp_parse_algo_attr_blob(const pgp_blob_t *blob, sc_cardctl_openpgp_keygen_info_
...
+		case SC_OPENPGP_KEYALGO_ECDSA:
+
+			/* SC_OPENPGP_KEYALGO_ECDH || SC_OPENPGP_KEYALGO_ECDSA */
+			key_info->algorithm = blob->data[0];
+
+			/* Create copy of oid from blob */
+			for (j=0; j < (blob->len-1); j++) {
+				oid.value[j] = blob->data[j+1]; /* ignore first byte of blob (algo ID) */
+				if (j == blob->len-2)
+					oid.value[j+1] = -1;
```
frankmorgner: I'm just curious: Why are you setting the last byte to 0xFF, why aren't you intializing the rest of `oid.value` to 0xFF?

src/libopensc/card-openpgp.c
```diff
@@ -2283,46 +2371,123 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
...
+	if (key_info->algorithm == SC_OPENPGP_KEYALGO_RSA) {
 
-	if (modulus == NULL || exponent == NULL || mlen == 0 || elen == 0) {
-		sc_log(card->ctx, "Null data (modulus or exponent)");
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
-	}
+		if (key_info->u.rsa.modulus == NULL 
+		   || key_info->u.rsa.exponent == NULL
+		   || (key_info->u.rsa.modulus_len >> 3) == 0
+		   || (key_info->u.rsa.exponent_len >> 3) == 0) {
```
frankmorgner: bad indent.

`key_info->u.rsa.modulus_len >> 3` looks strange. Don't you want to use some more readable comparison? Why are you enforcing these limits at all?

src/libopensc/card-openpgp.c
```diff
@@ -2283,46 +2371,123 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
...
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+	sc_log(card->ctx, "pk_packet_len is %lu", pk_packet_len);
 
 	fp_buffer_len = 3 + pk_packet_len;
 	p = fp_buffer = calloc(fp_buffer_len, 1);
 	if (!p) {
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_ENOUGH_MEMORY);
 	}
 
+	/* constructing public-key packet */
```
frankmorgner: Using `sc_asn1_encode` would avoid manual encoding...

src/libopensc/card-openpgp.c
```diff
@@ -2537,7 +2730,8 @@ pgp_gen_key(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t *key_info)
 	LOG_FUNC_CALLED(card->ctx);
 
 	/* temporary workaround: protect v3 cards against non-RSA */
-	if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA)
+	if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA
+	   && card->type < SC_CARD_TYPE_OPENPGP_V3)
```
frankmorgner: bad indent

src/libopensc/card-openpgp.c
```diff
@@ -2735,54 +2930,73 @@ pgp_build_extended_header_list(sc_card_t *card, sc_cardctl_openpgp_keystore_info
...
-	if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA)
-		LOG_FUNC_RETURN(ctx, SC_ERROR_NOT_SUPPORTED);
+	/* RSA */
+	if (key_info->algorithm == SC_OPENPGP_KEYALGO_RSA){
 
-	if (key_info->rsa.keyformat == SC_OPENPGP_KEYFORMAT_RSA_STDN
-		|| key_info->rsa.keyformat == SC_OPENPGP_KEYFORMAT_RSA_CRTN)
-		comp_to_add = 4;
+		if (key_info->u.rsa.keyformat == SC_OPENPGP_KEYFORMAT_RSA_STDN
+		   || key_info->u.rsa.keyformat == SC_OPENPGP_KEYFORMAT_RSA_CRTN)
```
frankmorgner: bad indent

src/libopensc/card-openpgp.c
```diff
@@ -2724,9 +2918,10 @@ pgp_build_extended_header_list(sc_card_t *card, sc_cardctl_openpgp_keystore_info
 	u8 *data = NULL;
 	size_t len = 0;
 	u8 *p = NULL;
-	u8 *components[] = {key_info->rsa.e, key_info->rsa.p, key_info->rsa.q, key_info->rsa.n};
-	size_t componentlens[] = {key_info->rsa.e_len, key_info->rsa.p_len,
-				  key_info->rsa.q_len, key_info->rsa.n_len};
+	u8 *components[] = {key_info->u.rsa.e, key_info->u.rsa.p,
+			    key_info->u.rsa.q, key_info->u.rsa.n};
+	size_t componentlens[] = {key_info->u.rsa.e_len, key_info->u.rsa.p_len,
+				  key_info->u.rsa.q_len, key_info->u.rsa.n_len};
```
frankmorgner: Shouldn't we initialize these values based on the key type?

src/libopensc/pkcs15-pubkey.c
```diff
@@ -1083,6 +1083,8 @@ sc_pkcs15_dup_pubkey(struct sc_context *ctx, struct sc_pkcs15_pubkey *key, struc
 	u8* alg;
 	size_t alglen;
 
+	sc_log(ctx, "Called");
```
frankmorgner: use `LOG_FUNC_CALLED`

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -185,46 +201,167 @@ static int openpgp_generate_key(sc_profile_t *profile, sc_pkcs15_card_t *p15card
...
-		goto out;
-	memcpy(pubkey->u.rsa.exponent.data, key_info.rsa.exponent, key_info.rsa.exponent_len>>3); /* 1/8 */
-
-out:
-	if (key_info.rsa.modulus)
-		free(key_info.rsa.modulus);
-	if (key_info.rsa.exponent)
-		free(key_info.rsa.exponent);
+		goto err;
+	memcpy(pubkey->u.rsa.exponent.data, key_info.u.rsa.exponent, key_info.u.rsa.exponent_len>>3); /* 1/8 */
```
frankmorgner: What happens if the length is not a multiple of 8?

frankmorgner:
@marschap do you have more functional changes in your branch that need to be considered for this PR?

marschap:
@frankmorgner I haven't compared in detail, but I'd rather have this PR accepted before I start rebasing my patches and see which parts have been superseded by this PR.
(I guess it will not be very much left ;-)

src/libopensc/card-openpgp.c
alex-nitrokey: I mostly reused previous code. I trusted this code more than myself. I try to stop this behaviour :wink:

I tried to add some (hopefully) proper checks now. Please have a look at it.

`SC_ERROR_ASN1_END_OF_CONTENTS` is imho included in
```
	r = sc_asn1_read_tag((const u8**)&part, data_len - (in - data), &cla, &tag, &len);
 	if (part == NULL)
 		r = SC_ERROR_ASN1_OBJECT_NOT_FOUND;
 	LOG_TEST_RET(card->ctx, r, "Unexpected end of contents.");
```

src/libopensc/card-openpgp.c
alex-nitrokey: This check was done before as well. Honestly, I am not totally sure if it is needed or not. I just hesitated to just delete it. The shifting is unnecessary and deleted now though.

src/libopensc/card-openpgp.c
alex-nitrokey: I thought about this as well. I just couldn't come up with a elegant way to do it. Do you have a suggestion? I don't like to put these values in a array separately...

src/libopensc/card-openpgp.c
alex-nitrokey: Just because I didn't think about this solution. I am using sc_init_oid instead now.

src/pkcs15init/pkcs15-openpgp.c
alex-nitrokey: What do you suggest? I don't like this bit pushing either, but I don't know how to prevent it (and just copied it from previous code). 
Do you like me to just test `len%8 == 0` each time? Or do you have an idea to fix this issue generally?

src/libopensc/card-openpgp.c
alex-nitrokey: I try to refactor the code these days. But it will probably take some time for me.

src/pkcs15init/pkcs15-openpgp.c
marschap: @alex-nitrokey To make sure there are enough bytes available even for bit lengths which are not a multiple of 8, simply use the formula: `number_of_bytes = (number_of_bits + 7) / 8`.

Maybe a macro can help to increase code legibility, e.g.:
```
#define BYTES4BITS(num)  (((num) + 7) / 8)    /* number of bytes necessary to hold 'num' bits */
```

dengert:
There are about 22 other places in libopensc that do (x+7)/8. If you create
a macro consider replacing these.

How many other places where this should also be done?

frankmorgner:
> Additional question: Would you find a patch set acceptable that changes the code style to anonymous unions for all of OpenSC?

*In theory* it could be useful to renew the codebase of OpenSC. However, I think this would be like opening Pandora's box since there are **many** things that newer C code would do differently. I think that there are other things that are more pressing and equally work intensive. For example, having a good documentation would be more useful than just having a nice look of the code.

src/pkcs15init/pkcs15-openpgp.c
alex-nitrokey: Fixed via e706b3ba
I was not sure, what the best location for the define is and I put it in internal.h.
Do you have better idea?

src/libopensc/card-openpgp.c
alex-nitrokey: I don't think 19d4a34 is a masterpiece, but this is the only (easiest?) way I could think of to solve this early initialisation.

src/libopensc/card-openpgp.c
alex-nitrokey: I am sorry, but I really don't know how to use `sc_asn1_encode` and don't feel comfortable with it. You may can help me (e.g. give some advice or hint how it was done similarly elsewhere) or you are fine with keeping it the way it was the last years.

alex-nitrokey:
I am sorry that I couldn't work on this for some weeks, but now I feel like I solved most of the issues.

As I told before this PR (how it is now) should make all the OpenPGP part compatible with ECC keys. Nevertheless, OpenSC is (as far I can see) not ready yet to:

* import ECC keys
* decrypt with ECC keys or
* other operations (?)

Therefore, I can not fully test the implementations. This PR is still an improvement even before these operations are supported because:

* ECC keys are correctly recognized now (see #1351)
* Generating ECC keys works

dengert:
ECC keys can not be used to decrypt. They can only sign  ECDASA or derive ECDH 

The closest thing to "decryption" is derivation where some form of ECDH is done on both sides. OpenSC has support for ECDH.  See  [NIST.SP.800-56Ar3.pdf](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf)
and PKCS\#11.

pkcs11-tool can be used to do key derivation, mostly as an example. 

Each party does the following Given the certificate of the other party,  as $FILE,  extract the public key from the certificate:

in a script do somrting like: 
`TMP=/tmp/derive.$$`    (define diferent directory if both parties are on the same machine )
`openssl x509 -noout -in $FILE -pubkey  | openssl ec -pubin -outform DER > $TMP.other.pubkey.der`
`pkcs11-tool -l --derive -m ECDH1-COFACTOR-DERIVE -O -d $ID -i $TMP.other.pubkey.der`
 
Both parties will produce the same string.

src/libopensc/card-openpgp.c
frankmorgner: nevermind, most of it isn't ASN.1

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -185,46 +202,167 @@ static int openpgp_generate_key(sc_profile_t *profile, sc_pkcs15_card_t *p15card
...
+	else
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+
+	for (i=0; (i < key_info.u.ec.oid_len) && (i+2 < info_ec->der.len); i++){
+		key_info.u.ec.oid.value[i] = info_ec->der.value[i+2];
+	}
+	key_info.u.ec.oid.value[key_info.u.ec.oid_len] = -1;
+
+	/* Prepare buffer */
+	key_info.u.ec.ecpoint_len = (required->field_length>>2) + 1;
```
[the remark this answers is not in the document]
alex-nitrokey: Honestly, you really got me here... I needed to think about this for a *really* long time and realized, that I probably just understood something wrong when I did this. Funnily I reversed it in card-openpgp.c afterwards anyway as you can see in the commit.

I looked through the usage of ecpoint_len very carefully and now I am quite confident that the issue is fixed.

Thanks for looking at the code thorougly.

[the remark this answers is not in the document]
frankmorgner: Does this mean that this code fragment had an undetected bug? Did you verify the correctness with your new change?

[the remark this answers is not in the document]
alex-nitrokey: In part, I would say it did.

The problem is that I am not able to fully test the *correctness* of the results as e.g. storing an ec key fails on behalf of OpenSC. Key generation on card worked and works fine though. This is due to the fact that I first divided the length by four and then tricked myself by multiplying 4 again (as seen in the last commit). I am sorry for this mistake.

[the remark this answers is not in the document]
frankmorgner: I see. Then, however, I'm asking myself if we need this code at all if nobody ever looks into what you've encoded on the card. Could this be verified with gpg?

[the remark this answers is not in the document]
alex-nitrokey: When I started to implement ECC key support I assumed that importing ECC keys is working in OpenSC. Thus I implemented both, key generation and import alike with all needed subfunction. We could disable the key import but I felt like it would not hurt to have the already implemented code until the import is working for OpenSC.

In general I - of course - excpect the code to work. I am not so sure about the fingerprint function and there might be some other issues though.

I can not compare/verify with gpg as long as I can not import keys. Even then it is a tricky task as there is no ready-to-go way to move a S/MIME based key to OpenPGP standard. (Currently, I am looking for a way to get this a bit more fluid in the long run.)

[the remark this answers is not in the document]
frankmorgner: If you plan to add the missing features in the future, then we can already merge the prerequisite now.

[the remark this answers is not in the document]
alex-nitrokey: That would be great. I can not promise that I'll be able to do this anytime soon though, I am afraid.

PR #1363, opened 2018-05-20
openpgp-tool extensions & fixes

marschap:
Hi,

the attached commits fix a few annoyances in openpgp-tool and a new option --card-info to display information on the card.

It has been tested with
* Zeitcontrol OpenPGP card v2.0
* Zeitcontrol OpenPGP card v3.3
* Yubikey NEO

##### Checklist
- [x] Documentation is added or updated

PR #1508, opened 2018-10-13
openpgp-tool extensions

marschap:
Hi,

this PR 
* replaces the option `-L`  / ` --key-length` which implicitly assumes RSA keys with a more general option `-t` / `--key-type` that allows passing a string as key-type. 
  Currently it only supports `rsa` and `rsaNNNN` where `NNNN` is the modulus length, but allows for easy extension.
* adds two new information options to openpgp-tool
  * `-C` / `--card-info`   to display card-specific details
  * `-K` `--key-info`    to display info about the keys on the card
* adds length checks to `prettify_...()` functions instead of blindly assuming that the arguments are \0-terminated strings

This PR is an extension of the withdrawn PR #1363 addressing the issues mentioned in that old PR's reviews.

Please consider including it in OpenSC master.

##### Checklist

### DIFF
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -332,7 +332,7 @@ static char *prettify_name(u8 *data, size_t length)
 			*dst = *src++;
 			length--;
 			if (*dst == '<') {
-				if (*src == '<') {
+				if (length > 0 && *src == '<') {
 					src++;
 					length--;
 				}
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -251,7 +251,7 @@ static char *prettify_date(u8 *data, size_t length)
 		static char result[64];	/* large enough */
 
 		tp = gmtime(&time);
-		strftime(result, sizeof(result), "%Y-%m-%d %T", tp);
+		strftime(result, sizeof(result), "%Y-%m-%d %H:%M:%S", tp);
 		return result;
 	}
 	return NULL;
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -97,7 +97,7 @@ static int opt_keyinfo = 0;
 static char *exec_program = NULL;
 static int opt_genkey = 0;
 static u8 key_id = 0;
-static unsigned int key_len = 2048;
+static char *keytype = NULL;
 static int opt_verify = 0;
 static char *verifytype = NULL;
 static int opt_pin = 0;
@@ -119,7 +119,7 @@ static const struct option options[] = {
 	{ "user-info", no_argument,       NULL, 'U'        },
 	{ "key-info",  no_argument,       NULL, 'K'        },
 	{ "gen-key",   required_argument, NULL, 'G'        },
-	{ "key-length",required_argument, NULL, 'L'        },
+	{ "key-type",  required_argument, NULL, 't'        },
 	{ "help",      no_argument,       NULL, 'h'        },
 	{ "verbose",   no_argument,       NULL, 'v'        },
 	{ "version",   no_argument,       NULL, 'V'        },
@@ -141,7 +141,7 @@ static const char *option_help[] = {
 /* U */	"Show card holder information",
 /* K */	"Show key information",
 /* G */ "Generate key",
-/* L */ "Key length (default 2048)",
+/* t */ "Key type (default: rsa2048)",
 /* h */	"Print this help message",
 /* v */	"Verbose operation. Use several times to enable debug output.",
 /* V */	"Show version number",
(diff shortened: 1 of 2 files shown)


src/tools/openpgp-tool.c
```diff
@@ -441,29 +664,62 @@ static int do_dump_do(sc_card_t *card, unsigned int tag)
...
-	key_info.algorithm = SC_OPENPGP_KEYALGO_RSA;
-	key_info.rsa.modulus_len = key_len;
-	key_info.rsa.modulus = malloc(key_len/8);
-	r = sc_card_ctl(card, SC_CARDCTL_OPENPGP_GENERATE_KEY, &key_info);
-	free(key_info.rsa.modulus);
-	if (r < 0) {
-		util_error("failed to generate key: %s", sc_strerror(r));
+
+	/* generate key depending on keytype passed */
+	if (strncasecmp("RSA", keytype, strlen("RSA")) == 0) {
```
frankmorgner: Using `sscanf()` would be easier, wouldn't it?

marschap:
@frankmorgner I fixed some of the issues in your review, and explain why some other cannot be changed.

Please indicate whether my changes & explanations are sufficient.

frankmorgner:
Your explanation is not visible in this PR, do you have a link?

src/tools/openpgp-tool.c
```diff
@@ -152,37 +193,173 @@ static void show_version(void)
...
+		}
+	}
+	return NULL;
+}
+
+
+/* prettify algorithm parameters */
+static char *prettify_algorithm(u8 *data, size_t length)
+{
+	if (data != NULL && length >= 1) {
```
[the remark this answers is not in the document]
marschap: I cannot use `sc_asn1_print_object_id()` here for two reasons:
* some algorithms and their attributes are not encoded as OIDs (e.g. RSA),
  which is contrary to `sc_asn1_print_object_id()`'s expectations, breaking it.
* the `prettify_...()` functions prepare a string - either for printing 
  or for setting environment variables, while `sc_asn1_print_object_id()` wants to print directly

src/tools/openpgp-tool.c
```diff
@@ -152,37 +193,173 @@ static void show_version(void)
...
+		}
+	}
+	return NULL;
+}
+
+
+/* prettify date/time */
+static char *prettify_date(u8 *data, size_t length)
+{
+	if (data != NULL && length == 4) {
```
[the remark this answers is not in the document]
marschap: I cannot use `sc_asn1_print_utctime()` here, because the `prettify_...()` functions prepare a string - either for printing or for setting environment variables - while `sc_asn1_print_utctime()` wants to print directly.

src/tools/openpgp-tool.c
marschap: Not really.
I am trying to parse a regular expression `/rsa(\d+)?/i`, i.e. a string
1) starting with the literal string "RSA" in any upper-/lower-case combination
2) optionally followed by a positive number, which terminates the string

Unfortunately `scanf ("%u",...)` has some strange effects when being passed negative numbers.
With all the checks to cover that plus the checks for the correct size, the difference in LoC is negligible.

marschap:
Ooops,

it looks like I pressed the wrong button ("Resolve conversation") after commenting each issue, effectively deleting the comment.

Tried it again - hopefully this time the comments remain.

Sorry for the fuss
Peter

src/tools/openpgp-tool.c
```diff
@@ -152,37 +193,173 @@ static void show_version(void)
...
-		char *dst = str;
+	if (data != NULL && length > 0) {
+		char *src = (char *) data;
+		char *dst = (char *) data;
 
-		while (*src != '\0') {
+		while (*src != '\0' && length > 0) {
 			*dst = *src++;
+			length--;
 			if (*dst == '<') {
```
frankmorgner: I think you need to check for `length > 0` here

marschap:
#@frankmorgner 
* the uppercasing in prettify_hex(), that you wanted to have changed, was removed in commit 18dd1912d3f10a9c82659d0399846c0c09ab737c
* length check in prettify_name() added in commit f6edb4a5497062aa23e2120ffb14c50ca241276a

Thanks for the review.

frankmorgner:
thanks!

frankmorgner:
@marschap could you once more look at this:

CID 325863 (#1 of 1): Unintended sign extension (SIGN_EXTENSION)sign_extension: Suspicious implicit sign extension: data[0] with type u8 (8 bits, unsigned) is promoted in (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3] to type int (32 bits, signed), then sign-extended to type long (64 bits, signed). If (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3] is greater than 0x7FFFFFFF, the upper bits of the result will all be 1.

PR #1511, opened 2018-10-15
Fix division by zero in SimCList when appending to an empty list.

### LINKED ISSUE #1509: Floating-point division by zero in SimCList
SimCList implements `list_append()` through the use of `list_findpos()` where (see `src/common/simclist.c:422`) it performs division by zero whenever a new item is appended to an empty list.

Being a floating-point division by zero, it does not raise an SIGFPE unless the process is set up for that specifically. Please consider the following example (it is the file `samples/ex1.c` from the latest distribution of SimCList with all but two lines added):

```
#include <stdio.h>

#include <fenv.h>

#include <simclist.h>   /* use the SimCList library */


int main() {
    list_t mylist;      /* declare a list */
    int userval;

    feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);

    list_init(& mylist);    /* initialize the list */

    printf("Insert your number: ");
    scanf("%d", & userval);

    list_append(& mylist, & userval);       /* add an element to the list */
    
    printf("The list now holds %u elements.\n", \
            list_size(& mylist));           /* get the size of the list */

    printf("Your number was: %d\n", \
            * (int*)list_get_at(& mylist, 0));  /* extract the first element of the list */

    list_destroy(&mylist);
    
    return 0;
}
```

Please copy this code into an OpenSC's source tree as `src/common/ex1.c` and compile it as follows
(having `src/common` as your current working directory):
```
cc -D_GNU_SOURCE -I . -o ex1 ex1.c simclist.c -lm
```
then execute the example:
```
user@host> ./ex1
Insert your number: 1
Floating point exception
```

I practice, the issue described above causes the process to dump core whenever OpenSC attempts to add the first reader to the list of readers, _when_ the process has SIGFPE handling set up as described above, otherwise `list_append()` works by a chance.

I notified mij, the author of SimCList, regarding that issue.

vvp2014:
This PR fixes #1509 ('Floating-point division by zero in SimCList').

### DIFF
--- a/src/common/simclist.c
+++ b/src/common/simclist.c
@@ -419,7 +419,7 @@ static simclist_inline struct list_entry_s *list_findpos(const list_t *simclist_
     /* accept 1 slot overflow for fetching head and tail sentinels */
     if (posstart < -1 || posstart > (int)l->numels) return NULL;
 
-    x = (float)(posstart+1) / l->numels;
+    x = l->numels ? (float)(posstart+1) / l->numels : 0;
     if (x <= 0.25) {
         /* first quarter: get to posstart from head */
         for (i = -1, ptr = l->head_sentinel; i < posstart; ptr = ptr->next, i++);


dengert:
Looks OK.

frankmorgner:
thanks

vvp2014:
Thank you.

Issue #1507, opened 2018-10-13
opensc-notify: forks inside `C_Initialize`


loskutov:
### Problem Description
Here's a stack backtrace:
```
#0  0x00007ff6b32026a0 in __GI___nanosleep (requested_time=requested_time@entry=0x7ffdc57bfd20, remaining=remaining@entry=0x0) at ../sysdeps/unix/sysv/linux/nanosleep.c:28
#1  0x00007ff6b32357d4 in usleep (useconds=<optimized out>) at ../sysdeps/posix/usleep.c:32
#2  0x00007ff6b36ff232 in  () at /usr/lib/x86_64-linux-gnu/libpkcs11-helper.so.1
#3  0x00007ff6b324d578 in __run_fork_handlers (who=who@entry=atfork_run_prepare) at register-atfork.c:122
#4  0x00007ff6b3202707 in __libc_fork () at ../sysdeps/nptl/fork.c:58
#5  0x00007ff6b1f0b712 in fork_exec_with_fds
    (intermediate_child=intermediate_child@entry=0, working_directory=working_directory@entry=0x0, argv=argv@entry=0x55be22fdf8e0, envp=envp@entry=0x0, close_descriptors=close_descriptors@entry=1, search_path=search_path@entry=1, search_path_from_envp=0, stdout_to_null=0, stderr_to_null=0, child_inherits_stdin=0, file_and_argv_zero=0, cloexec_pipes=0, child_setup=0x0, user_data=0x0, child_pid=0x7ffdc57c0008, child_close_fds=0x7ffdc57bff20, stdin_fd=-1, stdout_fd=5, stderr_fd=7, error=0x7ffdc57c01c8) at ../../../../glib/gspawn.c:1644
#6  0x00007ff6b1f0bfab in fork_exec_with_pipes
[...]

PR #1513, opened 2018-10-17
gio: avoid unneccessary unitialization

frankmorgner:
@loskutov could you check if this helps for https://github.com/OpenSC/OpenSC/issues/1507? During the first initialization however, I fear that `g_application_register()` still forks...

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/ui/notify.c
+++ b/src/ui/notify.c
@@ -442,9 +442,10 @@ static GApplication *application = NULL;
 
 void sc_notify_init(void)
 {
-	sc_notify_close();
-	application = g_application_new("org.opensc.notify", G_APPLICATION_NON_UNIQUE);
-	if (application) {
+	if (!application) {
+		application = g_application_new("org.opensc.notify", G_APPLICATION_NON_UNIQUE);
+	}
+	if (application && FALSE == g_application_get_is_registered(application)) {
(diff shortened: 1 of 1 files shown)


frankmorgner:
did you try the unsafe fork mode of pkcs11-helper?

src/ui/notify.c
```diff
@@ -442,9 +442,10 @@ static GApplication *application = NULL;
 
 void sc_notify_init(void)
 {
-	sc_notify_close();
-	application = g_application_new("org.opensc.notify", G_APPLICATION_NON_UNIQUE);
-	if (application) {
+	if (!application) {
+		application = g_application_new("org.opensc.notify", G_APPLICATION_NON_UNIQUE);
+	}
+	if (application && FALSE == g_application_get_is_registered(application)) {
```
loskutov: `g_application_new` isn't supposed to return `NULL` (at least the docs don't state `NULL` as a possible return value)
```suggestion
	if (FALSE == g_application_get_is_registered(application)) {
```

frankmorgner: I can't imagine how `g_application_new()` could *always* allocate enough memory. And even if that would be possible, checking for `application != NULL` doesn't hurt.

PR #1518, opened 2018-10-25
Fix files to allow build without openssl and openpace

jozsefd:
##### Checklist

- [ ] macOS tokend is tested

I wanted to build OpenSC without openssl and openpace on windows and found the following only 2 places, which prevented it:
- the npa driver uses openpace, so I can exclude it from the build, but additionally it should have been suppressed in ctx.c
- pkcs15_prkey_sign in framework-pkcs15.c references SHA_DIGEST_LENGTH even when openssl is not enabled, so I set it

I used CMake to generate Visual Studio 2017 Solution and Project files, with this patch I could build 32 and 64 bit binaries. I tested it with opensc-explorer on cards: Atos CardOS and StarCOS 3.4

### DIFF
--- a/src/libopensc/ctx.c
+++ b/src/libopensc/ctx.c
@@ -146,7 +146,9 @@ static const struct _sc_driver_entry internal_card_drivers[] = {
 #endif
 	{ "openpgp",	(void *(*)(void)) sc_get_openpgp_driver },
 	{ "jpki",	(void *(*)(void)) sc_get_jpki_driver },
+#ifdef ENABLE_OPENPACE
 	{ "npa",	(void *(*)(void)) sc_get_npa_driver },
+#endif
 	/* The default driver should be last, as it handles all the
 	 * unrecognized cards. */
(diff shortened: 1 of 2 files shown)


src/libopensc/ctx.c
```diff
@@ -146,7 +146,9 @@ static const struct _sc_driver_entry internal_card_drivers[] = {
 #endif
 	{ "openpgp",	(void *(*)(void)) sc_get_openpgp_driver },
 	{ "jpki",	(void *(*)(void)) sc_get_jpki_driver },
+#ifdef ENABLE_OPENPACE
```
frankmorgner: npa should be available without OpenPACE, the crypto may be done inside the card reader...

jozsefd:
Yes, you are correct, I have double checked it right now. I was confused by
the fact that it includes pace.h. On the other hand it cannot be built
without OpenSSL, as it requires SM, which in turn "Need OpenSSL".
Would it make sense to add an "#ifdef ENABLE_OPENSSL" around the npa driver?

PR #1519, opened 2018-10-28
OpenPGP: support v3 DO sizes on reading

marschap:
Hi,

this PR allows reading DOs with sizes larger than the - previously - hard-coded 2k bytes if the OpenPGP card supports larger DOs.

Please consider including it into OpenSC's master.

Thanks in advance
Peter


marschap:
I wonder why Appveyor fails the tests.
According to my reading, of the C99 standard, variable length arrays are OK (see C99 standard, section 6.7.5.2).
@frankmorgner What is the C standard to be used within the OpenSC project?

dengert:
Does this help:
https://stackoverflow.com/questions/36020806/c99-casting-to-a-multi-dimensional-array-in-msvc

frankmorgner:
I think we're mostly C89 compliant. Variable size arrays are not declared anywhere...

marschap:
@frankmorgner Rewritten in C89.
@dengert Awwww Microsoft .... :-)

src/libopensc/card-openpgp.c
```diff
@@ -1083,20 +1090,30 @@ static int
...
+	static u8 *buffer = NULL;
+	static size_t buf_size = 0;
 
 	if (blob->data != NULL)
 		return SC_SUCCESS;
 	if (blob->info == NULL)
 		return blob->status;
 
+	/* calculate buffer size required and allocate static buffer */
+	if (buffer == NULL) {
```
frankmorgner: how can this *not* be NULL?

src/libopensc/card-openpgp.c
```diff
@@ -1083,20 +1090,30 @@ static int
...
+		buf_size = MAX(priv->max_specialDO_size, 2048);
+		buffer = calloc(buf_size, sizeof(u8));
```
frankmorgner: When is the buffer `free`d?

src/libopensc/card-openpgp.c
marschap: Because of the properties of variables declared using the storage class `static`!
In C, static variables survive outside the scope in which they are declared.
When re-entering the scope, the variables are in the same state they were in when the scope was previously left.
The assignment of a value in the declaration of a `static` variable is performed only once, i.e. when entering the scope the first time.

What the code does, is:
1) it declares the static variables, giving them their initial state
2) if the variables are still in their initial state, it calculates the size of the buffer and allocates the memory for it.
  If the variables are not in their initial state any more (i.e. buffer already allocated), then nothing is done.

Net result: buffer is allocated with the required size only once during the lifetime of the process.

src/libopensc/card-openpgp.c
marschap: Because buffer is a `static` variable.
The buffer is not freed, because it is only allocated once during the whole lifetime of the process (see other comment).

I consider this strategy safer than the use of a dynamic buffer that needs to get allocated & freed with every call to that function, which would risk failing to allocate the memory with every call.
In this implementation, the buffer is allocated once, and every subsequent call, reuses it without any risk of failing to allocate the memory after the first call.

frankmorgner: I overlooked the `static` attribute. Why don't you just use plain old `malloc` and `free`? I don't see a reason for taking shortcuts here.

marschap:
I do not consider it a shortcut, but a strategy for reducing risk: limit the risk for failing to allocate the memory to exactly the 1st invocation.

In addition, from a code logic PoV it is the closest to the original logic, with the static buffer.

frankmorgner:
If you want to keep this "magic" - which I doubt is needed -, put the pointer into the card driver's private data and perform proper cleanup to avoid dangeling memory (think of PKCS#11 that gets loaded/unloaded multiple times).

marschap:
@frankmorgner Thanks for asking for a rework. This made me step back and see the hen & egg problem in the previous versions: setting the size only after having done the complete parsing of the DOs ;-(

The new version radically simplifies the approach by using the max. allowed number of bytes as a constant.

Thanks in advance for including it in master.

frankmorgner:
Allocating 64KB is not ideal in restricted applications. The problem is that it may now fail silently (or abruptly with a program crash). Since I think we've dropped embedded compatibility long ago, I'll include this change anyway if you want. Did you experience a problem with the existing limit?

marschap:
@frankmorgner I limited the size to 8k.

This quadruples the size compared to v2 cards, but may fail with cards that
* support more than these 8k and
* have DOs with more than 8k

It should give enough margin for the time being.

Please consider merging
PEter

PR #1520, opened 2018-10-28
pkcs15-tool cleanup

marschap:
* remove unnecessary width in `printf("%.*s", sizeof(str), str)`.
  In those cases `printf("%s", string)` is equivalent
* when listing public keys, do not cut object labels in compact mode
* when listing private keys in compact mode, left align labels
* make hex codes at least 2 chars wide by changing "0x%X" to "0x%02X"


frankmorgner:
`printf("%.*s", sizeof(str), str)` is only identical to `printf("%s", str)` if the string is garanteed to be NUL terminated. I think it would be better to keep the implementation as is.

marschap:
Adapted and removed the changes from `printf("%.*s", sizeof(str), str)` to `printf("%s", str)`.

Also found a way to get rid of the necessity for defining MAX(), minimizing the change even more.

Issue #1515, opened 2018-10-19
Problems using  OpenSSL 1.1.1 "CRYPTO_secure_*" routines


dengert:
### Problem Description
`pkcs11-tool --test --login` fails the Decryption tests when it tried to encrypt:

```
Decryption (currently only for RSA)
  testing key 0 (PIV AUTH key) 
    RSA-X-509: Encryption failed, returning
    RSA-PKCS: Encryption failed, returning
  testing key 1 (SIGN key) 
    RSA-X-509: Encryption failed, returning
    RSA-PKCS: Encryption failed, returning
  testing key 2 (KEY MAN key) 
    RSA-X-509: Encryption failed, returning
    RSA-PKCS: Encryption failed, returning
```

Commit ea6f7cfe1db8724b0952c35a51b5c02858707b89  added these lines to ctx.c:
```
 835 #ifdef ENABLE_OPENSSL
 836         if (!CRYPTO_secure_malloc_initialized()) {
 837                 /* XXX What's a reasonable amount of secure heap? */
 838                 CRYPTO_secure_malloc_init(4096, 32);
 839         }
 840 #endif
```
As best I can tell from the OpenSSL man pages, once the secure heap is created its size is fixed.
and is used by internally by OpenSSL. For example in  [RAND_DRBG](https://www.openssl.org/docs/man1.1.1/man3/RAND_DRBG_new.html) 

Debugging the problem by adding `ERR_print_errors_fp(stdout)`  to `pkcs11-tool.c` shows:
```
140737353951232:error:24074041:random number generator:rand_pool_new:malloc failure:../openssl/crypto/rand/rand_lib.c:447:
140737353951232:error:2406C06E:random number generator:RAND_DRBG_instantiate:error retrieving entropy:../openssl/crypto/rand/drbg_lib.c:320:
140737353951232:error:24074041:random number generator:rand_pool_new:malloc failure:../openssl/crypto/rand/rand_lib.c:447:
140737353951232:error:2406C06E:random number generator:RAND_DRBG_instantiate:error retrieving entropy:../openssl/crypto/rand/drbg_lib.c:320:
140737353951232:error:24074041:random number generator:rand_pool_new:malloc failure:../openssl/crypto/rand/rand_lib.c:447:
140737353951232:error:24074041:random number generator:rand_pool_new:malloc failure:../openssl/crypto/rand/rand_lib.c:447:
[...]

PR #1521, opened 2018-10-30
incomplete mechanims list reported after merging RSA-PSS in some drivers

Jakuje:
After merging the change #1435 (RSA-PSS emulation), the cardos card stopped reporting the hashed mechanisms (for example `SHA1-RSA-PKCS`). This is done by changing of the semantics of the constant `SC_ALGORITHM_RSA_HASH_NONE`. ~~The attached patch resolves the issue for CardOS, cards we have around, but we might need some more system solution, because there might be more affected card drivers I can not test.~~

~~I will leave the PR as it is now, but I will have a look tomorrow if there is a way how to resolve it systematically (sigh ... there is no sensible explanation what is the `SC_ALGORITHM_RSA_HASH_NONE` supposed to mean initially). Comments and proposals welcomed.~~

The current patch solves the issue in the mechanism registration function, which was wrong since the RSA-PSS implementation. Also other cards that depended on this behavior should start working with this PR.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -12,7 +12,8 @@ noinst_HEADERS = cards.h ctbcs.h internal.h esteid.h muscle.h muscle-filesystem.
 	errors.h types.h compression.h itacns.h iso7816.h \
 	authentic.h iasecc.h iasecc-sdo.h sm.h card-sc-hsm.h \
 	pace.h cwa14890.h cwa-dnie.h card-gids.h aux-data.h \
-	jpki.h sc-ossl-compat.h card-npa.h ccid-types.h reader-tr03119.h
+	jpki.h sc-ossl-compat.h card-npa.h ccid-types.h reader-tr03119.h \
+	card-cac-common.h
 
 AM_CPPFLAGS = -D'OPENSC_CONF_PATH="$(sysconfdir)/opensc.conf"' \
      -D'DEFAULT_SM_MODULE_PATH="$(DEFAULT_SM_MODULE_PATH)"' \
--- a/src/libopensc/card-coolkey.c
+++ b/src/libopensc/card-coolkey.c
@@ -784,18 +784,25 @@ size_t coolkey_list_meter(const void *el) {
 	return sizeof(sc_cardctl_coolkey_object_t);
 }
 
+static void coolkey_free_private_data(coolkey_private_data_t *priv);
+
 static coolkey_private_data_t *coolkey_new_private_data(void)
 {
 	coolkey_private_data_t *priv;
+
 	/* allocate priv and zero all the fields */
 	priv = calloc(1, sizeof(coolkey_private_data_t));
 	if (!priv)
 		return NULL;
+
 	/* set other fields as appropriate */
 	priv->key_id = COOLKEY_INVALID_KEY;
-	list_init(&priv->objects_list);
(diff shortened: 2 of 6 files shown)


Jakuje:
Update: It looks like I pinpointed the issue and I am able to get sensible list of mechanisms from the card again with a simple patch attached (7461c259c). Additionally I took the opportunity to fix some coverity reports, formatting code around the change and adding the SHA224 mechanism, which was missing from the PKCS#1.5 mechanisms.

Jakuje:
Hmm ...it might need some more love ... still looking into that.

Jakuje:
It looks like I am done here. Sorry for mixing more changes here, but while verifying the RSA-PSS functionality I noticed that the last version that got merged upstream had a bug (sigh ...).

With the attached patch set, I see all the mechanisms I would expect in the CardOS cards (I assume other cards will be fine too) and I am able to do the RSA-PSS signatures with PIV and CAC cards (as well as other mechanisms verified by the `p11test`).

sumit-bose:
Hi @Jakuje,

I had to add something like
```
diff --git a/src/libopensc/Makefile.am b/src/libopensc/Makefile.am
index 5a17df2b..4c024296 100644
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -12,7 +12,7 @@ noinst_HEADERS = cards.h ctbcs.h internal.h esteid.h muscle.h muscle-filesystem.
        errors.h types.h compression.h itacns.h iso7816.h \
        authentic.h iasecc.h iasecc-sdo.h sm.h card-sc-hsm.h \
        pace.h cwa14890.h cwa-dnie.h card-gids.h aux-data.h \
-       jpki.h sc-ossl-compat.h card-npa.h ccid-types.h reader-tr03119.h
+       jpki.h sc-ossl-compat.h card-npa.h ccid-types.h reader-tr03119.h card-cac-common.h
 
 AM_CPPFLAGS = -D'OPENSC_CONF_PATH="$(sysconfdir)/opensc.conf"' \
      -D'DEFAULT_SM_MODULE_PATH="$(DEFAULT_SM_MODULE_PATH)"' \
```

to be able to build from a tar-ball after 'make dist'.

When testing the build calling OpenSSL's RAND_bytes() failed which I call directly after calling C_SignInit(). Do you have an idea what might cause this (it might not be related to your patches since I use all other patches in the branch as well)? OpenSC-0.16 worked find on the same platform.

dengert:
Could be related to #1515 if you are using OpenSSL 1.1.1

Jakuje:
@sumit-bose Thank you. Added to the PR.

About the OpenSSL issue, it sounds like the issue pointed out by @dengert

dengert:
Try calling RAND_bytes before any pkcs11 call.

sumit-bose:
@dengert, thanks for the hint, you are right I tried to workaround from #1515 and added CRYPTO_secure_malloc_init(65536, 32) to my code and now it is working as expected.

Nevertheless I think it would be nice to find a solution on the opensc-pkcs11.so side. I didn't load opensc-pkcs11.so directly but used p11-kit and iterate over what p11-kit offers me. So chances are that other projects might run into this kind of issue as well.

dengert:
It is really an OpenSSL problem. The CRYPTO_secure pool is not expandable
and can get called internally by OpenSSL. I will look at reporting it as a
bug next week.

PR #1525, opened 2018-11-05
allow specifying the size of OpenSSL secure memory

frankmorgner:
... and set it for builds where we're linking OpenSSL statically (i.e.
Windows and macOS)

fixes https://github.com/OpenSC/OpenSC/issues/1515

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/MacOSX/build-package.in
+++ b/MacOSX/build-package.in
@@ -61,6 +61,7 @@ if ! test -e ${BUILDPATH}/target/$PREFIX/lib/pkgconfig; then
 		--sysconfdir=$PREFIX/etc \
 		--enable-cvcdir=$PREFIX/etc/cvc \
 		--enable-x509dir=$PREFIX/etc/x509 \
+		--enable-openssl-secure-malloc=65536 \
 		--disable-dependency-tracking \
 		--enable-shared \
 		--enable-static \
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,6 +41,7 @@ install:
         $env:ARTIFACT="${env:ARTIFACT}-Light"
       } Else {
         $env:NMAKE_EXTRA="OPENSSL_DEF=/DENABLE_OPENSSL ${env:NMAKE_EXTRA}"
+        $env:NMAKE_EXTRA="OPENSSL_EXTRA_CFLAGS=/DOPENSSL_SECURE_MALLOC_SIZE=65536 ${env:NMAKE_EXTRA}"
         If (!(Test-Path C:\zlib )) {
           appveyor DownloadFile "https://github.com/madler/zlib/archive/v${env:ZLIB_VER_DOT}.zip" -FileName zlib.zip
           7z x zlib.zip -oC:\
--- a/configure.ac
+++ b/configure.ac
@@ -172,10 +172,17 @@ AC_ARG_ENABLE(
 
 AC_ARG_ENABLE(
 	[openssl],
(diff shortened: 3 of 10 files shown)


mouse07410:
Is there a reason to mess with secure heap from within OpenSC at all, instead of following @dengert's suggestion to leave it to the application?

frankmorgner:
Well, do you know what the secure heap is used for?

frankmorgner:
sorry for being cocky. We need to do it in OpenSC if nobody else does it (for example when we're linking OpenSSL statically) otherwise there is the risk that secrets get swapped out to disk.

dengert:
I agree that writing secrets to disk is a problem and is a problem for many projects.  And what we will see is other projects will also try use it too.  

[CRYPTO_secure_malloc](https://www.openssl.org/docs/man1.1.0/crypto/CRYPTO_secure_malloc_init.html) man pages say: " help protect applications (particularly long-running servers)" and "The level and type of security guarantees depend on the operating system." It also goes on to say "private key BIGNUM values are stored there".

OpenSSL has made it easy to use their "secure heap", but their implementation is not flexible. It may have been good enough for a few BIGNUMs. But it looks like in 1.1.1 (or maybe earlier) OpenSSL started to store other things in there like their RAND_DRBG pools, that are much larger then a few BIGNUMs. They (or other projects) may be storing other things in there too.  

 I would rather see OpenSC try to get OpenSSL to improve their implementation to allow for expansion of the "secure heap". Either when requested by a library or upon a failed allocation attempt. I would like to see this become their problem with their solution and not OpenSCs problem with an OpenSC only solution.  

For example: Upon an attempt to use one of the `CRYPTO_secure_*` routines (not sure which one)  if env  `OPENSSL_SECURE_HEAP="<size>, <minsize>"` is set, create it. This would then work for all the libraries that that might want to use it and is set at the application level.

And the problem of exposing secrets goes beyond just libraries, and the OS. Consider a virtual machine that thinks it has non paged memory, but in reality the host may still be swapping the VM's memory to disk or not clearing  physical memory with secrets before ging it to another VM. Other projects may have addressed this in their own way, or tried using kernel memory that is non-paged.

frankmorgner:
Just some quick remarks on the misconceptions above:
1. Future random number need to be kept secret. Not swapping the pool/state of RNG is totally reasonable.
2. If you're doing Virtualization correct, locked memory won't be swapped.

Yes OpenSSL could do better and yes you could take the time to report this. But look at the development process of OpenSSL (or especialy the contribution about secure memory): The OpenSSL core developers don't always have the time to make every feature smooth and comfortable. Nevertheless, its useful and gets integrated and obviously it *is* used by some people despite the limitations. If you suggest (or even implement) a fix for OpenSSL, this will usually take years until this hits a release.

Until then we need to deal with problems (swapping) with the tools that are available (OpenSSL 1.1+). (And pushing all responsibility to the user by honoring an environment variable is not a good way of dealing with this).

mouse07410:
@frankmorgner either this PR does not work, or I'm doing something wrong:

```
$ ./configure --enable-openssl --openssl-secure-malloc=1048576
configure: error: unrecognized option: `--openssl-secure-malloc=1048576'
Try `./configure --help' for more information
$ fgrep secure-malloc configure.ac
AC_ARG_ENABLE([openssl-secure-malloc],
    [AC_HELP_STRING([--openssl-secure-malloc=<SIZE_IN_BYTES>],
$ fgrep secure-malloc configure
  --openssl-secure-malloc=<SIZE_IN_BYTES>
# Check whether --enable-openssl-secure-malloc was given.
$ 
```

PR #1528, opened 2018-11-09
Fix C_UnwrapKey related issues. Code formatting in framework-pkcs15.c

hhonkanen:
The following changes are included:

- Fixes an issue in pkcs15-lib: calling check_key_compatibility in sc_pkcs15init_store_secret_key did not give correct result when called from C_UnwrapKey, in case CKA_EXTRACTABLE was false. Use value_len argument which is set even if we don't have key data yet.
- Adds AES algorithms to tokenInfo in myeid_init_card. Previously this was done, when AES keys were stored on card. It is better to add the algorithm in init_card, because SO-PIN would be needed to update tokenInfo in C_UnwrapKey case.
- Add a new function based on sc_pksc15_get_supported_algo to pkcs15.h/pkcs15.c to make difference between AES algorithms with different key lengths. This is needed to update all cases to tokenInfo correctly. There are different OIDs for each key length.
- changed sequences of four spaces to tabs in framework-pkcs15 for code consistency as discussed in #1393 review. Four spaces are replaced with one tab in all simple cases. Didn't touch some cases with more complex formatting.

Function sc_pkcs15init_store_secret_key can be called from framework-pkcs15 and from pkcs15-init tool. Both cases have been tested after this change and they worked correctly. 

Tested also other basic pkcs#11 functionality after building.

##### Checklist

### DIFF
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -2574,6 +2574,31 @@ sc_pkcs15_get_supported_algo(struct sc_pkcs15_card *p15card, unsigned operation,
 	return info;
 }
 
+struct sc_supported_algo_info *
+sc_pkcs15_get_specific_supported_algo(struct sc_pkcs15_card *p15card, unsigned operation, unsigned mechanism, const struct sc_object_id *algo_oid)
+{
+	struct sc_context *ctx = p15card->card->ctx;
+	struct sc_supported_algo_info *info = NULL;
+	int ii;
+
+	if (algo_oid == NULL)
+		return NULL;
+
+	for (ii=0;ii<SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo->supported_algos[ii].reference; ii++)
+		if ((p15card->tokeninfo->supported_algos[ii].operations & operation)
+				&& (p15card->tokeninfo->supported_algos[ii].mechanism == mechanism)
+				&& sc_compare_oid(algo_oid, &p15card->tokeninfo->supported_algos[ii].algo_id))
+			break;
+
+	if (ii < SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo->supported_algos[ii].reference)   {
+		info = &p15card->tokeninfo->supported_algos[ii];
+		sc_log(ctx, "found supported algorithm (ref:%X,mech:%X,ops:%X,algo_ref:%X)",
+				info->reference, info->mechanism, info->operations, info->algo_ref);
+	}
+
+	return info;
+}
+
 int
 sc_pkcs15_get_generalized_time(struct sc_context *ctx, char **out)
 {
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -2587,7 +2587,7 @@ sc_pkcs15_get_specific_supported_algo(struct sc_pkcs15_card *p15card, unsigned o
 	for (ii=0;ii<SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo->supported_algos[ii].reference; ii++)
 		if ((p15card->tokeninfo->supported_algos[ii].operations & operation)
 				&& (p15card->tokeninfo->supported_algos[ii].mechanism == mechanism)
-				&& sc_compare_oid(algo_oid, &p15card->tokeninfo->supported_algos[ii].algo_id))
+				&& sc_compare_oid(algo_oid, &p15card->tokeninfo->supported_algos[ii].algo_id) == 1)
 			break;
 
 	if (ii < SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo->supported_algos[ii].reference)   {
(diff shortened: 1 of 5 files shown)


src/libopensc/pkcs15.c
```diff
@@ -2574,6 +2574,31 @@ sc_pkcs15_get_supported_algo(struct sc_pkcs15_card *p15card, unsigned operation,
...
+	struct sc_supported_algo_info *info = NULL;
+	int ii;
+
+	if (algo_oid == NULL)
+		return NULL;
+
+	for (ii=0;ii<SC_MAX_SUPPORTED_ALGORITHMS && p15card->tokeninfo->supported_algos[ii].reference; ii++)
+		if ((p15card->tokeninfo->supported_algos[ii].operations & operation)
+				&& (p15card->tokeninfo->supported_algos[ii].mechanism == mechanism)
+				&& sc_compare_oid(algo_oid, &p15card->tokeninfo->supported_algos[ii].algo_id))
```
frankmorgner: you need to check for `1 == sc_compare_oid`

src/pkcs11/framework-pkcs15.c
```diff
@@ -564,8 +564,8 @@ CK_RV C_GetTokenInfo(CK_SLOT_ID slotID, CK_TOKEN_INFO_PTR pInfo)
 	slot->token_info.flags &= ~(CKF_USER_PIN_COUNT_LOW|CKF_USER_PIN_FINAL_TRY|CKF_USER_PIN_LOCKED);
 	auth = slot_data_auth(slot->fw_data);
 	sc_log(context,
-	       "C_GetTokenInfo() auth. object %p, token-info flags 0x%lX", auth,
-	       slot->token_info.flags);
+		   "C_GetTokenInfo() auth. object %p, token-info flags 0x%lX", auth,
+		   slot->token_info.flags);
```
frankmorgner:
```
   "C_GetTokenInfo()
^^
```
why do you keep the spaces above?

src/pkcs11/framework-pkcs15.c
```diff
@@ -1979,7 +1979,7 @@ pkcs15_init_pin(struct sc_pkcs11_slot *slot, CK_CHAR_PTR pPin, CK_ULONG ulPinLen
 	}
 
 	sc_log(context, "Init PIN: pin %p:%lu; unblock style %i", pPin,
-	       ulPinLen, sc_pkcs11_conf.pin_unblock_style);
+		   ulPinLen, sc_pkcs11_conf.pin_unblock_style);
```
frankmorgner: again, why keep these spaces?
```
   ulPinLen
^^
```

src/pkcs11/framework-pkcs15.c
```diff
@@ -2782,8 +2781,8 @@ get_X509_usage_privk(CK_ATTRIBUTE_PTR pTempl, CK_ULONG ulCount, unsigned long *x
 			*x509_usage |= SC_PKCS15INIT_X509_NON_REPUDIATION;
 		if (typ == CKA_VERIFY || typ == CKA_WRAP || typ == CKA_ENCRYPT) {
 			sc_log(context,
-			       "get_X509_usage_privk(): invalid typ = 0x%0lx",
-			       typ);
+				   "get_X509_usage_privk(): invalid typ = 0x%0lx",
+				   typ);
```
frankmorgner: no need to keep the spaces in front

src/libopensc/pkcs15.c
hhonkanen: Explicit check for == 1 added. 

Checking for boolean return values without == operator is normal in C language and not always considered as bad practice. I guess you want to make a clear difference with functions, which return 0 to inidicate success? (because CKR_OK and SC_SUCCESS are 0)

I put it other way around though: if (sc_compare_oid(...) == 1)
This style is mostly used in OpenSC and i personally like it better, I think the code is easier to read this way.

hhonkanen:
I kept spaces in the sections you commented about, because they looked good in my editor like that. The second line started under the first letter of the first line, and the spaces didn't match a full tab.

I removed the spaces now.

src/libopensc/pkcs15.c
frankmorgner: Thanks!

Before this change, `sc_compare_oid` could have returned an error (negative), which would have been interpreted as success.

In general, using explicit checks is advisable in security related code... I haven't found the time yet to refactor OpenSC, however...

frankmorgner: Oh, and choosing `1 == x` over `x == 1` protects from overlooking the following typo `1 = x`, because it yields a compiler error...

hhonkanen: You are right, it could return SC_ERROR_INVALID_ARGUMENT so this check is required for it to work correctly, not only a style issue. I agree that it's good and more safe to use explicit checks. I'm more used to coding c++ where a bool function can only return true or false, not for example -1.

Which one is better, 1==x or x==1 could be a source of a neverending debate. I'd rather not go into it ;-)

PR #1536, opened 2018-11-21
SM: fix crash if nested GET RESPONSE issued during sm session fails o…

idrassi:
…r returns invalid MAC.

Crash discovered while implementing secure messaging for a new IAS v2 based smart card. During tests, if GET RESPONSE is issued during secure messaging session and if its MAC is wrong, OpenSC frees twice the SM context and we end up calling the function iso_rm_sm with sctx parameter as NULL pointer which caused a crash since their was no check on this parameter before its use.


frankmorgner:
thanks

PR #1537, opened 2018-11-21
Update reader-pcsc.c

hajikhorasani:
Error handling extended in case of changing the card reader

##### Checklist

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -339,6 +339,16 @@ static int refresh_attributes(sc_reader_t *reader)
 			reader->flags &= ~SC_READER_CARD_CHANGED;
 			LOG_FUNC_RETURN(reader->ctx, SC_SUCCESS);
 		}
+		
+		/* the system could not dectect any reader. It means, the prevoiusly attached reader is disconnected. */		
+		if (rv == SCARD_E_NO_READERS_AVAILABLE || rv == SCARD_E_SERVICE_STOPPED) {
+ 			if (old_flags & SC_READER_CARD_PRESENT) {
+ 				reader->flags |= SC_READER_CARD_CHANGED;
+ 			}
+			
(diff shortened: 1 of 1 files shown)


hajikhorasani:
Now, it's possible to detach and attach the card readers without unload and load the library.

src/libopensc/reader-pcsc.c
```diff
@@ -339,6 +339,16 @@ static int refresh_attributes(sc_reader_t *reader)
 			reader->flags &= ~SC_READER_CARD_CHANGED;
 			LOG_FUNC_RETURN(reader->ctx, SC_SUCCESS);
 		}
+		
+		/* the system could not dectect any reader. It means, the prevoiusly attached reader is disconnected. */		
+		if (rv == SCARD_E_NO_READERS_AVAILABLE || rv == SCARD_E_SERVICE_STOPPED) {
```
frankmorgner: cast to (LONG)

hajikhorasani:
Yes, it’s reproducible. I see the problem usually with Gemalto card readers with embedded SIM-sized smart cards.

frankmorgner:
how can it be reproduced?

though you've *resolved* the conversation above, the build problem persists...

hajikhorasani:
I’ll resolve the build problem ASAP.

hajikhorasani:
@frankmorgner I've resolved the build problem, but there is an issue in "Inno Setup" with HOST=mingw. How could I fix it ? Is it related to source code that I changed ?
Error:
wine: cannot find 'C:/Program Files (x86)/Inno Setup 5/ISCC.exe'
The command "if [ ! -z "$HOST" -a "${DO_COVERITY_SCAN}" != "yes" ]; then make install; wine "C:/Program Files (x86)/Inno Setup 5/ISCC.exe" win32/OpenSC.iss; fi" exited with 2.

PR #1539, opened 2018-11-26
Implemented handling of CKA_ALWAYS_AUTHENTICATE when creating key objects.

hhonkanen:
Implemented handling of CKA_ALWAYS_AUTHENTICATE attribute to framework-pkcs15.c when importing and creating keys. CKA_ALWAYS_AUTHENTICATE = CK_TRUE sets user_consent = 1 in PKCS#15 level.  
Added command line options to tools:   
In pkcs11-tool --always-auth sets CKA_ALWAYS_AUTHENTICATE to CK_TRUE.
In pkcs15-init, you can set for example --user-consent 1
pkcs15-init allows using also other values as defined in PKCS#15 specification.

Tested with both tools and a MyEID card. userConsent field was encoded into commonObjectAttributes and the user_consent value was passed to the driver. Verified that userConsent was effective when using the created keys.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -2153,6 +2153,9 @@ pkcs15_create_private_key(struct sc_pkcs11_slot *slot, struct sc_profile *profil
 		case CKA_OPENSC_NON_REPUDIATION:
 			args.usage |= pkcs15_check_bool_cka(attr, SC_PKCS15_PRKEY_USAGE_NONREPUDIATION);
 			break;
+		case CKA_ALWAYS_AUTHENTICATE:
+			args.user_consent = (int) (pkcs15_check_bool_cka(attr, 1));
+			break;
 		default:
 			/* ignore unknown attrs, or flag error? */
 			continue;
@@ -2318,7 +2321,10 @@ pkcs15_create_secret_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
 			break;
 		case CKA_EXTRACTABLE:
 			if (pkcs15_check_bool_cka(attr, 1))
-					args.access_flags |= SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE;
+				args.access_flags |= SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE;
+			break;
+		case CKA_ALWAYS_AUTHENTICATE:
+			args.user_consent = (int) (pkcs15_check_bool_cka(attr, 1));
 			break;
 		default:
 			/* ignore unknown attrs, or flag error? */
@@ -2931,6 +2937,7 @@ pkcs15_gen_keypair(struct sc_pkcs11_slot *slot, CK_MECHANISM_PTR pMechanism,
 	char		priv_label[SC_PKCS15_MAX_LABEL_SIZE];
 	int		rc;
 	CK_RV rv = CKR_OK;
+	CK_BBOOL always_auth = CK_FALSE;
 
 	sc_log(context, "Keypair generation, mech = 0x%0lx",
 		   pMechanism->mechanism);
@@ -3045,6 +3052,12 @@ pkcs15_gen_keypair(struct sc_pkcs11_slot *slot, CK_MECHANISM_PTR pMechanism,
 	if (rv != CKR_OK)
 		goto kpgen_done;
 	pub_args.x509_usage = keygen_args.prkey_args.x509_usage;
+
+	len = sizeof(always_auth);
+	rv = attr_find(pPrivTpl, ulPrivCnt, CKA_ALWAYS_AUTHENTICATE, &always_auth, &len);
+	if (rv == CKR_OK && always_auth == CK_TRUE) {
+		keygen_args.prkey_args.user_consent = 1;
+	}
 
 	/* 3.a Try on-card key pair generation */
 
@@ -4831,6 +4844,10 @@ pkcs15_skey_get_attribute(struct sc_pkcs11_session *session,
 		*(CK_BBOOL*)attr->pValue = (((skey->base.p15_object->flags & SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE) == SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE)
 					&& (skey->base.p15_object->flags & SC_PKCS15_PRKEY_ACCESS_NEVEREXTRACTABLE) == 0
 					&& (skey->base.p15_object->flags & SC_PKCS15_PRKEY_ACCESS_ALWAYSSENSITIVE) == 0) ? CK_TRUE : CK_FALSE;
+		break;
+	case CKA_ALWAYS_AUTHENTICATE:
+		check_attribute_buffer(attr, sizeof(CK_BBOOL));
+		*(CK_BBOOL*)attr->pValue = skey->base.p15_object->user_consent;
 		break;
 	case CKA_VALUE_LEN:
 		check_attribute_buffer(attr, sizeof(CK_ULONG));
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -3670,7 +3670,7 @@ pkcs15_prkey_get_attribute(struct sc_pkcs11_session *session,
 		break;
     case CKA_ALWAYS_AUTHENTICATE:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
-		*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent;
+		*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent >= 1 ? CK_TRUE : CK_FALSE;
 		break;
 	case CKA_PRIVATE:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
@@ -4847,7 +4847,7 @@ pkcs15_skey_get_attribute(struct sc_pkcs11_session *session,
 		break;
 	case CKA_OPENSC_ALWAYS_AUTH_ANY_OBJECT:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
-		*(CK_BBOOL*)attr->pValue = skey->base.p15_object->user_consent == 1 ? CK_TRUE : CK_FALSE;
+		*(CK_BBOOL*)attr->pValue = skey->base.p15_object->user_consent >= 1 ? CK_TRUE : CK_FALSE;
 		break;
 	case CKA_VALUE_LEN:
 		check_attribute_buffer(attr, sizeof(CK_ULONG));
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -2323,7 +2323,7 @@ pkcs15_create_secret_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
 			if (pkcs15_check_bool_cka(attr, 1))
 				args.access_flags |= SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE;
 			break;
-		case CKA_ALWAYS_AUTHENTICATE:
+		case CKA_OPENSC_ALWAYS_AUTH_ANY_OBJECT:
 			args.user_consent = (int) (pkcs15_check_bool_cka(attr, 1));
 			break;
 		default:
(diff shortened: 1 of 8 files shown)


frankmorgner:
Why do you want to force PIN verification in PKCS#11? Shouldn't this be done in the PKCS#15 layer if the card enforces this behavior?

please document new command line switches

dengert:
PKCS#11 2.40 shows `CKA_ALWAYS_AUTHENTICATE` is an attribute of only `CKO_PRIVATE_KEY` objects. It is not an attribute of `CKO_SECRET_KEY` or any other objects. So framework-pkcs15.c should not be calling `pkcs15_create_secret_key` to set `user consent`. 
The PKCS\#15 `user consent` is much more general then the PKCS\#11 `CKA_ALWAYS_AUTHENTICATE`. The code in `pkcs11-tool.c` tries to support `CKA_ALWAYS_AUTHENTICATE` for secret keys and any other objects. If you add this capability to the pkcs11-tool.c it should be well documented as non-standard and some PKCS\#11 modules may not accept it.
 
I would suggest that the changes the pkcs11-tool. c only support the  `CKA_ALWAYS_AUTHENTICATE` with `CKO_PRIVATE_KEY` objects and stick to the standards.  

There is no assumption that the pkcs11-tool is calling a PKCS\#15 aware module (other then some OpenSC cards  are PKCS\#15 based.)

If you really want to allow PKCS\#11 applications to set this attribute you should look at adding a vendor supported attribute. `opensc-pkcs11.so` has two: `CKA_OPENSC_NON_REPUDIATION` and `CKA_SPKI` as examples which are used internally. grep  for `CKA_VENDOR_DEFINED`

If you plan on adding a CKA_VENDOR_DEFINED to your own PKCS\#11` module, Do something like NSS and OpenSC do in ./pkcs11-opensc.h to grab a name space to them unique.

hhonkanen:
@dengert Thank you for checking this. I incorrectly assumed that CKA_ALWAYS_AUTHENTICATE could be set also for a secret key template. It's better to follow the spec on this, so I will change my implementation.

We actually need this functionality for secret keys too and must be able to set userConsent=1 using PKCS#11, so I would like to add a custom attribute to make it possible. On the PKCS#15 side userConsent can be set on any object which has access conditions specified, so someone else might find it useful as well.

I planned to do it like this:
/* In PKCS#11 CKA_ALWAYS_AUTHENTICATE attribute is only associated with private keys. The corresponding userConsent field in PKCS#15 is allowed for any object type. This attribute can be used
to set userConsent=1 for other objects than private keys via PKCS#11. */
#define CKA_OPENSC_ALWAYS_AUTH_ANY_OBJECT (CKA_VENDOR_DEFINED | SC_VENDOR_DEFINED | 3UL)

If you can find some more compact, but still obvious enough name for the attribute, please suggest.

I can restrict the --always-auth option to private keys only in pkcs11-tool. Whoever needs the custom attribute could then use it in PKCS#11 calls or use pkcs15init.

src/tools/pkcs15-init.c
```diff
@@ -218,6 +219,7 @@ const struct option	options[] = {
 	{ "wait",		no_argument, NULL,		'w' },
 	{ "help",		no_argument, NULL,		'h' },
 	{ "verbose",		no_argument, NULL,		'v' },
+	{ "user-consent",	required_argument, NULL, OPT_USER_CONSENT},
```
frankmorgner: please rename this to something like *always-authenticate* and move it near `OPT_PRIVATE` and `OPT_SENSITIVE`. Don't forget to update `doc/tools/pkcs11-tool.1.xml`

frankmorgner:
@hhonkanen Yes, that's a good plan!

src/tools/pkcs15-init.c
hhonkanen: I'll move the argument close to similar arguments and update pkcs11-tool.1.xml. Wasn't aware of this file, maybe that caused Travics CI build to fail?

Let me explain why I chose to use name "user-consent" in pkcs15-init.
This a PKCS#15 tool, so I though it would be good to use PKCS#15 terminology. User might be looking at PKCS#15/ISO7816-15 instead of PKCS#11 when using this tool. The meaning of this parameter is a also a bit different in pkcs#15: 

"The userConsent field gives, in the case of a private object (or an object for which access conditions has been specified), the number of times an application may access the object without explicit consent from the user (e.g. a value of 3 indicates that a new authentication will be required before the first, the 4th, the 7th, etc. access). A value of 1 means that a new authentication is required before each access.". 

It is an integer value and a nonzero value doesn't mean "always authenticate", unless it is 1. With pkcs15init you can for example set --user-consent 5, which means re-authentication is required every fifth usage. I'm don't know which cards support it this way, that's just what the spec says.

On the other hand, it would be good to use unified terms between OpenSC's tools when possible, this would support naming it as you suggest. I just kind of found more reasons to use the pkcs15 name in this tool. 

Please confirm, considering my explanation, are you sure you want it to be renamed?

frankmorgner: Oh, I'm terribly sorry. I was stuck believing that this change was done to pkcs11-tool (where you're correctly using `always-auth`). In pkcs15-init, you're perfectly right that using `user-consent` is more consistent with the PKCS#15 interpretation.

To sum up, please leave your implementation as is. You just need to
- extend the xml files, from which the man pages are generated
- use the vendor specific implementation regarding PKCS#11 as suggested by Doug

hhonkanen: No problem and thank you for taking the time to review this.

I'll make the needed modifications.

hhonkanen:
In pkcs15_prkey_get_attribute user_consent is currently converted to CK_BBOOL like this:

*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent;

IMO this is kind of vague. CK_TRUE is defined as 1 but user_consent is int and could be >1, so directly casting to CK_BBOOL doesn't go right. 
user_consent > 1 also doesn't match definition of CKA_ALWAYS_AUTHENTICATE = CK_TRUE in PKCS#11. In my undertanding there's no way to define user_consent>1 in PKCS#11 so it would be better to return CK_FALSE.

Do you agree it would be better like this:
*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent == 1 ? CK_TRUE : CK_FALSE;
?
I could change it in this PR. It could potentially cause a compatibility problem if someone is expecting CKA_ALWAYS_AUTHENTICATE to be CK_TRUE for user_consent==5 for example. Don't know if anyone is really using it like that.

dengert:
You have a point about the conversion of user_consent to CKA_ALWAYS_AUTHENTICATE only working when user_consent == 1.   

PKCS\#11 was written after PKCS\#15 does not say how to do the conversion. (Not that I know of). I don't think how to use user_consent was worked out in PKCS\#15. User_consent is the number of times an object can be used before the PIN must be verified again and each object can have different value.  One way to look at it: There should be a user_consent_counter for each object decremented each time the object is used. Which then also implies if the PIN is reentered, each object's user_consent_counter should be reset to its user_consent value.  It is also not clear if the card or middleware should enforce this. The user_consent really only makes sense on a private key (and maybe a secret key). Any other object can be read once by the application as the application does not have to go back to the card or middleware.  

I think the PKCS\#11 authors did not see a way to address the issues and simplified it by limiting it to a BOOL  attribute of a private key, CKA_ALWAYS_AUTHENTICATE. The name containing "ALWAYS",  implies (IMHO) it is a fixed attribute for a key, and does not change during a session. i.e. application could query it once not expecting it might change. 
  
It can also be enforced on a card, rather easily too. The PIV cards has the "PIN Always" rule. The card sets a flag when a verify is done, and reset it after any other APDU is done. A crypto operation using the a  key with "PIN ALways" checks the flag. (Thus all the talk in OpenSC discussions about interference from other applications and drivers so no other APDU is done to the card.) 

Other cards may enforce  CKA_ALWAYS_AUTHENTICATE in other ways, or may not, treating user_consent and CKA_ALWAYS_AUTHENTICATE as policies to be enforced by middleware. 

The PIV and other cards that are not PKCS\#15 are emulation cards as they construct internal OpenSC structures to look like they came from PKCS\#15. So if they set user_consent, they do it based in other information. The PIV sets it to one because the NIST Specs say the SIGN Key (9C)  is "PIN Always" and the card will enforce it and the PIV driver was written to follow the specs. 

You said: " In my understanding there's no way to define user_consent>1 in PKCS#11 so it would be better to return CK_FALSE."  I agree with the first half, but disagree with the conclusion.  (Its not a problem for PIV as user_consent is  0 or 1.) 

I would suggest that you grep the source looking for user_consent to see what code uses it and where did they get their information before making any changes. Pointing out where it may be greater then 1  will get the attention of developers of that code. (The git blame command can be very helpful.)

hhonkanen:
In MyEID card, handling of user_consent = 1 is implemented like this: We have a corresponding proprietary attribute in File Control Parameters structure used in CREATE FILE apdu. This attribute means "Clear AC #x after crypto-operation". A normal use case is that PIN #2 is a signature PIN and associated with a private key. A user authenticates PIN #2 and signs a hash with this key. The card clears AC #2 by itself, so the key cannot be used again without re-authentication, regardless of what the middleware does. We cannot handle user_consent > 1 in MyEID.

To user_consent > 1 to make sense, in my opinion the card itself should maintain the counter and enforce it. 

A real life example where for example user_consent = 10 would make sense, would be something similar like the contactless payment system used in card payment terminals. You can usually just show your card and pay a small amount, but the terminal SOMETIMES requires PIN verification.

We have also found usage for user consent in key wrapping operations. We want to ensure with user_consent = 1, that a specific key is wrapped only once per authentication. In this case, the object which has user consent set is not the one doing encryption, but the one that is being wrapped. This is in line with the PKCS#15 definition, which is a about accessing any private object.

I quickly searched the code for user_consent and looks like everyone is using it so that it's either 0 or 1. But if someone really has user_consent set to for example 5 using some other tools, it is parsed from PKCS#15, and the user gets the object template via PKCS#11 and looks for CKA_ALWAYS_AUTHENTICATE. We cannot know what they expect!

Would it be ok if I change it like this: 
*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent != 0 ? CK_TRUE : CK_FALSE;

In my opinion the current implementation is technically wrong. It's casted to CK_BBOOL simply like this:
*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent;

Let's do like this: 

CK_BBOOL val;
CK_ATTRIBUTE always_auth { CKA_ALWAYS_AUTHENTICATE, &val, sizeof(val));
pkcs15_prkey_get_attribute(obj, ses, &always_auth);

let's say user_consent = 5.
if we check: if (!val) ... the expression evaluates to false.
if we check: if (val == CK_FALSE) ... the expression evaluates to false, because val is not zero.
but if we check: if (val == CK_TRUE) ... it also evaluates to false, because val is not 1. 

I don't change it for now and leave this open to discussion. If you agree with my new suggestion, I can then add another commit.

hhonkanen:
Requested changes done and rebased on top of upstream/master

src/pkcs11/framework-pkcs15.c
```diff
@@ -4832,6 +4845,10 @@ pkcs15_skey_get_attribute(struct sc_pkcs11_session *session,
 					&& (skey->base.p15_object->flags & SC_PKCS15_PRKEY_ACCESS_NEVEREXTRACTABLE) == 0
 					&& (skey->base.p15_object->flags & SC_PKCS15_PRKEY_ACCESS_ALWAYSSENSITIVE) == 0) ? CK_TRUE : CK_FALSE;
 		break;
+	case CKA_OPENSC_ALWAYS_AUTH_ANY_OBJECT:
+		check_attribute_buffer(attr, sizeof(CK_BBOOL));
+		*(CK_BBOOL*)attr->pValue = skey->base.p15_object->user_consent == 1 ? CK_TRUE : CK_FALSE;
```
frankmorgner: Should be
```
*(CK_BBOOL*)attr->pValue = skey->base.p15_object->user_consent >= 1 ? CK_TRUE : CK_FALSE;
```

frankmorgner:
>Would it be ok if I change it like this:
>(CK_BBOOL)attr->pValue = prkey->prv_p15obj->user_consent != 0 ? CK_TRUE : CK_FALSE;

Yes, that would be the only correct way! Feel free to fix it here or in an other PR. Note that you have the same problem in your current code, see [comment above](https://github.com/OpenSC/OpenSC/pull/1539#discussion_r238657128).

I've seen other PKCS#11 providers that implement their own PIN query GUI to let the underlying layers decide whether or when to ask for the PIN, but that isn't ideal either. For now, it's OK if we risk unnecessary PIN validations. But as Doug pointed out, the application could also check the PIN status with `C_GetTokenInfo()`...

hhonkanen:
Fixed this in both places.

I chose to use >= instead of !=

There's a small functional difference. user_consent must be zero or positive. If it is <0, I'd consider this an erroneus/undefined value which cannot mean CKA_ALWAYS_AUTHENTICATE = TRUE.

frankmorgner:
thanks to all!

PR #1540, opened 2018-11-28
Fix some pinpad reader related issues

llogar:
1. Show pinpad reader capabilities even for uninitialised tokens. This way pinpad can be used during token initialisation.

2. Make possible to create SO PIN object during initialisation (using pkcs15-init app) even if no SO PIN was provided (on the command line) but pinpad reader is used and card profile contains SO PIN data.

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -771,6 +771,7 @@ sc_pkcs15init_add_app(struct sc_card *card, struct sc_profile *profile,
 	struct sc_app_info	*app;
 	struct sc_file		*df = profile->df_info->file;
 	int			r = SC_SUCCESS;
+	int			has_so_pin = 0;
 
 	LOG_FUNC_CALLED(ctx);
 	p15card->card = card;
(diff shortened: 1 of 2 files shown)


src/pkcs15init/pkcs15-lib.c
```diff
@@ -784,8 +785,15 @@ sc_pkcs15init_add_app(struct sc_card *card, struct sc_profile *profile,
...
+	/* In case of pinpad readers check if SO PIN is defined in a profile */
+	if (!args->so_pin_len && (card->reader->capabilities & SC_READER_CAP_PIN_PAD)) {
+		sc_profile_get_pin_info(profile, SC_PKCS15INIT_SO_PIN, &pin_ainfo);
+		/* If found, assume we want SO PIN */
+		has_so_pin = pin_ainfo.attrs.pin.reference != -1;
+	}
+
 	/* If the profile requires an SO PIN, check min/max length */
-	if (args->so_pin_len) {
+	if (args->so_pin_len || has_so_pin) {
```
frankmorgner: This looks like a hack: How can you check max/min length if `args->so_pin_len` is `0`? Shouldn't this look something like this:
```c
	/* In case of pinpad readers check if SO PIN is defined in a profile */
	if (!args->so_pin_len && (card->reader->capabilities & SC_READER_CAP_PIN_PAD)) {
		sc_profile_get_pin_info(profile, SC_PKCS15INIT_SO_PIN, &pin_ainfo);
		/* If found, assume we want SO PIN */
		has_so_pin = pin_ainfo.attrs.pin.reference != -1;
	} else {
		has_so_pin = args->so_pin_len != 0;
	}

	if (has_so_pin) {
		const char	*pin_label;

		/* If the profile requires an SO PIN, check min/max length */
		if (args->so_pin_len) {
			sc_profile_get_pin_info(profile, SC_PKCS15INIT_SO_PIN, &pin_ainfo);
			r = sc_pkcs15init_qualify_pin(card, "SO PIN", args->so_pin_len, &pin_ainfo);
			LOG_TEST_RET(ctx, r, "Failed to qualify SO PIN");
		}
```

llogar: Well, since `sc_pkcs15init_qualify_pin()` returns `SC_SUCCESS` for `args->so_pin_len == 0`, I thought additional `if` wasn't necessary...

Anyway, here's an updated PR.

frankmorgner:
thanks

PR #1541, opened 2018-11-28
Fix PKCS#11 token label formatting

llogar:
PKCS#15 token label may be padded with spaces, trim it when making a PKCS#11 token label in order not to loose closing parenthesis.

before:
$ pkcs11-tool --list-slots
Available slots:
Slot 0 (0x0): OMNIKEY AG CardMan 3121 00 00
  token label        : User PIN (myCard
...

after:
$ pkcs11-tool --list-slots
Available slots:
Slot 0 (0x0): OMNIKEY AG CardMan 3121 00 00
  token label        : User PIN (myCard)
...


frankmorgner:
thanks

PR #1542, opened 2018-11-28
Fix some pkcs15-init issues

llogar:
1. pkcs15-init is using XKU but it should use cert KU to check private key usage instead.
2. Don't mark imported keys as ALWAYSSENSITIVE and NEVEREXTRACTABLE as they are not.
3. When importing keys from PKCS#12 files (with several certs inside), use consecutive IDs for additional certificates (instead of starting from 45).


frankmorgner:
thanks

PR #1543, opened 2018-11-28
Add per card_driver config

llogar:
Make `pkcs11_enable_InitToken` and `user_pin_unblock_style` configurable per card_driver, so different cards can coexist peacefully


frankmorgner:
If I understand `user_pin_unblock_style` correctly, then this is an option that doesn't depend on a specific card. It depends on the calling application's interpretation of PKCS#11. If so, it should be kept as global option rather than card driver specific option. Similarly, `pkcs11_enable_InitToken` is a matter of whether you want to allow initializing the card via PKCS#11 rather than being capable of doing so (only the latter would depend on the actual card).

What's your use case for this change?

llogar:
It was my understanding that `pkcs11_enable_InitToken` is used to separate cards than **can** (rather than **may**, as you pointed out) be initialised via `C_InitToken()`. 
In any case, one could use _pkcs15-init_ to initialise the card irrespective of `pkcs11_enable_InitToken` setting, so why disabled by default?
Also, isn't the role of SO (and SO PIN) to prevent unauthorized token initialisation? 

As for `user_pin_unblock_style`, I just wanted to be able to use _pkcs11-tool_ (beeing the "default", included PKCS#11 utility) for PIN unblocking (with the default _opensc.conf_ settings) which was not the case. Hence the change to get the default config work with my card (IsoApplet) and leave all the other cards/settings as they were. I guess, I was wrong about this one.

Jakuje:
The `pkcs11_enable_InitToken` is basically used to display the card if no other card driver could recognize the card. I am not sure if it somehow plugs into the `C_InitToken()` -- the configuration option is used only in the card detection functions, when no other card reader picks this card so I do not think it makes any sense to add this option under any condition matching the card driver.

llogar:
That may be the case, but I have observed that with `pkcs11_enable_InitToken` set to `no`, uninitialised IsoApplet card is not even detected (`pkcs11-tool --list-slots` displays `slot 0: (empty)`). Only when setting `pkcs11_enable_InitToken` to `yes`, the card gets detected and properly displayed as `token state:   uninitialized`, so it can then be initialised with `pkcs11-tool --init-token ...`.

Again. I would expect, that one could use "out of the box" _pkcs11-tool_ to initialise and/or unblock cards without changing _opensc.conf_?

And that's partially why I made this PR - to be able to use `C_InitToken()` with IsoApplet cards (sure, I could use _pkcs15-init_, but why use two different utils), and leave settings unchanged for other cards.

frankmorgner:
So you agree that the default for `user_pin_unblock_style` should *not* be changed, right? In PKCS#11, unblocking a PIN is not clearly defined, this is why we have the configuration option. If you want to unblock the PIN with OpenSC, there is also `pkcs15-tool --unblock-pin`.

`pkcs11_enable_InitToken` was introduced with 77bed452836c56752e2c3dd2e610bae5237b5ed5 (0.13.1). Before this change one could argue that calling `SC_CARDCTL_PKCS11_INIT_TOKEN` was possible by default.

I think it would be good to
- change the default back to allow initialization
- remove `pkcs11_enable_InitToken` and just use the `read_only` configuration flag, i.e. we should abort `C_InitToken()` if `CKF_WRITE_PROTECTED` is set for the token.

llogar:
No, I agree, that there is no need for `user_pin_unblock_style` to be per card-driver configurable. But I think the default should be changed to `set_pin_in_unlogged_session`. A real world deployment would probably require a customized _opensc.conf_ anyway, but at least stock _pkcs11-tool_  (instead of, or alongside _pkcs15-tool_) can be used to unlock the PIN without the need to change any config files.

Perhaps during `C_Initalize()` a program/executable filename could be read and `user_pin_unblock_style` set accordingly?

I agree, that `pkcs11_enable_InitToken` default value should be changed to `yes` and that there is also no need to make it per card-driver configurable. However I don't know how this affects what Jakuje said above. Also, why make additional check against `read_only` flag? If the card does not support `C_InitToken()` or is truly read only then the initialisation will fail anyway.

Jakuje:
Can we summarize what are the outcomes, what changes you suggest and update the PR accordingly? My understanding is the following:

 * We do not need per-driver configuration
 * We should change the `user_pin_unblock_style` defaults (even though there were some reasons for and against)
 * We should change default of `pkcs11_enable_InitToken` to `yes`

From my reading through the code, I do not see that changing the value of `pkcs11_enable_InitToken` would allow or prevent you from calling per-driver SC_CARDCTL_PKCS11_INIT_TOKEN, but prevents you from calling the "generic initialization procedure", which could be quite dangerous if your card requires a different initialization procedure. This is a reason why I believe we should not change the defaults here.

For the PIN unblock, I do not have experience how widely supported this is among our drivers. If we have larger set of drivers, which we would make easier to use, it could make sense to flag the ones that do not support this inside of the driver.

dengert:
If you are not sure if a driver can do what you ask, and you are going to have to look at the drivers anyway, would it may more sense to add   SC_CARD_CAP_* flags?  For the init and the UNBLOCK.

Jakuje:
> If you are not sure if a driver can do what you ask, and you are going to have to look at the drivers anyway, would it may more sense to add SC_CARD_CAP_* flags? For the init and the UNBLOCK.

For the UNBLOCK certainly. After a bit more thinking, having similar thing for the INIT_TOKEN would make sense too. The driver can already define the handler for INIT_TOKEN if they need specific initialization (only sc-hsm does now). But if we know the card can be initialized using the generic method, the driver should advertise this, rather than global (previously per-atr) configuration option.

llogar:
@Jakuje what you suggest looks reasonable to me. 

However I would also like to see that one can use _pkcs11-tool_ to initalize and/or unblock the card (if supported by the driver, of course) without changing the _opensc.conf_ file first.

llogar:
As suggested, drivers can now announce `C_InitToken` support using the 
`SC_CARD_CAP_ENABLE_INIT_TOKEN` flag and also unblock style using one of the 
```
#define SC_CARD_CAP_PIN_UNBLOCK_UNLOGGED_SETPIN><------>0x00004000
#define SC_CARD_CAP_PIN_UNBLOCK_SCONTEXT_SETPIN><------>0x00008000
#define SC_CARD_CAP_PIN_UNBLOCK_SO_LOGGED_INITPIN<----->0x00010000
```
flags. 

However, this can still be overriden using the _opensc.conf_ file.
_IsoApplet_ driver is the first and only one at the moment to set `SC_CARD_CAP_ENABLE_INIT_TOKEN` flag, so one can use `pkcs11-tool --init-token ...` to initialize the card without any changes to _opensc.conf_ (that were required prior to this patch).

src/libopensc/opensc.h
```diff
@@ -562,6 +562,13 @@ struct sc_reader_operations {
 /* Card (or card driver) supports key unwrapping operations */
 #define SC_CARD_CAP_UNWRAP_KEY			0x00001000
 
+/* Card (or card driver) can be initialized using C_InitToken() */
+#define SC_CARD_CAP_ENABLE_INITTOKEN			0x00002000
```
frankmorgner, line 566: If I understand correct, then this flag indicates whether the card can be initialized using the generic initialization method from here:
https://github.com/OpenSC/OpenSC/blob/e98f6733d9410b64249e7ca9028071a69ed8f004/src/pkcs11/framework-pkcs15.c#L1887-L1892
And this works if we have a card specific initializer in pkcs15init, right?
https://github.com/OpenSC/OpenSC/blob/e98f6733d9410b64249e7ca9028071a69ed8f004/src/pkcs15init/pkcs15-lib.c#L138-L171
Shouldn't we then use the latter list to determine the intialization capabilities rather than setting a flag in the generic driver, should we?

src/libopensc/opensc.h
```diff
@@ -562,6 +562,13 @@ struct sc_reader_operations {
...
+/* Card (or card driver) pin unblock options */
+#define SC_CARD_CAP_PIN_UNBLOCK_UNLOGGED_SETPIN		0x00004000
+#define SC_CARD_CAP_PIN_UNBLOCK_SCONTEXT_SETPIN		0x00008000
+#define SC_CARD_CAP_PIN_UNBLOCK_SO_LOGGED_INITPIN	0x00010000
+
```
frankmorgner, line 571: If I see this correctly, then you're not using the new flags in any card driver, so they don't change anything meaningful at the momement, right?

I'm not sure if its a good idea to let the PKCS#11 stuff drip through to the card driver. If the usages of `C_SetPIN` or `C_InitPIN` are ambigious, then there should be an implicit contract between the PKCS#11 application and the PKCS#11 module on how to use it correctly. At the card driver level, however, the uppser layers (PKCS#11) should have sorted out whether a PUK or an SOPIN was passed.

dengert:
If we are going to let some init card code run for some browser we must be absolutely positive that the card in uninitialized. The last thing we want to happen, is some unrelated card is trashed by this init code unintentionally with default OpenSC configuration. 

IMHO at must at least:
-  require the user or computer admin to have intentionally turned on this feature.
-  No other cards are in any other reader, (or user or admin overides this too.)
-  Card has no unknown AIDs to the init code. i.e. if Java card, check what AIDs are on card, and don't initialize if the init code is not sure.
-  A list of know AIDs supported by OpenSC, for other applets should also be checked, and the init code not run on this card if it does not know what to do. 

In other words, put responsibility on user or local admin for any damage done by this init code.
 
We do this now by requiring pkcs15_init be run by user, who then takes responsibility for their actions.

mouse07410:
@dengert couldn't agree more. This is a minefield.

llogar:
Ok, then forget about it, and let's just add a bold warning everytime `pkcs11-tool --init-token` is executed that "**option --init-token is disabled. Enable it in _opensc.conf_ by setting pkcs11_enable_InitToken=yes**"? 

The same goes for unlocking the pin by using "`pkcs11-tool --unlock-pin`" which also does not work out of the box: "**option --unlock-pin is nonfunctional. Make it work by setting user_pin_unblock_style = set_pin_in_unlogged_session in _opensc.conf_**"

It happened to me more than once, that I was scratching my head why I can't initialize and/or unlock the token using `pkcs11-tool` before realizing I forgot to change these settings in config file.

frankmorgner:
My comments were actually going into the opposite direction, I'd rather like to have a single module with everything (including initialization) working out of the box. I don't think it is our responsibility to dictate all use cases and workflows. After all, this is a module which enables applications to implement whatever they want to do with our tokens.

Yes, I once accidentally deleted a certificate from a card with Firefox, which is why I made the module respect the "readonly" flag from PKCS#15 to fix this. Other cards can pick this up in their profile as well if they want to protect from unwanted access. But Firefox also allows uploading key/certificate to a card, which I'd really like to use for personalizing a card rather than some command line tool.

dengert:
https://github.com/llogar,  I agree with "let's just add a bold warning everytime"  That puts the onus on the user or admin, that there is a security risk in having that on by default.  It opens up the possibility that other PKCS11 applications like a browser, could also take advantage of the flag. 

You as a user or admin could also have a second opensc.conf used only for card administration that has flags like this set,  so you could have it both ways. An alias or script something like:
`OPENSC_CONF=/<path>/opensc-for-card-admin.conf pkcs11-tool $*`

I also agree with @frankmorgner on: "I made the module respect the "readonly" flag from PKCS#15 to fix this."

I am trying to look out for all the other users who do not want or expect OpenSC to be able trash their cards.

frankmorgner:
I think there is not enough need for this specific change.

If you think some default options should be different for certain card drivers, just customize the defaults in the card specific initialization routine or the PKCS#11 module itself.

PR #1549, opened 2018-12-05
PIV - Improved Card Matching for Dual CAC/PIV and PIVKEY cards.

dengert:
Not all PIV applets are the same. Different versions of NIST 800-73 and improperly implemented
or not implemented required features of NIST 800-73 cases problems. Have a look at the card_issues
listed in card-piv.c. The PIV driver has tried to detect the differences based on clues found in
the ATR historical bytes and vendor version numbers for some cards.

At the same time it has tried to support the possibility there are multiple applets
on a card that the user may want to use at the same time from different applications.
This has lead to some detection problems with Dual CAC/PIV cards. The same cards
sold by the vendor may have only a PIV applet that may not be the same PIV applet that
is on the Dual PIV/CAC cards.

http://www.cac.mil/Portals/53/Documents/CAC-utilziation-and-variation-matrix-v2.03-20May2016.doc
defines a number of official CAC cards in active service. A table of the ATRs for these is now used
to detect these cards. The PIV version of the CCC is also read to see if any CAC PKI objects
are defined in the CCC, indicating it is a Dual CAC/PIV, even if the ATR is not listed.

A more conservative approach to try and handle multiple applets on a card is used. Based
on issues with the implementation of the PIV applet this may not be possible to do.
So for many cards no additional detection will be done at the start of every transaction,
and the login state can not be detected correctly.

ATRs for PIVKEY are also in the match table, as these cards have a lot of issues.

Other PIV cards in the future or not yet tested may not be covered properly by this patch.
Extra debugging was added with "PIV_MATCH" to help with these other cards.
With "debug = 7;", `grep PIV_MATCH opensc-debug.log` can be used to see how a card
type and card_issues are derived.

On branch piv-improved-matching
 Changes to be committed:
	modified:   card-piv.c
	modified:   cards.h

Tested using NIST demo cards, Oberthur cards that have a CCC that lists CAC in CCC F0,
Oberthur PIV card with a CCC copied from a Dual CAC/PIV card, Yubico NEO, Yubkey 4, PIVKEY C980, T600 and T800. These last two are "Feitian eJAVA Tokens" with the Taglio PIVKEY PIV applet.
 

##### Checklist


src/libopensc/card-piv.c
```diff
@@ -2635,6 +2681,81 @@ static int piv_process_discovery(sc_card_t *card)
...
+						priv->ccc_flags  |= PIV_CCC_F3_CAC_PKI;
+				}
+				break;
+		}
+		p = p + *(p+1) + 2;
+	}
+
+err:
+	LOG_FUNC_RETURN(card->ctx, r);
+}
```
frankmorgner: This function contains multiple possibilities for out of bounds reads!

src/libopensc/card-piv.c
```diff
@@ -3006,8 +3143,9 @@ static int piv_match_card_continued(sc_card_t *card)
 			 */
 			else if (card->reader->atr_info.hist_bytes_len >= 4
 					&& !(memcmp(card->reader->atr_info.hist_bytes, "sfte", 4))) {
-				type = SC_CARD_TYPE_PIV_II_GI_DE;
+				type = SC_CARD_TYPE_PIV_II_GI_DE_DUAL_CAC;
 			}
+#endif
```
frankmorgner: delete dead code

frankmorgner:
The whole PIV situation looks like a big mess. How does it come that we are taking so much care when Apple, for example, is fine with such a simple [PIV driver](https://developer.apple.com/library/archive/samplecode/PIVToken/Introduction/Intro.html)? Are we doing something wrong, are we overly picky regarding features and implementation?

dengert:
We are overly picky with:
- Trying to support multiple applets on the same card
- Trying to allow different applications to access different applets on the same PIV card  Yubico with  PIV and OpenPGP  and PIV and CAC
- Trying to avoid caching the PIN, i.e. keep the login state
- Trying to also support PIN PAD readers
- Having OpenSC driver matching routines for other cards sending APDUs to the PIV cards causing problems with login state
- Trying to deal with all the different PIV applet implementations with bugs. There are a few in github...

src/libopensc/card-piv.c
dengert: Fixed

src/libopensc/card-piv.c
dengert: Fixed

src/libopensc/card-piv.c
```diff
@@ -3029,10 +3164,16 @@ static int piv_match_card_continued(sc_card_t *card)
 				}
 			}
 		}
+		sc_debug(card->ctx,SC_LOG_DEBUG_MATCH, "PIV_MATCH card->type:%d type:%d r:%d\n", card->type, type, r);
+
 		if (type == -1)
```
Jakuje: Missing braces around the following block

src/libopensc/card-piv.c
frankmorgner: All this byte buffer dancing and length checking would be way more readable if you'd use `sc_asn1_read_tag` also for simple tags. See https://github.com/OpenSC/OpenSC/blob/a7a9326f35a2c46ba06fe480bf0d53b6f29edf91/src/libopensc/iso7816.c#L336-L419 for example.

src/libopensc/card-piv.c
```diff
@@ -3064,38 +3206,105 @@ static int piv_match_card_continued(sc_card_t *card)
...
 	if (i < 0) {
 		/* Detect by selecting applet */
 		i = piv_find_aid(card);
 	}
 
+	sc_debug(card->ctx,SC_LOG_DEBUG_MATCH, "PIV_MATCH card->type:%d i:%d CI:%08x r:%d\n", card->type, i, priv->card_issues, r);
 	if (i >= 0) {
+		 /* We now know PIV AID is active, test CCC object  800-73-* say CCC is required */
+		switch (card->type)  {
+			int iccc = 0;
```
Jakuje: I can not build the source with the `gcc (GCC) 8.2.1 20181105 (Red Hat 8.2.1-5)`, which is causing the following error:
```
card-piv.c: In function 'piv_match_card_continued':
card-piv.c:3271:8: error: statement will never be executed [-Werror=switch-unreachable]
    int iccc = 0;
        ^~~~
```

Jakuje:
@dengert Thank you for the work through the specification to identify the correct cards. I added few comments to the code that need some more attention but after fixing them, I report success:
```
[root@rhel8-2 piv2]# pkcs11-tool --test --login
Logging in to "FLUORINE.JANE.F.2001441054".
Please enter User PIN:
  ERR: C_GenerateRandom failed: CKR_DATA_INVALID (0x20)
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
  testing key 0 (PIV AUTH key)
  testing signature mechanisms:
    RSA-X-509: OK
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
  testing key 1 (2048 bits, label=SIGN key) with 1 signature mechanism
Logging in to "FLUORINE.JANE.F.2001441054".
Please enter context specific PIN:
    RSA-X-509: OK
  testing key 2 (2048 bits, label=KEY MAN key) with 1 signature mechanism -- can't be used to sign/verify, skipping
  testing key 0 (PIV AUTH key)
    RSA-X-509: OK
    RSA-PKCS: OK
    SHA1-RSA-PKCS: OK
    MD5-RSA-PKCS: OK
    RIPEMD160-RSA-PKCS: OK
  testing key 1 (SIGN key) with 1 mechanism
Logging in to "FLUORINE.JANE.F.2001441054".
Please enter context specific PIN:
    RSA-X-509: OK
  testing key 2 (KEY MAN key) with 1 mechanism -- can't be used to sign/verify, skipping
Unwrap: not implemented
  testing key 0 (PIV AUTH key)
    RSA-X-509: OK
    RSA-PKCS: OK
  testing key 1 (SIGN key)
    RSA-X-509: Logging in to "FLUORINE.JANE.F.2001441054".
Please enter context specific PIN:
OK
    RSA-PKCS: Logging in to "FLUORINE.JANE.F.2001441054".
Please enter context specific PIN:
OK
  testing key 2 (KEY MAN key)
    RSA-X-509: OK
    RSA-PKCS: OK
1 errors
```
Only issue is with the `C_GenerateRandom`, which still reports errors. Could this be also blacklisted as nonworking for the dual cards?

dengert:
I could add the `CI_NO_RANDOM` to the `card_issues` for all the CAC cards. This PR   does turn off ``SC_CARD_CAP_RNG`. 

Note that PKCS\#11 says `C_SeedRandom`  can return `CKR_RANDOM_SEED_NOT_SUPPORTED` and  `CKR_RANDOM_NO_RNG`  and  `C_GenerateRandom` can return `CKR_RANDOM_NO_RNG`. But framework-pkcs15.c does not return either, and  `sc_to_cryptoki_error_conversion` does not have way to convert a `SC_ERROR_` code to either.  OpenSC has `SC_CARD_CAP_RNG` which cold be used to set `
CKR_RANDOM_NO_RNG` 

So to do this right needs some work in OpenSC internals and pkcs11-tool to address this correctly.

src/libopensc/card-piv.c
```diff
@@ -2635,6 +2681,92 @@ static int piv_process_discovery(sc_card_t *card)
...
+	body = rbuf;
+	if ((r = sc_asn1_read_tag(&body, rbuflen, &cla_out, &tag_out,  &bodylen)) != SC_SUCCESS) {
+		sc_log(card->ctx, "DER problem %d",r);
+		r = SC_ERROR_INVALID_ASN1_OBJECT;
+		goto err;
+	}
+
+	priv->ccc_flags |= PIV_CCC_FOUND;
+
+	/* CCC boby has 1 byte tags, and 1 byte lengths */
```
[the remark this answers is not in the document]
dengert: Fixed with rewrite of `piv_parse_ccc`

src/libopensc/card-piv.c
```diff
@@ -2635,6 +2681,92 @@ static int piv_process_discovery(sc_card_t *card)
...
+	p = body;
+	end = p + bodylen;
+	while (p <= (end - 2) && (next = (p + *(p + 1) + 2)) <= end ) {
```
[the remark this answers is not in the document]
dengert: piv_parse_ccc was rewritten to use sc_simpletlv_read_tag

src/libopensc/card-piv.c
dengert: Fixed added brackets.

src/libopensc/card-piv.c
dengert: Should be fixed, moved the declare statement.

dengert:
CI_NO_RANDOM added for Dual PIV/CAC cards.  The PIV "9B" admin key was used to get a challenge. Some cards do not have one.

src/libopensc/card-piv.c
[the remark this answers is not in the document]
dengert: Note: there is a cast in the code:
```
r = sc_simpletlv_read_tag((u8**)&body, end - body , &tag, &len);
```
because both of these are used to parse the CCC:
```
int sc_asn1_read_tag(const u8 ** buf, size_t buflen, unsigned int *cla_out,
                    unsigned int *tag_out, size_t *taglen);
```
```
int sc_simpletlv_read_tag(u8 **buf, size_t buflen, u8 *tag_out, size_t *taglen);
```

I would suggest that the `sc_simpletlv` use the same `const u8 ** buf` but could be a separate PR.

src/libopensc/card-piv.c
```diff
@@ -2635,6 +2681,92 @@ static int piv_process_discovery(sc_card_t *card)
...
+err:
+	LOG_FUNC_RETURN(card->ctx, r);
+}
+
+static int piv_process_ccc(sc_card_t *card)
+{
+	int r = 0;
+	u8 * rbuf = NULL;
+	size_t rbuflen = 0;
+	/* piv_private_data_t * priv = PIV_DATA(card); */
```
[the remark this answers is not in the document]
dengert: Fixed

src/libopensc/card-piv.c
```diff
@@ -2715,33 +2720,32 @@ static int piv_process_discovery(sc_card_t *card)
...
+	/* CCC  entries are simple tlv */
+	end = body + bodylen;
+	
+	for(; (body < end); body += len) {
+	
+		r = sc_simpletlv_read_tag((u8**)&body, end - body , &tag, &len);
+		if (r < 0)
+			goto err;
+		switch (tag) {
+			case PIV_CCC_TAG_F0:
```
Jakuje: nit: Switch and case can have the same indent to avoid too much indentation. I would recommend it in this case.

src/libopensc/card-piv.c
```diff
@@ -3064,38 +3226,104 @@ static int piv_match_card_continued(sc_card_t *card)
...
+		 * Some PIV don't support DISCOVERY and return 
+		 * SC_ERROR_INCORRECT_PARAMETERS. Any error 
+		 * including SC_ERROR_FILE_NOT_FOUND means we cannot use discovery 
 		 * to test for active AID.
 		 */
 		int i7e = piv_find_discovery(card);
 
-		if (i7e != 0 && i7e !=  SC_ERROR_FILE_NOT_FOUND) {
+		sc_debug(card->ctx,SC_LOG_DEBUG_MATCH, "PIV_MATCH card->type:%d i7e:%d CI:%08x r:%d\n", card->type, i7e, priv->card_issues, r);
+		if (i7e != 0) {
```
frankmorgner: `SC_SUCCESS` instead of `0`?

src/libopensc/card-piv.c
```diff
@@ -3064,38 +3226,104 @@ static int piv_match_card_continued(sc_card_t *card)
...
+			 * CCC is required and all Dual PIV/CAC will have a CCC
+			 * Currently Dual PIV/CAC are based on NIST 800-73-1 which does not have Discovery or History
+			 */
+			case SC_CARD_TYPE_PIV_II_GENERIC: /* i.e. really dont know what this is */
+			case SC_CARD_TYPE_PIV_II_HIST:
+			case SC_CARD_TYPE_PIV_II_GI_DE:
+			case SC_CARD_TYPE_PIV_II_GEMALTO:
+			case SC_CARD_TYPE_PIV_II_OBERTHUR:
+				iccc = piv_process_ccc(card);
+				sc_debug(card->ctx,SC_LOG_DEBUG_MATCH, "PIV_MATCH card->type:%d iccc:%d ccc_flags:%08x CI:%08x r:%d\n",
```
frankmorgner: saving the return to `iccc` is useless. `piv_process_ccc` will log entry/exit points itself. card issues will be logged later. Generally speaking, I'd be more conservative with debug messages. However, if it doesn't clutter up the output for my cards, then keep the messages if they are useful to you.

src/libopensc/card-piv.c
```diff
@@ -3029,10 +3183,17 @@ static int piv_match_card_continued(sc_card_t *card)
 				}
 			}
 		}
-		if (type == -1)
-			type = SC_CARD_TYPE_PIV_II_GENERIC;
+		sc_debug(card->ctx,SC_LOG_DEBUG_MATCH, "PIV_MATCH card->type:%d type:%d r:%d\n", card->type, type, r);
+
+		if (type == -1) {
+			/* use known ATRs  */
+			i = _sc_match_atr(card, piv_atrs, &type);
```
Jakuje: Unused return value.

mouse07410: I suggest printing it via debug output.

src/libopensc/card-piv.c
```diff
@@ -2222,11 +2266,33 @@ static int piv_get_challenge(sc_card_t *card, u8 *rnd, size_t len)
...
+	if (priv->card_issues & CI_NO_RANDOM) {
+		r = SC_ERROR_NOT_SUPPORTED;
+		LOG_TEST_GOTO_ERR(card->ctx, r, "No support for random data");
+	}
+
 	/* NIST 800-73-3 says use 9B, previous verisons used 00 */
 	r = piv_general_io(card, 0x87, 0x00, 0x9B, sbuf, sizeof sbuf, &rbuf, &rbuf_len);
+	/*
+	 * piv_get_challenge is called in a loop. 
+	 * some cards may allow 1 challenge expecting it to be part of
```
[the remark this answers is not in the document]
dengert: Fixed

src/libopensc/card-piv.c
```diff
@@ -2635,6 +2701,91 @@ static int piv_process_discovery(sc_card_t *card)
...
+		goto err;
+	}
+
+	priv->ccc_flags |= PIV_CCC_FOUND;
+
+	/* CCC  entries are simple tlv */
+	end = body + bodylen;
+	
+	for(; (body < end); body += len) {
+
```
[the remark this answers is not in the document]
dengert: Fixed

src/libopensc/card-piv.c
```diff
@@ -2635,6 +2701,91 @@ static int piv_process_discovery(sc_card_t *card)
...
+		sc_log(card->ctx, "DER problem %d",r);
+		r = SC_ERROR_INVALID_ASN1_OBJECT;
+		goto err;
+	}
+
+	priv->ccc_flags |= PIV_CCC_FOUND;
+
+	/* CCC  entries are simple tlv */
+	end = body + bodylen;
+
```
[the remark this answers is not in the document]
dengert: Fixed

src/libopensc/card-piv.c
dengert: Fixed, test i rather then type in next line.

src/libopensc/card-piv.c
dengert: Fixed use `if (i7e < 0) {` as is done in other places when testing  return of `piv_find_discovery(card)`

src/libopensc/card-piv.c
dengert: The sc_debug messages added in this PR are there so one can do 
`grep PIV_MATCH`  opensc-debug.log` to see how the the matching was done. 
I expect there may be problems with cards I do not have and this makes it easier for a user to submit a bug. 

And it is only for the PIV cards.

dengert:
@frankmorgner, I am dismissing your comment:  "Your code doesn't build due to compiler warnings." This was caused by Appveyor because I forced merged a new set of commits.  "All checks have passed" and  "This branch has no conflicts with the base branch"

Can we get this merged in soon. @Jakuje has approved it and I would like to get in into master so other can easily try it too. 

Thanks.

PR #1553, opened 2018-12-08
openpgp-tool: use binary OR for calculating 32-bit integers from byte

marschap:
This commit should also fix 
CID 325863 (#1 of 1): Unintended sign extension (SIGN_EXTENSION)sign_extension: Suspicious implicit sign extension: data[0] with type u8 (8 bits, unsigned) is promoted in (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3] to type int (32 bits, signed), then sign-extended to type long (64 bits, signed). If (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3] is greater than 0x7FFFFFFF, the upper bits of the result will all be 1.

### DIFF
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -239,7 +239,7 @@ static char *prettify_algorithm(u8 *data, size_t length)
 static char *prettify_date(u8 *data, size_t length)
 {
 	if (data != NULL && length == 4) {
-		time_t time = (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3];
+		time_t time = (time_t) (data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3]);
 		struct tm *tp;
 		static char result[64];	/* large enough */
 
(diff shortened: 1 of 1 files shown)


src/tools/openpgp-tool.c
```diff
@@ -312,9 +312,9 @@ static char *prettify_serialnumber(u8 *data, size_t length)
 {
 	if (data != NULL && length >= 4) {
 		static char result[15];	/* large enough for even 2*3 digits + separator */
-		unsigned int serial = (data[0] << 24) + (data[1] << 16) + (data[2] << 8) + data[3];
+		unsigned long serial = (unsigned long) (data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3]);
```
frankmorgner: I think you want `bebytes2ulong()`

marschap: @frankmorgner I need your help here:
- I #included "libopensc/internal.h"
- replaced the code in question with bebytes2ulong()
- I tried to correctly update the Makefile.am
But it fails to compile.

Can you give any hints (or even better: fix it), also for Makefile.mak, where I know even less.

frankmorgner: You have 2 options:
- either rewrite ulong2bebytes so that it is a macro instead of a function
- or add `bebytes2ulong` to `src/libopensc/libopensc.exports`

marschap: @frankmorgner I consider both options as sub-optimal compared to the original version of the patch.
* exporting bebytes2ulong()
  bebytes2ulong is currently not exported, and declared in a header file named `internal.h`.
  In my opinion these 2 facts give strong hints that this function should not be exported.
  In my opinion this makes very much sense, as the libraries may be is dynamically loaded into the name space of other programs, and additional exported functions not part of the original API, especially those not having a special prefix like `sc_...` may collide with functions in the loading program leading to strange effects.
* bebytes2ulong() as macro
  This removes the possibility for compile time checks on the arguments which I'd rather keep.

For that reason I'd rather have the original patch included - I'll re-push it

frankmorgner:
OK

Issue #1550, opened 2018-12-07
Regression: some code not compiling on Debian testing


larssilven:
### Problem Description

Error when compiling src/pkcs11/framework-pkcs15.c with gcc version 8.2.0 (Debian 8.2.0-9) from debian testing.

### Proposed Resolution

Fix the code. This is a regression from commit ba7704291180882dbfa18705f4169a88bc5d61ae
This patch makes it compile again:
```
@@ -1089,18 +1089,11 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
                        pin_info = NULL;
                }
                else   {
-                       if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
-                               /* Trim tokeninfo->label to make right parenthesis visible */
-                               char tokeninfo_label[sizeof label];
-                               int len;
-                               snprintf(tokeninfo_label, sizeof(tokeninfo_label), "%s", p15card->tokeninfo->label);
-                               for (len = strlen(tokeninfo_label) - 1; len >= 0 && tokeninfo_label[len] == ' '; len--) {
-                                       tokeninfo_label[len] = 0;
-                               }
+                       if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
                                snprintf(label, sizeof(label), "%.*s (%s)",
                                        (int) sizeof(auth->label), auth->label,
-                                       tokeninfo_label);
-                       } else
+                                       p15card->tokeninfo->label);
+                       else
                                /* The PIN label is empty or says just non-useful "PIN" */
                                snprintf(label, sizeof(label), "%s", p15card->tokeninfo->label);
                        slot->token_info.flags |= CKF_LOGIN_REQUIRED;
```

### Steps to reproduce

make

### Logs
```
make[3]: Leaving directory '/home/lars/work/PWE/OpenSC/src/libopensc'
Making all in pkcs11
make[3]: Entering directory '/home/lars/work/PWE/OpenSC/src/pkcs11'
  CC       opensc_pkcs11_la-framework-pkcs15.lo
framework-pkcs15.c: In function ‘pkcs15_init_slot’:
framework-pkcs15.c:1100:43: error: ‘%s’ directive output may be truncated writing up to 264 bytes into a region of size between 9 and 263 [-Werror=format-truncation=]
     snprintf(label, sizeof(label), "%.*s (%s)",
                                           ^~
framework-pkcs15.c:1102:6:
[...]

PR #1554, opened 2018-12-08
framework-pkcs15: fix warning

marschap:
This commit
* fixes `-Werror=format-truncation=` warning in `pkcs15_init_slot()`
  This fixes issue #1550
* flexibilizes the characters to truncate from the label in `pkcs15_init_slot()`:
  not only blanks but spaces according to `isspace()`


frankmorgner:
thanks

PR #1555, opened 2018-12-09
OpenPGP cleanup

marschap:
The commits in this PR harmonize `card-openpgp.c`'s error handling:
* they convert manual coding to the LOG_TEST_*() macros
* they initialize pointers to NULL
* they explicitly compare pointers with NULL

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2870,22 +2870,21 @@ out:
 static int
 pgp_store_key(sc_card_t *card, sc_cardctl_openpgp_keystore_info_t *key_info)
 {
-	sc_context_t *ctx = card->ctx;
 	sc_cardctl_openpgp_keygen_info_t pubkey;
 	u8 *data = NULL;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -2948,15 +2933,13 @@ pgp_store_key(sc_card_t *card, sc_cardctl_openpgp_keystore_info_t *key_info)
...
 	                           key_info->rsa.e, 8*key_info->rsa.e_len, key_info->key_id);
 
-	sc_log(ctx, "Update card algorithms.");
+	sc_log(card->ctx, "Update card algorithms.");
 	pgp_update_card_algorithms(card, &pubkey);
 
-out:
-	if (data) {
+err:
+	if (data != NULL)
```
frankmorgner: this check is technically not needed, `free()` works on `NULL`

marschap: @frankmorgner  issue addressed & fixed with the latest commit

PR #1566, opened 2019-01-01
OpenPGP: cleanup "magic" constants

marschap:
Hi,

the attached PR replaces magic constants & tricks relying on C's string semantics with 
* symbolic names of well known DOs and other objects
* proper assignments using ushort2bebytes()

Apart from those cleanups, which make the code easier to understand, it does not change the code logic

Thanks in advance for pulling this PR into master.
Peter

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -3066,7 +3066,7 @@ gnuk_delete_key(sc_card_t *card, u8 key_id)
 {
 	sc_context_t *ctx = card->ctx;
 	int r = SC_SUCCESS;
-	char *data = NULL;
+	u8 data[4] = { 0x4D, 0x02, 0x00, 0x00 };
 
 	LOG_FUNC_CALLED(ctx);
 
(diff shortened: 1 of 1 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -2819,21 +2816,20 @@ pgp_build_extended_header_list(sc_card_t *card, sc_cardctl_openpgp_keystore_info
...
+			break;
+		case SC_OPENPGP_KEY_AUTH:
+			ushort2bebytes(data, DO_AUTH);
+			break;
+		default:
+			sc_log(ctx, "Unknown key id %d.", key_info->key_id);
+			LOG_TEST_GOTO_ERR(ctx, SC_ERROR_INVALID_ARGUMENTS, "Invalid key id");
 	}
-	memcpy(data + 2, tlv_7f48, tlvlen_7f48);
+	memcpy(data + 2,               tlv_7f48, tlvlen_7f48);
```
Jakuje: Being late for the party, I would like to see these bogus spaces removed.
Additionally, having the PR open for more than 24 hours can allow others to catch these "typos".

frankmorgner: @marschap will say that `tlv_7f48` should be aligned with `tlv_5f48` below...

PR #1571, opened 2019-01-07
GoID personalization with fingerprint

frankmorgner:
This PR allows personalization of Bundesdruckerei's GoID card with a fingerprint. `sc-hsm-tool` is extended to configure the fingerprint for the PKI application.

##### Checklist


frankmorgner:
@CardContact FYI

PR #1573, opened 2019-01-07
Fix PKCS#11 Object Restrictions

dengert:
Framework-pkcs15.c silently ignores adding objects if MAX_OBJECTS
is exceeded while creating the fw_data objects. This simple fix
is to change the MAX_OBJECTS from 64 to 128. A better fix would
be to realloc the objects arrays as needed.

__pkcs15_create_data_object and __pkcs15_create_secret_key_object
now return rv like the other  __pkcs15_create_*_object routines.

pkcs15_dobj_get_value now calls sc_pkcs15_read_data_object just like
the other pkcs15_*_get_value routines. The problem was introduced
in 0c3412bb 2018-04-09 which added:
 `return sc_to_cryptoki_error(SC_SUCCESS, "C_GetAttributeValue");`
before trying to read the data object.

The MAX_OBJECT problem was discovered while trying to use a new PIV
card with 24 standard cert objects and 10 other objects for a total
of 106 objects. Each cert object corresponds to a cert, pubkey,
private key, and the cert object itself for a possible 112 data objects.

The pkcs15_dobj_get_value was found while running:
running pkcs11-tool -r -y data --application-id 2.16.840.1.101.3.7.2.1.1
using git bisect to locate the bad commit.  The pkcs11 data objects are
created last from the pkcs15 objects which are a linked list with no limits.

 On branch fix-object-restrictions

Tested with PIV test cards, using `pkcs11-tool --test --login`
and  `pkcs11-tool -r -y data --application-id 2.16.840.1.101.3.7.2.1.1 > /tmp/cert.object`

##### Checklist

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -58,7 +58,7 @@ struct pkcs15_slot_data {
 	}                                       \
 	attr->ulValueLen = size;
 
-#define MAX_OBJECTS	64
+#define MAX_OBJECTS	128
 struct pkcs15_fw_data {
 	struct sc_pkcs15_card *		p15_card;
 	struct pkcs15_any_object *	objects[MAX_OBJECTS];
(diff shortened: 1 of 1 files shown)


frankmorgner:
So this PR doesn't fix the actual issue but instead postpones it for later...

Nevertheless, it looks OK for me.

Issue #1483, opened 2018-09-18
Fix the card matching for Coolkey and Muscle drivers


Jakuje:
### Problem Description

In reference to the comment https://github.com/OpenSC/OpenSC/issues/1377#issuecomment-394778180 from @dengert  I just talked with Bob and this is an issue of some Java cards that can not be handled in the applet itself. The SELECT APDU is handled by the Card itself so the applet can not do anything about it. This used to be an issue also for CAC or PIV cards and they no longer depend just on the return code from SELECT APDU. 

### Proposed Resolution

I will try to propose a patch for Coolkey driver soon so it can behave better, avoid mismatches and therefore moved above muscle. I do not know quite much anything about muscle nor I have the applet so I can not help in this way, but if the muscle is considered as active, I assume some muscle developers can do the same for muscle.

### Steps to reproduce

Soon after boot (WaitForSlotEvent + plugging the card in) or when running several sessions soon after each other from (p11test), the SELECT APDU returns 90 00 regardless the applet is actually available on the card. This can mismatch coolkey card as muscle or vice versa.

dengert:
Is there any additional command to a muscle or coolkey to determine which
applet is active?

After a select aid all drivers shoild check for response with correct AID.
If no data returned muscle and coolkey drivers need to check for something
else.

dengert:
For example is there any Java card command that could return what applets
are on the card?

Or does muscle or coolkey respond. To 7E command?

Jakuje:
> Or does muscle or coolkey respond. To 7E command?

I did not find any of the applets explicitly responding to this instruction.

Reading through the source code of both applet, I noticed that the GetStatus has a difference in handling the P1:

 * The [coolkey applet](https://github.com/dogtagpki/coolkey/blob/master/applet/src/com/redhat/ckey/applet/CardEdge.java#L1333) has this argument reserved for call to `GetMemory()`, which is not implemented and returns invalid instruction (`0x6D00`)
* The [muscle applet](https://github.com/martinpaljak/MuscleApplet/blob/d005f36209bdd7020bac0d783b228243126fd2f8/src/com/musclecard/CardEdge/CardEdge.java#L2115) requires this argument is set to zero, otherwise it returns incorrect P1 (`0x9C10`)

This should allow as to make coolkey matching more robust and move the driver a bit up.

[...]

PR #1574, opened 2019-01-08
Select MyEID applet with AID in myeid_init() to ensure that it is sel…

### LINKED ISSUE #1563: Failed to connect to card: Card is invalid or cannot be handled - Aventra MyEID
### Problem Description

A previously working MyEID smartcard suddenly returns the error "Card is invalid or cannot be handled".

### Proposed Resolution

Smartcard recognised successfully.

### Steps to reproduce

Most commands return the above error, including pkcs15-tool --dump.

The key line seems to be:

0x1124d55c0 02:14:30.828 [pkcs15-tool] card-myeid.c:171:myeid_init: Failed to get MyEID applet information.: -1210 (Card is invalid or cannot be handled)

This triggers opensc to keep looking for other drivers, even though this ATR matched.

### Logs

https://gist.github.com/minfrin/d16273b4772bc259d67799d2f7feb16a

hhonkanen:
…ected before sending other commands.

Fixes #1563 

Some other process might have selected Card Manager applet before using a MyEID card with OpenSC. In this case APDUs from OpenSC were routed to Card Manager instead of MyEID. Added code to select MyEID applet with AID right after ATR matching.

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -146,13 +146,48 @@ static int myeid_match_card(struct sc_card *card)
 	return 1;
 }
 
+static int
+myeid_select_aid(struct sc_card *card, struct sc_aid *aid, unsigned char *out, size_t *out_len)
+{
+	struct sc_apdu apdu;
+	unsigned char apdu_resp[SC_MAX_APDU_BUFFER_SIZE];
+	int rv;
+
+	/* Select application (deselect previously selected application) */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0x04, 0x00);
+	apdu.lc = aid->len;
+	apdu.data = aid->value;
+	apdu.datalen = aid->len;
+	apdu.resplen = sizeof(apdu_resp);
+	apdu.resp = apdu_resp;
+
+	rv = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, rv, "APDU transmit failed");
+	rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	LOG_TEST_RET(card->ctx, rv, "Cannot select AID");
+
+	if (*out_len > 0) {
+		if (*out_len < apdu.resplen)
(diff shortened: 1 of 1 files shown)


src/libopensc/card-myeid.c
```diff
@@ -146,13 +146,44 @@ static int myeid_match_card(struct sc_card *card)
...
+	apdu.resp = apdu_resp;
+
+	rv = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, rv, "APDU transmit failed");
+	rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	LOG_TEST_RET(card->ctx, rv, "Cannot select AID");
+
+	if (*out_len < apdu.resplen)
+		LOG_TEST_RET(card->ctx, SC_ERROR_BUFFER_TOO_SMALL, "Cannot select AID");
+	memcpy(out, apdu.resp, apdu.resplen);
```
frankmorgner: why request response data when you don't need it?

hhonkanen: I copied the code from the ias/ecc driver and just edited it a little. Didn't bother to remove the code that returns the response buffer, because SELECT AID command returns the FCI structure. We don't need it for now but actually some of the data in this structure might be useful in future development. I modified it a little to make the output buffer optional to avoid the unnecessary memcpy, new commit coming soon.

dengert:
You said: "We don't need it for now".  There are problems with the Muscle and Coolkey applets that will return `9000` to any select SELECT command. If one of these cards is present. If you don't check the FCI for your AID your driver can get a false positive from one of these cards.  But if you next command is a MyEID only command it should catch it there. 
See #1483

hhonkanen:
@dengert myeid_init() where this new SELECT AID command is executed is called after myeid_match_card(). Card matching is based entirely on ATR. The problem that we want to fix is that when we have already correctly found a MyEID card with myeid_match_card(), MyEID applet has got unselected. In this phase, if it is a MyEID card the new command selects MyEID applet on the cards. If for some reason we have got here with some other card than MyEID, the next command should fail anyway because it probably doesn't return the expected data. The next command is a GET DATA with MyEID specific parameters and requests applet information. But because Coolkey and Muscle have different ATRs, we shouldn't even get into this point with these cards.

PR #1584, opened 2019-01-17
Remove postecert and infocamere support because no longer issued

opensignature:
##### Checklist

- [ ] macOS tokend is tested


src/libopensc/pkcs15-pubkey.c
```diff
@@ -1617,9 +1617,9 @@ sc_pkcs15_convert_pubkey(struct sc_pkcs15_pubkey *pkcs15_key, void *evp_key)
 			return SC_ERROR_INTERNAL;
 		X = BN_new();
 		Y = BN_new();
-		if (X && Y && EC_KEY_get0_group(eckey))
-			r = EC_POINT_get_affine_coordinates_GFp(EC_KEY_get0_group(eckey),
-					point, X, Y, NULL);
+//		if (X && Y && EC_KEY_get0_group(eckey))
+//			r = EC_POINT_get_affine_coordinates_GFp(EC_KEY_get0_group(eckey),
+//					point, X, Y, NULL);
```
frankmorgner: Why did you remove this code?

opensignature: Unfortunately the italian way of providing a signature device from 2009 onwards was to lock-in to proprietary library that comes along with the device, based on Secure Messaging symmetric keys embedded in the PKCS11 library.
The cards of which I ask the removal of the support have been considered unusable starting from 1/7/2011 (unfortunately I have not yet found a document in English).
For the lines commented on pkcs15-pubkey.c it's my mistake, now I put it back as it was.

frankmorgner: So am I right saying that today
- there aren't any cards issued that were compatible with OpenSC and
- there aren't any cards in the field that were compatible with OpenSC and still have a valid certificate?

And I assume that the new proprietary library also has support for old cards so that users are still able to decrypt their old mails. Correct?

opensignature: > * there aren't any cards issued that were compatible with OpenSC and
> * there aren't any cards in the field that were compatible with OpenSC and still have a valid certificate?

Yes
 
> And I assume that the new proprietary library also has support for old cards so that users are still able to decrypt their old mails. Correct?

This risk does not exist because the certificate inserted in those cards had only the use not-repudiation, so no sign, encrypt or decrypt

NEWS
```diff
@@ -837,9 +836,6 @@ New in 0.10.0; 2005-10-31; Andreas Jellinghaus
   via TeleTrust compatible PC/SC interface.
 * Fixed OpenSSL behaviour in the configure
   script.
-* PKCS#15 emulation layer improvements and
-  a new driver for the Italian postecert
-  card.
 * New API documentation and generic documentation
```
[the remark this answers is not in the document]
frankmorgner: There's still overall one line that has been deleted from NEWS...

frankmorgner:
please move the OpenSSL compatibility stuff to `src/libopensc/sc-ossl-compat.h`. I think in its current version it's not compatible with LibreSSL

paride:
I am a bit confused as this PR is dated January 2019, but I have a Postecert smartcard issued in July 2019, and valid until June 2025. In any case it seems to be true that Poste isn't issuing smartcards anymore, so I don't think it's worth spending time on it.

Moreover I tried to use an older version of OpenSC which in still had support for Postecert cards, but still the card isn't recognized, so something was off already.

frankmorgner:
> 

It is possible that in 2019 postecert issued smart cards that are different from what OpenSC knew as postecert profile...

PR #1600, opened 2019-02-01
Trustonic pkcs11

AlexandreGonzalo:
Hi,
In this pull request I am mainly adding the support for CKM_RSA_PKCS_OAEP for the "pkcs11-tool --test" command.
I am also fixing the "pkcs11-tool --verify" when the file to verify is larger than 1025 bytes.
Finally, I fixed a minor build issue because RIPEMD and CAST are not defined by my version of OpenSsl.
I have only tested with my own implementation of the PKCS#11 library.
Notice that this is my first contribution so let me know if I am not doing the things correctly.
Best Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5258,21 +5258,10 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 	int             failed;
 	CK_RV           rv;
 	int             pad;
-	CK_MECHANISM_TYPE hash_alg = CKM_SHA_1;
+	CK_MECHANISM_TYPE hash_alg;
 	CK_RSA_PKCS_OAEP_PARAMS oaep_params;
 
 	printf("    %s: ", p11_mechanism_to_name(mech_type));
-
-	if ((mech_type == CKM_RSA_PKCS) || (mech_type == CKM_RSA_PKCS_OAEP)) {
-		if (opt_hash_alg == 0) {
-			hash_alg = CKM_SHA_1;
-		} else if (opt_hash_alg != CKM_SHA_1) {
-			printf("Only CKM_SHA_1 supported\n");
-			return 0;
-		} else {
-			hash_alg = opt_hash_alg;
-		}
-	}
 
 	pkey = get_public_key(session, privKeyObject);
 	if (pkey == NULL)
@@ -5292,6 +5281,14 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		in_len = mod_len-11;
 		break;
 	case CKM_RSA_PKCS_OAEP: {
+		if (opt_hash_alg == 0) {
+			hash_alg = CKM_SHA_1;
+		} else if (opt_hash_alg != CKM_SHA_1) {
+			printf("Only CKM_RSA_PKCS_OAEP with CKM_SHA_1 supported\n");
+			return 0;
+		} else {
+			hash_alg = opt_hash_alg;
+		}
 		pad = RSA_PKCS1_OAEP_PADDING;
 		/* Limit the input length to <= mod_len-2-2*hlen */
 		size_t len = 2+2*hash_length(hash_alg);
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5334,18 +5334,6 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 			oaep_params.mgf = CKG_MGF1_SHA1;
 			break;
 		}
-		break;
-	case CKM_RSA_PKCS:
-		mech.pParameter = NULL;
-		mech.ulParameterLen = 0;
-		break;
-	default:
-		util_fatal("Mechanism %s illegal or not supported\n", p11_mechanism_to_name(mech_type));
-	}
-
-
-	/* If an RSA-OAEP mechanism, it needs parameters */
-	if (oaep_params.hashAlg) {
 		if (opt_mgf != 0)
 			oaep_params.mgf = opt_mgf;
 
@@ -5354,6 +5342,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		oaep_params.pSourceData = NULL; /* PKCS#11 standard: this must be NULLPTR */
 		oaep_params.ulSourceDataLen = 0; /* PKCS#11 standard: this must be 0 */
 
+		/* If an RSA-OAEP mechanism, it needs parameters */
 		mech.pParameter = &oaep_params;
 		mech.ulParameterLen = sizeof(oaep_params);
 
@@ -5363,9 +5352,14 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 			oaep_params.source,
 			oaep_params.pSourceData,
 			oaep_params.ulSourceDataLen);
-
+		break;
+	case CKM_RSA_PKCS:
+		mech.pParameter = NULL;
+		mech.ulParameterLen = 0;
+		break;
+	default:
+		util_fatal("Mechanism %s illegal or not supported\n", p11_mechanism_to_name(mech_type));
 	}
-
 
 	mech.mechanism = mech_type;
 	rv = p11->C_DecryptInit(session, &mech, privKeyObject);
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5340,7 +5340,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		mech.ulParameterLen = 0;
 		break;
 	default:
-		util_fatal("Mechanism %s illegal or not supported\n", p11_mechanism_to_name(opt_mechanism));
+		util_fatal("Mechanism %s illegal or not supported\n", p11_mechanism_to_name(mech_type));
 	}
 
 
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -1985,7 +1985,7 @@ static void verify_signature(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 			r = read(fd, in_buffer, sizeof(in_buffer));
 		} while (r > 0);
 
-		sig_len = sizeof(sig_buffer);
+		sig_len = r2;
 		rv = p11->C_VerifyFinal(session, sig_buffer, sig_len);
 		if (rv != CKR_OK)
 			p11_fatal("C_VerifyFinal", rv);
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5361,7 +5361,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		return 0;
 	}
 
-	if (in_len >= sizeof(orig_data)) {
+	if (in_len > sizeof(orig_data)) {
 		printf("Private key size is too long\n");
 		return 0;
 	}
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5371,12 +5371,19 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 			md = EVP_sha512();
 			break;
 		}
+#if defined(EVP_PKEY_CTX_set_rsa_oaep_md)
 		if (EVP_PKEY_CTX_set_rsa_oaep_md(ctx, md) <= 0) {
 			EVP_PKEY_CTX_free(ctx);
 			EVP_PKEY_free(pkey);
 			printf("set md failed, returning\n");
 			return 0;
 		}
+#else
+        if (hash_alg != CKM_SHA_1) {
+            printf("This version of OpenSsl only supports SHA1 for OAEP, returning\n");
+			return 0;
+        }
+#endif
 		switch (mgf) {
 		case CKG_MGF1_SHA1:
 			md = EVP_sha1();
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5461,6 +5461,17 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 		printf("No OpenSSL support, unable to validate decryption\n");
 #else
 		for (n = 0; n < num_mechs; n++) {
+			switch (mechs[n]) {
+			case CKM_RSA_PKCS:
+			case CKM_RSA_PKCS_OAEP:
+			case CKM_RSA_X_509:
+			//case CKM_RSA_PKCS_TPM_1_1:
+			//case CKM_RSA_PKCS_OAEP_TPM_1_1:
+				break;
+			default:
+				printf(" -- mechanism can't be used to decrypt, skipping\n");
+				continue;
+			}
 
 			errors += encrypt_decrypt(sess, mechs[n], privKeyObject);
 		}
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5287,7 +5287,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		CK_OBJECT_HANDLE privKeyObject)
 {
 	EVP_PKEY       *pkey;
-	unsigned char	orig_data[512] = {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', '\0'};
+	unsigned char	orig_data[512];
 	unsigned char	encrypted[512], data[512];
 	CK_MECHANISM	mech;
 	CK_ULONG	encrypted_len, data_len;
@@ -5299,6 +5299,8 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 	CK_RSA_PKCS_OAEP_PARAMS oaep_params;
 
 	printf("    %s: ", p11_mechanism_to_name(mech_type));
+
+	pseudo_randomize(orig_data, sizeof(orig_data));
 
 	pkey = get_public_key(session, privKeyObject);
 	if (pkey == NULL)
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5258,7 +5258,8 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 	int             failed;
 	CK_RV           rv;
 	int             pad;
-	CK_MECHANISM_TYPE hash_alg;
+	CK_MECHANISM_TYPE hash_alg = CKM_SHA256;
+	CK_RSA_PKCS_MGF_TYPE mgf;
 	CK_RSA_PKCS_OAEP_PARAMS oaep_params;
 
 	printf("    %s: ", p11_mechanism_to_name(mech_type));
@@ -5281,14 +5282,35 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		in_len = mod_len-11;
 		break;
 	case CKM_RSA_PKCS_OAEP: {
-		if (opt_hash_alg == 0) {
-			hash_alg = CKM_SHA_1;
-		} else if (opt_hash_alg != CKM_SHA_1) {
-			printf("Only CKM_RSA_PKCS_OAEP with CKM_SHA_1 supported\n");
-			return 0;
-		} else {
+		if (opt_hash_alg != 0) {
 			hash_alg = opt_hash_alg;
 		}
+		switch (hash_alg) {
+		case CKM_SHA_1:
+			mgf = CKG_MGF1_SHA1;
+			break;
+		case CKM_SHA224:
+			mgf = CKG_MGF1_SHA224;
+			break;
+		default:
+			printf("hash-algorithm %s unknown, defaulting to CKM_SHA256\n", p11_mechanism_to_name(hash_alg));
+			/* fall through */
+		case CKM_SHA256:
+			mgf = CKG_MGF1_SHA256;
+			break;
+		case CKM_SHA384:
+			mgf = CKG_MGF1_SHA384;
+			break;
+		case CKM_SHA512:
+			mgf = CKG_MGF1_SHA512;
+			break;
+		}
+		if (opt_mgf != 0) {
+			mgf = opt_mgf;
+		} else {
+			printf("mgf not set, defaulting to %s\n", p11_mgf_to_name(mgf));
+		}
+
 		pad = RSA_PKCS1_OAEP_PADDING;
 		/* Limit the input length to <= mod_len-2-2*hlen */
 		size_t len = 2+2*hash_length(hash_alg);
@@ -5305,7 +5327,7 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		in_len = mod_len;
 		break;
 	default:
-		printf("Unsupported mechanism, returning\n");
+		printf("Unsupported mechanism %s, returning\n", p11_mechanism_to_name(mech_type));
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -4650,7 +4652,9 @@ static int test_signature(CK_SESSION_HANDLE sess)
 		CKM_RSA_PKCS,
 		CKM_SHA1_RSA_PKCS,
 		CKM_MD5_RSA_PKCS,
+		#ifndef OPENSSL_NO_RIPEMD
```
Jakuje: Please, do not indent the `#ifdef` and `#endif` macros.

src/tools/pkcs11-tool.c
```diff
@@ -5231,7 +5235,9 @@ static int test_unwrap(CK_SESSION_HANDLE sess)
 		errors += wrap_unwrap(sess, EVP_des_cbc(), privKeyObject);
 		errors += wrap_unwrap(sess, EVP_des_ede3_cbc(), privKeyObject);
 		errors += wrap_unwrap(sess, EVP_bf_cbc(), privKeyObject);
+		#ifndef OPENSSL_NO_CAST
```
Jakuje: Please, do not indent the `#ifdef` and `#endif` macros.

src/tools/pkcs11-tool.c
```diff
@@ -5262,11 +5269,96 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(pkey);
+			printf("Encryption failed, returning\n");
+			return 0;
+		}
+		EVP_PKEY_CTX_free(ctx);
+		EVP_PKEY_free(pkey);
+		encrypted_len = outlen;
+
+	} else {
```
Jakuje: I think this should also explicitly check for `CKM_RSA_PKCS` as the decryption does so we can fail earlier if the user provided unsupported mechanism.

src/tools/pkcs11-tool.c
```diff
@@ -5363,6 +5455,18 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 		printf("No OpenSSL support, unable to validate decryption\n");
 #else
 		for (n = 0; n < num_mechs; n++) {
+			switch (mechs[n]) {
+			case CKM_RSA_PKCS:
+			case CKM_RSA_PKCS_OAEP:
+			case CKM_RSA_X_509:
```
Jakuje: `CKM_RSA_X_509` can be used for encryption/decryption, but you do not handle it above. Does it work?

src/tools/pkcs11-tool.c
```diff
@@ -5363,6 +5455,18 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
...
+			//case CKM_RSA_PKCS_TPM_1_1:
+			//case CKM_RSA_PKCS_OAEP_TPM_1_1:
```
Jakuje: Do we need these comments here? Do you have a pkcs11 that presents them? Does it work?

AlexandreGonzalo: These two mechanisms are defined in the "official" header files that I am using and described in the 2.40 specifications.
They don't exist in the header files of your project, this is why I have commented them.
My pkcs11 won't support them, so I won't be able to test.
I can remove these two lines.

src/tools/pkcs11-tool.c
AlexandreGonzalo: You are right, we need to check explicitly for CKM_RSA_PKCS and properly handle the case of CKM_RSA_X_509.
I will try to fix this issue during the week.

src/tools/pkcs11-tool.c
AlexandreGonzalo: Unfortunately, it is not working.
I will try to fix this issue during the week.

src/tools/pkcs11-tool.c
AlexandreGonzalo: 2 lines removed @https://github.com/OpenSC/OpenSC/pull/1600/commits/d25fbe3cecc55857541ad773899235e227409c47

AlexandreGonzalo:
I tested my new changes with 2 keys:
pkcs11-tool --list-objects
Public Key Object; RSA 1024 bits
  label:      NotSupported
  ID:         1024
  Usage:      encrypt, verify
Public Key Object; RSA 4096 bits
  label:      NotSupported
  ID:         4096
  Usage:      encrypt, verify
Private Key Object; RSA 
  label:      NotSupported
  ID:         1024
  Usage:      decrypt, sign
Private Key Object; RSA 
  label:      NotSupported
  ID:         4096
  Usage:      decrypt, sign
pkcs11-tool --test
  seems to be OK
Digests:
  SHA-1: OK
  testing key 0 (NotSupported) 
  testing signature mechanisms:
    SHA1-RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
  testing key 1 (4096 bits, label=NotSupported) with 1 signature mechanism
    SHA1-RSA-PKCS: OK
  testing key 0 (NotSupported)
    SHA1-RSA-PKCS: OK
  testing key 1 (NotSupported) with 1 mechanism
    SHA1-RSA-PKCS: OK
Unwrap: not implemented
  testing key 0 (NotSupported) 
    RSA-X-509: OK
    RSA-PKCS: OK
OAEP parameters: hashAlg=SHA-1, mgf=MGF1-SHA1, source_type=0, source_ptr=0x0, source_len=0
    RSA-PKCS-OAEP: OK
 -- mechanism can't be used to decrypt, skipping
 -- mechanism can't be used to decrypt, skipping
  testing key 1 (NotSupported) 
    RSA-X-509: OK
    RSA-PKCS: OK
OAEP parameters: hashAlg=SHA-1, mgf=MGF1-SHA1, source_type=0, source_ptr=0x0, source_len=0
    RSA-PKCS-OAEP: OK
 -- mechanism can't be used to decrypt, skipping
 -- mechanism can't be used to decrypt, skipping
No errors

AlexandreGonzalo:
With OPENSSL_VERSION=1.0.1j, the digest of OAEP is hardcoded to SHA-1.
So I can't test for other digests.

mouse07410:
OpenSSL 1.0.1 is outdated and should not be used in new designs. SHA1 has been deprecated and must not be used, let alone hardcoded.

Please change to OpenSSL 1.0.2 (at least) and SHA2.

AlexandreGonzalo:
Hi,
I totally agree that SHA1 is broken and should not be used anymore.
I am only using this version of OpenSsl for testing purpose and I have to update it to a newer version.
I will be able to finalize this pull request only when I will be able to test with this newer OpenSsl.
Thanks,
Alex.

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5272,125 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+		pad = RSA_PKCS1_PADDING;
+		/* Limit the input length to <= mod_len-11 */
+		in_len = mod_len-11;
+		break;
+	case CKM_RSA_PKCS_OAEP: {
+		if (opt_hash_alg == 0) {
+			hash_alg = CKM_SHA_1;
+		} else if (opt_hash_alg != CKM_SHA_1) {
+			printf("Only CKM_RSA_PKCS_OAEP with CKM_SHA_1 supported\n");
+			return 0;
```
Jakuje: If I see right, here you reject any other `opt_hash_alg` than `CKM_SHA1` (default) and on the line 5346 you again check for all the other possible hashes that can not happen. I think you should allow other hashes to be used too here for the sake of universality (might need some more switches for the openssl).

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5272,125 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+		/* If an RSA-OAEP mechanism, it needs parameters */
+		mech.pParameter = &oaep_params;
+		mech.ulParameterLen = sizeof(oaep_params);
+
+		fprintf(stderr, "OAEP parameters: hashAlg=%s, mgf=%s, source_type=%lu, source_ptr=%p, source_len=%lu\n",
+			p11_mechanism_to_name(oaep_params.hashAlg),
+			p11_mgf_to_name(oaep_params.mgf),
+			oaep_params.source,
+			oaep_params.pSourceData,
+			oaep_params.ulSourceDataLen);
```
Jakuje: Most of this is copied from the `decrypt_data()` function. Would it make sense to avoid code duplication and create a separate function that would create the `oaep_params`?

src/tools/pkcs11-tool.c
```diff
@@ -5363,6 +5482,16 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 		printf("No OpenSSL support, unable to validate decryption\n");
 #else
 		for (n = 0; n < num_mechs; n++) {
+			switch (mechs[n]) {
+			case CKM_RSA_PKCS:
+			case CKM_RSA_PKCS_OAEP:
+			case CKM_RSA_X_509:
+				break;
+			default:
+				printf(" -- mechanism can't be used to decrypt, skipping\n");
```
Jakuje: This output is not helpful since it does not show what mechanism can not be used to decrypt. You should print  at least the mechanism name before this message, ideally using `p11_mechanism_to_name().`

src/tools/pkcs11-tool.c
AlexandreGonzalo: I have changed this code now I am able to test against OpenSsl 1.1.a.

src/tools/pkcs11-tool.c
AlexandreGonzalo: I improved the output as suggested in my new changes.

AlexandreGonzalo:
I have noticed something weird with OpenSsl 1.1.1a.
I have a Segmentation fault when the pkcs11-tool is exiting.
I don't have this crash when I remove the OPENSSL_INIT_LOAD_CRYPTO_STRINGS flag to the call to OPENSSL_init_crypto().
I am wondering why this function is called since it is supposed to be done automatically with OpenSSL >= 1.1.0.
https://wiki.openssl.org/index.php/Library_Initialization#libcrypto_Initialization
_"If you are using OpenSSL 1.1.0 or above, then the library will initialize itself automatically"_

frankmorgner:
https://travis-ci.org/OpenSC/OpenSC/jobs/489960037:
```
pkcs11-tool.c: In function ‘encrypt_decrypt’:
pkcs11-tool.c:5384:3: error: implicit declaration of function ‘EVP_PKEY_CTX_set_rsa_oaep_md’ [-Werror=implicit-function-declaration]
   if (EVP_PKEY_CTX_set_rsa_oaep_md(ctx, md) <= 0) {
   ^
```

Can you debug the crash with gdb, valgrind or VS?

AlexandreGonzalo:
Hi Franck,
Thank you for pointing out this build issue.
I am a beginner in this project and I am a making many mistakes...
I have submitted a new patch to fix this build issue.
Unfortunately, this week I won't have much time to try to debug this crash.
It is most probably linked to my environment.
I am trying to use the P11 engine and my PKCS#11 library which is still under development.
It looks like a memory corruption which is "detected" when the buffers are freed during the process exit.
I will get back to you when I will have more information.
Regards,
Alexandre.

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5273,187 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+			md = EVP_sha256();
+			break;
+		case CKM_SHA384:
+			md = EVP_sha384();
+			break;
+		case CKM_SHA512:
+			md = EVP_sha512();
+			break;
+		}
+#if defined(EVP_PKEY_CTX_set_rsa_oaep_md)
```
Jakuje: I do not think this will work. You need to create a configure check to be able to test for a function this way. This will never get executed, even with the newer OpenSSL versions, if I am right.

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5273,187 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(pkey);
+			printf("set md failed, returning\n");
+			return 0;
+		}
+#else
+        if (hash_alg != CKM_SHA_1) {
+            printf("This version of OpenSsl only supports SHA1 for OAEP, returning\n");
+			return 0;
+        }
```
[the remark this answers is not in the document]
AlexandreGonzalo: Sorry, this is my editor which is defaulting to spaces instead of tabs as by default we don't use tabs in our software.
Do you have a kind of checkpatch or astyle script that I can run before submitting patches?
This is fixed by https://github.com/OpenSC/OpenSC/pull/1600/commits/9ae507c5f8dcc193f367e6ad536157cb679f3408

src/tools/pkcs11-tool.c
AlexandreGonzalo: Please have a look at https://github.com/openssl/openssl/blob/master/include/openssl/rsa.h
This is not a function but a macro for the EVP_PKEY_CTX_ctrl().
If you prefer I can  use #if OPENSSL_VERSION_NUMBER >= 0x... (I have to check in which version it was introduced).

src/tools/pkcs11-tool.c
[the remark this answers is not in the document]
Jakuje: Unfortunately not. If you have some recommendation for service or even just a script, contributions are welcomed.

src/tools/pkcs11-tool.c
Jakuje: Oh. You are right. Then it is probably fine. Thanks for clarification.

src/tools/pkcs11-tool.c
```diff
@@ -1946,6 +1946,42 @@ static void verify_signature(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+			 * So a "256" bit key has x and y of 32 bytes each
+			 * something like: "04 41 04||x||y"
+			 * Do simple size calculation based on DER encoding
+			 */
+			if ((len - 2) <= 127)
+				rs_len = len - 3;
+			else if ((len - 3) <= 255)
+				rs_len = len - 4;
+			else
+				util_fatal("Key not supported");
```
frankmorgner: isn't this length check performed by `sc_asn1_sig_value_sequence_to_rs`? I think that's the natural place to put it.

src/tools/pkcs11-tool.c
```diff
@@ -5244,12 +5287,16 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		CK_OBJECT_HANDLE privKeyObject)
 {
 	EVP_PKEY       *pkey;
-	unsigned char	orig_data[] = {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', '\0'};
+	unsigned char	orig_data[512] = {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', '\0'};
```
frankmorgner: I think using the rest of the 512 bytes will yield a warning of uninitialized data. You can use `pseudo_randomize` to initialize it properly for the test.

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5309,187 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
 	}
-	encrypted_len = EVP_PKEY_encrypt_old(encrypted, orig_data, sizeof(orig_data), pkey);
-	EVP_PKEY_free(pkey);
-	if (((int) encrypted_len) <= 0) {
+	size_t in_len;
+	CK_ULONG mod_len = (get_private_key_length(session, privKeyObject) + 7) / 8;
+	switch (mech_type) {
+	case CKM_RSA_PKCS:
+		pad = RSA_PKCS1_PADDING;
+		/* Limit the input length to <= mod_len-11 */
```
frankmorgner: why?

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5309,187 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+			break;
+		}
+		if (opt_mgf != 0) {
+			mgf = opt_mgf;
+		} else {
+			printf("mgf not set, defaulting to %s\n", p11_mgf_to_name(mgf));
+		}
+
+		pad = RSA_PKCS1_OAEP_PADDING;
+		/* Limit the input length to <= mod_len-2-2*hlen */
```
frankmorgner: why?

src/tools/pkcs11-tool.c
```diff
@@ -5262,12 +5309,187 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
+#if defined(EVP_PKEY_CTX_set_rsa_oaep_md)
+		if (EVP_PKEY_CTX_set_rsa_oaep_md(ctx, md) <= 0) {
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(pkey);
+			printf("set md failed, returning\n");
+			return 0;
+		}
+#else
+		if (hash_alg != CKM_SHA_1) {
+			printf("This version of OpenSsl only supports SHA1 for OAEP, returning\n");
```
frankmorgner: typo

src/tools/pkcs11-tool.c
frankmorgner: does it work now?

src/tools/pkcs11-tool.c
frankmorgner: I think that's not pushed to the PR, is it?

src/tools/pkcs11-tool.c
AlexandreGonzalo: This limitation is coming from pkcs11-curr-v2.40.pdf
Have a look a the table p25 (2.1.6 PKCS #1 v1.5 RSA).
The Output length for C_Decrypt() must be <= k-11
(k is the length in bytes of the RSA modulus)

src/tools/pkcs11-tool.c
AlexandreGonzalo: This limitation is coming from pkcs11-curr-v2.40.pdf
Have a look a the table p27 (2.1.8 PKCS #1 RSA OAEP).
The Output length for C_Decrypt() must be <= k-2-2hLen
(k is the length in bytes of the RSA modulus, hLen is the output length of the message digest algorithm)

src/tools/pkcs11-tool.c
AlexandreGonzalo: Fixed by https://github.com/OpenSC/OpenSC/pull/1600/commits/b327b76134b19e2495e01abce47ca1aef2d32034

AlexandreGonzalo:
> Jakub asked about unifying `encrypt_decrypt()` and `decrypt()`, what do you think about that?

It is true that I originally copied code from decrypt_data() but since I have done many changes and I don't think that it is worth to try to factorize it otherwise it will be less readable.

src/tools/pkcs11-tool.c
AlexandreGonzalo: Sorry, what change do you suggest?

src/tools/pkcs11-tool.c
AlexandreGonzalo: It seems that all the callers (card-isoApplet.c and card-myeid.c) to sc_asn1_sig_value_sequence_to_rs() have the same difficulty to determine the length (buflen) to give the function.
card-isoApplet.c is doing some calculation based on ec_field_length, card-myeid.c is decoding the ASN1...
It would be simpler if we do a first call to the function with a NULL buf, allocate the buffer and then call it a second time if the correct length.

src/tools/pkcs11-tool.c
frankmorgner: OpenSSL instead of OpenSsl

AlexandreGonzalo: Fixed @ 728d099

AlexandreGonzalo:
> Hi Frank,
> Thank you for pointing out this build issue.
> I am a beginner in this project and I am a making many mistakes...
> I have submitted a new patch to fix this build issue.
> Unfortunately, this week I won't have much time to try to debug this crash.
> It is most probably linked to my environment.
> I am trying to use the P11 engine and my PKCS#11 library which is still under development.
> It looks like a memory corruption which is "detected" when the buffers are freed during the process exit.
> I will get back to you when I will have more information.
> Regards,
> Alexandre.

I tried to figure out why I have a seg fault with the OpenSSL 1.1.1a.
In fact, it seems that the issue is in the libp11 engine.
It is crashing when ERR_unload_strings() is called in eng_err.c/ERR_unload_ENG_strings().

01-21 23:14:33.801 23399 23399 F DEBUG   : ABI: 'arm'
01-21 23:14:33.801 23399 23399 F DEBUG   : pid: 23396, tid: 23396, name: pkcs11-tool  >>> pkcs11-tool <<<
01-21 23:14:33.801 23399 23399 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
01-21 23:14:33.801 23399 23399 F DEBUG   : Cause: null pointer dereference
01-21 23:14:33.802 23399 23399 F DEBUG   :     r0  00000000  r1  00000001  r2  00000002  r3  f616cf3f
01-21 23:14:33.802 23399 23399 F DEBUG   :     r4  f5e144c8  r5  f5de8ad0  r6  f616ea8d  r7  ffd954a0
01-21 23:14:33.802 23399 23399 F DEBUG   :     r8  f5e14564  r9  f616ea8d  r10 f61f0034  r11 ffd95478
01-21 23:14:33.802 23399 23399 F DEBUG   :     ip  f5de7360  sp  ffd95478  lr  f5d56554  pc  f6174966
01-21 23:14:33.803 23399 23399 F DEBUG   : 
01-21 23:14:33.803 23399 23399 F DEBUG   : backtrace:
01-21 23:14:33.803 23399 23399 F DEBUG   :     #00 pc 00064966  /system/lib/libc.so (pthread_rwlock_wrlock)
01-21 23:14:33.803 23399 23399 F DEBUG   :     #01 pc 00185550  /system/lib/libcrypto.so.1.1 (CRYPTO_THREAD_write_lock+8)

frankmorgner:
@mtrojnar, any ideas regarding the problem with OpenSSL 1.1.1?

mtrojnar:
@frankmorgner It is hard to draw any conclusions without a complete stack backtrace.

AlexandreGonzalo:
> @frankmorgner It is hard to draw any conclusions without a complete stack backtrace.
I am seeing that err_cleanup() is called before engine_destroy().

frankmorgner:
both, pkcs11-tool and libp11 are calling `OPENSSL_init_crypto`, whereas the [man page](https://github.com/openssl/openssl/blob/master/doc/man3/OPENSSL_init_crypto.pod) indicates that there are some circumstances where it must not be used... However, I it's not clear to me what causes the problem in this particular case. I think using automatic initialization for 1.1.0 and later is the way to go if it solves the problem.

src/tools/pkcs11-tool.c
frankmorgner: thanks for the clearification.

The check for `(in_len <= sizeof orig_data)` is still missing, isn't it?

src/tools/pkcs11-tool.c
frankmorgner: ping?

AlexandreGonzalo:
Hi Frank,
I added the missing check in my latest commit.
Do you need more fixes or clarifications?
Regards,
Alexandre.

Jakuje:
When I try to build your branch I am getting the following failures:
```
pkcs11-tool.c: In function ‘encrypt_decrypt’:
pkcs11-tool.c:6524:6: error: ‘mgf’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   if (mi->mech == mgf)
      ^
pkcs11-tool.c:5290:23: note: ‘mgf’ was declared here
  CK_RSA_PKCS_MGF_TYPE mgf;
                       ^~~
```

Jakuje:
When I fixed the above and tried to run the OEAP tests with pkcs11-tool, I am still getting failures:
```
Using slot 0 with a present token (0x7ab4c146)
OAEP parameters: hashAlg=SHA256, mgf=MGF1-SHA256, source_type=0, source_ptr=(nil), source_len=0
error: PKCS11 function C_DecryptInit failed: rv = CKR_ARGUMENTS_BAD (0x7)
Aborting.
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
Signatures: not implemented
  testing key 0 (ECC_auth) -- non-RSA, skipping
  testing key 1 (RSA_auth) with 1 mechanism
    RSA-X-509: OK
  testing key 2 (ECC521) with 1 mechanism -- non-RSA, skipping
  testing key 3 (RSA2048) with 1 mechanism
    RSA-X-509: OK
Unwrap: not implemented
  testing key 0 (ECC_auth)  -- non-RSA, skipping
  testing key 1 (RSA_auth)
    RSA-PKCS: OK
    RSA-X-509: OK
    RSA-PKCS-OAEP: mgf not set, defaulting to MGF1-SHA256
```
(from `tests/test-pkcs11-tool-test.sh` with SoftHSM)

AlexandreGonzalo:
> When I try to build your branch I am getting the following failures:
> 
> ```
> pkcs11-tool.c: In function ‘encrypt_decrypt’:
> pkcs11-tool.c:6524:6: error: ‘mgf’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
>    if (mi->mech == mgf)
>       ^
> pkcs11-tool.c:5290:23: note: ‘mgf’ was declared here
>   CK_RSA_PKCS_MGF_TYPE mgf;
>                        ^~~
> cc1: all warnings being treated as errors
> ```

I am trying to understand why mgf is not initialized.
for the case CKM_RSA_PKCS_OAEP, it is set for all the cases.
for default case, we have a fall through which sets mgf = CKG_MGF1_SHA256;

AlexandreGonzalo:
> When I fixed the above and tried to run the OEAP tests with pkcs11-tool, I am still getting failures:
> 
> ```
> Using slot 0 with a present token (0x7ab4c146)
> OAEP parameters: hashAlg=SHA256, mgf=MGF1-SHA256, source_type=0, source_ptr=(nil), source_len=0
> error: PKCS11 function C_DecryptInit failed: rv = CKR_ARGUMENTS_BAD (0x7)
> Aborting.
> C_SeedRandom() and C_GenerateRandom():
>   seems to be OK
> Digests:
>   all 4 digest functions seem to work
>   MD5: OK
>   SHA-1: OK
> Signatures: not implemented
> Verify (currently only for RSA)
>   testing key 0 (ECC_auth) -- non-RSA, skipping
>   testing key 1 (RSA_auth) with 1 mechanism
>     RSA-X-509: OK
>   testing key 2 (ECC521) with 1 mechanism -- non-RSA, skipping
>   testing key 3 (RSA2048) with 1 mechanism
>     RSA-X-509: OK
> Unwrap: not implemented
> Decryption (currently only for RSA)
>   testing key 0 (ECC_auth)  -- non-RSA, skipping
>   testing key 1 (RSA_auth)
>     RSA-PKCS: OK
>     RSA-X-509: OK
>     RSA-PKCS-OAEP: mgf not set, defaulting to MGF1-SHA256
> ```
> (from `tests/test-pkcs11-tool-test.sh` with SoftHSM)

Maybe your PKCS11 implementation does not support this mechanism?

Jakuje:
OK, it looks like the SoftHSM supports only SHA1 hash and MGF1, which is the reason for the failure here. But the build issue should be addressed. From what I see, it really looks like false positive of my gcc, but assigning the default value during initialization of a variable is a good practice.

AlexandreGonzalo:
I initialized the variable as suggested in my latest commit.

Jakuje:
Thanks. I think we are good.

frankmorgner:
thanks everyone

AlexandreGonzalo:
Thanks for your help Jakub and Frank!

PR #1605, opened 2019-02-10
Fix build on cygwin in strict mode

lksj:
In strict mode (all warnings is errors) there is two cases build broken
First case:
comparison between signed and unsigned integer expressions error
Second case:
format '%lu' expects argument of type 'long unsigned int', but argument has type 'size_t'

In both cases add explicit type conversion

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2459,7 +2459,7 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
 	}
 	else
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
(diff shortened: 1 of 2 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -2459,7 +2459,7 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
 	}
 	else
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
-	sc_log(card->ctx, "pk_packet_len is %lu", pk_packet_len);
+	sc_log(card->ctx, "pk_packet_len is %lu", (unsigned long)pk_packet_len);
```
frankmorgner: This should use `"%"SC_FORMAT_LEN_SIZE_T"u` instead.

PR #1607, opened 2019-02-11
Add installer option to deselect tokend

metsma:

##### Checklist

- [X] macOS tokend is tested

### DIFF
--- a/MacOSX/Distribution.xml.in
+++ b/MacOSX/Distribution.xml.in
@@ -1,4 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
+<!--
+https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/DistributionDefinitionRef/
+-->
 <installer-gui-script minSpecVersion="2">
     <allowed-os-versions>
       <os-version min="10.10"/>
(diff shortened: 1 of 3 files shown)


frankmorgner:
What about an additional option to install OpenSCToken?

metsma:
It is currently not build by default?

frankmorgner:
It's currently not part of the installer, instead it's available as separate download. Do you have updates on the compatibility regarding CTK with apps that don't come from Apple?

metsma:
I don't see that there is any issue.
We distribute it with Open-EID installer https://github.com/open-eid/esteid-ctk-tokend and
https://github.com/open-eid/osx-installer/blob/master/distribution.xml#L152

martinpaljak:
LGTM, @frankmorgner ?

frankmorgner:
I wanted to see the change myself and maybe add the option for installing OpenSCToken on the way. Unfortunately, I did not yet have time for that.

metsma:
Lets merge this then And create another PR for OpenSCToken?
I wan't test sometimes nightly build, but really hate to install this old tokend

mouse07410:
> I don't see that there is any issue

I do. After this PR was merged, tokend does not get started any more when the card is inserted. Yes, I installed both `OpenSC.pkg` and `OpenSC-tokend.pkg`.

Nothing in system logs to give a hint on what happened.

metsma:
I don't see how this change can affect tokend functionality.
It will only split packages and add option to install separately.
We don't use Adobe and MS Office for signing.
Safari (macos > 10.13 with ECDSA token) and Chrome (macos > 10.12) can handle the CTK tokend.
Our plugins and signing application uses PKCS11 for signing.
esteid-ctk-tokned is separate macos extension to handle EstEID cards similar to OpenSCToken.
I don't think CTK tokend exposes CDSA interface.

mouse07410:
> I don't see how this change can affect tokend functionality. It will only split packages and add option to install separately.

In theory I don't either. In practice - OpenSC tokend was not invoked any more with this PR, and resurrected again when I reverted it in my fork. Go figure...

@metsma thanks for explaining your use case. Could you please confirm that Apple `pivtoken` does **not** work with EstEID cards (and that's your main reason for supplying your own `esteid-ctk-tokend`)?

metsma:
yes, EstEID card is not PIV compatible thats why we have own extension

PR #1615, opened 2019-02-28
Handle reader going missing

CendioOssman:
Make sure we report properly to the upper layers if PC/SC is reporting that the reader has gone missing. Otherwise applications won't be properly notified that something has happened.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -340,8 +340,8 @@ static int refresh_attributes(sc_reader_t *reader)
 			LOG_FUNC_RETURN(reader->ctx, SC_SUCCESS);
 		}
 		
-		/* the system could not dectect any reader. It means, the prevoiusly attached reader is disconnected. */		
-		if (rv == (LONG)SCARD_E_NO_READERS_AVAILABLE || rv == (LONG)SCARD_E_SERVICE_STOPPED) {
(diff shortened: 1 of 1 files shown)


CendioOssman:
Argh. It seems I used an out of date `master` for this. Some work has been done to handle this case, but not this specific error code. I'll update the patch.

src/libopensc/reader-pcsc.c
```diff
@@ -340,8 +340,8 @@ static int refresh_attributes(sc_reader_t *reader)
 			LOG_FUNC_RETURN(reader->ctx, SC_SUCCESS);
 		}
 		
-		/* the system could not dectect any reader. It means, the prevoiusly attached reader is disconnected. */		
-		if (rv == (LONG)SCARD_E_NO_READERS_AVAILABLE || rv == (LONG)SCARD_E_SERVICE_STOPPED) {
+		/* the system could not dectect the reader. It means, the prevoiusly attached reader is disconnected. */
```
plaes: Can you fix the "detect" typo too?

CendioOssman: Sure thing. :)

CendioOssman:
We found some more issues on this theme when we did further testing, so I'd like to suggest these two commits as well.

martinpaljak:
Thanks, these look like useful fixes!

PR #1616, opened 2019-03-01
Small memory leak fix (Fixes#1586 / CVE-2019-6502)

sgayou:
CVE-2019-6502 was assigned to what appears to be a very minor
memory leak that only occurs on an error-case in a CLI tool.
If util_connect_card fails, we still need to release the sc
context previously allocated by sc_context_create else memory
will leak.

### DIFF
--- a/src/tools/eidenv.c
+++ b/src/tools/eidenv.c
@@ -403,6 +403,7 @@ int main(int argc, char **argv)
 	r = util_connect_card(ctx, &card, opt_reader, opt_wait, 0);
 	if (r) {
 		fprintf(stderr, "Failed to connect to card: %s\n", sc_strerror(r));
+		sc_release_context(ctx);
 		return 1;
 	}
 


frankmorgner:
please check the other tools as well. I suspect that some will contain the same problem.

frankmorgner:
BTW, do you still call it a memory leak if it happens exactly when the program terminates (and the OS then frees all resources afterwards)?

sgayou:
> 
> 
> BTW, do you still call it a memory leak if it happens exactly when the program terminates (and the OS then frees all resources afterwards)?

Great question. I think this answers it the best: http://c-faq.com/malloc/freeb4exit.html

Presumably the only outcome of this fix is "good practice" and no one will (hopefully) report this specific Valgrind/ASAN error in the future. As stated in the ticket though, I don't think the CVE is necessarily valid for this issue as I don't see any way this could be exploited by an attacker to have any form of negative impact on availability, confidentiality, or integrity.

PR #1617, opened 2019-03-03
Restore "<token label> (<PIN label>)" format

muktupavels:
Commit 01287e393ebeed1db7ef2f45ee133b2cd3f069ae fixed some warnings,
but at same time introduced regression.

sizeof(p15card->tokeninfo->label) does not return label length causing
token label truncation. Use strlen to get real label size.

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1052,13 +1052,128 @@ pkcs15_add_object(struct sc_pkcs11_slot *slot, struct pkcs15_any_object *obj,
 }
 
 
+static char *
+trim_whitespace(const char *str)
+{
+	const char *start;
+	const char *end;
+	size_t new_len;
+	char *retval;
+
+	if (str == NULL || *str == '\0')
+		return NULL;
+
+	start = str;
+	while(isspace((unsigned char) *start))
+		start++;
+
+	if (*start == '\0')
+		return NULL;
+
+	end = start + strlen(start) - 1;
+	while(end > start && isspace((unsigned char)*end))
+		end--;
+
+	new_len = end - start + 1;
+
+	retval = malloc(new_len + 1);
+	if (retval == NULL)
+		return NULL;
+
+	memcpy(retval, start, new_len);
+	retval[new_len] = '\0';
+
+	return retval;
+}
+
+
+static char *
+truncate_string(const char *string, size_t max_len)
+{
+	size_t len = strlen(string);
+	char *retval;
+
+	if (len <= max_len)
+		return strdup(string);
+
+	retval = malloc(max_len + 1);
+	memcpy(retval, string, max_len);
+	retval[max_len] = '\0';
+
+	return retval;
+}
+
+#define MIN_TOKEN_LEN 4
+#define EXTRA_SPACE 3
+
+static char *
+generate_token_label(const char *token_label, const char *pin_label, size_t max_len)
+{
+	char *token;
+	size_t token_len;
+	char *pin;
+	size_t pin_len;
+	char *retval;
+	size_t retval_len;
+
+	token = trim_whitespace(token_label);
+	token_len = token ? strlen(token) : 0;
+
+	pin = trim_whitespace(pin_label);
+	pin_len = pin ? strlen(pin) : 0;
+
+	if (token == NULL && pin == NULL) {
+		return strdup("");
+	}
+	else if (token == NULL || pin_len > max_len - MIN_TOKEN_LEN - EXTRA_SPACE) {
+		/* Token label is not available or there is no space for it */
+		retval = truncate_string (pin, max_len);
+		free (token);
+		free (pin);
+
+		return retval;
+	}
+	else if (pin == NULL || strncmp(pin, "PIN", 4) == 0) {
+		/* The PIN label is not available or says just non-useful "PIN" */
+		retval = truncate_string (token, max_len);
+		free (token);
+		free (pin);
+
+		return retval;
+	}
+
+	if (token_len + pin_len + EXTRA_SPACE > max_len) {
+		size_t new_token_len;
+		char *new_token;
+
+		new_token_len = max_len - pin_len - EXTRA_SPACE;
+		new_token = truncate_string (token, new_token_len);
+		free (token);
+
+		token = new_token;
+	}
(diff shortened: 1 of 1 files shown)


muktupavels:
Oh, windows does not like my change... Before I try to fix that what do you think about following changes:
```
index c7c8f1b6..2098b81a 100644
@@ -1058,7 +1058,7 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 {
 	struct pkcs15_slot_data *fw_data;
 	struct sc_pkcs15_auth_info *pin_info = NULL;
-	char label[sizeof(auth->label) + sizeof(p15card->tokeninfo->label) + 10];
+	char label[33];
 	int write_protected;
 	scconf_block *atrblock;
 
@@ -1100,24 +1100,36 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 			pin_info = NULL;
 		}
 		else   {
-			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0) {
+			slot->token_info.flags |= CKF_LOGIN_REQUIRED;
+
+			if (!auth->label[0] || strncmp(auth->label, "PIN", 4) == 0) {
+				/* The PIN label is empty or says just non-useful "PIN" */
+				snprintf(label, sizeof(label), "%s",
+						p15card->tokeninfo ? p15card->tokeninfo->label : "");
+			} else if (!p15card->tokeninfo || strlen(auth->label) > 32 - 3 - 4) {
+				/* The tokeninfo is not available or there is no space for it */
+				snprintf(label, sizeof(label), "%.32s", auth->label);
+			} else {
 				/* Trim tokeninfo->label to make right parenthesis visible */
-				char tokeninfo_label[sizeof(p15card->tokeninfo->label)+1];
+				int max_tokeninfo_label = 32 - strlen(auth->label) - 3;
+				char tokeninfo_label[strlen(p15card->tokeninfo->label) + 1];
 				int len;
 				snprintf(tokeninfo_label, sizeof(tokeninfo_label), "%s",
-						p15card->tokeninfo ? p15card->tokeninfo->label : "");
+						p15card->tokeninfo->label);
 				tokeninfo_label[sizeof(tokeninfo_label)-1] = '\0';
 				for (len = strlen(tokeninfo_label) - 1; len >= 0 && isspace(tokeninfo_label[len]); len--) {
 					tokeninfo_label[len] = 0;
 				}
-				snprintf(label, sizeof(label), "%.*s (%s)",
-					(int) sizeof(auth->label), auth->label,
-					tokeninfo_label);
-			} else
-				/* The PIN label is empty or says just non-useful "PIN" */
-				snprintf(label, sizeof(label), "%s",
-						p15card->tokeninfo ? p15card->tokeninfo->label : "");
-			slot->token_info.flags |= CKF_LOGIN_REQUIRED;
+				if (strlen(tokeninfo_label) > (size_t) max_tokeninfo_label) {
+					len = max_tokeninfo_label - 1;
+					tokeninfo_label[len--] = '.';
+					tokeninfo_label[len--] = '.';
+					tokeninfo_label[len] = '.';
+				}
+				snprintf(label, sizeof(label), "%.*s (%.*s)",
+					(int) strlen(auth->label), auth->label,
+					max_tokeninfo_label, tokeninfo_label);
+			}
 		}
 	}
 	else   {
```

First there is no need to generate token label in `xxx (xxx)` format if `auth->label` size is bigger than 32 bytes. Also if `tokeninfo` is not available generating `xxx ()` seems useless to me. With my proposed changes `xxx (xxx)` format is used only if there is at least 4 byte space for `p15card->tokeninfo->label` and if that label is larger then available space it is ellipsized to indicate that it is truncated.

Still this does not fully fix token labels as code completely ignores fact that label might include utf8 characters in which case we might truncated label in "middle" of utf8 character. Does this project has function that counts characters? So we can properly truncate utf8 labels?

Oh, and looks like there is one more case that could produce undesired result... When `auth->label` is non-useful "PIN" we might create even more useless label "" if `tokeninfo` is not available.

frankmorgner:
Ah yeah, the usage of both, static arrays and pointers for labels is just confusing sometimes.

https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/22785777/job/sodooh8we6aqal9h#L839:
```
framework-pkcs15.c(1061): error C2057: expected constant expression
framework-pkcs15.c(1061): error C2466: cannot allocate an array of constant size 0
framework-pkcs15.c(1061): error C2133: 'label': unknown size
framework-pkcs15.c(1105): error C2057: expected constant expression
framework-pkcs15.c(1105): error C2466: cannot allocate an array of constant size 0
framework-pkcs15.c(1105): error C2133: 'tokeninfo_label': unknown size
framework-pkcs15.c(1107): warning C4034: sizeof returns 0
framework-pkcs15.c(1109): warning C4034: sizeof returns 0
framework-pkcs15.c(1113): warning C4034: sizeof returns 0
framework-pkcs15.c(1118): warning C4034: sizeof returns 0
framework-pkcs15.c(1124): warning C4034: sizeof returns 0
```

frankmorgner:
It should be done with malloc...

frankmorgner:
Your work regarding the truncation is a great starter (though it, again, needs malloc instead of an dynamic array)!

We also have this truncation problem for the other labels. You should truncate with ellipsis in a dedicated function.

Currently we don't have the need of detecting UTF-8 characters, but feel free to a add this.

frankmorgner:
... and you need to check for NULL dereferencing.

muktupavels:
I have updated pull request to generate token label in `<token label> (<PIN label>)` format. If necessarily then token label will be truncated to make sure there is space for PIN label.

In the end I am not sure if ellipsis is right thing to add. 3 dots would just waste space...

frankmorgner:
I think the three dots are good enough. The same expressed as utf8 character would have also taken 3 bytes (`"\u2026"`).

I think your code is a bit oversized to solve this problem. I've tried an alternative https://github.com/OpenSC/OpenSC/pull/1629, but did not yet have the capabilities to test it, which I hope to have next week.

muktupavels:
Ok. Let me know if I should update pull request to make truncation with ellipsis.

muktupavels:
Alternative solution has been merged.

PR #1621, opened 2019-03-05
notify: avoid forks in C_Initialize

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1507
fixes https://github.com/OpenSC/pkcs11-helper/issues/16

@loskutov, @rosly, could you please verify the fix?

##### Checklist


loskutov:
Unfortunately, the issue is still there: `sc_notify_init` is still called by `C_Initialize`, and it still forks inside.
Here's the stacktrace: https://paste.ubuntu.com/p/5FBnctyWxG/ (not that different from the previous ones, just with actual line numbers)

Issue #1271, opened 2018-02-20
Unused SC_SEC_OPERATION_AUTHENTICATE


frankmorgner:
### Problem Description

`opensc.h` contains `SC_SEC_OPERATION_AUTHENTICATE`, that's virtually unused. There is no way that `sc_set_security_environment()` is called with this type of operation; neither in PKCS#11, Minidriver, Tokend nor in any of the tools.

The following cards implement an MSE based on `SC_SEC_OPERATION_AUTHENTICATE`
- DNIe
- itacns
- iasecc
- starcos
- rutoken

Only iasecc and starcos are performing something usefull with it by translating `SC_SEC_OPERATION_SIGN` into `SC_SEC_OPERATION_AUTHENTICATE` in some cases. The other cards seem to be implementing some internal authentication, but as said above, I suspect this has never been tested.

@viktorTarasov, @germanblanco, @martinpaljak, @rickyepoderi @alonbl are you using this feature somewhere?

### Proposed Resolution

Remove the unused implementations in the card driver, but keep the flag and the implementation of iasecc and starcos.

alonbl:
I do not.

rickyepoderi:
Hi,

I think that in DNIe it is not used internally. The file `cwa14890.c` has another method  `cwa_set_security_env` which is the one that is used to start the secure channel. Despite all I've done a test removing `SC_SEC_OPERATION_AUTHENTICATE` from the switch in `card-dnie.c` and my tests still work ok.

Regards.

PR #1622, opened 2019-03-06
acos5: removed incomplete driver

frankmorgner:


martinpaljak:
LGTM, what's the story with authenticate?

frankmorgner:
See #1271 . I decided to only remove the flag where it can't be executed, but not to refactor the rest of the existing code.

PR #1624, opened 2019-03-06
tools: release context when card connection fails

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


martinpaljak:
Is it some random mishap with the vsmartcard setup or something is indeed broken with the test? As the change itself looks perfect (as is verified with several platforms) ?

martinpaljak:
Travis decided to change its mind :)

PR #1625, opened 2019-03-07
opensc-tool: do not use card driver to read ATR

### LINKED ISSUE #1606: "opensc-tool -a" not always prints ATR
Hi,

Symptom: "opensc-tool -a" is in some cases unable to print ATR

In default configuration "opensc-tool -a" prints correct ATR:
3b:be:18:00:00:41:05:10:00:00:00:00:00:00:00:00:00:90:00

If I configure some card driver to this ATR, several of card drivers prevent "opensc-tool -a" to print ATR (tested with drivers from src/libopensc/card-*.c):

atrust-acos: Segmentation fault
myeid: Failed to connect to card: Not allowed

authentic,belpic,cac,cac1,cardos,coolkey,epass2003,gemsafeV1,isoApplet,oberthur,openpgp:

Of course, forcing some card driver to another card is not OK, but basic functionality of "opensc-tool -a" should not depend on card driver.

I do not have solution for this behavior, either "opensc-tool" is to be fixed, or lot of card drivers must be fixed or, this behavior is tolerable. 

(At minimum .. segfault in atrust-acos should be fixed).

popovec:
If card driver fails to connect to card, 'opensc-tool -a' may fail to print
ATR even if ATR is available from card reader.  Before use of card driver,
do only card reader connect, then print ATR.  Only if it is neccesary, use
card driver for the rest of opensc-tool functions.  Fixes #1606

### DIFF
--- a/src/tools/opensc-tool.c
+++ b/src/tools/opensc-tool.c
@@ -712,6 +712,7 @@ int main(int argc, char *argv[])
 	const char *opt_conf_entry = NULL;
 	const char *opt_reset_type = NULL;
 	char **p;
+	struct sc_reader *reader = NULL;
 	sc_context_param_t ctx_param;
 
 	while (1) {
@@ -846,6 +847,24 @@ int main(int argc, char *argv[])
 			goto end;
 		action_count--;
 	}
+
+	err = util_connect_reader(ctx, &reader, opt_reader, opt_wait, verbose);
+	if (err) {
+		fprintf(stderr, "Failed to connect to reader: %s\n", sc_strerror(err));
+		err = 1;
+		goto end;
+	}
+	if (do_print_atr) {
+		if (verbose) {
+			printf("Card ATR:\n");
+			util_hex_dump_asc(stdout, reader->atr.value, reader->atr.len, -1);
+		} else {
+			char tmp[SC_MAX_ATR_SIZE*3];
+			sc_bin_to_hex(reader->atr.value, reader->atr.len, tmp, sizeof(tmp) - 1, ':');
+			fprintf(stdout,"%s\n",tmp);
+		}
+		action_count--;
+	}
 	if (action_count <= 0)
 		goto end;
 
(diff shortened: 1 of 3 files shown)


src/tools/opensc-tool.c
```diff
@@ -846,6 +847,22 @@ int main(int argc, char *argv[])
 			goto end;
 		action_count--;
 	}
+
+	err = util_connect_reader(ctx, &reader, opt_reader, opt_wait, verbose);
+	if (err)
+		goto end;
```
frankmorgner: please add the following:
```
		fprintf(stderr, "Failed to connect to reader: %s\n", sc_strerror(err));
		err = 1;

File: src/tools/util.c [hunk 2]
```diff
@@ -139,24 +137,36 @@ util_connect_card_ex(sc_context_t *ctx, sc_card_t **cardp,
...
-		if (sc_detect_card_presence(reader) <= 0) {
+		if (sc_detect_card_presence(*reader) <= 0) {
 			fprintf(stderr, "Card not present.\n");
 			return 3;
 		}
 	}
+	return 0;
+}
+int
+util_connect_card_ex(sc_context_t *ctx, sc_card_t **cardp,
```
martinpaljak: This function is internal, why not change the signature and not make a new _ex ?

popovec: The `util_connect_card_ex()`  is function with 6 arguments.  One specific argument is  `int do_lock`.  The `util_connect_card()` function is same as `util_connect_card_ex()` with `do_lock` set to 1.

There is 8 calls in whole opensc sources  to `util_connect_card_ex()` with `do_lock` set to  0,  and 9 calls to `util_connect_card()` (in 18 files).

IMHO, this PR should not change function `util_connect_card_ex ()` (because this introduces changes into files not affected by #1606).  If the `util_connect_card()` and `util_connect_card_ex ()` interface is not satisfactory, it should be solved in separate PR.

PR #1626, opened 2019-03-07
libopensc: recognize Latvian eID card

muktupavels:
This is enough to recognize Latvian eID card.

Known problem is PIN verifying and/or changing, but that does not seems to be Latvian eID card only problem. See - https://github.com/OpenSC/OpenSC/issues/1240.

Output (without keys and certificates part) from `pkcs15-tool --dump`: 
```
PKCS#15 Card [LATVIA ID]:
	Version        : 1
	Serial number  : (null)
	Manufacturer ID: Oberthur Technologies
	Last update    : 20170915121352Z
	Flags          : Login required, PRN generation

PIN [User PIN]
	Object Flags   : [0x00]
	Auth ID        : 02
	ID             : 01
	Flags          : [0x830], initialized, needs-padding, exchangeRefData
	Length         : min_len:4, max_len:64, stored_len:64
	Pad char       : 0xFF
	Reference      : 1 (0x01)

PIN [Global PUK]
	Object Flags   : [0x00]
	ID             : 02
	Flags          : [0x8F0], initialized, needs-padding, unblockingPin, soPin, exchangeRefData
	Length         : min_len:8, max_len:64, stored_len:64
	Pad char       : 0xFF
	Reference      : 2 (0x02)

PIN [Signature PIN]
	Object Flags   : [0x00]
	Auth ID        : 02
	ID             : 03
	Flags          : [0x832], local, initialized, needs-padding, exchangeRefData
	Length         : min_len:6, max_len:64, stored_len:64
	Pad char       : 0xFF
	Reference      : 129 (0x81)
	Path           : e828bd080ff2504f5420415750::
```


src/libopensc/card-iasecc.c
```diff
@@ -445,16 +447,22 @@ iasecc_oberthur_match(struct sc_card *card)
...
 	unsigned char *hist = card->reader->atr_info.hist_bytes;
+	size_t hist_bytes_len = card->reader->atr_info.hist_bytes_len;
 
 	LOG_FUNC_CALLED(ctx);
 
-	if (*hist != 0x80 || ((*(hist+1)&0xF0) != 0xF0))
-		LOG_FUNC_RETURN(ctx, SC_ERROR_OBJECT_NOT_FOUND);
-
-	sc_log_hex(ctx, "AID in historical_bytes", hist + 2, *(hist+1) & 0x0F);
+	if (*hist == 0x80 && ((*(hist+1)&0xF0) == 0xF0)) {
```
frankmorgner: Please fix the possible out of bounds access by checking `hist_bytes_len`. (I know, you didn't introduce it!)

src/libopensc/cards.h
```diff
@@ -204,6 +204,7 @@ enum {
 	SC_CARD_TYPE_IASECC_BASE = 25000,
 	SC_CARD_TYPE_IASECC_GEMALTO,
 	SC_CARD_TYPE_IASECC_OBERTHUR,
+	SC_CARD_TYPE_IASECC_LATVIA_EID,
```
frankmorgner: add your card type to the end of oberthur cards to preserve the existing ids

muktupavels: Sorry, not sure if I understand correctly. Should I move new type after  `SC_CARD_TYPE_IASECC_MI2`?

src/libopensc/card-iasecc.c
muktupavels: `&& (*(hist+1) & 0x0F) <= hist_bytes_len - 2`?

src/libopensc/cards.h
frankmorgner: yes, exactly

src/libopensc/card-iasecc.c
frankmorgner: `if (hist_bytes_len >= 2 && *hist == 0x80 && ((*(hist+1)&0xF0) == 0xF0))`

src/libopensc/card-iasecc.c
```diff
@@ -445,16 +447,22 @@ iasecc_oberthur_match(struct sc_card *card)
...
+		sc_log_hex(ctx, "AID in historical_bytes", hist + 2, *(hist+1) & 0x0F);
 
-	if (memcmp(hist + 2, OberthurIASECC_AID.value, *(hist+1) & 0x0F))
-		LOG_FUNC_RETURN(ctx, SC_ERROR_RECORD_NOT_FOUND);
+		if (memcmp(hist + 2, OberthurIASECC_AID.value, *(hist+1) & 0x0F))
```
frankmorgner: here's an other possibility for out of bounds reads

muktupavels:
@frankmorgner do you have more comments?

RayCic:
In my opinion this pull request should not be accepted.
Motivation: this Laivia eID support is incomplete and even **dangerous** - unsuspecting users can easily block their Latvian eID cards (they only have 3 PIN tries).
Correct solution should correctly implement PIN padding in iasecc driver or disable all PIN operations for Latvia eID card.

muktupavels:
Incomplete just like some other IAS-SEC cards... Of course PIN padding should be implemented, but in my opinion it should not be part of this pull request. PIN padding should be discussed in https://github.com/OpenSC/OpenSC/issues/1240.

RayCic:
Some comments about function "iasecc_oberthur_match":
https://github.com/OpenSC/OpenSC/blob/7768d16f6320c904255f652c8e79d566d3ca5230/src/libopensc/card-iasecc.c#L445-L478

1. in my opinion it is safer to write for Latvia eID its own match function (based on original "iasecc_oberthur_match")
2. are you sure there is no card which have  "hist" and "hist_bytes_len" parameters like Latvia eID card but  in reality it is some kind  new Oberthur card which may be incompatible with Latvia eID and vice versa (in other words - why you do not check "card->type")?
3. are you sure there is no card which have same ATR as current Latvia eID but have different AID and is unsupported (in other words - why you do not check Latvia eID AID)?
4. why you in Latvia eID case copy Oberthur's AID to "card->ef_atr->aid"? (I know, original Latvia eID driver did this, but IMHO it is bug)

IMHO, if you do not have access to all possible configurations of hardware you should not overgeneralize driver.

muktupavels:
Well I am new to smart cards... So no doubt that I might have done something wrong! Or maybe I understand things wrong.

From what I understand historical bytes are part of ATR. Are you suggesting that it is completely valid to have card with same ATR and be completely different card? I was thinking that I could check that historical bytes matches, but I was assuming that if ATR was already matched then in `iasecc_oberthur_match` it would be pointless to check if historical bytes are same.

About checking Latvia AID... Are you asking because Oberthur has check for AID in historical bytes? Latvia eID card does not have AID in historical bytes, it is proprietary format.

Because Latvia eID card has same AID. How that is bug?

Sure I can duplicate AID and match function if that is needed.

dengert:
I agree with  @RayCic on point (3) the matching. From your debug output, ATR is 
`3B DF 96 00 80 31 FE 45 00 31 B8 64 04 1F EC C1 73 94 01 80 82 90 00 EC` 

`https://smartcard-atr.appspot.com/parse?ATR=3bdf96008031fe450031b864041fecc173940180829000ec`
says "Possibly identified card: Ministry of Interior - France "Agent Card" 

If the applet has an AID conforming to ISO 7816-4 Section "8.1 Card identification"  (I assume it does and is defined as 'OberthurIASECC_AID') This is starts with 'AX' and is "International registration of application providers according to ISO/IEC 7816-5" and there for unique. 

See Section 8.2.2

Other parts of Section 8.1 say the AID **may** be in the ATR Historical bytes and/or in the EF.ATR Historical bytes. 

The other approach is to use Section 8.2.2.2. A special case of the SELECT FILE  is used in many OpenSC driver's match routines to SELECT the AID **and check the results of the SELECT to make sure it contains the AID**. If you card supports this, it would be the best way to match. In effect this means you are selecting the applet and the applet is not restricted to being on a specific type of card.

muktupavels:
That is not my debug output, your linked ATR is for `SC_CARD_TYPE_IASECC_MI`.

https://smartcard-atr.appspot.com/parse?ATR=3BDD18008131FE45904C41545649412D65494490008C
This is my card...

RayCic:
I am also new to smart cards... Started few days ago to investigate current state of things when Swedbank almost banned code cards.

Sorry, I was wrong about whole AID thing (Too much late night looking in original Latvia eID code).
Latvia eID use same AID as Oberthur.

Hmmm... If history data is derived solely from ATR and each card type (Oberthur & Latvia eID) have only one ATR entry in "iasecc_known_atrs" table, then following part is redundant:
https://github.com/OpenSC/OpenSC/blob/7768d16f6320c904255f652c8e79d566d3ca5230/src/libopensc/card-iasecc.c#L454-L467
This code is only useful if somebody will try to add card which is **almost** compatible with current Oberthur & Latvia eID cards.

muktupavels:
Perhaps I should just add `iasecc_latvia_eid_match` that works same as `iasecc_mi_match`? Of course trying to select `OberthurIASECC_AID` not `MIIASECC_AID`.

RayCic:
IMHO, you should also add "SC_CARD_TYPE_IASECC_LATVIA_EID" to:
/src/pkcs15init/pkcs15-iasecc.c
in same way like SC_CARD_TYPE_IASECC_OBERTHUR added

muktupavels:
Why? For example does our card support creation/deletion of the key slots? My initial goal is to get Latvia eID card recognized and be able to change PIN codes.

RayCic:
IMHO
Just for completeness.
If somebody in far future will find clean Latvian eID card, and try to add "init" functionality it will be hard to find why it not work (I suspect Latvian eID initialization is more like Oberthur way).

May be just some TODO type comment?

muktupavels:
Ok, then I guess we should start by fixing PIN verification / changing. Do you have any comment / suggestion for https://github.com/OpenSC/OpenSC/issues/1240#issuecomment-471192164?

dengert:
Sorry I assumed the debug output was yours. Did you or can you post some debug output with your card?

Since your card has the Historical  bytes in the ATR with "LATVIA-eID.." it should be for a "LATVIA-eID" card. 

But in general the card could be be used for multiple applets. That is one of the points in ISO 7816-4 Section "8.1 Card identification". The world has changed over the years. It used to be  a card and its applet where the same thing.  Today many different cards may have the same applet and may have multiple applets. (Even worse, someone can write their own version of an applet based on standards. Introducing new features and bugs and not implementing the required features of the apple.)

muktupavels:
@dengert what command should I use to get most useful debug output?

dengert:
The most useful debugging is done with `opensc.conf` 'debug = 7;' and'debug_file =<path to a file>;` 
Note it exposes the PIN and certificates. So you may want to change the pin or delete some lines. If you do, please mark that you have modified part of the output. 

Looking at your ATR, and your mode that adds:
```
	{ "3B:DD:18:00:81:31:FE:45:90:4C:41:54:56:49:41:2D:65:49:44:90:00:8C", NULL,
		"IAS/ECC v1.0.1 Latvia eID", SC_CARD_TYPE_IASECC_LATVIA_EID, 0, NULL },
```
The Historic bytes with "LATVIA-eID" make the ATR unique and `iasecc_match_card` routine called by OpenSC should match the ATR and set the `card->type = SC_CARD_TYPE_IASECC_LATVIA_EID.`

So as https://github.com/RayCic suggested, you may want to write your own `iasecc_init_latvia_eid` routine and not call `iasecc_init_oberthur`. 

You may also find `opensc-tool -c default` useful especially if a card does not work with any other driver. default is the  iso7816 driver. Use with some commands such as:

-  `--send-apdu` May be repeated to send more then one APDU s to the card. 

-  `--list-files` may also show something.

frankmorgner:

PR #1627, opened 2019-03-08
Remove un-lincenced header file

metsma:

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/pkcs15-esteid.c
```diff
@@ -49,6 +48,7 @@ set_string (char **strp, const char *value)
 static int
 sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
 {
+	static const size_t SC_ESTEID_PD_DOCUMENT_NR = 8;
```
frankmorgner: just curious: what's the benefit of adding `static` here?

metsma: It will only initialized once in program runtime.
Probably not big optimization gain here.
Thinking maybe define is even better and will not allocate size_t (if compiler will not optimize it out)

frankmorgner: With `const size_t` the compiler should already optimize it to something like a define internally without allocating any data, shouldn't it?

PR #1636, opened 2019-03-18
 Added LATVIA-eID smart card support (may  be fixes #1240)

### LINKED ISSUE #1240: PIN verification fails on IAS-SEC card

RayCic:
`opensc-tool -n` = 'IAS-ECC'
Overrides pull request #1626
May be fixes #1240

LATVIA-eID smart card is "IAS/ECC v1.0.1 Oberthur" compatible card (made by Oberthur for Latvian goverment) which have following differences from the original:
- different ATR format
- require PIN padding functionality

Pull request add generic pin padding support for almost **all** iasecc driver supported smart cards requiring it.

Pull request contain three commits with very detailed commit's comments.
Commits do not modify existing code. Only removes old conflicting redundant code and add new.

Differences from pull request #1626:
- support pin operations (for LATVIA-eID smart card)
- generic pin padding support (for almost **all** driver's cards)
- IMHO less intrusive (exclusive changes for LATVIA-eID smart card support consist of single small commit containing only two added lines (ATR definition))

##### Checklist

- [ ] macOS tokend is tested


RayCic:
I will separate my answer in several posts.

If you do not like code, it does not mean that one who wrote it is stupid or incompetent.
May be you do not have all information to judge?

Ok. Let's look at your objections.
BTW, did you even bothered to look at full commit message of commit which introduced this code?
And read full Issue #1240?

RayCic:
> It's unclear to me why you need an extra 150 lines of code to implement padding if sc_build_pin() has this already implemented (in less LOCs).

It's apples to oranges comparison.
My code workarounds serious bug (design flaw?)? My main intent was to make sure I do not brake anything that worked and do not introduce new bugs.
sc_build_pin() - almost unconditionally just copy data from buffer A to buffer B.

And I use this function where this fit.

BTW, I can bet my life on correctness of my code.
Interesting, developer of sc_build_pin() can do the same?
Let's look... After few minutes of blazing on function's source code...
Oh my God! At least 5(!) cases when application can CRASH! Not mere error message!!!
```
buf == NULL		// CRASH!
buflen < 1		// CRASH! only on some codepaths
pin == NULL		// CRASH!
pin->data == NULL	// CRASH! NULL for data is _valid_ value
pin->len < 0		// CRASH! only on some codepaths. -1 for pin is valid value (mean - PID must be checked by pinpad)
			//        and cast (size_t)-1 in function begining will create **huge positive** number 
```
If developer of sc_build_pin() function cared more of stability of his code then function would be larger, but my code smaller.

RayCic:
> card-iasecc.c implicitly uses sc_build_pin() by calling iso_ops->pin_cmd()

This comment show - you are unfamiliar with card-iasecc.c source code.

1) iso_ops->pin_cmd() is overridden by iasecc_pin_cmd() and in most cases iasecc_pin_cmd() is used.

2) Just look at following command's output:
```
grep -B50 -A4 'iso_ops->pin_cmd' card-iasecc.c
```
Notice functions' names where this code is found.
There is only three places where iso_ops->pin_cmd() can be found, and all three places are related to pin-pad's pin_cmd requests.
Such requests do not have PIN data.

3) globally, relaying on iso_ops->pin_cmd(), do not work.
Less then 10 drivers which use iso_ops DO NOT override iso_ops->pin_cmd(). Rest drivers (~30) use fully or partly its own implementation.

RayCic:
> so the real solution would be to correctly initialize the PIN flags for your card.

1) what do you mean by "initialize the PIN flags for your card"?

a) at production time ("initialize the PIN flags for your **clean** card")?
Sorry, but in most cases, cards are initialized by some authority (bank, government agency etc.).
I'm not employee of such entity. I can't be responsible for all such smart cards in universe.
I just mere eID card user who is unhappy with it's card support in Open Source.
And I'm not the only one. Look at Issue #1240

b) when card inserted in reader ("initialize the PIN flags for your card **when it is inserted in reader**")? 
Sorry, but PIN flags are burned-in at cards' initialization at factory (In most cases. Look at a) )
Mere mortals in most cases can not change such flags.

2) where did you get idea that flags are **incorrect**?

a) I made different tests with different OpenSC tools and did not saw **any** errors

b) you do not understand source of problem:
Each card support multiple different protocols/standards **at same time** (some times things they describe overlaps).
From bunch of protocols/standards my eID card support, it also supports:
- IAS ECC (respectable standard which describes communication with cards)
- PKCS#15 (respectable standard which describes information storage format)

They have overlapping area - pin data requirements. Both describe where and how to get such information.
Both describe how to get "minimal pin length" and "maximal pin length". For 'IAS ECC' it is (almost) all information you can get on topic.
BUT! PKCS#15 can provide "is pin padding required" and "padding length". 'IAS ECC' do not have means to express such information.
From 'IAS ECC' view point **all** cards in universe do not require padding.
And yes, 'IAS ECC' and PKCS#15 on my card agree on "minimal pin length" and "maximal pin length".
But card have also additional pin padding requirements from PKCS#15.

RayCic:
Last but not least:
# Real problem / Real solution

Real problem (~~is ***** OpenSC architecture.~~ Calm down, let's be polite) is OpenSC architecture which
do not have means for following entities to describe their pin processing capabilities and requirements to each other:
- card reader's drivers (Hmmm... First I thought they should be excluded, but then I saw some code in some drivers which mangle pin data)
- card's drivers (some drivers support only unpadded pins, some padded, some both. Some respect SC_PIN_CMD_NEED_PADDING flag and some not etc)
- cards itself (some cards (most modern?) support some configuration information storage in data transfer protocol independent way (hello PKCS#15!))

## Real solution: if OpenSC architecture do not have such means, they should be implemented.

Proposal (**very briefly**, there is too much to consider):

1) implement (let's call it) tri-state type, some thing like:
```C
typedef enum {Unknown = 0, Yes = 1, No = -1} TriState;
```

2) implement structure for pin capability and requirements storage, some thing like:
```C
struct pin_support {
// General padding capabilities and requirements
	TriState do_not_support_pin_padding;	// threat all pins as unpadded (iasecc)
	TriState always_pad_pin;		// pad all pins by itself uncoditionally
	TriState respect_need_padding_flag;	// pad pins by itself depending on SC_PIN_CMD_NEED_PADDING or
						// by quiering card by its own means (all drivers which do NOT overide iso_ops->pin_cmd())
	TriState require_external_padding;	// pin should be padded before sent to driver
	// etc...

// "Physical" limitations of entity
	int min_len;	// -1 == unkmown
	int max_len;	// -1 == unkmown
	int min_pad;	// -1 == irrelevant (if pin should be reciceived padded)
	int max_pad;	// -1 == irrelevant (if pin should be reciceived padded)
	// etc...
};
```

3) add member `'struct pin_support pins;'` to `'struct sc_card_driver'` and `'struct sc_reader_driver'`

4) fill this data in each driver with `Unknown` status

5) let drivers' developers fill real data

6) add to sec.c code which will analyse new & old data and consider
- pass pin_cmd() to card's driver or not (if reader or card's driver do not met requirements - reject)
- pad pin by itself or let it to the card's driver
- etc...

7) if all or part of relevant information is in `Unknown` state - reject, but show to user information how he can override default behaviour
This will require new variables in opensc.conf


## REMEMBER!!!
This pull request is not about OpenSC architecture change!
I just want my eID card supported by upstream as soon as possible.

RayCic:
Help! Somebody!

What mean those two failed CI checks?
I looked at error messages.
Both failed because of network problems -
both was unable to download some files for building environment.
Hmmm... It looks it is not problem of OpenSC or my code...

frankmorgner:
OK, this is definately going too far. I would never ever bet my life on any code! Let's agree on that I'm unfamiliar with this code and I'll let @viktorTarasov decide.

`sc_build_pin()` is an internal function and should be called with correct arguments. If you see any code paths where that's not the case, please let me know and I'll have a look.

RayCic:
> OK, this is definately going too far. I would never ever bet my life on any code!

Sorry, forgot `<SARCASM>` tag.

I just try to program in the way like it is code for autopilot for aircraft on which I will fly as a passenger.  :wink:

> sc_build_pin() is an internal function and should be called with correct arguments.

Whhyyyyyyyy???!!!!..... Seriously.
- where is description what `sc_build_pin()` considers as `correct arguments`?
- as I mentioned before, for example `pin.data == NULL` and `pin.len == -1` are generally  **correct values** for this data structure
- is it better to crash, than give descriptive error message?
- is this really costly to add required checks?
- do we have very strict memory restrictions?
- do we have very strict processing power restrictions?
- do we have very strict real time reaction restrictions?
- why some other checks are implemented?
- etc... etc... etc...

I do not need response to this post. Just think about this questions...

frankmorgner:

PR #1639, opened 2019-03-19
Change u8 *data to const because sc_apdu unsigned char *data is const

metsma:

##### Checklist

- [ ] macOS tokend is tested


metsma:
> changing to const is OK. But I think the opening brackets require two indents instead of one.

The ident in header was mess with spaces and tabs.
I followed style like rest of header file

martinpaljak:
I think clang tools are common enough and good ebnough that we can have a project based formatting configuration and maybe even commit hooks to just do it automatically? 

The formatting could be anything, as long as there is a easy to use integrated tooling to apply the common.

PR #1640, opened 2019-03-21
iso7816_read_binary: do not assume that 6282 is returned with data

martinpaljak:
Instead of a double check_sw call in case there is no data, assume
that a SW is properly sent by the card and do not expose
SC_ERROR_FILE_END_REACHED outside of the function
(most probably sc_pkcs15_read_file)

This is to facilitate Estonian eID 2018+ that instead of properly returning
6282 with trunkated data, 9000 is returned and next READ BINARY returns
6b00 (invalid p1/p2). The change should be generally harmless for well-behaving
cards and makes re-mapping 6B00 to 6282 work as expected.

Tested with Estonian eID (2018+)

Change-Id: I7511ab4841d3bcdf8d6f4a37a9315ea4ac569b10

##### Checklist


martinpaljak:
Ping on this one. Required for EstEID 2018+ driver.

PR #1642, opened 2019-03-22
Use strdup and fix all casts 

metsma:

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-esteid.c
+++ b/src/libopensc/pkcs15-esteid.c
@@ -30,20 +30,10 @@
 #include <stdio.h>
 
 #include "common/compat_strlcpy.h"
-#include "common/compat_strlcat.h"
 
 #include "internal.h"
 #include "opensc.h"
 #include "pkcs15.h"
-
-static void
-set_string (char **strp, const char *value)
-{
-	if (*strp)
-		free (*strp);
-	*strp = value ? strdup (value) : NULL;
-}
-
 
 static int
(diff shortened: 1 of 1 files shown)


frankmorgner:
Does the card have a valid EF.TokenInfo? If so, the synthetic profile parsing of EF.TokenInfo will allocate resources, which this PR will not free when the emulated profile is loaded.

metsma:
> Does the card have a valid EF.TokenInfo? If so, the synthetic profile parsing of EF.TokenInfo will allocate resources, which this PR will not free when the emulated profile is loaded.

No

martinpaljak:
I looked at the code dealing with the emulation parts and binding and think that life has changed from when that was written. options like turning emulation off and the order of detection and the number of emulated drivers... Maybe we should change that.

frankmorgner:
What do you have in mind? SC_CARD_TYPE_MCRD_ESTEID_V30 is already listed in sc_pkcs15_is_emulation_only...

src/libopensc/pkcs15-esteid.c
```diff
@@ -166,7 +150,7 @@ sc_pkcs15emu_esteid_init (sc_pkcs15_card_t * p15card)
 		pin_info.attrs.pin.stored_length = 12;
 		pin_info.attrs.pin.max_length = 12;
 		pin_info.attrs.pin.pad_char = '\0';
-		pin_info.tries_left = (int)tries_left;
+		pin_info.tries_left = buff[5];
```
frankmorgner: you need to check the length

PR #1643, opened 2019-03-24
fix sc_bin_to_hex: ckecking for SC_ERROR_BUFFER_TOO_SMALL and SC_ERRO…

carblue:
…R_INVALID_ARGUMENTS

Current code is incorrect regarding 4 issues/corner cases:
1. No error condition checking is done for case in_len==0. Then, with out == NULL, SIGSEGV occurs.

2. With in_len>1, SC_ERROR_BUFFER_TOO_SMALL may occur even if enough memory for out is allocated, depending on out_len passed

3. With in_len==1, SC_ERROR_BUFFER_TOO_SMALL may occur even if enough memory for out is allocated, depending on out_len and in_sep passed

4. Don't separate with nonprintable ASCII characters, thus any separator<' ' should suppress separation, same as it's the case for separate=0 currently

IMHO, at least out_len should be documented.

So far I didn't do testing the way I think the PR template asks for: Build opensc and execute with a card and driver (there still is no internal opensc driver for my card), BUT
the PR results from requirements/testing with diverse input to successfully (and with correct results) call sc_bin_to_hex in libopensc.so from Rust code.
The memory-safe Rust wrapping function around the call to sc_bin_to_hex allocates exactly as many bytes as required for out (depending on the other input arguments), but must cheat out_len (for the call to sc_bin_to_hex) to be sizeof(out) +1 or sizeof(out) +2 (depending on arguments) rather than the correct out_len=sizeof(out), in order to get around wrong SC_ERROR_BUFFER_TOO_SMALL issues.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/opensc.h
+++ b/src/libopensc/opensc.h
@@ -1378,7 +1378,31 @@ const sc_path_t *sc_get_mf_path(void);
 /********************************************************************/
 
 int sc_hex_to_bin(const char *in, u8 *out, size_t *outlen);
-int sc_bin_to_hex(const u8 *, size_t, char *, size_t, int separator);
+/**
+ * Converts an u8 array to a string representing the input as hexadecimal,
+ * human-readable/printable form. It's the inverse function of sc_hex_to_bin.
+ *
+ * @param in: The u8 array input to be interpreted, may be NULL iff in_len==0
+ * @param in_len: Less or equal to the amount of bytes available from in
+ * @param out: output buffer offered for the string representation, *MUST NOT*
+ *             be NULL and *MUST* be sufficiently sized, see out_len
+ * @param out_len: *MUST* be at least 1 and state the maximum of bytes available
+ *                 within out to be written, including the \0 termination byte
+ *                 that will be written unconditionally
+ * @param separator: The character to be used to separate the u8 string
+ *                   representations. Any value<32 (32 corresponds to ' '
+ *                   i.e. space) will suppress separation
+ * The algorithm will require for out_len (otherwise resulting in an error
+ * SC_ERROR_BUFFER_TOO_SMALL):
+ * 1 (\0 termination byte) + 2*in_len + optional_separation_bytes
+ * opt_separation_bytes = 0  if in_len<=1 or separator<' ',  otherwise
+ * opt_separation_bytes = in_len-1, i.e. there will be no trailing separator
+ * Example: input [0x3f], in_len=1, requiring an out_len>=3, will write to out:
+ * [0x33, 0x66, 0x00] which reads as "3f"
+ * Example: input [0x3f, 0x01], in_len=2, separator=':', req. an out_len>=6,
+ * writes to out: [0x33, 0x66, 0x3A, 0x30, 0x31, 0x00] which reads as "3f:01"
+ */
+int sc_bin_to_hex(const u8 *in, size_t in_len, char *out, size_t out_len, int separator);
 size_t sc_right_trim(u8 *buf, size_t len);
 scconf_block *sc_get_conf_block(sc_context_t *ctx, const char *name1, const char *name2, int priority);
 
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -121,12 +121,16 @@ int sc_bin_to_hex(const u8 *in, size_t in_len, char *out, size_t out_len,
 	unsigned int	n, sep_len;
 	char		*pos, *end, sep;
 
+	if (out == NULL || out_len == 0 ||
+		(in == NULL && in_len > 0))
+		return SC_ERROR_INVALID_ARGUMENTS;
 	sep = (char)in_sep;
-	sep_len = sep > 0 ? 1 : 0;
+	sep_len = sep > 31 ? 1 : 0;
 	pos = out;
+	/* end: position one-past allowed to write to */
 	end = out + out_len;
 	for (n = 0; n < in_len; n++) {
-		if (pos + 3 + sep_len >= end)
+		if (pos + 3 + (n && sep_len ? sep_len : 0) > end)
 			return SC_ERROR_BUFFER_TOO_SMALL;
 		if (n && sep_len)
 			*pos++ = sep;


carblue:
Testing libopensc.so build with PR applied (and not cheating about out_len when calling sc_bin_to_hex):

in= [0x3F, 0x00, 0x41, 0x00, 0x43, 0x00]
out allocation done in wrapping code as required by algorithm for in_len and separator:
sizeof(out) == 1+ in_len*2 + (separator>31 && in_len ? in_len-1 : 0);
out_len=sizeof(out) for all the following tests

All results are what's expected and no SC_ERROR_BUFFER_TOO_SMALL can be triggered any more iff out_len==sizeof(out's minimum required):

Parameter combinations tested:
in_len=0; separator='\n'; (out_len=1) results in: ""
in_len=1; separator='\n'; (out_len=3) results in: "3f"
in_len=2; separator='\n'; (out_len=5) results in: "3f00"
in_len=3; separator='\n'; (out_len=7) results in: "3f0042"

in_len=0; separator=':';  (out_len=1) results in: ""
in_len=1; separator=':';  (out_len=3) results in: "3f"
in_len=2; separator=':';  (out_len=6) results in: "3f:00"
in_len=3; separator=':';  (out_len=9) results in: "3f:00:42"

out=NULL;           results in: SC_ERROR_INVALID_ARGUMENTS
out_len=0;          results in: SC_ERROR_INVALID_ARGUMENTS
in=NULL; in_len=1;  results in: SC_ERROR_INVALID_ARGUMENTS

My first thought about code depending on this PR, glancing over output from grep -rnw sc_bin_to_hex 2>/dev/null:
There shouldn't be problems, except there is strange code relying on the prior incorrect SC_ERROR_BUFFER_TOO_SMALL triggering, which is relaxed now to the correct minimum limit.
Code that tends to offer more out_len than required, is not affected.

carblue:
fix typo:
in= [0x3F, 0x00, 0x42, 0x00, 0x43, 0x00]

frankmorgner:
If the caller wants to use a non printable character as seperator, he should be fine to do so, I would not force some ASCII table conversion requirement.

If we start rewriting this function, I think we could make it both, more readable and more efficient. What do you think about this:
```c
int sc_bin_to_hex(const u8 *in, size_t in_len, char *out, size_t out_len,
		  int in_sep)
{
	const char hex[] = "0123456789abcdef";

	if (in_sep > 0) {
		if (out_len < in_len*3 || out_len < 1)
			return SC_ERROR_BUFFER_TOO_SMALL;
	} else {
		if (out_len < in_len*2 + 1)
			return SC_ERROR_BUFFER_TOO_SMALL;
	}

	while (in_len) {
		unsigned char value = *in++;
		*out++ = hex[(value >> 4) & 0xF];
		*out++ = hex[ value       & 0xF];
		in_len--;
		if (in_len && in_sep > 0)
			*out++ = (char)in_sep;
	}
	*out = '\0';

	return SC_SUCCESS;
}
```
This is tested in PKCS#11, catches the corner cases and avoids sprintf and checks the length only once.

carblue:
I agree with Your arguing about nonprintable ASCII characters as separator. And I think we have to consider the inverse function sc_hex_to_bin as well, to be consistent (at first glance this one seems to care for colon ':' and space ' ' only and raise SC_ERROR_INVALID_ARGUMENTS for all other 'non-hexadecimal' ASCII chars). Thus that function should need fixes or a rewrite as well, and maybe also other code.

Yes, a more readable and more efficient rewrite is even better. I just limited myself to minimal invasive changes to get it working as expected, pointing to the problem, knowing that the repeated check against end is inefficient.

From just reading the code rewrite, it seems to be okay, exept:
```
if (out_len < in_len*3 || out_len < 1)
```
should also treat in_len==0 specifically, resulting in
```
if (out_len < (in_len>0 ? in_len*3 : 1))
```
I'll test with that code later and report.
Thanks for reviewing.

As we got in touch, another topic that's dear to my heart:
I would like to do more with/for opensc code, but the lack (not meaning absence but by far not enough) of API-documentation (or did I miss something outside the source base?) slows me down so much, that I always get stuck (e.g. writing a new driver). That's why I documented at length in opensc.h for sc_bin_to_hex, and that's what I would like to see everywhere in headers: I want to read from API-documentation what a function does exactly, how to use it safely, what exactly the parameters are for etc., so that I'm not obliged to read implementing code, except I want to, e.g. for details or for fixing something.
Currently, I have to read opensc code over and over again, very time consuming.
How are the chances that this situation will improve apart from want I'm willing to document myself?

Jakuje:
Would it make sense to automate the tests you ran in some unit test?

frankmorgner:
> From just reading the code rewrite, it seems to be okay, exept:
> 
> ```
> if (out_len < in_len*3 || out_len < 1)
> ```
> 
> should also treat in_len==0 specifically, resulting in
> 
> ```
> if (out_len < (in_len>0 ? in_len*3 : 1))
> ```

`in_len = 0` is handled correctly by checking whether `out` can hold at least one byte of `'\0'`. Hence, checking for `out_len < 1` is enough.

> As we got in touch, another topic that's dear to my heart:
> I would like to do more with/for opensc code, but the lack (not meaning absence but by far not enough) of API-documentation (or did I miss something outside the source base?) slows me down so much, that I always get stuck (e.g. writing a new driver). That's why I documented at length in opensc.h for sc_bin_to_hex, and that's what I would like to see everywhere in headers: I want to read from API-documentation what a function does exactly, how to use it safely, what exactly the parameters are for etc., so that I'm not obliged to read implementing code, except I want to, e.g. for details or for fixing something.
> Currently, I have to read opensc code over and over again, very time consuming.
> How are the chances that this situation will improve apart from want I'm willing to document myself?

Yes, you are absolutely right and although I'm working with OpenSC for years, I still discover new things that are poorly documented. A better API documentation is key for sustaining a long term and diverse community engagement. Unfortunately, there are still more important (and more exciting :wink: ) things to do, like testing/simulating, fuzzing/code analysis, simplifying user interaction, documenting user settings. I'm happy to integrate your documentation, but currently I'm not very optimistic, that the current state is easily fixed.

carblue:
@Jakuje 
>Would it make sense to automate the tests you ran in some unit test?

Yes, You are absolutely right; this function is a perfect candidate for unit testing and thus prevent regressions and my favorite language(s) D (and perhaps Rust; I'm still at learning level) do have unit testing integrated (in the language's core), that I do use sooner or later. But the unit test would best sit in the opensc code or close to it and I don't know if/how that is handled in C, nor whether OpenSC does use such a tool. If so, I'll be happy to feed in my test cases

P.S.
From the D ecosystem I know they even integrate unit tests and the built-in documentation generator:
E.g. for the standard library: One can simply specify in source code which unit tests (usually they sit very close to the implementations, within the source code) shall go into the documentation and the unit test(s) will then show up there as usage example(s) of function/struct etc., at the right place, all done by the compiler. I'm rarely looking into the standard library's sources, because I don't need to. All is in the docs. That's what I'm used to and so much missing here in OpenSC.

carblue:
Concerning in_len==0, Yes You are right, both if expressions are equivalent.
I tested Your code successfully, matching my results,
exept:
out=NULL;               <- SIGSEGV occurs
in=NULL; in_len=1;  <- SIGSEGV occurs

PR #1646, opened 2019-03-26
optimize bin/hex low parsing level functions

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


src/libopensc/libopensc.exports
```diff
@@ -128,6 +128,7 @@ sc_get_version
 sc_hex_dump
 sc_dump_hex
 sc_hex_to_bin
+sc_hex_to_bin_seperators
```
plaes: spelling: separator

PR #1648, opened 2019-04-01
Implement Secret Key write object

metsma:

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2437,7 +2437,7 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
 	CK_RV rv;
 
 	if (type != NULL) {
-		if (strncmp(type, "AES:", strlen("AES:")) == 0 || strncmp(type, "aes:", strlen("aes:")) == 0) {
+		if (strncasecmp(type, "AES:", strlen("AES:")) == 0) {
 			CK_MECHANISM_TYPE mtypes[] = {CKM_AES_KEY_GEN};
 			size_t mtypes_num = sizeof(mtypes)/sizeof(mtypes[0]);
 			const char *size = type + strlen("AES:");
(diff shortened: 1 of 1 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -3153,9 +3153,63 @@ static int write_object(CK_SESSION_HANDLE session)
...
-	if (opt_object_class == CKO_DATA) {
+		break;
+	case CKO_SECRET_KEY:
+		clazz = CKO_SECRET_KEY;
+		type = CKK_AES;
+
+		if (opt_key_type != 0) {
+			if (strncmp(opt_key_type, "AES:", strlen("AES:")) == 0 || strncmp(opt_key_type, "aes:", strlen("aes:")) == 0)
+				type = CKK_AES;
+			else if (strncmp(opt_key_type, "DES:", strlen("DES:")) == 0 || strncmp(opt_key_type, "des:", strlen("des:")) == 0)
```
martinpaljak: Not sure if any new code should be added for DES (as in single key DES) whereas there are plenty of systems that make use of 2 key 3DES (and would be nice to have explicit support for this in import level, without having to do manual key shuffling before)

metsma: I think this option list should to be equal whith generate key options.
https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L2460
Or will I remove this option from gen_key also?

src/tools/pkcs11-tool.c
```diff
@@ -3153,9 +3153,61 @@ static int write_object(CK_SESSION_HANDLE session)
...
-	if (opt_object_class == CKO_DATA) {
+		break;
+	case CKO_SECRET_KEY:
+		clazz = CKO_SECRET_KEY;
+		type = CKK_AES;
+
+		if (opt_key_type != 0) {
+			if (strncmp(opt_key_type, "AES:", strlen("AES:")) == 0 || strncmp(opt_key_type, "aes:", strlen("aes:")) == 0)
+				type = CKK_AES;
+			else if (strncmp(opt_key_type, "DES3:", strlen("DES3:")) == 0 || strncmp(opt_key_type, "des3:", strlen("des3:")) == 0)
```
frankmorgner: `strncasecmp()`?

src/tools/pkcs11-tool.c
```diff
@@ -3153,9 +3153,61 @@ static int write_object(CK_SESSION_HANDLE session)
...
+		if (opt_object_label != NULL) {
+			FILL_ATTR(seckey_templ[n_seckey_attr], CKA_LABEL, opt_object_label, strlen(opt_object_label));
+			n_seckey_attr++;
+		}
+		if (opt_object_id_len != 0)  {
+			FILL_ATTR(seckey_templ[n_seckey_attr], CKA_ID, opt_object_id, opt_object_id_len);
+			n_seckey_attr++;
+		}
+
+		rv = p11->C_CreateObject(session, seckey_templ, n_seckey_attr, &seckey_obj);
```
frankmorgner: for consistency, please move this part down to where the other objects are created.

PR #1649, opened 2019-04-01
Option to delete object by index

metsma:
When object does not have label and ID

##### Checklist

- [ ] macOS tokend is tested


martinpaljak:
HSM Vendors have this "nice" habit of relying on almost-persistent slot ID-s, but generally speaking, indexes are not persistent and depend on listing. Maybe you could locate objects via labels or ID-s instead?

frankmorgner:
thanks

PR #1650, opened 2019-04-02
Add enable-openssl with optional DIR for argument

metsma:
--enable-openssl=/usr/local/opt/openssl

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,6 @@ define([VS_FF_PRODUCT_URL], [https://github.com/OpenSC/OpenSC])
 
 m4_sinclude(version.m4.ci)
 
-m4_define([openssl_minimum_version], [0.9.8])
 
 AC_INIT([PRODUCT_NAME],[PACKAGE_VERSION_MAJOR.PACKAGE_VERSION_MINOR.PACKAGE_VERSION_FIX[]PACKAGE_SUFFIX],[PRODUCT_BUGREPORT],[PRODUCT_TARNAME],[PRODUCT_URL])
 AC_CONFIG_AUX_DIR([.])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -171,8 +171,8 @@ AC_ARG_ENABLE(
 
 AC_ARG_ENABLE(
 	[openssl],
-	[AS_HELP_STRING([--enable-openssl],[enable OpenSSL linkage @<:@detect@:>@])],
-	,
+	[AS_HELP_STRING([--enable-openssl[[=DIR]]],[enable OpenSSL linkage [installed at prefix DIR] @<:@detect@:>@])],
+	[OPENSSL_CFLAGS="-I${enableval}/include" OPENSSL_LIBS="-L${enableval}/lib -lcrypto"],
```
frankmorgner: OpenSSL may have other dependant libraries...

metsma: Is there possible assign prefix then to pkg-config search path?
Seems like the "enable-openssl" option is also wrong and should be "with-openssl".

metsma: I want avoid this
https://github.com/OpenSC/OpenSC/blob/master/MacOSX/build-package.in#L42-L43

frankmorgner: You may also need to set `PKG_CONFIG_SYSROOT_DIR`

metsma: Is there reason why we link with libssl?
```
otool -L /usr/local/lib/libopensc.dylib 
/usr/local/lib/libopensc.dylib:
	/usr/local/lib/libopensc.6.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
```

metsma:
Actually I don't care build-package content
I want use intuitive configure with parameters and without messing around OPENSSL_INCLUDE OPENSSL_LIB variables.
Every other project has nice ./configure --with-openssl=/usr/local/opt/openssl path option

frankmorgner:
I think the other projects were including some OpenSSL path, because it took some time until OpenSSL adopted a pkg-config configuration file. Nowadays, modifying `PKG_CONFIG_PATH`+`PKG_CONFIG_TOP_BUILD_DIR` in `MacOSX/build-package.in` is all you need.

But if you really care, please include https://www.gnu.org/software/autoconf-archive/ax_check_openssl.html

PR #1653, opened 2019-04-08
pkcs11-tool: Add keys access flags

dfortinskiy:
##### Checklist

- [ ] macOS tokend is tested


martinpaljak:
The way I read the specs always and never flags shoukd not be in ideal world settable by the user but be derived from existing flags. It does not hurt to have them in the tool but implementations are free to reject changing them.

PR #1657, opened 2019-04-12
Support for MyEID 4K RSA and clean ups

fabled:
This adds support for upcoming MyEID card version 4.5 which supports 4K RSA keys. Additional care was taken to simplify the driver by merging code where possible.

Tested using `pkcs11-tool` testing suite and key importing; and `pkcs15-tool` key generation on MyEID 4.0.2 (current production) and MyEID 4.5.2 (development prototype).


fabled:
@hhonkanen Please review. Perhaps you can also provide additional feedback results. Thanks.

fabled:
@frankmorgner @viktorTarasov @dengert Please, take a look at this at your convenience.

src/libopensc/card-myeid.c
```diff
@@ -123,27 +114,14 @@ static int myeid_get_card_caps(struct sc_card *card, myeid_card_caps_t* card_cap
...
-		if (memcmp(card->atr.value, defatr, len) != 0)
-			continue;
-		match = i;
-		break;
+	/* Normally the historical bytes are exactly "MyEID", but there might
+	 * be some historic units which have a small prefix byte sequence. */
+	if (card->reader->atr_info.hist_bytes_len >= 5 &&
+	    !memcmp(&card->reader->atr_info.hist_bytes[card->reader->atr_info.hist_bytes_len - 5], "MyEID", 5)) {
+		card->type = SC_CARD_TYPE_MYEID_GENERIC;
+		return 1;
```
frankmorgner: You may want to register your card for Windows' minidriver via `win32/customactions.cpp`.

dengert:
It looks good.

src/libopensc/card-myeid.c
fabled: Thanks. If done it's worth separate commit. I let @hhonkanen to figure out this as separate step if needed.

frankmorgner: In a similar case, I've decided to simply list all possible ATRs with an ATR mask:
https://github.com/OpenSC/OpenSC/blob/46930fc0c46a41c4a2e0121d7012cab6030f0dfd/src/libopensc/card-sc-hsm.c#L62-L131
and the corresonding code in the Windows installer:
https://github.com/OpenSC/OpenSC/blob/46930fc0c46a41c4a2e0121d7012cab6030f0dfd/win32/customactions.cpp#L77-L100

hhonkanen: I am not sure about registering for Windows minidriver. Aventra maintains a MyEID specific minidriver, which is tested using Microsoft Hardware Lab Kit tools upon each release. I don't have experience on using MyEID with OpenSC's minidriver and we ususally recommend our customers to use our MyEID Minidriver for CSP/CNG support on Windows, and OpenSC for PKCS#11. Because I don't know what OpenSC's minidriver is like compared to our own, I think it is better to consider this separately, and not within this PR. 

If somebody has installed our minidriver and then installs OpenSC for PKCS#11, could it accidentally override the previous minidriver installation? Is the minidriver option enabled by default in OpenSC's Windows installer?

fabled: > In a similar case, I've decided to simply list all possible ATRs with an ATR mask:
> 
> [OpenSC/src/libopensc/card-sc-hsm.c](https://github.com/OpenSC/OpenSC/blob/46930fc0c46a41c4a2e0121d7012cab6030f0dfd/src/libopensc/card-sc-hsm.c#L62-L131)

Other card drivers e.g. `card-piv.c` prefer historical bytes matching. As it's wider match and simpler code I chose this approach.

Seems there's some drivers that first try full ATR match and fall back to historical bytes matching. And some that do just ATR matching. So there's a variety of ways to do it in the tree.

If preferred I'm happy to convert the code to use instead _sc_match_atr(). Perhaps it would be good to document the preference somewhere? Or maybe I missed it?

dengert: In regard to PIV driver: Windows provides a builtin PIV driver, based on looking for the PIV AID. 

All PIV cards must respond to SELECT AID for the PIV AID.  There are many PIV  NIST approved cards and PIV look-a-like applets on various cards/tokens. Many of these don't implement required features correctly or add features.  Thus much of the OpenSC PIV match and init code deals with determining if this card/token could have multiple applets, is an approved or look-a-like card. Some of this can be determined from known ATRs or if the Historical bytes contain the PIV AID. 

Look-a-like applet/card vendors write their applets to work with Windows. So most people will use the windows driver. Thus not much testing has been done with the OpenSC minidriver with PIV.

hhonkanen:
The code looks good. Here are my test results:

Using MyEID 4.0.17: 
- unwrap secret key using RSA: OK
- unwrap a secret key with CKA_TOKEN=FALSE using RSA: OK
- unwrap a secret key using a secret key: OK
- wrap a secret key: OK
- generate 1024 bit RSA key using pkcs11-tool: OK
- generate 2048 bit RSA key using pkcs11-tool: OK
- signing and decrypting using pkcs11-tool: OK
- EC key generation and signing: OK

Using MyEID 4.5.0 development version:
- Generate 4K RSA key using pkcs11-tool: OK
- sign using 4K key, pkcs11-tool and pkcs15-crypt: OK

mescheryakov1:
> * Generate 4K RSA key using pkcs11-tool: OK
> * sign using 4K key, pkcs11-tool and pkcs15-crypt: OK

Sorry for the distracted question, 
how long does this card take to generate a 4k key and sign on this key?

hhonkanen:
@mescheryakov1 I don't have any accurate statistics on key gen/signing speed yet, but very roughly estimate would be around 30-60 seconds for key gen. Signing with a 4K key is not significantly slower than with a 2K key. I would say it takes a couple or a few seconds longer.

mescheryakov1:
@hhonkanen thank you very much!

Issue #1609, opened 2019-02-16
ECDH not working


popovec:
ECDH operation (myeid driver) with opensc from git, commit 20daced60575c508c81b3fbb3ea2209b138b71f7.

pkcs11-tool  --login --pin XXXXXXXX --derive -m ECDH1-DERIVE --id 10   --input-file tmp/ecdh-bob-pub.der   --output-file tmp/ecdh-bob-alices-shared.der
error: PKCS11 function C_DeriveKey failed: rv = CKR_GENERAL_ERROR (0x5) Aborting.

APDUs from debug.. 
00 A4 08 0C 04 50 15 4B 01                   < select file
00 22 41 A4 0A 80 01 04 81 02 4B 01 84 01 00 < set sec envir.
00 2A 80 86 32 00 04 CC 99 26 38 32 9A ....  < operation decrypt! not general auth

Last working commit: be2cc38565c040af0a36b671d59877984678aa19 generates correct APDU:
00 86 00 00 35 7C 33 85 31 04 8C D7 .... < general authenticate

Wrong APDU is selected by this condition (src/libopensc/card-myeid.c):

        if (priv->sec_env && priv->sec_env->algorithm == SC_ALGORITHM_EC
                && priv->sec_env->operation == SC_SEC_OPERATION_DERIVE
                && priv->sec_env->algorithm_flags & SC_ALGORITHM_ECDSA_RAW)

This fail on priv->sec_env->algorithm_flags & SC_ALGORITHM_ECDSA_RAW, because priv->sec_env->algorithm_flags is zero. 

priv->sec_env->algorithm_flags comes from padding.c:511:sc_get_encoding_flags:

Relevant part from log:

Wrong: (commit e5707b545e5a2dc33b0ca52a8bf63f36f71b3d85)
padding.c:472:sc_get_encoding_flags: iFlags 0x0, card capabilities 0x80300300
padding.c:511:sc_get_encoding_flags: pad flags 0x0, secure algorithm flags 0x0
padding.c:512:sc_get_encoding_flags: returning with: 0 (Success)

Correct: (commit be2cc38565c040af0a36b671d59877984678aa19)
padding.c:288:sc_get_encoding_flags: iFlags 0x0, card capabilities 0x80300300
padding.c:322:sc_get_encoding_flags: pad flags 0x0, secure algorithm flags 0x100000
padding.c:323:sc_get_encoding_flags: returning with: 0 (Success)

[...]

PR #1677, opened 2019-05-16
Unbreak derive for sc-hsm and myeid cards

Jakuje:
This should address the #1609, which was caused by refactoring of the mechanisms passing in the post-0.19.0 work.

I tested this with sc-hsm using the following test at this moment (from #1609):
```
$ ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so --login --pin 123456 --derive -m ECDH1-DERIVE --id 219824b5b02d2c228a892eec8acbf2de1a3ea5b2 --input-file cert.der --output-file ecdh-bob-alices-shared.der
Using derive algorithm 0x00001050 ECDH1-DERIVE
```
Feedback, comments welcomed.

What is missing:
- [x] Testing with actual MyEID
- [x] Testing with software implementation of MyEID (in CI?) [1]
- [x] Implementing automatic tests for the derive in `p11test` to run in CI

##### Checklist

- [ ] macOS tokend is tested

[1] https://sourceforge.net/projects/oseid/


Jakuje:
Ok, it will be more complicated. For some reason, this now breaks the hashed mechanisms for RSA and ECDSA. I will investigate later.

Jakuje:
The travis simulator failure looks like related to the modification of the pcscd that virtualsmartcard is somehow interfering with the OsEID emulator. I will try to check what can be done about that later.

popovec:
I tested  the `Jakuje:unbreak-derive` 30652d4  - ECDH operation is now fixed (MyEID).

Jakuje:
I think the CI build was failing because of #1685 (I rerun it to check if it will work now when it is merged).

Jakuje:
Yes, the current version passes the new target for oseid testing. Is there anything more that needs to be changed here or can this be merged?

frankmorgner:
using both simulators at the same time should not lead to conflicts. However, we could leave a cleanup for the future.

Jakuje:
@frankmorgner I did not investigate the problems very deeply, but I think something was hitting the limit of slots or readers (either of them was presenting too much of empty virtual readers or something like that).

Additionally, when I added the oseid to the same target, the output was longer than travis was able to handle and I had to list through the end of the log in separate window without the decorations.

I see advantage of having the separate emulators in separate builds so one can see on the first sight what got broken.

popovec:
Some lines from failed  travis simulation (https://api.travis-ci.org/v3/job/536406881/log.txt):

_++java -noverify -cp IsoApplet/src/:jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar com.licel.jcardsim.remote.VSmartCard isoapplet_jcardsim.cfg
Error: Could not find or load main class com.licel.jcardsim.remote.VSmartCard
++opensc-tool --card-driver default --send-apdu 80b800001a0cf276a288bcfba69d34f310010cf276a288bcfba69d34f3100100_

This is not an opensc (opensc-tool -a) bug. But it is not clear, why java is unable to load main class ...

frankmorgner:
Following your argument, we should have separate builds for each Java card applet. Don't get me wrong, I think this would be a great idea in terms of spotting the actual problem. However, this consumes extra build time, because doesn't allow to run tests separately from the build script. I think we should not overstrain Travis' free cpu time as our contribution to stop climate change

Jakuje:
Indeed, if they would start doing something more useful then just trying few APDU, it would make sense to separate them. But for now, it really does not make sense since it is good that they share the time needed to build the java dependencies. But the builds are parallel so splitting them might in the end even reduce the build time as the response to the commit, even though the actual "accounted" time would be higher.

I agree that it would be useful to share the artifacts from the build stage to the simulation stage and in simulation of different java applets reuse the java artifacts. If I read right, this should be possible to achieve https://github.com/travis-ci/travis-ci/issues/7590 if somebody is willing to experiment with the travis.

etc/opensc.conf.oseid
```diff
@@ -0,0 +1,14 @@
...
+                atrmask = "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff";
+                driver = "myeid";
+        }
+        # new ATR for OsEID T0, T1 protocol
+        card_atr 3b:f5:00:00:02:80:01:4f:73:45:49:44:00 {
+                #force_protocol = t0;
+                atrmask = "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff:00"
+                driver = "myeid";
+        }
+}
```
frankmorgner: Does the standard configuration not work?

Jakuje: This was part of the instructions I got from @popovec. It does not work without that:

https://travis-ci.org/Jakuje/OpenSC/builds/542173526

It would be better to have this part in the driver, rather than in separate configuration files. The MyEID card detection checks for this string in the ATR historical bytes so it fails for the OsEID historical bytest. But it should be pretty trivial to modify the myeid driver, then it works:

https://travis-ci.org/Jakuje/OpenSC/builds/542176862

If you agree, I will add the commits from there to this PR and squash them.

popovec: Somebody from the Aventra should comment on this.

Jakuje: For the record, there was a comment in [1] from @hhonkanen approving the changes with a suggestion to log what card was matched and set the card name to the OsEID. This is what I did in 
https://github.com/Jakuje/OpenSC/commit/79f63827bc74cf5d3d064a1e585d62200c28e213
I will add the changes also to this PR and then we are ready for merge.

[1] https://github.com/Jakuje/OpenSC/commit/94adc80037bad65b3990c11c6289c53c619b43f7#commitcomment-33906504

frankmorgner:
thanks!

PR #1686, opened 2019-05-25
gemsafeV1: remove redundant match card call to allow for opensc.conf match

nunojpg:
At the point when gemsafe_match_card is called, the card type is already known,
either because of a previous match at card.c, or because it is forced at opensc.conf.

With this redundant match it's not possible to force selection on opensc.conf.


nunojpg:
@l1k

PR #1711, opened 2019-06-18
sc-hsm-tool: Add options for public key authentication

frankbraun:
This PR extends the `sc-hsm-tool` with options to use Public Key Authentication ([n-of-m Authentication](https://www.smartcard-hsm.com/docs/SmartCard-HSM_n-of-m_Authentication_V1.0_2015-03-25.pdf)).

The new options allow the `sc-hsm-tool` to:

- Initialize a HSM for public key authentication (`--public-key-auth` and `--required-pub-keys`).
- To export a public key to be used for public key authentication (`--export-for-pub-key-auth`).
- To register such an exported public key with an HSM initialized for public key authentication (`--register--public-key`).
- Show the status of public key authentication (`public-key-auth-status`).

What's still missing is an option to do the actual authentication.

The SmartCard-HSM version 2.6 and SmartCard-HSM version 3.1 were used during testing.

#### Open issue

The CAR and CHR of CV Certificates generated with OpenSC are fixed values, see:

https://github.com/OpenSC/OpenSC/blob/19711d0a1301f0519a1348390bb996fd2925f766/src/pkcs15init/pkcs15-sc-hsm.c#L265-L266

This prevents registering multiple keys (created with OpenSC) for public key authentication due to name clashes.

This issue has been discussed with Andreas Schwier from @CardContact, but no solution has been found yet.

##### Checklist


frankmorgner:
CAR/CHR are tied to a limit of 7-16 bytes by BSI TR-03110.

CardContact:
Ideally, one could specify the CHR and inner CAR used in the certificate signing request signed by the SmartCard-HSM when generating a new key pair.

In our own PKCS#11 module we do this with proprietary CKA_ attributes that carry information like CHR, CAR, algorithm list, key use counter etc. Unfortunately OpenSC and the pkcs11-tool does not have the capability to specify custom attributes in the keygen template.

In Public Key Authentication (PKA), the public key is imported and validated in the CSR format and the CHR is used as unique key identifier. With a static CHR this doesn't work with more that one key.

Patch #1714 replaces the static CHR with an id based on the device serial number (which in fact is the CHR of the device certificate without serial 00000).  As a result, the CSR would contain a CHR that is unique per card, which should be sufficient for PKA.

src/tools/sc-hsm-tool.c
```diff
@@ -1667,6 +1693,298 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
...
+	u8 tag = SC_ASN1_TAG_CONSTRUCTED | SC_ASN1_TAG_SEQUENCE; /* 0x30 */
+	unsigned int cla_out, tag_out;
+	const u8 *buf;
+	const u8 *pk;
+	const u8 *devcert;
+	const u8 *dicacert;
+	size_t taglen, pk_len, devcert_len, dicacert_len;
+	int r;
+
+	if (!(in = fopen(inf, "rb"))) {
```
frankmorgner: you may want to use `fread_to_eof()`

src/tools/sc-hsm-tool.c
```diff
@@ -1667,6 +1693,298 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
...
+		fprintf(stderr, "Error reading ASN.1 sequence: %s\n", sc_strerror(r));
+		free(pka);
+		return -1;
+	}
+	devcert = buf;
+	devcert_len = taglen;
+	buf += taglen;
+
+	/* read device CA */
+	if ((r = sc_asn1_read_tag(&buf, sb.st_size, &cla_out, &tag_out, &taglen)) < 0) {
```
frankmorgner: check tag_out?

src/tools/sc-hsm-tool.c
```diff
@@ -1667,6 +1693,298 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
...
+	/* read ASN.1 sequence */
+	buf = pka;
+	if ((r = sc_asn1_read_tag(&buf, sb.st_size, &cla_out, &tag_out, &taglen)) < 0) {
+		fprintf(stderr, "Error reading ASN.1 sequence: %s\n", sc_strerror(r));
+		free(pka);
+		return -1;
+	}
+
+	/* read public key */
+	if ((r = sc_asn1_read_tag(&buf, sb.st_size, &cla_out, &tag_out, &taglen)) < 0) {
```
frankmorgner: check tag_out?

frankbraun:
Thanks for the feedback @frankmorgner, I'll address the remarks.

src/libopensc/card-sc-hsm.c
```diff
@@ -1279,9 +1287,7 @@ static int sc_hsm_initialize(sc_card_t *card, sc_cardctl_sc_hsm_init_param_t *pa
 		*p++ = params->bio2.len;
 		memcpy(p, params->bio2.value, params->bio2.len);
 		p += params->bio2.len;
-	}
-
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x50, 0x00, 0x00);
+  } sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x50, 0x00, 0x00);
```
Jakuje: Wrong indentation of closing bracket and the `sc_format_apdu()` should start on the new line. Generally, this diff does not change anything useful.

src/tools/sc-hsm-tool.c
frankbraun: Done, but it required build system changes. Would be great if you could review them.

src/libopensc/card-sc-hsm.c
frankbraun: Sorry, this slipped in by accident.

src/libopensc/card-sc-hsm.c
```diff
@@ -1424,10 +1432,215 @@ static int sc_hsm_unwrap_key(sc_card_t *card, sc_cardctl_sc_hsm_wrapped_key_t *p
...
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	/* verify dicacert */
+	r = verify_certificate(card, params->dicacert, params->dicacert_length, params->dicacert_chr, params->dicacert_chr_length);
+	LOG_TEST_RET(ctx, r, "device issuer certificate verification failed");
+
+	/* verify devcert */
+	r = verify_certificate(card, params->devcert, params->devcert_length, params->devcert_chr, params->devcert_chr_length);
+	LOG_TEST_RET(ctx, r, "device issuer certificate verification failed");
```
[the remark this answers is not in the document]
frankbraun: Yes, "device issuer verification failed" is the correct error message. Changed.

src/tools/sc-hsm-tool.c
frankbraun: I think I rather stick with the current code, I still need the size of the read buffer and using `fread_to_eof()` would also require build system changes.

src/tools/sc-hsm-tool.c
```diff
@@ -1667,6 +1693,298 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
...
+	ef_cert_len = sc_read_binary(card, 0, ef_cert, sizeof(ef_cert), 0);
+	if (ef_cert_len < 0) {
+		fprintf(stderr, "Error reading certificate %s. Skipping\n", sc_strerror(ef_cert_len));
+		ef_cert_len = 0;
+	} else {
+		ef_cert_len = determineLength(ef_cert, ef_cert_len);
+	}
+
+	/* C_DevAut */
+	fid[0] = 0x2F;
```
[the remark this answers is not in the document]
frankbraun: Doesn't seem to contain the same. Is it just the device certificate? I need the concatenation of device certificate and device issuer certificate.

src/libopensc/card-sc-hsm.c
```diff
@@ -1424,10 +1432,215 @@ static int sc_hsm_unwrap_key(sc_card_t *card, sc_cardctl_sc_hsm_wrapped_key_t *p
...
+
+    /* find embedded Certification Authority Reference (CAR) */
+	if (!(car = sc_asn1_find_tag(ctx, buf, buflen, 0x42, &taglen))) {
+		LOG_FUNC_RETURN(ctx, SC_ERROR_UNKNOWN);
+	}
+
+    /* return CAR */
+    strncpy((char*) carstr, (const char*) car, taglen);
+	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+}
```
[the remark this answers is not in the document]
frankbraun: I tried to parse this with `sc_pkcs15emu_sc_hsm_decode_cvc()`, but wasn't able to. I got the following error: 
```
Required ASN.1 object not found
```

frankbraun:
I tried to address all remarks. Is there anything else that needs to be changed? Should I squash commits to cleanup the commit history?

src/libopensc/card-sc-hsm.c
```diff
@@ -1424,10 +1432,190 @@ static int sc_hsm_unwrap_key(sc_card_t *card, sc_cardctl_sc_hsm_wrapped_key_t *p
...
+	buf = cb;
+	buflen = taglen;
+
+	/* find embedded Certification Authority Reference (CAR) */
+	if (!(car = sc_asn1_find_tag(ctx, buf, buflen, 0x42, &taglen))) {
+		LOG_FUNC_RETURN(ctx, SC_ERROR_UNKNOWN);
+	}
+
+	/* return CAR */
+	memcpy(carstr, car, taglen);
```
frankmorgner: possible out of bounds write. you need to check taglen against the length of your target buffer.

src/tools/sc-hsm-tool.c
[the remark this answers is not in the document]
frankmorgner: That's EF_C_DevAut, see
https://github.com/OpenSC/OpenSC/blob/93bdc8c826205637cf1dcafec24dc4f3615e0003/src/libopensc/card-sc-hsm.c#L509-L552

src/tools/sc-hsm-tool.c
```diff
@@ -1667,6 +1693,322 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
...
+		fclose(outfp);
+	if (out)
+		free(out);
+	if (data)
+		free(data);
+
+	return r;
+}
+
+static void get_CHR(char *chrstr, int is_cvc, sc_context_t *ctx, const u8 *buf, size_t buflen)
```
frankmorgner: please use sc_pkcs15emu_sc_hsm_decode_cvc for parsing

src/tools/sc-hsm-tool.c
frankmorgner:
```
int fread_to_eof(const char *file, unsigned char **buf, size_t *buflen);
```
`buflen` will contain the length of the allocated buffer, i.e. the read file.

The "build system changes" are as simple as adding *fread_to_eof.c* to sc-hsm-tool.

On the other hand, using `fread_to_eof()` would have saved me reviewing 50 lines of code and will save us maintaining them in the future.

frankbraun:
> The code looks OK, but you really need to use `sc_pkcs15emu_sc_hsm_decode_cvc()`!

I tried to refactor the code in order to use `sc_pkcs15emu_sc_hsm_decode_cvc()`.
I can successfully use it to parse the public key CVC, which has the following structure:

```
class=1, tag=7, compound=true, len=493, fullLen=497, sha1=9874645e2b9257125c1c7f8d58cc2fc08a2c9ca0
CVC parse
  tag=CV Certificate, cl=1, cmpnd=true, len=403
    tag=Certificate Body, cl=1, cmpnd=true, len=331
      tag=Certificate Profile Identifier, cl=1, cmpnd=false, len=1
      tag=Certification Authority Reference, cl=1, cmpnd=false, len=16
      CAR=DENK010141000000
      tag=Public Key, cl=1, cmpnd=true, len=285
        tag=6, cl=0, cmpnd=false, len=10
        tag=1, cl=2, cmpnd=false, len=32
        tag=2, cl=2, cmpnd=false, len=32
        tag=3, cl=2, cmpnd=false, len=32
        tag=4, cl=2, cmpnd=false, len=65
        tag=5, cl=2, cmpnd=false, len=32
        tag=6, cl=2, cmpnd=false, len=65
        tag=7, cl=2, cmpnd=false, len=1
      tag=Certificate Holder Reference, cl=1, cmpnd=false, len=16
      CHR=DENK010141000000
    tag=Signature, cl=1, cmpnd=false, len=64
  tag=Certification Authority Reference, cl=1, cmpnd=false, len=16
  CAR=DENK010141000000
  tag=Signature, cl=1, cmpnd=false, len=64
CVC end
```

However, the device certificate and the device issuer certificate have structures like this:

```
class=1, tag=33, compound=true, len=228, fullLen=232, sha1=4bcfcfe3784903f5a6de0d19bc9023484fa86e91
CVC parse
  tag=Certificate Body, cl=1, cmpnd=true, len=157
    tag=Certificate Profile Identifier, cl=1, cmpnd=false, len=1
    tag=Certification Authority Reference, cl=1, cmpnd=false, len=13
    CAR=DEDINK0100001
    tag=Public Key, cl=1, cmpnd=true, len=79
      tag=6, cl=0, cmpnd=false, len=10
      tag=6, cl=2, cmpnd=false, len=65
    tag=Certificate Holder Reference, cl=1, cmpnd=false, len=16
    CHR=DENK010141000000
    tag=Certificate Holder Authorization Template, cl=1, cmpnd=true, len=16
      tag=6, cl=0, cmpnd=false, len=11
      tag=Discretionary Data, cl=1, cmpnd=false, len=1
    tag=Certificate Effective Date, cl=1, cmpnd=false, len=6
    tag=Certificate Expiration Date, cl=1, cmpnd=false, len=6
  tag=Signature, cl=1, cmpnd=false, len=64
CVC end
```

which I cannot parse simply with `sc_pkcs15emu_sc_hsm_decode_cvc()`, I get the following error:

```
sc_card_ctl(*, SC_CARDCTL_SC_HSM_REGISTER_PUBLIC_KEY, *) failed with Required ASN.1 object not found
```

I could try wrapping them in a CV Certificate and add a fake outer CAR and Signature in order to also use `sc_pkcs15emu_sc_hsm_decode_cvc()` for them, but that seems a bit convoluted.

What should I do in that regard in order to get the PR into a mergeable state?

src/libopensc/card-sc-hsm.c
frankbraun: Fixed, thanks.

src/tools/sc-hsm-tool.c
frankbraun: Regarding EF_C_DevAut I can see that it is set in the static function `sc_hsm_perform_chip_authentication()`. `sc_hsm_perform_chip_authentication()` is only called from `sc_hsm_pin_cmd()`. That is, filling EF_C_DevAut currently requires PIN entry, which is not necessary for my implementation.

What would be a good way to use EF_C_DevAut, ideally without PIN entry?

frankbraun: Regarding `sc_pkcs15emu_sc_hsm_decode_cvc()`: I spent many hours trying to get it to work.
Like I wrote in https://github.com/OpenSC/OpenSC/pull/1711#issuecomment-546378453, I wasn't able to parse the device certificate and the device issuer certificate with it. For the public key, `sc_cvc_t.car` is filled correctly, but `sc_cvc_t.outer_car` contains garbage. Unfortunately, I couldn't figure out what the problem is.

frankmorgner:
Sorry for the late response.

OpenPACE has both, [`CVC_d2i_CVC_CERT`](https://github.com/frankmorgner/openpace/blob/master/src/eac/cv_cert.h#L330)for parsing a complete certificate and [`d2i_CVC_CERT_BODY`](https://github.com/frankmorgner/openpace/blob/master/src/eac/cv_cert.h#L184) to parse the body without the actual signature. That would be one solution.

The other one, would be to extend `pkcs15-sc-hsm.c` to add a function like `sc_pkcs15emu_sc_hsm_decode_cvc_body()`. This new function can be bootstrapped from `sc_pkcs15emu_sc_hsm_decode_cvc()` to only parse the certificate's body. Although this adds some duplication, I think it's better to make the CVC (body) parsing explicit instead of parsing some tags by hand.

src/tools/sc-hsm-tool.c
frankmorgner: > What would be a good way to use EF_C_DevAut, ideally without PIN entry?

The basic idea is in sc-hsm-tool to
1. force a sc-hsm card driver, as done [here](https://github.com/OpenSC/OpenSC/pull/1896/files#diff-1252f4f12c940c3ec84eebedcd816a8fR600-R604) and
2. then cast card->drv_data to `struct sc_hsm_private_data *` and use its `EF_C_DevAut` if it's available.

Additionally, you may want to transfer ownership of a newly read EF_C_DevAut to the driver for re-use. You don't need to perform CA to do this.

frankmorgner:
@frankbraun , you've put a lot of effort into this and I think it would be useful. However, there has been some inactivity and now there are merge conflicts, would you card to look at this again? Do you need some help regarding the comments above?

nickaknudson:
I'm interested in finishing this pull request. Aside from merge conflicts can someone help outline what work remains to be done? Was it just the two comments from Jan 17 2020?

frankmorgner:
> I'm interested in finishing this pull request. Aside from merge conflicts can someone help outline what work remains to be done? Was it just the two comments from Jan 17 2020?

Yes, indeed. The main problem is maintainability. You should use the CVC capabilities we already have. If I remember correctly, the rest of the code looked OK.

charredlot:
hi, also interested in this functionality. i rebased this branch and tweaked it to share code with `sc_pkcs15emu_sc_hsm_decode_cvc` for the CVC parsing here: https://github.com/charredlot/OpenSC/tree/pka

@nickaknudson looks like we ran into this at the same time, so if you have something ready, i'd be happy to defer. otherwise it'd be convenient to have this in upstream soon, so I'll make a new PR if there's no objections?

charredlot:
created https://github.com/OpenSC/OpenSC/pull/2301 in the interest of moving this forward

frankmorgner:
closed with https://github.com/OpenSC/OpenSC/pull/2301

PR #1714, opened 2019-06-19
sc-hsm: Use CHR in CSR based on device serial number

CardContact:
This patch replaces the static CHR in a certificate signing request generated internally in the SmartCard-HSM with a value derived from the device serial number.

This patch solves the issue mentioned in #1711.


frankbraun:
Thanks for the patch, seems to be a good enough fix for me, given that the proper solution isn't really possible in OpenSC.

PR #1717, opened 2019-06-20
Update pkcs15-sec.c

programatix:
When card supports SC_ALGORITHM_RSA_PAD_PKCS1 but not SC_ALGORITHM_RSA_HASH_NONE, then the DigestInfo need to be removed.

Current check requires the card to not support both SC_ALGORITHM_RSA_PAD_PKCS1 and SC_ALGORITHM_RSA_HASH_NONE to have the removal done.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -652,7 +652,8 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	 * block from PKCS1 padding. */
 	if ((flags == (SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE)) &&
 	    !(alg_info->flags & SC_ALGORITHM_RSA_RAW) &&
-	    !(alg_info->flags & (SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE))) {
+	    !(alg_info->flags & SC_ALGORITHM_RSA_HASH_NONE) &&
+	    (alg_info->flags & SC_ALGORITHM_RSA_PAD_PKCS1)) {
 		unsigned int algo;
 		size_t tmplen = sizeof(buf);
 


frankmorgner:
what's your configuration that lets you run into an error without this?

programatix:
The card is configured to support,
- ```SC_ALGORITHM_RSA_PAD_PKCS1```
- ```SC_ALGORITHM_RSA_HASH_SHA1```
- ```SC_ALGORITHM_RSA_HASH_SHA224```
- ```SC_ALGORITHM_RSA_HASH_SHA256```
- ```SC_ALGORITHM_RSA_HASH_SHA384```
- ```SC_ALGORITHM_RSA_HASH_SHA512```

The client sent data with [DigestInfo] + [hash data] with flag ```SC_ALGORITHM_RSA_PAD_PKCS1``` | ```SC_ALGORITHM_RSA_HASH_NONE```.

```sc_pkcs15_compute_signature``` proceeds to pass both the data unchanged and ```senv.algorithm_flags``` with ```SC_ALGORITHM_RSA_PAD_PKCS1``` | ```SC_ALGORITHM_RSA_HASH_NONE``` to the card driver.

It supposed to remove the DigestInfo and get the hash algo from it and update the ```senv.algorithm_flags``` with the correct hash flag before passing them to the card driver.

frankmorgner:
I'm not completely sure about the flags. e5707b545e5a2dc33b0ca52a8bf63f36f71b3d85 *adds* digest stripping when `!(alg_info->flags & SC_ALGORITHM_RSA_RAW)`, but doesn't change any other condition. Your wording seems to imply that your use case used to work at some point, but I don't know what your baseline (working commit) is. Did you test your code?

programatix:
I'm currently writing a new card driver. The card supports the algorithm configuration stated above. When I tested using PuTTY-CAC (http://risacher.org/putty-cac/) via opensc-pkcs11.dll, ```sc_pkcs15_compute_signature``` is called with the data in ```in``` containing the hash's DigestInfo + hash. The '''flags''' is 0x102 which means ```SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_PAD_PKCS1```.

As mentioned, the card capability does not include HASH_NONE. Taking a look at the comment in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/opensc.h line 126 to 128, it states that HASH_NONE + PAD_PKCS1 is valid and means that the incoming data is DigestInfo + hash.
>  For PKCS1 (v1.5) the hash is recorded in the padding, and HASH_NONE is a
>  valid value, meaning that the hash's DigestInfo has already been
>  prepended to the data, otherwise the hash id is put on the front.

Since the card capability does not include HASH_NONE, the DigestInfo should be removed and the correct hash algorithm should be obtained from the DigestInfo and set it in ```senv.algorithm_flags```. This means that, in this condition, ```sc_pkcs1_strip_digest_info_prefix``` must be called.

The current condition ```!(alg_info->flags & (SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE))``` is rather incorrect. It means, execute if card does not support ```SC_ALGORITHM_RSA_PAD_PKCS1``` **or** ```SC_ALGORITHM_RSA_HASH_NONE```. In my driver case, it would never be executed because the card supports ```SC_ALGORITHM_RSA_PAD_PKCS1```.

It should be execute if card supports ```SC_ALGORITHM_RSA_PAD_PKCS1``` but not ```SC_ALGORITHM_RSA_HASH_NONE```, which is inline with the comment referred above.

dengert:
There may be some confusion in the meaning of `SC_ALGORITHM_RSA_HASH_NONE`  Lines 117-126 of `opensc.c` appears to say the card can accept a hash, then add the digestInfo before padding and crypto.

But other  parts of the code expect `SC_ALGORITHM_RSA_HASH_NONE` to mean the absence of any hash.  Prior to e5707b545 tests were done using `SC_ALGORITHM_RSA_HASHES` to see if any of the `SC_ALGORITHM_RSA_HASH_*` were set.

Note that e5707b545 modified `#define SC_ALGORITHM_RSA_HASHES          0x0001FF00` which implies  `SC_ALGORITHM_RSA_HASH_NONE` is a hash (or the absence of any other hash), not that it is a separate flag having to do with adding a DigestInfo.  

We have 41 card drivers and `SC_ALGORITHM_RSA_HASH_NONE` is set in about 18 of these. I believe these cards are using `SC_ALGORITHM_RSA_HASH_NONE`  to mean no hash is set. 

We should look at changing `SC_ALGORITHM_RSA_HASH_NONE` to something like `SC_ALGORITHM_RSA_HASH_CARD_CAN_ADD_DIGESTINFO`, remove it from `SC_ALGORITHM_RSA_HASHES` and go back to using  `(flags && SC_ALGORITHM_RSA_HASHES == 0)` to mean no hash is defined.

programatix:
Any suggestion on how to resolve this on the driver level? The issue is, my driver needs to know the hash algo when ```sc_pkcs15_compute_signature``` calls ```xxx_set_security_env```. I can call ```sc_pkcs1_strip_digest_info_prefix``` there but it has no access to the incoming data, thus making it impossible. It's too late to do so when it calls ```xxx_compute_signature```.

Again, in case the original post (and the posts above) is not clear, my card supports PKCS1 padding and SHA1 to SHA-512 hash. The hash algo is sent to card during ```xxx_set_security_env``` It does not support digest info OID in the incoming data of compute signature command. Thus, for client which calls opensc with hash algo ```SC_ALGORITHM_RSA_HASH_NONE``` and digest info in the data, the digest info need to be stripped and the algo need to be updated based on the stripped hash algo.

dengert:
One way to do this. Have xxx_set_security_env save what it needs then
postpone the actual set_security_env.  Your xxx_compute_signature would
then look at data to be signed which has the digest info i.e. the OID of
the hash, strip it off and then do the set security env based on the hash
then do sign operation.

PR #1720, opened 2019-06-26
treewide: Fix compilation without deprecated OpenSSL APIs

neheb:


dengert:
Have a look at ` sc-ossl-compat.h`. It has all the backward compatibility routines and defines needed for older (or deprecated) OpenSSL APIs  so the base OpenSC code uses the most upto date OpenSSL API. 

Rather then adding all the \#ifdefs in each file for `EVP_CIPHER_CTX_cleanup` vs `EVP_CIPHER_CTX_reset` change all the `EVP_CIPHER_CTX_cleanup` to `EVP_CIPHER_CTX_reset` and add the \#ifdefs to `sc-ossl-compat.h`  and add  `#define EVP_CIPHER_CTX_reset(x) EVP_CIPHER_CTX_cleanup(x)` much like `sc-ossl-compat.h` does for many other routines. 


Why are card drivers and `pkcs15-prkey.c` calling `ERR_load_ERR_strings();` and `ERR_load_crypto_strings();`?  Should this be done in the application or when `libopensc.so`  or  `opensc-pkcs11.so` is dynamically loaded?

neheb:
I did so for card-piv.c

The issue with _reset is that it is defined to both _init and _cleanup.

dengert:
I would like to suggest that the attached diff could replace this PR. [OPENSSL_API_COMPAT.opensc.diff.txt](https://github.com/OpenSC/OpenSC/files/3340425/OPENSSL_API_COMPAT.opensc.diff.txt)

This was developed and partially tested with OpenSSL-1.1.0c compiled from git source. OpenSC was then compiled with `-DOPENSSL_API_COMPAT=0x10100000L` added to the OpenSC  CPPFLAGS. 

The previous function calls are left in the source code, and if compiled against 1.1.0 or later the 1.1.0 routines are used.  You said: "The issue with _reset is that it is defined to both _init and _cleanup." This should also be addressed by the diff. In 1.1.0 _reset does both _init and _cleanup.

neheb:
Works for me.

dengert:
Can you take the diff file and submit it as a PR? You are the one interested in compiling OpenSC without using any deprecated functions. This is not a problem today. 

 And to do this required adding `-DOPENSSL_API_COMPAT=0x10100000L` to build for OpenSC to force the issue. I might expect in some future version of OpenSSL they will add more deprecated functions (or remove them) so this may be an ongoing process. 

Note that the OpenSC  `internal.h` and most if not all the tools have `#include "libopensc/sc-ossl-compat.h"` so the include was not explicitly needed in card-piv.c. But was need in `pkcs15init/pkcs15-oberthur-awp.c`. 

Do you have a way to test these changes? I only have the PIV card to test and the change is only used with the `piv-tool` with the mutual authentication of an admin 3DES or AES key.  In other words this is not an easy modification to test.

neheb:
Force pushed.

Jakuje:
This is breaking building with OpenSSL 1.1.1 if I am right, since the openssl still defines these functions even though they are deprecated:
```
In file included from pkcs15-lib.c:59:
../../src/libopensc/sc-ossl-compat.h:51: error: "ERR_load_crypto_strings" redefined [-Werror]
   51 | #define ERR_load_crypto_strings(x) {}
      | 
In file included from pkcs15-lib.c:53:
/usr/include/openssl/err.h:255: note: this is the location of the previous definition
  255 | # define ERR_load_crypto_strings() \
      | 
```

neheb:
This is the main reason I didn't want to do this. It can probably be worked around by passing -Wno-error=macro-redefined

Or #undef before defining.

mouse07410:
> ... I didn't want to do this...

To do - what?

> ... can be worked around ...

Of course it can be. My point is - the simplest and IMHO the most useful/correct solution is to just revert that PR, as IMHO it serves no useful purpose and introduces more problems than it is trying to solve.

dengert:
Its more complicated then just doing a revert as suggested above and in https://github.com/OpenSC/OpenSC/issues/1755 It depends on how OpenSSL was built and how the calling application is built.  OpenSC  should be able to work with any distro or user build of  OpenSSL.  

Despite OpenSSL warnings about using deprecated routines, they did include some support to help with any transition, as they saw the same problems this PR was designed to address. 

I believe @neheb may have been using OpenSSL built with `no-deprecated` and @mouse07410 is using OpenSSL built without `no-deprecated`.  Can you both look at the `opensslconf.h` you have to see what is the definition of `OPENSSL_MIN_API`. 

Building 1.1.1c using `Configure ... no-deprecated ...` will cause the deprecated routines not be included.   `no-deprecated` caused OpenSSL to be compiled with `-DOPENSSL_API_COMPAT=0x10100000L` (Based one of 3 releases for 1.1.1c). For explaination see the `Configure` and `INSTALL` in the OpenSSL source and read up on `no-deprecated` and  `--api=x.y.z` 

A diff of building with and without `no-deprecated` shows:
```
diff -u /opt/ossl-1.1/include/openssl/opensslconf.h ./opensslconf.h
--- /opt/ossl-1.1/include/openssl/opensslconf.h	2019-06-28 07:00:50.570332815 -0500
+++ ./opensslconf.h	2019-07-28 14:23:04.907969863 -0500
@@ -24,6 +24,7 @@
  * OpenSSL was configured with the following options:
  */
 
+#define OPENSSL_MIN_API 0x10100000L
 #ifndef OPENSSL_NO_MD2
 # define OPENSSL_NO_MD2
 #endif
@@ -45,6 +46,9 @@
 #ifndef OPENSSL_NO_CRYPTO_MDEBUG_BACKTRACE
 # define OPENSSL_NO_CRYPTO_MDEBUG_BACKTRACE
 #endif
+#ifndef OPENSSL_NO_DEPRECATED
+# define OPENSSL_NO_DEPRECATED
+#endif
 #ifndef OPENSSL_NO_DEVCRYPTOENG
 # define OPENSSL_NO_DEVCRYPTOENG
 #endif
Only in /opt/ossl-1.1/include/openssl: sm2err.h
Only in /opt/ossl-1.1/include/openssl: sm2.h
```
Then when an applications is being compiled `./opensslconf.h` does:
```
#ifndef OPENSSL_MIN_API
# define OPENSSL_MIN_API 0
#endif

#if !defined(OPENSSL_API_COMPAT) || OPENSSL_API_COMPAT < OPENSSL_MIN_API
# undef OPENSSL_API_COMPAT
# define OPENSSL_API_COMPAT OPENSSL_MIN_API
#endif
```


If I am correct we should be adding defines in this PR only if they are not native functions of the release (for example 1.0.2) or are not included as deprecated or defined by OpenSSL.

Awaiting your input of what is the value of `OPENSSL_MIN_API`.

neheb:
```#define OPENSSL_MIN_API 0x10100000L```

**Comment by mouse07410** [comment]
> OpenSC should be able to work with any distro or user build of OpenSSL

I'm afraid this is too tall an order. First, as I said above, I don't think more than three main branches of OpenSSL need to be supported, and those are 1.0.2, 1.1.1, and 3. Second - I'm pretty sure you can create an OpenSSL build that OpenSC won't compile with, no matter how tricky we get with `ossl_compat.h`. Fortunately, it only matters for somebody building OpenSC from source - in which case that somebody can simply force OpenSC to download and compile its own OpenSSL copy, and statically link with it (that's what I'm doing at the office, BTW).

Relevant part of `/opt/local/include/openssl/opensslconf.h`:
```
#ifndef OPENSSL_MIN_API
# define OPENSSL_MIN_API 0
#endif

#if !defined(OPENSSL_API_COMPAT) || OPENSSL_API_COMPAT < OPENSSL_MIN_API
# undef OPENSSL_API_COMPAT
# define OPENSSL_API_COMPAT OPENSSL_MIN_API
#endif

/*
 * Do not deprecate things to be deprecated in version 1.2.0 before the
 * OpenSSL version number matches.
 */
#if OPENSSL_VERSION_NUMBER < 0x10200000L
# define DEPRECATEDIN_1_2_0(f)   f;
#elif OPENSSL_API_COMPAT < 0x10200000L
# define DEPRECATEDIN_1_2_0(f)   DECLARE_DEPRECATED(f)
#else
# define DEPRECATEDIN_1_2_0(f)
#endif

#if OPENSSL_API_COMPAT < 0x10100000L
# define DEPRECATEDIN_1_1_0(f)   DECLARE_DEPRECATED(f)
#else
# define DEPRECATEDIN_1_1_0(f)
#endif

#if OPENSSL_API_COMPAT < 0x10000000L
# define DEPRECATEDIN_1_0_0(f)   DECLARE_DEPRECATED(f)
#else
# define DEPRECATEDIN_1_0_0(f)
#endif
```

> If I am correct we should be adding defines in this PR only if they are not native functions of the release (for example 1.0.2) or are not included as deprecated or defined by OpenSSL

Yes, IMHO you're absolutely correct. 

> [neheb] Or `#undef` before defining.

Considering that an OpenSSL version may define these to something meaningful, bluntly/blindly `undef`ing them and re-defining sounds like a **very bad** idea.

PR #1721, opened 2019-06-28
card-rtecp: Fix list_files on T0 cards

dfortinskiy:
Rutoken ECP SC over T0 expects Get Response after SW1=61 which
is not called with zero le.

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/card-rtecp.c
```diff
@@ -577,8 +577,7 @@ static int rtecp_list_files(sc_card_t *card, u8 *buf, size_t buflen)
 	for (;;)
 	{
 		apdu.resp = rbuf;
-		apdu.resplen = sizeof(rbuf);
-		apdu.le = 256;
+		apdu.resplen = apdu.le = sizeof(rbuf);
```
frankmorgner: please use two lines.

dfortinskiy: fixed.

PR #1722, opened 2019-06-28
card-rtecp: Add Rutoken Lite

dfortinskiy:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-rtecp.c
+++ b/src/libopensc/card-rtecp.c
@@ -1,5 +1,5 @@
 /*
- * card-rtecp.c: Support for Rutoken ECP cards
+ * card-rtecp.c: Support for Rutoken ECP and Rutoken Lite cards
  *
  * Copyright (C) 2009  Aleksey Samsonov <samsonov@guardant.ru>
  *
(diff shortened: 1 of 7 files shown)


dfortinskiy:
> `decipher()` and `compute_signature()` are not implemented, so from a user's perspective there's not a lot to do with this card. Adding this card to OpenSC only makes sense if one of these would be implemented.

Rutoken Lite doesn't support any cryptographic operations (incl. `decipher()` and `compute_signature()`), so they are cut. But there are some people who use such cards for operations with data objects only with PIN protection. By happy occasion they also use OpenSC, so we have thought it could be nice to add such card.

If you really consider it should not be in upstream, than it's OK. 

> don't use `SC_CARD_TYPE_GENERIC_BASE`, instead use a card driver specific define. (`card-rtecp.c` already contains this problem for the other cards, I know).

OK.

dfortinskiy:
> don't use `SC_CARD_TYPE_GENERIC_BASE`, instead use a card driver specific define. (`card-rtecp.c` already contains this problem for the other cards, I know).

In https://github.com/OpenSC/OpenSC/pull/1721.

frankmorgner:
Again, I think it's a bit overkill to create a new card driver. Please check out https://github.com/OpenSC/OpenSC/pull/1728 if it works with your card.

dfortinskiy:
> Again, I think it's a bit overkill to create a new card driver. 

Sure.
> Please check out #1728 if it works with your card.

This one is even better!

Close this MR?

PR #1726, opened 2019-07-04
Update strings.c

programatix:
The check condition is obviously wrong. It should check for EQUAL. The original bitwise check caused any other language to turn into DE because as long as a bit is filtered, it will hit.

### DIFF
--- a/src/ui/strings.c
+++ b/src/ui/strings.c
@@ -157,7 +157,7 @@ const char *ui_get_str(struct sc_context *ctx, struct sc_atr *atr,
 			|| !find_lang_str(p15card->tokeninfo->preferred_language, &lang)) {
 #ifdef _WIN32
 		LANGID langid = GetUserDefaultUILanguage();
-		if (langid & LANG_GERMAN) {
+		if (langid == LANG_GERMAN) {
 			lang = DE;
 		}
 #else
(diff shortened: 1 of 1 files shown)


programatix:
LANG_GERMAN is actually a byte already. So, it would simply be,
```if ((langid & LANG_GERMAN) == LANG_GERMAN) {```
if you insist on using bitwise compare.

However, the value returned by GetUserDefaultUILanguage() is a numeric identifier and not bit flag. Also, the LANG_xxxx is also a numeric identifier, not bit flags, so I don't see why the need of using bitwise if.

frankmorgner:
https://docs.microsoft.com/de-de/windows/win32/intl/language-identifiers
https://docs.microsoft.com/de-de/windows/win32/intl/language-identifier-constants-and-strings

programatix:
If that is the case, then it should be,
```if ((langid & LANG_GERMAN & 0xFF) == LANG_GERMAN) {```
or to play safe,
```if ((langid & LANG_GERMAN & 0xFF) == (LANG_GERMAN & 0xFF)) {```

This is because the value returned by ```GetUserDefaultUILanguage()``` may include sub language id, thus the 0xFF mask should be applied to it instead.

Addtionally, in winnt.h (from Windows Kits 10.0.17763.0), ```LANG_GERMAN``` is 0x07. It's the "Prim. lang. identifier" column, not "Language identifier" of https://docs.microsoft.com/de-de/windows/win32/intl/language-identifier-constants-and-strings. Below is the screenshot of winnt.h
image

frankmorgner:
great, thanks!

programatix:
Happy to help.

PR #1732, opened 2019-07-10
pkcs15-tool: inconsistent -r option fix

### LINKED ISSUE #1704: inconsistent use of option -r in opensc tools
### Problem Description
**-r** switch is used in inconsistent way (info from manual pages, not from sources):

pkcs15-tool:
--reader arg
--read-certificate cert, -r cert

pkcs15-crypt:
--reader arg, -r arg

pkcs15-init:
--reader arg, -r arg

opensc-tool:
--reader arg, -r arg

opensc-explorer:
--reader arg, -r arg

other tools, that respect `--reader` and `-r` to specify reader:
cardos-tool, cryptoflex-tool, dnie-tool, eidenv, netkey-tool, openpgp-tool, piv-tool, sc-hsm-tool,westcos-tool 

iasecc-tool allows only long option `--reader`

### Proposed Resolution

Alternatives:
1. change pkcs15-tool, use `-r` to specify reader (same change for iasecc-tool )
2. remove `-r` from all tools 

Both alternatives may affect some user scripts, maybe new release of opensc may generate a warning if `-r` is used (i.e.  switch will be removed/changed in future release..).

popovec:
Option -r is used in other opensc tools to specify card reader.  pkcs15-tool
uses -r to specify cerfificate.  This fix intorduces warning message if -r
is used, and for future versions of pkcs15-tool -r is used to specify
reader. Fixes #1704

##### Checklist

- [x ] Documentation is added or updated

### DIFF
--- a/doc/tools/pkcs15-tool.1.xml
+++ b/doc/tools/pkcs15-tool.1.xml
@@ -209,8 +209,7 @@
 
 				<varlistentry>
 					<term>
-						<option>--read-certificate</option> <replaceable>cert</replaceable>,
-						<option>-r</option> <replaceable>cert</replaceable>
+						<option>--read-certificate</option> <replaceable>cert</replaceable>
 					</term>
 					<listitem><para>Reads the certificate with the given id.</para></listitem>
 				</varlistentry>
--- a/src/tools/pkcs15-tool.c
+++ b/src/tools/pkcs15-tool.c
@@ -107,6 +107,7 @@ enum {
 	OPT_RAW,
 	OPT_PRINT_VERSION,
 	OPT_LIST_INFO,
+	OPT_READ_CERT,
 };
 
 #define NELEMENTS(x)	(sizeof(x)/sizeof((x)[0]))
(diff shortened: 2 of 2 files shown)


martinpaljak:
I'm OK with harmonizing the usage of `-r` but reading certificates is a common operation and should have a shorthand. Would have proposed `-R` but that is already used by reading data objects. Maybe try to harmonize reading common objects with short args as well, -C (certificate) -D (data) etc ?

popovec:
I agree, short option for reading certificate is useful. But  first, let's be careful not to make any further collision.

Possible collisions (from man pages) .. 

openpgp-tool  -C  Show card information
opensc-tool    -D  List all installed card drivers.
piv-tool           -C Load a certificate onto the card
sc-hsm-tool    -C Create a DKEK share encrypted under a password and save it .....
pkcs15-init    -D Tells pkcs15-init to delete the specified object.
pkcs15-init    -C This tells pkcs15-init to create a PKCS 15 structure on the card
pkcs15-tool   -D List all card objects.
pkcs15-tool   -C List all data objects stored on the token.


I can add -r into iasecc-tool  (here only --reader is allowed for now, and -r is free for use). Is this acceptable ? 

@frankmorgner please, review this patch, is based on your suggestions, but I decided to create two version of code for 0.20.0 .. already 1st is enabled - as your suggest, hard error if -r is used and no card reader that match optarg is found.

frankmorgner:
thanks

PR #1734, opened 2019-07-10
Fix #1731 sc_format_apdu_cse_lc_le fails to set Le correctly

dengert:
This commit is not well tested and needs review.

Fixes #1731 See users comments on this patch. 

Changed four places where "<" should be "<=" so Le will be set correctly
Previous for 65K (extended) or 256 (short) Le is left set to 0.
This then caused Le to be to be not added to APDU as Le==0
Code later converts actual Le in APDU to be set to 0 to mean 256 or 65K.

SC_APDU_CASE_*_EXT are changed to SC_APDU_CASE_* so sc_detect_apdu_cse
to set the cse based on card capabilities as well as data chaining.

 On branch fix-1731
 Changes to be committed:
	modified:   src/libopensc/card.c

##### Checklist

- [X] PKCS#11 module is tested

PR #1742, opened 2019-07-15
Fix using environmental variables on WIN32.

kouril:
The comparison operator is wrong. The additional variable declaration doesn't make much sense to me.

### DIFF
--- a/src/libopensc/ctx.c
+++ b/src/libopensc/ctx.c
@@ -181,9 +181,9 @@ sc_ctx_win32_get_config_value(const char *name_env,
 		return SC_ERROR_INVALID_ARGUMENTS;
 
 	if (name_env)   {
-		char *value = value = getenv(name_env);
+		char *value = getenv(name_env);
 		if (value) {
-			if (strlen(value) < *out_len)
+			if (strlen(value) > *out_len)
 				return SC_ERROR_NOT_ENOUGH_MEMORY;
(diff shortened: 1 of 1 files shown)


kouril:
Could someone here please take a look at the checks to see why the CI jobs are failing? The checks are not passing Linux builds, while the change I made is in a WIN32-defined part of the code, which isn't getting built there. Also, after I tried and re-opened the PR I'm getting two fails instead on a single one that I got before :-). I suspect an issue in the building mechanisms.

popovec:
There is big difference on travis  for previous pull requests tests (for example #1734) --  travis is running Ubuntu 14.04.5 LTS, and  test run for this  pull request - Description:    Ubuntu 16.04.6 LTS


Seems there is problem, that some packages are not installed on travis, from failed OsEID simulation:

_The program 'socat' is currently not installed. To run 'socat' please ask your administrator to install the package 'socat'_

from failed simulation javacard:

_The program 'ant' is currently not installed. To run 'ant' please ask your administrator to install the package 'ant'_

You can add `ant` and `socat` package into `.travis.yml` ..

kouril:
> You can add `ant` and `socat` package into `.travis.yml` ..

Thanks for the investigation. Not sure who your're referring to with "You" but I'd appreciate if someone more familiar with the building systems would take a look :-).

frankmorgner:
thanks

PR #1748, opened 2019-07-20
Allow chaining for pkcs15-init --store-private-key EC keys

### LINKED ISSUE #1747: Import of secp384r1 private key in isoApplet card fails
### Problem Description

When trying to import a secp384r1 private key into an isoApplet card using pkcs15-init:
`pkcs15-init -S test.key  -i 1 -u sign -l tls -a ff`
I get an error message:
`Failed to store private key: Invalid arguments`

Digging deeper into this, it turns out that the apdu len was 357 bytes but isoApplet_put_data_prkey_ec tried to send it as a short apdu, enven though the card didn't allow extended apdus.

### Proposed Resolution

Enable APDU chaining if the length of the apdu is greater than 255 bytes and the card does not allow extended APDUs in isoApplet_put_data_prkey_ec, as is already checked in isoApplet_put_data_prkey_rsa.

I am preparing a pull request for this.

### Steps to reproduce

Create a EC private key on the NIST P-384 curve and try to import it onto a card with an isoApplet which does not have extended APDUs enabled.

```
openssl genpkey -algorithm EC -pkeyopt ec_paramgen_curve:P-384 -pkeyopt ec_param_enc:named_curve -out test.key
pkcs15-init -S test.key  -a ff
```

### Logs
```
0x7f2d088ad740 16:20:38.256 [pkcs15-init] card-isoApplet.c:872:isoApplet_put_data_prkey_ec: called
0x7f2d088ad740 16:20:38.256 [pkcs15-init] card-isoApplet.c:505:isoApplet_put_ec_params: called
0x7f2d088ad740 16:20:38.256 [pkcs15-init] card-isoApplet.c:538:isoApplet_put_ec_params: returning with: 0 (Success)
0x7f2d088ad740 16:20:38.256 [pkcs15-init] apdu.c:554:sc_transmit_apdu: called
0x7f2d088ad740 16:20:38.256 [pkcs15-init] apdu.c:261:sc_check_apdu: failed length check for short APDU
0x7f2d088ad740 16:20:38.256 [pkcs15-init] apdu.c:339:sc_check_apdu: Invalid Case 3 short APDU:
cse=03 cla=00 ins=db p1=3f p2=ff lc=357 le=0
resp=(nil) resplen=0 data=0x7ffca021b990 datalen=357
0x7f2d088ad740 16:20:38.256 [pkcs15-init] card-isoApplet.c:934:isoApplet_put_data_prkey_ec: APDU transmit failed
0x7f2d088ad740 16:20:38.256 [pkcs15-init] card-isoApplet.c:967:isoApplet_put_data_prkey_ec: returning with: -1300 (Invalid arguments)
```

jo-bitsch:
When importing a private key onto a pkcs15 card, if the card does not support
extended APDUs, we need to use chaining to store keys longer than 255 bytes.

While for RSA keys, this check was included, it was missing for EC keys.
This patch adds the SC_APDU_FLAGS_CHAINING flag to apdu.flags if data length is
greater than 255 and the card caps does not include SC_CARD_CAP_APDU_EXT.

This is particularly the case for secp384r1 keys.

This fixes #1747. 

##### Checklist

### DIFF
--- a/src/libopensc/card-isoApplet.c
+++ b/src/libopensc/card-isoApplet.c
@@ -928,6 +928,10 @@ isoApplet_put_data_prkey_ec(sc_card_t *card, sc_cardctl_isoApplet_import_key_t *
 	apdu.lc = p - sbuf;
 	apdu.datalen = p - sbuf;
 	apdu.data = sbuf;
+	if ((apdu.datalen > 255) && !(card->caps & SC_CARD_CAP_APDU_EXT))
+	{
+		apdu.flags |= SC_APDU_FLAGS_CHAINING;
+	}
 	r = sc_transmit_apdu(card, &apdu);
 	if(r < 0)
 	{


philipWendland:
I can test August 12th at the earliest. 
But the change looks good to me, even without testing.

frankmorgner:
that's enough for me

Issue #1735, opened 2019-07-11
CardOS 5.3 card does not work with windows minidriver


jarabizna:
### Problem Description
Internet Explorer or Google Chrome in Windows does not recognize CardOS5.3 card. Card is working correctly in Firefox. Issue is related to opensc version 0.19.

Card definition is missing in registry key for PKI minidriver - HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards

### Proposed Resolution
Modify windows installer for adding more registry entries in HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards key structure - example is content of exported  .reg file:
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards\CardOS v5.3]
"Crypto Provider"="OpenSC CSP"
"Smart Card Key Storage Provider"="Microsoft Smart Card Key Storage Provider"
"80000001"="C:\\Program Files\\OpenSC Project\\OpenSC\\minidriver\\opensc-minidriver.dll"
"ATR"=hex,3b,d2,18,00,81,31,fe,58,c9,03,16
"ATRMask"=hex,ff,ff,ff,ff,ff,ff,ff,ff,ff,ff,ff

### Steps to reproduce
certutil -scinfo -silent ends with error 
Cannot retrieve Provider Name for SCardGetCardTypeProviderName: <...related text translated to windows local language ...>. 0x2 (WIN32: 2 ERROR_FILE_NOT_FOUND)

Same issue will be with other CardOS versions > 5.0 (5.1, 5.2).

PR #1750, opened 2019-07-22
md: added missing cardos5 ATRs

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1735

##### Checklist

- [ ] macOS tokend is tested


win32/customactions.cpp
```diff
@@ -137,7 +137,11 @@ MD_REGISTRATION minidriver_registration[] = {
                                           13, {0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff}},
 	{TEXT("CardOS 4.4"),                      {0x3b,0xd2,0x18,0x02,0xc1,0x0a,0x31,0xfe,0x58,0xc8,0x0d,0x51},
                                           12, {0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff}},
-	{TEXT("CardOS v5.0"),                     {0x3b,0xd2,0x18,0x00,0x81,0x31,0xfe,0x58,0xc9,0x01,0x14},
+	{TEXT("CardOS v5.0 (a)"),                 {0x3b,0xd2,0x18,0x00,0x81,0x31,0xfe,0x58,0xc9,0x01,0x14},
+                                          11, {0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff}},
+	{TEXT("CardOS v5.0 (b)"),                 {0x3b,0xd2,0x18,0x00,0x81,0x31,0xfe,0x58,0xc9,0x02,0x17},
+                                          11, {0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff}},
+	{TEXT("CardOS v5.0 (c)"),                 {0x3b,0xd2,0x18,0x00,0x81,0x31,0xfe,0x58,0xc9,0x03,0x16},
```
Jakuje: I would mark these strings as CardOS 5.3 as initially marked in the cardos driver since these are really a new version (even though the `card->type` is the same since there was no difference in handling):

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-cardos.c#L54

For me, this is the first time I hear about this file so if we want to have the consistency between the ATRs in drivers and in md, I think we should write a simple script to check whether they match and update it globally, not just opportunistically when somebody notices.

Jakuje:
Looks good to me for the actual changes. Still waiting for the answer for the second part of my comment, but that is probably for new issue and PR if it would make sense.

frankmorgner:
I'll try to take the cardos ATRs directly from the driver into the installer if possible.

Jakuje:
@frankmorgner how did the attempts to take ATRs from driver to MD installer turned out? Did you do some work for that or should I create a new issue so this will not get forgotten?

frankmorgner:
created https://github.com/OpenSC/OpenSC/issues/1793

PR #1757, opened 2019-08-01
Implement OpenSC CI without HW cards

PL4typus:
### Problem Description

 - The CI with hardware is complicated, limits where the runners can be located (CAC export control, hardware requirement), which limits when it can be run. It is also less easy to reproduce hardware tests for contributors with limited ressources.
 - Travis CI doesn't offer a varied choice of up to date images to run tests. 
 
### Proposed Resolution

This proposal is about implementing OpenSC CI operating on virtual smart cards and using GitLab CI. This was done using Libcacard to emulate Common Access  Cards (CAC), but could be done with any kind of virtual smart card (for example, see the [vsmartcard project](https://github.com/frankmorgner/vsmartcard/)).
It is divided in three related sub-projects:

 - First of all, there is [virt_cacard](https://github.com/PL4typus/virt_cacard), a program using [libcacard](https://gitlab.freedesktop.org/spice/libcacard/), [virtualsmartcard's vpcd](https://github.com/frankmorgner/vsmartcard/tree/master/virtualsmartcard) and [softhsm2](https://github.com/opendnssec/SoftHSMv2) to present a virtual CAC to the user through the PC/SC interface. 
 
schema
 
 - Then, I integrated [virt_cacard as a submodule for OpenSC.](https://github.com/PL4typus/OpenSC/tree/virt_cacard) I made a mirror of this repository on GitLab and [wrote CI jobs that run the p11tests on the virtual CAC](https://gitlab.com/PL4typus/OpenSC/pipelines/74197004). 
 Mirroring pulls the changes from the GitHub repo automatically and runs the CI tests as set up. For instance, this could be done with pull requests.
 - Finally, those jobs are run by GitLab shared runners using container images built using GitLab-CI [in this project](https://gitlab.com/PL4typus/opensc-images) and stored in GitLab's container registry. This is done in a similar fashion to [GnuTLS](https://gitlab.com/gnutls/gnutls)'s CI and [build-images repository](https://gitlab.com/gnutls/build-images/). 
 This gives more flexibility to test on different platforms, and also presents to the user a clear method for setting up a build environment, or to use the provided container images to try things out. 
 
 
Integrating this project to upstream and increasing the test coverage of the OpenSC will avoid future regressions and breakages already in upstream.


Jakuje:
@frankmorgner Thank you for a comment. We tried to run this as part of the travis CI, but we were hitting the limitations of outdated Ubuntu images all the time so we decided to try the GitLab, which provides far more flexibility even for the price of maintaining the mirror on gitlab (I would be ofcourse happy to help with that, if that would be a burden for you).

martinpaljak:
Why not try out Github Actions, which is also based on Docker?

PL4typus:
Hello, and thank you for the comments!

@martinpaljak It looks like Github Actions is still in public beta. While it definitely sounds interesting, perhaps it would be more pertinent to use a released & stable solution for OpenSC?

@frankmorgner @Jakuje , it seems that in the meantime Travis made Ubuntu 18.04 available. Since I already have written the job for this release on GitLab-CI, I should be able to port it to Travis-CI. I will get on it.

Jakuje:
@PL4typus  Please, check the appveyor failures. It looks like it can not build your submodule on windows so I think the easiest thing would be to skip it there:
```
configure.ac:1100: error: required file 'src/tests/virt_cacard/Makefile.in' not found
```
Additionally, the travis build did not kick in for some reason. Isn't there some typo in the travis.yml?

frankmorgner:
A proper integration in Github (Travis CI) has the advantage that PRs get tested before being merged into master. However, I'm fine with only including the CI badge only and letting someone else handle the configuration (of Gitlab). But please don't add a submodule or modify OpenSC's internal files then. It also doesn't make sense to add a dependency to OpenSC from virt_cacard's perspective. It should work with any CAC capable software that accesses the card through PC/SC.

PL4typus:
> But please don't add a submodule or modify OpenSC's internal files then.

@frankmorgner This makes a lot of sense, I'm correcting that right now. 

> It also doesn't make sense to add a dependency to OpenSC from virt_cacard's perspective.

This is only the case because of the decision to make it a submodule within OpenSC, but there is a standalone branch for virt_cacard. I will update it and set it as default as part of the process of making it external to OpenSC.

PL4typus:
#### This commit should answer the requested changes. 
I tried not to touch too much of the existing structure with .travis.yml, but had to change one thing:

The *wine* package is named differently in Ubuntu Bionic (*wine-stable* or *wine-development* instead), so it was causing my build to fail in the install stage.

As virt_cacard has to be built on Bionic, I had to move the installation of *wine* to before_install and only for the jobs that need it. [I did it like this](https://github.com/PL4typus/OpenSC/blob/4f17e8d5a8b878600f4a5086d23094a034b17c7e/.travis.yml#L81).

README.md
```diff
@@ -11,6 +11,7 @@ Build and test status of specific cards:
 | Cards                                                               | Status                                                                                                                            |
 |---------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
 | CAC                                                                 | [![CAC](https://gitlab.com/redhat-crypto/OpenSC/badges/cac/build.svg)](https://gitlab.com/redhat-crypto/OpenSC/pipelines)         |
+| [virt_CACard](https://github.com/PL4typus/OpenSC/tree/virt_cacard)                                                         | [![virt_cacard](https://gitlab.com/PL4typus/OpenSC/badges/virt_cacard/pipeline.svg)](https://gitlab.com/PL4typus/OpenSC/pipelines)        |
```
Jakuje: This should probably link now to the travis results.

PL4typus: The travis badge is for the whole build, and not a specific job. Since there is already a travis badge for OpenSC, should I leave the one pointing specifically to virtual CAC testing, or remove it altogether ?

frankmorgner: I agree that the image should point to the Travis result. Hopefully, the service will allow linking to a specific job sometime soon, which would the link make more specific then.

frankmorgner:
Sorry for being misunderstanding, the image should be
```
[![virt_CACard](https://travis-ci.org/OpenSC/OpenSC.svg)](https://travis-ci.org/OpenSC/OpenSC/branches)
```
because it's tested within the OpenSC CI job.

PL4typus:
@frankmorgner Sorry, I misunderstood! I changed it again, should be good now!

frankmorgner:
thanks

frankmorgner:
@PL4typus you could also post the basic script [in the wiki](https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Simulation)

Issue #1731, opened 2019-07-07
Possible wrong Le handling / never ending loop


popovec:
### Problem Description

I have run some experiments with T1/T0 protocol and max_recv_size = 256. There is  difference in APDU handling for T0/T1 protocol (especialy, handling of *Le* in `sc_transmit()`, comments below).  

If *Le* is set to 0, I do not expect any data in receive buffer. But from log (T1 protocol) this call returns data:

```` 
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] apdu.c:518:sc_transmit: CLA:0, INS:B0, P1:0, P2:0, data(0) (nil) Le=0
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] apdu.c:363:sc_single_transmit: called
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] apdu.c:370:sc_single_transmit: CLA:0, INS:B0, P1:0, P2:0, data(0) (nil)
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] reader-pcsc.c:285:pcsc_transmit: reader 'OsEIDsim 00 00'
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] reader-pcsc.c:286:pcsc_transmit:
Outgoing APDU (5 bytes):
00 B0 00 00 00 .....
P:4675; T:0x139860042953152 15:22:22.934 [pkcs15-tool] reader-pcsc.c:214:pcsc_internal_transmit: called
P:4675; T:0x139860042953152 15:22:22.938 [pkcs15-tool] reader-pcsc.c:295:pcsc_transmit:
Incoming APDU (258 bytes):
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
[...]

PR #1761, opened 2019-08-04
Fix 1734

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-esteid2018.c
+++ b/src/libopensc/card-esteid2018.c
@@ -170,11 +170,11 @@ static int esteid_set_security_env(sc_card_t *card, const sc_security_env_t *env
 	sc_log(card->ctx, "algo: %d operation: %d keyref: %d", env->algorithm, env->operation, env->key_ref[0]);
 
 	if (env->algorithm == SC_ALGORITHM_EC && env->operation == SC_SEC_OPERATION_SIGN && env->key_ref[0] == 1) {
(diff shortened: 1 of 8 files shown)


popovec:
Ok, this fix the #1731 but  we still need #1734 to fix  errors in `sc_format_apdu_cse_lc_le()`.

frankmorgner:
Thanks for the testing. I've added the limitations into the documentation of `sc_format_apdu_ex()`, which is the only place where `sc_format_apdu_cse_lc_le()` is called.

dengert:
This PR may fix the specific  problem in #1731 , but IMHO `sc_format_apdu_ex` is only partially implemented. It looks like it will only work correctly with extended APDUs and it does not factor in data chaining and may have problems with T=0.  

This problem started when a developer assumed `sc_format_apdu_ex` could replace a lot of extra code and tried to use it with short APDUs.  I would like to use it too. 

If the intent of https://github.com/OpenSC/OpenSC/commit/27526de0219f42a666eaadf1e484ea96a9ba8db9 is to use `sc_format_apdu_ex` as a replacement for `sc_format_apdu`  everywhere then it needs to support T=0, T=1, data chaining, short APDUs and other things. Adding comments does not address the underlying problems introduced with `sc_format_apdu_ex` or help developers avoid the same problems. 

#1734 was only a partial fix. As I said in #1731 "Fix it however you want." But the problems still exist.

metsma:
I confirm, this PR fixes EstEID 2018 card communication with T=0

PR #1762, opened 2019-08-05
opensc-tool: do not connect card if not neccesary, fix util.c errors

popovec:
opensc-tool: for options --version, --list-readers, -D, etc. we do not
need to connect card/reader. This removes unnecessary error messages
if card is not present in card reader or if reader is not available.
util.c: use symbolic error codes, pass error codes to caller without change.


frankmorgner:
thanks

Issue #1659, opened 2019-04-15
Initializing OpenPGP card with OpenSC


frankmorgner:
### Problem Description

In order to extend the CI tests, I'm trying to initialize and use a OpenPGP card (i.e. https://github.com/Yubico/ykneo-openpgp/).

- The [wiki page](https://github.com/OpenSC/OpenSC/wiki/OpenPGP-card#3-generating-keys) suggests `openpgp-tool --verify CHV3 --pin 12345678 --gen-key 3` for generating keys, which doesn't work. It also promotes the argument `--key-len` which has been replaced by `--key-type`.
- `pkcs15-init --generate-key rsa/2048 --auth-id 3 --verify --pin 12345678` doesn't work either, although the tool should have some support for OpenPGP...

### Proposed Resolution

Ideally, fix OpenSC for a self sustained solution. Alternatively use gpg --card-edit (is this scriptable?).

### Logs

- with openpgp-tool: https://travis-ci.org/frankmorgner/OpenSC/jobs/519471598#L5675-L5678
- with pkcs15-init: https://travis-ci.org/frankmorgner/OpenSC/jobs/519582301#L5674-L5678

Jakuje:
The `gpg --card-edit` would be painful in the way that it would require also the `scd-pkcs11` to go through the OpenSC, otherwise it would be testing the OpenPGP only (at least for initialization).

dengert:
Any chance the RSA 1024 is not supported in ykneo-openpgp/applet/bin,  jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar or com.licel.jcardsim.remote.VSmartCard openpgp_jcardsim.cfg?

Does it work with 2024?

alex-nitrokey:
The [changes in the openpgp-tool](https://github.com/OpenSC/OpenSC/pull/1508) are not in any release yet. Therefore, it might not be good to reference to the new commands in the wiki already. On the other side, this indeed needs to be changed in future. I don't know why the command is not working right now, but it does work in 0.19.

`pkcs15-init --generate-key rsa/2048 --auth-id 3 --verify --pin 12345678` is working for 0.19 as well, but not on current master. (tested on ZeitControl with OpenPGP Card 3.3 with Nitrokey Pro).
I get the following output:
```
Using reader with a card: Nitrokey Nitrokey Pro (00005F120000000000000000) 00 00
Unable to find PIN code: Requested object not found
Failed to verify User PIN : Requested object not found
```

[...]

PR #1770, opened 2019-08-22
Fix handle OpenPGP cards without changeable algorithm information

frankmorgner:
See https://github.com/OpenSC/OpenSC/issues/1659 for details

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -216,6 +216,9 @@ script:
       sleep 5;
       opensc-tool --card-driver default --send-apdu 80b800002210D276000124010200000000000001000010D276000124010200000000000001000000;
       opensc-tool -n;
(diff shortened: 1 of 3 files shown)


Jakuje:
I think this one broke the travis CI. Since that, there are no runs in master.

Jakuje:
The comments inside of the script blocks do not work. But if you want to have it there, you can workaround it this way (but putting it in subshell). Not submitting a PR, but feel free to pull this change:
https://github.com/Jakuje/OpenSC/commit/c7391dbbe278a0f9d2d468842d3c4a3082794da1

frankmorgner:
Indeed, I'll remove the comment

PR #1772, opened 2019-08-26
Add support for Gemalto IDPrime smart cards

Jakuje:
The card is largely ISO 7816 compliant, but does not provide any simple way of listing the content which is supported by current PKCS#15 implementation therefore the PKCS#15 emulator had to be used.
    
The certificates are compressed in a similar way as in DNIE cards which complicates reading from the card and which I think could be moved to the shared ISO (or some other file since I saw that code already many times).

The card supports wide range of algorithms including RSA-PSS and RSA-OAEP padding schemes in-card for SHA2 variants. On the other hand, it does not allow raw RSA and SHA1 hashes on card anymore nor the raw-RSA.
    
The card is manufactured by Gemalto so it has strict ATR which can be used for detection.

This patch set also implements the basic support for RSA-OAEP in the OpenSC internals, which was tested through the PKCS#11 interface, but I am not familiar with the minidriver so it might need some more work there. Also I would aprecieate any comments for improvement or simplification.

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/pkcs15-cert.c
```diff
@@ -545,6 +549,88 @@ sc_pkcs15_encode_cdf_entry(sc_context_t *ctx, const struct sc_pkcs15_object *obj
...
+#define SC_PKCS15_X509_USAGE_DERIVE \
+	SC_X509_KEY_AGREEMENT
+#define SC_PKCS15_X509_USAGE_UNWRAP \
+	(SC_X509_KEY_ENCIPHERMENT | \
+	SC_X509_KEY_AGREEMENT)
+#define SC_PKCS15_X509_USAGE_DECRYPT \
+	(SC_X509_DATA_ENCIPHERMENT | \
+	SC_X509_ENCIPHER_ONLY)
+#define SC_PKCS15_X509_USAGE_NONREPUDIATION \
+	SC_X509_NON_REPUDIATION
```
frankmorgner: Some parts of the code use this (or similar):
```
#define USAGE_ANY_SIGN		(SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_NONREPUDIATION)
#define USAGE_ANY_DECIPHER	(SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP)
#define USAGE_ANY_AGREEMENT (SC_PKCS15_PRKEY_USAGE_DERIVE)
```
Could it be that you should actually use the `SC_PKCS15_PRKEY...` flags?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+	p15card->tokeninfo->manufacturer_id = strdup("Gemalto");
+
+	/*
+	 * get serial number
+	 */
+	memset(&serial, 0, sizeof(serial));
+	r = sc_card_ctl(card, SC_CARDCTL_GET_SERIALNR, &serial);
+	if (r < 0) {
+		sc_log(card->ctx, "sc_card_ctl rc=%d", r);
+		p15card->tokeninfo->serial_number = strdup("00000000");
```
frankmorgner: Wouldn't it be better to throw an error if there was no serial number?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+
+	pin_info.auth_type = SC_PKCS15_PIN_AUTH_TYPE_PIN;
+	sc_pkcs15_format_id(pin_id, &pin_info.auth_id);
+	pin_info.attrs.pin.reference     = 0x11;
+	pin_info.attrs.pin.flags         = SC_PKCS15_PIN_FLAG_INITIALIZED;
+	pin_info.attrs.pin.type          = SC_PKCS15_PIN_TYPE_ASCII_NUMERIC;
+	pin_info.attrs.pin.min_length    = 4;
+	pin_info.attrs.pin.stored_length = 8;
+	pin_info.attrs.pin.max_length    = 8;
+	sc_format_path("", &pin_info.path);
```
frankmorgner: This looks strange. Doesn't give `memset(&pin_info, 0, sizeof(pin_info));` a good enough path?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+		 * the key reference later used in the security environment */
+		prkey_info.path.len = 0;
+		prkey_info.path.aid.len = 0;
+		pubkey_info.key_reference = prkey_info.key_reference;
+		sc_log(card->ctx,  "Key ref r=%x", prkey_info.key_reference);
+
+		pubkey_info.native        = 1;
+		prkey_info.native        = 1;
+
+		memcpy(cert_obj.label, cert_label, sizeof(cert_label));
```
frankmorgner: These labels are rather useless. Would it make sense to add a counter to the label or to use the CN from the certificate as label?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		/* following will find the cached cert in cert_info */
+		r =  sc_pkcs15_read_certificate(p15card, &cert_info, &cert_out);
+		if (r < 0 || cert_out->key == NULL) {
+			sc_log(card->ctx,  "Failed to read/parse the certificate r=%d",r);
+			if (cert_out != NULL)
+				sc_pkcs15_free_certificate(cert_out);
+			continue;
+		}
```
frankmorgner: `sc_pkcs15_read_certificate()` includes handling of in-memory or on-file cache, which you've implemented above, doesn't it?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+
+		r = sc_pkcs15emu_add_x509_cert(p15card, &cert_obj, &cert_info);
+		if (r < 0) {
+			sc_log(card->ctx,  " Failed to add cert obj r=%d",r);
+			sc_pkcs15_free_certificate(cert_out);
+			continue;
```
frankmorgner: I think the memory in `cert_der` needs to be freed in case of an error.

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+			if (r == SC_SUCCESS) {
+				token_name = malloc (cn_len+1);
+				if (!token_name) {
+					free(cn_name);
+					r = SC_ERROR_OUT_OF_MEMORY;
+					goto fail;
+				}
+				memcpy(token_name, cn_name, cn_len);
+				free(cn_name);
+				token_name[cn_len] = 0;
```
frankmorgner: `'\0'`

src/libopensc/pkcs15-cac.c
```diff
@@ -407,7 +324,7 @@ static int sc_pkcs15emu_cac_init(sc_pkcs15_card_t *p15card)
 		if (r < 0) {
 			usage = 0xd9ULL; /* basic default usage */
```
frankmorgner: Could you use some speaking name instead of `0xd9ULL`?

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,268 @@
...
+
+		r = sc_pkcs15_encode_pubkey_as_spki(card->ctx, cert_out->key,
+			&pubkey_info.direct.spki.value, &pubkey_info.direct.spki.len);
+		if (r < 0)
+			goto fail;
+		pubkey_obj.emulated = cert_out->key;
+
+		r = sc_pkcs15_get_bitstring_extension(card->ctx, cert_out, &usage_type, &usage, NULL);
+		if (r < 0) {
+			usage = 0xd9ULL; /* basic default usage */
```
frankmorgner: Could you use some speaking name instead of `0xd9ULL`?

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+static int idprime_init(sc_card_t *card)
+{
+	int r;
+	unsigned long flags;
+
+	/* TODO We should get this from the MF FCI:
+	 * 00 A4 04 00 00 (select MF returning FCI, tag 9F65 under tag A5 from GP
+	 */
+	card->max_send_size = (255 - 12);
+	card->max_recv_size = 255;
```
frankmorgner: not 256, i.e. Le=0 => Ne=256?

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+static int idprime_match_card(sc_card_t *card)
+{
+	int i, r;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	i = _sc_match_atr(card, idprime_atrs, &card->type);
+	if (i < 0)
+		return 0;
+
+	r = idprime_find_and_initialize(card, 0);
```
frankmorgner: Ideally, you only modify the `card` object in `idprime_init`, i.e. you only add private data in card->init so that it definitely does get freed in card->finish. By the way, your card doesn't have a function for cleanup, where are you freeing the private data?

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+		if (r < 0) {
+			LOG_FUNC_RETURN(card->ctx, r);
+		}
+		if (r < 4) {
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_DATA);
+		}
+		if (buffer[0] == 1 && buffer[1] == 0) {
+#ifdef ENABLE_ZLIB
+			size_t expectedsize = buffer[2] + buffer[3] * 0x100;
+			r = sc_decompress_alloc(&priv->cache_buf, &(priv->cache_buf_len),
```
frankmorgner: Could you have a look at sc_decompress_zlib_alloc? It looks like there's are multiple problems with
- calling realloc with size 0 (possible double free)
- missing length checks when calculating the buffer sizes
- possible misconception of realloc (afaik, "shrinking down" a buffer is not possible)

This could eventually hit you when passing cache_buf and cache_buf_len without any checks.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+					 expectedsize, priv->cache_buf_len);
+				LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_DATA);
+			}
+#else
+			sc_log(card->ctx, "compression not supported, no zlib");
+			return SC_ERROR_NOT_SUPPORTED;
+#endif /* ENABLE_ZLIB */
+		} else {
+			/* assuming uncompressed certificate */
+			priv->cache_buf = malloc(r);
```
frankmorgner: free(priv->cache_buf)?

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+
+	/* INS: 0x2A  PERFORM SECURITY OPERATION
+	 * P1:  0x80  Resp: Plain value
+	 * P2:  0x86  Cmd: Padding indicator byte followed by cryptogram */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4, 0x2A, 0x80, 0x86);
+	apdu.resp    = out;
+	apdu.resplen = outlen;
+	apdu.le      = outlen;
+
+	sbuf[0] = 0x81; /* padding indicator byte, 0x81 = ? */
```
frankmorgner: grafik

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,698 @@
...
+
+	idprime_ops = *iso_ops;
+	idprime_ops.match_card = idprime_match_card;
+	idprime_ops.init = idprime_init;
+	idprime_ops.read_binary = idprime_read_binary;
+	idprime_ops.select_file = idprime_select_file;
+	idprime_ops.card_ctl = idprime_card_ctl;
+	idprime_ops.set_security_env = idprime_set_security_env;
+	idprime_ops.compute_signature = idprime_compute_signature;
+	idprime_ops.decipher = idprime_decipher;
```
frankmorgner: idprime_ops.finish?

src/libopensc/iso7816.c
```diff
@@ -48,6 +48,11 @@ static void fixup_transceive_length(const struct sc_card *card,
 		 * All other workarounds must be carried out by the upper layers. */
 		apdu->le = sc_get_max_recv_size(card);
 	}
+
+	/* When we shorten the APDU here, make sure to change the case too */
+	if (apdu->le <= 255 && apdu->lc <= 255) {
+		apdu->cse &= ~SC_APDU_EXT;
```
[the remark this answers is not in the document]
Jakuje: Changed. I hit this in the beginning of the development. I think this card does not support the extended APDUs and this APDUs were coming from some detection steps. But as I see now, it works for me also without this code so if you prefer, I can drop this commit.

src/libopensc/pkcs15-idprime.c
Jakuje: I do not have any documentation about the card so I would rather have it working if some things will happen do behave differently on other cards than failing hard. I can modify it to an error, but I would rather not.

src/libopensc/pkcs15-idprime.c
Jakuje: yes, you are right. Removed.

src/libopensc/pkcs15-idprime.c
Jakuje: The official driver lists the labels just as some undocumented hashes. I will add a counter since the CN names will show up in the subject field anyway.

src/libopensc/pkcs15-idprime.c
Jakuje: Yes. Because
 * the internals do not handle compressed certificates
 * the pkcs15 code can not handle the cards where only certificate is available (pkcs15 expects real separate key objects)

I did not find a way around this but if you will propose a better way to do that, I would be happy to change it.

src/libopensc/pkcs15-idprime.c
Jakuje: Yes, you are right. Fixed.

src/libopensc/pkcs15-idprime.c
Jakuje: Changed.

src/libopensc/pkcs15-idprime.c
Jakuje: Yes. Fixed.

src/libopensc/card-idprime.c
Jakuje: Surprisingly this looks like working now even without this code so I will try to drop that for now.

src/libopensc/card-idprime.c
Jakuje: It is as you write. The private data is modified only from the init() in similar manner as piv/cac drivers. In this case, I heavily depend on the card ATR and this one select command that is executed in the match is more like a safeguard.

Indeed, it could be rewritten to simplify this by calling `idprime_select_index()` directly from  the init and it would be probably nicer. Lets see

src/libopensc/card-idprime.c
Jakuje: I think the `cache_buf` member is freed when different applet is selected and therefore we should never hit any issues with realloc. It just gets the buffer with compressed certificate and decompress that into the `cache_buf`, which is allocated by `sc_decompress_zlib_alloc()`.

src/libopensc/card-idprime.c
Jakuje: This should not be needed. We are in condition `!priv->cached`, which means there should be nothing in the priv->cache_buf.

src/libopensc/card-idprime.c
Jakuje: Proprietary? Added. I did not see a lot of value in this, but probably better than a question mark.

src/libopensc/card-idprime.c
Jakuje: Added.

Jakuje:
Thank you for the review. I fixed the proposed changed and improvements. Fast scan through the Gemsafe token did not show anything familiar for me.

src/libopensc/pkcs15-cert.c
Jakuje: No, these are completely different values. One is the internal PKCS15 flag and the other X509 value (used for usage mapping):
```
src/libopensc/pkcs15.h:#define SC_X509_DIGITAL_SIGNATURE     0x0001UL
src/libopensc/pkcs15.h:#define SC_PKCS15_PRKEY_USAGE_SIGN               0x04
```

PavelKa:
Hello, after successful build and installation I was able to read information from card:
`pkcs15-tool -D
PKCS#15 Card [My Name]:
	Version        : 0
	Serial number  : XY
	Manufacturer ID: Gemalto
	Flags          : 

PIN [PIN]
	ID             : 11
	Flags          : [0x10], initialized
	Length         : min_len:4, max_len:8, stored_len:8
	Pad char       : 0x00
	Reference      : 17 (0x11)

Private RSA Key [Certificate 1]
	Usage          : [0x0E], decrypt, sign, signRecover
	Access Flags   : [0x1D], sensitive, alwaysSensitive, neverExtract, local
	ModLength      : 1024
	Key ref        : 49 (0x31)
	Native         : yes
	Auth ID        : 11
	ID             : 0001
	MD:guid        : some guid

Public RSA Key [Certificate 1]
	Object Flags   : [0x00]
	Usage          : [0xC1], encrypt, verify, verifyRecover
	Access Flags   : [0x02], extract
	ModLength      : 1024
	Key ref        : 49 (0x31)
	Native         : yes
	ID             : 0001
	DirectValue    : <absent>

X.509 Certificate [Certificate 1]
	Object Flags   : [0x00]
	Authority      : no
	Path           : 
	ID             : 0001
	Encoded serial : 02 13 some code`

Unfortunately I'm not able to use card for a authentication due to following error:
`P:7821; T:0x140245136546560 21:21:38.487 [opensc-pkcs11] framework-pkcs15.c:3919:pkcs15_prkey_sign: Initiating signing operation, mechanism 0x1.
P:7821; T:0x140245136546560 21:21:38.487 [opensc-pkcs11] card.c:473:sc_lock: called
P:7821; T:0x140245136546560 21:21:38.487 [opensc-pkcs11] reader-pcsc.c:651:pcsc_lock: called
P:7821; T:0x140245136546560 21:21:38.487 [opensc-pkcs11] card.c:513:sc_lock: returning with: 0 (Success)
P:7821; T:0x140245136546560 21:21:38.487 [opensc-pkcs11] framework-pkcs15.c:4050:pkcs15_prkey_sign: Selected flags 102. Now computing signature for 51 bytes. 128 bytes reserved.
P:7821; T:0x140245136546560 21:21:38.488 [opensc-pkcs11] pkcs15-sec.c:565:sc_pkcs15_compute_signature: called
P:7821; T:0x140245136546560 21:21:38.488 [opensc-pkcs11] pkcs15-sec.c:199:format_senv: Card does not support RSA with key length 1024
P:7821; T:0x140245136546560 21:21:38.488 [opensc-pkcs11] pkcs15-sec.c:200:format_senv: returning with: -1408 (Not supported)
P:7821; T:0x140245136546560 21:21:38.488 [opensc-pkcs11] pkcs15-sec.c:572:sc_pkcs15_compute_signature: Could not initialize security environment: -1408 (Not supported)
`
Do You have any idea how to resolve the problem.  Old version of driver  libidprimepkcs11 provided by gemalto is working on older linux  versions (but not supported on latest) so RSA with 1024 should be supported.

Jakuje:
Thank you for testing. I have only 2k RSA keys in my card. Now, I pushed an updated which should unbreak support for the 1k RSA keys.

I think I have Gemalto IDPrime 3810, which should support RSA keys up to 2048 bits [1], so I think we are fine with 1k and 2k keys for now. Not sure whether some newer versions will support something bigger. The specification also says that the card should support SHA-1, but it was not my observation.

https://safenet.gemalto.com/multi-factor-authentication/idprime-md-pki-smart-cards/

PavelKa:
Thank You for update . Authentication now fails  on :
`P:13354; T:0x140243878714112 13:04:05.709 [opensc-pkcs11] reader-pcsc.c:306:pcsc_transmit: 
Incoming APDU (2 bytes):
6A 88 j.
P:13354; T:0x140243878714112 13:04:05.709 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] apdu.c:535:sc_transmit: returning with: 0 (Success)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] card.c:524:sc_unlock: called
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] card-idprime.c:575:idprime_compute_signature: The initial APDU did not return the same data
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] card-idprime.c:576:idprime_compute_signature: returning with: -1400 (Internal error)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] sec.c:63:sc_compute_signature: returning with: -1400 (Internal error)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] card.c:524:sc_unlock: called
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] pkcs15-sec.c:707:sc_pkcs15_compute_signature: use_key() failed: -1400 (Internal error)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] pkcs15-sec.c:565:sc_pkcs15_compute_signature: called
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] pkcs15-sec.c:613:sc_pkcs15_compute_signature: supported algorithm flags 0xE0E032, private key usage 0xE
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] padding.c:470:sc_get_encoding_flags: called
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] padding.c:474:sc_get_encoding_flags: iFlags 0x2002, card capabilities 0xE0E032
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] padding.c:526:sc_get_encoding_flags: pad flags 0x0, secure algorithm flags 0x2002
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] padding.c:527:sc_get_encoding_flags: returning with: 0 (Success)
P:13354; T:0x140243878714112 13:04:05.710 [opensc-pkcs11] pkcs15-sec.c:671:sc_pkcs15_compute_signature: DEE flags:0x00002002 alg_info->flags:0x00e0e032 pad:0x00000000 sec:0x00002002`

Jakuje:
Can you share complete debug logs (even through email if you do not want them public) with redacted PIN. This looks like a difference from what I observed with my cards. I would be also interested in what ATR and what card are you using in case the driver will need adjustments.

PavelKa:
Thank You for update. Last commit works well with my card.

paweljasinski:
When testing with different type of IDPrime MD card, there is an error:
```
rejap@zed:~/github/OpenSC-IdPrime$ opensc-tool --verbose --serial --info --atr
OpenSC 0.20.0 [gcc  7.4.0]
Enabled features: locking zlib openssl pcsc(libpcsclite.so.1)
Card ATR:
3B 7F 96 00 00 80 31 80 65 B0 84 56 51 10 12 0F ;.....1.e..VQ...
FE 82 90 00                                     ....
Connecting to card in reader Alcor Micro AU9560 00 00...
Failed to connect to card: Wrong length
```
And the log with debug enabled.
[error.log](https://github.com/OpenSC/OpenSC/files/3985569/error.log)

Jakuje:
The log looks good up to the attempt to read the card index file, which fails with error  (Security status not satisfied). This sounds like a card with different access policy to the card objects that I saw before.

Checking the log, the important part of the FCI of the index file will be this one:
```
8C: Tag: Security attribute in compact format
 03: Length
 03 D8 4A: Value
    03 is "Access mode byte for data object" affecting GET and PUT DATA operations
    D8 is "Security condition byte" saying both "Secure Messaging" and  "External Authenticate" needs to be fulfilled using SEID 0x08
    4A is "Security condition byte" saying both "Secure Messaging" needs to be active using SEID 0x0A
```

So you were right in your email, that the communication is encrypted with the official driver. From here on, I am not sure if we will be able to do anything more for your card. I have never worked with any card that would require SM, but other maintainers might have some tips.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	unsigned short length;
+} idprime_object_t;
+
+/*
+ * IDPrime private data per card state
+ */
+typedef struct idprime_private_data {
+	u8 *cache_buf;			/* cached version of the currently selected file */
+	size_t cache_buf_len;		/* length of the cached selected file */
+	int cached;			/* is the cached selected file valid */
```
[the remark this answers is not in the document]
Jakuje: I was not able to make the cache sensibly working with the certificate decompression. Do you have some reference of this combination working in some other driver?

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+/* This select some index file, which is useful for enumerating other files
+ * on the card */
+static int idprime_select_index(sc_card_t *card)
+{
+	int r;
+	sc_file_t *file = NULL;
+	sc_path_t index_path;
+
+	/* First, we need to make sure the IDPrime AID is selected */
+	r = idprime_select_idprime(card);
```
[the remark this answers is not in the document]
Jakuje: This is the first select in the card detection (and then in the initialization). In the first case, we know it is not selected. In the second run, we already selected the index.

I think I wrote this really as a sanity check to make sure we talk to the correct card, but having a second look now, it might be better to leave this only in the card detection step. But I would have to check with real card.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	/* Returns FCI with expected length of data */
+	sc_format_path("0101", &index_path);
+	r = iso_ops->select_file(card, &index_path, &file);
+	if (r != SC_SUCCESS) {
+		sc_file_free(file);
+		LOG_FUNC_RETURN(card->ctx, r);
+	}
+	r = file->size;
+	sc_file_free(file);
+	return r;
```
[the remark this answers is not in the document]
Jakuje: True. I will rewrite it and add a check. Thanks for reminder.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+
+		/* This looks like non-standard extension listing pkcs11 token info label in my card */
+		} else if ((memcmp(&start[4], "tinfo", 6) == 0) && (memcmp(&start[12], "p11", 4) == 0)) {
+			memcpy(priv->tinfo_df, new_object.df, sizeof(priv->tinfo_df));
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found p11/tinfo object");
+		}
+	}
+	r = SC_SUCCESS;
+done:
+	if (buf)
```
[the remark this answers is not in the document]
Jakuje: True. I will remove it. Thanks.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+
+static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, int length)
+{
+	u8 *buf = NULL;
+	int r = SC_ERROR_OUT_OF_MEMORY;
+	int i, num_entries;
+	idprime_object_t new_object;
+
+	buf = malloc(length);
+	if (buf == NULL) {
```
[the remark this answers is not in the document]
Jakuje: No. The `priv->cache_buf` is used only for read_binary called from the other code expecting certificates instead of compressed blobs. This is internal function, which calls directly iso operation without any other caching.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	}
+	new_object.fd = 0;
+	for (i = 0; i < num_entries; i++) {
+		u8 *start = &buf[i*21+1];
+
+		/* First two bytes specify the object DF */
+		new_object.df[0] = start[0];
+		new_object.df[1] = start[1];
+		/* Second two bytes refer to the object size */
+                new_object.length = bebytes2ushort(&start[2]);
```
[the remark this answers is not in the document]
Jakuje: will fix. Thanks

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+				new_object.key_reference = start[8] + 1;
+			}
+			new_object.fd++;
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found certificate with fd=%d, key_ref=%d",
+				new_object.fd, new_object.key_reference);
+			idprime_add_object_to_list(&priv->pki_list, &new_object);
+
+		/* This looks like non-standard extension listing pkcs11 token info label in my card */
+		} else if ((memcmp(&start[4], "tinfo", 6) == 0) && (memcmp(&start[12], "p11", 4) == 0)) {
+			memcpy(priv->tinfo_df, new_object.df, sizeof(priv->tinfo_df));
```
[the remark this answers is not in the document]
Jakuje: You are right. But we do not actually create an emulated tinfo object with this FD, but use just the DF only internally to pull the cardholder name later.

On the other hand, we could create a data object here, which would be more clean. But at this moment I do not see any value in exposing this raw object since all its content is exposed in the tokeninfo->label (see pkcs15-idprime.c).

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+			if (card->type == SC_CARD_TYPE_IDPRIME_V2) {
+				/* The key reference starts from 0x11 */
+				new_object.key_reference = 0x10 + new_object.fd;
+			} else {
+				/* The key reference is one bigger than the value found here for some reason */
+				new_object.key_reference = start[8] + 1;
+			}
+			new_object.fd++;
+			sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found certificate with fd=%d, key_ref=%d",
+				new_object.fd, new_object.key_reference);
```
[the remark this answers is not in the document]
Jakuje: Not index, but something from which is the key ID derived. Not sure how I ended up having the `new_object.fd++` incremented twice but at this moment, it results in IDs from 2, 4, 6 ... I think this should get fixed.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+
+	r = idprime_process_index(card, priv, r);
+	if (r != SC_SUCCESS) {
+		idprime_free_private_data(priv);
+		LOG_FUNC_RETURN(card->ctx, r);
+	}
+
+	card->drv_data = priv;
+
+	card->name = "Gemalto IDPrime";
```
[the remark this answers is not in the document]
Jakuje: The last information I have is that there will be more differences between the IDPrime cards versions. The one differentiation I know of now is the byte in CPLC data as described in the code, but I hope I will get hand on some more different cards in future to be able to write better detection.

Anyway, this is a good idea at least for future debugging so the OS version is visible directly in most of the logs. Will add.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	sc_format_path("0201", &cardid_path);
+	r = iso_ops->select_file(card, &cardid_path, &file);
+	if (r != SC_SUCCESS || file->size != IDPRIME_CARDID_LEN) { /* The cardid is always 16 B */
+		sc_file_free(file);
+                LOG_FUNC_RETURN(card->ctx, SC_ERROR_WRONG_LENGTH);
+	}
+
+	r = iso_ops->read_binary(card, 0, buf, file->size, 0);
+	sc_file_free(file);
+	if (r < 1) {
```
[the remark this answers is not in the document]
Jakuje: Thanks! Will fix.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	int r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	/* XXX this is assumed to be cardid for windows. It can be read from the index file */
+	sc_format_path("0201", &cardid_path);
+	r = iso_ops->select_file(card, &cardid_path, &file);
+	if (r != SC_SUCCESS || file->size != IDPRIME_CARDID_LEN) { /* The cardid is always 16 B */
+		sc_file_free(file);
+                LOG_FUNC_RETURN(card->ctx, SC_ERROR_WRONG_LENGTH);
```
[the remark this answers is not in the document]
Jakuje: Will fix.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	u8 buf[2];
+	int r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	if (tname == NULL) {
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
+	}
+
+	if (memcmp(priv->tinfo_df, "\0\0", 2) == 0) {
```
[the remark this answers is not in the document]
Jakuje: DF in IDPrime cards started for me with `02 01`. This was supposed to be detection whether this was set or not. I will create a separate flag whether it should be used or not regardless of the value of the DF.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	int r, len;
+	idprime_private_data_t * priv = card->drv_data;
+	u8 data[HEADER_LEN];
+	size_t data_len = HEADER_LEN;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (priv) { /* don't record anything if we haven't been initialized yet */
+		/* forget any old cached values */
+		if (priv->cache_buf) {
```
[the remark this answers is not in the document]
Jakuje: True. This is probably some copy&paste leftover from older driver. I could use this function also in the card detection method, but I do not see this bringing any advantage as the changes only revolve around certificates so I will drop this check.

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	/* We should be signing hashes only so we should not reach this limit */
+	if (datalen + 2 > sizeof(sbuf)) {
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
+	}
+
+	p = sbuf;
+	*(p++) = 0x90;
+	*(p++) = datalen;
```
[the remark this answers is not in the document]
Jakuje: Right. This is also from ISO. As we do not support raw rsa, I think I can shrink the sbuf buffer and then this will be covered in the previous check (`datalen + 2 > sizeof(sbuf)`)

src/libopensc/card-idprime.c
```diff
@@ -0,0 +1,789 @@
...
+	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+
+	if (apdu.sw1 == 0x90 && apdu.sw2 == 0x00)
+		LOG_FUNC_RETURN(card->ctx, apdu.resplen);
+	else
+		LOG_FUNC_RETURN(card->ctx, sc_check_sw(card, apdu.sw1, apdu.sw2));
+}
+
+
+static struct sc_card_driver * sc_get_driver(void)
```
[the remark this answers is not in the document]
Jakuje: I think there was some historical reason why all the drivers do it this way and why this construction is in all the drivers. Though I do not insist on this construct if removing it will not break something.

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,282 @@
...
+			continue;
+		} else {
+			pubkey_info.modulus_length = cert_out->key->u.rsa.modulus.len * 8;
+			prkey_info.modulus_length = cert_out->key->u.rsa.modulus.len * 8;
+			r = sc_pkcs15emu_add_rsa_pubkey(p15card, &pubkey_obj, &pubkey_info);
+			sc_log(card->ctx,  "adding rsa public key r=%d usage=%x",r, pubkey_info.usage);
+			if (r < 0)
+				goto fail;
+			r = sc_pkcs15emu_add_rsa_prkey(p15card, &prkey_obj, &prkey_info);
+			sc_log(card->ctx,  "adding rsa private key r=%d usage=%x",r, prkey_info.usage);
```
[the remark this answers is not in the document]
Jakuje: Right. The r value should be checked too.

src/libopensc/pkcs15-idprime.c
```diff
@@ -0,0 +1,282 @@
...
+		prkey_info.path.len = 0;
+		prkey_info.path.aid.len = 0;
+		pubkey_info.key_reference = prkey_info.key_reference;
+		sc_log(card->ctx,  "Key ref r=%x", prkey_info.key_reference);
+
+		pubkey_info.native        = 1;
+		prkey_info.native        = 1;
+
+		snprintf(cert_obj.label, SC_PKCS15_MAX_LABEL_SIZE, CERT_LABEL_TEMPLATE, i+1);
+		snprintf(pubkey_obj.label, SC_PKCS15_MAX_LABEL_SIZE, CERT_LABEL_TEMPLATE, i+1);
```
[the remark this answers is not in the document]
Jakuje: Right. I overlooked this. Will be fixed with next rebase

src/libopensc/card-idprime.c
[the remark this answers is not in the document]
frankmorgner: Did you make progress with this?

src/libopensc/card-idprime.c
frankmorgner: My comment was not about your code, but about the implementation of `sc_decompress_alloc()` itself. Maybe this is better treated in a seperate issue....

I've opened https://github.com/OpenSC/OpenSC/issues/1905

Jakuje:
@PavelKa @paweljasinski do you still have the IDPrime cards? Can you test the proposed changes in #2666 if they keep working with your cards?

paweljasinski:
Hi,

the cards I have access to never worked at the opensc api level. The
communication is proprietary :-(
At the moment, the only functionality I can test is card forwarding via RDP
protocol (pcscd, pcsclite).

CHeers
--pawel

PavelKa:
I'm sorry I still have the card but I  don't use it anymore. Maybe I can do some basic test, if you can provide steps to do ... ?

PavelKa:
My steps to test: 
git clone https://github.com/OpenSC/OpenSC.git
 cd OpenSC/
 gh pr checkout 2666
 ./bootstrap
. /configure --prefix=/usr --sysconfdir=/etc/opensc
make
sudo make install 

Both commands pkcs11-tool -L, pkcs11-tool -O return the same output as previous version.

Command pkcs11-tool --test --login --pin XXXX   returns the same output as previous version (I don't know pin ) : 
Aborting.
Hope this helps...
How to validate that I'm testing correct version and command sudo make install was successful ?
pkcs11-tool -I
Cryptoki version 3.0
Manufacturer     OpenSC Project
Library          OpenSC smartcard framework (ver 0.23)

Issue #1764, opened 2019-08-07
PKCS11 function C_SignFinal failed: rv = CKR_ARGUMENTS_BAD with PKCS#15 Card [CardOS V5.3] using 3072 bit RSA key


saper:
### Problem Description
Using opensc 0.19.0 with pcsc-lite 1.8.24,2 on FreeBSD 11.2-STABLE

Applied a small patch to get key length 3072 and 4096 recognized and get some extra debug output:

````patch
--- src/libopensc/card-cardos.c.orig    2018-09-13 11:52:42 UTC
+++ src/libopensc/card-cardos.c
@@ -239,6 +239,8 @@ static int cardos_init(sc_card_t *card)
        }
 
        if (card->type == SC_CARD_TYPE_CARDOS_V5_0) {
+               _sc_card_add_rsa_alg(card, 3072, flags, 0);
+               _sc_card_add_rsa_alg(card, 4096, flags, 0);
                /* Starting with CardOS 5, the card supports PIN query commands */
                card->caps |= SC_CARD_CAP_ISO7816_PIN_INFO;
        }
@@ -888,6 +890,7 @@ cardos_compute_signature(sc_card_t *card
        assert(card != NULL && data != NULL && out != NULL);
        ctx = card->ctx;
        SC_FUNC_CALLED(ctx, SC_LOG_DEBUG_VERBOSE);
+       sc_log(card->ctx,  "datalen/outlen/max %zd/%zd/%d", datalen, outlen, SC_MAX_APDU_BUFFER_SIZE);
 
        if (datalen > SC_MAX_APDU_BUFFER_SIZE)
                LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
````

I try to sign the following uuencoded bytes:

````uuencode
begin 644 COPYRIGHT.hash
MIUF]LQ'?J!%"LOTGV]28=<X(>U>XR!`_X^8#P(#WSKVRRSD4P%73")*]L,!&
3FLE%NQ;J%6QK(_$$PPKT'E,5Y```
`
end
````

My opensc.conf:

````
app default {
        debug = 4;
        debug_file = /home/saper/pki/cardos/opensc-debug.txt;
        reader_driver pcsc {
                max_send_size = 32767;
                max_recv_size = 32767;
        }
}
````

My test script:

````sh
[...]

PR #1776, opened 2019-08-28
Add support for 4K RSA keys in CardOS 5

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1764

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
This needs the `#include <stdlib.h>` for the windows builds. For some reason, this is not needed in travis since it gets included from some of the common files, but not on windows. Might need some more investigation why is that.

Also I am wondering why the travis did not run for this PR.

Otherwise it looks fine for me.

saper:
Even [https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html POSIX] says that `malloc(3)` needs `#include <stdlib.h>`. Did travis ran?

Issue #1769, opened 2019-08-20
YubiKey PIN policy is not supported


ghost:
### Problem Description

I have a YubiKey 5 which supports PIN policies per slot [[1]](https://support.yubico.com/support/solutions/articles/15000014219-yubikey-5-series-technical-manual#Policiesf5yfro) [[2]](https://developers.yubico.com/PIV/Introduction/Yubico_extensions.html#_pin_policy). There are three different PIN policies: never required, required once and always required. However, it seems the first two are unsupported, as applications always ask for PIN regardless of the policy. This does not allow to authenticate the card only. An example of the desired behavior is when in pkcs11-tool the code responsible for logging in is disabled (opt_login forced to false).

### Proposed Resolution
Maybe CKF_LOGIN_REQUIRED should be ignored and login should be deferred until actually required (CKR_USER_NOT_LOGGED_IN). I don't really know, tell me the proper way this should be handled and I can try to implement it.

### Steps to reproduce

```
ykman piv import-key 9e key.key --pin-policy never
# requires PIN
pkcs11-tool --sign
# works as desired
patched/pkcs11-tool --sign
```

### Logs
[OPENSC_DEBUG=9 patched/pkcs11-tool --sign](https://gist.github.com/matejgroma/374445e6b6a9870470bd90239301811c)
[pkcs11-tool -L](https://gist.github.com/matejgroma/64fbe08513c09d07b3630179aee6d58f)
[pkcs11-tool -O](https://gist.github.com/matejgroma/b6eb23bcd000175d60e3d0bfa9724764)

dengert:
This sounds like some new Yubico extension.  The PIV specifications from NIST [NIST.SP.800-73-4](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf) defines fixed pin policies. 

A quick look at [1](https://support.yubico.com/support/solutions/articles/15000014219-yubikey-5-series-technical-manual#Policiesf5yfro) and [2](https://developers.yubico.com/PIV/Introduction/Yubico_extensions.html#_pin_policy) does not show how to read the policy for a specific key.

dengert:
By default key `9E` should not require a PIN. You said you have a "patched/pkcs11-tool" What is the patch?

matejgroma:
9E should not require PIN, but it still asks somehow to login into the card. The patch (for 0.19.0) is:
```
[...]

PR #1784, opened 2019-09-06
pkcs11-tool: allow tests with keys that don't require PIN

frankmorgner:
Basically, I want to use the 9e key slot of a PIV card in pkcs11-tool. This key doesn't require authentication of the PIN and `pkcs15-tool -D` shows that the key isn't associated with the PIN. However, `pkcs11-tool --test` can't use this key for a signature because `--test` doesn't activate a RW session, which is a hardcoded requirement in pkcs11-tool.c for signatures.

This PR activates an RW session when testing the card. However, I'm not completely sure whether that's the correct fix. Reading PKCS#11, it isn't really clear to me which functions (e.g. signature/decryption) are available in an RW/RO session. The standard only talks about objects that are available in both cases, not about functions.

Maybe you, @dengert @Jakuje, have some better insight...

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -777,7 +777,7 @@ int main(int argc, char * argv[])
 			opt_sig_format = optarg;
 			break;
 		case 't':
-			need_session |= NEED_SESSION_RO;
+			need_session |= NEED_SESSION_RW;
 			do_test = 1;
 			action_count++;
 			break;


dengert:
The intent of the PIV 9E key and cert is to allow the card to authenticate the card (not the user) to things like a physical door lock. Much like a non smart card badge. 

This is related to:  https://github.com/OpenSC/OpenSC/issues/1769#issuecomment-524000247 where I  suggested a possible fix, https://github.com/OpenSC/OpenSC/files/3531147/fix-1747.diff-v2.diff.txt but put the wrong issue number in the file name rather than 1769.

The patch allows pkcs11-tool to use the 9E key (PKCS\#11 ID=04)  when  the `pkcs11-tool  --id xxxx` parameter is set by looking for the object in a `CKS_RW_PUBLIC_SESSION`.  If not found, the C_Login is done which changes the session to a  becomes a `CKS_RW_USER_FUNCTIONS`  where all the keys including the one that does not require the login will show up.  The overhead is  an extra call to  `find_object` to look for a specific object before the login and following call to `find_object` 

You said: "Reading PKCS#11, it isn't really clear to me which functions (e.g. signature/decryption) are available in an RW/RO session."  It is probably more of  a  `CKS_RW_PUBLIC_SESSION` vs `CKS_RW_USER_FUNCTIONS` session. The PIV card does not not require a PIN login.


(I am on vacation for the next two week, but can read e-mail.)

frankmorgner:
I know about the purpose of this key/certificate, which is not the problem. It's a standard X.509 certificate with an associated private key so this should be usable from any PKCS#11 application.

For the Yubikey a single PKCS#11 slot is created for its single User PIN, hence the slot gets the flag `CKF_LOGIN_REQUIRED`. For some reason, I don't understand, however, the 9E key/certificate is associated with exactly this slot (instead of creating a new separate slot), which make the PKCS#11 application believe that it needs to verify the PIN before using the 9E key. I don't see how your suggested patch could fix this situation...

dengert:
On vacation. Will look at this tonight.

mouse07410:
Would it make more sense mark with `CKF_LOGIN_REQUIRED` an individual key instead of the entire slot? 

For PIV applications, it matters that several keys/certs belong to the same slot (aka are on the same token, and can be referred by the same token/slot address - with different key IDs).

frankmorgner:
No need to hurry, enjoy your vacation :sunrise:

My overall idea is broken anyway, since Apple's/Microsoft's default implementation are requesting the PIN even for 9E key anyway...

dengert:
The mod addresses the specific problem the user was having. It takes advantage of the difference between `CKS_RW_PUBLIC_SESSION`  and `CKS_RW_USER_FUNCTIONS`. Non PIN protected objects can be seen with a `CKS_RW_PUBLIC_SESSION`  But `pkcs11-tool` assumes a crypto operation always needs a PIN and does a `C_Login`.  The mod will skip the ``C_Login` if the specific key can be seen in the  `CKS_RW_PUBLIC_SESSION`

The question is, does OpenSC implement the correct PKCS\#11 behavior?

PR #1791, opened 2019-09-12
cardos5: fix reading of serial number

akemnade:
The corresponding GET DATA command only returns the serial,
nothing else.
Tested with CardOS 5.0 and 5.3 cards. The serial number
is the same as shown with other tools

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -1190,14 +1190,19 @@ static int cardos_get_serialnr(sc_card_t *card, sc_serial_number_t *serial)
 	LOG_TEST_RET(card->ctx, r,  "APDU transmit failed");
 	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
 		return SC_ERROR_INTERNAL;
-	if (apdu.resplen != 32) {
+	if ((apdu.resplen == 8) && (card->type == SC_CARD_TYPE_CARDOS_V5_0)) {
+		/* cache serial number */
+		memcpy(card->serialnr.value, rbuf, 8);
+		card->serialnr.len = 8;
+	} else if (apdu.resplen == 32) {
+		/* cache serial number */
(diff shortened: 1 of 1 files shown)


Jakuje:
Looks good to me. Verified with my CardOS 5.3 card that it returns 8B serial number for me (was failing before).

As already mentioned in the referenced issue, the travis issue is most probably unrelated to your changes..

Verified with
```
$ ./src/tools/opensc-tool --serial -r 1
02 04 B8 2D 00 0D 17 31 ...-...1
$ opensc-tool -s 00CA018100 -r 1
Sending: 00 CA 01 81 00 
02 04 B8 2D 00 0D 17 31 ...-...1
```
The `pkcs11-tool -L` is still using serial number from tokenInfo, which is different. This can be a bit confusing, but I am not sure what is the preferred serial number to use in this case inside of the pkcs11 structures:
```
Slot 1 (0x4): OMNIKEY AG CardMan 3121 01 00
  token label        : ** PKI Card (Card PIN)
  token manufacturer : ***
  token model        : PKCS#15
  token flags        : login required, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 06760082
  pin min/max        : 6/6
```
And debug log:
```
 asn1.c:1614:asn1_decode: Looking for 'serialNumber', tag 0x4, OPTIONAL
 asn1.c:1404:asn1_decode_entry:  decoding 'serialNumber', raw data:06760082
```

akemnade:
token serial vs. card serial: some software sets a different random number in the tokeninfo, and assigns a new one everytime the token is re-initialized. But the situation is same as with Atos 4.4.
But there were serious problems with  OpenSCToken if there is so card serial:
see
https://github.com/frankmorgner/OpenSCToken/pull/10

frankmorgner:
thanks

PR #1795, opened 2019-09-16
sc-hsm: Fix maximum APDU size to 4K version (Fixes #1794)

### LINKED ISSUE #1794: sc-hsm: Writing a large certificate fails with CKR_INVALID_DATA
C_CreateObject() fails with CKR_INVALID_DATA when writing a large certificate (e.g. >1K) to a SmartCard-HSM 4K (3.2).

Apparently the maximum send size is set for previous versions, however the 4K version supports only 1232 byte APDUs.

Use the test suite from the sc-hsm-workspace available at the CDN to reproduce the issue.

CardContact:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-sc-hsm.c
+++ b/src/libopensc/card-sc-hsm.c
@@ -1686,7 +1686,13 @@ static int sc_hsm_init(struct sc_card *card)
 	}
 	sc_file_free(file);
 
-	card->max_send_size = 1431;		// 1439 buffer size - 8 byte TLV because of odd ins in UPDATE BINARY
+	// APDU Buffer limits
+	//   JCOP 2.4.1r3    1462
+	//   JCOP 2.4.2r3    1454
+	//   JCOP 3          1232
+	//   Reiner SCT      1014
+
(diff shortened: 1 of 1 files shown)


frankmorgner:
A better way of encoding the card's limit would be to use the "Extended length information" in the FMD when selecting the applet. (By the way, the reader's limits should be handled in reader-pcsc.c if needed):
grafik

CardContact:
Yes, I know. But we need this to be a little bit more flexible. With the upcoming MicroSD card we need even smaller APDU sizes compared to the limit imposed by the applet.

So currently we plan to have a file that can be written during initialization.

In the meantime we are happy with the 4K buffer size, even for older devices that support slightly more.

frankmorgner:
The current idea uses a hard limit regardless of the cards properties, that's anything but flexible. If it's impossible to know the actual limits, this part of the FMD could be configured during pre-personalization. However, I'm fine with your suggested short term solution if it solves the problems.

PR #1797, opened 2019-09-17
simplify PIV I/O

frankmorgner:
This commit simplifies a bit of code and memory allocations. The sizes of the static buffers may be reworked, however (currently they are chosen to match what's currently used). Maybe, @dengert you want to have a look.

##### Checklist

- [ ] macOS tokend is tested


dengert:
The PR looks OK. But I can not test this till next week. 

Are you planing on adding this to 0.20.0? I would hold off as it does not fix any known problem.

frankmorgner:
Sure, I would only include it when the release is done.

frankmorgner:
applied in cfd5519b981c1aeb8f150a13b0f116fb281c8782

PR #1813, opened 2019-10-01
Fix issues reported by oss-fuzz

Jakuje:
This addresses several issues reported by the fuzzer.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -513,6 +513,7 @@ pgp_init(sc_card_t *card)
 		/* explicitly get the full aid */
 		r = get_full_pgp_aid(card, file);
 		if (r < 0) {
+			sc_file_free(file);
 			pgp_finish(card);
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
 		}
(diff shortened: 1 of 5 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -577,8 +578,10 @@ pgp_init(sc_card_t *card)
 	sc_format_path("3f00", &file->path);
 
 	/* set up the root of our fake file tree */
+	/* Transfers ovnership of the file to the priv->mf structure */
```
frankmorgner: typo

Jakuje: Fixed

Jakuje:
> Regarding the file size, I would not impose file size limitations everywhere. this would be a lot of work with not much benifit. I think it would be enough if we do a proper error handling when a subsequent `malloc()` fails.

I did not meant everywhere, but where the size is directly read from some of the structures, the size can reach insane values and when it is causing troubles to, for example, fuzzer. As in the top commit.  Indeed, in this case, the failure is correctly handled, but allocating this large files as read from smart card is an error on its own.

PR #1816, opened 2019-10-03
Minidriver hScard changed Posible fix for #1763

### LINKED ISSUE #1763: SC-HSM: Creating private key for ADCS and other certificates fails w/ SCARD_E_READ_ONLY_CARD when using OpenSC CSP or MS Smart Card KSP
### Problem Description

I was trying to set up ADCS using a Nitrokey HSM 2. When trying to use it to create a certificate for Active Directory Certificate Services using Windows Server 2019, the process fails with error SCARD_E_READ_ONLY_CARD. This error occurs choosing OpenSC CSP/sha1/2048 (#1749 related) as well as RSA#Microsoft Smart Card Key Storage Provider/sha256/2048.

For testing, I use the 0.19-derived OpenSC build mentioned in #1688 because otherwise I'd have the same issues mentioned in this issue (read only smart card which is not selectable).

### Proposed Resolution
I see that it is very much possible ADCS will never work using SC-HSM or with OpenSC in general because of reasons only Microsoft can tell us. I've seen that ADCS support is explicitly not advertised for the NK HSM2. Anyway, it would be nice if the issues could be worked around/needed features implemented or I could just get some background information why ADCS cannot work with smart cards, as I'm interested 😉.  I also tried to test this with a YubiKey 5 and Yubico's Minidriver which will also not work (although key generation and the setup process run fine or at least complete).

### Steps to reproduce
Initialize token (this one was actually initialized using dkek shares). Install Windows Server, Active Directory Certificate Services and the OpenSC build from #1688 .
Configure OpenSC like shown below and change the ATR for SC-HSM inside the registry.
Try to go through the setup assistant, choosing to generate a new CA with a new private key.
Choose OpenSC CSP or MS SC KSP. Allow Administrator Interaction.
Try to complete the assistant by choosing the NitroKey HSM smart card twice (note that there's no PIN request at all)

edit: This error is not specific to ADCS and can be easily reproduced on every recent Windows machine by running the New-SelfSignedCertificate PowerShell cmdlet (you may need to have administrative privileges):

```powershell
New-SelfSignedCertificate -DnsName "www.example.com" -CertStoreLocation cert:\LocalMachine\My" -Provider "OpenSC CSP"

New-SelfSignedCertificate -DnsName "www.example.com" -CertStoreLocation cert:\LocalMachine\My" -Provider "Microsoft Smart Card Key Storage Provider"
```

### Logs
Configuration (ATR was updated in registry as well):
```
app default {
	debug = 3;
	debug_file = c:/tmp/opensc-debug.txt;
	framework pkcs15 {
		# use_file_caching = true;
	}
	# SmartCard-HSM 4k with contact-based interface or USB-Stick
	card_atr 3b:de:18:ff:81:91:fe:1f:c3:80:31:81:54:48:53:4d:31:73:80:21:40:81:07:1c {
		driver = "sc-hsm";
		md_read_only = false;
		md_supports_X509_enrollment = true;
		md_supports_container_key_gen = true;
		md_guid_as_label = true;
	}
}
```
MS SC KSP log: https://gist.github.com/marcoh00/567331f26e0bbda9991031427367e469
OpenSC CSP log: https://gist.github.com/marcoh00/a4b90ebdf55f64331da54e5523e0887e

dengert:
If the hScard changes but hSCardCtx does not don't do call
reinit_card_for.

May fix #1763 because reinit_card_for will lose some data.

This is first cut. I don't have a way to test it.

 Changes to be committed:
	modified:   minidriver.c

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -423,6 +423,15 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
 		logprintf(pCardData, 1, "HANDLES CHANGED from 0x%08X 0x%08X\n",
 			  (unsigned int)vs->hSCardCtx,
 			  (unsigned int)vs->hScard);
+		if (pCardData->hSCardCtx == vs->hSCardCtx) {
+			/*  only hScard changed, set the provided reader and card handles into ctx */
+			vs->hScard = pCardData->hScard;
+			r = sc_ctx_use_reader(vs->ctx, &vs->hSCardCtx, &vs->hScard);
+			if (r != SC_SUCCESS) {
+				logprintf(pCardData, 0, "sc_ctx_use_reader() failed with %d\n", r);
+				return SCARD_E_COMM_DATA_LOST;
+			}
+		} else
 		return reinit_card_for(pCardData, name);
 	}
 


dengert:
As reported in #1763 this patch does not fix the problem. So I am closing this PR.

PR #1821, opened 2019-10-08
openpgp: Fixes for importing ECC keys

alex-nitrokey:
This is a follow up of #1506 
I wasn't able to test the import of ECC keys back then, therefore I needed to leave the import feature untested (see PR discussions).

The problem was that something in OpenSC did not parse the key that is to be imported correctly. This seems to got fixed meanwhile so that I could test the importing today. 

Some minor code needed to be changed which is documented in this PR. I'd be glad to see this PR merged before release of 0.20 as it enabled importing of ECC for OpenPGP Card devices. I will further test 0.20-rc2 and this PR on various OpenPGP cards tomorrow to make sure everything is fine (thus the `WIP`)


alex-nitrokey:
cc @enleth
please try to test the ECC import

alex-nitrokey:
Okay, I tested the import with `secp256v1` and `brainpoolP384r1` curves. To do that, I created keys with

`openssl ecparam -genkey -name "<curve>" -out myec.key.pem`

Then I created a self-signed certificate corresponding to the key. I imported both with either

`pkcs15-init --delete-objects privkey,pubkey --id 3 --store-private-key myec.p12 --format pkcs12 --auth-id 3 --verify-pin `
or
`pkcs15-init --delete-objects privkey,pubkey --id 3 --store-private-key myec.key.pem --auth-id 3 --verify-pin --id 3 && pkcs15-init --id 3 --store-certificate myec.cert.pem`

The import succeeds without any error and looking at `pkcs15-tool -c -k` everything *seems* to be right. The problem is, I am not sure how to make sure, that the key got imported *correctly*. **Can you give me a hint how I could test the imported key?**

Besides that this PR should work fine and enables the import of ECC keys. I can not reproduce the [last error that happened in Travis](https://travis-ci.org/OpenSC/OpenSC/builds/595688440?utm_source=github_status&utm_medium=notification) and nothing I changed should produce that error as it only affects ecc keys that are not tested. **Is it possible to provoke another Travis test without doing a new commit?**

dengert:
You could test the imported key, by using OpenSSL to sign and verify something with the the myec.key.pem using just software. 

Then using the opensc or the libp11 engine to sign the same something using the imported key on the card.  Then use openssl to verify the output of the opensc or libp11 with the myec.key.pem. 

IIRC The signatures in both ECDSA cases might be different, but verify should work.

src/libopensc/card-openpgp.c
```diff
@@ -719,10 +719,20 @@ pgp_parse_algo_attr_blob(const pgp_blob_t *blob, sc_cardctl_openpgp_keygen_info_
 			/* SC_OPENPGP_KEYALGO_ECDH || SC_OPENPGP_KEYALGO_ECDSA */
 			key_info->algorithm = blob->data[0];
 
+			/* last byte is only set if pubkey import is supported, empty otherwise*/
+			if (blob->data[blob->len] == SC_OPENPGP_KEYFORMAT_EC_STDPUB){
```
frankmorgner: shouldn't this be `blob->len-1`?

src/libopensc/card-openpgp.c
```diff
@@ -719,10 +719,20 @@ pgp_parse_algo_attr_blob(const pgp_blob_t *blob, sc_cardctl_openpgp_keygen_info_
...
+				key_info->u.ec.oid_len = blob->len - 2;
```
frankmorgner: The code needs to check that `blob->len` is at least 2 or 1 respectively

src/libopensc/card-openpgp.c
```diff
@@ -719,10 +719,20 @@ pgp_parse_algo_attr_blob(const pgp_blob_t *blob, sc_cardctl_openpgp_keygen_info_
...
+			else {
+				key_info->u.ec.oid_len = blob->len - 1;
+				key_info->u.ec.keyformat = SC_OPENPGP_KEYFORMAT_EC_STD;
+			}
+
 			sc_init_oid(&oid);
 			/* Create copy of oid from blob */
-			for (j=0; j < (blob->len-1); j++) {
-				oid.value[j] = blob->data[j+1]; /* ignore first byte of blob (algo ID) */
+			for (j=0; j < key_info->u.ec.oid_len; j++) {
```
frankmorgner: Again, length check missing

src/libopensc/card-openpgp.c
alex-nitrokey: good catch!

dengert:
I would like to see on OpenPGP use of the "oid_binary" also called  "binary_oid" be replace by the use of  `r = sc_asn1_decode_object_id(&cxdata[1], cxdata_len-1, &oid);` as is done in: https://github.com/OpenSC/OpenSC/pull/1960/commits/baa9f0021d3fccda1938f7b51e01ffc46fc34d7d in `pkcs15-openpgp.c` The same can be done in card-openpgp.c and https://github.com/OpenSC/OpenSC/commit/e45712bd29ebe405a8ee48aceac751c2f277e164
 And the need for two OIDs in the ec_curve tables would be eliminated.

alex-nitrokey:
Okay I look into it.

alex-nitrokey:
> You may want to add your import procedure to the CI tests by extending .travis.yml:

[ykneo-openpgp does not support ECC keys](https://github.com/Yubico/ykneo-openpgp/issues/12) and thus we can not inlcude it in our tests :(

frankmorgner:
> > You may want to add your import procedure to the CI tests by extending .travis.yml:
> 
> [ykneo-openpgp does not support ECC keys](https://github.com/Yubico/ykneo-openpgp/issues/12) and thus we can not inlcude it in our tests :(

I see, thanks for looking into this anyway.

frankmorgner:
@alex-nitrokey what's the status? did you find the time to look at all comments? the branch still has some conflicts...

alex-nitrokey:
I think I have resolved every comment with my last commits. Github tells me, that the branch has no conflicts. But I can merge current master, if you like.

frankmorgner:
> I think I have resolved every comment with my last commits. Github tells me, that the branch has no conflicts. But I can merge current master, if you like.

Ah, a merge commit will do. I'll have a look at the changes...

alex-nitrokey:
Okay, after merging master it does not build anymore locally on my pc. I am wondering why Travis works though. I try to find out what is happening...

alex-nitrokey:
The message is

```
make[3]: Entering directory '/home/nitrokey/git/opensc/src/OpenSC/src/tools'
  CC       opensc_notify-opensc-notify.o
opensc-notify.c: In function ‘sig_handler’:
opensc-notify.c:192:2: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
  192 |  write(cancellation_fd[1], &sig, sizeof sig);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[3]: *** [Makefile:1262: opensc_notify-opensc-notify.o] Error 1
make[3]: Leaving directory '/home/nitrokey/git/opensc/src/OpenSC/src/tools'
make[2]: *** [Makefile:461: all-recursive] Error 1
make[2]: Leaving directory '/home/nitrokey/git/opensc/src/OpenSC/src'
make[1]: *** [Makefile:604: all-recursive] Error 1
make[1]: Leaving directory '/home/nitrokey/git/opensc/src/OpenSC'
make: *** [Makefile:512: all] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
```

seems to be connected with this commit 2c26b7392d8d0d7e1db2b6143583376758074dda.

alex-nitrokey:
@frankmorgner do you know what might happen here?

frankmorgner:
It's fixed upstream

PR #1826, opened 2019-10-11
Simplify code and card detection

metsma:

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-mcrd.c
+++ b/src/libopensc/card-mcrd.c
@@ -51,9 +51,7 @@ static const struct sc_atr_table mcrd_atrs[] = {
 	{NULL, NULL, NULL, 0, 0, NULL}
 };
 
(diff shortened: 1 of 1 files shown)


src/libopensc/card-mcrd.c
```diff
@@ -26,15 +26,15 @@
...
-
 #include "internal.h"
 #include "asn1.h"
 #include "cardctl.h"
 #include "gp.h"
 
+#include <ctype.h>
+#include <string.h>
+#include <stdlib.h>
+
```
Jakuje: Why is this moved below? It is a good practice to have system includes on the top and local on the bottom, as it is in the rest of the code base.

metsma: Reverted,  but I follow the C++ guidlines:
* corresponded header file
* necessary project headers
* 3rd party libraries headers
* standard libraries headers
* system headers

PR #1828, opened 2019-10-15
Use onepin-opensc-pkcs11.so for Chromium

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-register.c
+++ b/src/tools/pkcs11-register.c
@@ -118,7 +118,7 @@ get_next_profile_path(const char **profiles_ini, const char *home, const char *b
 
 void
 add_module_pkcs11_txt(const char *profile_dir,
-		const char *module_path, const char *module_name)
+		const char *module_path, const char *module_name, const char *exclude_module_path)
 {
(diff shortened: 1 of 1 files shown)


src/tools/pkcs11-register.c
```diff
@@ -136,7 +136,8 @@ add_module_pkcs11_txt(const char *profile_dir,
 	p[pkcs11_txt_len] = '\0';
 	pkcs11_txt = p;
 
-	if (!strstr(pkcs11_txt, module_path)) {
+	if (!strstr(pkcs11_txt, module_path)
+			&& (exclude_module_path && !strstr(pkcs11_txt, exclude_module_path))) {
```
jimmytheneutrino: Should it be `&& (!exclude_module_path || !strstr(pkcs11_txt, exclude_module_path))` instead?

Because it seems that currently any invocation with an explicit command line argument will do nothing.

frankmorgner: Yes, you're right, thanks for the review

frankmorgner:
thanks for the feedback

PR #1829, opened 2019-10-15
pkcs11: Support for RSA PSS padding in verify

oskarwiksten:
* Explicitly copies the mechanism parameters during a PKCS#11 `C_VerifyInit`
  operation.
* Resolves issues where the calling application deallocates the `pParameter`
  pointer in the `CK_MECHANISM` struct between calls to `C_VerifyInit` and
  `C_Verify`.
* This commit copies the same fix that was applied to `sc_pkcs11_sign_init` in
  commit e5707b545e5a2dc33b0ca52a8bf63f36f71b3d85 for supporting RSASSA-PSS.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -566,6 +566,12 @@ sc_pkcs11_verif_init(struct sc_pkcs11_session *session, CK_MECHANISM_PTR pMechan
 		return rv;
 
 	memcpy(&operation->mechanism, pMechanism, sizeof(CK_MECHANISM));
+	if (pMechanism->pParameter) {
+		memcpy(&operation->mechanism_params, pMechanism->pParameter,
+			pMechanism->ulParameterLen);
+		operation->mechanism.pParameter = &operation->mechanism_params;
+	}
+
 	rv = mt->verif_init(operation, key);
 
 	if (rv != CKR_OK)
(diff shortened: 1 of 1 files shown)


Jakuje:
Thank you for your contribution. The change looks good to me. I see the same code is already used in the SignInit function. Would it make sense to modify also the encrypt/decrypt functions the same way, as they support the OAEP with parameters?

oskarwiksten:
Thanks for reviewing, @Jakuje . Sure, that sounds like a good idea - I guess the same issue could also happen when using those functions. Would you prefer that being done this same pr, or would you like to open a different pr for that?

Jakuje:
It is the same change for the other functions so I would probably do it in the same one, but it is really up to you.

oskarwiksten:
Ok, thanks for the feedback @Jakuje . Have pushed an updated commit onto this branch now, that makes the same change to `C_DecryptInit`.

I did not update `C_EncryptInit` since that does not seem to be supported yet? https://github.com/OpenSC/OpenSC/blob/3a1bd36e2018e7e4c7f78b1fa84d998a9d163616/src/pkcs11/pkcs11-object.c#L822

I have tested the change for `C_VerifyInit`, and verified that it resolves the issue we were seeing. The setup available here does not enable me to verify the change to `C_DecryptInit` however.

Jakuje:
Thank you. You are right, the encrypt is still not implemented so it does not apply here. This looks fine to me.

oskarwiksten:
@Jakuje is there anything else I need to do on this, or can this be merged for possible inclusion in 0.20.0-rc4?

Jakuje:
It is fine as it is and it should get into 0.20.0 release, but it needs to be merged by @frankmorgner or @dengert

PR #1830, opened 2019-10-17
Introduce unit tests and address more oss-fuzz issues  (mostly ASN1 parser)

Jakuje:
**Update**: Mostly introduction of unit tests is left after most of the issues got merged.

Finally, I took a time to check the ASN1 issues reported by oss-fuzz. In the end, I set up first unit tests to make sure the corner cases that I was handling really fix the issue I was fixing and not break any other use cases. Any comments or concerns are welcomed.

By this, I would like to encourage everyone who will be writing and fixing some (at least the low-level) code, to write some unit tests. It should be pretty straight forward, but if you will have some issues with that, please, let me know. It is reusing the cmocka testing framework used in p11tests now.

##### Checklist
- [ ] macOS tokend is tested


frankmorgner:
thanks especially for finding the issue with negative integers!

I've cherry-picked almost all of the fixes, leaving the unittest for the next release.

Jakuje:
Thank you. I will rebase the branch and update the description.

Jakuje:
@frankmorgner It turned out that the fixes were not perfect from the last oss-fuzz scan. We do not have results from the testcases that these fixes are addressing, but it turned out there were few new, which I hope are addressed by the most recent commits.

Jakuje:
@frankmorgner I think I got down to all the issues that are reported up to date. I would like to ask for review, most importantly about the change that I made to the fuzzer code, which I think causes several of the weird reports.

frankmorgner:
Thanks, I've cherry-picked the fixes again. This fuzzing stuff is great, but let's hope we can get some rest now to make the new release :wink:

frankmorgner:
I've applied the last two commits!

src/tests/unittests/asn1.c
```diff
@@ -0,0 +1,60 @@
...
+ * the too-large onese for now, rather than causing undefined overflow.
+ *
+ * https://oss-fuzz.com/testcase-detail/5673497895895040
+ */
+static void torture_large_oid(void **state)
+{
+	/* 2.5.4.18446744073709551619 (The last part is 64 bit overflow) */
+	u8 data1[] = {0x55, 0x04, 0x82, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x03};
+	/* 2.18446744073709551621.4.3 (The second part is 64 bit overflow) */
+	u8 data2[] = {0x82, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x55, 0x04, 0x03};
```
tomato42: those are two test scenarios, shouldn't they be separate test cases?
why 18446744073709551619 (0x10000000000000003) and not 18446744073709551616 (0x10000000000000000)? same for 18446744073709551621

shouldn't we have a test with 18446744073709551615 (0xffffffffffffffff) to verify that they _do_ work?

src/tests/unittests/asn1.c
```diff
@@ -47,12 +47,91 @@ static void torture_large_oid(void **state)
...
+	int value;
+	int rv = 0;
+
+	rv = sc_asn1_decode_integer(zero, sizeof(zero), &value);
+	assert_int_equal(rv, SC_SUCCESS);
+	assert_int_equal(value, 0);
+
+	rv = sc_asn1_decode_integer(one, sizeof(one), &value);
+	assert_int_equal(rv, SC_SUCCESS);
+	assert_int_equal(value, 1);
```
tomato42: again, separate scenarios, so they should be separate test cases

src/tests/unittests/asn1.c
```diff
@@ -47,12 +47,91 @@ static void torture_large_oid(void **state)
...
+		rv = sc_asn1_decode_integer(max4, sizeof(max4), &value);
+		assert_int_equal(rv, SC_SUCCESS);
+		assert_int_equal(value, 2147483647);
+
+		rv = sc_asn1_decode_integer(min4, sizeof(min4), &value);
+		assert_int_equal(rv, SC_SUCCESS);
+		assert_int_equal(value, -2147483648);
+
+		rv = sc_asn1_decode_integer(over, sizeof(over), &value);
+		assert_int_equal(rv, SC_ERROR_INVALID_ASN1_OBJECT);
```
tomato42: there really isn't a function that can handle integers bigger than INT_MAX? those methods won't be able to decode an ECDSA signature made with P-160, let alone anything actually secure... (or serial number in a typical X.509 certificate, or an RSA public key, etc.)

src/tests/unittests/asn1.c
```diff
@@ -47,12 +47,91 @@ static void torture_large_oid(void **state)
...
+{
+	/* Without the Tag and Length {0x02, 0x01} */
+	u8 zero[] = {0x00};
+	u8 one[] = {0x01};
+	u8 minus_one[] = {0xFF};
+	u8 max2[] = {0x7F, 0xFF};
+	u8 min2[] = {0x80, 0x00};
+	u8 max4[] = {0x7F, 0xFF, 0xFF, 0xFF};
+	u8 min4[] = {0x80, 0x00, 0x00, 0x00};
+	u8 over[] = {0x7F, 0xFF, 0xFF, 0xFF, 0xFF};
```
tomato42: negative test cases missing:
zero encoded with no bytes: `{};`
zero encoded with multiple bytes: `0x00, 0x00`
one encoded with multiple bytes: `0x00, 0x01`
minus one encoded with multiple bytes: `0xff, 0xff`

src/tests/unittests/asn1.c
tomato42: also, it's it more of a overflow rather than invalid asn1 object per-se?

src/tests/unittests/asn1.c
tomato42: but then we should also have a test case to verify that values padded so that they don't look negative, are accepted, as in: `0x00, 0x80` is decoded to 128 but `0x00, 0x7f` causes an error

src/tests/unittests/asn1.c
```diff
@@ -47,6 +47,61 @@ static void torture_large_oid(void **state)
...
+	/* Small OIDs */
+	u8 small[] = {0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
+	/* Limit what we can fit into the first byte */
+	u8 limit[] = {0x7F};
+	/* The second octet already oveflows to the second byte */
+	u8 two_byte[] = {0x81, 0x00};
+	/* Missing second byte, even though indicated with the first bit */
+	u8 missing[] = {0x81};
+	/* Missing second byte, even though indicated with the first bit */
+	u8 ecpubkey[] = {0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x02, 0x01};
```
tomato42: the comment doesn't match the value - this is a well-formed encoding of an (1, 2, 840, 10045, 2, 1) OID

src/tests/unittests/asn1.c
```diff
@@ -47,6 +47,61 @@ static void torture_large_oid(void **state)
...
+{
+	/* (without the tag and length {0x06, 0x06}) */
+	/* Small OIDs */
+	u8 small[] = {0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
+	/* Limit what we can fit into the first byte */
+	u8 limit[] = {0x7F};
+	/* The second octet already oveflows to the second byte */
+	u8 two_byte[] = {0x81, 0x00};
+	/* Missing second byte, even though indicated with the first bit */
+	u8 missing[] = {0x81};
```
tomato42: Something like this, but for the 3rd subidentifier may be also interesting

src/tests/unittests/asn1.c
```diff
@@ -47,6 +47,61 @@ static void torture_large_oid(void **state)
...
+	assert_int_equal(oid.value[4], 4);
+	assert_int_equal(oid.value[5], 5);
+	assert_int_equal(oid.value[6], 6);
+	assert_int_equal(oid.value[7], -1);
+
+	rv = sc_asn1_decode_object_id(limit, sizeof(limit), &oid);
+	assert_int_equal(rv, SC_SUCCESS);
+	assert_int_equal(oid.value[0], 2);
+	assert_int_equal(oid.value[1], 47);
+	assert_int_equal(oid.value[2], -1);
```
tomato42: again, separate test scenarios in the same unit test

src/tests/unittests/asn1.c
```diff
@@ -114,42 +131,80 @@ static void torture_integer(void **state)
...
+	free(buf);
 
 	rv = sc_asn1_decode_integer(one, sizeof(one), &value);
 	assert_int_equal(rv, SC_SUCCESS);
 	assert_int_equal(value, 1);
+	rv = asn1_encode_integer(value, &buf, &buflen);
+	assert_int_equal(rv, SC_SUCCESS);
+	assert_int_equal(buflen, sizeof(one));
+	assert_memory_equal(buf, one, buflen);
+	free(buf);
```
tomato42: test cases like this really are asking for parametrisation, there is a lot of code repetition here

src/libopensc/asn1.c
```diff
@@ -891,10 +891,13 @@ sc_asn1_encode_object_id(u8 **buf, size_t *buflen, const struct sc_object_id *id
...
-			if (k > 39)
+			if (k > 39 && id->value[0] < 2) {
 				return SC_ERROR_INVALID_ARGUMENTS;
-			*p++ += k;
-			break;
+			}
+			/* We can encode larger IDs to multiple bytes
+			 * similarly as the following IDs */
+			k += *p;
+			/* fall through */
```
tomato42: test coverage for this?

src/tests/unittests/asn1.c
```diff
@@ -234,6 +234,45 @@ static void torture_negative_int(void **state)
...
+ *
+ * In this example, invalid "unused bytes" value was used
+ */
+static void torture_bit_field(void **state)
+{
+	/* Without the Tag and Length {0x03, 0x02} */
+	u8 data0[] = {0x07, 0x80};
+	/* Without the Tag and Length {0x03, 0x05} */
+	/* u8 data1[] = {0x20, 0x20, 0xff, 0xff, 0xff}; original data */
+	u8 data1[] = {0x20, 0xff, 0xff, 0xff, 0xff};
```
tomato42: test case with bits set but non-zero truncate missing, something like this:
`{0x01, 0xff}` for negative and `{0x01, 0xfe` for positive test

src/libopensc/card-cac1.c
```diff
@@ -69,17 +69,16 @@ static int cac_cac1_get_certificate(sc_card_t *card, u8 **out_buf, size_t *out_l
...
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 	/* get the size */
 	size = left = *out_buf ? *out_len : sizeof(buf);
 	out_ptr = *out_buf ? *out_buf : buf;
 	sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, CAC_INS_GET_CERTIFICATE, 0, 0 );
-	next_len = MIN(left, 100);
-	for (; left > 0; left -= len, out_ptr += len) {
-		len = next_len;
+	len = MIN(left, 100);
+	for (; left > 0;) { /* Increments for readability in the end of the function */
```
tomato42: shouldn't it then be a `while() {}` loop?

tomato42:
@Jakuje could you explain what "pkcs15*: Fail if there is no fallback file_app" is attempting to do?

src/tests/unittests/asn1.c
Jakuje: If I see right, there are arbitrary large integers handled as octet strings. Other primitives are usually not handled in this asn1 library and go through the OpenSSL.
Good point about the error. I changed that to `NOT_SUPPORTED` to better reflect reality.

src/tests/unittests/asn1.c
Jakuje: Thank you. This was not handled in the code either. Added the testcase too.

Jakuje:
> @Jakuje could you explain what "pkcs15*: Fail if there is no fallback file_app" is attempting to do?

This is trying to address oss-fuzz issues, from the [fuzz_pkcs15_decode](https://github.com/OpenSC/OpenSC/blob/master/src/tests/fuzzing/fuzz_pkcs15_decode.c) fuzzer, which `p15card` without the `app_data`. Previous attempts to fix it did not turn out working that well.

@frankmorgner I tried to build the opensc again the the current container provided by oss-fuzz to try to reproduce some of the recent reports and I noticed one more issue in the fuzzer itself. I did not see any new email from oss-fuzz for few days (at least not referencing any new revisions) so I am wondering whether the oss-fuzz can even build the fuzzers now. But if it could not, I assume they would notify us somehow ...

src/tests/unittests/asn1.c
Jakuje: Here, you mean encode/decode as separate scenarios or different values/errors? If the first, I wanted to avoid having the "magic numbers" all over the place copied.

src/tests/unittests/asn1.c
Jakuje: *bug

src/tests/unittests/asn1.c
tomato42: I mean that
```

	rv = sc_asn1_decode_integer(zero, sizeof(zero), &value);
	assert_int_equal(rv, SC_SUCCESS);
	assert_int_equal(value, 0);
	rv = asn1_encode_integer(value, &buf, &buflen);
	assert_int_equal(rv, SC_SUCCESS);
	assert_int_equal(buflen, sizeof(zero));
	assert_memory_equal(buf, zero, buflen);
	free(buf);
```
is a single test case
then 
```

	rv = sc_asn1_decode_integer(one, sizeof(one), &value);
	assert_int_equal(rv, SC_SUCCESS);
	assert_int_equal(value, 1);

	rv = asn1_encode_integer(value, &buf, &buflen);
	assert_int_equal(rv, SC_SUCCESS);
	assert_int_equal(buflen, sizeof(one));
	assert_memory_equal(buf, one, buflen);
	free(buf);
```
is a second test case, and so on...

src/tests/unittests/asn1.c
tomato42: so a macro would still work, wouldn't it?

src/tests/unittests/asn1.c
tomato42: > If I see right, there are arbitrary large integers handled as octet strings.

the problem with this is that make the implementation accept malformed ones (like ones that use too many bytes to encode a value)

tomato42:
> This is trying to address oss-fuzz issues, from the fuzz_pkcs15_decode fuzzer, which p15card without the app_data. Previous attempts to fix it did not turn out working that well.

I get this much, but the message and code is rather cryptic. What's the issue it's fixing? What special case was not handled?

src/tests/unittests/asn1.c
tomato42: yes, for signed int 32 bit type, 0x7fffffff and 0x80000000 are the values we want to test as the one working and the one outside supported range

src/libopensc/asn1.c
```diff
@@ -807,10 +825,45 @@ sc_asn1_decode_object_id(const u8 *inbuf, size_t inlen, struct sc_object_id *id)
...
+		while (inlen && *p & 0x80) {
+			/* Limit the OID values to int size and do not overflow */
+			if (value > (UINT_MAX>>7)) {
+				sc_init_oid(id);
+				return SC_ERROR_NOT_SUPPORTED;
+			}
+			p++;
+			value <<= 7;
+			value |= *p & 0x7F;
+			inlen--;
```
tomato42: this looks like code duplication with the below code

wouldn't it be simpler to decode all subidentifiers and then split the first one into two and insert new values to the array? less code, less special cases = less potential bugs

src/tests/unittests/asn1.c
```diff
@@ -0,0 +1,379 @@
...
+	/* The second octet already oveflows to the second byte */
+	struct sc_object_id two_byte_oid = {{2, 48, -1}};
+	u8 two_byte[] = {0x81, 0x00};
+	/* Existing OID ec publickey */
+	struct sc_object_id ecpubkey_oid = {{1, 2, 840, 10045, 2, 1, -1}};
+	u8 ecpubkey[] = {0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x02, 0x01};
+	/* Missing second byte, even though indicated with the first bit */
+	u8 missing[] = {0x81};
+	/* Missing second byte in later identifiers */
+	u8 missing_second[] = {0x2A, 0x48, 0x81};
```
tomato42: what about non-minimal encoding of the subidentifier? something like `{0x2a, 0x80, 0x01}` (as an incorrect encoding of (1, 2, 1)) or `{0x80, 0x01}`

Jakuje:
@frankmorgner The error is actually now reported and shown on the badge on the main page:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18701

Jakuje:
> > This is trying to address oss-fuzz issues, from the fuzz_pkcs15_decode fuzzer, which p15card without the app_data. Previous attempts to fix it did not turn out working that well.
> 
> I get this much, but the message and code is rather cryptic. What's the issue it's fixing? What special case was not handled?

Originally, the `p15card->file_app` was dereferenced without any check in the code in several places. First attempt to fix it was to skip this in case of it is undefined, which resulted in using some other variables that should have been allocated by the call to `sc_pkcs15_make_absolute_path` from `file_app`.

frankmorgner:
> @frankmorgner The error is actually now reported and shown on the badge on the main page:
> 
> https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18701

thanks for the hint, I've pushed a change to fix this.

src/libopensc/asn1.c
Jakuje: True. I will simplify that.

src/tests/unittests/asn1.c
Jakuje: This is the  problem that the `asn1_decode_integer()` is used also from other places for parsing the size in less strict inputs. Enforcing the format in dbba826f already broke the functional tests so I will try to modify the code to be less strict about these kinds of errors if asked through argument.

src/tests/unittests/asn1.c
Jakuje: Already in as `data3` to `data6`.

src/tests/unittests/asn1.c
Jakuje: I think this got fixed on the way, even though I did not comment on this particular issue yet.

src/tests/unittests/asn1.c
tomato42: both DER and BER encodings require that there are no padding bytes in INTEGER encoding, so if that's too strict, I'd say that those specific usages should use a less strict parser rather than making the default parser non-compliant

src/tests/unittests/asn1.c
Jakuje: Resolved in the current branch

src/tests/unittests/asn1.c
Jakuje: I will add these

src/libopensc/pkcs15-pubkey.c
```diff
@@ -311,6 +311,10 @@ int sc_pkcs15_decode_pukdf_entry(struct sc_pkcs15_card *p15card,
 	memset(gostr3410_params, 0, sizeof(gostr3410_params));
 
 	r = sc_asn1_decode(ctx, asn1_pubkey, *buf, *buflen, buf, buflen);
+	if (r < 0) {
+		free(der->value);
```
frankmorgner: Has `der->value` reallly been reported as leaking? It's effectively [`obj->content`](https://github.com/OpenSC/OpenSC/blob/45d7338351b4e4e23b66a89c4801d959118066aa/src/libopensc/pkcs15-pubkey.c#L220), which should be freed [here](https://github.com/OpenSC/OpenSC/blob/8fea658fe9b5bdf51de7e321a1a549056ba4b150/src/tests/fuzzing/fuzz_pkcs15_decode.c#L61)...

Jakuje: The reports are too vague and reproducing it in docker does not always yield the expected results.

It is freed only if the `obj->type` is correctly set, which is not the case if we fail here. Type is set just few lines below this.

src/libopensc/pkcs15-pubkey.c
```diff
@@ -311,6 +311,10 @@ int sc_pkcs15_decode_pukdf_entry(struct sc_pkcs15_card *p15card,
...
+		free(info->subject.value);
```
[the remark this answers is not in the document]
Jakuje: You are probably right here. Thank you for review. I will drop this one.

src/libopensc/asn1.c
```diff
@@ -577,6 +577,9 @@ static int decode_bit_string(const u8 * inbuf, size_t inlen, void *outbuf,
 
 	if (inlen < 1)
 		return SC_ERROR_INVALID_ASN1_OBJECT;
+	/* ITU-T Rec. X.690 8.6.2.2: The number shall be in the range zero to seven. */
+	if ((*in & ~0x07) != 0)
+		return SC_ERROR_INVALID_ASN1_OBJECT;
```
tomato42: doesn't it also specify that if the bitstring is empty, the unused must be zero?

I think we will need also test coveragate to verify it

src/libopensc/asn1.c
```diff
@@ -706,17 +713,28 @@ static int encode_bit_field(const u8 *inbuf, size_t inlen,
 	return encode_bit_string(data, bits, outbuf, outlen, 1);
 }
 
-int sc_asn1_decode_integer(const u8 * inbuf, size_t inlen, int *out)
+int sc_asn1_decode_integer(const u8 * inbuf, size_t inlen, int *out, int strict)
```
tomato42: isn't it technically more of "non_compliant" rather than non "strict"?

DER is DER, the decoding and encoding of an object needs to always produce the same bytes...

src/tests/unittests/asn1.c
```diff
@@ -0,0 +1,275 @@
...
+		rv = decode_bit_field(data, datalen, &value, value_len); \
+		assert_int_equal(rv, error); \
+	}
+/* Without the Tag (0x03) and Length */
+/* Simple value 1 */
+TORTURE_BIT_FIELD(one, "\x07\x80", 1)
+/* This is the last value that can be represented in the unsigned int */
+TORTURE_BIT_FIELD(uint_max, "\x00\xff\xff\xff\xff", UINT_MAX)
+/* Valid padding */
+TORTURE_BIT_FIELD(padding, "\x01\xfe", 127)
```
tomato42: wait, so the decoded bitstring doesn't remember the size of the input string?
so `\x00\x00\x00` and `\x00\x00` are equal?!

Jakuje: Yes. In this case, when we request the decoding to integer type, there is no way how we can remember also the length. This is described in the header of the function in 

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/asn1.c#L663

There is also API to decode the BIT STRINGs into byte fields:

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/asn1.c#L569

src/libopensc/asn1.c
Jakuje: Yes. But it is already used in other places for very similar encoding, sometimes prefixed with zero bytes, or on fixed length bytes, ...). I think it started to be used instead of writing a new parsers for these particular use cases.
About the naming, I would like to avoid using more aggressive names as non_compliant.

src/libopensc/asn1.c
Jakuje: Yes, indeed. I will add that:

> 8.6.2.3 If the bitstring is empty, there shall be no subsequent octets, and the initial octet shall be zero. 

https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf

src/tests/unittests/asn1.c
tomato42: ah, ok then, thanks for the references

src/libopensc/asn1.c
tomato42: while I would like more of a scarlet colour on this "red flag", I guess that will have to be enough :smile:

src/libopensc/pkcs15-pubkey.c
frankmorgner: No, it also gets [`free`d when having an arbitrary type](https://github.com/OpenSC/OpenSC/blob/8fea658fe9b5bdf51de7e321a1a549056ba4b150/src/libopensc/pkcs15.c#L2536).

Jakuje: The backtrace points directly to this place, and the `der->value` is the only ASN1 member of `SC_ASN1_OCTET_STRING` type with `SC_ASN1_ALLOC` flags. I can reproduce it locally (in oss-fuzz container) to get more verbose backtrace:
```
#0  __GI___libc_malloc (bytes=0) at malloc.c:2902
#1  0x00000000006451a3 in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd79e8, obj=0x224273d "\244!\363", objlen=0, depth=5) at asn1.c:1559
#2  0x0000000000642fae in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd79c0, in=0x224273b "\240", len=5, newp=0x7ffeb7bd6000, len_left=0x7ffeb7bd5fe8, choice=1, depth=5) at asn1.c:1727
#3  0x0000000000642caa in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd7940, in=0x224273b "\240", len=5, newp=0x0, len_left=0x0, choice=0, depth=4) at asn1.c:1698
#4  0x0000000000644e4b in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd78f0, obj=0x224273b "\240", objlen=5, depth=3) at asn1.c:1490
#5  0x0000000000642fae in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd78f0, in=0x2242739 "0\005\240", len=10, newp=0x0, len_left=0x0, choice=0, depth=3) at asn1.c:1727
#6  0x0000000000644e4b in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd6d98, obj=0x2242739 "0\005\240", objlen=10, depth=2) at asn1.c:1490
#7  0x0000000000642fae in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd6d20, in=0x2242722 "0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", len=33, newp=0x0, len_left=0x0, choice=0, depth=2) at asn1.c:1727
#8  0x0000000000645f97 in asn1_decode_p15_object (ctx=0x2242830, in=0x2242722 "0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", len=33, obj=0x7ffeb7bd75d0, depth=1) at asn1.c:1411
#9  0x0000000000645568 in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd76d8, obj=0x2242722 "0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", objlen=33, depth=1) at asn1.c:1632
#10 0x0000000000642fae in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd7660, in=0x2242720 "\240!0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", len=262, newp=0x7ffeb7bd7410, len_left=0x7ffeb7bd73f8, choice=1, depth=1)
    at asn1.c:1727
#11 0x0000000000642caa in asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd7610, in=0x2242720 "\240!0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", len=262, newp=0x7ffeb7bd7d98, len_left=0x7ffeb7bd7d90, choice=0, depth=0)
    at asn1.c:1698
#12 0x0000000000642911 in sc_asn1_decode (ctx=0x2242830, asn1=0x7ffeb7bd7610, in=0x2242720 "\240!0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", len=262, newp=0x7ffeb7bd7d98, len_left=0x7ffeb7bd7d90) at asn1.c:1749
#13 0x00000000004da9eb in sc_pkcs15_decode_pukdf_entry (p15card=0x22430a0, obj=0x22434e0, buf=0x7ffeb7bd7d98, buflen=0x7ffeb7bd7d90) at pkcs15-pubkey.c:313
#14 0x00000000004b4fe3 in LLVMFuzzerTestOneInput (Data=0x2242720 "\240!0\004V\001\233\a0\r\004\003\023/\001\003\003\001)\004", Size=262) at fuzz_pkcs15_decode.c:57
#15 0x0000000000445672 in ExecuteCallback () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556
#16 0x0000000000430292 in RunOneTest () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:292
#17 0x0000000000435f4f in FuzzerDriver () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:774
#18 0x00000000004600c3 in main () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19
```
Also the `gdb` confirms we are currently handling the `direct` field of `publicECKeyAttributes`:
```
(gdb) f 1
#1  0x00000000006451a3 in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd79e8, obj=0x224273d "\244!\363", objlen=0, depth=5) at asn1.c:1559
(gdb) p *entry
$1 = {name = 0x6718b2 "direct", type = 4, tag = 553648128, flags = 6, parm = 0x2243fa0, arg = 0x2243fa8}
(gdb) f 4
#4  0x0000000000644e4b in asn1_decode_entry (ctx=0x2242830, entry=0x7ffeb7bd78f0, obj=0x224273b "\240", objlen=5, depth=3) at asn1.c:1490
1490				r = asn1_decode(ctx, (struct sc_asn1_entry *) parm, obj,
(gdb) p *entry
$4 = {name = 0x6746c5 "publicECKeyAttributes", type = 129, tag = 16777232, flags = 0, parm = 0x7ffeb7bd7940, arg = 0x0}
```
Actually, my patch would really not address the issue, since as I step further, the decode worked, but later, I see the following:
```
(gdb) p *der
$10 = {value = 0x2e7bcb0 "x\373\017\b@\177", len = 0}
```
So this looks like it is bigger issue that the length of allocation is not propagated to the calling objects. I will try to have a better look into that.

Jakuje: The important part is this one:
```
#0  __GI___libc_malloc (bytes=0) at malloc.c:2902
```
The asn1.c calls malloc of size 0, which is invalid, but not reported as error in the fuzzers. This is also the reason why it is not freed in cleanups.

Jakuje: @frankmorgner The fixes for this group of problems are already in this branch. I think also for one of the reports from today, which was hitting similar code path.

frankmorgner: malloc(0) is valid and returns either NULL or a valid pointer, i.e. though it's implementation dependent, you can always free the result... I'm not sure why malloc(0) would cause a leaking memory.

frankmorgner:
@Jakuje , we still have [some critical issue with fetching the certificates in CAC](https://oss-fuzz.com/testcase-detail/6392392311898112)... I was hoping to at least fix the security issues soon to finally get the release out, please have a look.

(unfortunately, there is also [this issue](https://oss-fuzz.com/testcase-detail/5755855475834880), for which I have no clue how to fix it. Maybe you can find the time to have a look at this, too.)

src/libopensc/pkcs15-pubkey.c
Jakuje: Yes, it is valid and it returns either of these two. But in case it returns NULL, the following condition would wrongly return out of memory, usually making all program fail, while zero-length bit string is perfectly valid, I think. It is causing memory leak probably only in the fuzzer implementation. But anyway, we should avoid implementation dependent behavior.

Jakuje:
> @Jakuje , we still have [some critical issue with fetching the certificates in CAC](https://oss-fuzz.com/testcase-detail/6392392311898112)... I was hoping to at least fix the security issues soon to finally get the release out, please have a look.

I do not think this is issue anymore after b75c002c -- it was reproduced only once and it will get closed in few days as flaky.

> (unfortunately, there is also [this issue](https://oss-fuzz.com/testcase-detail/5755855475834880), for which I have no clue how to fix it. Maybe you can find the time to have a look at this, too.)

I will check this one. It landed just yesterday.

src/libopensc/pkcs15-pubkey.c
frankmorgner: I see, thanks for the explanation.

src/libopensc/pkcs15-prkey.c
```diff
@@ -258,10 +258,6 @@ int sc_pkcs15_decode_prkdf_entry(struct sc_pkcs15_card *p15card,
 	memset(gostr3410_params, 0, sizeof(gostr3410_params));
 
 	r = sc_asn1_decode_choice(ctx, asn1_prkey, *buf, *buflen, buf, buflen);
-	if (r < 0) {
-		/* This might have allocated something. If so, clear it now */
-		free(info.subject.value);
```
frankmorgner: Hmm, where is this freed? Do you mean `sc_pkcs15_parse_df()`?

Jakuje: My bad. Thank you for double-checking. I though that this function is more similar to the one we talked about previously (`sc_pkcs15_decode_pukdf_entry()`, but now I see this one uses static allocation and then memcpy so this is actually needed.
To avoid later confusion, I would suggest to rewrite either of these functions so they are more similar in the internal semantics, probably the prkey one as it allocates as late as possible, making the whole function more readable and avoiding `goto`. What do you think? I will push a new commit.

Jakuje: After checking a bit more, the functions are still different in spite of allocations of other data so unless you would be strongly for refactoring one of them, I would leave them as they are now.

Jakuje:
I am not sure how to adjust the configure and makefile magic around the code coverate to work with appveyor (not even to not to use it there) so if somebody will be interested in that, I would be glad for pointers.

frankmorgner:
looks like a automake issue rather than an issue with appveyor. See https://github.com/tpm2-software/tpm2-tss/pull/1256/commits/4403f3bb15bd233d5d52a8e09ada71716c9a0465 for a possible fix.

Jakuje:
> looks like a automake issue rather than an issue with appveyor. See [tpm2-software/tpm2-tss@4403f3b](https://github.com/tpm2-software/tpm2-tss/commit/4403f3bb15bd233d5d52a8e09ada71716c9a0465) for a possible fix.

This is already in. It works fine on Linux, but not on windows in the appveyor.

Jakuje:
With the last changes, I think I am again on the bottom of the list of issues now.

frankmorgner:
I currently don't have the capacity to work on this, so thanks again for your help!

frankmorgner:
The memory leak in `sc_pkcs15_decode_prkdf_entry()` still persists, could you check whether https://github.com/OpenSC/OpenSC/commit/d3f1c4314320147168aa8ef73bcada95a2936bd6 is working?

Jakuje:
> The memory leak in `sc_pkcs15_decode_prkdf_entry()` still persists, could you check whether [d3f1c43](https://github.com/OpenSC/OpenSC/commit/d3f1c4314320147168aa8ef73bcada95a2936bd6) is working?

This looks like a good fix. I can test it on monday.

Jakuje:
@frankmorgner I verified the issue is gone with your fix. I pushed your change also to this branch to have it everything in one place.

src/libopensc/gp.h
```diff
@@ -32,9 +32,34 @@
...
+	u8 packing_data[2];
+	u8 icc_manufacturer[2];
+	u8 ic_embedding_data[2];
+	u8 pre_personalizer[2];
+	u8 ic_pre_personalization_data[2];
+	u8 ic_pre_personalization_id[4];
+	u8 ic_personalizaer[2];
+	u8 ic_personalization_data[2];
+	u8 ic_personalization_id[4];
+} global_platform_cplc_data_t;
```
frankmorgner: the attributes need to be packed, right?

frankmorgner: Ideally, this should return the interpreted data only, i.e. without the ASN.1 overhead...

Jakuje: Yes. The data is read directly into it from the response in APDU. So I believe it would be more safe to define that packet. I will add it in following commits (especially as it is not completely related to the issue that I was fixed).

frankmorgner:
great, thanks. I've cherry-picked the fixes

src/libopensc/gp.h
Jakuje: I don't see this doing anything with ASN.1.

Jakuje:
> great, thanks. I've cherry-picked the fixes

There is at least one issue that is not addressed by the attached patches, but I hope I will get to the root of the problem soon.

Jakuje:
I still could reproduce some of the issues in coolkey (get_attributes) locally, but I was not able to figure out what goes wrong there. The testcases got closed with latest revision for some reason so I assume there will jump up new ones soon.

Jakuje:
@frankmorgner I finally got back to this one. The recent commits should address all the remaining issues in oss-fuzz reported to date and hopefully it would be able to move on. Can you cherry-pick them if they are ok for you?

src/libopensc/gp.h
frankmorgner: What about tag and length? do they nominate something different than an ASN.1 header?

frankmorgner:
looks all good, thank you!

src/libopensc/gp.h
Jakuje: Ok, it is `BER-TLV` response from the GET DATA instruction, but since this one has pretty strict encoding (2B tag, 1B length), I do not think explicit parsing using ASN1 library is needed here. I would just check the length (already done), tag and internal length. See the last commit (rebased on master).

> When the value of P1-P2 lies in the range from ‘4000’ to ‘FFFF’, the value of P1-P2 shall be a BER-TLV tag on two bytes.

Jakuje:
@frankmorgner the recent commits should address two more issues that popped up in fuzzing. Additionally, I created corpuses for simple card detection of cards I have around and feeding them to the fuzzer (after fixing the ATR storing in the fuzzer itself) already reports more crashes further:
```
(gdb) bt
#0  0x00000000004e52d4 in sc_pkcs15_unwrap (p15card=0x2f3d9e0, key=0x2f51060, target_key=0x7fff46ffff08, flags=0, in=0x2f3d155 "S\201\370p\201\360\060\201a\363\365", inlen=10, 
    param=0x2f3d161 "\355\060\201\330\002\001", paramlen=245) at pkcs15-sec.c:397
#1  0x00000000004b5732 in LLVMFuzzerTestOneInput (Data=0x2f3c720 "\022", Size=3153) at fuzz_pkcs15_reader.c:237
#2  0x0000000000445672 in ExecuteCallback () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556
#3  0x0000000000430292 in RunOneTest () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:292
#4  0x0000000000435f4f in FuzzerDriver () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:774
#5  0x00000000004600c3 in main () at /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19
(gdb) p tkey
$1 = (const struct sc_pkcs15_skey_info *) 0x0
(gdb) p *target_key
$3 = {type = 0, label = '\000' <repeats 254 times>, flags = 0, auth_id = {value = '\000' <repeats 254 times>, len = 0}, usage_counter = 0, user_consent = 0, access_rules = {{access_mode = 0, 
      auth_id = {value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {
        value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {value = '\000' <repeats 254 times>, 
        len = 0}}, {access_mode = 0, auth_id = {value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {value = '\000' <repeats 254 times>, len = 0}}, {access_mode = 0, auth_id = {
        value = '\000' <repeats 254 times>, len = 0}}}, data = 0x0, emulated = 0x0, df = 0x0, next = 0x0, prev = 0x0, content = {value = 0x0, len = 0}, session_object = 0}
(gdb) p *target_key->data
Attempt to dereference a generic pointer.
(gdb) p target_key->data
$4 = (void *) 0x0
```
I was not sure whether this is a misuse of the `sc_pkcs15_unwrap()` API or whether it should check for the NULL pointers/uninitialized values as they are passed as arguments. Let me know what you think about this issue.

After we will have some more corpuses (do you have some cards around you can contribute corpuses from?) and the above bug fixed, I will modify the build scripts in oss-fuzz to pick them up.

These are yet only for the card detection, which looks like working fine. Generating right responses for the other steps and operations that you outlined in the fuzzer would be more work, but I would certainly like to have a look into that.

frankmorgner:
Thanks for looking into this again!

> I was not sure whether this is a misuse of the `sc_pkcs15_unwrap()` API or whether it should check for the NULL pointers/uninitialized values as they are passed as arguments. Let me know what you think about this issue.

Indeed the access to `sc_pkcs15_unwrap()` needs different input parameters, something [like this](https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L2401-L2438) may work:
```diff
diff --git a/src/tests/fuzzing/fuzz_pkcs15_reader.c b/src/tests/fuzzing/fuzz_pkcs15_reader.c
index b6ab09f8..65d31ce4 100644
--- a/src/tests/fuzzing/fuzz_pkcs15_reader.c
+++ b/src/tests/fuzzing/fuzz_pkcs15_reader.c
@@ -233,10 +233,24 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
             int wrap_flags[] = {0, SC_ALGORITHM_AES_ECB, SC_ALGORITHM_AES_CBC_PAD,
                 SC_ALGORITHM_AES_CBC};
             for (i = 0; i < sizeof wrap_flags/sizeof *wrap_flags; i++) {
+                unsigned long l = sizeof buf;
                 struct sc_pkcs15_object target_key;
+                struct sc_pkcs15_skey_info skey_info;
+                memset(&target_key, 0, sizeof target_key);
+                memset(&skey_info, 0, sizeof skey_info);
+                target_key.type = SC_PKCS15_TYPE_SKEY;
+                target_key.flags = 2; /* TODO not sure what these mean */
+                target_key.session_object = 1;
+                target_key.data = skey_info;
+                skey_info.usage = (unsigned int) args.usage;
+                skey_info.native = 0; /* card can not use this */
+                skey_info.access_flags = 0; /* looks like not needed */
+                skey_info.key_type = 0x1fUL; /* CKK_AES */
+                skey_info.value_len = 128;
+                fuzz_get_chunk(reader, &skey_info.data.value, &skey_info.data.len);
+
                 sc_pkcs15_unwrap(p15card, obj, &target_key, wrap_flags[i],
                         in, in_len, param, param_len);
-                unsigned long l = sizeof buf;
                 sc_pkcs15_wrap(p15card, obj, &target_key, wrap_flags[i],
                         buf, &l, in, in_len);
             }
```
> After we will have some more corpuses (do you have some cards around you can contribute corpuses from?) and the above bug fixed, I will modify the build scripts in oss-fuzz to pick them up.

How are those generated?

Jakuje:
> Indeed the access to `sc_pkcs15_unwrap()` needs different input parameters, something [like this](https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L2401-L2438) may work:

The proposed changes actually do not build. It will probably need some more love to handle pointers correctly. Especially `args.usage` is not defined.

> > After we will have some more corpuses (do you have some cards around you can contribute corpuses from?) and the above bug fixed, I will modify the build scripts in oss-fuzz to pick them up.
> 
> How are those generated?

See the commit  975d4a0  and d84a68e introducing the code generating them and describing this in more detail. The file names are sha1 sums of content as required by oss-fuzz documentation.

.travis.yml
```diff
@@ -48,6 +48,7 @@ env:
 
 addons:
   apt_packages:
+    - autoconf-archive
```
frankmorgner: please add the m4 scripts by copying them to the m4 directory instead of creating a new install dependency (we already have some of the ax scripts there).

Jakuje: I pushed the new version with the coverage module copied in the tree. Lets see how it will be handled now in the CI.

Jakuje:
It looks like I already hunted down all the autoconf dependencies for the code coverage and looks like it kind of works also in windows. Lets give some more time to CI, but generally it looks good.

Jakuje:
At this moment, all is green. But I need to polish the top commit. It passes in CI, because it is missing some dependencies and these tests are skipped (so the commit does not make a difference):
```
SKIP: test-pkcs11-tool-sign-verify.sh
SKIP: test-pkcs11-tool-test.sh
SKIP: test-pkcs11-tool-allowed-mechanisms.sh
```
I would probably like to see them running too.

tomato42:
> Looks like all requested changes are done. Do you agree, @tomato42 ?

looks like it to me

Jakuje:
> At this moment, all is green. But I need to polish the top commit. It passes in CI, because it is missing some dependencies and these tests are skipped (so the commit does not make a difference):
> 
> ```
> SKIP: test-pkcs11-tool-sign-verify.sh
> SKIP: test-pkcs11-tool-test.sh
> SKIP: test-pkcs11-tool-allowed-mechanisms.sh
> ```
> 
> I would probably like to see them running too.

Seems like I forgot to comment on the last changes. I fixed the Ubuntu paths to softhms and now it runs at least something. Unfortunately, the `pkcs11-tool --test` and `ALLOWED_MECHANISMS` are still broken with `softhms` in Ubuntu so they are skipped. I also modified the Sign & Verity test to skip broken tests on Ubuntu so at least some of them are executed:
```
PASS: test-manpage.sh
PASS: test-pkcs11-tool-sign-verify.sh
SKIP: test-pkcs11-tool-test.sh
SKIP: test-pkcs11-tool-allowed-mechanisms.sh
```

frankmorgner:
Great, thank you!

PR #1840, opened 2019-10-30
Fix build on ARM, PPC

pkubaj:
Those architectures use unsigned char by default, and command variable is later compared to -1.


frankmorgner:
thanks

frankmorgner:
I wonder if this is the only place where `char` is implicitly taken as `signed char`...

pkubaj:
There's no problem if all you use is the range 0-127. Then it doesn't matter whether it's signed or unsigned.

Issue #1822, opened 2019-10-09
Signing operations fail after a token has been removed and reinserted (Chrome/Firefox)


the-kernel:
### Problem Description

After performing an operation that causes the PIV token to be unlocked (in our case a Yubikey 4), removing the token and then reinserting it causes all subsequent operations to fail with:
```
pkcs15-pin.c:358:sc_pkcs15_verify_pin: sc_lock() failed: -1900 (Unknown error)
...
reader-pcsc.c:546:pcsc_lock: Yubico Yubikey 4 OTP+U2F+CCID 00 00:SCardBeginTransaction failed: 0x80100011
misc.c:61:sc_to_cryptoki_error_common: libopensc return value: -1900 (Unknown error)
```

There seems to be an issue in the interaction between opensc and nss, this has been tested with CACKey as the middleware which appears to work fine.

It is worth noting that if you access the Firefox Security Devices page (Preferences->Security Devices) while the token is removed it updates the state, which leads me to think there is an issue with caching, and stale slots not being handled correctly (possibly because both Chromium and Firefox no longer use the NSS SECMOD_WaitForAnyTokenEvent (which in turn calls C_WaitForSlotEvent), causing the slots not to be updated when the token is removed).

[NSS Bug Report](https://bugzilla.mozilla.org/show_bug.cgi?id=1585714)

Note: "Please read about [reporting bugs](https://github.com/OpenSC/OpenSC/wiki/How-to-report-bugs-so-that-they-can-be-fixed) before opening an issue." The link is dead :(

### Steps to reproduce

1. Unlock PIV card by performing some operation (i.e. going to settings -> manage certificates)
2. Remove card.
3. Reinsert card.
4. Perform another operation that requires unlocking the card i.e. an x509 login.
5. Login fails due to signing operation failure.

### Logs

[opensc-debug.log](https://github.com/OpenSC/OpenSC/files/3706141/opensc-debug.log)

Edit:
This has been tested against opensc-0.15 and master branch build on Ubuntu 16.04 and 19.04 (with the latest versions of Chrome/Firefox)

frankmorgner:
which version of PCSC-Lite are you using?

the-kernel:
I have tried it against pcsc-lite 1.8.14, and 1.8.24
thanks

Jakuje:
Can you share debug log from the current master or 0.19.0 version? In your log, I see some mixing of sessions from different processes (observe the first hexadecimal value on the log lines).

[...]

PR #1842, opened 2019-10-31
Handle Yubikey reinsertion from browsers more gracefully

Jakuje:
Based on the discussion in #1822 this patch seems to solve the issues for me most of the time in chromium (the opensc plugged in through p11-kit). The Firefox looks like having some issues with the tokens these days so I did not test with it.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
good enough, thanks

PR #1845, opened 2019-11-04
Fix some issues reported by Coverity

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -272,14 +272,6 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
 		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 	LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_ALLOWED);
 #else
-	/* check that user_consent_app exists. TODO: check if executable */
-	res = stat(GET_DNIE_UI_CTX(card).user_consent_app, &st_file);
-	if (res != 0) {
-		sc_log(card->ctx, "Invalid pinentry application: %s\n",
-		       GET_DNIE_UI_CTX(card).user_consent_app);
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
-	}
-
 	/* just a simple bidirectional pipe+fork+exec implementation */
 	/* In a pipe, xx[0] is for reading, xx[1] is for writing */
 	if (pipe(srv_send) < 0) {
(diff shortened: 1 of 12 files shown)


src/libopensc/card-dnie.c
```diff
@@ -304,10 +296,17 @@ int dnie_ask_user_consent(struct sc_card * card, const char *title, const char *
...
+		res = stat(GET_DNIE_UI_CTX(card).user_consent_app, &st_file);
+		if (res != 0) {
+			sc_log(card->ctx, "Invalid pinentry application: %s\n",
+					GET_DNIE_UI_CTX(card).user_consent_app);
+		} else {
+			/* call exec() with proper user_consent_app from configuration */
+			/* if ok should never return */
+			execlp(GET_DNIE_UI_CTX(card).user_consent_app, GET_DNIE_UI_CTX(card).user_consent_app, (char *)NULL);
+			sc_log(card->ctx, "execlp() error");
+		}
```
Jakuje: nit: The expression `GET_DNIE_UI_CTX(card).user_consent_app` is used in the above code already 4 times. I think it should already deserve temporary variable to improve readability.

src/tools/pkcs15-init.c
```diff
@@ -2987,9 +2987,13 @@ read_options_file(const char *filename)
...
 				optarg = strtok(NULL, "");
 				if (optarg) {
+					char *start = optarg;
 					while (isspace((int) *optarg))
 						optarg++;
-					optarg = strdup(optarg);
+					if (optarg != start) {
+						optarg = strdup(optarg);
+						free(start);
+					}
```
Jakuje: I don't think this is a correct fix. The issue here is that we have (global?) variable `optarg`, which is set here and possibly handled in `handle_option()` -- assigned to different variable, but in following cycle we are overwriting `strdup`-ed variable with a new one.

Your proposed solution will allow assigning pointers to the static buffer to these options, which will get overwritten in the next cycle.

src/pkcs15init/pkcs15-oberthur-awp.c
```diff
@@ -967,7 +967,7 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
 	 */
 	ptr = awp_get_commonName(x);
 	if (!ptr)
-		LOG_TEST_RET(ctx, SC_ERROR_INTERNAL, "AWP encode cert failed: cannot get CommonName");
+		LOG_TEST_GOTO_ERR(ctx, SC_ERROR_INTERNAL, "AWP encode cert failed: cannot get CommonName");
```
Jakuje: If I read the macro `LOG_TEST_GOTO_ERR` correctly, it does not set the `r` variable, which is used as a return value of the function. This means that the function is going to return the previous value of `r`, which is in this case `0`, even on errors from these conditions, that are not modifying `r`.

src/tools/pkcs15-init.c
frankmorgner: I've removed this commit; I didn't find an easy solution to this.

I think the easiest way to handle this issue is to deprecate `--options-file`, which isn't used in any other tool. It was invented to not expose PINs via `ps`, which we can use environment variables for nowadays...

Issue #1688, opened 2019-05-27
GlobalSign smart card enrollment not working


CardContact:
### Problem Description

As described [here](https://support.nitrokey.com/t/globalsign-no-cryptographic-service-provider-listed-nitrokey-hsm2-install-cert-via-csp-in-internet-explorer/1791), certificate enrollment using Internet Explorer does no work with OpenSC.

Most common causes are already ruled out, but maybe someone has a working setup or a test service that can be used to further diagnose the root cause of the problem.

### Steps to reproduce

Enroll for a GlobalSign e-mail certificate

### Logs

[log](https://gist.github.com/CardContact/6dd9af7f090b254c3f9b334d63af36ec)

szszszsz:
cc @alex-nitrokey @jans23

frankmorgner:
please try #1685

shivan:
Tried that. Now I get one step further. I can enter the pin. But it stops here:

![image](https://user-images.githubusercontent.com/1827692/58784738-68468300-85e4-11e9-9c76-01251d62c733.png)

Can be reproduced with free 30 day mail certificates, using Microsoft Internet Explorer at:
https://secure.comodo.net/products/frontpage?area=SecureEmailCertificate

szszszsz:
Linking here additional logs from the discussion on the forum: [logs]

[logs]: https://ametrasrentconcept-my.sharepoint.com/personal/markus_stein_remoso_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fmarkus%5Fstein%5Fremoso%5Fcom%2FDocuments%2FHSM%2Fmd%2Dmaster%2Dbuild%2Elog&parent=%2Fpersonal%2Fmarkus%5Fstein%5Fremoso%5Fcom%2FDocuments%2FHSM&cid=69602225-df7a-4a29-92f1-8c67bf6197a9

frankmorgner:
@szszszsz seems to require some MS account

szszszsz:
Works for me. Perhaps Javascript / some browser extension? I will reupload it to Github in a minute.

shivan:
Any useful informations in the log?

shivan:
Hi,

I currently cannot see any progress on that issue since 20 days.

Is there anything missing from my side? Or can I do anything else to support?

Jakuje:
Do you see some errors related to smart cars in the "event log" of windows?

shivan:
Hi,

nothing at 

[...]

Issue #447, opened 2015-04-23
windows setup improvement


vletoux:
I'd like to work on the setup program.
I need your feedback about this proposition
(I can spend some time tomorrow about that)

**1) look & feel**
I'd like to change the default red background provided by wix to another one.
(like this one: https://raw.githubusercontent.com/vletoux/OpenSC/7a7f8fb27471719a32f0cdcf819170771589791b/win32/dlgbmp.bmp)
If you have some logo / graphic, I can improve the proposed design.

**2) registering minidriver**
I'd like to add the registration of the minidriver for all cards (except piv - I can find the ATR & mask for all other drivers)
Indeed the minidriver needs a proper inf file to be install, but the driver will be unsigned and hard to design for each card (extract the historical bytes).
There is a trick to use a minidriver (including the ctrl-alt-del changing pin screen) on Windows 7 without having a driver install. You set exactly the same registry than for a minidriver installation, but the CSP points to another one (like OpenSC CSP) and this new CSP has been registered with the same dll than the Base CSP. I'll add the Base KSP registration data which is missing in existing files.

**optionally**
If you like, I can donate the source code for the program we called "smart card manager" & used in several of our projects (see http://download.mysmartlogon.com/EIDVirtual/EIDVirtual%20-%20User%20Documentation.pdf page 11 to 13). The user will be able to manage its card using the minidriver interface and check quickly if everything is fine for Windows. It is indeed more user friendly than "certutil -scinfo".

I'm waiting for your feedback.

regards,
Vincent

dengert:
On 4/23/2015 4:39 AM, vletoux wrote:
> I'd like to work on the setup program.
> I need your feedback about this proposition
> (I can spend some time tomorrow about that)
>
> ++1) look & feel++
> I'd like to change the default red background provided by wix to another one.
> (like this one: https://raw.githubusercontent.com/vletoux/OpenSC/7a7f8fb27471719a32f0cdcf819170771589791b/win32/dlgbmp.bmp)
> If you have some logo / graphic, I can improve the proposed design.

OK, red background bug me too.  Google for: red background color intimidating

>
> ++2) registering minidriver++
> I'd like to add the registration of the minidriver for all cards (except piv - I can find the ATR & mask for all other drivers)
> Indeed the minidriver needs a proper inf file to be install, but the driver will be unsigned and hard to design for each card (extract the historical bytes).
[...]

PR #1848, opened 2019-11-05
Reuse OpenSC reader when handles change.  Better error debug messages for Minidriver 

### LINKED ISSUE #1763: SC-HSM: Creating private key for ADCS and other certificates fails w/ SCARD_E_READ_ONLY_CARD when using OpenSC CSP or MS Smart Card KSP

dengert:
Possibly Fixes #1763  But needs testing by others.  

Windows CNG is managing the insertion and removal of the reader and the card and will call `CardAcquireContext` and `CardDeleteContext` If the card or reader actually change. But different processes or threads may establish different PCSC handles to the same reader and card but with different handles using the same Context created by `CardAcquireContext`. 

`minidriver.c` is being changed to still call `sc_ctx_use_reader` that calls `pcsc_use_reader` but no longer calls `reinit_card_for` when just the handles change. 

`reader-pcsc.c` in `pcsc_use_reader` will check if it has a reader and if so, only the handles are changed. otherwise a  new reader will be initialized using the handles passed. A new reader is only initialized from `CardAcquireContext`.  

In the old  `pcsc_use_reader` the `reader` was deleted which include the reader and card states.  `minidriver.c` would thus lost any information in memory when it called `disassociate_card` and `associate_card`. This may be the root problem seen in #1763
 
Tests were done using a SmartCard-HSM. 

`certutil -v -scinfo` works the same. 

A`power shell`  command was able to generate a key and a self signed certificate on the card that could be read from another computer.  (The script needs a lot of work but is a starting point.) 
```
New-SelfSignedCertificate `
 -Type Custom `
 -Subject "E=deengert@gmail,com,CN=Doug Engert-selfSignedTest" `
 -KeyExportPolicy NonExportable `
 -KeyUsage DigitalSignature `
 -KeySpec Signature `
 -KeyUsageProperty Sign `
 -SmimeCapabilities `
 -CertStoreLocation "Cert:\CurrentUser\My" `
 -Provider "OpenSC CSP" `
 -FriendlyName "selfSignedTest"
```

For easier debugging, `MD_FUNC_CALLED(pCardData, level)` and  `MD_FUNC_RETURN(pCardData, level, ...)` macros were created. 

Handles are type `__int3264` in VS2015 and are cast to  `size_t` when printing so all bytes of the handles are printed rather then only 4 bytes. `size_t` on Windows are also treated as 32 or 64 bits.
`SC_FORMAT_LEN_SIZE` is used in the printf format.
 (Works with VS2105 needs to be tested on other compilers.)

 Changes to be committed:
	modified:   src/libopensc/reader-pcsc.c
	modified:   src/minidriver/minidriver.c

Code based on 0.20.0-rc3 
Built on Windows 10 using VS2015 with OpenSSL 1.1.0 and  zlib but not openpace. 
Additional testing can be done with the artifacts of the Appveyor from the PR. 

##### Checklist

- [X] Windows minidriver is tested

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -71,6 +71,25 @@
 
 #include "cardmod.h"
 
+#define MD_FUNC_CALLED(pCardData, level) do { \
+	logprintf(pCardData, level, "MD_Function:%s:%d called\n",__func__, __LINE__); \
+	} while(0)
+
+
+#define MD_FUNC_CALLED_EXT(pCardData, level) do { \
+	logprintf(pCardData, level, "\nP:%lu T:%lu MD_Function:%s:%d called\n", \
+	(unsigned long)GetCurrentProcessId(),\
+	(unsigned long)GetCurrentThreadId(), \
+	__func__, __LINE__); \
+	} while(0)
+
+#define MD_FUNC_RETURN(pCardData, level, ...) do { \
+	DWORD _ret = __VA_ARGS__; \
+	logprintf(pCardData, level,\
+		"MD_Function:%s:%d returning with: 0x%08X\n", __func__, __LINE__, _ret); \
+	return _ret; \
+	} while(0)
+
 /* store the instance given at DllMain when attached to access internal resources */
 HINSTANCE g_inst;
 
@@ -309,13 +328,11 @@ static DWORD reinit_card(PCARD_DATA pCardData)
 	DWORD r;
 
 	if (!pCardData)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 	vs = (VENDOR_SPECIFIC *)(pCardData->pvVendorSpecific);
 	if (!vs)
-		return SCARD_E_INVALID_PARAMETER;
-
-	logprintf(pCardData, 2, "trying to reinit card\n");
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 	if (vs->initialized) {
 		disassociate_card(pCardData);
@@ -324,7 +341,7 @@ static DWORD reinit_card(PCARD_DATA pCardData)
 
 	r = associate_card(pCardData);
 	if (r != SCARD_S_SUCCESS)
-		return r;
+		MD_FUNC_RETURN(pCardData, 1,  r);
 
 	r = md_fs_init(pCardData);
 	if (r != SCARD_S_SUCCESS) {
@@ -332,10 +349,10 @@ static DWORD reinit_card(PCARD_DATA pCardData)
 			  "reinit_card md_fs_init failed, r = 0x%lX\n",
 			  (unsigned long)r);
 		disassociate_card(pCardData);
-		return r;
+		MD_FUNC_RETURN(pCardData, 1, r);
 	}
 
-	return SCARD_S_SUCCESS;
+	MD_FUNC_RETURN(pCardData, 1, SCARD_S_SUCCESS);
 }
 
 static BOOL lock(PCARD_DATA pCardData)
@@ -371,24 +388,26 @@ static DWORD reinit_card_for(PCARD_DATA pCardData, const char *name)
 			  "%s was called, but unable to initialize card, r = %u\n",
 			  name, (unsigned int)r);
 
-	return r;
+	MD_FUNC_RETURN(pCardData, 1, r);
 }
 
 static DWORD check_card_status(PCARD_DATA pCardData, const char *name)
 {
 	VENDOR_SPECIFIC *vs;
 
+	MD_FUNC_CALLED(pCardData, 3);
+
 	if (!pCardData)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 3, SCARD_E_INVALID_PARAMETER);
 
 	vs = (VENDOR_SPECIFIC *)(pCardData->pvVendorSpecific);
 	if (!vs)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 3, SCARD_E_INVALID_PARAMETER);
 
 	if (vs->initialized)
-		return SCARD_S_SUCCESS;
+		MD_FUNC_RETURN(pCardData, 3, SCARD_S_SUCCESS);
 
-	return reinit_card_for(pCardData, name);
+	MD_FUNC_RETURN(pCardData, 3, reinit_card_for(pCardData, name));
 }
 
 /*
@@ -403,27 +422,37 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
 	DWORD dwRet;
 	int r;
 
-	logprintf(pCardData, 4, "check_reader_status\n");
+	logprintf(pCardData, 1, "check_reader_status for %s\n", name);
 	if(!pCardData)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 	dwRet = check_card_status(pCardData, name);
 	if (dwRet != SCARD_S_SUCCESS)
-		return dwRet;
+		MD_FUNC_RETURN(pCardData, 3, dwRet);
 
 	vs = (VENDOR_SPECIFIC*)(pCardData->pvVendorSpecific);
 	if(!vs)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 3, SCARD_E_INVALID_PARAMETER);
 
-	logprintf(pCardData, 7, "pCardData->hSCardCtx:0x%08X hScard:0x%08X\n",
-		  (unsigned int)pCardData->hSCardCtx,
-		  (unsigned int)pCardData->hScard);
+	logprintf(pCardData, 7, "sizeof(size_t):%d sizeof(ULONG_PTR):%d sizeof(__int3264):%d sizeof pCardData->hSCardCtx:%d\n",
+		sizeof(size_t), sizeof(ULONG_PTR), sizeof(__int3264), sizeof pCardData->hSCardCtx);
+
+	logprintf(pCardData, 1, "pCardData->hSCardCtx:0x%08"SC_FORMAT_LEN_SIZE_T"X hScard:0x%08"SC_FORMAT_LEN_SIZE_T"X\n",
+		(size_t)pCardData->hSCardCtx,
+		(size_t)pCardData->hScard);
 
 	if (pCardData->hSCardCtx != vs->hSCardCtx || pCardData->hScard != vs->hScard) {
-		logprintf(pCardData, 1, "HANDLES CHANGED from 0x%08X 0x%08X\n",
-			  (unsigned int)vs->hSCardCtx,
-			  (unsigned int)vs->hScard);
-		return reinit_card_for(pCardData, name);
+		logprintf(pCardData, 1, "HANDLES CHANGED from  vs->hSCardCtx:0x%08"SC_FORMAT_LEN_SIZE_T"X vs->hScard:0x%08"SC_FORMAT_LEN_SIZE_T"X\n",
+			(size_t)vs->hSCardCtx,
+			(size_t)vs->hScard);
+		if (vs->ctx) {
+			vs->hScard = pCardData->hScard;
+			vs->hSCardCtx = pCardData->hSCardCtx;
+			r = sc_ctx_use_reader(vs->ctx, &vs->hSCardCtx, &vs->hScard);
+			logprintf(pCardData, 1, "sc_ctx_use_reader returned %d\n", r);
+			if (r)
+			    MD_FUNC_RETURN(pCardData, 1, SCARD_F_INTERNAL_ERROR);
+		}
 	}
 
 	/* This should always work, as BaseCSP should be checking for removal too */
@@ -432,8 +461,8 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
 		  "check_reader_status r=%d flags 0x%08X\n", r,
 		  (unsigned int)vs->reader->flags);
 	if (r < 0)
-		return md_translate_OpenSC_to_Windows_error(r,
-							    SCARD_F_INTERNAL_ERROR);
+		MD_FUNC_RETURN(pCardData, 1,
+			md_translate_OpenSC_to_Windows_error(r, SCARD_F_INTERNAL_ERROR));
 
 	if (!(r & SC_READER_CARD_PRESENT)) {
 		/*
@@ -442,10 +471,10 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
 		 * it anyway for completeness
 		 */
 		logprintf(pCardData, 1, "no card present? trying to reinit\n");
-		return reinit_card_for(pCardData, name);
+		MD_FUNC_RETURN(pCardData, 1, reinit_card_for(pCardData, name));
 	}
 
-	return SCARD_S_SUCCESS;
+	MD_FUNC_RETURN(pCardData, 1, SCARD_S_SUCCESS);
 }
 
 static DWORD
@@ -454,12 +483,14 @@ md_get_pin_by_role(PCARD_DATA pCardData, PIN_ID role, struct sc_pkcs15_object **
 	VENDOR_SPECIFIC *vs;
 	int rv;
 
+	MD_FUNC_CALLED(pCardData, 1);
+
 	if (!pCardData)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 	vs = (VENDOR_SPECIFIC*)(pCardData->pvVendorSpecific);
 	if (!ret_obj)
-		return SCARD_E_INVALID_PARAMETER;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 	/* please keep me in sync with _get_auth_object_by_name() in pkcs11/framework-pkcs15.c */
 	if (role == ROLE_USER) {
@@ -507,18 +538,18 @@ md_get_pin_by_role(PCARD_DATA pCardData, PIN_ID role, struct sc_pkcs15_object **
 		logprintf(pCardData, 2,
 			  "cannot get PIN object: unsupported role %u\n",
 			  (unsigned int)role);
-		return SCARD_E_UNSUPPORTED_FEATURE;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_UNSUPPORTED_FEATURE);
 	}
 
 	if (rv)
-		return SCARD_E_UNSUPPORTED_FEATURE;
+		MD_FUNC_RETURN(pCardData, 1, SCARD_E_UNSUPPORTED_FEATURE);
 
 	if (*ret_obj)
 		logprintf(pCardData, 7, "Returning PIN '%.*s' for role %u\n",
 			  (int) sizeof (*ret_obj)->label, (*ret_obj)->label,
 			  (unsigned int)role);
 
-	return SCARD_S_SUCCESS;
+	MD_FUNC_RETURN(pCardData, 1, SCARD_S_SUCCESS);
 }
 
 static const char *
@@ -653,6 +684,7 @@ md_get_config_bool(PCARD_DATA pCardData, char *flag_name, BOOL ret_default)
 			ret = scconf_get_bool(atrblock, flag_name, ret_default) ? TRUE : FALSE;
 	}
 
+	logprintf(pCardData, 2, "flag_name:%s:%s\n", flag_name, ret ? "TRUE": "FALSE");
 	return ret;
 }
 
@@ -1657,19 +1689,19 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
 
 	dwret = md_get_pin_by_role(pCardData, ROLE_USER, &vs->pin_objs[ROLE_USER]);
 	if (dwret != SCARD_S_SUCCESS)   {
-		logprintf(pCardData, 2, "Cannot get User PIN object");
+		logprintf(pCardData, 2, "Cannot get User PIN object\n");
 		return dwret;
 	}
 
 	dwret = md_get_pin_by_role(pCardData, MD_ROLE_USER_SIGN, &vs->pin_objs[MD_ROLE_USER_SIGN]);
 	if (dwret != SCARD_S_SUCCESS) {
-		logprintf(pCardData, 2, "Cannot get Sign PIN object -- ignored");
+		logprintf(pCardData, 2, "Cannot get Sign PIN object -- ignored\n");
 		vs->pin_objs[MD_ROLE_USER_SIGN] = NULL;
 	}
 
 	dwret = md_get_pin_by_role(pCardData, ROLE_ADMIN, &vs->pin_objs[ROLE_ADMIN]);
 	if (dwret != SCARD_S_SUCCESS) {
-		logprintf(pCardData, 2, "Cannot get Admin PIN object -- ignored");
+		logprintf(pCardData, 2, "Cannot get Admin PIN object -- ignored\n");
 		vs->pin_objs[ROLE_ADMIN] = NULL;
 	}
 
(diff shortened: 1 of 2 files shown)


frankmorgner:
`reinit_card_for()` performs `disassociate_card()` and `associate_card()`, i.e. it fully reinitializes the card. What's the advantage of simply changing the handles while leaving all initialized (and possibly obsolete) data as is?

Which card state, do you think will be lost when the reader is removed?

Try using `__FUNC__` instead of `__func__`.

dengert:
Because windows supplies new handles with the same PCARD_DATA does not mean it may be a different card or there was a reinsertion of the card or reader. What it does mean is Windows can have multiple PCSC connections to the same card from different threads.  **In other words the PCARD_DATA represents the card and contains transient data such as that created by `MdGenerateKey`  The caller expects that data to still be accessible after the handles are changed.**  

I believe if the caller of the minidriver thought the reader or card was reinserted or changed, it would have called CardDeleteContext and then started over. 

OpenSC was written to do all the error handling for PCSC

I believe the caller of the minidriver is expecting to handle many of these errors, such as removal of the card or reader. The minidriver should  return the PCSC error codes. The caller of the minidriver would then cleanup by calling CardDeleteContext.  The caller of the minidriver could then start over if it wanted to. 

Changing of the handles should is not an error.   

If you grep for HANDLES_CHANGED in the two files below, you will see lines like this:
```
 pCardData->hSCardCtx:0xCD00000100000002 hScard:0xEA00000500000000
HANDLES CHANGED from  vs->hSCardCtx:0xCD00000100000002 vs->hScard:0xEA00000400000000
sc_ctx_use_reader returned 0
check_reader_status r=5 flags 0x00000005
MD_Function:check_card_reader_status:481 returning with: 0x00000000
```
`sc_ctx_use_reader` causes just the handles in the existing  OpenSC `reader` to be changed. This then leaves the rest of `reader` the same including existing state.  In the old code the new `reader` looked like the card was reinserted as the previous state was 0. In the new code the  previous state is the previous state. 

(How to return real PCSC encountered errors is not addressed by this PR.)

Here are debug logs from a version I built on Windows 10 using VS2015 from essentially the same source in the PR. 

[md-1105-selfsign.log](https://github.com/OpenSC/OpenSC/files/3814453/md-1105-selfsign.log)
[opensc-debug-1105-selfsign.txt](https://github.com/OpenSC/OpenSC/files/3814454/opensc-debug-1105-selfsign.txt)

I am hoping https://github.com/marcoh00 or https://github.com/gnugnug will give this a try to see if it solves their problem. 

(I will try __FUNC__ with next update.)

frankmorgner:
OK, makes sense, although I'd feel more comfortable if we had some MS' documentation that says we should keep the internal state when the handles are changing...

@marcoh00 @gnugnug ?

src/minidriver/minidriver.c
```diff
@@ -403,27 +422,37 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
...
-			  (unsigned int)vs->hSCardCtx,
-			  (unsigned int)vs->hScard);
-		return reinit_card_for(pCardData, name);
+		logprintf(pCardData, 1, "HANDLES CHANGED from  vs->hSCardCtx:0x%08"SC_FORMAT_LEN_SIZE_T"X vs->hScard:0x%08"SC_FORMAT_LEN_SIZE_T"X\n",
+			(size_t)vs->hSCardCtx,
+			(size_t)vs->hScard);
+		if (vs->ctx) {
+			vs->hScard = pCardData->hScard;
+			vs->hSCardCtx = pCardData->hSCardCtx;
+			r = sc_ctx_use_reader(vs->ctx, &vs->hSCardCtx, &vs->hScard);
```
frankmorgner: I think it would be good to add a new optional `sc_reader_t *` parameter to `sc_ctx_use_reader()`. If it is passed, this reader's handles are modified; no further action is done. If this parameter is `NULL`, then `pcsc_use_reader()` can still do its dance about deleting and adding the internal readers.

dengert: Within the minidriver there can only be one reader. `reader-pcsc.c` already has many tests for  
`priv->gpriv->cardmod`  that do not allow `SCardConnect` and  `SCardDisconnect`  (But does allow `SCardReconnect` which is probable a bug.)   
`pcsc_lock` and `pcsc_unlock`  always return `SC_SUCCESS` without doing anything because "minidriver_specs_v7.07" section 2.3.1 say:
•	A card minidriver should assume that transactions are handled by the caller, if it uses SCRM to access the card.
•	The card minidriver can assume that all entry points except CardDeleteContext are called by holding the card transaction. This cannot be assumed in CardDeleteContext because the card might have been removed or it is being called as part of a cleanup procedure.

So the whole concept of the minidriver is to use the provided handles and let the caller deal with errors.  The minidriver OpenSC has never been able to have more then one `reader_t` with different handles  because it can not do the `SCardConnect`, `SCardDisconnect`,  `SCardBeginTransaction` or `SCardEndTransaction`.

In effect  sc_ctx_use_reader() does what you ask. The test in `reader-pcsc.c` `pcsc_use_reader` 
`if (sc_ctx_get_reader_count(ctx) > 0) {` is testing if the reader_t exists. 

note that in `minidriver.c` `associate_card`   line in 7046 `sc_ctx_get_reader_count(vs->ctx)`
and line 7048 `vs->reader = sc_ctx_get_reader(vs->ctx, 0);`  list number of readers and set `vs->reader` to the first (and only) reader.

frankmorgner: From a functional point of view, you may be right. From a software architecture view, `sc_ctx_use_reader()` claims some sort of universal applicability otherwise we would not need to add the `use_reader` callback to every reader implementation. However, I feel that your code has too many references to the minidriver. My suggestion was intended to maintain this universality.

However, thinking about this, it may be better to simply remove the `use_reader` callback entirely and really create a function in reader_pcsc.c specifically for the interaction with the minidriver. This would simplify the overall architecture and remove some clutter from the mind of a programmer unfamiliar with the whole entity of the code.

dengert: `pcsc_use_reader` **is** the " function in `reader_pcsc.c` specifically for the interaction with the minidriver."  

"remove some clutter from the mind of a programmer unfamiliar with the whole entity of the code." we could  use code you added:
```
fcc8ea593 (Frank Morgner        2016-11-15 22:48:48 +0100  797)         if(strcmp(ctx->app_name, "cardmod") == 0) {
fcc8ea593 (Frank Morgner        2016-11-15 22:48:48 +0100  798)                 gpriv->cardmod = 1;
fcc8ea593 (Frank Morgner        2016-11-15 22:48:48 +0100  799)         }
```
and add:
```
diff --git a/src/libopensc/reader-pcsc.c b/src/libopensc/reader-pcsc.c
index 32bf9934..7303fd5f 100644
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -2382,6 +2382,11 @@ int pcsc_use_reader(sc_context_t *ctx, void * pcsc_context_handle, void * pcsc_c
                goto out;
        }
 
+       if (!gpriv->cardmod) {
+               ret = SC_ERROR_INTERNAL;
+               goto err;
+       }
+
        /* Only minidriver calls this and only uses one reader */
        /* if we already have a reader, use it*/
        if (sc_ctx_get_reader_count(ctx) > 0) {
```

If a programmer tried to use pcsc_use_reader, it would fail during development testing. 


0e03c6d5c 2010

dengert:
The closest  "MS' documentation" I have found is [sc-minidriver_specs_V7.07](http://download.microsoft.com/download/3/3/2/332fd70b-f04d-470a-a135-040350b9563f/sc-minidriver_specs_v7.07.docx)  which states:

"2.3.1	Transaction Management
•	A card minidriver should assume that transactions are handled by the caller, if it uses SCRM to access the card.
•	The card minidriver can assume that all entry points except CardDeleteContext are called by holding the card transaction. This cannot be assumed in CardDeleteContext because the card might have been removed or it is being called as part of a cleanup procedure.
•	Multiple contexts can exist in a single process. Calling CardDeleteContext on one process should not prevent the other context from functioning.
•	Handling the authentication state of the card is also the responsibility of the caller, not the card minidriver."

The caller has done a SCardBeginTransaction using the SCRM (i.e. PCSC) handles passed. The caller has already matched the ATR to a minidriver.  Subsequent calls to the minidriver with the same context are requests for the same card and the caller is handling the login state too.  

 OpenSC has always used one handle for a reader and one handle to a card. But Windows says it may use multiple handles to access the same reader with the same card even in the same process. Windows is doing a lot of processing to make it "easier" to write a minidriver from scratch. But it makes it harder to turn OpenSC inside out to use the handles provided and think like Microsoft.

dengert:
One Travis-ci  build failed with:
```
++java -noverify -cp IsoApplet/src/:jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar com.licel.jcardsim.remote.VSmartCard isoapplet_jcardsim.cfg
Error: Could not find or load main class com.licel.jcardsim.remote.VSmartCard
```
This looks unrelated to this PR.

frankmorgner:
I just realized that the same callbacks are used in OpenSCToken on macOS (`reader-cryptotokenkit.m`). So not the architecture needs to be simplified, as I've suggested above. Instead, the documentation should be extended. Your code can stay as is.

@marcoh00 @gnugnug @jans23 did you have a look at the fix?

dengert:
Some additional comments.

The PR appears to allow a SHA1 certificate to be created and stored on the SC-HSM token. But not a SHA256 certificate. From what I can deduced from reading various MS documents, it looks like this is caused be a number of MS applications and libraries being updated for SHA256 but not all.

SHA256 was added to CAPI2 but not CAPI1 

The Minidriver specs define how write support for  CSP and KSP (i.e. CNG). 

The registry entries for  HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards\SmartCard-HSM
define `Crypto Provider` which leads to loading `msclmd.dll` i.e. a CSP module that will call the minidriver and `Smart Card Key Storage Provider` that is a flag meaning the minidriver also supports KSP. 

CAPI1 will only use a CSP minidriver.  CAPI2 can use either CSP or KSP. 

So trying to generate a self signed certificate using  `PowerShell` `New-SelfSignedCertificate` can accept `-HashAlgorithm sha256` and create a private key on the card which does not need to use the hash. But then  somewhere in the com or .net routines, it appears that CAPI1 is used to sign the certificate that is created. This is where it fails to sign the certificate because CAPI1 will not accept SHA256 even though the minidriver can accept it.  I don't know how to get around this as I am not a com or .net programmer.

frankmorgner:
@marcoh00 @gnugnug @jans23, any updates?

frankmorgner:
@dengert your approach looks sane, but since this doesn't solve any actual problem and we don't get feedback from @marcoh00, @gnugnug, or @jans23 for #1688, I have the impression we're somewhere stuck half the way. How do you think should we proceed?

jans23:
I do follow this discussion but don't have any opinion on this (yet).

dengert:
I would like to see  @marcoh00, @gnugnug, or @jans23  test the 0.20.0 release, which has PR  #1866  but no this PR. This PR addresses what appears to be excessive overhead put on a card by forcing the disassociate_card and associate_card when this is not needed. This may be a problem they have seen. 

It does not fix the SHA256 https://github.com/OpenSC/OpenSC/issues/1688 issue with certificate creation which appears to me  to be a Microsoft problem with code they may be dropping. 

It does add better debugging information that might help with #1688 to show it is a Microsoft problem.  

Now that 0.20.0 is out we could commit this PR to master, to see if anyone has problems with it, as it does cut out a lot of overhead.

marcoh00:
Hi @dengert ! Sorry for my rather long absence!

The 0.20.0 release works using Microsoft Smart Card Key Storage Provider. OpenSC CSP returns a `NTE_BAD_KEYSET: The keyset does not exist`.

[md.log](https://github.com/OpenSC/OpenSC/files/4053159/md.log)
[opensc-debug.txt](https://github.com/OpenSC/OpenSC/files/4053161/opensc-debug.txt)

dengert:
When you say it works, what works? From a quick look at https://github.com/OpenSC/OpenSC/files/4053159/md.log and https://github.com/OpenSC/OpenSC/files/4053161/opensc-debug.txt line 1923 has:
```
MdGenerateKey(): generated key(idx:0,id:0e1f041f7a928c4f0c06643ae83440b13d60f6c3,guid:te-aaafb4da-8288-4333-873b-00c0447c5cc5)
key generated
```
and line starting at 2065 show the  SubjectPublicKeyInfo with the RSA public key. 

But it does not show any use of the key to sign a certificate request. (CSR)

NTE_BAD_KEYSET is 0x80090016 is vague, and I cant find a specific definition. Keyset does not exist could be because the creation of a certificate using the generated key did not complete. 

I still think the problem of using SHA256 vs SHA1 in the Microsoft code. To test this. create a CSR that uses SHA1 for the signature, but not a CSR that uses SHA256.  This might require the CA to accept a SHA1 signature for the CSR and create a certificate using SHA1 that would then be loaded on the card. 

Certificate enrollment with SHA256 may still work for software certs/keys, but not smart cards that rely on CSP vs KSP because they did not fix the SHA256 correctly for KSP. (or back porting to CSP)

You could also test by generating a key pair and CSR on MacOS, Unix or using OpenSSL and OpenSC on Windows, (anything but the CSP/KSP based enrollment) get the the CA to create the certificate, load the certificate on the card then test from certutil.

marcoh00:
Hi!

I also think that this is a problem inside the Microsoft code - but unfortunately it seems to be a prevalent one. My guess is that there's a switch somewhere that uses SHA1 for (legacy) CSPs/CAPI and only accepts SHA2-algorithms for KSPs/CNG.

Unfortunately I still don't think I really understand their driver model and how those CSPs/KSPs interact with each other. On one side, it seems to be possible to implement both the KSP and CSP interfaces, on the other side the Microsoft Base CSP and KSP seem to interact with OpenSC as well. Intrestingly, using the MS Smart Card KSP seems to bring better results than directly using the OpenSC KSP while at the same time they both seem to make use of OpenSC somehow (otherwise I suppose the SC-HSM would not work at all).

Anyway, when I say "it works", I mean that when using the MS Smart Card KSP I experienced no problems whatsoever and a new key pair + sha-256-signed certificate are successfully generated and stored on the SC-HSM. When using "OpenSC CSP", the above error occurs (at least using the 0.20.0 release, the CI build of this PR had no problems in this regard).

If it helps I could do some tests with generating a SHA256-signature-based certificate using the MSKSP and try to use it via OpenSC CSP. Unfortunately I don't think there is much certutil can do in terms of concrete cryptographic operations apart from some simple self-tests, can it? Do you have an idea which tools also use the MS Crypto API which could help with testing?

dengert:
I think the key diagrams for CSP/KSP/CAPI1/CAPI2 can be found at [smart-card-minidrivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/smartcard/smart-card-minidrivers) 

CAPI API only calls the baseCSP.DLL  which does not support SHA256.

CAPI2 API can call CAPI to call baseCSP.DLL, or CNG to call SCKSP.DLL that does support SHA256. 

So there are two paths to get to the "CardVvendor1.DLL" i.e. OpenSC minidriver. 

Look at:
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Defaults\Provider\OpenSC CSP`
and
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Defaults\Provider\Microsoft Base Smart Card Crypto Provider`
Both point at basecsp.dll, but OpenSC CSP does not have all the other value. I suspect that the basecsp.dll looks at the Microsoft Base Smart Card Crypto Provider entry to find them.

For every card OpenSC installs such as SmartCard-HSM 4K:
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards\SmartCard-HSM 4K`
there is a "Crypto Provider" with value "OpenSC CSP" and a "Smart Card Key Storage Provider" with value "Microsoft Smart Card Key Storage Provider" which looks more like a flag that says the driver also supports CNG, and to load SCKSP.DLL.

You say you are using the "MS Smart Card KSP".  Can you explain? is this sckcp.dll? If so how are you calling it.  Is it using CAPI2?

I think the problem for SHA256 is in the PowerShell applets that end up calling CAPI  and not CAPI2. 

0.20.0 does contain change for the NITRO card so that may get around some of your problems.

frankmorgner:
@marcoh00, we have https://github.com/OpenSC/OpenSC/blob/master/src/tests/opensc-minidriver-test.c that performs some basic tests on the CNG API with OpenSC...

I only see the additional DWORD setting `SignInFile` in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Defaults\Provider\Microsoft Base Smart Card Crypto Provider`, what other value do you mean @dengert?

dengert:
Here is what my Windows 10  64 bit 1909 build 18363.535 has for `Microsoft Base Smart Card Crypto Provider`.  I don't recall making any changes. I may have changes the timeout. 

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Defaults\Provider\Microsoft Base Smart Card Crypto Provider]
"SigInFile"=dword:00000000
"Type"=dword:00000001
"AllowPrivateExchangeKeyImport"=dword:00000000
"AllowPrivateSignatureKeyImport"=dword:00000000
"DefaultPrivateKeyLenBits"=dword:00000400
"Image Path"="basecsp.dll"
"RequireOnCardPrivateKeyGen"=dword:00000000
"TransactionTimeoutMilliseconds"=dword:00001388
```
Two questions:
Does the "OpenSC CSP" need any of these values too?  `OpenSC CSP` provider name appears to be an alias for `basecsp.dll`.

Are there additional entries that are missing? Like to support SHA256?

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Defaults\Provider\OpenSC CSP]
"Image Path"="basecsp.dll"
"Type"=dword:00000001
```

dengert:
This might also be helpful: 
[base-csp-and-smart-card-ksp-registry-keys](https://docs.microsoft.com/en-us/windows/security/identity-protection/smart-cards/smart-card-group-policy-and-registry-settings#base-csp-and-smart-card-ksp-registry-keys)

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\Cryptography\Providers\Microsoft Smart Card Key Storage Provider]
"Aliases"=hex(7):4d,00,69,00,63,00,72,00,6f,00,73,00,6f,00,66,00,74,00,20,00,\
  42,00,61,00,73,00,65,00,20,00,53,00,6d,00,61,00,72,00,74,00,20,00,43,00,61,\
  00,72,00,64,00,20,00,43,00,72,00,79,00,70,00,74,00,6f,00,20,00,50,00,72,00,\
  6f,00,76,00,69,00,64,00,65,00,72,00,00,00,4d,00,69,00,63,00,72,00,6f,00,73,\
  00,6f,00,66,00,74,00,20,00,53,00,6d,00,61,00,72,00,74,00,20,00,43,00,61,00,\
  72,00,64,00,20,00,4b,00,65,00,79,00,20,00,53,00,74,00,6f,00,72,00,61,00,67,\
  00,65,00,20,00,50,00,72,00,6f,00,76,00,69,00,64,00,65,00,72,00,00,00,00,00
"DefaultPrivateKeyLenBits"=dword:00000400
"RequireOnCardPrivateKeyGen"=dword:00000000
"AllowPrivateECDSAKeyImport"=dword:00000000
"AllowPrivateECDHEKeyImport"=dword:00000000
"TransactionTimeoutMilliseconds"=dword:000005dc
"AllowPrivateSignatureKeyImport"=dword:00000000
"AllowPrivateExchangeKeyImport"=dword:00000000

[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\Cryptography\Providers\Microsoft Smart Card Key Storage Provider\UM]
"Image"="scksp.dll"

[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\Cryptography\Providers\Microsoft Smart Card Key Storage Provider\UM\00010001]
@="CRYPT_KEY_STORAGE_INTERFACE"
"Flags"=dword:00000000
"Functions"=hex(7):4b,00,45,00,59,00,5f,00,53,00,54,00,4f,00,52,00,41,00,47,00,\
  45,00,00,00,00,00
```

frankmorgner:
I don't think these options are relevant.

https://github.com/OpenSC/OpenSC/issues/447 explains that we choose basecsp.dll to avoid signing and uploading the installer to MS, which is normally forced by the smartcard plug and play service.

frankmorgner:
@vletoux,can have a quick look here or into https://github.com/OpenSC/OpenSC/issues/1763 to check if we are on the right track?

vletoux:
Quick answer:
First you should use api monitor (http://www.rohitab.com/apimonitor) to trace ADCS certificate flow and it’s use of CAPI/CNG

Then the OpenSC Csp is a hack I introduced to disable the auto detection of cards. It’s an alias of the Base Smart card CSP. You cannot do it with the KSP because the KSP checks its name.
You can force the return of the Minidriver to avoid this issue.

vletoux:
Also the « alias » key for KSP is used  in smart card logon scenarios when a CSP name is passed and a KSP need to be called. It’s a Unicode string that you can display. As I said, it’s checked inside the KSP in a list of hard coded CSP

marcoh00:
Hello again,

unfortunately I have no idea what APIs the PowerShell Cmdlet calls in the background but I could offer to try to find out. I just call the PowerShell cmdlet like shown inside the issue, one time by specifying `-Provider "OpenSC CSP"` and one time by `-Provider "Microsoft Smart Card Key Storage Provider"`. Obviously, both of them can yield different results but I *always* find opensc + md logfiles after doing this. This, of course, makes sense, because it does something using the attached SC-HSM and Windows itself has no support whatsoever for this device.

While I don't know much about the CNG-APIs I know a little bit of PowerShell, C# and how to use a decompiler. I will also have a look at this API Monitor application as soon as the website stops to throw `Error establishing a database connection` errors at me :slightly_smiling_face: I will also try to build the test application and see what I can do with it.

dengert:
Based on https://github.com/OpenSC/OpenSC/issues/1763#issuecomment-577186705 using provider "OpenSC CSP" only works with CAPI, but using provider "Microsoft Smart Card Key Storage Provider" works better. 

@vletoux  in https://github.com/OpenSC/OpenSC/pull/1848#issuecomment-575799706 you said:
 "As I said, it’s checked inside the KSP in a list of hard coded CSP"

Is the KSP really hard coded, or is it in: `HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Control\Cryptography\Providers\Microsoft Smart Card Key Storage Provider` which has an Aliases key with multibyte strings: "Microsoft Base Smart Card Crypto Provider" and "Microsoft Smart Card Key Storage Provider". 

Would adding "OpenSC CSP" to the aliases allow one to use -provider "OpenSC CSP" ?

Or should we suggest we never to use "OpenSC CSP"  and always use  "Microsoft Smart Card Key Storage Provider"?

dengert:
Also see https://github.com/OpenSC/OpenSC/issues/1763#issuecomment-577237817

Issue #1715, opened 2019-06-19
Extend Gnuk ECC support (curve25519, secp256k1)


alex-nitrokey:
### Problem Description

The Gnuk/Nitrokey Start is a software implementation of the OpenPGP Card standard. It differs in some terms though. As these devices do support two curves not defined in the standard, it might be interesting to add these in OpenSC as well.

### Proposed Resolution
Add the curves for Gnuk devices only. This shall not be done [here](https://github.com/OpenSC/OpenSC/blob/19711d0a1301f0519a1348390bb996fd2925f766/src/libopensc/card-openpgp.c#L158), but in a switch-case when [actually added](https://github.com/OpenSC/OpenSC/blob/19711d0a1301f0519a1348390bb996fd2925f766/src/libopensc/card-openpgp.c#L636). Please note that the comment in said code part is not accurate anymore. Gnuk devices are recognized as such.

### Further work to do
I might implement the code myself in the future. Meanwhile this issue is a reminder for anyone eager to implement this and for others to may give head ups. For example: I do not know right now if `curve25519` and `secp256k1` are handled by OpenSC anyway etc.

szszszsz:
Connected: https://github.com/OpenSC/OpenSC/pull/319

popovec:
Hi, 

I have tested OpenSC with secp256k1 key - only ESDSA-SHA1 is working for me. MyEID card driver, OsEID card/simulation. OsEID simulation is already used to test OpenSC on travis. Example sign test with secp256k1 key  is already in OsEID-tool with small patch:

````
diff --git a/OsEID-tool b/OsEID-tool
index 9825ba7..abd2642 100755
--- a/OsEID-tool
+++ b/OsEID-tool
@@ -489,6 +489,7 @@ echo storing EC key
 cat <<EOP
 prime192v1 1.2.840.10045.3.1.1
 prime256v1 1.2.840.10045.3.1.7
+secp256k1 1.3.132.0.10
 EOP
 if [ $ECC_SIZE_MAX -ge 384 ]; then echo "secp384r1  1.3.132.0.34"; fi
 if [ $ECC_SIZE_MAX -ge 521 ]; then echo "secp521r1  1.3.132.0.35"; fi
````


OpenSC need to be patched too:
````
diff --git a/src/libopensc/card-myeid.c b/src/libopensc/card-myeid.c
index d4ed17c4..733b2331 100644
--- a/src/libopensc/card-myeid.c
[...]

PR #1853, opened 2019-11-07
openpgp: add nistp256 and secp251k1 curves for gnuk devices

alex-nitrokey:
This patch adds support for nistp256 and secp251k1 curves for Gnuk devices. This partly solves #1715, besides curve25519 support. I will open another issue for this task shortly as it can not be solved in openpgp driver only, but needs to be addressed in OpenSC in general.

This was tested with a Nitrokey Start (equivalent to Gnuk) in firmware version 1.2.14.


Jakuje:
Looks good.

PR #1868, opened 2019-11-25
Fix macOS uninstallation

frankmorgner:
##### Checklist

- [x] macOS tokend is tested


metsma:
I think these applications should be in  /Applications/Utilities folder

frankmorgner:
On my system, I can only see internal Software in that directory. Can you link some Apple documentation for application developers regarding the use of the system folders?

metsma:
oh I see this has changed since 10.15 but from Finder you can see both system and installed applications

PR #1872, opened 2019-11-26
Fixing invalid signature with 3072 RSA bits in GemsafeV1

### LINKED ISSUE #1871: Invalid signature with 3072 RSA bits portuguese eID cards (GemsafeV1)
### Problem Description
After #1858 it's possible to sign data with a 3072 or 4096 bit RSA key, but the signature is invalid.
Using OpenSC nightly build 2019-11-23_ce71b171.

### Proposed Resolution
Increase buffer size to 518 bytes.
```
--- a/src/libopensc/card-gemsafeV1.c
+++ b/src/libopensc/card-gemsafeV1.c
@@ -455,8 +455,8 @@ static int gemsafe_compute_signature(struct sc_card *card, const u8 * data,
 {
        int r, len;
        struct sc_apdu apdu;
-       u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
-       u8 sbuf[SC_MAX_APDU_BUFFER_SIZE];
+       u8 rbuf[SC_MAX_APDU_BUFFER_SIZE + 257];
+       u8 sbuf[SC_MAX_APDU_BUFFER_SIZE + 257];
        sc_context_t *ctx = card->ctx;

        SC_FUNC_CALLED(ctx, SC_LOG_DEBUG_VERBOSE);
```
If this change is ok I can create a PR.

### Steps to reproduce
1) Sign a PDF file with a 2019 portuguese eID card (3072 RSA bits)
2) Open PDF in Adobe Reader, signature will be invalid with error: "Internal cryptographic library error, Error Code 0x2719"

### Notes
I increased buffer size to 518 bytes, sign a PDF with a 2019 portuguese  eID card (citizen card) then opened it in Adobe Reader (2019.021.20056) and the signature was valid, screenshot attached.
I also tested with an older card from 2015 and it keeps working.

image

AdriaoNeves:
Fixes #1871
Card: GemSAFE V1
PTEID

### DIFF
--- a/src/libopensc/card-gemsafeV1.c
+++ b/src/libopensc/card-gemsafeV1.c
@@ -32,6 +32,7 @@
 #define GEMSAFEV3_ALG_REF_FREEFORM	0x02
 #define GEMSAFEV3_ALG_REF_SHA1		0x12
 #define GEMSAFEV3_ALG_REF_SHA256	0x42
+#define MAX_RESP_BUFFER_SIZE 2048
 
 static struct sc_card_operations gemsafe_ops;
 static struct sc_card_operations *iso_ops = NULL;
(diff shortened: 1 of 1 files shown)


Jakuje:
I would rather put there directly value 1024 or 2048, rather than this suspicious addition. Or a local define such as
```
#define MAX_RESP_BUFFER_SIZE 2048
```
Otherwise it looks good, assuming it works for you :)

AdriaoNeves:
New commit with your sugestion. I also changed other functions to use new buffer size.
I tested with two cards (with 1024 and 3072 bits) and signatures are valid.
Thank you for the quick answer :)

PR #1878, opened 2019-12-02
Starcos3x extensions

jozsefd:
##### Checklist

- [x ] Windows minidriver is tested
- [ ] macOS tokend is tested

I have adjusted the actual starcos driver for the cards of a large German federal organisation. They have StarCOS 3.4 and 3.5 cards in use, I changed the following:
- the cards have alphanumeric PINs, so they use ASCII encoding instead of GLP. I didn't want to add a complex logic to determine this feature, instead, I added a card driver specific option "pin_format_ascii"
- added a new StarCOS 3.5 ATR (v3.52 afaik)
- added option for 3072 bit keys on StarCOS 3.5
- some card reader devices (e.g. Alcor Micro USB Smart Card Reader) may fail with long extended APDUs, so I added an option to suppress parsing the EF.ATR ("disable_parse_atr") in order to use the configured/default transmit sizes of StarCOS 3.5 cards
- the path caching has caused problems in several cases (e.g.: unlock windows with smart card), so added an option to disable it ("disable_path_cache")
- adjusted the path fixing "unify path (the first FID should be MF)" in function starcos_select_file. StarCOS 3.5 works now the same way as v3.4
I have tested it with StarCOS 3.4 and StarCOS 3.5 on Ubuntu with pkcs15-tool and on Windows 7, the minidriver and the PKCS#11 Module in Firefox.


frankmorgner:
Thanks for your feedback! I hope we can use your insights for a fix that works for most users.

- If you want to avoid path caching, you should set `keep_alive` in the config, which invalidates the cache once the card is unlocked. Since path caching has already caused problems in the past, you may want to remove path caching entirely from the driver.
- If extended length APDUs are causing trouble, you should limit the size in the `reader pcsc` section (i.e. `max_send_size`/`max_recv_size`)
- I'd appreciate it if you could look again into parsing the card's attributes regarding the PIN format rather than using an option. Your solution works if and only if the user has deeper knowledge about her card; the average user most likely won't have this understanding.
- If adding new options, you should document them in `doc/files/opensc.conf.5.xml.in` `opensc.example.conf`

jozsefd:
Frank, thank you for your quick reply! 
- I will check the keep_alive option, if it works then I can revoke the option for disabling path caching. On the other hand if path caching is unstable then we may want to set it to optional, i.e. disabled by default which can be turned on per config
- As far as I can remember, limiting the size in the config did not work in this driver, the value from EF.ATR has overridden it.
- OK, I can add card attribute parsing, first I didn't want to add extra complexity and additional transmits which may increase the initialization time.
- Thanks for the hint, I will expand the doc as well!

frankmorgner:
maybe you can send a log with the problem for reading EF.ATR

jozsefd:
It seems that I cannot reproduce the EF.ATR (and at the end a max_send/recv_size) issue anymore, which happened by a reader where the feature query failed (quote from an older log):
> Probing PC/SC reader
> Adding new PC/SC reader 'Alcor Micro USB Smart Card Reader 0'
> Alcor Micro USB Smart Card Reader 0 check
> reader-pcsc.c:1068:detect_reader_features: called
> Requesting reader features ...
> Alcor Micro USB Smart Card Reader 0:SCardControl failed: 0x00000001

then the card reported a large transmit size despite the setting of pcsc/max_send_size|max_recv_size:

> card info name:'STARCOS 3.5', type:7003, flags:0x0, max_send/recv_size:1920/1920

which caused a binary read APDU:

> reader 'Alcor Micro USB Smart Card Reader 0'
> Outgoing APDU (7 bytes):
> 00 B0 00 00 00 07 80 .......

with a Wrong length error:

> Incoming APDU (2 bytes):
> 67 00 g.

Anyway, it seems none of the new starcos options make sense, so I am going to add the PIN format detection and submit a new PR.

jozsefd:
I'd like to revoke this PR, please see the new one: Starcos3x pin format #1882

PR #1879, opened 2019-12-02
Card IsoApplet - add support for no_padding algortihm RSA-X-509 (no padding)/raw RSA) …

gesteur:
RSA-X-509 (CKM_RSA_X_509) seems to be used by OpenSSL during the authentication handshake for OpenVPN with hardware tokens. 
OpenSSL pre-pads the data to be sent to the token. without the support of this algorithm OpenVPN client cannot connect using IsoApplet Card.

With normal PKCS#1 padding I have the same issue as described by another user in the following link 
http://openssl.6102.n7.nabble.com/Issue-with-smartcard-authentication-for-openvpn-td76415.html

The RSA-X-509 mechanism (mech=3) is used with OpenSSL 1.1.1 and TLS version >= 1.2

- [ x] PKCS#11 module is tested

### DIFF
--- a/src/libopensc/card-isoApplet.c
+++ b/src/libopensc/card-isoApplet.c
@@ -31,6 +31,7 @@
 
 #define ISOAPPLET_ALG_REF_ECDSA 0x21
 #define ISOAPPLET_ALG_REF_RSA_PAD_PKCS1 0x11
+#define ISOAPPLET_ALG_REF_RSA_NOPAD 0x12
 
 #define ISOAPPLET_API_VERSION_MAJOR 0x00
 #define ISOAPPLET_API_VERSION_MINOR 0x06
@@ -249,6 +250,7 @@ isoApplet_init(sc_card_t *card)
 	flags = 0;
 	/* Padding schemes: */
 	flags |= SC_ALGORITHM_RSA_PAD_PKCS1;
+	flags |= SC_ALGORITHM_RSA_RAW;
 	/* Hashes are to be done by the host for RSA */
 	flags |= SC_ALGORITHM_RSA_HASH_NONE;
 	/* Key-generation: */
@@ -1106,6 +1108,10 @@ isoApplet_set_security_env(sc_card_t *card,
 			if( env->algorithm_flags & SC_ALGORITHM_RSA_PAD_PKCS1 )
 			{
 				drvdata->sec_env_alg_ref = ISOAPPLET_ALG_REF_RSA_PAD_PKCS1;
+			}
+		   	else if( env->algorithm_flags & SC_ALGORITHM_RSA_RAW )
+			{
+				drvdata->sec_env_alg_ref = ISOAPPLET_ALG_REF_RSA_NOPAD;
 			}
 			else
 			{


Jakuje:
This breaks the isoApplet tests in CI, which worked before. Can you clarify what isoApplet are you using? The one we use really does not support anything but PKCS#1 padding:

https://github.com/philipWendland/IsoApplet/blob/master/src/net/pwendland/javacard/pki/isoapplet/IsoApplet.java#L1157

gesteur:
I have modified the IsoApplet to support RSA no padding in my fork and created a pull request, waiting for it to be merged, there is almost no activity on the official repo. 

https://github.com/philipWendland/IsoApplet/pull/19

i'm sure this will benefit other users too which are struggling with this card.

Jakuje:
I am not sure if this is the way to go. The RAW RSA operations are insecure and can lead to exposing of the private key. I think the isoApplet has the limitation intentionally to avoid these security issues. If you are using TLS, I think you might want to have a look into the implementation of RSA-PSS padding, which is secure and supported by all of the newer TLS and OpenSC internals,  rather than downgrading security of the isoApplet by allowing raw RSA operations.

gesteur:
I understand that, but I cant really change openSSL to force them using one algo instead of an other.
There must be a reason why with the newer openSSL version they are using the CKM_RSA_X_509 (no padding), right? it seems openSSL does the padding. 
looking around forums it seems that openSSL is using CKM_RSA_X_509 padding mechanism, so I'm a bit stuck because i cant influence that.
if you have any suggestions, I m happy to look into it.

Jakuje:
It is using the RSA_X_509 because it negotiated some algorithm that is not directly supported by pkcs11 library/token/driver (again, assuming RSA-PSS, but it can be some other -- you will find it in the logs). The RSA_X_509 is the raw RSA, which is generally used as a fallback, but it does not mean that enabling this is the only way to go. Unless the change will be accepted in the isoApplet, I do not think we should have this modification in OpenSC.

gesteur:
thanks for the reply, at the moment it doesnt seem pkcs11-helper nor libp11 do support PSS, they only support RSA_PKCS1_PADDING and RSA_X_509 (as far as I can see from the source code).
RSA_PKCS1_PADDING doensn't seem to be used by newer version of openssl and as you say therefore it might default to RSA_X_509. 

I'm not an expert at all, was just trying to get my smartcard to work with openvpn, I saw other tokens had similar issues.
e.g. https://github.com/OpenSC/libp11/issues/176#issuecomment-325909510 (commit is not in master though and it's not pkcs11-helper which is standard lib used by openssl)

thanks again for your help with this, I guess we just need to wait till the helper libs get support for PSS. I ll close this PR in the meantime.

Jakuje:
I think libp11 should support RSA PSS out of the box:

https://github.com/OpenSC/libp11/blob/master/tests/rsa-pss-sign.c#L229

It also worked for me when I tried it last time with PIV cards (they support the RSA_X_590, but the operations went through the RSA_PSS mechanism) from curl (built with openssl and configured with engine from libp11). So the only missing bits is really the applet (isoApplet) and OpenSC driver.

I did not implement anything in the isoApplet, but the RSA-PSS padding should not be that complicated to add. The driver in OpenSC needs just minor corrections and calculation of padding in card should not hurt that much.

For example I have quite-ISO-compliant Gemalto IDPrime smart card (#1772) which supports RSA-PSS padding in card (see the flags definition and set security environment functions).

For examples how the RSA-PSS padding is implemented in OpenSC (or you can also read through specs or OpenSSL version):

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/padding.c#L279

It should be pretty straight-forward to reimplement it in Java.

philipWendland:
As I wrote in the original issue (https://github.com/philipWendland/IsoApplet/issues/18), I agree with @Jakuje - RAW RSA should not be exposed by the token.

The most secure way would be to use JavaCard primitives for the padding and not do it in Java(card) itself. 
The JC 2.2.2 API does only allow PSS with SHA1 (for ciphers, OAEP would be less problematic).

After having a quick look -- the JC 3.1 API does not allow pre-hashed data with PSS padding as well.  I am wondering - does implementing PSS padding in Java(card) introduce possible attacks vectors, e.g. side-channel attacks?

Jakuje:
@philipWendland Right. Using JavaCard API would be preferred way. RSA-PSS with SHA1 is probably better than nothing. OAEP would not be useful here at all as it is not that easy to fake the signature as with the PKCS 1.5 counterpart.

I think adding padding to digest can not introduce any possible attack vectors since:
 * The input is already digest (irreversible) which prevents any way of retrieving plaintext
 * The operation does not use any part of private key that could leak (just the modulus length)

As for the suspicion what changed in OpenVPN, they probably disabled PKCS#1.5 schemes (as it is no longer supported in TLS 1.3 either) and the fallback to raw RSA. But most of the libraries do not check whether the PKCS#11 device can use the (earlier negotiated) mechanism and just fail.

gesteur:
It would be really useful if we could use standard openvpn,opensc repos for javacard token using isoApplet. Unless there are other applets which implement more modern padding schemes.

Do I understand correctly that the only way to use openVPN with IsoApplet is to code the PSS padding in the java code?

Jakuje:
I have OpenSSL 1.1.1d on Fedora.

One option might be to use different applet, that already supports the RSA-PSS or raw RSA. One example might be PivApplet, which supports raw RSA even for the possibility of lower security for your keys:

https://github.com/arekinath/PivApplet/

dengert:
Maybe the author of pkcs11-helper @alonbl  would like to comment on this. It looks like pkcs11-helper could accept PSS as being valid. 

A spy trace of the pkcs11 calls would help show what mechanisms are presented to the helper or OpenSSL.

philipWendland:
> It would be really useful if we could use standard openvpn,opensc repos for javacard token using isoApplet. Unless there are other applets which implement more modern padding schemes.

Yes, I agree. 

> Do I understand correctly that the only way to use openVPN with IsoApplet is to code the PSS padding in the java code?

It looks like it.
For 3.x JavaCards, I remember something about pre-hashing. But I am not sure, it has been some years.
It could be possible to pre-hash the data in OpenSC and perform the last SHA2 hashing step in the IsoApplet.

> Found other references
> [openssl/openssl#7968](https://github.com/openssl/openssl/issues/7968)
> 
> @Jakuje which openssl version were you using ?
> 
> Btw the RSA_X_509 is not insecure per se, it shouldn't be used to pass user data, but is perfectly fine if the data passed to it is already padded (which openssl does).
> If this shouldnt be used, i m pretty sure openssl wouldnt allow it.

I know. The point is to limit security risks *at the token level*.

gesteur:
Hi again,

could you guys clarify something to me as I'm a bit puzzled.
Does OpenSC do the padding itself when we provide a RSA-PSS algo to the pcks11-tool and then pass raw RSA ( RSA_X_509 / RSA_NO_PADDING) to the token? 

I'm asking because without RSA_X_509 support I can't use the following parameters to sign data:
`pkcs11-tool --id 01 -s  -m RSA-PKCS-PSS --hash-algorithm SHA256 --mgf MGF1-SHA256   ....`

result:
```
Using signature algorithm RSA-PKCS-PSS
PSS parameters: hashAlg=SHA256, mgf=MGF1-SHA256, salt_len=32 B
```

obviously no signature created.


But if I activate the "SC_ALGORITHM_RSA_RAW" flag in the card driver, I get much more supported algos with pkcs11-tool -M
and the following param works:
`pkcs11-tool --id 01 -s  -m RSA-PKCS-PSS --hash-algorithm SHA256 --mgf MGF1-SHA256`   ....
result:
```
Using signature algorithm RSA-PKCS-PSS
PSS parameters: hashAlg=SHA256, mgf=MGF1-SHA256, salt_len=32 B
```

I can also verify the signature using:
```
openssl dgst -keyform DER -verify pubkey.der -sha256 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -sigopt rsa_mgf1_md:sha256 -signature data2.sig data
Verified OK
```

This leads me to believe that the padding must be done in software since clearly JavaCard and in this case isoApplet dont do any RSA-PSS padding on card.
So we might think that the token is actually doing RSA-PSS but instead it's just doing RAW RSA on pre-padded data, I would like to hear your opinions about this, since it's a bit confusing to me

Jakuje:
Yes. The OpenSC (master) can provide the padding (RSA-PSS or PKCS1.5) if the card does not natively support that, but supports raw RSA.

gesteur:
> It is using the RSA_X_509 because it negotiated some algorithm that is not directly supported by pkcs11 library/token/driver (again, assuming RSA-PSS, but it can be some other -- you will find it in the logs). The RSA_X_509 is the raw RSA, which is generally used as a fallback, but it does not mean that enabling this is the only way to go. Unless the change will be accepted in the isoApplet, I do not think we should have this modification in OpenSC.

hi @Jakuje 
I don't believe it's because the token doesn't support it, it's more with the upgrade to OpenSSL 1.1.1 which uses RSA_PKCS1_PSS_PADDING instead of RSA_PKCS1_PADDING.

Investigating deeper the RSA_PKCS1_PSS_PADDING doesn't call the encryption function with that padding, instead it calls it with RSA_NO_PADDING, suggesting the data is pre-padded.

So to me it seems strange that it's possible to actually use RSA-PSS native on a token using OpenSSL, as OpenSSL would clearly not pass RSA-PSS padding algo to the token. 

Just thought to share my findings, please correct me if I misunderstood anything. 


[https://github.com/openssl/openssl/blob/dd63f9bbfc95931e4f41558fbc7a2ffb68b0eef0/crypto/rsa/rsa_pmeth.c#L268](https://github.com/openssl/openssl/blob/dd63f9bbfc95931e4f41558fbc7a2ffb68b0eef0/crypto/rsa/rsa_pmeth.c#L268)

else if (rctx->pad_mode == RSA_PKCS1_PSS_PADDING) {
            if (!setup_tbuf(rctx, ctx))
                  return -1;
              if (!RSA_padding_add_PKCS1_PSS_mgf1(rsa,
                                                rctx->tbuf, tbs,
                                                rctx->md, rctx->mgf1md,
                                                rctx->saltlen))
                return -1;
            ret = RSA_private_encrypt(RSA_size(rsa), rctx->tbuf,
                                      sig, rsa, RSA_NO_PADDING);

PR #1896, opened 2019-12-29
opensc-explorer: make '--card-driver ?' list all available drivers

marschap:
Make opensc-explorer a bit more user friendly by treating the question mark
given as argument to option '--card-driver' special: list all available
drivers instead of stupidly bailing out.

##### Checklist

### DIFF
--- a/src/tools/piv-tool.c
+++ b/src/tools/piv-tool.c
@@ -52,6 +52,7 @@
 
 #include "libopensc/opensc.h"
 #include "libopensc/cardctl.h"
+#include "libopensc/cards.h"
 #include "libopensc/asn1.h"
 #include "util.h"
 #include "libopensc/sc-ossl-compat.h"
@@ -80,7 +81,6 @@ static const struct option options[] = {
 	{ "in",			1, NULL, 		'i' },
 	{ "send-apdu",		1, NULL,		's' },
 	{ "reader",		1, NULL,		'r' },
-	{ "card-driver",	1, NULL,		'c' },
 	{ "wait",		0, NULL,		'w' },
 	{ "verbose",		0, NULL,		'v' },
 	{ NULL, 0, NULL, 0 }
@@ -98,7 +98,6 @@ static const char *option_help[] = {
 	"Input file for cert",
 	"Sends an APDU in format AA:BB:CC:DD:EE:FF...",
 	"Uses reader number <arg> [0]",
-	"Forces the use of driver <arg> [auto-detect; '?' for list]",
 	"Wait for a card to be inserted",
 	"Verbose operation. Use several times to enable debug output.",
 };
@@ -469,7 +468,7 @@ static void print_serial(sc_card_t *in_card)
 
 int main(int argc, char *argv[])
 {
-	int err = 0, r, c, long_optind = 0;
+	int err = 0, r, c;
 	int do_send_apdu = 0;
 	int do_admin_mode = 0;
 	int do_gen_key = 0;
@@ -478,9 +477,7 @@ int main(int argc, char *argv[])
 	int compress_cert = 0;
 	int do_print_serial = 0;
 	int do_print_name = 0;
-	int do_list_card_drivers = 0;
 	int action_count = 0;
-	const char *opt_driver = NULL;
 	const char *out_file = NULL;
 	const char *in_file = NULL;
 	const char *cert_id = NULL;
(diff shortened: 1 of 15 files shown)


frankmorgner:
`opensc-tool` has `--list-drivers` for this, while many other tools also have `--card-driver` without the special value of `?`. I don't have a special preference, but the options and their use should be unified across the tools.

marschap:
@frankmorgner 
I opted to implement '--card-driver ?' functionality for all tools using option '--card-driver' (and consolidated 'dnie-tool' to use '--card-driver' instead of '--driver' while at it).

The reason behind it is simple: we should not force our users to search for a different tool
[opensc-tool] to get the list of drivers available.

Please consider merging for the next release.

frankmorgner:
I see, makes sense (at least for the generic tools).

However, for card specific tools (piv-tool, dnie-tool, cardos-tool) this option isn't very usefull. From the user's perspective, the opposite should be done. For example, piv-tool should force the PIV driver and abort if the inserted card is not compatible.

marschap:
@frankmorgner I removed the option '--card-driver' from the card specific tools, and instead forced the correct card driver to load and to fail if no appropriate card was found.

While at it I did some slight re-factoring of the 3 tools, even correcting a - in my opinion larger error - error in the piv-tool.

I did it explicitly as additional commits to allow retrieving the code should it be necessary at some point in the future.

frankmorgner:
Thanks for taking the time for cleaning this a little bit. What error in piv-tool do you mean?

Are you sure about checking for EOF for getopt_long? I think it should be `-1`:
>Return Value
>
>If an option was successfully found, then getopt() returns the option character. If all command-line options have been parsed, then getopt() returns -1. If getopt() encounters an option character that was not in optstring, then '?' is returned. If getopt() encounters an option with a missing argument, then the return value depends on the first character in optstring: if it is ':', then ':' is returned; otherwise '?' is returned.
>
>getopt_long() and getopt_long_only() also return the option character when a short option is recognized. For a long option, they return val if flag is NULL, and 0 otherwise. Error and -1 returns are the same as for getopt(), plus '?' for an ambiguous match or an extraneous parameter.

marschap:
Hi @frankmorgner, 

The piv-tool error was that it ignored the option -r / --reader, because "r:" was missing in the getopt_long parameters.

I fixed the getopt_long() result checks to correctly use -1 instead of EOF
Bonus: not only for the initial ones, but for all tools.

src/tools/piv-tool.c
```diff
@@ -603,19 +596,25 @@ int main(int argc, char *argv[])
...
 	if (err)
 		goto end;
 
+	/* fail if card is not a PIV card */
+	if (card->type < SC_CARD_TYPE_PIV_II_BASE || card->type >= SC_CARD_TYPE_PIV_II_BASE+1000) {
+		fprintf(stderr, "Card type %X: not a PIV card\n", card->type);
+		err = 1;
+		goto end;
+	}
+
```
frankmorgner: I think you don't need to check the type of the driver if you force it with `sc_set_card_driver()`. Rather, you should print the error that "no compatible PIV card has been found"  above where `util_connect_card()` cannot connect to a card.

(Same for dnie-tool and cardos-tool...)

marschap: Sorry I need to disagree.

At least the `dnie` driver will happily connect to the card [i.e. `util_connect_card()` succeeds], and spew unknown APDUs at  unsuspecting cards of any type if this check is left out. 
As the additional check does not hurt if `util_connect_card()` fails, I rather opt for safety here.

Please commit.

frankmorgner: util_connect_card calls sc_connect_card, which calls dnie_init, which errors if a card doesn't support the DNIe commands, doesn't it?

marschap: I do not know about the internals of the `dnie` driver.
But when I compiled `dnie-tool`and ran it against my Yubikey (which definitely is no DNIe card), `util_connect_card()` definitely succeeded and without the additional check, `dnie-tool` spat it's APDUs to my Yubikey.
That's why I think the checks are necessary. In any case they do not hurt.

Please merge!

dengert: Another approach is to set the default value  for `opt_driver` to the name of the driver for card specific tools.  So piv-tool.c would have `const char *opt_driver = "PIV-II";` and dnie-tool.c would have ` const char *opt_driver = "dnie";` User could still override this for what ever reason. But only to appropriate driver would be used. 

Six tools support `--card-driver` which have `static const char *opt_driver = NULL;`  So another approach in the card specific drivers: `dnie-tool.c`, `cardos-tool.c` `piv-tool-c` and `westcos-tool.c`  change to something like: `static const char *opt_driver = "dnie";` or `const char *opt_driver = "PIV-II";`. User could still use the `--card-driver` if they wanted too. 

The same could be done in tools that do not have `--card-driver` option but are card specific. These are: `cryptoflex-tool.c`, `egk-tool.c`, `egk-tool-cmdline.c`, `gids-tool.c`, `goid-tool.c`, `goid-tool-cmdline.c`, `iasecc-tool.c`, `netkey-tool.c`, `npa-tool.c`, `npa-tool-cmdline.c` and `openpgp-tool.c`. 

If you go with testing the card type, the above list of card specific tools should check for their card types too.

marschap: @dengert of course there are many ways how to do it. 
Frank asked me to extend the original patch, which had very little to do with the topics at hand, in one specific way. Which I did.
I am willing to write another patch, but only after this one gets merged and you and Frank agree on the approach.

frankmorgner: npa-tool and goid-tool don't rely on a specific card driver.

@dengert feel free to set the card driver as well in the other tools you mentioned. The code in this PR is a good starting point.

PR #1897, opened 2019-12-31
tcos: fix indentation

jmastr:
Increase readability. No content-based changes.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-tcos.c
+++ b/src/libopensc/card-tcos.c
@@ -86,7 +86,7 @@ static int tcos_match_card(sc_card_t *card)
 
 static int tcos_init(sc_card_t *card)
 {
-        unsigned long flags;
+	unsigned long flags;
 
 	tcos_data *data = malloc(sizeof(tcos_data));
 	if (!data) return SC_ERROR_OUT_OF_MEMORY;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-tcos.c
```diff
@@ -124,88 +124,88 @@ static int tcos_construct_fci(const sc_file_t *file,
...
-                memcpy (buf+12, "\xd6\x00\x00\x00\xff\xff", 6); /* upd bin */
-                memcpy (buf+18, "\x60\x00\x00\x00\xff\xff", 6); /* admin grp*/
-                n = 24;
-        }
-        sc_asn1_put_tag(0x86, buf, n, p, sizeof (buf), &p);
-
-        
-        /* fixup length of FCI */
-        out[1] = p - out - 2;
+	else {
```
Jakuje: I tried to find the last discussion about coding style, but I was not successful. The only thing I found was some ancient wiki referencing linux kernel coding style, which allows only `} else {` on single line so I would try to stick with that as we rewrite things.

https://github.com/OpenSC/OpenSC/wiki/Development-Policy

src/libopensc/card-tcos.c
```diff
@@ -351,28 +351,28 @@ static int tcos_select_file(sc_card_t *card,
...
+			apdu.p1 = 8;
+			if (pathlen >= 2 && memcmp(path, "\x3F\x00", 2) == 0) path += 2, pathlen -= 2;
+			if (pathlen == 0) apdu.p1 = 0;
+			break;
+		case SC_PATH_TYPE_PARENT:
+			apdu.p1 = 3;
+			pathlen = 0;
+			break;
+		default:
+			SC_FUNC_RETURN(ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_INVALID_ARGUMENTS);
```
Jakuje: The same as above -- the switch and case should be on the same indentation level according to kernel coding style:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst#n47

src/libopensc/card-tcos.c
```diff
@@ -436,9 +436,9 @@ static int tcos_list_files(sc_card_t *card, u8 *buf, size_t buflen)
 		r = sc_check_sw(card, apdu.sw1, apdu.sw2);
 		LOG_TEST_RET(ctx, r, "List Dir failed");
 		if (apdu.resplen > buflen) return SC_ERROR_BUFFER_TOO_SMALL;
-		sc_log(ctx, 
-			 "got %"SC_FORMAT_LEN_SIZE_T"u %s-FileIDs\n",
-			 apdu.resplen / 2, p1 == 1 ? "DF" : "EF");
+		sc_log(ctx,
+				"got %"SC_FORMAT_LEN_SIZE_T"u %s-FileIDs\n",
+				apdu.resplen / 2, p1 == 1 ? "DF" : "EF");
```
Jakuje: +1 for removing the bogus spaces, but this seems to me as over-intentation. Similar problems are in the following diffs.

src/libopensc/card-tcos.c
```diff
@@ -698,10 +698,10 @@ static int tcos_get_serialnr(sc_card_t *card, sc_serial_number_t *serial)
...
 {
 	switch (cmd) {
-	case SC_CARDCTL_TCOS_SETPERM:
-		return tcos_setperm(card, !!ptr);
-	case SC_CARDCTL_GET_SERIALNR:
-		return tcos_get_serialnr(card, (sc_serial_number_t *)ptr);
+		case SC_CARDCTL_TCOS_SETPERM:
+			return tcos_setperm(card, !!ptr);
+		case SC_CARDCTL_GET_SERIALNR:
+			return tcos_get_serialnr(card, (sc_serial_number_t *)ptr);
```
Jakuje: Again, over-indentation as mentioned in the previous comments

jmastr:
Hey @Jakuje Thanks for your feedback. I incorporated your comments. Please check the diff again.

According to https://www.kernel.org/doc/html/v4.12/process/coding-style.html#placing-braces-and-spaces:

```
This does not apply if only one branch of a conditional statement is a single statement; in the latter case use braces in both branches:

if (condition) {
        do_this();
        do_that();
} else {
        otherwise();
}
```

So I added a few brackets.

PR #1900, opened 2020-01-07
Check if card name is null in acos atrust driver

lewisporter:
`opensc-tool` segfaults when trying to read a card that doesn’t populate the name field. This commit adds a null check before calling `strcmp()`.

The card in question is unknown and subject to research. This is probably quite easy to reproduce, though.

### DIFF
--- a/src/libopensc/card-atrust-acos.c
+++ b/src/libopensc/card-atrust-acos.c
@@ -123,7 +123,7 @@ static int atrust_acos_init(struct sc_card *card)
 		| SC_ALGORITHM_RSA_HASH_RIPEMD160
 		| SC_ALGORITHM_RSA_HASH_MD5_SHA1;
 
-	if (!strcmp(card->name, ACOS_EMV_A05))
+	if (card->name != NULL && !strcmp(card->name, ACOS_EMV_A05))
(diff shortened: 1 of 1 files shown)


lewisporter:
Some further investigation shows that `atrust_acos_init` is called before `sc_connect_card` is, and so the line you cited does not have an effect yet. I have rewritten `atrust_acos_match_card` to use `_sc_match_atr`, which appears to be more in-line with the convention for this codebase. 

Does it make sense for you that I modify `atrust_acos_match_card` to set a name there based on flags in the `sc_atr_table` struct, or is there a more fundamental problem that I am missing?

frankmorgner:
Indeed, using `_sc_match_atr` would be the cleaner approach (`atrust_acos_match_card` currently seems to basically duplicate the code of `_sc_match_atr`). However, regarding the issue with the card's name, I'm not sure if it will be solved, because the card's name will still be initialized in the same manner even with the change to `_sc_match_atr`.

Anyway, if you see the problem disappear, this change would be good to go.

frankmorgner:
card.c will only add the driver name for card name when init is finished. so your current change is good enough.

please make a new pull request with the change for using _sc_match_card to avoid code duplication, thanks.

PR #1914, opened 2020-01-26
PGPtool fixes

marschap:
Please find attached a series of commits that
* refactor `card-openpgp`
  - split out type definitions and constants into a header file
    so that the info can be used consistently in `openpgp-tool`
  - correctly determine the max. DO size for special DOs
    With v3+ cards this size is given in the "extended capabilities" DO 00C0
* update `openpgp-tool`
  - force use of the `openpgp` driver, similar to other tools recently updated
  - use the  max_specialDO_size variable from the driver to correctly deal with v3+ cards.

##### Checklist


Jakuje:
LGTM. The CI failure is related to recent changes in libcacard, which was addressed on monday so we should be good now

https://gitlab.freedesktop.org/spice/libcacard/merge_requests/12

frankmorgner:
thanks

Issue #1913, opened 2020-01-25
pkcs11-tool: Insecure default template attributes allow WRAP & DECRYPT attack


crtxcr:
### Problem Description
pkcs11-tool sets wrap and unwrap to true. There is no way to unset wrap and unwrap when generating keys. 
Furthermore, this default template is a classic PKCS#11 attack, see e. g. https://link.springer.com/content/pdf/10.1007%2F978-3-540-45238-6_32.pdf section 2.3

### Proposed Resolution
- Do not set CKA_WRAP and CKA_UNWRAP with the value of to true in templates by default
- Allow specifcing by command line whether CKA_WRAP and CKA_UNWRAP should be set to true or not.

frankmorgner:
Thanks for the links. OpenSC has just released the first version which actually supports a card that's capable of wrapping a key, so I guess, you're talking about using pkcs11-tool with different pkcs11 modules. Anyway,

I'll look at this for the next release. Feel free to make a pull request if you find the time.

frankmorgner:
@hongquan the issues mostly relate to the token's implementation, you may want to have a look...

frankmorgner:
I've created https://github.com/OpenSC/OpenSC/pull/1915 which uses the proposed solution.

crtxcr:
>so I guess, you're talking about using pkcs11-tool with different pkcs11 modules

Correct. I am implementing (parts of) a pkcs11 token right now and I am testing some things with pkcs11-tool. This token prevents the WRAP & DECRYPT attack above by checking whether wrap and decryption related flags are set, although it will not implement the wrap & unwrap functionality soon. But I at least needed the option to disable disable wrap attribute, because my token would not allow me to create the keys.

The question now of course is whether OpenSC itself wants to take a similar approach. It probably was never relevant since it didn't support C_WrapKey until recently. It is unfortunate that by following PKCS#11 strictly you introduce problems like this though. But maybe this should be discussed in a separate issue?

dengert:
A corollary to this issue is: Unless the card can prevent  the WRAP & DECRYPT attack,(which your card can do,  adding OpenSC support to prevent the attack does not stop a hacker from using other software or a modified OpenSC to still do the attack.

PR #1915, opened 2020-01-27
pkcs11-tool: add --usage-wrap (disabled by default)

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/1913

##### Checklist

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -137,6 +137,7 @@ enum {
 	OPT_KEY_USAGE_SIGN,
 	OPT_KEY_USAGE_DECRYPT,
 	OPT_KEY_USAGE_DERIVE,
+	OPT_KEY_USAGE_WRAP,
 	OPT_PRIVATE,
 	OPT_SENSITIVE,
 	OPT_EXTRACTABLE,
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -197,6 +198,7 @@ static const struct option options[] = {
 	{ "usage-sign",		0, NULL,		OPT_KEY_USAGE_SIGN },
 	{ "usage-decrypt",	0, NULL,		OPT_KEY_USAGE_DECRYPT },
 	{ "usage-derive",	0, NULL,		OPT_KEY_USAGE_DERIVE },
+	{ "usage-wrap",	0, NULL,		OPT_KEY_USAGE_WRAP },
```
crtxcr: Tabs are not aligned here

crtxcr:
I wonder whether a separate option for unwrap could be useful? So you could have a keypair that can wrap, but not unwrap. However, I am sure whether this case can be expected to occur in practise though.

frankmorgner:
> I wonder whether a separate option for unwrap could be useful? So you could have a keypair that can wrap, but not unwrap. However, I am sure whether this case can be expected to occur in practise though.

I think I'll leave it as is to keep it simple...

PR #1921, opened 2020-01-29
Fix current oss-fuzz issues

Jakuje:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-cac.c
+++ b/src/libopensc/card-cac.c
@@ -105,6 +105,8 @@
 #define CAC_ACR_AMP                   0x20
 #define CAC_ACR_SERVICE               0x21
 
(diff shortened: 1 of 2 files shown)


frankmorgner:
Although this solves the problem of looping infinitesimally, don't you think separating processing from parsing (without card commands) would be a good idea? Is it possible that CCC is present multiple times?

Jakuje:
I was checking the specification, but there was really no specific pointer whether there can be more of the "NEXT CCC" elements in the file nor how deep it can go. I think neither of the cards that I used had this so I decided for this solution.

https://www.nist.gov/publications/government-smart-card-interoperability-specification-version-21

PR #1923, opened 2020-01-30
opensc-notify: handle reader insertion/removal events 

### LINKED ISSUE #1874: opensc-notify.exe(from OpenSC 0.20.0-rc2) - Win10( ver1809) - too high processor load
### Problem Description
Too high processor load on win10 enterprise (ver 1809, build 17763.805 ) by opensc-notify.exe without activity. Still, continuous 30%.
My CPU: Intel Core i5 - 5300U 2,3GHz

### Proposed Resolution
I don't know if the opensc-notify process can be disabled in the opensc.conf configuration. That would be temporary help.

### Steps to reproduce
Only install OpenSC-0.20.0-rc2 version.

### Logs
obrazek

frankmorgner:
- If readers are attatched, the new reader is probed for a card to check
if a notification needs to be sent
- removal of readers are not notified to the user, we assume that PC/SC
sends the correct card removal event
- The list of readers to be monitored is adjusted once a reader (dis)appears
- On macOS, without PnP notification, we always check for new/removed
readers with SCardListReaders

fixes #1874

##### Checklist

- [x] macOS (Catalina)
- [x] Linux (Ubuntu 19.10)
- [x] Windows

### DIFF
--- a/src/libopensc/errors.c
+++ b/src/libopensc/errors.c
@@ -45,8 +45,8 @@ const char *sc_strerror(int error)
 		"Message too long (keypad)",
 		"Timeout while waiting for event from card reader",
 		"Unresponsive card (correctly inserted?)",
-		"Reader detached (hotplug device?)",
-		"Reader reattached (hotplug device?)",
+		"Reader detached",
+		"Reader reattached",
 		"Reader in use by another application"
 	};
 	const int rdr_base = -SC_ERROR_READER;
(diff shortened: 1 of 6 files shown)


frankmorgner:
macOS is working now.

Unfortunately, Windows has a high load after disconnecting a reader and I don't currently have the time to debug this...

frankmorgner:
I finally had time to test this across all platforms. It turns out, that there are more differences than I have initially thought. It also seems that sc_wait_for_event hasn't been tested anywhere else than in PKCS#11... Anyway, it all now looks good!

PR #1931, opened 2020-02-04
asn1: prevent null-dereference

tobhe:
`tagp == NULL` and `tag != SC_ASN1_TAG_EOC` are a valid result of `sc_asn1_read_tag` (see the "end of data reached" comment in `sc_asn1_read_tag`).

In this case error handling is skipped and `tagp` is used in `hlen = tagp - p`.

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -374,7 +374,7 @@ static void print_tags_recursive(const u8 * buf0, const u8 * buf,
 		size_t len;
 
 		r = sc_asn1_read_tag(&tagp, bytesleft, &cla, &tag, &len);
-		if (r != SC_SUCCESS || (tagp == NULL && tag != SC_ASN1_TAG_EOC)) {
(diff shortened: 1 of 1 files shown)


frankmorgner:
I have to admit, the code is not easily readable, but there will be no NULL dereference; the function will return, because `cla` is left at `0` and `tag` is `0` (`SC_ASN1_TAG_EOC`):
https://github.com/OpenSC/OpenSC/blob/34dad7f543f006ad269ce1f935a9e8d3e3a83db7/src/libopensc/asn1.c#L382-L385
Although `hlen` underflows, this doesn't have any consequence...

mfriedl:
I would also consider the other two calls to sc_asn1_read_tag()...

mfriedl:
Oh, sorry, I was reading an older revison :/

tobhe:
> I have to admit, the code is not easily readable, but there will be no NULL dereference; the function will return, because `cla` is left at `0` and `tag` is `0` (`SC_ASN1_TAG_EOC`):

You're right, my mistake! Thanks for the fast response

PR #1933, opened 2020-02-05
Allow disabling -Werror when strict compile mode is enabled

julianaito:
Hi,

In some cases, strict compile mode is wanted, but not `-Werror`. This PR allows that, without changing the default CFLAGS behaviour.

As a side effect, this avoids extra patching by downstream to make it build with funny compilers ;)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1072,10 +1072,10 @@ AS_IF([test "${enable_shared}" = "yes"], [AC_DEFINE([ENABLE_SHARED], [1], [Enabl
 
 if test "${enable_pedantic}" = "yes"; then
 	enable_strict="yes";
-	CFLAGS="${CFLAGS} -pedantic"
+	CFLAGS="-pedantic ${CFLAGS}"
 fi
 if test "${enable_strict}" = "yes"; then
-	CFLAGS="${CFLAGS} -Wall -Wextra -Wno-unused-parameter -Werror"
+	CFLAGS="-Wall -Wextra -Wno-unused-parameter -Werror ${CFLAGS}"
 fi
 
 AC_CONFIG_FILES([


frankmorgner:
I'd rather not like to add yet more configuration options. You can do all this fine tuning by passing flags manually. We use strict in order to detect problems early. 

What kind of compiler problem do you have exactly?

julianaito:
> I'd rather not like to add yet more configuration options. You can do all this fine tuning by passing flags manually. We use strict in order to detect problems early.

I can't override `-Werror` with `-Wno-error` because user supplied flags are put before the hardcoded flags, if you're fine with the idea, i can try to swap positions, and will change my PR in that sense.

Sure i can disable strict mode and put the other strict CFLAGS in our port Makefile, but it's prone to be unsynced with time.

> What kind of compiler problem do you have exactly?

OpenBSD is still using old GCC versions on old architectures, but in that case, it's because these GCCs make no compromise on `-Werror`. It's failing because we've the latest version of glib2, and some parts of the code uses deprecated fonctions, see http://build-failures.rhaalovely.net/sparc64/2020-02-01/security/opensc.log

frankmorgner:
So all eyes should actually go to glib2 then?!

Anyway, putting user supplied flags at top priority in OpenSC is a good idea. If you could change this in your PR I can accept that.

julianaito:
> So all eyes should actually go to glib2 then?!

Yes. It looks like they're reworking their code but still releasing new versions meanwhile.

> Anyway, putting user supplied flags at top priority in OpenSC is a good idea. If you could change this in your PR I can accept that.

I've committed that in my fork because it makes me easier to test changes with our actual port that way. I'll tell you when i'm done with testing

julianaito:
That works for me :)

frankmorgner:
thanks

julianaito:
You're welcome!

PR #1938, opened 2020-02-09
iso7816: extend iso7816_process_fci()

marschap:
Hi,

this PR extends iso7816_process_fci().  In detail, it
* defines file type SC_FILE_TYPE_UNKNOWN
* explicitly sets file->type to SC_FILE_TYPE_UNKNOWN for unkown files
* stores full-length file type attributes via sc_file_set_type_attr(),
  not only the first byte
* parses # of records for record-oriented EFs,
  if given in the file type attributes
* parses record length for for EFs with fixed-size records,
  if given in the file type attributes

Note:
I am not sure, parsing the record length only for EFs with fixed-size records is the correct approach.
My interpretation of 7816-4 is slightly different, but doing it this way seems
to be in-line what's currently in opensc:
- there's a comment hinting at that interpretation
- otherwise variable size records fail to be read in opensc-explorer
So I leave it this way for now.

As this PR extends the features of the default driver, it helps researching unknown cards.
I intend to use the additional information in opensc-explorer.

Thanks in advance for merging
Peter

### DIFF
--- a/src/libopensc/iso7816.c
+++ b/src/libopensc/iso7816.c
@@ -387,13 +387,36 @@ iso7816_process_fci(struct sc_card *card, struct sc_file *file,
 							file->type = SC_FILE_TYPE_DF;
 							break;
 						default:
+							file->type = SC_FILE_TYPE_UNKNOWN;
 							type = "unknown";
 							break;
 					}
 					sc_log(ctx, "  type: %s", type);
 					sc_log(ctx, "  EF structure: %d", byte & 0x07);
 					sc_log(ctx, "  tag 0x82: 0x%02x", byte);
-					if (SC_SUCCESS != sc_file_set_type_attr(file, &byte, 1))
+
+					/* if possible, get additional information for non-DFs */
+					if (file->type != SC_FILE_TYPE_DF) {
+						/* record length for fixed size records */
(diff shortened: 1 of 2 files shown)


frankmorgner:
What would be your interpretation of handling fixed size records? I think your implementation pretty much does what's defined in 7816-4...

marschap:
@frankmorgner: 
My reading of the standard is that FCI tag 83 contains the _**maximum** record size_ in byte 3 resp. bytes 3 and 4.  I.e.
* For fixed-size records, I think opensc does the right thing.
* For variable sized-records, I think opensc should also take the value from FCI tag 83 byte 3 resp. bytes 3 and 4 as an upper limit of the record size possible for the respective card.

My PR does not set `record_length` for files with variable-sized records because
* of the comment next to the definition of `record_length` in `src/libnopensc/types.h` hints in the direction that `record_length` is only valid for files with fixed-sized records (at least in my reading)
* opensc-explorer failed when I left out the "fixed size condition" `byte & 0x02` (I did not investigate further)

So the PR above is - as you and @Jakuje agreed - an improvement.
But I think there is space for further improvements in additional PRs, once this one has been merged ;-)

Hope this helps & Sorry for being not clear enough in the original comment to the PR
Peter

PR #1947, opened 2020-02-14
Improved Handling of CK_SLOT_ID values returned to C_GetSlotList

### LINKED ISSUE #1935: Firefox:  Stuck on invalidated PKCS#11 session handles
### Problem description
In NSS there is a problem with handling CKR_SESSION_HANDLE_INVALID as result of some PKCS#11 operation. When NSS opens a session and asks for the password, then CKR_SESSION_HANDLE_INVALID as result of C_Login correctly sets the internal state to CK_INVALID_SESSION. However, the opened (and invalidated) session is still used for further operation with the PKCS#11 module. This results in (a potentially infinite) number of password prompts although a token has long been removed.

Steps to reproduce:

1.   Configure PKCS#11 module in Firefox
2.   Insert token (it's shown in "Security Devices" as "not logged in")
3.   Go to a Website that requests a client certificate and wait for the password prompt for the token
4.    Remove the token
5.    Enter a password or abort it to close the password prompt
6.    Go to a Website that requests a client certificate and wait for the password prompt for the token
7.    Go to step 5. and repeat.

This bug reproduces with Firefox 72, OpenSC and a Yubikey, for example. After Step 4, the slot for the token still exists and the token is correctly shown as "not present" in the list of "security devices". However, the open (but invalid) session is never closed and PK11_Authenticate constantly tries to use the session to verify a password.

Note, that in Step 4, the PKCS#11 module is forced to invalidate all existing PKCS#11 sessions to a slot:

>(quote PKCS#11 v2.30 "Session events":)
>When the device is removed, all sessions of all applications are automatically logged out. Furthermore, all sessions any applications have with the device are closed (this latter behavior was not present in Version 1.0 of Cryptoki)"an application cannot have a session with a token that is not present. Realistically, Cryptoki may not be constantly monitoring whether or not the token is present, and so the token's absence could conceivably not be noticed until a Cryptoki function is executed. If the token is re-inserted into the slot before that, Cryptoki might never know that it was missing.
>(quote end)

Hence, all further operation with existing sessions to a removed token result in CKR_SESSION_HANDLE_INVALID

### Actual results:

Depending on the number of open TLS session, the user cannot get rid of constantly popping password prompts and is forced to close Firefox.

### Expected results:

If the PKCS#11 module returns CKR_SESSION_HANDLE_INVALID, NSS needs to close the session and re-open a new one.

### LINKED ISSUE #1945: PKCS11 changes CK_SLOT_IDs between calls to C_GetSlotList when devices are added or removed
### Problem Description
While addressing issue https://github.com/OpenSC/OpenSC/issues/1935 it appears  OpenSC whill changed the CK_SLOT_IDs between calls to C_GetSlotList when devices are inserted or removed. 

One would except the CK_SLOT_ID for a slot that has not changed to remain the same.

`pkcs11-tool --test-hotplug -v` with different tokens and readers can be used to show what happens.

dengert:
Partially Fixes #1945  
Partially Fixes #1935 

Firefox 72.0.2 still does not handle deleting of slots. See https://bugzilla.mozilla.org/show_bug.cgi?id=1613632

Previously the CK_SLOT_IDs where derived from the place on the
virtual_slot list. This means that if a reader was removed, the position
on the list could change. An application such as Firefox is also keeping
track of data based on the previous slot list. So the application
could get confused.

PKCS11 v2.30 code in in OpenSC to keep deleted slots on the list until the
application has see the slot once since itmiht have been deleted. But the
slot also looked like an available hotplug which might g reused before
the application ha seen it. This was fixed so it could not be reused until
the application had a chance to see it.

The CK_SLOT_ID now starts at 0, and increased by 1 each time a new slot
is created which occues when a reader is inserted. There is an upper limit
of 2^32 calls to create_slot. In practics this will never be reached but
the code has a "TODO" to add code to reuse slot IDs.

Pkcs11-tool.c has improved wording for listing slots.
 On branch PKCS11-SLOTS
 Changes to be committed:
	modified:   pkcs11-global.c
	modified:   sc-pkcs11.h
	modified:   slot.c

##### Checklist

- [X] PKCS#11 module is tested using FireFox
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -50,6 +50,8 @@ sc_context_t *context = NULL;
 struct sc_pkcs11_config sc_pkcs11_conf;
 list_t sessions;
 list_t virtual_slots;
+CK_SLOT_ID next_slot_id = 0;
+int slot_id_wrapped = 0;
 #if !defined(_WIN32)
 pid_t initialized_pid = (pid_t)-1;
 #endif
@@ -486,21 +488,6 @@ CK_RV C_GetSlotList(CK_BBOOL       tokenPresent,  /* only slots with token prese
 			slot->flags |= SC_PKCS11_SLOT_FLAG_SEEN;
 		}
 		prev_reader = slot->reader;
-	}
-
-	/* Slot list can only change in v2.20 */
-	if (pSlotList == NULL_PTR) {
-		/* slot->id is derived from its location in the list virtual_slots.
-		 * When the slot list changes, so does slot->id, so we reindex the
-		 * slots here the same way it is done in `create_slot()`
-		 *
-		 * TODO use a persistent CK_SLOT_ID, e.g. by using something like
-		 * `slot->id = sc_crc32(slot, sizeof *slot);` (this example, however,
-		 * is currently not thread safe).  */
-		for (i=0; i<list_size(&virtual_slots); i++) {
-			slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, i);
-			slot->id = (CK_SLOT_ID) list_locate(&virtual_slots, slot);
-		}
 	}
 
 	if (pSlotList == NULL_PTR) {
--- a/src/pkcs11/sc-pkcs11.h
+++ b/src/pkcs11/sc-pkcs11.h
@@ -344,6 +344,8 @@ extern struct sc_pkcs11_config sc_pkcs11_conf;
 extern list_t sessions;
 extern list_t virtual_slots;
 extern list_t cards;
+extern CK_SLOT_ID next_slot_id;
+extern int slot_id_wrapped;
 
 /* Framework definitions */
 extern struct sc_pkcs11_framework_ops framework_pkcs15;
--- a/src/pkcs11/slot.c
+++ b/src/pkcs11/slot.c
@@ -44,6 +44,8 @@ static struct sc_pkcs11_slot * reader_get_slot(sc_reader_t *reader)
 	/* Locate a slot related to the reader */
 	for (i = 0; i<list_size(&virtual_slots); i++) {
 		sc_pkcs11_slot_t *slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, i);
+		if (!reader && slot->flags & SC_PKCS11_SLOT_FLAG_SEEN)
+			continue;
 		if (slot->reader == reader)
 			return slot;
 	}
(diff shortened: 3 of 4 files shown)


frankmorgner:
- As far as I can see, the slot id doesn't need to be unique nor is it required to have static slot ids. Every meaningful operation is done with the session handle, not with a slot id.
- preemptively changing OpenSC for working around Firefox's errors is not good. Not really solving the issue is even worse. We should wait for a fix (or even a response) on Firefox' side

dengert:
I am still looking at this. I was able to get Firefox to not go into an infinite loop. It took a second cancel of the password prompt to end firefox.  That was why I said partial fix. 

The slot id needs to be persistent. 

You can play around with current `pkcs11-tool --test-hotplug -v` with 2 or 3 devices and see the Slot indexs and slot_ids in hex. You can test both C_GetSlotList and if you type "x" it moves on to testing  WaitForSlotEvent.  

 I expect Firefox  will not do anything for a while. and only handle added devices.

dengert:
Using single card  and ` pkcs11-tool --test-hotplug -v`

OpenSC 0.17.0 (default in my Ubuntu system) and current master both show:
```
Testing card detection using C_GetSlotList()
Please press return to continue, x to exit: 
Available slots:
  manufacturer:  SCM Microsystems Inc.
  hardware ver:  2.4
  firmware ver:  0.0
  flags:         token present, removable device, hardware slot
  token label        : PIV Card Holder pin (PIV_II)
  token manufacturer : piv_II
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : b8c3080027dbba7c
  pin min/max        : 4/8
Please press return to continue, x to exit:
```

But with 0.17.0 if I remove reader with card still in reader and hit enter it shows:
```
Available slots:
Slot 0 (0x0): Virtual hotplug slot
  manufacturer:  OpenSC Project
  hardware ver:  0.17
  firmware ver:  0.0
  flags:         removable device, hardware slot
Please press return to continue, x to exit: 
```

But Current master if I remove reader with card and hot enter it shows:
```
Available slots:
Please press return to continue, x to exit: 
```
With 0.17.0 and firefox,  C_GetSlotList  would show one slot, but as a hot plug slot.
With 0.20.0 and firefox , C_GetSlotList would show no slots, but firefox would ignore that there were no slots but then try and use the slot_id  0.  `SC_PKCS11_SLOT_FLAG_SEEN` code may have changed this. 

Firefox may fail in different ways in both cases when the card is removed. One causes a loop, the other does not.

frankmorgner:
I didn't find a reason why a virtual slot would make sense for itself (hotplugging is implemented by adding slots at runtime). We introduced a possible shrinking of the slot list in 7fb72ccf7bf3ca2ff3979b7ffbb690eed41ddb5f, which also eliminates the virtual slot as placeholder for slots that the application has seen, but which aren't valid anymore (as seen in 0.17.0). My reading is that this is perfectly valid in terms of PKCS#11 and for users/applications it makes sense to remove this clutter.

If we don't find a proper workaround for all the related problems in Firefox, a partial fix doesn't make sense. Mozilla will have to change its code anyway, so why not implement the proper fix**es** in Firefox?

dengert:
Yes I would like to hear from them.

dengert:
OpenSSH  mail list has related issue. from 2/21/2020, 
Re-adding PKCS#11 key in ssh-agent produces "agent refused operation" error.

dengert:
See #1956 for improved debugging of  virtual_slots.  
This PR #1947  does not resolve many of the issuer we have with with multiple tokens being inserted and removed while long running applications like Firefox are running.

dengert:
Many of the comments in this PR may be helpful, but the PR itself is not the complete answer.

PR #1956, opened 2020-02-22
Improved Debugging of PKCS11 virtual_slots

dengert:
Improved Debugging of PKCS11 virtual_slots

The handling of these virtual slots with different applications has proved to be elusive. Hopefully the improved debugging will help in understanding how the current code works or does not work.

This PR consist of 2 commits. 

The first commit:
Debug PKCS11 virtual slots
    
Macro DEBUG_VSS and routine _debug_virtual_slots were added.
    
DEBUG_VSS(slot, "printf like format string" [,args...]) will print the virtual_slots
to the opensc-debug.log showing were it was called from.
If slot is not NULL and found in the table it will be highlighted
with an "*".
    
In gdb: `call _debug_virtual_slots(slot)` can be used along with
another window to tail opensc-debug.log anywhere to get the same effect.

The second commit:

Add calls to DEBUG_VSS  

These calls are placed at many places in the PKCS11 code to help show have the virtual_slots are changed.

Several issues are related: #1907, #1935 , #1945 and #1947. 

An example changed that improves FireFox, but not completely
[sample-slot-diff.txt](https://github.com/OpenSC/OpenSC/files/4239737/sample-slot-diff.txt)

Running without this sample shows that `SC_PKCS11_SLOT_FLAG_SEEN` appears to have no effect.  Slots are always deleted. No slot is converted to a "HotPlug" slot.

With this sample a removed reader results in a ""HotPlug" show showing up, but this is not enough if one tries to use more then one token. 

Firefox only expects slots to to increase between calls to C_GetSlotList. It can not handle the situation where all the slots are deleted. For example when a token with built in reader such as a Yubikey is removed.  This is different from a card being removed but the reader is still present.

#1947 is being withdrawn, as it is not complete. 

##### Checklist

- [ X] PKCS#11 module is tested with `Firefox` and `pkcs11-tool --test-hotplug -v`

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -449,6 +449,7 @@ CK_RV C_GetSlotList(CK_BBOOL       tokenPresent,  /* only slots with token prese
 
 	sc_log(context, "C_GetSlotList(token=%d, %s)", tokenPresent,
 			pSlotList==NULL_PTR? "plug-n-play":"refresh");
+	DEBUG_VSS(NULL, "C_GetSlotList before ctx_detect_detect");
 
 	/* Slot list can only change in v2.20 */
 	if (pSlotList == NULL_PTR) {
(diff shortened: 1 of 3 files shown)


frankmorgner:
merged with 8d519ed3c4a8c4cedd9ea7e9c11a6348d447d452 at a higher debug level 2a6bf1e5771800f87691fd6f8532c3d057a0e090

PR #1961, opened 2020-02-26
Improved Handling of PKCS11 Slots

dengert:
OpenSC PKCS11 now retains slots even when the reader is removed.
It can do this because existing OpenSC reader handling in ctx.c,
reader-pcsc.c and PC/SC allow OpenSC to do this.

This simplifies the code, and allow a reader to be reinserted
and use the existing slot. This matching is actually done
in reader-pcsc.c because PC/SC returns the unique ids based on
the OS reader names. This is then used as the manufacturerID

By not deleting slots the slot list can only increase which is a
restriction of Firefox. It does not fix all the Firefox issues, but
it does not go into a loop, when all the readers are removed.

The defaults in opensc.conf for max-virtual-readers and slots-per-card
allow for 4 different readers used during one session.

 On branch PKCS11-SLOTS-3
 Changes to be committed:
	modified:   sc-pkcs11.h
	modified:   slot.c

Fixes: #1945 
Fixes: #1935 

Replaces: #1947 

##### Checklist

- [X ] PKCS#11 module is tested Firefox and pkcs11-tool

### DIFF
--- a/src/pkcs11/sc-pkcs11.h
+++ b/src/pkcs11/sc-pkcs11.h
@@ -206,6 +206,11 @@ struct sc_pkcs11_card {
  * the application calls `C_GetSlotList` with `NULL`. This flag tracks the
  * visibility to the application */
 #define SC_PKCS11_SLOT_FLAG_SEEN 1
+/* reader-pcsc.c can reuse a removed reader, as the ctx->reader list contains
+ * readers which have been removed retain removed readers.
+ * Take advantage of this feature to allow for reinsertion of a reader*/
+#define SC_PKCS11_SLOT_FLAG_READER_REMOVED 2
+
 
 struct sc_pkcs11_slot {
 	CK_SLOT_ID id;			/* ID of the slot */
--- a/src/pkcs11/slot.c
+++ b/src/pkcs11/slot.c
@@ -133,18 +133,8 @@ CK_RV create_slot(sc_reader_t *reader)
 void empty_slot(struct sc_pkcs11_slot *slot)
 {
 	if (slot) {
-		if (slot->flags & SC_PKCS11_SLOT_FLAG_SEEN) {
-			/* Keep the slot visible to the application. The slot's state has
-			 * already been reset by `slot_token_removed()`, lists have been
-			 * emptied. We replace the reader with a virtual hotplug slot. */
-			slot->reader = NULL;
-			init_slot_info(&slot->slot_info, NULL);
-		} else {
-			list_destroy(&slot->objects);
-			list_destroy(&slot->logins);
-			list_delete(&virtual_slots, slot);
-			free(slot);
-		}
+		list_clear(&slot->objects);
+		list_clear(&slot->logins);
 	}
 }
 
@@ -383,21 +373,38 @@ fail:
 CK_RV
 card_detect_all(void)
 {
-	unsigned int i;
+	unsigned int i, j;
 
 	sc_log(context, "Detect all cards");
 	/* Detect cards in all initialized readers */
 	for (i=0; i< sc_ctx_get_reader_count(context); i++) {
 		sc_reader_t *reader = sc_ctx_get_reader(context, i);
+		int removed = 0; /* have we called card_removed for this reader */
+
 		if (reader->flags & SC_READER_REMOVED) {
 			struct sc_pkcs11_slot *slot;
-			card_removed(reader);
-			while ((slot = reader_get_slot(reader))) {
-				empty_slot(slot);
+			/* look at all slots to call card_removed amd empty_slot */
+			for (j = 0; j<list_size(&virtual_slots); j++) {
+				slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, j);
+				if (slot->reader == reader) {
+					if (!removed) {
+						card_removed(reader);
+						removed = 1;  /* only need to call once for this reader */
+					}
+					if (slot->flags & SC_PKCS11_SLOT_FLAG_READER_REMOVED) {
+						empty_slot(slot);
+						slot->flags |= SC_PKCS11_SLOT_FLAG_READER_REMOVED;
+					}
+				}
 			}
-			_sc_delete_reader(context, reader);
-			i--;
 		} else {
+			struct sc_pkcs11_slot *slot;
+			for (j = 0; j<list_size(&virtual_slots); j++) {
+				slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, j);
+				if (slot->reader == reader)
+					slot->flags &= ~SC_PKCS11_SLOT_FLAG_READER_REMOVED;
+			}
+
 			if (!reader_get_slot(reader))
 				initialize_reader(reader);
 			else


dengert:
I think my PR fixes #1935  It does not get stuck, it may take 2 hitting cancel twice.

frankmorgner:
> I think my PR fixes #1935 It does not get stuck, it may take 2 hitting cancel twice.

Hmm, I've verified my patch, but maybe you're right with your change. I'll keep that in mind. But still we need to wait for Mozilla's response and keep the workaround NSS specific

dengert:
@frankmorgner  In you response to https://github.com/OpenSC/OpenSC/wiki/Using-OpenSC (which I saw via e-mail, but not in bugzilla yet),  you said:"but Firefox still shows a non-existent token to the user".

This PR  #1961 fixes that problem, as simulates the removal of the card/token but not the slot/reader.  It shows the slot has not having a card present, and forces NSS to treat the slot as empty. 

Thus if you examine the "Security Devices" you will see all the previous seen readers, including 
reader for cards or  readers for physical tokens with built in readers like Yubiky or SmartCard-HSM. 
The reader will appear to be empty i.e. no PKCS11 token.

So  physical tokens with built in readers like Yubiky or SmartCard-HSM when first seen  appears as both a slot with a PKCS11 token. But if removed, the slot is still there by no PKCS#11 token. 

It also does not keep the slots "static". By not removing a slot, but retaining relationship of the slot to the same reader (PC/SC reports this and PKCS11 and it is used for Description or ManuacftureID"). If the reader is plugged back in, it will be associated with the reader which OpenSC maintains in ctx.c and thus with an existing PKCS11 slot.

If PC/SC reports a different name for the reader, it will appear as a new reader and get a new slot. 

So from NSS view point, the slot list can only increase and removal of a card or physical tokens with built in readers like Yubiky or SmartCard-HSM appear as removal of a card, which NSS can handle with current code.

I think OpenSC made a mistake in 0.15.0 when it added code in PKCS11 slot.c to delete readers from the OpenSC list of readers and also removed the slots.  This PR reverses the behavior by not deleting the reader or slot, but does delete the PKCS11 token and login state.


**So Please** have a second look at this PR as it solves many of the issues we are having.

dengert:
Correction: bug report was:    https://bugzilla.mozilla.org/show_bug.cgi?id=1613632

frankmorgner:
Hmm, I am using your PR,
1. insert yubikey
2. remove yubikey
3. goto website with client cert auth
4. get prompted for PIN
5. goto 3. and repeat

... I am always getting prompted for the PIN although the status of the slot shows as "not present".

On the positive side, the slot gets reused once I'm inserting the Yubikey again. However, I think this only works because reader-pcsc.c reuses sc_reader_t objects, which is not true for the other I/O interfaces.

dengert:
Thanks for testing this.  When prompted for the PIN again, a few cancel of the PIN should get out of the loop and recover some what.  But see "**BUT**"  below...

NSS  still does not handle a removed card very well.   But it was worse if slot was removed. So the PR was designed to treat a removed reader and card i.e. Yubico, by causing CKR_TOKEN_NOT_PRESENT to be returned. 

I get the same response if removing a Yubico as removing a real card  from a reader without removing the reader.  

**BUT** I do not have a real  web site to test against. I have been using OpenSSL s_server. It  works  but if the connection fails it does not restart. It looks like the second  cancel to the PIN prompt followed a different code path because the server did not respond. This then caused the second step (3) to got out of the loop. 

If I restart the s_server aftert every failure, I get in to the loop just like you did.

So as you also see the main effect of the PR is to allow you to reinsert the Yubico and get out of the loop. 
You said: "However, I think this only works because reader-pcsc.c reuses sc_reader_t objects, which is not true for the other I/O interfaces."

I agree with the first part. I don't agree with the second part. 

_sc_delete_reader is only called in reader_pcsc.c in pcsc_use_reader only if  pcsc_add_reader fails.
use_reader is only  called by minidriver, and  and supported by reader-cryptotokenkit.m
and reader-pcsc.c. See opensc.h:
 505         /* Used to pass in PC/SC handles to minidriver */
 506         int (*use_reader)(struct sc_context *ctx, void *pcsc_context_handle, void *pcsc_card_hand     le);

and in ctx.c in sc_release_context when done.
 






I have been using this script to do some testing with and without gdb:
```
#!/bin/bash
export NSS_DEBUG_PKCS11_MODULE="OPenSC"
export NSPR_LOG_MODULES=nss_mod_log:4
export NSPR_LOG_FILE=/tmp/nspr_log_file

/usr/bin/firefox
#/usr/bin/firefox -g -d gdb -a "-x /home/doug/gdb-firefox-ex"
```
The output in  NSPR_LOG_FILE is their version of pkcs11-spy, and very handy. 

The gdb-firefox-ex has:
```
 set logging file /home/doug/gdb.log
 set logging on

 set pagination off

# set breakpoint pending on
# break pkcs11-session.c:113
# commands
# backtrace 15
# cont
# end
#break PK11_CheckPassword
#break PK11_DoPassword
#break PK11_HandlePasswordCheck
#break SECMOD_UpdateSlotList
#break PK11_Authenticate
```

frankmorgner:
My previous point was that the user experience is still terrible with or without your PR. This must finally be only solved within NSS. But, of course, we have no choice but to accept your PR, to change the situation from "unusable" to "usable with some drawbacks".

But if we fully commit to retaining a slot once it's created, even if the reader was removed, we can simplify the code. [We don't need additional state tracking, the slot object is enough](https://github.com/OpenSC/OpenSC/pull/1970/commits/cf4f21236060c35e566de2712a6a54ea0e2eabe5). [And I changed reclaiming a slot wit a reattached reader based on the reader's and slot's description instead of using the reader object](https://github.com/OpenSC/OpenSC/pull/1970/commits/cc4d4ff0b069c5517cce1ccbd374b6a18d61e020).

Maybe you can have a look at https://github.com/OpenSC/OpenSC/pull/1970/, which adds those changes to your PR.

dengert:
PCSC appears to keep track of the readers by name. This is then copied to the OpenSC reader_t.
reader-pcsc.c:1334  `if ((reader->name = strdup(reader_name)) == NULL) {`

The name is then used for the slot description:
pkcs11/slot.c:56    `strcpy_bp(pInfo->slotDescription, reader->name, 64);`

So it is not clear what you gain with `reader_reclaim_slot()`. reader-pcsc.c is doing it for you, by reusing the existing reader_t based on the pcsc name. 

The original PR disconnects for the card by calling empty_slot:
```
                list_clear(&slot->objects);
		list_clear(&slot->logins);
```
but #1970 removes empty_slot. 
If a user has multiple tokens, with the same reader name, it may look like a reinsertion, but may be a different device with different certificates, keys and pins. It is not clear if this will cause confusion with the application or worse. That maybe a FireFox issue. Not clear how they handle this.

The call to empty_slot is the conservative way to handle what may or may not be a reinsertion by forcing PKCS11 to treat it as a new token which happens to be in a reader with the same name.
i.e. like a card removed. 

I can not test my version (or looked closely at your version) on MacOS. reader-pcsc.c appears to have extra code for MacOS.

I don't know if you looked at #1956. It is all debugging to show the status of the virtual slots. It is independent of the the actual PRs. I learned a lot from using it, as there are so many dependencies between the ordering of the virtual slots, the slot id, the slot_list and entries pointing at the same reader have to be in order.     

We should also up the default number of virtual slots, to something large like 256. With the current code the real number of different readers is really 4 with 4 virtual slots per reader. It is not clear why the number is so low. 

I have spent way to much time on this to try to accommodate NSS failings. So if it works for you its OK with me.

frankmorgner:
The reason for implementing this in the PKCS#11 level (`reader_reclaim_slot()`) is a separation of concerns, which makes the whole code maintainable. We should not assume anything else than what's given via the APIs to the lower level.

We don't need `empty_slot()` (anymore), because the logins/objects lists are already cleaned in `slot_token_removed()`, which is called by `card_removed()`, which is called by `card_detect_all()` when the reader was removed (`SC_READER_REMOVED`)

You need a slot for each PIN on the card. I think that no card has and should have more than 4 PINs. Actually, one could rewrite the virtual slots to a dynamically growing list, but the benefit is minimal and the refactoring work is quite big...

I think I know the reasoning behind the `virtual_slots`. #1956 blows up the log a bit too much. Maybe raising the required debug level to something like `SC_LOG_DEBUG_MATCH` or above could solve this.

Thanks for taking the time to look into this; in the end it is your PR which fixes the issue with NSS.

frankmorgner:
closed with https://github.com/OpenSC/OpenSC/pull/1970

PR #1969, opened 2020-03-02
More fixes for issues reported by oss-fuzz, coverity and lgtm

Jakuje:
I went through the backlog of the recent reports and tried to fix most of them. But the coverity UI is awful so I was not able to load half of these issues.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
nice! I already had a few of those fixed locally, but forgot to publish them :see_no_evil:

PR #1970, opened 2020-03-03
Improved Handling of PKCS11 Slots (2)

frankmorgner:
This PR extends https://github.com/OpenSC/OpenSC/pull/1961 to

- simplify the PKCS#11 state tracking
- remove duplicated initialization code
- identifies a slot's reader based on the reader description instead of the reader's handle

Fixes: #1945
Fixes: #1935

Replaces: #1947

### DIFF
--- a/src/libopensc/pkcs15-gemsafeGPK.c
+++ b/src/libopensc/pkcs15-gemsafeGPK.c
@@ -360,14 +360,14 @@ static int sc_pkcs15emu_gemsafeGPK_init(sc_pkcs15_card_t *p15card)
 			}
 
 			if ( gsdata[idx1] == 0x30 &&
-				is_seq(gsdata + idx1, &seq_size1, &seq_len1) &&
-			 	is_seq(gsdata + idx1 + seq_size1, &seq_size2, &seq_len2) &&
-			    gsdata[idx1 + seq_size1 + seq_size2 + 0] == 0xa0 &&
-				gsdata[idx1 + seq_size1 + seq_size2 + 1] == 0x03 &&
-				gsdata[idx1 + seq_size1 + seq_size2 + 2] == 0x02 &&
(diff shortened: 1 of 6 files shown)


Jakuje:
This changed the slot IDs and now I am getting slots 0, 1 and 2 instaed of 0, 4 and 8 (with the virtual slot ids reserved). Not sure if this is good or bad, but this is a change that I do not see described here.

I can think of a case that we would have reader 1 with PIV card and reader 2 with other card, giving slot 0 and 1 respectively. If we swap the PIV card in first slot for OpenPGP card, which creates 2 virtual slots, I am not sure what will happen. Previously, the virtual slot ids were allocated, now it is not clear to me if it will use non-consecutive slot ids, it will orverlap or it will do something els.

This broke the CAC/Coolkey CI, which was expecting these particular slot IDs, so I would like to know whether this was intentional and I should update the configuration or it is a bug.

dengert:
#1961 kept the sc_reader_t and sc_pkcs11_slot_t for all the virtual slots and slot IDs even if physical reader was removed.   If the same reader is reinserted (and the OS returns the same name to pcsc), the same sc_reader_t,  sc_pkcs11_slot_t  and slot_ids will be used. If not a new sc_reader_t,  sc_pkcs11_slot_t for the virtual readers and slot IDs will be allocated.

frankmorgner:
I've [restored the original slot ids](https://github.com/OpenSC/OpenSC/commit/7e7458edce303ac3e8f421e52dacc350c908274c).

(However, an application should not rely on any hard coded slot ids...)

Jakuje:
I am not saying the applications should rely on hard-coded slot ids. This is test system running nightly tests. But I was not sure how the above mentioned use case can be handled without having virtual slots the way how they worked before so I asked. As well as I did not see this mentioned in any commit message.

PR #1977, opened 2020-03-08
opensc-explorer: add function get_record

marschap:
While analyzing an unknown card, I needed to get individual records from it for further investigation on my computer.
As `opensc-explorer` did not support it, I added this function.

Please merge into master.

##### Checklist

### DIFF
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -217,16 +217,21 @@ static struct command	cmds[] = {
 };
 
 
-static char *path_to_filename(const sc_path_t *path, const char sep)
+static char *path_to_filename(const sc_path_t *path, const char sep, size_t rec)
 {
-	static char buf[2*SC_MAX_PATH_STRING_SIZE];
+	static char buf[2*SC_MAX_PATH_STRING_SIZE+10];
 	size_t i, j;
 
+	/* build file name from path elements */
 	for (i = 0, j = 0; path != NULL && i < path->len; i++) {
 		if (sep != '\0' && i > 0 && (i & 1) == 0)
 			j += sprintf(buf+j, "%c", sep);
 		j += sprintf(buf+j, "%02X", path->value[i]);
 	}
+	/* single record: append record-number */
+	if (rec > 0)
+		j += sprintf(buf+j, "%c%"SC_FORMAT_LEN_SIZE_T"u",
+				(sep == '-') ? '_' : '-', rec);
 	buf[j] = '\0';
 
 	return buf;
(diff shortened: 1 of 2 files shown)


src/tools/opensc-explorer.c
```diff
@@ -217,16 +221,21 @@ static struct command	cmds[] = {
...
 	size_t i, j;
 
+	/* build file name from path elements */
 	for (i = 0, j = 0; path != NULL && i < path->len; i++) {
 		if (sep != '\0' && i > 0 && (i & 1) == 0)
 			j += sprintf(buf+j, "%c", sep);
 		j += sprintf(buf+j, "%02X", path->value[i]);
 	}
+	/* single record: append record-number */
+	if (rec > 0)
```
frankmorgner: rec needs to be `int` then

src/tools/opensc-explorer.c
```diff
@@ -1469,6 +1478,92 @@ static int do_get(int argc, char **argv)
...
+	if (argc < 2 || argc > 3)
+		return usage(do_get);
+	if (arg_to_path(argv[0], &path, 0) != 0)
+		return usage(do_get);
+	rec = strtoul(argv[1], NULL, 10);
+
+	filename = (argc == 3) ? argv[2] : path_to_filename(&path, '_', rec);
+	outf = (strcmp(filename, "-") == 0)
+		? stdout
+		: fopen(filename, "wb");
```
frankmorgner: `setmode(fileno(stdout), O_BINARY)`?

src/tools/opensc-explorer.c
```diff
@@ -1469,6 +1478,92 @@ static int do_get(int argc, char **argv)
...
+	if (outf == NULL) {
+		perror(filename);
+		goto err;
+	}
+
+	r = sc_lock(card);
+	if (r == SC_SUCCESS)
+		r = sc_select_file(card, &path, &file);
+	sc_unlock(card);
```
frankmorgner: sc_unlock behind sc_read_record to bind the read file to the selected one?

marschap: @frankmorgner  I am sorry, I do not understand.  This piece of code was copied verbatim from `do_get()`, and the pattern highlighted in these four lines is used all across `opensc-explorer.c`.

src/tools/opensc-explorer.c
marschap: @frankmorgner Thanks for the hint. I will add it also to the other functions writing potentially binary data to stdout.

src/tools/opensc-explorer.c
frankmorgner: Technically there is no problem at all. However, if you want to avoid being deselected between the SELECT and the GET RECORD, then you may want to lock before SELECT and unlock when the GET RECORD is done. This may also improve the pattern for the other operations, but I'll leave it to you whether or not to tie the both commands together.

PR #1978, opened 2020-03-09
pkcs11: don't try to allocate 0 byte with calloc

ueno:
This should suppress a warning from Electric Fence as reported in:
https://lists.libreswan.org/pipermail/swan-dev/2020-February/003626.html.

While it is is safe to call calloc() with 0 size on glibc systems,
POSIX says the behavior is implementation-defined.

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -458,7 +458,8 @@ CK_RV C_GetSlotList(CK_BBOOL       tokenPresent,  /* only slots with token prese
 
 	DEBUG_VSS(NULL, "C_GetSlotList after card_detect_all");
 
-	found = calloc(list_size(&virtual_slots), sizeof(CK_SLOT_ID));
+	/* alloc 1 more, to handle case of no virtual_slots */
+	found = calloc(list_size(&virtual_slots) + 1, sizeof(CK_SLOT_ID));
 
 	if (found == NULL) {
 		rv = CKR_HOST_MEMORY;


dengert:
But line 433 has `CK_SLOT_ID_PTR found = NULL;` 
so if you do not allocate found, line 464-466 will treat is as a `CKR_HOST_MEMORY` error.

ueno:
> But line 433 has `CK_SLOT_ID_PTR found = NULL;`
> so if you do not allocate found, line 464-466 will treat is as a `CKR_HOST_MEMORY` error.

There seems to be a confusion; the problem is that as soon as calloc() is called with 0, EF aborts the process; see the actual error linked from the commit message:
https://swantest.libreswan.fi/gdb-stdout.txt

dengert:
I was making the point, that your mod would introduce a different problem. When there are no virtual slots, it would endup up returning  CKR_HOST_MEMORY. 

The case of zero slots is common if there are no smart card readers when C_GetSlotList is first called. 

`found` is only used within `C_GetSlotList`.  I would propose you do something like  the following:
```
diff --git a/src/pkcs11/pkcs11-global.c b/src/pkcs11/pkcs11-global.c
index cdfbca41..b70e80f4 100644
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -491,7 +491,8 @@ CK_RV C_GetSlotList(CK_BBOOL       tokenPresent,  /* only slots with token prese
 
        DEBUG_VSS(NULL, "C_GetSlotList after card_detect_all");
 
-       found = calloc(list_size(&virtual_slots), sizeof(CK_SLOT_ID));
+       /* alloc 1 more, to handle case of no virtual_slots */
+       found = calloc(list_size(&virtual_slots) + 1 , sizeof(CK_SLOT_ID));
 
        if (found == NULL) {
                rv = CKR_HOST_MEMORY;
```

`found` contains the maximum number of slot_ids that **could** be returned. 
Later in code  `numMatches`  will be the number of slot_ids to return  via  `memcpy(pSlotList, found, numMatches * sizeof(CK_SLOT_ID));` and  `*pulCount = numMatches;`

ueno:
> When there are no virtual slots, it would endup up returning CKR_HOST_MEMORY.

This is not the case (I believe), because that code path is guarded with `if (!list_empty(&virtual_slots)) { ... }`.

> Later in code numMatches will be the number of slot_ids to return via memcpy(pSlotList, found, numMatches * sizeof(CK_SLOT_ID)); and *pulCount = numMatches;

This is actually a problem: if `found` is NULL, that memcpy call will cause an undefined behavior. I will go with your suggestion then; thank you for that!

PR #1993, opened 2020-03-27
From the PKCS#11 specification chapter C_SetPIN:

larssilven:
C_SetPIN modifies the PIN of the user that is currently logged in, or
the CKU_USER PIN if the session is not logged in. ....

This was not true for "if the session is not logged in" before this fix.

lars@brandt:~$ opensc-tool -n
Using reader with a card: ACS APG8201-B2 00 00
MyEID 4.0.17

##### Checklist

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1668,22 +1668,6 @@ pkcs15_login(struct sc_pkcs11_slot *slot, CK_USER_TYPE userType,
 
 	if (!p11card)
 		return CKR_TOKEN_NOT_RECOGNIZED;
-	if (p11card->card->reader->capabilities & SC_READER_CAP_PIN_PAD
-			|| (p15card->card->caps & SC_CARD_CAP_PROTECTED_AUTHENTICATION_PATH)) {
-		/* pPin should be NULL in case of a pin pad reader, but
-		 * some apps (e.g. older Netscapes) don't know about it.
-		 * So we don't require that pPin == NULL, but set it to
-		 * NULL ourselves. This way, you can supply an empty (if
-		 * possible) or fake PIN if an application asks a PIN).
-		 */
-		/* But we want to be able to specify a PIN on the command
-		 * line (e.g. for the test scripts). So we don't do anything
-		 * here - this gives the user the choice of entering
-		 * an empty pin (which makes us use the pin pad) or
-		 * a valid pin (which is processed normally). --okir */
-		if (ulPinLen == 0)
-			pPin = NULL;
-	}
 
 	/* By default, we make the reader resource manager keep other
 	 * processes from accessing the card while we're logged in.
@@ -1843,18 +1827,7 @@ pkcs15_change_pin(struct sc_pkcs11_slot *slot,
 		return CKR_USER_PIN_NOT_INITIALIZED;
 
 	sc_log(context, "Change '%.*s' (ref:%i,type:%i)", (int) sizeof pin_obj->label, pin_obj->label, auth_info->attrs.pin.reference, login_user);
-	if ((p11card->card->reader->capabilities & SC_READER_CAP_PIN_PAD)
-			|| (p15card->card->caps & SC_CARD_CAP_PROTECTED_AUTHENTICATION_PATH)) {
-		/* pPin should be NULL in case of a pin pad reader, but
-		 * some apps (e.g. older Netscapes) don't know about it.
-		 * So we don't require that pPin == NULL, but set it to
-		 * NULL ourselves. This way, you can supply an empty (if
-		 * possible) or fake PIN if an application asks a PIN).
-		 */
-		pOldPin = pNewPin = NULL;
-		ulOldLen = ulNewLen = 0;
-	}
-	else if (ulNewLen < auth_info->attrs.pin.min_length || ulNewLen > auth_info->attrs.pin.max_length)  {
+	if (pNewPin && (ulNewLen < auth_info->attrs.pin.min_length || ulNewLen > auth_info->attrs.pin.max_length)) {
 		return CKR_PIN_LEN_RANGE;
 	}
 
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1858,11 +1858,7 @@ pkcs15_change_pin(struct sc_pkcs11_slot *slot,
 		return CKR_PIN_LEN_RANGE;
 	}
 
-	if (login_user < 0) {
-		if (sc_pkcs11_conf.pin_unblock_style != SC_PKCS11_PIN_UNBLOCK_UNLOGGED_SETPIN) {
-			sc_log(context, "PIN unlock is not allowed in unlogged session");
-			return CKR_FUNCTION_NOT_SUPPORTED;
-		}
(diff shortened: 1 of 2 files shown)


frankmorgner:
I'm fine with the first commit, but the workaround removed in b00e716d17b1aa8bd650c566a8b0a523b02c58dc was added specifically to work around some apps' limitations... Is this a common case to have a PIN-pad-reader without wanting to use the PIN pad?

larssilven:
Don't know how common it is but there are situations when the application know a PIN but not the card owner. In my case it is a PIN that has been set when the card was initialized but before any keys and data and certificates has been written to it. I want the user to be forced to change PIN before the card has any sensitive data and the user should not be bothered with any initial PIN that the card was initialized with.
Anyway the pkcs11 specification says that PIN must only be fetched by other means than the parameter if the parameter is a null pointer. I think it is more important to follow the standard than to have a workaround for "older Netscapes". I also believe that other PKCS#11 implementations for cards are following the standard. It should be possible to substitute one implementation with another.
But if it is really needed there could be a parameter in opensc.conf that is enabling this exception from the standard.
I'm currently working with pkcs11-tool to support this as well so don't merge this second commit yet.
Maybe it is actually better to wait with the first commit as well. I will then just do one commit for everything.

dengert:
I don't believe you situation is common. Pin Pad readers are used to enforce a "policy" that the PIN will not be exposed to the host computer. We should try and support that policy by default. 

A "really good" pinpad reader will not accept a standard PIN command from the host  
to enforce the "policy".  So your idea may not work at all.   

A user or hacker can always violate the "policy" by swapping readers or hacking the code.  Nothing we can do about that in software. 

Depending on your card and application you are not constrained to using pkcs11. Without logging in you application can check if certificates are present or not using pkcs11. If you don't have a "really good" pin pad reader, your application can send verify command and change pin commands. 
Part of a script I have used with a card using 8 bype pins with FF padding and digits can use this:
```
# Change userpin using pin:
#       00 24 00 80 10 oldpin newpin
# Change global pin using pin (only works on some cards):
#       00 24 00 00 10 oldpin newpin
# Change puk using puk
#       00 24 00 81 10 oldpuk newpuk 

# Change userpin using puk
#   00 2C 00 80 10 oldpuk newpin
# Change global pin using puk (does not work on some cards)
#   00 2C 00 00 10 oldpuk newpin
#
opensc-tool -v  -s 00:24:00:80:10:$OLDPIN:$NEWPIN

#example old pin is 123456, NewPin is 20200330 
opensc-tool -v  -s 00:24:00:80:10:31:32:33:34:34:36:FF:FF:32:30:32:30:30:33:33:30
```
You can test if you have a "really good" pin pad reader with something like this too.

I don't think pkcs11-tool should be changed to support what you want. You can use pkcs11-tool.c as an and example for your own application.

larssilven:
Doug,
Did not read you comment until my last commit.
Some comments on you comment:
There will be no keyboard available on the device to which the PIN pad reader is connected to. So the user is forced to enter PIN on the reader.
The SW to be used will also use null instead of a PIN string when using the PKCS#11 API except for this initial setting of PIN. The new PIN will be null in the C_SetPIN call it is only the old PIN that will be a string (tested working).
Further on I can not see that the security will be better by not allowing PIN strings when when using a PIN pad reader. As you said yourself it is just to change reader or make the changes I have made yourself the the opensc pkcs11. 
The important thing is to instruct the user that he must not enter his PIN anywhere but on his own smart card reader. This is regardless of whatever PIN could be sent to the reader or not. In our case the user must also know that the reader must not be plugged in anywhere, just in devices supplied by us.
And also there might be other use cases where not all PINs are "owned" by the user. It could be that certain functions of the card should just be activated when connecting to a trusted server. The trusted server knows this PIN but not fake servers that want use or get information from your card.

Everything is working as I want it now. I have tested it. It will be fine for me to add some switch in opensc.conf that will enable the current none pkcs#11 compliant implementation. It could be enabled by default.
Cheers,
Lars

dengert:
You have a special device, my concern is making changes to OpenSC just for you special situation which may cause problems for everyone else.

larssilven:
Yes I got a special device. But there are a lot other special devices out there and for some of them it could be usable to select whether PIN should be supplied from the PIN pad of the reader or whether PIN should be supplied from the application.
I got another argument for using using both PIN from application and PIN from pad:
What if you got an application that is not aware of the feature that the reader could provide the PIN and is using the current implementation? Then the application will first prompt the user for the PIN (cli or gui) and then call the pkcs11 function with the PIN that will be ignored and the user will be prompted again on the pad for a PIN. In some cases it might be acceptable to use the PIN from the application instead to prevent the user of getting confused.
I could change my request to make this configurable in opensc.conf . Then we could also add an alternative that is failing with CKR_FUNCTION_NOT_SUPPORTED if a PIN string is used when a the reader has a PIN pad. This will then notify the user that he ought to have his application updated or use an ordinary reader.

frankmorgner:
There are some use cases for passing a PIN despite having a PIN pad readers and there are readers out there which have hard coded exceptions for certain cards when filtering VERIFY commands.

Note, however, that we're still using an [empty PIN for detecting whether or not to use the PIN pad in tokend](https://github.com/OpenSC/OpenSC.tokend/blob/e878b8ddef4dd8a41476bd7729dfc7c07e621e00/OpenSC/OpenSCToken.cpp#L160-L184
), because tokend doesn't handle external PIN pads at all...

I think @larssilven suggestion is en par with the specification. The issue brought up by the source code comment seemed to be solved by now (Firefox doesn't have any issues with PIN pad readers). So I think we're good to merge. What do you think, @dengert ?

saper:
> Is this a common case to have a PIN-pad-reader without wanting to use the PIN pad?

Two reasons:

1. I have a problem getting the D-Trust V3.1 card to work, because I cannot get OpenSC to select a proper PIN to choose. A pinpad reader displays only a "PIN" prompt at the display and I have to guess which one it really means. OpenSC CLI at least asks me for "Signature PIN", "Signature PUK","Card PIN" as they are named on the card so I can tell which one is really asked for.

2. PIN unlock does not seem to work for me when asking to use a pinpad, therefore I have to use the host entry.

dengert:
@frank The code looks, OK.

dengert:
@saper The "Signature PIN"  most likely comes from `pkcs15-openpgp.c`  and copied to a `sc_pkcs15_object_t`  `strlcpy(pin_obj.label, pin_cfg[i].label, sizeof(pin_obj.label));`

Look at `pkcs15-pin.c` for  `prompt = `  (10 of these).  You may be able to submit a PR that uses the `label` in the prompt.  I have not looked at PIN PAD code in a long time, and there may be restrictions on the size if the prompt shown on the screen.

saper:
Thank you @dengert - will check it. The pinpad of my cyberJack displays just generic "PIN" and there is certainly not enough space for such long strings.

frankmorgner:
@saper, pin pad strings are most likely hard coded in the reader. Typically, the application should show an additional prompt telling the user what type of PIN to enter on the pin pad...

frankmorgner:
@saper if you see an open question here, please open an issue

saper:
@frankmorgner sure, sorry for hijacking!

larssilven:
Thank you all for accepting these changes.
It will help me much not having to patch opensc.
Also I believe that the standard must be followed; then an application could use different devices with different pkcs#11 implementations.

PR #2004, opened 2020-04-14
Add OpenSCToken to dmg

frankmorgner:
This PR uses my developer id to create and sign a CTK token, which is included in OpenSC's dmg.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
thanks for having a look.

One more thing I'd like to tweak, is to make the xcode signing flags configurable. Right now, the developer ID is hard coded, for example, which isn't so nice for local use.

frankmorgner:
I'll also clean up the commits when it's ready.

PR #2006, opened 2020-04-16
Starcos 3.2 support

rumpelsepp:
This is a rebase of #1347 against the current master. I have tested it right now; the card works. Disclaimer: I am not very aware of the code base right now, I just did git magic. But maybe I can incorporate further feedback to eventually get this merged. I hope the CI does not blame me for breaking everything. **cross fingers**

cc: @jpecholt


frankmorgner:
Thanks for looking into this again. The good news is that your code seems to compile 👍

Did you also have time to look at the comments found in https://github.com/OpenSC/OpenSC/pull/1347? The main problem was that the code looks like to be implemented for a specific card *configuration* rather than a generic solution for starcos. Back then I've suggested to parse EF.KEYD, for example...

rumpelsepp:
Yeah and it also works for my card. That's good news I guess…

> Back then I've suggested to parse EF.KEYD, for example...

Hmm ok. I need to look for that manual and do some reading.

frankmorgner:

PR #2011, opened 2020-04-21
pkcs11-tool: Fix mismatched --help text

CyberShadow:
The addition of --usage-wrap did not add a corresponding help string, which caused all help text for the options below it to be shifted by one.

Just a small blemish I noticed :)

I used the --usage-wrap help text from the XML file.


frankmorgner:
thanks

PR #2026, opened 2020-05-07
Address recent coverity issues

Jakuje:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -750,7 +750,7 @@ __pkcs15_create_pubkey_object(struct pkcs15_fw_data *fw_data,
 		object->pub_data = p15_key;
 		if (p15_key && object->pub_info->modulus_length == 0 && p15_key->algorithm == SC_ALGORITHM_RSA)
 			object->pub_info->modulus_length = 8 * p15_key->u.rsa.modulus.len;
-	} else if (!(pubkey->emulated && (fw_data->p15_card->flags & SC_PKCS15_CARD_FLAG_EMULATED))) {
+	} else {
 		sc_pkcs15_free_pubkey(p15_key);
 	}
 	if (object && object->pub_data) {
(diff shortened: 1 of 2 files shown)


src/pkcs11/framework-pkcs15.c
```diff
@@ -750,7 +750,7 @@ __pkcs15_create_pubkey_object(struct pkcs15_fw_data *fw_data,
 		object->pub_data = p15_key;
 		if (p15_key && object->pub_info->modulus_length == 0 && p15_key->algorithm == SC_ALGORITHM_RSA)
 			object->pub_info->modulus_length = 8 * p15_key->u.rsa.modulus.len;
-	} else if (!(pubkey->emulated && (fw_data->p15_card->flags & SC_PKCS15_CARD_FLAG_EMULATED))) {
```
frankmorgner, line 753: I think the assumption used to be that this refers to a static object. If I checked correctly, this is only used in the PKCS#15 emulators of PIV, CAC and IDPrime so that your change should work, right?

Jakuje: The condition was same as is on line 733 until 35c5eb0. Now it is negated, which does not make sense as far as I read it. The `pubkey->emulated` was probably intended to contain static object, but it is not the case now. All PIV, CAC and IDPrime, coolkey now assign dynamically allocated objects if I read it correctly. More I think about it, I believe it was intended to transfer this memory to the object, but in that case, it should remove the pointer from the pubkey object, otherwise it causes only memory mess when freeing and is very prone to double frees.

This branch was actually originally introduced in c5e4012 to avoid memory leak but in very cumbersome way -- the pointers to this structure would be kept in both p15key and in new object, which is very prone to double frees.

The behavior with dup and explicit free on error and cleanup of the emulated pointer on free (in 75be610ea). Sounds to me cleaner.

Jakuje:
Rebased on top of current master (dropping one of the changes that doug already fixed in master).

Issue #2028, opened 2020-05-07
javacard simulation CI target broken


Jakuje:
### Problem Description

Recently, all builds using javacard simulation fail to run PivApplet. I started investigating what is going on and it turns out (after unrolling the raw log in travis), that the build of PivApplet fails since some changes in upstream repository
```
BUILD FAILED
/home/travis/build/Jakuje/OpenSC/PivApplet/build.xml:66: Compile failed; see the compiler error output for details.
```
Investigating further, it looks like also the build of `yubico-piv-tool` fails later:
```
ykcs11.c:1838:5: error: â€˜forâ€™ loop initial declarations are only allowed in C99 mode
     for (CK_ULONG j = 0; j < ulCount; j++) {
     ^
make[2]: *** [ykcs11.lo] Error 1
make[2]: Leaving directory `/home/travis/build/Jakuje/OpenSC/yubico-piv-tool/ykcs11'
make[1]: *** [install-recursive] Error 1
make[1]: Leaving directory `/home/travis/build/Jakuje/OpenSC/yubico-piv-tool/ykcs11'
make: *** [install-recursive] Error 1
travis_time:end:2ae5cfef:start=1588869529247751602,finish=1588869598104316046,duration=68856564444,event=before_script
```
But both of these failures are ignored and we see only the failure of the jcardsim to start the applet.

The javacard simulation runs on oldes Ubuntu (14.04) -- the rest was already migrated to bionic (18.04). I did not update last time, because the jcardsim was not able to build with anything newer. There are some commits in https://github.com/arekinath/jcardsim which allow me to build it on my Fedora, but it still fails on Ubuntu for unknown reasons.

I put together few commits in the way to address this issue, but can not continue fiddling with that right now so if somebody would like to continue, feel free to do so:

https://github.com/Jakuje/OpenSC/commits/travis-kill

For others, this is a note that the javacard tests are broken now and its not your code that broke it :)

Jakuje:
For the record, now I got up to the place where ecdsa keygen fails with the above branch:
```
++yubico-piv-tool -v 9999 -r 'Virtual PCD 00 00' -P 123456 -s 9a -a generate -A ECCP256
Connect reader 'Virtual PCD 00 00' matching 'Virtual PCD 00 00'.
00098606 APDU: 00 A4 04 00 05 A0 00 00 03 08 
[...]

PR #2033, opened 2020-05-12
card-piv.c error in using sc_asn1_put_tag

dengert:
In piv_general_mutual_authenticate sc_asn1_put_tag is not used correctly.

 On branch piv-sc_asn1_put_tag-error
 Changes to be committed:
	modified:   card-piv.c

This is only used by piv-tool when doing piv_general_mutual_authenticate.

Checklist

- [X] piv-tool to create EC key. 

Problem found while testing real PIV card for #2028 See https://github.com/OpenSC/OpenSC/issues/2028#issuecomment-627579440


frankmorgner:
thanks

PR #2035, opened 2020-05-18
tests: Fix card present check

torin-carey:
Check `SC_READER_CARD_PRESENT` flag rather than` == 1`.
Having no card present on the first loop and then inserting a card will
return `rc = CARD_PRESENT | CARD_CHANGED (= 3)`. SEGFAULT ensures when we mistake
the unset `opt_reader` as having a present card.

### DIFF
--- a/src/tests/sc-test.c
+++ b/src/tests/sc-test.c
@@ -83,18 +83,22 @@ int sc_test_init(int *argc, char *argv[])
 			if (rc < 0)
 				return rc;
 		} else {
-			for (i = rc = 0; rc != 1 && i < (int) sc_ctx_get_reader_count(ctx); i++)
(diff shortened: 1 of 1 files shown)


src/tests/sc-test.c
```diff
@@ -85,16 +85,19 @@ int sc_test_init(int *argc, char *argv[])
 		} else {
 			for (i = rc = 0; rc != 1 && i < (int) sc_ctx_get_reader_count(ctx); i++)
```
Jakuje, line 86: Should not the check be updated also here in `rc != 1`?

torin-carey: Amended, thank you

frankmorgner:
thanks

PR #2040, opened 2020-05-27
pkcs11: If user consent is ignored through configuration, do not pres…

### LINKED ISSUE #2039: PIN has to be entered twice with pkcs11 keys
### Problem Description

When using Yubikey tokens with `pkcs11:` key URIs, the PIN has to be input twice: once for the "_token PIN for PIV Card Holder pin_" and a second time for "_key PIN for SIGN key_".  If the PIN is passed in the URI, such as `pkcs11:pin-value=123456`, it is only used for the token PIN and the key PIN still has to be typed in manually.  For signing a large number of files, such as the Linux kernel modules, this prevents any automation of the process.

This might be a duplicate of https://github.com/OpenSC/OpenSC/issues/1269 although the suggestion of setting `pin_cache_ignore_user_consent=true;` in `/etc/opensc/opensc.conf` does not seem to make any difference in the behaviour.

### Proposed Resolution
Replay the PIN provided in the `pin-value` or provide an alternate way to specify the token PIN separately from the key PIN.

### Steps to reproduce
Extract the public key certificate from the yubikey and convert it to DER:
```
% yubico-piv-tool -s 9c -a read-certificate -o cert.pem
% openssl x509 -outform der -in cert.pem -out cert.crt
```

Try to sign a kernel module with it, which uses the pkcs11 engine and the DER public key certificate:
```
% kmodsign sha256  'pkcs11:'  cert.crt unsigned.ko signed.ko
Enter PKCS#11 token PIN for PIV Card Holder pin (PIV_II):
Enter PKCS#11 key PIN for SIGN key:
% kmodsign sha256  'pkcs11:pin-value=123456'  cert.crt unsigned.ko signed.ko
Enter PKCS#11 key PIN for SIGN key:
```

The same result happens when signing with OpenSSL:
```
% echo foo | openssl dgst -engine pkcs11 -keyform engine -sign 'pkcs11:' -out foo3.sig
engine "pkcs11" set.
Enter PKCS#11 token PIN for PIV Card Holder pin (PIV_II):
Enter PKCS#11 key PIN for SIGN key:
% echo foo | openssl dgst -engine pkcs11 -keyform engine -sign 'pkcs11:pin-value=123456' -out foo2.sig
engine "pkcs11" set.
Enter PKCS#11 key PIN for SIGN key:
```

However, `pkcs11-tool` is able to perform signatures without requiring the PIN twice if it is specified on the command line (otherwise it does ask for both PINs):
```
% pkcs11-tool --sign -m RSA-PKCS  -i <(echo foo)
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter User PIN: 
Using signature algorithm RSA-PKCS
Logging in to "PIV Card Holder pin (PIV_II)".
Please enter context specific PIN:
....
% pkcs11-tool --pin 123456 --sign -m RSA-PKCS  -i <(echo foo) > foo.sig
Using signature algorithm RSA-PKCS
```

### Logs

Log files attached:
* [kmodsign.txt](https://github.com/OpenSC/OpenSC/files/4663771/kmodsign.txt)
* [pkcs11-tool.txt](https://github.com/OpenSC/OpenSC/files/4663773/pkcs11-tool.txt)

Jakuje:
…ent ALWAYS_AUTHENTICATE attribute

Fixes #2039

This extends the function of `pin_cache_ignore_user_consent` to prevent presenting `CKA_ALWAYS_AUTHENTICATE` to applications, which understand this attribute and issue useless second PIN prompt (even though pin is already cached in OpenSC or not needed at all).

I believe current documentation is fine (man opensc.conf), but if you believe it needs to be updated, please suggest.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -3773,9 +3773,13 @@ pkcs15_prkey_get_attribute(struct sc_pkcs11_session *session,
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
 		*(CK_BBOOL*)attr->pValue = (prkey->prv_info->access_flags & SC_PKCS15_PRKEY_ACCESS_LOCAL) != 0;
 		break;
-    case CKA_ALWAYS_AUTHENTICATE:
+	case CKA_ALWAYS_AUTHENTICATE:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
-		*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent >= 1 ? CK_TRUE : CK_FALSE;
+		if (fw_data->p15_card->opts.pin_cache_ignore_user_consent) {
+			*(CK_BBOOL*)attr->pValue = CK_FALSE;
+		} else {
+			*(CK_BBOOL*)attr->pValue = prkey->prv_p15obj->user_consent >= 1 ? CK_TRUE : CK_FALSE;
+		}
 		break;
 	case CKA_PRIVATE:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
--- a/doc/files/opensc.conf.5.xml.in
+++ b/doc/files/opensc.conf.5.xml.in
@@ -1163,6 +1163,11 @@ app <replaceable>application</replaceable> {
 							<literal>CKA_ALWAYS_AUTHENTICATE</literal> may
 							need to set this to get signatures to work with
 							some cards (Default: <literal>false</literal>).
+					</para>
+					<para>
+							It is recommended to enable also PIN caching using
+							<literal>use_pin_caching</literal> option for OpenSC
+							to be able to provide PIN for the card when needed.
 					</para></listitem>
 				</varlistentry>
 				<varlistentry>
--- a/etc/opensc.conf.example.in
+++ b/etc/opensc.conf.example.in
@@ -903,6 +903,8 @@ app default {
 		# Older PKCS#11 applications not supporting CKA_ALWAYS_AUTHENTICATE
 		# may need to set this to get signatures to work with some cards.
 		# Default: false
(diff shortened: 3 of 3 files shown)


src/pkcs11/framework-pkcs15.c
```diff
@@ -3730,6 +3730,10 @@ pkcs15_prkey_get_attribute(struct sc_pkcs11_session *session,
 		break;
     case CKA_ALWAYS_AUTHENTICATE:
 		check_attribute_buffer(attr, sizeof(CK_BBOOL));
+		if (fw_data->p15_card->opts.pin_cache_ignore_user_consent) {
+			*(CK_BBOOL*)attr->pValue = CK_FALSE;
+			break;
```
frankmorgner, line 3735: putting the `break` here, doesn't seem so common. If you would just use an `else` branch, it looks more obvious, at least for me.

Jakuje: No problem in changing this detail. I would be more interested it if works for the reporter use case and whether you are conceptually fine with this.

frankmorgner: I'm fine with it. A card, that *always* forces a re-validation of the PIN will be rendered useless if PIN-caching is disabled, but if the user sets this option manually, then I hope she knows what she's doing.

Jakuje:
I tested this with yk, but I would like to see this tested also with PIV test cards, which should enforce reauthentication and therefore the cached pin should get used. But I do not have these at hand now.

Jakuje:
After reading through https://github.com/OpenSC/OpenSC/issues/2039#issuecomment-635949560 I believe this should happen only in combination with `use_pin_caching = yes`. I always forget about this option. What do you think?

Jakuje:
Thanks for suggestion. I added this note both to manual pages and example configuration file.

frankmorgner:
thank you

PR #2045, opened 2020-05-29
Fix C_Login() crashing when using uninitialized card

llogar:
`pkcs11-tool --init-pin ...` crashes if the card hasn't been initialized beforehand with `pkcs11-tool --init-token ...` (or by any other means) as p15card == NULL

Tested on IsoApplet javacard.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,7 +153,7 @@ before_script:
       javac -classpath jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar IsoApplet/src/net/pwendland/javacard/pki/isoapplet/*.java;
       echo "com.licel.jcardsim.card.applet.0.AID=F276A288BCFBA69D34F31001" > isoapplet_jcardsim.cfg;
       echo "com.licel.jcardsim.card.applet.0.Class=net.pwendland.javacard.pki.isoapplet.IsoApplet" >> isoapplet_jcardsim.cfg;
-      echo "com.licel.jcardsim.card.ATR=3B80800101" >> isoapplet_jcardsim.cfg;
+      echo "com.licel.jcardsim.card.ATR=3B8B8001805949736F4170706C6574BA" >> isoapplet_jcardsim.cfg;
       echo "com.licel.jcardsim.vsmartcard.host=localhost" >> isoapplet_jcardsim.cfg;
       echo "com.licel.jcardsim.vsmartcard.port=35963" >> isoapplet_jcardsim.cfg;
 
@@ -254,6 +254,8 @@ script:
       opensc-tool --card-driver default --send-apdu 80b800001a0cf276a288bcfba69d34f310010cf276a288bcfba69d34f3100100;
       opensc-tool -n;
       pkcs15-init --create-pkcs15 --so-pin 123456 --so-puk 0123456789abcdef;
+      pkcs15-tool --change-pin --pin 123456 --new-pin 654321;
+      pkcs15-tool --unblock-pin --puk 0123456789abcdef --new-pin 123456;
       pkcs15-init --generate-key rsa/2048     --id 1 --key-usage decrypt,sign --auth-id FF --pin 123456;
       pkcs15-init --generate-key rsa/2048     --id 2 --key-usage decrypt      --auth-id FF --pin 123456;
       pkcs15-init --generate-key ec/secp256r1 --id 3 --key-usage sign         --auth-id FF --pin 123456;
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1603,6 +1603,9 @@ pkcs15_login(struct sc_pkcs11_slot *slot, CK_USER_TYPE userType,
 	if (!fw_data)
 		return sc_to_cryptoki_error(SC_ERROR_INTERNAL, "C_Login");
 	p15card = fw_data->p15_card;
+	if (p15card == NULL) {
+		return CKR_FUNCTION_FAILED;
+	}
 
 	sc_log(context, "pkcs15-login: userType 0x%lX, PIN length %li", userType, ulPinLen);
 	switch (userType) {


Jakuje:
The code change looks good.

At this moment, the tests for iso-applet are not very extensive, but would it make sense to try to add a regression test with a reproducer?

The least work would be dumping few commands into the [.travis.yml](https://github.com/OpenSC/OpenSC/blob/master/.travis.yml#L254), but much better would be creating some bash script containing these tests with comments that would be invoked from the `.travis.yml`. You are certainly more familiar with ISO applet so I believe you would be able to come up with few more use cases that would be nice to test.

frankmorgner:
Did you check the other PKCS#11 entry points for this possible problem?

llogar:
@Jakuje:
> At this moment, the tests for iso-applet are not very extensive, but would it make sense to try to add a regression test with a reproducer?

I don't know. I've only noticed this as I am using a modified version of _IsoApplet_ where among other improvements I can use PKCS#11 functions to initialise the token (`C_InitToken` & `C_InitPin`) instead of the IsoApplet "official" `pkcs15-init --create-pkcs15 ...`. And also due to the issue below you actually couldn't invoke a `C_InitToken()` on an applet running in a simulator (and catch this bug):

ATR used in jcardsim config (3B:80:80:01:01) clashes with _SmartCard-HSM with fingerprint sensor and PIN pad_ from `opensc.conf` which can have some undesired side effects - for example `pkcs11_enable_InitToken` setting is not honored in `card_detect()`. It would probably be wise to change it to something else.

> The least work would be dumping few commands into the [.travis.yml](https://github.com/OpenSC/OpenSC/blob/master/.travis.yml#L254), but much better would be creating some bash script containing these tests with comments that would be invoked from the `.travis.yml`. You are certainly more familiar with ISO applet so I believe you would be able to come up with few more use cases that would be nice to test.

Perhaps adding pin change and pin unblock tests?

```
pkcs15-tool --change-pin --pin 123456 --new-pin 654321
pkcs15-tool --unblock-pin --puk 0123456789abcdef --new-pin 123456
```
Apart from these, I think the basics are pretty much covered...

@frankmorgner: No, but now that you have mentioned it, I have tried a couple more possible problematic scenarios (very unsystematic, I know) and it didn't crash. So I hope this was the only one...

Jakuje:
Can you add these two cheks to the test and rebase on current master to get the CI green?

Jakuje:
Good to see the results green and check added. Can you rebase the change on top of current master (instead of the merge commit you did?). Just `git rebase upstream/master` in your branch should do that.

.travis.yml
```diff
@@ -153,7 +153,7 @@ before_script:
       javac -classpath jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar IsoApplet/src/net/pwendland/javacard/pki/isoapplet/*.java;
       echo "com.licel.jcardsim.card.applet.0.AID=F276A288BCFBA69D34F31001" > isoapplet_jcardsim.cfg;
       echo "com.licel.jcardsim.card.applet.0.Class=net.pwendland.javacard.pki.isoapplet.IsoApplet" >> isoapplet_jcardsim.cfg;
-      echo "com.licel.jcardsim.card.ATR=3B80800101" >> isoapplet_jcardsim.cfg;
+      echo "com.licel.jcardsim.card.ATR=3B8B8001805949736F4170706C6574BA" >> isoapplet_jcardsim.cfg;
```
frankmorgner, line 156: Why do you choose this ATR? I think, IsoApplet doesn't require any specific one...

frankmorgner:
Sigh, this seems to be a generic problem. Almost nowhere there's a check for `p15_card != NULL` 🙄 

Also, I think we should to change the return code to `CKR_TOKEN_NOT_RECOGNIZED`

frankmorgner:
I think I've found all the missing checks. Could you check if https://github.com/OpenSC/OpenSC/pull/2049 works as well? I've also changed back the ATR; having the same ATR for different cards with different applets should not be a problem.

.travis.yml
llogar: I chose this specific ATR for no reason, any unique one would do. 
As I explained in my answer to @Jakuje, If you leave the stock ATR, then you can not initialize the token using _PKCS#11_ functions (C_InitToken & C_InitPin) as `pkcs11_enable_InitToken` is taken from the `card_atr 3B:80:80:01:01` section which "belongs" to one of the SmartCard-HSMs, unless of course you add  `pkcs11_enable_InitToken=yes` to that section...

llogar:
> Also, I think we should to change the return code to `CKR_TOKEN_NOT_RECOGNIZED`

per specification `CKR_TOKEN_NOT_RECOGNIZED `is not among the possible return codes of C_Login(). I don't know how strict are we about this...

frankmorgner:
> ATR used in jcardsim config (3B:80:80:01:01) clashes with SmartCard-HSM with fingerprint sensor and PIN pad from opensc.conf which can have some undesired side effects - for example pkcs11_enable_InitToken setting is not honored in card_detect(). It would probably be wise to change it to something else.

`3B:80:80:01:01` is an unspecific ATR, which is why [we're catching this ATR in card-sc-hsm.c only for specific stuff](https://github.com/OpenSC/OpenSC/commit/3e7f7e632c8562db7ffbe9d712dd020cb806825d), but not for locking this into a specific card type. We have removed this from the default configuration. I think currently there's no way to avoid modifying opensc.conf for initializing a token via PKCS#11. This would mean that you can use any (specific or unspecific ATR to do this), right?

frankmorgner:
> per specification `CKR_TOKEN_NOT_RECOGNIZED `is not among the possible return codes of C_Login(). I don't know how strict are we about this...

Thanks for the pointer, I've now checked the return of `CKR_TOKEN_NOT_RECOGNIZED` to match the specified return values.

llogar:
> `3B:80:80:01:01` is an unspecific ATR, which is why [we're catching this ATR in card-sc-hsm.c only for specific stuff](https://github.com/OpenSC/OpenSC/commit/3e7f7e632c8562db7ffbe9d712dd020cb806825d), but not for locking this into a specific card type. We have removed this from the default configuration. I think currently there's no way to avoid modifying opensc.conf for initializing a token via PKCS#11. This would mean that you can use any (specific or unspecific ATR to do this), right?

Right. And this brings me back to #1543.

PR #2050, opened 2020-06-04
Addressing new issues from oss-fuzz

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-tcos.c
+++ b/src/libopensc/card-tcos.c
@@ -611,40 +611,43 @@ static int tcos_decipher(sc_card_t *card, const u8 * crgram, size_t crgram_len, 
 		return SC_ERROR_INVALID_ARGUMENTS;
 	}
 	ctx = card->ctx;
-	tcos3=(card->type==SC_CARD_TYPE_TCOS_V3);
-	data=(tcos_data *)card->drv_data;
+	tcos3 = (card->type == SC_CARD_TYPE_TCOS_V3);
+	data = (tcos_data *)card->drv_data;
 
 	LOG_FUNC_CALLED(ctx);
 	sc_log(ctx,
-		"TCOS3:%d PKCS1:%d\n",tcos3,
+		"TCOS3:%d PKCS1:%d\n", tcos3,
 		!!(data->pad_flags & SC_ALGORITHM_RSA_PAD_PKCS1));
 
-	sc_format_apdu(card, &apdu, crgram_len>255 ? SC_APDU_CASE_4_EXT : SC_APDU_CASE_4_SHORT, 0x2A, 0x80, 0x86);
+	sc_format_apdu(card, &apdu, crgram_len > 255 ? SC_APDU_CASE_4_EXT : SC_APDU_CASE_4_SHORT, 0x2A, 0x80, 0x86);
 	apdu.resp = rbuf;
 	apdu.resplen = sizeof(rbuf);
 	apdu.le = crgram_len;
 
 	apdu.data = sbuf;
-	apdu.lc = apdu.datalen = crgram_len+1;
+	apdu.lc = apdu.datalen = crgram_len + 1;
 	sbuf[0] = tcos3 ? 0x00 : ((data->pad_flags & SC_ALGORITHM_RSA_PAD_PKCS1) ? 0x81 : 0x02);
 	if (sizeof sbuf - 1 < crgram_len)
 		return SC_ERROR_INVALID_ARGUMENTS;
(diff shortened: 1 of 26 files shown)


src/libopensc/pkcs15-esteid2018.c
```diff
@@ -61,6 +61,8 @@ static int sc_pkcs15emu_esteid2018_init(sc_pkcs15_card_t *p15card) {
 	for (j = 0; j < taglen; j++)
 		if (!isalnum(tag[j]))
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
+	if (p15card->tokeninfo->serial_number)
+		free(p15card->tokeninfo->serial_number);
```
metsma, line 65: You can free safely null pointer, if check is redundant.

> The free() function shall cause the space pointed to by ptr to be deallocated; that is, made available for further allocation. If ptr is a null pointer, no action shall occur.

Jakuje: Right. I did it this way as it was done similarly in all the other places, but I will try to adjust that, probably using `set_string` function, which is used in some of the drivers.

Jakuje:
seems like @frankmorgner already fixed some of these in the meantime so I will rebase the changes on top of his.

src/libopensc/card-tcos.c
```diff
@@ -571,13 +586,23 @@ static int tcos_compute_signature(sc_card_t *card, const u8 * data, size_t datal
...
+			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_INVALID_ARGUMENTS);
+		}
+
 		sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0x2A,0x80,0x86);
-		for(i=0; i<sizeof(sbuf);++i) sbuf[i]=0xff;
-		sbuf[0]=0x02; sbuf[1]=0x00; sbuf[2]=0x01; sbuf[keylen-datalen]=0x00;
-		memcpy(sbuf+keylen-datalen+1, data, datalen);
-		dlen=keylen+1;
+		for (i = 0; i < sizeof(sbuf); ++i) {
+			sbuf[i] = 0xff;
```
frankmorgner, line 598: `memset(sbuf, 0xff, sizeof sbuf)`?

Jakuje: This is part of reformat-only commit where I did not want to introduce any functional changes. But as we are touching it, it should be fine to change this too. Thanks for suggestion. I will add it.

Jakuje:
> Looks good, thanks. I've been a bit quick resolving some of the issues on my own and pushing them to master. If you wish, I'll revert them and take your PR instead; your changes go even deeper, even respecting the formatting.

Np. Your changes look good. I decided to tweak just a few from what you already fixed.

Pushed now on top of current master addressing recommendations from @metsma and adding your suggested improvement.

Jakuje:
Rebased on current master and added one more coverity issue reported recently

src/libopensc/sc.c
```diff
@@ -237,6 +237,12 @@ unsigned long lebytes2ulong(const u8 *buf)
 		| (unsigned long)buf[0];
 }
 
+void set_string(char **strp, const char *value)
+{
+	free(*strp);
+	*strp = value ? strdup(value) : NULL;
```
frankmorgner, line 243: Couls you surround this with `if (strp)`, just to be extra sure?

src/libopensc/pkcs15.c
```diff
@@ -220,6 +220,8 @@ int sc_pkcs15_parse_tokeninfo(sc_context_t *ctx,
 	LOG_TEST_RET(ctx, r, "ASN.1 parsing of EF(TokenInfo) failed");
 
 	if (asn1_toki_attrs[1].flags & SC_ASN1_PRESENT && serial_len > 0)   {
+		if (ti->serial_number)
```
frankmorgner, line 223: this check could also be omitted

src/libopensc/sc.c
Jakuje: Thanks. Added
```
if (strp == NULL) {
        return;
}
```
to avoid too much indentation.

src/libopensc/pkcs15.c
Jakuje: Thanks for review. Removed now.

Jakuje:
All comments fixed and rebased on current master again. There is also new commit from this morning addressing another issue from oss-fuzz.

PR #2051, opened 2020-06-05
Fix Hotplug Events

### LINKED ISSUE #2021: C_WaitForSlotEvent not returning.
### Problem Description
C_WaitForSlotEvent is not returning when a card is inserted or removed since commit 2c26b7392d8d0d7e1db2b6143583376758074dda to the master branch.

### Proposed Resolution
Fix the bug.
### Steps to reproduce
Call C_WaitForSlotEvent with the slot ID of a reader without a card.
The function will not return when a card is inserted.
### Logs
PKCS#11 spy logs before and after the commit referenced above.
Version before the commit:
[0.20.0-215-g7893d286.log](https://github.com/OpenSC/OpenSC/files/4552334/0.20.0-215-g7893d286.log)
The commited version:
[0.20.0-216-g2c26b739.log](https://github.com/OpenSC/OpenSC/files/4552335/0.20.0-216-g2c26b739.log)

frankmorgner:
Fixes #2021 

##### Checklist

- [x] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -1741,6 +1741,10 @@ static int pcsc_wait_for_event(sc_context_t *ctx, unsigned int event_mask, sc_re
 			}
 		}
 
+		/* if a reader was detected, we need to create a new list of readers */
+		if (detected_hotplug)
+			goto out;
+
 		/* Set the timeout if caller wants to time out */
 		if (timeout == -1) {
 			dwtimeout = INFINITE;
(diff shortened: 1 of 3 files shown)


Jakuje:
Tested with one physical card reader with card and yubikey. Works much better now, but the yubikey insertion yields two events (or the state is not correctly updated after the reader insertion?) if I see right. The second call to `C_WaitForSlotEvent` after inserting yubikey returns immediately with event on slot 1, while the first call has event on different slot ID (hotplug ID?). Not sure if this is intentional or not.

Attaching PKCS11 trace:
[yk_hotplug.log](https://github.com/OpenSC/OpenSC/files/4736565/yk_hotplug.log)

frankmorgner:
I've pushed a fix for detecting a new card twice. Could you please check again?

Jakuje:
Looks good now. Weirdly enough, when I tried if it will detect insertion of empty reader, it did not, but given that empty reader is not useful for anything, I think it is fine. Hopefully it will not break anything else.

Tested also with `opensc-notify`, which looks like doing detecting the same events too.

PR #2055, opened 2020-06-10
One more bunch of fixes for oss-fuzz

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-coolkey.c
+++ b/src/libopensc/card-coolkey.c
@@ -1675,55 +1675,43 @@ typedef struct coolkey_compute_ecc_params {
 	u8 buf[MAX_COMPUTE_BUF];
 } coolkey_compute_ecc_params_t;
 
-static int coolkey_rsa_op(sc_card_t *card,
-					const u8 * data, size_t datalen,
-					u8 * out, size_t max_out_len)
+static int coolkey_rsa_op(sc_card_t *card, const u8 * data, size_t datalen,
+	u8 * out, size_t max_out_len)
 {
 	int r;
-	const u8 *crypt_in;
-	u8 **crypt_out_p;
-	size_t crypt_in_len, *crypt_out_len_p;
-	coolkey_private_data_t * priv = COOLKEY_DATA(card);
+	u8 **crypt_out_p = NULL;
+	size_t crypt_out_len_p;
+	coolkey_private_data_t *priv = COOLKEY_DATA(card);
 	coolkey_compute_crypt_params_t params;
 	u8 key_number;
 	size_t params_len;
-	size_t buf_len;
-	u8 buf[MAX_COMPUTE_BUF+2];
-	u8 *buf_out;
+	u8 buf[MAX_COMPUTE_BUF + 2];
 
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
-	sc_log(card->ctx, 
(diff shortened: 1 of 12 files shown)


src/libopensc/card-coolkey.c
```diff
@@ -1757,6 +1757,9 @@ static int coolkey_rsa_op(sc_card_t *card,
 		u8 len_buf[2];
 		size_t out_length;
 
+		/* Free card response -- nothing useful -- result is in export object */
+		free(*crypt_out_p);
```
frankmorgner, line 1761: Are you sure that this is correct? It looks like there is some case where this calls free on a buffer on the heap:
https://github.com/OpenSC/OpenSC/blob/0d958c1f7d8ea34e3e127bf560dcbf963b9147a9/src/libopensc/card-coolkey.c#L1702-L1703

Jakuje: The code is very confusing, but most important is that `crypt_in_len = datalen;` and therefore the conditions `if (crypt_in_len > MAX_COMPUTE_BUF) {` and `if (datalen > MAX_COMPUTE_BUF) {` are equal. In the first one we are setting

https://github.com/OpenSC/OpenSC/blob/0d958c1f7d8ea34e3e127bf560dcbf963b9147a9/src/libopensc/card-coolkey.c#L1725

which keeps the io function to allocate the memory. This probably does not happen with real cards, but fuzzer can return some mess somethimes.

frankmorgner: `crypt_out_p` is only reset `if (crypt_in_len > MAX_COMPUTE_BUF)`, so there seems to be the other case where it is *not* set to `NULL`. Would it be possible to clearify the code, e.g. by using different variables?

Jakuje: I will try to improve that part.

Jakuje: Rewritten in a8101c26. Does it look better for you now?

Jakuje: The final version is in 6b66c43 as the initial had some issues to build. Looks fine now.

Jakuje:
I added one more report from oss-fuzz today and rebased on top of current master.

src/libopensc/pkcs15-sc-hsm.c
```diff
@@ -862,6 +864,7 @@ static int sc_pkcs15emu_sc_hsm_init (sc_pkcs15_card_t * p15card)
 	appinfo->aid = sc_hsm_aid;
 
 	appinfo->ddo.aid = sc_hsm_aid;
+	sc_pkcs15_free_app(p15card);
```
frankmorgner, line 867: I assume some other initialization routine initialized this, but failed in the end (`sc_pkcs15_bind_internal()`?) This routine would be a better place for cleaning this up, wouldn't it?

frankmorgner:
(Big thank you, for keeping up with the reported issues :tulip: )

src/libopensc/pkcs15-sc-hsm.c
Jakuje: Actually, this one was more shot in the dark. Correct fix was 31e442f3 as you propose in pkcs15. I will drop ddea3c73 and 7d40797 as they do not seem to be much helpful.

PR #2059, opened 2020-06-11
SECURITY.md: Introduce security reporting process

Jakuje:
As part of CII Best Practices evaluation, there is a step for making sure we have documentation how to report security vulnerabilities. Github proposes the `SECURITY.md` to do that. I tried to fill their template, but I would like to hear if it reflects previously used processes, if there is something to change or improve in this direction.

https://bestpractices.coreinfrastructure.org/en/projects/3908/

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+OpenSC releases are made roughly once a year, unless important security is discovered.
+
+OpenSC does not release micro updates for previously released versions and does not
+backport security fixes into them.
+
+| Version  | Supported          |
+| -------- | ------------------ |
+| 0.20.0   | :white_check_mark: |
+| < 0.20.0 | :x:                |
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


Jakuje:
I would like also others members to review the questionnaire (everyone with commit access here should be able to update it), whether they see some space for improvement, if I missed something or there is something to add. After having the security process,I believe we will have the checklist complete and we can add the badge.

dengert:
"It is SUGGESTED that the test suite cover most (or ideally all) the code branches, input fields, and function"

very large code base and poor testing history makes this hard to achieve yet.

Could also say project does not have all smart cards for testing. And not able to tell if cards are still being used.

Jakuje:
> "It is SUGGESTED that the test suite cover most (or ideally all) the code branches, input fields, and function"
> 
> very large code base and poor testing history makes this hard to achieve yet.
> 
> Could also say project does not have all smart cards for testing. And not able to tell if cards are still being used.

Thanks for suggestion. I added that part.

frankmorgner:
thank you!

PR #2062, opened 2020-06-16
Add sc_free()

### LINKED ISSUE #2054: Missing sc_free() or similar
### Problem Description

When using the function sc_asn1_sig_value_rs_to_sequence() on Windows using the DLL, I cannot free the returned buffer.

The reason is that the provide OpenSC DLL use a different C runtime then the program loading the DLL. So calling a free in my program might use a different heap implementation then the DLL has compiled in, which leads to problems. (Might even crash).

Actually to the OpenSC 32bit DLL references only Windows system DLLs, so there is no chance to get the correct free() implementation as all is internal to the OpenSC DLL.

You have a similar problem if you load the DLL using cffi, then you need a C runtime function to free the buffer. (this is not only Windows)

For other data structure there exists corresponding functions (sc_file_new and sc_file_free, and others). But there is no function to return the buffer I get with sc_asn1_sig_value_rs_to_sequence.

### Proposed Resolution

Add to sc.c a new function and export it in the DLL interface:

`void sc_free(void *p)
{
    free(p);
}`

So I can free any buffer opensc has allocated by using a dedicated opensc function, which calls the correct free implementation.

There might need to be added  also some hints at function documention of sc_asn1_sig_value_rs_to_sequence (and others ?) to use sc_free to free the returned buffer.

### Steps to reproduce

Create a small test program using the function above and compile it with MinGW (http://http://mingw-w64.org). MinGW will link against MSVCRT and use its free function.
If run in debugger it stops at the free with "Thread 1 received signal SIGTRAP, Trace/breakpoint trap."
Without debugger you get a message box like this "Access violation at address 76FD68EA in module 'ntdll.dll'. Read of address C2D71BFD."

rliebscher:
Add sc_free() function

Fixes #2054

### DIFF
--- a/src/libopensc/libopensc.exports
+++ b/src/libopensc/libopensc.exports
@@ -381,3 +381,4 @@ sc_notify
 sc_notify_id
 sc_color_fprintf
 iso7816_update_binary_sfid
+sc_free
--- a/src/libopensc/opensc.h
+++ b/src/libopensc/opensc.h
@@ -1656,6 +1656,15 @@ int iso7816_update_binary_sfid(sc_card_t *card, unsigned char sfid,
  * */
 int iso7816_logout(sc_card_t *card, unsigned char pin_reference);
 
+/**
+ * Free a buffer returned by OpenSC.
+ * Use this instead your C libraries free() to free memory allocated by OpenSC.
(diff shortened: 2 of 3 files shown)


Jakuje:
The AppVeyor failure is nothing important so, it looks good to me.

PR #2063, opened 2020-06-16
Move to macOS Utilities folder

metsma:

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
@frankmorgner can you check this one if it is fine to merge?

frankmorgner:
Isn't this folder reserved for the builtin system utilities?

metsma:
Adobe and friends add also items there and since macOS 10.15 it is separated into 2 folders.
/Applications/Utilities for third party and /System/Applications/Utilities for system readonly folder

frankmorgner:
nice to know :+1: Then we should only install it to Utilities on 10.15 and later, right?

metsma:
You can install previous versions also. Adobe has done so always.
Screenshot 2020-06-23 09 18 20

PR #2064, opened 2020-06-16
Remove compiler warnings/errors

rliebscher:
Recent compilers have activated some additional
checks which let the build fail. (at least with cygwin)
(Normally it would be warnings but opensc compiles
with -Werror)

GCC 9.3:
```
In file included from profile.c:27:
profile.c: In function '__expr_get':
profile.c:2273:18: error: array subscript has type 'char' [-Werror=char-subscripts]
 2273 |   while (isspace(*s))
      |                  ^~
```
clang 8.0.1:
```
compat_getopt_main.c:102:22: error: array subscript is of type 'char' [-Werror,-Wchar-subscripts]
                rc = toupper(rc);
                     ^~~~~~~~~~~
/usr/include/ctype.h:161:25: note: expanded from macro 'toupper'
      (void) __CTYPE_PTR[__x]; (toupper) (__x);})
                        ^~~~
```
Actually the code is correct as isspace and others
are used here with data type char, and are to be used
with data type int.

So either the compiler should have deactivated
this error, or the ctype.h macros have to be
written so the compiler no longer complains.

As there is also a simple workaround by casting
char to int, there is no need to wait for one
of the former options to be happen sometime.

### DIFF
--- a/src/common/compat_getopt_main.c
+++ b/src/common/compat_getopt_main.c
@@ -99,7 +99,7 @@ handle(char *progname,
             {
               char rc = letters[(match - letters + rotate) % 26];
               if (isupper(c))
-                rc = toupper(rc);
+                rc = toupper((unsigned char)rc);
               c = rc;
             }
         }
--- a/src/libopensc/card-sc-hsm.c
+++ b/src/libopensc/card-sc-hsm.c
@@ -284,7 +284,7 @@ static int sc_hsm_match_card(struct sc_card *card)
 static int sc_hsm_encode_sopin(const u8 *sopin, u8 *sopinbin)
 {
 	int i;
-	char digit;
+	unsigned char digit;
 
 	memset(sopinbin, 0, 8);
 	for (i = 0; i < 16; i++) {
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -274,7 +274,7 @@ int sc_format_oid(struct sc_object_id *oid, const char *in)
 		if (!*q)
 			break;
 
-		if (!(q[0] == '.' && isdigit(q[1])))
+		if (!(q[0] == '.' && isdigit((unsigned char)q[1])))
 			goto out;
 
 		p = q + 1;
--- a/src/pkcs15init/profile.c
+++ b/src/pkcs15init/profile.c
@@ -2123,7 +2123,7 @@ get_authid(struct state *cur, const char *value,
 	char	temp[16];
 	size_t	n;
 
-	if (isdigit((int) *value)) {
+	if (isdigit((unsigned char) *value)) {
 		*num = 0;
 		return get_uint(cur, value, type);
 	}
@@ -2183,7 +2183,7 @@ map_str2int(struct state *cur, const char *value,
 	unsigned int	n;
 	const char	*what;
 
-	if (isdigit((int) *value))
+	if (isdigit((unsigned char) *value))
 		return get_uint(cur, value, vp);
 	for (n = 0; map[n].name; n++) {
 		if (!strcasecmp(value, map[n].name)) {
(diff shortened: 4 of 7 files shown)


Jakuje:
Interesting. We have both clang and gcc building opensc in CI (well ... older version) and in Fedora we built already with GCC 10 and I did not see this error.

My manual page shows that `-Wchar-subscripts` is enabled with `-Wall` (which is enabled by default). Not sure what is different in your gcc than in mine. Could it be a bug in gcc 9.3.0?

I tried to rebuild current master with `-Wall` (gcc 9.3.1 and clang 9.0.1 on Fedora)
```
CFLAGS="-Wall" ./configure && make clean && make -j4
CC=clang ./configure && make clean && make -j4
```
and both of these work just fine without any errors. Or could that be an issue of c runtime on windows in cygwin?

rliebscher:
I tried both opensc-0.20.0 and current master with same result.

I guess it depends from the ctype.h implemention of these macros. Probably cygwin has to do here some work to get rid of these warnings.

But here a minimal example to let both compilers fail (when using the same CFLAGS as configure outputs at end):
```
#include <ctype.h>

extern char* x;

int main()
{
	return tolower(*x) + isdigit(*x);
}
```
gcc:
```
$ gcc -Wall -Wextra -Wno-unused-parameter -Werror -g -O2 a.c
In file included from a.c:1:
a.c: In function 'main':
a.c:7:31: error: array subscript has type 'char' [-Werror=char-subscripts]
    7 |  return tolower(*x) + isdigit(*x);
      |                               ^~
```
clang
```
$ clang -Wall -Wextra -Wno-unused-parameter -Werror -g -O2 a.c
a.c:7:9: error: array subscript is of type 'char' [-Werror,-Wchar-subscripts]
        return tolower(*x);
               ^~~~~~~~~~~
/usr/include/ctype.h:164:25: note: expanded from macro 'tolower'
      (void) __CTYPE_PTR[__x]; (tolower) (__x);})
                        ^~~~
1 error generated.
```

rliebscher:
BTW: If found some comment in my ctype.h, so they trigger the warning intentionally.
```
/* These macros are intentionally written in a manner that will trigger
   a gcc -Wall warning if the user mistakenly passes a 'char' instead
   of an int containing an 'unsigned char'.  Note that the sizeof will
   always be 1, which is what we want for mapping EOF to __CTYPE_PTR[0];
   the use of a raw index inside the sizeof triggers the gcc warning if
   __c was of type char, and sizeof masks side effects of the extra __c.
   Meanwhile, the real index to __CTYPE_PTR+1 must be cast to int,
   since isalpha(0x100000001LL) must equal isalpha(1), rather than being
   an out-of-bounds reference on a 64-bit machine.  */
```

Jakuje:
Thank you for the pointer. It looks like cygwins ctype.h specialty. There is some discussion about this in glibc [1], where the table indices are expanded also to the negative values [2] so this is not an issue with glibc.

But strictly according to the POSIX, it is indeed correct to make sure the value is cast to correct data type. Although my manual page recommends the following:

> The standards require that the argument c for these functions is either EOF or a value that is representable in the type unsigned char.  If the argument c is of type char, it must be cast to unsigned char, as in the following example:
> 
>            char c;
>            ...
>            res = toupper((unsigned char) c);

So even though the signature is indeed int, we should probably use the above to strictly limit the range of input. Would that work for you?

[1] https://sourceware.org/bugzilla/show_bug.cgi?id=11435 
[2] https://sourceware.org/git/?p=glibc.git;a=blob;f=locale/C-ctype.c;h=06c0d1ee404edf639eea27b90dbf62f001695631;hb=HEAD#l238

rliebscher:
Changed to unsigned char now and tested cygwin still compiles.

PR #2066, opened 2020-06-18
YubiKey5 PIV changes using Attestation

dengert:
Yubico version 4 and 5 devices allow a pin and touch policy to be set
during generation or importation of a key. They also support the creation
of an attestation certificate for a key with certificate extensions, one
of which has the pin and touch policies for the key.

NIST 800-73 PIV specifications says the 9C key will require the
PIN for every use of the key. PIV cards enforce this "PIN Always" policy
on the card. (the 9E key does not require a PIN.)

But the Yubico devices have extensions that do not always follow the
NIST specifications. This has lead to caching of PINs and changes
to opensc.conf to be able to "turn off" the "PIN Always" policy.

If there is an attestation certificate for the 9C PIV Signing key,
and the pin policy extension is not the default=0, or "always=3" this
means the card will not enforce the NIST PIV "PIN Always".
card-piv.c and pkcs15-piv.c will then mark the 9C key as not
requiring "PIN Always" aka "user_consent" aka "CKA_ALWAYS_AUTHENTICATE"
and the key will be treated as any other key.

This PR could have gone further, and checked and set pin "NEVER"
much like the PIV 9E key and touch polices, but is not clear how to
handle via PKCS#11 how to tell user to touch the button with or without
a PIN.

Tested with Yubikey 5 NFC version 5.2.6. and pkcs11-tool --test --login

The 5.2.6 device appears to have fixed all the other "card_issues"
as listed in card-piv.c for other cards and keeps the login state correctly.

See:
https://developers.yubico.com/PIV/Introduction/PIV_attestation.html
https://developers.yubico.com/yubico-piv-tool/Manuals/yubico-piv-tool.1.html
Or contact Yubico support.

 On branch YubiKey5
 Changes to be committed:
	modified:   card-piv.c
	modified:   cardctl.h
	modified:   cards.h
	modified:   pkcs15-piv.c

Fixes:#2039 But requires the policy flags to be set when generating or importing the key
and the attestation certificate need to be created. 


Jakuje:
lgtm, though I can not test it as I have YK 4 only.

dengert:
Yubico says It should work on a 4.3 and above.

Jakuje:
> Yubico says It should work on a 4.3 and above.

Oh right, mine is 4.3.something. Tested locally and reading the debug log shows it is correctly used (though the default is always authenticate and I do not plan to re-generate the key now).

src/libopensc/pkcs15-piv.c
```diff
@@ -1143,7 +1143,56 @@ sc_log(card->ctx,  "DEE Adding pin %d label=%s",i, label);
...
+				r = sc_pkcs15_read_certificate(p15card, attcert_info, &attcert_out);
+				if (r >= 0 && attcert_out && attcert_out->extensions) {
+					r = sc_pkcs15_get_extension(card->ctx, attcert_out,
+							&pp_oid, &pp_val, &pp_len, &pp_critical);
+					/* pp_val[0] pin policy default = 0, none = 1, session = 2, always = 3 */
+					/* pp_val[1] touch policy. TODO can we support this in some way */
+					if (r >= 0 && pp_len == 2 && pp_val) {
+						if (*pp_val != 0 && *pp_val != 3) {
+							prkey_obj.user_consent = 0; /* turn off */
+							sc_log(card->ctx, "user_consent turned off for 0x%2.2x", prkeys[i].ref);
```
frankmorgner, line 1186: Would it be possible to evaluate the touch policy in the card driver? Everytime that key is used (and the user needs to touch), the card driver could send a notification so that the user is aware that he needs to do something. Similarily, [I've used `sc_notify()` in OpenSCTokenApp to tell the user to look at the PIN-pad of his reader](https://github.com/frankmorgner/OpenSCToken/blob/5359efaad81f2593526b7b30c3d423b840b21aa4/OpenSCToken/TokenSession.m#L128-L139). You could use the Yubikey's ATR to customize `MD_PINPAD_DLG_CONTENT_USER` and `MD_PINPAD_DLG_MAIN` with text for the touch button.

dengert: Yes it could be done
My goal with this PR was to address the non standard PIV card's ability to set and enforce policies on the card for the 9c signing key.  I was not trying to address the touch policy or other keys. 

It could also be expanded to look at all keys and the "none" key policy. (PIV standard has this for the 9e key). 

(At my request, they sent me a YubiYey 5 NFC device with version 5.2.6 to use in testing of the Attestation certificate.)
Yubico is adding in 5.3 a simpler way to get the pin and touch policy without using the Attestation certificate. 

With Yubico devices (identified from their ATR) card-piv.c reads the Yubico version number to further distinguish differences in their devices. 

You could add this on the wish list.

frankmorgner: I'd like to avoid moving this from "not supported"" to "somewhat supported"; this only confuses users. For example, applications like MS Outlook are hanging during PIN verification. Adding a notification is the least we can do to get the user's attention so he has some clue of what he's supposed to do. I'd even go for a popup Window, but that would require even more work.

dengert:
In https://github.com/OpenSC/OpenSC/pull/2066/commits/06a967b8ab5f27206d7d75285773afc5f8423993 I took the simple approach to send hint via the pin prompt. The text can easily be changed. Suggestions are welcome. 

Unfortunately older Yubico devices may have touch policies but there is no way to know in advance. 

@frankmorgner , You asked about sending a notification. 
A more general solution for older Yubico and other devices that may require being touched could be added to 
a timer for 10 seconds (or so)  if crypto operation has not completed, to notification to user, 

Yubico also says touch policy can be used with OpenPGP applet. Other none Yubico devices may have similar capabilities, so this is more than a PIV only problem. 

 I will leave that up to you to address this.

frankmorgner:
> In [06a967b](https://github.com/OpenSC/OpenSC/commit/06a967b8ab5f27206d7d75285773afc5f8423993) I took the simple approach to send hint via the pin prompt. The text can easily be changed. Suggestions are welcome.
> 
> Unfortunately older Yubico devices may have touch policies but there is no way to know in advance.
> 
> @frankmorgner , You asked about sending a notification.
> A more general solution for older Yubico and other devices that may require being touched could be added to
> a timer for 10 seconds (or so) if crypto operation has not completed, to notification to user,
> 
> Yubico also says touch policy can be used with OpenPGP applet. Other none Yubico devices may have similar capabilities, so this is more than a PIV only problem.
> 
> I will leave that up to you to address this.

I think that the touch policy is only detectable in PIV, because there's no vendor extension of PGP for this.

Your solution of adding this information to the PIN label looks good to me. However, I would slightly change the wording to make it more digestable for the average user. What do you think about *XXX with touch required*. This is the wording, which [Yubico uses for the Yubikey manager](https://support.yubico.com/support/solutions/articles/15000006419-using-your-yubikey-with-authenticator-codes).

src/libopensc/pkcs15-piv.c
```diff
@@ -1200,6 +1220,61 @@ sc_log(card->ctx,  "DEE Adding pin %d label=%s",i, label);
...
+					&pin_ref) == 0 &&
+				pin_ref == 0x00) { /* must be 80 for PIV pin, or 00 for Global PIN */
+			pin_info.attrs.pin.reference = pin_ref;
+			pin_info.attrs.pin.flags &= ~SC_PKCS15_PIN_FLAG_LOCAL;
+			strncpy(label, "Global PIN", SC_PKCS15_MAX_LABEL_SIZE -1);
+		}
+		if (has_yubikey_touch_policy) {
+			int offset = strlen(label);
+			strncpy(label + offset, ",Yubikey has touch policy", SC_PKCS15_MAX_LABEL_SIZE - offset - 1);
+		}
```
frankmorgner, line 1258: I think there are some corner cases when copying the label around.

since `label` is uninitialized it may not contain a termination. When you're copying the label (line 1246), this is still the case if the label is completely filled, because you're omitting the label's termination. This will then lead to a bad calculation of the length of `label`. Also, the calculation of index and lenght looks fishy.

Why don't you use strlcat?

frankmorgner (changes requested):

dengert:
Right now this is on hold. Only 3 of us have commented, no users pushing for it. 

It also may have conflicts with #2053 which I am much more interested in getting in next release first.

frankmorgner (changes requested):
Some more comments. Note that there are still some comments left from the previous review.

dengert:
I have not updated this for a number of reasons:

-  no interest was shown for this PR
-  in later versions of Yubikey, there is a simpler way to do this
-  the button does not blink when using NFC, so better prompting is needed
-  the PIV Secure Messaging  #2053 code may have conflicts with this PR.
-  I  want to see the #2053 added before spending time on this PR.

frankmorgner:

dengert:
The status has not changed  since: https://github.com/OpenSC/OpenSC/pull/2066#issuecomment-733808054
No interest has been shown and conflicts exist newer changes.  So I am closing it.

PR #2067, opened 2020-06-26
PKCS11 C_Initialize locking

### LINKED ISSUE #2032: Opensc 0.20.1.4 and FF ver. 74
Not sure if related to https://github.com/OpenSC/OpenSC/issues/1999 or not. Symptoms are a little bit different. 

The testing was made with the new latest OpenSC 0.20.1.4 and FF 74. 
- Firefox "not responding" but didn't close itself
- Every tab, what is being made brings up "not responding" response

Some things, what we're noticed:
1. Delay because of the smart card?
```
Outgoing APDU (4 bytes):
00 A4 00 0C ....
P:2124; T:2460 2020-05-08 16:05:34.100 [onepin-opensc-pkcs11] reader-pcsc.c:242:pcsc_internal_transmit: called
```
2. Interesting error codes

`P:2124; T:2460 2020-05-08 16:05:29.600 [onepin-opensc-pkcs11] OMNIKEY CardMan 1021 0:SCardTransmit/Control failed: 0x0000045d`

```
ERROR_IO_DEVICE
1117 (0x45D)
The request could not be performed because of an I/O device error.
```

3. Problem is starting near:

```
P:2124; T:2460 2020-05-08 15:55:52.950 [onepin-opensc-pkcs11] reader-pcsc.c:324:pcsc_transmit: 
Outgoing APDU (4 bytes):
00 A4 00 0C ....
P:2124; T:2460 2020-05-08 15:55:52.950 [onepin-opensc-pkcs11] reader-pcsc.c:242:pcsc_internal_transmit: called
P:2124; T:2460 2020-05-08 15:55:57.059 [onepin-opensc-pkcs11] OMNIKEY CardMan 1021 0:SCardTransmit/Control failed: 0x0000045d
```
Added log file in attachment. Hope something comes up.

[opensc-debug.txt](https://github.com/OpenSC/OpenSC/files/4608904/opensc-debug.txt)

dengert:
C_Initialize may be called by multiple threads.
But while trying to setup an OpenSC context, setup global_locking
and detect cards, it is possible that multiple threads may be trying
to do this, as the only test is "if (context == NULL)" but this
may not be set  until it is too late, and  things may be overwritten
such as existing context, causing additional problems, with pcsc.

FireFox appears to do this see #2032

The PR adds a mutex or Critical section to make sure only one
thread creates the context, and does the initial detect cards, etc.

pkcs11-tool.c adds a --test_threads option.  When given only once,
CK_C_INITIALIZE_ARGS  c_initialize_args = {NULL, NULL, NULL, NULL, CKF_OS_LOCKING_OK, NULL};
is passed to C_Initialize. If --test_threads is present more then once,
additional threads are created to test a module's C_Initialize.

Tests on Linux show multiple process do indeed create multiple contexts.

Not yet tested on Windows.

 On branch C_Initialize-locking
 Changes to be committed:
	modified:   ../pkcs11/pkcs11-global.c

May fix #2032 and other recent FireFox issues.

##### Checklist

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -61,6 +61,11 @@ extern CK_FUNCTION_LIST_3_0 pkcs11_function_list_3_0;
 
 #if defined(HAVE_PTHREAD)
 
+/* mutex used to control C_Initilize creation of mutexes */
+static pthread_mutex_t c_initialize_m = PTHREAD_MUTEX_INITIALIZER;
+#define C_INITIALIZE_M_LOCK  pthread_mutex_lock(&c_initialize_m);
+#define C_INITIALIZE_M_UNLOCK pthread_mutex_unlock(&c_initialize_m);
+
 CK_RV mutex_create(void **mutex)
 {
 	pthread_mutex_t *m;
@@ -101,6 +106,9 @@ static CK_C_INITIALIZE_ARGS _def_locks = {
 #define HAVE_OS_LOCKING
 
 #elif defined(_WIN32)
+CRITICAL_SECTION c_initialize_cs = {0};
+#define C_INITIALIZE_M_LOCK EnterCriticalSection(&c_initialize_cs);
+#define C_INITIALIZE_M_UNLOCK LeaveCriticalSection(&c_initialize_cs);
 
 CK_RV mutex_create(void **mutex)
 {
@@ -140,6 +148,10 @@ static CK_C_INITIALIZE_ARGS _def_locks = {
 #define HAVE_OS_LOCKING
 
 #endif
+
+#else /* PKCS11_THREAD_LOCKING */
+#define C_INITIALIZE_M_LOCK
+#define C_INITIALIZE_M_UNLOCK
 
 #endif /* PKCS11_THREAD_LOCKING */
 
@@ -221,6 +233,9 @@ __attribute__((constructor))
 #endif
 int module_init()
 {
+#ifdef _WIN32
+	InitializeCriticalSection(&c_initialize_cs);
+#endif
 	sc_notify_init();
 	return 1;
 }
@@ -236,6 +251,9 @@ int module_close()
 #endif
 #ifdef ENABLE_OPENPACE
 	EAC_cleanup();
+#endif
+#ifdef _WIN32
+	DeleteCriticalSection(&c_initialize_cs);
 #endif
 	return 1;
 }
@@ -283,8 +301,12 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
 	in_finalize = 0;
 #endif
 
+	/* protect from multiple threads tryng to setup locking */
+	C_INITIALIZE_M_LOCK
+
 	if (context != NULL) {
 		sc_log(context, "C_Initialize(): Cryptoki already initialized\n");
+		C_INITIALIZE_M_UNLOCK
 		return CKR_CRYPTOKI_ALREADY_INITIALIZED;
 	}
 
@@ -335,6 +357,9 @@ out:
 		/* Release and destroy the mutex */
 		sc_pkcs11_free_lock();
 	}
+
+	/* protect from multiple threads tryng to setup locking */
+	C_INITIALIZE_M_UNLOCK
 
 	return rv;
 }
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -426,6 +426,22 @@
 
 				<varlistentry>
 					<term>
+						<option>--use-locking</option>
+					</term>
+					<listitem><para>Tell pkcs11 module it should use OS thread locking.
+					</para></listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>
+						<option>--test-threads</option> <replaceable>options</replaceable>
+					</term>
+					<listitem><para>Test a pkcs11 module's thread implication. (See source code).
+					</para></listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>
 						<option>--token-label</option> <replaceable>label</replaceable>
 					</term>
 					<listitem><para>Specify the label of token.
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -55,9 +55,10 @@ opensc_explorer_LDADD = $(OPTIONAL_READLINE_LIBS)
 pkcs15_tool_SOURCES = pkcs15-tool.c util.c ../pkcs11/pkcs11-display.c ../pkcs11/pkcs11-display.h
 pkcs15_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 pkcs11_tool_SOURCES = pkcs11-tool.c util.c
+pkcs11_tool_CFLAGS = $(OPTIONAL_OPENSSL_CFLAGS) $(PTHREAD_CFLAGS)
 pkcs11_tool_LDADD = \
 	$(top_builddir)/src/common/libpkcs11.la \
-	$(OPTIONAL_OPENSSL_LIBS)
+	$(OPTIONAL_OPENSSL_LIBS) $(PTHREAD_CFLAGS)
 if ENABLE_SHARED
 else
 pkcs11_tool_LDADD += \
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -29,6 +29,9 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#ifdef HAVE_PTHREAD
+#include <pthread.h>
+#endif
 #else
 #include <windows.h>
 #include <io.h>
@@ -70,6 +73,10 @@
 
 #ifndef ENABLE_SHARED
 extern CK_FUNCTION_LIST_3_0 pkcs11_function_list_3_0;
+#endif
+
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+#define MAX_TEST_THREADS 10
 #endif
 
 #define NEED_SESSION_RO	0x01
@@ -151,6 +158,11 @@ enum {
 	OPT_DERIVE_PASS_DER,
 	OPT_DECRYPT,
 	OPT_TEST_FORK,
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	OPT_TEST_THREADS,
+	OPT_USE_LOCKING,
+#endif
+
 	OPT_GENERATE_KEY,
 	OPT_GENERATE_RANDOM,
 	OPT_HASH_ALGORITHM,
@@ -237,6 +249,10 @@ static const struct option options[] = {
 #ifndef _WIN32
 	{ "test-fork",		0, NULL,		OPT_TEST_FORK },
 #endif
+	{ "use-locking",	0, NULL,		OPT_USE_LOCKING },
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	{ "test-threads",	1, NULL,		OPT_TEST_THREADS },
+#endif
 	{ "generate-random",	1, NULL,		OPT_GENERATE_RANDOM },
 	{ "allow-sw",		0, NULL,		OPT_ALLOW_SW },
 
@@ -315,8 +331,12 @@ static const char *option_help[] = {
 #ifndef _WIN32
 	"Test forking and calling C_Initialize() in the child",
 #endif
+	"Call C_initialize() with CKF_OS_LOCKING_OK.",
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	"Test threads. Multiple times to start additional threads, arg is string or 2 byte commands",
+#endif
 	"Generate given amount of random data",
-	"Allow using software mechanisms (without CKF_HW)"
+	"Allow using software mechanisms (without CKF_HW)",
 };
 
 static const char *	app_name = "pkcs11-tool"; /* for utils.c */
@@ -380,6 +400,24 @@ static CK_FUNCTION_LIST_3_0_PTR p11 = NULL;
 static CK_SLOT_ID_PTR p11_slots = NULL;
 static CK_ULONG p11_num_slots = 0;
 static int suppress_warn = 0;
+static CK_C_INITIALIZE_ARGS_PTR  c_initialize_args_ptr = NULL;
+
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+static CK_C_INITIALIZE_ARGS  c_initialize_args_OS = {NULL, NULL, NULL, NULL, CKF_OS_LOCKING_OK, NULL};
+#ifdef _WIN32
+static HANDLE test_threads_handles[MAX_TEST_THREADS];
+#else
+static pthread_t test_threads_handles[MAX_TEST_THREADS];
+#endif
+struct test_threads_data {
+	int tnum;
+	char * tests;
+	volatile int state;
+	volatile CK_RV rv;
+};
+static struct test_threads_data test_threads_datas[MAX_TEST_THREADS];
+static int test_threads_num = 0;
+#endif /* defined(_WIN32) || defined(HAVE_PTHREAD) */
 
 struct flag_info {
 	CK_FLAGS	value;
@@ -480,6 +518,16 @@ static void		test_ec(CK_SLOT_ID slot, CK_SESSION_HANDLE session);
 #ifndef _WIN32
 static void		test_fork(void);
 #endif
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+static void		test_threads();
+static int		test_threads_start(int tnum);
+static int		test_threads_cleanup();
+#ifdef _WIN32
+static DWORD WINAPI	test_threads_run(_In_ LPVOID pttd);
+#else
+static void *		test_threads_run(void * pttd);
+#endif
+#endif /* defined(_WIN32) || defiend(HAVE_PTHREAD) */
 static void		generate_random(CK_SESSION_HANDLE session);
 static CK_RV		find_object_with_attributes(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *out,
 				CK_ATTRIBUTE *attrs, CK_ULONG attrsLen, CK_ULONG obj_index);
@@ -572,7 +620,6 @@ VARATTR_METHOD(EC_POINT, unsigned char);		/* getEC_POINT */
 VARATTR_METHOD(EC_PARAMS, unsigned char);		/* getEC_PARAMS */
 VARATTR_METHOD(ALLOWED_MECHANISMS, CK_MECHANISM_TYPE);	/* getALLOWED_MECHANISMS */
 
-
 int main(int argc, char * argv[])
 {
 	CK_SESSION_HANDLE session = CK_INVALID_HANDLE;
@@ -600,6 +647,9 @@ int main(int argc, char * argv[])
 	int do_test_ec = 0;
 #ifndef _WIN32
 	int do_test_fork = 0;
+#endif
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	int do_test_threads = 0;
 #endif
 	int need_session = 0;
 	int opt_login = 0;
@@ -929,6 +979,23 @@ int main(int argc, char * argv[])
 			action_count++;
 			break;
 #endif
+		case OPT_USE_LOCKING:
+			c_initialize_args_ptr = &c_initialize_args_OS;
+			break;
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+		case OPT_TEST_THREADS:
+			do_test_threads = 1;
+			if (test_threads_num < MAX_TEST_THREADS) {
+				test_threads_datas[test_threads_num].tnum = test_threads_num;
+				test_threads_datas[test_threads_num].tests = optarg;
+				test_threads_num++;
+			} else {
+				fprintf(stderr,"Too many --test_threads options limit is %d\n", MAX_TEST_THREADS);
+				util_print_usage_and_die(app_name, options, option_help, NULL);
+			}
+			action_count++;
+			break;
+#endif
 		case OPT_GENERATE_RANDOM:
 			need_session |= NEED_SESSION_RO;
 			opt_random_bytes = strtoul(optarg, NULL, 0);
@@ -994,10 +1061,23 @@ int main(int argc, char * argv[])
 	if (do_list_interfaces)
 		list_interfaces();
 
-	rv = p11->C_Initialize(NULL);
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	if (do_test_threads)
+		test_threads();
+#endif
+
+	rv = p11->C_Initialize(c_initialize_args_ptr);
+
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	if (do_test_threads || rv != CKR_OK)
+		fprintf(stderr,"Main C_Initialize(%s) rv:%s\n",
+				(c_initialize_args_ptr) ? "CKF_OS_LOCKING_OK" : "NULL",  CKR2Str(rv));
+#else
 	if (rv == CKR_CRYPTOKI_ALREADY_INITIALIZED)
 		fprintf(stderr, "\n*** Cryptoki library has already been initialized ***\n");
-	else if (rv != CKR_OK)
+#endif /* defined(_WIN32) || defined(HAVE_PTHREAD) */
+
+	if (rv != CKR_OK && rv != CKR_CRYPTOKI_ALREADY_INITIALIZED)
 		p11_fatal("C_Initialize", rv);
 
 #ifndef _WIN32
@@ -1241,6 +1321,11 @@ end:
 		if (rv != CKR_OK)
 			p11_fatal("C_CloseSession", rv);
 	}
+
+#if defined(_WIN32) || defined(HAVE_PTHREAD)
+	if (do_test_threads)
+		test_threads_cleanup();
+#endif /* defined(_WIN32) || defiend(HAVE_PTHREAD) */
 
 	if (p11)
 		p11->C_Finalize(NULL_PTR);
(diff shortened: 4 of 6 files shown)


dengert:
An example of running the modified pkcs11-tool
`./pkcs11-tool --test-threads --test-threads --test-threads --test-threads --login --test`
 **without** the pkcs11-global.c changes, the first 32 lines of the log show
4 threads have gotten past the simple `if (context == NULL)` test and overwritten the context, and are
all in the `pcsc_detect_readers` code. 

Timing may vary, from what is seen in FireFox but this should not happen.

```
 1 P:11002; T:0x140583396749760 09:35:17.837 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
   2 P:11002; T:0x140583396749760 09:35:17.838 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.20.0
   3 P:11002; T:0x140583396749760 09:35:17.838 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:892:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 rec     onnect_action=0 enable_pinpad=1 enable_pace=1
   4 P:11002; T:0x140583324649216 09:35:17.838 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
   5 P:11002; T:0x140583324649216 09:35:17.838 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.20.0
   6 P:11002; T:0x140583324649216 09:35:17.838 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:892:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 rec     onnect_action=0 enable_pinpad=1 enable_pace=1
   7 P:11002; T:0x140583307863808 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
   8 P:11002; T:0x140583307863808 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.20.0
   9 P:11002; T:0x140583307863808 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:892:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 rec     onnect_action=0 enable_pinpad=1 enable_pace=1
  10 P:11002; T:0x140583396749760 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
  11 P:11002; T:0x140583396749760 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
  12 P:11002; T:0x140583396749760 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
  13 P:11002; T:0x140583324649216 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
  14 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
  15 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.20.0
  16 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:892:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 rec     onnect_action=0 enable_pinpad=1 enable_pace=1
  17 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
  18 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
  19 P:11002; T:0x140583316256512 09:35:17.841 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
  20 P:11002; T:0x140583307863808 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
  21 P:11002; T:0x140583307863808 09:35:17.848 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
  22 P:11002; T:0x140583307863808 09:35:17.848 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
  23 P:11002; T:0x140583324649216 09:35:17.839 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
  24 P:11002; T:0x140583324649216 09:35:17.848 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
  25 P:11002; T:0x140583396749760 09:35:20.268 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1323:pcsc_add_reader: Adding new PC/SC reader 'Nitrokey Nitrokey Pro (0000000000000000000080B8) 00      00'
  26 P:11002; T:0x140583396749760 09:35:20.268 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:360:refresh_attributes: Nitrokey Nitrokey Pro (0000000000000000000080B8) 00 00 check
  27 P:11002; T:0x140583316256512 09:35:20.268 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1323:pcsc_add_reader: Adding new PC/SC reader 'Nitrokey Nitrokey Pro (0000000000000000000080B8) 00      00'
  28 P:11002; T:0x140583316256512 09:35:20.268 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:360:refresh_attributes: Nitrokey Nitrokey Pro (0000000000000000000080B8) 00 00 check
  29 P:11002; T:0x140583396749760 09:35:20.269 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:407:refresh_attributes: current  state: 0x00000022
  30 P:11002; T:0x140583307863808 09:35:20.269 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1323:pcsc_add_reader: Adding new PC/SC reader 'Nitrokey Nitrokey Pro (0000000000000000000080B8) 00      00'
  31 P:11002; T:0x140583396749760 09:35:20.269 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:408:refresh_attributes: previous state: 0x00000000
  32 P:11002; T:0x140583307863808 09:35:20.269 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:360:refresh_attributes: Nitrokey Nitrokey Pro (0000000000000000000080B8) 00 00 check
```

src/pkcs11/pkcs11-global.c
```diff
@@ -282,8 +296,12 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
 	in_finalize = 0;
```
Jakuje, line 296: The above block already changes global context by calling C_Finalize() or changing initialized_pid. Should not lock come even before this block? Rather safe than sorry.

Jakuje:
The CI is failing because you added a new switch to pkcs11-tool, but did not update manual page. Could you add the switch to manual page too?

dengert:
I will look at  the man pages.

dengert:
[threads-and-fork-think-twice-before-using-them](https://www.linuxprogrammingblog.com/threads-and-fork-think-twice-before-using-them) has a nice summary of issues with threads and fork.  [does-posix-guarantee-signals-will-not-be-delivered-to-a-partially-initialized-th](https://stackoverflow.com/questions/4227383/does-posix-guarantee-signals-will-not-be-delivered-to-a-partially-initialized-th) is another horror story. 

The use of `pthread_atfork` with `child` function would  be a much better place to do any cleanup of OpenSC, rather than waiting for the child process to call C_Initialize which may never happen.

OpenSC pkcs11 uses just a single mutex, in sc_pkcs11_lock using the locking parameters passed by the C_Initialize caller. Note that the caller does not pass a mutex, but only how to create, destroy, lock and unlock a mutex created by  C_Initialize, as well as the CKF_OS_LOCKING_OK flag. This indicates the caller will be using threads of some type. 

This PR was designed to address calling C_Initiatize from multiple threads at the same time. The one  opensc debug log I have seen showed this was happening, but there were multiple processes running and one was FireFox, but it is not possible to tell if  it was FireFox or it was some other process.  

I was able to test on Windows 10 the pkcs11-tool from this PR against a OpenSC 0.20.0.123 05 Feb 2020,
and FireFox only called C_Initialize once. The was using 2 tokens, but I don't have one that was failing.  Also tested the whole PR with FireFox 77.0.1 without problems.

Therefore, I am reluctant to commit this PR at this time as it does not appear to fix a real problem. It does show there is a potential problem, and also shows there may by better ways to handle cleanup after a fork if `pthread_atfork` was used with `static pthread_mutex_t c_initialize_m = PTHREAD_MUTEX_INITIALIZER;`.  On Windows a static `CRITICAL_SECTION` in the DLLmain could be used. Both of these do not require and create operation to be done before they can be used as a lock.

frankmorgner:
Concurrency is hard. Indeed I've thought about problems like this before after reading the source code. But I hesitated changing this stuff, exactly because there's a lot that could go wrong. Indeed, I'd like to wait with such a change until we have a test case that can provoke an error and a solution that works on all OSs.

The ideal solution would be to avoid global data entirely...

frankmorgner (changes requested):
As stated before, concurrency is hard. Before we change anything, we should have a test case which errors without the change, but which is fixed afterwards...

smuehlst:
> C_Initialize may be called by multiple threads.

I described in issue #2170 that this assumption is wrong. This would be an illegal use of the PKCS#11 API. For completeness I'm quoting the relevant sections from the PKCS#11 specification also here:

http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755803

> `C_Initialize` should be the first Cryptoki call made by an application, except for calls to `C_GetFunctionList`. ... Every call to `C_Initialize` should (eventually) be succeeded by a single call to `C_Finalize`.

http://docs.oasis-open.org/pkcs11/pkcs11-ug/v2.40/cn02/pkcs11-ug-v2.40-cn02.html#_Toc406759986

> To Cryptoki, an application consists of a single address space and all the threads of control running in it.  An application becomes a “Cryptoki application” by calling the Cryptoki function `C_Initialize` from one of its threads; after this call is made, the application can call other Cryptoki functions.  When the application has finished using Cryptoki, it calls the Cryptoki function `C_Finalize` and ceases to be a Cryptoki application.

dengert:
OK, I see your point. Once an application calls C_Initialize, it can call other functions, but not C_initialize until either C_Finalize is called or if the process was forked, the child process can call  C_initialize followed by  C_Finalize to clean up. 

But #2032 showed FireFox 74 did not follow this and called C_Initialize from multiple threads and OpenSC got confused.  

So this solution that this PR was trying to address is not to accept C_Initialize calls from multiple threads, but to catch the error early and return CKR_CRYPTOKI_ALREADY_INITIALIZED or CKR_FUNCTION_FAILED, if the first C_Initialize is still initializing.  

Or to  address the #2032 problem (which we have not seen since), lock until the first  C_Initialize finishes, then return CKR_CRYPTOKI_ALREADY_INITIALIZED or even CKR_OK.

smuehlst:
> But #2032 showed FireFox 74 did not follow this and called C_Initialize from multiple threads and OpenSC got confused.

This is clearly a bug in Firefox.

> Or to address the #2032 problem (which we have not seen since), lock until the first C_Initialize finishes, then return CKR_CRYPTOKI_ALREADY_INITIALIZED or even CKR_OK.

I wonder whether this can even be implemented in a manner that is conforming to the PKCS#11 specification. `C_Initialize` is the mechanism that instructs the PKCS#11 library how locking is to be performed. One scenario is that the application does not pass the CKF_OS_LOCKING_OK flag while providing function pointers for the locking functions. The specification says this:

http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755803

> A call to C_Initialize specifies one of four different ways to support multi-threaded access via the value of the CKF_OS_LOCKING_OK flag in the flags field and the values of the CreateMutex, DestroyMutex, LockMutex, and UnlockMutex function pointer fields:
...
> 3.     If the flag isn’t set, and the function pointer fields are supplied (i.e., they all have non-NULL_PTR values), that means that the application will be performing multi-threaded Cryptoki access, and the library needs to use the supplied function pointers for mutex-handling to ensure safe multi-threaded access.  If the library is unable to do this, C_Initialize should return with the value CKR_CANT_LOCK.

This would mean that `C_Initialize` must use the supplied function pointers to create a mutex and to lock the mutex, as it is not allowed to use the native operating system primitives for locking. This is a chicken-and-egg situation: Two parallel calls to `C_Initialize` would create two separate mutexes, which then wouldn't lock correctly as each thread would use its own mutex.

frankmorgner:
We don't know what the problem with Firefox is exactly. Could someone please get some clearification from the Firefox Team whether calling C_Initialize from two different threads is intended? @dengert , you currently have the most insights regarding the low level calls that lead to the problem. Could you open a bug report at Mozilla?

dengert:
@smuehlst  @frankmorgner  - I am going to change my opinion to:  Calling C_Initialize multiple times is **not** a bug.
I am going back to the original statement in this PR which says OpenSC does not handle it correctly with threads. 
 
PKCS11 specs clearly state that C_Initialize can return CKR_CRYPTOKI_ALREADY_INITIALIZED. So this allows an application to  initialize the library if it is not already initialized. It does not say anything about threads. 

Based on #2170  comments this would also imply that the same values for CreateMutex, DestroyMutex, LockMutex, UnlockMutex, flags, and pReserved must be provided. If not, I would suggest returning CKR_ARGUMENTS_BAD. 

PKCS11-V2.40 says C_Initialize  can have these Return values: CKR_ARGUMENTS_BAD, CKR_CANT_LOCK,
CKR_CRYPTOKI_ALREADY_INITIALIZED, CKR_FUNCTION_FAILED, CKR_GENERAL_ERROR,
CKR_HOST_MEMORY, CKR_NEED_TO_CREATE_THREADS, CKR_OK.

This PR may need revisions.

smuehlst:
> @smuehlst @frankmorgner - I am going to change my opinion to: Calling C_Initialize multiple times is **not** a bug.
> I am going back to the original statement in this PR which says OpenSC does not handle it correctly with threads.

You are right that calling `C_Initialize` multiple times per se is not a bug. But what is clearly not allowed is when multiple calls to `C_Initialize` from different threads  overlap concurrently. This follows from the following sentence from the PKCS#11 spec that was quoted above already:

>  An application becomes a “Cryptoki application” by calling the Cryptoki function C_Initialize from one of its threads; after this call is made, the application can call other Cryptoki functions.

dengert:
This is not the way I read it. It says C_Initialize must be called before calling other C_* functions. 

You say: "But what is clearly not allowed is when multiple calls to C_Initialize from different threads overlap concurrently."

Yes that is the problem, but it is not clear to everyone.  And we have to assume application developers could see it either way.  
I lean toward being robustness 
One could argue or assume C_Initialize will lock internally and use the values from CreateMutex, DestroyMutex, LockMutex, UnlockMutex or something else.  That is what this PR is trying to do, using:

static pthread_mutex_t c_initialize_m = PTHREAD_MUTEX_INITIALIZER;
or 
CRITICAL_SECTION c_initialize_cs = {0};
It uses these only within C_Initialize to protect from "multiple calls to C_Initialize from different threads overlap concurrently".

https://stackoverflow.com/questions/14320041/pthread-mutex-initializer-vs-pthread-mutex-init-mutex-param

 https://stackoverflow.com/questions/12853554/equivalent-of-pthread-mutex-initializer-on-windows

**On a lighter side**

 I lean toward being robust in what we accept. But if one thinks we can get Mozilla to change, we can send this IETF draft to them:
https://tools.ietf.org/html/draft-iab-protocol-maintenance-01
Which says: "The flaw in Postel's logic originates from the presumption of an inability to affect change in a system the size of the Internet."
(Later drafts do not have this line.)

One could change " Internet." to "Mozilla".
Note:  the author and who he works for...

frankmorgner (changes requested):
@dengert , I think you are right about the problem. Have you been able to reproduce a deadlock with pkcs11-tool and soft-HSM? Your test parameters don't look like they are able of running into a problem (with only 10 threads). Have you tried using 1000 or more threads?

Also, I suggest you use `module_init()`/`module_close()` for (de-)initializing the global lock (`C_INITIALIZE_M_LOCK`), because they are exactly once when loaded/unloaded on both, unix and windows. Note, that this change would also handle lpReserved correctly (which you are currently ignoring).

dengert:
From my cell phone. Can give longer response latter.

 On  windows with quad processor i have been able to show. Opensc 0.19 can
fail but PR build works as expected.

On linux with 2 processors running as virtual box guest with windows 10
host with 4 processors i can get  unmodified code to fail. And PR code to
work.

dengert:
Moved the InitializeCriticalSection(&c_initialize_cs); and InitializeCriticalSection(&c_initialize_cs);

On Windows it creates the CRITICAL_SECTION from DLLMain when attached by a process DLL_PROCESS_ATTACH and  when detached by process DLL_PROCESS_DETACH. So it is only initialized once. 

With pthreads  it uses pthread_mutex_t c_initialize_m = PTHREAD_MUTEX_INITIALIZER; so the lock is ready to use when module is loaded and nothing needs to be done when module is closed. 

The critical section/Mutex is not the PKCS11 global lock, it is designed to protect the code that creates the global lock,
and the OpenSC context creation.   This allows the application to still pass in its own mutex methods, pass in the CKF_OS_LOCKING_OK or pass in NULL to C_Initialize.  Even if NULL is passed, it still protects C_Initialize.  

To test if C_Initialize  can protect itself when not passing NULL:

First thread calls C_initialize with NULL,
Second thread pauses for 0 seconds C_initialize(NULL), C_GetInfo 
Then main  will call C_initialize as normal, i.e. NULL, and  an -L list info. 

Depending on how the OS schedules threads ,you can from the printf output if threads overlap or fail.  
Note all but first C_Initialize returns with CKR_CRYPTOKI_ALREADY_INITIALIZED
Test of this PR:
(Not sure about  GLib-GObject-CRITICAL **: 16:28:32.848: g_object_unref: assertion 'old_ref > 0' failed`
I have not seen this before. )

```
./opensc-tool -i
OpenSC 0.21.0 [gcc  9.3.0]
Enabled features: locking zlib openssl pcsc(/usr/lib/libpcsclite.so)
./pkcs11-tool --test-threads IN --test-threads P0:IN:GI -L
Test thread 0 started with options:IN
Test thread 0 C_Initialize(NULL)

(process:27491): GLib-GObject-CRITICAL **: 16:28:32.848: g_object_unref: assertion 'old_ref > 0' failed
Test thread 1 started with options:P0:IN:GI
Test thread 1 pauseing for 0 seconds
Test thread 1 C_Initialize(NULL)
Test thread Main C_Initialize(NULL) rv:CKR_CRYPTOKI_ALREADY_INITIALIZED
Test thread 1 C_Initialize returned CKR_CRYPTOKI_ALREADY_INITIALIZED
Test thread 1 C_GetInfo
Test thread 1 C_GetInfo returned CKR_OK
Test thread 1 returning rv = 0
Test thread 0 C_Initialize returned CKR_OK
Test thread 0 returning rv = 0
Available slots:
  token label        : Test Cardholder
  token manufacturer : piv_II
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : c9d45c86501843e2
  pin min/max        : 4/8
test_threads cleanup starting
test_threads all threads have ended with CKR_OK
test_threads #0 state:-1, rv:CKR_OK
test_threads #1 state:-1, rv:CKR_OK
test_threads cleanup finished
```

Using opensc-pkcs11.so from Ubuntu 20.4 it ends up segfaulting:
```
./pkcs11-tool --module /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so --test-threads IN --test-threads P0:IN:GI -L
Test thread 1 started with options:P0:IN:GI
Test thread 1 pauseing for 0 seconds
Test thread 1 C_Initialize(NULL)
Test thread 0 started with options:IN
Test thread 0 C_Initialize(NULL)
Test thread 0 C_Initialize returned CKR_OK
Test thread 0 returning rv = 0
Test thread 1 C_Initialize returned CKR_OK
Test thread 1 C_GetInfo
Test thread 1 C_GetInfo returned CKR_OK
Test thread 1 returning rv = 0
Test thread Main C_Initialize(NULL) rv:CKR_OK
Available slots:
  token label        : Test Cardholder
  token manufacturer : piv_II
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : c9d45c86501843e2
  pin min/max        : 4/8
test_threads cleanup starting
test_threads all threads have ended with CKR_OK
test_threads #0 state:-1, rv:CKR_OK
test_threads #1 state:-1, rv:CKR_OK
test_threads cleanup finished
```
Note how: sc_context_create is called by each C_Initialize as there is no locking.
```
    1 P:27710; T:0x139831899146304 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
    2 P:27710; T:0x139831899146304 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.21.0
    3 P:27710; T:0x139831880062720 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
    4 P:27710; T:0x139831899146304 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:885:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 trans      action_end_action=0 reconnect_action=0 enable_pinpad=1 enable_pace=1
    5 P:27710; T:0x139831880062720 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.21.0
    6 P:27710; T:0x139831880062720 16:57:24.162 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:885:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 trans      action_end_action=0 reconnect_action=0 enable_pinpad=1 enable_pace=1
    7 P:27710; T:0x139831899146304 16:57:24.163 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
    8 P:27710; T:0x139831899146304 16:57:24.163 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
    9 P:27710; T:0x139831899146304 16:57:24.163 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
   10 P:27710; T:0x139831880062720 16:57:24.163 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
   11 P:27710; T:0x139831880062720 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
   12 P:27710; T:0x139831880062720 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
   13 P:27710; T:0x139831871670016 16:57:24.164 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:855:sc_context_create: ===================================
   14 P:27710; T:0x139831871670016 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/ctx.c:856:sc_context_create: opensc version: 0.21.0
   15 P:27710; T:0x139831871670016 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:885:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 trans      action_end_action=0 reconnect_action=0 enable_pinpad=1 enable_pace=1
   16 P:27710; T:0x139831871670016 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1374:pcsc_detect_readers: called
   17 P:27710; T:0x139831871670016 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1387:pcsc_detect_readers: Probing PC/SC readers
   18 P:27710; T:0x139831871670016 16:57:24.165 [opensc-pkcs11] ../../../OpenSC/src/libopensc/reader-pcsc.c:1440:pcsc_detect_readers: Establish PC/SC context
```

The tests are a work in progress...

frankmorgner (commented):
There is still some problem with the includes. pkcs11-tool.c can't find pthread_create in the fuzzing environment. Apart from that this change looks good. Do you want to clean up the commits or shall we squash all commits with a merge?

src/pkcs11/pkcs11-global.c
dengert: I don't think so. 
The DeleteCriticalSection is only used on windows, and the module_close is called from DLLMain when the process is unloading the DLL.  the C_Finalize is called the application before the application unloads a module, as PKCS11 does not define how modules are loaded/linked. 

The code is designed to have a static mutex already setup when C_Initialize is called to protect the code that sets up the current locking for pkcs11, either: none, OS locking or what ever the calling application provides. This static mutex is is not the PKCS11 global_lock. (I am not even sure if the call has to be made when the DLL is being deleted.) 

The pthread version is  also in static storage, and initialized via PTHREAD_MUTEX_INITIALIZER which is done by the OS and pthread when linking or loading, so no extra code is needed.

dengert: OR are you saying  line 304  `C_INITIALIZE_M_LOCK` should come before line 300 `in_finalize = 0;` ?

Issue #2068, opened 2020-06-29
OpenSC build error in opensc-notify


misterzed88:
### Problem Description

OpenSC fails to build from the master branch, with error:
```
  CCLD     opensc-notify
/usr/bin/ld: opensc_notify-opensc-notify.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

The problem is that opensc_notify uses pthread functions without linking with the library.

The build was done with:
`gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)`

### Proposed Resolution
```
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -98,7 +98,7 @@ endif
 
 opensc_notify_SOURCES = opensc-notify.c opensc-notify-cmdline.c
 opensc_notify_LDADD = $(top_builddir)/src/libopensc/libopensc.la $(OPTIONAL_NOTIFY_LIBS)
-opensc_notify_CFLAGS = -I$(top_srcdir)/src $(OPTIONAL_NOTIFY_CFLAGS)
+opensc_notify_CFLAGS = -I$(top_srcdir)/src $(PTHREAD_CFLAGS) $(OPTIONAL_NOTIFY_CFLAGS)
 opensc_notify_CFLAGS += -Wno-unused-but-set-variable
 if HAVE_UNKNOWN_WARNING_OPTION
 opensc_notify_CFLAGS += -Wno-unknown-warning-option
```

Jakuje:
The proposed change looks good. Can you submit a PR? Wondering why it does not fail in CI? Is the notify support built in CI at all?

misterzed88:
PR submitted.

I have no why the error doesn't show up in a CI build. My manual build was done using plain bootstrap and configure (without any options). Is there anything in the env. setup of the CI that makes a difference, making it include the pthread lib by default?

Jakuje:
Fixed with #2069.

PR #2069, opened 2020-06-30
Fixes build error in opensc-notify (issue #2068).

misterzed88:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -98,7 +98,7 @@ endif
 
 opensc_notify_SOURCES = opensc-notify.c opensc-notify-cmdline.c
 opensc_notify_LDADD = $(top_builddir)/src/libopensc/libopensc.la $(OPTIONAL_NOTIFY_LIBS)
-opensc_notify_CFLAGS = -I$(top_srcdir)/src $(OPTIONAL_NOTIFY_CFLAGS)
+opensc_notify_CFLAGS = -I$(top_srcdir)/src $(PTHREAD_CFLAGS) $(OPTIONAL_NOTIFY_CFLAGS)
 opensc_notify_CFLAGS += -Wno-unused-but-set-variable
 if HAVE_UNKNOWN_WARNING_OPTION
 opensc_notify_CFLAGS += -Wno-unknown-warning-option


Jakuje:
Can you check/share your `config.log` before this change, where is the `-lpthread` mentioned and generated Makefile (in src/tools/) in your case?

For me (on Fedora), `-lpthread` comes from pkgconfig's `GIO2_CFLAGS`, which are already in the notify CFLAGS.

misterzed88:
On 7/1/20 3:07 PM, Jakub Jelen wrote:
> Can you check/share your |config.log| before this change, where is the 
> |-lpthread| mentioned and generated Makefile (in src/tools/) in your case?
> 
> For me (on Fedora), |-lpthread| comes from pkgconfig's |GIO2_CFLAGS|, 
> which are already in the notify CFLAGS.

My GIO2_CFLAGS from configure is empty:
GIO2_CFLAGS=''

Which is no surprise since my system does not have GIO installed. Quote 
from config.log:
configure:20258: $PKG_CONFIG --exists --print-errors "gio-2.0"
Package gio-2.0 was not found in the pkg-config search path.

I am running the following distro (lsb-release -da), installed with 
Qubes OS:
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.12 (stretch)
Release:	9.12
Codename:	stretch

So this explains why the CI build works while my build breaks. (It is 
really nice to go to the bottom of a problem to fully understand what's 
happening.)

Since gio-2.0 lib availibility is optional in the OpenSC build, notify 
should not rely on it to provide the pthread lib option (which would be 
an indirect dependency, anyway). So the correct solution should still be 
to add the direct dependency as proposed in the PR.

Or what do you think?

Best regards,
Zoltan

Jakuje:
Right. This is a good solution for now.

On the other hand, the opensc-notify is app designated only to show notification in gui and if there is no GUI library (GIO on Linux) available, it basically does not do anything. I would argue that id there was `--disable-notify` configure switch or the GUI dependencies were not found, the `opensc-notify` tool should not build at all, but not sure if it would be more or less confusing than the current situation. What do you think @frankmorgner ?

frankmorgner:
... interesting I've just checked the generated Makefile. It looks like AM_CFLAGS (which contains -lpthread) is not used once we specify `opensc_notify_CFLAGS`. This doesn't show up in CI, because opensc-notify is linked to libgio, which also requires pthread.

Note that in your case, where libgio is not found, opensc-notify gets built with your change, but it should never emit any notification...

frankmorgner:
Oh... and yes, disabling opensc-notify in this case, would be good from a user's perspective

Issue #1999, opened 2020-04-07
Firefox 74.0 Crash


Counter178:
Firefox crashes with OpenSC 0.20

Firstly I apologize if my bug report isn't very detailed in text-based but I'm forwarding feedback about Firefox Crash with OpenSC 0.20. When asking for more detailed answer the crash report was given:
1st feedback: https://bugzilla.mozilla.org/show_bug.cgi?id=1560486
2nd feedback:  https://crash-stats.mozilla.org/report/index/4b860515-e5f4-4fef-b440-29bff0200318#tab-details 

Question: In what circumstances does this error happen?

dengert:
This may have been solved by: https://github.com/OpenSC/OpenSC/issues/1935 and https://github.com/OpenSC/OpenSC/pull/1970

frankmorgner:
Dustin L. Howett [commented](https://bugzilla.mozilla.org/show_bug.cgi?id=1560486#c20):
```
0:000> k
 # Child-SP          RetAddr           Call Site
00 00000061`1ddfc090 00007fff`d996df9c opensc_pkcs11!_invoke_watson+0x17 [f:\dd\vctools\crt\crtw32\misc\invarg.c @ 132] 
01 00000061`1ddfc0c0 00007fff`d996dfb9 opensc_pkcs11!_invalid_parameter+0x64 [f:\dd\vctools\crt\crtw32\misc\invarg.c @ 85] 
02 00000061`1ddfc100 00007fff`d9972f65 opensc_pkcs11!_invalid_parameter_noinfo+0x19 [f:\dd\vctools\crt\crtw32\misc\invarg.c @ 97] 
03 00000061`1ddfc140 00007fff`d997ed23 opensc_pkcs11!strcpy_s+0x29 [f:\dd\vctools\crt\crtw32\h\tcscpy_s.inl @ 18] 
04 00000061`1ddfc170 00007fff`d983c31c opensc_pkcs11!_strdup+0x3f [f:\dd\vctools\crt\crtw32\string\strdup.c @ 80] 
05 00000061`1ddfc1b0 00007fff`d983c99d opensc_pkcs11!pcsc_add_reader+0x90 [c:\projects\opensc\src\libopensc\reader-pcsc.c @ 1274] 
06 00000061`1ddfc1e0 00007fff`d97aa395 opensc_pkcs11!pcsc_detect_readers+0x27d [c:\projects\opensc\src\libopensc\reader-pcsc.c @ 1403] 
[...]

PR #2077, opened 2020-07-08
Fix for context handle management in PCSC reader driver

misterzed88:
Proposed fix for issue #1999, taken a minimalistic change approach.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -1424,7 +1424,7 @@ static int pcsc_detect_readers(sc_context_t *ctx)
 
 			if ((rv == (LONG)SCARD_E_NO_SERVICE) || (rv == (LONG)SCARD_E_SERVICE_STOPPED)) {
 				gpriv->SCardReleaseContext(gpriv->pcsc_ctx);
-				gpriv->pcsc_ctx = 0;
+				gpriv->pcsc_ctx = -1;
 				gpriv->pcsc_wait_ctx = -1;
 				/* reconnecting below may may restart PC/SC service */
 				rv = SCARD_E_INVALID_HANDLE;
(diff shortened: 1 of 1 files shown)


frankmorgner:
Thanks a lot for tracking this down! It's good as is.

Using `-1` for an invalid SCARDCONTEXT (instead of `0`) was first introduced in e2375747 without justification. @LudovicRousseau do you know any good reason for doing so?

misterzed88:
So this bug has been lurking there since 2008, over a decade, before surfacing? Interesting!

I am not sure about the motivations for the original code but will try to make a qualified guess.

The original code used -1 since it relied on SCardListReaders to return SCARD_E_INVALID_HANDLE in order to call SCardEstablishContext. Since SCardListReaders accepts 0 as input in Windows, the author chose -1, assuming it would be rejected on all platforms.

This was bad coding practice for multiple reasons:
- Relying on an error code for important functionality.
- Using an arbitary value without API support (a PCSC implementation may chose to return -1 as a valid context).
- Assuming that SCardEstablishContext would leave the output parameter unchanged (as we now know, it was changed to 0, but any value is possible in theory).

The author updated the code in 4fc6c498edf91c03601d8f2dc640c14001144ba3, explicitly recognizing an uninitialized context instead of relying on an error from SCardListReaders. Now it would have been a good time to switch to 0 as the invalid context value, but the value -1 was kept.

LudovicRousseau:
0 is not less or more invalid than -1. On pcsc-lite the SCARDCONTEXT is a random number in [0, RAND_MAX] (see rand(3)).
I did not find a constant value that would define an invalid SCARDCONTEXT.

If you want to check a PC/SC context is valid you can/should use the API function [SCardIsValidContext()](https://pcsclite.apdu.fr/api/group__API.html#ga722eb66bcc44d391f700ff9065cc080b).

misterzed88:
You're right, my bad. (Being too quick when assuming that 0 was reserved in both Windows and pcsc-lite).

This means that the proposed solution must be updated to not assume anything about the returned handle. I will get back with an updated fix ASAP.

misterzed88:
A new fix was proposed, starting fresh, discarding the previous commits in this change branch.

The new fix is even more surgical - changing only the problem with the altered output parameter and adding the zero terminators as a safeguard. The -1 value is kept as the invalid handle value. Even though there are no guarantees from the API for this value being reserved, keeping it for now minimizes the risk for any unexpected side effects.

frankmorgner:
thanks!

PR #2079, opened 2020-07-21
ASN1 cleanup part 1

dengert:
ASN1 tags are represented in two many ways within OpenSC.
This is a trivial change to simplify one aspect of this.
It also makes the code more readable.

SC_ASN1_CLASS_MASK, SC_ASN1_APP, SC_ASN1_CTX, SC_ASN1_PRV,
SC_ASN1_CONS are changed, and SC_ASN1_CLASS_MASK is added.

These then align with the bits defined by SC_ASN1_TAG_CLASS,
SC_ASN1_TAG_APPLICATION, SC_ASN1_TAG_CONTEXT, SC_ASN1_TAG_PRIVATE,
and SC_ASN1_TAG_CONSTRUCTED.

(SC_ASN1_UNI and SC_ASN1_TAG_UNIVERSAL are both 0x00 thus no change
is needed).

(No sign of a right shift of SC_ASN1_CTX or SC_ASN1_PRV causeing
problems has been seen in the code.) If found, can be solved.)

Close examination of the OpenSC code base shows all uses of tags
used by routines and sc_asn1_entry use the defines.

This could allows 26 lines of code in sc_asn1_skip_tag used to test
the 3 CLASS and CONSTRUCTED bits to be replaced by:

	if (((cla << 24) | tag) != tag_in)
		return NULL;

The 26 lines still work as will any other code in OpenSC
that tests the bits using the defines. It also allows new code
to be simplified.

Problem identified while looking at better way to check response
on GET_DATA (0xCB) that returns TLV as used in card-piv.c

Changes tested using pkcs11-tool --test --login with PIV, SC_HSM
and OpenPGP cards.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/asn1.h
+++ b/src/libopensc/asn1.h
@@ -127,13 +127,16 @@ int sc_asn1_sig_value_sequence_to_rs(struct sc_context *ctx,
 		const unsigned char *in, size_t inlen,
                 unsigned char *buf, size_t buflen);
 
-#define SC_ASN1_CLASS_MASK		0x30000000
+/* long form tags use these */
+/* Same as  SC_ASN1_TAG_* shifted left by 24 bits  */
+#define SC_ASN1_CLASS_MASK		0xC0000000
 #define SC_ASN1_UNI			0x00000000 /* Universal */
-#define SC_ASN1_APP			0x10000000 /* Application */
-#define SC_ASN1_CTX			0x20000000 /* Context */
-#define SC_ASN1_PRV			0x30000000 /* Private */
-#define SC_ASN1_CONS			0x01000000
+#define SC_ASN1_APP			0x40000000 /* Application */
+#define SC_ASN1_CTX			0x80000000 /* Context */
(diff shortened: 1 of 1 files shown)


dengert:
@frank I recall you recently said something about cleaning up the ASN.1 routines. This PR could be a start.

other simple improvement could include:

- in asn1.h prototypes for routines which use "tag" to replace with: "tag_long", "tag_short", or "tag_only"
to  indicate  CLASS+CONSTRUCTED in high order byte with the rest of tag in lower bytes,  CLASS+CONSTRUCTED is with first byte of rest of tag in lower order bytes, or  CLASS+CONSTRUCTED removed  in tag bytes.  This makes it a little more readable to. 

- With this PR, 4 byte tags can fit in "tag_long".  (But I have never even seen a 3 byte tag yet.)  

- Add conversion routines between "tag_long" and "tag_short"

- look closer at the  "sc_asn1_verify_tag" Its only used once.

frankmorgner (approved):

PR #2080, opened 2020-07-23
Card API cleanup

misterzed88:
Improve card API design without breaking anything and without changing any functionality.

The card API (especially sc_pin_cmd_data) has become cluttered with fields with unclear purposes, or even with fields which are used only privately in some card drivers. Some fields have also become obsoleted by recent changes in master.

The existence of these fields makes it more complex than necessary to implement new card drivers, since it is not clear how the fields should be handled. The excess fields also make the API harder to use, since it is not obvious which of the fields can be relied upon to be available or not.

In short: this pull request is an attempt at cleaning up, removing unused- and private fields from the public interface and improving public comments.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-authentic.c
+++ b/src/libopensc/card-authentic.c
@@ -93,7 +93,7 @@ unsigned char aid_AuthentIC_3_2[] = {
 static int authentic_select_file(struct sc_card *card, const struct sc_path *path, struct sc_file **file_out);
 static int authentic_process_fci(struct sc_card *card, struct sc_file *file, const unsigned char *buf, size_t buflen);
 static int authentic_get_serialnr(struct sc_card *card, struct sc_serial_number *serial);
-static int authentic_pin_get_policy (struct sc_card *card, struct sc_pin_cmd_data *data);
+static int authentic_pin_get_policy (struct sc_card *card, struct sc_pin_cmd_data *data, struct sc_acl_entry *acls);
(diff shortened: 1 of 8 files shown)


Jakuje:
I have no idea how the OSX build passed for this PR as it looks like `length_offset` is still used in the `src/libopensc/reader-cryptotokenkit.m` and it is failing now:

https://travis-ci.org/github/OpenSC/OpenSC/jobs/720459495

I am not sure what it is used there for as I am not familiar with the osx code. Can you check if we should revert 48d939b0 or there is a way to get it working on OSX too?

misterzed88:
Oops, I was not aware of the OSX Cryptotokenkit based reader module and it was therefore not kept in synch when making the PIN-pad changes in reader-pcsc.c. The PR test build does not complain since it builds without Cryptotokenkit support (and does therefore not build this module):
```CryptoTokenKit support:  no```

I created a fix in PR #2099.  This is the preferred way of fixing it (instead of reverting commit 48d939b05719e0e47af7891bf438c09647929f64). In this quick timeframe I have not had any chance to build and test on OSX.

In about 2 weeks timeframe I should be able to do a real test run on OSX (which would also include testing with PIN-pad readers).

Jakuje:
Thank you for investigation and for a quick fix. As I do not know much about the OSX, I already asked Frank for review.

Issue #2081, opened 2020-07-26
Commit c4d7bb1 does not do what it claims to do


marschap:
Hi,

commit https://github.com/OpenSC/OpenSC/commit/c4d7bb1a7ba80d155021f7a311b7fa6b1ce54e15 does not do what it claims to do.

It embraces a `LOG_TEST_RET()` with an `if` clause that triggers on a condition which `LOG_TEST_RET()` ignores:
* `LOG_TEST_RET()` only exits a function if its parameter `r` is lower than zero.
* trying to call it with the OR-ed condition `obj.data == NULL` will not cause `LOG_TEST_RET()` to do anything as long as `r >= 0`.

### Proposed Resolution

If the function should be exited at that point based on the condition `obj.data == NULL`, the correct way to do it, is to set `r` to a negative value **before** the call to `LOG_TEST_RET()`, e.g. by 
`if (obj.date == NULL)
      r = SC_ERROR_OBJECT_NOT_FOUND;`

Jakuje:
Thank you for the review. Your suggestion sounds good to me. Please, submit a PR.

Jakuje:
oh, I see it is already in #2082. Sorry for the noise.

Jakuje:
Fixed by #2082.

PR #2082, opened 2020-07-26
Fix for c4d7bb1 (#2081)

marschap:
fix issue 2081,
and fix a typo in a log message I detected while I was at it.

### DIFF
--- a/src/libopensc/pkcs15-sc-hsm.c
+++ b/src/libopensc/pkcs15-sc-hsm.c
@@ -788,7 +788,7 @@ static int sc_pkcs15emu_sc_hsm_add_cd(sc_pkcs15_card_t * p15card, u8 id) {
 	/* Try to select a related EF containing the PKCS#15 description of the data */
 	len = sizeof efbin;
 	r = read_file(p15card, fid, efbin, &len, 1);
-	LOG_TEST_RET(card->ctx, r, "Skipping optional EF.DCOD");
+	LOG_TEST_RET(card->ctx, r, "Skipping optional EF.CDF");
 
 	ptr = efbin;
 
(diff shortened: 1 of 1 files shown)


src/libopensc/pkcs15-sc-hsm.c
```diff
@@ -788,15 +788,15 @@ static int sc_pkcs15emu_sc_hsm_add_cd(sc_pkcs15_card_t * p15card, u8 id) {
...
+	LOG_TEST_RET(card->ctx, r, "Skipping optional EF.CDF");
 
 	ptr = efbin;
 
 	memset(&obj, 0, sizeof(obj));
 	r = sc_pkcs15_decode_cdf_entry(p15card, &obj, &ptr, &len);
-	if (obj.data == NULL || r != SC_SUCCESS) {
-		LOG_TEST_RET(card->ctx, r, "Skipping optional EF.CDOD");
-	}
+	if (obj.data == NULL && r >= SC_SUCCESS)
```
Jakuje, line 797: I think this will not work for you -- the src/libopensc/pkcs15-cert.c:509 returns 0 (`SC_SUCCESS`), if it is ignoring the certificate, keeping the `obj->data` empty and therefore you will not avoid the crash in the original issue.

marschap: @Jakuje Hi Jakub, 
please have another look.

In the original code (i.e. before https://github.com/OpenSC/OpenSC/commit/c4d7bb1a7ba80d155021f7a311b7fa6b1ce54e15 ), all errors, as indicated by `r < 0`, were caught by the `LOG_TEST_RET()`.
But the case of  'obj.data == NULL` despite `r >= 0` was not caught.

Please note: `SC_SUCCESS`is defined as `0`, so if `sc_pkcs15_decode_cdf_entry()` returns SC_SUCCESS in the non-error case, this is covered by the check `r >= 0` )

https://github.com/OpenSC/OpenSC/commit/c4d7bb1a7ba80d155021f7a311b7fa6b1ce54e15 tried to address it, but failed, because 'LOG_TEST_RET()' only looked at 'r', completely ignoring 'obj.data == NULL'.

This patch now addresses the missing case of obj.data == NULL' respite `r >= 0` and changes `r' (sic!) to an error value in this case, which then allows `LOG_TEST_RET()` to catch this error case as well.

So in essence, this patch is closing the original hole.

Please be so kind to approve or merge this change if you agree
Peter

Jakuje: Thanks for explanation. You are indeed right. I approved the PR now and if there will be no more concerns, I will merge it tomorrow.

PR #2090, opened 2020-08-09
Explorer records

marschap:
During analysis of an unknown card, I found it helpful to have opensc-explorer's commands `asn1`  and `cat` to be slightly extended.

Using the commits in this PR
- `cat` can be limited to only show one given record in a record oriented EF
- `asn1` can parse individual records of record-oriented EFs as well as parsing the data from a given offset, instead of the default start of the EF / record.

Please approve and merge
Peter

### DIFF
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -206,7 +206,7 @@ static struct command	cmds[] = {
 		"apdu",	"<data> ...",
 		"send a custom apdu command"		},
 	{ do_asn1,
-		"asn1",	"[<file-id> [<rec-no>]]",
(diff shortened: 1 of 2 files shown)


src/tools/opensc-explorer.c
```diff
@@ -53,6 +53,11 @@
 
 #define DIM(v) (sizeof(v)/sizeof((v)[0]))
 
+#ifndef MIN
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
+#endif
```
Jakuje, line 58: This is already defined in `src/libopensc/internal.h`. Would it work to use the definition from there, rather than redefining it all over again in each tool? Other tools already include `internal.h`.

Jakuje:
Thank you.

PR #2091, opened 2020-08-12
Fix recent oss-fuzz and coverity issues

Jakuje:
##### Checklist
- [ ] macOS tokend is tested


Jakuje:
I added few more fixes and it looks like I got to the bottom of the list again. Review/comments welcomed.

Issue #1860, opened 2019-11-21
PKCS #11 3.0 in OpenSC


Jakuje:
In regards of #1854, I would like to bump version of PKCS #11 library to 3.0 and implement the new required interfaces in coming months.

The public review drafts are already available on OASIS website and I assume there will be no significant changes before the final version:

https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/pkcs11-base-v3.0.html

So far, these are my thoughts about advantages of 3.0:
 * Support for new key types (ed25519 -- see #1854)
 * It is backward compatible -- new functions are available through `C_GetInterfaceList` and `C_GetInterface`
 * Profile objects can resolve the issues with NSS and its handling of OpenSC as non-friendly by announcing the `CKP_PUBLIC_CERTIFICATES_TOKEN` profile (from https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.0/pkcs11-profiles-v3.0.html )
 * SHA3 support
 
I will probably go through the draft and try to create issues for particular tasks if it will be easy to separate them so anyone can help. I would probably like to create it as a project, but it looks like I do not have appropriate permissions to do that.

Any thoughts?

frankmorgner:
Although we're currently often using 2.4 as reference for PKCS#11, we're announcing compatibility with 2.20 only:
https://github.com/OpenSC/OpenSC/blob/ee1c8073c2ecd0d7644d64ef3f2a2be4b5752c6e/src/pkcs11/pkcs11-global.c#L408-L409

I'm not sure what's the implication in terms of backward compatibility...

Jakuje:
I think this is mostly done now. We might need some implementation of the function stubs, we might bring more algorithms as needed, but the base is in place.

Issue #1854, opened 2019-11-07
Add curve25519 support in OpenSC


alex-nitrokey:
Currently, there is no support for curve25519 in OpenSC. There is support for the curves shown [here](https://github.com/CardContact/OpenSC/blob/19711d0a1301f0519a1348390bb996fd2925f766/src/tools/pkcs11-tool.c#L85) and [here](https://github.com/OpenSC/OpenSC/blob/a8de0070fd351b56895191653bf35888d0d3ab39/src/libopensc/pkcs15-pubkey.c#L1482) only (as far as I can see). It is possible that there are just not many cards out there that support this curve, but there are people looking out for this curve. So this is a newer version of an [old request](https://github.com/OpenSC/OpenSC/issues/924).

I can not oversee the work that needs to be done to make curve25519 work, therefore I created this issue. I hope for discussion and help. If anyone wants to implement this, I would be happy. Nitrokey can provide free Nitrokey Start devices (equivalent to Gnuk devices, which support curve25519) to test code. The openpgp card driver used by Gnuk [is already prepared](https://github.com/OpenSC/OpenSC/issues/1715#issuecomment-551170110).

ya-isakov:
Yubikey with firmware 5.2.3+ is also supporting Ed25519/Curve25519, and these kind of curves are more and more popular.

Jakuje:
The problem with Ed25519/Curve25519 is that it is not supported in the current version of PKCS #11 specification without vendor specific extension (see the discussion in softhsm PR: https://github.com/opendnssec/SoftHSMv2/pull/324 ). The IDs for these should come in PKCS#11 3.0, but it is not out yet (even though used in SoftHSM: https://github.com/opendnssec/SoftHSMv2/blob/develop/src/lib/pkcs11/pkcs11.h#L889 )

I think I have old yubikey that does not support these keys, but it would be interesting to implement support for these (at least keys -- not sure about the use of the pure curve25519).

jans23:
@Jakuje If it helps, we could ship you a free Nitrokey Start.

Jakuje:
@jans23 That would be nice. Please, drop me an email. Unfortunately, I can not estimate when I will be able to work on these, but I would be certainly interested in improving OpenSC in this way.

Jakuje:
for the record, this is already WIP, but I still do not have anything that can be published as it is touching a lot of things all over the code base, but I hope in coming weeks I will be able to open a PR.

PR #2096, opened 2020-08-19
More PKCS #11 3.0 features

Jakuje:
As announced in #1860 there are bunch of new features in PKCS #11 3.0 which are needed in OpenSC to support the new version. The EdDSA keys are already implemented in #1854.

This pull request is about the rest of the PKCS #11 3.0 features outlined in the previously mentioned issue, allowing the OpenSC to work with applications talking PKCS #11 3.0 and allowing the pkcs11-tool to talk with new PKCS #11 3.0 modules.

There are also implemented new profile objects (`CKP_PUBLIC_CERTIFICATES_TOKEN`), which should prevent issues with NSS/Firefox and unneeded logins if the certificates are publicly accessible.

The last but not least function coming in PKCS #11 3.0 is that library can support and application can request fork-safe interface. I am not sure how opensc is fork-safe (@frankmorgner ?) but if it is, we can announce it in the interfaces interface (not done yet).

References:
https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html
https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html
https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.0/pkcs11-profiles-v3.0.html
https://github.com/oasis-tcs/pkcs11/tree/master/working/3-00-current

##### Checklist

- [ ] macOS tokend is tested


src/tools/pkcs11-tool.c
```diff
@@ -1251,6 +1264,41 @@ static void show_cryptoki_info(void)
...
+	}
+
+	rv = p11->C_GetInterfaceList(NULL, &count);
+	if (rv != CKR_OK) {
+		p11_fatal("C_GetInterfaceList(size inquire)", rv);
+	}
+
+	interfaces = malloc(count * sizeof(CK_INTERFACE));
+	if (interfaces == NULL) {
+			perror("calloc failed");
```
simo5, line 1287: perror  says calloc, but you used malloc

src/pkcs11/pkcs11f.h
```diff
@@ -0,0 +1,85 @@
...
+   This file is free software; as a special exception the author gives
+   unlimited permission to copy and/or distribute it, with or without
+   modifications, as long as this notice is preserved.
+
+   This file is distributed in the hope that it will be useful, but
+   WITHOUT ANY WARRANTY, to the extent permitted by law; without even
+   the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+   PURPOSE.  */
+
+/* This is a PKCS #11 2.x function list for inclusion in both legacy and
```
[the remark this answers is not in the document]
simo5: Sounds good to me

simo5:
I have no other comments, LGTM

Jakuje:
> I have no other comments, LGTM

Thank you.

frankmorgner:
Quoting from v3 regarding `CKF_INTERFACE_FORK_SAFE`:

>The returned interface will have fork tolerant semantics. When the application forks, each process will get its own copy of all session objects, session states, login states, and encryption states. Each process will also maintain access to token objects with their previously supplied handles.

Regarding the duplication of objects and other states, we are compliant since we are not using shared memory or similar. With the duplication of the process' memory, `fork` itself will duplicate every handle or state that OpenSC is using.

However, we need to review our locking process to not run into a deadlock (`sc_lock`/`sc_pkcs11_lock`). pkcs11-tool has some test for forking, so some work has already been done in this regard.

frankmorgner (changes requested):
This looks good.

Do you know if there's a diff to the previous version(s) the API calls (e.g. clarification or changes in their behavior)?

src/pkcs11/framework-pkcs15.c
```diff
@@ -811,6 +818,32 @@ __pkcs15_create_data_object(struct pkcs15_fw_data *fw_data,
...
+	struct sc_pkcs15_object *obj = NULL;
+	int rv;
+
+	obj = calloc(1, sizeof(struct sc_pkcs15_object));
+
+	rv = __pkcs15_create_object(fw_data, (struct pkcs15_any_object **) &pobj,
+			obj, &pkcs15_profile_ops, sizeof(struct pkcs15_profile_object));
+	if (rv >= 0) {
+		pobj->profile_id = public_certificates ? CKP_PUBLIC_CERTIFICATES_TOKEN : CKP_AUTHENTICATION_TOKEN;
+	}
```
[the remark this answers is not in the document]
Jakuje: Right. Fixed and reformatted a bit.

Jakuje:
> Quoting from v3 regarding `CKF_INTERFACE_FORK_SAFE`:
> 
> > The returned interface will have fork tolerant semantics. When the application forks, each process will get its own copy of all session objects, session states, login states, and encryption states. Each process will also maintain access to token objects with their previously supplied handles.
> 
> Regarding the duplication of objects and other states, we are compliant since we are not using shared memory or similar. With the duplication of the process' memory, `fork` itself will duplicate every handle or state that OpenSC is using.
> 
> However, we need to review our locking process to not run into a deadlock (`sc_lock`/`sc_pkcs11_lock`). pkcs11-tool has some test for forking, so some work has already been done in this regard.

Thanks for checking. I know that we discussed this before, but I was not sure about the exact state so I rather kept this out. Once we will be ready it is easy to add another interface with the above flag here:

https://github.com/OpenSC/OpenSC/pull/2096/files#diff-aa05497d497dd4fa0c8ffa689f3810a95ff40a973e3b85e4369c370e3459ff18R787-R790

Jakuje:
> Do you know if there's a diff to the previous version(s) the API calls (e.g. clarification or changes in their behavior)?

The 3.0 is backward compatible with the 2.4 -- it still presents the getFunctionList which returns 2.x version API. The new api getInterfacesList can return both of them but as visible from the changest, the new functions are just appended in the end of the 2.x list, keeping it also backward compatible. I am not aware of any change in behavior of any functions.

src/pkcs11/framework-pkcs15.c
```diff
@@ -1446,8 +1487,17 @@ _add_public_objects(struct sc_pkcs11_slot *slot, struct pkcs15_fw_data *fw_data)
...
-		if (obj->p15_object->flags & SC_PKCS15_CO_FLAG_PRIVATE)
+		if (obj->p15_object->flags & SC_PKCS15_CO_FLAG_PRIVATE) {
+			/* If we found some non-accessible public object,
+			 * we can no longer claim Public Ceritificate Token conformance */
+			if ((obj->p15_object->type >= SC_PKCS15_TYPE_PUBKEY &&
+				obj->p15_object->type <= SC_PKCS15_TYPE_PUBKEY_EC) ||
+				(obj->p15_object->type >= SC_PKCS15_TYPE_CERT &&
+				obj->p15_object->type <= SC_PKCS15_TYPE_CERT_SPKI)) {
+				public_certificates = 0;
+			}
```
frankmorgner, line 1498: I think it would be better to have an explicit check for all non-public objects here rather than relying on the specific order of OpenSC flags. You could, for example, use a switch statement with "fallthrough" cases

frankmorgner (approved):

frankmorgner:
An include guard for pkcs11f.h would be nice (would also solve the lgtm complaint)

Jakuje:
> An include guard for pkcs11f.h would be nice (would also solve the lgtm complaint)

The problem is that it is included twice so not including it is intentional.

src/pkcs11/framework-pkcs15.c
Jakuje: Having better look, the `SC_PKCS15_TYPE_PUBKEY` and `SC_PKCS15_TYPE_CERT` are bitmasks of all the following types so the checking against the bitmask should be enough.

frankmorgner:
> > An include guard for pkcs11f.h would be nice (would also solve the lgtm complaint)
> 
> The problem is that it is included twice so not including it is intentional.

I agree with @simo5 that this trick is not so nice. Although the official PKCS#11 files also contain it, I think we should not use it here. We already have a different version of `pkcs11.h`, because of different coding guidelines. Here, simply copying the the lines of pkcs11f.h into pkcs11.h twice would be a simple and better choice, in my opinion.

Jakuje:
> > > An include guard for pkcs11f.h would be nice (would also solve the lgtm complaint)
> > 
> > 
> > The problem is that it is included twice so not including it is intentional.
> 
> I agree with @simo5 that this trick is not so nice. Although the official PKCS#11 files also contain it, I think we should not use it here. We already have a different version of `pkcs11.h`, because of different coding guidelines. Here, simply copying the the lines of pkcs11f.h into pkcs11.h twice would be a simple and better choice, in my opinion.

OK. I dropped that part.

frankmorgner (approved):

Issue #767, opened 2016-05-27
epass 2003 - Failed to erase card


jursonovicst:
### Expected behaviour

What should happen?

epass 2003 should be erased
### Actual behaviour

What happens instead?

`pkcs15-init -E` gives the following error:

```
root@ubuntu:~# pkcs15-init -E
Using reader with a card: Feitian ePass2003 00 00
Failed to erase card: Security status not satisfied
```
### Steps to reproduce
1. connect epass 2003
2. pkcs15-init -E
### Logs

[pkcs15-init -E -vvv logs](https://gist.github.com/jursonovicst/482ae5e628cc12db7db95b2daf94f2e9.js)

vDorst:
Please contact @FeitianSmartcardReader, they can provide you a tool to fix that problem.
Still on my todo list to update the Wiki.

FeitianSmartcardReader:
@jursonovicst Please send mail to Leon@ftsafe.com, he will give you a tool, thanks

jursonovicst:
Thank you...

aurabindo:
@FeitianSmartcardReader  it would be convenient if you could post the tool here. I also have the same problem, and that you send me the tool. I've sent you a mail already. kindly respond to that.

alistairmackenzie:
Also having this issue with two different devices. @FeitianSmartcardReader dropped the address above an email.

FeitianSmartcardReader:
did you send mail to us? We didn't receive your mail, please send mail to us, thanks
leon@ftsafe.com and copy hongbin@ftsafe.com

alistairmackenzie:
@FeitianSmartcardReader Have sent again to both addresses. Should be coming from alistair.mackenzie@v2.pw

FeitianSmartcardReader:
@alistairmackenzie We finish linux tool, check and download from below:
[Fix_tool](http://download.ftsafe.com/files/ePass/Fix_Tool.tar.gz)

persmule:
Source code for Fix_tool is needed, otherwise we cannot know whether the Fix_tool do something other than fixing the token.

zweipunktnull:
Hey ;)
I am trying to run: 
sudo pkcs15-init --store-certificate cs.pem --auth-id 01 --id 01 --format pem -vvv
but it gives me the following output...


00 20 00 F4 08 31 32 33 34 35 36 37 38 . ...12345678
[...]

PR #2097, opened 2020-08-19
epass2003: Fix erase sequence.

riastradh:
The previous erase sequence did not always work.  For example:

```
% pkcs15-init -C
Using reader with a card: Feitian ePass2003 00 00
New User PIN.
Please enter User PIN: 1234
Please type again to verify: 1234
Unblock Code for New User PIN (Optional - press return for no PIN).
Please enter User unblocking PIN (PUK):
Failed to create PKCS #15 meta structure: Security status not satisfied

% pkcs15-init -E
Using reader with a card: Feitian ePass2003 00 00
Failed to erase card: Security status not satisfied
```

This apparently bricked many people's ePass2003 devices:

https://github.com/OpenSC/OpenSC/issues/767
https://sourceforge.net/p/opensc/mailman/message/33621883/
https://github.com/OpenSC/OpenSC/wiki/Feitian-ePass2003

Feitian provided a proprietary binary blob called `FIX_TOOL` to recover devices from this state, but declined to offer source code when asked:

https://download.ftsafe.com/files/ePass/Fix_Tool.tar.gz
https://download.ftsafe.com/files/reader/SDK/Fix_Tool_20200604.zip

With reverse-engineering help by Saleem Rashid (@saleemrashid on Github), I was able to find the sequence of three APDUs that the tool submits to the device to erase it.  The mechanism seems to be:

1. Install a magic PIN.  This is like install_secret_key, as used by internal_install_pin, but with a few different magic constants.

2. Verify the magic PIN.

3. Delete the MF file, without selecting anything first.

With this patch, `pkcs15-init -E` successfully erases my ePass2003, and I am able to initialize it with `pkcs15-init -C -p pkcs15+onepin` if I set both a user pin and a PUK.  (This patch does not prevent the ePass2003 from getting into the state which could not be erased by the old erase sequence.)

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -2518,12 +2518,48 @@ epass2003_gen_key(struct sc_card *card, sc_epass2003_gen_key_data * data)
 static int
 epass2003_erase_card(struct sc_card *card)
 {
+	static const unsigned char install_magic_pin[26] = {
+		/* compare install_secret_key */
+		0x06,0x01,0x10,0x16, 0x16,0x16,0x00,0x0f, 0xff,0x66,
+		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
+		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
+	};
+	static const unsigned char magic_pin[16] = "1234567812345678";
+	static const unsigned char mf_path[2] = { 0x3f, 0x00 };
+	sc_apdu_t apdu;
 	int r;
 
 	LOG_FUNC_CALLED(card->ctx);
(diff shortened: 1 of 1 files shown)


Jakuje:
I don't know much about these cards, but maybe @FeitianSmartcardReader can review if it is safe to do or if there is something else needed.

frankmorgner (approved):
I wonder what else could be done with "the magic PIN"...

@FeitianSmartcardReader could you comment?

FeitianSmartcardReader:
@frankmorgner after talk internal, the answer is yes, we agree merge this pull request, thanks

frankmorgner:
great, thanks for the feedback!

PR #2103, opened 2020-08-30
Fixes #1399, 3 certs support OpenPGP (pksc15-tool)

### LINKED ISSUE #1399: OpenPGP v3.3: Support multiple certificates
With v3.3 of the OpenPGP spec, there are now 3 slots for certificates on the card, matching the three private keys you can store. It would be great if opensc could be updated to allow these to be read and loaded onto the card.
GnuPG does not have support for this so far as far as I'm aware.

daringer:
Tested with:

```bash
$ opensc-tool -n
Using reader with a card: Nitrokey Nitrokey Pro (000000000000000000008161) 00 00
OpenPGP card v3.3 (0005 00008161)
```

##### Checklist

My impression is that the other checklist items do not apply, please correct me if I am wrong.


src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -419,29 +419,38 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
 		}
+
+
 		/* Just update the certificate DO */
-		sc_format_path("7F21", path);
+		u8 param = (u8) (2 - (cid->value[0] - 1));
+		sc_card_ctl(card, SC_CARDCTL_OPENPGP_SELECT_DATA, &param);
+
+        sc_format_path("7F21", path);
```
Jakuje, line 442: indentation

Jakuje (changes requested):
generally looks good, but there are some cosmetic issues that should be fixed before merging this

daringer:
thx for the feedback, 
all changes understood and will fix them shortly!

src/libopensc/card-openpgp.c
```diff
@@ -129,6 +129,8 @@ static void		pgp_free_blob(pgp_blob_t *);
 static int		pgp_get_pubkey(sc_card_t *, unsigned int, u8 *, size_t);
 static int		pgp_get_pubkey_pem(sc_card_t *, unsigned int, u8 *, size_t);
 
+//static int      pgp_get_data_next(sc_card_t *, unsigned int, u8 *, size_t );
+
```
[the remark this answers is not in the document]
daringer: nope, slipped through during cleaned-up, removing....

src/libopensc/card-openpgp.c
```diff
@@ -3273,13 +3314,15 @@ pgp_card_ctl(sc_card_t *card, unsigned long cmd, void *ptr)
 		memmove((sc_serial_number_t *) ptr, &card->serialnr, sizeof(card->serialnr));
 		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 		break;
-
+	case SC_CARDCTL_OPENPGP_SELECT_DATA:
+		pgp_select_data(card, *((u8 *) ptr));
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
```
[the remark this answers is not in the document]
daringer: yap, 
* assigning return value to `r` now and returning it
* followed the call into `pgp_select_data` and found the same fail, thus fixed it there, too
* the `pgp_select_data()` return value is now the return value of the underlying `sc_transmit_apdu()` call

src/libopensc/pkcs15-openpgp.c
```diff
@@ -353,42 +363,69 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
 		}
 	}
 
-	/* Check if certificate DO 7F21 holds data */
-	sc_format_path("7F21", &path);
+
+    const u8 NUM_CERTS = 3;
```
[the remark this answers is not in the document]
daringer: caps was there to remind me to move it to a `#define`,
hasn't helped obviously, but is done now... 
```#define PGP_MAX_NUM_CERTS 3```

File: src/libopensc/pkcs15-openpgp.c [hunk 5]
```diff
@@ -353,42 +363,69 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+    const size_t buf_len = MAX_OPENPGP_DO_SIZE;
+    const cdata certs[] = {
+		{"AUT certificate", 0, "3F007F21", "3", SC_PKCS15_CO_FLAG_MODIFIABLE},
+		{"DEC certificate", 0, "3F007F21", "2", SC_PKCS15_CO_FLAG_MODIFIABLE},
+		{"SIG certificate", 0, "3F007F21", "1", SC_PKCS15_CO_FLAG_MODIFIABLE},
+		{NULL, 0, NULL, NULL, 0}
+	};
+
+    sc_path_t path;
```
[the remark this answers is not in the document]
daringer: mmmh, maybe this is my silent protest against this rule from the 80s :smile:
but no discussion, confirmed, changes:
* move `certs` into global scope and make it `static` similar to other constructs
* move `path` all the way up to the beginning of the scope
* moved all declarations inside the following `for`-loop to the start of its scope

src/libopensc/pkcs15-openpgp.c
```diff
@@ -353,42 +363,69 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+    sc_format_path("7F21", &path);
 	r = sc_select_file(card, &path, &file);
 	if (r < 0)
-		goto failed;
+    	goto failed;
```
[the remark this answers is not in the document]
daringer: check

src/pkcs15init/pkcs15-openpgp.c
daringer: check

daringer:
all change requests resolved and some more details to integrate better into the existing code base.

src/libopensc/card-openpgp.c
```diff
@@ -1644,14 +1642,14 @@ pgp_select_data(sc_card_t *card, u8 p1){
...
-	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+	LOG_FUNC_RETURN(card->ctx, r);
 }
 
 
 /**
  * ABI: ISO 7816-4 GET DATA - get contents of a DO.
  */
-static int
+	static int
```
Jakuje, line 1652: Here should be no indent

src/libopensc/card-openpgp.c
```diff
@@ -3302,36 +3300,36 @@ pgp_erase_card(sc_card_t *card)
 /**
  * ABI: ISO 7816-9 CARD CTL - perform special card-specific operations.
  */
-static int
+	static int
```
Jakuje, line 3303: no indent should be here either.

src/libopensc/card-openpgp.c
```diff
@@ -3302,36 +3300,36 @@ pgp_erase_card(sc_card_t *card)
...
-		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
-		break;
+		case SC_CARDCTL_GET_SERIALNR:
+			memmove((sc_serial_number_t *) ptr, &card->serialnr, sizeof(card->serialnr));
+			LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+			break;
+		case SC_CARDCTL_OPENPGP_SELECT_DATA:
+			r = pgp_select_data(card, *((u8 *) ptr));
+			LOG_FUNC_RETURN(card->ctx, r);
+			break;
```
Jakuje, line 3318: I do not remember I suggested indenting the `case` blocks inside the switches. I think the current coding guidelines should allow both, but less indent is always better in my opinion. But I still did not manage to get the final `clang-format` in (see #1641).
Reformatting the whole blocks with other changes in single commit makes it hard to gasp what you actually changed.

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -409,85 +409,85 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
-			r = sc_update_binary(p15card->card, 0,
-			                     (const unsigned char *) content->value,
-			                     content->len, 0);
-		}
-		break;
 
-	default:
-		r = SC_ERROR_NOT_IMPLEMENTED;
+		default:
+			r = SC_ERROR_NOT_IMPLEMENTED;
```
Jakuje, line 490: same here.

Jakuje (commented):
Thank you for addressing the review comments.This looks good, but there are still few nits, that I would like to get cleared before merging.

daringer:
yeah, sorry for this being so painful, would love to just throw `clang-format` on it, 
but will give my best to resolve this by hand:
*  function annotation indent is clearly a no-go
* good that you come up with the case/switch point, saw both in the code-base, will try to ensure a minimal diff, to avoid non-functional changes
next try follows shortly...

src/libopensc/card-openpgp.c
daringer: check

src/libopensc/card-openpgp.c
daringer: check

src/libopensc/card-openpgp.c
daringer: case in same indent as switch, check

src/pkcs15init/pkcs15-openpgp.c
daringer: fixed

Jakuje (approved):
Thanks! Looks good to me now. I will keep it open for some time if others will have some more comments.

daringer:
pushed the changes, 
* cleaned up various indents to stay as they were before to reduce the diff size
* [commit-diff](https://github.com/OpenSC/OpenSC/commit/9e8c530bb9b0191909efdbb0d662aa696d78563f) looks much cleaner now 
* the bigger block within `libopensc/pkcs15-openpgp.c` was mostly rewritten anyways

src/libopensc/pkcs15-openpgp.c
```diff
@@ -353,42 +370,58 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
 	r = sc_select_file(card, &path, &file);
 	if (r < 0)
 		goto failed;
 
-	/* If DO 7F21 holds data, we declare a cert object for pkcs15 */
-	if (file->size > 0) {
+	for(u8 i=0; i<PGP_MAX_NUM_CERTS; i++) {
 		struct sc_pkcs15_cert_info cert_info;
 		struct sc_pkcs15_object    cert_obj;
+		u8* buffer = malloc(MAX_OPENPGP_DO_SIZE);
```
frankmorgner, line 382: please add error checking

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -419,29 +419,38 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
 			sc_log(card->ctx,
 			       "This version does not support certificate ID = %d (only ID=3 is supported).",
 			       cid->value[0]);
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
 		}
+
+
 		/* Just update the certificate DO */
+		u8 param = (u8) (2 - (cid->value[0] - 1));
+		sc_card_ctl(card, SC_CARDCTL_OPENPGP_SELECT_DATA, &param);
```
frankmorgner, line 440: please add error checking

frankmorgner (changes requested):
Thank you, this would solve a long standing issue.

Is this backward compatible with v2 (where only a single certificate could be stored?

src/libopensc/pkcs15-openpgp.c
daringer: checked for `NULL` here

src/pkcs15init/pkcs15-openpgp.c
daringer: checking `sc_card_ctl`'s return value for `< 0` (also other occurrence in `libopensc/pkcs15-openpgp.c`)

daringer:
Yes, believe it should be backwards compatible:
* `openpgp_store_data` will check for the OpenPGP version and store only in `id==3` for `p15card->card->type < SC_CARD_TYPE_OPENPGP_V3`
* thus no check is necessary during readout of the certs, uhm at least before, now the error check I just added will lead to problems, means I will do an explicit check inside the readout, too

daringer:
Now also added explicit checks for the OpenPGP version inside the certificate readout and making sure the loop is only iterated once for `OpenPGP < v3`

frankmorgner:
Could it be that you forgot to push some commit? The error checking is not visible here https://github.com/OpenSC/OpenSC/pull/2103/files

daringer:
sorry pushed into the wrong branch, now it's better, will be updated again in a minute because I again messed up the indentation :roll_eyes:

src/libopensc/card-openpgp.c
```diff
@@ -247,8 +247,11 @@ static pgp_do_info_t	pgp34_objects[] = {	/**** OpenPGP card spec 3.4 ****/
 	{ 0x5f48, CONSTRUCTED, READ_NEVER  | WRITE_PIN3,  NULL,               sc_put_data },
 	{ 0x5f50, SIMPLE,      READ_ALWAYS | WRITE_PIN3,  sc_get_data,        sc_put_data },
 	{ 0x5f52, SIMPLE,      READ_ALWAYS | WRITE_NEVER, sc_get_data,        NULL        },
+
 	/* DO 7F21 is CONSTRUCTED in spec; we treat it as SIMPLE: no need to parse TLV */
-	{ DO_CERT, SIMPLE,      READ_ALWAYS | WRITE_PIN3,  sc_get_data,        sc_put_data },
+	/* kept for backwards compatibility (reads ID3), reading certs manually */
```
alex-nitrokey, line 252: Honestly, I did not follow the whole conversation here, but I do not understand this addition without it anyway, so... what is it for?

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -419,29 +419,42 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
+		if (cid->value[0] != 3 && p15card->card->type < SC_CARD_TYPE_OPENPGP_V3) {
 			sc_log(card->ctx,
 			       "This version does not support certificate ID = %d (only ID=3 is supported).",
 			       cid->value[0]);
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
 		}
+
+
 		/* Just update the certificate DO */
+		u8 param = (u8) (2 - (cid->value[0] - 1));
```
alex-nitrokey, line 439: A check for `SC_CARD_TYPE_OPENPGP_V3` is needed here as `SC_CARDCTL_OPENPGP_SELECT_DATA` does not exist in older cards (import fails indeed, I checked it).

You may put it in an else block above.

alex-nitrokey (changes requested):
I tested only importing and listing for OPCv2 and OPCv3 so far, but LGTM besides the mentioned problem in pkcs-openpgp.c for OPCv2

src/pkcs15init/pkcs15-openpgp.c
alex-nitrokey, line 439: Please explain this calculation in code. If you do not know the specification this looks like magic. Maybe we use constants here instead?

The idea is bascially:
```
AUT: cid-value[0] = 3; param=0
DEC: cid-value[0] = 2; param=1
SIG: cid-value[0] = 1; param=2
```

while I agree with this calculation, the output is 255 for the AUT key-cert-pair in specific cases, and I am not sure why (see next comment). I am not sure yet if it is happening here, we should check param though.

alex-nitrokey:
I realized the following:

When importing a key-cert pair (aka a .p12 file) with `pkcs15-init --delete-objects privkey,pubkey --id 1 --store-private-key key-cert.pfx --format pkcs12 --auth-id 3 --verify-pin` the store is iterated multiple times, depending on the slot used.

* openpgp_store_data (thus writing at 0, doesn't it?)
* select_data with 1 (!)
* openpgp_store_data
* select_data with 0 (!)
* openpgp_store_data

Instead it should do the following (imho):

* select_data with 2
* openpgp_store_data

When checking the stored certificates I get three instances which leaves me puzzled, because I would have guessed two instances, in AUT and in DEC...

But it doesn't stop here, if using `--id 2` it iterates two times and does the following:

* openpgp_store_data
* select_data with 0 (!)
* openpgp_store_data
* select_data with 255 (!) -> fails, of course (overflow error)

Here again, I get two instances and I would have guessed only one (as both storing attempts should end up in the slot for AUT).

Consequently, it works if using `--id 3` as it firstly use openpgp_store_data which stores in default AUT slot and then fails with a select_data attempt on 255 again.

I couldn't find the reason in the code yet, does anyone has any idea? Can you reproduce?

frankmorgner:
@daringer , could you please look at the comments from @alex-nitrokey ?

daringer:
@frankmorgner yeah, of course, already in the queue, will update the PR accordingly

src/pkcs15init/pkcs15-openpgp.c
daringer: Isn't this already enforced by the if-statement before:
```C
/* OpenPGP card v.2 contains only 1 certificate */
if (cid->value[0] != 3 && p15card->card->type < SC_CARD_TYPE_OPENPGP_V3) {
    sc_log(card->ctx,
        "This version does not support certificate ID = %d (only ID=3 is supported).",
        cid->value[0]);
}
```
this should ensure the following lines won't be reach w/o a valid `card->type` and `cid->value[0]` combination, or am I misunderstanding this?

src/libopensc/card-openpgp.c
daringer: that's a misleading comment, will remove it

daringer:
@alex-nitrokey looks like I cannot reproduce, let's automate a little.
Creating a `certkey.pfx` like this first:
```bash
openssl genrsa -out rsa_key.pem
openssl genrsa -out server.pem
openssl req -new -key rsa_key.pem -out cert.csr
openssl x509 -req -in cert.csr -signkey server.pem -out cert_signed.crt
openssl pkcs12 -export -out certkey.pfx -inkey server.pem -in cert_signed.crt
```
then run the following based on sources from this branch:
```bash
$ openpgp-tool --erase
```

alex-nitrokey:
I scratched my head and found the solution of the puzzle...

I used a S/MIME cert-key pair of Commodo that I created in the past. In it are two more certificates (CA chain certs). That is to say, using `pkcs15-init --delete-objects privkey,pubkey --id 1 --store-private-key myprivate.p12 --format pkcs12 --auth-id 3 --verify-pin --ignore-ca-certificates` works perfectly fine (emphasizing on `--ignore-ca-certificates`.

This raises the question: What is the recommended behaviour on handling multiple certificates (in OpenSC)? Honestly, I would prefer to always use the first certs for OpenPGP Cards as it does not really make sense to store the CA chain on these devices as well (maybe it does on others, thus having `ignore-ca-certificates` instead of the other way round like `--import-ca-certificates` in OpenSC).

So what is the best implementation for this? Currently, the multiple certs are not handled proberly (importing 3 if `--id 1` is used, 2 if `--id 2` is used and 1 if `--id3` is used). So should we rather always only import one or import all properly if `ignore-ca-certificates` is not used?

src/pkcs15init/pkcs15-openpgp.c
alex-nitrokey: The following code expects the full command range, but older cards do not have `SC_CARDCTL_OPENPGP_SELECT_DATA`. Thus, the following code fails for older cards (see line 440), no matter if they fullfill the `cid->value[0] == 3` in the mentioned if clause. That is to say, you have to differentiate your code for both types or change the code so that `SC_CARDCTL_OPENPGP_SELECT_DATA` is not used for key id 3 keys.

Do you see, what I mean?

alex-nitrokey:
Please check my latest comment in code, this PR is breaking stuff otherwise imho.

src/libopensc/card-openpgp.c
frankmorgner: you did not yet, did you?

frankmorgner (commented):
@daringer did you check the rest of the inline comments from @alex-nitrokey ?

daringer:
Nope, didn't do it yet. Alex and I made a session and came to the conclusion that we do not know how to properly handle multiple certificates passed to OpenSC ... see https://github.com/OpenSC/OpenSC/pull/2103#issuecomment-762281018 ...

Any thoughts on that?

frankmorgner:
Thanks for the pointer. I'm only using pkcs15-init for debugging purposes, not for production use, so I cannot exactly tell what's the best approach. @Jakuje , could you have a look?

Jakuje:
I am not using pkcs15-tool for production either, but if you are writing file that contains multilple certificates, I would assume it is a certificate chain with CAs and you should write only the first one, assuming the others are some (intermediate) CAs. I would even suggest to have the `--ignore-ca-certificates` on by default, but it would be probably too intrusive.

On normal PKCS#15 cards, it is quite possible to have CA certs so I would probably suggest to fix the overflow and assume users will know what they are doing (will not try to import certificate chains to openpgp cards).

dengert:
I believe this PR addresses  https://github.com/OpenSC/OpenSC/issues/1399#issuecomment-398579881 where OpenPGP 3.3 writes 3 certificates to one file on the card. It is not a certificate chain.

So some of the details of the card restrictions are now exposed in the pkcs15 layer in pkcs15-tool. These could be addressed, if the card driver took care of managing the details of how to add one cert to the card and in any order.

alex-nitrokey:
@daringer I would suggest to follow Jakubs advice then and fix the overflow and only import the first cert. I can not think of a really good use case where you would want to store the whole chain in the OpenPGP Card anyway.

Does this sound right?

ya-isakov:
This MR doesn't work for my Yubikey 5 NFC (trying it with pkcs11-tool -O)
```
P:1128981; T:0x139820467156928 17:56:35.473 [opensc-pkcs11] card-openpgp.c:1756:pgp_select_data: called
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] card-openpgp.c:1758:pgp_select_data: select data with: 0
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] apdu.c:548:sc_transmit_apdu: called
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] card.c:473:sc_lock: called
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] card.c:513:sc_lock: returning with: 0 (Success)
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] apdu.c:515:sc_transmit: called
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] apdu.c:363:sc_single_transmit: called
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] apdu.c:367:sc_single_transmit: CLA:0, INS:A5, P1:0, P2:4, data(6) 0x7ffd29e87bb2
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] reader-pcsc.c:323:pcsc_transmit: reader 'Yubico YubiKey FIDO+CCID 00 00'
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] reader-pcsc.c:324:pcsc_transmit:
Outgoing APDU (11 bytes):
00 A5 00 04 06 60 04 5C 02 7F 21 .....`.\..!
P:1128981; T:0x139820467156928 17:56:35.474 [opensc-pkcs11] reader-pcsc.c:242:pcsc_internal_transmit: called
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] reader-pcsc.c:333:pcsc_transmit:
Incoming APDU (2 bytes):
6A 80 j.
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] apdu.c:537:sc_transmit: returning with: 0 (Success)
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] card.c:523:sc_unlock: called
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] iso7816.c:128:iso7816_check_sw: Incorrect parameters in the data field
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] card-openpgp.c:1775:pgp_select_data: Card returned error: -1205 (Incorrect parameters in APDU)
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] card-openpgp.c:3527:pgp_card_ctl: returning with: -1205 (Incorrect parameters in APDU)
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] card.c:1065:sc_card_ctl: returning with: -1205 (Incorrect parameters in APDU)
P:1128981; T:0x139820467156928 17:56:35.475 [opensc-pkcs11] pkcs15-openpgp.c:547:sc_pkcs15emu_openpgp_init: Failed OpenPGP - select data: -1205 (Incorrect parameters in APDU
)
```
But yubikey-manager, which issues the same SELECT DATA, works without any problems

ya-isakov:
After checking APDUs of yubikey-manager, it seems that it's prepending length with data, already started with length. So, APDU sent by this MR is
00 A5 00 04 06 60 04 5C 02 7F 21
and yubikey-manager sent
00 A5 00 04 07 06 60 04 5C 02 7F 21
After I've changed pgp_select_data to have `apdu_data[0] = 0x06;` (and all other index shifted by one), and used 7 as datalen, pkcs15-tool -c showed me certificates:
```
X.509 Certificate [AUT certificate]
	Authority      : no
	Path           : 3f007f21
	ID             : 03

X.509 Certificate [DEC certificate]
	Authority      : no
	Path           : 3f007f21
	ID             : 02
	Encoded serial : 02 10 01C5396BD120336259440AE796C184FC
```

ya-isakov:
With my changes from https://github.com/OpenSC/OpenSC/pull/2288 (and wrapping APDU data, as I wrote in https://github.com/OpenSC/OpenSC/pull/2103#issuecomment-812895413), this MR shows me all ED25519/X25519 keys in pkcs11-tool.

frankmorgner:

src/libopensc/card-openpgp.c
daringer: done

src/pkcs15init/pkcs15-openpgp.c
daringer: done by ensuring SELECT DATA is not called for < v3 openpgp cards

daringer: done by explaining the calculation (mapping in a comment)

daringer:
Revived this PR, rebased on top of master and resolved the comments accordingly:

* removed [misleading comments](https://github.com/OpenSC/OpenSC/pull/2103#pullrequestreview-598842273)
* ensure `SELECT DATA` is not used for older OpenPGP Cards (<v3)
* in any case always [just import a single cert](https://github.com/OpenSC/OpenSC/pull/2103#issuecomment-801753130) 
* explained the conversion from `--id` to `param` in `pkcs15init/pkcs15-openpgp.c`

PR cleanups: squashed into one commit, various minor additional lines removed and some whitespaces in favor of tabs removed

Jakuje:
I would love to get this merged for 0.23.0. The code looks ok and I believe it is handling all the use cases discussed above except the workaround for old Yubikey as mentioned by @ya-isakov (https://github.com/Yubico/yubikey-manager/commit/0ba0ad0e3fd23059367415b7e7c116cb3c05a175). Do we want to implement this?

The code is also not rebased on current master and there are no CI results (can you rebase?). As part of the CI we run few tests against some old yubico openpgp applet: https://github.com/OpenSC/OpenSC/blob/master/.github/test-openpgp.sh I think this one implements only much older versions of the applet so we can not test this case. Is somebody aware of some newer applet that could be used for testing given that yubico no longer releases the source of the yubico applets?

daringer:
> I would love to get this merged for 0.23.0. 

affirmative, taking this as a motivation to keep the pace up...

> The code is also not rebased on current master and there are no CI results (can you rebase?). 

rebase on top of master: :heavy_check_mark: 

> Do we want to implement this?

Very good question, no personal preference on that.
Essentially, it looks like "SELECT DATA" needs to handle a "special case" for Yubikeys, technically no problem from a first glance - but I have no means to test such a change, no yubikey here. But I could easily realize that if someone could give me a hint for what (exactly) to check in order to branch into this "special case".

amaccuish:
Just compiled tested on Fedora with a v3.3 card and could only see the AUT cert. Loading the extra certs appears to work as I can read them back out using gpg-card readcert x > file.pem. Does the card have to be initialised a certain way?

Happy to provide ADPU logs, a card for testing (I have a few lying round). Unfortunately not a C programmer :/

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -443,29 +443,43 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
 			       "This version does not support certificate ID = %d (only ID=3 is supported).",
 			       cid->value[0]);
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
 		}
-		/* Just update the certificate DO */
+
+		/* OpenPGP card < v.3 does not support SELECT DATA calls */
+		if (p15card->card->type >= SC_CARD_TYPE_OPENPGP_V3) {
+			/* Mapping [3..1] passed --id to [0..2] for param */
+			u8 param = (u8) (2 - (cid->value[0] - 1));
```
frankmorgner, line 464: do we need to do some bound checking to avoid an "integer" underflow?

frankmorgner (changes requested):

daringer:
@amaccuish thx a lot for testing and finding this, shot my foot during the last update! should be fixed by now - as you correctly observed the certificates are in fact added, but not correctly listed nor readable using opensc, this should be fixed now.

For you and maybe others to test, I use the following:
```
# erase openpgp-card
openpgp-tool --erase 
# add certificate with id
pkcs15-init --store-certificate some-cert.pem --id 1
# list certificates
pkcs15-tool --list-certificates
```
for me this now lists the certificates properly, you can extract it like this:
```
pkcs15-tool --read-certificate 1
```
any permutation of ids works for me, too, looking for someone to verify ...

src/pkcs15init/pkcs15-openpgp.c
daringer: sounds good, will verify `param` is within expected bounds...

src/libopensc/pkcs15-openpgp.c
```diff
@@ -529,26 +544,55 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+
 		/* Certificate ID. We use the same ID as the authentication key */
-		cert_info.id.value[0] = 3;
-		cert_info.id.len = 1;
-		/* Authority, flag is zero */
-		/* The path following which PKCS15 will find the content of the object */
-		sc_format_path("3F007F21", &cert_info.path);
+		sc_pkcs15_format_id(certs[i].id, &cert_info.id);
+
+		resp_len = sc_get_data(card, 0x7F21, buffer, MAX_OPENPGP_DO_SIZE);
```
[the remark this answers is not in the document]
daringer: thx, will update (change `resp_len` to `int` and catch `< 0`)

src/pkcs15init/pkcs15-openpgp.c
daringer: done

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
daringer: done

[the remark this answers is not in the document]
daringer: maybe the `free(buffer)` is not needed here, as other `goto failed` also omit this, any opinion on this?

amaccuish:
@daringer works great! Amazing 😍

I guess it's not possible to override the key usage? For example, slot 2 can never be used for TLS since the card refuses to perform the required operations? The key usage flags in the x509 certificate are ignored correct?

Otherwise I've tested slot 1 and 3 for S/MIME and TLS auth and they work 🎉❤️

daringer:
@amaccuish good to hear, thanks for testing! 

> For example, slot 2 can never be used for TLS since the card refuses to perform the required operations? The key usage flags in the x509 certificate are ignored correct?

I have to admit that I cannot tell in full detail. The OpenPGPCard spec defines SIG, DEC and AUT and for which respective operations they shall be used (and can optionally be used), so I would assume that client applications simply obey the spec - and partly the smartcard implementation also enforces that (e.g., `PSO: COMPUTE DIGITAL SIGNATURE` may only use key 1). That's likely also the reason why the x509 flags are not taken into account.

src/pkcs15init/pkcs15-openpgp.c
frankmorgner: thanks

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
frankmorgner: thanks for fixing the error checking.

`free`ing the buffer is actually needed to avoid a memory leak. And actually, if you go back to line 563 you also need to `free` `buffer` :wink: 

There also is the preemptive exit on line 597, where sc_pkcs15emu_add_x509_cert fails. Here, the error indicates that the function failed and with that, the responsibility to free the buffer is still in the hands of the caller. please add another `free` here as well. thank you!

frankmorgner (changes requested):

src/pkcs15init/pkcs15-openpgp.c
```diff
@@ -443,29 +443,48 @@ static int openpgp_store_data(struct sc_pkcs15_card *p15card, struct sc_profile
...
+			if (param > 2) {
+				LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+			}
+
+			/* Just update the certificate DO */
+			r = sc_card_ctl(card, SC_CARDCTL_OPENPGP_SELECT_DATA, &param);
+			if (r < 0) {
+				LOG_TEST_RET(card->ctx, r, "Failed OpenPGP - select data");
+				LOG_FUNC_RETURN(card->ctx, r);
+			}
```
Jakuje, line 475: this looks at least suspicious. First you check for error, then you check again and return on error (=always) and return and then you  return anyway. I think these three lines are functionally equivalent to the second one, if I see right:
```suggestion
			LOG_TEST_RET(card->ctx, r, "Failed OpenPGP - select data");
```

daringer:
uh nice, thanks for the thorough reviews! fixes are on the way

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
daringer: * `goto failed` line 563 -> `free(buffer)` added
* `goto failed` line 597 -> `free(buffer)` added

src/libopensc/card-openpgp.c
```diff
@@ -1767,6 +1767,43 @@ pgp_get_pubkey_pem(sc_card_t *card, unsigned int tag, u8 *buf, size_t buf_len)
...
+	// create apdu data (taken from spec: SELECT DATA 7.2.5.)
+	apdu_data[0] = 0x60;
+	apdu_data[1] = 0x04;
+	apdu_data[2] = 0x5c;
+	apdu_data[3] = 0x02;
+	apdu_data[4] = 0x7f;
+	apdu_data[5] = 0x21;
+
+	// apdu, cla, ins, p1, p2, data, datalen, resp, resplen
+	sc_format_apdu_ex(&apdu, 0x00, 0xA5, p1, 0x04, apdu_data, 6, NULL, 0);
```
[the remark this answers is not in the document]
daringer: done

src/pkcs15init/pkcs15-openpgp.c
daringer: done

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
daringer: line 564 returns on failure, I could remove the `goto failed` afterwards, right? 
because currently the `goto failed` will never be reached, or am I seeing this wrong?

src/libopensc/pkcs15-openpgp.c
```diff
@@ -529,26 +544,61 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+			goto failed;
 
 		memset(&cert_info, 0, sizeof(cert_info));
 		memset(&cert_obj,  0, sizeof(cert_obj));
 
+		/* only try to SELECT DATA for OpenPGP >= v3 */
+		if (card->type >= SC_CARD_TYPE_OPENPGP_V3) {
+			r = sc_card_ctl(card, SC_CARDCTL_OPENPGP_SELECT_DATA, &i);
+			if (r < 0) {
+				LOG_TEST_RET(card->ctx, r, "Failed OpenPGP - select data");
```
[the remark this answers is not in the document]
daringer: done

src/libopensc/pkcs15-openpgp.c
```diff
@@ -529,26 +544,64 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+			goto failed;
 
 		memset(&cert_info, 0, sizeof(cert_info));
 		memset(&cert_obj,  0, sizeof(cert_obj));
 
+		/* only try to SELECT DATA for OpenPGP >= v3 */
+		if (card->type >= SC_CARD_TYPE_OPENPGP_V3) {
+			r = sc_card_ctl(card, SC_CARDCTL_OPENPGP_SELECT_DATA, &i);
+			if (r < 0) {
+                free(buffer);
```
Jakuje, line 563: wrong indent (spaces instead of tabs)

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
Jakuje: right. The goto on line 565 can not be reached now.

src/libopensc/pkcs15-openpgp.c
```diff
@@ -529,26 +544,64 @@ sc_pkcs15emu_openpgp_init(sc_pkcs15_card_t *p15card)
...
+		cert_obj.flags = certs[i].obj_flags;
+
 		/* Object label */
-		strlcpy(cert_obj.label, "Cardholder certificate", sizeof(cert_obj.label));
+		strlcpy(cert_obj.label, certs[i].label, sizeof(cert_obj.label));
 
 		r = sc_pkcs15emu_add_x509_cert(p15card, &cert_obj, &cert_info);
-		if (r < 0)
+		if (r < 0) {
+            free(buffer);
```
Jakuje, line 598:
```suggestion
			free(buffer);
```

daringer: damn, sry, fixed!

src/libopensc/pkcs15-openpgp.c
daringer: fixed...

src/libopensc/pkcs15-openpgp.c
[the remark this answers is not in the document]
daringer: removed it

frankmorgner (approved):
Looks good now, thank you

Jakuje (approved):

Jakuje:
Thank you. Merging for next rc/final release of 0.23.0

PR #2104, opened 2020-09-03
prompt for env vars in bash completion for PIN/password options

eighthave:
This adds a new case of options in the bash completion and normalized the docs regarding the PIN options.  I tested this on Debian/buster.  I wrote these files originally in #238

##### Checklist


Jakuje:
Should also the [npa-tool](https://github.com/OpenSC/OpenSC/blob/master/doc/tools/npa-tool.1.xml#L79) documentation be updated to work with this? Otherwise it looks sensible.

eighthave:
I know nothing about _npa-tool_, but looking at the XML, it seems like it should match the others.

Jakuje:
I don't know much about that either, bu ti was running grep through the other occurrences of PIN in documentation. We can probably leave it out as it looks like it has its own way for getting PINs from environment.

Jakuje (approved):

frankmorgner:
npa-tool is slightly different with hard coded names. Thank you!

PR #2107, opened 2020-09-09
PKCS#11 C_FindObjectsInit does not see private objects when logged in as CKU_SO

avrx1985:
PKCS#11 C_FindObjectsInit does not see private objects when logged in as CKU_SO.
However in case CKU_USER all private objects are visible, but we can't manage them.
For example, when we want to remove stored key pair (using C_DestroyObject) and then re-create it with C_GenerateKeyPair, it is required to log in as SO, but we don't see private key then and can't destroy it, so the call fails.

### DIFF
--- a/src/pkcs11/pkcs11-object.c
+++ b/src/pkcs11/pkcs11-object.c
@@ -375,7 +375,7 @@ C_FindObjectsInit(CK_SESSION_HANDLE hSession,	/* the session's handle */
 
 	/* Check whether we should hide private objects */
 	hide_private = 0;
-	if (slot->login_user != CKU_USER && (slot->token_info.flags & CKF_LOGIN_REQUIRED))
+	if ((slot->login_user != CKU_SO) && (slot->login_user != CKU_USER) && (slot->token_info.flags & CKF_LOGIN_REQUIRED))
 		hide_private = 1;
 
 	/* For each object in token do */


src/pkcs11/pkcs11-object.c
```diff
@@ -375,7 +375,7 @@ C_FindObjectsInit(CK_SESSION_HANDLE hSession,	/* the session's handle */
 
 	/* Check whether we should hide private objects */
 	hide_private = 0;
-	if (slot->login_user != CKU_USER && (slot->token_info.flags & CKF_LOGIN_REQUIRED))
+	if ((slot->login_user != CKU_SO) && (slot->login_user != CKU_USER) && (slot->token_info.flags & CKF_LOGIN_REQUIRED))
```
Jakuje, line 378: There are 4 values the `slot->login_user` could have:
 * `-1` for not logged in user
 * `0` for `CKU_SO` login
 * `1` for `CKU_USER` login
 * `2` for `CKU_CONTEXT_SPECIFIC` (AFAIK this does not get propagated to the above variable, but is used for logins).

Your suggested change makes it very hard to understand what is going on there. Would it make sense to replace it with:

    if ((slot->login_user == -1) && (slot->token_info.flags & CKF_LOGIN_REQUIRED)

dpondeljak:
Kaj je to

src/pkcs11/pkcs11-object.c
avrx1985: Yes, you're probably right. It would be better to check against not logged in at all.

PR #2112, opened 2020-09-17
More fixups for fuzzing

Jakuje:
Tested with my coolkey cards.

##### Checklist
- [ ] macOS tokend is tested


frankmorgner:
Thank you

PR #2113, opened 2020-09-19
pkcs11-tool: allow using SW tokens

conradoplg:
I wanted to test SoftHSM with `pkcs11-tool --test` and had to remove `CKF_HW` from the mechanism filter, which is what this PR does.

I can't think of a reason for only allowing HW mechanisms, but let me know if I'm missing something.

### DIFF
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -612,6 +612,15 @@
 					</para></listitem>
 				</varlistentry>
 
+				<varlistentry>
+					<term>
+						<option>--allow-sw</option>
+					</term>
+					<listitem><para>Allow using software mechanisms that do not have the CKF_HW flag set.
+					May be required when using software tokens and emulators.
+					</para></listitem>
+				</varlistentry>
+
 			</variablelist>
 		</para>
 	</refsect1>
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -160,7 +160,8 @@ enum {
 	OPT_SIGNATURE_FILE,
 	OPT_ALWAYS_AUTH,
 	OPT_ALLOWED_MECHANISMS,
-	OPT_OBJECT_INDEX
+	OPT_OBJECT_INDEX,
+	OPT_ALLOW_SW
 };
 
 static const struct option options[] = {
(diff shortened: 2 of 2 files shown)


dengert:
I can think of a reason. pkcs11 is for accessing hardware devices that provide the real security. The hardware provides additional physical security. 

pkcs11 Base v2.40 Introduction:
The PKCS#11 standard specifies an application programming interface (API), called “Cryptoki,” for
devices that hold cryptographic information and perform cryptographic functions. Cryptoki follows a
simple object based approach, addressing the goals of technology independence (any kind of device) and
resource sharing (multiple applications accessing multiple devices), presenting to applications a common,
logical view of the device called a “cryptographic token”.

If you wanted to submit a PR with a parameter to ignored the CKF_HW flag that would be more acceptable. 
But the default should be use CKF_HW.

conradoplg:
I agree that hardware provides additional security. But should pkcs11-tool care? (In particular, should `--test` care?) I don't see a reason to prevent the user from using pkcs11-tool it with a software token if they want to. Maybe as a sanity check to prevent them from accidentally using a software token by accident, but seems unlikely...

I see these options:

- Force CKF_HW (current behavior)
- Remove CKF_HW (current PR)
- Remove CKF_HW from `--test` only
- Remove CKF_HW when a flag is specified (`--allow-sw`)?

Let me know what's best and I'll change it accordingly!

dengert:
I would vote for "Remove CKF_HW when a flag is specified" 

I would like to hear from others and what is the name of the option.

mouse07410:
I am OK with either one of these:
- Remove CKF_HW (current PR) *preferred*, or
- Remove CKF_HW when a flag is specified (`--allow-sw`)

I do **not** see value in `pkcs11-tool` trying to do the enforcement.

popovec:
I vote for:
* Remove CKF_HW when a flag is specified (--allow-sw)

frankmorgner:
prefering a HW slot is what the tool should do. you may implement the commandline switch for SW tokens as suggested above.

Jakuje:
I am fine with either. The pkcs11-tool is just a tool to do some operations on pkcs11 library, whatever it provides.

Indeed it would be good if it would prefer hw variants of the operations (not sure if it is possible/common to have the same mechanism in HW and not in HW in the same pkcs11 library).

In any case, I consider pkcs11-tool as a testing tool rather than something that should be used for any production critical operations. We already have quite a lot configuration switches (which is good for testing various libraries, cards), but sometimes pain for testing pkcs11-tool itself and using.

conradoplg:
I've updated the PR with `--allow-sw`

conradoplg:
Note that each mechanism can be reported with either CKF_HW or not, you can't have both, so it's not possible to pick up a software implementation instead of a hardware implementation.

frankmorgner (approved):

Jakuje (approved):
Thanks for the contribution and clarification. Looks good to me.

PR #1924, opened 2020-01-30
avoid calling sc_*_binary recursively

frankmorgner:
- lock the card early to avoid deselection of the file
- check on integer overflows of indices

fixes https://github.com/OpenSC/OpenSC/issues/1919

Since this affects many cards, I'll not integrate this until 0.21.0 is out.

##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [ ] PKCS#11 module is tested
- [ ] Windows minidriver is tested
- [ ] macOS tokend is tested

PR #2116, opened 2020-09-24
idprime: Unbreak reading of compressed certificates

Jakuje:
The midair collision of #1924 and IDPrime support (#1772) caused that the IDPrime driver did not work when was merged in master. I was notified by Pavel over email and short review of the logs turned out that the iso operations are no longer called recursively (for getting something larger that does not find into APDU).

The attached patch solves the issue for me (I will ask also the reporter to confirm).

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-idprime.c
+++ b/src/libopensc/card-idprime.c
@@ -526,24 +526,31 @@ static int idprime_read_binary(sc_card_t *card, unsigned int offset,
 	unsigned char *buf, size_t count, unsigned long flags)
 {
 	struct idprime_private_data *priv = card->drv_data;
-	int r;
+	int r = 0;
 	int size;
 
 	sc_log(card->ctx, "called; %"SC_FORMAT_LEN_SIZE_T"u bytes at offset %d",
 		count, offset);
 
 	if (!priv->cached && offset == 0) {
+		/* Read what was reported by FCI from select command */
+		int left = priv->file_size;
+		size_t read = 0;
+
 		// this function is called to read and uncompress the certificate
 		u8 buffer[SC_MAX_EXT_APDU_BUFFER_SIZE];
 		if (sizeof(buffer) < count) {
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
 		}
-		/* Read what was reported by FCI from select command */
-		r = iso_ops->read_binary(card, 0, buffer, priv->file_size, flags);
-		if (r < 0) {
-			LOG_FUNC_RETURN(card->ctx, r);
+		while (left > 0) {
+			r = iso_ops->read_binary(card, read, buffer + read, priv->file_size - read, flags);
+			if (r <= 0) {
+				LOG_FUNC_RETURN(card->ctx, r);
+			}
+			left -= r;
+			read += r;
 		}
-		if (r < 4) {
+		if (read < 4 || read != priv->file_size) {
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_DATA);
 		}
 		if (buffer[0] == 1 && buffer[1] == 0) {


frankmorgner:
How/where are you reading uncompressed certificates? Although not being called recursively, sc_read_binary should still get the whole file if the output buffer is big enough... I.e. idprime_ read_binary would be called multiple times from sc_read_binary with an increased offset if there are still some bytes available. No need to implement this on the driver level.

Jakuje:
As far as I know, we do not know if it is compressed or not before we read first two bytes (uncompressed certificate is in the else branch, just couple of lines below):
```
 		if (buffer[0] == 1 && buffer[1] == 0) {
```
This information could be present somewhere else too, but as I have encountered only cards with compressed certificates, I can not confirm this for certain.

I could refactor this and separate reading uncompressed and compressed certificate if you prefer, but I am not sure if it would be more readable.

frankmorgner:
If this minimal change works for you, I'm fine with it, too. It just *looks* like using one of the high level functions could simplify the code...

Jakuje:
Yes, this change works for me. Can you clarify which high-level function I should use to simplify the code? I do not see the `sc_decompress_*` called from anywhere else than from existing card drivers.

frankmorgner:
Actually, I was simply thinking of sc_read_binary. However, now I realize that you're already decompressing in idprime_read_binary, so that is not straight forward to implement. You could do the following: 

In idprime_select_file:
save the information whether the selected file is compressed or not in something like `priv->file_is_compressed`

In idprime_read_binary:
`if (!priv->file_is_compressed)`, just forward the request to `iso_ops->read_binary`
`else { priv->file_is_compressed = 0; sc_read_binary(...); priv->file_is_compressed = 1;` and then decompress and return the data. `}`

But as said above, I'm fine with your PR.

With a broader look at this issue, I wonder whether we should rather handle the compressed certificates in the PKCS#15 layer. Do you know, why this isn't done? Isn't it standardized enough?

dengert:
In response to: https://github.com/OpenSC/OpenSC/pull/2116#issuecomment-698658316 select file FCI will return the size of the file, and may be used by caller to allocate a buffer. But if decompressing is being done on the fly while the data is being read the decompressed size will be larger than the compressed size  and the read binary returns to few bytes. 

i.e. if it is compressed, the reading (and caching of decompressed data) needs to be done as part of select file so FCI contains the uncompressed length. The read card driver's read binary can use  the correct length and access the cached data. 

PIV driver had this problem in 2006,  as certificate objects may contain compressed certificates. So it decompresses the contained certificate and caches it and emulates select file and returns uncompressed size, and card driver emulates read_binary.

frankmorgner:
@dengert, please look at
https://github.com/OpenSC/OpenSC/blob/57a391f16cd69cedf02d5b326a689d19a985de11/src/libopensc/card-idprime.c#L491-L522

... When the file is selected, a compressed file is detected and the *uncompressed* size is used as file size

dengert:
OK

Jakuje:
It was also one of my thoughts that it could be moved into the PKCS#15 layer, but I think the compression markers are not well standardized (especially with cards further from the PKCS#15 standard as PIV).

For example CAC, has compression marked in [CERT_INFO tag's third byte](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-cac.c#L579), PIV has [different byte](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L1039) and here we have just 2 prefix bytes.

So unless there is somebody with better overview of PKCS#15 standard, who is able to bring more light into this, I would keep it as it is now. 

This works with my card, but I am still in contact with Pavel, who originally reported this to me through email. He has a card with different OS release version and after certificate exchange, the `key_id` reference no longer works for him, so I will be probably updating this PR with more changes after we will figure out what is the issue.

dengert:
Can not find "compress" or "zip" in PKCS\#15-v1.1  iso 7816-15 may have something.

PIV driver will still decompress certificates to get the public key from certificate so it know private key size and algo 
 and curve.

frankmorgner:
Indeed, I can't find it, too. However, nothing is hindering us from adding a compression flag to `sc_pkcs15_object_t`, which can be handled at the end of `sc_pkcs15_read_file()`. Let's leave this to one of the future releases.

Issue #2061, opened 2020-06-15
Towards new release 0.21.0


Jakuje:
There have been talks about new release for quite some time as there have been many quite important fixes in master since 0.20.0.

I would like to kick off some testing and summarizing of changelog to get a new version out of door soon. Feel free to edit the following draft, add what I missed or propose missing issues/PRs. Let me know if there is something else where I can help towards the next release.

------------------------

### General Improvements

 * Bump minimal required OpenSSL version to 1.0.1 (#1658)
 * Implement basic unit tests for asn1 library, compression and simpletlv parser (#1830)
 * Allow generating code coverage
 * Improve fuzzing by providing corpus from real cards (#1830)
 * Implement support for OAEP encryption
 * New separate debug level for PIN commands (d06f23e8)
 * Fix handling of card/reader insertion/removal events in pcscd
 * Many bugfixes reported by oss-fuzz, coverity and lgtm.com
 * Fixes of removed readers handling (#1970)
 * Fix Firefox crash because of invalid pcsc context (#2077)

### PKCS #11

 * Return CKR_TOKEN_NOT_RECOGNIZED for not recognized cards (#2030)
 * Propagate ignore_user_content to PKCS#11 layer not to confuse applications (#2040)

### MacOS

 * Add installer signing for PR and master
 * Move OpenSC to MacOS Utilities folder (#2063)

### pkcs11-tool

 * Make SHA256 default for OAEP encryption

### opensc-explorer

 * `asn1` accepts offsets and decode records (#2090)
 * `cat` accepts records (#2090)

### OpenPGP

 * Add new ec curves supported by GNUK (#1853)
 * First steps supporting OpenPGP 3.4
 * Add support for EC key import (#1821)

### CardOS

 * Improve detection of various CardOS 5 configurations (#1987)

### ePass2003

 * Improve ECC support (#1859)
 * Fixed erase sequence (#2097)

### IDPrime

 * New driver for Gemalto IDPrime (only some types) (#1772)

### eDo

 * New driver with initial support for Polish eID card (e-dowód, eDO) (#2023)

### MCRD

 * Remove unused and broken RSA EstEID support (#2095)

### TCOS

 * Add missing encryption certificates (#2083)

### IAS-ECC (#2070):

 * Fixed support for Idemia Cosmo cards with AWP middleware interoperability (previously broken).
 * Added support for Idemia Cosmo v8 cards.
 * PIN padding settings are now used from PKCS #15 info when available.
[...]

PR #2109, opened 2020-09-12
Issue 2105: Add new DNIe CA structure for the secure channel

rickyepoderi:
Fixes #2105.

The police department in Spain (DGP) has changed the certificate structure for the secure channel recently. DNIe cards since serial number `BMP100001 ` (`IDESP` field in `dnie-tool -a`) are using new certs, keys and keys references to establish the secure channel. The new values for the new configuration have not been published by the DGP yet, they were discovered dumping and parsing the official binary distribution.

The two different configurations are now setup into a [configuration array](https://github.com/rickyepoderi/OpenSC/blob/dnie-2/src/libopensc/cwa-dnie.c#L417). The exact channel data to use is saved in [a new field inside the card private data](https://github.com/rickyepoderi/OpenSC/blob/dnie-2/src/libopensc/cwa-dnie.h#L91). And there are two new methods to assign and retrieve the channel data to use ([here](https://github.com/rickyepoderi/OpenSC/blob/dnie-2/src/libopensc/cwa-dnie.c#L608-L659)). As @2opremio suggested in the issue, the `dnie_set_channel_data` uses the `issuer` of the `icc_intermediate_ca_cert` to know which infrastructure the current card has to use. Please let me know if you agree with this idea or if you prefer another approach.

Don't merge it yet. We need to test this with different DNIe cards as I only have the new one. @miguel-cv please test with previous DNIe 3.0 and 2.0 if you can (maybe I have done a mistake copying all the data, it's a mess). But wait until the current PR is more or less accepted. I don't want you to work twice.

Issue #1796, opened 2019-09-17
pkcs11-tool -t: unwrap and encryption/decryption still broken (see also issue #1786)


carblue:
### Problem Description

There are still errors with pkcs11-tool -t  (see output below)

1. The "fix #1786" by commit 25bc8fc was incomplete, in that now (for me) there is a total match within the first 10 bytes of Original/Decrypted but the length differ and thus result in error(s) (see output below in the end)

2. "Key unwrap" testing seems to be not fully implemented in OpenSC. With the following (incomplete) patch I succeed to at least get a call to card's sc_card_operations:unwrap function, otherwise that would never be called:

### Proposed Resolution
For encryption/decryption: nothing supplied.
For unwrap maybe something similar to:
```
user@host:~/workspace/libopensc/C/contribute_opensc/OpenSC-1$ diff -u src/tools/pkcs11-tool.c_orig src/tools/pkcs11-tool.c                                              
--- src/tools/pkcs11-tool.c_orig        2019-09-16 17:45:14.000000000 +0200
+++ src/tools/pkcs11-tool.c     2019-09-17 13:14:11.564117967 +0200
@@ -5383,9 +5383,17 @@
        unsigned char   iv[32], ciphered[1024], cleartext[1024];
        int             ciphered_len, cleartext_len, len;
        CK_MECHANISM    mech;
-       CK_ULONG        key_type = CKM_DES_CBC;
+       CK_ULONG        key_type = CKK_DES;//CKM_DES_CBC;
        CK_ULONG key_len_ul;
-       CK_ATTRIBUTE    key_template = { CKA_KEY_TYPE, &key_type, sizeof(key_type) };
+       CK_BBOOL false = FALSE;
+       CK_OBJECT_CLASS  class = CKO_SECRET_KEY;
+       CK_BYTE id = 0xFF; // this is not suitable generally to be a fixed id ?
+       CK_ATTRIBUTE    key_template[4] = {
+               { CKA_KEY_TYPE, &key_type, sizeof(key_type) },
+               { CKA_TOKEN, &false, sizeof(false) }, /* session only object */
+               { CKA_CLASS, &class, sizeof(class) },
+               { CKA_ID,    &id, sizeof(id) }
+       };
 
        pkey = get_public_key(session, privKeyObject);
        if (pkey == NULL)
@@ -5426,7 +5434,7 @@
        mech.mechanism = CKM_RSA_PKCS;
[...]

PR #2117, opened 2020-09-25
Prepare RC1 of 0.21.0

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


Jakuje:
Generally looks good. I would just add Fixes #2061. If you took the list of changes from there, I am not sure how up-to-date is that and what all landed since I was composing that list (also worth checking again).

While grepping through the code, I found parts disabling wrap/unwrap tests in pkcs11-tool, that will get enabled by bumping version, referencing the issue #1796: 
```
#if OPENSC_VERSION_MAJOR == 0 && OPENSC_VERSION_MINOR <= 20
```
I am not sure what is the status of that, nor what is the actual problem, but it might be something worth looking into as it just got a new comment few days back.

frankmorgner:
1. I've update the description of #2061 (check the edit history)
2. I've copied the description into NEWS

Only the mention of #2109 should now be missing

I'll have a look at the wrapping tests...

PR #2118, opened 2020-09-25
itacns: fixed length checking

frankmorgner:
fixes https://oss-fuzz.com/testcase-detail/5116660103774208

I hope this is not breaking anything.

I think the correct solution would rather be to use an ASN.1 parser.

Could someone familiar with the itacns have a look, please? @nunojpg


Jakuje (approved):
Changes look good to me.

I am trying to add the links to the chromium issue tracker, which becomes public after the embargo expires, rather than to the oss-fuzz testcase, which will stay private forever (if I understand right). Could you add this too?

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25935

frankmorgner:
I was not aware. sure, I can do.

PR #2122, opened 2020-09-29
card-rtecp: Add Rutoken ECP SC NFC ATR

lolyonok:
Added new ATR for card over NFC. It has the same functionality as the ISO one.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
thank you

PR #2124, opened 2020-10-01
Create codeql-analysis.yml

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
If I'm not mistaken, Semmle is behind both LGTM and Github's CodeQL. So I'm not sure if we need both services...

PR #2126, opened 2020-10-02
idprime: Unbreak OS version 2 after key&cert renewal

Jakuje:
Fix the IDPrime key reference for OS version 2.

One year later after implementing the initial support for IDPrime, I got contacted by an owner of another IDPrime card that it no longer works after certificate renewal. This card had only on certificate so there was quite a lot of guesswork. After some more digging, I updated the code to work also with renewed certificates (with different IDs), which is confirmed it works.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-idprime.c
+++ b/src/libopensc/card-idprime.c
@@ -196,8 +196,12 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, 
 			&& (memcmp(&start[12], "mscp", 5) == 0)) {
 			new_object.fd++;
 			if (card->type == SC_CARD_TYPE_IDPRIME_V2) {
-				/* The key reference starts from 0x11 */
-				new_object.key_reference = 0x10 + new_object.fd;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-idprime.c
```diff
@@ -196,8 +196,9 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
 			&& (memcmp(&start[12], "mscp", 5) == 0)) {
 			new_object.fd++;
 			if (card->type == SC_CARD_TYPE_IDPRIME_V2) {
-				/* The key reference starts from 0x11 */
-				new_object.key_reference = 0x10 + new_object.fd;
+				/* The key reference starts from 0x11 and increments by the key id (ASCII) */
+				int key_id = start[8] - '0';
```
frankmorgner, line 200: I suppose an integer underflow isn't critical here, right?

Jakuje: the value is passed to the set security environment APDU so it would always be better to be safe. I will add a safeguard.

PR #2129, opened 2020-10-06
Support changing PIN with CAC Alt tokens

Jakuje:
Some CAC Alt tokens support PIN change, but the APDU is not following the ISO specification closely (the P1 is different in the APDU).

I am not aware of any other CAC card that would support PIN change at this moment so I am changing this generally in the CAC driver. It will probably not work with generic CAC or dual CAC-PIV cards (but it is probably not working now either).

Tested locally with my testing HID CAC Alt token.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
If only a specific set of tokens are supporting the change of PIN, shouldn't we allow changing the PIN only for those tokens? I think your current approach suggests that it's available for all CAC tokens, which may brick the card if the user tries changing the PIN too often if it not implemented or differently implemented.

Jakuje:
Not sure where my comment got lost, but I added a check and new card type that is affected by this change.

Jakuje:
@frankmorgner any more comments on the current version of the patch or are you fine merging this also for 0.21.0?

frankmorgner:
go ahead if you want to

frankmorgner (approved):

PR #2132, opened 2020-10-12
westcos-tool.c:fix memory leaks in dst->modulus.data and dst->exponen…

whzhe51:
…t.data

Signed-off-by: whzhe <wanghongzhe@huawei.com>

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/westcos-tool.c
+++ b/src/tools/westcos-tool.c
@@ -673,11 +673,16 @@ int main(int argc, char *argv[])
 
 			if (!do_convert_bignum(&dst->modulus, rsa_n)
 			 || !do_convert_bignum(&dst->exponent, rsa_e))
+			{
+				free(dst->modulus.data);
+				free(dst->exponent.data);
 				goto out;
-
+			}
 		}
 
 		r = sc_pkcs15_encode_pubkey(ctx, &key, &pdata, &lg);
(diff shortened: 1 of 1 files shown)


src/tools/westcos-tool.c
```diff
@@ -678,6 +678,10 @@ int main(int argc, char *argv[])
 		}
```
Jakuje, line 678: `dst->modulus` already leaks if the condition on line 675 fails. It would be worth adding the free there too.

src/tools/westcos-tool.c
```diff
@@ -678,6 +678,10 @@ int main(int argc, char *argv[])
...
 
 		r = sc_pkcs15_encode_pubkey(ctx, &key, &pdata, &lg);
+		if(dst->modulus.data)
```
Jakuje, line 681: On line 582, the `key` structure is initialized to zeroes so this condition is bogus. Calling just free should be enough. In the following condition too.

Jakuje (changes requested):

whzhe51:
@Jakuje OK，I fixed the mr.

Jakuje (approved):
lgtm

frankmorgner:
thanks!

Issue #2135, opened 2020-10-16
There may be a memory leak in main in /src/tools/netkey-tool.c


hifoolno:
https://github.com/OpenSC/OpenSC/blob/26a404d6a53bfab7b8ea764a8763e23727c2986e/src/tools/netkey-tool.c#L533-L566

Hello, I submitted a pull request yesterday (https://github.com/OpenSC/OpenSC/pull/2134), and now I find another similar problem. I think in the fragment above, when failed to access reader, cxt needs to be released before exiting the program. Like in the patch o [CVE-2019-6502](https://github.com/OpenSC/OpenSC/issues/1586), a sc_release_context(ctx) is needed before every exit(), or a memory leak may occur. 
Look forward to your reply.

frankmorgner:
It depends. Typically, the OS destroys the process, when `exit()` is called so all memory is freed anyway. But calling `sc_release_context()` doesn't hurt. If you could submit a PR, we can merge it.

hifoolno:
Thank you for your kind reply. I just submitted a PR.

PR #2137, opened 2020-10-18
Add sc_release_context() before exit()

hifoolno:
Fixes https://github.com/OpenSC/OpenSC/issues/2135

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/netkey-tool.c
+++ b/src/tools/netkey-tool.c
@@ -542,17 +542,20 @@ int main(
 		if(!strcmp("tcos", ctx->card_drivers[i]->short_name)) break;
 	if(!ctx->card_drivers[i]){
 		fprintf(stderr,"Context does not support TCOS-cards\n");
+		sc_release_context(ctx);
 		exit(1);
 	}
 
 	printf("%d Readers detected\n", sc_ctx_get_reader_count(ctx));
 	if(reader < 0 || reader >= (int)sc_ctx_get_reader_count(ctx)){
 		fprintf(stderr,"Cannot open reader %d\n", reader);
+		sc_release_context(ctx);
 		exit(1);
 	}
 
 	if((r = sc_connect_card(sc_ctx_get_reader(ctx, 0), &card))<0){
 		fprintf(stderr,"Connect-Card failed: %s\n", sc_strerror(r));
+		sc_release_context(ctx);
 		exit(1);
 	}
 	printf("\nCard detected (driver: %s)\nATR:", card->driver->name);
@@ -562,6 +565,7 @@ int main(
 
 	if((r = sc_lock(card))<0){
 		fprintf(stderr,"Lock failed: %s\n", sc_strerror(r));
+		sc_release_context(ctx);
 		exit(1);
 	}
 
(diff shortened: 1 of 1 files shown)


src/tools/netkey-tool.c
```diff
@@ -542,17 +542,20 @@ int main(
 		if(!strcmp("tcos", ctx->card_drivers[i]->short_name)) break;
 	if(!ctx->card_drivers[i]){
 		fprintf(stderr,"Context does not support TCOS-cards\n");
+		sc_release_context(ctx);
```
Jakuje, line 545: You should initialize the `ctx` to NULL on line 440 to make sure we are not accessing uninitialized memory if something goes wrong. It is not a must, but it would be great as we are already touching the code.

src/tools/netkey-tool.c
```diff
@@ -562,6 +565,7 @@ int main(
 
 	if((r = sc_lock(card))<0){
 		fprintf(stderr,"Lock failed: %s\n", sc_strerror(r));
+		sc_release_context(ctx);
```
Jakuje, line 568: Here you need to call `sc_card_free(card)` too as `sc_connect_card()` allocates new `sc_card_t` structure.

Jakuje: Initializing it to NULL on a line 442 is also a good idea just to be sure.

Jakuje (changes requested):

src/tools/netkey-tool.c
hifoolno: Hello~ I followed your advice and submitted a new PR(https://github.com/OpenSC/OpenSC/pull/2138), but the LGTM analysis was not successful. I found that the `sc_card_free()` was not defined in src/tools/netkey-tool.c . Maybe you mean `sc_disconnect_card()` ?

Jakuje: You don't need to submit a new PR -- you can push more commits to this one. Right. The `sc_card_free` is not public. `sc_disconnect_card` is the right function. Sorry for a confusion.

hifoolno: I'm so sorry to disturb you again. I found that no `sc_disconnect_card` is called when connect-card failed in other scripts (e.g. `main` in src\tools\npa-tool.c), so I am still a little confused about calling `sc_disconnect_card` here.

Jakuje: It is the way as @frankmorgner wrote in #2135 -- calling exit frees all the memory so making sure we free that explicitly is mostly a good custom here and protection on "bad" platforms that might not do that. For example, `opensc-tool` calls this in the only in the `end` block  (reached from all error condition using `goto`). This is similar in all the other tools.

hifoolno: Thank you very much for taking the time to reply! I've pushed another commit to this PR~ (This was just my second attempt to submit a PR, and I did learn a lot. Thanks again!)

Jakuje: You are welcomed. Change looks good. Thank you for your contribution!

Jakuje (approved):

PR #2140, opened 2020-10-20
pkcs11-tool support key-gen for GENERIC secret key

### LINKED ISSUE #2139: How to keygen a generic secret? pkcs11-tool error if using --keygen --key-type secrkey:512 -M GENERIC-SECRET-KEY-GEN
### Problem Description

How can I use pkcs11-tool to generate a generic secret for use in HmacSHA256? Running --list-mechanisms on SoftHSM2 v2.5.0 for Windows, I see that PKCS#11 driver supports GENERIC-SECRET-KEY-GEN. However, when I run pkcs11-tool with --keygen --key-type secrkey:512 parameters I get "error: Unknown key type secrkey:512".

I am unclear if this error is a bug or limitation of pkcs11-tool, or if there is something missing from --help documentation.

### Proposed Resolution

The --help output for --key-gen parameter could add extra examples. This is the current output from pkcs11-tool for OpenSC:
- Specify the type and length of the key to create, for example rsa:1024 or EC:prime256v1 or GOSTR3410-2012-256:B

It would be useful to add more common examples like AES:32 and SECRKEY:512.

If bug or limitation related to --keygen of a secrkey, please fix the code.

### Steps to reproduce
From --help I figured out how to generate RSA-3027.
- "C:\Program Files\OpenSC Project\OpenSC\tools\pkcs11-tool.exe" --module C:\SoftHSM2\lib\softhsm2-x64.dll --slot-index 0 --pin 0000 --login --label rsa3072 --keypairgen --key-type rsa:3072 --mechanism RSA-PKCS-KEY-PAIR-GEN
Using slot with index 0 (0x1e94de1b)
Key pair generated:
Private Key Object; RSA
  label:      rsa3072
  Usage:      decrypt, sign, unwrap
  Access:     sensitive, always sensitive, never extractable, local
Public Key Object; RSA 3072 bits
  label:      rsa3072
  Usage:      encrypt, verify, wrap
  Access:     local

There was no AES example for --keygen, but --list-mechanisms shows AES-KEY-GEN with range 16-32 so I guessed aes:32.
- "C:\Program Files\OpenSC Project\OpenSC\tools\pkcs11-tool.exe" --module C:\SoftHSM2\lib\softhsm2-x64.dll --slot-index 0 --pin 0000 --login --label aes256 --keygen --key-type aes:32 --mechanism AES-KEY-GEN
Using slot with index 0 (0x1e94de1b)
Key generated:
Secret Key Object; AES length 32
warning: PKCS11 function C_GetAttributeValue(VALUE) failed: rv = CKR_ATTRIBUTE_SENSITIVE (0x11)

  label:      aes256
  Usage:      encrypt, decrypt, verify, wrap, unwrap
  Access:     never extractable, local

There was no SECRKEY example for --keygen, but --list-mechanisms shows GENERIC-SECRET-KEY-GEN with length up to 2048 so I guessed secrkey:512. It doesn't work.
- C:\SoftHSM2\bin>"C:\Program Files\OpenSC Project\OpenSC\tools\pkcs11-tool.exe" --module C:\SoftHSM2\lib\softhsm2-x64.dll --slot-index 0 --pin 0000 --login --label hmacSha256 --keygen --key-type secrkey:512 --mechanism GENERIC-SECRET-KEY-GEN
**Using slot with index 0 (0x1e94de1b)
error: Unknown key type secrkey:512
Aborting.**

### Logs

I enabled debug logging for --keygen with --key-type secrkey:512, but debug log file is empty.

dengert:
Fixes #2139

Added code to support  mechanism GENERIC-SECRET-KEY-GEN.

Improved --help because key gen of symmetric keys pass
CKA_VALUE_LEN which is length of key in bytes.

Tested with:
```
./pkcs11-tool --module /usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so \
 --login --label generic-64 --keygen --key-type GENERIC:64 \
 --mechanism GENERIC-SECRET-KEY-GEN

./pkcs11-tool --module /usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so --login -O
```

 On branch pkcs11-tool-generic
 Changes to be committed:
	modified:   pkcs11-tool.c

##### Checklist

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -272,7 +272,7 @@ static const char *option_help[] = {
 	"Unlock User PIN (without '--login' unlock in logged in session; otherwise '--login-type' has to be 'context-specific')",
 	"Key pair generation",
 	"Key generation",
-	"Specify the type and length of the key to create, for example rsa:1024 or EC:prime256v1 or GOSTR3410-2012-256:B",
+	"Specify the type and length (bytes if symmetric) of the key to create, for example rsa:1024, EC:prime256v1, GOSTR3410-2012-256:B, AES:16 or GENERIC:64",
 	"Specify 'sign' key usage flag (sets SIGN in privkey, sets VERIFY in pubkey)",
 	"Specify 'decrypt' key usage flag (RSA only, set DECRYPT privkey, ENCRYPT in pubkey)",
 	"Specify 'derive' key usage flag (EC only)",
@@ -2340,7 +2340,7 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 				n_privkey_attr++;
 			}
 		}
-		else if (!strncmp(type, "EC:", 3))   {
+		else if (strncmp(type, "EC:", strlen("EC:")) == 0 || strncmp(type, "ec:", strlen("ec:")) == 0)  {
 			CK_MECHANISM_TYPE mtypes[] = {CKM_EC_KEY_PAIR_GEN};
 			size_t mtypes_num = sizeof(mtypes)/sizeof(mtypes[0]);
 			int ii;
@@ -2639,6 +2639,26 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
 			key_length = (unsigned long)atol(size);
 			if (key_length == 0)
 				key_length = 16;
+
+			FILL_ATTR(keyTemplate[n_attr], CKA_KEY_TYPE, &key_type, sizeof(key_type));
+			n_attr++;
+		}
+		else if (strncasecmp(type, "GENERIC:", strlen("GENERIC:")) == 0) {
+			CK_MECHANISM_TYPE mtypes[] = {CKM_GENERIC_SECRET_KEY_GEN};
+			size_t mtypes_num = sizeof(mtypes)/sizeof(mtypes[0]);
+			const char *size = type + strlen("GENERIC:");
+
+			key_type = CKK_GENERIC_SECRET;
+
+			if (!opt_mechanism_used)
+				if (!find_mechanism(slot, CKF_GENERATE, mtypes, mtypes_num, &opt_mechanism))
+					util_fatal("Generate Key mechanism not supported\n");
+
+			if (size == NULL)
+				util_fatal("Unknown key type %s", type);
+			key_length = (unsigned long)atol(size);
+			if (key_length == 0)
+				key_length = 32;
 
 			FILL_ATTR(keyTemplate[n_attr], CKA_KEY_TYPE, &key_type, sizeof(key_type));
 			n_attr++;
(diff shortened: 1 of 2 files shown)


frankmorgner (changes requested):
Looks good, but could you add some documentation in `doc/tools/pkcs11-tool.1.xml`?

src/tools/pkcs11-tool.c
```diff
@@ -2640,6 +2640,26 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
 			FILL_ATTR(keyTemplate[n_attr], CKA_KEY_TYPE, &key_type, sizeof(key_type));
 			n_attr++;
 		}
+		else if (strncasecmp(type, "GENERIC:", strlen(":GENERIC")) == 0) {
```
Jakuje, line 2643: why `GENERIC:` and `:GENERIC` (not the colon).

src/tools/pkcs11-tool.c
```diff
@@ -271,8 +271,8 @@ static const char *option_help[] = {
 	"Change User PIN",
 	"Unlock User PIN (without '--login' unlock in logged in session; otherwise '--login-type' has to be 'context-specific')",
 	"Key pair generation",
-	"Key generation",
-	"Specify the type and length of the key to create, for example rsa:1024 or EC:prime256v1 or GOSTR3410-2012-256:B",
+	"Key generation ",
```
Jakuje, line 274: bogus trailing space

Jakuje (commented):
As frank mentioned, the manual page should be updated too.

src/tools/pkcs11-tool.c
dengert: fixed in rebased squashed commit. pkcs11-tool.1.xml was updated that updates man page.

dengert: fixed

src/tools/pkcs11-tool.c
dengert: fixed

Jakuje (approved):

frankmorgner (commented):
I'm a bit puzzled. `gen_keypair() uses:
```
		if (strncmp(type, "RSA:", strlen("RSA:")) == 0 || strncmp(type, "rsa:", strlen("rsa:")) == 
```
while `gen_key()` uses:
```
		if (strncasecmp(type, "AES:", strlen("AES:")) == 0) {
```
Could you add an other commit making this consistent?

dengert:
The gen_keypair  uses strncmp and does upper and lower tests for all the asymmetric keys: RSA, EC and GOSTR3410.  But the GOST3410 code will only accept upper A, B or C after ":". 

The gen_key uses  strncasecmp for all the symmetric keys: AES, DES, DES3 and GENERIC. 

The only reason I changed the EC test was because it would only accept upper, and followed the gen_keypair convention of using 2  strncmp tests. 

So to make it consistent, the DES and DES3 would also need changes. I can do that if you want. 
Note: PKCS11 always uses upper case for key names. pkcs11-tool is being less restrictive.

PKCS11 also defines: CKK_DES2, CKM_DES2_KEY_GEN which we do not support, other the by DES3 with first 8 bytes == last 8 bytes.

PR #2142, opened 2020-10-22
PIV global pin bug

### LINKED ISSUE #2141: PIN preference set incorrectly for PIV Card
For PIV Cards, OpenSC does not set the PIN preference to the be the Global PIN if the card supports on-card biometric (OCC) comparison or the virtual contact interface (VCI).

Under SP 800-73-3, the first byte of the PIN Usage Policy in the Discovery object could only be 0x40 or 0x60. So, lines 2606 - 2608 of https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c set the pin_preference to be the Global PIN if byte 1 of the PIN Usage Policy is 0x60 and byte two is 0x20:
```
if (*pinp == 0x60 && *(pinp+1) == 0x20) { /* use Global pin */
    sc_log(card->ctx, "Pin Preference - Global");
    priv->pin_preference = 0x00;
}
```
However, this does not work with SP 800-73-4, since SP 800-73-4 allows for additional bits to be set in the first byte of the PIN Usage Policy.

In order to work with SP 800-73-4, OpenSC should only look at bit 6 of the first byte of the PIN Usage Policy. So, something like:
```
if ((*pinp & 0x20) == 0x20 && *(pinp+1) == 0x20) { /* use Global pin */
    sc_log(card->ctx, "Pin Preference - Global");
    priv->pin_preference = 0x00;
}
```

dengert:
Fixes #2141

NIST 800-73-3 based cards only had 2 bits set in first pin policy byte.
NIST 800-73-4 defines additions bits in first pin policy byte.
When one of these bit is set, and the policy prefers the Global pin,
it is not recognized and the local pin is used.

 On branch PIV-global-pin-bug
 Changes to be committed:

##### Checklist

- [X] PKCS#11 module is tested  using IDEMIA test card with pin policy 6820 
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2603,7 +2603,7 @@ static int piv_parse_discovery(sc_card_t *card, u8 * rbuf, size_t rbuflen, int a
 				if (pinp && pinplen == 2) {
 					sc_log(card->ctx, "Discovery pinp flags=0x%2.2x 0x%2.2x",*pinp, *(pinp+1));
 					r = SC_SUCCESS;
-					if (*pinp == 0x60 && *(pinp+1) == 0x20) { /* use Global pin */
+					if ((*pinp & 0x60) == 0x60 && *(pinp+1) == 0x20) { /* use Global pin */
 						sc_log(card->ctx, "Pin Preference - Global");
 						priv->pin_preference = 0x00;
 					}


Jakuje (approved):
lgtm

Issue #2060, opened 2020-06-12
pkcs11-register not recognizing symbolic links


henning-schild:
### Problem Description

I observed the problem on a Linux machine that already has opensc modules in my various `pkcs11.txt`. The helper does not detect that and writes to the file without using `modutil` - the _only_ allowed tool to mess with that stuff.

Multiple issues occur. First i get the module entered twice. Second newlines are missing so the file is corrupted and `modutil` can not find/remove entries anymore.

The helper does not seem to know that there are conflicts between modules.  `onepin-opensc-pkcs11.so` vs ` opensc-pkcs11.so` vs. `p11-kit-proxy.so`

### Proposed Resolution

The tool should make use of `modutil` instead of writing to those files itself. It should not overwrite previously existing entries if unsure. It should know about conflicting modules ... Or it should just stay away from those files!

### Steps to reproduce

Just one example here.

`.pki/nssdb/pkcs11.txt` before
```
library=
name=NSS Internal PKCS #11 Module
parameters=configdir='sql:/home/henning/.pki/nssdb' certPrefix='' keyPrefix='' secmod='secmod.db' flags=optimizeSpace updatedir='' updateCertPrefix='' updateKeyPrefix='' updateid='' updateTokenDescription='' 
NSS=Flags=internal,critical trustOrder=75 cipherOrder=100 slotParams=(1={slotFlags=[ECC,RSA,DSA,DH,RC2,RC4,DES,RANDOM,SHA1,MD5,MD2,SSL,TLS,AES,Camellia,SEED,SHA256,SHA512] askpw=any timeout=30})

library=/usr/lib64/pkcs11/p11-kit-trust.so
name=my-p11-kit-trust.so
NSS=trustOrder=100    

library=/usr/lib64/pkcs11/onepin-opensc-pkcs11.so
name=my-onepin-opensc-pkcs11.so

```
after `pkcs11-register`

```
library=
name=NSS Internal PKCS #11 Module
parameters=configdir='sql:/home/henning/.pki/nssdb' certPrefix='' keyPrefix='' secmod='secmod.db' flags=optimizeSpace updatedir='' updateCertPrefix='' updateKeyPrefix='' updateid='' updateTokenDescription='' 
[...]

PR #2144, opened 2020-11-01
configure: disable "autostart-items" by default

henning-schild:
Related-to: https://github.com/OpenSC/OpenSC/issues/2060
Signed-off-by: Henning Schild <henning@hennsch.de>

##### Checklist

- [x] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,9 +273,9 @@ AC_ARG_ENABLE(
 
 AC_ARG_ENABLE(
 	[autostart-items],
-	[AS_HELP_STRING([--enable-autostart-items],[enable autostart items @<:@enabled@:>@])],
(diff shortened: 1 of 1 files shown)


frankmorgner:
A real solution would be to recognize symlinks instead...

frankmorgner (changes requested):
A real solution would be to recognize symlinks instead...

henning-schild:
Sure, and hard-links, bind-mounts, all other pkcs#11 plugins in the world, that opensc can not possibly know about ... browsers that users do not want to have configured for opensc.

The tool needs more work before it can be considered safe for unconditional regular automatic use. And making that the default means distros will likely inherit it ... once they go >=0.20 (which many of the big ones have not yet done)

frankmorgner:
Feel free to rework and make a new PR

PR #2148, opened 2020-11-10
Use more standard location for file cache

Jakuje:
Currently, opensc uses non-standard `~/.eid/cache` directory to cache files. These days, there is a bit more standardization around the places which should be used for these things and from my reading of [1], `XDG_CACHE_HOME` is a place where we should place our cache by default.

The location is still possible to adjust in the configuration file (but it makes it hard as it does not support any replacement characters/environment variables, especially in multi-user systems.

[1] https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/doc/files/files.html
+++ b/doc/files/files.html
@@ -724,9 +724,11 @@ app <em class="replaceable"><code>application</code></em> {
 					</span></dt><dd><p>
 							Where to cache the card's files. The default values are:
 							</p><div class="itemizedlist"><ul class="itemizedlist" style="list-style-type: disc; "><li class="listitem"><p>
-										<code class="filename"><code class="envar">HOME</code>/.eid/cache/</code> (Unix)
+										<code class="filename"><code class="envar">$XDG_CACHE_HOME</code>/opensc/</code> (if defined)
 									</p></li><li class="listitem"><p>
-										<code class="filename"><code class="envar">USERPROFILE</code>\.eid-cache\</code> (Windows)
+										<code class="filename"><code class="envar">$HOME</code>/.cache/opensc/</code> (Unix)
+									</p></li><li class="listitem"><p>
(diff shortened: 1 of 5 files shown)


dengert:
Sounds good.  What will happen to users who already have a .eid directory? 

The PIV driver can use it for off card certs as listed in the history object.  But it is very unlikely it is being used, as newer cards have more memory to store the certs on the card and extracting the URL out of the history object to download the certs is not straight forward. 

Any other drivers using it in strange ways?

Jakuje:
I think this option is mostly used to cache certificates from card on disk on slow cards. If users have .eid directory, it will just create a new cache (for some performance penalty). I am not sure if it is worth to implement some moving logic inside of opensc as it can be quite racy and the old directory can contain many unused files on older systems.

We tested this configuration option enabled with all cards we have and we did not notice any issues so far, except for performance improvements. Surprisingly, the slowest cards, which lead us to investigation of this path were the PIV cards as they also have the largest speedup with file caching, but we saw improvements also in the CardOS cards (pkcs15) and in others. I do not mind going into more details, investigating the strange ways drivers using it and fixing possible issues. 

If it was used in different location for some off-card certs, it will certainly break and there will be no other way out than moving the directory to new location. I am not sure how widely is that used though these days but I would guess very little. If there is a better way how to change the location of the cache directory (or you prefer the old one) or any other suggestions, I would be happy to consider that. This was the first idea to use more standard location in Linux.

frankmorgner:
If I'm not mistaken, then the option for storing historic PIV certificates on disk is not even documented. I doubt that it is even used in practice. I think using `XDG_CACHE_HOME` is a good idea and recreating old cached data is OK.

PR #2152, opened 2020-11-16
Regular dose of fixed issues reported by oss-fuzz

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-tcos.c
+++ b/src/libopensc/pkcs15-tcos.c
@@ -139,7 +139,7 @@ static int insert_key(
 			"Searching for Key-Ref %02X\n", key_reference);
 		while ((r = sc_read_record(card, ++rec_no, buf, sizeof(buf), SC_RECORD_BY_REC_NR)) > 0) {
 			int found = 0;
-			if (buf[0] != 0xA0)
+			if (buf[0] != 0xA0 || r < 2)
 				continue;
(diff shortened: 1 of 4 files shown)


src/libopensc/pkcs15-tcos.c
```diff
@@ -141,7 +141,7 @@ static int insert_key(
 			int found = 0;
 			if (buf[0] != 0xA0)
 				continue;
-			for (i = 2; i < buf[1] + 2; i += 2 + buf[i + 1]) {
+			for (i = 2; i < buf[1] + 2 && i < r - 2; i += 2 + buf[i + 1]) {
```
frankmorgner, line 144: If `r` is exactly `1`, then `r - 2` is `-1`. We're lucky that `i` is of type `int`, which makes your fix work. However, if `i` was a `size_t` then `-1` would be translated to a very large number, which would not fix the out of bounds access. Do you think we should reformulate the fix for clearity?

Jakuje: You are indeed right. We need to check `sc_read_record()` returns at least 2 bytes in the `buf` buffer. For `r = 1`, accessing `buf[1]` would be also out of bounds read. I added a check in the top commit (can squash later).

frankmorgner (approved):
Thank you, I think we're good.

PR #2166, opened 2020-11-25
pkcs11-tool: Avoid calloc with 0 argument

Jakuje:
Even though the behavior of calloc with 0 argument is quite safe (from `man calloc`):

> If nmemb or size is 0, then calloc() returns either NULL, or a unique pointer value that can later be successfully passed to free().

It can cause various issues with memory sanitizers as previously reported in #1978. This is another occurrence where the similar pattern is happening. If we do not have any slots, it does not make sense to allocate anything in pkcs11-tool if I am right.

##### Checklist
- [-] PKCS#11 module is tested
- [-] Windows minidriver is tested
- [-] macOS tokend is tested


t184256 (approved):
Looks sensible.

frankmorgner:
thanks

PR #2183, opened 2020-12-11
Friday fuzzing and testing

Jakuje:
Fixing couple of more fuzzing issues.

On the way, I tried to make working the clang-tidy, which is in place for quite some time, but really does not report much useful results (because configuration yields many warnings as memcpy and memset is used all over the place) and it was actually not installed in the CI machines.

I noticed also couple of duplicate symbols in `.exports`, which probably did not cause any issues, but my pedantic mind wanted to clear it.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ addons:
     - ant
     - socat
     - cmake
+    - clang-tidy
 
 before_install:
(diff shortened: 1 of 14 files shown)


frankmorgner (approved):

Jakuje:
Seems like this fails in the osx  build. Let me see if this can be fixed.

PR #2186, opened 2020-12-16
man page update, info about generating EC keys

popovec:
modified:   doc/tools/pkcs15-init.1.xml

##### Checklist

### DIFF
--- a/doc/tools/pkcs15-init.1.xml
+++ b/doc/tools/pkcs15-init.1.xml
@@ -136,11 +136,12 @@
 				<command>pkcs15-init --generate-key " keyspec " --auth-id " nn</command>
 			</para>
 			<para>
-				where <replaceable>keyspec</replaceable> describes the algorithm and length of the
-				key to be created, such as <literal>rsa/512</literal>. This will create a 512 bit
-				RSA key. Currently, only RSA key generation is supported. Note that cards
-				usually support just a few different key lengths. Almost all cards will support
-				512 and 1024 bit keys, some will support 768 or 2048 as well.
+				where <replaceable>keyspec</replaceable> describes the algorithm and the parameters
+				of the key to be created. For example, <literal>rsa:2048</literal> generates a RSA key
+				with 2048-bit modulus. If you are generating an EC key, the curve designation must
+				be specified, for example <literal>ec:prime256v1</literal>. For symmetric key,
+				the length of key is specified in bytes, for example  <literal>AES:32</literal>
+				or <literal>DES3:24</literal>.
 			</para>
 			<para>
 				<replaceable>nn</replaceable> is the ID of a user PIN installed previously,
@@ -242,7 +243,7 @@
 				you would use
 			</para>
 			<para>
-				<command>pkcs15-init --store-secret-key /dev/urandom --secret-key-algorithm aes/256 --auth-id 01</command>
+				<command>pkcs15-init --store-secret-key /dev/urandom --secret-key-algorithm aes:256 --auth-id 01</command>
 			</para>
 			<para>
 				By default a random ID is generated for the secret key. You may specify an ID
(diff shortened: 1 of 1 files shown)


doc/tools/pkcs15-init.1.xml
```diff
@@ -138,9 +138,10 @@
 			<para>
 				where <replaceable>keyspec</replaceable> describes the algorithm and length of the
 				key to be created, such as <literal>rsa/512</literal>. This will create a 512 bit
-				RSA key. Currently, only RSA key generation is supported. Note that cards
-				usually support just a few different key lengths. Almost all cards will support
-				512 and 1024 bit keys, some will support 768 or 2048 as well.
+				RSA key. If you are generating an EC key, you must enter a curve specification,
+				for example <literal>ec-prime256v1</literal> or <literal>ec-secp384r1</literal>.
+				Note that cards usually support just a few different key lengths. Almost all cards
+				will support 512 and 1024 bit keys, some will support 768 or 2048 as well.
```
Jakuje, line 144: The last sentence sounds misplaced. It should be just after the RSA  sentence as it refers to the RSA key lengths.

popovec: Ok,   before I change this patch and do force push, I'll wait a few days to see if anyone else has comment on this text. 

New version:
`````
     where <replaceable>keyspec</replaceable> describes the algorithm and length of the
     key to be created, such as <literal>rsa/512</literal>. This will create a 512 bit
     RSA key. Note that cards usually support just a few different key lengths. Almost 
     all cards will support 512 and 1024 bit keys, some will support 768 or 2048 as well.
     If you are generating an EC key, you must enter a curve specification,
     for example <literal>ec-prime256v1</literal> or <literal>ec-secp384r1</literal>.
`````

dengert: RSA keys should be at least 2048 these days. Please use 2048 as an example. Some cards can go as high as 4096.  I would drop all the comments about key sizes for a card other then the example. 
"curve specification"  should use the term "named curve" or "namedCurve"  

The RSA section says: "rsa/512" but the ec section uses "ec-prime256v1"  can you use the same separators. pkcs11-tool uses ":" and can accept upper or lower case for the keytype.

popovec: Yes, you're right with RSA, I'll move it to 2048 key size..

I considered using the term "namedCurve" - but I assume that the manual is read by someone who does not know the terminology. 

Thanks for the info about the separator, I will use `rsa:2048` and `ec:prime256v1`  . There is also info about `Secret Key Upload`, here `aes/256` is used .. It would be appropriate to unify the separator here as well. (There is also in `OPTIONS` section in `--generate-key` paragraph information about "slash" as separator .. it will of course need to be changed to "colon"..  )

New version:
`````
     where <replaceable>keyspec</replaceable> describes the algorithm and the parameters of 
     the key to be created. For example, <literal>rsa:2048</literal> generates a RSA key with 
     2048-bit modulus. If you are generating an EC key, the curve designation must be specified,
     for example <literal>ec:prime256v1</literal>.
`````

dengert: Much better. And it could support symmetric keys too. Also see: #2140
and  updated [pkcs11-tool.1.xml](https://github.com/OpenSC/OpesnSC/pull/2140/files#diff-eed650dde8d69b3bf8fc029b5b54b132c5988523024d66d95ee2c2fc0da53e17)

If / was acceptable in the past, maybe it should accept either "/" or ":" ?

popovec: It  seems there are more characters that can be used as separator slash `/`,colon `:`, or hyphen `-` (src/tools/pkcs15-init.c line 755 _if (*spec == '/' || *spec == '-' || *spec == ':')_) .   I will list them all in `OPTIONS` section. In the examples I will use `:` to match the syntax of `pkcs11-tool`.

Updated version:
`````
     where <replaceable>keyspec</replaceable> describes the algorithm and the parameters of
     the key to be created. For example, <literal>rsa:2048</literal> generates a RSA key with
     2048-bit modulus. If you are generating an EC key, the curve designation must be specified,
     for example <literal>ec:prime256v1</literal>. For symmetric key the length of key is specified
     in bytes, for example  <literal>AES:32</literal> or <literal>DES3:24</literal>.
`````

Jakuje: looks good for me

frankmorgner:
❤️ thanks a lot!

Issue #2199, opened 2021-01-08
opensc 0.21.0 break my openvpn config using a Aladdin Knowledge Systems Token JC PKI


EricV:
As soon as I install opensc 0.21.0, the detection of this PKI token by openvp is broken. 

Bus 005 Device 003: ID 0529:0620 Aladdin Knowledge Systems Token JC

### Steps to reproduce
It seems the PKI key is no more detected and the code keeps asking for inserting the token and the window to enter Pin code never displayed. Downgrading to opensc 0.20.0, the problem vanishes.

Note that for this PKI, there is a specific library to use libeToken.so but pkcs11-tool works when instructed to use this module.

It is probably a bad interaction via libpkcs11-helper but I have no clue yet.

Could you provide some direction to help debugging the problem.

dengert:
Start with Debug log and SPY trace. See:  [Using-OpenSC](https://github.com/OpenSC/OpenSC/wiki/Using-OpenSC)

frankmorgner:
Please create a debug log of your problem as described in the [comments of our issue template](https://raw.githubusercontent.com/OpenSC/OpenSC/master/.github/ISSUE_TEMPLATE.md). You may also want to look at the hints on [how to write a good bug report](https://github.com/OpenSC/OpenSC/wiki/How-to-write-a-good-bug-report).

EricV:
Thanks for the people who responded. I probably need to give you some more context to see why the general way of debugging is hard to use (at least I think so) in my case.

First the openvpn process is started by systemd and is used as a slave process that receives its command via a communication socket (See /usr/share/doc/openvpn/management-notes.txt.gz for command
output format and real-time message format). Then a graphical front end, display the progress and ask for PKI PIN code or barks if it does not detect the key, probably from real notification send by openvpn process through the socket channel.

So the command line option, opensc log and even the spy library trick is hardly relevant because at first, I do not call or use opensc directly, then although I know what pkcs11 library should be used to access my dongle, is never explicitly set it and I do not even know where the path for the library is given to openvpn.

First do you have a clue on were the interaction between openvpn and opensc-pkcs-11 could take place. Only installing opensc-pkcs-11 is sufficient to break my openvpn setup.

dpkg -L opensc-pkcs11 
/.
/usr
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libopensc.so.6.0.0
[...]

PR #2200, opened 2021-01-11
Fixed C_GetSlotInfo handling of unknown token

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2199

(handling of pkcs#11 module in p11-kit)

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -578,7 +578,7 @@ CK_RV C_GetSlotInfo(CK_SLOT_ID slotID, CK_SLOT_INFO_PTR pInfo)
 			now = get_current_time();
 			if (now >= slot->slot_state_expires || now == 0) {
 				/* Update slot status */
-				rv = card_detect(slot->reader);
+				rv = card_detect(slot->reader, "C_GetSlotInfo");
 				sc_log(context, "C_GetSlotInfo() card detect rv 0x%lX", rv);
 
 				if (rv == CKR_TOKEN_NOT_RECOGNIZED || rv == CKR_OK)
@@ -616,7 +616,7 @@ CK_RV C_GetMechanismList(CK_SLOT_ID slotID,
 	if (rv != CKR_OK)
 		return rv;
 
-	rv = slot_get_token(slotID, &slot);
+	rv = slot_get_token(slotID, &slot, "C_GetMechanismList");
 	if (rv == CKR_OK)
 		rv = sc_pkcs11_get_mechanism_list(slot->p11card, pMechanismList, pulCount);
 
@@ -638,7 +638,7 @@ CK_RV C_GetMechanismInfo(CK_SLOT_ID slotID,
 	if (rv != CKR_OK)
 		return rv;
 
-	rv = slot_get_token(slotID, &slot);
+	rv = slot_get_token(slotID, &slot, "C_GetMechanismInfo");
 	if (rv == CKR_OK)
 		rv = sc_pkcs11_get_mechanism_info(slot->p11card, type, pInfo);
 
@@ -661,7 +661,7 @@ CK_RV C_InitToken(CK_SLOT_ID slotID,
 	if (rv != CKR_OK)
 		return rv;
 
-	rv = slot_get_token(slotID, &slot);
+	rv = slot_get_token(slotID, &slot, "C_InitToken");
 	if (rv != CKR_OK)   {
 		sc_log(context, "C_InitToken() get token error 0x%lX", rv);
 		goto out;
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -301,7 +301,7 @@ get_fw_data(struct sc_pkcs11_card *p11card, struct sc_app_info *app_info, int *o
 
 /* PKCS#15 Framework */
 static CK_RV
-pkcs15_bind(struct sc_pkcs11_card *p11card, struct sc_app_info *app_info)
+pkcs15_bind(struct sc_pkcs11_card *p11card, struct sc_app_info *app_info, const char *ctx)
 {
 	struct pkcs15_fw_data *fw_data = NULL;
 	struct sc_aid *aid = app_info ? &app_info->aid : NULL;
@@ -324,7 +324,7 @@ pkcs15_bind(struct sc_pkcs11_card *p11card, struct sc_app_info *app_info)
 	rc = sc_pkcs15_bind(p11card->card, aid, &fw_data->p15_card);
 	if (rc != SC_SUCCESS) {
 		sc_log(context, "sc_pkcs15_bind failed: %d", rc);
-		return sc_to_cryptoki_error(rc, NULL);
+		return sc_to_cryptoki_error(rc, ctx);
 	}
 
 	/* Mechanisms are registered globally per card. Checking
@@ -550,7 +550,7 @@ CK_RV C_GetTokenInfo(CK_SLOT_ID slotID, CK_TOKEN_INFO_PTR pInfo)
 	if (rv != CKR_OK)
 		return rv;
 
-	rv = slot_get_token(slotID, &slot);
+	rv = slot_get_token(slotID, &slot, "C_GetTokenInfo");
 	if (rv != CKR_OK)   {
 		sc_log(context, "C_GetTokenInfo() get token: rv 0x%lX", rv);
 		goto out;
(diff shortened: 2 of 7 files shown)


src/pkcs11/pkcs11-global.c
```diff
@@ -590,7 +590,12 @@ CK_RV C_GetSlotInfo(CK_SLOT_ID slotID, CK_SLOT_INFO_PTR pInfo)
 		}
 	}
 
-	if (rv == CKR_TOKEN_NOT_PRESENT || rv == CKR_TOKEN_NOT_RECOGNIZED)
+	if (rv == CKR_TOKEN_NOT_PRESENT || rv == CKR_TOKEN_NOT_RECOGNIZED
+			/* The sub-routines are loosing the context of "C_GetSlotInfo" and
+			 * return a different default value for SC_ERROR_INVALID_CARD than
+			 * CKR_TOKEN_NOT_RECOGNIZED.  Explicitly check for that default
+			 * value, here */
+			|| rv == sc_to_cryptoki_error(SC_ERROR_INVALID_CARD, NULL))
```
Jakuje, line 598: should not this have the second argument `"C_GetSlotInfo"` so it can get the context?

Jakuje: In any case, what you wrote is equivalent to `rv == CKR_FUNCTION_NOT_SUPPORTED`, isn't it?

frankmorgner: yes, it's intended to translate the default value for SC_ERROR_INVALID_CARD (currently CKR_FUNCTION_NOT_SUPPORTED), because SC_ERROR_INVALID_CARD should not return an error in C_GetSlotInfo.

dengert:
Testing #2200 With a PIV card and Yubikey and running to simulate an unknom card to OpenSC:
```
export OPENSC_DRIVER=myeid
export PKCS11SPY_OUTPUT=/tmp/spy.txt
export PKCS11SPY=/opt/ossl-1.1/lib/opensc-pkcs11.so
/opt/ossl-1.1/bin/pkcs11-tool --module /opt/ossl-1.1/lib/pkcs11-spy.so -L
Available slots:
Slot 0 (0x0): (GetSlotInfo failed, CKR_GENERAL_ERROR)
Slot 1 (0x4): (GetSlotInfo failed, CKR_GENERAL_ERROR)
```
[pr-2200.opensc-debug.log](https://github.com/OpenSC/OpenSC/files/5836313/pr-2200.opensc-debug.log)
[pr-2200.spy.txt](https://github.com/OpenSC/OpenSC/files/5836315/pr-2200.spy.txt)

The C_GetSlotList works, but C_GetSlotInfo returns CKR_GENERAL_ERROR
It should return CKR_TOKEN_NOT_RECOGNIZED as it used to do.

frankmorgner:
Thanks for having a look!

> The C_GetSlotList works, but C_GetSlotInfo returns CKR_GENERAL_ERROR
> It should return CKR_TOKEN_NOT_RECOGNIZED as it used to do.

[PKCS#11 2.40](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html) does not allow `CKR_TOKEN_NOT_RECOGNIZED` for `C_GetSlotInfo`:
> Return values: CKR_ARGUMENTS_BAD, CKR_BUFFER_TOO_SMALL, CKR_CRYPTOKI_NOT_INITIALIZED, CKR_FUNCTION_FAILED, CKR_GENERAL_ERROR, CKR_HOST_MEMORY, CKR_OK.

dengert:
Here is spy trace of Ubuntu opensc-pkcs11.so
C_GetSlotInfo returns OK, and its C_GetTokenInfo that returns CKR_TOKEN_NOT_RECOGNIZED
The slot/reader is ok, its the token that is not recognized.


[pr-2200-working.spy.txt](https://github.com/OpenSC/OpenSC/files/5836412/pr-2200-working.spy.txt)

frankmorgner:
I'm puzzled: In your test case, do you want C_GetSlotInfo to return CKR_TOKEN_NOT_RECOGNIZED or CKR_OK?

dengert:
I should **not** have said "It should return CKR_TOKEN_NOT_RECOGNIZED as it used to do."

C_GetSlotInfo should return OK, as the slot is OK. 

So when you fix C_GetSlotInfo to return OK, then we/you can look at what C_GetTokenInfo returns. 
C_GetTokenInfo that  should return CKR_TOKEN_NOT_RECOGNIZED

frankmorgner:
Now I think what you mean. I've added a change to fix that.

dengert:
OK, that looks better.
[pr-2200-1-spy.txt](https://github.com/OpenSC/OpenSC/files/5836670/pr-2200-1-spy.txt)

dengert:
One more comment. C_GetSlotInfo should not need special treatment.  It has to do with when OpenSC connects to the the card. C_GetSlotInfo really **only** cares if a token is in the slot. It does not care if it is usable or not. But OpenSC  connects to the card when  C_GetSlotInfo is called. 


This means that if an application can bypass calling C_GetTokenInfo it may not know if the token is good or not. 

For example pkcs11-tool -O does not call C_GetTokenInfo
[pr-2000-1-O-spy.txt](https://github.com/OpenSC/OpenSC/files/5836818/pr-2000-1-O-spy.txt)

pkcs11-tool --test --login does call C_GetTokenInfo.

My concern with this PR is there are a lot of changes which may not have been checked and even if we fix the regression for the C_GetSlotList, C_GetSlotInfo and C_GetTokenInfo there may be other routines which now return  different return code such as  CKR_GENERAL_ERROR then they did before.

mouse07410:
I don't like this - it's a solution to a non-existent (in practice) problem. More so, this change is likely to be an elephant in a china shop.

I don't want to chase new incompatibilities and broken code that this is bound to create - like in the OpenVPN case.

frankmorgner:
I share your concern that low level changes like this may have unpredictable consequences. This, however, must not indicate that we should go back to a behavior that's not conforming to the PKCS#11 standard. Since my time is limited, I can only propose the changes given here, which I think are minimal, conforming to the standard, and fixing the actual problem of #2199. If I had more time, I would have added something to the test infrastructure which prevents problems like this in the future. Unfortunately, I can't do that right now.

As you can see, I did not handle C_GetSlotInfo seperately before 7cbcdde22ffc2001c94757d1edd503645e9f3574, because its return values don't differ compared with the default values. Since C_GetSlotInfo seperately handles CKR_TOKEN_NOT_RECOGNIZED (which it must not return to the application), I decided to allow CKR_TOKEN_NOT_RECOGNIZED in the low level operations, again.

The real question, is what set of return values one considers to  be a default set. I think the original intent of `sc_to_cryptoki_error` was to unify the error code handling so that only allowed values are returned to the application. When this is done too early, we are loosing valuable information. The other way to solve the issue in question, would be to only call sc_to_cryptoki_error() in the top layer, all other functions should be changed to return the internal (OpenSC) error codes. This change, would have been huge compared to this PR. (e.g. detect_card should return an internal error code rather than a PKCS#11 one)

dengert:
I agree with @mouse07410. This is a regression, and the original PR should be reverted, and 0.21.1 should be released.  See comments in https://github.com/OpenSC/OpenSC/issues/2199

The approach in the original  PR and this has been to default everything to some  return code like CKR_GENERAL_ERROR. The approach should have been to only change specific errors as they were found.

frankmorgner:
I agree, that reverting 1bb2547 is an option, too. However, please also add code that respects PKCS#11' allowed return values. Would anyone of you care to do this?

mouse07410:
Standard is there to facilitate interoperability. Nothing more.

If interoperability is already there, standard's job is **done**. 

This code's purpose is **not** to serve as a "validation gateway" for HW and SW implementations.

dengert:
I would like to see something, especially when opensc-pkcs11 can not handle a card. With more and more use of p11-kit to provide multiple pkcs11 modules, opensc-pkcs11 must not cause problems or overhead when it is asked to look at a token it can not support. in the example above of pkcs11-tool -O,  every function returned CKR_OK.  The call to  C_OpenSession should have returned  CKR_TOKEN_NOT_RECOGNIZED. PKCS11 2.40 says: 
"Make one or more calls to C_OpenSession to obtain one or more sessions with the token." There are 7 references to CKR_TOKEN_NOT_RECOGNIZED on 2.40. I have not looked at all of them. 

The card connect should flag a slot with an unsupported token. OpenSC maybe configured on a machine where the user never uses a card supported by OpenSC.  As long as the token is not removed, it should  not keep trying to match it to a driver (and sending APDU to the token) which is extra overhead and possible interference with an other modules use of the token.

dengert:
Sorry hit the wrong button.

dengert:
@frankmorgner you asked what should we do about this problem. @mouse07410 and I have given our options. I would like to hear from @Jakuje as he works for a large distro who may also have an option also.

frankmorgner:
I've decided to end this discussion and created https://github.com/OpenSC/OpenSC/pull/2207 instead.

PR #2216, opened 2021-02-01
IASECC: log AID selection

vjardin:
Record the selection of the AID for better debugging

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -339,6 +339,8 @@ iasecc_select_aid(struct sc_card *card, struct sc_aid *aid, unsigned char *out, 
 	unsigned char apdu_resp[SC_MAX_APDU_BUFFER_SIZE];
 	int rv;
 
+	LOG_FUNC_CALLED(card->ctx);
+
 	/* Select application (deselect previously selected application) */
 	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0x04, 0x00);
 	apdu.lc = aid->len;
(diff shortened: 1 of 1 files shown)


vjardin:
The failure does not seem to be related to this patch. We get a "failed to upload cache" for an OSX environment.

dengert:
Some cards have been knows to return 9000 for selection of any AID. (muscle cards). If this select AID is done during card matching the response should also be checked as the response should contain the AID requested.

vjardin:
Hi @dengert , please, can you be more explicit ?
For instance, when I check:
```
  OpenSC/src/libopensc$ grep -i aid *muscle*[ch]
```

I do not see any AID reference with the muscle cards. What am I missing ?

dengert:
I am not saying that this problem is the problem in this issue, but it is a problem in the same area of the code. 

The issue is a SELECT AID is a special case of a SELECT command, and some cards such as muscle and coolkey treat it as an ordinary SELECT command and can just return 9000.

The AID is specified in ISO 7816-4 section 12.2.3. and table 121 where the first nibble is "A" meaning "International registration of application providers according to ISO/IEC 7816-5".  "D", "E" and "F" are also defined, as National, Standard and Proprietary. 
Table 122 — Interindustry DOs for application identification and selection shows what should be returned. 
 
So when trying to determine if an unidentified card is one your card drive can support, you need to make sure the return from the unidentified card is what is expected from your card. This will help prevent false positives causing problems when driver thinks it has your card but does not. You won't see the problem, but other users who are not using your card or driver may. 

I note in the card driver, is saves the return, but does not check if it is what is expected.   

The grep fails because the muscle card does not use AID so does not search for one. 
The problem is in the card, not honoring iso7816-4.  

This problem has shown up in the past:
https://github.com/OpenSC/OpenSC/issues/1483
https://github.com/OpenSC/OpenSC/pull/1574

And this looks like the bug muscle code. It returns SW_NO_ERROR.

https://github.com/JavaCardOS/MuscleCard-Applet/blob/master/musclecard/src/com/musclecard/CardEdge/CardEdge.java#L321

vjardin:
Thanks for these pointers, there are some very interesting readings. But the purpose of this commit is not to fix the issues you are rising, but first, it is only to add some logs.

The fixes you are looking for should be part of a specific set of pull request: it  shall need a specific deep dive in order to get it.

Jakuje (approved):

PR #2217, opened 2021-02-01
IASECC: Add support for CPx cards

vjardin:
The French CPx Healthcare cards are designed to support the IASECC
standard.


vjardin:
(record notes)

Thanks to this patch, we can get its SDO:
```
./iasecc-tool --list-sdos 1
Using reader with a card: Microchip SEC1110 [CCID Interface] (XXXXX) 00 00
Found SDO class 1, reference 1
	contact ACLs:	F3:00:00:13:FF:00
	tries maximum:	03
	tries remaining:	03
Found SDO class 1, reference 2
	contact ACLs:	F3:FF:00:FF:FF:00
	tries maximum:	0A
	tries remaining:	0A
	usage remaining:	00:64
```

and its list of applications:
```
./iasecc-tool --list-applications
Using reader with a card: Microchip SEC1110 [CCID Interface] (XXXXX) 00 00
Application 'CPX IAS':
	AID: E828BD080F8025000001FF0010
	DDO: 4F0DE828BD080F8025000001FF0010

Application 'CPX IAS CL':
	AID: E828BD080F8025000001FF0020
	DDO: 4F0DE828BD080F8025000001FF0010
```

src/libopensc/card-iasecc.c
```diff
@@ -594,6 +606,42 @@ iasecc_init_amos_or_sagem(struct sc_card *card)
...
+		         SC_ALGORITHM_RSA_HASH_SHA256;
+
+	_sc_card_add_rsa_alg(card, 1024, flags, 0);
+	_sc_card_add_rsa_alg(card, 2048, flags, 0);
+	_sc_card_add_rsa_alg(card, 512, flags, 0);
+#else
+	_sc_card_add_rsa_alg(card, 1024, IASECC_CARD_DEFAULT_FLAGS, 0x10001);
+	_sc_card_add_rsa_alg(card, 2048, IASECC_CARD_DEFAULT_FLAGS, 0x10001);
+	_sc_card_add_rsa_alg(card, 512, IASECC_CARD_DEFAULT_FLAGS, 0x10001);
+#endif
```
Jakuje, line 637: Do you still need the `#if 1` or was it only for debugging/progress?

src/libopensc/card-iasecc.c
```diff
@@ -620,6 +668,10 @@ iasecc_init(struct sc_card *card)
 		rv = iasecc_init_amos_or_sagem(card);
 	else if (card->type == SC_CARD_TYPE_IASECC_MI)
 		rv = iasecc_init_amos_or_sagem(card);
+	else if (card->type == SC_CARD_TYPE_IASECC_CPX)
+		rv = iasecc_init_cpx(card);
+	else if (card->type == SC_CARD_TYPE_IASECC_CPXCL)
+		rv = iasecc_init_cpx(card);
```
Jakuje, line 674: I see this fillows the previous  branches, but I would rather see it in one block such as follows so one does not have to investigate what is different in each of the blocks.

```suggestion
	else if (card->type == SC_CARD_TYPE_IASECC_CPX ||
		  card->type == SC_CARD_TYPE_IASECC_CPXCL)
		rv = iasecc_init_cpx(card);
```
Maybe even better would be the use of `switch`/`case`.

Jakuje:
Thank you for the contribution. I put couple of minor comments inline. Does the `pkcs11-tool --test` work with your configured mechanisms? Does listing of the objects work with `pkcs11-tool -O`?

src/libopensc/card-iasecc.c
vjardin: no, it is was a place holder, I'll move into the git logs.

src/libopensc/card-iasecc.c
vjardin: I agree that a `switch`/`case` would be a better design, but I do not want to "change" the design logic. It was already a `if`/`else` chain. Once this pull request will be merged, I can provide an update that replaces this `if`/`else` chain with a proper `switch`/`case`.

vjardin:
@jakuje enclosed some outputs per your request:
```
./pkcs11-tool -O
Data object 2966466816
  label:          'CPX_DIR'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966451920
  label:          'CPX_ATR'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966455072
  label:          'CPX_TECH_1'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966462480
  label:          'CPX_TECH_2'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966458800
  label:          'CPX_ID_TECH'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966459472
  label:          'CPX_ID_CARD'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966446240
  label:          'CPX_DATA'
  application:    'CPX'
  app_id:         <empty>
  flags:           modifiable
Data object 2966467184
  label:          'CPX_NAME_PS'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966457680
  label:          'CPX_LANG_PS'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Data object 2966453120
  label:          'CPX_INFO_PS'
  application:    'CPX'
  app_id:         <empty>
  flags:          <empty>
Profile object 2966451040
  profile_id:          '3'
```

regarding the self tests:
```
./pkcs11-tool --test
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
error: PKCS11 function C_DigestInit failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
```
which is an expected behavior for the current support.

Jakuje:
The above commands show that you have only data objects on the card so then what is the point registering the RSA mechanisms if it does not have any keys? Or is it WIP to add support for the keys?

I am not sure what is the purpose of that card and what do you want to use it for with OpenSC.

vjardin:
> The above commands show that you have only data objects on the card so then what is the point registering the RSA mechanisms if it does not have any keys? Or is it WIP to add support for the keys?

@Jakuje right, it is WIP.

> I am not sure what is the purpose of that card and what do you want to use it for with OpenSC.

I need OpenSC to support a widely used IASECC card in France by about 1.4M+ users (healthcare users). Till now, it was using a framework that has never been upstreamed.

frankmorgner:
I think to make this more valuable, you need to support some private key and PIN operation to make this meaningful for some users.

I'm not happy with the changes to ef-atr.c, but let's discuss this in #2220...

vjardin:
> you need to check if rv > 1

done https://github.com/OpenSC/OpenSC/pull/2217/commits/b7b5cde518b29866c49b2479967804c02ed171d5#diff-d9ab7bee953b722791ab5aa28ac145b0149fec6cbd3082ce3f5b5c92b44bf953R171

thanks for pointing this issue.

src/tools/opensc-explorer.c
```diff
@@ -2176,6 +2177,10 @@ static int do_asn1(int argc, char **argv)
 		}
 	}
 
+	/* workaround when the issuer of a card does prefix the EF.ATR payload with 0x80 */
+	if ((buf[0] == ISO7816_II_CATEGORY_TLV) &&
+	    (memcmp(path.value, "\x3f\x00\x2f\x01", 4) == 0))
```
frankmorgner, line 2182: Checking for the path of EF.ATR is a good workaround. just one more corner case: if the user specified an offset, then we should not automatically increment it

frankmorgner (changes requested):
Looks OK in general, please let me know if you were able to test some private key operation.

vjardin:
> Looks OK in general, please let me know if you were able to test some private key operation.

I can perform properly a `pkcs11-tool -O --login --ping ABCD` but regarding some deep pkcs11 tests, I'd like to provide them thru a new serie of pull request. From the current analysis, for instance, if I compare OpenSC with the PKCS11 driver provided with the drivers, the `C_CreateObject()` callback should be unset (`NULL`).

Currently, with OpenSC, it is always:
```
(gdb) p *card->framework
$4 = {bind = 0x7ffff5649c00 <pkcs15_bind>, unbind = 0x7ffff5649ad0 <pkcs15_unbind>, create_tokens = 0x7ffff564e290 <pkcs15_create_tokens>, release_token = 0x7ffff56431e0 <pkcs15_release_token>, 
  login = 0x7ffff5648500 <pkcs15_login>, logout = 0x7ffff5643740 <pkcs15_logout>, change_pin = 0x7ffff56433e0 <pkcs15_change_pin>, init_token = 0x7ffff5647480 <pkcs15_initialize>, 
  init_pin = 0x7ffff56470e0 <pkcs15_init_pin>, create_object = 0x7ffff5648180 <pkcs15_create_object>, gen_keypair = 0x7ffff5644ca0 <pkcs15_gen_keypair>, get_random = 0x7ffff5644ba0 <pkcs15_get_random>}
```

while any calls to `C_CreateObject()` should return `CKR_FUNCTION_NOT_SUPPORTED`, at least, based on my current understanding. So, I feel that it is a new set of topics about "to do" or "not to do" thru another thread.

dengert:
card->framework is used by all cards. There is also the pkcs15init director that deals with making changes to a card.  There are 3 iasecc files there.   You may also need to make changes to pkcs15-iasecc.c if there is still some issue with create_object for your card.

vjardin:
I need to understand the logics that are under the hood of the pkcs15-iasecc in order to get the proper Oasis C_xyz() callbacks to be set and unset ; for instance C_CreateObject() should be unset. The more I dig, the more I feel it should be part of another pull request once we can close this current pull request.

dengert:
PKCS11 is an API. 
PKCS15 is a description of a file system. 
OpenSC and framework-pkcs15.c takes PKCS11 calls and and and uses PKCS15 routines to access the files on a token.

For cards that have a full PKCS15 filesystem, `pkcs15-<card>.c` does not do much at all. But for cards that do not or have only partial or improperly implemented PKCS15, `pkcs15-<card>.c` takes care of issues. Or 
emulate a PKCS15 file system (or subset). 

Many  `pkcs15-<card>.c` has a `sc_pkcs15emu_<card>_init_ex()` and/or a `sc_pkcs15emu_<card>_init()`.
in your case `sc_pkcs15emu_<card>_init` calls `sc_pkcs15_bind_internal` then overwrites 
p15card->ops.parse_df = _iasecc_parse_df; in the `sc_pkcs15_operations` structure.  

OpenSC pkcs15 routines call the card driver to access the card, via the the routines listed in 

`card-<card>.c` routines implement ISO7816 routines to access the token and the applet(s) on the card.
struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
Then can provide card specific function. In your `card-iasecc.c `driver see the `sc_get_drive()` that overwrites 8 of these.

I don't think you need to worry about trying to create objects for two reasons:
1. Your card should be protecting itself from hacking and only allow write to the card with some special pin or authentication with a 3DES or AES key known only to the card issuer.
2. PKCS11 allows session objects that may be on the token, or in memory, so only try an block a write in the card driver if you must, by modified  your `sc_get_drive()` to reject the operation.

src/libopensc/card-iasecc.c
```diff
@@ -594,6 +609,36 @@ iasecc_init_amos_or_sagem(struct sc_card *card)
...
+	unsigned int flags; /* TBC it is not IASECC_CARD_DEFAULT_FLAGS */
+
+	LOG_FUNC_CALLED(ctx);
+
+	card->caps = SC_CARD_CAP_RNG; /* TBC it is not IASECC_CARD_DEFAULT_CAPS */
+
+	flags = SC_ALGORITHM_RSA_PAD_PKCS1;
+	flags |= SC_ALGORITHM_RSA_RAW;
+
+	/* No signature with contactless mode */
```
Jakuje, line 625: I think what you do here does not reflect this comment. What you do allows raw RSA and RSA with PKCS1 padding. If you do not want to allow any signatures in contactless mode, you should not set flags at all and not call `_sc_card_add_rsa_alg`.

src/libopensc/card-iasecc.c
```diff
@@ -594,6 +609,36 @@ iasecc_init_amos_or_sagem(struct sc_card *card)
...
+	if (card->type == SC_CARD_TYPE_IASECC_CPXCL)
+		flags |= SC_ALGORITHM_RSA_HASH_NONE;
+	else
+		flags |= SC_ALGORITHM_RSA_HASH_SHA1    |
+		         SC_ALGORITHM_RSA_HASH_SHA256;
+
+	_sc_card_add_rsa_alg(card, 1024, flags, 0);
+	_sc_card_add_rsa_alg(card, 2048, flags, 0);
+	_sc_card_add_rsa_alg(card, 512, flags, 0);
```
Jakuje, line 634: nit: would be more readable it the values 1024, 2048 and 512 would be ordered, either ascending or descending.

src/libopensc/card-iasecc.c
```diff
@@ -916,7 +969,9 @@ iasecc_select_file(struct sc_card *card, const struct sc_path *path,
 
 	sc_print_cache(card);
 
-	if (card->cache.valid && card->cache.current_df && lpath.type == SC_PATH_TYPE_DF_NAME
+	if ((card->type != SC_CARD_TYPE_IASECC_CPX) &&
+	    (card->type != SC_CARD_TYPE_IASECC_CPXCL)  &&
```
Jakuje, line 973: This is already second time you have the same condition. I think it would be a great candidate for some helper function `inline static int iasecc_is_cpx(card)` to make the changes here less intrusive.

src/libopensc/ef-atr.c
```diff
@@ -162,8 +163,16 @@ int sc_parse_ef_atr(struct sc_card *card)
...
-	
-	rv = sc_parse_ef_atr_content(card, buf, rv);
+
+	/* Workaround: Some cards seem to have a buggy storage of the EF.ATR */
+	if ((card->type == SC_CARD_TYPE_IASECC_CPX) ||
+	    (card->type == SC_CARD_TYPE_IASECC_CPXCL)) {
+		/* Let's keep the first byte */
+		if ((rv > 1) &&
+		    (buf[0] == ISO7816_II_CATEGORY_TLV))
+		off++;
```
Jakuje, line 173: wrong indentation

vjardin: fixed: https://github.com/OpenSC/OpenSC/pull/2217/commits/756fc2e83c2244488a877bd669990189c0e2c6d1#diff-d69b111a62a36d4981628e54ba9e9606d0ddee8a16ac7542a7ffec173b15b07aR2183

src/libopensc/card-iasecc.c
vjardin: Thanks, I factorized it with a specific commit in order to track the updates. https://github.com/OpenSC/OpenSC/pull/2217/commits/656d203e195ef6216f9d00cc264fb0dcf1faeaa4

src/libopensc/card-iasecc.c
vjardin: done: https://github.com/OpenSC/OpenSC/pull/2217/commits/3867bd357a9122f78aa92bfa4a49c584e4e47b7c

src/libopensc/card-iasecc.c
vjardin: Right, currently, I am focusing on the Contact mode. The contactless mode is supposed to be close to the contactless mode, but with a different AID and no PIN code. For the time being, I'll focus on getting a full support of the contact mode.

So let's focus on this scope:
   https://github.com/OpenSC/OpenSC/pull/2217/commits/3867bd357a9122f78aa92bfa4a49c584e4e47b7c

vjardin:
The new serie includes the comments from @Jakuje

Jakuje (approved):

vjardin:
Please @frankmorgner , what's next ? See my comments - https://github.com/OpenSC/OpenSC/pull/2217#issuecomment-775459448 -, I'd like to have such update for an another set of pull request.

Jakuje:
If there will be no other comments, we will merge your changes soon. I would like Frank to have a look into the changes too. But that should not stop you from creating another PR based on this one.

frankmorgner (approved):

src/libopensc/card-iasecc.c
```diff
@@ -867,8 +923,11 @@ iasecc_select_file(struct sc_card *card, const struct sc_path *path,
 		rv = iasecc_select_mf(card, file_out);
 		LOG_TEST_RET(ctx, rv, "MF selection error");
 
-		memmove(&lpath.value[0], &lpath.value[2], lpath.len - 2);
-		lpath.len -=  2;
+		if (!iasecc_is_cpx(card)) {
+			sc_log(ctx, "iasecc_select_mf done, remove 3f00 from the path");
+			memmove(&lpath.value[0], &lpath.value[2], lpath.len - 2);
+			lpath.len -=  2;
```
frankmorgner, line 929: If you *don't* remove 3F00 from the path, then you can remove the prior iasecc_select_mf()...

frankmorgner (changes requested):

src/libopensc/card-iasecc.c
vjardin: I'll perform more tests with your advise and I'll provide some feedbacks.

vjardin: Thanks @frankmorgner  for the tips. Your are right, I can apply your advice and it works. Digging deeper on this topic, it allows to remove the patches 6a64f328499ec20e8d08f8dbcb1c77157b9f7065 and a60f4988f5fc108bae29945105b1e6e7b95b2331 from this serie too. I'll build an update of this pull request.

dengert:
With BER vs DER I don't have a problem,  but the lax bit string change is making some assumptions. 
What is the specific case you were trying to address? 

Some examples that would pass the lax bit string code.

- bit string was meant to be 8 bits of all ones  `03 02 00 FF` and creator created it as `03 01 FF`. This is invalid as len <1, but the lax code change would return 0x00. 
- bit string was meant to be 16 bits of all ones `03 03 00 FF FF` and was created as `03 02 FF FF` This is invalid, as first byte is at most 07. the lax code would return 1, because first FF would say 7 unused bits so FF would be 80 and reordered as 01.  

They are probable other examples. 

I would rather have these caught, and addressed in code changes, then to return the wrong bits. 

They are probable other examples. 

If there is a specific error created by the card issuer, I would rather see the change added to the card driver to handle just this error. If this code is in asn1.c, the variable "strict" should be renamed to "asn1_flags" as ulong with bits to indicate what specific error is acceptable. so asn1_flags == 0 means follow ASN.1 strict standards, and something like `asn1_flags & C_ASN1_LAX_BITSTRING_MISSING_LEADING_BYTE`  address the examples above. 

BER vs DER could also be a bit too. DER is really only needed if data is signed so signatures are always the same.

vjardin:
> With BER vs DER I don't have a problem, but the lax bit string change is making some assumptions.
> What is the specific case you were trying to address?

I do not understand your/@dengert comments: some examples of content parsing are provided into the traces of the issue #2224, does it help ?

dengert:
What I am saying is how do you know what bits should bein the string?   87a35b170b8571db3a274cfda26aab838f097dfb trades off one or two errors for a different error. Length has to be 2 or more, and first has to be <=7. 

Creator of the string sets first byte, which should between 0 and 7 saying how many unused bits in last byte. But  what did creator really mean if they don't understand how to code a bit string?

Form the two bit string in error, the firsts is 03 01 08` and second is `03 02 08 00` as best I can tell,
In both cases your fix interprets these a zero length bit string. But was that intent of the creator?

The creator may have left off the first byte and should have said   `03 02 00 08` to mean "BIT STRING (8 bit) 00001000"  and in second  as `03 03 00 08 00` "BIT STRING (16 bit) 0000100000000000"

And maybe the creator meant to use 02 for integer or 04 for octet string? I don't know what you are decoding.
You can look at the source code. 

So you need to understand what the creator of the error meant to say and apply this only to this one case in your driver.
You don't let errors in other cards asn1 that would have been caught to slip by.

vjardin:
Thanks, I get your point now:

> So you need to understand what the creator of the error meant to say and apply this only to this one case in your driver.
> You don't let errors in other cards asn1 that would have been caught to slip by.

my logic is another path/way. Before the the commit 8ddfafe057fa4960fed5c427bea2c48d68e67655 and c1814571bd09440e09b7c32285367d580d65650a, it was working fine for those cards. I mean that some cards are not "strict" with the ITU recommendations. We could blame them, but there are 1.4M which are spread in France :( for a decade.

So, the design of this flexibility is the same than the commit aaa302ca3576607c7bf3b80730113b588da60d8d which introduces a concept of "non strict" integer parsing.

vjardin:
> OpenSC asn1 parsing has a way to use SC_ASN1_CALLBACK so caller can get control or pass in another SC_ASN1_* bit. It might be possible to add this in `asn1_decode_p15_object` for this one bad bit string to set the two flags for this one card.
> 
> The least I would do is stop calling this "lax" and call it something like "know_error" and then fix up the flags in the driver.

Good idea, I'll try to add such call back capabilities that would be specific to the French IASECC healthcare cards. Please, do you have some helpers/pointers into the code in order to help me with the support of such specific callback in order to support such workaround on some specific cards ?

For instance, from the code, I do not see any other drivers using such callback:
```
src/libopensc/pkcs15-cert.c:		{ "subjectPublicKeyInfo",SC_ASN1_CALLBACK, SC_ASN1_TAG_SEQUENCE | SC_ASN1_CONS, 0, sc_pkcs15_pubkey_from_spki_fields,  &pubkey },
src/libopensc/pkcs15-pubkey.c:			{ "subjectPublicKeyInfo", SC_ASN1_CALLBACK, SC_ASN1_TAG_SEQUENCE | SC_ASN1_CONS, 0, sc_pkcs15_pubkey_from_spki_fields, &pubkey},
src/libopensc/asn1.c:	case SC_ASN1_CALLBACK:
```

dengert:
https://github.com/OpenSC/OpenSC/issues/2224#issuecomment-775103434 did list the hexdump  of the two objects,
with `03 01 08` and `03 02 08 00`  as the bit strings in error.  

I would propose adding a SC_ASN1_BIT_FIELD_XXXX  and SC_ERROR_XXXX 
modify asn1.c to add it like 

1548         case SC_ASN1_BIT_FIELD:
1549                 if (parm != NULL)
1550                         r = decode_bit_field(obj, objlen, (u8 *) parm, *len);

but if r ==  SC_ERROR_XXXX  print error message, and set r = len (or 0)?
then modify you patch to return SC_ERROR_XXXX if  `03 01 80` or `03 02 08 00`

and in asn1.c:
1366         { "flags", SC_ASN1_BIT_FIELD_XXXX, SC_ASN1_TAG_BIT_STRING, SC_ASN1_OPTIONAL, NULL, NULL },

So only in an object commonObjectAttributes if the flags are `03 01 80` or `03 02 08 00` will it be acceptable.

vjardin:
> Right, callback will not work.

Thanks for your ack't.

> 
> In OpenSC, if sc_context is passed, a routine can write to debug log.
> If sc_card is passed, the card type can be tested.
> Problem is most asn.1 routines don't have access to these, but can set new error codes, like SC_ERROR_XXXX can be tested, printed, and modified in upper levels.

My concern is :  the error SC_ERROR_INVALID_ASN1_OBJECT gets up to sc_asn1_print_bit_string(), which is a 'static void' function (so it does not return any SC_ERROR_XXXX) to the upper layers. So, I am not able to apply your advises.

As you stated, the sc_card/sc_context are not "useable" within the asn1 code.

vjardin:
Big thanks, I am giving a try. I'll provide some analysis and traces ASAP.

Regarding your patch, why do you add:
```
+++ b/src/tools/pkcs15-tool.c
@@ -262,6 +262,9 @@ static void print_common_flags(const struct sc_pkcs15_object *obj)
 		}
 	}
 	printf("\n");
+
+	if (obj->user_consent)
+		printf("\tUser Consent   : %d\n", obj->user_consent);
 }
```
it seems related to another purpose, doesn't it ?

dengert:
Yes the pkcs15-tool.c was related to another bug, but it prints out user_consent if found for an object. it was not committed when I switched to master.
It would have made it easier to see the problem

vjardin:
> Big thanks, I am giving a try. I'll provide some analysis and traces ASAP.

Done, see: https://github.com/OpenSC/OpenSC/pull/2217/commits/a0ca9ef976ab6cb0ce28662404d729941e539076

Your suggestion was very nice, thank you.

dengert:
You will need to change   decode_bit_string  to return SC_ERROR_ASN1_BAD_BITSTRING  if the  "strict" code found  one of the bad  encodings 03 01 08 or 03 02 08 00  Any other bad encoding is something we don't know about so return SC_ERROR_INVALID_ASN1_OBJECT.

frankmorgner:
I don't like adding more (undocumented) flags globally, because that makes the code even harder to read. In my opinion, having a relaxed parsing as default is just good enough.

Actually, instead of introducing new flags, new error codes, we could add a warning to the log that the ASN.1 encoding was incorrect given a strict interpretation of the standard and just continue parsing. Obviously, card manufacturers are making a lot of such errors, so we must cope with this type of problem. Yes, we may trade one error for another, as Doug said, but either way, the user observes an error and will most likely create a log. The latter will then show the exact trace of the error including a possible warning for badly encoded ASN.1...

Let's wait for @Jakuje 's comment...

frankmorgner:
Maintaining software is difficult and implementation has always multiple ways to solve a problem... Thanks for your patience!

dengert:
Lets take a step back here. 
@vjardin is trying to use French Healthcare cards, where  PKCS\#15 commonObjectAttributes of the Flags does not follow ASN.1 standards imposed by 2 recent patches to asn1.c in enforce the bit string standards. He found this problem using opensc-explorer. There may be other problems like this too.

A solution to the opensc-explorer where is is parsing the ASN.1 and would like to skip errors and continue is one thing. But fixing up bad ASN.1 is another, as information may be lost or reported incorrectly. 

OpenSC needs to work with existing cards with the bad ASN.1 which is loaded via pkcs15.c. There are no hooks, for a card driver or a pkcs15-card.c to ask to "fix" the error while pkcs15 routines loads all the pkcs15 data. 

We know that the 2 enforcement patches may cause existing cards to stop working if these are included in next release. 
To avoid or mitigate problems, we should also have a way to turn off these enforcement patches from opensc.conf and let users have a way to fallback if needed, while we fix the problem in following release.  

That said, if we know there is an ASN.1 problem in a specific card in a specific ASN.1 encoding and maybe only in a specify object, we would need a way to let card driver or pkcs15 driver fix the problem. 

But OpenSC is not setup to allow a card driver to control the reading of PKCS\#15 files with bad ASN.1.  Decode routines don't ever get passed `sc_context_t`, or sc_card_t. The strict/lax flag would work if it can be set by a card driver for the specific card. i.e. with PKCS\#11 lax should be set for only that one card.  

So in other words, it looks to difficult to do the SC_ERROR_ASN1_BAD_BITSTRING.

dengert:
The one thing the about enforcing the "correct" ASN.1 has shown that this card has invalid ASN.1 for the  commonObjectAttributes Flag bit string. 

So far @vjardin has shown he can get opensc-explorer to parse the invalid bitstring, but the value of the bits is not what was intended. He still has to show he can get it to work with the incorrect bits, or how to set the bits some other way. 

@vjardin  what does `./pkcs11-tool -O --login` show?
Can you try these commands with and without `--verify-pin:` `pkcs15-tool -D',  `pkcs15-tool -c`, `pkcs15-tool -k` and `pkcs15-tool --list-public-keys` 
`private` object flag is one that appears to not be set correctly, and would show up only when PIN was verified.

mouse07410:
> The one thing the about enforcing the "correct" ASN.1 has shown that this card has invalid ASN.1 for the commonObjectAttributes Flag bit string.

Great. And how did it help us? Will we tell the vendor to fix the format?

dengert:
I hope so. And we know enough to look at other bit stings on these cards that may also be invalid.
But not clear if the French Healthcare card issuers will listen.

mouse07410:
Yubico listened, and their PIV applet is one of the closest to the standard...

Would French _government supplier_ listen? I rather doubt - but let's see.

vjardin:
> @vjardin what does `./pkcs11-tool -O --login` show?
> Can you try these commands with and without `--verify-pin:` `pkcs15-tool -D', `pkcs15-tool -c`, `pkcs15-tool -k`

see the enclosed logs
[pkcs15-tool-c+verify-pin.log.gz](https://github.com/OpenSC/OpenSC/files/6119363/pkcs15-tool-c%2Bverify-pin.log.gz)
[pkcs15-tool-D.log.gz](https://github.com/OpenSC/OpenSC/files/6119364/pkcs15-tool-D.log.gz)
[pkcs15-tool-D+verify-pin.log.gz](https://github.com/OpenSC/OpenSC/files/6119365/pkcs15-tool-D%2Bverify-pin.log.gz)
[pkcs15-tool-k.log.gz](https://github.com/OpenSC/OpenSC/files/6119366/pkcs15-tool-k.log.gz)
[pkcs15-tool-k+verify-pin.log.gz](https://github.com/OpenSC/OpenSC/files/6119367/pkcs15-tool-k%2Bverify-pin.log.gz)
[pkcs15-tool-c.log.gz](https://github.com/OpenSC/OpenSC/files/6119368/pkcs15-tool-c.log.gz)

vjardin:
> I hope so. And we know enough to look at other bit stings on these cards that may also be invalid.

It seems that there are only 2 issues, for the 2 flags of these 2 files.

> But not clear if the French Healthcare card issuers will listen.

Even if they "listen", it is too late the million of cards are already shipped for many years. I prefer that we live with a proper workaround that fits all the cases.

mouse07410:
> Does Yubico supports the IASECC standard ???

No, of course not.

It means that Yuhikeys had problems in their PIV implementation, we pointed that out to them, and Yubico improved their implementation greatly.

vjardin:
Please what's next ? How to tackle it ?

dengert:
What's next? Not sure. 
 OpenSC has been developed by card vendors, people interested in using (or need to use) their work or government issued cards on non-windows systems, linux distros wanting to support many smartcards, open source developers wanting user controllable security, delelopers wanting to develeope a better applet, programmers interested in a challenge and I may have missed some other too.  None of these people have more then a handful of cards for testing.

French CPx Healthcare cards are not of much interest outside of France. It looks like a pkcs15 type card, but we have found it does not encode a bit string properly.  @Jakuje in  https://github.com/OpenSC/OpenSC/pull/2217#issuecomment-772697956 said it looks like the card has only objects and no keys or certificates. That could be true but it looks like the bit string issue caused pkcs15 routines to not find the keys or certificates.  There may be other ANS.1 or invalid PKCS15 structures that will also cause problems.  

As with other government issues cards, they most likely did not publish the code or documents about the cards. Without these you may have to resort to PC/SC or USB tracing of APDUs.  You may want to look at #2244 where I was talking to @the-docta about how to do some traces, and do simple command line issuing of APDUs.

There is no general  consensus right now on how to handle the ASN1 bit string bug. Either drop the commits the enforce it, or make same major changes in the asn1 parsing code that will effect every card. 

So if you wish to pursue this, I would suggest for your testing,  revert the two commits and see how far you can get. If you can get the card working that is a case to solve the bit string problem.

If this card is as simple as I would suspect, with 1 or 2 certificates and private keys on the card and RSA only, it might be possible write a new driver, that emulates PKCS15 and avoids any unnecessary parsing of bas ASN1 data.

Personally I fall into the "interested is using their work/government issued card on linux" and "programmers interested in a challenge".  But with no one else is interested in  CPx Healthcare card, I don't want to spend much time on it.

So find some documentation on the card, buy (or find) copy of ISO-7816-4,ISO-7816-8, PKCS15 (or ISO-7816-15) and get some PC/SC traces of a card on windows.

frankmorgner:
please leave this PR as is for now.

I suggest to remove a0ca9ef976ab6cb0ce28662404d729941e539076 from this PR and merge it. Always having a non-strict parsing is in line with what we're currently using with `sc_asn1_decode_integer`.

When we're done with this PR, we should discuss elsewhere whether we want to continue adding these type of flags for ASN.1 parsing...

vjardin:
> please leave this PR as is for now.
> 
> I suggest to remove [a0ca9ef](https://github.com/OpenSC/OpenSC/commit/a0ca9ef976ab6cb0ce28662404d729941e539076) from this PR and merge it. Always having a non-strict parsing is in line with what we're currently using with `sc_asn1_decode_integer`.

done. Thanks for the advises.

Jakuje (approved):
sorry for a delay. Catching up with the emails after last week. I think this PR is fine to merge

frankmorgner (approved):

vjardin:
thank you.

PR #2219, opened 2021-02-03
Weekly dose of fixes for issues reported by oss-fuzz

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -159,7 +159,7 @@ static int cardos_have_2048bit_package(sc_card_t *card)
 	sc_apdu_t apdu;
         u8        rbuf[SC_MAX_APDU_BUFFER_SIZE];
         int       r;
-	const u8  *p = rbuf, *q;
+	const u8  *p = rbuf, *q, *pp;
 	size_t    len, tlen = 0, ilen = 0;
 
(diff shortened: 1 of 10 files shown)


vjardin:
LGTM (code review).

src/libopensc/card-cardos.c
```diff
@@ -175,10 +175,10 @@ static int cardos_have_2048bit_package(sc_card_t *card)
 		return 0;
 
 	while (len != 0) {
-		p = sc_asn1_find_tag(card->ctx, p, len, 0xe1, &tlen);
+		pp = sc_asn1_find_tag(card->ctx, p, len, 0xe1, &tlen);
 		if (p == NULL)
```
frankmorgner, line 179: `if (pp == NULL)`?

frankmorgner (changes requested):

src/libopensc/card-cardos.c
Jakuje: Right. Good catch. Will fix.

Jakuje:
@vjardin I see you working on iasecc now. The top commit is fixing one issue reported by oss-fuzz when the ACLs returns empty zero-length data, there was buffer overflow. Can you check my change is not breaking anything for you?

Jakuje:
(I bundled up couple of new fixes for coverity too)

Jakuje:
@frankmorgner did you manage to re-review this PR?

frankmorgner (approved):

PR #2221, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.


frankmorgner (approved):
Thank you, looks good

PR #2230, opened 2021-02-10
fix potential buffer overflow.

ihsinme:
elsewhere, record size is tightly controlled.
in this place it seemed to me that no.

I suggest a quick and easy way to fix it.

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -585,7 +585,7 @@ static void initialize_uid(sc_reader_t *reader)
 		if (SC_SUCCESS == pcsc_transmit(reader, &apdu)
 				&& apdu.sw1 == 0x90 && apdu.sw2 == 0x00) {
 			reader->uid.len = apdu.resplen;
-			memcpy(reader->uid.value, apdu.resp, reader->uid.len);
+			memcpy(reader->uid.value, apdu.resp, MIN(sizeof(reader->uid.value), reader->uid.len));
 			sc_log_hex(reader->ctx, "UID",
 					reader->uid.value, reader->uid.len);
 		} else {


src/libopensc/reader-pcsc.c
```diff
@@ -585,7 +585,7 @@ static void initialize_uid(sc_reader_t *reader)
 		if (SC_SUCCESS == pcsc_transmit(reader, &apdu)
 				&& apdu.sw1 == 0x90 && apdu.sw2 == 0x00) {
 			reader->uid.len = apdu.resplen;
-			memcpy(reader->uid.value, apdu.resp, reader->uid.len);
+			memcpy(reader->uid.value, apdu.resp, MIN(sizeof(reader->uid.value), reader->uid.len));
```
frankmorgner, line 588: uid.len is 0 at this point, so your fix won't work. Also, there are some corner cases that aren't catched

frankmorgner (changes requested):
Thanks for noticing this issue. However, I've proposed a different approach in https://github.com/OpenSC/OpenSC/pull/2232/ , which should hopefully catch all corner cases...

src/libopensc/reader-pcsc.c
Jakuje: This is missing the adjustment of the `reader->uid.len` value if we get longer response. I like the #2232 better as it handles the issue outside of the block.

ihsinme:
your fix looks like other such places in your code.
this is probably the best solution at this stage.
in any case, the main thing is to get a more secure version.

frankmorgner:
thank you

ihsinme:
Patch reward submission 408: OpenSC/OpenSC buffer overflow for OpenSC/OpenSC

PR #2234, opened 2021-02-17
Minidriver RSA-PSS signing not working

llogar:
I am using a somewhat modified version of IsoApplet. Up till now it worked fine. However recently I stumbled upon a web site that
forces a client cert auth with RSA-PSS. And (at least on windows, using minidriver) it didn't work. It looks to me, that it's a bug
in the PSS support code in minidriver, as I cannot find any place where a MGF1 padding scheme is specified. And since none is specified signing fails. This patch fixes this. It assumes, that the same hash is used for hashing and padding.


Jakuje (approved):

frankmorgner (approved):

frankmorgner:
thank you!

PR #2241, opened 2021-02-23
Fix for running after build with gcc11

Jakuje:
This made most of the applications crashing in Fedora 34 when
smart card was plugged in.

The suggested patch makes the code path more obvious for gcc to
handle.

https://bugzilla.redhat.com/show_bug.cgi?id=1930652

##### Checklist

- [X] PKCS#11 module is tested (roughly)
- [ ] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1091,7 +1091,7 @@ if test "${enable_pedantic}" = "yes"; then
 	CFLAGS="-pedantic ${CFLAGS}"
 fi
 if test "${enable_strict}" = "yes"; then
-	CFLAGS="-Wall -Wextra -Wno-unused-parameter -Werror ${CFLAGS}"
+	CFLAGS="-Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing=2 ${CFLAGS}"
 fi
 
 AC_CONFIG_FILES([
--- a/src/libopensc/pkcs15-westcos.c
+++ b/src/libopensc/pkcs15-westcos.c
@@ -124,18 +124,17 @@ static int sc_pkcs15emu_westcos_init(sc_pkcs15_card_t * p15card)
 		struct sc_pkcs15_pubkey_info pubkey_info;
 		struct sc_pkcs15_object pubkey_obj;
 		struct sc_pkcs15_pubkey *pkey = NULL;
+		sc_pkcs15_cert_t *cert = NULL;
+
 		memset(&cert_info, 0, sizeof(cert_info));
 		memset(&cert_obj, 0, sizeof(cert_obj));
 		cert_info.id.len = 1;
 		cert_info.id.value[0] = 0x45;
 		cert_info.authority = 0;
 		cert_info.path = path;
-		r = sc_pkcs15_read_certificate(p15card, &cert_info,
-					       (sc_pkcs15_cert_t
-						**) (&cert_obj.data));
+		r = sc_pkcs15_read_certificate(p15card, &cert_info, &cert);
+		cert_obj.data = (void *) cert;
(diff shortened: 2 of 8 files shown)


Jakuje:
Updated the title -- the build passed just fine without any warning, but it was crashing in runtime.

frankmorgner:
Thank you!

Is there some mechanism which could warn us when the strict aliasing leads to an optimisation that could be problematic?

How did you hunt down the other pieces of code with this problem? Maybe, we have similar stuff that's not related to `pkcs15_prkey_object`, but still has its stored value accessed other than by an lvalue of an allowable type.

Jakuje:
> Thank you!
> 
> Is there some mechanism which could warn us when the strict aliasing leads to an optimisation that could be problematic?

Jakub Jelinek suggested the `-Wstrict-aliasing=2` should warn about these:

https://bugzilla.redhat.com/show_bug.cgi?id=1930652#c17

Just re-running with these shows more issues like that for example:
```
pkcs15-westcos.c: In function 'sc_pkcs15emu_westcos_init':
pkcs15-westcos.c:135:12: error: dereferencing type-punned pointer might break strict-aliasing rules [-Werror=strict-aliasing]
  135 |       **) (&cert_obj.data));
      |           ~^~~~~~~~~~~~~~~
```
Let me update the PR with the other fixes.

> How did you hunt down the other pieces of code with this problem?

These were all having the same pattern and are on the most common code paths. I mostly tried to rerun the minimal test case using `pkcs11-tool` with virtual smart card. After that I run couple of more complicated tests and the issue did not show anymore so at least it fixes the issues on the most common code paths. But as mentioned above, it will still require some more love.

> Maybe, we have similar stuff that's not related to `pkcs15_prkey_object`, but still has its stored value accessed other than by an lvalue of an allowable type.

Yeah. Let me update the PR with more fixes and probably add the flag to our CIs ...

frankmorgner:
Yes, good plan, thank you!

Jakuje:
Seems like there is couple of more issues on top of what is fixed with the above commits. Couple of them show only on 32b systems, and couple of them are reported by Travis (with GCC 5!) so let me give it one more try.

Jakuje:
I am not sure what is the gcc meaning with the following errors:
```
reader-pcsc.c: In function ‘detect_reader_features’:
reader-pcsc.c:1162:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  pcsc_tlv = (PCSC_TLV_STRUCTURE *)feature_buf;
  ^
reader-pcsc.c:1227:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     PIN_PROPERTIES_STRUCTURE_v5 *caps = (PIN_PROPERTIES_STRUCTURE_v5 *)rbuf;
     ^
reader-pcsc.c:1236:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     PIN_PROPERTIES_STRUCTURE *caps = (PIN_PROPERTIES_STRUCTURE *)rbuf;
     ^
```
These assignments look totally fine for me and also do not raise issues in new gcc. I will try to get some answers from gcc developers if they will be helpful before merging this.

frankmorgner:
I love FOSS to learn new things every day. Here's an excelent answer, which is exactly about the issue which puzzled you and me: https://stackoverflow.com/questions/98650/what-is-the-strict-aliasing-rule/99010#99010

Jakuje:
Thanks! I pushed the new commits fixing the pcsc issues detected by the old gcc. Lets see if there will pop up something else.

PR #2246, opened 2021-03-01
oberthur: Handle more memory issues during initialization

Jakuje:
This will hopefully finalize dealing with the oberthur driver.

Unfortunately, I do not have any oberthur cards to verify it still works but all the sorts of issues mentioned in oss fuzz in similar functions should be addressed now.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):
Thanks for taking the time

PR #2248, opened 2021-03-02
pkcs11-global: Obey to the tokenPresent parameter on C_GetSlotList

3v1n0:
Since commit dba0f56 the tokenPresent parameter is ignored in case the
slot has been already seen.

This breaks the API expectations as we may return a slot that has no
token inserted.

So, only consider the SC_PKCS11_SLOT_FLAG_SEEN if tokenPresent is false

---

Tested with:

CNS card

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -519,9 +519,10 @@ CK_RV C_GetSlotList(CK_BBOOL       tokenPresent,  /* only slots with token prese
 		 * - any slot with token;
 		 * - any slot that has already been seen;
 		 */
-		if ((!tokenPresent && slot->reader != prev_reader)
-				|| (slot->slot_info.flags & CKF_TOKEN_PRESENT)
-				|| (slot->flags & SC_PKCS11_SLOT_FLAG_SEEN)) {
+		if ((!tokenPresent &&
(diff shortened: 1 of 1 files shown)


Jakuje:
Could thisbe related to #2199 ?

3v1n0:
> Could thisbe related to #2199 ?

Mh, not sure... I didn't see the calling code, but this is not causing toubles to callers waiting for a card. But it indeed does create troubles if you assume that if you request only slots with a token to be returned as in the case in the [gnome-settings-daemon port to p11-kit](https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/merge_requests/208).

Indeed we could filter this out too, but I'm quite sure the module is wrong here. Logic can probably change though as the current module implementation is a bit convoluted.

Jakuje:
Yeah. This was not to question the correctness of the patch, but the fact that we already had one issue with the last version which looked similarly and was traced in the end to different changeset that ended reverted so I would certainly like @frankmorgner to have a look.

frankmorgner (approved):
It doesn't look like it's related to #2199, but since we still haven't found the exact place causing this problem, I'm not totally sure.  

This PR looks good, thank you!

PR #2260, opened 2021-03-16
More fixes related to strict aliasing with GCC 8

Jakuje:
Thanks @popovec for reporting this. I took the liberty to use better naming for variables and slightly reorder the lines. Additionally, I would like to try to fix the osx test failure investigating what is going on there.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):

vjardin:
Note: mandatory for Ubuntu 20.04:
```
gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)
```

Jakuje:
(note that this is an issue only with strict builds and can be disabled with `./configure --disable-strict`. But we want to merge this anyway)

PR #2261, opened 2021-03-17
oberthur: fixed Heap-buffer-overflow

frankmorgner:
fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32149

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-oberthur.c
+++ b/src/libopensc/pkcs15-oberthur.c
@@ -616,12 +616,14 @@ sc_pkcs15emu_oberthur_add_pubkey(struct sc_pkcs15_card *p15card,
 	offs += 2 + len;
 
 	/* ID */
-	if (offs > info_len) {
+	if (offs + 2 > info_len) {
 		free(info_blob);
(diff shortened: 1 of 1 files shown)


frankmorgner:
This should fix one of the two remaining security related issues left. @Jakuje , could you have a look at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28855 ?

Jakuje:
> This should fix one of the two remaining security related issues left. @Jakuje , could you have a look at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28855 ?

This one says it is not reproducible according to the description. It comes from coolkey so I assume it is something that we fixed when things went crazy in there.

In the meantime, there landed one more.

This change looks good to me.

Jakuje (approved):

Jakuje:
The fix for the other new issue is here: https://github.com/Jakuje/OpenSC/commit/0f660f3b

Tested that both of them are gone with the fixes.

Issue #2262, opened 2021-03-17
IASECC: error/regression ACL since oss-fuzz fix


vjardin:
Since the commit ae1cf0be90396fb6c0be95829bf0d3eecbd2fd1c, when we get:
```
./opensc-explorer -r 0
OpenSC Explorer version 0.21.0
unable to select MF: Requested object not found
```
the logs are:
```
 [opensc-explorer] reader-pcsc.c:324:pcsc_transmit:
Outgoing APDU (5 bytes):
00 C0 00 00 20 ....
 [opensc-explorer] reader-pcsc.c:242:pcsc_internal_transmit: called
 [opensc-explorer] reader-pcsc.c:297:pcsc_internal_transmit: returning with: 0 (Success)
 [opensc-explorer] reader-pcsc.c:333:pcsc_transmit:
Incoming APDU (34 bytes):
62 1E 82 01 38 83 02 3F 00 84 08 80 25 00 00 01 b...8..?....%...
FF 01 00 A1 08 8C 06 7A FF FF FF FF 45 8A 01 05 .......z....E...
90 00                                           ..
 [opensc-explorer] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
 [opensc-explorer] apdu.c:537:sc_transmit: returning with: 0 (Success)
 [opensc-explorer] card.c:523:sc_unlock: called
 [opensc-explorer] apdu.c:499:sc_get_response: returning with: 0 (Success)
 [opensc-explorer] apdu.c:537:sc_transmit: returning with: 0 (Success)
 [opensc-explorer] card.c:523:sc_unlock: called
 [opensc-explorer] card-iasecc.c:1109:iasecc_select_file: iasecc_select_file() apdu.resp 32
 [opensc-explorer] card-iasecc.c:1113:iasecc_select_file: apdu.resp 62:1E:82...
 [opensc-explorer] card-iasecc.c:1213:iasecc_process_fci: called
 [opensc-explorer] card-iasecc.c:1216:iasecc_process_fci: processing FCI: 0x6F tag (nil)
 [opensc-explorer] card-iasecc.c:1224:iasecc_process_fci: processing FCI: 0x62 tag 0x7fff5608f092
 [opensc-explorer] card-iasecc.c:1226:iasecc_process_fci:   FCP length 30
 [opensc-explorer] iso7816.c:366:iso7816_process_fci:   shareable: no
[...]

PR #2264, opened 2021-03-17
iasecc: Fix ACLs support when length is 6

vjardin:
ACLs with length < 6 are allowed, depending on the mask of the offset 0.
For instance, when the offset 0 is 0x7B, then length can be up to 7
when the offset 0 is 0x7A, the loop was never performing any access to
the acls[7] thanks to:
  if (!(mask & acls[0]))
    continue;

However, the oss-fuzz tools cannot guess such behavior. So let's have a
robust boundary check.

Fix: issue #2262
Fix: ae1cf0be90396f 'Prevent stack buffer overflow when empty ACL is returned'

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1189,7 +1189,8 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 {
 	struct sc_context *ctx = card->ctx;
 	size_t taglen;
-	int rv, ii, offs;
+	size_t offs;
+	int rv, ii;
 	const unsigned char *acls = NULL, *tag = NULL;
 	unsigned char mask;
 	unsigned char ops_DF[7] = {
@@ -1248,7 +1249,7 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 		if (!(mask & acls[0]))
 			continue;
 
-		sc_log(ctx, "ACLs mask 0x%X, offs %i, op 0x%X, acls[offs] 0x%X", mask, offs, op, acls[offs]);
+		sc_log(ctx, "ACLs mask 0x%X, offs %"SC_FORMAT_LEN_SIZE_T"u, op 0x%X, acls[offs] 0x%X", mask, offs, op, acls[offs]);
 		if (op == 0xFF)   {
 			;
 		}
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1246,13 +1246,13 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 	for (ii = 0; ii < 7; ii++, mask /= 2)  {
 		unsigned char op = file->type == SC_FILE_TYPE_DF ? ops_DF[ii] : ops_EF[ii];
 
-		if (!(mask & acls[0]))
-			continue;
 		/* avoid any access to acls[offs] beyond the taglen */
 		if (offs >= taglen) {
 			sc_log(ctx, "Warning: Invalid offset reached during ACL parsing");
 			break;
 		}
+		if (!(mask & acls[0]))
+			continue;
 
 		sc_log(ctx, "ACLs mask 0x%X, offs %"SC_FORMAT_LEN_SIZE_T"u, op 0x%X, acls[offs] 0x%X", mask, offs, op, acls[offs]);
 		if (op == 0xFF)   {
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1188,9 +1188,8 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 		 const unsigned char *buf, size_t buflen)
 {
 	struct sc_context *ctx = card->ctx;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-iasecc.c
```diff
@@ -1232,7 +1232,7 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 	else
 		acls = sc_asn1_find_tag(ctx, buf, buflen, IASECC_DOCP_TAG_ACLS_CONTACT, &taglen);
 
-	if (!acls || taglen < 7)   {
+	if (!acls)   {
```
Jakuje, line 1235: I think that this will still crash for acls that have  exactly one byte. This condition will be skipped and on line 1251 you will be accessing acls[1] unconditionally.

src/libopensc/card-iasecc.c
```diff
@@ -1270,6 +1270,9 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 		}
 
 		offs++;
+		/* avoid any access to acls[offs] beyond the taglen */
+		if (offs >= (int)taglen)
+			break;
```
Jakuje, line 1275: I would probably add this condition to the beginning of the cycle (to avoid the issue mentioned above) and add some log message which would simplify debugging if things would go wrong.

Jakuje (changes requested):

src/libopensc/card-iasecc.c
frankmorgner: exactly! This check belongs into line 1247, before anything from acls is accessed. here, offs is 1 in the very first run and then increased.

frankmorgner (changes requested):
please also change offs and ii to type size_t

src/libopensc/card-iasecc.c
vjardin: Done: the warning is only for the case of length 0, which would be strange because the mask should always be the first byte.

vjardin:
> please also change offs and ii to type size_t

thanks for the comments, done f8d33df

src/libopensc/card-iasecc.c
```diff
@@ -1245,10 +1246,17 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 	for (ii = 0; ii < 7; ii++, mask /= 2)  {
 		unsigned char op = file->type == SC_FILE_TYPE_DF ? ops_DF[ii] : ops_EF[ii];
 
+		/* avoid any access to acls[offs] beyond the taglen */
+		if (offs >= taglen) {
+			if (taglen == 0)
+				sc_log(ctx, "Warning: ACL is empty, it is valid but surprising");
```
Jakuje, line 1252: I think this log should show up also if `taglen != 0` -- this means inconsistency in the returned data (the first byte is not matching the content bytes following if I understand what is done here). It is probably up to you whether it should be just log or error (i am fine with both).

vjardin: No, other lengths are valid. It is tight to the mask. For instance, when the length is 6, the mast is 7A, see the logs from the issue #2262

Jakuje: Other lengths are valid, but only with valid mask! But here, we are catching buffer overruns, which are not valid. If we have other length, lets say 4 and mask indicating the length should be 6, we end here too, which is error and should be logged (at least).

If you have valid length and valid mask, you should not be ending here, given that you move this block after the block continue statement below, which makes sure the mask is taken into the effect.

vjardin: I got your point, see: https://github.com/vjardin/OpenSC/commit/fed485b723c436e1ec08d4410b6149feb9e61294

Jakuje: I had in my mind something simpler like this:
```
		if (!(mask & acls[0])) {
			continue;
		}
		/* avoid any access to acls[offs] beyond the taglen */
		if (offs >= taglen) {
			sc_log(ctx, "Warning: Invalid offset reached during ACL parsing");
			break;
		}
```
The first part is still safe with invalid `offs` and simplifies the conditions in the second block. But your version should work the same way if I see right.

vjardin: I agree, I prefer your suggestion, it avoids too many if() else logic. See https://github.com/vjardin/OpenSC/commit/3b2581587de1c4f5007c12def8568af3c760aa4d

src/libopensc/card-iasecc.c
```diff
@@ -1247,8 +1248,13 @@ iasecc_process_fci(struct sc_card *card, struct sc_file *file,
 
 		if (!(mask & acls[0]))
 			continue;
+		/* avoid any access to acls[offs] beyond the taglen */
+		if (offs >= taglen) {
```
frankmorgner, line 1252: sorry, but you need to check this *before* accessing the first byte of acls in line 1248!

frankmorgner (changes requested):

frankmorgner:
I've added my suggestions on top of your changes...

vjardin:
> I've added my suggestions on top of your changes...

thanks, LGTM

PR #2268, opened 2021-03-21
pkcs11-tool: symmetric encrypt/decrypt

popovec:
supported mechanisms: AES-ECB, AES-CBC, AES-CBC-PAD
support for initialization vector

Tested - softhsm2, (test-pkcs11-tool-sym-crypt-test.sh)

	modified:   doc/tools/pkcs11-tool.1.xml
	modified:   tests/Makefile.am
	new file:   tests/test-pkcs11-tool-sym-crypt-test.sh

### DIFF
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -387,6 +387,12 @@
 				</varlistentry>
 				<varlistentry>
 					<term>
+						<option>--wrap</option>,
+					</term>
+					<listitem><para>Wrap key.</para></listitem>
+				</varlistentry>
+				<varlistentry>
+					<term>
 						<option>--derive</option>,
 					</term>
 					<listitem><para>Derive a secret key using another key and some data.</para></listitem>
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -375,6 +375,12 @@
 
 				<varlistentry>
 					<term>
+						<option>--encrypt</option>,
+					</term>
+					<listitem><para>Encrypt some data.</para></listitem>
+				</varlistentry>
+				<varlistentry>
+					<term>
 						<option>--derive</option>,
 					</term>
 					<listitem><para>Derive a secret key using another key and some data.</para></listitem>
@@ -646,6 +652,15 @@
 					</para></listitem>
 				</varlistentry>
 
+				<varlistentry>
+					<term>
+						<option>--iv</option> <replaceable>data</replaceable>
+					</term>
+					<listitem><para>Initialization vector for symmetric ciphers.
+					The <replaceable>data</replaceable> is hexadecimal number, i.e. "000013aa7bffa0".
+					</para></listitem>
+				</varlistentry>
+
 			</variablelist>
 		</para>
 	</refsect1>
@@ -668,6 +683,13 @@
 			using the private key with ID <replaceable>ID</replaceable> and
 			using the RSA-PKCS mechanism:
 				<programlisting>pkcs11-tool --sign --id ID --mechanism RSA-PKCS --input-file data --output-file data.sig</programlisting>
+
+			To encrypt file using the AES key with ID 85 and using mechanism AES-CBC with padding:
+				<programlisting>
+pkcs11-tool --encrypt --id 85 -m AES-CBC-PAD \
+ --iv "00000000000000000000000000000000" \
+ -i file.txt -o encrypted_file.data
+				</programlisting>
 		</para>
 	</refsect1>
 
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -381,6 +381,12 @@
 				</varlistentry>
 				<varlistentry>
 					<term>
+						<option>--unwrap</option>,
+					</term>
+					<listitem><para>Unwrap key.</para></listitem>
+				</varlistentry>
+				<varlistentry>
+					<term>
 						<option>--derive</option>,
 					</term>
 					<listitem><para>Derive a secret key using another key and some data.</para></listitem>
@@ -689,6 +695,18 @@
 pkcs11-tool --encrypt --id 85 -m AES-CBC-PAD \
  --iv "00000000000000000000000000000000" \
  -i file.txt -o encrypted_file.data
+				</programlisting>
+
+			Use the key with ID 22 and mechanism RSA-PKCS to unwrap key from file
+			<replaceable>aes_wrapped.key</replaceable>. After a successful unwrap operation,
+			a new AES key is created on token. ID of this key is set to 90 and label of this
+			key is set to <replaceable>unwrapped-key</replaceable>
+			Note: for the MyEID card, the AES key size must be present in key
+			specification i.e. AES:16
+				<programlisting>
+pkcs11-tool --unwrap --mechanism RSA-PKCS --id 22 \
+  -i aes_wrapped.key --key-type AES: \
+  --application-id 90 --applicatin-label unwrapped-key
 				</programlisting>
 		</para>
 	</refsect1>
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -166,6 +166,7 @@ enum {
 	OPT_DECRYPT,
 	OPT_ENCRYPT,
 	OPT_UNWRAP,
+	OPT_WRAP,
 	OPT_TEST_FORK,
 #if defined(_WIN32) || defined(HAVE_PTHREAD)
 	OPT_TEST_THREADS,
@@ -201,6 +202,7 @@ static const struct option options[] = {
 	{ "decrypt",		0, NULL,		OPT_DECRYPT },
 	{ "encrypt",		0, NULL,		OPT_ENCRYPT },
 	{ "unwrap",		0, NULL,		OPT_UNWRAP },
+	{ "wrap",		0, NULL,		OPT_WRAP },
 	{ "hash",		0, NULL,		'h' },
 	{ "derive",		0, NULL,		OPT_DERIVE },
 	{ "derive-pass-der",	0, NULL,		OPT_DERIVE_PASS_DER },
@@ -286,6 +288,7 @@ static const char *option_help[] = {
 	"Decrypt some data",
 	"Encrypt some data",
 	"Unwrap key",
+	"Wrap key",
 	"Hash some data",
 	"Derive a secret key using another key and some data",
 	"Derive ECDHpass DER encoded pubkey for compatibility with some PKCS#11 implementations",
@@ -526,6 +529,7 @@ static int		gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE,
 				CK_OBJECT_HANDLE *, CK_OBJECT_HANDLE *, const char *);
 static int		gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE, CK_OBJECT_HANDLE *, const char *, char *);
 static int		unwrap_key(CK_SESSION_HANDLE session);
+static int		wrap_key(CK_SESSION_HANDLE session);
 
 static int		write_object(CK_SESSION_HANDLE session);
 static int		read_object(CK_SESSION_HANDLE session);
@@ -684,6 +688,7 @@ int main(int argc, char * argv[])
 	int do_decrypt = 0;
 	int do_encrypt = 0;
 	int do_unwrap = 0;
+	int do_wrap = 0;
 	int do_hash = 0;
 	int do_derive = 0;
 	int do_gen_keypair = 0;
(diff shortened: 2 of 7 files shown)


tests/test-pkcs11-tool-sym-crypt-test.sh
```diff
@@ -0,0 +1,140 @@
...
+softhsm2-util --import aes_128.key --aes --token "SC test" --pin "$PIN" --label import_aes_128 --id "$ID2"
+assert $? "Fail, unable to import key"
+
+$PKCS11_TOOL --module="$P11LIB" --list-objects -l --pin=$PIN  2>/dev/null |tee > objects.list
+assert $? "Failed to list objects"
+
+VECTOR="00000000000000000000000000000000"
+echo "======================================================="
+echo " AES-CBC-PAD"
+echo " Opensssl encrypt, pkcs11-tool decrypt"
```
Jakuje, line 41:
```suggestion
echo " OpenSSL encrypt, pkcs11-tool decrypt"
```

tests/test-pkcs11-tool-sym-crypt-test.sh
```diff
@@ -0,0 +1,140 @@
...
+assert $? "Fail, AES-ECB must not work if the input is not a multiple of the block size"
+! $PKCS11_TOOL --module="$P11LIB" --pin "$PIN" --encrypt --id "$ID2" -m AES-CBC --iv "${VECTOR}" \
+	--input-file aes_plain.data --output-file aes_ciphertext_pkcs11.data 2>/dev/null
+assert $? "Fail, AES-CBC must not work if the input is not a multiple of the block size"
+
+echo -n "UUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUU" > aes_plain.data
+
+echo "======================================================="
+echo " AES-ECB"
+echo " Opensssl encrypt, pkcs11-tool decrypt"
```
Jakuje, line 75:
```suggestion
echo " OpenSSL encrypt, pkcs11-tool decrypt"
```

tests/test-pkcs11-tool-sym-crypt-test.sh
```diff
@@ -0,0 +1,140 @@
...
+assert $? "Fail, AES-ECB - wrong decrypt"
+
+$PKCS11_TOOL --module="$P11LIB" --pin "$PIN" --encrypt --id "$ID2" -m AES-ECB --input-file aes_plain.data --output-file aes_ciphertext_pkcs11.data 2>/dev/null
+assert $? "Fail/pkcs11-tool encrypt"
+cmp aes_ciphertext_pkcs11.data aes_ciphertext_openssl.data >/dev/null 2>/dev/null
+assert $? "Fail, AES-ECB - wrong encrypt"
+
+echo "======================================================="
+echo " AES-CBC"
+echo " Opensssl encrypt, pkcs11-tool decrypt"
```
Jakuje, line 93:
```suggestion
echo " OpenSSL encrypt, pkcs11-tool decrypt"
```

tests/test-pkcs11-tool-sym-crypt-test.sh
```diff
@@ -0,0 +1,140 @@
...
+$PKCS11_TOOL --module="$P11LIB" --pin "$PIN" --encrypt --id "$ID2" -m AES-CBC --iv "${VECTOR}" \
+	--input-file aes_plain.data --output-file aes_ciphertext_pkcs11.data 2>/dev/null
+assert $? "Fail/pkcs11-tool encrypt"
+cmp  aes_ciphertext_pkcs11.data aes_ciphertext_openssl.data >/dev/null 2>/dev/null
+assert $? "Fail, AES-CBC - wrong encrypt"
+
+VECTOR="000102030405060708090a0b0c0d0e0f"
+echo "======================================================="
+echo " AES-CBC, another IV"
+echo " Opensssl encrypt, pkcs11-tool decrypt"
```
Jakuje, line 114:
```suggestion
echo " OpenSSL encrypt, pkcs11-tool decrypt"
```

Jakuje (commented):
noticed just couple of typos, but otherwise looks good.

popovec:
@Jakuje Thanks for the review, I will do a force push with more features - key unwrap and wrap.

doc/tools/pkcs11-tool.1.xml
```diff
@@ -373,6 +373,19 @@
...
 
+				<varlistentry>
+					<term>
+						<option>--encrypt</option>,
+					</term>
+					<listitem><para>Encrypt some data.</para></listitem>
+				</varlistentry>
+				<varlistentry>
+					<term>
+						<option>--Unwrap</option>,
```
Jakuje, line 384: lowercase 'u'

popovec: Thanks for the review, I'll fix it. 

But I have question.. Is command line format (for unwrap)  acceptable (especially use of `--application-id`  `--application-label`)?  Use key id 22 to unwrap file input.key as AES key with ID 90 and label `unwraped-key`:
```
pkcs11-tool --unwrap --mechanism RSA-PKCS --id 22 -i input.key --key-type AES: --application-id 90 --applicatin-label unwraped-key
```
If appropriate, I can add another switch for specification of the unwrapped key.

For the wrap operation, I have similar question..  I suggest the following command line format (use key id 22 to wrap key id 90):
```
pkcs11-tool --wrap --mechanism RSA-PKCS --id 22 --application-id 90 -o output.key
```

I'll add the  the wrap code soon...

doc/tools/pkcs11-tool.1.xml
```diff
@@ -646,6 +664,15 @@
 					</para></listitem>
 				</varlistentry>
 
+				<varlistentry>
+					<term>
+						<option>--iv</option> <replaceable>data</replaceable>
+					</term>
+					<listitem><para>Initialization vector for symmetric ciphers.
+					<replaceable>data</replaceable> - hexadecimal numbers i.e. "000013aa7bffa0".
```
Jakuje, line 672: I would suggest that the description of the options in the manual pages should be in the whole sentences, such as "The `data` is hexadecimal number, i.e. ..."

doc/tools/pkcs11-tool.1.xml
```diff
@@ -668,6 +695,24 @@
...
+
+			To encrypt file using the AES key with ID 85 and using mechanism AES-CBC with padding:
+				<programlisting>
+pkcs11-tool --encrypt --id 85 -m AES-CBC-PAD \
+ --iv "00000000000000000000000000000000" \
+ -i file.txt -o encrypted_file.data
+				</programlisting>
+
+			Use the key with ID 22 and mechanism RSA-PKCS to unwrap key from file
+			<replaceable>aes_wrapped_key</replaceable> as AES key. This unwrapped
```
Jakuje, line 707: I would suggest the file should have some extension to be more obvious it is a file and not just a name or something. For example `aes_wrapped.key`.

doc/tools/pkcs11-tool.1.xml
```diff
@@ -668,6 +695,24 @@
...
+			AES key becomes label <replaceable>unwrapeed-key</replaceable> and ID 90.
```
Jakuje, line 708: I would rewrite this sentence to something like "After unwrapping the key is stored on token with a label ..."

src/tools/pkcs11-tool.c
```diff
@@ -1238,6 +1291,94 @@ int main(int argc, char * argv[])
...
+			util_fatal("failed to open %s: %m", opt_output);
+
+		r = write(fd, pWrappedKey, pulWrappedKeyLen);
+
+		if (r < 0)
+			util_fatal("Failed to write to %s: %m", opt_output);
+		if (fd != 1)
+			close(fd);
+		free(iv);
+	}
```
Jakuje, line 1381: please, create a separate function for wrap_key. This function is already long enough.

tests/test-pkcs11-tool-unwrap-wrap-test.sh
```diff
@@ -0,0 +1,83 @@
...
+
+echo "======================================================="
+echo "Setup SoftHSM"
+echo "======================================================="
+if [[ ! -f $P11LIB ]]; then
+    echo "WARNINIG: The SoftHSM is not installed. Can not run this test"
+    exit 77;
+fi
+# The Ubuntu has old softhsm version not supporting this feature
+grep "Ubuntu 18.04" /etc/issue && echo "WARNING: Not supported on Ubuntu 18.04" && exit 77
```
Jakuje, line 12: Travis already supports also Ubuntu 20.04: https://docs.travis-ci.com/user/reference/focal/ Does it already work there?

Would it make sense to try to migrate the CI to new version or add one more target which would execute also this test case? Or just write a new one in github actions, which should allow running in more up-to-date systems?

popovec:
Tests for symmetric encryption and wrap / unwrap operations are fully functional, tested on Ubuntu 20, against softhsm2. 
More at https://travis-ci.org/github/OpenSC/OpenSC/jobs/765595173 

One of the tests that is not affected by this  PR  does not work (FAIL: test-pkcs11-tool-allowed-mechanisms.sh).
The bug is probably on the softhsm2 side, probably a bug in the key listing. The following modification works for me on the Debian 10 system: 

```
diff --git a/tests/test-pkcs11-tool-allowed-mechanisms.sh b/tests/test-pkcs11-tool-allowed-mechanisms.sh
index 4c9c62d5..7dcb5590 100755
--- a/tests/test-pkcs11-tool-allowed-mechanisms.sh
+++ b/tests/test-pkcs11-tool-allowed-mechanisms.sh
@@ -22,7 +22,7 @@ MECHANISMS="RSA-PKCS,SHA1-RSA-PKCS,RSA-PKCS-PSS"
 # Generate key pair
 $PKCS11_TOOL --keypairgen --key-type="RSA:" --login --pin=$PIN \
        --module="$P11LIB" --label="test" --id="$ID" \
-       --allowed-mechanisms="$MECHANISMS"
+       --allowed-mechanisms="${MECHANISMS},SHA384-RSA-PKCS"
 assert $? "Failed to Generate RSA key pair"
 
 # Check the attributes are visible

```
If necessary, I can include it in this PR.

Jakuje:
Thank you for checking the new version. It looks like some bug in softhsm in Ubuntu 20. Locally in Fedora 34 it works for me fine (with or without your patch) so please, include it in the PR so we have it passing in CI.

I also see that even though the check fails, the pipeline passes, which is wrong. If I am right, it is because of the log is printed, which mangles the exit code. Something like this should address this problem:
```
diff --git a/.travis.yml b/.travis.yml
index e449f182..b87751d7 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -251,7 +251,7 @@ script:
       fi;
     fi
   - if [ -z "$HOST" -a "${DO_COVERITY_SCAN}" != "yes" -a -z "$DO_SIMULATION" ]; then
-      make check && make dist || cat tests/*log;
+      make check && make dist || (cat tests/*log src/tests/unittests/*log && exit 1);
     fi
   - if [ ! -z "$HOST" -a "${DO_COVERITY_SCAN}" != "yes" ]; then
       make install;
@@ -336,7 +336,7 @@ script:
     fi
   - if [ "${DO_SIMULATION}" = "cac" ]; then
       cd $TRAVIS_BUILD_DIR;
-      make check && sudo make install || (cat tests/*log src/tests/unittests/*log && exit);
+      make check && sudo make install || (cat tests/*log src/tests/unittests/*log && exit 1);
       export LD_LIBRARY_PATH=/usr/local/lib;
       cd src/tests/p11test/;
       ./p11test -s 0 -p 12345678 -i &
```

doc/tools/pkcs11-tool.1.xml
Jakuje: (sorry for a delay)
I do not like much the reuse of applicatin-id here. This is originally used for data objects. I think we will need a new switches for this not to confuse users. But I am not sure what would be the best terminology. `--new-id` and `--new-label` or `--wrapped-key-id` and `--wrapped-key-label` are just some ideas.

tests/test-pkcs11-tool-unwrap-wrap-test.sh
Jakuje: We have new Github Actions in place already, based on Ubuntu 20 so this should work there.

Jakuje:
Can you rebase the changes on current master, which introduced HMAC sigh&verify to pkcs11-tool? Otherwise I think this is good to go for the next release.

frankmorgner (approved):

doc/tools/pkcs11-tool.1.xml
popovec: I am aware that `--application_id`  and  `--application_label`is not used in its original meaning. As an alternative, I still like it the most:

a) unwrap: use the `--new-label` and `--new-id` switches (alternative `--new-key-label`, `--new-key-id`)
example:
```
pkcs11-tool --unwrap --mechanism RSA-PKCS --id 22 -i input.key --key-type AES: --new-key-id 90 --new-key-label unwraped-key

```
b) wrap:  it is possible to use the keyword `--wrap` with an argument, where either the `id` or `label` of the key that will be the subject of the wrapping can be used.

example:
```
pkcs11-tool --wrap 90 --mechanism RSA-PKCS --id 22  -o output.key

pkcs11-tool --wrap key_sign1  --mechanism RSA-PKCS --id 22  -o output.key
```

I would like other OpenSC developers to comment on this before applying this PR.

Thanks.

Jakuje: @frankmorgner what do you think?

PR #2269, opened 2021-03-22
Oss fuzz

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -644,8 +644,6 @@ iasecc_init_cpx(struct sc_card *card)
 	_sc_card_add_rsa_alg(card, 1024, flags, 0);
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
 
-	LOG_TEST_RET(ctx, sc_enum_apps(card), "Enumerate apps failed");
-
 	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
 }
 
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -2183,10 +2183,11 @@ static int do_asn1(int argc, char **argv)
 	}
 
 	/* workaround when the issuer of a card does prefix the EF.ATR payload with 0x80 */
-	if ((offsu == 0) && /* do not apply the workaround if any offset */
-	    (buf[0] == ISO7816_II_CATEGORY_TLV) &&
-	    (path.len >= 4) &&
-	    (memcmp(path.value, "\x3f\x00\x2f\x01", 4) == 0))
+	if (offsu == 0 /* do not apply the workaround if any offset */
+			&& r >= 1
+			&& buf[0] == ISO7816_II_CATEGORY_TLV
+			&& path.len >= 4
+			&& memcmp(path.value, "\x3f\x00\x2f\x01", 4) == 0)
 		offs++;
 	/* if offset does not exceed the length read from file/record, ... */
 	if (offs <= (unsigned int) r) {


frankmorgner:
@vjardin could you check if your card still works with the above changes?

Jakuje (approved):
Verified manually that it solves the mentioned oss-fuzz issue. Might make sense to add CID 367545 to the other commit to clarify that this issue was reported by coverity.

Jakuje:
For the failied travis-ci build, please pull https://github.com/Jakuje/OpenSC/commit/b7857fdc0ff627b2bec07e8c3fee20c221f05025 The issue was introduced by the ed25519 support, which I did not check with the recent compiler flags.

frankmorgner:
included the specific CID in the commit message

vjardin:
> @vjardin could you check if your card still works with the above changes?

You welcome. I'll provide some feedbacks asap (tonight).

src/libopensc/card-iasecc.c
```diff
@@ -644,8 +644,6 @@ iasecc_init_cpx(struct sc_card *card)
 	_sc_card_add_rsa_alg(card, 1024, flags, 0);
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
 
-	LOG_TEST_RET(ctx, sc_enum_apps(card), "Enumerate apps failed");
-
```
vjardin, line 648: I get a "permission denied" when opening the chromium.org URL.

I do confirm that it works with this patch. I am confused because some other drivers (for example edo_init()) does the same calls.

OK to remove this line, but please, if you can provide some backgrounds about "why is it called by other drivers ?", it would be welcomed.

src/tools/opensc-explorer.c
```diff
@@ -2183,10 +2183,11 @@ static int do_asn1(int argc, char **argv)
...
 	/* workaround when the issuer of a card does prefix the EF.ATR payload with 0x80 */
-	if ((offsu == 0) && /* do not apply the workaround if any offset */
-	    (buf[0] == ISO7816_II_CATEGORY_TLV) &&
-	    (path.len >= 4) &&
-	    (memcmp(path.value, "\x3f\x00\x2f\x01", 4) == 0))
+	if (offsu == 0 /* do not apply the workaround if any offset */
+			&& r >= 1
+			&& buf[0] == ISO7816_II_CATEGORY_TLV
+			&& path.len >= 4
+			&& memcmp(path.value, "\x3f\x00\x2f\x01", 4) == 0)
```
vjardin, line 2190: LGTM, manual checks OK too.

vjardin (approved):
OK

src/libopensc/card-iasecc.c
Jakuje: Yes. The bugs are restricted for the developers but they will open 30 days after releasing fix.

frankmorgner:
@vjardin basically, if `iasecc_init_cpx` fails, you should not modify the card object so that initialization of a different card type can be done. In your case, file path caching leads to a file object that will not be correctly cleaned up...

vjardin:
> @vjardin basically, if `iasecc_init_cpx` fails, you should not modify the card object so that initialization of a different card type can be done. In your case, file path caching leads to a file object that will not be correctly cleaned up...

Thanks @frankmorgner , it makes sense, but then I am very confused because my understanding was that first the match_atr from:
```
static int
iasecc_match_card(struct sc_card *card)
{
        struct sc_context *ctx = card->ctx;
        int i;

        i = _sc_match_atr(card, iasecc_known_atrs, &card->type);
        if (i < 0)   {
                sc_log(ctx, "card not matched");
                return 0;
        }
```
is performed, then in case of any mismatch of the ATR, the `iasecc_init()` would never be called.

I am concerned from your feedbacks, how can the `iasecc_init()` function be called for *different card type* ?

Jakuje:
The input for this crash is just 85 B long. The input contains the ATRs too so the first attempt matches the iasecc card:
```
P:8; T:0x139698219808704 10:05:42.246 [fuzz] card.c:1230:match_atr_table: ATR     : 2b:8f:80:01:00:31:b8:64:00:00:ec:c0:73:94:01:80:82:90:00:0e
...
P:8; T:0x139698219808704 10:05:42.247 [fuzz] card.c:1241:match_atr_table: ATR try : 2B:8F:80:01:00:31:B8:64:04:B0:EC:C1:73:94:01:80:82:90:00:0E
P:8; T:0x139698219808704 10:05:42.247 [fuzz] card.c:1248:match_atr_table: ATR mask: FF:FF:FF:FF:FF:FF:FF:FF:00:00:FF:C0:FF:FF:FF:FF:FF:FF:FF:FF
P:8; T:0x139698219808704 10:05:42.247 [fuzz] card-iasecc.c:389:iasecc_match_card: 'IAS/ECC CPxCL' card matched
```
but couple of lines below the init fails:
```
P:8; T:0x139698219808704 10:05:42.249 [fuzz] dir.c:182:sc_enum_apps: EF(DIR) is not a working EF.: -1210 (Card is invalid or cannot be handled)
P:8; T:0x139698219808704 10:05:42.249 [fuzz] card-iasecc.c:647:iasecc_init_cpx: Enumerate apps failed: -1210 (Card is invalid or cannot be handled)
P:8; T:0x139698219808704 10:05:42.249 [fuzz] card-iasecc.c:720:iasecc_init: returning with: -1210 (Card is invalid or cannot be handled)
P:8; T:0x139698219808704 10:05:42.249 [fuzz] card.c:360:sc_connect_card: driver 'IAS-ECC' init() failed: Card is invalid or cannot be handled
P:8; T:0x139698219808704 10:05:42.249 [fuzz] card.c:341:sc_connect_card: trying driver 'belpic'
...
```
which falls back to the trying other drivers.

From now on, it looks like we ran out of the fuzzing input so any further APDU returns errors:
```
P:8; T:0x139698219808704 10:05:42.261 [fuzz] apdu.c:370:sc_single_transmit: CLA:0, INS:A4, P1:4, P2:0, data(7) 0x7f0e06b75820
P:8; T:0x139698219808704 10:05:42.262 [fuzz] fuzz_pkcs15_reader.c:63:fuzz_get_chunk: Invalid Arguments
P:8; T:0x139698219808704 10:05:42.262 [fuzz] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
```
I think the issue is that the fuzzers run the same inputs several times to trigger even non-deterministic issues so after the failed card detection, context is destroyed and the same input is tried once more:
```
P:8; T:0x139698219808704 10:05:42.302 [fuzz] ctx.c:716:process_config_file: scconf_parse failed: Unable to open "/usr/local/etc/opensc.conf": No such file or directory
P:8; T:0x139698219808704 10:05:42.302 [fuzz] ctx.c:855:sc_context_create: ===================================
P:8; T:0x139698219808704 10:05:42.302 [fuzz] ctx.c:856:sc_context_create: opensc version: 0.21.0
P:8; T:0x139698219808704 10:05:42.302 [fuzz] card.c:254:sc_connect_card: called
P:8; T:0x139698219808704 10:05:42.302 [fuzz] fuzz_pkcs15_reader.c:83:fuzz_get_chunk: 
Returning fuzzing chunk (20 bytes):
2B 8F 80 01 00 31 B8 64 00 00 EC C0 73 94 01 80 +....1.d....s...
82 90 00 0E                                     ....
```
with the same results as described above, but after the second run, the memory leak is reported:
```
==8==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 464 byte(s) in 1 object(s) allocated from:
    #0 0x5256f2 in calloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:154:3
    #1 0xa6285e in sc_file_new /src/opensc/src/libopensc/sc.c:608:33
    #2 0xa62e99 in sc_file_dup /src/opensc/src/libopensc/sc.c:644:9
    #3 0x89dcd2 in iasecc_select_file /src/opensc/src/libopensc/card-iasecc.c:1152:5
    #4 0x574709 in sc_select_file /src/opensc/src/libopensc/card.c:865:6
    #5 0x57d619 in sc_enum_apps /src/opensc/src/libopensc/dir.c:175:6
    #6 0x8ac5de in iasecc_init_cpx /src/opensc/src/libopensc/card-iasecc.c:647:2
    #7 0x89683d in iasecc_init /src/opensc/src/libopensc/card-iasecc.c:680:8
    #8 0x569231 in sc_connect_card /src/opensc/src/libopensc/card.c:358:8
    #9 0x5587b7 in LLVMFuzzerTestOneInput /src/opensc/src/tests/fuzzing/fuzz_pkcs15_reader.c:209:5
    #10 0x45c093 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:599:15
    #11 0x445f52 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:323:6
    #12 0x44bd3a in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:856:9
    #13 0x477512 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #14 0x7f0e0515a83f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)

DEDUP_TOKEN: calloc--sc_file_new--sc_file_dup
Direct leak of 384 byte(s) in 1 object(s) allocated from:
    #0 0x525853 in realloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:164:3
    #1 0x578417 in _sc_card_add_algorithm /src/opensc/src/libopensc/card.c:1075:30
    #2 0x57981b in _sc_card_add_rsa_alg /src/opensc/src/libopensc/card.c:1200:9
    #3 0x8ac5d5 in iasecc_init_cpx /src/opensc/src/libopensc/card-iasecc.c:645:2
    #4 0x89683d in iasecc_init /src/opensc/src/libopensc/card-iasecc.c:680:8
    #5 0x569231 in sc_connect_card /src/opensc/src/libopensc/card.c:358:8
    #6 0x5587b7 in LLVMFuzzerTestOneInput /src/opensc/src/tests/fuzzing/fuzz_pkcs15_reader.c:209:5
    #7 0x45c093 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:599:15
    #8 0x445f52 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:323:6
    #9 0x44bd3a in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:856:9
    #10 0x477512 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #11 0x7f0e0515a83f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)
```
I would argue that invocation of fuzzer input should be self-contained and tied to the context, but it looks like there is still something preserved between input invocations.

Running through the code, it looks like this comment and will have something to do with it:

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L325-L336

I think it is the place where the allocated memory is leaked, especially if the init function is written the way how the iasecc was (allocate stuff through `_sc_card_add_rsa_alg()` and then call `sc_enum_apps()` which can fail).

Issue #2267, opened 2021-03-19
IASECC: Algo_refs not set, pkcs15_prkey_can_do() CKR_FUNCTION_NOT_SUPPORTED


vjardin:
When parsing `3f00/0001/5032`, we get the algRef = 66:
```
$ pkcs15-tool -k --verify-pin --pin PIN
[...]
 [pkcs15-tool] asn1.c:1499:asn1_decode_entry:    decoding 'algRef', raw data:42
 [pkcs15-tool] asn1.c:1527:asn1_decode_entry:    decoding 'algRef' returned 66
 [pkcs15-tool] asn1.c:1766:asn1_decode: returning with: 0 (Success)
```
because, from opensc-explorer's asn1 parser, it is:
```
      30 SEQUENCE (26 bytes)
         02 INTEGER (1 byte): 4
         02 INTEGER (1 byte): 64
         05 NULL
         03 BIT STRING (2 bytes): 10
         06 OBJECT IDENTIFIER (9 bytes):  1.2.840.113549.1.1.11, RSA-SHA256
         02 INTEGER (1 byte): 66
```

but as a bottom line, `Algo_refs` is not set:
```
pkcs15-tool -k --verify-pin --pin 1234 
Using reader with a card: ACS ACR33U-A1 3SAM ICC Reader 00 00
Private RSA Key [CPS_PRIV_SIG]
	Object Flags   : [0x01], private
	Usage          : [0x200], nonRepudiation
	Access Flags   : [0x0D], sensitive, alwaysSensitive, neverExtract
	Algo_refs      : 0
	Access Rules   : pso_cds:01;
	ModLength      : 2048
	Key ref        : 129 (0x81)
	Native         : yes
	Path           : e828bd080f8025000001ff0010::
	Auth ID        : 01
	ID             : e828bd080f8025000001ff001001
	MD:guid        : e7aab727-f2af-e673-37bb-7d43867a6349

Private RSA Key [CPS_PRIV_AUT]
	Object Flags   : [0x07], private, modifiable
	Usage          : [0x06], decrypt, sign
	Access Flags   : [0x0D], sensitive, alwaysSensitive, neverExtract
	Algo_refs      : 0
	Access Rules   : pso_decrypt:01; int_auth:01;
	ModLength      : 2048
	Key ref        : 130 (0x82)
	Native         : yes
	Path           : e828bd080f8025000001ff0010::
	Auth ID        : 01
	ID             : e828bd080f8025000001ff001002
[...]

PR #2270, opened 2021-03-22
IASECC/CPX: Fix AlgoRefs

vjardin:
Same than Gemalto's IASECC, the CPX cards need a workaround since
the PrKey does not have its Algo_regs.

Currently, it is a WIP in order to share the updates. Please, do not apply it yet.

We get:
```
pkcs15-tool -k --verify-pin --pin 1234
Using reader with a card: ACS ACR33U-A1 3SAM ICC Reader 00 00
Private RSA Key [CPS_PRIV_SIG]
	Usage          : [0x200], nonRepudiation
	Access Flags   : [0x0D], sensitive, alwaysSensitive, neverExtract
	Algo_refs      : 0
	Access Rules   : pso_cds:01;
	ModLength      : 2048
	Key ref        : 129 (0x81)
	Native         : yes
	Path           : e828bd080f8025000001ff0010::
	Auth ID        : 01
	ID             : e828bd080f8025000001ff001001
	MD:guid        : e7aab727-f2af-e673-37bb-7d43867a6349

Private RSA Key [CPS_PRIV_AUT]
	Object Flags   : [0x07], private, modifiable
	Usage          : [0x06], decrypt, sign
	Access Flags   : [0x0D], sensitive, alwaysSensitive, neverExtract
	Algo_refs      : 0
	Access Rules   : pso_decrypt:01; int_auth:01;
	ModLength      : 2048
	Key ref        : 130 (0x82)
	Native         : yes
	Path           : e828bd080f8025000001ff0010::
	Auth ID        : 01
	ID             : e828bd080f8025000001ff001002
	MD:guid        : 2b6bf284-225c-80bc-8cbe-1c791db33543
```

We need to get `Algo_regs` to be set to something that is not 0.

Fix: issue #2267

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -628,21 +628,23 @@ static int
 iasecc_init_cpx(struct sc_card *card)
 {
 	struct sc_context *ctx = card->ctx;
-	unsigned int flags; /* TBC it is not IASECC_CARD_DEFAULT_FLAGS */
+	unsigned int flags;
+	int rv = 0;
 
 	LOG_FUNC_CALLED(ctx);
 
-	card->caps = SC_CARD_CAP_RNG; /* TBC it is not IASECC_CARD_DEFAULT_CAPS */
+	card->caps = IASECC_CARD_DEFAULT_CAPS;
 
-	flags = SC_ALGORITHM_RSA_PAD_PKCS1;
-	flags |= SC_ALGORITHM_RSA_RAW;
-
-	flags |= SC_ALGORITHM_RSA_HASH_SHA1    |
-	         SC_ALGORITHM_RSA_HASH_SHA256;
+	flags = IASECC_CARD_DEFAULT_FLAGS;
 
 	_sc_card_add_rsa_alg(card, 512, flags, 0);
 	_sc_card_add_rsa_alg(card, 1024, flags, 0);
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
+
+	rv = iasecc_parse_ef_atr(card);
+	if (rv)
+		sc_invalidate_cache(card); /* avoid memory leakage */
+	LOG_TEST_RET(ctx, rv, "Parse EF.ATR");
 
 	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
 }
--- a/src/libopensc/iasecc.h
+++ b/src/libopensc/iasecc.h
@@ -133,4 +133,5 @@ struct iasecc_private_data {
 
 	struct iasecc_se_info *se_info;
 };
+
 #endif
--- a/src/libopensc/pkcs15-iasecc.c
+++ b/src/libopensc/pkcs15-iasecc.c
@@ -34,6 +34,7 @@
 
 #include "internal.h"
 #include "pkcs15.h"
+#include "../pkcs15init/pkcs15-iasecc.h"
 #include "iasecc.h"
 #include "aux-data.h"
 
--- a/src/libopensc/pkcs15-iasecc.c
+++ b/src/libopensc/pkcs15-iasecc.c
@@ -104,6 +104,38 @@ _iasecc_md_update_keyinfo(struct sc_pkcs15_card *p15card, struct sc_pkcs15_objec
 }
 
 
+/*
+ * CPx cards have an undocumented issue: they lack of
+ * Algorithm's reference into their PKCS's ASN1 encoding.
+ */
+static int
+_iasecc_cpx_fixup_prkdf(struct sc_pkcs15_card *p15card)
+{
+	struct sc_context * const ctx = p15card->card->ctx;
+	struct sc_pkcs15_object *pkobjs[32];
+	int ii, count;
+	int rv = SC_SUCCESS;
+
+	LOG_FUNC_CALLED(ctx);
+
+	rv = sc_pkcs15_get_objects(p15card, SC_PKCS15_TYPE_PRKEY, pkobjs, sizeof(pkobjs)/sizeof(pkobjs[0]));
+	LOG_TEST_RET(ctx, rv, "Cannot get PRKEY objects list");
+
+	count = rv;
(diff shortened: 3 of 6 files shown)


Jakuje:
I marked the PR as WIP so it does not pop up. Please, remove the status once it will be ready.

vjardin:
I do not understand yet how to fix this issue for Windows:
```
..\..\src\scconf\scconf.lib  ..\..\src\common\common.lib  ..\..\src\common\libscdl.lib  ..\..\src\ui\strings.lib  ..\..\src\ui\notify.lib  ..\..\src\sm\libsmiso.lib  ..\..\src\sm\libsmeac.lib  ..\..\src\pkcs15init\pkcs15init.lib    gdi32.lib Comctl32.lib Shell32.lib user32.lib advapi32.lib ws2_32.lib
opensc.def : error LNK2001: unresolved external symbol iasecc_pkcs15_encode_supported_algos
opensc.lib : fatal error LNK1120: 1 unresolved externals
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\link.EXE"' : return code '0x460'
Stop.
```
I already tried to add:
```
src/libopensc/libopensc.exports:iasecc_pkcs15_encode_supported_algos
```
but it is not enough.

Please, any tips ? I am not a windows/Visual expert.

Jakuje:
No windows expert here either, but you try to add the function `iasecc_pkcs15_encode_supported_algos` into the libopensc, while it is defined only in the `src/pkcs15init/pkcs15-iasecc.c` which is completely out of the scope of the libopensc.

Do you even need to export this function, when it is used only inside of the libopensc?

dengert:
Try this instead:
[iasecc.h.diff.txt](https://github.com/OpenSC/OpenSC/files/6244629/iasecc.h.diff.txt)
Don't know if it will work or not, but every other header file has it including pkcs15/init/pkcs15init.h. Windows, has windows had multiple calling conventions.

vjardin:
Thanks for the tips, the linkage issues are different now. I do not understand what is so specific with these environments :( I'll check...

If you know any OpenSC background / historical cases with the same compilation issues, they are welcomed ;)

src/libopensc/pkcs15-iasecc.c
```diff
@@ -103,6 +103,44 @@ _iasecc_md_update_keyinfo(struct sc_pkcs15_card *p15card, struct sc_pkcs15_objec
...
+	count = rv;
+	for(ii=0; ii<count; ii++)   {
+		rv = iasecc_pkcs15_encode_supported_algos(p15card, pkobjs[ii]);
+		LOG_TEST_RET(ctx, rv, "Cannot fix suported_algos");
+	}
+
+	LOG_FUNC_RETURN(ctx, rv);
+#else
+	LOG_FUNC_CALLED(ctx);
+	sc_log(ctx, "%s: missing ENABLE_OPENSSL", __func__);
```
Jakuje, line 133: I would suggest to use similar message as in other cases, something along the lines with "OpenSC was built without OpenSSL support: skipping", rather than referencing internal constants.

src/libopensc/pkcs15-iasecc.c
```diff
@@ -103,6 +103,44 @@ _iasecc_md_update_keyinfo(struct sc_pkcs15_card *p15card, struct sc_pkcs15_objec
...
+	for(ii=0; ii<count; ii++)   {
+		rv = iasecc_pkcs15_encode_supported_algos(p15card, pkobjs[ii]);
+		LOG_TEST_RET(ctx, rv, "Cannot fix suported_algos");
+	}
+
+	LOG_FUNC_RETURN(ctx, rv);
+#else
+	LOG_FUNC_CALLED(ctx);
+	sc_log(ctx, "OpenSC was built without OpenSSL support: skipping");
+	LOG_FUNC_RETURN(ctx, rv);
```
Jakuje, line 134: one more note, sorry for not noticing on the first sight. You can move the `LOG_FUNC_RETURN` out of the ifdef to avoid code duplication. as it is the same in both branches.

Jakuje (approved):

src/libopensc/pkcs15-iasecc.c
vjardin: done, I could have guessed it myself ;)

vjardin:
It fails because of:
image
how to kick again the ci ?

frankmorgner (changes requested):

src/libopensc/card-iasecc.c
```diff
@@ -628,22 +628,20 @@ static int
...
 
-	flags |= SC_ALGORITHM_RSA_HASH_SHA1    |
-	         SC_ALGORITHM_RSA_HASH_SHA256;
+	flags = IASECC_CARD_DEFAULT_FLAGS;
 
 	_sc_card_add_rsa_alg(card, 512, flags, 0);
 	_sc_card_add_rsa_alg(card, 1024, flags, 0);
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
 
+	LOG_TEST_RET(ctx, iasecc_parse_ef_atr(card), "Parse EF.ATR");
```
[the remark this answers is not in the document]
vjardin: Included

src/libopensc/iasecc.h
```diff
@@ -133,4 +138,10 @@ struct iasecc_private_data {
 
 	struct iasecc_se_info *se_info;
 };
+
+int iasecc_pkcs15_encode_supported_algos(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *object);
```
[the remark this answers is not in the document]
vjardin: Good idea. I updated the serie with this logic.

vjardin:
Hi @frankmorgner : I git-push-force-d this serie with your comments. Is it ok ?

vjardin:
Please, can we close this topic based on the enclosed serie ?

vjardin:
Thank you.

PR #2271, opened 2021-03-23
minor fixup after the eddsa merge

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -5582,16 +5582,18 @@ get_ec_pubkey_params(struct sc_pkcs15_pubkey *key, CK_ATTRIBUTE_PTR attr)
 			return sc_to_cryptoki_error(r, NULL);
 		}
 
-		attr->ulValueLen = (unsigned long)value_size;
 		if (attr->pValue == NULL_PTR) {
+			attr->ulValueLen = (unsigned long)value_size;
 			free(value);
 			return CKR_OK;
 		}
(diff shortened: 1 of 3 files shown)


src/pkcs11/framework-pkcs15.c
```diff
@@ -5577,22 +5577,22 @@ get_ec_pubkey_params(struct sc_pkcs15_pubkey *key, CK_ATTRIBUTE_PTR attr)
...
 		}
 
-		attr->ulValueLen = value_len;
+		attr->ulValueLen = (unsigned long)value_size;
 		if (attr->pValue == NULL_PTR) {
 			free(value);
 			return CKR_OK;
 		}
-		if (attr->ulValueLen < value_len) {
+		if (attr->ulValueLen < (unsigned long)value_size) {
```
frankmorgner, line 5590: Your commit doesn't introduce this problem, but as it now strikes my eye, we should take a closer look:

1. with `attr->ulValueLen = (unsigned long)value_size;`
2. this will always be false: `attr->ulValueLen < (unsigned long)value_size`

Jakuje: Right. Thank you for having a look! Indeed, it is incorrect and caused by my attempt to "optimize" some assignments. The first assignments needs to be in all the exit code paths but not before we know we can fit the output to the allocated buffer to follow pkcs11 semantics similarly as in the other functions. I will submit a fix shortly.

frankmorgner (approved):

PR #2277, opened 2021-03-29
Fix EOL for pkcs11-tool --test

vjardin:
The \n is not required since all the previous displays already
include an EOL.

For instance, without this fix, we were getting:
```
[...]
  testing key 0 (CPS_PRIV_SIG) -- can't be used to decrypt, skipping
  testing key 1 (CPS_PRIV_AUT)
 -- mechanism can't be used to decrypt, skipping
    RSA-PKCS: OK
No errors
```
instead of
```
[...]
  testing key 0 (CPS_PRIV_SIG) -- can't be used to decrypt, skipping
  testing key 1 (CPS_PRIV_AUT) -- mechanism can't be used to decrypt, skipping
    RSA-PKCS: OK
No errors
```

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6255,8 +6255,6 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 			continue;
 		}
 
-		printf("\n");
-
 #ifndef ENABLE_OPENSSL
 		printf("No OpenSSL support, unable to validate decryption\n");
 #else
@@ -6271,6 +6269,7 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 				continue;
 			}
 
+			/* each display will end with a \n */
 			errors += encrypt_decrypt(sess, mechs[n], privKeyObject);
 		}
 #endif


src/tools/pkcs11-tool.c
```diff
@@ -6255,8 +6255,6 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 			continue;
 		}
 
-		printf("\n");
-
```
Jakuje, line 6259: I do not think this is the correct fix. The above messages refer to the objects that are skipped for some reason (all followed with continue). If we get here, we start testing the key with different mechanisms and it needs to follow on a new line.

src/tools/pkcs11-tool.c
```diff
@@ -6271,6 +6269,7 @@ static int test_decrypt(CK_SESSION_HANDLE sess)
 				continue;
```
Jakuje, line 6269: The above message, should not be follow-up of the above line, but it should look like
```
printf("    %s: mechanism can't be used to decrypt, skipping\n", p11_mechanism_to_name(mechs[n]));
```
similarly as the first print inside of the `encrypt_decrypt()`.

The current version says the mechanism does not work for decrypt, but it does not say which one.

frankmorgner:
@vjardin , could you please check the comments above?

vjardin:
> @vjardin , could you please check the comments above?

I am a bit confused since the test from today's master with a CPX is leading to:

```
./pkcs11-tool --test
  seems to be OK
Digests:
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
Signatures: no private key found in this slot
  No private key found for testing
No errors
```

It seems that it performs less tests than it was used to be, how can it be ?! 

FYI, I performed some deep tests using google-chrome and opensc: I get the same access to all HTTPS French governmental web sites with both OpenSC and with the "private" driver from our French government. So, I do not "know" how important is this test tool ?! Maybe we can close this topic if it is a "non" issue.

Jakuje:
> Signatures: no private key found in this slot

It says it was not able to find a private key for some reason. But it is weird when all other uses work for you. The master version of `pkcs11-tool --test` still works for me with my yubikey.

So unless you want to update the PR, I would suggest to close (and reopen when there will be something new).

vjardin:
ok, let's close it.

Issue #2275, opened 2021-03-28
IASECC/CPX: get SE data  error - 6A 88 - check SM TBC


vjardin:
### Problem Description

iasecc_se_get_info() fails with the following logs:
```
 [opensc-pkcs11] apdu.c:367:sc_single_transmit: CLA:0, INS:CB, P1:3F, P2:FF, data(10) 0x7fffffffcaa6
 [opensc-pkcs11] reader-pcsc.c:323:pcsc_transmit: reader 'Ingenico TL TELIUM (25005334) 00 02'
 [opensc-pkcs11] reader-pcsc.c:324:pcsc_transmit:
Outgoing APDU (15 bytes):
00 CB 3F FF 0A 4D 08 70 06 BF FB 05 02 7B 80 ..?..M.p.....{.
 [opensc-pkcs11] reader-pcsc.c:242:pcsc_internal_transmit: called
 [opensc-pkcs11] reader-pcsc.c:333:pcsc_transmit:
Incoming APDU (2 bytes):
6A 88 j.
 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
 [opensc-pkcs11] apdu.c:507:xxxvincent: called
 [opensc-pkcs11] apdu.c:548:sc_transmit: returning with: 0 (Success)
 [opensc-pkcs11] card.c:523:sc_unlock: called
 [opensc-pkcs11] iso7816.c:128:iso7816_check_sw: Referenced data not found
 [opensc-pkcs11] card-iasecc.c:1729:iasecc_se_get_info: get SE data  error: -1216 (Data object not found)
 [opensc-pkcs11] iasecc-sm.c:235:iasecc_sm_se_mutual_authentication: Get SE info error: -1216 (Data object not found)
 [opensc-pkcs11] iasecc-sm.c:285:iasecc_sm_initialize: iasecc_sm_initialize() MUTUAL AUTHENTICATION failed: -1216 (Data object not found)
 [opensc-pkcs11] iasecc-sm.c:605:iasecc_sm_read_binary: iasecc_sm_read_binary() SM INITIALIZE failed: -1216 (Data object not found)
 [opensc-pkcs11] card-iasecc.c:819:_iasecc_sm_read_binary: returning with: -1216 (Data object not found)
 [opensc-pkcs11] card.c:638:sc_read_binary: returning with: -1216 (Data object not found)
[...]

PR #2278, opened 2021-03-29
IASECC/CPX: fix path errors

vjardin:
From runtime, it shows that the path needs to be set specifically in order to read these objects.

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -2871,6 +2871,34 @@ iasecc_sdo_get_tagged_data(struct sc_card *card, int sdo_tag, struct iasecc_sdo 
 
 	LOG_FUNC_CALLED(ctx);
 
+	sc_log(ctx, "sdo_tag=0x%x sdo_ref=0x%x sdo_class=0x%x", sdo_tag,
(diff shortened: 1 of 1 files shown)


frankmorgner:
I'm puzzled! I thought that by including https://github.com/OpenSC/OpenSC/pull/2217, the basic functionality of your card would be working (i.e. you can use all your private keys with `pkcs11-tool --test --login`). Is this not the case?

Looking at `master`, what is the missing functionality? You're constantly opening little pull requests making minor modifications to your card driver. You should stick to those little changes in terms of commits, but when making a pull request there should be some significant improvement in terms of functionality or quality, that's worth *my* time looking at your code.

vjardin:
Right, using #2217, it was good enough in order to use some legacy PKCS11 applications (such as php-pkcs11). Then, I started to deep dive into the logs and I found some strange behaviors (eg 6A 88) that are not show stopper. Digging deeper, these patches became required in order to avoid some fallback behavior of the iasecc driver.

I have isolated these 2 patches in order to ease the review and comments in order to fix the issue "IASECC/CPX: get SE data error - 6A 88" #2275 and to be focus only with this issue #2275

vjardin:
Great. Thank you.

PR #2279, opened 2021-03-29
Gemalto IAS ECC DUAL ID ONE Cosmo

vjardin:
Fix support for cards from:
   http://cartesapuce-discount.com/fr/cartes-a-puce-ias-ecc/146-cartes-a-puce-protiva-ias-ecc-tpc.html


vjardin:
For the records, the outputs are:
```
$ pkcs15-tool --list-info --reader 3       
PKCS#15 Card [ECC eID]:
	Version        : 1
	Serial number  : 990000000101000F
	Manufacturer ID: Oberthur Technologies
	Last update    : 20000101000000Z
	Profile        : IAS ECC Generic A05
	Flags          : Login required
		 sc_supported_algo_info[0]:
			 reference  : 1 (0x01)
			 mechanism  : [0x220] CKM_SHA_1                    
			 operations : [0x40], hash
			 algo_id    : 1.3.14.3.2.26
			 algo_ref   : [0x10]
		 sc_supported_algo_info[1]:
			 reference  : 2 (0x02)
			 mechanism  : [0x250] CKM_SHA256                   
			 operations : [0x40], hash
			 algo_id    : 2.16.840.1.101.3.4.2.1
			 algo_ref   : [0x40]
		 sc_supported_algo_info[2]:
			 reference  : 3 (0x03)
			 mechanism  : [0x06] CKM_SHA1_RSA_PKCS            
			 operations : [0x02], compute_signature
			 algo_id    : 1.2.840.113549.1.1.5
			 algo_ref   : [0x12]
		 sc_supported_algo_info[3]:
			 reference  : 4 (0x04)
			 mechanism  : [0x40] CKM_SHA256_RSA_PKCS          
			 operations : [0x02], compute_signature
			 algo_id    : 1.2.840.113549.1.1.11
			 algo_ref   : [0x42]
		 sc_supported_algo_info[4]:
			 reference  : 5 (0x05)
			 mechanism  : [0x01] CKM_RSA_PKCS                 
			 operations : [0x02], compute_signature
			 algo_id    : 1.2.840.113549.1.1.1
			 algo_ref   : [0x02]
		 sc_supported_algo_info[5]:
			 reference  : 6 (0x06)
			 mechanism  : [0x01] CKM_RSA_PKCS                 
			 operations : [0x20], decipher
			 algo_id    : 1.2.840.113549.1.1.1
			 algo_ref   : [0x1a]


$ pkcs15-tool --list-pins  --reader 3      
PIN [Card PIN]
	Auth ID        : 02
	ID             : c1
	Flags          : [0x810], initialized, exchangeRefData
	Length         : min_len:4, max_len:4, stored_len:4
	Pad char       : 0x00
	Reference      : 1 (0x01)

PIN [Card PUK]
	Object Flags   : [0x00]
	ID             : 02
	Flags          : [0x5C], change-disabled, unblock-disabled, initialized, unblockingPin
	Length         : min_len:8, max_len:8, stored_len:8
	Pad char       : 0x00
	Reference      : 2 (0x02)

$ pkcs15-tool --list-applications --reader 3
Application 'ECC Generic PKI':
	AID: E828BD080FD25047656E65726963

Application 'ECC eID':
	AID: E828BD080FD2504543432D654944
```
Since it is easy to buy any samples from this IAS/ECC card, I hope it'll help anyone that would like to start with the IASECC framework.

It seems that the specs were available few years ago from:
  https://web.archive.org/web/20170902215543/http://www.unsads.com/specs/IASECC/
but the PDFs from this URL have been removed !

frankmorgner (changes requested):
please do not give some magic offsets here. In the first patch, we've skipped the category indicator. Now, you want to skip an additional byte without understanding why it's there in the first place (note that you don't check the length of the return buffer in this case)

vjardin:
> please do not give some magic offsets here. In the first patch, we've skipped the category indicator. Now, you want to skip an additional byte without understanding why it's there in the first place (note that you don't check the length of the return buffer in this case)

You are right, I could not find any proper reference of this logic into the ISO standard. I'll remove this patch from this serie and I'll try to figure-out. Currently, only the card-iasecc leverages sc_parse_ef_atr() that was introduced by bf4aa793af8bdbeca37af02bdb7614ababe7cad4.

In fact, I am even wondering it it makes sense to patch asn1's CLI from opensc-explorer. I feel that it would be a better option to add a CLI command that calls/uses sc_parse_ef_atr() and then that displays the output.

Nevertheless, first, I'll remove this patch tonight and I'll prepare a specific call to sc_parse_ef_atr().

dengert:
The title of this PR is misleading.  

I believe "ID One Cosmo" is a trademark of "Oberthur" now owned by "IDEMIA" https://en.wikipedia.org/wiki/Oberthur_Technologies.

"Gemalto" is a separate company https://en.wikipedia.org/wiki/Gemalto

Are you saying that the IAS ECC applet maybe available on two different cards?

frankmorgner:
how does the second byte in EF.ATR look like?

vjardin:
> how does the second byte in EF.ATR look like?

see the commit logs: https://github.com/OpenSC/OpenSC/pull/2279/commits/5c2be6d62f3a13c459843594f00b25503de6aee8
```
00000000: 80 00 43 01 B8 46 04 40 70 14 10 47 03 94 01 80 ..C..F.@p..G....
```

dengert:
Back in olden days, vendors sold cards with preinstalled vendor software. Much like there were only IBM PC with DOS, and Apple with its OS, IBM had MVS and DEC had VMS.  Now you can run almost any OS on any type of CPU chip.  Same with smart cards, anyone can buy a blank card, and install their own applet.  
The ATR  does not define a card or manufacture, it is most electrical and low level capabilities. So they are not unique and the OpenSC matching does not match all the bits.  Cards can have more then one applet. OpenSC "card drivers" are in some cases really "applet drivers", such as PIV, CAC, OpenPGP. These match cards based on the response to a select file for their AID, they may use the ATR, but the AID must always match. 

The IAS/ECC sound more like a standard for an Applet. The first ATRs, SC_CARD_TYPE_IASECC_GEMALTO and SC_CARD_TYPE_IASECC_OBERTHUR are most likely vendor implementation of the the IAS/ECC specs. There are 6 more ATR in the match, and your two new ones. As you have found out, they all have their own idiosyncrasies. And now you say there is another "CARTES IAS ECC DUAL ID ONE Cosmo" and it will have its own ATR. And worse yet,  anyone can write an Applet to whatever spec they can find. 

https://www.ia.nato.int/niapc/Product/ID-One-IAS-Large-D_503 is interesting. It points out "ID-One" is a trademark  of Oberthur Technologies, and this might be the card being sold as "CARTES IAS ECC DUAL ID ONE Cosmo"

vjardin:
@dengert and @frankmorgner , see below an interesting combination test for 4 cases with the command:
```
$ iasecc-tool --list-application --reader 3
```
It is OK when I get a list of application (3 cases) and it is NOK when it failed (2 cases). Beside the case from this serie (Gemalto) which would be OK thanks to this serie, 2 cases are OK and 2 are failed. You will notice that the Oberthur case is NOK !!!

So, I do not have any history about the `iasecc_known_atrs[]` definition, but I was puzzle that SC_CARD_TYPE_IASECC_GEMALTO was defined into `iasecc_select_file()` https://github.com/OpenSC/OpenSC/blob/83162c5c87e88a5f911ca5349320938a52dbc1df/src/libopensc/card-iasecc.c#L1010-L1015  but never implemented into the code except for the init part (`iasecc_init_gemalto()`  - https://github.com/OpenSC/OpenSC/blob/83162c5c87e88a5f911ca5349320938a52dbc1df/src/libopensc/card-iasecc.c#L446 and https://github.com/OpenSC/OpenSC/blob/83162c5c87e88a5f911ca5349320938a52dbc1df/src/libopensc/card-iasecc.c#L667)

# cases OK

## GEMALTO
It requires the enclosed pull request, but is it the proper patch ? TBC.

## AMOS
```
@@ -79,7 +79,8 @@ static struct sc_card_driver iasecc_drv = {
 static const struct sc_atr_table +++ b/src/libopensc/card-iasecc.c
@@ -79,7 +79,8 @@ static struct sc_card_driver iasecc_drv = {
 static const struct sc_atr_table iasecc_known_atrs[] = {
 	{ "3B:7F:96:00:00:00:31:B8:64:40:70:14:10:73:94:01:80:82:90:00",
 	  "FF:FF:FF:FF:FF:FF:FF:FE:FF:FF:00:00:FF:FF:FF:FF:FF:FF:FF:FF",
-		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
+		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_SAGEM,  0, NULL },
+		//"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
[] = {
 	{ "3B:7F:96:00:00:00:31:B8:64:40:70:14:10:73:94:01:80:82:90:00",
 	  "FF:FF:FF:FF:FF:FF:FF:FE:FF:FF:00:00:FF:FF:FF:FF:FF:FF:FF:FF",
-		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
+		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_AMOS,  0, NULL },
+		//"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
```
Logs: [opensc-debug.logSC_CARD_TYPE_IASECC_AMOS.gz](https://github.com/OpenSC/OpenSC/files/6239078/opensc-debug.logSC_CARD_TYPE_IASECC_AMOS.gz)

## MI
```
@@ -79,7 +79,8 @@ static struct sc_card_driver iasecc_drv = {
 static const struct sc_atr_table iasecc_known_atrs[] = {
 	{ "3B:7F:96:00:00:00:31:B8:64:40:70:14:10:73:94:01:80:82:90:00",
 	  "FF:FF:FF:FF:FF:FF:FF:FE:FF:FF:00:00:FF:FF:FF:FF:FF:FF:FF:FF",
-		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
+		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_MI,  0, NULL },
+		//"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
```
Logs: [opensc-debug.logSC_CARD_TYPE_IASECC_MI.gz](https://github.com/OpenSC/OpenSC/files/6239077/opensc-debug.logSC_CARD_TYPE_IASECC_MI.gz)

# cases NOK

## Oberthur
```
@@ -79,7 +79,8 @@ static struct sc_card_driver iasecc_drv = {
 static const struct sc_atr_table iasecc_known_atrs[] = {
 	{ "3B:7F:96:00:00:00:31:B8:64:40:70:14:10:73:94:01:80:82:90:00",
 	  "FF:FF:FF:FF:FF:FF:FF:FE:FF:FF:00:00:FF:FF:FF:FF:FF:FF:FF:FF",
-		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
+		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_OBERTHUR,  0, NULL },
+		//"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
```
Logs: [opensc-debug.logSC_CARD_TYPE_IASECC_OBERTHUR.gz](https://github.com/OpenSC/OpenSC/files/6239089/opensc-debug.logSC_CARD_TYPE_IASECC_OBERTHUR.gz)

## SAGEM
```
@@ -79,7 +79,8 @@ static struct sc_card_driver iasecc_drv = {
 static const struct sc_atr_table iasecc_known_atrs[] = {
 	{ "3B:7F:96:00:00:00:31:B8:64:40:70:14:10:73:94:01:80:82:90:00",
 	  "FF:FF:FF:FF:FF:FF:FF:FE:FF:FF:00:00:FF:FF:FF:FF:FF:FF:FF:FF",
-		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
+		"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_SAGEM,  0, NULL },
+		//"IAS/ECC Gemalto", SC_CARD_TYPE_IASECC_GEMALTO,  0, NULL },
```
Logs: 
[opensc-debug.logSC_CARD_TYPE_IASECC_SAGEM.gz](https://github.com/OpenSC/OpenSC/files/6239094/opensc-debug.logSC_CARD_TYPE_IASECC_SAGEM.gz)

dengert:
I am not sure what you are asking, but keep in mind:

- be careful not to change code that you can not test. Others with these cards may depend on it. 

- A card may have more then one applet or application on the card and your driver should ignore any it does not understand. 

- Don't count on the ATR to much, you may still have to look at other data on a card to determine what application or version of an application it has. 

- OpenSC matches a card to a driver by letting each driver look at the card, and if the driver can not handle return and let OpenSC  try the next driver. 

- You should always assume your driver is looking  at a card you can not support in the match and _init() routines.

- An AID is a special form of a filename and can be registered internationally, so if you know you driver is supporting a specific AID always send a select file for for the AID, and check the results which should contain the AID as some cards have been know to respond with `90 00` for AID the card does not understand.  

Here is one list of known AIDs:
See https://www.eftlab.com/knowledge-base/211-emv-aid-rid-pix/

vjardin:
> * be careful not to change code that you can not test. Others with these cards may depend on it.

That's why my current patch does not change any existing code, I just add the "gemalto" case for the if() cases. The `iasecc_init_gemalto()` is there (https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-iasecc.c#L445), it is called (see the logs) for this card but the `if()` cases for Gemalto were missing from the code.

> * OpenSC matches a card to a driver by letting each driver look at the card, and if the driver can not handle return and let OpenSC  try the next driver.

My previous comment was just to highlight that the `SC_CARD_TYPE_IASECC_OBERTHUR` case of the `card-iasecc.c` driver cannot work while from a marketing point of view this card is supposed to be provided by OBERTHUR but `pcsc_scan` and the code proves that it does not work if we enforce the OBERTHUR case ! 

Since this IAS/ECC card can be purchased from Internet by anyone, feel free to double checks my comments. I would be pleased that an OpenSC expert double checks this surprising outcome.

frankmorgner:
Maybe @viktorTarasov could check with other types of ECCs?

frankmorgner (approved):
@vjardin , could you please check the merge conflict? The pull request itself looks fine in its current state.

vjardin:
> The pull request itself looks fine in its current state.

Thanks @frankmorgner, it is rebased, we should get soon the outcome of the CI's test too.

frankmorgner:
thanks

vjardin:
You welcome. It is a pleasure to get it reviewed thoroughly.

PR #2280, opened 2021-03-30
update configure.ac to be less noisy

pprindeville:
Seeing noise when running `autoreconf -fi` (i.e. `./bootstrap`).


frankmorgner (approved):

Jakuje (approved):

Jakuje:
I do not think the test failure is related to this change. Should we also bring this to the next release?

pprindeville:
Thanks!

Issue #2283, opened 2021-03-31
C_Sign for ECDSA fails if input is larger than key size


CardContact:
### Problem Description

When doing a routinely retest of the OpenSC master against the SmartCard-HSM we've come across a failing signature created with C_Sign using CKM_ECDSA_SHA1 on a secp192 curve. Apparently the produced signature fails verification if the input length exceeds 24 byte, which equals the key length.

This issue is not present in the latest 0.21.0 release, so it's probably introduced after that release.

### Proposed Resolution

Find and revise the commit that introduced the problem.

### Steps to reproduce

Using p11-test from the SmartCard-HSM workspace in the starterkit.

CardContact:
Apparently the input is truncated along the way:

33: C_Sign
2021-03-31 17:26:59.092
[in] hSession = 0x7fef7807ad10
[in] pData[ulDataLen] 00007fef5ce00050 / 255
    00000000  00 C5 8B 5D EF 9D 12 77 4A 04 7B 02 03 36 05 D8  ...]...wJ.{..6..
    00000010  A5 92 D7 46 48 F0 E5 BD 68 2F EE 83 8D 48 29 62  ...FH...h/...H)b
    00000020  52 E1 09 EB 93 A8 4E A0 24 34 95 70 40 0F 15 87  R.....N.$4.p@...
    00000030  00 B2 8D 0A 40 B6 0C 19 44 23 C5 BF 5D E2 D6 A7  ....@...D#..]...
    00000040  81 AF FB 79 E7 DD E4 4D 0E E8 D3 46 9D A2 E8 77  ...y...M...F...w
    00000050  AD 22 E3 BE 87 44 4B 94 71 60 12 A1 DF 1E B6 71  ."...DK.q`.....q
    00000060  B7 01 E5 12 A4 FB 6D 55 45 79 B1 6F 78 47 84 E0  ......mUEy.oxG..
    00000070  45 F9 A4 3E 92 01 D5 89 83 F1 FE BC 57 25 2B 5D  E..>........W%+]
    00000080  44 88 91 09 B8 86 C1 39 B4 E0 FD CA 83 18 F1 A1  D......9........
    00000090  11 6F 33 EA 1C EE D6 80 01 C0 C3 6D EE 0A 6E D5  .o3........m..n.
    000000A0  61 D4 B3 14 9B F9 3E 98 71 CF 19 EB A0 B2 8B 8B  a.....>.q.......
[...]

PR #2285, opened 2021-04-01
Fix for #2283 C_Sign fails ECDSA when card can do HASH on card

dengert:
Do not truncate ECDSA input to size of key if card or driver will do HASH.

Fixes: #2283

 On branch Fix_for_2283_ECDSA
 Changes to be committed:
	modified:   src/libopensc/pkcs15-sec.c

Tested with:

  token label        : SmartCard-HSM (UserPIN)
  token manufacturer : www.CardContact.de
  hardware version   : 24.13
  firmware version   : 1.2
  serial num         : DECC0100248
  pin min/max        : 6/15


Jakuje (approved):

frankmorgner:
@CardContact ?

vjardin:
+1 LGTM

PR #1960, opened 2020-02-25
Add support for (X)EdDSA keys in OpenPGP driver

Jakuje:
Fixes #1854

This patch set implements first batch of PKCS#11 3.0 features, mostly Ed25519 and curve25519 keys and related algorithms in the OpenSC itself and in the OpenPGP driver (as @jans23 provided me with NitroKey Start, which is equipped with GNUK). The card was initialized using `gpg` so the card initialization is not in place (yet).

There are many changes across the board including the `p11test` testsuite, which worked for me fine, but more testing would be appreciated. This is a transcript of `p11test` with current branch testing sign and derive mechanisms using these new keys:
```
[jjelen@t470s p11test (eddsa)]$ ./p11test -p 123456
[==========] Running 8 test(s).
[ RUN      ] wait_test
To test wait, run in interactive mode (-i switch).
[  SKIPPED ] wait_test
[ RUN      ] supported_mechanisms_test
[      MECHANISM      ] [ KEY SIZE ] [  FLAGS   ]
[        CKM_*        ] [ MIN][ MAX] [          ]
[SHA_1                ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[SHA224               ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[SHA256               ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[SHA384               ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[SHA512               ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[MD5                  ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[RIPEMD160            ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[GOSTR3411            ] [   0][   0] [CKF_DIGEST] CKF_DIGEST
[EDDSA                ] [ 255][ 255] [0x00000801] CKF_HW CKF_SIGN
[ECDH1_DERIVE         ] [ 255][ 255] [0x00080001] CKF_HW CKF_DERIVE
[       OK ] supported_mechanisms_test
[ RUN      ] readonly_tests
[KEY ID] [LABEL]
[ TYPE ] [ SIZE ] [PUBLIC] [SIGN&VERIFY] [ENC&DECRYPT] [WRAP&UNWR] [ DERIVE ]

[03    ] [Authentication key]
[ EC_E ] [   255] [      ] [[./]  [./] ] [[  ]  [  ] ] [[  ] [  ]] [[  ][  ]]
[...]

PR #2286, opened 2021-04-03
OpenPGP 3.4+: Enable ed25519/curve25519 support

ya-isakov:
This MR add support of 25519-series of curves to OpenSC. It extends support added in https://github.com/OpenSC/OpenSC/pull/1960.

Tested with Yubikey 5 NFC, Firmware 5.2.4.

It includes fix for Yubikey, which returns extra 0 at the end of OID in Algorithm Attribute DO. I've changed sc_compare_oid to return in it found -1 in any OID. This way, all trailing bytes will be ignored for comparison, including Import-Format.

p11test works, the only issue is with test_usages, which complains that derive should not be set on public key. I think, this should be an issue for all ECDH supporting tokens, as it's set in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-openpgp.c#L480 and https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-openpgp.c#L484.

##### Checklist

- [ ] macOS tokend is tested

P.S. Right now I'm just checking if version of card is more or equal to 3.4, but it seems, that there is a better way to detect list of supported algorithms, via Algorithm Information DO added in spec version 3.4. I've addressed that in https://github.com/OpenSC/OpenSC/pull/2287

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -613,7 +613,7 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
 			/* SC_OPENPGP_KEYALGO_ECDH || SC_OPENPGP_KEYALGO_ECDSA || SC_OPENPGP_KEYALGO_EDDSA */
 			key_info->algorithm = blob->data[0];
 
-			/* last byte is only set if pubkey import is supported, empty otherwise*/
+			/* last byte is set to 0xFF if pubkey import is supported */
 			if (blob->data[blob->len-1] == SC_OPENPGP_KEYFORMAT_EC_STDPUB){
 				if (blob->len < 3)
 					return SC_ERROR_INCORRECT_PARAMETERS;
@@ -621,9 +621,14 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
 				key_info->u.ec.keyformat = SC_OPENPGP_KEYFORMAT_EC_STDPUB;
 			}
 			else {
+				/* otherwise, last byte could be 00, so let's ignore it, as
+				 * it is not part of OID */
 				if (blob->len < 2)
 					return SC_ERROR_INCORRECT_PARAMETERS;
-				key_info->u.ec.oid_len = blob->len - 1;
+				if (blob->data[blob->len-1] == SC_OPENPGP_KEYFORMAT_EC_STD)
+					key_info->u.ec.oid_len = blob->len - 2;
+				else
+					key_info->u.ec.oid_len = blob->len - 1;
 				key_info->u.ec.keyformat = SC_OPENPGP_KEYFORMAT_EC_STD;
 			}
 
@@ -1635,7 +1640,6 @@ pgp_get_pubkey_pem(sc_card_t *card, unsigned int tag, u8 *buf, size_t buf_len)
 				/* PKCS#11 3.0: 2.3.5 Edwards EC public keys only support the use
 				 * of the curveName selection to specify a curve name as defined
 				 * in [RFC 8032] */
-
 				r = sc_pkcs15_encode_pubkey_as_spki(card->ctx, &p15pubkey, &data, &len);
 				break;
 			case SC_OPENPGP_KEYALGO_ECDH:
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -98,6 +98,19 @@ static pgp_ec_curves_t	ec_curves_openpgp[] = {
 	{{{-1}}, 0} /* This entry must not be touched. */
 };
 
+/* v3.0+ supports: [RFC 4880 & 6637] 0x12 = ECDH, 0x13 = ECDSA */
+static pgp_ec_curves_t  ec_curves_openpgp34[] = {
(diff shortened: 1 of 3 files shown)


dengert:
In regards to: "derive should not be set on public key." I disagree. Other drivers in OpenSC set
the usage for ECDH pub key to derive. (It is not used directly, but the peer uses it as input for a derive operation. 

I can not find directly in PKCS11 if it is appropriate or not. But an EC private key can be used for Sign ECDSA or derive ECDH. X509 cert also have usage and PKCS15 also has usage.
If used for Sign the public key has a usage of verify. 
My argument is if a certificate or the private key has a usage of derive, this fact should also be expressed in the usage of the public key.  

Policies may require some EC keys to be used only for signatures and others may only be used for derivation. If you don't allow for a usage of derive on public key,  you can not tell from the public key if the private key is allowed for derive.

We need a PKCS11 or other authoritative reference.

mouse07410:
Of course, @dengert is right. What idiot wrote a test that declares invalid "derive" for ECDH keys?! What Key Usage would _you_ assign to such a key?

src/libopensc/card-openpgp.c
```diff
@@ -98,6 +98,19 @@ static pgp_ec_curves_t	ec_curves_openpgp[] = {
 	{{{-1}}, 0} /* This entry must not be touched. */
 };
 
+/* v3.0+ supports: [RFC 4880 & 6637] 0x12 = ECDH, 0x13 = ECDSA */
```
Jakuje, line 101: copy&paste comment which no longer reflects the reality.

src/libopensc/card-openpgp.c
```diff
@@ -98,6 +98,19 @@ static pgp_ec_curves_t	ec_curves_openpgp[] = {
...
+        {{{1, 3, 132, 0, 34, -1}}, 384}, /* ansiX9p384r1 */
+        {{{1, 3, 132, 0, 35, -1}}, 521}, /* ansiX9p521r1 */
+        {{{1, 3, 36, 3, 3, 2, 8, 1, 1, 7, -1}}, 256}, /* brainpoolP256r1 */
+        {{{1, 3, 36, 3, 3, 2, 8, 1, 1, 11, -1}}, 384}, /* brainpoolP384r1 */
+        {{{1, 3, 36, 3, 3, 2, 8, 1, 1, 13, -1}}, 512}, /* brainpoolP512r1 */
+        {{{1, 3, 6, 1, 4, 1, 3029, 1, 5, 1, -1}}, 256}, /* curve25519 for encryption => CKK_EC_MONTGOMERY */
+        {{{1, 3, 6, 1, 4, 1, 11591, 15, 1, -1}}, 256}, /* ed25519 for signatures => CKK_EC_EDWARDS */
+        {{{-1}}, 0} /* This entry must not be touched. */
+};
+
```
Jakuje, line 113: This is the same as the original `ec_curves_openpgp` aside of the curve25519 and ed25519 addition so I am wondering if we could simplify this somehow to avoid list duplication. Either by prepending these new to the original list and using `ec_curves_openpgp` as a pointer to the third element or some other creative (but readable) way?

src/libopensc/card-openpgp.c
ya-isakov: Thank you! And what is reality now? I thought that these RFCs are still valid

src/libopensc/card-openpgp.c
ya-isakov: Thank you! I'm changing it to be similar to a pgp_objects

Jakuje (approved):

ya-isakov:
Appveyor problems are not related to my changes, it seems, but I don't know how to retrigger it...

Jakuje:
> Of course, @dengert is right. What idiot wrote a test that declares invalid "derive" for ECDH keys?! What Key Usage would _you_ assign to such a key?

Contributions and reviews are always welcomed.

ya-isakov:
I'm reworking OID comparison, as I don't like the idea of changing the function, which worked before. While researching on it, I've found a bug in parsing of last byte in Algorithm Attribute - if it's 00, which is a mark that no pubkey import is supported, it;s not ignored, and parsed as a part of oid. While fixing that, I realized that parse of Algorithm Attributes in DO 6E should be done, even if DO FA was parsed, as in Yubikey, DO FA entries has no last byte set to FF or 00 at all. So, to be on the safe side, I'm adding the logic that will parse DO 6E, and if algo was added previously, it will update it's keyformat.

Not a fan of this solution, but it's better than to make some wrong assumptions.

ya-isakov:
So, I reverted change of OID comparison logic, and instead, strip trailing zero byte, in case if it was read from DOs. Same logic as is used by gnupg (https://github.com/gpg/gnupg/blob/1c16878efd0bcf036f56abef93d64ac41ce9e95b/scd/app-openpgp.c#L5969-L5971 and https://github.com/gpg/gnupg/blob/1c16878efd0bcf036f56abef93d64ac41ce9e95b/scd/app-openpgp.c#L5969-L5971)
This logic was in place, but it ignored last byte only when it was 0xFF, and this was checked only in reading Algorithm Attributes while parsing list of supported curves, but same DO is read while parsing Public/Private keys, so I've added it there.

ya-isakov:
@Jakuje I'm happy with algorithm now, so, should I wait for someone with merge rights to merge it?

PR #2287, opened 2021-04-04
OpenPGP 3.4+: Get list of supported algorithms from Algorithm Information

ya-isakov:
OpenPGP Card 3.4+ supports getting list of algos from the new Algoritm Information DO
https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.pdf, 4.4.3.11
If at least one algo is in this list, use it, and ignore hardcoded.

Tested on Yubikey 5 NFC with FW 5.2.4. Handled few problems with Yubikey not
following the spec.

Also, fixed problem with algorithm find, which will return EC algorithm with the same
key size, even if curve is not the same (e.g., curves 1.3.132.0.34 and 1.3.36.3.3.2.8.1.1.11 has
the same keysize of 384 bits, so it was possible that one returned in place of other)

This is needed for Yubikey to support more algorithms, than are used in Algorithm Attributes DO,
in case it was initialized with gpg, and not pkcs15-tool.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -1246,12 +1246,14 @@ pgp_enumerate_blob(sc_card_t *card, pgp_blob_t *blob)
 
 		r = sc_asn1_read_tag(&data, blob->len - (in - blob->data),
 					&cla, &tag, &len);
-		if (data == NULL) {
-			sc_log(card->ctx, "Unexpected end of contents");
-			return SC_ERROR_OBJECT_NOT_VALID;
-		}
 		if (r == SC_ERROR_INVALID_ASN1_OBJECT) {
 			sc_log(card->ctx, "Invalid ASN.1 object");
+			return SC_ERROR_OBJECT_NOT_VALID;
+		}
+		/* Check for unknown error, or empty data */
+		if (((r < 0) && (r != SC_ERROR_ASN1_END_OF_CONTENTS)) ||
+		    (data == NULL)) {
+			sc_log(card->ctx, "Unexpected end of contents");
 			return SC_ERROR_OBJECT_NOT_VALID;
 		}
 
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -685,58 +685,61 @@ int _pgp_add_algo(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t key_info, si
 {
 	unsigned long flags = 0, ext_flags = 0;
 
-	/* RSA [RFC 4880] */
+	/* [RFC 4880], [draft-ietf-openpgp-crypto-refresh] */
 	switch (key_info.algorithm) {
-		case SC_OPENPGP_KEYALGO_RSA:
-			/* OpenPGP card spec 1.1 & 2.x, section 7.2.9 & 7.2.10 /
-			 * v3.x section 7.2.11 & 7.2.12 */
-			flags = SC_ALGORITHM_RSA_PAD_PKCS1 |
-				SC_ALGORITHM_RSA_HASH_NONE |
-				SC_ALGORITHM_ONBOARD_KEY_GEN;	/* key gen on card */
+	case SC_OPENPGP_KEYALGO_RSA:
+		/* OpenPGP card spec 1.1 & 2.x, section 7.2.9 & 7.2.10 /
+		 * v3.x section 7.2.11 & 7.2.12 */
+		flags = SC_ALGORITHM_RSA_PAD_PKCS1 |
+			SC_ALGORITHM_RSA_HASH_NONE |
+			SC_ALGORITHM_ONBOARD_KEY_GEN;	/* key gen on card */
 
-			_sc_card_add_rsa_alg(card, key_info.u.rsa.modulus_len, flags, 0);
-			sc_log(card->ctx, "DO %zX: Added RSA algorithm, mod_len = %"
-				SC_FORMAT_LEN_SIZE_T"u",
-				do_num, key_info.u.rsa.modulus_len);
-			LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
-		case SC_OPENPGP_KEYALGO_ECDH:
-			/* The montgomery curve (curve25519) needs to go through
-			 * different paths, otherwise we handle it as a normal EC key */
-			if (_pgp_handle_curve25519(card, key_info, do_num))
-			    LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
-			/* fall through */
-		case SC_OPENPGP_KEYALGO_ECDSA:
-			/* v3.0+: ECC [RFC 4880 & 6637] */
-			/* EdDSA from draft-ietf-openpgp-rfc4880bis-08 */
+		_sc_card_add_rsa_alg(card, key_info.u.rsa.modulus_len, flags, 0);
+		sc_log(card->ctx, "DO %zX: Added RSA algorithm, mod_len = %"
+			SC_FORMAT_LEN_SIZE_T"u",
+			do_num, key_info.u.rsa.modulus_len);
+		break;
+	case SC_OPENPGP_KEYALGO_ECDH:
+		/* The montgomery curve (curve25519) needs to go through
+		 * different paths, otherwise we handle it as a normal EC key */
+		if (_pgp_handle_curve25519(card, key_info, do_num))
+			break;
+		/* fall through */
+	case SC_OPENPGP_KEYALGO_ECDSA:
+		/* v3.0+: ECC [RFC 4880 & 6637] */
+		/* EdDSA from draft-ietf-openpgp-rfc4880bis-08 */
 
-			/* Allow curve to be used by both ECDH and ECDSA.
-                         * pgp_init set these flags the same way */
-			flags = SC_ALGORITHM_ECDH_CDH_RAW;
-			flags |= SC_ALGORITHM_ECDSA_RAW;
-			flags |= SC_ALGORITHM_ECDSA_HASH_NONE;
-			flags |= SC_ALGORITHM_ONBOARD_KEY_GEN;
-			ext_flags = SC_ALGORITHM_EXT_EC_NAMEDCURVE;
+		/* Allow curve to be used by both ECDH and ECDSA.
+		 * pgp_init set these flags the same way */
+		flags = SC_ALGORITHM_ECDH_CDH_RAW;
+		flags |= SC_ALGORITHM_ECDSA_RAW;
+		flags |= SC_ALGORITHM_ECDSA_HASH_NONE;
+		flags |= SC_ALGORITHM_ONBOARD_KEY_GEN;
+		ext_flags = SC_ALGORITHM_EXT_EC_NAMEDCURVE;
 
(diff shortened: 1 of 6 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -650,6 +653,80 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
...
+	return 1;
+}
+
+int _pgp_add_algo(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t key_info, size_t do_num)
+{
+	unsigned long flags = 0, ext_flags = 0;
+
+	/* RSA [RFC 4880] */
+	switch (key_info.algorithm) {
+		case SC_OPENPGP_KEYALGO_RSA:
```
Jakuje, line 678: you can unindent the `case` lines as it was in the original code to avoid too long lines.

src/libopensc/card-openpgp.c
```diff
@@ -650,6 +653,80 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
...
+	sc_log(card->ctx, "DO %zX: Added XEDDSA algorithm (%d), mod_len = %d",
+	    do_num, SC_ALGORITHM_XEDDSA, key_info.u.ec.key_length);
+	return 1;
+}
+
+int _pgp_add_algo(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t key_info, size_t do_num)
+{
+	unsigned long flags = 0, ext_flags = 0;
+
+	/* RSA [RFC 4880] */
```
Jakuje, line 676: the comment looks misplaced

src/libopensc/card-openpgp.c
```diff
@@ -650,6 +653,80 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
...
+			 * different paths, otherwise we handle it as a normal EC key */
+			if (_pgp_handle_curve25519(card, key_info, do_num))
+			    LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+			/* fall through */
+		case SC_OPENPGP_KEYALGO_ECDSA:
+			/* v3.0+: ECC [RFC 4880 & 6637] */
+			/* EdDSA from draft-ietf-openpgp-rfc4880bis-08 */
+
+			/* Allow curve to be used by both ECDH and ECDSA.
+                         * pgp_init set these flags the same way */
```
Jakuje, line 701: wrong indentation -- we use tabs.

Jakuje (commented):
Looks good. Just couple of comments around the code style. Tested with Nitrokey Start and I did not notice any issues.

src/libopensc/card-openpgp.c
ya-isakov: Thank you!

src/libopensc/card-openpgp.c
ya-isakov: thank you!

src/libopensc/card-openpgp.c
ya-isakov: It was like that before (https://github.com/OpenSC/OpenSC/pull/2287/files#diff-36989530cd06078fd8bce35f8f4c950abceca45e73980a5fddcd3731658db775L804), but I'll change it to mention https://tools.ietf.org/html/draft-ietf-openpgp-crypto-refresh

ya-isakov:
@Jakuje thank you for review. IIRC, Nitrokeys are based on GNUK, which OpenPGP Card 2.0 with some extensions, so, my code should not work for it. But it's good to know that I haven't broke anything for other tokens :) I've changed logic a little bit (now, DO FA should be checked independent from DO 6E, but, still, if any algo is found by it, code in DO 6E will ignore info from DOs 7E -> C1-C3. Could you, please, test it again?

Jakuje:
> @Jakuje thank you for review. IIRC, Nitrokeys are based on GNUK, which OpenPGP Card 2.0 with some extensions, so, my code should not work for it.

Right. I was writing the initial support for ed25519 in opensc around the gnuk.

> But it's good to know that I haven't broke anything for other tokens :) I've changed logic a little bit (now, DO FA should be checked independent from DO 6E, but, still, if any algo is found by it, code in DO 6E will ignore info from DOs 7E -> C1-C3. Could you, please, test it again?

Retested current version and looks working fine.

src/libopensc/card-openpgp.c
```diff
@@ -1188,20 +1234,29 @@ pgp_enumerate_blob(sc_card_t *card, pgp_blob_t *blob)
 
 		r = sc_asn1_read_tag(&data, blob->len - (in - blob->data),
 					&cla, &tag, &len);
-		if (r < 0 || data == NULL) {
-			sc_log(card->ctx,
-				 "Unexpected end of contents\n");
+		if (data == NULL) {
```
Jakuje, line 1237: it is safer to keep both checks here -- some static analyzers might get confused by checking only one of these.

ya-isakov: I've split this check into three - `data = NULL` is in the same place, then `r == SC_ERROR_INVALID_ASN1_OBJECT`, to have a better message, than "Unexpected end of contents", and `r == SC_ERROR_ASN1_END_OF_CONTENTS`, to handle Yubikey issue. `sc_asn1_read_tag` can return only SC_ERROR_INVALID_ASN1_OBJECT and SC_ERROR_ASN1_END_OF_CONTENTS, so, I think, I covered all cases?

ya-isakov: Or do you mean, that static analyzers are not aware, that there are only two possible errors in called function, and want that calling function should check for r<0 anyway?

Jakuje: Sorry, I missed that part that you moved the `r` checking to the other places. Right now, it returns these two errors, but it might change in the future and we might get different return values which would go unnoticed. So I would be for adding a fallback branch for all the `r < 0` anyway.

ya-isakov: Got it, will add, thank you!

ya-isakov:
After some research, I found that OpenSC is parsing Import-Format byte, but then it's not storing it anywhere. So, it's set in [pgp_parse_algo_attr_blob](https://github.com/OpenSC/OpenSC/blob/b8266a4c86e06bec9e3d150e688efcd96b0721e1/src/libopensc/card-openpgp.c#L614), but the struct, where it is set, is created in [pgp_get_card_features](https://github.com/OpenSC/OpenSC/blob/b8266a4c86e06bec9e3d150e688efcd96b0721e1/src/libopensc/card-openpgp.c#L795), but nowhere In this function I can see usage of keyformat field (e.g, it is not added as algo field) Then, keyformat is checked only once - in [pgp_build_extended_header_list](https://github.com/OpenSC/OpenSC/blob/b8266a4c86e06bec9e3d150e688efcd96b0721e1/src/libopensc/card-openpgp.c#L3151), and this function is called in [openpgp_store_key](https://github.com/OpenSC/OpenSC/blob/b8266a4c86e06bec9e3d150e688efcd96b0721e1/src/pkcs15init/pkcs15-openpgp.c#L166). In the same function, key_info is created, but no keyformat field is set.

So, I think I can assume, that as this setting is ignored, my current approach is fine - for tokens < 3.4, the setting will be read from DO 6E, and ignored. For 3.4+ tokens, if DO FA has this byte set for any algo, it will be ignored, again...

dengert:
Looks like the keyformat deals with the format of the private key on the card.
```
#define SC_OPENPGP_KEYFORMAT_RSA_STD	0    /* See 4.3.3.6 Algorithm Attributes */
#define SC_OPENPGP_KEYFORMAT_RSA_STDN	1    /* OpenPGP card spec v2 */
#define SC_OPENPGP_KEYFORMAT_RSA_CRT	2
#define SC_OPENPGP_KEYFORMAT_RSA_CRTN	3
#define SC_OPENPGP_KEYFORMAT_EC_STD	0
#define SC_OPENPGP_KEYFORMAT_EC_STDPUB	0xFF
```
And in OpenPGP-3.4.1 it shows how the STD and CRT are used.

So unless you are loading or reading a private key, you should not care, because you will never see it.   `src/tools/openpgp-tool.c` uses it in  `pgp_store_key` and only supports some of them.

ya-isakov:
I've fixed problem with multiple ECDH-DERIVE mechanisms, which surfaced after DO FA parsing. Now, derive works (with MRs https://github.com/OpenSC/OpenSC/pull/2286 and https://github.com/OpenSC/OpenSC/pull/2288)

src/libopensc/card-openpgp.c
```diff
@@ -650,6 +653,83 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
...
+		break;
+	case SC_OPENPGP_KEYALGO_ECDH:
+		/* The montgomery curve (curve25519) needs to go through
+		 * different paths, otherwise we handle it as a normal EC key */
+		if (_pgp_handle_curve25519(card, key_info, do_num))
+			break;
+		/* fall through */
+	case SC_OPENPGP_KEYALGO_ECDSA:
+		/* v3.0+: ECC [RFC 4880 & 6637] */
+		/* EdDSA from draft-ietf-openpgp-rfc4880bis-08 */
```
Jakuje, line 698: i think this comment is misplaced as this case is about ecdsa.

src/libopensc/card.c
```diff
@@ -1154,16 +1154,13 @@ sc_algorithm_info_t * sc_card_find_alg(sc_card_t *card,
```
frankmorgner, line 1154: The indenting is broken here, isn't it?

frankmorgner (approved):

frankmorgner (commented):
The code looks good, just a formatting issue

Jakuje:
@ya-isakov can you rebase the patch set on the current master and resolve the conflicts?

ya-isakov:
@Jakuje, Sure, I'll do it on this weekend.

src/libopensc/card.c
ya-isakov: Sorry, how it is broken? Cannot see it.

ya-isakov: Oh, because I dropped one level of ifs, I see.

ya-isakov:
I've rebased my branch, fixed leftover comments, but I need to check, if it is still working :) Will do it soon.

Jakuje (approved):
just minor nits. I think we are good to go with this. Tested locally with Nitrokey start and all looks fine.

szszszsz:
> @Jakuje thank you for review. IIRC, Nitrokeys are based on GNUK, which OpenPGP Card 2.0 with some extensions, so, my code should not work for it.

A bit off topic: Nitrokey Pro and Nitrokey Storage contain OpenPGP v3.3 card. Nitrokey Start is indeed v2.0+ software OpenPGP implementation, based on GNUK.

Jakuje:
@szszszsz thanks for clarification and thanks for reminder I wanted to merge this finally.

It would be great if we could get the nits I mentioned last month fixed. But anyway, if there will be no further concerns, I will merge this during the next week.

frankmorgner (approved):

Jakuje:
Thank you!

PR #2288, opened 2021-04-05
OpenPGP: Fix read of certs with Ed25519 key

ya-isakov:
Proper Ed25519 certs have pubkey algo with OID 1.3.101.112, according to
RFC8410. This commit add this OID, and also fixes pubkey parsing - according
to the same RFC, it's just a bytestring, without ASN.1 wrapping.

##### Checklist

- [ ] macOS tokend is tested


ya-isakov:
Closed, as https://github.com/OpenSC/OpenSC/pull/2293 has a better version

PR #2292, opened 2021-04-07
Public keys can have CKA_DERIVE attribute

dengert:
in response to https://github.com/OpenSC/OpenSC/pull/2286#issuecomment-814734597

This removes the error message:

fprintf(stderr, " [ ERROR %s ] Derive should not be set on public key\n",

PKCS11-base-v2.40 says:
Table 22, Common Key Attributes  Common key attributes:
    CKA_DERIVE  | CK_BBOOL | CK_TRUE if key supports key derivation (i.e., if other keys can be derived from this one (default CK_FALSE)

Table 24, Mapping of X.509 key usage flags to Cryptoki attributes for public keys:
   Key usage flags for public keys in X.509 public key certificates
      Corresponding cryptoki attributes for public keys.
         keyAgreement  CKA_DERIVE

NIST demo PIV card 4 with EC keys was used with p11test. The error message not shown.
```
[ RUN      ] usage_test
[KEY ID] [LABEL]
[ TYPE ] [ SIZE ] [PUBLIC] [SIGN&VERIFY] [ENC&DECRYPT] [WRAP&UNWR] [ DERIVE ] [ALWAYS_AUTH]

[07    ] [Retired Certificate for Key Management 3]
[ RSA  ] [  2048] [  ./  ] [[  ]  [  ] ] [[./]  [./] ] [[./] [./]] [[  ][  ]] [    [  ]   ]

[01    ] [Certificate for PIV Authentication]
[  EC  ] [   256] [  ./  ] [[./]  [./] ] [[  ]  [  ] ] [[  ] [  ]] [[  ][  ]] [    [  ]   ]

[02    ] [Certificate for Digital Signature]
[  EC  ] [   256] [  ./  ] [[./]  [./] ] [[  ]  [  ] ] [[  ] [  ]] [[  ][  ]] [    [./]   ]

[03    ] [Certificate for Key Management]
[  EC  ] [   256] [  ./  ] [[  ]  [  ] ] [[  ]  [  ] ] [[  ] [  ]] [[./][./]] [    [  ]   ]

[05    ] [Retired Certificate for Key Management 1]
[  EC  ] [   256] [  ./  ] [[  ]  [  ] ] [[  ]  [  ] ] [[  ] [  ]] [[./][./]] [    [  ]   ]

[06    ] [Retired Certificate for Key Management 2]
[  EC  ] [   256] [  ./  ] [[  ]  [  ] ] [[  ]  [  ] ] [[  ] [  ]] [[./][./]] [    [  ]   ]

[04    ] [Certificate for Card Authentication]
[  EC  ] [   256] [  ./  ] [[./]  [./] ] [[  ]  [  ] ] [[  ] [  ]] [[  ][  ]] [    [  ]   ]
 Public == Cert -----^       ^-----^       ^-----^       ^----^      ^---^
 Sign & Verify Attributes ------'             |            |           |
 Encrypt & Decrypt Attributes ----------------'            |           |
 Wrap & Unwrap Attributes ---------------------------------'           |
 Public and Private key Derive Attributes -----------------------------'
[       OK ] usage_test
```

### DIFF
--- a/src/tests/p11test/p11test_case_usage.c
+++ b/src/tests/p11test/p11test_case_usage.c
@@ -54,11 +54,6 @@ void usage_test(void **state) {
 			fprintf(stderr, " [ ERROR %s ] If Unwrap is set, Wrap should be set too.\n",
 			    objects.data[i].id_str);
 		}
-		if (objects.data[i].derive_pub) {
-			errors++;
-			fprintf(stderr, " [ ERROR %s ] Derive should not be set on public key\n",
-			    objects.data[i].id_str);
-		}
 
 		/* We have at least one usage flag for every key group */
 		if (! objects.data[i].sign       && ! objects.data[i].verify &&


ya-isakov:
The check for derive usage was there, before change, it was like that https://github.com/OpenSC/OpenSC/blob/32ec1f92b9f1ff134ee74677a449d87c68d1c01d/src/tests/p11test/p11test_case_usage.c#L57-L61.
Do you think, that it was also invalid?

dengert:
You are correct. The check says they should both be set.
I will withdraw the PR. But you should make sure the OpenPGP set the bits.


PKCS11-base-v2.40 says:
Table 22, Common Key Attributes Common key attributes:
CKA_DERIVE | CK_BBOOL | CK_TRUE if key supports key derivation (i.e., if other keys can be derived from this one (default CK_FALSE)

Table 24, Mapping of X.509 key usage flags to Cryptoki attributes for public keys:
Key usage flags for public keys in X.509 public key certificates
Corresponding cryptoki attributes for public keys.
keyAgreement CKA_DERIVE

dengert:
I misread the check it is valid and works as expected. The assertion in #2286 that public keys should not have the CKA_DERIVE is the problem.

dengert:
The confusion about this PR was caused by @ya-isakov  list of the source in https://github.com/OpenSC/OpenSC/pull/2292#issuecomment-815151779 . I was looking at master, but the source above is  from 32ec1f92b9 which is a better fix, as both keys should have CKA_DERIVE.

dengert:
I think the two lines 57 and 59 from #485b6cff4 (Jakub Jelen 2020-02-25 10:44:27 +0100  57
should look like:
```
		if (objects.data[i].derive_pub != objects.data[i].derive_priv) {

			fprintf(stderr, " [ ERROR %s ] Derive should be set on both private and public part.\n",
```

which would revert these 2 lines from #485b6cff4

My PR remove lines 57-61 which was too much. Pkcs11 allows CKA_DERIVE to be set on both public and private keys, and both are used during a derive operation, the private key by the local application and the public key sent to the peer application. So this also implies CKA_DERIVE should be the same in the key pair. And certificate keyUsage would also indicate to the peer that the public key can be used for derivation.   

I will leave it up to you to handle this.

Jakuje:
I merged #2345 so we can close this one.

Issue #2291, opened 2021-04-07
On Windows, use LoadLibraryEx with LOAD_WITH_ALTERED_SEARCH_PATH if absolute path is given


qpernil:
### Problem Description

If a PKCS11 module is specified with an absolute path on Windows, and it has dependencies in the same directory (i.e. other DLLs it uses) those will not be found by LoadLibrary, which only searches the standard library path plus the directory the application that loads it was launched from, so it will fail to load the PKCS11 module with no indication why. 

### Proposed Resolution

Use LoadLibraryEx with the LOAD_WITH_ALTERED_SEARCH_PATH flag if the path given is absolute (LOAD_WITH_ALTERED_SEARCH_PATH requires an absolute path according to docs), and with a 0 flag if not. This will work the same as now for non-absolute paths, but will fix the above problem.

### Steps to reproduce

See problem description

### Logs

See for example https://github.com/Pkcs11Interop/Pkcs11Interop/pull/143

frankmorgner:
as far as I can see, this is only a problem for pkcs11-tool. OpenSC is linking all requirements statically to avoid these kinds of problems.

would you care to make a pull request?

qpernil:
I fixed the pull request, it seems to build now (except the last build failed but that looks temporary)

PR #2294, opened 2021-04-09
Use LoadLibrarEx with LOAD_WITH_ALTERED_SEARCH_PATH for absolute paths

qpernil:
If an absolute path is given set flags to find dependencies in the same directory, otherwise we can't load libraries that are in a directory that is not in the PATH.

Fixes https://github.com/OpenSC/OpenSC/issues/2291

### DIFF
--- a/src/common/libscdl.c
+++ b/src/common/libscdl.c
@@ -27,6 +27,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #include <shlwapi.h>
+#pragma comment(lib, "shlwapi")
 void *sc_dlopen(const char *filename)
 {
 	DWORD flags = PathIsRelativeA(filename) ? 0 : LOAD_WITH_ALTERED_SEARCH_PATH;
--- a/src/common/libscdl.c
+++ b/src/common/libscdl.c
@@ -27,7 +27,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #include <shlwapi.h>
-#pragma comment(lib, "shlwapi")
+
 void *sc_dlopen(const char *filename)
 {
 	DWORD flags = PathIsRelativeA(filename) ? 0 : LOAD_WITH_ALTERED_SEARCH_PATH;
--- a/src/common/libscdl.c
+++ b/src/common/libscdl.c
@@ -26,9 +26,11 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#include <shlwapi.h>
 void *sc_dlopen(const char *filename)
 {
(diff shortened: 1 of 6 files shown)


qpernil:
Using the pragma to link shlwapi might not be the way to go. I found that shlwapi.h was included in one other place, but couldn't find any mention of the library, so did it like this to make the PR small and self-contained.

vjardin:
Please, could you update your patch in order to fix:
https://travis-ci.org/github/OpenSC/OpenSC/jobs/766455267#L1511
```
libscdl.c:30:0: error: ignoring #pragma comment  [-Werror=unknown-pragmas]
 #pragma comment(lib, "shlwapi")
```
?

qpernil:
> Please, could you update your patch in order to fix:
> https://travis-ci.org/github/OpenSC/OpenSC/jobs/766455267#L1511
> 
> ```
> libscdl.c:30:0: error: ignoring #pragma comment  [-Werror=unknown-pragmas]
>  #pragma comment(lib, "shlwapi")
> ```
> 

OK, it's not being built by Visual Studio build tools I take it. I'll look into it.

qpernil:
I think this builds OK now, the fail seems to be unrelated to my change (and probably temporary)

qpernil:
Any comments on this, is it good to go ?

frankmorgner (approved):

Jakuje:
one more comment/suggestion. I see that `shlwapi.lib` is added to dozen of build targets in the makefiles and mostly appended to the same list of other libs. Would it make sense to add it to some variable that is used in all of them to avoid the repetition of the same list in all the targets? It would make also the lines a bit shorter and changes easier to review.

This is not a must, but I think it should make the windows makefiles a bit more readable.

frankmorgner:
Unfortunately, all Windows Makefiles are hand crafted right now. I'd leave this as is.

I've seen some cmake projects lately, which are doing a better job of supporting multiple platforms with different IDEs. Maybe this is something OpenSC could use, too. Also, there's a very need [package manager](https://conan.io/) available which integrates nicely...

Jakuje:
Hmm. This looks like breaking the mingw build. I am not sure if it is because of some missing dependency or it needs to be somehow gated for mingw. Can you check?

https://github.com/OpenSC/OpenSC/runs/3534149314

Jakuje:
This should be fixed with https://github.com/OpenSC/OpenSC/pull/2381/commits/520c88d6fdc0c1fe7cb88da9a475d3e3bac7ef77 from #2381 when merged.

PR #2295, opened 2021-04-09
If object doesn't found, pkcs11-spy show it's id

lo1ol:
Hello,
I've noticed that pkcs11-spy doesn't print id of an unknown object. Here is a fix for it.

Notice that inside this patch lookup_enum function returns malloced string. Do you have any idea how to avoid this?


Jakuje:
do you have an example of the output from the pkcs11-spy?

lo1ol:
```bash
26: C_DeriveKey
2021-04-02 23:15:53.735
[in] hSession = 0x1
pMechanism->type=(null)
[in] hBaseKey = 0x81c2b841
[in] pTemplate[7]: 
    CKA_CLASS             CKO_SECRET_KEY       
    CKA_KEY_TYPE          Value D4321006 not found for type CK_KEY_TYPE
    CKA_TOKEN             False
    CKA_MODIFIABLE        True
    CKA_PRIVATE           True
    CKA_EXTRACTABLE       True
    CKA_SENSITIVE         False
```

Jakuje:
I am no big fan of pkcs11-spy handling too much dynamic memory (for dumping each and every option). I would probably suggest to write it as it was written in some other cases, such as:
```
const char *name = lookup_enum(MEC_T, type);
if (name)
	fprintf(spy_output, "[in] type = %30s\n", name);
else
	fprintf(spy_output, "[in] type = Unknown Mechanism (%08lx)\n", type);
```
If used in more places, this can be encapsulated in some macro for consistency, such as

const char *name = lookup_enum(MEC_T, type);

lo1ol:
yes, I think so. Using templating make this code more nice:
#define FPRINTF_LOOKUP_ENUM(fmt, category, type)\
{\
        const char *name = lookup_enum((category), (type));\
        if (name)\
                fprintf(spy_output, (fmt), (name));\
        else {\
                size_t needed = snprintf(NULL, 0, "Unknown(0x%08lX)", (type)) + 1;\
                char  *buffer = malloc(needed);\
                sprintf(buffer, "Unknown(0x%08lX)", (type));\
                fprintf(spy_output, (fmt), buffer);\
                free(buffer);\
        }\
}

What do you think about it?

Jakuje:
> yes, I think so. Using templating make this code more nice:
> #define FPRINTF_LOOKUP_ENUM(fmt, category, type)
> {
> const char *name = lookup_enum((category), (type));
> if (name)
> fprintf(spy_output, (fmt), (name));
> else {
> size_t needed = snprintf(NULL, 0, "Unknown(0x%08lX)", (type)) + 1;
> char *buffer = malloc(needed);

Please, check the allocation.

> sprintf(buffer, "Unknown(0x%08lX)", (type));
> fprintf(spy_output, (fmt), buffer);
> free(buffer);
> }
> }

Make the outer `do { ... } while(0)` for compatibility.

> 
> What do you think about it?

I think it will make the change smaller and less complicated for both runtime and reviews.

lo1ol:
What do you mean here?
> Please, check the allocation.

Yes, ```c do { ... } while(0)``` statement will be more safe.

dengert:
I would say SPY was working correctly. If "type" is not found, it prints the HEX value of the type. 
PKCS11 allows for "vendor defined" values for a number of fields. In these the high order bit is set. This appears to be the case for  `D4321006`. 

SPY understands a few of these, "vendor provided" values for Netscape, where the value starts with `0xCE534350` See  https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/pkcs11-display.c#L32

I have never heard of a "D e r i  v e d .  K u z n  e c h i  k . t w  i n . k  e y" If it becomes a PKCS11 standard key type, then the type should be added to the above.

lo1ol:
> If "type" is not found, it prints the HEX value of the type.

But it's not work for all cases. You can notice that mechanism ID prints as Null:
> pMechanism->type=(null)

lo1ol:
May you please restart test?

lo1ol:
Can you please check my patch? Everething is ok?

src/pkcs11/pkcs11-object.c
```diff
@@ -281,8 +296,15 @@ C_GetAttributeValue(CK_SESSION_HANDLE hSession,	/* the session's handle */
 		}
 	}
 
-out:	sc_log(context, "C_GetAttributeValue(hSession=0x%lx, hObject=0x%lx) = %s",
+out:
+	;	
+	const char* name = lookup_enum ( RV_T, rv );
```
Jakuje, line 301: the variable definition can come only in the beginning of the block in the current C standard OpenSC is using.

src/pkcs11/pkcs11-session.c
```diff
@@ -27,6 +27,22 @@
...
+                size_t needed = snprintf(NULL, 0, "0x%08lX", (rv)) + 1;\
+                char *buffer = malloc(needed);\
+                if (buffer) {\
+                        sprintf(buffer, "0x%08lX", (rv));\
+                        sc_log(context, (fmt), buffer);\
+                        free(buffer);\
+                }\
+        }\
+} while(0)
+
```
Jakuje, line 45: Can you implement this macro in some header file to avoid code duplication? I see it already for the third time.

lo1ol:
any other changes?

Jakuje (approved):

lo1ol:
still waiting for...
image

Jakuje:
Sorry for a delay. Can you rebase the changes on top of current master? There were some changes that cause conflicts now.

lo1ol:
Resolve conflicts and fix typoes

src/pkcs11/pkcs11-session.c
```diff
@@ -251,6 +251,7 @@ CK_RV C_GetSessionInfo(CK_SESSION_HANDLE hSession,	/* the session's handle */
 	struct sc_pkcs11_session *session;
 	struct sc_pkcs11_slot *slot;
 	int logged_out;
+	const char* name;
```
Jakuje, line 254: we use the following style in opensc:
```suggestion
	const char *name;
```
The same pattern appears on couple of more places in this PR.

src/pkcs11/pkcs11-global.c
```diff
@@ -640,7 +641,11 @@ CK_RV C_GetSlotInfo(CK_SLOT_ID slotID, CK_SLOT_INFO_PTR pInfo)
 		memcpy(pInfo, &slot->slot_info, sizeof(CK_SLOT_INFO));
 
 	sc_log(context, "C_GetSlotInfo() flags 0x%lX", pInfo->flags);
-	sc_log(context, "C_GetSlotInfo(0x%lx) = %s", slotID, lookup_enum( RV_T, rv));
+	name = lookup_enum( RV_T, rv);
```
Jakuje, line 644:
```suggestion
	name = lookup_enum(RV_T, rv);
```

src/pkcs11/pkcs11-global.c
```diff
@@ -640,7 +641,11 @@ CK_RV C_GetSlotInfo(CK_SLOT_ID slotID, CK_SLOT_INFO_PTR pInfo)
...
+	if (name)
+		sc_log(context, "C_GetSlotInfo(0x%lx) = %s", slotID, lookup_enum( RV_T, rv));
```
Jakuje, line 646:
```suggestion
		sc_log(context, "C_GetSlotInfo(0x%lx) = %s", slotID, name);
```

src/pkcs11/pkcs11-object.c
```diff
@@ -281,8 +281,15 @@ C_GetAttributeValue(CK_SESSION_HANDLE hSession,	/* the session's handle */
 		}
 	}
 
-out:	sc_log(context, "C_GetAttributeValue(hSession=0x%lx, hObject=0x%lx) = %s",
+out:
+	name = lookup_enum ( RV_T, rv );
```
Jakuje, line 285: couple of bogus whitespace characters
```suggestion
	name = lookup_enum(RV_T, rv);
```

src/pkcs11/pkcs11-object.c
```diff
@@ -281,8 +281,15 @@ C_GetAttributeValue(CK_SESSION_HANDLE hSession,	/* the session's handle */
...
+	if (name)
+		sc_log(context, "C_GetAttributeValue(hSession=0x%lx, hObject=0x%lx) = %s",
 			hSession, hObject, lookup_enum ( RV_T, rv ));
```
Jakuje, line 288:
```suggestion
			hSession, hObject, name);
```

src/pkcs11/pkcs11-spy.c
```diff
@@ -404,13 +404,26 @@ static void
...
+	const char *mec_name;
 
 	if (!pMechanism) {
 		fprintf(spy_output, "[in] %s = NULL\n", name);
 		return;
 	}
 
-	fprintf(spy_output, "[in] %s->type = %s\n", name, lookup_enum(MEC_T, pMechanism->mechanism));
+	mec_name = lookup_enum(MEC_T, pMechanism->mechanism);
+	if (name)
```
Jakuje, line 415:
```suggestion
	if (mech_name)
```

lo1ol:
Thanks. I remove similar typoes according whole commits

Jakuje (approved):
looks good. Thanks for your contribution! If there will be no more comments, I will merge it in couple days.

lo1ol:
Thanks:)

Jakuje:
Maybe one bonus item would be adding some tests using pkcs11 spy into the testsuite under `tests/` here in the repository. At this moment, we have couple of scripts to run various use cases mostly with softhsm, so inserting the pkcs11-spy between pkcs11-tool and softhsm to at least check that it works in these cases should be enough. Checking the output to make sure the IDs or what you are after are in place would be the best.

It can certainly land in separate PR or later or I will probably get to implement something like that, but if you would be interested to check this, I would be happy to review and accept such change.

lo1ol:
Sorry, I havn't time for that now :( But if it would, i will try to help

PR #2302, opened 2021-04-20
iasecc: use proper printf format specifiers for size_t

marschap:
Do not hard-code the printf format specifier for size_t: use the macro instead.

This fixes compilation on 32-bit architectures.


vjardin:
LGTM, ok for this fix.

frankmorgner (approved):

PR #2311, opened 2021-04-27
Remove the SC_SEC_ENV_FILE_REF_PRESENT flag for Rutoken ECP cards

lolyonok:
Rutoken ECP cards have no default SE file. Previous cards ignored
MSE with restoring default SE, but new cards don't. This requires
SC_SEC_ENV_FILE_REF_PRESENT to be removed from env flags.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-rtecp.c
+++ b/src/libopensc/card-rtecp.c
@@ -366,6 +366,19 @@ static int rtecp_logout(sc_card_t *card)
 	SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, r);
 }
 
+static int rtecp_set_security_env(	struct sc_card *card,
+									const struct sc_security_env *env,
+									int se_num)
+{
+	struct sc_security_env se_env;
+	if(!env)
+		return SC_ERROR_INVALID_ARGUMENTS;
+
+	se_env= *env;
+	se_env.flags &= ~SC_SEC_ENV_FILE_REF_PRESENT;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-rtecp.c
```diff
@@ -366,6 +366,17 @@ static int rtecp_logout(sc_card_t *card)
...
 }
 
+static int rtecp_set_security_env(	struct sc_card *card,
+									const struct sc_security_env *env,
+									int se_num)
+{
+	assert(env);
+	struct sc_security_env se_env = *env;
+	se_env.flags &= ~SC_SEC_ENV_FILE_REF_PRESENT;
+	assert(iso_ops && iso_ops->set_security_env);
```
frankmorgner, line 376: please avoid `assert()` and use proper error handling

frankmorgner (changes requested):
Which cards did you test?

src/libopensc/card-rtecp.c
lolyonok: fixed args handling. removed assertion here cause didn't find any normal error for that. assertion seemed the best one

lolyonok:
> Which cards did you test?

this fixes apply only to rutoken cards and tokens.
we were testing our new cards (with same atr) and found this issue.
old cards and tokens work fine after this fix.

frankmorgner (approved):
thank you

PR #2314, opened 2021-04-30
Allow signing more than 512 bytes of data

### LINKED ISSUE #2300: [RFC] OpenSC should allow to sign data with length more than 512 bytes.
As a continuation of discussion in #2293, I've found that at least Yubikey is (almost certanly) supports ed25519 without any prehashing possible. This should work for small data, but won't work for e.g. X509 cettficates, which could be really longer than 512 bytes.

From RFC 8410, there are only OIDs for eddsa in pure mode, and no other hashing is defined, so we cannot prehash certificates ourselves.

I've managed to sign X509 using Yubikey and XCA with my changes, and openssl verifies this certificate, so I could assume that my understanding is correct.

So, one proposed solution is to change length of two buffers in OpenSC to be, for exampe, 2048 bytes, which should fit most X509 certificates, and still fits into max size of APDU

frankmorgner:
Fixes #2300

Note, that I've used `sc_mem_secure_alloc()` for saving the input data.

##### Checklist
- [x] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -581,8 +581,8 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	sc_security_env_t senv;
 	sc_algorithm_info_t *alg_info;
 	const struct sc_pkcs15_prkey_info *prkey = (const struct sc_pkcs15_prkey_info *) obj->data;
-	u8 buf[1024], *tmp;
-	size_t modlen;
+	u8 *buf = NULL, *tmp;
+	size_t modlen = 0, buflen = 0;
 	unsigned long pad_flags = 0, sec_flags = 0;
 
 	LOG_FUNC_CALLED(ctx);
@@ -612,9 +612,13 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	}
 
 	/* Probably never happens, but better make sure */
-	if (inlen > sizeof(buf) || outlen < modlen)
+	if (outlen < modlen)
 		LOG_FUNC_RETURN(ctx, SC_ERROR_BUFFER_TOO_SMALL);
 
+	buflen = inlen + 256;
+	buf = sc_mem_secure_alloc(buflen);
+	if (buf == NULL)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_OUT_OF_MEMORY);
 	memcpy(buf, in, inlen);
 
 	/* revert data to sign when signing with the GOST key.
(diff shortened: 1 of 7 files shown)


src/libopensc/pkcs15-sec.c
```diff
@@ -754,7 +757,8 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 		r = modlen;
 	}
 
-	sc_mem_clear(buf, sizeof(buf));
+err:
+	sc_mem_secure_free(buf, buflen);
```
Jakuje, line 761: would it make sense to keep the `sc_mem_clear()` here to zeroize the memory before freeing?

src/libopensc/pkcs15-sec.c
```diff
@@ -711,19 +714,19 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
...
+		LOG_TEST_GOTO_ERR(ctx, r, "Unable to add padding");
 		inlen = tmplen;
 	}
 	else if ( senv.algorithm == SC_ALGORITHM_RSA &&
 	          (flags & SC_ALGORITHM_RSA_PADS) == SC_ALGORITHM_RSA_PAD_NONE) {
 		/* Add zero-padding if input is shorter than the modulus */
 		if (inlen < modlen) {
-			if (modlen > sizeof(buf))
+			if (modlen > buflen)
 				return SC_ERROR_BUFFER_TOO_SMALL;
```
Jakuje, line 730: goto err;

src/pkcs11/mechanism.c
```diff
@@ -38,11 +38,51 @@ struct hash_signature_info {
...
+	return calloc(1, sizeof(struct signature_data));
+}
+
+static void
+signature_data_release(struct signature_data *data)
+{
+	if (!data)
+		return;
+	sc_pkcs11_release_operation(&data->md);
+	sc_mem_secure_free(data->buffer, data->buffer_len);
```
Jakuje, line 58: would it make sense to do `sc_mem_clear()` here too as it was in the previous implementation or using the secure memory implies this somehow (might be worth mentioning in comment?)?

Jakuje:
@frankmorgner did you get back to my review comments?

frankmorgner:
Thanks for the reminder, all comments should be fixed now.

There isn't a single case where we should call sc_secure_free without the need for clearing the memory. Therefor I've added `sc_secure_clear_free()`, which is used instead.

Jakuje (approved):

Jakuje:
I got back to this to verify it works as expected. I am able to sign arbitrary sizes with SoftHSM, but with Nitrokey (through OpenSC), I get only up to 256 bytes, for 384 and 512 bytes, I am getting `CKR_FUNCTION_FAILED` and for 640 and larger, I have just `CKR_GENERAL_ERROR`.

I was not able to figure out if the gnuk/nitrokey is supposed to support chaining or not from its source code. Certainly the card response weirdly to the chained APDUs, which is something I would like to investigate later.

But this PR is fine anyway.

Issue #2317, opened 2021-04-30
missing test-duplicate-symbols.sh


vjardin:
### Problem Description

`make distcheck` fails because of missing `test-duplicate-symbols.sh`

vjardin:
@frankmorgner performed the same fix at the same time,
  https://github.com/OpenSC/OpenSC/commit/32004e74ceeb2ce346ef8e0cf37fbe663320ccf0

let's close it and give up the pull request.

PR #2318, opened 2021-04-30
Fix missing test-duplicate-symbols.sh

vjardin:
make distcheck fails because this script is missing from the package.

Fix: commit b486965, issue #2317


frankmorgner:
thanks you. I've been working on this, too. it's already fixed.

PR #2322, opened 2021-05-01
Enable out-of-source build

### LINKED ISSUE #2319: make distcheck failures
### Problem Description

`make distcheck` fails
```
make[4]: Entering directory 'OpenSC/opensc-0.22.0-rc1/_build/sub/tests'
FAIL: test-manpage.sh
PASS: test-duplicate-symbols.sh
SKIP: test-pkcs11-tool-sign-verify.sh
SKIP: test-pkcs11-tool-test.sh
SKIP: test-pkcs11-tool-test-threads.sh
SKIP: test-pkcs11-tool-allowed-mechanisms.sh
============================================================================
Testsuite summary for OpenSC 0.22.0-rc1
============================================================================
# TOTAL: 6
# PASS:  1
# SKIP:  4
# XFAIL: 0
# FAIL:  1
```
because of
```
SKIP: test-pkcs11-tool-sign-verify.sh
=====================================

../../../tests/test-pkcs11-tool-sign-verify.sh: line 3: common.sh: No such file or directory
```
 since the commit 8c99e5076a829af618d0cb7ebb2b0b80e51831e1 from @Jakuje 

 - the path for common.sh is not properly set (main issue)
 - chmod 755 common.sh is missing (minor)

Jakuje:
Fixes #2319 

Tested with `make distcheck`, which is also added to the travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -252,7 +252,7 @@ script:
       fi;
     fi
   - if [ -z "$HOST" -a "${DO_COVERITY_SCAN}" != "yes" -a -z "$DO_SIMULATION" ]; then
-      make check && make dist || (cat tests/*log src/tests/unittests/*log && exit 1);
+      make check && make distcheck || (cat tests/*log src/tests/unittests/*log && exit 1);
     fi
   - if [ ! -z "$HOST" -a "${DO_COVERITY_SCAN}" != "yes" ]; then
       make install;
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -43,4 +43,4 @@ TIDY_FILES = \
 	libpkcs11.c libscdl.c
 
 check-local:
-	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(TIDY_FILES) -- $(TIDY_FLAGS); fi
+	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(addprefix $(srcdir)/,$(TIDY_FILES)) -- $(TIDY_FLAGS); fi
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -147,4 +147,4 @@ TIDY_FILES = \
 	#$(SOURCES)
 
 check-local:
-	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(TIDY_FILES) -- $(TIDY_FLAGS); fi
+	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(addprefix $(srcdir)/,$(TIDY_FILES)) -- $(TIDY_FLAGS); fi
--- a/src/pkcs11/Makefile.am
+++ b/src/pkcs11/Makefile.am
@@ -87,4 +87,4 @@ TIDY_FILES = \
 			 framework-pkcs15init.c debug.c
 
 check-local:
-	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(TIDY_FILES) -- $(TIDY_FLAGS); fi
+	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' -header-filter=.* $(addprefix $(srcdir)/,$(TIDY_FILES)) -- $(TIDY_FLAGS); fi
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -182,4 +182,4 @@ noinst_DATA = org.opensc-project.mac.pkcs11-register.plist org.opensc-project.ma
 	$(AM_V_GEN)$(do_subst) < $< > $@
 
 clean-local:
-	rm -f $(abs_builddir)/npa-tool.ggo $(abs_builddir)/opensc-notify.ggo $(abs_builddir)/opensc-asn1.ggo $(abs_builddir)/goid-tool.ggo $(abs_builddir)/egk-tool.ggo org.opensc.notify.desktop pkcs11-register.desktop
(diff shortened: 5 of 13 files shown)


Jakuje:
sgih ... looks like something got broken with the latest ubuntu in travis ... let me check that later.

@vjardin can you check this solves the issues for you?

LudovicRousseau:
You should not need to add `$(srcdir)/` everywhere in the `Makefile.am` files.
The problem was the missing `pkcs15-iasecc.h` file. That problem is already fixed with  32004e74ceeb2ce346ef8e0cf37fbe663320ccf0

Also `make distcheck` already does `make dist`. So using `make dist && make distcheck` is redundant.

Since you run `make check` in the source code directory you should not have to modify the scripts to add `$SOURCE_PATH/`

I propose to NOT merge this PR.

Jakuje:
> You should not need to add `$(srcdir)/` everywhere in the `Makefile.am` files.
> The problem was the missing `pkcs15-iasecc.h` file. That problem is already fixed with [32004e7](https://github.com/OpenSC/OpenSC/commit/32004e74ceeb2ce346ef8e0cf37fbe663320ccf0)

This fixes problem of `make dist`, but this fixes the problem of running `distcheck`, which builds the package out-of-source, but the tests scripts were initially not written for that. Nor the clang-tidy support. If there is some better way to add it to the generic script (=non-magic variables used by autotools), feel free to suggest better way how to handle this. I did not find a better way.

> Also `make distcheck` already does `make dist`. So using `make dist && make distcheck` is redundant.

Fair point. I will remove the dist.

> Since you run `make check` in the source code directory you should not have to modify the scripts to add `$SOURCE_PATH/`

The `make check` does not have to be running from the source code directory. If you run build out-of-source as `distcheck` does, you are running `make check` from build directory.

frankmorgner:
I agree with @Jakuje . the local checks with clang-tidy and the bash scripts are executed from the build directory and the variables they are fed with are used as is (i.e. the source files as they are, without the directory).

frankmorgner:
(ideally we should feed clang-tidy with the correct variable that contains the source files of the build (which are prepended with the directory, automatically, I think). but as far as i know we had a lot of warnings with some files which was the reason to exclude some of them)

vjardin:
Thanks, it is better.

FYI I am still getting some errors ([test-suite.log.gz](https://github.com/OpenSC/OpenSC/files/6410968/test-suite.log.gz)) but I guess they are not related. If you believe so, let's postpone this issue into another issue I'll open.

What's about you ?
Maybe I'll need to update the Ubuntu's SoftHSM2.
```
FAIL: test-pkcs11-tool-sign-verify.sh
=====================================

Testing /usr/local/lib/softhsm/libsofthsm2.so
Testing /usr/lib/softhsm/libsofthsm2.so
Setting P11LIB=/usr/lib/softhsm/libsofthsm2.so
Setup SoftHSM
The token has been initialized and is reassigned to slot 1940784900
Using slot 0 with a present token (0xff50e7c)
Key pair generated:
Private Key Object; RSA
  label:      RSA_auth
  ID:         01
  Usage:      decrypt, sign, unwrap
  Access:     sensitive, always sensitive, never extractable, local
Public Key Object; RSA 1024 bits
  label:      RSA_auth
  ID:         01
  Usage:      encrypt, verify, wrap
  Access:     local
Using slot 0 with a present token (0xff50e7c)
writing RSA key
Using slot 0 with a present token (0xff50e7c)
error: PKCS11 function C_GenerateKeyPair failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
Couldn't generate RSA:2048 key pair
Main C_Initialize(NULL) rv:CKR_GENERAL_ERROR
error: PKCS11 function C_Initialize failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
Couldn't generate EC:secp256r1 key pair
Main C_Initialize(NULL) rv:CKR_GENERAL_ERROR
error: PKCS11 function C_Initialize failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
Couldn't generate EC:secp521r1 key pair
```

LudovicRousseau:
I understand now why you need to update the tests scripts.
`make distcheck` will also call `make check` (or something similar).

But you should not have to modify the `Makefile.am` to add `$(srcdir)/`

Jakuje:
> Thanks, it is better.
> 
> FYI I am still getting some errors ([test-suite.log.gz](https://github.com/OpenSC/OpenSC/files/6410968/test-suite.log.gz)) but I guess they are not related. If you believe so, let's postpone this issue into another issue I'll open.
> 
> What's about you ?
> Maybe I'll need to update the Ubuntu's SoftHSM2.

What Ubuntu version and softhsm version are you using? It works fine with the one that executes this in travis (bionic) as well as for me with Fedora 33.

> But you should not have to modify the `Makefile.am` to add `$(srcdir)/`

Ok, after some more thinking about the issue, the other (maybe better and easier) option might be to do `cd $(srcdir)` before calling the `clang-tidy` to make sure it can find the files. Let me see if it will work tomorrow.

Jakuje:
Changing to the source directory did not work as it can not find the `config.h` so this is final version, unless somebody else will suggest better way than including the $(srcdir) for each file for clang-tidy.

LudovicRousseau:
If I am correct, clang-tidy is **not** started by `make distcheck`.
So you can use `make distcheck` to verify that all the files are present in the archive.
And use a normal `./configure` in the source directory when you use clang-tidy. No?

Yes, you would run `./configure` multiple times with different configurations but is that really a problem for CI/CD?

Jakuje:
> If I am correct, clang-tidy is **not** started by `make distcheck`.

It is, if you have `clang-tidy` installed. It is invoked from the `check-local` in the above changed makefiles, which is invoked by `check`.

LudovicRousseau:
> It is, if you have `clang-tidy` installed. It is invoked from the `check-local` in the above changed makefiles, which is invoked by `check`.

Maybe it is possible to run `make distcheck` with clang-tidy checks _disabled_.
Something like (untested) `CLANGTIDY=invalid make distcheck`.

vjardin:
@LudovicRousseau , @Jakuje , what do you propose as a final "solution" to be tested/checked on Linux distributions ? I am ready to provide some helps in running / testing `make distcheck`, just tell me which patch I should cherry pick.

Jakuje:
I searched a bit more and find what I did not find initially: `addprefix` function in makefile, which does exactly what I needed so I rebased the current branch, dropping the modification of all the files and doing this only on the clang-tidy command-line.

frankmorgner:
Nice find! It looks like [addprefix](https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html) will always be available in GNU make, but I'm not so sure about FreeBSD's (macOS') make, for example...

Although this is more code, using a Automake feature (`$(srcdir)`) should be more portable. What do you think, @LudovicRousseau ?

Jakuje:
Given that this is not part of the main build, but only used in the check and when clang-tidy is available, I think it is a reasonable tradeoff for less "code".

LudovicRousseau:
OK for me.
Nice work.

PR #2323, opened 2021-05-02
fix possible access outside the array.

ihsinme:
if 5000 bytes are read, then at the end of the array we will write zero beyond its boundaries, damaging the stack.
Here's a simple solution. if you see the need to increase the array itself, let me know.


Jakuje (approved):

frankmorgner:
thanks

PR #2324, opened 2021-05-03
pkcs15-tool: Write data objects in binary mode

divinehawk:
Writing out data objects to a file in win32 results in all instances of 0x0a being replaced with 0x0d0a since the output file is opened in text mode.


vjardin:
LGTM

Jakuje (approved):

frankmorgner:
thanks

Issue #2321, opened 2021-05-01
certutil -scinfo fails


vjardin:
### Problem Description

On windows, most tools work (pkcs11-tool --test OK, firefox OK, ssh-keygen -D + ssh -I OK), but `certutil -scinfo` fails with the following error:
```
C:\Program Files\OpenSC Project\OpenSC\pkcs11>certutil -scinfo
The Microsoft Smart Card Resource Manager is running.
Current reader/card status:
Readers: 1
  0: Gemalto USB SmartCard Reader 0
--- Reader: Gemalto USB SmartCard Reader 0
--- Status: SCARD_STATE_PRESENT | SCARD_STATE_INUSE
--- Status: The card is being shared by a process.
---   Card:
---    ATR:
        3b ac 00 40 2a 00 12 25  00 64 80 00 03 10 00 90   ;..@*..%.d......
        00                                                 .


=======================================================
Analyzing card in reader: Gemalto USB SmartCard Reader 0
SCardGetCardTypeProviderName: The system cannot find the file specified. 0x2 (WIN32: 2 ERROR_FILE_NOT_FOUND)
Cannot retrieve Provider Name for SCardGetCardTypeProviderName: The system cannot find the file specified. 0x2 (WIN32: 2 ERROR_FILE_NOT_FOUND)
Cannot retrieve Provider Name for
--------------===========================--------------
CertUtil: -SCInfo command FAILED: 0x2 (WIN32: 2 ERROR_FILE_NOT_FOUND)
CertUtil: The system cannot find the file specified.
```

I do not have any idea how to debug/trace it.

dengert:
`./win32/customactions.cpp` needs to be updated with ATRs for your cards and driver.

vjardin:
Thanks, I'll try it. For my information, why all drivers from OpenSC do not get their ATRs included automatically ? Is it on purpose ?

dengert:
Windows is 74% of desktop market according to: https://gs.statcounter.com/os-market-share/desktop/worldwide so vendors will provide a windows driver. Some actually add support to OpenSC so they don't have to write their own! But most don't want to put in the effort to support anything but Windows and maybe Mac OS.  Microsoft provides a Microsoft PIV driver that looks for PIV AID  not the ATR.  MacOS also does or used to support PIV driver. (Someone else can comment on that.)  Gov issued ID or health smartcards have a vendor provided driver and users use or have access to a windows machine to uses these cards. 

[...]

PR #2329, opened 2021-05-05
CPx: add registration for Windows/minidrivers

vjardin:
Let's OpenSC be able to support the IAS/ECC CPx cards.

Suggested-by: Doug Engert <deengert@gmail.com>

Fix: issue #2321


Jakuje:
lgtm.

This issue was raised before that we usually update ATRs in the libopensc, but not in the minidriver, which causes these weird issues. I was suggesting we should have some automated check to verify that all the ATR mapping in card drivers are available also in the minidriver, but as I am not a Windows user, it is not high priority for me. I think either simple test in C or bash script that would be executed during `make check` should be fine.

frankmorgner:
Difficult...

1. With ATR masking this could be quite tricky to detect whether a card driver is also registered via Minidriver
2. Most vendors provide a proprietary implementation for Windows, which may be preferred by users. Unfortunately, Windows just picks the first minidriver it sees without letting the user decide which implementation to use.

Making this configuration part of our installer, would be my preferred approach, but I don't currently have time for this...

@vjardin is there a proprietary driver for your card thay may be conflicting with OpenSC?

vjardin:
I agree with @frankmorgner about the checks. I guess some comments into the OpenSC's minidriver code would be a good start.

Yes there is a proprietary driver but I do not understand the issues: anyone can decide not to install OpenSC but get the proprietary one. Of course if both are installed, it becomes an "undefined" state, but it is up to the users that decide to install or not these drivers.

Jakuje (approved):

frankmorgner:
OK, then let's agree on adding CPx to OpenSC's minidriver configuration.

However, I agree with Doug that Microsoft isn't very verbose regarding the ATR mask. @vjardin did you verify that your comment is actually correct?

vjardin:
Thanks @dengert : how should we proceed ? Should I cut/paste these lines as comments into the `minidriver.c` file ? I feel it is the best option in order to keep some awareness about it.

Or B-plan, what's about adding a section regarding Windows minidriver into the wiki (https://github.com/OpenSC/OpenSC/wiki/Adding-a-new-card-driver) ?

dengert:
Add URL would work.

frankmorgner:
done, thank you

vjardin:
Thanks. Some users will like it ;)

PR #2333, opened 2021-05-07
pkcs11.h: avoid C++ comments

alonbl:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11.h
+++ b/src/pkcs11/pkcs11.h
@@ -374,8 +374,10 @@ typedef unsigned long ck_key_type_t;
 #define CKK_EC_MONTGOMERY	(0x41UL)
 #define CKK_VENDOR_DEFINED	(1UL << 31)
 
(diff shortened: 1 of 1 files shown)


frankmorgner (commented):
`//` is allowed since c99, why do you want this change?

alonbl:
On Fri, 7 May 2021 at 15:33 Frank Morgner ***@***.***> wrote:

> ***@***.**** commented on this pull request.
>
> // is allowed since c99, why do you want this change?
>

Hi Frank,
This file originating from scute is now used in many open source projects.
As OpenSC  continues to maintain it while scute is unmaintained, it would
be very nice if it can be taken as is without modifications to other
projects which still uses pedantic compatibility.
The only issue is this comment… So I guess we can provide this service.
Regards.
Alon

frankmorgner:
OK

Issue #2308, opened 2021-04-24
Building OpenSC with OpenSSL 3.0.0 


dengert:
### Problem Description

OpenSSL version 3.0.0 is now in alpha testing, and will soon be in beta. OpenSC can be built without OpenSSL, but many card drivers depend on some version of OpenSSL. 

OpenSC can used multiple versions of OpenSSL by following the 1.1 API and accommodates older versions and LibreSSL with a few \#ifdefs and [sc-ossl-compat.h](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/sc-ossl-compat.h) which defines some macros and inline code to handle the backward compatibility. 

[OpenSSL_3.0](https://wiki.openssl.org/index.php/OpenSSL_3.0) makes some major changes. and now includes some backward compatibility features similar to sc-ossl-compat.h and allows for compiling with multiple APIs using `CPPFLAGS="-DOPENSSL_API_COMPAT=0x10101000L"` for 1.1.1.

The following tests were all run with a modified `sc-ossl-compat.h` which was easier then I had though it would be:
[sc-ossl-compat.c.diff.txt](https://github.com/OpenSC/OpenSC/files/6370897/sc-ossl-compat.c.diff.txt)

On Ubuntu 20.04, OpenSSL 3.0.0 was built from git and installed into `/opt/ossl-3.0.0`
https://www.openssl.org/source/gitrepo.html I used  git://git.openssl.org/openssl.git

To include the legacy code for SHA1, DES, RIPEMD160 and other code used by OpenSC, /opt/ossl-3.0.0/ssl/openssl.cnf was modified as suggested in: https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers 

OpenSC was then built using:
```
CPPFLAGS="-DOPENSSL_USE_DEPRECATED -DDEBUG -DOPENSSL_API_COMPAT=0x10101000L
CFLAGS=""
LDFLAGS="-Wl,-rpath,/opt/ossl-3.0.0/lib"
../OpenSC/configure --disable-silent-rules --prefix=/opt/ossl-3.0.0 --enable-pcsc \
 --with-completiondir=/opt/ossl-3.0.0/etc/bash_completion.d \
 --with-pcsc-provider=/usr/lib/libpcsclite.so --enable-doc --disable-strict
```
results in 16 warnings, but `./pkcs11-tool --test --login` using PIV demo card works.

### Proposed Resolution Questions

[...]

PR #2337, opened 2021-05-12
First draft of migration to new OpenSSL 3.0 API

Jakuje:
Related: #2308

What works:
 * Digest are done using EVP_Digest. This is compatible with the old versions
 * Using EVP_* interface for 3DES encryption/decryption (not converted everywhere yet). Does not work with older OpenSSL.
 * Using EVP_* interface for DES encryption with legacy provider
 * Encoders and decoders for key structures should work (I did not change in in all places though yet

What does not:
 * Accessing key parameters from EVP_PKEY API (RSA keys -- fails for me with the first check `evp_pkey_is_provided()`)
 * Accessing key parameters from EVP_PKEY API for EC keys (I did not find a way how to get `EC_GROUP` structure from the `EVP_PKEY` or modify/get anything from the group structure
 * iasecc is doing some [weird stuff with SHA hashes](https://github.com/OpenSC/OpenSC/blob/f0c059ede81871e4d4e388d51e2b8353cfdd9182/src/libopensc/card-iasecc.c#L3195). @vjardin could you check what it is and if there is a way to write it in some sensible way without touching OpenSSL internal structures? 
 * gost card is using some deprecated engine calls, which I am not familiar with at all

The PR is indeed incomplete and more like a draft. It will require quite some work to get it into the shape, working with new API, probably also on the OpenSSL side as it looks like some things are not possible yet (or I do not know how to do them).

I tried to rewrite at least most of the bits of the `p11tests`, to identify the most common issues and verified them against softhsm. I also wanted to submit the PR to see how the (old) CI systems will handle it.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-authentic.c
+++ b/src/libopensc/card-authentic.c
@@ -1318,9 +1318,9 @@ authentic_pin_verify(struct sc_card *card, struct sc_pin_cmd_data *pin_cmd)
 	}
 
 	if (pin_cmd->pin1.data)
-		SHA1(pin_cmd->pin1.data, pin_cmd->pin1.len, pin_sha1);
+		EVP_Digest(pin_cmd->pin1.data, pin_cmd->pin1.len, pin_sha1, NULL, EVP_sha1(), NULL);
 	else
-		SHA1((unsigned char *)"", 0, pin_sha1);
+		EVP_Digest((unsigned char *)"", 0, pin_sha1, NULL, EVP_sha1(), NULL);
 
 	if (!memcmp(pin_sha1, prv_data->pins_sha1[pin_cmd->pin_reference], SHA_DIGEST_LENGTH))   {
 		sc_log(ctx, "Already verified");
--- a/src/libopensc/card-cac.c
+++ b/src/libopensc/card-cac.c
@@ -44,6 +44,7 @@
 
 #ifdef ENABLE_OPENSSL
 #include <openssl/sha.h>
+#include <openssl/evp.h>
 #endif /* ENABLE_OPENSSL */
 
 #include "internal.h"
@@ -1707,7 +1708,7 @@ static int cac_populate_cac_alt(sc_card_t *card, int index, cac_private_data_t *
 		if (priv->cac_id == NULL) {
 			return SC_ERROR_OUT_OF_MEMORY;
 		}
-		SHA1(val, val_len, priv->cac_id);
+		EVP_Digest(val, val_len, priv->cac_id, NULL, EVP_sha1(), NULL);
 		priv->cac_id_len = 20;
 		sc_debug_hex(card->ctx, SC_LOG_DEBUG_VERBOSE,
 		    "cuid", priv->cac_id, priv->cac_id_len);
--- a/src/libopensc/card-cac1.c
+++ b/src/libopensc/card-cac1.c
@@ -44,6 +44,7 @@
 
 #ifdef ENABLE_OPENSSL
 #include <openssl/sha.h>
+#include <openssl/evp.h>
 #endif /* ENABLE_OPENSSL */
 
 #include "internal.h"
@@ -441,7 +442,7 @@ static int cac_populate_cac1(sc_card_t *card, int index, cac_private_data_t *pri
 			return SC_ERROR_OUT_OF_MEMORY;
 		}
 #ifdef ENABLE_OPENSSL
-		SHA1(val, val_len, priv->cac_id);
+		EVP_Digest(val, val_len, priv->cac_id, NULL, EVP_sha1(), NULL);
 		priv->cac_id_len = 20;
 		sc_debug_hex(card->ctx, SC_LOG_DEBUG_VERBOSE,
 			"cuid", priv->cac_id, priv->cac_id_len);
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -183,7 +183,7 @@ iasecc_chv_cache_verified(struct sc_card *card, struct sc_pin_cmd_data *pin_cmd)
 
 	pin_status->reference = pin_cmd->pin_reference;
 	if (pin_cmd->pin1.data)
-		SHA1(pin_cmd->pin1.data, pin_cmd->pin1.len, pin_status->sha1);
+		EVP_Digest(pin_cmd->pin1.data, pin_cmd->pin1.len, pin_status->sha1, NULL, EVP_sha1(), NULL);
 	else
 		memset(pin_status->sha1, 0, SHA_DIGEST_LENGTH);
 
@@ -246,7 +246,7 @@ iasecc_chv_cache_is_verified(struct sc_card *card, struct sc_pin_cmd_data *pin_c
 	LOG_FUNC_CALLED(ctx);
 
 	if (pin_cmd->pin1.data)
-		SHA1(pin_cmd->pin1.data, pin_cmd->pin1.len, data_sha1);
+		EVP_Digest(pin_cmd->pin1.data, pin_cmd->pin1.len, data_sha1, NULL, EVP_sha1(), NULL);
 	else
 		memset(data_sha1, 0, SHA_DIGEST_LENGTH);
 	sc_log_hex(ctx, "data_sha1: %s", data_sha1, SHA_DIGEST_LENGTH);
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -50,6 +50,7 @@
 #include "errors.h"
 #ifdef ENABLE_OPENSSL
 #include <openssl/sha.h>
+#include <openssl/evp.h>
 #endif /* ENABLE_OPENSSL */
 
 #include "card-openpgp.h"
@@ -2647,7 +2648,7 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
 	p = NULL;
 
 	/* hash with SHA-1 */
-	SHA1(fp_buffer, fp_buffer_len, fingerprint);
+	EVP_Digest(fp_buffer, fp_buffer_len, fingerprint, NULL, EVP_sha1(), NULL);
 	free(fp_buffer);
 
 	/* store to DO */
--- a/src/libopensc/cwa14890.c
+++ b/src/libopensc/cwa14890.c
@@ -41,6 +41,7 @@
 #include <openssl/x509.h>
 #include <openssl/des.h>
 #include <openssl/rand.h>
+#include <openssl/evp.h>
 #include "cwa14890.h"
 #include "cwa-dnie.h"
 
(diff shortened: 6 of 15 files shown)


vjardin:
> * iasecc is doing some [weird stuff with SHA hashes](https://github.com/OpenSC/OpenSC/blob/f0c059ede81871e4d4e388d51e2b8353cfdd9182/src/libopensc/card-iasecc.c#L3195). @vjardin could you check what it is and if there is a way to write it in some sensible way without touching OpenSSL internal structures?

Same, I do not understand what it does. Is @viktorTarasov still getting these messages ? I'll try to better understand the code, but it is not my best expertise.

Regarding SM: I have some cards that are supposed to support SM, but I have not been able to understand yet how to get it running. For instance, https://github.com/OpenSC/OpenSC/blob/master/etc/opensc.conf.example.in#L823, I do not understand how to get/set the keyset_xyz values such as "RO_ENC_TEST_KEY_" ?! Once I could get it running, I'll be able to test this serie.

dengert:
I too have been looking at the problem. 

- Alpha-15 had SHA1 as deprecated.  master has reversed that: 4d49b68504cc494e552b
"Crypto: Add deprecation compatibility declarations for SHA* message digest functions            
        Also add hints to SHA256_Init.pod and CHANGES.md how to replace SHA256() etc."

- Have been trying to build with 3.0.0 API. A lot of RSA, DSA, and EC stuff is also deprecated. 

-  having to do something kike this is pkcs15-prkey.c: 
```
77 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
678                 BIGNUM *src_n = NULL, *src_e = NULL, *src_d = NULL,
679                                 *src_p = NULL, *src_q = NULL,
680                                 *src_iqmp = NULL, *src_dmp1 = NULL, *src_dmq1 = NULL;
681                 OSSL_PARAM rsa_params[] = {
682                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_N, &src_n, sizeof(src_n)),
683                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_E, &src_e, sizeof(src_e)),
684                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_D, &src_d, sizeof(src_d)),
685                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_FACTOR1, &src_p, sizeof(src_p)),
686                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_FACTOR2, &src_q, sizeof(src_q)),
687                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_EXPONENT1, &src_dmp1, sizeof(src_dmp1)),
688                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_EXPONENT2, &src_dmq1, sizeof(src_dmq1)),
689                         OSSL_PARAM_BN(OSSL_PKEY_PARAM_RSA_COEFFICIENT1, &src_iqmp, sizeof(src_iqmp)),
690                         OSSL_PARAM_END
691                 };
692 
693                 if (EVP_PKEY_todata(pk, EVP_PKEY_KEY_PARAMETERS, &&rsa_params) != 1) {
694                         r = SC_ERROR_NOT_SUPPORTED;
695                         goto err_rsa;
696                 }
697 #else
It compiles, but DSA and EC also need to be conveted in a lot of places too

vjardin:
@Jakuje : please, could you extract `sm.c` from this pull request and have a specific one that could be included into OpenSC independently of the migration to OpenSSL3.0 ? It would even be better to merge this `sm.c` unittest prior to this serie.

viktorTarasov:
> > * iasecc is doing some [weird stuff with SHA hashes](https://github.com/OpenSC/OpenSC/blob/f0c059ede81871e4d4e388d51e2b8353cfdd9182/src/libopensc/card-iasecc.c#L3195). @vjardin could you check what it is and if there is a way to write it in some sensible way without touching OpenSSL internal structures?
> 
> Same, I do not understand what it does. Is @viktorTarasov still getting these messages ? I'll try to better understand the code, but it is not my best expertise.
> 
> Regarding SM: I have some cards that are supposed to support SM, but I have not been able to understand yet how to get it running. For instance, https://github.com/OpenSC/OpenSC/blob/master/etc/opensc.conf.example.in#L823, I do not understand how to get/set the keyset_xyz values such as "RO_ENC_TEST_KEY_" ?! Once I could get it running, I'll be able to test this serie.

1.
In IASECC qualified signature (normally it's a signature with non-repudiation)  the last hash iteration has to be done by card. 
That's why in the QSign APDU data there is pre-hash (hash of input data without the last segment)  and last segment of input data.
The internal openssl structures where used to get the pre-hash value. 

2.
_RO_ENC_TEST_KEY__ is value of keyset key, ENC or MAC.
In configuratioon example there are default values of different demo IASECC cards and profiles. 
This value has to be seen as 16 bytes binary value.
The value is to be secretely shared between card and remote service .
The long name of the keyset key contains the ADF (for keyset SDO created on application level) and keyset SDO index . 
Name of keyset created on MF level has no ADF part.

Jakuje:
@dengert I tried something like you with https://github.com/OpenSC/OpenSC/pull/2337/files#diff-425869f5411dfdcf6053cef2ed9c6d5567d4dd96b59cf79ce40aaa866ab9a0b9R329-R330 but it did not work. Let me try your way if it works.

Thank you for the info about the reversal of the deprecation. I do not follow the releases that closely. If that is not indeed deprecated, we should be fine to use it further (or just apply it to have peace in mind for the future releases which might deprecate it? Or deprecate the new EVP_Digest? Who knows ...).

vjardin:
> > iasecc is doing some weird stuff with SHA hashes.

> In IASECC qualified signature (normally it's a signature with non-repudiation) the last hash iteration has to be done by card.
> That's why in the QSign APDU data there is pre-hash (hash of input data without the last segment) and last segment of input data.
> The internal openssl structures where used to get the pre-hash value.

Thanks for this tips. It helps a lot. Just in case, do you have some ideas how it could be "updated" without using the OpenSSL internal structures ?

> > Regarding SM:

> _RO_ENC_TEST_KEY__ is value of keyset key, ENC or MAC.
> The value is to be secretly shared between card and remote service .
> The long name of the keyset key contains the ADF (for keyset SDO created on application level) and keyset SDO index .
> Name of keyset created on MF level has no ADF part.

Interesting: since it is shared between the cards and the remote service, assuming there are 1M cards (same supplier) with the same "secret" ; it could mean that thousands of systems are deployed with this same secret so how "safe" is this secret/secure messaging mechanism ?

dengert:
The OSSL_PARAM stuff compiled, but I don't know if it works or not. too many other errors. 

Also note 3DES is not deprecated, but they removed the DES_LONG, DES_cblock and DES_key_schedule. I had added them to sc-ossl-compat.h. But I like the replacement of any symmetric key with "unsigned char *"  which is what they are doing. 

https://www.openssl.org/docs/manmaster/man3/EVP_des_ede3_cbc.html  Single des is provided by legacy provider, but it looks like 2DES and 3DES are provided by default provider if using EVP. 

What I was trying to do, is get a subset of OpenSC to work by building OpenSSL  with "no-ghost" and  and also with "no-gost no-deprecated". Then built OpenSC against both with combinations of:  -DOPENSSL_USE_DEPRECATED, -DOPENSSL_API_COMPAT=0x10002000L, -DOPENSSL_API_COMPAT=0x10101000L or -DOPENSSL_API_COMPAT=30000 

I also modified  21 card drivers and SM files to build dummy files so I was left with a few drivers and files that I could test using cards I have. 

This is going to be a big project, and will need testing by people who have cards to test with.  I have not gotten as far as you have.

viktorTarasov:
> Interesting: since it is shared between the cards and the remote service, assuming there are 1M cards (same supplier) with the same "secret" ; it could mean that thousands of systems (remote service/servers) are deployed with this same secret so how "safe" is this secret/secure messaging mechanism ?

What you see in a public project are the default keysets values of cards for testing .
Beside this scope there is the procedure of card personalisation for particular customer. (IASECC applet is loaded into JAVA card, personalised, then locked) .
Customer can do this procedure himself. 
During this procedure the keysets are diversified for each card, using, for example serial number .

viktorTarasov:
> Just in case, do you have some ideas how it could be "updated" without using the OpenSSL internal structures ?

Only some dumb idea - include HASH procedure into OpenSC code.

I imagine that around us there are a lot of such implementaitons.
For example in 'MbedTLS', 'WolfSSL' crypto libraries ...
The code in these last two, IMHO, is much more comprehensible then in OpenSSL .
 
At the limit one can venture to imagine to completely replace OpenSSL with MbedTLS ...

vjardin:
> (IASECC applet is loaded into JAVA card, personalized, then locked) .
> During this procedure the keysets are diversified for each card, using, for example serial number .

@viktorTarasov , please, just in case, how/where can we buy some cards with a "public" shared secret that I could use for testing+using the IASECC framework ? I have some IAS/ECC cards, but I do not know their shared secret, so no way for me to use them in order to check this "secure messaging" support.

Jakuje:
My bad. I had very outdated alpha13. Now, I updated to alpha15, which should be hopefully more feature-complete.

But anyway, the only use of `EVP_PKEY_todata()` I was able to find is in the openssl tests and which looks completely different than your example:

https://github.com/openssl/openssl/blob/master/test/evp_extra_test2.c#L633

I tried to use it inside of the p11test (its the place where I can test it the easiest), but I ended up with openssl crashing so I filled the openssl/openssl#15290 to fix the crash and get some clarifications.

Using some different crypto backend might cause some issues as it will always be less common dependency than OpenSSL. OpenSSL is available quite much everywhere. For example mbedTLS is not available in RHEL at all (only EPEL). Rewriting all the legacy code to it might be also quite some feat, making sure it will keep working as expected with completely different API (but the OpenSSL 3.0 API is quite some way from previous versions).

dengert:
I was running off a clone from: git@github.com:openssl/openssl.git from 
"8be513ae46 (HEAD -> master, origin/master, origin/HEAD) Mark pop/clear error stack in der2key_decode_p8"
30 or so commits ahead of: tag: openssl-3.0.0-alpha16

The OSSL_PARAMS compiles, but I have not gotten much further. As you said: "OpenSSL 3.0 API is quite some way from previous versions."

They have a OSSL_PARAM_construct_*  to build a parameter list and a a way to define the array static which I was using. 

They are adding ASN.1 code too if one could request an SPKI , Opensc could decode the ASN.1. 

They are in alpha, and still writing code.

Jakuje:
This looks like it will get delayed so I will try to split out what can be merged now into separate PR.

Jakuje:
Closing as the more complete changes are ready for reviews in #2438.

dengert:
See comment: https://github.com/OpenSC/OpenSC/pull/2506#issuecomment-1031527137 on OpenSSL 3.0.0 changed need to expose SAH data

PR #2345, opened 2021-05-21
p11test: Expect DERIVE to be set on both private and public key

Jakuje:
Basically reverts part of 485b6cf, which turned out to be wrong.

Alternative to #2292


dengert:
Looks good.

PR #2349, opened 2021-06-03
Add support for applet 1.8

yoe:
New ID card types are being issued in Belgium, with secp384r1 ECC keys rather than RSA ones and a new ATR. Update card-belpic.c to support those cards, too.

TODO: understand better how to select the correct algorithms for ECC. Help?

### DIFF
--- a/src/libopensc/card-belpic.c
+++ b/src/libopensc/card-belpic.c
@@ -146,6 +146,8 @@ static long t1, t2, tot_read = 0, tot_dur = 0, dur;
 static size_t next_idx = (size_t)-1;
 
 static const struct sc_atr_table belpic_atrs[] = {
+	/* Applet v1.8 */
+	{ "3B:7F:96:00:00:80:31:80:65:B0:85:04:01:20:12:0F:FF:82:90:00", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
 	/* Applet V1.1 */
 	{ "3B:98:13:40:0A:A5:03:01:01:01:AD:13:11", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
 	/* Applet V1.0 with new EMV-compatible ATR */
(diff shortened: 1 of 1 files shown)


dengert:
EC keys can be used for signature ECDSA and for key derivation ECDH. https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman

sc_format_oid does not format a curvename.

You are passing the pointer to the sc_object_id. This needs to be a static. Grep for _sc_card_add_ec_alg. You will note only  ./card-myeid.c and ./card-openpgp.c pass a address of static curve OID. Most other card drivers use NULL.

Jakuje:
@yoe was there any progress? I recently merged #2455 which added the ATR you used, but not the EC parts.

yoe:
> @yoe was there any progress?

Sorry, no. I've been too busy doing other things.

> I recently merged #2455 which added the ATR you used, but not the EC parts.

Thanks.

If someone else is interested in tackling this, the documentation on how the applet 1.8 cards act, and how they differ from the 1.7 ones, can be found at https://github.com/Fedict/eid-mw/blob/master/doc/sdk/documentation/Applet%201.8%20eID%20Cards/Public_Belpic_Applet_v1%208_Ref_Manual_v1%20.pdf

If nobody else manages, I might get around to doing this eventually, but given how busy I am with other things, don't hold your breath...

yoe:
Nothing has changed since my last comment on this PR.

I got stuck because the changes I made didn't work in ways that tell me that I don't really understand the internals of OpenSC very well, mostly because I haven't don't much OpenSC work in the past (the one fire-and-forget patch that increased the RSA key length for applet 1.7 didn't really count), so in order to make this work I'd have to spend a lot of time that I don't really have to understand those bits.

I do have a pretty good understanding of how the Belgian eid works, so if someone wants to undertake this work I'm happy to help with testing and acting in an advisory capacity. I can maybe even set up some CI system that can interact with a card (although I'd have to confirm), but I'm not going to have the time to implement this myself anymore.

frankmorgner:
Thanks for the update! We'll leave this open for now.

Speaking of CI testing, in OpenSC we are currently testing against a variatey of smart card emulators. If there is some service or infrastructure to test the Belgian ID card, we would be happy to test against that in our CI pipeline...

frankmorgner:

PR #2357, opened 2021-06-23
pkcs11-tool: allow setting CKA_EXTRACTABLE during keypair generation

a-dma:
Section 4.9 of the PKCS#11 v2.40 specification [1], mentions
CKA_EXTRACTABLE as a valid attribute for Private Key objects. However,
when calling `pkcs11-tool` with the `--exportable` option, the
attribute is not set as part of the private key template.

[1]: http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html

Tested using a YubiHSM.

##### Checklist


Jakuje (approved):

frankmorgner:
thanks

PR #2364, opened 2021-07-07
Avoid black/whitelist terms in documentation and comments

Jakuje:
As part of inclusive efforts, we try to eliminate some terms in the software we are using.

There are a lot of cases that we can not change, mostly PKCS#11 constants or terms as master file in several drivers or references to may git branches, but there are at least couple of changes that are not needed and can be quite safely changed as in this PR.

##### Checklist


frankmorgner (approved):

Jakuje:
rebased & pushed one more place where "whitelist" was used.

PR #2366, opened 2021-07-14
PIV Improved parsing of data from the card

dengert:
Based on Fuzz testing, many of the calls to sc_asn1_find_tag were replaced
with sc_asn1_read_tag. The input is also tested that the
expected tag is the first byte. Additional tests are also add.

sc_asn1_find_tag will skip 0X00 or 0Xff if found. NIST sp800-73-x specs
do not allow these extra bytes.

 On branch PIV-improved-parsing
 Changes to be committed:
	modified:   card-piv.c

##### Checklist

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2372,7 +2372,7 @@ static int piv_validate_general_authentication(sc_card_t *card,
 	r = sc_asn1_read_tag(&p2, r, &cla, &tag, &bodylen);
 	if (p2 == NULL || r < 0 || bodylen == 0 || (cla|tag) != 0x7C) {
 		LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_INVALID_DATA, "Can't find 0x7C");
-        }
+	}
 
 	r = sc_asn1_read_tag(&p2, bodylen, &cla, &tag, &taglen);
 	if (p2 == NULL || r < 0 || taglen == 0 || (cla|tag) != 0x82) {
@@ -2453,7 +2453,7 @@ piv_compute_signature(sc_card_t *card, const u8 * data, size_t datalen,
 				ptemp++;
 				templen--;
 			}
-			memcpy(out + nLen*i + nLen - templen , ptemp, templen);
+			memcpy(out + nLen*i + nLen - templen, ptemp, templen);
 			pint += intlen; /* next integer */
 			
 		}
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2449,8 +2449,6 @@ piv_compute_signature(sc_card_t *card, const u8 * data, size_t datalen,
 			if (intlen > nLen) { /* drop leading 00 if present */
 				if (*ptemp != 0x00) {
 					LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_INVALID_DATA,"Signature too long");
-					r = SC_ERROR_INVALID_DATA;
-					goto err;
 				}
 				ptemp++;
 				templen--;
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -608,14 +608,12 @@ static int piv_generate_key(sc_card_t *card,
 		const u8 *cp;
 		keydata->exponent = 0;
 
-		/* expected tag is 7f49.  */
-		/* we will whatever tag is present */
-
 		cp = rbuf;
 		in_len = r;
 
+		/* expected tag is 0x7f49,returned as cla_out == 0x60 and tag_out = 0x1F49 */
 		r = sc_asn1_read_tag(&cp, in_len, &cla_out, &tag_out, &in_len);
-		if (cp == NULL) {
+		if (cp == NULL || in_len == 0 || cla_out != 0x60 || tag_out != 0x1f49) {
 			r = SC_ERROR_ASN1_OBJECT_NOT_FOUND;
 		}
 		if (r != SC_SUCCESS) {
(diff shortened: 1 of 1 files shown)


src/libopensc/card-piv.c
```diff
@@ -608,14 +608,12 @@ static int piv_generate_key(sc_card_t *card,
...
-		/* expected tag is 7f49.  */
-		/* we will whatever tag is present */
-
 		cp = rbuf;
 		in_len = r;
 
+		/* expected tag is 0x7f49,returned as cla_out == 0x60 and tag_out = 0x1F49 */
 		r = sc_asn1_read_tag(&cp, in_len, &cla_out, &tag_out, &in_len);
-		if (cp == NULL) {
+		if (cp == NULL || in_len == 0 || cla_out != 0x60 || tag_out != 0x1f49) {
```
Jakuje, line 616: would it be more readable to add the check similarly as in the other cases as `(cla_out|tag_out) != 0x7f49`?

dengert: I suggest we add #2366  to 0.22.0-rc2  and (see below)  I will rework  #2053 to build on top of it so it has the needed fix for the piv_compute_signature.

It could be replaced by `(cla_out<<8 | tag_out) != 0x7f49`

This goes back to a note I sent to fuzz, but I do not see it in any comments. It said in part:

> In the asn1.c routines we have multiple ways to handle cla and tag.
See https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/asn1.h#L130

>Some routines return it as the low order byte in an integer. Some return the cla_out as a 4 byte integer i.e shift cla<<24.
and all return the rest of the tag as the remaining bytes without the cla.
So although I am looking for 0x7f49 sc_asn1_read_tag returns cla_out=0x60 and tag_out=1f49.

>will add comment like"
 /* 0x7F39 retuened as cla_out == 0x60 and tag_out = 0x1F39 */
cla_out<<8 == 0x6000
0x6000|0x1f49 == 0x7f49

>sc_asn1_find_tag does this:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/asn1.c#L498-L505

>sc_asn1_skip_tag does this:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/asn1.c#L529-L554
(Should use the same code as sc_asn1_find_tag.)

>Also note a single byte tag: `(tag | cla) == somevalue` will work such as in tools/goid-tool.c or other parts of card-piv.c


I also forgot that in #2053 (That has been sitting around for quite some time) I rewrote the buffering and handling of IO from the card and addressed the short vs long `cla` issue in at least the PIV driver. And addressed tag 0x7f49 issue:

https://github.com/OpenSC/OpenSC/pull/2053/files#diff-14d802e467876e00ece49c8d6621ec74eaf7d02384642ac0694de83397d9bbfc

in card-piv.c lines 2620 - 2623

```
r = sc_asn1_read_tag(&cp, in_len, &cla_out, &tag_out, &in_len);
		if (r < 0 || cp == NULL || !piv_is_expected_tag(cla_out, tag_out, SC_ASN1_CTX | SC_ASN1_CONS | 0x1F49)) {
			r = SC_ERROR_ASN1_OBJECT_NOT_FOUND;
		}
```
which added a piv_is_expected_tag using the long tag formats. This is not much clearer then
then `(cla_out<<8|tag_out) != 0x7f49`

#2053 made substantial changes to PIV from the card but needs more work to address the issues fixed by this PR. The two PRs do not rebase well.

Jakuje: Oh, right. Multibyte tags are confusing. Comment should be enough for now I think. 

I am fine with adding this to 0.22.0.

src/libopensc/card-piv.c
```diff
@@ -2424,32 +2430,32 @@ piv_compute_signature(sc_card_t *card, const u8 * data, size_t datalen,
...
+				if (*ptemp != 0x00) {
+					LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_INVALID_DATA,"Signature too long");
 				}
-			} else  {
-				r = SC_ERROR_INVALID_DATA;
-				goto err;
+				ptemp++;
+				templen--;
 			}
+			memcpy(out + nLen*i + nLen - templen , ptemp, templen);
```
Jakuje, line 2456: nit: could also remove the whitespace before the comma.

Jakuje (approved):
just minor nits, otherwise I think this is ready for 0.22.0

src/libopensc/card-piv.c
```diff
@@ -2367,20 +2365,28 @@ static int piv_validate_general_authentication(sc_card_t *card,
...
-				r = taglen;
-			} else
-				r = SC_ERROR_INVALID_DATA;
-		} else
-			r = SC_ERROR_INVALID_DATA;
+	p2 = rbuf;
+	r = sc_asn1_read_tag(&p2, r, &cla, &tag, &bodylen);
+	if (p2 == NULL || r < 0 || bodylen == 0 || (cla|tag) != 0x7C) {
+		LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_INVALID_DATA, "Can't find 0x7C");
+        }
```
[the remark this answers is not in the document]
dengert: Fixed.

src/libopensc/card-piv.c
dengert: Fixed

Jakuje:
Looks good. I see you already pulled the github actions, which looks like working quite good already (aside of the openpgp). If there will be no more comments, I will merge it tomorrow.

dengert:
I did not pull the the github actions, my upstream master is at d90048e5bb8ee85afab048020fd879be9c68aebc. The statement: "This branch has no conflicts with the base branch"  would imply the PR is tested against the github base master which would have your  github actions changes.

PR #2374, opened 2021-08-03
Remove old card drivers for MioCOS and JCOP

xhanulik:
This PR removes old card drivers, which were disabled in #1175.

There was no user and developer activity since 2003
* `card-miocos.c` 30c0943
* `card-jcop.c` f761d15

except some general fixes on both drivers.


frankmorgner (approved):

Jakuje:
I know that https://github.com/OpenSC/OpenSC/projects/2 talks about disabling compilation of the old drivers, but I do not see a point in keeping some unused code in the repository. If somebody will need these drivers for something, they can be pulled from git history.

frankmorgner:
> I know that https://github.com/OpenSC/OpenSC/projects/2 talks about disabling compilation of the old drivers, but I do not see a point in keeping some unused code in the repository. If somebody will need these drivers for something, they can be pulled from git history.

We want to have the old drivers still available for at least one release cycle where they are disabled (not removed). If they are needed, they can be easily enabled again by the users and they will hopefully yield an issue here. For MioCOS and JCOP, there were no complaints, so they're fine to be removed.

PR #2375, opened 2021-08-04
pkcs11-spy: fix random crash in print_generic()

vesajaaskelainen:
Depending on stack state print_generic() could cause crash or spurious garbage
in logs.

Example crash:

*** buffer overflow detected ***: pkcs11test terminated

Signed-off-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11-display.c
+++ b/src/pkcs11/pkcs11-display.c
@@ -141,7 +141,8 @@ print_generic(FILE *f, CK_LONG type, CK_VOID_PTR value, CK_ULONG size, CK_VOID_P
 	CK_ULONG i;
 
 	if((CK_LONG)size != -1 && value != NULL) {
-		char hex[16*3+1], ascii[16+1];
+		char hex[16*3+1] = {0};
+		char ascii[16+1];
 		char *hex_ptr = hex, *ascii_ptr = ascii;
 		int offset = 0;
 


Jakuje:
Good catch. Would not it be easier just to terminate the `hex` buffer with NULL byte, rather than zeroize the whole buffer with each cycle?

vesajaaskelainen:
> Good catch. Would not it be easier just to terminate the `hex` buffer with NULL byte, rather than zeroize the whole buffer with each cycle?

Easier is subjective :)

Have been customary to initialize buffers with zero.

Can change both to `hex[0] = 0;` if you want?

vesajaaskelainen:
> Well, zeroizing buffer with every block is ineffective (even though pkcs11-spy is just a debug tool where we do not strive for too much efficiency). We are "broken" only on the last byte, which is actually never overridden after the initialization and shorter inputs are always padded to the same size so just initialization of the buffer to zero should solve the issue if I am right:
> 
> ```
> char hex[16*3+1] = {0};
> ```

Should work. Works in desktop will still run test on device where it was originally noticed.

vesajaaskelainen:
Updated the PR.

Jakuje (approved):

PR #2368, opened 2021-07-17
pkcs11-spy: Fix crash for pMechanism=NULL

vesajaaskelainen:
Some test suites are excercising pMechanism==NULL case and this causes
crash when using pkcs11-spy for logging in between.

Centralize logging for pMechanism and handle NULL case.

Signed-off-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>

##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [ ] PKCS#11 module is tested
- [ ] Windows minidriver is tested
- [ ] macOS tokend is tested

PR #2377, opened 2021-08-04
pkcs11-spy: Fix more crashes with NULL arguments

vesajaaskelainen:
This PR fixes last crashes noticed with [Google's pkcs11test](https://github.com/google/pkcs11test).

Now it works with SoftHSMv2 (v2.6.1) and with OP-TEE PKCS#11 (devel) without crashing.

Following PRs are also needed:
- https://github.com/OpenSC/OpenSC/pull/2368
- https://github.com/OpenSC/OpenSC/pull/2375

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/pkcs11-display.c
+++ b/src/pkcs11/pkcs11-display.c
@@ -1035,6 +1035,9 @@ print_attribute_list(FILE *f, CK_ATTRIBUTE_PTR pTemplate, CK_ULONG  ulCount)
 	CK_ULONG j, k;
 	int found;
 
+	if (!pTemplate)
+		return;
+
 	for(j = 0; j < ulCount ; j++) {
 		found = 0;
 		for(k = 0; k < ck_attribute_num; k++) {
@@ -1065,6 +1068,9 @@ print_attribute_list_req(FILE *f, CK_ATTRIBUTE_PTR pTemplate, CK_ULONG  ulCount)
 {
 	CK_ULONG j, k;
 	int found;
+
+	if (!pTemplate)
+		return;
 
 	for(j = 0; j < ulCount ; j++) {
 		found = 0;
--- a/src/pkcs11/pkcs11-spy.c
+++ b/src/pkcs11/pkcs11-spy.c
@@ -416,6 +416,8 @@ CK_RV C_GetFunctionList
 	}
 
 	enter("C_GetFunctionList");
+	if (ppFunctionList == NULL)
+		return retne(CKR_ARGUMENTS_BAD);
 	*ppFunctionList = pkcs11_spy;
 	return retne(CKR_OK);
 }
--- a/src/pkcs11/pkcs11-spy.c
+++ b/src/pkcs11/pkcs11-spy.c
@@ -611,7 +611,10 @@ C_OpenSession(CK_SLOT_ID  slotID, CK_FLAGS  flags, CK_VOID_PTR  pApplication,
 	fprintf(spy_output, "[in] pApplication = %p\n", pApplication);
 	fprintf(spy_output, "[in] Notify = %p\n", (void *)Notify);
 	rv = po->C_OpenSession(slotID, flags, pApplication, Notify, phSession);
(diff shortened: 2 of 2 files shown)


Jakuje (approved):

vesajaaskelainen:
Rebased to master -- still seems that isoapplet tests are failing thou -- I don't think they are related to changes in here (or other PRs that I have opened).

Seems like python 2 is being used instead of python 3:
```
Byte-compiling python modules...
  File "/usr/local/lib/python2.7/dist-packages/virtualsmartcard/cards/RelayMiddleman.py", line 8
    def handleInPDU(self, inPDU: bytes):
                               ^
SyntaxError: invalid syntax
```

Can't comment on actual java tests.

Jakuje:
> Rebased to master -- still seems that isoapplet tests are failing thou -- I don't think they are related to changes in here (or other PRs that I have opened).
> 
> Seems like python 2 is being used instead of python 3:
> 
> ```
> Byte-compiling python modules...
>   File "/usr/local/lib/python2.7/dist-packages/virtualsmartcard/cards/RelayMiddleman.py", line 8
>     def handleInPDU(self, inPDU: bytes):
>                                ^
> SyntaxError: invalid syntax
> ```
> 
> Can't comment on actual java tests.

Interestingly enough, this is not the failure that breaks the test (and I missed it previously). In any case, this part of vsmartcard is probably not used. We should use some newer image, but I think I had issues making the rest of the tools working with newer Java on the other hand.

The final failure is the following, but it does not look like related to your changes:
```
Error:  Errors: 
Error:    com.licel.jcardsim.crypto.KeyAgreementImplTest#testGenerateSecretDH ArrayIndexOutOfBoundsException
```
the jcardsim tests are unfortunately still flaky and for me as a non-Java developer, they are hard to comprehend and fix.

Jakuje:
Thank you for your contributions! They will be in upcoming release. One more thing, could you write some reproduceres for the crashes in automated tests? We already have bunch of these in `tests/` for other use cases, but nothing for pkcs11-spy. Having them automated, will help us to prevent breaking the use case in the future.

vesajaaskelainen:
> Thank you for your contributions! They will be in upcoming release. One more thing, could you write some reproduceres for the crashes in automated tests? We already have bunch of these in `tests/` for other use cases, but nothing for pkcs11-spy. Having them automated, will help us to prevent breaking the use case in the future.

Let me see what I can do.

vesajaaskelainen:
> > Thank you for your contributions! They will be in upcoming release. One more thing, could you write some reproduceres for the crashes in automated tests? We already have bunch of these in `tests/` for other use cases, but nothing for pkcs11-spy. Having them automated, will help us to prevent breaking the use case in the future.
> 
> Let me see what I can do.

The base code is kinda missing in here. There is no good entry point where to add "source level" tests.

I suppose one option would be to have PKCS#11 test suite that gets executed -- nothing one solution is better that others seems to be there.

Existing choices that could could be easy to extended I suppose would be based tests suite on [python-pkcs11](https://github.com/danni/python-pkcs11) -- requires python 3 usage for tests -- also has own pytest suite, Google's [pkcs11tests](https://github.com/google/pkcs11test) is one option -- that is in C++ so would need to be compiled and extending would probably mean contributing back in that project.

Don't know how well those work on other than Linux.

Jakuje:
At this moment, we have unit tests in [`src/tests/unittests`](https://github.com/OpenSC/OpenSC/tree/master/src/tests/unittests). For pkcs11 functional testing (but it does not test many corner cases) we use [`src/tests/p11test`](https://github.com/OpenSC/OpenSC/tree/master/src/tests/p11test), but neither of that checks the pkcs11-spy. but it should be easy to add it.
Running either of these projects as part of the CI would be also a possibility. We have couple of emulators in CI, but for testing pkcs11-spy, it should be fine to run against softhsm as you initially did.

If it would not work on anything but Linux, it is fine. Most of our testing already happens on Linux anyway.

PR #2381, opened 2021-08-10
Try to document release process + improve PIV testing

Jakuje:


tests/test-pkcs11-tool-test-threads.sh
```diff
@@ -14,16 +14,16 @@ fi
...
-$PKCS11_TOOL --test-threads IN -L
+$PKCS11_TOOL --test-threads IN -L --module="$P11LIB"
 assert $? "Failed running tests"
 
 
 echo "======================================================="
 echo "Test pkcs11 threads OSILGISLT0 "
 echo "======================================================="
-$PKCS11_TOOL --test-threads OSILGISLT0 -L
+$PKCS11_TOOL --test-threads OSILGISLT0 -L --module="$P11LIB"
```
Jakuje, line 26: @dengert The `pkcs11-tool --test-threads` here warns about unknown operation `OS`. Can you check what you had in your mind when writing this test?

dengert: It looks like OS was the old way of initialize with OS locking. But the code was change to use:
- IN call C_Initialize with NULL 
- IL  call C_Initialize with CKF_OS_LOCKING_OK

The OS was left in the test script input parameter and comment . It should look like this: 
```
./pkcs11-tool --test-threads ILGISLT0 -L
Test thread 0 started with options:ILGISLT0
Test thread 0 C_Initialize CKF_OS_LOCKING_OK 
Main C_Initialize(NULL) rv:CKR_OK
Test thread 0 C_Initialize  returned CKR_CRYPTOKI_ALREADY_INITIALIZED
Test thread 0 C_GetInfo
Test thread 0 C_GetInfo returned CKR_OK
Test thread 0 C_GetSlotList to get l_p11_num_slots
Available slots:
No slots.
test_threads cleanup starting
test_threads threads stills active:1
test_threads thread:0 state:2, rv:CKR_OK

test_threads waiting for 30 seconds ...
Test thread 0 C_GetSlotList returned CKR_OK
Test thread 0 l_p11_num_slots:0
Test thread 0 C_GetTokenInfo from slot_index 0 using show_token
```

But without a reader and card, there was a segfault that I will correct.
Without a slot, the test code called `show_token(p11_slots[(*(pctest + 1) - '0')]);` line 7512 should be an error message.

I must not have tested without a reader.

Jakuje: Thanks. Let me know if you would like to submit it as a separate PR there will be some commit I should pull.

dengert: Submitted #2388 to fix both.

frankmorgner (commented):
I'm fine about the source code changes, but I have some general comments

1. We have already some information available in the wiki, which is sometimes more detailed than CONTRIBUTING.md...

- https://github.com/OpenSC/OpenSC/wiki/OpenSC-Release-Howto
- https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Release-Testing

This information needs to be synchronized. One thing that's missing from the current release is to use annotated tags instead of lightweight tags for releases (unfortunately the latter is the default when creating a release via the webpage on Github).

Also, it would be better to have this information in one single place, instead of having to synchronize the changes all the time.

2. I would expect the content of CONTRIBUTING.md to be more generic, potentially linking to subchapters, e.g. about testing with podman/docker. I doubt that everyone contributing here is using containers. A quick google search spit out https://mozillascience.github.io/working-open-workshop/contributing/ for me, which looks promising...

If I'm not mistaken, adding `containers/README.md` would make Github render that file when browsing this source code directory. This file could, of course be, be linked from a more generic CONTRIBUTING.md. 

3. Would it make sense to move the container directory into tests directory?

Jakuje:
Right. It does not make sense to have the stuff in two places. But it needs to be linked from one place. In wiki, it is very hard to find the stuff. Having all usful stuff linked from `CONTRIBUTING.md` would be useful.

The `CONTRIBUTING.md` was also not meant to be complete, but when I started putting the notes together, there was no better place. I will try to update the wiki and keep this just linking it. The issue with this might be that we do not have single history as with git. Yeah, the wiki has some history, but it is strictly tied to github and not backed up with the code.

Using `containers/README.md` sounds also like a good idea.

Jakuje:
I cleaned up the branch a bit and also hopefully addressed the problem of random jcardsim tests failures.

I think it would be even better to keep the jar file in some cache instead of building it 4 times in CI. But lets keep this for the future.

CONTRIBUTING.md
```diff
@@ -0,0 +1,10 @@
+# Testing locally
+
+To learn how to run the tests from Github actions locally in containers, see
+[`containers`](https://github.com/OpenSC/OpenSC/master/containers/).
```
frankmorgner, line 4: Github [supports relative links](https://github.blog/2013-01-31-relative-links-in-markup-files/); should look a bit nicer.

.github/setup-java.sh
```diff
@@ -13,12 +13,16 @@ env | grep -i openjdk
...
+if [ ! -d "oracle_javacard_sdks" ]; then
+	git clone https://github.com/martinpaljak/oracle_javacard_sdks.git
+fi
 export JC_HOME=$PWD/oracle_javacard_sdks/jc222_kit
 export JC_CLASSIC_HOME=$PWD/oracle_javacard_sdks/jc305u3_kit
 
 # jCardSim
-git clone https://github.com/arekinath/jcardsim.git
+if [ ! -d "jcardsim" ]; then
+	git clone https://github.com/Jakuje/jcardsim.git
```
frankmorgner, line 24: The bug you've found should be reported back.

Jakuje: It is reported to the parent project: https://github.com/licel/jcardsim/pull/174 -- I hope if it will be accepted, it can be pulled by @arekinath later. But I can report it also there if there will be no follow-up for some time.

CONTRIBUTING.md
Jakuje: ok. Changed to relative.

PR #2388, opened 2021-08-17
Fix pkcs11-tool and test-pkcs11-tool-test-threads.sh

dengert:
test-pkcs11-tool-test-threads.sh had the wrong parameters
for one of the tests. And extra "OS" was left in the parameter.

pkcs11-tool when testing threads could segfault if no reader present.
Error checking for slot index improved.

Fixes:#2381

 On branch pkcs11-tool-and-test
 Changes to be committed:
	modified:   pkcs11-tool.c
	modified:   ../../tests/test-pkcs11-tool-test-threads.sh

##### Checklist


Jakuje (approved):
looks good. The failing tests are probably caused by yet another change in the ubuntu images. I will try to investigate it later.

dengert:
See #2395 looks like failing test caused by change to vsmartcard on 8/13 and https://github.com/dengert/OpenSC/actions/runs/1153542375

PR #2394, opened 2021-08-20
prepare package signing for macOS

frankmorgner:


frankmorgner:
Hmm, how are the actions activated for this PR?

Jakuje:
Actions on PR are activated only when any c or h code is changed at this moment. Feel free this part to test your changes

https://github.com/OpenSC/OpenSC/blob/088a8a1302f8b6ff3224bf12526f14a8722afd8f/.github/workflows/macos.yml#L5-L7

dengert:
how about adding  "- '.github/workflow/*'" to   .github/workflow/macros.yml  line 8 so it looks like:
``
on:
  pull_request:
    paths:
      - '**.c'
      - '**.h'
      - '.github/workflow/*'    
```  
This would thee run the actions anytime an action was changed.

**Comment by Jakuje** [comment]
That should do it too.

**Comment by frankmorgner** [comment]
I've added the suggested changes.

What's strange is that the (raw) log in [macOS](https://github.com/OpenSC/OpenSC/runs/3394734980) says:
```
2021-08-22T19:24:59.2925010Z Cloning into 'OpenSCToken'...
```
followed by
```
2021-08-22T19:25:12.7786860Z ##[group]Run .github/build.sh
```

It looks like the `git clone` is not successful or timed out, but I can't find any clue in the log...

**Comment by dengert** [comment]
You could add a `set -x` in the setup-macos.sh to get a better log. 
then add a ls -l OpenSCToken

Note" `if [ "$GITHUB_EVENT_NAME" == "pull_request" -a -n "$encrypted_3b9f0b9d36d1_key" ]; then`  will give different results when you push to your repro vs when you submit a PR. 

https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/macos.yml#L14-L16 says:

**Comment by Jakuje** [comment]
Please, apply the following change to unbreak the push-artifacts:
```
diff --git a/.github/workflows/macos.yml b/.github/workflows/macos.yml
index 77725d01..cadc02ac 100644
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -20,6 +20,7 @@ jobs:
           name: opensc-build-macos
           path:
             OpenSC*.dmg
+      - run: .github/remove_signing_key.sh; rm -f .github/secrets.tar
 
   push-artifacts:
     runs-on: macos-latest
@@ -34,6 +35,3 @@ jobs:
       - run: git config --global user.name "Github Actions";
       - run: .github/push_artifacts.sh "Github Actions ${GITHUB_REF}"
         if: ${{ github.event_name != 'pull_request' && github.repository == 'OpenSC/OpenSC' }}
-      - run: .github/remove_signing_key.sh; rm -f .github/secrets.tar
```
The secret is only in the first job, not in the second one. That was my bad when I migrated it from travis.

Jakuje:
If the clonning would not be successful, it would probably fail, but the script is missing the `-e` to make sure we fail early enough. As propsoed by Doug, adding the `-x` to see what is executed would also make sense for debugging.

Jakuje (approved):

Jakuje:
If this is final, it looks good to me. Not sure why there is a conflict. Can you try to rebase on current master?

frankmorgner:
done. populating some signing credentials to the build step was missing. now everything seems to be fine

Issue #2390, opened 2021-08-18
github actions problems with test scripts starting /usr/bin/pcscd -f 


dengert:
Some of the github actions test scripts running in Ubuntu fail with 
`00000000 [140024449624000] utils.c:81:GetDaemonPid() Can't open /run/pcscd/pcscd.pid: No such file or directory`

The problems appears in the [test-piv.sh lines 26-29](https://github.com/OpenSC/OpenSC/blob/master/.github/test-piv.sh#L26-L29)
./test-isoapplet.sh, ./test-gidsapplet.sh and ./test-openpgp.sh all do the same.

This first problem is `PCSCD_PID=$!` returns the PID of the sudo command, not the /usr/sbin/pcscd -f
Because /usr/sbin/pcscd  is started by sudo, not be the shell, and sudo still running when     https://github.com/OpenSC/OpenSC/blob/master/.github/test-piv.sh#L45 kills the sudo, but not the /usr/sbin/pcscd keeps running. So after a few tests, there are multiple  "/usr/sbin/pcscd -f" running.
Since "bash -e" is used, line 45 may never be executed if there is a failure before. 

The second problem is a timing issue because /usr/sbin/pcscd takes some time to get started before it sets `/run/pcscd/pcscd.pid` so any following command may fail.

On my Ubuntu-20.04, (/var/run is symlink to /run)  and systemctl shows after stopping and starting to clean up from some testing:
```
doug@XUbuntu-20:/tmp$ sudo systemctl stop pcscd.service pcscd.socket
doug@XUbuntu-20:/tmp$ sudo systemctl start pcscd.service pcscd.socket
doug@XUbuntu-20:/tmp$ sudo systemctl status pcscd
● pcscd.service - PC/SC Smart Card Daemon
     Loaded: loaded (/lib/systemd/system/pcscd.service; indirect; vendor preset: enabled)
     Active: active (running) since Wed 2021-08-18 10:04:19 CDT; 16s ago
TriggeredBy: ● pcscd.socket
       Docs: man:pcscd(8)
   Main PID: 6646 (pcscd)
      Tasks: 3 (limit: 9485)
     Memory: 692.0K
     CGroup: /system.slice/pcscd.service
             └─6646 /usr/sbin/pcscd --foreground --auto-exit

Aug 18 10:04:19 XUbuntu-20.04 systemd[1]: Started PC/SC Smart Card Daemon.
[...]

PR #2395, opened 2021-08-21
test-actions-2 vsmartcard-0.8 

dengert:
[vsmartcard](https://github.com/frankmorgner/vsmartcard) was updated on 8/13 a few days before we start noticing Github actions failing for several tests.

This temporary change  checkout out vsmartcard-0.8  avoids this commit: https://github.com/frankmorgner/vsmartcard/commit/5a6b330b154a6b6b3ad30c4b69fae83a031d0f36 

All the action tests now run.

### DIFF
--- a/.github/setup-vsmartcard.sh
+++ b/.github/setup-vsmartcard.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
-
+set -x
 if [ ! -d "vsmartcard" ]; then
 	git clone https://github.com/frankmorgner/vsmartcard.git
 fi
 pushd vsmartcard/virtualsmartcard
+git checkout -b  tag-0.8 virtualsmartcard-0.8
 autoreconf -vis && ./configure && make -j2 && sudo make install
 popd


Jakuje:
Good catch. I was indeed too tired from looking into it for too long to notice this change. But as I mentioned in the https://github.com/OpenSC/OpenSC/issues/2390#issuecomment-902901549, this would be the next place I would look into. You can see the working github actions in your fork:

https://github.com/dengert/OpenSC/actions/runs/1153542375

I think we can use for now this one as a temporary workaround, but the change in https://github.com/frankmorgner/vsmartcard/pull/202 should be reviewed as it is obviously causing issues not only to us, see https://github.com/frankmorgner/vsmartcard/issues/204

I will have a better look probably on Monday.

Jakuje (approved):

Jakuje:
I will merge this one at least to have something working for now. Once the https://github.com/frankmorgner/vsmartcard/pull/206 will be merged (and we will have some assurance that this will not repeat that frequently such as CI), we can back to the testing of master.

PR #2401, opened 2021-08-30
ci: no need to focus on non-PRs

frankmorgner:
secrets are only available for collaborators, so just fire everything up

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -46,4 +46,3 @@ jobs:
       - run: git config --global user.email "builds@github.com"
       - run: git config --global user.name "Github Actions";
       - run: .github/push_artifacts.sh "Github Actions ${GITHUB_REF}"
-        if: ${{ secrets.GH_TOKEN != '' }}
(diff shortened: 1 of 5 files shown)


.github/workflows/macos.yml
```diff
@@ -46,4 +46,3 @@ jobs:
       - run: git config --global user.email "builds@github.com"
       - run: git config --global user.name "Github Actions";
       - run: .github/push_artifacts.sh "Github Actions ${GITHUB_REF}"
-        if: ${{ github.event_name != 'pull_request' && github.repository == 'OpenSC/OpenSC' }}
```
Jakuje, line 49: I would probably keep at least the second part `github.repository == 'OpenSC/OpenSC'` to avoid noise. The `GH_TOKEN` can be set also on forks and then it would just lead to unnecessary noise when these forks will try to push to nightly repository and fail.

Jakuje (approved):

PR #2405, opened 2021-09-11
openpgp-tool: fix buffer overflow on serials with MSb set

vpelletier:
Fixes the following crash:
  $ openpgp-tool --card-info
  Using reader with a card: Linux Foundation Multifunction Composite Gadget - vincent [python-usb-f-ccid] 00 00
  AID:             d2:76:00:01:24:01:03:41:ff:65:a4:9b:68:64:00:00
  Version:         3.41
  Manufacturer:    unmanaged S/N range
  *** buffer overflow detected ***: terminated
  Abandon (core dumped)
"a4:9b:68:64" from the AID being the serial, and:
  (gdb) print (data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3])
  $12 = -1533319068
but
  (gdb) print (unsigned long) (data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3])
  $13 = 18446744072176232548

Avoid the shifts and cast altogether.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -25,7 +25,8 @@ EXTRA_DIST = Makefile.mak versioninfo-tools.rc.in versioninfo-opensc-notify.rc.i
 
 noinst_HEADERS = util.h fread_to_eof.h \
 	egk-tool-cmdline.h goid-tool-cmdline.h npa-tool-cmdline.h \
-	opensc-asn1-cmdline.h opensc-notify-cmdline.h pkcs11-register-cmdline.h
+	opensc-asn1-cmdline.h opensc-notify-cmdline.h pkcs11-register-cmdline.h \
+	openpgp-tool-helpers.h
 noinst_PROGRAMS = sceac-example
 bin_PROGRAMS = opensc-tool opensc-explorer opensc-notify \
 	pkcs15-tool pkcs15-crypt pkcs11-tool pkcs11-register \
@@ -77,7 +78,7 @@ netkey_tool_SOURCES = netkey-tool.c
 netkey_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 westcos_tool_SOURCES = westcos-tool.c util.c
 westcos_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
-openpgp_tool_SOURCES = openpgp-tool.c util.c
+openpgp_tool_SOURCES = openpgp-tool.c util.c openpgp-tool-helpers.c
 openpgp_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 iasecc_tool_SOURCES = iasecc-tool.c util.c
 iasecc_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
--- a/src/tools/openpgp-tool.c
+++ b/src/tools/openpgp-tool.c
@@ -49,6 +49,8 @@
 #include "libopensc/log.h"
 #include "libopensc/card-openpgp.h"
 
+#include "openpgp-tool-helpers.h"
+
 #define OPT_RAW     256
 #define OPT_PRETTY  257
 #define OPT_VERIFY  258
@@ -74,15 +76,6 @@ struct ef_name_map {
 
 /* declare functions */
 static void show_version(void);
-static char *prettify_hex(u8 *data, size_t length, char *buffer, size_t buflen);
-static char *prettify_algorithm(u8 *data, size_t length);
-static char *prettify_date(u8 *data, size_t length);
-static char *prettify_version(u8 *data, size_t length);
-static char *prettify_manufacturer(u8 *data, size_t length);
-static char *prettify_serialnumber(u8 *data, size_t length);
-static char *prettify_name(u8 *data, size_t length);
-static char *prettify_language(u8 *data, size_t length);
-static char *prettify_gender(u8 *data, size_t length);
 static void display_data(const struct ef_name_map *mapping, u8 *data, size_t length);
 static int decode_options(int argc, char **argv);
 static int do_info(sc_card_t *card, const struct ef_name_map *map);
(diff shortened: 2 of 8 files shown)


Jakuje (approved):

Jakuje:
One more thing I am trying to evangelize here is a bit more testing, given the amount of legacy code in here.

This is quite self-contained function (even though static), which could be quite easily tested with basic unit test, would you be interested in looking into writing some basic unit tests under `src/tests/unittests/`, testing at least the corner cases of this function?

It would be awesome if you could test also some other functions of the openpgp-tool to check if there are not similar issues around, but that would be non-mandatory.

vpelletier:
Are you sure you want to ask the pythonist who is not sure why a C cast is behaving in some way to write a test (especially starting a test file from scratch) ? :)

Joke aside, here is how I think I will go, so you can correct me early:
- de-static-ify and move all `prettify_`* functions to a header (`src/tools/openpgp-tool.h` ?) included both in the test and in `src/tools/openpgp-tool.c`
- then exercise them all with a list of pairs of input args + expected output each defined in the test source

Jakuje:
> Are you sure you want to ask the pythonist who is not sure why a C cast is behaving in some way to write a test (especially starting a test file from scratch) ? :)

Only if you are interested in learning something new in C, autotools and friends :) The change is totally fine as it is.

> Joke aside, here is how I think I will go, so you can correct me early:
> 
>     * de-static-ify and move all `prettify_`* functions to a header (`src/tools/openpgp-tool.h` ?) included both in the test and in `src/tools/openpgp-tool.c`

It would require some creative linking anyway.

We don't have any test for tools yet. I though about directly including the `tools/openpgp-tool.c` into the the test file and `#ifdef`-ing the unused functions in the tool to be able to build the test (which is not nice but works).

It would be nicer if these functions would be in separate file, such as `openpgp-tool-prettify.c` with separate header file, but we don't have yet anything like this in other tools except for the generated `-cmdline` suffixes. And then this c file could be directly included in the test again or do the proper linking, but its ugly in autotools between different directories.

>     * then exercise them all with a list of pairs of input args + expected output each defined in the test source

Right. That's same as in python, isn't it? ;)

vpelletier:
I pushed my first attempt (just checking `prettify_serialnumber` for now) in [a branch](/vpelletier/OpenSC/tree/openpgp_tool_tests) based on the one for this merge request, so that I don't block its merging (and don't consume auto-checker resources every time I amend).

I intend to also constify the `prettify_*(u8 data` argument, which is trivial in all but `language` and `name` cases, and should be not too hard there either.

Jakuje:
Thank you! I only do not like having the function definitions in header files. The headers should really contain only function declarations (unless they are inline). Therefore I would suggest to split the `openpgp.h` by moving the definitions in file like `openpgp-tool-helpers.c`, declarations into `openpgp-tool-helpers.h`

vpelletier:
I just pushed to my branch a few more changes, but I only now notice your latest message.

> Therefore I would suggest to split the openpgp.h by moving the definitions in file like openpgp-tool-helpers.c, declarations into openpgp-tool-helpers.h

Doing, I will force-push shortly.

vpelletier:
> Doing, I will force-push shortly.

Done, with a fix for const-ification.

```
OpenSC/src/tests/unittests$ make && ./openpgp-tool
make: rien à faire pour « all ».
[==========] Running 8 test(s).
[ RUN      ] torture_prettify_algorithm
[       OK ] torture_prettify_algorithm
[ RUN      ] torture_prettify_date
[       OK ] torture_prettify_date
[ RUN      ] torture_prettify_version
[       OK ] torture_prettify_version
[ RUN      ] torture_prettify_manufacturer
[       OK ] torture_prettify_manufacturer
[ RUN      ] torture_prettify_serialnumber
[       OK ] torture_prettify_serialnumber
[ RUN      ] torture_prettify_name
[       OK ] torture_prettify_name
[ RUN      ] torture_prettify_language
[       OK ] torture_prettify_language
[ RUN      ] torture_prettify_gender
[       OK ] torture_prettify_gender
[==========] 8 test(s) run.
[  PASSED  ] 8 test(s).
```

Jakuje:
Looks great! Thanks!

One thing (which is again not completely related to the changes, but was surfaced by me having a better look into whats going on there) on a side note, there is a new algorithm `SC_OPENPGP_KEYALGO_EDDSA` (0x16), which is handled by the driver, but not by the prettify function so once you touch that, it would be great if you could add it:

https://github.com/OpenSC/OpenSC/blob/a00ead5bcaeef9761287572f7b62065d62c211d9/src/libopensc/cardctl.h#L937

I think you will also have to update the `src/tools/Makefile.mak` for windows builds (the tests are not running in windows).

And last nit might be the rewrite of the `prettify_language` function in b1f0d8077eab0396ff38e89f93082469e6ff3eb6 which does not mention it in the commit message.

vpelletier:
> there is a new algorithm `SC_OPENPGP_KEYALGO_EDDSA` (0x16)

Added. I followed the existing code and put the literal directly in the code, rather than using the macro. Let me know if I should swap all the literals for their corresponding macros.

> I think you will also have to update the `src/tools/Makefile.mak` for windows builds (the tests are not running in windows).

Done. I ran a cygwin build, which I believe exercised this part of the makefile.

> And last nit might be the rewrite of the `prettify_language` function in b1f0d80 which does not mention it in the commit message.

Ah thanks, this reminded me that I wanted to check about my change collapsing all contiguous `<` chars in `prettify_name`. According to wikipedia, there should be no more than two consecutive `<`, so I reverted that part and dropped the corresponding unittest entry.

I also documented the changes to `prettify_language` (and `prettify_name`) in the commit message.

I pushed to this merge request's branch, as I believe there should be only few further changes to do to these commits, to get them all tested.

vpelletier:
I angered the windows build:

```
openpgp-tool-helpers.c
	cl /GS /W3 /WX /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_WARNINGS /MT /nologo /DHAVE_CONFIG_H /I..\..\win32 /I..\..\src /IC:\openpace-Win32\src /IC:\OpenSSL-v111-Win32\include /DOPENSSL_SECURE_MALLOC_SIZE=65536 /IC:\zlib-Win32   "/IC:\Program Files (x86)\Windows Kits\10\Cryptographic Provider Development Kit\Include" "/IC:\Program Files (x86)\Microsoft CNG Development Kit\Include" "/IC:\Program Files (x86)\WiX Toolset v3.11\\SDK\VS2015\inc" /DWINVER=0x0601 /D_WIN32_WINNT=0x0601 /DWIN32_LEAN_AND_MEAN /DENABLE_OPENPACE /DENABLE_OPENSSL /DENABLE_ZLIB /DENABLE_MINIDRIVER /DENABLE_SM  /DOPENSC_FEATURES="\"pcsc openssl zlib\"" /Zi /c openpgp-tool.c
openpgp-tool.c
	link /NOLOGO /INCREMENTAL:NO /MACHINE:x86 /NODEFAULTLIB:MSVCRTD  /NODEFAULTLIB:MSVCRT /NXCOMPAT /DYNAMICBASE /NODEFAULTLIB:LIBCMTD /DEBUG /OPT:REF /OPT:ICF /pdb:openpgp-tool.pdb /out:openpgp-tool.exe openpgp-tool.obj openpgp-tool-helpers.obj ..\..\src\common\common.lib  ..\..\src\scconf\scconf.lib  ..\..\src\libopensc\opensc.lib  ..\..\src\pkcs15init\pkcs15init.lib  ..\..\src\common\libpkcs11.lib  ..\..\src\common\libscdl.lib gdi32.lib shell32.lib User32.lib ws2_32.lib shlwapi.lib
openpgp-tool-helpers.obj : error LNK2005: _prettify_hex already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_algorithm already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_date already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_version already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_manufacturer already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_serialnumber already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_name already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_language already defined in openpgp-tool.obj
openpgp-tool-helpers.obj : error LNK2005: _prettify_gender already defined in openpgp-tool.obj
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_hex_dump_asc referenced in function _do_dump_do
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_print_usage_and_die referenced in function _decode_options
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_error referenced in function _decode_options
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_fatal referenced in function _main
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_connect_card referenced in function _main
openpgp-tool.obj : error LNK2019: unresolved external symbol _util_get_pin referenced in function _decode_options
openpgp-tool.exe : fatal error LNK1120: 6 unresolved externals
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\link.EXE"' : return code '0x460'
Stop.
```

I pushed a fix which seems to work better (one VisualStudio 2015 build passed): after all it seems I did not need to teach Makefile.mak about the new `.c` file... Or maybe I do, just not in the way I initially wrote it. Guidance very welcome.

src/tools/openpgp-tool.c
```diff
@@ -49,6 +49,8 @@
 #include "libopensc/log.h"
 #include "libopensc/card-openpgp.h"
 
+#include "openpgp-tool-helpers.c"
```
Jakuje, line 52: This needs to be `openpgp-tool-helpers.h`.

src/tools/Makefile.am
```diff
@@ -25,7 +25,8 @@ EXTRA_DIST = Makefile.mak versioninfo-tools.rc.in versioninfo-opensc-notify.rc.i
 
 noinst_HEADERS = util.h fread_to_eof.h \
 	egk-tool-cmdline.h goid-tool-cmdline.h npa-tool-cmdline.h \
-	opensc-asn1-cmdline.h opensc-notify-cmdline.h pkcs11-register-cmdline.h
+	opensc-asn1-cmdline.h opensc-notify-cmdline.h pkcs11-register-cmdline.h \
+	openpgp-tool-helpers.c
```
Jakuje, line 29: This should be also the header file `openpgp-tool-helpers.h`.

Jakuje: later, you will probably need to modify the following line:
```
-openpgp_tool_SOURCES = openpgp-tool.c util.c
+openpgp_tool_SOURCES = openpgp-tool.c util.c openpgp-tool-helpers.c
```

Jakuje:
> > there is a new algorithm `SC_OPENPGP_KEYALGO_EDDSA` (0x16)
> 
> Added. I followed the existing code and put the literal directly in the code, rather than using the macro. Let me know if I should swap all the literals for their corresponding macros.

I think this is fine. To use macros, you would have to include the header files defining them and I do not think it is available in there.

> > I think you will also have to update the `src/tools/Makefile.mak` for windows builds (the tests are not running in windows).
> 
> Done. I ran a cygwin build, which I believe exercised this part of the makefile.
> 
> > And last nit might be the rewrite of the `prettify_language` function in [b1f0d80](https://github.com/OpenSC/OpenSC/commit/b1f0d8077eab0396ff38e89f93082469e6ff3eb6) which does not mention it in the commit message.
> 
> Ah thanks, this reminded me that I wanted to check about my change collapsing all contiguous `<` chars in `prettify_name`. According to wikipedia, there should be no more than two consecutive `<`, so I reverted that part and dropped the corresponding unittest entry.
> 
> I also documented the changes to `prettify_language` (and `prettify_name`) in the commit message.
>
> I pushed to this merge request's branch, as I believe there should be only few further changes to do to these commits, to get them all tested.

Thanks!

> I pushed a fix which seems to work better (one VisualStudio 2015 build passed): after all it seems I did not need to teach Makefile.mak about the new .c file... Or maybe I do, just not in the way I initially wrote it. Guidance very welcome.

I think this is because you included directly the .c file, which indeed works, but it is not a correct way to build c programs. See the inline comments. With the `#include` of the .h files, you will need most probably need a modification of the windows makefile. Including the c file in the unittest is fine.

vpelletier:
> With the `#include` of the .h files, you will need most probably need a modification of the windows makefile.

Makes sense indeed.

> Including the c file in the unittest is fine.

I tried to make it include the .h and am happy to see that it builds just by specifying the .c as a dependency, and that the object was automatically added in its rule:

```
am_openpgp_tool_OBJECTS = openpgp-tool.$(OBJEXT) \
        $(top_builddir)/src/tools/openpgp-tool-helpers.$(OBJEXT)
```

I pushed, let's see if the tests are happy.

Jakuje (approved):

frankmorgner (approved):

PR #2406, opened 2021-09-11
Fix some lgtm warnings and add its configuration

Jakuje:


Jakuje:
For some reason, this configuration is causing the javascript analysis fail no matter what I do ... so I am not sure if this will help us as it would probably fail every CI run :/

frankmorgner (approved):

PR #2407, opened 2021-09-12
Fix typos found by codespell

DimitriPapadopoulos:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,7 +7,7 @@ NEWS for OpenSC -- History of user visible changes
  * Add threading test to `pkcs11-tool` (#2067)
  * Add support to generate generic secret keys (#2140)
  * `opensc-explorer`: Print information about LCS (Life cycle status byte) (#2195)
- * Add support for Apple's arm64 (M1) binaries, removed TokenD. A seperate installer with TokenD (and without arm64 binaries) will be available (#2179).
+ * Add support for Apple's arm64 (M1) binaries, removed TokenD. A separate installer with TokenD (and without arm64 binaries) will be available (#2179).
  * Support for gcc11 and its new strict aliasing rules (#2241, #2260)
  * Initial support for building with OpenSSL 3.0 (#2343)
  * pkcs15-tool: Write data objects in binary mode (#2324)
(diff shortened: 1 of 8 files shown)


NEWS
```diff
@@ -7,7 +7,7 @@ NEWS for OpenSC -- History of user visible changes
  * Add threading test to `pkcs11-tool` (#2067)
  * Add support to generate generic secret keys (#2140)
  * `opensc-explorer`: Print information about LCS (Life cycle status byte) (#2195)
- * Add support for Apple's arm64 (M1) binaries, removed TokenD. A seperate installer with TokenD (and without arm64 binaries) will be available (#2179).
+ * Add support for Apple's arm64 (M1) binaries, removed TokenD. A separate installer with TokenD (and without arm64 binaries) will be available (#2179).
```
Jakuje, line 10: I think the `NEWS` file should be append only. I was already turned down with one typo fix in there, but lest see what others will think.

Jakuje (approved):

NEWS
DimitriPapadopoulos: OK, I can revert that if needed. I can understand NEWS is a pristine copy of past NEWS, with their typos. Wait and see?

frankmorgner (approved):

DimitriPapadopoulos:
I could add a codespell GitHub Action if you're interested - in another PR.

Jakuje:
I am all for having this in action in OpenSC. But I am sad that github still does not support the third result of the actions "warning" that would not make the whole check fail, but it would be simply to distinguish if there is an issue or not.

How does the action work? Does it check all code base or files changed in the PR?

@frankmorgner what about the the NEWS file? Last time I tried to [fix](https://github.com/OpenSC/OpenSC/pull/1960#discussion_r530191835) something in there, it was turned down.

DimitriPapadopoulos:
I think the action checks the whole code base. I haven't looked into the details.

frankmorgner:
here's what we *should* have in NEWS... https://www.gnu.org/prep/standards/html_node/NEWS-File.html

Looking back at this today, I think we should be fine correcting typos...

Jakuje:
> I think the action checks the whole code base. I haven't looked into the details.

That would mean that we would have to have a lot of "exceptions" in dictionary to have some useful reports in the PRs, which I am not sure how viable is. I don't have time to experiment with it right now, but if you could draft a something that will work, I think we can make use of it.

DimitriPapadopoulos:
I'll do that in a different PR.

Jakuje:
> I'll do that in a different PR. Is that OK?

Yes please.

I merged this one as all the outstanding comments are resolved. Thank you!

PR #2409, opened 2021-09-15
Codespell

DimitriPapadopoulos:
* Fix remaining typos
* Add to CI with a list of words to ignore and skipping vendored `compat_*` files

##### Checklist


src/libopensc/asn1.c
```diff
@@ -1689,7 +1689,7 @@ static int asn1_decode(sc_context_t *ctx, struct sc_asn1_entry *asn1,
 	size_t left = len, objlen;
 
 	sc_debug(ctx, SC_LOG_DEBUG_ASN1,
-		 "%*.*scalled, left=%"SC_FORMAT_LEN_SIZE_T"u, depth %d%s\n",
+		 "%*.*s called, left=%"SC_FORMAT_LEN_SIZE_T"u, depth %d%s\n",
```
DimitriPapadopoulos, line 1692: Not sure about this fix. Can you have a look?

Jakuje: afaik this works as expected, adding indentation based on the depth we are in. Adding one more space indents the whole log message even more. I think the right solution here would be to split these words with the quotes so codespell is happy but the log message is not modified such as
```
-		 "%*.*scalled, left=%"SC_FORMAT_LEN_SIZE_T"u, depth %d%s\n",
+		 "%*.*s""called, left=%"SC_FORMAT_LEN_SIZE_T"u, depth %d%s\n",
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,37 @@
...
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: codespell-project/actions-codespell@master
+        with:
+          skip: opensc.conf.example.in
+          ignore_words_list:
```
Jakuje, line 18: I would rather have this in separate file (ideally named in a way that local codespell can use it). The yml does not look like accepting comments well

src/libopensc/asn1.c
DimitriPapadopoulos: Fixed as suggested.

.github/workflows/codespell.yml
DimitriPapadopoulos: Sure. Where do I put the `codespell_ignore_words_file.txt` file?

DimitriPapadopoulos: In `.github`?

DimitriPapadopoulos:
By the way, I see 2 files with CRLF Windows line endings:
```
src/smm/smm-local.exports
src/smm/sm-common.exports
```
All other files have LF Unix line endings. Can I fix that? Or does this exception have a specific purpose?

.github/workflows/codespell.yml
Jakuje: either in `.github` or in root directory. I would probably like it in root, unless there would be any strong opinions against this. Ideally, with a new section about the way how to run the codespell locally in `CONTRIBUTING.md`.

.github/codespell_ignore_words_file.txt
```diff
@@ -0,0 +1,23 @@
+aci
```
Jakuje, line 1: Where is this used? I can not find it in the sources anywhere?
Does this file support some kind of comments to give some comments why the unclear words are here?

Jakuje: I see obvious groups of words in other languages (Portuguese?, German, Russian), which look on the first sight as a typo.

.github/workflows/codespell.yml
DimitriPapadopoulos: I prefer having this file in root too, I was afraid you would find the root would get too crowded. If needed, I can make it a hidden file, for example **`codespell_ignore_words_file.txt`** → **`.codespell_ignore_words_file.txt`**.

Jakuje:
> By the way, I see 2 files with CRLF Windows line endings:
> 
> ```
> src/smm/smm-local.exports
> src/smm/sm-common.exports
> ```
> 
> All other files have LF Unix line endings. Can I fix that? Or does this exception have a specific purpose?

I think this is some historic leftover. I think we should be fine with standard uinx line endings there.

.github/codespell_ignore_words_file.txt
DimitriPapadopoulos: https://github.com/OpenSC/OpenSC/blob/7a6032d3/src/libopensc/card-setcos.c#L439

.github/workflows/codespell.yml
DimitriPapadopoulos: I have added a few words in `CONTRIBUTING.md`.

src/smm/sm-common.exports
```diff
@@ -1,3 +1,3 @@
-sm_cwa_get_mac
-sm_cwa_securize_apdu
-
+sm_cwa_get_mac
+sm_cwa_securize_apdu
+
```
Jakuje, line 3: nit: Maybe the trailing newline is not needed

DimitriPapadopoulos: I can add a commit to remove the empty lines at end of 85 files.
```
$ find . -type f -size +0 -exec gawk 'ENDFILE{if ($0 == "") print FILENAME}' {} +
./configure.ac
./src/tools/westcos-tool.c
./src/tools/dnie-tool.c
./src/common/compat_strlcat.c
./src/common/compat_getopt.c
./src/common/compat_getopt.txt
./src/common/compat_getopt.3
./src/common/simclist.h
./src/common/simclist.c
./src/tests/regression/crypt0007
./src/tests/regression/init0004
./src/tests/regression/erase
./src/tests/regression/init0001
./src/tests/regression/pin0002
[...]
$ 
```

DimitriPapadopoulos: I can also remove spaces at the end of 1617 lines:
```
$ grep -r '[[:blank:]]$' .
./.git/packed-refs:# pack-refs with: peeled fully-peeled sorted 
./src/tools/npa-tool.c:ASN1_ITEM_TEMPLATE(ASN1_AUXILIARY_DATA_NPA_TOOL) = 
./src/tools/npa-tool.c:/** 
./src/tools/npa-tool.c: * 
./src/tools/pkcs15-crypt.c:		
./src/tools/pkcs15-crypt.c:		 * and CKU_CONTEXT_SPECIFIC login to let driver know this verify will be followed by 
./src/tools/pkcs15-crypt.c:		 * to do any special handling. 
./src/tools/npa-tool-cmdline.c:  /usr/bin/gengetopt --file-name=npa-tool-cmdline --output-dir=. 
./src/tools/npa-tool-cmdline.c:  
[...]
$ 
```

DimitriPapadopoulos: Perhaps in a different PR?

Jakuje:
Seems like it already works:
image

I think we can also ignore words only in uris (--uri-ignore-words-list), which would be helpful, but it looks like it is not exported to the action itself:
https://github.com/codespell-project/codespell/blob/ecd1e81915cba81f1caa9544adfd188b37fe5deb/codespell_lib/_codespell.py#L313-L319

Reading through the code, the ignore-words file does not look like supporting comments, but if you create a row for example starting with `#`, it should be just ignored and not match anything. It would be great if this would be documented though. I filled https://github.com/codespell-project/codespell/issues/2063

.github/codespell_ignore_words_file.txt
dengert: Looks like  https://github.com/OpenSC/OpenSC/blob/7a6032d3/src/libopensc/card-setcos.c#L439 should be "ACL" not "ACI".  So aci not needed in codespell_ignore_words_file.txt

src/smm/sm-common.exports
Jakuje: There is probably too much of them. We wanted to avoid mass modifications before #2017 but it is getting delayed. So I would just skip it for now. Maybe just drop it from the file you modified right now.

.github/codespell_ignore_words_file.txt
Jakuje: Good catch. Indeed sounds like a typo from the context to me too

DimitriPapadopoulos:
😳 Fixed the typo.

.github/codespell_ignore_words_file.txt
Jakuje: @DimitriPapadopoulos can you fix also the ACI typo to ACL? I think this will not be either in the default dictionary either.

DimitriPapadopoulos: Fixed.

src/smm/sm-common.exports
DimitriPapadopoulos: Removed the final empty line from `src/smm/sm-common.exports`.

Jakuje (approved):

frankmorgner (approved):

PR #2414, opened 2021-09-24
pkcs11: Unbreak detection of unenrolled cards

Jakuje:
This was broken since 58b03b68, which tried to sanitize some states,
but caused C_GetTokenInfo returning CKR_TOKEN_NOT_RECOGNIZED instead
of empty token information.

Note, that this has effect only if the configuration options
enable_default_driver and pkcs11_enable_InitToken are turned on.
Otherwise it still returns CKR_TOKEN_NOT_RECOGNIZED.

I am still considering writing some tests for this as some applications depend on this.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/restart-pcscd.sh
+++ b/.github/restart-pcscd.sh
@@ -7,7 +7,7 @@
 function pcscd_cleanup {
 	echo "Process terminated: resetting pcscd"
 	sudo pkill pcscd
-	if which systemctl; then
+	if which systemctl && systemctl is-system-running; then
 		sudo systemctl start pcscd.socket
 	fi
 }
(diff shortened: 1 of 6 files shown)


frankmorgner (approved):

kloczek:
Just please make new release even without fixing possible CVE.
If flow of the chengas is enought high it is alwayt better to make more frequent releases with smaller batches of the changes because iless changes overlaps.
As usually no one expects that each release will be poerfect. Only new re;lease must be enough fast reaction of found issue.

Thank you for your work :)
PS. according to my priv observation ~ąll maintainer if already have accumulated +100 commits has growing difficulties to make new release because growing risk of overlaping impacts of the changes.

Issue #2415, opened 2021-10-01
C_WaitForSlotEvent(CKF_DONT_BLOCK) no longer works properly


CendioOssman:
### Problem Description

After upgrading from OpenSC 0.19 to 0.22, we've discovered that `C_WaitForSlotEvent(CKF_DONT_BLOCK)` no longer works properly for reader removal. Card removal works fine though.

I suspect the cause of this regression is #1970, but I have not verified this.

The problem can be reproduced with `pkcs11-tool --test-hotplug` if you apply this patch:

```diff
diff --git a/src/tools/pkcs11-tool.c b/src/tools/pkcs11-tool.c
index 881768f1..9184cde9 100644
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6457,7 +6457,19 @@ static int test_card_detection(int wait_for_event)
                if (wait_for_event) {
                        printf("Calling C_WaitForSlotEvent: ");
                        fflush(stdout);
-                       rv = p11->C_WaitForSlotEvent(0, &slot_id, NULL);
+
+#if 1
+                       CK_ULONG num_slots;
+                       rv = p11->C_GetSlotList(CK_FALSE, NULL, &num_slots);
+                       if (rv != CKR_OK)
+                               p11_fatal("C_GetSlotList(NULL)", rv);
+#endif
+
+                       rv = p11->C_WaitForSlotEvent(CKF_DONT_BLOCK, &slot_id, NULL);
+                       if (rv == CKR_NO_EVENT) {
+                               printf("no event.\n");
+                               continue;
+                       }
                        if (rv != CKR_OK) {
                                printf("failed.\n");
                                p11_perror("C_WaitForSlotEvent", rv);
```

The call to `C_GetSlotList()` is required by the spec. in our understanding, and this is also supported in practice. With the `C_GetSlotList()` in place OpenSC will report reader addition but not reader removal. Removing that call reverses the situation, i.e. OpenSC will report reader removal but not reader addition.

### Proposed Resolution

Unknown. I'm not familiar enough with this code.

### Steps to reproduce

1. Apply above patch
2. Run `pkcs11-tool --test-hotplug`
3. Keep stepping through `C_WaitForSlotEvent()`
4. Plug in reader with card
5. Remove reader with card

### Logs

Log from poll when using OpenSC 0.22.0:

https://gist.github.com/CendioOssman/d25873fb09c3958cafca4e254284d682

[...]

PR #2418, opened 2021-10-08
Detached reader macOS

frifl:
This fixes an issue where a detached reader wasn't noticed when running C_WaitForSlotEvent(CKF_DONT_BLOCK) on macOS 11.6. This was found when trying the work around mentioned in issue #2415:

(We're going to see if we can work around this by doing a more complex tango with OpenSC:

 1. `C_WaitForSlotEvent(CKF_DONT_BLOCK)`
 2. `C_GetSlotList(NULL)`
 3. `C_WaitForSlotEvent(CKF_DONT_BLOCK)`

_Originally posted by @CendioOssman in https://github.com/OpenSC/OpenSC/issues/2415#issuecomment-932120601_)

After my changes, I have tested that the following events are observed with the work around mentioned above: 
✓ Remove card
✓ Detach reader with card in
✓ Insert card
✓ Attach reader with card in
✓ Change to another card in the reader

Tested on both Fedora 34 and macOS 11.6. The last point about changing to another card did not work as expected on macOS. I have tested this before my changes and it didn't work then either.

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -415,7 +415,8 @@ static int refresh_attributes(sc_reader_t *reader)
 		 * XXX: We'll hit it again, as no readers are removed currently.
 		 */
 		reader->flags &= ~(SC_READER_CARD_PRESENT);
-		return SC_ERROR_READER_DETACHED;
+		sc_log(reader->ctx, "Reader unknown: %s", sc_strerror(SC_ERROR_READER_DETACHED));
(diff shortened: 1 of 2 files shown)


frifl:
I have now added a section in the release testing for detaching and attaching reader with the card in it. I did the instructions using the ThinLinc client because that's what I used when finding this issue. It might be possible to test with another application that lists certificates in real time in a non-blocking way, but I have not investigated this.

PR #2425, opened 2021-10-29
Jcop4 NQ-Applet

jozsefd:
##### Checklist

- [ ] macOS tokend is tested

## Cards Supported
This PR contains the card driver and the PKCS15 emulator for the JCOP4 based cards released with the NQ Applet by Bundesagentur für Arbeit, a German Federal Agency. The NQ Applet contains 2 3072-Bit RSA private keys with the associated certificates, intended for non qualified PKI operations. The card may include a Q Applet for qualified signatures,  not included in this PR.

The NQ Applet has a limited ISO compatible APDU set, for instance neither SELECT FILE nor MSE is supported.

## 3072-Bit Raw RSA
The applet can execute only raw RSA calculations, thus the input and the output of such calculations require buffers for 384 octets. Unfortunately the function sc_pkcs15_compute_signature() in pkcs15-sec.c defined the buffer size as `buflen = inlen + 256;` which may result an SC_ERROR_BUFFER_TOO_SMALL during padding, like in sc_pkcs1_add_01_padding().

I changed the definition of this buffer size to: `buflen = inlen + modlen;`

### DIFF
--- a/src/libopensc/card-nqApplet.c
+++ b/src/libopensc/card-nqApplet.c
@@ -1,0 +1,553 @@
+/*
+ * Support for the JCOP4 Cards with NQ-Applet
+ *
+ * Copyright (C) 2020 jozsefd <jozsef.dojcsak@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "asn1.h"
+#include "cardctl.h"
+#include "internal.h"
+#include "log.h"
+
+#define APPLET_VERSION_LEN 2
+#define APPLET_MEMTYPE_LEN 1
+#define APPLET_SERIALNR_LEN 8
+
+/* card constants */
+static const struct sc_atr_table nqapplet_atrs[] = 
+{
+	{ "3b:d5:18:ff:81:91:fe:1f:c3:80:73:c8:21:10:0a", NULL, NULL, SC_CARD_TYPE_NQ_APPLET, 0, NULL },
+	{ NULL, NULL, NULL, 0, 0, NULL }
+};
+
+static const u8 nqapplet_aid[] = {0xd2, 0x76, 0x00, 0x01, 0x80, 0xBA, 0x01, 0x44, 0x02, 0x01, 0x00};
+
+static struct sc_card_operations nqapplet_operations;
+static struct sc_card_operations *iso_operations = NULL;
+
+#define KEY_REFERENCE_NO_KEY	0x00
+#define KEY_REFERENCE_AUTH_KEY	0x01
+#define KEY_REFERENCE_ENCR_KEY	0x02
+
+struct nqapplet_driver_data
+{
+	u8 version_minor;
+	u8 version_major;
+	u8 key_reference;
+	u8 serial_nr[APPLET_SERIALNR_LEN];
+};
+typedef struct nqapplet_driver_data * nqapplet_driver_data_ptr;
+
+static struct sc_card_driver nqapplet_driver = 
+{
+	"NQ-Applet",      // name
+	"nqapplet",      // short name
+	&nqapplet_operations,	// operations
+	NULL,           // atr table
+    0,              // nr of atr
+    NULL            // dll?
+};
+
+static const struct sc_card_error nqapplet_errors[] = 
+{
+	{ 0x6700, SC_ERROR_WRONG_LENGTH, "Invalid LC or LE"}, 
+	{ 0x6982, SC_ERROR_SECURITY_STATUS_NOT_SATISFIED, "Security status not satisfied" }, //TODO MK/DK??
+	{ 0x6985, SC_ERROR_NOT_ALLOWED,	"Invalid PIN or key" },
+	{ 0x6986, SC_ERROR_NOT_ALLOWED,	"Conditions of use not satisfied" },
+	{ 0x6A80, SC_ERROR_INVALID_ARGUMENTS, "Invalid parameters" },
+	{ 0x6A82, SC_ERROR_OBJECT_NOT_FOUND, "Data object not found"},
+	{ 0x6A84, SC_ERROR_NOT_ENOUGH_MEMORY, "Not enough memory" },
+	{ 0x6A86, SC_ERROR_INCORRECT_PARAMETERS, "Invalid P1 or P2" }, 
+	{ 0x6A88, SC_ERROR_INVALID_ARGUMENTS, "Wrong key ID" },
+	{ 0x6D00, SC_ERROR_FILE_NOT_FOUND, "Applet not found" }
+};
+
+/* convenience functions */
+
+static int init_driver_data(sc_card_t *card, u8 version_major, u8 version_minor, u8 * serial_nr, size_t cb_serial_nr)
+{
+	nqapplet_driver_data_ptr data = calloc(1, sizeof(struct nqapplet_driver_data));
+	if (data == NULL)
+	{
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
+	}
+
+	data->version_major = version_major;
+	data->version_minor = version_minor;
+	data->key_reference = KEY_REFERENCE_NO_KEY;
+	memcpy(data->serial_nr, serial_nr, min(cb_serial_nr, sizeof(data->serial_nr)));
+	card->drv_data = (void*)data;
+	return SC_SUCCESS;
+}
+
+/**
+ * SELECT NQ-Applet, on success it returns the applet version and card serial nr.
+ *
+ * @param[in]     	card
+ * @param[out,opt]  version_major	Version major of the applet
+ * @param[out,opt]  version_minor	Version minor of the applet
+ * @param[out,opt]  serial_nr		Buffer to receive serial number octets
+ * @param[in]  		cb_serial_nr	Size of buffer in octets
+ * @param[out,opt]	serial_nr_len	The actual number of octet copied into serial_nr buffer
+ *
+ * @return SC_SUCCESS: The applet is present and selected.
+ * 
+ */
+static int select_nqapplet(sc_card_t *card, u8 *version_major, u8 *version_minor, u8 *serial_nr, size_t cb_serial_nr, size_t *serial_nr_len)
+{
+	int rv;
+	sc_context_t *ctx = card->ctx;
+	sc_apdu_t apdu;
+	u8 buffer[APPLET_VERSION_LEN+APPLET_MEMTYPE_LEN+APPLET_SERIALNR_LEN+2];
+	size_t cb_buffer = sizeof(buffer);
+	size_t cb_aid = sizeof(nqapplet_aid);
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0x04, 0x00);
+	apdu.lc = cb_aid;
+	apdu.data = nqapplet_aid;
+	apdu.datalen = cb_aid;
+	apdu.resp = buffer;
+	apdu.resplen = cb_buffer;
+	apdu.le = cb_buffer;
+
+	rv = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(ctx, rv, "APDU transmit failure.");
+
+	rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	LOG_TEST_RET(card->ctx, rv, "Card returned error");
+
+	if ( version_major != NULL )
+	{
+		*version_major = buffer[0];
+	}
+	if ( version_minor != NULL )
+	{
+		*version_minor = buffer[1];
+	}
+	if ( serial_nr != NULL && cb_serial_nr > 0 && serial_nr_len != NULL )
+	{
+		size_t cb = MIN(APPLET_SERIALNR_LEN, cb_serial_nr);
+		memcpy(serial_nr, buffer+3, cb);
+		*serial_nr_len = cb;
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+
+/* driver operations API */
+static int nqapplet_match_card(struct sc_card *card)
+{
+	int rv = _sc_match_atr(card, nqapplet_atrs, &card->type);
+	return (rv >= 0);
+}
+
+static int nqapplet_init(struct sc_card *card)
+{
+	u8 version_major;
+	u8 version_minor;
+	u8 serial_nr[APPLET_SERIALNR_LEN];
+	size_t cb_serial_nr = sizeof(serial_nr);
+	unsigned long rsa_flags = 0;
+
+	LOG_FUNC_CALLED(card->ctx);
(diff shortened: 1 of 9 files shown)


src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+		}
+		break;
+	}
+	return SC_ERROR_NOT_SUPPORTED;
+}
+
+
+struct sc_card_driver * sc_get_nqApplet_driver(void)
+{
+	sc_card_driver_t * iso_driver = sc_get_iso7816_driver();
```
frankmorgner, line 501: some of the functionality above looks almost like implemented in the iso7816 driver (e.g. get response, get data, get challenge). do you see some limitations in the iso driver that could be improved?

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+int nqapplet_set_security_env(struct sc_card *card, const struct sc_security_env *env, int se_num)
+{
+	/* Note: the NQ-Applet does not have APDU for SET SECURITY ENV, 
+	this function checks the intended parameters and sets card_data.key_reference */
+	nqapplet_driver_data_ptr data;
+	
+	assert(card != NULL && env != NULL);
```
frankmorgner, line 288: please return an error code instead

frankmorgner (changes requested):
This looks very nice. Is there some product information for the card available on the internet?

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+
+/* convenience functions */
+
+static int init_driver_data(sc_card_t *card, u8 version_major, u8 version_minor, u8 * serial_nr, size_t cb_serial_nr)
+{
+	nqapplet_driver_data_ptr data = calloc(1, sizeof(struct nqapplet_driver_data));
+	if (data == NULL)
+	{
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
+	}
```
Jakuje, line 90: Please, use the braces formatting as in the rest of the code. Some formal hints/references might be found in #2017, even though they are not finalized yet.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+	apdu.resplen = cb_buffer;
+	apdu.le = cb_buffer;
+
+	rv = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(ctx, rv, "APDU transmit failure.");
+
+	rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	LOG_TEST_RET(card->ctx, rv, "Card returned error");
+
+	if ( version_major != NULL )
```
Jakuje, line 138: you should check first that the returned Lc was large enough otherwise you are reading uninitialized memory.

src/libopensc/pkcs15-nqApplet.c
```diff
@@ -0,0 +1,238 @@
...
+
+	// 2.1) CH.Auth
+	rv = add_nqapplet_certificate(p15card, "1", "C.CH.Auth", 0x00);
+	LOG_TEST_RET(card->ctx, rv, "Failed to add Auth. certificate");
+
+	// 2.2) PrK.CH.Auth
+	rv = add_nqapplet_private_key(p15card, "1", 0x01, "PrK.CH.Auth", "1", SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_ENCRYPT | SC_PKCS15_PRKEY_USAGE_DECRYPT);
+	LOG_TEST_RET(card->ctx, rv, "Failed to add Auth. private key");
+
+	// 2.1) CH.Auth
```
Jakuje, line 168: I assume this should say `3.1 C.CH.Encr`. Similarly comment bellow.

src/libopensc/pkcs15-nqApplet.c
```diff
@@ -0,0 +1,238 @@
...
+	// 1) User PIN
+	rv = add_nqapplet_pin(p15card, "1", 0x01);
+	LOG_TEST_RET(card->ctx, rv, "Failed to add PIN 1");
+
+	// 2.1) CH.Auth
+	rv = add_nqapplet_certificate(p15card, "1", "C.CH.Auth", 0x00);
+	LOG_TEST_RET(card->ctx, rv, "Failed to add Auth. certificate");
+
+	// 2.2) PrK.CH.Auth
+	rv = add_nqapplet_private_key(p15card, "1", 0x01, "PrK.CH.Auth", "1", SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_ENCRYPT | SC_PKCS15_PRKEY_USAGE_DECRYPT);
```
Jakuje, line 165: I think the `SC_PKCS15_PRKEY_USAGE_ENCRYPT` should not be here as its not a private key operation.

src/libopensc/pkcs15-nqApplet.c
```diff
@@ -0,0 +1,238 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "common/compat_strlcpy.h"
+#include "log.h"
+#include "pkcs15.h"
+#include "cards.h"
+
+static const char name_Card[] = "NQ-Applet";
+static const char name_Vendor[] = "JCOP4";
```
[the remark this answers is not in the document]
jozsefd: As far as I know the Chip vendor is NXP and the Card is produced by Secunet AG for Bundesagentur für Arbeit (BA). The card is referenced in all documents as JCOP4, that's why I used this name. I can amend it to NXP (or BA?), if you wish.

src/libopensc/pkcs15-nqApplet.c
```diff
@@ -0,0 +1,238 @@
...
+	memset(&cert_info, 0, sizeof(cert_info));
+	memset(&cert_obj, 0, sizeof(cert_obj));
+
+	sc_pkcs15_format_id(id, &cert_info.id);
+	cert_info.authority = 0;
+	rv = get_nqapplet_certificate(card, data_id, &cert_info.value);
+	LOG_TEST_RET(card->ctx, rv, "Failed to get certificate");
+
+	strlcpy(cert_obj.label, name, sizeof(cert_obj.label));
+	cert_obj.flags = 0;
```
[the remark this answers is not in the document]
jozsefd: As far as I know some compilers initialize a struct defined on the stack with unspecified values and this struct has lots of members. I just wanted to avoid unexpected side effects in sc_pkcs15emu_add_x509_cert.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+	int rv;
+	sc_context_t *ctx = card->ctx;
+	sc_apdu_t apdu;
+	u8 buffer[APPLET_VERSION_LEN+APPLET_MEMTYPE_LEN+APPLET_SERIALNR_LEN+2];
+	size_t cb_buffer = sizeof(buffer);
+	size_t cb_aid = sizeof(nqapplet_aid);
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0x04, 0x00);
```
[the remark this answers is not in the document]
jozsefd: I wanted to express that this APDU sends data and expects response data (explicit Case 4), I changed it to sc_format_apdu_ex.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+	if ( version_minor != NULL )
+	{
+		*version_minor = buffer[1];
+	}
+	if ( serial_nr != NULL && cb_serial_nr > 0 && serial_nr_len != NULL )
+	{
+		size_t cb = MIN(APPLET_SERIALNR_LEN, cb_serial_nr);
+		memcpy(serial_nr, buffer+3, cb);
+		*serial_nr_len = cb;
+	}
```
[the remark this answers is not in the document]
jozsefd: I check the sw1/sw2 before using apdu.resp and according to the card specs. if no error reported then the response data is present. Anyway, it is is a good idea to check the length of the response, so I added it.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,556 @@
...
+#define KEY_REFERENCE_NO_KEY	0x00
+#define KEY_REFERENCE_AUTH_KEY	0x01
+#define KEY_REFERENCE_ENCR_KEY	0x02
+
+struct nqapplet_driver_data
+{
+	u8 version_minor;
+	u8 version_major;
+	u8 key_reference;
+	u8 serial_nr[APPLET_SERIALNR_LEN];
```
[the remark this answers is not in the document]
jozsefd: I was trying to encapsulate all available information about the applet in struct nqapplet_driver_data, but you are right, this serial_nr field is kind of redundant here, removed it.

src/libopensc/card-nqApplet.c
jozsefd: I tried to use the iso7816 driver where it was possible, in fact I implemented only those functions which failed during testing. For instance the difference between iso7816_get_response and my nqapplet_get_response is the "cla" value. Unfortunately the applet uses 3 different cla values for its APDUs, so I could not rely on setting the card->cla.
This applet is not even trying to be iso compatible, so all of these incompatibilities are caused by the applet implementation and not the iso driver.

src/libopensc/card-nqApplet.c
jozsefd: It seems that other card drivers don't bother with parameter checking, so I can assume that at this point these parameters are correct. Removed the asserts.

src/libopensc/card-nqApplet.c
jozsefd: Adjusted braces formatting as in rest of the code.

src/libopensc/card-nqApplet.c
jozsefd: Fixed it, added checking of apdu.resplen

src/libopensc/pkcs15-nqApplet.c
jozsefd: Oops, forgotten to adjust comments. Fixed it.

src/libopensc/pkcs15-nqApplet.c
jozsefd: You are right, removed it.

src/libopensc/pkcs15-nqApplet.c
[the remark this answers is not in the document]
jozsefd: Changed vendor to NXP

jozsefd:
> 
> This looks very nice. Is there some product information for the card available on the internet?
> 
The issuer of the card is [Vertrauensdiensteanbieter der Bundesagentur für Arbeit](https://www.pki.arbeitsagentur.de/)

Additional information for the vendor/chip/os:
- [JCOP 4 P71](https://www.commoncriteriaportal.org/files/epfiles/[ST-LITE][v3.4.1]%20JCOP4_P71_SecurityTargetLite_v3.4.1.pdf)
- [Certification Report](https://www.commoncriteriaportal.org/files/epfiles/Certification%20Report%20NSCIB-CC-180212-CR3.pdf)
- [Assurance Continuity Maintenance Report](https://www.commoncriteriaportal.org/files/epfiles/NSCIB-CC-180212-MA.pdf)

frankmorgner (approved):
Thanks for looking at the comments, the code looks good now.

With your research, it seems like *Vertrauensdiensteanbieter der Bundesagentur für Arbeit* should work best as "vendor". According to [this internal directive](https://www.arbeitsagentur.de/datei/Weisung201708015_ba014579.pdf) your card may be the *Digitale Dienstkarte (dDk)*, but I don't know for sure.

jozsefd:
> Thanks for looking at the comments, the code looks good now.
> 
> With your research, it seems like _Vertrauensdiensteanbieter der Bundesagentur für Arbeit_ should work best as "vendor". According to [this internal directive](https://www.arbeitsagentur.de/datei/Weisung201708015_ba014579.pdf) your card may be the _Digitale Dienstkarte (dDk)_, but I don't know for sure.

Yes, I have also found this internal directive, but it lacks of technical details. The card can be either a "dDK" which is issued for the employees of the organization (over 100k cards!), or a guest/mass-signature card.

src/libopensc/pkcs15-nqApplet.c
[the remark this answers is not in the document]
Jakuje: This was not resolved yet. Setting flats to `0` is not needed when you previously called memset on the whole structure. If you would not, it would indeed be needed (ideally for all the members). This way, it will probably get optimized out, but it unneeded step. The same for the `cert_info.authority` above.

src/libopensc/pkcs15-nqApplet.c
```diff
@@ -0,0 +1,235 @@
...
+			sc_format_path("3f00", &file->path);
+			sc_file_free(p15card->file_app);
+			p15card->file_app = file;
+		}
+	}
+
+	if (p15card->tokeninfo != NULL)
+	{
+		sc_pkcs15_free_tokeninfo(p15card->tokeninfo);
+	}
```
Jakuje, line 216: The check for NULL is not needed. It is already checked inside of the `sc_pkcs15_free_tokeninfo()` function.

src/libopensc/card-nqApplet.c
Jakuje: I do not see this addressed yet. You can run the `clang-format` with the configuration attached to the #2017 to get close to the expected formatting.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,555 @@
...
+
+	if (count < apdu.resplen)
+	{
+		return (int) count;
+	}
+   
+	return (int) apdu.resplen;
+}
+
+static int nqapplet_logout(struct sc_card *card)
```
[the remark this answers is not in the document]
jozsefd: The only option in this applet to reset the PIN status is selecting the applet again, that's why nqapplet_logout uses this command.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,555 @@
...
+{
+	/* Note: the NQ-Applet does not have APDU for SET SECURITY ENV, 
+	this function checks the intended parameters and sets card_data.key_reference */
+	nqapplet_driver_data_ptr data;
+	
+	LOG_FUNC_CALLED(card->ctx);
+
+	data = (nqapplet_driver_data_ptr)card->drv_data;
+	data->key_reference = KEY_REFERENCE_NO_KEY;
+	u8 key_reference = KEY_REFERENCE_NO_KEY;
```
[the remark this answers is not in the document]
jozsefd: Oops, I got too much C++ recently, I moved the definition to the right place.

src/libopensc/card-nqApplet.c
```diff
@@ -0,0 +1,555 @@
...
+	nqapplet_operations.reset_retry_counter = NULL;
+	nqapplet_operations.create_file = NULL;
+	nqapplet_operations.delete_file = NULL;
+	nqapplet_operations.list_files = NULL;
+	nqapplet_operations.process_fci = NULL;
+	nqapplet_operations.construct_fci = NULL;
+	nqapplet_operations.put_data = NULL;
+	nqapplet_operations.delete_record = NULL;
+	nqapplet_operations.read_public_key = NULL;
+	nqapplet_operations.card_ctl = nqapplet_card_ctl;
```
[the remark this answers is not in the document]
jozsefd: You are right, moved it.

src/libopensc/pkcs15-nqApplet.c
jozsefd: Removed it.

src/libopensc/card-nqApplet.c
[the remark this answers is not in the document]
Jakuje: Can you add this to the comment for this function?

src/libopensc/pkcs15-nqApplet.c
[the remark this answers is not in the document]
jozsefd: I suppose it would be optimized out, but anyway, removed the unnecessary code.

src/libopensc/card-nqApplet.c
jozsefd: I had another clang-format configuration, updated it and reformatted the files.

src/libopensc/card-nqApplet.c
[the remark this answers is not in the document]
jozsefd: I have already added a comment to this function:
`/* selecting NQ-Applet again will reset the applet status and unauthorize PINs */`

Jakuje (approved):

frankmorgner (approved):

Issue #1595, opened 2019-01-25
Unused selection of padding mechanism in pkcs15_prkey_unwrap


frankmorgner:
### Problem Description

There's an assignment to the variable `flags`, which doesn't get used:

https://github.com/OpenSC/OpenSC/blob/70d690ace7a5ddc5a534e7455bdc1800e26cb807/src/pkcs11/framework-pkcs15.c#L4031-L4041

### Proposed Resolution

@hhonkanen please check what's going on there... and/or if the above sniplet can be removed.

hhonkanen:
I think we should not remove this code, but rather pass flags to sc_pkcs15_unwrap(). Looks like we have just forgotten to pass it forward, and it works fine in our tests regardless of this. We now just pass zero to flags.  

Both mechanisms should be tested after fixing this. I have tested the code only with CKM_RSA_PKCS, and it works correctly event though flag SC_ALGORITHM_RSA_PAD_PKCS1 is not passed forwards.

Jakuje:
@hhonkanen The code is commented out since 2019. Do you have some particular idea why these flags would be used in `sc_pkcs15_unwrap()`? If not, I would propose to remove the commented code in
https://github.com/OpenSC/OpenSC/blob/a00ead5bcaeef9761287572f7b62065d62c211d9/src/pkcs11/framework-pkcs15.c#L4308-L4321

hhonkanen:
@Jakuje I would prefer to not remove the code, but rather pass the flags to sc_pkcs15_unwrap() at lines 4329-4330. The idea of the flags argument is to pass details about the RSA encrypted key data to the card, for example which padding has been used. 
The cards that implement unwrap operation may need this information from the software that is using them. Two flags have been defined there, but there might be need to extend this and support more mechansisms/flags in future. 

The flags argument should be passed after targetKeyObj->p15_object like this:

	/* Call the card to do the unwrap operation */
	rv = sc_pkcs15_unwrap(fw_data->p15_card, prkey->prv_p15obj, targetKeyObj->p15_object, flags,
		pWrappedKey, ulWrappedKeyLen, NULL, 0);

The reason why the flags argument is unused is that I have just forgotten to pass it too sc_pkcs15_unwrap, and it happens to work anyway, with flags being zero, with MyEID and the test data we used. 

My opensc dev environment is currently out of order, and I haven't installed the tools yet on my new workstation. Could you please fix the code and commit it? 

[...]

PR #2430, opened 2021-11-02
Symmetric keys support in p11test + rough edges of symmetric keys support in OpenSC

Jakuje:
I started writing this as an intention to test wrapping and unwrapping as discussed in #1595 where we needed some verification of couple of changes that went in recently. But it turned out that I did not know how these keys are working and there was no support in the testing framework for this so I started with symmetric keys first, which already turned up couple of issues, which are fixed on the way.

I got down up to the key wrapping, but not yet to the unwrapping part which is why I started. The changeset is growing so I wanted to part it somewhere, got some early feedback and hopefully continue with the original target (either in this PR or in another one).

There are also couple of interleaved irrelevant patches on the way, I can split away if needed.

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
@popovec @hhonkanen @fabled you all contributed to the support of symmetric keys in OpenSC. This PR is doing some pkcs11-level tests on these so I would be glad for feedback, as this is basically my first use of secret keys in pkcs11.

popovec:
Unfortunately, there is no real support for symmetric enc / decrypt in OpenSC for MyEID card. Currently, the MyEID driver only supports uploading a symmetric key to the card. (Test of this operation can be added to github actions, with OsEID simulation, it works without problems). 

It is about one year ago - a have tested the @carblue patch and my MyEID driver patch -  outdated  version of can be found on github.. https://github.com/OpenSC/OpenSC/issues/1796#issuecomment-753611243  However, this solution is not complete - there is support for C_EncryptInit (), C_Encrypt () but there is no support for C_EncryptUpdate () and C_EncryptFinal ()... (similar for decrypt - no support  for C_DecryptUpdate .. C_DecryptFinal .. ). 

Due to the incompleteness of this solution, I have not yet offered my MyEID patch as a PR for OpenSC. 

If someone can add symmetric encryption support to the OpenSC (at least similar to the one outlined by @carblue), I will prepare a patch for the MyEID driver that will use it. I would like the solution to be complete -  it must also include C_EncryptFinal, C_Encrypt_Update, and the solution must have correspondingly designed interface for the card driver. Personally, I currently don't have enough time to start writing such a big patch. 

I would like to comment on problem #1796. This test is incomplete and strongly limited to the use of the CKM_DES_CBC mechanism. If any card supports AES-KEY-WRAP but does not have DES_CBC support for unwrap/wrap .. it will not work. Here we need to add a listing of wrap/unwrap mechanisms and perform a series of tests accordingly to the available wrap/unwrap mechanisms.

PR #2431, opened 2021-11-04
sc-hsm-tool: Add --transport-pin option

saper:
SmartCard HSM can be initialized with a Transport PIN option. 

While this change seems to work, we cannot use other OpenSC commands to manage the card right after the initialization. due to https://github.com/OpenSC/OpenSC/issues/2432, therefore marking as "draft".


frankmorgner:
Won't fix, see https://github.com/OpenSC/OpenSC/issues/2432

PR #2435, opened 2021-11-09
PIV: Use a single (truncated) AID for selection

frankmorgner:
Simplify the code so that only the AID without version information is
used for applet selection. Note that no other AID has ever been used.

##### Checklist

- [ ] macOS tokend is tested


dengert:
This is another nitpik change that you you are tyring to add to card-piv.c before adding #2053. 

See my comment https://github.com/OpenSC/OpenSC/pull/2053#issuecomment-964229161

frankmorgner:

PR #2439, opened 2021-11-10
PIV: split up card matching from initializing

frankmorgner:
- Separate detection of card type from card initialization; extract
initialization of the private data from `piv_match_card_continued()`
into `piv_init_priv()`
- Clearify lock handling; the card is unlocked in the function where it
is locked to avoid responsibility passing
- `piv_activate()` makes sure that the PIV applet is selected during
card matching and initialization.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -468,6 +468,7 @@ static struct sc_card_driver piv_drv = {
 };
 
 static int piv_match_card_continued(sc_card_t *card);
+static int piv_activate(sc_card_t *card);
 
 static int
 piv_find_obj_by_containerid(sc_card_t *card, const u8 * str)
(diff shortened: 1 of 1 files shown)


Jakuje (approved):

dengert:
@frankmorgner @Jakuje  You do realize that #2053 has major changes in this same area, and I have started to address the your comments starting at  https://github.com/OpenSC/OpenSC/pull/2053#issuecomment-964943760 unless this is a bug that you are trying to fix having both of us trying to change the same code is making it harder not easier for both of us. 

I would like to ask that you look at the #2053 changes  in this area before committing this change to master.

frankmorgner:
we can wait until we've finalized #2053

frankmorgner:

Jakuje:
Right. But its always easier to review smaller and self-contained changes than this large diff. I wanted to clean up the backlog of approved changes before getting to this. I will try to have a look to the #2053 now.

PR #2441, opened 2021-11-11
PIV: Fixed detection of PIV AID via empty discovery

frankmorgner:
Since we are using the discovery object to detect whether it's a proper
PIV application, `piv_parse_discovery()` should not return SC_SUCCESS on
empty files or on files with *some* ASN.1 structure... Instead, at least
the encoding of the PIV AID should be strictly according NIST SP 800-73-4

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2573,57 +2573,37 @@ static int piv_select_file(sc_card_t *card, const sc_path_t *in_path,
 
 }
 
-static int piv_parse_discovery(sc_card_t *card, u8 * rbuf, size_t rbuflen, int aid_only)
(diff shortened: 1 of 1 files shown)


dengert:
This was in #2053

frankmorgner:
Well, partially. `piv_parse_discovery()` still [returns `SC_SUCCESS` for an empty file](https://github.com/OpenSC/OpenSC/blob/4f60c2fdc1028a45417abbcff8e0b5479356f7b4/src/libopensc/card-piv.c#L4686).

Jakuje (changes requested):

frankmorgner:

PR #2448, opened 2021-11-19
Simplify PIV card driver

frankmorgner:
This PR is work in progress. The changes are meant to simplify the PIV card driver, more commits may be added later... Feedback is welcome at any time.

##### Checklist

- [ ] macOS tokend is tested


dengert:
@frankmorgner You have a lot of changes to PIV driver. If you feel these are really worth committing, go ahead. I don't have much time over the next month and a half to look at these.

mouse07410:
PIV driver evolved with workarounds for many weird behavior aberrations manifested in different tokens on the market.

By now probably even @dengert doesn't remember exactly what was done for what specific problem. 

It would be unfortunate if these simplifications "simplified out" a workaround, whose effect would surface several months after deployment, and then take several more months to hunt and fix (again). After people already spent plenty of time and efforts to mail those down - and who has free time nowadays? This is my major concern.

_I understand the value of the code being manageable with less efforts. But I prefer better-working and less-manageable to very-manageable and broken._

frankmorgner:
> @frankmorgner You have a lot of changes to PIV driver. If you feel these are really worth committing, go ahead. I don't have much time over the next month and a half to look at these.

I'm not in a hurry. We can merge these changes when SM is done.

frankmorgner:
> _I understand the value of the code being manageable with less efforts. But I prefer better-working and less-manageable to very-manageable and broken._

Your concerns are heard. I have local tests and we have CI tests. Feel free to run some tests yourself or (better) extend the CI tests with the missing use cases.

mouse07410:
> Your concerns are heard. 

I appreciate it.

> I have local tests and we have CI tests. 

I strongly doubt that every corner case and every workaround implemented during the life of this driver has a corresponding test. It would probably be impossible to even write such a test for a generic PIV - it would likely require the same specific physical token, for which the workaround was implemented.

> Feel free to run some tests yourself or (better) extend the CI tests with the missing use cases.

There are two problems. One is time - the same reason you'd rather spend less time and efforts on maintaining the code. I simply don't have enough time to become a co-maintainer of the PIV driver (and very likely don't have enough expertise - my primary interest was getting the whole thing to work with the PIV and CAC tokens).
Second problem - at I said, I don't quite see how you can test for a quirk of a specific device, unless you run the code against that specific device. And I don't even remember all the quirks of Yubikey and CAC - I'm sure there are other PIV implementations around with different incompatibilities. 

If the purpose of this driver (and OpenSC at large) was to validate various tokens, it would be trivially easy: just implement the strict PIV, and if a token fails - is a bad non-compliant token. But the goal is to make all these non-compliant (to different degrees) tokens work. As you yourself remember, it was pretty hard... I for one don't have time now to redo all that work... And I can't even judge the changes, saying without having to offer supporting evidence: this change is on but that one is likely to break token X...
Those are the concerns...

PR #2451, opened 2021-11-24
pkcs15-init: Set algorithm after parsing spec

xhanulik:
Move the algorithm setting in `do_generate_skey()` after parsing the spec.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs15-init.c
+++ b/src/tools/pkcs15-init.c
@@ -1793,13 +1793,14 @@ do_generate_skey(struct sc_profile *profile, const char *spec)
 
 	if ((r = init_skeyargs(&skey_args)) < 0)
 		return r;
-	skey_args.algorithm = algorithm;
 
 	algorithm = parse_alg_spec(alg_types_sym, spec, &keybits, 0);
 	if (algorithm < 0) {
 		util_error("Invalid symmetric key spec: \"%s\"", spec);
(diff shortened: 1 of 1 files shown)


Jakuje:
This is very similar pattern as in `do_store_secret_key()`, where this is handled correctly. I assume this was just mistype.

I actually did not try to generate symmetric key with pkcs15init on my myeid card yet. @fabled or @popovec can you check the change looks fine for you? Does OseID support key generation on card? Do you have it covered with tests Peter?

Jakuje (approved):

popovec:
MyEID/OsEID card is not affected by this PR. There is no support for symmetric key generation on the OsEID card. It is possible to upload a symmetric key to this card. This conforms to the MyEID card specification as described in the MyEID Java Card 2.4.1 Reference Manual.

Jakuje:
> MyEID/OsEID card is not affected by this PR. There is no support for symmetric key generation on the OsEID card. It is possible to upload a symmetric key to this card. This conforms to the MyEID card specification as described in the MyEID Java Card 2.4.1 Reference Manual.

Thanks for checking. So it looks like a unused code. I do not see symmetric key support in any other driver right now so this is likely dead code for opensc.

fabled:
LGTM.

PR #2453, opened 2021-11-25
OpenSSL FIPS_mode not defined in 3.0.0

dengert:
When trying to test with OpenSSL-3.0.0 the use of  `FIPS_mode()`
in two places pkcs11-tool.c and framework-pkcs11.c
causes a run time error like:
"./pkcs11-tool: symbol lookup error: /opt/ossl-3.0.0/lib/opensc-pkcs11.so: undefined symbol: FIPS_mode"

sc-ossl-sc-ossl-compat.h was modified to define it as a macro returning 0.
pkcs11-tool.c was changed to check the OpenSSL version

 On branch master
 Your branch is up to date with 'upstream/master'.

 Changes to be committed:
	modified:   sc-ossl-compat.h

##### Checklist

- [X ] PKCS#11 module is tested with both 1.1.1 and 3.0.0

### DIFF
--- a/src/libopensc/sc-ossl-compat.h
+++ b/src/libopensc/sc-ossl-compat.h
@@ -96,8 +96,11 @@ extern "C" {
 #endif
 
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/provider.h>
 #define EC_POINT_get_affine_coordinates_GFp     EC_POINT_get_affine_coordinates
 #define EC_POINT_set_affine_coordinates_GFp     EC_POINT_set_affine_coordinates
+#define EVP_PKEY_CTX_set_rsa_keygen_pubexp      EVP_PKEY_CTX_set1_rsa_keygen_pubexp
+#define FIPS_mode()                             OSSL_PROVIDER_available(NULL, "fips")
 #endif
 
 /*


Jakuje:
This is unfortunately not that easy as checking for the OpenSSL version. This depends on if the OpenSSL is built with FIPS module or not. Our OpenSSL 3.0 in CentOS/RHEL 9 Beta or Fedora 36/Rawhide has this function so this is not the correct fix.

I think this should be already addressed in #2438, because we previously hit this issue (but there is dozen of other changes that still need a review).

dengert:
The version of 3.0.0 I am using is from github  and tagged as openssl-3.0.0 https://github.com/openssl/openssl/commit/89cd17a031e022211684eb7eb41190cf1910f9fa
The simple `int FIPS_mode(void)` is not defined but the compiler assumes it will be found at link or load time.

But with 3.0.0 the FIPS provider can be used by individual calls using "fips=yes". 
So they no longer support a `int FIPS_mode(void)`.  

So I agree that this not easily solved. 

eb8a381cd and 8fc822aed 2021-08-11 introduced the problem by testing for FIPS_mode, which is fine for pre 3.0, but does not work after 3.0.0 

Without this PR, no further testing of OpenSC's usage of OpenSSL 3.0 can be done. pkcs11-tool fails, and opensc libs will not load. 

And 3.0.0  version of the FIPS provider is not yet validated. When it is, it may not be validated until some later version, and never in 3.0.0.
https://www.openssl.org/blog/blog/2021/09/22/OpenSSL3-fips-submission/

dengert:
I agree, it looks like 1272521059eadf19e3b58b275efb4b41465804df will solve this problem can this be committed master soon?

Jakuje:
You usually do not validate openssl on its own. What we do we certify the RHEL with the crypto modules including openssl, yet this is still a long run ahead of us, especially with the changes coming to FIPS 140-3 (previously, we did 140-2, which is discontinued).

But all this aside, we want to have applications working in FIPS mode (the system in FIPS mode which automatically loads fips module in openssl instead of the default one) and therefore we need to detect this mode somehow and either skip the non-approved crypto or workaround it with fips=no (intended for non-security uses) or by loading some different provider.

I would love to get the OpenSSL 3.0 support merged soon. I put my comments, which Veronika addressed. I think the changes are quite final by today. But more reviews are welcomed so if you can skim through the code and add your comments, it would help to get it merged, including the github actions running against OpenSSL 3.0.

dengert:
My point in eb8a381 and 8fc822a are in master branch.  https://github.com/OpenSC/OpenSC/commit/1272521059eadf19e3b58b275efb4b41465804df is not. 
So trying to build master using OpenSSL-3.0.0 fails because of  eb8a381 and 8fc822a 
Looking close https://github.com/OpenSC/OpenSC/commit/1272521059eadf19e3b58b275efb4b41465804df at it defines FIPS_mode() as a macro as does this PR

I or someone could modify this PR  to change
#define FIPS_mode() 0
to 
#define FIPS_mode()                             OSSL_PROVIDER_available(NULL, "fips")

in pkcs11-tool.c and sc_ossl-compat.h  to get master working again.

Jakuje:
> My point in [eb8a381](https://github.com/OpenSC/OpenSC/commit/eb8a381cd87dd91ad5e9cf66cc709325c52ab9bb) and [8fc822a](https://github.com/OpenSC/OpenSC/commit/8fc822aedca2f837c072f65bc16a73deeacc13c5) are in master branch. [1272521](https://github.com/OpenSC/OpenSC/commit/1272521059eadf19e3b58b275efb4b41465804df) is not. So trying to build master using OpenSSL-3.0.0 fails because of [eb8a381](https://github.com/OpenSC/OpenSC/commit/eb8a381cd87dd91ad5e9cf66cc709325c52ab9bb) and [8fc822a](https://github.com/OpenSC/OpenSC/commit/8fc822aedca2f837c072f65bc16a73deeacc13c5) Looking close [1272521](https://github.com/OpenSC/OpenSC/commit/1272521059eadf19e3b58b275efb4b41465804df) at it defines FIPS_mode() as a macro as does this PR
> 
> I or someone could modify this PR to change #define FIPS_mode() 0 to #define FIPS_mode() OSSL_PROVIDER_available(NULL, "fips")
> 
> in pkcs11-tool.c and sc_ossl-compat.h to get master working again.

indeed, but I think you do not need the change in `pkcs11-tool.c`. As it is now, it is not correct and does not work in OpenSSL 3.0 with FIPS module. The sc_ossl-compat.h is not included in the pkcs11-tool?

dengert:
`sc-ossl-compat.h` is included in the `pkcs11-tool.c` so only sc-ossl-compat.c needs the change

```
diff --git a/src/libopensc/sc-ossl-compat.h b/src/libopensc/sc-ossl-compat.h
index ba4a0c51..3e4e0080 100644
--- a/src/libopensc/sc-ossl-compat.h
+++ b/src/libopensc/sc-ossl-compat.h
@@ -96,8 +96,10 @@ extern "C" {
 #endif
 
+#include <openssl/provider.h>
 #define EC_POINT_get_affine_coordinates_GFp     EC_POINT_get_affine_coordinates
 #define EC_POINT_set_affine_coordinates_GFp     EC_POINT_set_affine_coordinates
+#define FIPS_mode()                             OSSL_PROVIDER_available(NULL, "fips")
 #endif
 ```
Wil fix PR tommorrow.

dengert:
Forced pushed a single commit to sc-ossl-compat.h that matches what is in https://github.com/OpenSC/OpenSC/commit/1272521059eadf19e3b58b275efb4b41465804df

@xhanulik  if this gets in your way, we can wait until #2438. I can keep it as a local patch.

Jakuje (approved):

Issue #2456, opened 2021-11-26
github actions build-mingw and build-build-mingw32 are failing


dengert:
### Problem Description
github actions build-mingw and build-build-mingw32 started failing with several  PR checks. 

> The following packages have unmet dependencies:
>  wine32:i386 : Depends: libwine:i386 (= 5.0-3ubuntu1) but it is not going to be installed
> E: Unable to correct problems, you have held broken packages.
> Error: Process completed with exit code 100.
> The following packages have unmet dependencies:
>  wine32:i386 : Depends: libwine:i386 (= 5.0-3ubuntu1) but it is not going to be installed
> E: Unable to correct problems, you have held broken packages.
> Error: Process completed with exit code 100.

On my XUbuntu-20.4, synaptic has a dependencies "Breaks:wine32" which may be a clue.
(The OS was upgraded over the years.) 

Could the "held broken packages" be caused by doing upgrades of the actions? 

https://wiki.debian.org/Wine
lists a different set of packages to install. 

https://github.com/OpenSC/OpenSC/blob/master/.github/setup-linux.sh#L28

These might help in figuring out what to do:
https://askubuntu.com/questions/1204823/wine32i386-depends-libwinei386-3-0-1ubuntu1-but-it-is-not-going-to-be-i
which points at:
https://wiki.winehq.org/Ubuntu

### Proposed Resolution
none

Jakuje:
I tried to look into this last week, but without any success so far. I tried to run the same scripts in podman container and all worked fine (the same as it worked until last week in github actions) so I assume the github/azure mirror is somehow broken, providing broken dependencies, but I am not expert on debian packaging.

I also tried to get some more debug logs from apt, but it was not much helpful so far:

https://github.com/Jakuje/OpenSC/runs/4351244711?check_suite_focus=true
https://github.com/Jakuje/OpenSC/commit/d090ba6ed18ff63c85c822d1dc3cdf0e771f3525

Suggestions/patches/ideas how to solve this are welcomed.

Jakuje:
For the record, I tried a lot of stuff, but there is something badly broken in the github images.

It changed between https://github.com/OpenSC/OpenSC/runs/4316990114?check_suite_focus=true (good) and https://github.com/OpenSC/OpenSC/runs/4325865048?check_suite_focus=true (bad), which updated image:

    13 Version: 20211114.1                      |   13 Version: 20211122.1 

[...]

PR #2458, opened 2021-11-29
.github: Workaround broken Github image

Jakuje:
Fixes: #2456

Related: https://github.com/actions/virtual-environments/issues/4589

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/setup-linux.sh
+++ b/.github/setup-linux.sh
@@ -25,6 +25,12 @@ elif [ "$1" == "piv" -o "$1" == "isoapplet" -o "$1" == "gidsapplet" -o "$1" == "
 	fi
 	DEPS="$DEPS ant openjdk-8-jdk maven"
 elif [ "$1" == "mingw" -o "$1" == "mingw32" ]; then
+	# The Github's Ubuntu images since 20211122.1 are broken
+	# https://github.com/actions/virtual-environments/issues/4589
+	sudo apt install -y --allow-downgrades libpcre2-8-0=10.34-7
+
+	# Note, that this list is somehow magic and adding libwine, libwine:i386 or wine64
+	# will make the following sections break without any useful logs. See GH#2458
 	DEPS="$DEPS wine wine32 xvfb wget"
 	sudo dpkg --add-architecture i386
 	if [ "$1" == "mingw" ]; then
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,6 +6,9 @@ on:
       - '**.c'
       - '**.h'
       - .github/workflows/linux.yml
+      - .github/setup-linux.sh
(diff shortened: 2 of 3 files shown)


Jakuje:
For the record, this looks like it is not yet complete solution as it looks like the following step is failing silently
```
+ wine isetup-5.5.6.exe /SILENT /VERYSILENT /SP- /SUPPRESSMSGBOXES /NORESTART
```
as well as the fiollowing one, because the installation above probably failed:
```
+ wine 'C:/Program Files/Inno Setup 5/ISCC.exe' win32/OpenSC.iss
wine: cannot find 'C:/Program Files/Inno Setup 5/ISCC.exe
```
and then we are failing to upload artifacts
```
Warning: No files were found with the provided path: win32/Output/OpenSC*.exe. No artifacts will be uploaded.
```
and the `push-artifacts` task fails because there are no artifacts to pull from cache.

This happens for both 64b and 32b versions.

https://github.com/Jakuje/OpenSC/actions/runs/1517493522

Jakuje:
Not sure what `/SILENT /VERYSILENT` does but it looks like it makes the wine installation hang for no apparent reason.

@dengert if you have some idea what might be going on there, any tips would be appreciated. Hopefully we will get some working image in week or two from Github, but there still might some other issue.

Jakuje:
no idea ... the installation just fails. No warnings from wine are useful so I give up for today.

Jakuje:
lol. Now with 3b18674 where I added exactly the same package list as before, technically dropping `libwine libwine:i386 wine64` (against suggestion from the Debian docs), it looks like it works fine again. Let me wait for the jobs to finish and squash that.

Jakuje:
@dengert unless you will have some concerns about this, I will merge this version soon to get the green-tick again and to make the CI working in PRs again.

dengert:
No concerns. I was not so much interested in `mingw` but in not having the green ticks!

I had said my xubuntu-20.04 system could not find the `wine32` package. But I was using the synaptic "quick" search, which could  not find it. The full search and  `apt show wine32` found it. 

I think Github or Ubuntu fixed the problem. I suspect it was in the repositories or their order or caching.

Jakuje:
> No concerns. I was not so much interested in `mingw` but in not having the green ticks!
> 
> I had said my xubuntu-20.04 system could not find the `wine32` package. But I was using the synaptic "quick" search, which could not find it. The full search and `apt show wine32` found it.

I think you would need to add the 32 arch to find it.

> I think Github or Ubuntu fixed the problem. I suspect it was in the repositories or their order or caching.

Not yet. What we have here is a workaround that we will have to eventually remove, after they will fix it, but I do not suspect it will be earlier than in a week.

Issue #2461, opened 2021-12-09
[0.22.0] npa card not detected


tuxmaster5000:
### Problem Description
The npa card are supported, but not detected, 

### Steps to reproduce
1. insert the card to the reader
2. call opensc-tool -n

using the card with the "Ausweisapp2" will work. So the reader and the card are ok.

### Logs

https://pastebin.com/wENAiFi8

Jakuje:
The NPA card detection part is here:
```
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] card.c:341:sc_connect_card: trying driver 'npa'
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] dir.c:171:sc_enum_apps: called
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] card.c:844:sc_select_file: called; type=2, path=3f002f00
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] apdu.c:548:sc_transmit_apdu: called
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] card.c:473:sc_lock: called
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] reader-pcsc.c:685:pcsc_lock: called
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] card.c:513:sc_lock: returning with: 0 (Success)
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] apdu.c:515:sc_transmit: called
P:13372; T:0x140398909413184 16:44:49.124 [opensc-tool] apdu.c:363:sc_single_transmit: called
P:13372; T:0x140398909413184 16:44:49.125 [opensc-tool] apdu.c:367:sc_single_transmit: CLA:0, INS:A4, P1:8, P2:0, data(2) 0x7ffe0f1cc082
P:13372; T:0x140398909413184 16:44:49.125 [opensc-tool] reader-pcsc.c:323:pcsc_transmit: reader 'REINER SCT cyberJack RFID komfort (XXXXXX) 00 00'
P:13372; T:0x140398909413184 16:44:49.125 [opensc-tool] reader-pcsc.c:324:pcsc_transmit: 
Outgoing APDU (8 bytes):
00 A4 08 00 02 2F 00 00 ...../..
 
P:13372; T:0x140398909413184 16:44:49.125 [opensc-tool] reader-pcsc.c:242:pcsc_internal_transmit: called
[...]

PR #2463, opened 2021-12-10
Unbreak nPA detection

frankmorgner:
Fixes https://github.com/OpenSC/OpenSC/issues/2461

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-npa.c
+++ b/src/libopensc/card-npa.c
@@ -136,51 +136,33 @@ err:
 	return r;
 }
 
+unsigned char dir_content_ref[] = {
+	0x61, 0x32, 0x4F, 0x0F, 0xE8, 0x28, 0xBD, 0x08, 0x0F, 0xA0, 0x00, 0x00,
+	0x01, 0x67, 0x45, 0x53, 0x49, 0x47, 0x4E, 0x50, 0x0F, 0x43, 0x49, 0x41,
+	0x20, 0x7A, 0x75, 0x20, 0x44, 0x46, 0x2E, 0x65, 0x53, 0x69, 0x67, 0x6E,
+	0x51, 0x00, 0x73, 0x0C, 0x4F, 0x0A, 0xA0, 0x00, 0x00, 0x01, 0x67, 0x45,
+	0x53, 0x49, 0x47, 0x4E, 0x61, 0x09, 0x4F, 0x07, 0xA0, 0x00, 0x00, 0x02,
+	0x47, 0x10, 0x01, 0x61, 0x0B, 0x4F, 0x09, 0xE8, 0x07, 0x04, 0x00, 0x7F,
+	0x00, 0x07, 0x03, 0x02, 0x61, 0x0C, 0x4F, 0x0A, 0xA0, 0x00, 0x00, 0x01,
+	0x67, 0x45, 0x53, 0x49, 0x47, 0x4E,
+};
+
 static int npa_match_card(sc_card_t * card)
 {
-	int r = 0;
+	unsigned char dir_content[sizeof dir_content_ref];
+	unsigned char id[] = {0x2F, 0x00};
+	sc_apdu_t select_ef_dir;
 
-	if (SC_SUCCESS == sc_enum_apps(card)) {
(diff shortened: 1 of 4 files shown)


src/libopensc/card-npa.c
```diff
@@ -37,6 +37,8 @@
 
 static int fread_to_eof(const char *file, unsigned char **buf, size_t *buflen);
 #include "../tools/fread_to_eof.c"
+/* from dir.c */
+int parse_dir_record(sc_card_t *card, u8 ** buf, size_t *buflen, int rec_nr);
```
Jakuje, line 41: This should go to opensc.h

src/libopensc/card-npa.c
```diff
@@ -415,6 +399,18 @@ static int npa_init(sc_card_t * card)
 	if (r != SC_SUCCESS)
 		goto err;
 
+	size_t bufsize = sizeof dir_content_ref;
+	u8 *p = dir_content_ref;
+	while (bufsize > 0) {
+		if (card->app_count == SC_MAX_CARD_APPS) {
```
Jakuje, line 405: is this branch reachable when you removed the `sc_enum_apps()`? Or is this called from some general match/detect/init code?

Jakuje (changes requested):

src/libopensc/card-npa.c
frankmorgner: indeed.

but I think I could do better here. Since after this change, the reference data is hard coded, I could also hardcode the initialized card apps without parsing the buffer. I'l look into this later...

src/libopensc/card-npa.c
frankmorgner: not needed anymore

src/libopensc/card-npa.c
frankmorgner: done

src/libopensc/dir.c
```diff
@@ -60,7 +60,7 @@ static const struct sc_asn1_entry c_asn1_dir[] = {
 };
 
 
-static int
+int
 parse_dir_record(sc_card_t *card, u8 ** buf, size_t *buflen, int rec_nr)
```
Jakuje, line 64: if it is not needed anymore, please, keep this static.

frankmorgner: done

Jakuje (approved):
Do you want to squash the fixup commits before merging the changes or are we fine merging it as it is?

frankmorgner:
> Do you want to squash the fixup commits before merging the changes or are we fine merging it as it is?

Done. The only change I was not able to verify is whether or not https://github.com/OpenSC/OpenSC/pull/2463/commits/043a7b09450a5c894fc4820efd0621866076f6cd is the correct fix to the strict aliasing problem...

Jakuje:
> > Do you want to squash the fixup commits before merging the changes or are we fine merging it as it is?
> 
> Done. The only change I was not able to verify is whether or not [043a7b0](https://github.com/OpenSC/OpenSC/commit/043a7b09450a5c894fc4820efd0621866076f6cd) is the correct fix to the strict aliasing problem...

I think it is correct. These are very similar changes I did in 2fa6700599f358bc97d69ff923b7bde9304b7dc7 when I was tackling this issue in the rest of the code. But I do not build the openpace so I probably did not hit this issue.

Did you try with `-Wstrict-aliasing=2` in CFLAGS? I think this was enough to trigger the warning/error (added now into `--enable-strict` in configure.ac with 544dcc6827.

frankmorgner:
Sorry for the late feedback. No, I wasn't able to see the warnings for the unmodified code with gcc 10.3.0 and the above flags. Could we make this strict build part of the CI process?

Jakuje:
np. I was adding the `-Wstrict-aliasing=2` in the above mentioned commit, but the linux builds do not use openpace so that parts are not executed. It should be executed on OSX though. I see the flags used there, but I am not sure what gcc/clang/compiler is used there and whether it supports this warnings correctly.

The simplest would probably be to add one more target to linux with building also openpace (It is not in the current ubuntu version in GH).

Issue #2460, opened 2021-12-09
[0.22.0] egk-tool don't detect the eGK G2


tuxmaster5000:
### Problem Description
The G2 card of the eGK are not detected, 

### Steps to reproduce
1. insert the card to the reader
2. start egk-tool

### Logs
https://pastebin.com/mnnv4U7f

frankmorgner:
please try the following change:
```diff
diff --git a/src/libopensc/card-starcos.c b/src/libopensc/card-starcos.c
index 22ca49ba..f2c471e7 100644
--- a/src/libopensc/card-starcos.c
+++ b/src/libopensc/card-starcos.c
@@ -44,6 +44,7 @@ static const struct sc_atr_table starcos_atrs[] = {
        { "3B:DF:96:FF:81:31:FE:45:80:5B:44:45:2E:42:4E:4F:54:4B:31:30:30:81:05:A0", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_5, 0, NULL },
        { "3B:D9:96:FF:81:31:FE:45:80:31:B8:73:86:01:E0:81:05:22", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_5, 0, NULL },
        { "3B:D0:97:FF:81:B1:FE:45:1F:07:2B", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_4, 0, NULL },
+       { "3B:D0:96:FF:81:B1:FE:45:1F:07:2A", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_4, 0, NULL },
        { "3b:df:96:ff:81:31:fe:45:80:5b:44:45:2e:42:41:5f:53:43:33:35:32:81:05:b5", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_5, 0, NULL },
        { NULL, NULL, NULL, 0, 0, NULL }
 };
```

tuxmaster5000:
Now,
Health Care Application (HCA):
      --pd             Show 'Persönliche Versicherungsdaten' (XML)
                         (default=off)
      --vd             Show 'Allgemeine Versicherungsdaten' (XML)
                         (default=off)
      --gvd            Show 'Geschützte Versicherungsdaten' (XML)
                         (default=off)
      --vsd-status     Show 'Versichertenstammdaten-Status'  (default=off
will work. :)
Thanks

PR #2464, opened 2021-12-11
starcos: added one more ATR of eGK

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2460

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/card-starcos.c
```diff
@@ -44,6 +44,7 @@ static const struct sc_atr_table starcos_atrs[] = {
 	{ "3B:DF:96:FF:81:31:FE:45:80:5B:44:45:2E:42:4E:4F:54:4B:31:30:30:81:05:A0", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_5, 0, NULL },
 	{ "3B:D9:96:FF:81:31:FE:45:80:31:B8:73:86:01:E0:81:05:22", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_5, 0, NULL },
 	{ "3B:D0:97:FF:81:B1:FE:45:1F:07:2B", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_4, 0, NULL },
+	{ "3B:D0:96:FF:81:B1:FE:45:1F:07:2A", NULL, NULL, SC_CARD_TYPE_STARCOS_V3_4, 0, NULL },
```
Jakuje, line 47: Should we add this new ATR also to win32/customactions.cpp for minidriver?

Jakuje: @frankmorgner ?

frankmorgner: I don't think that's needed, because the cards don't currently have usable keys.

Jakuje: Thanks for clarification.

PR #2466, opened 2021-12-13
Build fixes

Pekkari:
The following patches provides fixes to warnings treated as errors spotted
on the process of creating an opensc package for buildroot, where the testing
tools builds the project against multiple toolchains.

### DIFF
--- a/src/libopensc/pkcs15-coolkey.c
+++ b/src/libopensc/pkcs15-coolkey.c
@@ -147,9 +147,9 @@ coolkey_find_matching_cert(sc_card_t *card, sc_cardctl_coolkey_object_t *in_obj,
 static int
 coolkey_get_attribute_ulong(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE type, CK_ULONG *value)
 {
-	const u8 *val;
-	size_t val_len;
-	u8 data_type;
+	const u8 *val = NULL;
+	size_t val_len = 0;
+	u8 data_type = 0;
 	int r;
 
 	r  = coolkey_get_attribute(card, obj, type, &val, &val_len, &data_type);
@@ -168,8 +168,8 @@ static int
 coolkey_get_attribute_boolean(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE attr_type)
 {
 	int r;
-	const u8 *val;
-	size_t val_len;
+	const u8 *val = NULL;
+	size_t val_len = 0;
 
 	r = coolkey_get_attribute(card, obj, attr_type, &val, &val_len, NULL);
 	if (r < 0) {
@@ -186,7 +186,7 @@ static int
 coolkey_get_attribute_bytes(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE type, u8 *data, size_t *data_len, size_t max_data_len)
 {
 	const u8 *val;
-	size_t val_len;
+	size_t val_len = 0;
 	int r;
 
 	r = coolkey_get_attribute(card, obj, type, &val, &val_len, NULL);
--- a/src/pkcs15init/pkcs15-asepcos.c
+++ b/src/pkcs15init/pkcs15-asepcos.c
@@ -221,7 +221,7 @@ static int asepcos_do_store_pin(sc_profile_t *profile, sc_card_t *card,
 {
 	sc_file_t *nfile = NULL;
 	u8  buf[64], sbuf[64], *p = buf, *q = sbuf;
-	int r, akn;
+	int r, akn = 0;
 
 	if (auth_info->auth_type != SC_PKCS15_PIN_AUTH_TYPE_PIN)
 		return SC_ERROR_OBJECT_NOT_VALID;
--- a/src/tools/opensc-explorer.c
+++ b/src/tools/opensc-explorer.c
@@ -2472,7 +2472,7 @@ int main(int argc, char *argv[])
 		char *line;
 		int cargc;
 		char *cargv[260];
-		int multiple;
+		int multiple = 0;
 		struct command *cmd;
 		char prompt[3*SC_MAX_PATH_STRING_SIZE];
 


frankmorgner (changes requested):
You have multiple memory leaks in your code.

What build environment are you using. At least the Windows stuff looks wrong.

Pekkari:
I'm using the test-pkg tool on buildroot to check the consistency of the package,
though I didn't run the code afterwards in valgrind to check what you say, since I
have not all the platforms available on my end. Any improvement is appreciated.

frankmorgner:
can we test this somehow? ideally, you propose an CI integration. we're using github actions and docker atm, for example...

Pekkari:
It is a bit of a chiken and egg problem. Buildroot wants me to upstream the patches before the
package would be accesible to you, you want me to give a testing case that doesn't come easy
without upstreaming the package. If the package were upstreamed, running:
```
./utils/test-pkg -c opensc.cfg -p opensc
```
with opensc as:
```
$ cat opensc.cfg 
BR2_PACKAGE_OPENSC=y
```
On a buildroot checkout, would do it.

src/libopensc/pkcs15-coolkey.c
```diff
@@ -147,9 +147,9 @@ coolkey_find_matching_cert(sc_card_t *card, sc_cardctl_coolkey_object_t *in_obj,
 static int
 coolkey_get_attribute_ulong(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE type, CK_ULONG *value)
 {
-	const u8 *val;
-	size_t val_len;
-	u8 data_type;
+	const u8 *val = malloc(sizeof(u8));
```
Jakuje, line 150: please, initialize the value here with NULL. Otherwise you are leaking the memory. The `coolkey_get_attribute()` overrides whatever you pass in as the `val` attribute.

src/libopensc/pkcs15-coolkey.c
```diff
@@ -168,8 +168,8 @@ static int
 coolkey_get_attribute_boolean(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE attr_type)
 {
 	int r;
-	const u8 *val;
-	size_t val_len;
+	const u8 *val = malloc(sizeof(u8));
```
Jakuje, line 171: Same here

src/libopensc/pkcs15-coolkey.c
```diff
@@ -186,7 +186,7 @@ static int
 coolkey_get_attribute_bytes(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_ATTRIBUTE_TYPE type, u8 *data, size_t *data_len, size_t max_data_len)
 {
 	const u8 *val;
```
Jakuje, line 188: this case does not ring the "maybe uninitialized errors" for you?

src/libopensc/card-gpk.c
```diff
@@ -242,7 +243,7 @@ match_path(sc_card_t *card, unsigned short int **pathptr, size_t *pathlen,
 
 	/* Make sure path starts with MF.
 	 * Note the cached path should always begin with MF. */
-	if (ptr[0] != GPK_FID_MF || curptr[0] != GPK_FID_MF)
+	if (ptr[0] != (GPK_FID_MF >> 8) || curptr[0] != (GPK_FID_MF >> 8))
```
Jakuje, line 246: I think this change requires some more description what is changed and why and why it fixes your issue.

Jakuje (changes requested):

Jakuje:
I also approved the CI run for you, which immediately found the memory leaks through CIFuzz:

https://github.com/OpenSC/OpenSC/runs/4508348631?check_suite_focus=true

Please, review the report, comments and fix them.

src/libopensc/pkcs15-coolkey.c
Pekkari: ack

src/libopensc/pkcs15-coolkey.c
Pekkari: ack

src/libopensc/pkcs15-coolkey.c
Pekkari: No, sorry, this doesn't.

src/libopensc/card-gids.c
```diff
@@ -691,7 +691,8 @@ static int gids_finish(sc_card_t *card)
 // the cmap file is used to detect the key algorithm / size
 static int gids_get_crypto_identifier_from_key_ref(sc_card_t *card, const unsigned char keyref, unsigned char *cryptoidentifier) {
 	struct gids_private_data *data = (struct gids_private_data *) card->drv_data;
-	PCONTAINER_MAP_RECORD records = (PCONTAINER_MAP_RECORD) data->cmapfile;
+	PCONTAINER_MAP_RECORD records = malloc(sizeof(CONTAINER_MAP_RECORD));
```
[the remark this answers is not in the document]
Pekkari: this is the error:
```
card-gids.c:694:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  PCONTAINER_MAP_RECORD records = (PCONTAINER_MAP_RECORD) data->cmapfile;
  ^
```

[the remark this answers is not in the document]
Jakuje: The fix is not correct as the `data->cmapfile` can contain multiple `CONTAINER_MAP_RECORD` records if I read right, but you malloc & copy just one.

But right now, I do not see a simple way how to fix this. I read this assignment from `u8` ("alias" to `unsigned char` in types.h) `cmapfile` to `PCONTAINER_MAP_RECORD` valid. The other option I see suggested would be use of union to reference same data with different types, but I think it would require static size.

Pekkari:
I added a buildroot repository under my account where the patch for the opensc
package is applied. To test the toolchains, just clone https://github.com/Pekkari/builroot
and follow [this](https://github.com/OpenSC/OpenSC/pull/2466#issuecomment-992344482).

If you want to see the errors in the toolchains, move the patches under pakcages/opensc
to some other place, and rerun test-pkg.

Pekkari:
I have removed the buggy patch from the PR and fixed the first patch that was more
or less clear. I'll open a new PR when I have some solution for the strict aliasing problems.

src/libopensc/card-gids.c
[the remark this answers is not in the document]
Pekkari: Ack, any hints how to solve this appreciated. I'll push this patch to another PR though, so that
we can fix the maybe uninitialized warnings treated as errors.

frankmorgner (approved):
Thank you

Jakuje (approved):

PR #2467, opened 2021-12-13
Add support for SHA3* in pkcs11-tool.

AlexandreGonzalo:
Hi OpenSC team,
It's been a long time since my last PR.
Last year, I added the support for many algorithms with SHA3 in our Trustonic TEE PKCS#11 HSM.
For my own testing, I modified the pkcs11-tool and libp11.
In this PR, I am pushing all the changes for the pkcs11-tool.
Is it something that you want?
Should I also start the PR for the libp11?
Best Regards,
Alexandre.


Jakuje (approved):

Jakuje:
Looks good. Indeed, the SHA3 support is a useful as it is finally standardized. The smart card adoption of this algorithm did not get far yet though, but the pkcs11-tool should be considered more generic so having this in makes sense. The same for libp11.

I think the test failure is just flaky so lets ignore it for now.

AlexandreGonzalo:
> Looks good. Indeed, the SHA3 support is a useful as it is finally standardized. The smart card adoption of this algorithm did not get far yet though, but the pkcs11-tool should be considered more generic so having this in makes sense. The same for libp11.
> 
We don't know yet when our customers will start to use the SHA3 algorithms, it may start sooner than expected.
I will prepare the PR for libp11 tomorrow.
> I think the test failure is just flaky so lets ignore it for now.
This test passed for my second PR which includes the same patch for SHA3. I don't think that this failures is linked to my changes.
Thanks for the review.

frankmorgner (approved):

PR #2468, opened 2021-12-13
Trustonic pkcs11 fix build

AlexandreGonzalo:
Hi OpenSC team,
In my build environment, I don't always have pthread, thus sometimes, I don't have HAVE_PTHREAD defined.
I had a build issue.
Regards,
Alexandre.


src/tools/pkcs11-tool.c
```diff
@@ -263,8 +263,8 @@ static const struct option options[] = {
 #ifndef _WIN32
 	{ "test-fork",		0, NULL,		OPT_TEST_FORK },
 #endif
-	{ "use-locking",	0, NULL,		OPT_USE_LOCKING },
 #if defined(_WIN32) || defined(HAVE_PTHREAD)
+	{ "use-locking",	0, NULL,		OPT_USE_LOCKING },
```
Jakuje, line 267: I think this should stay as it is, but the check should come inside of the handling of this operation to provide better error than "unknown option".

Jakuje (approved):

src/tools/pkcs11-tool.c
AlexandreGonzalo: The issue is that OPT_USE_LOCKING is not defined when HAVE_PTHREAD is not defined.

Jakuje: Right. It would need some more modifications. So lets keep it this way.

Jakuje:
can you rebase this on top of the current master so I can cleanly merge the outstanding patch?

AlexandreGonzalo:
> can you rebase this on top of the current master so I can cleanly merge the outstanding patch?

Sure, it is done.

PR #2473, opened 2021-12-23
Sym crypt hw

popovec:
symmetric encryption

New code:
pkcs#11: C_EncryptInit, C_Encrypt, C_EncryptUpdate, C_EncryptFinal
pkcs#15: pkcs15_skey_encrypt, sc_pkcs15_encrypt_sym
also implements: sc_encrypt_sym

MyEID driver: support for symmetric crypt

Tested:  MyEID 4.0.1, 4.5.5 card, pkcs11-tool --encrypt, 128 bit AES key (small file encryption by C_encrypt, and big file by C_EncryptUpdate and C_EncrytpFinal), all supported mechanisms: AES-ECB, AEC-CBC, AES-CBC-PAD.

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -84,6 +84,13 @@ typedef struct myeid_private_data {
 	 after this pair of calls and must not be used elsewhere. */
 	const struct sc_security_env* sec_env;
 	int disable_hw_pkcs1_padding;
+	/* buffers for AES(DES) block cipher */
+	uint8_t sym_crypt_buffer[16];
+	uint8_t sym_plain_buffer[16];
+	uint8_t sym_crypt_buffer_len;
+	uint8_t sym_plain_buffer_len;
+	/* PSO for AES/DES need algo+flags from sec env */
+	unsigned int algorithm, algorithm_flags;
 } myeid_private_data_t;
 
 typedef struct myeid_card_caps {
@@ -713,6 +720,8 @@ static int myeid_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 	LOG_FUNC_RETURN(card->ctx, iso_ops->pin_cmd(card, data, tries_left));
 }
 
+#define IS_SYMETRIC_CRYPT(x) ((x) == SC_SEC_OPERATION_ENCRYPT_SYM || (x) == SC_SEC_OPERATION_DECRYPT_SYM)
+
 static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *env,
 		int se_num)
 {
@@ -756,6 +765,14 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
 		apdu.p1 = 0x81;
 		apdu.p2 = 0xB8;
 		break;
+	case SC_SEC_OPERATION_ENCRYPT_SYM:
+		apdu.p1 = 0x81;
+		apdu.p2 = 0xB8;
+		break;
+	case SC_SEC_OPERATION_DECRYPT_SYM:
+		apdu.p1 = 0x41;
+		apdu.p2 = 0xB8;
+		break;
 	default:
 		return SC_ERROR_INVALID_ARGUMENTS;
 	}
@@ -774,9 +791,16 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
 		memcpy(p, env->file_ref.value, 2);
 		p += 2;
 	}
+	/* symmetric operations: we need to set the key reference */
+	if (IS_SYMETRIC_CRYPT(env->operation)) {
+		*p++ = 0x83;
+		*p++ = 1;
+		*p++ = 0;
+	}
 	if (env->flags & SC_SEC_ENV_KEY_REF_PRESENT && env->operation != SC_SEC_OPERATION_UNWRAP &&
-		env->operation != SC_SEC_OPERATION_WRAP)
-	{
+			env->operation != SC_SEC_OPERATION_WRAP &&
+			env->operation != SC_SEC_OPERATION_ENCRYPT_SYM &&
+			env->operation != SC_SEC_OPERATION_DECRYPT_SYM) {
 		*p++ = 0x84;
 		*p++ = 1;
 		*p++ = 0;
@@ -795,11 +819,13 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
 			break;
 	    }
 
-	if (env->operation ==  SC_SEC_OPERATION_UNWRAP || env->operation == SC_SEC_OPERATION_WRAP)
-	{
-	    /* add IV if present */
+	r = 0;
+	if (env->operation == SC_SEC_OPERATION_UNWRAP || env->operation == SC_SEC_OPERATION_WRAP ||
+			IS_SYMETRIC_CRYPT(env->operation)) {
+		/* add IV if present */
 		for (i = 0; i < SC_SEC_ENV_MAX_PARAMS; i++)
 			if (env->params[i].param_type == SC_SEC_ENV_PARAM_IV) {
+				r = 1;
 				*p++ = 0x87;
 				*p++ = (unsigned char) env->params[i].value_len;
 				if (p + env->params[i].value_len >= sbuf + SC_MAX_APDU_BUFFER_SIZE) {
@@ -811,6 +837,15 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
 				break;
 			}
 	}
+	/* for AES_ECB we need to reset the IV but we respect if the IV is already present */
+	if (IS_SYMETRIC_CRYPT(env->operation) && env->algorithm == SC_ALGORITHM_AES &&
+			env->algorithm_flags == SC_ALGORITHM_AES_ECB && r == 0) {
+		*p++ = 0x87;
+		*p++ = 16;
+		memset(p, 0, 16);
+		p += 16;
+	}
+
 	r = p - sbuf;
 	apdu.lc = r;
 	apdu.datalen = r;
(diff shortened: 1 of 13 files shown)


src/libopensc/card-myeid.c
```diff
@@ -757,8 +772,17 @@ static int myeid_set_security_env_rsa(sc_card_t *card, const sc_security_env_t *
 		memcpy(p, env->file_ref.value, 2);
 		p += 2;
 	}
+	/* symmetric operations: we need set the key reference */
```
Jakuje, line 775:
```suggestion
	/* symmetric operations: we need to set the key reference */
```

src/libopensc/card-myeid.c
```diff
@@ -1799,6 +1848,243 @@ static int myeid_finish(sc_card_t * card)
...
+		memcpy(out, apdu.resp, apdu.resplen);
+		out += apdu.resplen;
+		return_len += apdu.resplen;
+	}
+	if (decipher) {
+		/* last block is stored in buffer and is returned to application
+		 * in next call to C_DecryptUpdate or C_DecryptFinal. This allow us
+	         * to compute how many bytes is to be returned after padding removal.
+		 * Whole handling of this is here, because "data" and "out" buffer
+	         * can be in the same place.
```
Jakuje, line 2049: there is some indentation problem

src/pkcs11/mechanism.c
```diff
@@ -1134,6 +1259,151 @@ sc_pkcs11_deri(struct sc_pkcs11_session *session,
...
 
 
+/*
+ * Initialize a encrypt operation
+ */
+static CK_RV
+sc_pkcs11_encrypt_init(sc_pkcs11_operation_t *operation,
+			struct sc_pkcs11_object *key)
+{
+	struct signature_data *data;
```
Jakuje, line 1269: The usage of `signature_data` is confusing here. If I see right, only the `key` part is used. Would it make sense to assign the key directly into the `operation->priv_data`? Or create generic `operation_data` if we envision there will be some more data needed to share between the functions?

popovec: We already have same confusing name in `sc_pkcs11_decrypt_init()`. It doesn't make sense to fix this in this PR just because of the new operation `sc_pkcs11_encrypt_init ()`, we have to fix it globally (so also for the function `sc_pkcs11_decrypt_init ()`) 

`````
static CK_RV
sc_pkcs11_decrypt_init(sc_pkcs11_operation_t *operation,
                        struct sc_pkcs11_object *key)
{
        struct signature_data *data;
        CK_RV rv;

        if (!(data = new_signature_data()))
                return CKR_HOST_MEMORY;
`````
> Would it make sense to assign the key directly into the operation->priv_data?

This solution complicates the freeing of memory (in `sc_pkcs11_new_fw_mechanism ()` we use `mt-> release = sc_pkcs11_signature_release;`)

It might be a good idea to rename `struct signature_data` to` operation_data` and similarly rename functions that operate with it: `new_signature_data()`, `signature_data_release()`, `signature_data_buffer_append()` and of course `sc_pkcs11_signature_release()`.  

The question is whether this renaming will not impair the readability of the code in sign and verify operations. I believe that this renaming should be done outside this PR.

File: src/libopensc/card-myeid.c [hunk 4]
```diff
@@ -965,6 +1007,13 @@ static int myeid_set_security_env(struct sc_card *card,
...
+			/* Tag 0x80 algorithm_ref - value 0x80 or 0x8A is working only for UNWRAP/WRAP
+			 * AES is supported from version 4.0 but without pkcs#7 padding.
+			 * For SC_SEC_OPERATION_ENCRYPT_SYM and SC_SEC_OPERATION_DECRYPT_SYM we running
+			 * PKCS#7 in software, here we fix the algorithm_ref variable.
+			 */
+			if (tmp.operation == SC_SEC_OPERATION_ENCRYPT_SYM || tmp.operation == SC_SEC_OPERATION_DECRYPT_SYM)
+				tmp.algorithm_ref &= ~0x80;		/* do not handle padding in card */
 
 			/* from this point, there's no difference to RSA SE */
 			return myeid_set_security_env_rsa(card, &tmp, se_num);
```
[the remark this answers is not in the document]
popovec: `myeid_set_security_env_rsa ()` is already used for wrap/unwrap operation with AES key:
`````
c217b254f       (Hannu Honkanen 2017-11-17 16:10:50 +0200       1018)                   /* from this point, there's no difference to RSA SE */
550d4eb03       (Hannu Honkanen 2018-04-06 14:15:10 +0300       1019)                   return myeid_set_security_env_rsa(card, &tmp, se_num);
`````

If this renaming is necessary, I would like to leave it to the developers from Aventra.

**Comment by Jakuje** [comment]
Then it looks ok to me, even though I did not go into detail of the operations.

One more thing, can you rebase on top of current master to get also CI results with OpenSSL 3.0?

In any case, I would like also somebody else to go through this before merging.

**Comment by popovec** [comment]
Rebased, CI tests with openssl 3.0 and libressl are OK. 

This PR is designed so as not to interfere with current asymmetric operations in OpenSC. Code for symmetric decryption may affect asymmetric decipher operation and therefore decipher code is planned to be submitted as separate PR .  Preliminary code for symmetric decipher  is available here: https://github.com/popovec/OpenSC/tree/sym_crypt_hw_test 

Some details about implementation:

Symmetric encrypt allow us to encrypt large amount of data, there is no way to accumulate all data in buffer and then encrypt them. The sc * functions must allow us to initialize encryption and then allow us to repetitive call (to update data) from `C_EncryptUpdate` and  get last part of result - `C_EncryptFinal`.

However, only one version of `sc_encrypt_sym ()` is enough, because each of the calls C_EncryptInit, C_EncryptUpdate and `C_EncryptFinal` sets the parameters in `sc_encrypt_sym ()` differently (similarly in `sc_pkcs15_encrypt_sym ()`). 

C_EncryptInit  is mapped to sc_pkcs15_encrypt_sym (...., NULL, 0, NULL,NULL)
C_EncryptUpdate  is mapped to sc_pkcs15_encrypt_sym (...., data, len, encdata, &enclen)
C_EncryptFinal  is mapped to sc_pkcs15_encrypt_sym (...., NULL, 0, encdata, &enclen)
C_Encrypt is constructed from C_EncryptUpdate and C_EncryptFinal.

If the above conversion is unsatisfactory for some reason, I would be happy if someone suggests another way to map the pkcs11 function to the pkcs15 interface.

File: src/pkcs11/mechanism.c (hunk 3 above)
Comment (frankmorgner) [reply src/pkcs11/mechanism.c:1269]: could you create a PR with the suggested fix to the naming?

**Review by frankmorgner — APPROVED** [review]

**Review by Jakuje — APPROVED** [review]
thank you! Look good

File: src/libopensc/card-myeid.c [hunk 5]
```diff
@@ -1799,6 +1848,243 @@ static int myeid_finish(sc_card_t * card)
...
+			LOG_FUNC_RETURN(ctx, SC_ERROR_WRONG_LENGTH);
+		memcpy(out, apdu.resp, apdu.resplen);
+		out += apdu.resplen;
+		return_len += apdu.resplen;
+	}
+	/* last block is stored in buffer and is returned to application
+	 * in next call to C_DecryptUpdate or C_DecryptFinal. This allow us
+         * to compute how many bytes is to be returned after padding removal.
+	 * Whole handling of this is here, because "data" and "out" buffer
+         * can be in the same place.
```
Jakuje, line 2048: indent. Also two lines above

Jakuje (commented):
I just went through the code and noticed only couple of formatting issues, which I would like to tackle before merging. You can use as a help the clang-format from #2017, but do not take every suggestion as a rule as we did not finalize the formatting rules yet.

popovec:
Rebased, formating fixed to match definitions in `.clang-format` from https://github.com/OpenSC/OpenSC/pull/2017 except reformatting of large structures in `src/pkcs11/openssl.c`, `src/pkcs11/framework-pkcs15.c`,`src/libopensc/iso7816.c` and `src/pkcs11/pkcs11-object.c`. (Is it appropriate to add another patch that reformats these structures as well?)

> this condition is repeated at least twice throughout this function. Would it make sense to make some helper condition or macro such as is_wrap_or_encrypt(env->operation) to make the code more readable?

This is fixed by macro `IS_SYMETRIC_CRYPT(x)` (I don't want to mix wrap/unwrap operations and symmetric encryption/decryption operations in one function/macro).

Jakuje:
@frankmorgner do you have some more comments regarding this or can I merge this for the 0.23.0 release?

popovec:
This patch introduced two defects reported by "coverity scan". I'm not sure if these are false negative reports or if they are real bugs. I will look into both reported issues.
````
*** CID 380538:  Memory - illegal accesses  (OVERRUN)
/src/libopensc/card-myeid.c: 1971 in myeid_enc_dec_sym()
1965     
1966                                    sc_log(ctx, "Found padding byte %02x", pad_byte);
1967                                    if (pad_byte == 0 || pad_byte > block_size)
1968                                            LOG_FUNC_RETURN(ctx, SC_ERROR_WRONG_PADDING);
1969                                    sdata = priv->sym_plain_buffer + block_size - pad_byte;
1970                                    for (i = 0; i < pad_byte; i++)
>>>     CID 380538:  Memory - illegal accesses  (OVERRUN)
>>>     Overrunning array of 16 bytes at byte offset 30 by dereferencing pointer "sdata + i".
1971                                            if (sdata[i] != pad_byte)
1972                                                    LOG_FUNC_RETURN(ctx, SC_ERROR_WRONG_PADDING);
1973                                    return_len = block_size - pad_byte;
1974                            }
1975                            *outlen = return_len;
1976                            if (return_len > *outlen)
````
````*** CID 380537:  Null pointer dereferences  (FORWARD_NULL)
/src/pkcs11/framework-pkcs15.c: 5586 in pkcs15_skey_encrypt()
5580            if (rv < 0)
5581                    return sc_to_cryptoki_error(rv, "C_Encrypt...");
5582     
5583            /* pointer CK_ULONG_PTR to size_t conversion */
5584            lpEncryptedDataLen = pulEncryptedDataLen ? &lEncryptedDataLen : NULL;
5585     
>>>     CID 380537:  Null pointer dereferences  (FORWARD_NULL)
>>>     Dereferencing null pointer "skey".
5586            rv = sc_pkcs15_encrypt_sym(fw_data->p15_card, skey->prv_p15obj, flags,
5587                            pData, ulDataLen, pEncryptedData, lpEncryptedDataLen,
5588                            pMechanism->pParameter, pMechanism->ulParameterLen);
5589     
5590            if (pulEncryptedDataLen)
5591                    *pulEncryptedDataLen = *lpEncryptedDataLen;
````

Issue #2465, opened 2021-12-12
Inconsistent implementation and lack of documentation of key ID parameter parsing of pkcs11-tool


egassemcinatas:
The hex_to_bin() function in pkcs11-tool.c contains an inconsistency:  The
parsing tolerates colon separators, but the check for the number of hex digits
at the start of the function does not account for them.  As a consequence,
using the --id option of pkcs11-tool with an ID with an even number of bytes
separated by colons will result in a failed operation.

To reproduce, try for example a decryption operation with a key ID with an even
number of bytes given with colon separators:

    pkcs11-tool  -l --decrypt --id 12:34:56:78 -i infile -o outfile

Or run this test program showing the input and output of hex_to_bin():
https://volkerschatz.com/tmp/hex2bintest.c

    in: 12:34:56:78
    out: 01 02 34 56 78
    in: 1:234:56
    out: 01 23 04 56

The test program also demonstrates another issue, that hex_to_bin() tolerates
more than two hex digits between colons and produces a result that is probably
not expected by the user.

What type of input to accept is a design decision on you part.  But whatever
you choose, please document the valid parameter format for the -e and -d
options (which use hex_to_bin) in the pkcs11-tool manual page.

dengert:
Looks like a bug. 
```
in: 12:34:56:78
out: 01 02 34 56 78
```
is definitely incorrect. 

OpenSC has a `sc_hex_to_bin`  and `pkcs11-tool.c` also has a `hex_to_bin` and pkcs11-tool.c uses both.
Other tools use the sc_hex_to_bin, that handles the ":" correctly. 

looks like the `hex_to_bin` should be removed, and `sc_hex_to_bin` used in its place.

popovec:
sc_hex_to_bin() output:
`````
in: 12:34:56:78
out: 12 34 56 78 
in: 1:234:56
out: 12 34 56
`````

I am not sure if second input is valid and if we should report a warning/error in this case as well.

dengert:
If the intent was ":" (or spaces?)  separates strings of hex digits which may have a missing leading "0" , then:
"in: 1:234:56" 
should be 
"out: 01 02 34 56"?

[code](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/sc.c#L102-L108) has special case for a single hex digit.  It is assumed to have a leading "0". (What ever we do we should make sure this still works.) 

[...]

PR #2476, opened 2021-12-30
The hex_to_bin() is replaced by sc_hex_to_bin()

popovec:
It is not advisable to have two functions that perform the same operation.
The hex_to_bin() is used only in pkcs11-tool, but some cases generates wrong
output. This function is removed in favor of the sc_hex_to_bin() function.

Fixes: #2465

### DIFF
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -82,8 +82,13 @@ int sc_hex_to_bin(const char *in, u8 *out, size_t *outlen)
 		else if ('A' <= c && c <= 'F')
 			nibble = c - 'A' + 10;
 		else {
-			if (strchr(sc_hex_to_bin_separators, (int) c))
+			if (strchr(sc_hex_to_bin_separators, (int) c)) {
+				if (byte_needs_nibble) {
+					r = SC_ERROR_INVALID_ARGUMENTS;
+					goto err;
+				}
 				continue;
+			}
 			r = SC_ERROR_INVALID_ARGUMENTS;
 			goto err;
 		}
--- a/src/tests/unittests/hextobin.c
+++ b/src/tests/unittests/hextobin.c
@@ -1,0 +1,125 @@
+/*
+ * hextobin.c: Test suite for sc_hex_to_bin()
+ *
+ * Copyright (C) 2022 Peter Popovec <popovec.peter@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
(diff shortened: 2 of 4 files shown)


frankmorgner:
Thank you!

Would it be possible to write a test case that verifies fixing the parsing issue?

src/tests/unittests/hextobin.c
```diff
@@ -0,0 +1,123 @@
...
+#include <string.h>
+#include "libopensc/opensc.h"
+
+#define LEN 15
+
+int main()
+{
+	char *t;
+	char test[][LEN] = {
+/* result length (0x60 error string, 0x70 end of array), string, result string */
```
Jakuje, line 31: I would suggest to use some struct instead of two-dimensional array with comment to describe its functionality. Something like this would make both the structure and the code more readable: 
```
struct {
	int result_len;
	const char *input;
	const char *output;
} test[] = {
	1, "0", "\x00",
	1, ":0", "\x00",
	...
};
```
(or even a bit more verbose with separating results and expected output length with proper types)

src/tests/unittests/hextobin.c
```diff
@@ -0,0 +1,123 @@
...
+
+		{1}, "10", "\x10",
+		{1}, " 90", "\x90",
+		{1}, ":a0", "\xa0",
+		{1}, " :B0 :", "\xb0",
+
+		{1}, " 11:", "\x11",
+		{1}, " :11:", "\x11",
+		{1}, " :1 1:", "\x11",
+		{1}, "1:1", "\x11",
```
Jakuje, line 60: I am wondering if this is the expected output. I would expect `\x01\x01`, but it might be a matter of convention/taste/mood.

src/tests/unittests/hextobin.c
```diff
@@ -0,0 +1,123 @@
...
+		{0}, ":", "",
+		{0}, " ", "",
+		{0}, "", "",
+
+		/* incorrect strings */
+		{0x60}, "111", "",
+		{0x60}, "b:211 :2", "",
+		{0x60}, "G", "",
+		{0x60}, " z", "",
+		{0x60}, "1:11", "",
```
Jakuje, line 85: I am wondering whats on `1:11` more wrong than on `1:234:56`, which is supposed to be a valid input.

Jakuje:
Please, see @dengert [comment in #2465](https://github.com/OpenSC/OpenSC/issues/2465#issuecomment-1003041806). I think this function needs a bit more love. There are obviously some odds in the parsing and some expectations, which might not be completely clear for the users and which cause issues such as #2465. The least we should have a comments in the unit tests saying why the particular questionable corner cases are "correct" or "wrong" so next time somebody will question some decisions or change the behavior, we will have some reference.

src/libopensc/sc.c
```diff
@@ -72,6 +72,13 @@ int sc_hex_to_bin(const char *in, u8 *out, size_t *outlen)
 	int r = SC_SUCCESS;
 	size_t left = *outlen;
 	u8 byte = 0;
+	/* skip initial separators */
```
frankmorgner, line 75: Initial seperators should be covered in the loop. In the first run byte_needs_nibble is 0 so that the loop will be continued in line 97, won't it?

frankmorgner (approved):
Great work, thanks for diving in so deep :wink:

Jakuje (approved):

PR #2481, opened 2022-01-08
error code mapping: SC_ERROR_WRONG_PADDING -> CKR_ENCRYPTED_DATA_INVALID

popovec:
modified:   src/pkcs11/misc.c

### DIFF
--- a/src/pkcs11/misc.c
+++ b/src/pkcs11/misc.c
@@ -115,6 +115,8 @@ static CK_RV sc_to_cryptoki_error_common(int rc)
 		return CKR_DEVICE_MEMORY;
 	case SC_ERROR_MEMORY_FAILURE:	/* EEPROM has failed */
 		return CKR_DEVICE_ERROR;
+	case SC_ERROR_WRONG_PADDING:
+		return CKR_ENCRYPTED_DATA_INVALID;
 	}
 	return CKR_GENERAL_ERROR;
 }


frankmorgner:
`strip_*_padding()` is the only place where SC_ERROR_WRONG_PADDING is returned. The padding functions are called after data was deciphered or signed by the card (and the padding could't be removed). I'm not sure if CKR_ENCRYPTED_DATA_INVALID is the best translation. What is the problem you're trying to fix here?

popovec:
If an error occurs while removing the padding, C Decrypt currently returns CKR_GENERAL_ERROR, which I don't think is the best error code. According to PKCS # 11, one of the possible errors is CKR_ENCRYPTED_DATA_INVALID, which best answers the problem (wrong padding).
 
There is one more option - CKR_ENCRYPTED_DATA_LEN_RANGE - it can also match the padding removal error. 

But .. from PKCS#11:

·         CKR_ENCRYPTED_DATA_INVALID: The encrypted input to a decryption operation has been determined to be invalid ciphertext.  This return value has lower priority than CKR_ENCRYPTED_DATA_LEN_RANGE.

·         CKR_ENCRYPTED_DATA_LEN_RANGE: The ciphertext input to a decryption operation has been determined to be invalid ciphertext solely on the basis of its length.  Depending on the operation’s mechanism, this could mean that the ciphertext is too short, too long, or is not a multiple of some particular block size.  This return value has higher priority than CKR_ENCRYPTED_DATA_INVALID.
 
It is possible that in the future, when symmetric encryption is implemented in OpenSC (PR is already available), we will have to correct this return value, but for asymmetric encryption it is best to report CKR_ENCRYPTED_DATA_INVALID in case of a padding error. 

Discussion of this PR is welcome, the goal is to meet the PKCS#11 standard as best as possible.

frankmorgner:
On second view, a possible error of `strip_*_padding()` on the input when creating a signature is not returned to the (PKCS#11) application but instead handled in the respective card drivers (cardos, incrypto). This leaves deciphering as only use case. So I guess that your change is OK...

Jakuje (approved):

frankmorgner (approved):

PR #2484, opened 2022-01-12
Oaep label

popovec:
(pkcs11-tool -l -t, MyEID card)

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6507,7 +6507,8 @@ static int test_unwrap(CK_SESSION_HANDLE sess)
 #ifdef ENABLE_OPENSSL
 static int encrypt_decrypt(CK_SESSION_HANDLE session,
 		CK_MECHANISM_TYPE mech_type,
-		CK_OBJECT_HANDLE privKeyObject)
+		CK_OBJECT_HANDLE privKeyObject,
+		char *param, unsigned long param_len)
 {
 	EVP_PKEY       *pkey;
 	unsigned char	orig_data[512];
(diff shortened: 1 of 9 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -6716,6 +6717,18 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 			printf("set mgf1 md failed, returning\n");
 			return 0;
 		}
+		if (param_len != 0 && param != NULL) {
+			/* label is ownership of openssl, do not free this ptr! */
```
Jakuje, line 6721:
```suggestion
			/* label is in ownership of openssl, do not free this ptr! */
```

src/tools/pkcs11-tool.c
```diff
@@ -6741,21 +6754,28 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
 		oaep_params.pSourceData = NULL; /* PKCS#11 standard: this must be NULLPTR */
 		oaep_params.ulSourceDataLen = 0; /* PKCS#11 standard: this must be 0 */
 
-		/* If an RSA-OAEP mechanism, it needs parameters */
+		fprintf(stderr, "OAEP parameters: hashAlg=%s, mgf=%s, ",
+			p11_mechanism_to_name(oaep_params.hashAlg),
+			p11_mgf_to_name(oaep_params.mgf));
+
+		if (param != NULL && param_len > 0) {
+			oaep_params.source = 1UL;
```
Jakuje, line 6766: I think we should use in this case the pkcs11 constant `CKZ_DATA_SPECIFIED` instead of the magic number 1. Similarly as with the other places where the `source` is used.

Jakuje:
Do you know if this is used in some real-world application/protocol outside of the tests right now? Is there some other pkcs11 module implementing this?

The implementation looks quite simple and straight-forward.

popovec:
I don't know any other pkcs#11 module with OAEP support which also includes support for "label".

I tested the use of "label" in OAEP encryption using openssl (from debian, 1.1.1k-1 + deb11u1), where an undocumented switch is available that allows you to specify a "label" using a hexadecimal string, for example: -pkeyopt rsa_oaep_label: 314F

I tested oaep encryption, including  "label" with openssl and decryption with OpenSC's pkcs#11 module (using this PR).

I plan to add code to "pkcs11-tool" so that "label" can be entered on the command line. 

This PR is already fixed as you suggested and is rebased to current master.

Jakuje (approved):

Jakuje:
> I plan to add code to "pkcs11-tool" so that "label" can be entered on the command line.

Thanks. Looks good. Do you plan to add this API to this PR or a next one after we will merge this one?

popovec:
I think this PR should be merged. I will prepare the changes in pkcs11-tool as a separate PR.

PR #2492, opened 2022-01-19
Cleanup of sc-ossl-compat.h for OpenSSL versions 1.1.1 and above

dengert:
With discussion to only support OpenSSL 1.1.1 and 3.0 above and
most of sc-ossl-compat.h can be removed.

This also removes support for LibreSSL.

 On branch improved-sc-ossl-compat
 Changes to be committed:
	modified:   sc-ossl-compat.h

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/sc-ossl-compat.h
+++ b/src/libopensc/sc-ossl-compat.h
@@ -29,82 +29,20 @@ extern "C" {
 
 #include <openssl/opensslv.h>
 #include <openssl/opensslconf.h>
+
 /*
- * Provide backward compatibility to older versions of OpenSSL
- * while using most of OpenSSL 1.1  API
+ * Drop support of old versions of OpenSSL and LibreSSL
+ * while using most of OpenSSL 1.1.1  API
+ * and working on supporting OpenSSL 3.0
  *
  * LibreSSL is a fork of OpenSSL from 2014
  * In its version of openssl/opensslv.h it defines:
  * OPENSSL_VERSION_NUMBER  0x20000000L (Will not change)
  * LIBRESSL_VERSION_NUMBER  0x2050000fL (changes with its versions.
- * The LibreSSL appears to follow the OpenSSL-1.0.1 API
- *
  */
 
-/*
- * 1.1.0 deprecated ERR_load_crypto_strings(), SSL_load_error_strings(), ERR_free_strings()
- * and ENGINE_load_dynamic.EVP_CIPHER_CTX_cleanup and EVP_CIPHER_CTX_init are replaced
- * by EVP_CIPHER_CTX_reset.
- * But for compatibility with LibreSSL and older OpenSSL. OpenSC uses the older functions
- */
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L  && !defined(LIBRESSL_VERSION_NUMBER)
-# if !(defined(OPENSSL_API_COMPAT) || OPENSSL_API_COMPAT >= 0x10100000L)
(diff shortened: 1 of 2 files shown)


Jakuje:
This looks like it breaks osx build. Can you check what openssl version is used there?
```
ctx.c:848:7: error: implicit declaration of function 'CRYPTO_secure_malloc_initialized' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        if (!CRYPTO_secure_malloc_initialized()) {
             ^
ctx.c:849:3: error: implicit declaration of function 'CRYPTO_secure_malloc_init' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                CRYPTO_secure_malloc_init(OPENSSL_SECURE_MALLOC_SIZE, OPENSSL_SECURE_MALLOC_SIZE/8);
                ^
ctx.c:849:3: note: did you mean 'CRYPTO_secure_malloc_initialized'?
ctx.c:848:7: note: 'CRYPTO_secure_malloc_initialized' declared here
        if (!CRYPTO_secure_malloc_initialized()) {
             ^
2 errors generated.
```
Similarly the windows builds in appveyor:
```
ctx.c
ctx.c(848): error C2220: warning treated as error - no 'object' file generated
ctx.c(848): warning C4013: 'CRYPTO_secure_malloc_initialized' undefined; assuming extern returning int
ctx.c(849): warning C4013: 'CRYPTO_secure_malloc_init' undefined; assuming extern returning int
```

Jakuje:
In any case, I would wait with this one after #2495 so we can see in which state are we in the LibreSSL waters. None of us has strong need to support LibreSSL, but it looks like it will not be that complicated in the end so if Frank can bring the PR to working stage with CI, we should be able to keep up with LibreSSL.

dengert:
The `CRYPTO_secure_malloc_initialized`, `CRYPTO_secure_malloc_initialized` and `CRYPTO_secure_malloc_initialized` were all deleted by this PR. They could be added back in. The were add as inline functions, but it looks like they could have been added 
as defines or macros:  `CRYPTO_secure_malloc_initialized` and  `CRYPTO_secure_malloc_initialized` and return 0. `CRYPTO_secure_malloc_initialized` is void. 

They where defined if this is true: `#if OPENSSL_VERSION_NUMBER < 0x10100000L || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2050300fL)` 

The are not listed in LibreSSL includes from 3.4.2, which has: `#define LIBRESSL_VERSION_NUMBER 0x3040100fL`   OpenSSL > 1.0.1 defines them and LibreSSL does not. So above `#if` is not correct.  

We will see how #2495 goes which says:
> it's a low hanging fruit and there actually are some distros which use LibreSSL.

The tree is growing fast...

mouse07410:
I don't think LibreSSL is a "low-hanging" fruit, and don't think it's worth time and effort.

Jakuje:
Can you rebase the branch on top of the current master and fix the build issues?

dengert:
Yes but not till next week.

dengert:
The PR as submitted, dropped all support for LibreSSL and OpenSSL below 1.1.1 which allowed most of code to be deleted. (removed 242 lines and added 7).  

But https://github.com/OpenSC/OpenSC/pull/2495 has updated support for LibreSSL so I am closing 2492 and may submit a new PR to clean up parts of `sc-ossl-compat.h` in the future.

PR #2502, opened 2022-01-29
pkcs11.h: do not use generic 'interface' variable

alonbl:
Hi,

interface term conflicts with Windows basetyps.h macro.

See: https://github.com/OpenSC/pkcs11-helper/issues/48

As this is the source of truth for this file for many projects, please accept the fix so that MSC people may use pkcs11.h in application.

Thanks to @selvanair for reporting.

Thanks,
Alon


frankmorgner (approved):

Jakuje:
Thank you. Looks good.

PR #2506, opened 2022-02-02
sc-ossl-compat.h cleanup for OpenSSL and LibreSSL

dengert:
Remove unused code and misplaced defines from sc-ossl-compat.h
to support OpenSSL 1.1.1 and 3.0.0+ and LibreSSL 3.4.2+

The "inline" routines are no longer needed and have been removed, as new functions replaced them. 

Many of the changes to previous version of `sc-ossl-compat.h` added many `#ifndef  ...  #define ... #endif` but made the assumption that the the header files where already included. But this depended on the order of include files in each source file.  This had lead to confusion. This dependency has been removed.

The proper way to add `#ifndef ...  #define... #endif` to `sc-ossl-compat.h` would be to include the needed header files in the source files before `sc-ossl-compat.h` is included, or to add the include to `sc-ossl-compat.h`.

Several other source files were modified to include additional header files
or use newer names for functions or macros which are defined in OpenSSL and LibreSSL.

This replaces #2492 which did not include LibreSSL. 

 Date:      Tue Feb 1 20:58:08 2022 -0600
 On branch sc-ossl-compat-cleanup
 Changes to be committed:
	modified:   sc-ossl-compat.h
	modified:   ../pkcs11/openssl.c
	modified:   ../pkcs15init/pkcs15-westcos.c
	modified:   ../tools/piv-tool.c
	modified:   ../tools/pkcs15-init.c

 A spreed sheet of comments and actions: [sc-ossl-compat.pdf](https://github.com/OpenSC/OpenSC/files/7987846/sc-ossl-compat.pdf) You will note that many of the routines used in all three are deprecated in 3.0.0, so these also need to be addressed in future changes.

Tested using PIV card with OpenSSL-1.1.1m, OpenSSL-3.0.1 and LibreSSL-3.4.2

##### Checklist


Jakuje:
Both OSX and AppVeyor builds still seem to fail because of some missing API.

dengert:
Anyone want to look at Appveyer configuration?

These look like conflicting configuration for cygwin: `--disable-openssl`,  `/DOPENSSL_SECURE_MALLOC_SIZE=65536` and `/DENABLE_OPENSSL`

Half of the Appveyer  builds work, half fail.

for example:
Environment: APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2015, VCVARSALL=%VS140COMNTOOLS%/../../VC/vcvarsall.bat, DO_PUSH_ARTIFACT=yes; Configuration: Release; Platform: x86

https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/42429370/job/6vwlqiljmgfjy63h

bash -c "exec 0</dev/null && ./configure --with-cygwin-native **--disable-openssl** --disable-readline --disable-zlib || cat config.log"

cl /O1 /GS /W3 /WX /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_WARNINGS /MT /nologo /DHAVE_CONFIG_H /I..\..\win32 /I..\..\src /IC:\openpace-Win32\src /IC:\OpenSSL-v111-Win32\include **/DOPENSSL_SECURE_MALLOC_SIZE=65536** /IC:\zlib-Win32   "/IC:\Program Files (x86)\Windows Kits\10\Cryptographic Provider Development Kit\Include" "/IC:\Program Files (x86)\Microsoft CNG Development Kit\Include" "/IC:\Program Files (x86)\WiX Toolset v3.11\\SDK\VS2015\inc" /DWINVER=0x0601 /D_WIN32_WINNT=0x0601 /DWIN32_LEAN_AND_MEAN /DENABLE_OPENPACE **/DENABLE_OPENSSL** /DENABLE_ZLIB /DENABLE_MINIDRIVER /DENABLE_SM  /DOPENSC_FEATURES="\"pcsc openssl zlib\"" /Zi /c sc.c ctx.c log.c errors.c asn1.c base64.c sec.c card.c iso7816.c dir.c ef-atr.c ef-gdo.c padding.c apdu.c simpletlv.c gp.c pkcs15.c pkcs15-cert.c pkcs15-data.c pkcs15-pin.c pkcs15-prkey.c pkcs15-pubkey.c pkcs15-skey.c pkcs15-sec.c pkcs15-algo.c pkcs15-cache.c pkcs15-syn.c pkcs15-emulator-filter.c muscle.c muscle-filesystem.c ctbcs.c reader-ctapi.c reader-pcsc.c reader-openct.c 


ctx.c(848): error C2220: warning treated as error - no 'object' file generated
ctx.c(848): warning C4013: 'CRYPTO_secure_malloc_initialized' undefined; assuming extern returning int
ctx.c(849): warning C4013: 'CRYPTO_secure_malloc_init' undefined; assuming extern returning int
log.c
errors.c

and ctx.c:847: `#if defined(ENABLE_OPENSSL) && defined(OPENSSL_SECURE_MALLOC_SIZE)`

src/pkcs15init/pkcs15-westcos.c
```diff
@@ -246,7 +246,8 @@ static int westcos_pkcs15init_generate_key(sc_profile_t *profile,
 	if (BN_set_word(bn, RSA_F4) != 1 ||
 	    EVP_PKEY_keygen_init(pctx) != 1 ||
 	    EVP_PKEY_CTX_set_rsa_keygen_bits(pctx, key_info->modulus_length) != 1 ||
-	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
+	    EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn) != 1 ||
+	    (bn = NULL) == NULL || /* pctx will free bn */
```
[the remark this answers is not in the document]
dengert: If "EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn)"  fails , then bn is still set and needs to be freed. If  "EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn)" passes -  man page says: "The pubexp pointer is used internally by this function so it should not be modified or freed after the call."

frankmorgner:
support for secure memory was added with https://github.com/openssl/openssl/commit/74924dcb3802640d7e2ae2e80ca6515d0a53de7a , i.e. it was added in 1.1.0 and not present in 1.0.2

dengert:
> support for secure memory was added with openssl/openssl@74924dc , i.e. it was added in 1.1.0 and not present in 1.0.2

So what is your point?

src/pkcs15init/pkcs15-westcos.c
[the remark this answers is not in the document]
Jakuje: > EVP_PKEY_CTX_ctrl() and its macros return a positive value for success and 0 or a negative value for failure. In particular a return value of -2 indicates the operation is not supported by the public key algorithm.

We should check for `< 1` to get it right in all cases.

But what I wanted to point out initially was that `(bn = NULL) == NULL` is always true so you will always get to the `r = SC_ERROR_UNKNOWN;` branch. I think you wanted to use `(bn = NULL) != NULL` or something similar.

src/libopensc/sc-ossl-compat.h
```diff
@@ -114,208 +55,11 @@ extern "C" {
 #define EVP_PKEY_get_raw_public_key(p, pu, l)   (0)
 #endif
 
+/* OpenSSL 1.1.1 has FIPS_mode function */
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
-#define EC_POINT_get_affine_coordinates_GFp     EC_POINT_get_affine_coordinates
-#define EC_POINT_set_affine_coordinates_GFp     EC_POINT_set_affine_coordinates
-
-# ifndef FIPS_mode
```
Jakuje, line 121: please, keep this `ifndef`. The OpenSSL 3.0 when built with FIPS module has this macro/function too.

src/pkcs15init/pkcs15-westcos.c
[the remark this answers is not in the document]
mouse07410: Neither `(bn = NULL) == NULL` (always true), nor `(bn = NULL) != NULL` (always false) make any sense.

Please explain what you're trying to accomplish there.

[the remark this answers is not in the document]
dengert: You are right, fix committed.

src/libopensc/sc-ossl-compat.h
dengert: FIPS_mode does not show up in include/openssl/  is 3.0.0 or 3.0.1.

But late in new code is:
```
 59 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 60 #define FIPS_mode()                             EVP_default_properties_is_fips_enabled(NULL)
 61 #endif 
```

I think this  does the same thing.

src/pkcs15init/pkcs15-westcos.c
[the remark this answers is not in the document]
dengert: In reply to @mouse07410  
EVP_PKEY_CTX_set_rsa_keygen_pubexp  takes a pointer to a BN and adds it to the pctx. "The pubexp pointer is used internally by this function so it should not be modified or freed after the call."

The preexisting if statement expression is a series of logical or "||" operations. The operands are function calls and tests that set true or false. As long as segment is false it will try the next. If any return true it will get to the `r = SC_ERROR_UNKNOWN`.

The `EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn) != 1` is the segment that stores the `bn` pointer value in the pctx.  The next segment, `(bn = NULL) != NULL` has an assigment and a test.  `(bn = NULL)` clears the pointer so it is not freed at the end of the function. The result of the `(bn = NULL) != NULL`  is "(always false)"  so the next segment of the if, `EVP_PKEY_keygen(pctx, &key) != 1` will be run.

src/libopensc/sc-ossl-compat.h
```diff
@@ -114,208 +55,11 @@ extern "C" {
...
-#  define EVP_PKT_EXCH    0x0040
-#  define EVP_PKS_RSA     0x0100
-#  define EVP_PKS_DSA     0x0200
-#  define EVP_PKS_EC      0x0400
-# endif
-#endif
-
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
-#define EVP_PKEY_eq                             EVP_PKEY_cmp
-#define EVP_PKEY_CTX_set1_rsa_keygen_pubexp     EVP_PKEY_CTX_set_rsa_keygen_pubexp
```
Jakuje, line 142: This function is deprecated in OpenSSL 3.0 so we need to keep this define.

src/libopensc/sc-ossl-compat.h
```diff
@@ -114,208 +55,11 @@ extern "C" {
 #define EVP_PKEY_get_raw_public_key(p, pu, l)   (0)
 #endif
 
+/* OpenSSL 1.1.1 has FIPS_mode function */
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
-#define EC_POINT_get_affine_coordinates_GFp     EC_POINT_get_affine_coordinates
-#define EC_POINT_set_affine_coordinates_GFp     EC_POINT_set_affine_coordinates
```
Jakuje, line 119: These two are also deprecated in OpenSSL 3.0 so we need to keep this defines.

src/libopensc/sc-ossl-compat.h
Jakuje: Thanks for clarification. We have for some reason this compat macro defined in `openssl/fips.h` in el9, but it looks like it does not complain about redefinition so we are fine.

src/libopensc/sc-ossl-compat.h
```diff
@@ -114,208 +55,11 @@ extern "C" {
...
-#  define EVP_PKT_ENC     0x0020
-#  define EVP_PKT_EXCH    0x0040
-#  define EVP_PKS_RSA     0x0100
-#  define EVP_PKS_DSA     0x0200
-#  define EVP_PKS_EC      0x0400
-# endif
-#endif
-
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
-#define EVP_PKEY_eq                             EVP_PKEY_cmp
```
Jakuje, line 141: The `EVP_PKEY_cmp` is also deprecated in OpenSSL 3.0 so we need to use the new version there.

src/libopensc/sc-ossl-compat.h
```diff
@@ -114,208 +55,11 @@ extern "C" {
...
-#  define EVP_PK_RSA      0x0001
-#  define EVP_PK_DSA      0x0002
-#  define EVP_PK_DH       0x0004
-#  define EVP_PK_EC       0x0008
-#  define EVP_PKT_SIGN    0x0010
-#  define EVP_PKT_ENC     0x0020
-#  define EVP_PKT_EXCH    0x0040
-#  define EVP_PKS_RSA     0x0100
-#  define EVP_PKS_DSA     0x0200
-#  define EVP_PKS_EC      0x0400
```
xhanulik, line 136: Some of those are still used in `p11test` and will be missing if OpenSSL is built with `no-deprecated`.

src/libopensc/sc-ossl-compat.h
dengert: This PR changed these other files too:
modified: sc-ossl-compat.h
modified: ../pkcs11/framework-pkcs15.c
modified: ../pkcs11/openssl.c
modified: ../pkcs15init/pkcs15-westcos.c
modified: ../tools/piv-tool.c
modified: ../tools/pkcs15-init.c

EC_POINT_set_affine_coordinates_GFp   was changed to  EC_POINT_set_affine_coordinates in the few places that used it. 

And there is a pdf:  https://github.com/OpenSC/OpenSC/files/7987846/sc-ossl-compat.pdf in the original conversation of a spread sheet that listed  how some of these old names that have changed.  

The PR has also assumed that we are only going to support OpenSC 1.1.1 and 3.0.0 and above. And also support LibreSSL 3.4.2 (or so) and above. So many of the old names have changed. Look at the PDF

src/libopensc/sc-ossl-compat.h
dengert: See the https://github.com/OpenSC/OpenSC/files/7987846/sc-ossl-compat.pdf and other files that are changed in this PR.

src/libopensc/sc-ossl-compat.h
dengert: There are a lot of warning about deprecated functions, as we have been targeting 1.1.1. All of those are in evp.h which is not included in some of the files that have changed.

```
doug@XUbuntu-20:/afs/anl.gov/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-ossl-3.0.1$ grep warning MA1.LOG 
../../../src/src/pkcs15init/pkcs15-westcos.c:249:6: warning: ‘EVP_PKEY_CTX_set_rsa_keygen_pubexp’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/pkcs15-pubkey.c:1858:5: warning: ‘EC_POINT_get_affine_coordinates_GFp’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3184:2: warning: ‘SHA1_Init’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3185:2: warning: ‘SHA1_Update’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3210:2: warning: ‘SHA1_Final’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3236:2: warning: ‘SHA256_Init’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3237:2: warning: ‘SHA256_Update’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/libopensc/card-iasecc.c:3262:2: warning: ‘SHA256_Final’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:232:2: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:244:3: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:247:4: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:249:3: warning: ‘ENGINE_ctrl_cmd_string’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:250:6: warning: ‘ENGINE_ctrl_cmd_string’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:251:4: warning: ‘ENGINE_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:257:3: warning: ‘ENGINE_set_default’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:258:3: warning: ‘ENGINE_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:232:2: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:244:3: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:247:4: warning: ‘ENGINE_by_id’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:249:3: warning: ‘ENGINE_ctrl_cmd_string’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:250:6: warning: ‘ENGINE_ctrl_cmd_string’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:251:4: warning: ‘ENGINE_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:257:3: warning: ‘ENGINE_set_default’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/pkcs11/openssl.c:258:3: warning: ‘ENGINE_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
../../../src/src/tools/pkcs15-init.c:1346:3: warning: ‘EVP_PKEY_cmp’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
```

src/libopensc/sc-ossl-compat.h
dengert: Code was changed  EVP_PKEY_eq  to   EVP_PKEY_cmp and 
EVP_PKEY_CTX_set1_rsa_keygen_pubexp    to  EVP_PKEY_CTX_set_rsa_keygen_pubexp

dengert:
As a general note this PR was not trying to address deprecated functions in 3.0.0, but rather it was trying to get ride of parts of sc-ossl-compat.h that are for versions < 1.1.1 and work with LibreSSL 3.4.2 and above as well as OpenSSL 1.1.1 and OpenSSL 3.0.0. More changes will be needed as we address building without deprecated 3.0.0 functions. 

Note at all the checks are clean, accept for one that is not related to the PR:
 https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/42440190/job/mbuaf50nxxc79q2c

src/libopensc/sc-ossl-compat.h
dengert: Fixed in pkcs15-init.c to use EVP_PKEY_eq in 3.0.0+ and EVP_PKEY_cmp in 1.1.1 and LibreSSL

src/libopensc/sc-ossl-compat.h
dengert: As pointed out, EVP_PK_RSA and EVP_PK_EC are deprecated in 3.0.0. They are only used in p11test in these routines:
 
./src/tests/p11test/p11test_case_common.c
./src/tests/p11test/p11test_case_readonly.c
./src/tests/p11test/p11test_case_pss_oaep.c

and only used to set "test_cert_t  type" in p11test_case_common.c:

 261         if (EVP_PKEY_base_id(evp) == EVP_PKEY_RSA) {
 262                 o->key = evp;
 263                 o->type = EVP_PK_RSA;
 264                 o->bits = EVP_PKEY_bits(evp);
 265
 266         } else if (EVP_PKEY_base_id(evp) == EVP_PKEY_EC) {
 267                 o->key = evp;
 268                 o->type = EVP_PK_EC;
 269                 o->bits = EVP_PKEY_bits(evp);
 270 
In other words the EVP_PK_* are only used internally to set a type and print error messages.

I would suggest either these "types" be maintained in p11tests or better still, 
EVP_PKEY_RSA and EVP_PKEY_EC be used instead and drop the use of these deprecated defines now.  

These are defined in evp.h in versions we are going to support: LibreSSL and OpenSSL 1.1.1 and 3.0.0:

# define EVP_PKEY_RSA    NID_rsaEncryption
# define EVP_PKEY_RSA2   NID_rsa
# define EVP_PKEY_RSA_PSS NID_rsassaPss
# define EVP_PKEY_EC	 NID_X9_62_id_ecPublicKey

Also note that EVP_PKEY_get_base_id() returns the type of pkey: EVP_PKEY_RSA
and EVP_PKEY_type() returns a specific NID. This might be useful for EVP_PKEY_RSA_PSS?

I would also suggest the evp.h be included specifically in the source code were needed.
Only p11test_case_wrap.c and p11test_case_pss_oaep.c include evp.h

dengert:
While writing the last commit, https://github.com/OpenSC/OpenSC/pull/2506/commits/b0b7cc30e8a59d267a6dd6be945fc6096b096108 showed that LibreSSL may have started as a fork of 1.0.2 but LibreSSL 3.4.2 these days is more like OpenSSL 1.1.1.

```
man -M /opt/ossl-libressl3.4.2/share/man ERR_load_CRYPTO_strings
man -M /opt/ossl-libressl3.4.2/share/man ERR_load_crypto_strings
man -M /opt/ossl-libressl3.4.2/share/man  OPENSSL_init_crypto

"These functions are deprecated.  It is never useful for an application program to call either of them
     explicitly".
```
```
man -M /opt/ossl-1.1.1m/share/man  OPENSSL_init_crypto
man -M /opt/ossl-3.0.1/share/man  OPENSSL_init_crypto

"As of version 1.1.0 OpenSSL will automatically allocate all resources that it needs so no explicit
       initialisation is required."
```
LibreSSL and OpenSSL all says `OPENSSL_INIT_LOAD_CRYPTO_STRINGS` and `OPENSSL_init_crypto`
are deprecated, and are not needed. OpenSSL leaves there use open for special situations
but OpenSC does not appears to have any special situations. 

OpenSC calls ERR_load_crypto_strings in 9 source files and calls OPENSSL_init_crypto in 4 tools. 

So https://github.com/OpenSC/OpenSC/pull/2506/commits/b0b7cc30e8a59d267a6dd6be945fc6096b096108  removes all usages of  ERR_load_CRYPTO_strings and friends, including OPENSSL_config is also not needed and deprecated in OpenSSL 1.1.0

Both LibreSSL and OpenSSL have been providing macros and dummy functions for some time. But if
OPENSSL_NO_DEPRECATED_1_1_0 or OPENSSL_NO_DEPRECATED_3_0 are set these macros or dummy functions are not defined.  

We may be close to building OpenSC with OpenSSL-3.0.x with  `no-deprecated`. Building this PR and using make -k only  `card-iasecc.c` has many errors. It is using SHA_CTX, SHA_LONG,  SHA256_CTX and members of the structures in `iasecc_qsign_data_sha1` and `iasecc_qsign_data_sha256` and does not compile and thus libopensc is not built. But all other files compile. 

I suspect `card-iasecc.c` does  the last part of the hash on the card, but it not clear how to do that in OpenSSL 3.0. 

Anyone want to take a look at `card-iasecc.c`?

src/pkcs15init/pkcs15-oberthur-awp.c
```diff
@@ -1078,10 +1065,9 @@ awp_encode_cert_info(struct sc_pkcs15_card *p15card, struct sc_pkcs15_object *ob
 err:
 	ERR_print_errors_fp(stderr);
 	ERR_clear_error();
-	ERR_free_strings();
 	if (pubkey.exponent.data) free(pubkey.exponent.data);
 	if (pubkey.modulus.data) free(pubkey.modulus.data);
-	if (x) 		X509_free(x);
+	if (x)		X509_free(x);
```
Jakuje, line 1070:
```suggestion
	if (x) X509_free(x);
```
i would suggest to keep a space here instead of the tab. Or just expand it to separate lines to keep it more readable when you touch this part already.

src/pkcs15init/pkcs15-westcos.c
```diff
@@ -246,7 +246,12 @@ static int westcos_pkcs15init_generate_key(sc_profile_t *profile,
 	if (BN_set_word(bn, RSA_F4) != 1 ||
 	    EVP_PKEY_keygen_init(pctx) != 1 ||
 	    EVP_PKEY_CTX_set_rsa_keygen_bits(pctx, key_info->modulus_length) != 1 ||
-	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+		EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
```
Jakuje, line 250:
```suggestion
	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
```
please, indent the second part with spaces to keep the alignment with the rest of the code regardless the tab width

Jakuje:
Thank you for polishing this PR and taking care to remove the old cruft and moving the rest of the code. Added some minor comments.

> We may be close to building OpenSC with OpenSSL-3.0.x with `no-deprecated`. Building this PR and using make -k only `card-iasecc.c` has many errors. It is using SHA_CTX, SHA_LONG, SHA256_CTX and members of the structures in `iasecc_qsign_data_sha1` and `iasecc_qsign_data_sha256` and does not compile and thus libopensc is not built. But all other files compile.

If that is so, can you remove the `--disable-strict` from the openssl 3.0 pipeline in CI (as well as the added CFLAGS):

https://github.com/OpenSC/OpenSC/blob/b71ca16b28c4e1a8d35a326f65bed29a8dec2183/.github/build.sh#L52

> I suspect `card-iasecc.c` does the last part of the hash on the card, but it not clear how to do that in OpenSSL 3.0.
> 
> Anyone want to take a look at `card-iasecc.c`?

Yes. We discussed this in https://github.com/OpenSC/OpenSC/pull/2337#issuecomment-840697055 last year. I think it is not possible to do that right now with 3.0, but I did not ask OpenSSL developers yet. I did not look into that recently, but if you (or @vjardin or @viktorTarasov) have the overview of the use case, it would be great to summarize it in an issue. We might hear there is already a way to do that, they might propose some new API or come up with some other solution.

dengert:
I have been looking at card-iasec.c too. It might be possible to update it.
Will look closer in next few days.

src/libopensc/card-iasecc.c
```diff
@@ -3667,12 +3670,24 @@ sc_get_driver(void)
 	return &iasecc_drv;
 }
 
+#endif /* OPENSC_NODEPRECATE_3_0 */
```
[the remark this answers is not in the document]
dengert: Removed with the revert TEMPORARY PATCH 8e0c3f27dbcf77ccea5c4c3d18f693eccba2bff9

src/pkcs15init/pkcs15-oberthur-awp.c
dengert: OK

src/pkcs15init/pkcs15-westcos.c
dengert: OK

dengert:
The last set of commits includes https://github.com/OpenSC/OpenSC/pull/2506/commits/1f3df786738431fa00fab23a6a830f32eb3e24e0

I Have not tested it, but it compiles on LibreSSL, OpenSSL 1.1.1m and 3.0.1 and 3.0.1-no-deprecated

@Jakuje you commented on talking to OpenSSL developers. 
As noted in the commit message, there are arguments for a discussion with OpenSSL developers to not deprecated the SHA_CTX and SHA256_CTX structures that we use.  (There may be other hash structures that other may also want to use.) 

- 3.0.0 provides `EVP_MD_CTX_get0_md_data`  to get pointer to these structures.
- What good is the pointer if you don't have the definition of the structure?
- ISO/IEC 7816-8  has a valid reason to access the structure in "5.5 HASH operation"
    "The HASH operation initiates the computation of a hash-code by performing:
            either the complete computation inside the card or
           **a partial computation inside the card (e.g. last round of hashing).**"
- Access is required as the SHA "h" integers from the the next to last round is send to the card to start the last round.
- The structures define the basic SHA data and are not expected to change. 

The change may be as simple as moving two  #ifdefs:
[OpenSSL-3-sha.diff.txt](https://github.com/OpenSC/OpenSC/files/8015941/OpenSSL-3-sha.diff.txt)

Jakuje:
Thank you for looking into that. The changes to iasec look reasonable, but still it looks pretty fragile depending on internal structure. I assume the more correct way would be using some OSSL_PARAMS to get the the content of the internal strucutre, but I did not see any OSSL_PARAM that would provide this info.

I added one comment (hope it arrived correctly) to the use of #ifs, which would be better in the sc-ossl-compat.h.

dengert:
> Could this be handled with a compat define in sc-ossl-compat.h to avoid these ifs?
Yes. But this is the only place in the code that needs to get at the SHA_CTX structures. 

Can we get this code tested before trying to move the code to sc-ossl-compat.h?
I do not have a iasecc card.
@viktorTarasov  @vjardin @xhanulik  can you test this code?

dengert:
In response to:
> Thank you for looking into that. The changes to iasec look reasonable, but still it looks pretty fragile depending on internal structure. I assume the more correct way would be using some OSSL_PARAMS to get the the content of the internal structure, but I did not see any OSSL_PARAM that would provide this info.

I submitted an issue to OpenSSL on the deprecation of the SHA_CTX in 3.0.0 See: https://github.com/openssl/openssl/issues/17688

dengert:
@viktorTarasov do you have any comments on this PR?

Based on the response in https://github.com/openssl/openssl/issues/17688 it is not likely OpenSSL will provide access to internal hash algorithm data. 

So we have some choices:

1. Drop support for `card-iasecc.c`  functions  `iasecc_qsign_data_sha1` and `iasecc_qsign_data_sha256` if not supported in the OpenSSL/LibreSSL. i.e. OpenSSL 3.0 and above when no-deprecated is supported. 

2. Include an internal SHA256 function used only for cards doing this last round of hash on card. 

It is not clear if these cards are still in use and these days SHA1 should not be used for signatures so if we want to continue to uses these cards, we would only need to address SHA256. 

If we do (2.) there are a number of open source routines or examples. These have license issues. Anyone want to pick one?

So in any case, I will rework the PR to do  (1.).  If we decide to do (2.) or OpenSSL provides a way to use their internal  hash algorithm data.  that can be added.

dengert:
Any one have  IAS-ECC smart card? Developer or user?
This PR needs testing with OpenSSL 3.0, as well as OpenSSL-1.1.1 and Libressl.

Based on OpenSSL's responses to https://github.com/openssl/openssl/issues/17688 it is likely the card when used with OpenSSL 3.0 will not do signatures.

The last commit  https://github.com/OpenSC/OpenSC/pull/2506/commits/541bfe7ec5c38949870553cd05c0de58d3b5b98f
implements the Option (1.) in https://github.com/OpenSC/OpenSC/pull/2506#issuecomment-1039381365 which may not be acceptable to many.

Option (2.)  is still possible, by using an existing implementation of SHA256, or OpenSC could write one.

dengert:
So far no developer or user with interest in the `card-isaecc.c` driver has responded to my 3 comments in the last ten days for help in testing. 

The card drivers  implements i "IAS ECC Identification Authentication Signature European Citizen Card", "Technical Specifications Revision: 1.0.1",  "21/03/2008"  which requires partial hash on card. 
Also see: https://github.com/OpenSC/OpenSC/wiki/IAS-ECC

Does the lack of interest mean there are no active cards that require this driver and it "qualified signatures"?  I can find what appears to be newer specifications for sale. Being in the U.S.A. I have no personal stack in this driver. 

But for now I am intending to clean up this PR with the "qualified signatures" returning not supported if linked with OpenSSL 3.0.

P.S. 
I have written a SHA256 implementation from FISP-180-4 that could be used for this card is really needed. (Option (2. as listed above.) With some testing it gives the same results as the OpenSSL SHA256 version.

dengert:
Force-pushed this PR to remove a commit and its revert, squashed another and changed name of another.

 I believe all this review issues have been addressed, accept for the need for SHA_CTX and SHA256_CTX.  These are only needed in the `card-iasecc.c` driver to support partial hash in software and last round  of hash on the card. So these these changes where not moved to `sc-ossl-compat.h`. 

`card-iasecc.c` when complied with OpenSSL > 3.0 will not sign data due to OpenSSL changes to not expose  SHA_CTX and SHA256_CTX. This driver may not be used with any current IAS ECC cards. If it is any additional changes can be in an other PR.

src/pkcs15init/pkcs15-oberthur-awp.c
Jakuje: I do not see this minor issue fixed in the final verison

src/libopensc/card-iasecc.c
```diff
@@ -3181,45 +3208,91 @@ iasecc_qsign_data_sha1(struct sc_context *ctx, const unsigned char *in, size_t i
...
+		goto err;
+	}
+	
+#if defined(LIBRESSL_VERSION_NUMBER)
+	md_data = (SHA_CTX *)mdctx->md_data;
+#elif  OPENSSL_VERSION_NUMBER < 0x30000000L
+	md_data = EVP_MD_CTX_md_data(mdctx);
+#else
+	md_data = EVP_MD_CTX_get0_md_data(mdctx);
+#endif
```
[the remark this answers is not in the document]
dengert: I will fix the white space.
In regards to `card-iasecc.c` I know you wanted the defines in `sc-ossl-compat.h`.  But here is my thinking on `card-iasecc.c` and `sc-ossl-compat,h`:

- No one as said they even want to test with the current code the `card-iasecc.c` code. 
- The Specs for this card are from 2008 and appeard to be: https://dvv.fi/documents/16079645/17324992/IAS+ECC+v1_0_1UK.pdf/84043361-56ad-bc26-8223-37e0d79355d8/IAS+ECC+v1_0_1UK.pdf?t=1622729405351
- This may be a unused driver as the IAE-ECC are for gov issued cards.
- `card-iasecc.c`  is the only source code that looks at the `md_data`.
- OpenSSL-3.0.0 developers do not want to expose the `md_data` Today, `md_data = `EVP_MD_CTX_get0_md_data(mdctx);` returns NULL, but they also said `EVP_MD_CTX_get0_md_data` should be deprecated This could happen soon. 
- `md_data = EVP_MD_CTX_md_data(mdctx);` compiles with OpenSSL 1.1.1
- `md_data = (SHA_CTX *)mdctx->md_data;`  compiles with LibreSSL
- If we do option (2.) "Include an internal SHA256 function used only for cards doing this last round of hash on card." it should be used for these two routines and not use LibreSSL or OpeenSSL-1.1.1  so no \#ifdefs would be needed. 
- The code is already hard to read, you need to understand what is in `SHA_CTX`, `SHA256_CTX` and  the specs listed above in  "6.1.1.3 Protocol steps" and "9.6 Perform security operation commands" that describes the data sent and the APDU. 

If I was going to change `card-iasecc.c` I would put `#if OPENSSL_VERSION_NUMBER >= 0x30000000L`  in `iasecc_qsign_data_sha1` and `iasecc_qsign_data_sha256` and  return SC_ERROR_NOT_SUPPORTED 

And since `card-iasecc.c` appears to be the last issue for supporting OpenSSL-3.0.0, I did not want to pollute `sc-ossl-compat.h` with defines we will most like drop if we drop the driver or add an OpenSC SHA256.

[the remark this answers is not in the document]
dengert: See https://github.com/OpenSC/OpenSC/pull/2506/commits/a9234ee8fdb7e0e0f872de7fd4a793fa86a79f73  which will run as before, but when compiled with OpenSSL 3.0 returns SC_ERROR_NOT_SUPPORTED.

Jakuje (approved):
as long as we do not have any feedback from any card owners, this looks ok to me.

dengert:
Thanks. Feedback and testing are needed.

frankmorgner (approved):

frankmorgner:
I've tested with IAS/ECC and SM (despite https://github.com/OpenSC/OpenSC/pull/2522 it looks good)

Jakuje:
Can you resolve the conflicts and rebase on current master? Is there something to address in this issue or should we merge it for the next 0.23.0 release?

dengert:
> Can you resolve the conflicts and rebase on current master?

Rebased against master of today.  I also added https://github.com/OpenSC/OpenSC/pull/2506/commits/d4c4a98c89c88171372fe24aae39e0234259d9d9 to address use of engine for GOST #2586 

OpenSC builds using OpenSSL-1.1.1q, OpenSSL 3.0.5 and LibreSSL-3.4.1.

OpenSC builds on XUbuntu-22.04 using Ubuntu provided OpenSSL 3.0.2 15 Mar 2022. 

>  Is there something to address in this issue or should we merge it for the next 0.23.0 release?

This is a lot here, as it gets rid of support for OpenSSL < 1.1.1. So when do you want to drop that too?

frankmorgner (approved):
lgtm, I didn't notice the change regarding the gost engine, though...

frankmorgner:
nevermind, I just saw that check which disabled gost on openssl >= 3.0

src/pkcs11/openssl.c
```diff
@@ -227,6 +227,11 @@ void
 sc_pkcs11_register_openssl_mechanisms(struct sc_pkcs11_card *p11card)
 {
 	sc_pkcs11_mechanism_type_t *mt = NULL;
+/*
+ * Engine support is being dropped in 3.0. OpenSC loads GOST as engine.
```
Jakuje, line 231:
```suggestion
 * Engine support is being deprecated in 3.0. OpenSC loads GOST as engine.
```

src/pkcs11/openssl.c
```diff
@@ -272,6 +277,7 @@ sc_pkcs11_register_openssl_mechanisms(struct sc_pkcs11_card *p11card)
 		CRYPTO_set_locking_callback(locking_cb);
 #endif
 #endif /* !defined(OPENSSL_NO_ENGINE) */
+#endif /* OPENSSL_VERSION_NUMBER >= 0x30000000L && !defined(LIBRESSL_VERSION_NUMBER) */
```
Jakuje, line 280: The part `!defined(LIBRESSL_VERSION_NUMBER)` should not go here, but on the line 268 with the `!OPENSSL_NO_STATIC_ENGINE && !OPENSSL_NO_GOST ` condition to match the opening.

src/pkcs11/openssl.c
dengert: Done in https://github.com/OpenSC/OpenSC/pull/2506/commits/723f559db0bf6d4e7731afe2b77e973662867b7c

src/pkcs11/openssl.c
dengert: \#endif changed to match starting \#if  in https://github.com/OpenSC/OpenSC/pull/2506/commits/723f559db0bf6d4e7731afe2b77e973662867b7c

Jakuje:
Sorry, one more thing, now when the OpenSSL 3.0 build works without the warnings, we should remove the workaround in 
https://github.com/OpenSC/OpenSC/blob/58d8099cb65d02589b4c54fbb5c820b0f4183aa5/.github/build.sh#L52
and build again with `-Werror` as on any other platform. Then I think we are good to merge this for the 0.23.0 release.

Jakuje:
> This is a lot here, as it gets rid of support for OpenSSL < 1.1.1. So when do you want to drop that too?

I think dropping support for OpenSSL <1.1.1 can wait and will be problematic because of libressl which does not support much of the newer APIs as far as I know. We can create a new issue for this to keep track of that.

If there will be no more comments, I will merge it in few days.

dengert:
While developing the sc-compact-cleanup I was surprised that  Libressl has been keeping up with newer versions of OpenSSL.   I was using LibreSSL 3.4.1 in February from October 14th, 2021. The latest stable release is 3.5.3 from May 22  See https://www.libressl.org/releases.html

The only warning  where a number of these:
/usr/bin/ld: ../../src/libopensc/.libs/libopensc.so: warning: EVP_EncryptFinal is often misused, please use EVP_EncryptFinal_ex and EVP_CIPHER_CTX_cleanup

PR #2507, opened 2022-02-03
Default value of SOURCE_PATH in tests was duplicatingn '/'s. Fixed it.

frt:
Just a very small fix to the tests scripts.


Jakuje:
Thank you. Looks good!

PR #2510, opened 2022-02-08
removed dead code

frankmorgner:
fixes coverity scan 374950 Logically dead code

Note that althoug this now is *correct* code, it looks that `init_gost_params` is
actually broken for OpenSSL 3.0 and later, because no are initialized...

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs15-init.c
+++ b/src/tools/pkcs15-init.c
@@ -1879,7 +1879,7 @@ init_gost_params(struct sc_pkcs15init_keyarg_gost_params *params, EVP_PKEY *pkey
 #else
 	char name[256]; size_t name_len = 0;
 #endif
-	int nid = 0;
+	int nid = NID_undef;
 
 	assert(pkey);
 	if (EVP_PKEY_id(pkey) == NID_id_GostR3410_2001) {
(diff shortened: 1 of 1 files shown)


frankmorgner:
@Jakuje @xhanulik, do you have an idea how a correct fix for OpenSSL 3.0 could look like?

Jakuje (approved):
looks good. The gost will probably not work with 3.0 right now though, but there is not much ways to test it for non-Russians.

src/tools/pkcs15-init.c
```diff
@@ -1892,7 +1892,6 @@ init_gost_params(struct sc_pkcs15init_keyarg_gost_params *params, EVP_PKEY *pkey
 #else
```
xhanulik, line 1892: I think that `nid` should be set above as 
```
nid = EC_GROUP_get_curve_name(EC_KEY_get0_group(key));
```
so both `#endif` can remain as they were and the `switch` statement will work appropriately.

Jakuje (approved):
looks good. Do you want to squash the changes before we will merge it?

frankmorgner:
Done. Thanks, @xhanulik , for the pointer.

PR #2513, opened 2022-02-23
Fixed more problems reported by coverity

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -1135,19 +1135,8 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 	sc_context_t *ctx = reader->ctx;
 	struct pcsc_global_private_data *gpriv = (struct pcsc_global_private_data *) ctx->reader_drv_data;
 	struct pcsc_private_data *priv = reader->drv_data;
-	DWORD rcount, feature_len, i;
-	PCSC_TLV_STRUCTURE *pcsc_tlv;
-	union {
-		PCSC_TLV_STRUCTURE msg;
-		u8 buf[256];
-	} feature_buf;
-	union {
-#ifdef PIN_PROPERTIES_v5
-		PIN_PROPERTIES_STRUCTURE_v5 capsv5;
-#endif
-		PIN_PROPERTIES_STRUCTURE caps;
-		u8 buf[SC_MAX_APDU_BUFFER_SIZE];
-	} rbuf;
+	DWORD rcount, i;
+	u8 buf[256];
 	LONG rv;
 	const char *log_disabled = "but it's disabled in configuration file";
 	int id_vendor = 0, id_product = 0;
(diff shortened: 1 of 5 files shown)


popovec (commented):
The coverity scan appears to indicate an error on two lines: 
`````
1032     
>>>     CID 376511:    (STRING_NULL)
>>>     Passing unterminated string "alg" to "print_ssh_key", which expects a null-terminated string.
1033                    print_ssh_key(outf, alg, obj, buf, len);
`````
`````
1095     
>>>     CID 376511:    (STRING_NULL)
>>>     Passing unterminated string "alg" to "print_ssh_key", which expects a null-terminated string.
1096                    print_ssh_key(outf, alg, obj, buf, len);
`````
I don't think it makes sense to check snprintf for a negative return value. snprintf always  terminates the string with a zero. It seems that the coverity scan is mistaken using memcpy. 

I tried the following patch, coverity scan reported: "Defects eliminated: 1" 

`````
diff --git a/src/tools/pkcs15-tool.c b/src/tools/pkcs15-tool.c
index f4900be4..3bbd15fa 100644
--- a/src/tools/pkcs15-tool.c
+++ b/src/tools/pkcs15-tool.c
@@ -1018,8 +1018,8 @@ static int read_ssh_key(void)
                buf[0] = 0;
                buf[1] = 0;
                buf[2] = 0;
-               len = snprintf((char *) buf+4, 20, "ssh-ed25519");
-               memcpy(alg, buf+4, 20);
+               len = snprintf((char *) buf + 4, sizeof(alg) , "ssh-ed25519");
+               strcpy(alg, (char *) buf + 4);
                buf[3] = len;
                len += 4;
 
@@ -1069,8 +1069,8 @@ static int read_ssh_key(void)
                buf[0] = 0;
                buf[1] = 0;
                buf[2] = 0;
-               len = snprintf((char *) buf+4, 20, "ecdsa-sha2-nistp%d", n);
-               memcpy(alg, buf+4, 20);
+               len = snprintf((char *) buf + 4, sizeof(alg), "ecdsa-sha2-nistp%d", n);
+               strcpy(alg, (char *) buf + 4);
                buf[3] = len;
 
                len += 4;
`````

**Comment by frankmorgner** [comment]
Thanks for checking this, I'll have a look at the second case with the same problem. From a logical point of view, I think it's better to first create the appropriate string in `alg` and then copy the result into buffer. This avoids a typecast and coverity will know that `alg` is definately NUL terminated. (in other words, I'd like to keep my suggested change to fix CID 376511. I don't have a local copy of coverity scan, could you check if my change also fixes this problem?)

According to the documentation, snprintf may return an error, which is negative. So while we have our hands on this part of the code, we should also consider the error handling.

**Comment by popovec** [comment]
It looks good, I tried whole #2513, report from  Coverity scan:  "Defects eliminated: 6"

File: src/tools/pkcs15-tool.c [hunk 1]
```diff
@@ -1066,14 +1072,18 @@ static int read_ssh_key(void)
 			goto fail2;
 		}
 
+		len = snprintf(alg, sizeof alg, "ecdsa-sha2-nistp%d", n);
+		if(len < 0) {
```
Jakuje, line 1076:
```suggestion
		if (len < 0) {
```

src/tools/pkcs15-tool.c
```diff
@@ -1007,20 +1007,25 @@ static int read_ssh_key(void)
...
+		unsigned int n;
+		int len;
 
 		n = pubkey->u.eddsa.pubkey.len;
 		if (n != 32) {
 			fprintf(stderr, "Wrong public key length\n");
 			goto fail2;
 		}
 
+		len = snprintf(alg, sizeof alg, "ssh-ed25519");
```
Jakuje, line 1019: in this case, the `alg` could be a static string (unlike in the ecdsa case) to simplify things even more (the same case for RSA and DSA), but I do not insist on this change.

Jakuje (approved):

src/libopensc/reader-pcsc.c
```diff
@@ -1159,44 +1148,41 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
...
+		} else if (pcsc_tlv->tag == FEATURE_IFD_PIN_PROPERTIES) {
+			priv->pin_properties_ioctl = ntohl(pcsc_tlv->value);
+		} else if (pcsc_tlv->tag == FEATURE_GET_TLV_PROPERTIES)  {
+			priv->get_tlv_properties = ntohl(pcsc_tlv->value);
+		} else if (pcsc_tlv->tag == FEATURE_EXECUTE_PACE) {
+			priv->pace_ioctl = ntohl(pcsc_tlv->value);
 		} else {
-			sc_log(ctx, "Reader feature %02x is not supported", pcsc_tlv[i].tag);
+			sc_log(ctx, "Reader feature %02x is not supported", pcsc_tlv->tag);
 		}
```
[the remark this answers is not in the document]
frankmorgner: `switch` would use 3 lines per case, with `if` we're at 2 lines per case. It's simple enough, I'd like to leave it as is..

src/tools/pkcs15-tool.c
frankmorgner: done

src/tools/pkcs15-tool.c
```diff
@@ -1002,25 +1002,26 @@ static int read_ssh_key(void)
 	if (pubkey->algorithm == SC_ALGORITHM_EDDSA) {
 		// SSH supports only ed25519 key now
 
-		char alg[20];
+		const char alg[] = "ssh-ed25505";
```
Jakuje, line 1005:
```suggestion
		const char alg[] = "ssh-ed25519";
```
one last typo

frankmorgner:
I cleaned up the commits; I think they are ready to be merged now.

PR #2516, opened 2022-02-28
Memory management in mechanism registering

### LINKED ISSUE #2508: Memory management inside sc_pkcs11_register_mechanism
### Problem Description

ede23473b changed sc_pkcs11_register_mechanism so that its parameter `mt` is freed on error. Previously, sc_pkcs11_register_mechanism only transferred the ownership of the memory held in `mt` in case of success.

For one, `mt` is *not* always freed on error; there is one exception:
https://github.com/OpenSC/OpenSC/blob/870f21a931772327ba514662abb8f54cc08f9673/src/pkcs11/mechanism.c#L140

Additionaly Coverity Scan reports a possible use of `mt` in register_ec_mechanisms after it has been freed by sc_pkcs11_register_mechanism...
grafik

@ya-isakov @Jakuje

### Proposed Resolution

Unsure... please review the memory handling.

xhanulik:
Fixes #2508

This PR removes freeing of the `mt` attribute, that was added in https://github.com/OpenSC/OpenSC/commit/ede23473b49958af4d5a2d1a8a1eb72aaf5884a2. Now, the `mt` is copied in `sc_pkcs11_register_mechanism()`, so it can be freed after calling the function. Since `sc_pkcs11_mechanism_type_t` contains mechanism private data, the structure is extended by a pointer to a function for copying this data.

 If needed, the copy can be returned for further use (e.g. in `sc_pkcs11_register_sign_and_hash_mechanism()`).

Also, the changes touch duplicates in openssl.c of SHA mechanisms in `sc_pkcs11_register_openssl_mechanisms()` - they are first stored in a variable and released after copying in `sc_pkcs11_register_mechanism()`.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -121,7 +121,7 @@ sc_pkcs11_copy_mechanism(sc_pkcs11_mechanism_type_t *mt,
  */
 CK_RV
 sc_pkcs11_register_mechanism(struct sc_pkcs11_card *p11card,
-				sc_pkcs11_mechanism_type_t *mt)
+				sc_pkcs11_mechanism_type_t *mt, sc_pkcs11_mechanism_type_t **result_mt)
 {
 	sc_pkcs11_mechanism_type_t *existing_mt;
 	sc_pkcs11_mechanism_type_t *copy_mt = NULL;
@@ -168,6 +168,9 @@ sc_pkcs11_register_mechanism(struct sc_pkcs11_card *p11card,
 	p11card->mechanisms = p;
 	p[p11card->nmechanisms++] = copy_mt;
 	p[p11card->nmechanisms] = NULL;
+	/* Return registered mechanism for further use */
+	if (result_mt)
+		*result_mt = copy_mt;
 	return CKR_OK;
 }
 
@@ -1376,7 +1379,7 @@ sc_pkcs11_register_sign_and_hash_mechanism(struct sc_pkcs11_card *p11card,
 		return CKR_HOST_MEMORY;
 	}
 
-	rv = sc_pkcs11_register_mechanism(p11card, new_type);
+	rv = sc_pkcs11_register_mechanism(p11card, new_type, NULL);
 	if (CKR_OK != rv) {
 		new_type->free_mech_data(new_type->mech_data);
 		free(new_type);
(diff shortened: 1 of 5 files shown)


src/pkcs11/mechanism.c
```diff
@@ -1315,6 +1315,16 @@ sc_pkcs11_new_fw_mechanism(CK_MECHANISM_TYPE mech,
...
 }
 
+void sc_pkcs11_free_mechanism(sc_pkcs11_mechanism_type_t *mt)
+{
+	if (!mt)
+		return;
+	if (mt->free_mech_data)
+		mt->free_mech_data(mt->mech_data);
+	free(mt);
+	mt = NULL;
```
frankmorgner, line 1325: `mt = NULL` will most likely be optimized out, because the code doesn't do anything with the new value

src/pkcs11/mechanism.c
```diff
@@ -1300,6 +1342,16 @@ void free_info(const void *info)
 	free((void *) info);
 }
 
+CK_RV copy_hash_signature_info(const void *mech_data, void **new_data)
+{
+	*new_data = calloc(1, sizeof(struct hash_signature_info));
```
frankmorgner, line 1347: please add a check for `mech_data != NULL`

frankmorgner (changes requested):
I think the approach is good. I've added some minor comments, though.

mouse07410:
I suggest keeping `mt + NULL;`, regardless of whether some compiler at some level of optimization may leave it out.

src/pkcs11/mechanism.c
xhanulik: Removed

src/pkcs11/mechanism.c
```diff
@@ -125,22 +152,27 @@ sc_pkcs11_register_mechanism(struct sc_pkcs11_card *p11card,
...
 		return CKR_BUFFER_TOO_SMALL;
 	}
 
 	p = (sc_pkcs11_mechanism_type_t **) realloc(p11card->mechanisms,
 			(p11card->nmechanisms + 2) * sizeof(*p));
 	if (p == NULL)
 		return CKR_HOST_MEMORY;
+	if (sc_pkcs11_copy_mechanism(mt, &copy_mt) != CKR_OK) {
+		free(p);
+		return CKR_HOST_MEMORY;
```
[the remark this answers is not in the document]
xhanulik: Fixed to return return value from `sc_pkcs11_copy_mechanism`

src/pkcs11/mechanism.c
xhanulik: Fixed

Jakuje (approved):

frankmorgner (approved):

PR #2520, opened 2022-03-09
Fuzzing of encoding/decoding functions

xhanulik:
This PR adds fuzzers for encoding functions for PKCS#15 objects and generic `sc_*()` functions.
* `fuzz_pkcs15_encode` iterates over objects on `p15card` and calls corresponding `sc_pkcs15_encode_*_entry()`.
* `fuzz_card` calls some functions from `libopensc/sec.c` and `libopensc/card.c` that are not covered according to OSS-Fuzz coverage reports.

Also, there are some changes in the `fuzz_pkcs15_decode`. The setting of algorithms for the `sc_pkcs15_decode_pubkey()` in a cycle extends testing to other algorithms besides RSA. Some of the `sc_pkcs15_decode_*_entry()` internally use `p15card->app` - these paths were not covered, since `p15card` was not initialized via `sc_pkcs15_bind()`. That is fixed with the usage of the reader from `fuzz_pkcs15_reader`.

After local run, `fuzz_card` encounter asserts in `muscle.c` - these are exchanged for explicit checking the arguments.

##### Checklist


src/tests/fuzzing/fuzz_pkcs15_decode.c
```diff
@@ -20,60 +20,92 @@
...
 #include "libopensc/internal.h"
-#include <stdlib.h>
-#include <string.h>
 
-static struct sc_context *ctx = NULL;
-static struct sc_pkcs15_card *p15card = NULL;
-static sc_card_t card = {0};
+const char *__asan_default_options() {
+  return "verbosity=0:mallocator_may_return_null=1";
+}
```
Jakuje, line 29: This part deserves a comment why it is needed now.

src/tests/fuzzing/fuzz_pkcs15_decode.c
```diff
@@ -20,60 +20,92 @@
...
 
-    struct sc_pkcs15_tokeninfo *tokeninfo = sc_pkcs15_tokeninfo_new();
-    sc_pkcs15_parse_tokeninfo(ctx, tokeninfo, Data, Size);
+    fuzz_get_chunk(reader, &buf, &buf_len);
+    tokeninfo = sc_pkcs15_tokeninfo_new();
+    sc_pkcs15_parse_tokeninfo(ctx, tokeninfo, buf, buf_len);
     sc_pkcs15_free_tokeninfo(tokeninfo);
 
-    sc_pkcs15_parse_unusedspace(Data, Size, p15card);
+    sc_pkcs15_parse_unusedspace(buf, buf_len, p15card);
```
Jakuje, line 107: I think this should use a new `buf` and `buf_len` gathered from `fuzz_get_chunk()`. This is aliased with the data parsed by `sc_pkcs15_parse_tokeninfo()`.

src/tests/fuzzing/fuzz_pkcs15_decode.c
xhanulik: Yes, I believe it should be in the fuzzers that use some memory allocation (with this flag, the allocator will return 0 on out-of-memory, according to [SanitizerCommonFlags](https://github.com/google/sanitizers/wiki/SanitizerCommonFlags)) - I already added it into the remaining two with some describing comments.

src/tests/fuzzing/fuzz_pkcs15_decode.c
xhanulik: Indeed, thank you.

src/tests/fuzzing/fuzz_pkcs15_decode.c
Jakuje: And do we need the `verbosity` one? It should be default. The other should be called `allocator_may_return_null` (without the `m`), but I am afraid this will cause lot of failures during the allocations before we will get to a useful code, even though it might increase code coverate with the error code paths.

I would probably keep this out, at least from this PR and at least for now.

xhanulik: Ok, understood, I removed it.

src/tests/fuzzing/fuzz_pkcs15_encode.c
```diff
@@ -0,0 +1,104 @@
...
+    while (list_size(&ctx->readers)) {
+        sc_reader_t *rdr = (sc_reader_t *) list_get_at(&ctx->readers, 0);
+        _sc_delete_reader(ctx, rdr);
+    }
+    if (ctx->reader_driver->ops->finish != NULL)
+        ctx->reader_driver->ops->finish(ctx);
+
+    ctx->reader_driver = sc_get_fuzz_driver();
+
+    fuzz_add_reader(ctx, Data, Size);
```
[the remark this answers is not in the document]
xhanulik: Yes, it's more convenient, it's moved into `fuzzer_reader.c`.

src/tests/fuzzing/fuzz_card.c
```diff
@@ -0,0 +1,104 @@
...
+
+	flag = *((unsigned long *) data);
+	len = *(data + sizeof(unsigned long));
+
+	/* Establish context for fuzz app*/
+	sc_establish_context(&ctx, "fuzz");
+	if (!ctx)
+		return 0;
+
+	if (fuzz_connect_card(ctx, &card, &reader, data, size) != SC_SUCCESS)
```
Jakuje, line 58: I think this `data` and `size` should be moved by the size of the already read components on lines 50 and 51 so it is not aliased with the card data.

src/tests/fuzzing/fuzz_pkcs15_encode.c
```diff
@@ -0,0 +1,90 @@
...
+	sc_pkcs15_bind(card, NULL, &p15card);
+	if (!p15card)
+		goto err;
+
+	for (obj = p15card->obj_list; obj != NULL; obj = obj->next) {
+		u8 *buf = NULL;
+		size_t buf_len = 0;
+		struct sc_pkcs15_object *key_object = NULL;
+		sc_pkcs15_pubkey_t *pkey = NULL;
+		switch(obj->type & SC_PKCS15_TYPE_CLASS_MASK) {
```
Jakuje, line 56:
```suggestion
		switch (obj->type & SC_PKCS15_TYPE_CLASS_MASK) {
```

Jakuje (approved):

Jakuje:
@frankmorgner can you check this PR if there is something you would like to do differently before I will merge it (after resolving the conflict indeed)?

src/tests/fuzzing/fuzz_pkcs15_decode.c
```diff
@@ -20,56 +20,80 @@
...
+	fuzz_get_chunk(reader, &buf, &buf_len);
+	sc_pkcs15_parse_unusedspace(buf, buf_len, p15card);
+
+	sc_pkcs15_card_free(p15card);
+
+err:
+	sc_disconnect_card(card);
+	sc_release_context(ctx);
+	return 0;
 }
```
frankmorgner, line 99: For the encoding/decoding operations that are independent from each other, it would maybe be better to either use separate fuzzing tests for them or to *not* advance the buffer with the fuzzing input (`buf`/`buf_len` before `sc_pkcs15_parse_tokeninfo`, `sc_pkcs15_parse_unusedspace`, for example). This allows a wider range of fuzzing input to be useful for actual testing...

src/tests/fuzzing/fuzz_pkcs15_encode.c
```diff
@@ -0,0 +1,90 @@
...
+		case SC_PKCS15_TYPE_AUTH:
+			sc_pkcs15_encode_aodf_entry(ctx, obj, &buf, &buf_len);
+			break;
+		case SC_PKCS15_TYPE_CERT:
+			sc_pkcs15_encode_cdf_entry(ctx, obj, &buf, &buf_len);
+			sc_pkcs15_prkey_attrs_from_cert(p15card, obj, &key_object);
+			break;
+		}
+		free(buf);
+	}
```
frankmorgner, line 80: You do pkcs15 bind and then you are using the parsed object to encode this to a buffer, right?

src/tests/fuzzing/fuzzer_reader.c
```diff
@@ -77,96 +77,124 @@ void fuzz_get_chunk(sc_reader_t *reader, const uint8_t **chunk, uint16_t *chunk_
...
-    reader->drv_data = data;
-    snprintf(name, sizeof name - 1, "%zu random byte%s reader (%p)",
-            Size, Size == 1 ? "" : "s", Data);
-    reader->name = strdup(name);
+	if (reader_out)
+		*reader_out = reader;
 
-    reader->ctx = ctx;
-    list_append(&ctx->readers, reader);
+	return SC_SUCCESS;
```
frankmorgner, line 199: I think the last time, we excluded the formatting changes until https://github.com/OpenSC/OpenSC/pull/2017 is merged, but I'm fine either way, with or without changing this here.

frankmorgner (changes requested):
I like the ideas for integrating the encoding/decoding functions with the fuzzing input, good job!

Some minor things could be improved, though - please see the inline comments.

src/tests/fuzzing/fuzz_pkcs15_encode.c
xhanulik: Yes, after the card initialization it just iterates over objects stored in `p15card` and tries to encode it (with corresponding function).

src/tests/fuzzing/fuzzer_reader.c
xhanulik: I moved the formatting changes into separate commit, so can be removed.

xhanulik: Indeed, sorry, removed.

Jakuje:
> I like the ideas for integrating the encoding/decoding functions with the fuzzing input, good job!
> 
> Some minor things could be improved, though - please see the inline comments.

@frankmorgner can you check this PR if there is something to fix from your point of view or are we ready to merge this?

PR #2521, opened 2022-03-09
Fix more problems reported by coverity

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tests/fuzzing/fuzzer.c
+++ b/src/tests/fuzzing/fuzzer.c
@@ -28,27 +28,40 @@ int main (int argc, char **argv)
 {
     printf("Testing one input:\n");
     FILE *fd = NULL;
-    size_t len = 0;
+    long int len = 0;
     unsigned char *buffer = NULL;
+    int r = 1;
 
     if (argc < 2) {
         fprintf(stderr, "No arguments, passing NULL\n");
         len = 0;
     } else {
-        fd = fopen(argv[1], "r");
-        fseek(fd, 0, SEEK_END);
-        len = ftell(fd);
-        rewind(fd);
-        buffer = (unsigned char*) malloc(len);
-        if (fread(buffer, 1, len, fd) != len) {
-            fprintf(stderr, "fread failed\n");
+        if ((fd = fopen(argv[1], "r")) == NULL
+                || fseek(fd, 0, SEEK_END) != 0
+                || (len = ftell(fd)) < 0) {
+            fprintf(stderr, "fopen/fseek failed\n");
+            goto err;
         }
-        fclose(fd);
+        rewind(fd);
+        if ((buffer = (unsigned char*) malloc(len)) == NULL) {
+            fprintf(stderr, "malloc failed\n");
+            goto err;
+        }
+
+        if (fread(buffer, 1, len, fd) != (size_t)len) {
+            fprintf(stderr, "fread failed\n");
+            goto err;
+        }
     }
 
     LLVMFuzzerTestOneInput(buffer, len);
+    r = 0;
+
+err:
+    if (fd)
+        fclose(fd);
     if (buffer)
         free(buffer);
     printf("Done!\n");
-    return 0;
+    return r;
 }


Jakuje (approved):
otherwise lgtm

src/pkcs11/framework-pkcs15.c
```diff
@@ -5828,7 +5828,7 @@ static CK_RV register_ec_mechanisms(struct sc_pkcs11_card *p11card, int flags,
 		unsigned long ext_flags, CK_ULONG min_key_size, CK_ULONG max_key_size)
 {
 	CK_MECHANISM_INFO mech_info;
-	sc_pkcs11_mechanism_type_t *mt, *registered_mt;
+	sc_pkcs11_mechanism_type_t *mt = NULL, *registered_mt = NULL;
```
xhanulik, line 5831: I think that instead of the initialization of `mt` and `registered_mt` to NULL, the `mt` should be set to the `registred_mt` if the registration is successful below.

On line 5855, new `mt` is created as 
```
mt = sc_pkcs11_new_fw_mechanism(CKM_ECDSA, &mech_info, CKK_EC, NULL, NULL, NULL);
```
and on lines 5858 - 5862 it is registered if only the condition is true - the `registered_mt` can then be used in calling `sc_pkcs11_register_sign_and_hash_mechanism()`. However, when the condition is false, `registered_mt` remains NULL.

I would suggest add
```
mt = registered_mt;
```
after the registration and release of the `mt` on line 5860. Then the `mt` can be used in `sc_pkcs11_register_sign_and_hash_mechanism()` no matter if it was registered above or not. 
(Or at least I think that was the way `mt` was used [before](https://github.com/OpenSC/OpenSC/blob/f1cac4700b8062998131d640c46f6f3ac50d0fa5/src/pkcs11/framework-pkcs15.c#L5852-L5867))

Jakuje: Initializing pointers to NULL is always a good idea. Strictly speaking, the initialization of the `mt` to NULL is not needed as it is always initialized before use, but it should not hurt.

But you are right that in case the `flags & SC_ALGORITHM_ECDSA_HASH_NONE` is evaluated to false, the `registered_mt` is kept either uninitialized (or NULL with the proposed patch) and calling `sc_pkcs11_register_sign_and_hash_mechanism()` with the last NULL argument would crash (or fail with the currently suggested patch).

In that case, we either need to update `mt = registered_mt` on line 5860 and change all the uses of the `registered_mt` below or add else branch, which will update `registered_mt = mt` in case the registration on line 5855 is skipped.

src/tests/fuzzing/fuzz_pkcs15init.c
```diff
@@ -359,17 +359,17 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
+        sc_pkcs15init_unbind(profile);
+        if (tmp_p15_data != p15card)
+            sc_pkcs15_unbind(tmp_p15_data);
+    }
+    if (p15card) {
+        sc_pkcs15_unbind(p15card);
+    }
     if (card)
-	    sc_disconnect_card(card);
+        sc_disconnect_card(card);
```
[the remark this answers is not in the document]
frankmorgner: for some reason the complete file uses spaces instead of tabs. my changes unifies the use of spaces...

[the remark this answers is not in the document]
Jakuje: Yeah. I missed that initially. Its another reason why to have this automated using clang-format. I would either reformat the whole file or drop it from this PR.

src/pkcs11/framework-pkcs15.c
frankmorgner: I'm a bit confused, because I don't see that `mt` is used if I would add `mt = registered_mt;` on line 5860.

I suggest to keep the suggested change as is so that the code fails instead of crashing. And I'd be happy if you could submit a more sophisticated fix to make things work instead of failing. OK?

Jakuje: `mt` is not used, but the `registered_mt` is used in `sc_pkcs11_register_sign_and_hash_mechanism`. But I am fine to resolve this in separate PR.

src/tests/fuzzing/fuzz_pkcs15init.c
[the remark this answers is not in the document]
frankmorgner: OK, I've removed this commit

Jakuje:
Thank you. Merging this and @xhanulik will have a look into the remaining issue with the mechanism registration.

PR #2522, opened 2022-03-12
Revert "IASECC/Gemalto: add support"

frankmorgner:
This reverts commit e93bd3983ca135f63dc8860febca3ee7f702853a.

It broke support for Gemalto MultiApp IAS/ECC v1.0.1, which can't select
files anymore. Propably we need a better way to distinguesh this card
from CARTE IAS ECC DUAL ID ONE COSMO.

@vjardin , do you maybe have an idea how to distinguish both cards so that they are both supported? I'm not an expert on IAS/ECC... all I know is that my card doesn't work anymore with the reverted commit.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -920,7 +920,6 @@ iasecc_select_file(struct sc_card *card, const struct sc_path *path,
 
 	sc_print_cache(card);
 	if ((!iasecc_is_cpx(card)) &&
-	    (card->type != SC_CARD_TYPE_IASECC_GEMALTO) &&
 	    (path->type != SC_PATH_TYPE_DF_NAME
 			&& lpath.len >= 2
 			&& lpath.value[0] == 0x3F && lpath.value[1] == 0x00))   {
@@ -1018,7 +1017,6 @@ iasecc_select_file(struct sc_card *card, const struct sc_path *path,
 			    card->type == SC_CARD_TYPE_IASECC_AMOS ||
 			    card->type == SC_CARD_TYPE_IASECC_MI ||
 			    card->type == SC_CARD_TYPE_IASECC_MI2 ||
-			    card->type == SC_CARD_TYPE_IASECC_GEMALTO ||
 			    iasecc_is_cpx(card)
 			    )   {
 				apdu.p2 = 0x04;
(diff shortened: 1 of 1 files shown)


vjardin:
Hi @frankmorgner ,

I am puzzle since I do not have good ideas yet in order to distinguish both cards.
It was easy for me to buy the Cosmo one (per my commit log). Where can I buy/get a "Gemalto MultiApp IAS/ECC v1.0.1" ?

frankmorgner:
I'm not quite sure how to distinguish the two cards. I got mine from @viktorTarasov a couple of years ago as sample card. Attached, you'll find a log of runnint `pkcs11-tool --test --login` with my card. Maybe you can spot some bits that could be used...
[iasecc_gemalto.txt](https://github.com/OpenSC/OpenSC/files/8252886/iasecc_gemalto.txt)

vjardin:
Thanks @frankmorgner 

your `iasecc_gemalto.txt` : please, can you provide the logs of the same code (same version of opensc) when it works and does not work (so with and without your patch), the logs should be without the timestamps `P:46835; T:0x140114622256960 12:40:09.011` ?

I'd like to do a diff of both traces.

Meanwhile, I'd be very interested to get such card too. I could not find it. Please, how did @viktorTarasov get it ? @viktorTarasov , can you help me to get one ?

frankmorgner:
[iasecc_gemalto_bad.txt](https://github.com/OpenSC/OpenSC/files/8302976/iasecc_gemalto_bad.txt)
Here you have a failing log.

frankmorgner:
In vim, I've removed the process/thread IDs via `:%s/.*\[opensc-pkcs11\] //g`:
[iasecc_gemalto_bad_clean.txt](https://github.com/OpenSC/OpenSC/files/8303000/iasecc_gemalto_bad_clean.txt)
[iasecc_gemalto_clean.txt](https://github.com/OpenSC/OpenSC/files/8303001/iasecc_gemalto_clean.txt)

frankmorgner:
@vjardin any ideas?

dengert:
Doing a diff of the two logs above, I am surprised at how many lines differ.
The revert only changes 1 file.  yest there are line numbers that differ in many other files. 
 For example, ctx.c has some changes to the line numbers:
```
-ctx.c:858:sc_context_create: ===================================
-ctx.c:859:sc_context_create: opensc version: 0.22.0
+ctx.c:854:sc_context_create: ===================================
+ctx.c:855:sc_context_create: opensc version: 0.22.0
```

Can you checkout master build and run the test, then apply the revert PR to the same master and build and test again?

I suspect that you developed the PR against an older version of master with your PR and it fixed you problem. But your problem may have already been fixed in the github version of master via changes in other files.

Jakuje:
Was there any progress on this issue?

frankmorgner:
@vjardin has a card that requires the changes currently in `master` while I have an older sample card from @viktorTarasov , which requires the change in question to be reverted. The problem is that we can't distinguish the two cards so I'm unsure how to proceed. It would be nice to keep my sample card functioning for future release tests. Maybe Viktor has some idea on how to continue.

frankmorgner:
@vjardin could you please generate a log as i did above with your card for comparison?

Issue #2517, opened 2022-03-01
Minidriver Reinit Problem


jozsefd:
### Problem Description

I have been struggling with various minidriver issues since version 0.20. The problem usually happens in non typical environments, for instance if more than one card reader is connected or when the card is used via Citrix VDI, that's why it took so long to identify the possible cause. 

Well, it has something to do with the faulty caching policy of the Smartcard Subsystem in Windows, but before Version 0.20, the minidriver could handle the scenarios like plugging a card from one card reader to another.

The commit which introduced this issue was [#111246f1d23f6705368d1e615b2a7b3b930a189a
](https://github.com/OpenSC/OpenSC/commit/111246f1d23f6705368d1e615b2a7b3b930a189a)
Among other things, it has changed the function check_card_reader_status() and instead of calling reinit_card_for(), it calls sc_ctx_use_reader(). This solution may work in most cases, but causes issues in some special cases.

### Proposed Resolution

If I revert the following hunks of #111246f
@@ -401,27 +420,37 @@ check_card_reader_status(PCARD_DATA pCardData, const char *name)
@@ -2411,11 +2411,22 @@ int pcsc_use_reader(sc_context_t *ctx, void * pcsc_context_handle, void * pcsc_c
then the issue seems to be fixed.

#### Problem by plugging a card into a second card reader
- more than one card reader is connected to a Windows 10 system
- trying to log on or unlock Windows with the card, Windows shows the card credentials, no PIN entered
- unplug the card and plug into the second reader: Windows is unable to connect to the card, shows an error

(It may be reproduced in logged in state too, but it requires certificate loading tool other than the standard Windows service)

#### Card via Citrix VDI
- a system with a single card reader, connected to a VDI host, card shared
- Windows shows card details correctly (like certutil -scinfo)
- replace the card with the new one
- Windows shows the card data of the first card incorrectly

#### The problem with the second card reader (note: at this point the card was in Omnikey 3121):
```
[cardmod] CardReadFile
[cardmod] pszDirectoryName = mscp, pszFileName = cmapfile, dwFlags = 0, pcbData=000000299A6FEB80, ppbData=000000299A6FEB88
[cardmod] check_reader_status for CardReadFile
[cardmod] MD_Function:check_card_status:396 called
[cardmod] MD_Function:check_card_status:406 returning with: 0x00000000
[...]

PR #2523, opened 2022-03-13
Minidriver fixes for ECC keys

dengert:
Fix for https://github.com/arekinath/PivApplet/issues/64

Minidriver fixes for ECC keys.

Cards with ECC keys need the KSP.

A minidriver is considered to support both CSP and KSP if registry has both:
"Crypto Provider"="Microsoft Base Smart Card Crypto Provider" and 
 "Smart Card Key Storage Provider"="Microsoft Smart Card Key Storage Provider"

Rather then change "Crypto Provider" a new registry key was added:
"InstalledBy"="OpenSC" which is used during uninstall to remove OpenSC registry.
Users and vendors are not expected to use this key so their entries are not deleted by OpenSC.

MD_MAX_KEY_CONTAINERS changed from 13 to 32 to support known cards that have more the 12 keys. 
    
"CardGetProperty" "CP_CARD_KEYSIZES" passes key_type in dwFlags
    
SC_ALGORITHM_EXT_EC_NAMEDCURVE in not in the flags, but in ext_flags
so remove from the test. If any test is needed it should be to match
the curve name supported by Windows.

 On branch minidriver-ECC
 Changes to be committed:
	modified:   minidriver.c

Tested on Windows 10 using Idemia PIV card with 8 ECC keys. 

Implementation of this PR will affect existing  installations, as the Microsoft cert store saves "Provider = OpenSC CSP" This may require uses to cleanup the cert store so "Provider = Microsoft Base Smart Card Crypto Provider" for RSA. Some entries in cert store have "Provider = Microsoft Smart Card Key Storage Provider" for ECC keys if found on a card.  

In other words we should never have tampered with the "Crypto Provider"="Microsoft Base Smart Card Crypto Provider" in the registry.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -260,6 +260,15 @@ static const struct sc_atr_table piv_atrs[] = {
 	{ NULL, NULL, NULL, 0, 0, NULL }
 };
 
+static struct piv_supported_ec_curves {
+	struct sc_object_id oid;
+	size_t size;
+} ec_curves[] = {
+	{{{1, 2, 840, 10045, 3, 1, 7, -1}},     256}, /* secp256r1, nistp256, prime256v1, ansiX9p256r1 */
+	{{{1, 3, 132, 0, 34, -1}},              384}, /* secp384r1, nistp384, prime384v1, ansiX9p384r1 */
+	{{{-1}}, 0} /* This entry must not be touched. */
+};
+
 /* all have same AID */
 static struct piv_aid piv_aids[] = {
 	{SC_CARD_TYPE_PIV_II_GENERIC, /* TODO not really card type but what PIV AID is supported */
@@ -3457,12 +3466,14 @@ static int piv_init(sc_card_t *card)
 	_sc_card_add_rsa_alg(card, 3072, flags, 0); /* optional */
 
 	if (!(priv->card_issues & CI_NO_EC)) {
+		int i;
 		flags = SC_ALGORITHM_ECDSA_RAW | SC_ALGORITHM_ECDH_CDH_RAW | SC_ALGORITHM_ECDSA_HASH_NONE;
 		ext_flags = SC_ALGORITHM_EXT_EC_NAMEDCURVE | SC_ALGORITHM_EXT_EC_UNCOMPRESES;
 
-		_sc_card_add_ec_alg(card, 256, flags, ext_flags, NULL);
-		if (!(priv->card_issues & CI_NO_EC384))
-			_sc_card_add_ec_alg(card, 384, flags, ext_flags, NULL);
+		for (i = 0; ec_curves[i].oid.value[0] >= 0; i++) {
+			if (!(priv->card_issues & CI_NO_EC384 && ec_curves[i].size == 384))
+				_sc_card_add_ec_alg(card, ec_curves[i].size, flags, ext_flags, &ec_curves[i].oid);
+		}
 	}
 
 	if (!(priv->card_issues & CI_NO_RANDOM))
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -3690,8 +3690,10 @@ DWORD WINAPI CardGetContainerInfo(__in PCARD_DATA pCardData, __in BYTE bContaine
 				memcpy(((PBYTE)publicKey) + sizeof(BCRYPT_ECCKEY_BLOB),  pubkey_der.value + 3,  pubkey_der.len -3);
 
 				logprintf(pCardData, 3,
-					  "return info on ECC SIGN_CONTAINER_INDEX %u\n",
-					  (unsigned int)bContainerIndex);
+					  "return info on ECC SIGN_CONTAINER_INDEX %u cbKey:%u dwMagic:%u\n",
+					  (unsigned int)bContainerIndex,
+					  (unsigned int)publicKey->cbKey,
+					  (unsigned int)publicKey->dwMagic);
 			}
 			if (cont->size_key_exchange)   {
 				sz = (DWORD) (sizeof(BCRYPT_ECCKEY_BLOB) +  pubkey_der.len -3);
@@ -3729,13 +3731,22 @@ DWORD WINAPI CardGetContainerInfo(__in PCARD_DATA pCardData, __in BYTE bContaine
 				memcpy(((PBYTE)publicKey) + sizeof(BCRYPT_ECCKEY_BLOB),  pubkey_der.value + 3,  pubkey_der.len -3);
 
 				logprintf(pCardData, 3,
-					  "return info on ECC KEYX_CONTAINER_INDEX %u\n",
-					  (unsigned int)bContainerIndex);
+					  "return info on ECC KEYX_CONTAINER_INDEX %u cbKey:%u dwMagic:%u\n",
+					  (unsigned int)bContainerIndex,
+					  (unsigned int)publicKey->cbKey,
+					  (unsigned int)publicKey->dwMagic);
 			}
 		}
 	}
 	logprintf(pCardData, 7, "returns container(idx:%u) info\n",
 		  (unsigned int)bContainerIndex);
+
+	logprintf(pCardData, 1,
+		  "CardGetContainerInfo bContainerIndex=%u, dwFlags=0x%08X, dwVersion=%lu, cbSigPublicKey=%lu, cbKeyExPublicKey=%lu\n",
+		  (unsigned int)bContainerIndex, (unsigned int)dwFlags,
+		  (unsigned long)pContainerInfo->dwVersion,
+		  (unsigned long)pContainerInfo->cbSigPublicKey,
+		  (unsigned long)pContainerInfo->cbKeyExPublicKey);
 
 err:
 	free(pubkey_der.value);
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -2788,7 +2788,8 @@ md_query_key_sizes(PCARD_DATA pCardData, DWORD dwKeySpec, CARD_KEY_SIZES *pKeySi
 		for (i = 0; i < count; i++) {
 			algo_info = vs->p15card->card->algorithms + i;
 			if (algo_info->algorithm == SC_ALGORITHM_EC) {
-				flag = SC_ALGORITHM_ECDH_CDH_RAW | SC_ALGORITHM_EXT_EC_NAMEDCURVE;
+				flag = SC_ALGORITHM_ECDH_CDH_RAW;
+				/* TODO  check if namedCurve matches Windows supported curves */
 				/* ECDHE */
 				if ((dwKeySpec == AT_ECDHE_P256) && (algo_info->key_length == 256) && (algo_info->flags & flag)) {
 					keysize = 256;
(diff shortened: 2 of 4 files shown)


vletoux:
my first advice would be to test the minidriver with cmck.exe (can provide it to you)

vletoux:
my second advice would be to test it with API monitor (http://www.rohitab.com/apimonitor) to trace KSP / CSP calls

dengert:
Yes please provide cmck.exe.  I will have a look at apimonitor.
@vletoux 
Here is log from minidriver when card is first inserted followed by then when certutil -scinfo is run.
Note that the certificates are loaded into the cert store. 

Using change msi file from this PR which don't appear to make any difference.

[openssl-i3.md.log.txt](https://github.com/OpenSC/OpenSC/files/8254960/openssl-i3.md.log.txt)

metsma:
I remember having same issue with curve key sizes with estonian id card minidriver.
It has been while when I looked the code maybe you can get some hints here

https://github.com/open-eid/minidriver/blob/master/esteidcm.cpp#L330-L345
https://github.com/open-eid/minidriver/blob/master/esteidcm.cpp#L734-L741
https://github.com/open-eid/minidriver/blob/master/esteidcm.cpp#L1248-L1253

dengert:
This experimental fix for https://github.com/arekinath/PivApplet/issues/64 has some minor changes but it looks like the main problem is in how OpenSC minidriver is installed. 

https://github.com/OpenSC/OpenSC/commit/0b683472b07 "win32 setup: add basic minidriver card registration and unregistration via a custom action" replaces in the normal:
```
[HKEY_LOCAL_MACHINE\SOFTWARE\\Microsoft\Cryptography\Calais\SmartCards\<some name>]
"Crypto Provider"="Microsoft Base Smart Card Crypto Provider"
```
with
``` 
"Crypto Provider"="OpenSC CSP"
```

When a card has EC keys, the combination of: "Crypto Provider"="Microsoft Base Smart Card Crypto Provider" and "Smart Card Key Storage Provider"="Microsoft Smart Card Key Storage Provider" says the minidriver can do both. Without this only the CSP provider is used. 

A temp solution, to use cards with EC keys is to use change registry entries:
`[HKEY_LOCAL_MACHINE\SOFTWARE\\Microsoft\Cryptography\Calais\SmartCards\<some name>]` and
`[HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Cryptography\Calais\SmartCards\<some name>]` to use `"Crypto Provider"="Microsoft Base Smart Card Crypto Provider"`.

These test were done on a Windows 10 PRO 64 bit system using the VS 2015 release versions built from this PR. Using am Idemia PIV test card with 8 EC keys and certs. 

Since the use of the `"Crypto Provider"="OpenSC CSP"` has been in the code for 7 years now, there must not be much use of EC keys with the minidriver. 

A long term fix would be to add some other" key" = "value"  like "installedByOpenSC" = "VERSION". 

`certutil -v -scinfo" now  works.
```
The Microsoft Smart Card Resource Manager is running.
Current reader/card status:
Readers: 1
  0: SCM Microsystems Inc. SCR35xx v2.0 USB SC Reader 0
--- Reader: SCM Microsystems Inc. SCR35xx v2.0 USB SC Reader 0
--- Status: SCARD_STATE_PRESENT | SCARD_STATE_UNPOWERED
--- Status: The card is available for use.
---   Card: OPENSC_PIV(2)
---    ATR:
	3b d6 96 00 81 b1 fe 45  1f 87 80 31 c1 52 41 12   ;......E...1.RA.
	22                                                 "


Analyzing card in reader: SCM Microsystems Inc. SCR35xx v2.0 USB SC Reader 0
Microsoft Base Smart Card Crypto Provider: Missing stored keyset

--------------===========================--------------
================ Certificate 0 ================
--- Reader: SCM Microsystems Inc. SCR35xx v2.0 USB SC Reader 0
---   Card: OPENSC_PIV(2)
Provider = Microsoft Smart Card Key Storage Provider
Key Container = 60470002580003013312345678901470

X509 Certificate:
...
```

dengert:
@vletoux @metsma 
Can you look at this PR, especially in the customactions.cpp and the comment in OpenSC.wxs.in:
```
<!-- install an alias for the Base smart card CSP. Using a different CSP in minidriver installation deactivate the plug and play feature
                  but not all other components like the pin change screen available after ctrl+alt+del.
                  It is because the "80000001" entry is still returning the minidriver dll.-->
```

dengert:
Implementation of this PR will affect existing installations because the "Crypto Provider" is saved in the cert store as "Provider".
The `certutil -Silent -store -user My` can be used list the certificates in the windows store for a user. (-Silent cause the test of the private key to be skipped and the "Encryption test FAILED" 
message.)   https://github.com/OpenSC/OpenSC/issues/2517 which is affected by the use of "Crypto Provider=OpenSC CSP" in the registry.

If we want to use ECC keys will will eventually need make the changes. 


For example from some tests to use a self signed cert on a card:
```
================ Certificate 118 ================
Serial Number: 1fef699450b7d8af4b3b0ac3364e0d32
Issuer: CN=Doug Engert-selfSignedTest-20200206-SHA256
 NotBefore: 2/7/2020 4:40 PM
 NotAfter: 2/7/2021 5:00 PM
Subject: CN=Doug Engert-selfSignedTest-20200206-SHA256
Signature matches Public Key
Root Certificate: Subject matches Issuer
Cert Hash(sha1): 57d1c02abfcd55767df5980d72c0c1290552e258
  Key Container = b3f65093-7384-1651-00ee-908132546647
  Provider = OpenSC CSP
Encryption test FAILED
```
IIRC I was trying to use an ECC key, but went back to test with RSA.

annu0412:
Sorry for the late reply
When i execute powershell, i got...

Analyzing card in reader: OMNIKEY CardMan 5x21 0
Microsoft Base Smart Card Crypto Provider: Missing stored keyset

--------------===========================--------------
================ Certificate 0 ================
--- Reader: OMNIKEY CardMan 5x21 0
---   Card: Identity Device (NIST SP 800-73 [PIV])
Provider = Microsoft Smart Card Key Storage Provider
Key Container = 00000000000000010000000000000000

X509 Certificate:

Here you can see key container has strange value?
My windows login still not working on the other hand when i put same certificate on yubikey it worked.

annu0412:
is there any way to enable  on windows?

dengert:
What  powershell command did you use?


"Microsoft Base Smart Card Crypto Provider" only supports RSA See:  https://docs.microsoft.com/en-us/windows-hardware/drivers/smartcard/smart-card-minidrivers "Microsoft Smart Card Key Storage Provider" is the KSP.

So it looks like it called both as expected. 

The container ID should be unique on the machine so a cached certificate can identify a card so user can be told to insert the card.  But  it looks like container ID was derived from a card unique ID ie.e. the ID private, public and certs share. It could still work. 

What type of card are you using?
If this was a Yubkey, did you create run the yubico-piv-tool with action "set-chuid"?

NIST does not define a serial number for a PIV card. The OpenSC PIV driver  and minidriver uses either the FASC-N or GUID contained in the CHUID to derive a serial number for the card, which can then be used by minidriver to create the Container ID for each cert/key pair. (IIRC the default serial number is all zeros, and the one character ID is then inserted in the middle, thus "00000000000000010000000000000000"

The yubikey minidriver may use the Yubikey token serial number, and thus work. 

If you use the same certificate on different cards windows can get confused as it can only cache one of them with the containerID it saw when first cached. . You will need to delete the cert(s) from the "my" cert store. Easy way is control panel->internet options->content->certificates->personal  then find and remove the old certificates that also contain the container ID. 

As I don't have access to AD, to test login, any additional testing would be helpful. See
https://github.com/OpenSC/OpenSC/wiki/Using-OpenSC  on how to get OpenSC debug logs on Windows.

dengert:
While reviewing this PR and #2575  the minidriver is called at:
https://github.com/OpenSC/OpenSC/pull/2523/files#diff-0472a0d91e9f422e4d90a09074f95d9edf67b14a2df432fcb3aacca29b9a3b39L4424

which calls [md_query_key_sizes](https://github.com/OpenSC/OpenSC/pull/2523/files#diff-0472a0d91e9f422e4d90a09074f95d9edf67b14a2df432fcb3aacca29b9a3b39L2723) with the dwKeySpec 

dwKeySpec  is a combination  of: ECC key, a specific usage (ECDSA or ECDHE), a specific keysize and a specific named curve.
as used starting here:

https://github.com/OpenSC/OpenSC/pull/2523/files#diff-0472a0d91e9f422e4d90a09074f95d9edf67b14a2df432fcb3aacca29b9a3b39L2786

But the OpenSC  looks at `vs->p15card->card->algorithms` for just the keysize.  
https://github.com/OpenSC/OpenSC/pull/2523/files#diff-0472a0d91e9f422e4d90a09074f95d9edf67b14a2df432fcb3aacca29b9a3b39L2782


`struct sc_algorithm_info* algo_info;`  has `parameters`  which for ECC is the  namedCurve OID, which could be tested . But the OID for different supported curves are not filled in by most card drivers when they call `sc_card_add_ec_alg`

To see these, run  in `libopensc`  `grep -A 1 _sc_card_add_ec_alg *`

Microsoft has added additional curves and has defined the OIDs in CNG docs from April 2021  for example:
https://docs.microsoft.com/en-us/windows/win32/seccng/cng-portal
https://docs.microsoft.com/en-us/previous-versions/dn468735(v=vs.85)
https://docs.microsoft.com/en-us/windows/win32/seccng/cng-named-elliptic-curves

In the last reference above they do list multiple curves with the same size. 

When and if these will be passed to a minidriver as `dwKeySpec` is not clear. Given that a certificate in the SPKI has the OID for the curve, they may never need to call the minidrive to get the size, assuming if the cert came from the card, it would support the curve.

dengert:
I am not referring to  CardCreateContainer but to CardQueryKeySizes:
CardQueryKeySizes

which says:
"Key size information is returned in the following structure. For ECC, minimum, default, and maximum are a specific value. Increment is 1."

the OpenSC minidriver  implemantaion in master: https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L4430-L4464
which calls https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L2728-L2835
tries to return a specific values. 

But this only works for some cards. Many do not fill in the vs->p15card->card->algorithms and card drivers do not fill in the OID when calling `sc_card_add_ec_alg`

Jakuje:
I dont have a way to test this, but the only comment I would have here would be the typo marked by the codespell. The remaining codespell reports are fixed already in #2598.

Jakuje (approved):

Jakuje:
@dengert any update on this? Did you manage to test it? Ready to merge?

dengert:
In other words it is ready to commit. It does not break anything, only made improvements.

Jakuje:
Thank you! Merging.

dengert:
If you use OpenSC on Windows, especially if you were having problems with EC keys, can you try  installing OpenSC from https://github.com/OpenSC/Nightly 

Uninstall any current OpenSC  packages both 64 and 32 bit. Then install the nightly versions of the .msi files.

Please report if it works or does not work, along with the card type and ATR.
'opensc-tool -a` can be used to get the ATR.

It is not clear if certificates already registered in the Microsoft Certificate Store will have to be deleted or not.  Previous versions of OpenSC set in registry "Provider = OpenSC" which caused EC keys to not work. #2523 fixes this to use "Provider = Microsoft Base Smart Card Crypto Provider" or "Provider = Microsoft Smart Card Key Storage Provider"

`certutil.exe -v -SCInfo` or a Microsoft browser to a site requeuing  the use of the smart card can be used to verify if #2523 is working.   

Thanks.

dengert:
One more issue is cached ATR that may cause Microsoft minidriver for PIV to
get called.

PR #2526, opened 2022-03-15
Fix registering of mechanisms

xhanulik:
This PR fixes calling `sc_pkcs11_register_sign_and_hash_mechanism()` to take proper mechanism as `sign_type`.

After creating new mechanism in `register_ec_mechanisms()` it can be either registered before `sc_pkcs11_register_sign_and_hash_mechanism()` for `SC_ALGORITHM_ECDSA_HASH_NONE` or not registered. In that case, original `mt` is used as sign_type.

Also, the pointer to `sign_type` in `struct hash_signature_info` is removed from the struct since it is not used anywhere. That also fixes the leaking of `sign_type` in the case when it is not registered before, and therefore not released with `p11card`.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -5794,7 +5794,7 @@ register_gost_mechanisms(struct sc_pkcs11_card *p11card, int flags)
 		if (!mt)
 			return CKR_HOST_MEMORY;
 		rc = sc_pkcs11_register_mechanism(p11card, mt, NULL);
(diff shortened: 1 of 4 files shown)


Jakuje (approved):

Jakuje:
@frankmorgner this is the follow-up from the discussion in  #2521. Can you check if it looks ok to you?

frankmorgner (approved):

PR #2528, opened 2022-03-18
Integrate Packit to test build on Fedora

xhanulik:
Current CI in GitHub actions does not cover testing build on Fedora. This causes subsequent patching issues in Fedora.
[Packit](https://packit.dev/) provides a convenient way to test that via building in Copr. The integration requires configuration file and spec file for RPM.

Config file `.packit.yaml` contains basic configuration - setting spec file and jobs. Created OpenSC spec file `packaging/opensc.spec` is similar to the one on Fedora dist-git except patches that were omitted. The configuration was tested on Fedora.

Packit can be used in some simple way locally - generating srpm ([`packit srpm`](https://packit.dev/docs/cli/srpm/)), local build ([`packit local-build`](https://packit.dev/docs/cli/local-build/)).

To integrate Packit into GitHub actions, it is required to _install Packit-as-a-Service_ into the OpenSC from GitHub Marketplace. After the installation, the project needs to be approved. The whole integration process is described in [Packit documentation](https://packit.dev/docs/packit-service/#integrating-packit-as-a-service-into-your-project-or-organization-from-github-marketplace).

The current `.packit.yaml` config file has only jobs for testing build triggered by pull request. It is also possible to set build on commit into the master branch. Targets are set only to `fedora-development` (can be extended - `fedora-stable`, `epel`, ...).

Other links:
* Packit CLI - https://packit.dev/docs/cli/
* Packit Service - https://packit.dev/docs/packit-service/
* Packit configuration - https://packit.dev/docs/configuration/

Also added fixes after running local test build - removed unused variable and forced static build for `fuzz_pkcs15init` (since the specfile uses `--disable-static`).

### DIFF
--- a/packaging/opensc.spec
+++ b/packaging/opensc.spec
@@ -1,0 +1,201 @@
+%define opensc_module "OpenSC PKCS #11 Module"
+%define nssdb %{_sysconfdir}/pki/nssdb
+
+Name:           opensc
+Version:        0.1.0
+Release:        1%{?dist}
+Summary:        Smart card library and applications
+
+License:        LGPLv2+
+URL:            https://github.com/OpenSC/OpenSC/wiki
+Source0:        opensc-0.1.0.tar.gz
+Source1:        opensc.module
+
+BuildRequires: make
+BuildRequires:  pcsc-lite-devel
+BuildRequires:  readline-devel
+BuildRequires:  openssl-devel
+BuildRequires:  /usr/bin/xsltproc
+BuildRequires:  docbook-style-xsl
+BuildRequires:  autoconf automake libtool gcc
+BuildRequires:  bash-completion
+BuildRequires:  zlib-devel
+# For tests
+BuildRequires:  libcmocka-devel
+%if ! 0%{?rhel}
+BuildRequires:  softhsm
+%endif
+BuildRequires:  openssl
+Requires:       pcsc-lite-libs%{?_isa}
+Requires:       pcsc-lite
+Obsoletes:      mozilla-opensc-signer < 0.12.0
+Obsoletes:      opensc-devel < 0.12.0
+Obsoletes:      coolkey <= 1.1.0-36
+# The simclist is bundled in upstream
+Provides:       bundled(simclist) = 1.5
+
+%description
+OpenSC provides a set of libraries and utilities to work with smart cards. Its
+main focus is on cards that support cryptographic operations, and facilitate
+their use in security applications such as authentication, mail encryption and
+digital signatures. OpenSC implements the PKCS#11 API so applications
+supporting this API (such as Mozilla Firefox and Thunderbird) can use it. On
+the card OpenSC implements the PKCS#15 standard and aims to be compatible with
+every software/card that does so, too.
+
+
+%prep
+%setup -q
+
+# The test-pkcs11-tool-allowed-mechanisms already works in Fedora
+sed -i -e '/XFAIL_TESTS/,$ {
+  s/XFAIL_TESTS.*/XFAIL_TESTS=test-pkcs11-tool-test-threads.sh/
+  q
+}' tests/Makefile.am
+
+
+cp -p src/pkcs15init/README ./README.pkcs15init
+cp -p src/scconf/README.scconf .
+# No {_libdir} here to avoid multilib conflicts; it's just an example
+sed -i -e 's|/usr/local/towitoko/lib/|/usr/lib/ctapi/|' etc/opensc.conf.example.in
+
+
+%build
+autoreconf -fvi
+%ifarch %{ix86}
+sed -i -e 's/opensc.conf/opensc-%{_arch}.conf/g' src/libopensc/Makefile.in
+%endif
+sed -i -e 's|"/lib /usr/lib\b|"/%{_lib} %{_libdir}|' configure # lib64 rpaths
+%set_build_flags
+CFLAGS="$CFLAGS -Wstrict-aliasing=2 -Wno-deprecated-declarations"
+%configure --disable-static\
+  --disable-autostart-items \
+  --disable-notify \
+  --disable-assert \
+  --enable-pcsc \
+  --enable-cmocka \
+  --enable-sm \
+  --with-pcsc-provider=libpcsclite.so.1
+%make_build
+
+
+%check
+make check
+
+
+%install
+%make_install
+install -Dpm 644 %{SOURCE1} $RPM_BUILD_ROOT%{_datadir}/p11-kit/modules/opensc.module
+
+%ifarch %{ix86}
+# To avoid multilib issues, move these files on 32b intel architectures
+rm -f $RPM_BUILD_ROOT%{_sysconfdir}/opensc.conf
+install -Dpm 644 etc/opensc.conf $RPM_BUILD_ROOT%{_sysconfdir}/opensc-%{_arch}.conf
+rm -f $RPM_BUILD_ROOT%{_mandir}/man5/opensc.conf.5
+install -Dpm 644 doc/files/opensc.conf.5 $RPM_BUILD_ROOT%{_mandir}/man5/opensc-%{_arch}.conf.5
+# use NEWS file timestamp as reference for configuration file
+touch -r NEWS $RPM_BUILD_ROOT%{_sysconfdir}/opensc-%{_arch}.conf
+touch -r NEWS $RPM_BUILD_ROOT%{_mandir}/man5/opensc-%{_arch}.conf.5
+%else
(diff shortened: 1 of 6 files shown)


packaging/opensc.spec
```diff
@@ -0,0 +1,201 @@
...
+# Upstream considers libopensc API internal and no longer ships
+# public headers and pkgconfig files.
+# Remove the symlink as nothing is supposed to link against libopensc.
+rm -f $RPM_BUILD_ROOT%{_libdir}/libopensc.so
+# remove the .pc file so we do not confuse users #1673139
+rm -f $RPM_BUILD_ROOT%{_libdir}/pkgconfig/*.pc
+rm -f $RPM_BUILD_ROOT%{_libdir}/libsmm-local.so
+%if 0%{?rhel} && 0%{?rhel} < 7
+rm -rf %{buildroot}%{_datadir}/bash-completion/
+%endif
```
Jakuje, line 117: This is also relevant only for RHEL6 so we do not need it for Fedora. Similar condition can be removed on line 136.

packaging/opensc.spec
```diff
@@ -0,0 +1,201 @@
...
+%{_bindir}/sc-hsm-tool
+%{_bindir}/dnie-tool
+%{_bindir}/westcos-tool
+%{_bindir}/egk-tool
+%{_bindir}/goid-tool
+%{_libdir}/lib*.so.*
+%{_libdir}/opensc-pkcs11.so
+%{_libdir}/pkcs11-spy.so
+%{_libdir}/onepin-opensc-pkcs11.so
+%%dir %{_libdir}/pkcs11
```
Jakuje, line 176: I assume the double-percent is not needed here either.

src/tools/pkcs15-init.c
```diff
@@ -1890,7 +1890,7 @@ init_gost_params(struct sc_pkcs15init_keyarg_gost_params *params, EVP_PKEY *pkey
 		assert(EC_KEY_get0_group(key));
 		nid = EC_GROUP_get_curve_name(EC_KEY_get0_group(key));
 #else
-		assert(EVP_PKEY_get_group_name(pkey, name ,sizeof(name), &name_len));
+		assert(EVP_PKEY_get_group_name(pkey, name ,sizeof(name), NULL));
```
Jakuje, line 1893:
```suggestion
		assert(EVP_PKEY_get_group_name(pkey, name, sizeof(name), NULL));
```

Jakuje:
This requires somebody from @OpenSC/core to install the app to OpenSC in https://github.com/marketplace/packit-as-a-service

This can help us to get better test coverage on PRs on other linux than Ubuntu we have in GH Actions. We use this in various other projects already, but if you have questions or concerns, please let us know.

frankmorgner:
Linux distributions are highly fragmented and it's hard to keep track of all the changes. The OpenSC debian template file has not been touched and it seems not to be used anywhere. By adding a Fedora definition, I fear we will sooner or later also loose track of this.

I wish there was a more generic service that's capable of creating a/multiple package(s) for different distributions...

For example, https://github.com/valentindavid/pkcs11-demo creates a snap that works in different distributions. However, the snap's sandbox restrictions only allow exporting PKCS#11 via the [p11-kit proxy](https://p11-glue.github.io/p11-glue/p11-kit/manual/sharing.html)...

Jakuje:
I agree that Linux distributions are fragmented, but the majority is Debian/Ubuntu (we have some test coverage for that already)  or Fedora/RHEL based. These are just two packaging formats, deb and rpm. Then there are couple of Arch based and Suse (rpm too) for completeness.

If you would go to the containers waters (yeah, snap is just a container with some wiring around), this is something considered cross-platform, but in reality, it works only on Ubuntu. In Fedora/RHEL, we have flatpak, which serves the similar purposes (but can access smart cards on the pcsc layer). But neither of these is very suitable for such low-level library as OpenSC, which is supposed to be used by other applications on the system. Ignoring the stability of the solution itself -- new Ubuntu starter using Firefox in snaps and since then I see it crashing all the time, the access to filesystem is limited and it has only one cursor icon.

But note that adding a spec file is not a goal of this PR. Having the tests running on different configuration (than the always outdated ubuntu image in Github Actions) is the goal.

I think I previously mentioned that there is a service to build for more different distros from Suse, called OpenBuildService, but I see that more like a publishing service, than something we can do CI on. See for example the darktable repository there: https://build.opensuse.org/project/show/graphics:darktable. In any case, if we would get interest from other distros we can indeed try something like that, but in Fedora/RHEL, we are more striving to keep things working together with the rest of the os, rather than having something generic, where one-size-fits-all, because it usually does not, as we could see recently with the `pkcs11-register` in Debian.

frankmorgner:
I don't mind having the packaging stuff here and having it actually maintained is even better :wink:

Regarding the packages that this project publishes, I'd like to keep the user friction low. You named two packages formats, but that needs to be multiplied with the number of possible architectures and that doesn't even consider the possible problems with run time dependencies.  Deciding which download to choose can be quite challenging, which is why I would opt for a one-size-fits-all solution for the releases, *on this website*.

But with regard to the official distribution channels, I fully agree that things should hand in hand. But, again, that's typically managed by a different stakeholder...

https://openbuildservice.org/2021/05/31/scm-integration/ at least looks like it could be integrated into CI some day...

Jakuje:
So getting back to this PR, do you have some concerns about running packit as part of CI? It has at least advantage of running on fresh dependencies (glibc, gcc, ...) and different architectures. I hope last couple of years showed that we in Red Hat are able to maintain opensc :)

If there are no concerns, can somebody add the application (or give me the access to do so?)? I am in the opensc-maintainers, but not in core or owners of the repo to be able to do it myself.

frankmorgner:
I'll try to get you into the correct groups

frankmorgner:
Thanks @xhanulik for setting up the integration!

As far as I can tell, the free plan for packit-as-a-service can only be set up for personal accounts, not for organizations (i.e. OpenSC). @Jakuje, please have a second look, thank you!

TomasTomecek:
@frankmorgner hi Frank, Packit dev here.

Our GitHub app is completely free and can be installed into the whole GitHub organization - most of our users utilize it like this.

I just tried the process for one of the orgs I'm part of:
1. Went to https://github.com/marketplace/packit-as-a-service
2. Navigated to "Pricing and setup"
3. Picked "Account: <the-org>"
4. Clicked "Install it for free" which got me to a new page that confirmed it's indeed $0.00
5. Clicked install
6. Confirmed the organization repositories I want to install the app into

Haven't done this in a while and can completely see how complicated it can feel.

Please let me know if this helps.

P.S. thanks for the pointer to OBS, didn't know about their SCM integration

frankmorgner:
I tried that and as organization the charge would be 5$...

frankmorgner:
As account I've chosen the Organization OpenSC

LudovicRousseau:
I am part of different github organisations. "OpenSC team" is the only one that is not free (for packit) but 5$ as mentioned by @frankmorgner 

I also see a "Current monthly bill $5.00" in https://github.com/organizations/OpenSC/settings/billing

I guess it is because we switched from the free github account a few years back to get higher download quotas or something like that.

TomasTomecek:
Wow, this is the first time I see an upper-tier GitHub org. Please let me know if I can help in any way because this looks like it's outside of our control.

Jakuje:
> I guess it is because we switched from the free github account a few years back to get higher download quotas or something like that.

AFAIK this was because of the macos build artifacts that were larger than the 50MB (?) limit. But I think in the end we switched back and started splitting the binaries to the the 50MB chunks in https://github.com/OpenSC/OpenSC/blob/master/.github/push_artifacts.sh#L20

So are we actually in some different plan than the default/free one? Given that we are still open-source project, I do not see why we should be charted for using packit. Is there some channel to raise the issue with github for clarifications as a packit package owner?

TomasTomecek:
I can see that GitHub has community support: https://support.github.com/features/organizations

Are you also being charged for using other apps? I.e. AppVeyor and LGTM analysis

Jakuje:
@frankmorgner can you clarify if we still have the non-free github organization and if there is still a reason for that? I would be happy to look into that issue, but the state of the OpenSC organization is still not completely clear to me.

Jakuje:
@TomasTomecek is there a way to integrate the packit without the installing the app, through the github actions?

We are mostly interested in running build based on the spec file in Fedora anyway, rather than in other more advanced features Packit provides.  I read briefly through the integration options and given that there is a integration with the gitlab just with the Gitlab CI, it should not be that complicated.

My question is somebody tried this before and if I miss something obvious that would prevent it from working (as it looks like the installation of the app is not an option here).

PR #2403, opened 2021-09-07
Epass2003 init

dengert:
When selecting a file from the token, recreated the OpenSC acls from the sec_attrs which are the acls use by the token. 

Fixes #2397 

https://github.com/OpenSC/OpenSC/commit/ce7fa4295896ad71d14fdd00a592217924c2fe6f introduced code that replaced using the file from the profile to use the file from the token. The file from the token did not have the `file->acl`. The `epass2003_process_fci` now recreates the `file ->acl` 

Tested with- pkcs15-init using Feitian epass2003:
- ./pkcs15-init -E
- ./pkcs15-init --create-pkcs15 --profile pkcs15+onepin --use-default-transport-key --pin 12345678 --puk entersafe  --label "A label"
- /pkcs15-init --generate-key rsa/2048 --key-usage sign,decrypt --auth-id 01 --label "RSA1"

Issue #2397, opened 2021-08-26
OpenSC 0.22 newly fails to initialize a Feitian ePass2003 Auto USB device


muks:
### Problem Description

After upgrading from `opensc-0.21.0-4.fc34.x86_64` to `opensc-0.22.0-1.fc34` on Fedora 34, I am unable to initialize a Feitian ePass2003 Auto USB device using `pkcs15-init`. The error that I'm newly seeing is:

```
Failed to create PKCS #15 meta structure: EF offset too large
```

This used to work fine with `opensc-0.21.0-4.fc34.x86_64`, and reverting back to the older package version makes it work again. There were no other package changes on the machine.

### Proposed Resolution

None from my side (I'm just an end user). The `pkcs15-init` command should not error.

### Steps to reproduce

The commands used to initialize are:
```
# Erase the smartcard
pkcs15-init -E

# Initialize the smartcard (this prints an error now in opensc-0.22.0-1.fc34)
pkcs15-init --create-pkcs15 --profile pkcs15+onepin --use-default-transport-key --pin 1234 --puk 12345678 --label "Token Label"
```

The device is:
```
Bus 004 Device 015: ID 096e:080a Feitian Technologies, Inc. ePass2003Auto
```

I currently do not have an unused device to capture more information right now - we issued the devices we had. I should be receiving more devices tomorrow if any additional information is desired.

Thank you.

dengert:
I have been looking at this issue and #2398 this weekend using my old epass2003 from December 2014. (This was from Gooze.) 

(I suggest for now, uses a shorter label.)

At the start I found 0.21.0 as installed on Ubuntu-20.04 both 
 `OPENSC_DEBUG=7  OPENSC_DRIVER=epass2003 pkcs15-init -E` and
` pkcs15-init --create-pkcs15 --profile pkcs15+onepin --use-default-transport-key --pin 123456 --puk entersafe --label "1"` worked, and I was able to use `pkcs15-tool ---list-info`

But after some debugging and getting ready to run git bisect. things got worse. 

I believe this is related to: https://github.com/OpenSC/OpenSC/commit/5a369a8f3

I am getting the same "Failed to erase card: Security status not satisfied` message from 0.21.00 `pkcs15-init -E` that worked before. 

I even tried the FIX_tool from Feitian  as listed in the above commit. It appeared to work at least twice.
But nothing is working now. 

And my epass2003 appears to be bricked. I suspect erase card may be losing memory and not really erasing the card. 

[...]

Issue #1737, opened 2019-07-13
Aventra MyEID: pkcs15-init --create-pkcs15 - Failed to format card: User PIN required


minfrin:
### Problem Description

When attempting to format a MyEID card that had been previously erased, I am asked for a user PIN, which makes no sense.

```
Little-Net:~ minfrin$ pkcs15-init --create-pkcs15 --use-default-transport-keys --profile pkcs15+onepin --label "Mercury" --so-pin 1111 --so-puk 1111
Using reader with a card: ACS ACR 38U-CCID
User PIN required.
Please enter User PIN: 
User PIN required.
Please enter User PIN: 
User PIN required.
Please enter User PIN: 
```

### Proposed Resolution

Don't ask for a user pin when formatting a card.

frankmorgner:
@popovec ?

popovec:
From https://github.com/OpenSC/OpenSC/wiki/Aventra-MyEID-PKI-card

```
TIP: When initializing cards, specify the PIN and PUK (with “—pin” and “—puk” parameters) to prevent OpenSC from unnecessarily asking for it several times. You can use any values, because the PIN is not created here.
pkcs15-init -C --pin 1111 --puk 1111
```

But ok, this workaround is wrong, lifecycle information from card can be used to prevent this. I'm not sure whether the repair will be simple, pkcs15-init is common to all cards and lifecycle information is not available from each card.

popovec:
I have solution for this issue, but can somebody review `sc_pkcs15init_update_file()` code?  I need lifecycle information from selected file, but this information is not always available. `sc_pkcs15init_authenticate()` is called with `file` but `file` in some cases points to wrong structure (incomplete structure).  I think it's better to call `sc_pkcs15init_authenticate()` with the `selected_file` parameter.


```
diff --git a/src/pkcs15init/pkcs15-lib.c b/src/pkcs15init/pkcs15-lib.c
index 830c2007..27e1c050 100644
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -3874,6 +3874,10 @@ sc_pkcs15init_authenticate(struct sc_profile *profile, struct sc_pkcs15_card *p1
                        sc_log(ctx, "unknown acl method");
                        break;
                }
+               else if (file->status == SC_FILE_STATUS_CREATION) {
+                       sc_log(ctx, "File id (%X) status SC_FILE_STATUS_CREATION", file->id);
[...]

PR #2529, opened 2022-03-27
Update pkcs15-lib.c

xaqfan:
It will verify the pin or not  by arguemnt alc  of file(selected_file). Otherwise it will return fail ,when update the binary file in function sc_pkcs15init_authenticate.
I use ePass2003 hardware by feitian.
now i change the argument from selected_filed to file.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -4147,7 +4147,7 @@ sc_pkcs15init_update_file(struct sc_profile *profile,
 	}
 
 	/* Present authentication info needed */
-	r = sc_pkcs15init_authenticate(profile, p15card, selected_file, SC_AC_OP_UPDATE);
+	r = sc_pkcs15init_authenticate(profile, p15card, file, SC_AC_OP_UPDATE);
 	if (r >= 0 && datalen)
 		r = sc_update_binary(p15card->card, 0, (const unsigned char *) data, datalen, 0);
 


popovec:
This commit attempts to revert the changes listed in ce7fa4295896ad71d14fdd00a592217924c2fe6f

If this commit is applied, the card initialization operation will be violated for MyEID cards (commit f46b61739711f400fe28c7274c8b49db412a0515 removed the need to use a PIN during initialization, this was reported as an issue https://github.com/OpenSC/OpenSC/issues/1737)

Jakuje:
Can you clarify what is the issue you see with your card and why it needs the different file?

dengert:
What version of OpenSC are you using? Can you build from master?

Also look at https://github.com/OpenSC/OpenSC/issues/2397 where https://github.com/OpenSC/OpenSC/commit/ce7fa4295896ad71d14fdd00a592217924c2fe6f is referenced 
and https://github.com/OpenSC/OpenSC/pull/2403 to fix epass2003  problems with acls.

#2403 was merged on Jan 25, so you may need to build from master.

Could the problem be resolved by modifying one of these (and maybe others):  https://github.com/OpenSC/OpenSC/blob/master/src/pkcs15init/epass2003.profile
and https://github.com/OpenSC/OpenSC/blob/master/src/pkcs15init/myeid.profile ?

dengert:
@xaqfan did https://github.com/OpenSC/OpenSC/pull/2403 solve your problem?

If not, I have an Epass2003.  Can you list the steps to recreate your problem? Please test with git master.

PR #2534, opened 2022-04-05
Fix issues reported by OSS-Fuzz

xhanulik:
PR fixes some recent OSS-Fuzz issues, mainly leaks and buffer-overflows. 

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/profile.c
+++ b/src/pkcs15init/profile.c
@@ -2491,7 +2491,10 @@ expr_eval(struct num_exp_ctx *ctx, unsigned int *vp, unsigned int pri)
 		expr_eval(ctx, &right, new_pri + 1);
 		switch (op) {
 		case '*': left *= right; break;
-		case '/': left /= right; break;
+		case '/':
+			if (right == 0)
+				expr_fail(ctx);
+			left /= right; break;
 		case '+': left += right; break;
 		case '-': left -= right; break;
 		case '&': left &= right; break;
--- a/src/pkcs15init/profile.c
+++ b/src/pkcs15init/profile.c
@@ -2359,8 +2359,9 @@ __expr_get(struct num_exp_ctx *ctx, int eof_okay)
 	}
 
 	ctx->j = 0;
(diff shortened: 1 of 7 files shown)


Jakuje (approved):

src/pkcs15init/profile.c
```diff
@@ -2359,8 +2359,9 @@ __expr_get(struct num_exp_ctx *ctx, int eof_okay)
 	}
 
 	ctx->j = 0;
+	s = ctx->str;
 	do {
-		if ((s = ctx->str) == NULL || *s == '\0') {
+		if (s == NULL || *s == '\0') {
```
Jakuje, line 2365: I am not sure if this does not change the meaning of the code.

The `ctx->str` is set also on line 2369 and your change and then only the `s` is moved. In next iteration, the `s` is not reset to the `ctx->str`, which is a difference.

In any case, it would probably make sense to write some unit tests for this code and figure out what it is actually supposed to be doing. The code was written almost 20 years ago.

It is also not completely clear to me what is timing out in this case in the fuzzer.

xhanulik: The setting of `ctx->str` at the beginning of the cycle is actually the problem causing the timeout - for example, if `ctx->str` initially contains only whitespace (and  `\0` for termination), the `s` is always set back to `ctx->str` in the `if` (before the check for `*s == '\0'`) even though the whitespaces were removed on 2374.

I think that the cycle here should get the value of `s` without additional prepended spaces, so the `s` is first set to `ctx->str`, then it can get value from `ctx->argv`, and the spaces remain removed also in the next iteration.

Jakuje: Makes sense. Thank you for the clarification.

Jakuje:
Tested locally both piv test and builds with openssl 3 and all look good. Not sure what the github does not like in there, but it if is not going to give us logs, there is nothing we could fix :/

PR #2539, opened 2022-04-14
Minor issues either reported by coverity or noticed while reviewing ci results

Jakuje:
Minor issues either reported by coverity or noticed while reviewing ci results

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/tests/fuzzing/fuzz_card.c
+++ b/src/tests/fuzzing/fuzz_card.c
@@ -38,7 +38,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 	uint8_t           len = 0;
 	u8               *rnd = NULL, *wrap_buf = NULL, *unwrap_buf = NULL;
 	size_t            wrap_buf_len = 0, unwrap_buf_len = 0;
-	int               reset = 0;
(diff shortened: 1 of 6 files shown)


src/tests/fuzzing/fuzz_card.c
```diff
@@ -86,8 +86,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 	sc_list_files(card, files, sizeof(files));
 
 	/* Get challenge */
-	if ((rnd = malloc(len)))
-	sc_get_challenge(card, rnd, len);
+	rnd = malloc(len);
+	if (rnd == NULL)
+		goto err;
+	if (sc_get_challenge(card, rnd, len) != SC_SUCCESS)
```
xhanulik, line 92: Is it necessary to check whether `sc_get_challenge()` fails here? The return values are not tested for the other functions since they are independent.

Jakuje: coverity complains about not checking return values. I think just logging something in the condition might be enough to make coverity happy

xhanulik: Ok, looks good.

PR #2540, opened 2022-04-19
Unbreak builds with --disable-static which now fail in master on osx

Jakuje:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -49,9 +49,11 @@ else
 
 	if [ "$1" == "ossl3" -o "$2" == "ossl3" ]; then
 		# without -Werror, because of rest of deprecated API
-		./configure  --disable-dependency-tracking --disable-strict CFLAGS="-Wall -Wextra -Wno-unused-parameter -Wstrict-aliasing=2"
+		./configure --disable-dependency-tracking --disable-strict CFLAGS="-Wall -Wextra -Wno-unused-parameter -Wstrict-aliasing=2"
+	elif [ "$1" == "no-shared" ]; then
+		./configure --disable-shared
 	else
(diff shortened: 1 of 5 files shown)


Jakuje:
Just for the curious, I did a diff of the last working osx run and the failing and the difference was like this:

```
  1   CCLD     opensc-pkcs11.la                                                                                               |    CCLD     opensc-pkcs11.la
  2 ../../libtool: line 2639: func__fatal_error: command not found                                                            |  libtool:   error: cannot build a shared library See the libtool documentation for more information. Fatal configuration error.
  3 libtool: warning: '/Users/runner/work/OpenSC/OpenSC/openpace_bin/Library/OpenSC/lib/libeac.la' seems to be moved          |  make[3]: *** [opensc-pkcs11.la] Error 1
  4 libtool: warning: library '/Users/runner/work/OpenSC/OpenSC/openpace_bin/Library/OpenSC/lib/libeac.la' was moved.         |  make[2]: *** [all-recursive] Error 1
  5   CC       pkcs11_spy_la-pkcs11-spy.lo                                                                                    |  make[1]: *** [all-recursive] Error 1
  6   CC       pkcs11_spy_la-pkcs11-display.lo                                                                                |  make: *** [all] Error 2
  7   CCLD     pkcs11-spy.la                                                                                                  |  ##[error]Process completed with exit code 2.

```
The old/pass was failing on unknown fail function and the new/fail was correctly failing so I assume this is something that was recently fixed in the osx image, which undercovered the problem that the `--disable-shared` build never worked in OpenSC.

frankmorgner (approved):

PR #2542, opened 2022-04-20
card-epass2003.c

xaqfan:
To support feitian epass2003 keys with  cos of fips.

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -41,6 +41,10 @@
 
 #include <openssl/evp.h>
 #include <openssl/sha.h>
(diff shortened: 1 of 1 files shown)


frankmorgner (changes requested):
Your code often lacks error checking and in some cases, even removes the existing error checking. Some changes are modifying magic numbers, which look like they are breaking support of existing cards/configurations...

xaqfan:
> Your code often lacks error checking and in some cases, even removes the existing error checking. Some changes are modifying magic numbers, which look like they are breaking support of existing cards/configurations...

Sorry ,I will check and  resubmit it .
now  close this.

PR #2547, opened 2022-04-29
Unbreak openssl 3.0 and libressl pipelines

Jakuje:
I think the main problem with these pipelines was the removal of openssl which caused some watchdog to not work and the builds were timing out.

This also updates the actions to v3, cleans up some of the pipelines and hides a lot of unneeded output from the openssl/libressl build.

I also tried to update libressl ot 3.5.2 but build was failing for some reason. This might be something to explore further or I will open a new issue for this.

### DIFF
--- a/.github/setup-libressl.sh
+++ b/.github/setup-libressl.sh
@@ -4,7 +4,7 @@ set -ex -o xtrace
 
 V=libressl-3.4.2
 
-sudo apt-get remove -y openssl libssl-dev java8-runtime-headless default-jre-headless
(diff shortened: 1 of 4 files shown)


Jakuje:
@frankmorgner @xhanulik It would be great to have this merged so we can test the new/outstanding PRs against the OpenSSL 3.0 and libressl too.

frankmorgner (approved):

xhanulik:
Looks good to me.

Jakuje:
Thank you for review.

PR #2550, opened 2022-05-09
Fuzz target for PKCS#11 API

xhanulik:
This PR adds a fuzz target for PKCS#11 API. To simplify and shorten the fuzzing input, the target tests only one operation (procedure) at a time regarding the first byte of the fuzzing input.
It is link statically with libopensc-pkcs11, PKCS#11 functions are accessed with `pkcs11_function_list_3_0` (as it is done in `pkcs11-tool`).

First, fuzz target extracts needed arguments from fuzzing input (such as PINs and labels) or fills the key template (needed, for example, for key generation). The bytes in the input data decide whether different options will be used.

It uses the same reader object as was introduced in `fuzz_pkcs15_reader`. The initialization is done via `C_Initialize`, after which the slots are removed and supplied with 'virtual reader'. The procedure of connecting the card is similar to the `card_detect()` function.

The corpus was created directly from fuzz target - when setting the `FUZZING` macro to 0, usage of the 'virtual reader' is omitted, and the APDU can be obtained as described in [README](https://github.com/OpenSC/OpenSC/tree/master/src/tests/fuzzing#how-to-generate-corpus-files-from-existing-cards) file.

##### Checklist


src/tests/fuzzing/fuzz_pkcs11.c
```diff
@@ -0,0 +1,1076 @@
...
+	p11->C_WrapKey(session, &mech, hWrappingKey, hkey, pWrappedKey, &pulWrappedKeyLen);
+
+fin:
+	p11->C_CloseSession(session);
+	p11->C_Finalize(NULL);
+end:
+	free(pin);
+}
+
+#define FILL_ATTR(attr, typ, val, len) {(attr).type=(typ); (attr).pValue=(val); (attr).ulValueLen=len;}
```
Jakuje, line 625: please, expand this macro to multiple lines to make it more readable.

xhanulik: Ok

src/tests/fuzzing/Makefile.am
```diff
@@ -44,3 +50,8 @@ fuzz_pkcs15_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 fuzz_pkcs15_crypt_SOURCES = fuzz_pkcs15_crypt.c fuzzer_reader.c fuzzer_tool.c $(ADDITIONAL_SRC) \
 							../../tools/util.c
 fuzz_pkcs15_crypt_LDADD = $(OPTIONAL_OPENSSL_LIBS)
+fuzz_pkcs11_SOURCES = fuzz_pkcs11.c fuzzer_reader.c fuzzer_tool.c $(ADDITIONAL_SRC)
+fuzz_pkcs11_LDADD = \
+	$(top_builddir)/src/common/libpkcs11.la \
+	$(OPTIONAL_OPENSSL_LIBS) $(PTHREAD_CFLAGS) \
```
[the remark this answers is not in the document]
xhanulik: Fixed

src/tests/fuzzing/Makefile.am
```diff
@@ -18,6 +19,11 @@ if ENABLE_OPENSSL
 noinst_PROGRAMS += fuzz_piv_tool
 endif
 
+if ENABLE_SHARED
+else
+noinst_PROGRAMS += fuzz_pkcs11
+endif
```
Jakuje, line 25: would negative condition like this work?
```suggestion
if ! ENABLE_SHARED
noinst_PROGRAMS += fuzz_pkcs11
endif
```
Is there something preventing the build when the shared build is enabled? Should not this be in `if ENABLE_STATIC` above?

xhanulik: Yes, it can be simplified, but for the second comment, the fuzz target is linked with `libopensc-pkcs11.la`, which is built only when the shared build is not enabled (this part is similar to [`pkcs11-tool`](https://github.com/OpenSC/OpenSC/blob/master/src/tools/Makefile.am#L64-L68)).

Jakuje: Thanks fr clarification.

Jakuje (approved):

PR #2556, opened 2022-05-20
pkcs11-session: fix C_GetSessionInfo in 'atomic' mode

fabled:
The device is always in logged-out mode with PIN cached, so ignore
the logged-in check.

Since the token is now guaranteed to be logged-in or have sessions
removed, simplify the further checks to match the session state.

Fixes: 4bd8cda96604 "pkcs11-session: When we notice logout in lower layers..."

### DIFF
--- a/src/pkcs11/pkcs11-session.c
+++ b/src/pkcs11/pkcs11-session.c
@@ -252,7 +252,6 @@ CK_RV C_GetSessionInfo(CK_SESSION_HANDLE hSession,	/* the session's handle */
 	CK_RV rv;
 	struct sc_pkcs11_session *session;
 	struct sc_pkcs11_slot *slot;
-	int logged_out;
 	const char *name;
 
 	if (pInfo == NULL_PTR)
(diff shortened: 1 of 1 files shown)


Jakuje:
Can you clarify what you mean by "atomic" mode and how does it work?

fabled:
> Can you clarify what you mean by "atomic" mode and how does it work?

I suggest reading opensc documentation on this configuration variable. https://github.com/OpenSC/OpenSC/blob/5d6f7cbfef47746fe60ccb5e89ed6101ac20f543/doc/files/opensc.conf.5.xml.in#L1479

Jakuje:
Thank you for clarification. I never saw this option over the last years, but fortunately it is well documented. It would be great if we could have some automated test for that too to make sure it keeps working.

I retested with the steps I used for the implementation of the previously mentioned commit in and all looks working ok with and without the `atomic` option enabled so I think we are good here.

https://github.com/OpenSC/OpenSC/issues/1822#issuecomment-548416241

Thank you for having a look into this!

Jakuje (approved):

Jakuje (approved):

PR #2574, opened 2022-07-04
OSS-Fuzz issues

xhanulik:
Fixes bugs recently reported by OSS-Fuzz

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-pubkey.c
+++ b/src/libopensc/pkcs15-pubkey.c
@@ -638,7 +638,7 @@ sc_pkcs15_decode_pubkey_ec(sc_context_t *ctx,
 		const u8 *buf, size_t buflen)
 {
 	int r;
-	u8 * ecpoint_data;
+	u8 * ecpoint_data = NULL;
 	size_t ecpoint_len;
 	struct sc_asn1_entry asn1_ec_pointQ[C_ASN1_EC_POINTQ_SIZE];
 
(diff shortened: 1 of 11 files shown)


src/libopensc/pkcs15-pubkey.c
```diff
@@ -646,11 +646,17 @@ sc_pkcs15_decode_pubkey_ec(sc_context_t *ctx,
 	sc_copy_asn1_entry(c_asn1_ec_pointQ, asn1_ec_pointQ);
 	sc_format_asn1_entry(asn1_ec_pointQ + 0, &ecpoint_data, &ecpoint_len, 1);
 	r = sc_asn1_decode(ctx, asn1_ec_pointQ, buf, buflen, NULL, NULL);
-	if (r < 0)
+	if (r < 0) {
+		if (ecpoint_data)
+			free(ecpoint_data);
```
Jakuje, line 651: the `if` check is not needed here as `free(NULL)` is completely valid. But please initialize the pointer to NULL on line 641 to make sure it is really initialized in all cases.

The same below.

src/pkcs15init/profile.c
```diff
@@ -1431,6 +1431,9 @@ do_content(struct state *cur, int argc, char **argv)
 	}
 	file = cur->file->file;
 
+	if (file->encoded_content)
+		free(file->encoded_content);
```
Jakuje, line 1435: no need for `if`

Jakuje (approved):

src/tools/pkcs15-tool.c
```diff
@@ -1961,8 +1961,11 @@ static int test_update(sc_card_t *in_card)
 			}
 			/* other tag */
 			i += 2 + rbuf[2+i+1]; /* length of this tag*/
+			if (2+i+1 >= sizeof(rbuf)) {
+				break;
+			}
 		}
-		if (rbuf[2+i+1] < 9 || 2+i+2+9 > apdu.resplen) {
+		if (2+i+1 >= sizeof(rbuf) || rbuf[2+i+1] < 9 || 2+i+2+9 > apdu.resplen) {
```
Jakuje, line 1968: Having another look, I think the comparison here as well as above should be done against the `apdu.resplen` instead of `sizeof(rbuf)` to capture also the cases where the fuzzer returns less data than the size of `rbuf` and the rest of the `rbuf` stays uninitialized.

xhanulik: Indeed, fixed

Jakuje:
Thank you. Merging to move on with the fuzzing.

PR #2578, opened 2022-07-18
Fix OSS-Fuzz and Coverity issues

xhanulik:
Fixing bugs reported by OSS-Fuzz and recent issues from Coverity.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/sm/sm-common.c
+++ b/src/sm/sm-common.c
@@ -519,7 +519,7 @@ sm_encrypt_des_cbc3(struct sc_context *ctx, unsigned char *key,
 	       data_len, sc_dump_hex(data, data_len));
 
 	*out_len = data_len;
-	*out = malloc(data_len + 8);
+	*out = calloc(data_len + 8, sizeof(unsigned char));
 	if (*out == NULL) {
 		free(data);
 		LOG_TEST_RET(ctx, SC_ERROR_OUT_OF_MEMORY, "SM encrypt_des_cbc3: failure");
--- a/src/tools/pkcs15-tool.c
+++ b/src/tools/pkcs15-tool.c
@@ -1377,7 +1377,7 @@ static int test_session_pin(void)
 	unsigned int  auth_method;
 	unsigned char		*pin;
 	int r;
-	unsigned char sessionpin[SC_MAX_PIN_SIZE];
+	unsigned char sessionpin[SC_MAX_PIN_SIZE] = {0};
 	size_t sessionpinlen = sizeof sessionpin;
 
 	if (!opt_auth_id)   {
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -877,6 +877,7 @@ construct_data_tlv(struct sc_card *card, struct sc_apdu *apdu, unsigned char *ap
 	size_t tlv_more;	/* increased tlv length */
 	unsigned char iv[16] = { 0 };
 	epass2003_exdata *exdata = NULL;
+	int r = 0;
 
 	if (!card->drv_data) 
 		return SC_ERROR_INVALID_ARGUMENTS;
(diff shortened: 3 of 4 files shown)


frankmorgner:
Two runs are failing due to 
```
error while loading shared libraries: libcrypto.so.3: cannot open shared object file: No such file or directory
```
which seems to be unrelated

src/libopensc/card.c
```diff
@@ -412,9 +412,23 @@ int sc_disconnect_card(sc_card_t *card)
...
+	
 	if (card->lock_count != 0)
-		return SC_ERROR_NOT_ALLOWED;
+		LOG_FUNC_RETURN(ctx, SC_ERROR_NOT_ALLOWED);
+
+	sc_force_disconnect_card(card);
+	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+}
+
+int sc_force_disconnect_card(sc_card_t *card)
```
frankmorgner, line 423: I'm not sure if handling this case globally is the right approach. Your commit message states the following:
>When performing context specific login with PIV card,
PIV driver does one extra sc_lock() call and expects
sc_unlock() to be called by following cryptographic operation.
If an error occurs in between, sc_disconnect_card will be called,
but will not release the card due to remaining locks.

So if I understand correctly, the additional lock is set and unset in different call backs of the PIV driver where `priv->context_specific` is handled. In this case, the card driver would be responsible for releasing resources that it has allocated. At the very least, `piv_finish()` could detect and release this additional lock, right @dengert ?

frankmorgner (changes requested):

src/libopensc/card.c
dengert: It could work if  https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L3502-L3507 was copied 
after https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L3008

But `sc_card_disconnect`  needs these lines switched: 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L416-L417
need to be after these lines:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card.c#L416-L417

In the general case, if `sc_card_disconnect` is called with `card->lock_count != 0` should it just release the locks?

xhanulik: Thank you for looking into this. As suggested, I removed the lock count check and added the unlock into `piv_finish`

frankmorgner: thank you

frankmorgner (approved):

src/sm/sm-common.c
```diff
@@ -524,6 +524,7 @@ sm_encrypt_des_cbc3(struct sc_context *ctx, unsigned char *key,
 		free(data);
 		LOG_TEST_RET(ctx, SC_ERROR_OUT_OF_MEMORY, "SM encrypt_des_cbc3: failure");
 	}
+	memset(*out, 0, data_len + 8);
```
Jakuje, line 527: would not it be easier to use `calloc()` instead of `malloc()` on line 522?

Jakuje (commented):
Some minor improvements can be done, but generally looks good. I will probably manually merge the other commits so we can move on with the fuzzing a bit.

Jakuje:
Pushed manually most of the commits. Please, rebase on top of current master to fix the mingw builds and resolve the conflict.

src/sm/sm-common.c
xhanulik: Fixed, thanks for review.

src/tools/pkcs15-tool.c
```diff
@@ -1380,6 +1380,8 @@ static int test_session_pin(void)
 	unsigned char sessionpin[SC_MAX_PIN_SIZE];
 	size_t sessionpinlen = sizeof sessionpin;
 
+	memset(sessionpin, 0, SC_MAX_PIN_SIZE);
+
```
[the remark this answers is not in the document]
Jakuje: last nit would be bogus newline added here. Otherwise good.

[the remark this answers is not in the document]
xhanulik: indeed, removed

Jakuje:
Thank you. Looks good.

PR #2579, opened 2022-07-18
Do not require a R/W session for --login, ...

### LINKED ISSUE #2182: pkcs11-tool '--login' option always tries to create a read/write session
### Problem Description

When the `--login` option is specified for `pkcs11-tool`, it always sets the the flag `CKF_RW_SESSION` in the `flags` parameter for the `C_OpenSession` call:

https://github.com/OpenSC/OpenSC/blob/f8af905574a55036f7c8e079188992be10ceb73e/src/tools/pkcs11-tool.c#L731-L734
https://github.com/OpenSC/OpenSC/blob/f8af905574a55036f7c8e079188992be10ceb73e/src/tools/pkcs11-tool.c#L1076-L1085

Because of this behavior it is impossible to login into a write-protected token. The concrete use case is to list all objects on a write-protected token including PKCS#11 private objects (having attribute `CKA_PRIVATE` set to `CK_TRUE`), which are visible only after authenticating to the token.

### Proposed Resolution

The presence of the `--login` option on the command line of `pkcs11-tool` should not automatically trigger the creation of a read/write session. A PKCS#11 session is needed but a read-write session isn't required per se. For example signing and listing of objects is allowed in read-only sessions.

### Steps to reproduce

```
$ pkcs11-tool --module <path to PKCS#11 library> -O --login
Using slot 0 with a present token (0x3e8)
error: PKCS11 function C_OpenSession failed: rv = CKR_TOKEN_WRITE_PROTECTED (0xe2)
Aborting.
```

ulrichb:
This allows to perform commands like `--login --pin XXX --list-objects` or `--sign ...` on a read-only token (which otherwise could return an  `CKR_TOKEN_WRITE_PROTECTED` error when calling `C_OpenSession()`).

Fixes #2182.

### Open questions
* [x] Should we switch more operations to `NEED_SESSION_RO`? Yes:
  * `--verify`
  * `--encrypt` and `--decrypt`
  * `--wrap`
* [x] Should we add a `--session-rw` flag as discussed in https://github.com/OpenSC/OpenSC/issues/2182#issuecomment-756239151?
    * DONE

### Checklist

- [n/a] Documentation is added or updated
- [n/a] New files have a LGPL 2.1 license statement
  - With our [SignPath](https://about.signpath.io/) PKCS#11 Cryptoprovider
- [n/a] Windows minidriver is tested
- [n/a] macOS tokend is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,6 +18,13 @@ jobs:
       - uses: actions/checkout@v2
       - run: .github/setup-linux.sh
       - run: .github/build.sh dist
+      - name: Upload test logs
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: ubuntu-test-logs
+          path:
+            tests/*.log
       - uses: actions/cache@v2
         id: cache-build
         with:
@@ -29,12 +36,6 @@ jobs:
           name: opensc-build
           path:
             opensc*.tar.gz
-      - name: Upload test logs
-        uses: actions/upload-artifact@v2
-        with:
-          name: ubuntu-test-logs
-          path:
-            tests/*.log
 
   build-no-shared:
     runs-on: ubuntu-latest
@@ -56,17 +57,18 @@ jobs:
       - uses: actions/checkout@v2
       - run: .github/setup-linux.sh
       - run: .github/build.sh
+      - name: Upload test logs
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: ubuntu-18-test-logs
+          path:
+            tests/*.log
       - uses: actions/cache@v2
         id: cache-build
         with:
           path: ./*
           key: ${{ runner.os }}-18-${{ github.sha }}
-      - name: Upload test logs
-        uses: actions/upload-artifact@v2
-        with:
-          name: ubuntu-18-test-logs
-          path:
-            tests/*.log
 
   build-mingw:
     runs-on: ubuntu-latest
@@ -195,8 +197,9 @@ jobs:
       - run: .github/setup-linux.sh ossl3
       - run: .github/build.sh dist ossl3
       - uses: actions/upload-artifact@v3
+        if: failure()
         with:
-          name: logs
+          name: openssl-3-logs
           path: |
             tests/*.log
             config.log
@@ -250,8 +253,9 @@ jobs:
       - run: .github/setup-linux.sh libressl
       - run: .github/build.sh dist libressl
       - uses: actions/upload-artifact@v3
+        if: failure()
         with:
-          name: logs
+          name: libressl-logs
           path: |
             tests/test-suite.log
             config.log
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -236,6 +236,13 @@
 
 				<varlistentry>
 					<term>
+						<option>--session-rw</option>,
+					</term>
+					<listitem><para>Forces to open the PKCS#11 session with CKF_RW_SESSION.</para></listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>
 						<option>--login</option>,
 						<option>-l</option>
 					</term>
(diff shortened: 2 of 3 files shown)


Jakuje:
I would be for creating a new switch `--session-rw` and trying to use RO session for at least the three operations you mentioned to 1) have a way back if the pkcs11 module does not work without the RW 2) try to use least possible permissions by default.

ulrichb:
@Jakuje Both done.

ulrichb:
@Jakuje Now ... also change to `NEED_SESSION_RO` for:
* `--wrap`
* `--unwrap`
* `--test`
?

popovec:
The `unwrap` operation is used to store wrapped key to card. For this operation RW session is necessary.

ulrichb:
@popovec Okaym then consequently also `--test` because it also does a "unwrap" operation. What about `--wrap`?

ulrichb:
@popovec Okay, also changed `--wrap` to RO, and left `--test` with RW as it will need unwrap after #1796.

Jakuje:
This looks like it breaks the tests in https://github.com/OpenSC/OpenSC/runs/7646167936?check_suite_focus=true

If I read the output right, its the `FAIL: test-pkcs11-tool-sign-verify.sh` and `FAIL test-pkcs11-tool-unwrap-wrap-test.sh`, but I did not check details. The test runs some pkcs11-tool operations against softhsm which might (wrongly?) require the RW session for some of the operations (or it already fails during setup?). Can you check locally?

popovec:
> This looks like it breaks the tests in https://github.com/OpenSC/OpenSC/runs/7646167936?check_suite_focus=true
> 
> If I read the output right, its the `FAIL: test-pkcs11-tool-sign-verify.sh` and `FAIL test-pkcs11-tool-unwrap-wrap-test.sh`, but I did not check details. The test runs some pkcs11-tool operations against softhsm which might (wrongly?) require the RW session for some of the operations (or it already fails during setup?). Can you check locally?

Interesting, only `build-openssl-3` is falling, same tests in `build-libressl` and `build-clang-tidy`  went without errors.

ulrichb:
Hmm, ... cannot reproduce the failing test, ... ran `make check` on Ubuntu 22.04 with OpenSSL 3.0.2 with the following result.

```plain
PASS: test-pkcs11-tool-sign-verify.sh
XFAIL: test-pkcs11-tool-test.sh
XFAIL: test-pkcs11-tool-test-threads.sh
PASS: test-pkcs11-tool-allowed-mechanisms.sh
PASS: test-pkcs11-tool-sym-crypt-test.sh
PASS: test-pkcs11-tool-unwrap-wrap-test.sh
```

ulrichb:
... can we get somehow the `test-pkcs11-tool-sign-verify.sh.log` from the CI run?

ulrichb:
... one suspicion I have, is that we use `openssl rsautl` and `openssl dgst` for verification within `test-pkcs11-tool-sign-verify.sh`. Maybe this broke the verification part (with OpenSSL 3), not the actual `pkcs11-tool` part.

Note that I also get in my local run _"The command rsautl was deprecated in version 3.0. Use 'pkeyutl' instead"_ warnings (although the test passes).

ulrichb:
@Jakuje Could you maybe trigger a build on master to get a "baseline"?

Jakuje:
> @Jakuje Could you maybe trigger a build on master to get a "baseline"?

We have build from yesterday on master which worked just ok, which makes me thinking that it has to be something else:

https://github.com/OpenSC/OpenSC/runs/7649380818?check_suite_focus=true

Actually, there are logs uploaded to github as artifacts in https://github.com/OpenSC/OpenSC/actions/runs/2784469746 but the naming is not great (and the upload is skipped because of the build failed, which makes it useless). I just pushed a change to your branch which shold upload the logs.

Jakuje:
Hmm. Now it works so it was probably just some intermittent issue, maybe bad openssl build ...

Jakuje (approved):

ulrichb:
Oh. Many thanks @Jakuje !

ulrichb:
@popovec Ready to merge?

popovec (approved):

Jakuje:

ulrichb:
Thanks for the review and for merging!

PR #2583, opened 2022-07-27
Add uninstall hooks

### LINKED ISSUE #2372: `sudo make uninstall` does not un-install all installed directories/files.
### Problem Description
`sudo make uninstall` does not uninstall all the directories and files that it installed.

### Proposed Resolution
Un-install the directory and files listed below.

### Steps to reproduce

I followed (more or less) the directions on <https://github.com/OpenSC/OpenSC/wiki/Compiling-and-Installing-on-Unix-flavors> to build OpenSC on Linux and then install it, namely, 
```
./bootstrap
./configure 
make
sudo make install
```
this put the installed files into `/usr/local`.  So far, so good.

The problem that I noticed is that not all of the installed directories/files are uninstalled by
```
sudo make uninstall
```
In particular, the leftover directories/files are:
```
/usr/local/lib/pkcs11/
/usr/local/lib/pkcs11/pkcs11-spy.so
/usr/local/lib/pkcs11/opensc-pkcs11.so
/usr/local/lib/pkcs11/onepin-opensc-pkcs11.so
```

xhanulik:
Fixes #2372
This PR adds an uninstall script to remove files created in the `install-exec-hook` in makefiles.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -9,6 +9,7 @@ on:
       - .github/setup-linux.sh
       - .github/build.sh
       - .github/push-artifacts.sh
+      - '**.am'
   push:
 
 jobs:
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -11,6 +11,7 @@ on:
       - .github/cleanup-macos.sh
       - .github/build.sh
       - .github/push-artifacts.sh
+      - '**.am'
   push:
 
 jobs:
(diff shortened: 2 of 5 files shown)


frankmorgner (approved):

Jakuje (approved):

Jakuje:
Can we make some regression test to make sure this keeps working and we prevent re-introducing similar issues in the future? I am not sure if this would work as part of the autotools check or as part of the `.github/build.sh` directly or somewhere else?

Additionally, I see that this change did not trigger the github actions as we trigger only on c and h files modifications (which would probably make sense to change also to add Makefiles modifications). Otherwise i

xhanulik:
Github actions should now run on makefiles as well. I'll look into doing some testing, either with autotools or some other way.

Jakuje:
Thank you. Merged.

PR #2584, opened 2022-07-29
Update the actions/virtual-environments#4589 workaround

swissbit-csteuer:
Fixes the Linux CI setup as described here actions/virtual-environments#4589

The fix removes/downgrades libraries built by GitHub Images for which no i386 version exists so that wine can install.


Jakuje:
Thank you for the PR fixing the CI for us!

PR #2594, opened 2022-08-15
Oss-fuzz issues

xhanulik:
Fixing another batch of recent issues reported by OSS-Fuzz, mainly memory leaks.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-westcos.c
+++ b/src/pkcs15init/pkcs15-westcos.c
@@ -247,12 +247,17 @@ static int westcos_pkcs15init_generate_key(sc_profile_t *profile,
 	    EVP_PKEY_keygen_init(pctx) != 1 ||
 	    EVP_PKEY_CTX_set_rsa_keygen_bits(pctx, key_info->modulus_length) != 1 ||
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
-	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
+	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1) {
 #else
-	    EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn) != 1 ||
-	    (bn = NULL) || /* pctx will free bn */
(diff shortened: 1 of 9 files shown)


src/pkcs15init/profile.c
```diff
@@ -835,6 +835,8 @@ init_state(struct state *cur_state, struct state *new_state)
 static int
 do_card_driver(struct state *cur, int argc, char **argv)
 {
+	if (cur->profile->driver)
+		free(cur->profile->driver);
```
Jakuje, line 839: no need to check for null here. If it is null, the `free` will be no operation.

src/tools/pkcs15-tool.c
```diff
@@ -2373,7 +2373,7 @@ int main(int argc, char *argv[])
 		aid.len = sizeof(aid.value);
 		if (sc_hex_to_bin(opt_bind_to_aid, aid.value, &aid.len))   {
 			fprintf(stderr, "Invalid AID value: '%s'\n", opt_bind_to_aid);
-			return 1;
+			goto end;
```
Jakuje, line 2376: I think you should set here also the `err` to some non-zero value.

Jakuje (approved):

xhanulik:
Added commits fixing new issues reported by Coverity.

src/pkcs15init/pkcs15-westcos.c
```diff
@@ -250,7 +250,7 @@ static int westcos_pkcs15init_generate_key(sc_profile_t *profile,
 	    EVP_PKEY_CTX_set1_rsa_keygen_pubexp(pctx, bn) != 1 ||
 #else
 	    EVP_PKEY_CTX_set_rsa_keygen_pubexp(pctx, bn) != 1 ||
-	    (bn = NULL) || /* pctx will free bn */
+	    (bn = NULL) != NULL || /* pctx will free bn */
```
Jakuje, line 253: not sure if this will not result in some report "condition is always true". Can we break this long condition to more blocks and use the `bn = NULL;` as normal assignement?

Jakuje (approved):

PR #2606, opened 2022-09-18
Fix for 'Coverity scan' CID 380537

popovec:

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -5552,7 +5552,14 @@ pkcs15_skey_encrypt(struct sc_pkcs11_session *session, void *obj,
 	   can be an init operation or final operation..*/
 
 	if (skey && !(skey->info->usage & SC_PKCS15_PRKEY_USAGE_ENCRYPT))
+		skey = NULL;
+
+	/* TODO: should we look for a compatible key automatically? prv_next not implemented yet. */
+	/* skey = skey->prv_next; */
+
+	if (skey == NULL)
 		return CKR_KEY_FUNCTION_NOT_PERMITTED;
+
 	sc_log(context, "Using mechanism %lx.", pMechanism->mechanism);
 
 	switch (pMechanism->mechanism) {


popovec:
test-cac and test-oseid - both tests fail at this point:
opensc-tool: error while loading shared libraries: libcrypto.so.47: cannot open shared object file: No such file or directory

but this patch passed these tests - see  link:
https://github.com/popovec/OpenSC/actions/runs/3075911508

dengert:
The failure is `pkcs11-tool` can not load `libcrypto.so.47` This appears to be from LibreSSL version  3.4.1.  OpenSSL libs look like librcypto.so.N.N.N

It looks like there is a mis-match of `pkcs11-tool` and the SSL libraries. 


Are we running multiple tests using the same system? 

i.e. `build-libressl` and `build-openssl-3`
`test-cac-openssl-3` and `test-cac-libressl`

Are we installing into the same directories i.e. /usr/local? 

Do we use a LD_LIBRARY_PATH path"

I build multiple versions of OpenSSL  using `PREFIX=/opt/ossl-<version>`
and  `LDFLAGS="$LDFLAGS -Wl,-rpath,$PREFIX/lib"` Then build OpenSC, libp11  into these same directories. i.e. basically multiple versions of `/usr/local`.

src/pkcs11/framework-pkcs15.c
```diff
@@ -5552,7 +5552,13 @@ pkcs15_skey_encrypt(struct sc_pkcs11_session *session, void *obj,
 	   can be an init operation or final operation..*/
 
 	if (skey && !(skey->info->usage & SC_PKCS15_PRKEY_USAGE_ENCRYPT))
+		skey = NULL;
+
+	/* Please read comments in pkcs15_skey_unwrap() and pkcs15_skey_wrap() */
+
+	if (skey == NULL)
 		return CKR_KEY_FUNCTION_NOT_PERMITTED;
```
Jakuje, line 5560: Isn't this doing the same as the following, with less obfuscation?

```suggestion
	if (!skey || !(skey->info->usage & SC_PKCS15_PRKEY_USAGE_ENCRYPT))
		return CKR_KEY_FUNCTION_NOT_PERMITTED;
```

I am also not sure what the comment is trying to say here, whether it is related to the change you made or to something else in this function.

Jakuje:
The CI failure happens from time to time, but I never had time  to investigate it better. There is caching going on between the stages of the build. I recall there were some issues with the prefixes so when I used cache keys, I used `${{ runner.os }}-18-${{ github.sha }}`, while the libressl and openssl 3.0 now use keys like `${{ runner.os }}-${{ github.sha }}-ossl3`, which might override the previously created cache values for the main build in `${{ runner.os }}-${{ github.sha }}` that is restored in these two tests. I would try with #2609 if this will help.

src/pkcs11/framework-pkcs15.c
popovec: We need to prevent dereferencingf NULL pointer (in case `skey` is NULL). But, there is already similar code in another functions:
````
	/* Check whether this key supports unwrap */
	if (skey && !(skey->info->usage & SC_PKCS15_PRKEY_USAGE_UNWRAP))
		skey = NULL;

	/* TODO: should we look for a compatible key automatically? prv_next not implemented yet. */
	/* skey = skey->prv_next; */

	if (skey == NULL)
		return CKR_KEY_FUNCTION_NOT_PERMITTED;
````
Either we will use the current PR, which we will update in the future (there will be a search for another possible key), or I will change this PR to just a simple test like you suggested.

Jakuje: If the comment between the statements will be as in the other functions, I am ok with the code as it is. 

But the current comment really does not help understanding why the statement is broken into separate blocks without reading through comments in other places.

Jakuje (approved):

frankmorgner (approved):

PR #2612, opened 2022-09-30
Fixes for OSS-Fuzz issues

xhanulik:
Fixing recent issues reported by OSS-Fuzz.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -496,7 +496,7 @@ static int cardos_list_files(sc_card_t *card, u8 *buf, size_t buflen)
 {
 	sc_apdu_t apdu;
 	u8        rbuf[256], offset = 0;
-	const u8  *p = rbuf, *q, *tag;
+	const u8  *p, *q, *tag;
 	int       r;
 	size_t    fids = 0, len;
 
@@ -521,6 +521,7 @@ get_next_part:
 		sc_log(card->ctx,  "directory listing > 256 bytes, cutting");
 	}
 
+	p = rbuf;
 	len = apdu.resplen;
 	while (len != 0) {
 		size_t   tlen = 0, ilen = 0;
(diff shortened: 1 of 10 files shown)


src/libopensc/card-cardos.c
```diff
@@ -524,6 +524,11 @@ static int cardos_list_files(sc_card_t *card, u8 *buf, size_t buflen)
 	len = apdu.resplen;
 	while (len != 0) {
 		size_t   tlen = 0, ilen = 0;
+		/* check for finding tag out of bounds of rbuf */
+		if (p + len > rbuf + sizeof(rbuf)) {
+			sc_log(card->ctx,  "error parsing tag");
+			return SC_ERROR_INTERNAL;
+		}
```
Jakuje, line 531: I think this is not correct fix. Similar as the previous comment, this looks like mending the place where the crash happens, but not the original problem, that caused the issue, which is a some miscalculation of the `p` or `len` when going into the second round of this cycle. I tried to get some more debug logs and I am getting the following misalignment:
```
[fuzz] card-cardos.c:527:cardos_list_files: p=0x7f29dca9b0b0, len=245, offset=0
[fuzz] card-cardos.c:539:cardos_list_files: len=245, tag=0x7f29dca9b0b2, tlen=50, p=0x7f29dca9b0b0
[...]
[fuzz] card-cardos.c:527:cardos_list_files: p=0x7f29dca9b0e4, len=206, offset=52

[fuzz] card-cardos.c:531:cardos_list_files: error parsing tag
```
Only after first couple of looks into that I noticed that this goes through the `goto` to `get_next_part:` to pull new data, but it is not updating the `p` pointer to point back to the start of the `rbuf` allowing the `sc_asn1_find_tag` going behind the end of the buffer. Hopefully, this will be the correct solution, but I dont have this card to try.

Jakuje (changes requested):

Jakuje (approved):
looks good. It might be even better to move the new function into the asn1.c as it is really parsing of special ASN1 value (sorry for guiding you to the sc.c), but that can be done later.

xhanulik:
Alright, moved to asn1.c

Jakuje (approved):

PR #2616, opened 2022-10-13
macOS: Correct LaunchAgent paths in opensc-uninstall script

Eitot:
This fixes an omission in e05574d94217e6fbd9186509a1b494fb30ee1431. The names of these files have changed, so they are not deregistered from LaunchServices and uninstalled when running the opensc-uninstall script.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -12,8 +12,7 @@ on:
       - .github/build.sh
       - .github/push-artifacts.sh
       - '**.am'
-      - MacOSX/build-package.in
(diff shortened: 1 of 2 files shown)


Jakuje:
The CI for macos did not trigger. Can you update also the https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/macos.yml#L16 to include any change in the MacOSX directory as it looks like all of them are related to the package build?

Jakuje (approved):
looks good to me. I dont have a way to test this though.

frankmorgner (approved):

PR #2617, opened 2022-10-14
Do not build tools when crucial dependencies are missing

### LINKED ISSUE #2074: Do not build empty/useless tools with missing dependencies
`opensc-notify`, `npa-tool` and maybe more tools are built even if crucial dependencies are not met (GUI libraries, openpace, ...). It would be nice to adjust build system to not build these if they are not usable.

Oh... and yes, disabling opensc-notify in this case, would be good from a user's perspective

_Originally posted by @frankmorgner in https://github.com/OpenSC/OpenSC/pull/2069#issuecomment-652948763_

xhanulik:
Fixes #2074
This PR adds checking for `ENABLE_NOTIFY` and `ENABLE_OPENPACE` macros into Makefile for tools. When the libraries are missing, the tools which are useless without them are not built.

##### Checklist
- [ ] macOS tokend is tested


Jakuje (approved):

win32/Make.rules.mak
```diff
@@ -126,6 +126,10 @@ OPENPACE_DIR = C:\openpace
 !ENDIF
 OPENPACE_INCL_DIR = /I$(OPENPACE_DIR)\src
 OPENPACE_LIB = $(OPENPACE_DIR)\src\libeac.lib
+!IF "$(OPENSSL_DEF)" == "/DENABLE_OPENSSL"
```
frankmorgner, line 129: This should check for `OPENPACE_DEF`, I think

frankmorgner (changes requested):

win32/Make.rules.mak
xhanulik: This is the case when `npa-tool` needs OpenSSL and OpenPACE, as in the `Makefile.am`

https://github.com/OpenSC/OpenSC/blob/ee0c1fbacc4dc75ce52910564ef619db5503716a/src/tools/Makefile.am#L34-L40

Jakuje:
Thank you for your contribution. Merging.

PR #2621, opened 2022-10-18
pkcs11-tool: disable wrap/unwrap test until OpenSC issue #1796 is resolved

popovec:


frankmorgner (approved):

Jakuje:
Thank you for keeping eye on this. It would be great to update the #1796 with that is now needed to make it working as last comment there is from January.

PR #2622, opened 2022-10-20
Run CI with Ubuntu 20.04

### LINKED ISSUE #2618: Update CI to run on newer ubuntu than 18.04
We are currently using Ubuntu 18 in CI because we need old java to run tests for javacards if I am right. 

Github will be removing these images on 1st December 2022 (or on 4th of March 2023? They are quite inconsistent in actions/runner-images/issues/6002 ) so we should try to see if the CI can work on newer images and migrate if possible.

xhanulik:
Fixes #2618

Switch the `ubuntu-18.04` to `ubuntu-20.04` in Linux CI workflows, as Ubuntu 18 will be unsupported from April 1st, 2023, according to https://github.com/actions/runner-images/issues/6002.


xhanulik:
Tests for `ubuntu-latest` are merged now. I also rewrote the job for testing OpenSSL 3.0 to use Ubuntu 22.04 as it already has OpenSSL3 and removed unneeded scripts for setting up OpenSSL.

.github/workflows/linux.yml
```diff
@@ -219,29 +199,29 @@ jobs:
             opensc*.tar.gz
 
   test-cac-openssl-3:
```
Jakuje, line 201: I would also update the test name to see it follows the  `build-ubuntu-22`.

Jakuje:
Now, we have a bogus file `.github/workflows/.linux.yml.swp` in git. Please, remove it.

Jakuje:
One more thing that mentions the ossl3 and setup-openssl.sh scripts is in the `containers/` directory.

xhanulik:
I missed the leftovers in `containers/`, thanks

Jakuje (approved):

frankmorgner (approved):

Issue #2626, opened 2022-10-24
Missing minidriver registry Keys for CardOS V4.2C


pl31:
### Problem Description

I installed OpenSC Version 0.22.0.0. It works pretty good using CardOS V5.3 SmartCards, but V4.2C SmartCards do not work properly.

"certutil -scinfo" reads ATR fine but fails afterwords (WIN32: 2 ERROR_FILE_NOT_FOUND).  After some research I found, that the minidriver registry entries are missing for CardOS V4.2 C.

I tried adding a registry key for all V4.2 cards, using the "mask" from card-cardos.c
```
{ "3b:f2:18:00:ff:c1:0a:31:fe:55:c8:06:8a", "ff:ff:0f:ff:00:ff:00:ff:ff:00:00:00:00", NULL, SC_CARD_TYPE_CARDOS_M4_2, 0, NULL }
```
but it still failed. Using explicit ATR finally worked.

### Proposed Resolution

I checked a propriatary SmartCard driver and found V4.2B, V4.2C and V4.3B missing. I finally added those entries manually to my system. Somehow V4.2 and V4.3 cards should be added to the installation procedure.

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards\CardOS 4.2 b]
"Crypto Provider"="OpenSC CSP"
"Smart Card Key Storage Provider"="Microsoft Smart Card Key Storage Provider"
"80000001"="C:\\Program Files\\OpenSC Project\\OpenSC\\minidriver\\opensc-minidriver.dll"
"ATR"=hex:3b,f2,18,00,02,c1,0a,31,fe,58,c8,09,75
"ATRMask"=hex:ff,ff,ff,ff,ff,ff,ff,ff,ff,ff,ff,ff,ff

[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards\CardOS 4.2 c]
"Crypto Provider"="OpenSC CSP"
"Smart Card Key Storage Provider"="Microsoft Smart Card Key Storage Provider"
"80000001"="C:\\Program Files\\OpenSC Project\\OpenSC\\minidriver\\opensc-minidriver.dll"
"ATR"=hex:3b,f2,18,00,02,c1,0a,31,fe,58,c8,0b,77
[...]

PR #2628, opened 2022-10-26
customactions.cpp - fix ATRs before adding to registry

dengert:
Fixes: #2625

OpenSC definitions of ATRs have been lax in "sc_atr_table" entries by allowing 1 bits in the ATR that need to be 0 bits when used with Windows compare Do the equivalent reduction of the table ATR done in card.c by `tbin[s] = (tbin[s] & mbin[s]);` before adding to registry.

This will then allow lax definitions of ATRs to be copied to customactions.cpp and fixed before adding to registry.

 Changes to be committed:
	modified:   win32/customactions.cpp

##### Checklist

- [todo] Windows minidriver is tested


dengert:
This PR was tested by @pl31 and reported  to work in https://github.com/OpenSC/OpenSC/issues/2626#issuecomment-1294654505

Jakuje (approved):

PR #2637, opened 2022-11-07
OSS-Fuzz issues

xhanulik:
Fixing latest issues reported by OSS-Fuzz

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -1679,6 +1679,38 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
 	return 0;
 }
 
+static void sc_free_entry(struct sc_asn1_entry *asn1) {
+	int idx = 0;
+	struct sc_asn1_entry *entry = asn1;
+
+	if (!asn1)
+		return;
+
+	for (idx = 0; asn1[idx].name != NULL; idx++) {
+		entry = &asn1[idx];
+		switch (entry->type) {
+		case SC_ASN1_CHOICE:
+		case SC_ASN1_STRUCT:
+			sc_free_entry((struct sc_asn1_entry *) entry->parm);
+			break;
+		case SC_ASN1_OCTET_STRING:
+		case SC_ASN1_BIT_STRING_NI:
+		case SC_ASN1_BIT_STRING:
+		case SC_ASN1_GENERALIZEDTIME:
+		case SC_ASN1_PRINTABLESTRING:
+		case SC_ASN1_UTF8STRING:
+			if ((entry->flags & SC_ASN1_ALLOC) && (entry->flags & SC_ASN1_PRESENT)) {
+				u8 **buf = (u8 **)entry->parm;
+				free(*buf);
+				*buf = NULL;
+			}
+			break;
+		default:
+			break;
+		}
+	}
+}
+
 static int asn1_decode(sc_context_t *ctx, struct sc_asn1_entry *asn1,
 		       const u8 *in, size_t len, const u8 **newp, size_t *len_left,
 		       int choice, int depth)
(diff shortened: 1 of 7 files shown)


src/pkcs15init/profile.c
```diff
@@ -2006,6 +2019,9 @@ build_argv(struct state *cur, const char *cmdname,
 		if (list == mac->value) {
 			return SC_ERROR_SYNTAX_ERROR;
 		}
+		if (check_macro_cycle(list, mac->value, cur)) {
```
Jakuje, line 2027: This change is quite hard to understand. It would be great if this would have some comment describing what is the issue we are trying to solve and how we are solving it.

My understanding is that this means there are some macros that recursively calls itself causing the infinite loop/recursion (maybe better word than cycle?) evaluation.

src/libopensc/card-openpgp.c
```diff
@@ -1150,18 +1150,16 @@ pgp_free_blob(pgp_blob_t *blob)
...
-			while (child != NULL) {
-				pgp_blob_t *next = child->next;
+		while (child != NULL) {
+			pgp_blob_t *next = child->next;
 
-				pgp_iterate_blobs(child, level-1, func);
-				child = next;
-			}
+			pgp_iterate_blobs(child, func);
+			child = next;
```
Jakuje, line 1162: I understand the rest of this change that is freeing the data, but not this first part which is removing the `level` argument. Isn't this going to cause infinite recursion?

src/pkcs15init/pkcs15-entersafe.c
```diff
@@ -130,6 +130,8 @@ static int entersafe_init_card(sc_profile_t *profile, sc_pkcs15_card_t *p15card)
 
 
 		 /* fill file by 0 */
+		 if (size > MAX_FILE_SIZE)
+			 LOG_TEST_RET(card->ctx, SC_ERROR_INVALID_DATA, "Initialize EF(DIR) failed with invalid file size");
```
Jakuje, line 134: maybe better error explaining the invalid size is too large?

xhanulik: Fixed

src/libopensc/asn1.c
```diff
@@ -1679,6 +1679,38 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
...
+static void sc_free_entry(struct sc_asn1_entry *asn1) {
+	int idx = 0;
+	struct sc_asn1_entry *entry = asn1;
+
+	if (!asn1)
+		return;
+
+	for (idx = 0; asn1[idx].name != NULL; idx++) {
+		entry = &asn1[idx];
+		switch(entry->type) {
```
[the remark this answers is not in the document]
xhanulik: Fixed

src/libopensc/card-openpgp.c
xhanulik: I believe it shouldn't - the `while (child != NULL)` is here as the 'termination condition'. Picturing the filesystem as a tree, blobs in leaves don't have any children, so the function starts here with freeing the blobs and returning back. With some enormous height of the tree, it would probably cause some stack overflow, but stating only higher `levels` would be overcome by fuzzing one day.

src/libopensc/card-authentic.c
```diff
@@ -194,6 +194,9 @@ authentic_get_tagged_data(struct sc_context *ctx, unsigned char *in, size_t in_l
 			tag_len = 1;
 		}
 
+		if (in + offs + tag_len >= in + in_len)
```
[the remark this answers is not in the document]
xhanulik: Fixed and checks added for `authentic_parse_size()`.

src/libopensc/card-openpgp.c
Jakuje: I think the `level` here is some kind of precaution before some infinite loops. We are freeing the blobs only after we walk through all of them depth-first, but if there is a cycle, this will never happen.

xhanulik: Yes, `level` denotes the maximum depth where the blobs are released, meaning that the deeper levels would leak memory. I think adjusting the `level` is insufficient, as the fuzzing will probably find the same problem with a higher tree in the future. 

However, building the blobs tree looks like there _should not_ be any cycles in the structure, so just removing the level limitation looks ok (otherwise, also `pgp_seek_blob()`, which recursively traverses the tree would run into some infinite loop). I also added commit message with some clarification.

Jakuje: Sounds good. Thanks for the clarification!

Jakuje (approved):

Jakuje:
Thank you!

PR #2630, opened 2022-10-28
[pkcs11-tool] CKO display serial number

AlexandreGonzalo:
I am using the pkcs11-tool to list the certificates stored in my PKCS#11 token.
I have this patch which displays the certificate serial number and maybe that's something which could be merged to master.

Certificate Object; type = X.509 cert
  subject:    DN: C=FR, ST=PACA, L=Valbonne, O=Trustonic, CN=*.trustonic.com
  serial:     54E67E7F3DB8F07D11D473717C71E057FD7D22C9
  ID:         01

##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [ ] PKCS#11 module is tested
- [ ] Windows minidriver is tested
- [ ] macOS tokend is tested

PR #2641, opened 2022-11-14
feat(pkcs11-tool): print serial number when reading certificates

AlexandreGonzalo:
I am using the pkcs11-tool to list the certificates stored in my PKCS#11 token.
I have this patch which displays the certificate serial number and maybe that's something which could be merged to master.

subject: DN: C=FR, ST=PACA, L=Valbonne, O=Trustonic, CN=*.trustonic.com
serial: 54E67E7F3DB8F07D11D473717C71E057FD7D22C9
ID: 01

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
From answer to @frankmorgner comment in #2630
>>  can we do this without being dependant on OpenSSL?
>
> I did the same as getSUBJECT() above and it is an easy win.
> I don't think that it is worth to try to avoid OpenSSL.

I think this is ok as we already use the openssl to parse the SUBJECT. In any case, most of the builds are using OpenSSL anyway now.

I think this is good to go also for 0.23.0

Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you. Merging.

PR #2642, opened 2022-11-14
isoapplet offcard hash

swissbit-csteuer:
Hi,

currently the IsoApplet and OpenSC only support ECDSA-SHA1 because the hash is computed on-card.
This pull request adds support for ECDSA, ECDSA-SHA224, ECDSA-SHA256, ECDSA-SHA384, ECDSA-SHA512 by computing the hash off-card.

The IsoApplet needs to be changed: https://github.com/philipWendland/IsoApplet/compare/master...swissbit-eis:IsoApplet:master

jcardsim also requires some changes to support raw ECDSA: https://github.com/swissbit-eis/jcardsim/commit/00e646a634dc7cc491364b6eff61810c6c20bdab

The changes have been tested with the following card: Swissbit Secure USB PU-50n SE/PE (99000010EF000279) 01 00

[Result of p11test](https://github.com/OpenSC/OpenSC/files/10019674/isoapplet.json.txt)
[Result of pkcs11-tool --test](https://github.com/OpenSC/OpenSC/files/10019675/pkcs11-tool--test-output.txt)

##### Checklist


swissbit-csteuer:
Hi @philipWendland , since this PR also affects the ISOApplet, could you take a look at it?

frankmorgner:
In your PR you're *replacing* on-card-hashing with off-card-hashing. Would it be possible to support both?

Additionally, I suggest you open a pull request with your changes at the ISOApplet repository so it can be reviewed...

swissbit-csteuer:
> In your PR you're _replacing_ on-card-hashing with off-card-hashing. Would it be possible to support both?

I guess we could check the version number of the ISOApplet and use SHA1 on-card-hashing for old versions and off-card-hashing for the new ISOApplet version. 

> Additionally, I suggest you open a pull request with your changes at the ISOApplet repository so it can be reviewed...

Done.

philipWendland:
Hi,
thank you for the contribution.

Currently, the change of the IsoApplet major version number in the card-isoApplet.c breaks all existing deployments because the driver refuses to work with older applet versions (see [here](https://github.com/OpenSC/OpenSC/blob/a5995ffb428e0664a5280c212f04c04de0d4f7cc/src/libopensc/card-isoApplet.c#L176). Hence, this PR should currently not be merged.
However, I think we can make this work.

ECC with off-card hashes requires JavaCard 3+ and I want to keep supporting v2.2.2 cards with IsoApplet.
It is likely necessary for me to maintain a "legacy" branch of the IsoApplet for JavaCards v2.2.2 (without off-card-ECC support) and one newer (v3+) cards. I will report back here once the IsoApplet-side is ready.

philipWendland:
Could you please explain the one-line change in pkcs11-tool to me?

swissbit-csteuer:
> Could you please explain the one-line change in pkcs11-tool to me?

This was necessary to compute raw ECDSA signatures with the pkcs11-tool like its defined in the [PKCS#11 spec (2.3.6 ECDSA without hashing)](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/os/pkcs11-curr-v2.40-os.html):
> Input the entire raw digest. Internally, this will be truncated to the appropriate number of bits.

popovec:
In my opinion, it is not necessary to change anything in the `pkcs11-tool.c` code. `pkcs11-toll.c` code for ECDSA (raw) is fully functional and is tested by github actions on the OsEID simulation.

Testing is done by calling `pkcs11-tool` approximately as follows:

` pkcs11-tool --sign -m ECDSA --signature-format "sequence" --input-file tmp/testfile.txt --output-file tmp/testfile.txt.pkcs11.sig --id $keyID`

Result is verified both with `pkcs11-tool` and with `openssl`

Example: if we calculate ECDSA (without applying the hash function) for `secp521r1`, the input string is exactly 66 bytes long -> _r=66,  sizeof(in_buffer) = 1025_ . The condition `r < sizeof(in_buffer)` will be met. Any test for mechanism makes no sense here.

dengert:
I believe the change is needed.

2.3.6 says: "(This mechanism corresponds only to the part of ECDSA that processes the hash value, which should not be longer than 1024 bits; it does not compute the hash value.)"  Which says: "**should**". 

If the input is larger the 1024 bytes, the size of the input buffer + 1, the operation will fail in the pkcs11-tool. PKCS11 also says in 
2.3.6 foot note 3 "Input the entire raw digest. Internally, this will be truncated to the appropriate number of bits." 

So pkcs11-tool is making some assumptions on the max size of the input data i.e. < 1024 bytes but that that should be up to the PKCS11 module to do the truncation based on the size of the key. 

One could argue that there are no ECDSA keys or hashes today larger the 1024 bytes, and hashes and key are measured in bits.

src/libopensc/card-isoApplet.c
```diff
@@ -1195,21 +1235,24 @@ isoApplet_compute_signature(struct sc_card *card,
...
-			memcpy(out, p, len);
+			memcpy(seqbuf, p, len);
 			r = len;
 		}
 
 		free(p);
 	}
+	if ((size_t) r > outlen)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_BUFFER_TOO_SMALL);
+	memcpy(out, seqbuf, r);
```
Jakuje, line 1255: I am wondering if we need these two memcpy. It will probably not be large overhead, but copying data back and forth might be at least confusing and error prone.

src/libopensc/card-isoApplet.c
```diff
@@ -210,22 +215,33 @@ isoApplet_init(sc_card_t *card)
...
 	/* Obtain applet version and specific features */
-	if (0 > isoApplet_select_applet(card, isoApplet_aid, ISOAPPLET_AID_LEN, rbuf, &rlen)) {
-		free(card->drv_data);
-		card->drv_data = NULL;
-		LOG_TEST_RET(card->ctx, SC_ERROR_INVALID_CARD, "Error obtaining applet version.");
+	r = isoApplet_get_info(card, drvdata);
+	LOG_TEST_GOTO_ERR(card->ctx, r, "Error obtaining information about applet.");
+
+
+	unsigned int major_version = drvdata->isoapplet_version & 0xFF00;
```
Jakuje, line 222: I would probably advocate to move this variable declaration before any any other statements.

.github/test-isoapplet.sh
```diff
@@ -11,13 +20,13 @@ export LD_LIBRARY_PATH=/usr/local/lib
 
 # The ISO applet
 if [ ! -d IsoApplet ]; then
-	git clone https://github.com/philipWendland/IsoApplet.git
-	# enable IsoApplet key import patch
-	sed "s/DEF_PRIVATE_KEY_IMPORT_ALLOWED = false/DEF_PRIVATE_KEY_IMPORT_ALLOWED = true/g" -i IsoApplet/src/net/pwendland/javacard/pki/isoapplet/IsoApplet.java
+    git clone https://github.com/philipWendland/IsoApplet.git --branch $isoapplet_branch
+    # enable IsoApplet key import patch
+    sed "s/DEF_PRIVATE_KEY_IMPORT_ALLOWED = false/DEF_PRIVATE_KEY_IMPORT_ALLOWED = true/g" -i "IsoApplet/src/${isoapplet_pkgdir}/IsoApplet.java"
```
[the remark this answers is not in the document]
swissbit-csteuer: I converted it back to tabs.

src/tests/p11test/isoapplet.json
```diff
@@ -0,0 +1,421 @@
+{
```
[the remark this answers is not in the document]
swissbit-csteuer: You're right. That file was added by accident. Removed it.

src/libopensc/card-isoApplet.c
swissbit-csteuer: Since we added an additional branch for RSA-PSS I refactored the code to handle each case (RSA, RSA-PSS and ECDSA) in its own branch. We only need seqbuf in the ECDSA branch and now we do not need additional memcpy's.

src/libopensc/card-isoApplet.c
swissbit-csteuer: Done

swissbit-csteuer:
I added some more changes to enable RSA 4096 support with the new v1 version of the IsoApplet. Now everything new function of the v1 applet should be covered. See https://github.com/philipWendland/IsoApplet/pull/28

philipWendland:
My plan is to review and test those changes carefully (also the changes to IsoApplet). Depending on the effort, it might take me until the end of this month.

Then, the main branch of IsoApplet will be Version 1.0.0 of the applet, for JavaCards >= 3.0.4, with the following new features:
- RSA 4096
- RSA PSS
- ECDSA off-card hashes

There will be a (maintained) legacy branch of IsoApplet, so that JavaCards v.2.2.2 are still supported. Both versions of the IsoApplet will be supported by the OpenSC driver. 

I will report back here as soon as my tests are done and the changes to IsoApplet are ready to be merged.

philipWendland:
From my point of view, this PR can be merged. 
I will update the IsoApplet main branch as soon as this PR is merged.

Jakuje (approved):
looks good. I put there one minor comment, but we can live without that. Please rebase on top of current master to get rid of the CI failures too.

.github/test-isoapplet.sh
```diff
@@ -11,13 +20,13 @@ export LD_LIBRARY_PATH=/usr/local/lib
 
 # The ISO applet
 if [ ! -d IsoApplet ]; then
-	git clone https://github.com/philipWendland/IsoApplet.git
+	git clone https://github.com/philipWendland/IsoApplet.git --branch $isoapplet_branch
```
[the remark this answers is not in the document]
swissbit-csteuer: Good idea. We can optimize this further by doing a shallow clone with `--depth 1` which implies `--single-branch`.

swissbit-csteuer:
> looks good. I put there one minor comment, but we can live without that. Please rebase on top of current master to get rid of the CI failures too.

Thanks for the review.

I have rebased onto the current master.
The test-isoapplet.sh script now fails when using the v1 branch.
I'm trying to fix it but can't reproduce it locally... so it might take a bit longer.

Jakuje:
>  The test-isoapplet.sh script now fails when using the v1 branch.

is it deterministic in ci? Did you try to rerun it? Some tests sometimes fail for no reason. There is already too many moving parts with java, jcardsim, ...

swissbit-csteuer:
Yes, it's deterministic and I can now reproduce it locally.

Before, I used a different `opensc.conf` where I forced the maximum send and receive size for pcsc to 400 bytes as a workaround for an issue with my test hardware.
When I do not force the maximum, send and receive size, I get the same problem with the simulator. This is what happens:

1. test-isoapplet.sh calls `pkcs11-tool --login --test ...`
2. One of the tests calls `iso7816_decipher` with in-len 256 and out-len 512
3. `iso7816_decipher` calls `fixup_transceive_length` which sets `apdu->flags |= SC_APDU_FLAGS_CHAINING` because `sc_get_max_send_size` returns 255 but we want to send 256 byte
4. `sc_transmit_apdu` sends a command chaining APDU (CLA = 0x10)
5. The IsoApplet v1 rejects that APDU with 0x6884 (SW_COMMAND_CHAINING_NOT_SUPPORTED).

The problem is that `sc_get_max_send_size` returns 255, so that OpenSC cannot send an extended length APDU and falls back to command chaining, which the new IsoApplet does not support.

The virtual smartcard ifd handler (ifd-vpcd) [does not support](https://github.com/frankmorgner/vsmartcard/blob/master/virtualsmartcard/src/ifd-vpcd/ifd-vpcd.c#L181) `IFDHControl`. `IFDHControll` is called by `SCardControl`, which in turn is called by OpenSC to [detect the reader properties](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L1151). Since that fails, [OpenSC assumes that the reader supports only short APDUs](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L1348) where the maximum send size  is 255.

I'm not sure about the best way to resolve this issue.
The easiest fix is forcing max send/recv size to 65536 with a custom 
OpenSC config when testing v1 IsoApplet with the javacard simulator.

What do you think @Jakuje ?

frankmorgner:
You may fix this right away in ifd-vpcd and I can have a look!

Jakuje:
>  The easiest fix is forcing max send/recv size to 65536 with a custom
> OpenSC config when testing v1 IsoApplet with the javacard simulator.

This sounds to me as a good testing workaround before we will have fix in vpcd. I am not sure how hard would it be to implement this in vpcd from top of my head

swissbit-csteuer:
I gave it a shot and opened a PR https://github.com/frankmorgner/vsmartcard/pull/245

swissbit-csteuer:
The IsoApplet CI tests are now running without issues.
I rebased onto the current master and updated the reference JSON (+RSA 4096, +PKCS_PSS).

The last commit also changes the order in which the test result JSON files are passed to the diff command. Now the reference/expected file is the FROM file and the actual file the TO file. If a line is missing in the actual file, it's marked with '-' in the diff output instead of '+' and vice versa.

The only CI task that fails is the AppVeyor build, but the failure does not seem the be related to changes in this PR.

Jakuje (approved):

swissbit-csteuer:
Is there more that needs to be done before this can be merged?

Jakuje:
I think we are good to merge, unless there would be some more comments from @frankmorgner . If not, I will try to merge the PR later this week.

philipWendland:
Hi Jakuje,
thank you for merging. I updated the main branch of IsoApplet as a consequence.

This means that the test scripts need to be updated, please see https://github.com/OpenSC/OpenSC/pull/2688

PR #2645, opened 2022-11-15
feat(pkcs11-tool): add new --undestroyable

AlexandreGonzalo:
Hi All,

In this PR, I am adding a new option which can be used to mark the certificates as undestroyable.
Please, let me know if you want this option for the pkcs11-tool.
Regards,
Alexandre.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -167,6 +167,7 @@ enum {
 	OPT_PRIVATE,
 	OPT_SENSITIVE,
 	OPT_EXTRACTABLE,
+	OPT_UNDESTROYABLE,
 	OPT_TEST_HOTPLUG,
 	OPT_UNLOCK_PIN,
 	OPT_PUK,
@@ -272,6 +273,7 @@ static const struct option options[] = {
 	{ "private",		0, NULL,		OPT_PRIVATE },
 	{ "sensitive",		0, NULL,		OPT_SENSITIVE },
 	{ "extractable",	0, NULL,		OPT_EXTRACTABLE },
+	{ "undestroyable",	0, NULL,		OPT_UNDESTROYABLE },
 	{ "always-auth",	0, NULL,		OPT_ALWAYS_AUTH },
 	{ "test-ec",		0, NULL,		OPT_TEST_EC },
 #ifndef _WIN32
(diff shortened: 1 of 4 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -272,6 +273,7 @@ static const struct option options[] = {
 	{ "private",		0, NULL,		OPT_PRIVATE },
 	{ "sensitive",		0, NULL,		OPT_SENSITIVE },
 	{ "extractable",	0, NULL,		OPT_EXTRACTABLE },
+	{ "undestroyable",	0, NULL,		OPT_UNDESTROYABLE },
```
Jakuje, line 276: This is missing a documentation in the manual page for the pkcs11-tool (hence the CI failure)

Jakuje (changes requested):

src/tools/pkcs11-tool.c
AlexandreGonzalo: Good catch!
Sorry for the mistake.

Jakuje: I meant the manual page in `doc/tools/pkcs11-tool.1.xml`.

Jakuje:
Code-wise looks good. Can you squash the changes into single commit?

AlexandreGonzalo:
> Code-wise looks good. Can you squash the changes into single commit?

Done

Jakuje (approved):

frankmorgner (approved):

Jakuje:

zhangxiaohui1:
I noticed that the --undestroyable option is currently only implemented for certificate objects (when using --write-object with --type cert). However, it does not seem to be supported for other object types such as privkey, pubkey, secrkey, or data.

This behavior is not documented in the help text of --undestroyable, which gives the impression that it should work universally across all writable object types.

Could you clarify:

Is this limitation intentional for security or technical reasons?
Would the project be open to extending --undestroyable support to all object types?
This feature would be valuable for use cases where long-term key material (e.g., device identity keys) must be protected from accidental or malicious deletion.

Thank you for your work on this essential tool!

Jakuje:
Sounds like this was the author original use case to have undestroyable certificate. I think it makes sense to support this for other object types so feel free to open a PR with the changes to do so!

AlexandreGonzalo:
CKA_DESTROYABLE can be set for any storage object, it is not limited to certificates.
The limitation is because I could only test it with certificates.
Feel free to generalize it for all the storage objects.

PR #2654, opened 2022-11-30
Fedora packaging update

### LINKED ISSUE #2633: Clearifications on RPM defaults
### Problem description

For the RPM, OpenSC is configured by default like this:
https://github.com/OpenSC/OpenSC/blob/7eb0ec34c87fee640d977fdc396feca3bbb30217/packaging/opensc.spec#L66-L74

Why do you disable notifications by default? Is this feature unwanted, typically? Do we need a dedicated configuration option for that?

Why do you build the test cases for the production environment (cmocka)?

Why do you use `--with-pcsc-provider=libpcsclite.so.1`? If you're specifiying this explicitly, it should be better to specify an absolute path, shouldn't it?

Thank you!

@Jakuje

Jakuje:
Fixes #2633

This is sync with Fedora spec file (missing dependency for the wrap/unwrap test) and clean up of unneeded configure switches. I prefer to keep the rest of explicit enable/disable configure options to make the configuration more explicit and not depending only on the installed dependencies (as these can change under our hands).

### DIFF
--- a/.github/workflows/packit.yaml
+++ b/.github/workflows/packit.yaml
@@ -6,6 +6,7 @@ on:
       - '**.h'
       - '**.am'
       - .github/workflows/packit.yaml
+      - packaging/opensc.spec
   push:
 
 jobs:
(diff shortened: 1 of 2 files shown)


xhanulik:
Looks good to me

PR #2661, opened 2022-12-09
Update actions/virtual-environments#4589 workaround

swissbit-csteuer:
The workaround for i386 that is necessary to install wine has to be changed again.
See PR #2584 and https://github.com/actions/runner-images/issues/4589#issuecomment-1260157764 for reference.

### DIFF
--- a/.github/setup-linux.sh
+++ b/.github/setup-linux.sh
@@ -41,7 +41,6 @@ fi
 if [ "$1" == "mingw" -o "$1" == "mingw32" -o "$1" == "ix86" ]; then
 	sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
 	sudo apt-get update -qq
-	sudo apt-get install -yqq --allow-downgrades libgd3/focal libpcre2-8-0/focal libpcre2-16-0/focal libpcre2-32-0/focal libpcre2-posix2/focal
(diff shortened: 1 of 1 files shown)


swissbit-csteuer:
Sorry. Just noticed that PR#2660 fixes the same problem.

dengert:
I like your patch better. If it works It removes more then I did I only did the minimum needed.

Jakuje (approved):

Jakuje:
Thank you for taking care of this!

istr:
This workaround is not stable. See, for example:

https://github.com/frankmorgner/OpenSC/actions/runs/3995313248/jobs/6854040241

https://github.com/OpenSC/OpenSC/actions/runs/3990200149/jobs/6844167067

@dengert @frankmorgner please re-open.

Jakuje:
Looks like its broken again. Feel free to propose a new PR. There is no way to reopen PR other than reverting that and I believe that is not going to solve the current issue.

istr:
Probably this idea is currently the best bet
https://github.com/WOnder93/r128gain/commit/81c6ad7cfcba99f0e85eb2bcc0ec2f6d528f5e28

> Work around [1] by not using GitHub's Ubuntu environment, but instead
use a Fedora container, where it's easier to get a working 32-bit Wine.

PR #2663, opened 2022-12-12
Fix recent coverity issues and couple of assorted changes in p11test

Jakuje:
including some of minor reformatting and trailing whitespace removal in separate commits.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/sm/sm-common.c
+++ b/src/sm/sm-common.c
@@ -363,10 +363,8 @@ sm_encrypt_des_ecb3(unsigned char *key, unsigned char *data, int data_len,
 	int tmplen;
 #endif
 
-
 	if (!out || !out_len)
 		return -1;
-
 
 	*out_len = data_len + 7;
 	*out_len -= *out_len % 8;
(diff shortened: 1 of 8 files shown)


src/smm/sm-card-authentic.c
```diff
@@ -80,19 +79,23 @@ sm_oberthur_diversify_keyset(struct sc_context *ctx, struct sm_info *sm_info,
 			sc_debug(ctx, SC_LOG_DEBUG_SM, "key_buf:%s", sc_dump_hex(key_buff, 16));
 
 			rv = sm_encrypt_des_ecb3(master_key, key_buff, sizeof(key_buff), &tmp, &tmp_len);
-			LOG_TEST_RET(ctx, rv, "GP init session: cannot derive key");
+			if (rv != 0) {
+				free(tmp);
```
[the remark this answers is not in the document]
Jakuje: Good point. I will fix that. Thanks for checking!

src/sm/sm-common.c
```diff
@@ -385,29 +383,32 @@ sm_encrypt_des_ecb3(unsigned char *key, unsigned char *data, int data_len,
 	for (ii=0; ii<data_len; ii+=8)
 		DES_ecb2_encrypt( (sm_des_cblock *)(data + ii),
 				(sm_des_cblock *)(*out + ii), &ks, &ks2, DES_ENCRYPT);
+
+	return 0;
```
frankmorgner, line 387: better use SC_SUCCESS here

src/sm/sm-common.c
```diff
@@ -385,29 +383,32 @@ sm_encrypt_des_ecb3(unsigned char *key, unsigned char *data, int data_len,
...
 #else
 	cctx = EVP_CIPHER_CTX_new();
 	if (!EVP_EncryptInit_ex2(cctx, EVP_des_ede_ecb(), key, NULL, NULL)) {
```
frankmorgner, line 390: please add a check for `NULL == cctx`

src/sm/sm-common.c
```diff
@@ -385,29 +383,32 @@ sm_encrypt_des_ecb3(unsigned char *key, unsigned char *data, int data_len,
...
+		goto err;
 	}
 	*out_len += tmplen;
 	EVP_CIPHER_CTX_free(cctx);
-#endif
-
 	return 0;
+
+err:
+	EVP_CIPHER_CTX_free(cctx);
```
frankmorgner, line 408: please check with `if (cctx)`

frankmorgner (changes requested):

src/sm/sm-common.c
Jakuje: Unfortunately, the manual page does not specify this, but OpenSSL API is getting sane semantics and NULL arguments are ok for the free functions:
https://github.com/openssl/openssl/blob/f5b06306b7727f830407c639f8a7bf1698f8aaa5/crypto/evp/evp_enc.c#L84

src/sm/sm-common.c
Jakuje: Thanks for review. Changed to use SC_SUCCESS also in other cases of this function.

src/sm/sm-common.c
frankmorgner: Nice to know; is this a general approach for all APIs? Historically, they typically didn't check any input parameters...

Jakuje: I am not aware of specific OpenSSL guidelines, but I think it follows the least surprise principle as the `free()` ignores the `NULL` pointers and people are (finally) getting use to that too as it simplifies a lot of needless null checks on cleanup.

frankmorgner (approved):

Jakuje:
Thank you for the review. Merging to move on with this.

Issue #2127, opened 2020-10-02
Switch from simclist to sys/queue.h?


frankmorgner:
### Problem Description

While handling some memory leaks in simclist, I've noticed that the code is quite complex considering that we are just using ordinary lists in OpenSC. It also looks like, [simclist](http://mij.oltrelinux.com/devel/simclist/) hasn't received any updates since 10 years, nor does it seem to be widely used in other projects.

### Proposed Resolution

BSD based systems have [`sys/queue.h`](https://github.com/freebsd/freebsd/blob/master/sys/sys/queue.h), which offers simple linked lists. I think this is all we need here. Copying only the header file to `src/compat` would allow us using this on other platforms, it doesn't use any other bsd dependencies.

POSIX has [`search.h`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/search.h.html#tag_13_40) for hash tables and binary trees, though some of that API [is not so well designed](https://stackoverflow.com/a/14012047), but I don't think we need this complexity.

Jakuje:
Do you have some particular memory leak in mind? The simclist is not updated, but from my experience it just works. There are couple of changes over last years though. On the other hand, using sys queues is more standard these days also in Linux world. The advantage is that it is just macro based.

https://linux.die.net/man/3/queue

Grepping through the code, the lists are used only in src/pkcs11/pkcs11-global.c, src/pkcs11/slot.c, src/libopensc/ctx.c and cac, coolkey and idprime card drivers. Replacing this shold not be that hard.

frankmorgner:
most of the [last changes on simclist.c](https://github.com/OpenSC/OpenSC/commits/master/src/common/simclist.c) were bug fixes, but that's not the point. The main problem is that the implementation is complex and none of us has deep insights for maintaining it.

PR #2666, opened 2022-12-15
Support for IDPrime MD 830, 930 and 940 cards

xhanulik:
Driver `card-idprime.c` does not support some of the IDPrime cards - MD 830, 930 and 940.

The main problem is not matching key references that are assigned during the processing of the index file. The key references were recently set according to the OS version. However, some of the tested cards shared the same OS version but different key references of the private keys. The card matching is now done via ATR for particular card types, and the key reference offsets are distinguished according to `card->type`.

The `idprime_process_index()` function for processing the `0x0101` file adds the found certificate objects to the card. The index file contains entries which (probably) differ as the certificates were imported to the card with different tools. The entries can be
* `ksc` or `kxc` alone - card should have corresponding private and public key
* `ksc`/`kxc` with `pubkxs`/`pubkxc` - certificate is missing corresponding private key (CA certificates)
* `priprk`, `ksc`/`kxc` and `pubkxs`/`pubkxc` - card should have the corresponding private and public key

Each certificate entry contains an index and the offset used for getting the private key reference.
Certificates that are missing a private key on the card are also added to the card but with some dummy key reference, which is then processed in `sc_pkcs15emu_idprime_init()`, where only the public key is added.

IDPrime 930 and 940 cards can work with 4096b RSA keys, which is added to the card initialization.

Cards also support EC keys and certificates - the `sc_pkcs15emu_idprime_init()` is extended by adding EC keys to the card. When signing with ECDSA keys, the data for signature needs to be padded (probably to some multiple of 8 bytes) before sending to the card.

The changes were tested with cards
* IDPrime 3810 with RSA keys
* IDPrime MD 830-FIPS with RSA keys
* IDPrime 930 with RSA keys
* IDPrime 940 cards with 4096b RSA key and EC key

There are still some issues which need to be addressed
* key references (and offsets in `idprime_process_index()`) of EC keys may differ from the key references of RSA key
* IDPrime 940 cards support CC certificates which are listed as common certificates in the index file


xhanulik:
Added reading of cmap file, which contains records holding private key container names and index of corresponding certificates. It is used for the detection of CA certificates on the card.

For IDPrime 940, there is also a second PIN (Digital Signature PIN) used for accessing CC certificates on the card. File `0x0005` contains key references of private keys with PIN index and index of corresponding certificates. This is used for the classification of key objects to correct PIN.

src/libopensc/card-idprime.c
```diff
@@ -156,13 +438,19 @@ static int idprime_select_index(sc_card_t *card)
 	return r;
 }
 
-static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, int length)
+static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, int length, idprime_container_t *containers, idprime_keyref_t *keyrefmap)
```
Jakuje, line 441: can we store these in `idprime_private_data_t` instead of passing them as additional arguments?

src/libopensc/card-idprime.c
```diff
@@ -198,30 +486,56 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
 		/* in minidriver, mscp/kxcNN or kscNN lists certificates */
 		if (((memcmp(&start[4], "ksc", 3) == 0) || memcmp(&start[4], "kxc", 3) == 0)
 			&& (memcmp(&start[12], "mscp", 5) == 0)) {
+			int cert_id = 0;
+			idprime_container_t *container = NULL;
+
+			if (start[8] >= '0' && start[8] <= '9') {
```
Jakuje, line 492: should we check the bounds of `start[7]` too here?

xhanulik: I will fix this, but it should not be a problem as `kscXX` and `kxcXX` (hopefully) should have two-digit number index.

src/libopensc/card-idprime.c
```diff
@@ -128,6 +174,242 @@ static int idprime_select_idprime(sc_card_t *card)
...
+		}
+		new_keyref->index = start[2];
+		new_keyref->key_reference = start[1];
+		new_keyref->pin_index = start[0];
+		sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found key reference with index=%d, pin=%d, keyref=%d", new_keyref->index, new_keyref->pin_index, new_keyref->key_reference);
+
+		if (!current) {
+			*keyrefmap = new_keyref;
+		} else {
+			current->next = new_keyref;
```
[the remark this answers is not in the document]
xhanulik: Nothing special, there is just no need of re-sizing/setting some fixed size of the array.

src/libopensc/card-idprime.c
xhanulik: fixed

Jakuje (approved):

src/libopensc/card-idprime.c
```diff
@@ -65,33 +85,84 @@ static const sc_path_t idprime_path = {
...
+
+static void idprime_free_keyref(idprime_keyref_t *keyrefmap)
+{
+	idprime_keyref_t *next = NULL;
+	while (keyrefmap) {
+		next = keyrefmap->next;
+		free(keyrefmap);
+		keyrefmap = next;
+	}
+}
```
frankmorgner, line 158: idprime_keyref_t and idprime_container_t are defining list types. Wouldn't it be better to use the existing list types from simclist?

note that there is also https://github.com/OpenSC/OpenSC/issues/2127

xhanulik: Yes, rewritten in 6eb67f876c203c615498016c255036f9e1dd15d5.

src/libopensc/pkcs15.h
```diff
@@ -386,6 +386,8 @@ struct sc_pkcs15_prkey_info {
 
 	/* Non-pkcs15 data, like MD CMAP record */
 	struct sc_auxiliary_data *aux_data;
+
+	char *pin_id;
```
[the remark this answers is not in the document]
xhanulik: I removed the changes of `sc_pkcs15_prkey_info` and extended the `idprime_card_ctl()` to retrieve PIN of the current processed object.

src/libopensc/card-idprime.c
```diff
@@ -143,17 +267,130 @@ static int idprime_select_index(sc_card_t *card)
...
+		if (start[0] == 0) /* Empty key ref */
+			continue;
+	
+		new_keyref.index = start[2];
+		new_keyref.key_reference = start[1];
+		new_keyref.pin_index = start[7];
+		sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found key reference with index=%d, pin=%d, keyref=%d", new_keyref.index, new_keyref.pin_index, new_keyref.key_reference);
+
+		if (idprime_add_keyref_to_list(&priv->keyrefmap, &new_keyref) != SC_SUCCESS) {
+			LOG_FUNC_RETURN(card->ctx, r);
```
Jakuje, line 387:
```suggestion
			goto done;
```

src/libopensc/cardctl.h
```diff
@@ -305,6 +305,7 @@ enum {
 	SC_CARDCTL_IDPRIME_GET_NEXT_OBJECT,
 	SC_CARDCTL_IDPRIME_FINAL_GET_OBJECTS,
 	SC_CARDCTL_IDPRIME_GET_TOKEN_NAME,
+	SC_CARDCTL_IDPRIME_FINAL_GET_PIN_ID,
```
Jakuje, line 308: what is the significance of the "final" in this action? I think it should be just `SC_CARDCTL_IDPRIME_GET_PIN_ID`?

src/libopensc/card-idprime.c
```diff
@@ -111,6 +203,23 @@ idprime_private_data_t *idprime_new_private_data(void)
 		return NULL;
 	}
 
+	/* Initialize container list */
+	if (list_init(&priv->containers) != 0 ||
+	    list_attributes_comparator(&priv->containers, list_comparator_uint8_t) != 0 ||
```
Jakuje, line 208: I think you still need a comparator as you need to compare the member `index`, not the whole `idprime_container_t` casted to `uint8_t`. Same for the keyrefmap

Jakuje (commented):
Hopefully last comments

src/libopensc/cardctl.h
xhanulik: No significance, only copy-paste-leftover, thanks for spotting.

src/libopensc/card-idprime.c
```diff
@@ -143,17 +267,130 @@ static int idprime_select_index(sc_card_t *card)
...
+		for (j = 0; j < MAX_CONTAINER_NAME_LEN + 1; j++) {
+			if (start[2 * j] == 0)
+				break;
+			new_container.guid[j] = start[2 * j];
+		}
+
+		sc_debug(card->ctx, SC_LOG_DEBUG_VERBOSE, "Found container with index=%d, guid=%s", new_container.index, new_container.guid);
+
+		if (idprime_add_container_to_list(&priv->containers, &new_container) != SC_SUCCESS) {
+			LOG_FUNC_RETURN(card->ctx, r);
```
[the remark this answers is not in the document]
Jakuje: this is still unresolved

src/libopensc/card-idprime.c
```diff
@@ -405,11 +707,26 @@ static int idprime_fill_prkey_info(list_t *list, idprime_object_t **entry, sc_pk
...
 	return SC_SUCCESS;
 }
 
+/* get PIN id of the current object on the list */
+static int idprime_get_pin_id(list_t *list, idprime_object_t **entry, char **pin_id)
+{
+	if (pin_id == NULL || entry == NULL) {
+		return SC_ERROR_INVALID_ARGUMENTS;
+	}
+	*pin_id = "11"; // normal PIN id
```
frankmorgner, line 724: maybe you want to make `pin_id` a `const char **`  to indicate that the returned memory does not need to be `free`d.

frankmorgner (changes requested):
Please look at @Jakuje comment about the possible memory leak, thank you.

frankmorgner (approved):

Jakuje:
Thank you!

Issue #208, opened 2014-01-17
A Slovakian eID card driver?


martinvarga:
I am just started writing a driver for Slovakian eID for fun, despite there is a promise from authorities to provide some Linux support in future. 

Because I am a naive newbie in opensc and smart cards and till now I have no technical specification for this card, I appreciate any information and help from community. 

For now, the plan is to find suitable implementation of function from existing drivers and integrating them into the  driver. 

If you are interested you are welcome to join. 

For now, there is only detection of card and only default iso 7816 functions. 

/tools/opensc-tool --reader 0 --atr
3b:df:18:00:81:31:fe:58:00:31:b9:64:05:0e:01:00:73:b4:01:d3:00:00:00:22

varga@ntb-varga:src$ ./tools/opensc-tool --reader 0 --name
Slovakian eID

Martin

andrewshadura:
The newly published closed-source Linux Slovak eID client contains 8 XML files which may be of some interest. They all begin with
```
<iso:CardInfo xmlns:iso="urn:iso:std:iso-iec:24727:tech:schema"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="urn:iso:std:iso-iec:24727:tech:schema CardInfo.xsd">
```
and describe APDUs like this:
```
      <iso:CardCall>
        <!-- 3. Check eSign.PIN state -->
        <iso:CommandAPDU>00200081</iso:CommandAPDU>
        <iso:ResponseAPDU>
          <iso:Trailer>6A88</iso:Trailer>
        </iso:ResponseAPDU>
        <iso:ResponseAPDU>
          <iso:Trailer>6984</iso:Trailer>
        </iso:ResponseAPDU>
      </iso:CardCall>
```

frankmorgner:
Sounds like it's a software according to ISO 24727, similar to https://github.com/ecsec/open-ecard. The `CardInfo` files indeed say something about the card, but unfortunately don't tell the whole story.

But I could have a look at it, maybe the card is similar to the German ID card...

andrewshadura:
Yes, I found that already… Yet that Java application doesn't work here for a reason.

[...]

PR #2672, opened 2022-12-31
support for Slovak eID cards

jurajsarinay:
##### Checklist

- [ ] macOS tokend is tested

Please see https://github.com/OpenSC/OpenSC/issues/208#issuecomment-1319429080.

I propose to add a new driver to handle eID cards issued by Slovakia from mid-2021 until November 2022.

## Driver

The card runs CardOS 5.4, the new driver is therefore a stripped-down version of ```card-cardos.c``` The only place where I have to diverge from the original driver is ```set_security_env```, because the card expects ```MSE RESTORE``` instead of ```MSE SET```. I abuse ```key_ref``` to store the corresponding ```seIdentifier```, as our pkcs\#15 structures do not include the entry.

Because the card shares the ATR with other CardOS 5.4 cards, the new driver precedes ```cardos-driver``` within ```internal_card_drivers[]```.

## PKCS\#15 emulation

Within EF.DIR there are 5 applications. The last two carry (apparently not entirely usable) PKCS\#15 structures. 

```pkcs15-skeid.c``` binds the _fourth_ application in the list. Otherwise ```sc_pkcs15_bind_internal``` would get called and create an unusable token. In the case of the fifth application this is prevented by ```SC_PKCS11_FRAMEWORK_DATA_MAX_NUM = 4```.

Because there is no point in calling ```sc_pkcs15_bind_internal``` for this card, I added it to ```sc_pkcs15_is_emulation_only```. This does not prevent ```sc_pkcs15_bind_internal``` from [getting called](https://github.com/OpenSC/OpenSC/blob/70771735ae10180bb039043b9a1b00b66bf00fc1/src/libopensc/pkcs15.c#L1296
) though (if synthetic binding was unsuccessful).

I consider this behaviour a bit counterintuitive. I mention it here to report on and justify what I have done, it has no noticeable effect on my driver (any more). Let me know if ```sc_pkcs15_is_emulation_only``` warrants a separate GitHub issue.

## PINs

There is a global User PIN labeled BOK. The qualified certificate (key) requires user consent and a separate (local) Signature PIN labeled KEP. The "official" proprietary PKCS\#11 module requires both the codes for every signature. Fortunately, the card is happy with the Signature PIN only, as there seems to be no convenient way to have multiple PIN codes per slot. 

I considered emulating (parts of) the "official" behaviour wihtin a custom ```pin_cmd``` that contained the following:

```C
    if (data->pin_reference == 0x87 && data->cmd != SC_PIN_CMD_CHANGE && data->pin_type != SC_AC_CONTEXT_SPECIFIC)
        {
               sc_log(card->ctx, "Non-specific KEP PIN encountered, handling it as BOK instead.");
               data->pin_reference = 0x03;
       }

```
I ultimately decided against the idea. It adds complexity (or confusion) and provides little benefit. I mention the issue because it is connected to a failure in ```pkcs11-tool --test --slot 1 --login```: [log](https://github.com/OpenSC/OpenSC/files/10326667/pkcs11-tool_test_slot_1.log).

Because the local Signature PIN is used for the session, ```test_verify()``` fails [here](https://github.com/OpenSC/OpenSC/blob/70771735ae10180bb039043b9a1b00b66bf00fc1/src/tools/pkcs11-tool.c#L6639).

The card enforces CKA_ALWAYS_AUTHENTICATE and therefore reports that the Signature PIN is (no longer) verified, apparently because signatures have been computed during ```test_signature()```. The only effect of this is that the built-in test fails even though the token works (reasonably) well.

The above ```pin_cmd``` hack would result in a passed ```pkcs11-tool --test --slot 1 --login```.

I include this information mainly to justify a PR with a failed test attached. If the behaviour of ```pkcs11-tool --test``` in the context of a local Signature PIN and ```user_consent``` warrants a separate GitHub issue, do please let me know.

For completeness,  [here](https://github.com/OpenSC/OpenSC/files/10326703/pkcs11-tool_test_slot_0.log) is the output of ```pkcs11-tool --test --slot 0 --login```.


src/libopensc/pkcs15-skeid.c
```diff
@@ -0,0 +1,189 @@
...
+		sc_bin_to_hex(p15card->card->serialnr.value, sn_len, buf,
+			2 * sn_len + 1, 0);
+		p15card->tokeninfo->serial_number = buf;
+	}
+
+	p15card->tokeninfo->flags = SC_PKCS15_TOKEN_PRN_GENERATION | SC_PKCS15_TOKEN_READONLY;
+
+	for (i = 0; i < 3; i++) {
+		const char *skeid_cert_names[3] = {
+			"Kvalifikovany certifikat",
```
martinpaljak, line 61: Not related to this PR, but it would be lovely to get some kind of i18n in OpenSC... A reminder-issue, not likely someone will work on this tho :)

src/libopensc/pkcs15-skeid.c
```diff
@@ -0,0 +1,189 @@
...
+
+		r = sc_pkcs15emu_add_x509_cert(p15card, &cert_obj, &cert_info);
+
+		LOG_TEST_RET(p15card->card->ctx, r, "Error adding certificate.");
+	}
+
+	/* add pins */
+	for (i = 0; i < 2; i++) {
+		const char *skeid_pin_names[2] = {
+			"BOK",
```
martinpaljak, line 92: Are the terms widely used? Or are they from specification? See my comment above - I know from experience that as these values are presented to users, they better be "average joe readable" (even if not precise)

martinpaljak:
@jurajsarinay don't take the tests at gold standard value, if it can be fixed in a generalized way to take into account "always authenticate" behavior, would be a lovely addition together with this PR.

src/libopensc/pkcs15-skeid.c
martinpaljak: Nobody outside of EU/EIDAS bureaucrats and related techies call what is in human speak a "signature certificate" (maybe even a signature key) a "qualified certificate". EU bureaucrats also want us to use "qualified website certificates" but that is not happening for other technical/political/sane reasons. I'd recommend "precise enough but human readable labels" unless you claim that everyone in .sk speaks "legally valid terminology, also early in the morning before coffee" in which case I'll start using qualified website authentication certificates, also before morning coffee.

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,180 @@
...
+
+	assert(card != NULL && env != NULL);
+
+	if (!(env->flags & SC_SEC_ENV_KEY_REF_PRESENT) || env->key_ref_len != 1) {
+		sc_log(card->ctx, "No or invalid key reference\n");
+		return SC_ERROR_INVALID_ARGUMENTS;
+	}
+
+	key_id = env->key_ref[0];
+	r = sc_restore_security_env(card, key_id);
```
Jakuje, line 155: This is at least weird. I would be probably for reimplementing the `iso7816_restore_security_env` (its just one APDU) with some comments, instead of reusing it for setting the security environment (with the key id instead of security environment number). The restore of security environment is completely different operation and that it works is probably pure coincidence.

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,180 @@
...
+	LOG_TEST_RET(card->ctx, r,  "APDU transmit failed");
+	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
+		return SC_ERROR_INTERNAL;
+	if (apdu.resplen == 8) {
+		/* cache serial number */
+		memcpy(card->serialnr.value, rbuf, 8);
+		card->serialnr.len = 8;
+	} else {
+		sc_log(card->ctx,  "unexpected response to GET DATA serial"
+			" number\n");
```
Jakuje, line 98: no need to break a line here in the middle of the string. And again, the `\n` should not be needed.

Jakuje (changes requested):

src/libopensc/pkcs15-skeid.c
martinpaljak: Just wondering - is the "signing certificate" actually used for authentication? (technically a signing operation, but not qualified) ? If/how does the card support (online) authentication?

src/libopensc/pkcs15-skeid.c
```diff
@@ -0,0 +1,189 @@
...
+
+static int sc_pkcs15emu_skeid_init(sc_pkcs15_card_t * p15card)
+{
+	int r;
+	int i;
+	int sn_len;
+	char *buf;
+
+	set_string(&p15card->tokeninfo->label, "eID karta");
+	set_string(&p15card->tokeninfo->manufacturer_id, "Ministerstvo vnutra SR");
```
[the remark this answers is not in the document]
martinpaljak: This is just leftover from old structures and people putting there things that "make sense". From product perspective - do you expect to see there the silicon manufacturer? module manufacturer? card manufacturer? government-tendered service provider? Anything fits, I somehow recall it was mandatory field (at least in OpenSC code)?

src/libopensc/pkcs15-skeid.c
frankmorgner: I have started to support multiple languages in https://github.com/OpenSC/OpenSC/blob/master/src/ui/strings.h
Text is adjusted according to locale or via opensc.conf

Jakuje: I dont think the language here is an issue as the eID cards are issued only to people with at least basic understanding of the Slovak language.

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,180 @@
...
+		const sc_security_env_t *env,
+		int se_num)
+{
+	int key_id;
+	int r;
+
+	assert(card != NULL && env != NULL);
+
+	if (!(env->flags & SC_SEC_ENV_KEY_REF_PRESENT) || env->key_ref_len != 1) {
+		sc_log(card->ctx, "No or invalid key reference\n");
```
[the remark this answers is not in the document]
Jakuje: yeah. There is a lot of old code, but there is no point bringing the issues from the old code to the new one.

frankmorgner:
I wonder if card-skeid.c is needed at if the difference from card-cardos.c is so small. card-cardos.c is written in a generic way, which the slowak card doesn't need. I think it should be better to use skeid_known_url() in cardos_init to get the card type and later to use the card type in cardos_set_security_env to implement the workaround for the slowak card.

@Jakuje , since you are familiar with the cardos specification, do you know where the use of MSE RESTORE originates and if it could be integrated in a general way into card-cardos.c?

frankmorgner:
> Apart from using a specific sc_set_security_env() I also simply use iso7816_compute_signature() and iso7816_decipher() instead of the CardOS-specific functions.

OK, fair enough, card-skeid.c is simple enough to be maintained seperately.

src/libopensc/ctx.c
```diff
@@ -98,6 +98,7 @@ struct _sc_driver_entry {
 };
 
 static const struct _sc_driver_entry internal_card_drivers[] = {
+	{ "skeid",	(void *(*)(void)) sc_get_skeid_driver },
```
frankmorgner, line 101: please add a comment here why skeid needs to be before cardos

frankmorgner (changes requested):

src/libopensc/pkcs15-skeid.c
andrewshadura: @Jakuje, not true actually: the same chip is in the residence permits issues to foreigners, and also in the "alternative authenticator" cards issued to those who doesn’t have a residence permit.

andrewshadura:
@jurajsarinay, you should probably rebase your commits on top of the current `master` instead of merging it in.

andrewshadura:
Oh, I never used that button, I always rebase and force-push manually 🙂

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,183 @@
...
+	u8 buf[SKEID_KNOWN_URL_LEN];
+
+	sc_path_t url_path;
+
+	int r = SC_ERROR_WRONG_CARD;
+
+	sc_path_set(&url_path, SC_PATH_TYPE_DF_NAME, skeid_aid_eid.value, skeid_aid_eid.len, 0, 0);
+
+	if (sc_select_file(card, &url_path, NULL) == SC_SUCCESS
+		&&  sc_get_data(card, 0x7F62, buf, SKEID_KNOWN_URL_LEN) == SKEID_KNOWN_URL_LEN
```
Jakuje, line 65:
```suggestion
		&& sc_get_data(card, 0x7F62, buf, SKEID_KNOWN_URL_LEN) == SKEID_KNOWN_URL_LEN
```

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,183 @@
...
+	sc_log(card->ctx,  "Slovak eID card v3 (CardOS 5.4)");
+
+	return 1;
+}
+
+static int skeid_get_serialnr(sc_card_t *card)
+{
+	int r;
+	sc_apdu_t apdu;
+	u8  rbuf[SC_MAX_APDU_BUFFER_SIZE];
```
Jakuje, line 86:
```suggestion
	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
```

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,183 @@
...
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, 0xca, 0x01, 0x81);
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+	apdu.le = 256;
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r,  "APDU transmit failed");
```
Jakuje, line 93:
```suggestion
	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
```

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,183 @@
...
+	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
+		return SC_ERROR_INTERNAL;
+	if (apdu.resplen == 8) {
+		/* cache serial number */
+		memcpy(card->serialnr.value, rbuf, 8);
+		card->serialnr.len = 8;
+	} else {
+		sc_log(card->ctx,  "unexpected response to GET DATA serial number");
```
Jakuje, line 101:
```suggestion
		sc_log(card->ctx, "unexpected response to GET DATA serial number");
```

src/libopensc/card-skeid.c
Jakuje: If others dont have issue with this. I am ok with keeping this as long as it will contain some comment in the code that this is what the official driver does. Without comment it still looks like an error (and comments on github are not guaranteed to be accessible forever).

src/libopensc/card-skeid.c
```diff
@@ -0,0 +1,183 @@
...
+}
+
+struct sc_card_driver * sc_get_skeid_driver(void)
+{
+	if (iso_ops == NULL) iso_ops = sc_get_iso7816_driver()->ops;
+	skeid_ops = *iso_ops;
+	skeid_ops.match_card = skeid_match_card;
+	skeid_ops.init = skeid_init;
+	skeid_ops.set_security_env = skeid_set_security_env;
+	skeid_ops.logout  = skeid_logout;
```
Jakuje, line 181:
```suggestion
	skeid_ops.logout = skeid_logout;
```

src/libopensc/pkcs15-skeid.c
```diff
@@ -0,0 +1,188 @@
...
+
+		r = sc_pkcs15emu_add_rsa_prkey(p15card, &prkey_obj, &prkey_info);
+
+		LOG_TEST_RET(p15card->card->ctx, r, "Error adding private key.");
+	}
+	LOG_FUNC_RETURN(p15card->card->ctx, SC_SUCCESS);
+}
+
+int sc_pkcs15emu_skeid_init_ex(sc_pkcs15_card_t * p15card,
+			struct sc_aid *aid)
```
Jakuje, line 179: needless line break and some whitespace cleanup
```suggestion
int sc_pkcs15emu_skeid_init_ex(sc_pkcs15_card_t *p15card, struct sc_aid *aid)
```

Jakuje (commented):
minor whitespace nits, but otherwise looks good

frankmorgner (approved):

Jakuje (approved):

frankmorgner:
done, thank you

PR #2674, opened 2023-01-09
EAC: Fix compatiility with OpenSSL 3.0 and remove "APDU firewall"

frankmorgner:
In #2646 the EAC implementation has been tested with OpenSSL 3.0 and it was found that the APDU encoding doesn't work for OpenSSL 3+ anymore (see [this discussion](https://github.com/OpenSC/OpenSC/pull/2646#discussion_r1035462442)). In order to fix this problem, the encoding/decoding has now been rewritten with OpenSC's ASN.1 parser.

Additionally this PR removes some sanity checks when transmitting encrypted APDUs. This is only useful when OpenSC is used as "transport layer" for communicating with a card, e.g. in [ccid-emulator](https://frankmorgner.github.io/vsmartcard/ccid/README.html). Since this is a niche use case, sanity checks are removed from OpenSC and should now be done in the calling application if needed.

Also, note the small fixes to the build system.

##### Checklist

- [ ] macOS tokend is tested


llogar:
I can confirm this fixes the problem mentioned in #2646. I hope it gets accepted soon, so I can then rebase the eOI support.

PR #2681, opened 2023-01-21
Remove onepin-opensc-pkcs11

frankmorgner:
The onepin-opensc-pkcs11 constantly leads to confusion with users, because it is not obvious which opensc module to choose. This PR uses the running program's name to check whether the onepin configuration is needed, i.e. it checks whether the PKCS#11 module is running inside firefox. 

The OpenSC tools are refactored so that they can output debug messages even before to a reader (i.e. directly after `sc_context_create()`.

##### Checklist

- [x] Documentation updated
- [x] Tools tested
- [x] Windows Firefox tested
- [x] Linux Firefox tested
- [x] macOS Firefox tested


frankmorgner:
I've now successfully tested all major platforms. Everything works as expected. I've noticed that the PKCS#11 module generates too many notifications for removal/insertion events. That's removed now as well.

@Jakuje, @dengert, please let me know if you think that removing onepin-opensc-pkcs11 may cause any trouble for users.

metsma:
I will notifiy our test team

src/tools/goid-tool.c
```diff
@@ -855,23 +855,21 @@ main(int argc, char **argv)
     memset(&ctx_param, 0, sizeof(ctx_param));
     ctx_param.ver      = 0;
     ctx_param.app_name = app_name;
+	ctx_param.debug    = cmdline.verbose_given;
+    if (cmdline.verbose_given > 1)
+		ctx_param.debug_file = stderr;
```
Jakuje, line 860: there is some mixture of tabs and spaces for indentation.

src/libopensc/ctx.c
```diff
@@ -804,6 +808,41 @@ int sc_context_repair(sc_context_t **ctx_out)
 	return SC_SUCCESS;
 }
 
+static char *get_exe_path()
+{
+	/* Find the executable's path which runs this code.
+	 * See * https://github.com/gpakosz/whereami/ for
```
Jakuje, line 814:
```suggestion
	 * See https://github.com/gpakosz/whereami/ for
```

Jakuje (commented):
please, fix the typos. Otherwise it looks good.

src/tools/goid-tool.c
frankmorgner: hmm, this file uses spaces only. I'll change the PR to stick to this.

frankmorgner:
typos are fixed.

Jakuje:
Please, remove also the onepin from the spec file in https://github.com/OpenSC/OpenSC/blob/master/packaging/opensc.spec#L164 so we can have the CI passing.

dengert:
PIV should be fine. Only uses 1 pin

Jakuje (approved):

frankmorgner:
Let's still wait for @metsma 's feedback

metsma:
> Let's still wait for @metsma 's feedback

Can we filter Chrome to?

henning-schild:
I did not look into the details, but this sounds like a user interface change that should not be ignored.

So i suggest a "symlink" or a "deprecation wrapper". If the module disappears ... we will have an issue with any config in any application that is configured to use it.

frankmorgner:
> I did not look into the details, but this sounds like a user interface change that should not be ignored.

Indeed it is, that's why [I asked for opinions above](https://github.com/OpenSC/OpenSC/pull/2681#issuecomment-1401564566). We can surely implement some temporary workarounds, but I fear that those are hard to maintain given the different platforms we support and that I don't see them go away once they are in place. Hence, I personally would opt for a possible breaking change nevertheless.

frankmorgner:
> Can we filter Chrome to?

With all the browser forks it could get crazy quickly, but it is possible, for sure. But, AFAIK, problems have only been reported for Firefox....

metsma:
> > Can we filter Chrome to?
> 
> With all the browser forks it could get crazy quickly, but it is possible, for sure. But, AFAIK, problems have only been reported for Firefox....

We have only official chrome in testing matrix. Chromium uses SNAP in Ubuntu and we don't supprt this configuration.

henning-schild:
> Indeed it is, that's why [I asked for opinions above](https://github.com/OpenSC/OpenSC/pull/2681#issuecomment-1401564566). We can surely implement some temporary workarounds, but I fear that those are hard to maintain given the different platforms we support and that I don't see them go away once they are in place. Hence, I personally would opt for a possible breaking change nevertheless.

Indeed such things are hard to get rid of. But i would still like to ask for some help for end users and distributors.

But these modules are mentioned in all sorts of management scripts that configure browsers (many of them) and other applications, some of them proprietary ... Think of ansible or puppet and alike. These scripts will have to work over generations and flavors of Linux, and maybe other OSs. And they will have to deal with creating new configuration and updating old.

Not even to mention manual configuration that people have somewhere, did after reading whatever google did spit out ... a forum/blog.

I am afraid the user-interface change might be big and hard to handle for distros, admins, and "did something funny" end-users.

frankmorgner:
@henning-schild I've added the onepin module as link/copy to the installation. I need to verify the Windows installer, however.

henning-schild:
> @henning-schild I've added the onepin module as link/copy to the installation. I need to verify the Windows installer, however.

Thanks so much. If i read that last commit correctly it is only about compatibility and not about deprecation.

I sure hope the old name can be deprecated one day, or kept forever and not causing too much trouble over time.

frankmorgner:
@metsma , could you please elaborate what you mean by
> Can we filter Chrome to?

Do you think OpenSC should behave like the onepin module in Chrome as well? If so, why?

Jakuje:
> @metsma , could you please elaborate what you mean by
> 
> > Can we filter Chrome to?
> 
> Do you think OpenSC should behave like the onepin module in Chrome as well? If so, why?

Because chrome is using NSS for accessing hardware tokens as far as I know.

metsma:
Yes. On linux Chrome uses NSS

frankmorgner:
I don't think it makes sense to behave like the onepin module in any application that uses NSS. There haven't been any reports about chrome on Linux having problems with multiple pins. I'd expect NSS should always have been configured with opensc-pkcs11 (i.e. **not** the onepin behavior), because of its wider use. But still, no reports for chrome having problems in Linux.

metsma:
Our testers can verify Chrome prompts PIN2 and PIN1.
The reason is that NSS thinks Certificate object is read protected and one must before reading use C_Login.
There is flag that can alter this behavior PKCS11_PUB_READABLE_CERT_FLAG.
https://github.com/open-eid/firefox-pkcs11-loader/blob/master/webextension/background.js#L30
You can observe this flag behavior is in effect when Firefox/Chrome asks certificate first and then PIN prompt.
But there is another issue that NSS configuration forgets this flag after some time.
Thats the whole reason why the onepin solution was created.

Jakuje:
We also have PKCS#11 3.0 profiles, where `CKP_PUBLIC_CERTIFICATES_TOKEN` [1] should tell NSS that it does not have to login to the slot before reading the certificates and NSS should follow this. But this needs to be implemented in OpenSC per-driver as we can not make this assumption opensc-wise. What token/driver do you use for your tests?

[1] https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.1/csd02/pkcs11-profiles-v3.1-csd02.pdf

metsma:
OpenSC driver and EstEID 2015/2018 cards.

Jakuje:
> OpenSC driver and EstEID 2015/2018 cards.

if you can make these assumptions from the driver side that all public keys are visible before login, it should present this profile to NSS automatically:

https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L1561-L1564

metsma:
If you follow the NSS code then they don't look CKP_PUBLIC_CERTIFICATES_TOKEN.
Only some custom SECMOD_FRIENDLY_FLAG (matches our JS definition const PKCS11_PUB_READABLE_CERT_FLAG = 0x1<<28) or internal module.
Any other case leads to Slot login, thats where it tries to login PIN1 and PIN2 slot.

https://github.com/servo/nss/blob/949eb9848f4fa5f83756f3ab7fdf9b0d3f20d37f/cmd/certutil/certutil.c#L433
```
    if (!PK11_IsFriendly(slot) && PK11_NeedLogin(slot)) {
        SECStatus newrv = PK11_Authenticate(slot, PR_TRUE, pwarg);
```
https://github.com/servo/nss/blob/949eb9848f4fa5f83756f3ab7fdf9b0d3f20d37f/lib/pk11wrap/pk11slot.c#L1590
```
PK11_IsFriendly(PK11SlotInfo *slot)
{
    /* internal slot always has public readable certs */
    return (PRBool)(slot->isInternal || 
		    ((slot->defaultFlags & SECMOD_FRIENDLY_FLAG) == 
		     SECMOD_FRIENDLY_FLAG));
```
https://github.com/servo/nss/blob/949eb9848f4fa5f83756f3ab7fdf9b0d3f20d37f/lib/pk11wrap/pk11slot.c#L1581
https://github.com/servo/nss/blob/949eb9848f4fa5f83756f3ab7fdf9b0d3f20d37f/lib/pk11wrap/pk11slot.c#L1123
```
    slot->needLogin = ((tokenInfo.flags & CKF_LOGIN_REQUIRED) ? 
							PR_TRUE : PR_FALSE);
```

frankmorgner:
@metsma thanks for sharing insights on why the onepin configuration was actually created. With your understanding of NSS and the possibility to reproduce the problem in Chrome, could you suggest a fix by setting the correct flags (i.e. CKP_PUBLIC_CERTIFICATES_TOKEN or SECMOD_FRIENDLY_FLAG) within OpenSC?

Jakuje:
The NSS code linked is 10 years old. The current code of `PK11_IsFriendly()` checks the `CKP_PUBLIC_CERTIFICATES_TOKEN` profile:

https://github.com/nss-dev/nss/blob/master/lib/pk11wrap/pk11slot.c#L1793-L1799

metsma:
I wonder which version then chrome uses.
https://chromium.googlesource.com/chromium/deps/nss/

TheElectronWill:
> Do you think OpenSC should behave like the onepin module in Chrome as well? If so, why?

Hello!
I might be a "did something funny" user, so I apologize for that in advance.

I've got an Atos CardOS SmartCard with two "extra PIN". If I use `opensc-pkcs11.so`, I get at least 3 PIN prompts instead of one, in Firefox _and_ in Chrome. If I use the `libcardos11.so`, I get 2 PIN prompts. _Using the onepin module was the only workaround I've found to get only one PIN prompt_ (on Windows, I only get one PIN prompt), and it works very well. Again, this is both in Chrome and Firefox. But I might be talking about something that has been fixed in a recent PR? :thinking:

Jakuje:
> > Do you think OpenSC should behave like the onepin module in Chrome as well? If so, why?
> 
> Hello! I might be a "did something funny" user, so I apologize for that in advance.
> 
> I've got an Atos CardOS SmartCard with two "extra PIN". If I use `opensc-pkcs11.so`, I get at least 3 PIN prompts instead of one, in Firefox _and_ in Chrome. If I use the `libcardos11.so`, I get 2 PIN prompts. _Using the onepin module was the only workaround I've found to get only one PIN prompt_ (on Windows, I only get one PIN prompt), and it works very well. Again, this is both in Chrome and Firefox. But I might be talking about something that has been fixed in a recent PR? thinking

Thank you! I think the Chrome should be handled too. Not sure if the `chromium` binary will need also exception, similarly as the other rebuilds ...

This PR should keep it working as it was, while keeping only one `libopensc-pkcs11.so`. For now, the `*onepin` library is the same as `libopensc-pkcs11.so`, but built with different parameters so you use case should keep working. Either by making a symlink from the old location where the `onepin` was or by using the `opensc-pkcs11.so` directly.

TheElectronWill:
Addendum: gnome evolution uses the SmartCard too, I think it does so via the NSS database (like Chrome). Maybe it should be handled as well? But who knows how many apps use the nssdb...

src/tools/pkcs11-register.c
```diff
@@ -235,11 +235,6 @@ add_module_firefox(const char *module_path, const char *module_name, const char
 #endif
 	};
 
-	if (0 == strcmp(module_path, default_pkcs11_provider)) {
-		module_path = default_onepin_pkcs11_provider;
-		exclude_module_path = default_pkcs11_provider;
-	}
-
```
henning-schild, line 242: did you try what will happen if any old `pkcs11.txt` still has

```
library=/usr/lib64/onepin-opensc-pkcs11.so
name=OpenSC smartcard framework (0.22)
```

One would expect the old entry to be removed and replaced with a new one.

I guess you did, but i really do not trust this application since it is very broken.

When i just looked at my firefox config i again found a conflict ...

```
library=/usr/lib64/p11-kit-proxy.so
name=my-nss-p11-kit-proxy.so
library=/usr/lib64/onepin-opensc-pkcs11.so
name=OpenSC smartcard framework (0.22)
```

where `my-nss-p11-kit-proxy.so` deals with the smartcard and `pkcs11-register` is doing all the wrong things

frankmorgner: This change is related to *always* adding `opensc-pkcs11.so` instead of `onepin-opensc-pkcs11.so` regardless of the hosting app since the application is the same with this PR. this change is not related to limitations of pkcs11-register. If you have new input regarding the latter, please use the existing issues or a new issue.

frankmorgner:
With #2733 being merged, I wonder if the special treatment of Firefox (or any other NSS powered app) is still needed. OpenSC now behaves *friendly* in terms of having publicly readable certificates! @metsma could you confirm this by testing `master`?

frankmorgner:
@metsma, do you have any updates regarding OpenSC's "friendlyness", especially with https://github.com/OpenSC/OpenSC/pull/2733 being merged?

metsma:
@frankmorgner our test team reports that Firefox and Chrome still request 2 pins (PIN1, PIN2) on authentication.

frankmorgner:
> @frankmorgner our test team reports that Firefox and Chrome still request 2 pins (PIN1, PIN2) on authentication.

I assume you mean that FF/Chrome are requesting the two PINs although only one would be needed for authenticating to a specific site. If so, it sseems that although OpenSC is considered to have the certificates friendly-ish readable, the user experience is still poor.

Then, I would try to extend this PR to also recognize `chrome` and `chromium`. Do you know (platform independant) names of other browsers, that are using NSS?

metsma:
> > @frankmorgner our test team reports that Firefox and Chrome still request 2 pins (PIN1, PIN2) on authentication.
> 
> I assume you mean that FF/Chrome are requesting the two PINs although only one would be needed for authenticating to a specific site. If so, it sseems that although OpenSC is considered to have the certificates friendly-ish readable, the user experience is still poor.

Yes

> 
> Then, I would try to extend this PR to also recognize `chrome` and `chromium`. Do you know (platform independant) names of other browsers, that are using NSS?

I think this will cover most. Maybe add configure option to extend this list. There are other forks available (Tor, Opera etc)

TheElectronWill:
> Then, I would try to extend this PR to also recognize chrome and chromium. Do you know (platform independant) names of other browsers, that are using NSS?

Wouldn't the problem show up in a potentially unlimited number of applications that use NSS? E.g. opera, thunderbird, evolution, some vpn clients, electron-based apps, etc. Listing them all in opensc seems impossible, perhaps there could be a config somewhere :S

frankmorgner:
I think a reasonable good default regular expression for matching the executable's name would work and could be made configurable.

Matching all NSS programs, would not make sense, because in E-Mail applications you want to do signing *and* encryption even if multiple PINs are involved.

frankmorgner:
- resolved merge conflicts
- extended one-slot configuration (onepin) to chrome and chromium
- extended opensc.conf to use the apps path for a configuration block. Which allows setting `slots_per_card` to `1` also for other browsers while leaving the general PKCS#11 configuration alone.

frankmorgner:
@metsma , could you trigger your test team, please?

PR #2682, opened 2023-01-21
Handle reader limits for SC Card unwrap operations - Fixes #2514

### LINKED ISSUE #2514: sc-hsm-tool unable to unwrap 4096 bit keys in SmartCard-HSM 4K chip card - crashes with "Card command failed"
### Problem Description
The SmartCard-HSM 4K can generate and hold 4096 bit RSA keys.

However, they cannot be imported into a second smart card using the wrap/unwrap mechanism unlike smaller keys.
wrap / unwrap works with key sizes up to 3072 bits but fails with 4096 bits.

Re-Import into the same card fails as well as into a second card that was initialized with the same dkek.

After the failure the card must be re-inserted into the reader to be accessible again.

Tried with packaged version 0.22 (debian) as well as build from source (master).
The error occurs on smartcards with firmware 3.4 (as shipped) as well as on an upgraded smartcard with firmware 3.5.
Used a ReinerSCT cyberJack reader.

### Proposed Resolution

Have someone with deep knowledge of OpenSC internals and access to SmartCard-HSM 4K chip cards debug the problem.

### Steps to reproduce

1. Buy at least two SmartCard-HSM 4K smartcards (https://www.cardomatic.de/epages/64510967.sf/de_DE/?ObjectPath=/Shops/64510967/Products/SmartCard-HSM-4K-SIM)
2. Buy a ReinerSCT cyberJack (https://shop.reiner-sct.com/chipkartenleser-fuer-die-sicherheitsklasse-3/cyberjack-rfid-komfort-usb)
3. Initialize the smartcard HSM, e.g. `sc-hsm-tool -X -s 3`
4. create a large key `pkcs11-tool --module opensc-pkcs11.so -l --keypairgen --key-type rsa:4096 --id 10 --label large`
5. for reference, create a smaller key `pkcs11-tool --module opensc-pkcs11.so -l --keypairgen --key-type rsa:3072 --id 20 --label small`
6. wrap the keys `sc-hsm-tool -i 1 -W large.wrap` and `sc-hsm-tool -i 2 -W small.wrap`
7. try to unwrap

Unwrap fails with the long key
```
sc-hsm-tool -r 0 -i 8 -U large.wrap 
Wrapped key contains:
  Key blob
  Private Key Description (PRKD)
  Certificate
Enter User PIN : 

sc_card_ctl(*, SC_CARDCTL_SC_HSM_UNWRAP_KEY, *) failed with Card command failed
```
Unwrap succeeds with the smaller key
```
sc-hsm-tool -r 0 -i 11 -U small.wrap 
Wrapped key contains:
  Key blob
  Private Key Description (PRKD)
  Certificate
Enter User PIN : 

Key successfully imported
```

### Logs

I will disclose the full logs only if necessary (because I do not know which parts of the payload are sensitive).

The core of the problem seems to be this part (only in the log in the failing case):
```
P:103751; T:0x140054954984512 20:52:12.408 [sc-hsm-tool] reader-pcsc.c:242:pcsc_internal_transmit: called
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] reader-pcsc.c:333:pcsc_transmit: 
Incoming APDU (2 bytes):
***
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] apdu.c:537:sc_transmit: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] card.c:523:sc_unlock: called
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] reader-pcsc.c:737:pcsc_unlock: called
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] card-coolkey.c:991:coolkey_apdu_io: result r=0 apdu.resplen=0 sw1=6a sw2=86
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] card-coolkey.c:872:coolkey_check_sw: sw1 = 0x6a, sw2 = 0x86
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] iso7816.c:128:iso7816_check_sw: Incorrect parameters P1-P2
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] card-coolkey.c:1001:coolkey_apdu_io: Transmit failed
P:103751; T:0x140054954984512 20:52:12.421 [sc-hsm-tool] card-coolkey.c:1019:coolkey_apdu_io: returning with: -1205 (Incorrect parameters in APDU)
```

The failing log ends with: 

```
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] card-sc-hsm.c:1416:sc_hsm_unwrap_key: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] apdu.c:548:sc_transmit_apdu: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] card.c:473:sc_lock: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] reader-pcsc.c:685:pcsc_lock: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] card.c:513:sc_lock: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] apdu.c:515:sc_transmit: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] apdu.c:363:sc_single_transmit: called
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] apdu.c:367:sc_single_transmit: CLA:80, INS:74, P1:8, P2:93, data(1099) 0x7ffe039d2a38
P:103751; T:0x140054954984512 20:52:17.408 [sc-hsm-tool] reader-pcsc.c:323:pcsc_transmit: reader 'REINER SCT cyberJack RFID komfort (4639352370) 00 00'
P:103751; T:0x140054954984512 20:52:17.409 [sc-hsm-tool] reader-pcsc.c:324:pcsc_transmit: 
Outgoing APDU (1106 bytes):
***
P:103751; T:0x140054954984512 20:52:17.409 [sc-hsm-tool] reader-pcsc.c:242:pcsc_internal_transmit: called
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] reader-pcsc.c:333:pcsc_transmit: 
Incoming APDU (2 bytes):
***
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] apdu.c:537:sc_transmit: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] card.c:523:sc_unlock: called
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] reader-pcsc.c:737:pcsc_unlock: called
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] iso7816.c:128:iso7816_check_sw: No precise diagnosis
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] card-sc-hsm.c:1429:sc_hsm_unwrap_key: Check SW error: -1200 (Card command failed)
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] card.c:1065:sc_card_ctl: returning with: -1200 (Card command failed)
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] card.c:414:sc_disconnect_card: called
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] reader-pcsc.c:670:pcsc_disconnect: REINER SCT cyberJack RFID komfort (4639352370) 00 00:SCardDisconnect returned: 0x00000000
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] card.c:436:sc_disconnect_card: returning with: 0 (Success)
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] ctx.c:931:sc_release_context: called
P:103751; T:0x140054954984512 20:52:17.411 [sc-hsm-tool] reader-pcsc.c:974:pcsc_finish: called
```

istr:
Fixes #2514

Using reader with a card: REINER SCT cyberJack RFID komfort (4639352370) 00 00
SmartCard-HSM version 3.4

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-sc-hsm.c
+++ b/src/libopensc/card-sc-hsm.c
@@ -145,9 +145,7 @@ static int sc_hsm_select_file_ex(sc_card_t *card,
 
 	if (file_out == NULL) {				// Versions before 0.16 of the SmartCard-HSM do not support P2='0C'
 		rv = sc_hsm_select_file_ex(card, in_path, forceselect, &file);
-		if (file != NULL) {
-			sc_file_free(file);
-		}
+		sc_file_free(file);
 		return rv;
 	}
 
@@ -181,9 +179,7 @@ static int sc_hsm_select_file_ex(sc_card_t *card,
 			LOG_TEST_RET(card->ctx, rv, "Could not select SmartCard-HSM application");
 
 			if (priv) {
-				if (priv->dffcp != NULL) {
-					sc_file_free(priv->dffcp);
-				}
+				sc_file_free(priv->dffcp);
 				// Cache the FCP returned when selecting the applet
 				sc_file_dup(&priv->dffcp, *file_out);
 			}
@@ -730,12 +726,12 @@ static int sc_hsm_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 			u8 recvbuf[SC_MAX_APDU_BUFFER_SIZE];
 #ifdef ENABLE_SM
 			if (card->sm_ctx.sm_mode != SM_MODE_TRANSMIT) {
-				sc_log(card->ctx, 
+				sc_log(card->ctx,
 						"Session PIN generation only supported in SM");
 				LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 			}
 #else
-			sc_log(card->ctx, 
+			sc_log(card->ctx,
 					"Session PIN generation only supported in SM");
 			LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 #endif
@@ -746,7 +742,7 @@ static int sc_hsm_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 			apdu.le = 0;
 			if (sc_transmit_apdu(card, &apdu) != SC_SUCCESS
 					|| sc_check_sw(card, apdu.sw1, apdu.sw2) != SC_SUCCESS) {
-				sc_log(card->ctx, 
+				sc_log(card->ctx,
 						"Generating session PIN failed");
 				LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 			}
@@ -756,12 +752,12 @@ static int sc_hsm_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
 							apdu.resplen);
 					data->pin2.len = apdu.resplen;
 				} else {
-					sc_log(card->ctx, 
+					sc_log(card->ctx,
 							"Buffer too small for session PIN");
 				}
 			}
 		} else {
-			sc_log(card->ctx, 
+			sc_log(card->ctx,
 					"Session PIN not supported for this PIN (0x%02X)",
 					data->pin_reference);
 		}
@@ -848,45 +844,59 @@ static int sc_hsm_write_ef(sc_card_t *card,
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
 	}
 
-	p = cmdbuff;
-	*p++ = 0x54;
-	*p++ = 0x02;
-	*p++ = (idx >> 8) & 0xFF;
-	*p++ = idx & 0xFF;
-	*p++ = 0x53;
-	if (count < 128) {
-		*p++ = (u8) count;
-		len = 6;
-	} else if (count < 256) {
-		*p++ = 0x81;
-		*p++ = (u8) count;
-		len = 7;
-	} else {
-		*p++ = 0x82;
-		*p++ = (count >> 8) & 0xFF;
-		*p++ = count & 0xFF;
-		len = 8;
-	}
+	size_t bytes_left = count;
+	// 8 bytes are required for T54(4) and T53(4)
+	size_t blk_size = card->max_send_size - 8;
+	size_t to_send = 0;
+	size_t offset = (size_t) idx;
+	do {
+		len = 0;
+		to_send = bytes_left >= blk_size ? blk_size : bytes_left;
+		p = cmdbuff;
+		// ASN1 0x54 offset
+		*p++ = 0x54;
+		*p++ = 0x02;
+		*p++ = (offset >> 8) & 0xFF;
+		*p++ = offset & 0xFF;
+		// ASN1 0x53 to_send
+		*p++ = 0x53;
+		if (to_send < 128) {
+			*p++ = (u8)to_send;
+			len = 6;
+		} else if (to_send < 256) {
+			*p++ = 0x81;
+			*p++ = (u8)to_send;
+			len = 7;
+		} else {
+			*p++ = 0x82;
+			*p++ = (to_send >> 8) & 0xFF;
+			*p++ = to_send & 0xFF;
+			len = 8;
+		}
 
-	if (buf != NULL)
-		memcpy(p, buf, count);
-	len += count;
+		if (buf != NULL)
+			memcpy(p, buf+offset, to_send);
+		len += to_send;
 
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_3, 0xD7, fid >> 8, fid & 0xFF);
-	apdu.data = cmdbuff;
-	apdu.datalen = len;
-	apdu.lc = len;
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3, 0xD7, fid >> 8, fid & 0xFF);
+		apdu.data = cmdbuff;
+		apdu.datalen = len;
+		apdu.lc = len;
 
-	r = sc_transmit_apdu(card, &apdu);
+		r = sc_transmit_apdu(card, &apdu);
+		LOG_TEST_GOTO_ERR(ctx, r, "APDU transmit failed");
+		r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		LOG_TEST_GOTO_ERR(ctx, r, "Check SW error");
+
+		bytes_left -= to_send;
+		offset += to_send;
+	} while (0 < bytes_left);
+
+err:
 	free(cmdbuff);
-	LOG_TEST_RET(ctx, r, "APDU transmit failed");
-
-	r =  sc_check_sw(card, apdu.sw1, apdu.sw2);
-	LOG_TEST_RET(ctx, r, "Check SW error");
 
 	LOG_FUNC_RETURN(ctx, count);
 }
-
 
 
 static int sc_hsm_update_binary(sc_card_t *card,
@@ -1227,7 +1237,7 @@ static int sc_hsm_initialize(sc_card_t *card, sc_cardctl_sc_hsm_init_param_t *pa
 		return SC_ERROR_INVALID_ARGUMENTS;
 	}
 	*p++ = 0x81;	// User PIN
-	*p++ = (u8) params->user_pin_len;
+	*p++ = (u8)params->user_pin_len;
 	memcpy(p, params->user_pin, params->user_pin_len);
 	p += params->user_pin_len;
 
@@ -1400,12 +1410,11 @@ static int sc_hsm_unwrap_key(sc_card_t *card, sc_cardctl_sc_hsm_wrapped_key_t *p
 
 	LOG_FUNC_CALLED(card->ctx);
 
-	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_EXT, 0x74, params->key_id, 0x93);
-	apdu.cla = 0x80;
-	apdu.lc = params->wrapped_key_length;
-	apdu.data = params->wrapped_key;
-	apdu.datalen = params->wrapped_key_length;
+	r = sc_hsm_write_ef(card, 0x2F10, 0, params->wrapped_key, params->wrapped_key_length);
+	LOG_TEST_RET(card->ctx, r, "Create EF failed");
 
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_1, 0x74, params->key_id, 0x93);
+	apdu.cla = 0x80;
 	r = sc_transmit_apdu(card, &apdu);
 	LOG_TEST_RET(ctx, r, "APDU transmit failed");
 
@@ -1765,16 +1774,9 @@ static int sc_hsm_init(struct sc_card *card)
 	int flags,ext_flags;
 	sc_file_t *file = NULL;
 	sc_path_t path;
-	sc_hsm_private_data_t *priv = card->drv_data;
+	sc_hsm_private_data_t *priv = NULL;
 
 	LOG_FUNC_CALLED(card->ctx);
-
-	if (!priv) {
-		priv = calloc(1, sizeof(sc_hsm_private_data_t));
-		if (!priv)
-			LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
-		card->drv_data = priv;
-	}
 
 	flags = SC_ALGORITHM_RSA_RAW|SC_ALGORITHM_RSA_PAD_PSS|SC_ALGORITHM_ONBOARD_KEY_GEN;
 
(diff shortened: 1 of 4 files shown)


frankmorgner (approved):

frankmorgner (changes requested):
please check the compiler warnings from CI
```
card-sc-hsm.c:852:9: error: variable 'sent' set but not used [-Werror,-Wunused-but-set-variable]
```

istr:
I see. Will fix the build issues.

istr:
@frankmorgner note that the build fails using gcc 12. So I needed to configure using `--disable-strict` for the build to pass (probably this is why I did not see the problems you mentioned in my local build).
```
gcc --version
gcc (Debian 12.2.0-13) 12.2.0
```

```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../.. -I../../src -I../../src/include -I../../src -Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing=2 -g -O2 -c sm-eac.c  -fPIC -DPIC -o .libs/libsmeac_la-sm-eac.o
sm-eac.c: In function 'eac_sm_pre_transmit':
sm-eac.c:2228:76: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
 2228 |                                                         (unsigned char **) &eacsmctx->auxiliary_data->data);
      |                                                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Hold on, I may need a few hours to have the build fixed.

istr:
@frankmorgner the following change did the trick for me
```
diff --git a/configure.ac b/configure.ac
index d7d9cb6b..8e9ffd01 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1124,7 +1124,7 @@ if test "${enable_pedantic}" = "yes"; then
        CFLAGS="-pedantic ${CFLAGS}"
 fi
 if test "${enable_strict}" = "yes"; then
-       CFLAGS="-Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing=2 ${CFLAGS}"
+       CFLAGS="-Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing ${CFLAGS}"
 fi
 
 AC_CONFIG_FILES([
```

However, this change has nothing to do with the PR at hand. You should perhaps deal with it elsewhere or correct the type punning in `sm-eac.c` otherwise. For me this looks like a false positive. At least I have no idea how to handle the `char *` to `unsigned char*` "type conversion" differently in this situation (and with this direction - I am pretty sure that the other way around would work).

frankmorgner:
There is also a memory leak here https://github.com/OpenSC/OpenSC/blob/d21b6ab4be50e99b43211cef7bbf7d447e204d5b/src/libopensc/card-sc-hsm.c#L1728, see https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:90

istr:
> There is also a memory leak here
> 
> https://github.com/OpenSC/OpenSC/blob/d21b6ab4be50e99b43211cef7bbf7d447e204d5b/src/libopensc/card-sc-hsm.c#L1728
> 
> , see https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:90

Interesting that it occurs _here_, because the code in question (`sc_hsm_generate_keypair`) is not modified in this PR.

This one is harder to tackle, because the fuzzer seems to re-use previous state and I need to figure out how to run the equivalent of `.github/workflows/cifuzz.yml`. Using `https://github.com/nektos/act` does not work out-of-the-box because the images are not usable outside of github. `pull access denied for act-google-oss-fuzz-infra-cifuzz-actions-build-fuzzers-master-infra-cifuzz-actions-build-fuzzers-dockeraction, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

istr:
> There is also a memory leak here
> 
> https://github.com/OpenSC/OpenSC/blob/d21b6ab4be50e99b43211cef7bbf7d447e204d5b/src/libopensc/card-sc-hsm.c#L1728
> 
> , see https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:90

Sure that you pointed to the correct line in the code?

```
==30==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x531b4e in __interceptor_calloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:77:3
    #1 0x[90](https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:91)b092 in sc_hsm_init /src/opensc/src/libopensc/card-sc-hsm.c:1784:10
    #2 0x57ded1 in sc_connect_card /src/opensc/src/libopensc/card.c:358:8
    #3 0x56f06b in fuzz_connect_card /src/opensc/src/tests/fuzzing/fuzzer_reader.c:205:9
    #4 0x56cacd in LLVMFuzzerTestOneInput /src/opensc/src/tests/fuzzing/fuzz_pkcs15_encode.c:44:6
    #5 0x4407e3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #6 0x43ffca in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #7 0x441e34 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:826:7
    #8 0x442069 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:857:3
    #9 0x4316cf in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:[91](https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:92)2:6
    #10 0x45ad22 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #11 0x7fe9c1c7b082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 1878e6b475720c7c51[96](https://github.com/OpenSC/OpenSC/actions/runs/3990200153/jobs/6844166672#step:5:97)9e69ab2d276fae6d1dee)

DEDUP_TOKEN: __interceptor_calloc--sc_hsm_init--sc_connect_card
SUMMARY: AddressSanitizer: 56 byte(s) leaked in 1 allocation(s).
```

frankmorgner:
Indeed... this may fix the problem:
```diff
diff --git a/src/tests/fuzzing/fuzz_pkcs15_decode.c b/src/tests/fuzzing/fuzz_pkcs15_decode.c
index a83c719cb..e5758ba4d 100644
--- a/src/tests/fuzzing/fuzz_pkcs15_decode.c
+++ b/src/tests/fuzzing/fuzz_pkcs15_decode.c
@@ -108,9 +108,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 
        sc_pkcs15_parse_unusedspace(buf, buf_len, p15card);
 
-       sc_pkcs15_card_free(p15card);
-
 err:
+       sc_pkcs15_card_free(p15card);
        sc_disconnect_card(card);
        sc_release_context(ctx);
        return 0;
diff --git a/src/tests/fuzzing/fuzz_pkcs15_encode.c b/src/tests/fuzzing/fuzz_pkcs15_encode.c
index eb3436dae..a10ecf564 100644
--- a/src/tests/fuzzing/fuzz_pkcs15_encode.c
+++ b/src/tests/fuzzing/fuzz_pkcs15_encode.c
@@ -80,8 +80,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
        sc_pkcs15_encode_unusedspace(ctx, p15card, &unused_space, &unused_space_len);
        free(unused_space);
 
-       sc_pkcs15_card_free(p15card);
 err:
+       sc_pkcs15_card_free(p15card);
        sc_disconnect_card(card);
        sc_release_context(ctx);
```

istr:
@frankmorgner rebased, squashed. Hopefully it is now ready for review. Let's see what the CI checks tell us.

istr:
appveyor job needs a restart
```
Build started
git clone -q https://github.com/OpenSC/OpenSC.git C:\projects\opensc
      1 [main] git-remote-https 47[2](https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/46039192/job/m2yxotck7dk340vh#L2) child_info_fork::abort: \??\C:\cygwin\bin\cygcrypto-1.1.dll: Loaded to different address: parent(0x1020000) != child(0xD60000)
error: cannot fork() for fetch-pack: Resource temporarily unavailable
Command exited with code 128
```

istr:
@frankmorgner  Hm... Probably this one hits:
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card.c#L325-L335

It seems that the initialization fails straight away while fuzzing:
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card.c#L358

The driver allocates some private data
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card-sc-hsm.c#L1773

which is never `free`d due to the direct call to `sc_card_free` instead of `sc_disconnect_card` here:
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card.c#L398-L402

because the internal data is only `free`d in the operation `finish` for `card-sc-hsm.c`.
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card-sc-hsm.c#L1880-L1896

That is to say, probably `card.c` might be better off calling `sc_disconnect_card` (which, in turn calls the `finish` operation) instead of `sc_card_free` in 
https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card.c#L402

But I cannot tell if this change would have significant negative impact on the other driver implementations.

istr:
I simply gave it a try with d868efb6e5f7b42bbfaaa2ee8fc3db6efbbd5089

frankmorgner:
> I simply gave it a try with [d868efb](https://github.com/OpenSC/OpenSC/commit/d868efb6e5f7b42bbfaaa2ee8fc3db6efbbd5089)

In principle, this change is a good idea. However, sc_disconnect_card then requires some more caution. For example, `card->reader->ops->disconnect`, will always be called regardless of whether the pointers are NULL or not. Also, sc_connect_card can then be simplified to not check whether a reader is `connected` or not since this is disconnected in sc_disconnect_card anyway.

istr:
> However, sc_disconnect_card then requires some more caution.

Indeed. Alas, this would also mean refactoring some of the other drivers'  `finish` code, because a call is unsafe in this situation as well. For example

https://github.com/OpenSC/OpenSC/blob/f2497d97f406a2810de9cbeaea5117cbc0d7eed8/src/libopensc/card-iasecc.c#L1520-L1521
(see last fuzzer error)

To lock and limit the scope of this PR I will try to find a solution within `card-sc-hsm.c` only and then open up another issue (and maybe PR) for the proposed refactoring.

frankmorgner (approved):

frankmorgner:
> To lock and limit the scope of this PR I will try to find a solution within `card-sc-hsm.c` only and then open up another issue (and maybe PR) for the proposed refactoring.

agreed, thank you

istr:
@CardContact Please review this PR.
@frankmorgner Thanks a lot for your intensive support to moving this forward.

src/libopensc/reader-pcsc.c
```diff
@@ -1314,8 +1335,8 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 }
 
 int pcsc_add_reader(sc_context_t *ctx,
-	   	char *reader_name, size_t reader_name_len,
-		sc_reader_t **out_reader)
+	char *reader_name, size_t reader_name_len,
+	sc_reader_t **out_reader)
```
Jakuje, line 1339: the continuation should be two tabs. See #2017 for more information.

src/libopensc/card-sc-hsm.c
```diff
@@ -1883,14 +1910,18 @@ static int sc_hsm_finish(sc_card_t * card)
...
+	if (priv) {
+		if (priv->serialno) {
+			free(priv->serialno);
+		}
+		if (priv->dffcp) {
+			sc_file_free(priv->dffcp);
+		}
+		if (priv->EF_C_DevAut) {
+			free(priv->EF_C_DevAut);
+		}
```
Jakuje, line 1922: the null check like this is useless as `free` handles NULL arguments without any issues and when the member already contains some junk data, it will not save you.

Jakuje (commented):
mostly nits. Please, squash the fixup and revert commits to have clean commit series before merging.

istr:
> mostly nits. Please, squash the fixup and revert commits to have clean commit series before merging.

Ok. Probably we should add a pre-commit linter that enforces the expected code format. Usually, I go for https://pre-commit.com/. Maybe this in combination with https://github.com/pocc/pre-commit-hooks is an option here.

src/libopensc/card-sc-hsm.c
istr: Yes. This was already in place before this PR.

There are lots of other occurrences in the code, esp. when calling `sc_file_free`. I will remove these spurious checks as well. However, the added `if (priv) {` check prevents to dereference a `NULL` value of `priv` in `free(priv->serialno);`.

istr:
@Jakuje squashed / force pushed again.

src/libopensc/reader-pcsc.c
istr: Code formatting in reader-pcsc.c is already rather inconsistent before this PR. For example `if(` vs `if (` feels like 50:50. :smiley:

Jakuje:
> Ok. Probably we should add a pre-commit linter that enforces the expected code format. Usually, I go for https://pre-commit.com/. Maybe this in combination with https://github.com/pocc/pre-commit-hooks is an option here.

It is a problem that we are not on the same page what would be the "correct" format and configuration to achieve it, see #2017. I would be all for something like this as soon as we will have the final .clang-format to use.

Jakuje (approved):

istr:
Rebased to 20899a14a38a8be036e284573a4b4c3908793b1c, force-pushed.

Jakuje:
Thanks!

PR #2687, opened 2023-01-26
Add support for basic attribute filtering to list objects

### LINKED ISSUE #2627: pkcs11-tool: list_objects not passing in object class template?
### Problem Description

Is there a reason pkcs11-tool list_objects doesn't pass in a template with the object class directly rather than getting all objects and then checking if the class matches? Seems like this is what the API intended all along.

https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L4613

Related to this, it seems impossible to list a reduced item set matching other criteria. Would it be desirable to take input parameters such as label, id, type, etc. and assemble a proper template to pass for the search and allow listing objects matching the sub-criteria?

dlegaultbbry:
NOTE: Not sure if anything needs to be mentioned in the docs about this new capability.

fixes #2627

```
Generate some objects

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --write-object /data/keystore/ec_privkey.pem --type privkey --id 2222 --label eckey
Created private key:
Private Key Object; EC
  label:      eckey
  ID:         2222
  Usage:      none
  Access:     sensitive, never extractable

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --keygen --key-type AES:32 --label aeskey --id 1111
Key generated:
Secret Key Object; AES length 32
  VALUE:      e42d2476fe21337ce66d4f9c073770fbba8a70b171f41dbd5cf6d2cdadbf6cbb
  label:      aeskey
  ID:         1111
  Usage:      encrypt, decrypt
  Access:     never extractable, local

List all objects

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --list-objects
Secret Key Object; AES length 32
  VALUE:      e42d2476fe21337ce66d4f9c073770fbba8a70b171f41dbd5cf6d2cdadbf6cbb
  label:      aeskey
  ID:         1111
  Usage:      encrypt, decrypt
  Access:     never extractable, local
Private Key Object; EC
  label:      eckey
  ID:         2222
  Usage:      none
  Access:     sensitive, never extractable

Filter by key type

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --list-objects --type privkey
Private Key Object; EC
  label:      eckey
  ID:         2222
  Usage:      none
  Access:     sensitive, never extractable

Filter by id

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --list-objects --id 1111
Secret Key Object; AES length 32
  VALUE:      e42d2476fe21337ce66d4f9c073770fbba8a70b171f41dbd5cf6d2cdadbf6cbb
  label:      aeskey
  ID:         1111
  Usage:      encrypt, decrypt
  Access:     never extractable, local

```

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -208,7 +208,12 @@
 						<option>--list-objects</option>,
 						<option>-O</option>
 					</term>
-					<listitem><para>Display a list of objects.</para></listitem>
+					<listitem>
+						<para>Display a list of objects.</para>
+						<para>The options <option>--keytype</option>, <option>--label</option>
+						, <option>--id</option> or <option>--application-id</option> can be
+						used to filter the listed objects.</para>
+					</listitem>
 				</varlistentry>
 
 				<varlistentry>
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -526,7 +526,7 @@ static void		show_cryptoki_info(void);
 static void		list_slots(int, int, int);
 static void		show_token(CK_SLOT_ID);
 static void		list_mechs(CK_SLOT_ID);
-static void		list_objects(CK_SESSION_HANDLE, CK_OBJECT_CLASS);
+static void		list_objects(CK_SESSION_HANDLE);
 static void		list_interfaces(void);
 static int		login(CK_SESSION_HANDLE, int);
 static void		init_token(CK_SLOT_ID);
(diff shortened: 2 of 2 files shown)


frankmorgner (changes requested):
This change looks good, but please add the possibility for combination of the arguments to `doc/tools/pkcs11-tool.1.xml`

Jakuje (approved):

frankmorgner (approved):

PR #2689, opened 2023-01-27
Revert "Remove redundant -lglib-2.0 library"

saper:
Fixes failure to link libopensc on FreeBSD

````
ld: error: unable to find library -lgio-2.0
ld: error: unable to find library -lgobject-2.0
````

This reverts commit f43169a8d7dc9dc9df99c15bcec0c804b8efd856.

- [ ] macOS tokend is tested

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -554,10 +554,6 @@ case "${host}" in
 							   have_notify="no"
 							   have_gio2="no"  ])
 			LIBS="$saved_LIBS"
-			if test "${have_gio2}" = "yes"; then
-				# we do not need lglib-2.0
-				GIO2_LIBS="-lgio-2.0 -lgobject-2.0"
-			fi
 		;;
 esac
 


Jakuje:
is there some way to intergrate or emulate the freebsd builds we we can avoid such breakages in the future?

Jakuje (approved):

Jakuje:
One more case, it would be great to update the workflow files to trigger also only on the `configure.ac` change. But it can land later/separately.

saper:
> workflow files

which are the "workflow files"?

dengert:
The workflow files are in the repository under .github  thus  you can test and submit PRs for them.

Jakuje:
Basically https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/linux.yml#L5 macos.yml and probably also packit.yml should include this path

frankmorgner (approved):

Jakuje:
Merged #2698 instead. Thank you for bringing the issue up and helping to solve this!

Issue #2125, opened 2020-10-02
Handle certificate decompression in pkcs15 layer, rather than in drivers


Jakuje:
### Problem Description

The certificate decompression is no handled inside of read_binary() functions of particular drivers. It is error-prone, hard to get right and unnecessary code duplication that needs to handle corner many cases.

### Proposed Resolution

In #2116 it was suggested we can move the decompression to pkcs15-layer (even though there is no standard for using compressed certificates there) and decompress "on the fly" if the driver will detect the certificate is compressed and sets some flag in `sc_pkcs15_object_t`.

### Steps to reproduce

grep for `sc_decompress` in drivers. At this time, the drivers `cac`, `cac1`, `coolkey`, `dnie`, `gids`, `idprime` and `piv` support compressed certificates and each of them handles them a bit differently.

### Logs

--

frankmorgner:
Before leaving `sc_pkcs15_read_file()` would be a good place to handle decompression with an `sc_pkcs15_object_t` flagged as compressed. This then should cover certificates and other objects (files) exposed via PKCS#15.

dengert:
All or most of  the driver listed above are emulating pkcs15 They already handle decompression.  As I have said before PIV driver need to decompress certificates (which are contained in certificate objects that include other data)  to emulate public key and private keys at card and and pkcs15 levels. 

There maybe pkcs15 cards with compressed objects. As far as we know, pkcs15 does not say anything about this. Thus any application that knows objects have data that must be decompressed, already knows how to handle it. 

I would be very reluctant to try and out guess the cards/applications handling of  compressed data.  

I do not have a copy of iso 7816-15 to see if it says anything about compression. If it does then we should implement what it supports. If not we should propose how it could be done in pkcs15.

PR #2690, opened 2023-01-30
Handle certificate decompression in pkcs15 layer

xhanulik:
This PR moves decompression from drivers to the pkcs15 layer (`sc_pkcs15_read_file()`), as discussed in #2125.

Decompression is removed from `*read_binary()` functions from drivers and moved after calling `sc_read_binary()` in `sc_pkcs15_read_file()`. Whether the data read by `sc_read_binary()` are compressed is denoted by `unsigned long *flags` - the driver sets the corresponding mask `SC_FILE_COMPRESSED`. The former `unsigned long flags` parameter is changed to a pointer to propagate the information about compression out of `sc_read_binary()`.

Changes tested with cards
- Gemalto IDPrime 3810 (OSv1)
- PIV Card

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -1078,7 +1078,7 @@ static int dnie_fill_cache(sc_card_t * card,unsigned long *flags)
  */
 static int dnie_read_binary(struct sc_card *card,
 			    unsigned int idx,
-			    u8 * buf, size_t count, unsigned long flags)
+			    u8 * buf, size_t count, unsigned long *flags)
 {
 	int res = 0;
 	sc_context_t *ctx = NULL;
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -38,7 +38,6 @@
 #include "opensc.h"
 #include "cardctl.h"
 #include "internal.h"
-#include "compression.h"
 #include "cwa14890.h"
 #include "cwa-dnie.h"
 
@@ -904,34 +903,32 @@ static int dnie_finish(struct sc_card *card)
 /* ISO 7816-4 functions */
 
 /**
- * Uncompress data if in compressed format.
+ * Check whether data are compressed.
  *
  * @param card pointer to sc_card_t structure
  * @param from buffer to get data from
- * @param len pointer to buffer length
- * @return uncompressed or original buffer; len points to new buffer length
- *        on error return null
+ * @param len buffer length
+ * @return 1 if data are compressed, 0 otherwise; len points to expected length of decompressed data
  */
-static u8 *dnie_uncompress(sc_card_t * card, u8 * from, size_t *len)
+
+static int dnie_is_compressed(sc_card_t * card, u8 * from, size_t len)
 {
-	u8 *upt = from;
 #ifdef ENABLE_ZLIB
-	int res = SC_SUCCESS;
 	size_t uncompressed = 0L;
 	size_t compressed = 0L;
 
 	if (!card || !card->ctx || !from || !len)
-		return NULL;
+		return 0;
 	LOG_FUNC_CALLED(card->ctx);
 
 	/* if data size not enough for compression header assume uncompressed */
-	if (*len < 8)
+	if (len < 8)
 		goto compress_exit;
 	/* evaluate compressed an uncompressed sizes (little endian format) */
 	uncompressed = lebytes2ulong(from);
 	compressed = lebytes2ulong(from + 4);
 	/* if compressed size doesn't match data length assume not compressed */
-	if (compressed != (*len) - 8)
+	if (compressed != len - 8)
 		goto compress_exit;
 	/* if compressed size greater than uncompressed, assume uncompressed data */
 	if (uncompressed < compressed)
@@ -940,32 +937,14 @@ static u8 *dnie_uncompress(sc_card_t * card, u8 * from, size_t *len)
 	if (uncompressed > MAX_FILE_SIZE)
 		goto compress_exit;
 
-	sc_log(card->ctx, "Data seems to be compressed. calling uncompress");
-	/* ok: data seems to be compressed */
-	upt = calloc(uncompressed, sizeof(u8));
-	if (!upt) {
-		sc_log(card->ctx, "alloc() for uncompressed buffer failed");
-		return NULL;
-	}
-	*len = uncompressed;
-	res = sc_decompress(upt,	/* try to uncompress by calling sc_xx routine */
-			    len,
-			    from + 8, (size_t) compressed, COMPRESSION_ZLIB);
-	if (res != SC_SUCCESS) {
-		sc_log(card->ctx, "Uncompress() failed or data not compressed");
-		goto compress_exit;	/* assume not need uncompression */
-	}
-	/* Done; update buffer len and return pt to uncompressed data */
-	sc_log_hex(card->ctx, "Compressed data", from + 8, compressed);
-	sc_log_hex(card->ctx, "Uncompressed data", upt, uncompressed);
- compress_exit:
-
+	sc_log(card->ctx, "Data seems to be compressed.");
+	return 1;
+compress_exit:
 #endif
 
-	sc_log(card->ctx, "uncompress: returning with%s de-compression ",
-	       (upt == from) ? "out" : "");
-	return upt;
-}
+	sc_log(card->ctx, "Data not compressed.");
+	return 0;
+}	
 
 /**
  * Fill file cache for read_binary() operation.
@@ -985,14 +964,14 @@ static u8 *dnie_uncompress(sc_card_t * card, u8 * from, size_t *len)
  * @param card Pointer to card structure
  * @return SC_SUCCESS if OK; else error code
  */
-static int dnie_fill_cache(sc_card_t * card)
+static int dnie_fill_cache(sc_card_t * card,unsigned long *flags)
 {
 	u8 tmp[MAX_RESP_BUFFER_SIZE];
 	sc_apdu_t apdu;
 	size_t count = 0;
 	size_t len = 0;
 	u8 *buffer = NULL;
-	u8 *pt = NULL, *p;
+	u8 *p;
 	sc_context_t *ctx = NULL;
 
 	if (!card || !card->ctx)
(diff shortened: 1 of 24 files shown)


src/libopensc/pkcs15-gemsafeV1.c
```diff
@@ -236,8 +237,9 @@ static int gemsafe_get_cert_len(sc_card_t *card)
 	 */
 	iptr = ibuf + GEMSAFE_READ_QUANTUM;
 	while ((size_t)(iptr - ibuf) < objlen) {
+		unsigned long flags = 0;
```
Jakuje, line 240: this variable `flags` shadows the previous declaration of `flags`. I think just rest it to 0 here should be enough.

Jakuje:
Otherwise looks good. Turned out much nicer than I was expecting when I was initially filling the issue.

src/libopensc/pkcs15-gemsafeV1.c
xhanulik: Thanks for review, fixed.

Jakuje (approved):

Jakuje:
@frankmorgner @dengert what do you think? We don't have all the cards to test this change.

xhanulik:
With CAC cards, the changes should work properly as well.

dengert:
It looked OK. I assume the PIV card used in the test has a compressed cert.

Jakuje:
> It looked OK. I assume the PIV card used in the test has a compressed cert.

yes, we tested with one of the NIST test cards with compressed certificates. Veronika can check which ones if needed.

src/libopensc/card-dnie.c
```diff
@@ -940,32 +937,14 @@ static u8 *dnie_uncompress(sc_card_t * card, u8 * from, size_t *len)
...
-	/* ok: data seems to be compressed */
-	upt = calloc(uncompressed, sizeof(u8));
-	if (!upt) {
-		sc_log(card->ctx, "alloc() for uncompressed buffer failed");
-		return NULL;
-	}
-	*len = uncompressed;
-	res = sc_decompress(upt,	/* try to uncompress by calling sc_xx routine */
-			    len,
-			    from + 8, (size_t) compressed, COMPRESSION_ZLIB);
```
frankmorgner, line 953: The buffer to decompress will start at offset 8. The first 8 bytes, however, will still be returned with dnie_read_binary. As consequence, this should give a decompression error.

If the file is decompressed, I think, the first 8 bytes need to be removed from the output buffer.

src/libopensc/opensc.h
```diff
@@ -658,7 +658,7 @@ struct sc_card_operations {
 	 * @see sc_read_binary()
 	 */
 	int (*read_binary)(struct sc_card *card, unsigned int idx,
-			u8 * buf, size_t count, unsigned long flags);
+			u8 * buf, size_t count, unsigned long *flags);
```
frankmorgner, line 661: Suggestion: If you would document `flags` as optional argument, then the card driver should check if flags is not NULL before setting it to `SC_FILE_COMPRESSED`. If used this way, all card drivers that currently ignore `flags` don't need to be touched with this PR since they can continue to pass `0` as `flags` to `sc_read_binary` (where `0` is what's the definition of `NULL`).

xhanulik:
@frankmorgner  I agree that this change only moves the decompression from the drivers to the upper layer. However, I'm not sure if I understand your idea correctly.

For distinguishing whether the certificate is compressed or not, the initialization in PKCS#15 emulators would require reading at least some first bytes, and then the check would differ per card as it now does in card drivers. This information should be stored somewhere along with the offset number where compressed certificate data starts - this could probably be stored in `sc_pkcs15_cert_info`, which is then around when `sc_pkcs15_read_file()` is called. Then `sc_pkcs15_read_file()` would need some flag, whether the data are compressed, and the offset length and decompression itself would move out of the `read_binary` functions into `sc_pkcs15_read_file()`.

Is that something like you thought of?

frankmorgner:
Yes, exactly. I still look at sc_read_binary as being related to card command reading the data. `sc_pkcs15_read_file()` is an abstraction for the file type (cached, elementary, record based).

It is true, that some code reads the data and checks the first bytes to check whether it is compressed or not. However, I assume this was most likely implemented because it is the easiest thing to do at the card driver level without much knowledge of the context. For PIV, for example, [you should look at the CertInfo byte (tag `0x71`)](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf), which is different from the actual certificate (tag `0x70`). I would expect to have this kind of meta information in the PKCS#15 emulation layer rather than at the card driver level.

dengert:
In regard to "you should look at the CertInfo byte (tag 0x71)", card-piv.c takes the PIV certificate object, and extracts the certificate. and presents to the pkcs15 layer two files: the certificate object, as path xxxx and the extracted certificate (formally uncompressed but with this PR compressed)  as  path xxxxCECE.

card-piv.c also looks at the certificate object's tag 0x71 and sets "compressed" 
 https://github.com/OpenSC/OpenSC/pull/2690/files#diff-14d802e467876e00ece49c8d6621ec74eaf7d02384642ac0694de83397d9bbfcL1091-R1090

which then sets PIV_OBJ_CACHE_COMPRESSED, which is then passed in https://github.com/OpenSC/OpenSC/pull/2690/files#diff-14d802e467876e00ece49c8d6621ec74eaf7d02384642ac0694de83397d9bbfcL1150-R1135

So a caller of

src/libopensc/pkcs15.c
```diff
@@ -2391,6 +2395,23 @@ sc_pkcs15_parse_unusedspace(const unsigned char *buf, size_t buflen, struct sc_p
...
+
+	rv = sc_decompress_alloc(out, outlen, buf, buflen, COMPRESSION_AUTO);
+	if (rv != SC_SUCCESS) {
+		sc_log(card->ctx,  "Decompression failed: %d", rv);
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_DATA);
+	}
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+#endif
+	sc_log(card->ctx, "Compression not supported, no zlib");
+	LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
```
frankmorgner, line 2413: please enclose the return of `SC_ERROR_NOT_SUPPORTED` in `#else`/`#endif`, otherwise static code analysis will complain that the last part is not reachable

xhanulik:
For now, it is rebased onto current master and I removed all the unnecessary NULL setting. I will look why the CIFuzz is failing and also try to look into the drivers and how they handle the decompression to move it into the emulators.

frankmorgner (changes requested):
Currently, it seems to universally work if decompress_file just uses `COMPRESSION_AUTO`. However, maybe it would be usefull to at least add a comment to `SC_FILE_COMPRESSED` that we may want to pass the actual compression algorithm via the flag as well.

xhanulik:
Thanks to @dengert for clarification about the PIV driver.

I think by PIV, the PKCS15 emu can use the certificate object path (without `cece`), read it and parse the certinfo byte to get information about whether the (extracted) certificate (read from the path with `cece`) is compressed or not.

I do not see such _a simple_ solution by some other drivers, for example, the CAC driver. It also checks the certinfo tag when [reading binary](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-cac.c#L569-L579). However, the output is only the extracted certificate, so I'm not sure how to deal with the tag on the PKCS15 level without much intervention into the driver.

xhanulik:
I resolved all of the review comments and problems found by fuzzing. The changes are only moving decompression out of drivers into `sc_pkcs15_read_file()` function.

I also tried to reimplement (not in this PR) the decompression handling as described in https://github.com/OpenSC/OpenSC/pull/2690#issuecomment-1439789855 and https://github.com/OpenSC/OpenSC/pull/2690#issuecomment-1497407611 - the decompression would be done in `sc_pkcs15_read_file()`, which would take a flag, whether the read binary should be decompressed or not. Checking whether the binary is compressed would take place before calling `sc_pkcs15_read_file()`.

I successfully implement this for [IDPrime](https://github.com/xhanulik/OpenSC/commit/ea53ec90f2821873a39b374ccd2f850d23e091dd), as reading some first bytes of the binary suffices to check the compression. For [PIV](https://github.com/xhanulik/OpenSC/commit/0f424b1e96afee2a49cc37d9104565de3ad0361b), we can distinguish between getting a certificate object (with certinfo byte) and only extracted certificate. Reading binary for CAC is done similarly as by PIV - `read_binary()` returns only extracted certificate, so this could be handled in the same manner as for PIV by [adding some _bogus_ path](https://github.com/xhanulik/OpenSC/commit/84adfeaa1e7d484d59bac69442064e6faf26d664) which would serve for also getting the certinfo byte.

However, by DNIE, I encountered a problem with such approach. In `sc_pkcs15emu_dnie_init()`, there is no direct call to `sc_pkcs15_read_file()`, but files are handled via `sc_pkcs15_parse_df()` where the file reading takes place. I'm not sure how to do the checking for decompressed objects in such general function without knowing which objects are being read.

src/libopensc/card-dnie.c
Jakuje: I think this is now handled in `dnie_fill_cache()`.

src/libopensc/opensc.h
Jakuje: We should update the doc text above to reflect this change. Now it says `currently not used`. So replacing with `optional` might be enough.

src/libopensc/types.h
```diff
@@ -225,6 +225,11 @@ typedef struct sc_acl_entry {
 #define SC_FILE_EF_CYCLIC		0x06
 #define SC_FILE_EF_CYCLIC_TLV		0x07
 
+/* Data compression */
+#define SC_FILE_COMPRESSED_AUTO		0x08
+#define SC_FILE_COMPRESSED_ZLIB		0x09
+#define SC_FILE_COMPRESSED_GZIP		0x0a
```
Jakuje, line 231: I would  probably prefer to have them in separate "namespace" like `SC_FILE_FLAG_*`. Given that you set them as a bitfiled, they should not be overlapping values.

src/libopensc/types.h
```diff
@@ -225,10 +225,10 @@ typedef struct sc_acl_entry {
...
 #define SC_FILE_EF_CYCLIC_TLV		0x07
 
-/* Data compression */
-#define SC_FILE_COMPRESSED_AUTO		0x08
-#define SC_FILE_COMPRESSED_ZLIB		0x09
-#define SC_FILE_COMPRESSED_GZIP		0x0a
+/* File flags */
+#define SC_FILE_FLAG_COMPRESSED_AUTO		0x08
+#define SC_FILE_FLAG_COMPRESSED_ZLIB		0x09
+#define SC_FILE_FLAG_COMPRESSED_GZIP		0x0a
```
Jakuje, line 231:
```suggestion
#define SC_FILE_FLAG_COMPRESSED_AUTO		0x01
#define SC_FILE_FLAG_COMPRESSED_ZLIB		0x02
#define SC_FILE_FLAG_COMPRESSED_GZIP		0x04
```

xhanulik: Thanks for review, fixed & rebased onto current master.

frankmorgner (approved):

PR #2696, opened 2023-02-08
westcos: Generate 2k RSA keys by default

Jakuje:
Bump the default RSA key size in the westcos tool. The 2k is supported by the driver and using smaller values is no longer recommended and flagged by the github scanning tool as a potential security issue:

https://github.com/OpenSC/OpenSC/security/code-scanning/5


Jakuje:
> please also update the tool's man page

thanks for a reminder. Updated now.

frankmorgner (approved):

PR #2615, opened 2022-10-10
Remove unneeded linking and flags from compilation

xhanulik:
This PR removes uneeded libraries and flags from linking of tools and opensc libraries. The `-ldl` is moved from `LIBS` to `LDL_LIBS` and is linked only when needed. Also `GIO_LIBS` contains only `-lgio-2.0 -lgobject-2.0`, as `lglib-2.0` is not needed.

Fixes  #2566, but there still may be redundant (transitive) dependencies as described in https://github.com/OpenSC/OpenSC/issues/2566#issuecomment-1206535919.

##### Checklist
- [ ] PKCS#11 module is tested
- [ ] Windows minidriver is tested
- [ ] macOS tokend is tested

Issue #2566, opened 2022-06-07
overlinked binaries


mikhailnov:
[spec-helper](http://abf.io/import/spec-helper) found that binaries are overlinked (linked with not used libraries):

```
+ /usr/share/spec-helper/check_elf_files
Warning: unused libraries in /usr/bin/npa-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/usr/lib64/libcrypto.so.1.1
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/gids-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/dnie-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/usr/lib64/libcrypto.so.1.1
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/sc-hsm-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/westcos-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/piv-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
	/lib64/libdl.so.2
Warning: unused libraries in /usr/bin/netkey-tool: 
	/usr/lib64/libz.so.1
	/usr/lib64/libgio-2.0.so.0
	/usr/lib64/libgobject-2.0.so.0
	/usr/lib64/libglib-2.0.so.0
[...]

PR #2698, opened 2023-02-09
Remove only -lglib-2.0 from GIO_LIBS2

xhanulik:
This is the alternative PR to #2689, fixing the problems described in #2615. The `pkg-config` sets `GIO2_LIBS` variable, and then only the unneeded `-lglib-2.0` is removed.

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -11,6 +11,7 @@ on:
       - .github/push-artifacts.sh
       - '**.am'
       - doc/**
+      - configure.ac
   push:
 
 permissions:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,6 +12,7 @@ on:
       - .github/push-artifacts.sh
       - '**.am'
       - doc/**
+      - configure.ac
   push:
 
 permissions:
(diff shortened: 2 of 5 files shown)


xhanulik:
@saper could you try whether this would work also for FreeBSD?

Jakuje:
Can you update the workflow files as mentioned in the following comment so we can receive a CI results on configure.ac changes? https://github.com/OpenSC/OpenSC/pull/2689#issuecomment-1413789336

Jakuje (approved):

Jakuje:
Yes, this came from @mikhailnov in #2566 without much comments. We came to conclusion that some of them are real issues, some of them kind of expected, but wanted to clean up a bit. Thinking about it a bit more I have no strict preference between this and #2689 (while the #2689 being simpler and less code might be a plus though). Any other opinions?

For the particular Glib linking issue, I assume that this is just we not using all the stuff they would like us to use and therefore having unused libraries. I do not think this can be simply addressed in Glib upstream.

Jakuje:
@frankmorgner any opinions or preferences for this and #2689 so we can merge one and move on? I would prefer this solution as it looks like working in all cases we know so far. But would like to hear opinions from others too as this is stalled for some time.

frankmorgner:
OK, fine for me

frankmorgner (approved):

PR #2712, opened 2023-02-15
Introduce use of custom ossl libctx with OpenSSL >= 3.0

### LINKED ISSUE #2711: OPenSc should use a custom libctx to avoid loops
### Problem Description

OpenSC is used indirectly by many applications as it provides a pkcs11 interface.
This interface may be used directly by an application or indirectly through some other library.

For example, with the introduction of providers in OpenSSL 3.0, there is now a path where opensc-pkcs11 is used by an openssl provider.

Because opensc also can use openssl itself it is necessary for opensc to start creating its own openssl libctx to avoid several surprises.

Using the "Default" OpenSSL context poses several risk of opensc inheriting configurations changes made by the application that are incompatible with the goals of opensc, additional, in the case of providers it can lead to loops.

### Proposed Resolution

Change opensc code to add a custom libctx configured to use just the default provider, and nay other custom configuration change that may be necessary.

### Example of issues

As an example this issue arises when using pkcs11-provider:
https://github.com/latchset/pkcs11-provider/issues/187

simo5:
This is a first shot at adding a custom (private) OpenSSL libctx in libopensc.

Fixes #2711

##### Checklist

- [ ] macOS tokend is tested


simo5:
I marked this as draft because it will probably require some discussion to figure out if the approach is correct.

I ran make check locally and it did not regress the number of failures (my local make check is not clean with the master tree).

@dengert PTAL.

dengert:
OpenSC can be built using OpenSSL 1.1x and  OpenSSL-3.0.x  LiberSSL so differet ifdefs are needed. 

I have identified in OpenSC, these OpenSSL routines (or the _ex version) can use a libctx:
BIO_new
BN_CTX_new
EC_GROUP_new_by_curve_name
EC_KEY_new_by_curve_name
EC_KEY_new
EVP_default_properties_is_fips_enabled
EVP_MAC_fetch
EVP_PKEY_CTX_new_from_name
EVP_PKEY_new_raw_public_key
OSSL_PROVIDER_try_load
RAND_bytes

Some of these are used in many places. We may niot need to worry about BIO_new or BN_CTX_new. 

What we have done in the past is to add defines or macros in src/libopensc/sc-ossl-compat.h so if OpenSSL 3 is being used the define or macro would then add point to a libctx and call the appropiate OpenSSL 3 routine.

simo5:
@dengert please check the commits
I added a separate ossl.c for the wrappers, but the function definitions are in opensc.h
Can easily move them to sc-ossl-compat.h if you think that's a better place, just let me know.

simo5:
Btw you mised the two most important changes needed
EVP_MD_fetch() and EVP_CIPHER_fetch(0 which are the things that actually allow you to select which provider to use (based on libctx) for digest and encryption operations.
See the patches I already committed, they handle those cases.

simo5:
EVP_default_properties_is_fips_enabled) is used directly at a high level in the tools, so it probably does not need its own context, applications can be left unchanged.
The libraries/modules are the problem.

dengert:
What I did was for `grep "(OSSL_LIB_CTX " include/openssl/*` There were 79 functions  [osslinc-2.txt](https://github.com/OpenSC/OpenSC/files/10747305/osslinc-2.txt) found

Then greped for these 79 functions  and 24 functions without "_ex"  to see what is in  OpenSC source. 

So there might be others that should be included.

dengert:
> EVP_default_properties_is_fips_enabled) is used directly at a high level in the tools, so it probably does not need its own context, applications can be left unchanged.

`EVP_default_properties_is_fips_enabled`  was the command that caused OSSL_CRYPTO_init to be called. And it looks out of place.  ./tools/pkcs11-tool.c line 5079 as `FIPS_mode` is define as `EVP_default_properties_is_fips_enabled(NULL)`

Could `OSSL_LIB_CTX_get0_global_default()` be used inplace of a libctx? What exactly the `global_default`?

simo5:
It cause the issue because:
pkcs11-tool -> openssl -> pkcs11-provider -> opensc-pkcs11 -> openssl ->pkcs11-provider -> ...

But once we cut the loop in opensc-pkcs11 we will be fine, there is really no need to modify pkcs11-tool itself, only libraries/modules.

simo5:
The call ```OSSL_LIB_CTX_get0_global_default()``` basically returns the default context, it is the call that tells you what context you get when you pass NULL in functions that can specify a libctx.

From OpenSSL code:
```
OSSL_LIB_CTX *OSSL_LIB_CTX_get0_global_default(void)
{
    if (!RUN_ONCE(&default_context_init, default_context_do_init))
        return NULL;

    return &default_context_int;
}
```

Passing the context returned by that call is equal to passing NULL.

simo5:
> So there might be others that should be included.

Be nice if you could take a quick look at the patches I am piling on, to see if I am  going in the right direction ... there should be enough meat now to get a sense of what this change entails.

simo5:
@dengert how stable need libsm interface need to be?

This library has two problems.

1. it is loading the legacy provider arbitrarily in the default context, this is a big no-no no matter what else
2. it expose public intefaces that do not take a sc_context_t pointer so I can't get to the custom openssl libctx unless I change the functions signatures.

simo5:
Looks to me that libsm is a static library while libssm is a public one, so I will liberally change also libsm functions ...

simo5:
Unfortunately the same problem is present in libssm ... I will break it and rthen we can figure out how to pick up the pieces

dengert:
Checked out #2712 bd1fc42a7d7e9d643f1806c28e57b478530930ca
and applied to OpenSC master

```
/opt/ossl-3.0.8/bin/pkcs11-tool --module /opt/ossl-3.0.8/lib/opensc-pkcs11.so --verify -m SHA384-RSA-PKCS-PSS -y cert -d 02 --signature-file /tmp/derive.132501.text.sig -i /tmp/derive.132501.text
```

```
0  futex_wait (private=0, expected=2, futex_word=0x7ffff7f8b420 <c_initialize_m>)
    at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x7ffff7f8b420 <c_initialize_m>, private=0)
    at ./nptl/lowlevellock.c:49
#2  0x00007ffff7298082 in lll_mutex_lock_optimized (mutex=0x7ffff7f8b420 <c_initialize_m>)
    at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x7ffff7f8b420 <c_initialize_m>)
    at ./nptl/pthread_mutex_lock.c:93
#4  0x00007ffff7f49c36 in C_Initialize (pInitArgs=0x7fffffffd0b0)
    at ../../../src/src/pkcs11/pkcs11-global.c:310
#5  0x00007ffff7ee82d4 in p11prov_Initialize (ctx=0x5555555db8d0, pInitArgs=0x7fffffffd0b0)
    at ../../src/src/interface.gen.c:10
#6  0x00007ffff7ef62eb in p11prov_module_init (ctx=0x5555555db8d0)
    at ../../src/src/provider.c:1011
#7  0x00007ffff7ef6973 in OSSL_provider_init (handle=0x5555555dad30, 
    in=0x7ffff7e66fc0 <core_dispatch_>, out=0x7fffffffd2c8, provctx=0x7fffffffd2d0)
    at ../../src/src/provider.c:1103
#8  0x00007ffff7c289ee in provider_init (prov=0x5555555dad30)
    at ../openssl/crypto/provider_core.c:930
#9  provider_activate (prov=prov@entry=0x5555555dad30, lock=0, lock@entry=1, 
    upcalls=upcalls@entry=1) at ../openssl/crypto/provider_core.c:1128
#10 0x00007ffff7c29b5c in ossl_provider_activate (aschild=0, upcalls=1, prov=0x5555555dad30)
    at ../openssl/crypto/provider_core.c:1257
--Type <RET> for more, q to quit, c to continue without paging--
#11 ossl_provider_activate (prov=prov@entry=0x5555555dad30, upcalls=upcalls@entry=1, 
    aschild=aschild@entry=0) at ../openssl/crypto/provider_core.c:1243
#12 0x00007ffff7c28051 in provider_conf_load (cnf=0x5555555d1560, 
    value=0x5555555d32d0 "pkcs11_sect", name=0x5555555d32b0 "pkcs11", libctx=0x0)
    at ../openssl/crypto/provider_conf.c:222
#13 provider_conf_init (md=<optimized out>, cnf=0x5555555d1560)
    at ../openssl/crypto/provider_conf.c:304
#14 0x00007ffff7b45bf6 in module_init (cnf=0x5555555d1560, 
    value=0x5555555d2fe0 "provider_sect", name=0x5555555d2f90 "providers", 
    pmod=0x5555555d1c20) at ../openssl/crypto/conf/conf_mod.c:421
#15 module_run (flags=<optimized out>, value=<optimized out>, 
    name=0x5555555d2f90 "providers", cnf=0x5555555d1560)
    at ../openssl/crypto/conf/conf_mod.c:266
#16 CONF_modules_load (flags=<optimized out>, appname=<optimized out>, cnf=0x5555555d1560)
    at ../openssl/crypto/conf/conf_mod.c:165
#17 CONF_modules_load (cnf=0x5555555d1560, appname=<optimized out>, flags=<optimized out>)
    at ../openssl/crypto/conf/conf_mod.c:117
#18 0x00007ffff7b466b4 in CONF_modules_load_file_ex (libctx=libctx@entry=0x0, filename=0x0, 
    appname=0x0, flags=50) at ../openssl/crypto/conf/conf_mod.c:208
#19 0x00007ffff7b467f4 in CONF_modules_load_file (filename=<optimized out>, 
    appname=<optimized out>, flags=<optimized out>) at ../openssl/crypto/conf/conf_mod.c:230
#20 0x00007ffff7b46a0a in ossl_config_int (settings=settings@entry=0x0)
    at ../openssl/crypto/conf/conf_sap.c:68
--Type <RET> for more, q to quit, c to continue without paging--
#21 0x00007ffff7c1d46f in ossl_init_config () at ../openssl/crypto/init.c:249
#22 ossl_init_config_ossl_ () at ../openssl/crypto/init.c:247
#23 0x00007ffff7299f68 in __pthread_once_slow (once_control=0x7ffff7e7e1e0 <config>, 
    init_routine=0x7ffff7c1d460 <ossl_init_config_ossl_>) at ./nptl/pthread_once.c:116
#24 0x00007ffff7c2c7ad in CRYPTO_THREAD_run_once (once=once@entry=0x7ffff7e7e1e0 <config>, 
    init=init@entry=0x7ffff7c1d460 <ossl_init_config_ossl_>)
    at ../openssl/crypto/threads_pthread.c:156
#25 0x00007ffff7c1dc49 in OPENSSL_init_crypto (opts=opts@entry=64, 
    settings=settings@entry=0x0) at ../openssl/crypto/init.c:588
#26 0x00007ffff7c5c79a in ossl_method_store_fetch (store=store@entry=0x5555555906c0, 
    nid=nid@entry=24577, prop_query=0x7ffff7d85b74 "", 
    prov_rw=prov_rw@entry=0x7fffffffd710, method=method@entry=0x7fffffffd610)
    at ../openssl/crypto/property/property.c:514
#27 0x00007ffff7bf0e3f in get_evp_method_from_store (store=0x5555555906c0, 
    prov=0x7fffffffd710, data=0x7fffffffd760) at ../openssl/crypto/evp/evp_fetch.c:172
#28 0x00007ffff7c1a2bb in ossl_method_construct (libctx=<optimized out>, 
    operation_id=operation_id@entry=1, provider_rw=provider_rw@entry=0x7fffffffd710, 
    force_store=force_store@entry=0, mcm=mcm@entry=0x7fffffffd720, 
    mcm_data=mcm_data@entry=0x7fffffffd760) at ../openssl/crypto/core_fetch.c:168
#29 0x00007ffff7bf149e in inner_evp_generic_fetch (
    free_method=0x7ffff7bd9860 <evp_md_free>, up_ref_method=0x7ffff7bd90c0 <evp_md_up_ref>, 
    new_method=0x7ffff7bd9870 <evp_md_from_algorithm>, properties=0x0, 
    name=0x7ffff7f78ef0 "sha1", name_id=<optimized out>, operation_id=1, 
--Type <RET> for more, q to quit, c to continue without paging--
    prov=<optimized out>, methdata=0x7fffffffd760) at ../openssl/crypto/evp/evp_fetch.c:344
#30 evp_generic_fetch (libctx=<optimized out>, operation_id=operation_id@entry=1, 
    name=0x7ffff7f78ef0 "sha1", properties=0x0, 
    new_method=new_method@entry=0x7ffff7bd9870 <evp_md_from_algorithm>, 
    up_ref_method=up_ref_method@entry=0x7ffff7bd90c0 <evp_md_up_ref>, 
    free_method=0x7ffff7bd9860 <evp_md_free>) at ../openssl/crypto/evp/evp_fetch.c:396
#31 0x00007ffff7bd907e in EVP_MD_fetch (ctx=<optimized out>, algorithm=<optimized out>, 
    properties=<optimized out>) at ../openssl/crypto/evp/digest.c:1073
#32 0x00007ffff7627b7f in sc_evp_md (context=0x55555558e250, 
    algorithm=0x7ffff7f78ef0 "sha1") at ../../../src/src/libopensc/ossl.c:32
#33 0x00007ffff7f5b416 in sc_pkcs11_register_openssl_mechanisms (p11card=0x555555597310)
    at ../../../src/src/pkcs11/openssl.c:307
#34 0x00007ffff7f5b112 in sc_pkcs11_register_generic_mechanisms (p11card=0x555555597310)
    at ../../../src/src/pkcs11/mechanism.c:1739
#35 0x00007ffff7f7066a in register_mechanisms (p11card=0x555555597310)
    at ../../../src/src/pkcs11/framework-pkcs15.c:6347
#36 0x00007ffff7f5d959 in pkcs15_bind (p11card=0x555555597310, app_info=0x0)
    at ../../../src/src/pkcs11/framework-pkcs15.c:345
#37 0x00007ffff7f541a6 in card_detect (reader=0x555555596550)
    at ../../../src/src/pkcs11/slot.c:323
#38 0x00007ffff7f5473d in card_detect_all () at ../../../src/src/pkcs11/slot.c:429
#39 0x00007ffff7f49d9b in C_Initialize (pInitArgs=0x0)
    at ../../../src/src/pkcs11/pkcs11-global.c:351
--Type <RET> for more, q to quit, c to continue without paging--
#40 0x000055555555fafe in main (argc=14, argv=0x7fffffffddb8)
    at ../../../src/src/tools/pkcs11-tool.c:1169
(gdb) 
```
We really need to get pkcs11-tool to force the OpenSSL to init before pkcs11-tool calls C_Initialize. 
Waiting till the opensc-pkcs11 is loaded it to late. 

 Since pkcs11-tool loads libopensc but only to access some hex print routines and other utility functions,
I think it could access your OSSL commands to create a libctx in pkcs11-tool.

simo5:
> We really need to get pkcs11-tool to force the OpenSSL to init before pkcs11-tool calls C_Initialize.

Yes, that we have to do, but it should be quite simple, any call into openssl should force that.
Calling OSSL_LIB_CTX_get0_global_default() will force initialization if it didn't happen already.
The result can then be discarded for now.

simo5:
Note that the pkcs11-tool commit is just a temporary workaround, it is *not* the correct fix, because it does not help any other application that may load opensc-pkcs11 in simialr situations (ie before openssl is initialized).
I believe the solution in this case will require us to completely forgot the standard openssl configuration file.
It should be fine because there isn't much this code cares about in the main openssl.cnf

simo5:
An option could be to provide a custom configuration file for opensc-pkcs11, that would have the advantage of letting users customize it if they need, including adding or removing the legacy provider.
That may be a better option than manually loading the providers as the current code does.

Alternatively the configuration file might be optional and could be sourced from opensc.conf only if needed ...

simo5:
Sadly it seems that the only way to init openssl without reading the config would then interfere with an application expecting a normal config to happen ...

I will leave this here for now, I think this patch is still something that needs to be introduced, if for nothing else because the current code is mucking with the default config by surreptitiously loading the legacy provider.

But this code alone does not, in fact, resolve the potential C_initialize loop ...
Will need more thinking about that.

dengert:
Using `OSSL_LIB_CTX_get0_global_default()` does not cause  `OPENSSL_init_crypto` to be run. But `EVP_default_properties_is_fips_enabled(NULL);`  does:
[pkcs11-tool-poke-openssl.diff.txt](https://github.com/OpenSC/OpenSC/files/10751037/pkcs11-tool-poke-openssl.diff.txt) The loop does not occur when but for strange reason.

pkcs11-tool->EVP_default_properties_is_fips_enabled->...->OPENSSL_init_crypto->loads pkcs11-provider->loads-opensc-pkcs11-> pkcs11-provider->C_Initialize ... C_initialize returns   OPENSSL_init_returns  EVP_default_properties_is_fips_enabled returns pkcs11-tool loads opensc-pkcs11, then calls C_Initialize which returns 
```
Main C_Initialize(NULL) rv:CKR_CRYPTOKI_ALREADY_INITIALIZED
Using signature algorithm SHA384-RSA-PKCS-PSS
PSS parameters: hashAlg=SHA384, mgf=MGF1-SHA384, salt_len=48 B
``` 
then goes on to says "Invalid signature"

Without this change the path was opensc-pkcs11 while in C_Initialize to cause OpenSSL to init, anp provider, load opensc-pkcs11 and provider calls C_Initialize which is holding a lock. 

PKCS11 C_Initialize can only return: CKR_ARGUMENTS_BAD, CKR_CANT_LOCK, CKR_CRYPTOKI_ALREADY_INITIALIZED, CKR_FUNCTION_FAILED, CKR_GENERAL_ERROR, CKR_HOST_MEMORY, CKR_NEED_TO_CREATE_THREADS, CKR_OK.
And it does not expect to be called by the same thread while in C_Initialize which was what was happing before. 

And it does make any sense to me, that a PKCS11 application would ever use the pkcs11-provider to load the same PKCS11 module the application has loaded. But there maybe a way around that. If while in C_Initialize, C_Initalize is called by same thread, it should return  CKR_FUNCTION_FAILED or maybe some vender specific CKR, which would then cause the provider not run.

I can look at this Thursday afternoon or Friday.
   








 while in C_initialize causes provider to load

simo5:
I think another way to deal with this could be to just call OPENSSL_init_crypto(0 in C_initialize, before it takes the lock.

It will return CKR_CRYPTOKI_ALREADY_INITIALIZED if nothing else is done.
The question is whether an error should be returned, or if some magic should be done such that both initializations end up returning CKR_OK.

Ideally we have a way to avoid getting OpenSSL to initialize, but the the code that detect cards always will end up calling something that makes openssl initialize apparently.

Sounds like a no-win scenario :-(

simo5:
@dengert after thinking a bit about this I think it should be relatively easy to change C_Initialize so that we make the second initialization in OpenSSL either fail or return an alredy initialized error which pkcs11-provider currently accept as ok.

What should be the actual behavior?
I am thinking that perhaps we should make the second nested initialization actually always fail with a hard error, as you said if an application is using a pkcs11 module directly, openssl should probably not have access to it.

However we should not cause hard error if two initialization happen and they are not nested IMO (because otherwise you might get a hard error in the application if openssl initializes first through some other library load).

I'll propose a patch that handles just the nest (same process) case.

simo5:
@dengert I realized we can do thread protection for nesting checks as well by simply having two separate locking sections, where the one that protect from nesting does not cause, under any circumstance nesting itself. SoI added another patch to that effect.

src/pkcs11/pkcs11-global.c
```diff
@@ -306,12 +307,25 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
...
+	C_INITIALIZE_M_LOCK
+        nesting++;
+        if (nesting > 1) {
+            sc_log(context, "C_Initialize(): Nested init detected");
+            nesting--;
+	    C_INITIALIZE_M_UNLOCK
+            return CKR_GENERAL_ERROR;
+        }
+	C_INITIALIZE_M_UNLOCK
+        /* protect from nesting */
```
frankmorgner, line 320: the above 10 lines look like that's all what is needed to work around the original problem. Do we still need to use libctx then?

frankmorgner (changes requested):

src/libopensc/card-westcos.c
```diff
@@ -659,9 +659,8 @@ static int westcos_get_crypte_challenge(sc_card_t * card, const u8 * key,
 	int r;
 #ifdef ENABLE_OPENSSL
 	EVP_CIPHER_CTX *cctx = NULL;
+        EVP_CIPHER *alg = NULL;
 	int tmplen = 0;
-	if ((cctx = EVP_CIPHER_CTX_new()) == NULL)
-		return SC_ERROR_INTERNAL;
```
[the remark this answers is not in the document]
simo5: Indeed, will fix

src/pkcs11/pkcs11-global.c
```diff
@@ -306,12 +307,25 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
 	in_finalize = 0;
 #endif
 
+        /* protect from nesting */
+	C_INITIALIZE_M_LOCK
+        nesting++;
+        if (nesting > 1) {
+            sc_log(context, "C_Initialize(): Nested init detected");
+            nesting--;
```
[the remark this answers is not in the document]
simo5: It should probably be done at the end of the second initialization indeed, will fix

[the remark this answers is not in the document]
simo5: Actually, I already do it there, I just moved it inside the lock though, it is not ok to nesting-- it outside the lock as another thread may race in i theory.

src/libopensc/opensc.h
```diff
@@ -889,6 +896,14 @@ typedef struct sc_context {
 	sc_thread_context_t	*thread_ctx;
 	void *mutex;
 
+#ifdef ENABLE_OPENSSL
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+        OSSL_LIB_CTX *osslctx;
+        OSSL_PROVIDER *default_provider;
+        OSSL_PROVIDER *legacy_provider;
```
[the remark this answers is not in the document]
simo5: The point of this stuff here is to load these providers in the custom libctx.
providers should not be loaded in the main context and in general loading them randomly in places is not a great thing. We should keep the loading of providers near the libctx creation so that this fact is somewhat under control.

What uses these context is the openssl library, not us directly, we just need to keep handlers to be able to unload them when we free the custom library context

src/pkcs11/pkcs11-global.c
simo5: Absolutely we do, see for example this comment that reiterates modules should use their own context:
https://github.com/openssl/openssl/issues/20315#issuecomment-1436554304

Quoting:
> Ideally, all modules would be using a custom context anyway

simo5:
Rebased the PR to fix the reported issues and two fix indentation as I was made aware this project uses tabs and all my code was spaces

.github/codespell_ignore_words.txt
```diff
@@ -20,3 +20,4 @@ responde
 sie
 signatur
 standarts
+Sorce
```
simo5, line 23: why is this not working to deal with the codespell action ?

src/libopensc/opensc.h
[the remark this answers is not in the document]
frankmorgner: The point I was trying to make is that we're calling OpenSSL APIs to *request* the `OSSL_PROVIDER`s without ever using the pointers within OpenSC. But if I get your response right, you're saying that by requesting that pointer, we are actually loading the providers into our custom `OSSL_LIB_CTX`, right?

src/pkcs11/pkcs11-global.c
frankmorgner: thanks for the clearification

src/libopensc/opensc.h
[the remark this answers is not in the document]
simo5: That's correct, as you can see OSSL_PROVIDER_load() takes the custom context as first parameter. So what happens here is that we load those providers in the specific context (instead of the default context when NULL is passed).
And that's the whole point. We keep around the pointers because providers need to be freed later when we destroy the custom context.
See the calls to OSSL_PROVIDER_unload().

frankmorgner:
I'm thinking about how this could be integrated a little less invasive, but I have not made up my mind completely, so please look at the following as suggestions...

1. As header file I'd use sc-ossl-compat.h and implementation should be put into sc-ossl-compat.c
2. the new functions you're introducing (`sc_evp_cipher` and the like), are so simple that they could be implemented as static and possibly inlined functions in sc-ossl-compat.h. This way, you don't need to extend the exports from libopensc
3. creating something like an `sc_openssl_init()` would allow handling of the OpenSSL version specific stuf outside of ctx.c. Also, this function would allow creating a local static variable, which holds the pointer to the default/legacy provider without exposing it in sc_context_t. Additionally, a local static pointer to a libctx could be held here, too, which will be then added to any sc_context_t that the library creates until some `sc_openssl_close()` resets this local pointer to NULL (i.e. a third function `sc_openssl_get_libctx()`. `sc_openssl_init()`/`sc_openssl_close()`/`sc_openssl_get_libctx()` can be left as pure internal functions of libopensc - if some external library needs the context, it consults the pointer from sc_context_t
4. I wonder if initialization with sc_context_create() is the best place for initializing libctx. For some functionality of this type, we've used shared library constructors to implement this. In the minidriver and the PKCS#11 module we've already added the initialization/termination of notifications, OpenSSL's secure memory and initialization of libeac. It seems more natural to do this once when loading the library instead of every time when we are creating a context.

src/libopensc/card-epass2003.c
```diff
@@ -456,36 +461,50 @@ aes128_encrypt_cmac(const unsigned char *key, int keysize,
...
 }
 
 
 static int
-aes128_decrypt_cbc(const unsigned char *key, int keysize, unsigned char iv[16],
+aes128_decrypt_cbc(struct sc_card *card, const unsigned char *key, int keysize, unsigned char iv[16],
 		const unsigned char *input, size_t length, unsigned char *output)
 {
-	return openssl_dec(EVP_aes_128_cbc(), key, iv, input, length, output);
+	EVP_CIPHER *alg = sc_evp_cipher(card->ctx, "AES-128-CBC");
```
Jakuje, line 492: do we need some error checking in this case or can the openssl functions handle NULL `EVP_CIPHER` in reasonable way (like returning an error instead of crashing).

src/libopensc/card-oberthur.c
```diff
@@ -1345,12 +1346,16 @@ auth_update_component(struct sc_card *card, struct auth_update_component_info *a
...
-		if (args->len == 24)
-			EVP_EncryptInit_ex(ctx, EVP_des_ede(), NULL, args->data, NULL);
-		else
-			EVP_EncryptInit_ex(ctx, EVP_des_ecb(), NULL, args->data, NULL);
+		if (args->len == 24) {
+			alg = sc_evp_cipher(card->ctx, "DES-EDE");
+			EVP_EncryptInit_ex(ctx, alg, NULL, args->data, NULL);
+		} else {
+			alg = sc_evp_cipher(card->ctx, "DES-ECB");
+			EVP_EncryptInit_ex(ctx, alg, NULL, args->data, NULL);
```
Jakuje, line 1354: nit: the second line in both branches is the same so I would prefer to keep it out of the branching once we already touch this.

src/libopensc/cwa14890.c
```diff
@@ -1896,6 +1883,9 @@ int cwa_decode_response(sc_card_t * card,
 	}
 
 	/* finally apply 3DES to result */
+	sc_evp_cipher_free(alg);
+	alg = sc_evp_cipher(card->ctx, "DES-EDE-ECB");
+
 	if (EVP_EncryptInit_ex(cctx, EVP_des_ede_ecb(), NULL, key, NULL) != 1 ||
```
Jakuje, line 1889: this is missing the `alg` usage as defined on the previous line:
```suggestion
	if (EVP_EncryptInit_ex(cctx, alg, NULL, key, NULL) != 1 ||
```

src/libopensc/cwa14890.c
```diff
@@ -1952,6 +1942,9 @@ int cwa_decode_response(sc_card_t * card,
 
 		/* decrypt into response buffer
 		 * by using 3DES CBC by mean of kenc and iv={0,...0} */
+		sc_evp_cipher_free(alg);
+		alg = sc_evp_cipher(card->ctx, "DES-EDE-CBC");
+
 		if (EVP_DecryptInit_ex(cctx, EVP_des_ede_cbc(), NULL, key, iv) != 1 ||
```
Jakuje, line 1948: same here
```suggestion
		if (EVP_DecryptInit_ex(cctx, alg, NULL, key, iv) != 1 ||
```

src/libopensc/opensc.h
```diff
@@ -1731,6 +1746,20 @@ iso7816_build_pin_apdu(struct sc_card *card, struct sc_apdu *apdu,
  */
 void sc_free(void *p);
 
+#ifdef ENABLE_OPENSSL
+#include <openssl/evp.h>
```
Jakuje, line 1750: I would vote for having the includes on the top of the file, instead of having them buried in here, but if others are ok with this, I can live with that.

simo5:
> 

For 1, 2, 3 I am ok making the changes. If you tell me to do them I will do, the change is simple.

Please let me know if I should rebase the PR with the above changes.

For 4 I am not sure I agree, generally you want a context that you can use in parallel with other uses w/o conflicts. I was under the impression you could end having two separate sc_context_t in an application, but perhaps in the end you always end up with a global one.
But in either case it is not clear to me why you would create a context at library initialization, doing things in constructors is always iffy and risk bringin in hard to debug issues in my experience. so I personally try to avoid adding stuff to library constructors unless it is really necessary.

src/libopensc/card-epass2003.c
simo5: Not really, the next function will fail if we pass in a NULL cipher anyway, so I thought that is sufficient, these are all hard-coded names, so it is extremely unlikely a failure would happen anyway.

simo5: (It is no more likely to fail than the previous version where a function was called inlined in the Init() calls.

src/libopensc/card-oberthur.c
simo5: ok

src/libopensc/cwa14890.c
simo5: good catch

simo5:
Just pushed a rebase to address the technical issues found by @Jakuje 
Next I will rebase once more to implement the changes asked by @frankmorgner (expect change 4 which I would like to see if really needed)

simo5:
Rereading 3 it may be a little too much, but working on at least parts of it.

frankmorgner:
Yes, different sc_context_t are possible in a single application, e.g. PKCS#11 app that uses multiple threads each having an initialized session with the library. But it is not needed to have a separated libctx for every one of these.

Anyway, I'm still thinking what would be the most elegant/efficient/effective way of this change...

simo5:
I changed the code to move functions to static inlines, and by keeping two helper functions in ctx.c ended up with not needing a new .c file at all nor any changes to exports files.

You may see a coupl odd defines for fucntiosn that take an osslctx, I opted that way while doing the rebase because dereferencing the sc_context_t directly in the header file was causing issues as opensc.h definitions of sc_context_t come after sc-ossl-compat.h is already included.

I could expose directly the ossl3ctx_t pointer but then I would need to define a dummy one for < 3.0.x callers to maintain the same interface.

It is a compromise it can be changed as taste dictates.

frankmorgner:
I think this looks like a good compromise.

Let's wait if someone else has an opinion about it, especially on where/when to do the OpenSSL initialization.

frankmorgner:
There's a build problem on Windows, but I can't exactly see what is going wrong...
https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/46313686/job/qfgxsnsvaqnwyvxn

Jakuje:
> There's a build problem on Windows, but I can't exactly see what is going wrong... https://ci.appveyor.com/project/LudovicRousseau/opensc/builds/46313686/job/qfgxsnsvaqnwyvxn

This is a build without openssl so I assume there is some issue in detection if the openssl support should be built or not. It looks like also the linux build is now failing.

simo5:
Indeed it looks like it is trying to build the openssl header, but I do not see how, it is all wrapped in #ifdef ENABLE_OPENSSL ...

simo5:
#include <openssl/evp.h> was missing from the else branch, but I still do not understand why that file would be built at all on windows, given ENABLE_OPENSSL should be off and nothing in that file should be built.

simo5:
The macOS build also seem choking on a file that should not be built, seem like the ENABLE_OPENSSL guard is not working?

simo5:
Am I causing the isoapplet v1 to fail somehow? I do not see anything related to my changes in the error it spits (fail to make install)

simo5:
Btw looking that isoapplet failure it says one of the github actions is deprecated and will be removed soon, someone may want to take a look at that.

Jakuje:
The macos and isoapplet is fixed with #2717 -- not your fault.

Jakuje:
> The macos and isoapplet is fixed with #2717 -- not your fault.

The CI should be fixed now if your rebase on current master.

simo5:
@Jakuje @frankmorgner is there anything more I should do in this PR?

src/libopensc/card-entersafe.c
```diff
@@ -220,20 +222,27 @@ static int entersafe_cipher_apdu(sc_card_t *card, sc_apdu_t *apdu,
...
-		  EVP_EncryptInit_ex(ctx, EVP_des_ecb(), NULL, key, iv);
-	 else if (keylen == 16) 
-		  EVP_EncryptInit_ex(ctx, EVP_des_ede(), NULL, key, iv);
-	 else
+	if (keylen == 8) {
+	 	alg = sc_evp_cipher(card->ctx, "DES-ECB");
+		EVP_EncryptInit_ex(ctx, alg, NULL, key, iv);
+	} else if (keylen == 16) {
+	 	alg = sc_evp_cipher(card->ctx, "DES-EDE");
+		EVP_EncryptInit_ex(ctx, alg, NULL, key, iv);
```
Jakuje, line 230: maybe one more nit that the the line `EVP_EncryptInit_ex(ctx, alg, NULL, key, iv);` is same in both of the condition so it can be moved out to avoid code duplication (similarly as in the previously mentioned case)

src/libopensc/ctx.c
```diff
@@ -923,6 +955,17 @@ int sc_wait_for_event(sc_context_t *ctx, unsigned int event_mask, sc_reader_t **
...
+void sc_openssl3_deinit(sc_context_t *ctx)
+{
+	if (ctx->ossl3ctx.legacyprov) OSSL_PROVIDER_unload(ctx->ossl3ctx.legacyprov);
+	if (ctx->ossl3ctx.defprov) OSSL_PROVIDER_unload(ctx->ossl3ctx.defprov);
+	if (ctx->ossl3ctx.libctx) OSSL_LIB_CTX_free(ctx->ossl3ctx.libctx);
+	ctx->ossl3ctx.legacyprov = NULL;
+	ctx->ossl3ctx.defprov = NULL;
+	ctx->ossl3ctx.libctx = NULL;
+}
+#endif
```
Jakuje, line 968: this is also nit, but I would prefer to keep move the deinit function to the same ifdef where the init lives to keep them together. But I understand the idea keeping that close to the "release context" functions.

src/libopensc/ctx.c
```diff
@@ -804,6 +804,29 @@ int sc_context_repair(sc_context_t **ctx_out)
 	return SC_SUCCESS;
 }
 
+#ifdef USE_OPENSSL3_LIBCTX
+int sc_openssl3_init(sc_context_t *ctx)
```
Jakuje, line 808: If I see right, this is used only here, so I would prefer to keep it static.

Jakuje (commented):
> @Jakuje @frankmorgner is there anything more I should do in this PR?

I added few comments, but these are mostly nits or personal preferences so I do not insist on them.
In any case, please, rebase the changes on current master to resolve the CI issues. I think that after that we should be good to go.

src/libopensc/card-entersafe.c
simo5: TBH the quality of the code of this card is soo poor I did not want to change the structure at all. It doesn't even do error checking, and I already fixed a memory leak here ... sounds like the code of this driver could use an overhaul.
I'd prefer to leave as is.

src/libopensc/ctx.c
simo5: Yeah sounds better to me to keep the code close together, we'd only go down from 4 ifdefs blocks to 3, doesn't look like we'd gain much.

src/libopensc/ctx.c
simo5: ok fixing this and the previous point given they are the same area of code

simo5:
Rebased and ctx.c nits addressed

src/libopensc/card-entersafe.c
Jakuje: np. Lets keep it this way

Jakuje:
For some reason the MacOS build failed  
```
  CC       opensc_asn1-opensc-asn1.o
In file included from opensc-asn1.c:25:
In file included from ../../src/libopensc/asn1.h:28:
In file included from ../../src/libopensc/opensc.h:47:
../../src/libopensc/sc-ossl-compat.h:30:10: fatal error: 'openssl/opensslv.h' file not found
#include <openssl/opensslv.h>
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```
I think this was not an issue on master couple of days back. But I do not see you changing that part of the file, which confuses me why it should be failing. @frankmorgner any ideas?

mouse07410:
I observed such errors when the path to OpenSSL header files got lost by the build tools.

Configuring with `--disable-silent-rules` could help determining the problem source.

frankmorgner:
This pull request is the first which includes OpenSSL dependencies into libopensc's headers that are considered as external interface by adding `#include "sc-ossl-compat.h" to `opensc.h`. Conseuently, the build of `src/tool/opensc-asn1.c` fails, because it doesn't need any OpenSSL for parsing ASN.1 data.

The trivial way to solve this would be to add the missing OPENSSL_CFLAGS to `opensc-asn1`.  The other option would be to hide the OpenSSL types in an dynamically allocated buffer as done in https://github.com/OpenSC/OpenSC/commit/a9d9f2fda424115fdf8b4369cf33ffbb5e95d300. I'd opt to go for the second option...

Jakuje:
Thanks for the pointer. My reading is that the `ENABLE_OPENSSL` should not come from `OPENSSL_CFLAGS`, but from `config.h`, which is included in the `opensc-asn1.c` so I am not sure if the first solution would solve this.

dengert:
Question: In OpenSC code uses  `EVP_DigestInit_ex` vs `EVP_DigestInit_ex2` https://www.openssl.org/docs/man3.1/man3/EVP_DigestInit.html defines these as:
```
int EVP_DigestInit_ex(EVP_MD_CTX *ctx, const EVP_MD *type, ENGINE *impl);
int EVP_DigestInit_ex2(EVP_MD_CTX *ctx, const EVP_MD *type,
                       const OSSL_PARAM params[]);
```
Should `EVP_DigestInit_ex2` be used instead?

simo5:
Only if you need to pass parameters, internally they use the same initialization function.

Jakuje:
As expected, this did not fix the OSX build so I would propose to use the solution proposed by Frank:

https://github.com/OpenSC/OpenSC/commit/a9d9f2fda424115fdf8b4369cf33ffbb5e95d300#diff-00a07a0f9b8dac152ab519cf04b8da9c54e33d0593f1e00c7f33a3dc7fa0b748

simo5:
Working on it

simo5:
Changed as requested, compiles here, but haven't run tests myself.

Jakuje (approved):

frankmorgner (approved):
thank you

Jakuje:
Thank you for the patience. Tested locally and I did not notice any issues. Merging.

dengert:
@Jakuje  You my also want to commit #2715

PR #2715, opened 2023-02-21
pkcs11-tool to use OpenSSL libctx

dengert:
This is in addition to "Introduce use of custom ossl libctx with OpenSSL >= 3.0" https://github.com/OpenSC/OpenSC/pull/2712

pkcs11-tool uses some functions found in libopensc, but does not create a sc_context like other OpenSC tools as the pkcs11 module can be any pkscs11 module.

There is one OpenSSL function "d2i_PUBKEY_bio" that does not have an equivalent "d2i_PUBKEY_ex_bio" in 3.0.8.  It is listed in OpenSSL master. See: https://github.com/openssl/openssl/pull/18427

 On branch ossl_lib_ctx-pkcs11-tool
 Changes to be committed:

##### Checklist

- [X] PKCS#11 module is tested with OpenSSL 3.0.8 and 1.1.1t
- [ ] macOS tokend is tested
- [X] With OpenSSL 3.0.8 with pkcs11-provider and #2712

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -73,8 +73,13 @@
 #include "util.h"
 #include "libopensc/sc-ossl-compat.h"
 
+/* pkcs11-tool uses libopensc routines that do not use an sc_context
+ * but does use some OpenSSL routines
+ */
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 	static OSSL_PROVIDER *legacy_provider = NULL;
+	static OSSL_PROVIDER *default_provider = NULL;
+	static OSSL_LIB_CTX *osslctx = NULL;
 #endif
 
 #ifdef _WIN32
@@ -751,6 +756,19 @@ int main(int argc, char * argv[])
 		util_fatal("Cannot set FMODE to O_BINARY");
 	if(_setmode(_fileno(stdin), _O_BINARY ) == -1)
 		util_fatal("Cannot set FMODE to O_BINARY");
+#endif
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+		if (!osslctx) {
+			if (!(osslctx = OSSL_LIB_CTX_new())) {
+				util_fatal("Failed to create OpenSSL OSSL_LIB_CTX\n");
+			}
+		}
+		if (!default_provider) {
+			if (!(default_provider = OSSL_PROVIDER_load(osslctx, "default"))) {
+				util_fatal("Failed to load OpenSSL \"default\" provider\n");
+			}
+		}
 #endif
 
 	while (1) {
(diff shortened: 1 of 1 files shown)


dengert:
https://github.com/simo5 can you have a look at this PR. And if you want, add to #2712?

simo5:
I am in favor of this if you find it useful, but pkcs11-tool is an end-user applciation, so it can definitely use the default context as its user will never conflict with anything else (except bad libraries perhaps).

src/tools/pkcs11-tool.c
```diff
@@ -4755,7 +4790,12 @@ derive_ec_key(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key, CK_MECHANISM_TYPE
 	if (BIO_read_filename(bio_in, opt_input) <= 0)
 		util_fatal("Cannot open %s: %m", opt_input);
 
+#if OPENSSL_VERSION_NUMBER >= 0x30200000L
+/* TODO following is in OpenSSL master not  3.0.8 */
```
Jakuje, line 4794: please, add a link to opensc commit or pull request implementing this (if you find it) so it is easy to recheck and update the condition accordingly when it will be in some released version.

dengert: Comments added https://github.com/OpenSC/OpenSC/pull/2715/commits/f216aa86fe3315d8cb7d97e3709655581cde19cc
Fix is in master, should be in 3.2.0.

src/tools/pkcs11-tool.c
```diff
@@ -753,6 +758,19 @@ int main(int argc, char * argv[])
 		util_fatal("Cannot set FMODE to O_BINARY");
 #endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+		if (!osslctx) {
```
Jakuje, line 762: we are in main. Is this check needed? I think in this case, it will be always true, unless we will introduce some wild goto logic. The same for the `default_provider` check below

src/tools/pkcs11-tool.c
```diff
@@ -5630,6 +5680,7 @@ static int read_object(CK_SESSION_HANDLE session)
 					util_fatal("cannot parse EC_PARAMS");
 				EVP_PKEY_assign_EC_KEY(pkey, ec);
 #else
+/* TODO needs debugging */
```
Jakuje, line 5683: does this mean that it does not work now or something else? The comment is not much self-explanatory.

Jakuje (approved):
two minor comments, but generally looks ok. I did not check though all the OpenSSL uses.

src/tools/pkcs11-tool.c
dengert: The tests could be removed if you want.

src/tools/pkcs11-tool.c
dengert: It was meant that I did not test this and `pkcs11-tool --test` does not test ECDSA.  I will  try `pkcs11-tool -sign` with an ec key and verify the signature.

dengert:
I addressed the two comments from above and forced pushed an update. 

Also tested with OpenSSL 3.0.8 and development built with "no-deprecated":
`/usr/bin/perl ../src/Configure --prefix=/opt/ossl-dev_ND -g -Wl,--enable-new-dtags,-rpath=$(LIBRPATH) -g no-deprecated linux-x86_64 shared`

```
#!/bin/bash
set -xv 

# test pkcs11-tool to sign and verify with ECDSA
# Using NIST Beta Test Card 5

PATH=/opt/ossl-dev_ND/bin:$PATH

echo "Hello world!" > /tmp/sign-demo.data.txt

gdb --args pkcs11-tool --sign --login --id 01  --mechanism ECDSA-SHA256 --input /tmp/sign-demo.data.txt \
        --output-file /tmp/sign-demo.signature --signature-format openssl

pkcs11-tool --verify --id 01 --mechanism ECDSA-SHA256 --input /tmp/sign-demo.data.txt \
        --signature-file /tmp/sign-demo.signature --signature-format openssl

pkcs15-tool --read-public-key 01  > /tmp/sign-demo.pubkey.pem

openssl dgst -sha256  -verify /tmp/sign-demo.pubkey.pem \
        -signature /tmp/sign-demo.signature < /tmp/sign-demo.data.txt
```

Jakuje (approved):
lgtm. Thanks for the help!

PR #2717, opened 2023-02-22
update ax_pthread to version 31 (autoconf-archive-2022.02.11)

Jakuje:
The OSX build was failing with the old version giving the following errors:

```
autoreconf: running: /usr/local/Cellar/autoconf/2.71/bin/autoconf --force
configure.ac:485: warning: $as_echo is obsolete; use AS_ECHO(["message"]) instead
```

### DIFF
--- a/.github/setup-macos.sh
+++ b/.github/setup-macos.sh
@@ -2,7 +2,7 @@
 
 set -ex -o xtrace
 
-brew install automake gengetopt help2man
+brew install automake gengetopt help2man pkgconfig
 
 # openSCToken
 export PATH="/usr/local/opt/ccache/libexec:$PATH"
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -11,6 +11,7 @@ on:
       - .github/push-artifacts.sh
       - '**.am'
       - doc/**
+      - configure.ac
   push:
 
 permissions:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,6 +12,7 @@ on:
       - .github/push-artifacts.sh
       - '**.am'
       - doc/**
+      - configure.ac
   push:
 
 permissions:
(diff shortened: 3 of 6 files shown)


Jakuje:
Added the changes to the github workflow files to trigger the ci on the configure change from #2698.

frankmorgner:
Github seems to have runners that are configured differently. In these two builds of this PR, pkg-config is preinstalled in one image, but missing in the other (check the step that runs setup-macos.sh):
- https://github.com/OpenSC/OpenSC/actions/runs/4246825455/jobs/7384116145
- https://github.com/OpenSC/OpenSC/actions/runs/4246635128/jobs/7383689408

Your suggestion to always install pkg-config should work in both cases now

frankmorgner:
One option of handling the warning about MACOSX_DEPLOYMENT_TARGET would be to be raise the requirement. A migration tool is provided at https://github.com/flutter/flutter/pull/107689, but I'm unsure if it would be applicable for our projects. An other option would be to use Xcode 14, which still ships the old SDK by pinning the Github Runner to [macos-12](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md)

Jakuje:
> One option of handling the warning about MACOSX_DEPLOYMENT_TARGET would be to be raise the requirement. A migration tool is provided at [flutter/flutter#107689](https://github.com/flutter/flutter/pull/107689), but I'm unsure if it would be applicable for our projects. An other option would be to use Xcode 14, which still ships the old SDK by pinning the Github Runner to [macos-12](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md)

I don't understand the pros and cons of either of the ways. Can you have a look into this? Should I open a new issue to track this?

I think this can be merged so we do not fail all poeple's PR's at least. If not, please comment.

frankmorgner:
I've created https://github.com/OpenSC/OpenSC/pull/2720

frankmorgner (approved):

PR #2720, opened 2023-02-24
CI: pin macos-12 to allow building for macOS 10.12

frankmorgner:
Xcode 14 is the last version which ships the SDK for 10.12. So we are pinning the Github Runner to macos-12, which still has this version installed. This avoids the following warning:

/Users/runner/work/OpenSC/OpenSC/OpenSCToken/OpenSCTokenApp.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.12, but the range of supported deployment target versions is 10.13 to 13.1.99. (in target 'OpenSCTokenApp' from project 'OpenSCTokenApp')
warning: ONLY_ACTIVE_ARCH=YES requested with multiple ARCHS and no active architecture could be computed; building for all applicable architectures (in target 'OpenSCTokenApp' from project 'OpenSCTokenApp')

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -20,7 +20,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-12
     steps:
       - uses: actions/checkout@v3
       - run: .github/setup-macos.sh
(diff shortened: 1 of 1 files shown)


frankmorgner:
I think the problem is more related to https://github.com/OpenSC/OpenSC/pull/2729

PR #2721, opened 2023-02-24
minidriver: fix support for pinless keys

llogar:
A call to `CardGetProperty('PIN Information', ...)` should not fail for keys that are not protected with PINs (`PinId = ROLE_EVERYONE`). Fixes #2719

##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [ ] PKCS#11 module is tested
- [x] Windows minidriver is tested
- [ ] macOS tokend is tested

Issue #2718, opened 2023-02-23
minidriver with ECC signing failing


llogar:
### Problem Description
As already mentioned in https://github.com/OpenSC/OpenSC/pull/2646#issuecomment-1325438662, padding info is added in `CardSignData()` even when non-RSA keys are being used, which makes no sense.

As a consequence signing (when using minidriver and EC keys) fails 
### Proposed Resolution
I guess the padding info should be included only if RSA keys are being used, surrounding the padding code in 
`sc_pkcs15_compute_signature()` with
```
if (pkey->type == SC_PKCS15_TYPE_PRKEY_RSA) {
...
}
```
### Steps to reproduce
Try to sign data with EC keys using minidriver...
### Logs
```
P:20188; T:12200 2023-02-23 09:27:00.202 [cardmod] CardSignData dwVersion=2, bContainerIndex=0, dwKeySpec=7, dwSigningFlags=0x00000000, aiHashAlg=0x00000000
...
P:20188; T:12200 2023-02-23 09:27:00.214 [cardmod] CARD_PADDING_INFO_PRESENT not set
...
P:20188; T:12200 2023-02-23 09:27:00.227 [cardmod] pkcs15-sec.c:607:sc_pkcs15_compute_signature: called
P:20188; T:12200 2023-02-23 09:27:00.230 [cardmod] pkcs15-sec.c:660:sc_pkcs15_compute_signature: supported algorithm flags 0x100100, private key usage 0x104
P:20188; T:12200 2023-02-23 09:27:00.232 [cardmod] pkcs15-sec.c:718:sc_pkcs15_compute_signature: ECDSA using SC_ALGORITHM_ECDSA_RAW flags before 0x00000002
P:20188; T:12200 2023-02-23 09:27:00.235 [cardmod] padding.c:618:sc_get_encoding_flags: called
P:20188; T:12200 2023-02-23 09:27:00.238 [cardmod] padding.c:622:sc_get_encoding_flags: iFlags 0x100002, card capabilities 0x100100
P:20188; T:12200 2023-02-23 09:27:00.240 [cardmod] padding.c:678:sc_get_encoding_flags: pad flags 0x1, secure algorithm flags 0x102
P:20188; T:12200 2023-02-23 09:27:00.243 [cardmod] padding.c:679:sc_get_encoding_flags: returning with: 0 (Success)
[...]

PR #2722, opened 2023-02-26
Minidriver EC Changes for private keys  that do not need a PIN and/or ECDSA failure to sign

dengert:
@llogar has been working on EOI cards with minidriver and has run into two problems which can be problem with other cards such as PIV.

> A private key that does not need a PIN does not work.
> ECDSA signatures failing because minidriver passed  SC_ALGORITHM_RSA_PAD_PKCS1 or other  SC_ALGORITHM_RSA_PAD_* flags to sc_pkcs15_compute_signature. 

This PR is being submitted to take advantage of the Appveyor build for testing with PIV card.   

##### Checklist

- [todo] Windows minidriver is tested
- [ ] macOS tokend is tested

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -6381,7 +6381,7 @@ DWORD WINAPI CardGetProperty(__in PCARD_DATA pCardData,
 		if (dwFlags >= MD_MAX_PINS)
 			MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
-		if (!vs->pin_objs[dwFlags])
+		if (dwFlags != ROLE_EVERYONE && vs->pin_objs[dwFlags] == NULL)
 			MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
 		p->PinType = vs->reader->capabilities & SC_READER_CAP_PIN_PAD
@@ -6389,6 +6389,18 @@ DWORD WINAPI CardGetProperty(__in PCARD_DATA pCardData,
 			? ExternalPinType : AlphaNumericPinType;
 		p->dwFlags = 0;
 		switch (dwFlags)   {
+			case ROLE_EVERYONE:
+				logprintf(pCardData, 2,
+					"returning info on PIN ROLE_EVERYONE [%lu]\n",
+					(unsigned long)dwFlags);
+				p->PinType = 0;   /* There is no pin, so don't need reader capabilities */
+				p->PinPurpose = 0; /* It can not be PrimaryCardPin */
+				p->PinCachePolicy.dwVersion = PIN_CACHE_POLICY_CURRENT_VERSION;
+				p->PinCachePolicy.PinCachePolicyType = PinCacheNone;
+				p->PinCachePolicy.dwPinCachePolicyInfo = 0;
+				p->dwChangePermission = 0;
+				break;
+
 			case ROLE_ADMIN:
 				logprintf(pCardData, 2,
 					  "returning info on PIN ROLE_ADMIN ( Unblock ) [%lu]\n",
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -4696,7 +4696,7 @@ DWORD WINAPI CardSignData(__in PCARD_DATA pCardData, __inout PCARD_SIGNING_INFO 
 	ALG_ID hashAlg;
 	sc_pkcs15_prkey_info_t *prkey_info;
 	BYTE dataToSign[0x200];
-	int opt_crypt_flags;
+	int opt_crypt_flags = 0;
 	size_t dataToSignLen = sizeof(dataToSign);
 	sc_pkcs15_object_t *pkey;
 
@@ -4788,7 +4788,7 @@ DWORD WINAPI CardSignData(__in PCARD_DATA pCardData, __inout PCARD_SIGNING_INFO 
 		 * padding and use the value in aiHashAlg. */
 		logprintf(pCardData, 3, "CARD_PADDING_INFO_PRESENT not set\n");
 
-		opt_crypt_flags = SC_ALGORITHM_RSA_PAD_PKCS1;
+		opt_crypt_flags = SC_ALGORITHM_RSA_PAD_PKCS1;  /* turn off later if key is EC */
 		if (hashAlg == CALG_MD5)
 			opt_crypt_flags |= SC_ALGORITHM_RSA_HASH_MD5;
 		else if (hashAlg == CALG_SHA1)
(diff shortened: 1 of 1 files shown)


src/libopensc/padding.c
```diff
@@ -653,7 +653,7 @@ int sc_get_encoding_flags(sc_context_t *ctx,
 		*sflags = SC_ALGORITHM_RSA_PAD_NONE;
 		*pflags = iflags;
 
-	} else if ((caps & (SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE)) &&
+	} else if ((caps & SC_ALGORITHM_RSA_PAD_PKCS1) && (caps & SC_ALGORITHM_RSA_HASH_NONE) &&
```
llogar, line 656: My problem with encoding flags is not covered with this. See my log in #2718:
```
P:20188; T:12200 2023-02-23 09:27:00.238 [cardmod] padding.c:622:sc_get_encoding_flags: iFlags 0x100002, card capabilities 0x100100
```
`caps` doesn't have `SC_ALGORITHM_RSA_PAD_PKCS1` set.

I'd rather fix this in minidriver code with something like:
```
--- minidriver.c
+++ minidriver.c
@@ -4978,6 +4978,7 @@
         dwret = SCARD_E_INVALID_VALUE;
         goto err;
     }
+    opt_crypt_flags &= ~SC_ALGORITHM_RSA_PADS;
   } else {
     logprintf(pCardData, 0, "invalid private key\n");
     dwret = SCARD_E_INVALID_VALUE;
```
and exclude any padding flags there when doing ECDSA?

dengert: Well it does.  And it looks like minidriver sets it when it should not. 

In minidriver.c (line numbers):
4692 DWORD WINAPI CardSignData(__in PCARD_DATA pCardData, __inout PCARD_SIGNING_INFO pInfo)
4699         int opt_crypt_flags;

In one of you logs I see: "CARD_PADDING_INFO_PRESENT not set:

4789                 logprintf(pCardData, 3, "CARD_PADDING_INFO_PRESENT not set\n");
4791                 opt_crypt_flags = SC_ALGORITHM_RSA_PAD_PKCS1;

This code is from cardmod when Windows only supported RSA. This looks like the problem.

4947                 r = sc_pkcs15_compute_signature(vs->p15card, pkey, opt_crypt_flags, dataToSign, dataToSignLen, pbuf, lg, NULL);

In logL sc_pkcs15_compute_signature: ECDSA using SC_ALGORITHM_ECDSA_RAW flags before 0x00000002

the "before" is the flags before they are change. which is SC_ALGORITHM_RSA_PAD_PKCS1

So my change sort of fixed it (but may break something else), but the real fix need to be in minidriver around line 4791  which should not not do any of that if not RSA.

Is that where you wanted to make the change?

llogar: Yes, See this commit:

https://github.com/llogar/OpenSC/commit/4c31a1d4213093886859d8bff7c03dc1f212b137

Altough this clears padding flags post festum rather than prevents setting them in the first place...

src/minidriver/minidriver.c
```diff
@@ -6381,14 +6381,26 @@ DWORD WINAPI CardGetProperty(__in PCARD_DATA pCardData,
 		if (dwFlags >= MD_MAX_PINS)
 			MD_FUNC_RETURN(pCardData, 1, SCARD_E_INVALID_PARAMETER);
 
-		if (!vs->pin_objs[dwFlags])
+		if (!dwFlags && !vs->pin_objs[dwFlags])
```
llogar, line 6385: At least in my case, when there is no PIN, it still returns `SCARD_E_INVALID_PARAMETER`, as this condition is met (`dwFlags == 0 && vs->pin_objs[dwFlags] == NULL`)

Possibly it should return "empty" PIN info for all `vs->pin_objs[dwFlags] == NULL` cases, as there may even be cards with ROLE_ADMIN and/or ROLE_USER (and not just ROLE_EVERYONE) without a PIN?

dengert: OK, it should be:
`if (dwFlags && !vs->pin_objs[dwFlags])`
or to make it more readable:
`if (dwFlags != ROLE_EVERYONE)  && (vs->pin_objs[dwFlags] == NULL))`

As for none real pins `vs->pin_objs` is filled in from PKCS15 structures. 

Then  if `dwFlags == ROLE_EVERYONE` it will fall through to the  case ROLE_EVERYONE:'  and log a message and endup 
6510 with another log message and hexdump or *p

Will update and try again today.

llogar:
Thanks, It works now.

dengert:
With updated (forced pushed) if statement: md_log shows:

```
returning info on PIN ROLE_EVERYONE [0]
returns 'PIN Information' --- 00000265ED666550:36
 0000  06000000 00000000 00000000 00000000  00000000 06000000 02000000 00000000
 0020  00000000 
MD_Function:CardGetProperty:6511 returning with: 0x00000000
```

`certutil -scinfo` shows all 4 certificates, where Certificate 3 does not require a PIN
Note: "Private key verifies" 
```
================ Certificate 3 ================
--- Reader: Identiv SCR3500 A Contact Reader 0
---   Card: NIST DEMO Oberthur test cards
Provider = Microsoft Smart Card Key Storage Provider
Key Container = Certificate for Card Authentication

Serial Number: 01f6
Issuer: CN=Test ECC P-384 CA for Test PIV Cards, OU=Test CA, O=Test Certificates 2010, C=US
 NotBefore: 10/1/2010 2:30 AM
 NotAfter: 10/1/2030 2:30 AM
Subject: SERIALNUMBER=D650185A13422C2267829D916CD89080501E649C86501843E2, OU=Test Agency, OU=Test Department, O=Test Government, C=US
Non-root Certificate
Cert Hash(sha1): 59618196f7713320e2c758117903141d415282f0

Performing  public key matching test...
Public key matching test succeeded
  Key Container = Certificate for Card Authentication
  Provider = Microsoft Smart Card Key Storage Provider
  ProviderType = 0
  Flags = 1
    0x1 (1)
  KeySpec = 0 -- XCN_AT_NONE
Private key verifies
Microsoft Smart Card Key Storage Provider: KeySpec=0
AES256+RSAES_OAEP(ECC:CNG) test skipped
```

Jakuje:
Whats the status of this PR? Should we merge the ECC part and drop/postpone the discussed/disputed role_everyone as it is already discussed in #2721?

dengert:
This PR replaces #2721 and adds some additional changes. It was tested using PIV card  and `certutil.exe` as noted in https://github.com/OpenSC/OpenSC/pull/2722#issuecomment-1450190067

@llogar  author of #2712  says it worked in https://github.com/OpenSC/OpenSC/pull/2722#issuecomment-1450156668

 I think it is ready to to be committed.

Jakuje:
ok, I am not much happy how the flags are handled in the minidriver, but I do not know anything about the minidriver so if you believe it is working, please merge this. Its in the limbo for long enough already.

dengert:
Main problem with old code is:`#define  ROLE_EVERYONE   0` which means there is no `PIN_ID`. This PR takes this into account.

PR #2724, opened 2023-03-01
fix OpenSSL library detection if pkgconfig not found

swissbit-csteuer:
If pkg-config fails to find OpenSSL, autoconf tries to check for the presence of the RSA_version symbol in the libcrypto library. However, it turns out that RSA_version was never actually part of OpenSSL, so this check will always fail.
To fix this issue, autoconf can search for the RSA_get_version symbol instead, which should be present in OpenSSL since version 1.1.1


Jakuje (approved):

Jakuje:

PR #2729, opened 2023-03-08
Fix compiler warnings on macOS

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -605,8 +605,8 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, unsigned int pubkey_len
 				int r;
 				r = sc_asn1_sig_value_rs_to_sequence(NULL, signat,
 						signat_len, &signat_tmp, &signat_len_tmp);
-				if (r == 0) {
-					res = EVP_VerifyFinal(md_ctx, signat_tmp, signat_len_tmp, pkey);
+				if (r == 0 && signat_len_tmp < UINT_MAX) {
+					res = EVP_VerifyFinal(md_ctx, signat_tmp, (unsigned int) signat_len_tmp, pkey);
 				} else {
 					sc_log(context, "sc_asn1_sig_value_rs_to_sequence failed r:%d",r);
 					res = -1;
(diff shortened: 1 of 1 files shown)


Jakuje:
Thanks. The changes look good. But it still bothers me that these errors are hidden behind the large wall of the text and do not cause the build failing immediately. Is it something that could be changed by changing some of the macos build scripts? Also I am not sure why we do not get these in the pull requests at all.

frankmorgner:
Why we don't get this warning in all builds? I don't know... maybe this has to do with the runners we observed that are configured differently.

Can xcodebuild output be made more readably? It's not supported natively. There is something like https://github.com/xcpretty/xcpretty, but I'm not sure if it was updated to support newer xcode versions. I don't think it's worth the effort integrating this into the CI pipeline.

You could look for the keyword *error* to search for the exact problem.

Jakuje:
> Why we don't get this warning in all builds? I don't know... maybe this has to do with the runners we observed that are configured differently.

I think this would not be that systematic. Over the last two months, all master runs failed, but all PR runs (without some exceptions when there were different issues) worked. There must be something different how it is executed in master and PR. I think I was mostly copying the instructions from the travis, but I can have a second look as I was putting the pipelines together initially. I did a diff of the logs and the difference is only in the presence of the secrets and some random order of some commands during the build, which is propagated to some of the `xcodebuild` commands up until we get to [`test -e OpenSCToken -a -n '***' -a -n ***`](https://github.com/OpenSC/OpenSC/blob/26cd7f227720a560df113be1c7d620b1fb7cd98a/MacOSX/build-package.in#L177): where we build OpenSC under the OpenSCToken (only in master) and where these errors come from.

So the question is if this part can be executed in the PR without the signing keys. This landed in commit 4d6ed77a4a423929899356ec1896bcd97e40acff without any PR for discussion.

> Can xcodebuild output be made more readably? It's not supported natively. There is something like https://github.com/xcpretty/xcpretty, but I'm not sure if it was updated to support newer xcode versions. I don't think it's worth the effort integrating this into the CI pipeline.

I did not mean more readable, but to exist one one stage fails. You can see that the line `2 errors generated.` is written several lines before the end of the file. Prettify might help, using some less verbose output might help too. Just ideas.

In the latest master run there is another error:
```
/Users/runner/work/OpenSC/OpenSC/OpenSCToken/OpenSC/src/pkcs11/mechanism.c:70:42: error: implicit conversion loses integer precision: 'unsigned long' to 'unsigned int' [-Werror,-Wshorten-64-to-32]
        unsigned int new_len = data->buffer_len + in_len;
                     ~~~~~~~   ~~~~~~~~~~~~~~~~~^~~~~~~~
/Users/runner/work/OpenSC/OpenSC/OpenSCToken/OpenSC/src/pkcs11/mechanism.c:136:79: error: implicit conversion loses integer precision: 'CK_FLAGS' (aka 'unsigned long') to 'unsigned int' [-Werror,-Wshorten-64-to-32]
        if ((existing_mt = sc_pkcs11_find_mechanism(p11card, mt->mech, mt->mech_info.flags))) {
                           ~~~~~~~~~~~~~~~~~~~~~~~~                    ~~~~~~~~~~~~~~^~~~~
```
Can you have a look also into this one?

frankmorgner:
I've opened https://github.com/OpenSC/OpenSC/pull/2731 with the requested extension so that this PR will be built with XCode as well.

PR #2730, opened 2023-03-08
Fixing OSS-Fuzz issues

xhanulik:
Another batch of fixed recently reported issues.

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/profile.c
+++ b/src/pkcs15init/profile.c
@@ -1976,24 +1976,23 @@ static struct block	root_ops = {
 };
 
 static int
-check_macro_reference_loop(scconf_list *start, scconf_list *current, struct state *cur, int depth) {
+check_macro_reference_loop(char *start, char *str, struct state *cur, int depth) {
 	sc_macro_t *mac = NULL;
-	const char *str = NULL;
+	char *mac_name = NULL;
 
-	if (!start || !current || !cur)
+	if (!start || !str || !cur)
 		return 1;
 
 	if (depth == 16)
 		return 1;
 
-	str = current->data;
-	if (str[0] != '$')
+	if (!(mac_name = strchr(str, '$')))
 		return 0;
-	if (!(mac = find_macro(cur->profile, str + 1)))
+	if (!(mac = find_macro(cur->profile, mac_name + 1)))
 		return 0;
-	if (!strcmp(mac->name, start->data + 1))
+	if (!strcmp(mac->name, start + 1))
 		return 1;
-	return check_macro_reference_loop(start, mac->value, cur, depth + 1);
+	return check_macro_reference_loop(start, mac->value->data, cur, depth + 1);
 }
 
 static int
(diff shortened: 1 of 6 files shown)


src/libopensc/card-sc-hsm.c
```diff
@@ -875,7 +875,7 @@ static int sc_hsm_write_ef(sc_card_t *card,
 		}
 
 		if (buf != NULL)
-			memcpy(p, buf+offset, to_send);
+			memcpy(p, buf, to_send);
```
frankmorgner, line 878: This looks like this is a general misconception about the `idx` parameter of this function. This should refer to the **index of the file** of the card we want to write to rather than to the index of the buffer we are supplying.

@istr could please you have a look in detail whether this has any impact of the rest of the code that was added in #2682?

frankmorgner: sc_update_binary is meant to write a buffer to a card's file with a specific offset, just like `man 2 pwrite`. Hence, its second parameter which is sometimes named `offs` is meant to specify the offset of the card's file. Please note that at the end of the `while` loop in `sc_update_binary` both, `idx` and `buf` are incremented by the number of bytes written. This means, that the card drivers' `update_binary` (i.e. `sc_hsm_update_binary`) should write `buf` at the file's offset `idx`.

istr: @frankmorgner 
> This means, that the card drivers' update_binary (i.e. sc_hsm_update_binary) should write buf at the file's offset idx.

Ok, I see. Then this needs to be fixed. The two offsets into `buf` (needs to start at `0`) and into the card's file (needs to start at `idx`) need to be tracked independently in this inner loop as well. I will provide a PR for that.

@Jakuje After this has been fixed, it should be sufficient to rely on the buffer limit check that is already performed in `sc_update_binary`.

frankmorgner: OK, then it would be best to remove this change from this PR and give you the time to look into this for a proper fix. thank you.

xhanulik: Thank you for looking into this, I will remove this commit.

src/pkcs15init/profile.c
```diff
@@ -2288,6 +2291,9 @@ get_authid(struct state *cur, const char *value,
 		return get_uint(cur, value, type);
 	}
 
+	if (strlen(value) >= sizeof(temp))
+		return 1;
+
```
Jakuje, line 2296: Note, these two issues were closed on the oss-fuzz tracker yesterday. I think these are false positives as the strlcpy should handle the bounds without this explicit check (as it has the size argument). The traces show they are using some of their reimplementation of the strlcpy instead of our internal one, which might have been the culprit. 

But having explicit check should not hurt.

Jakuje (approved):

PR #2731, opened 2023-03-09
Fix compiler warnings on macOS

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -24,6 +24,7 @@
 #include <string.h>
 
 #include "sc-pkcs11.h"
+#include "common/compat_overflow.h"
 
 /* Also used for verification data */
 struct hash_signature_info {
@@ -67,7 +68,9 @@ signature_data_buffer_append(struct operation_data *data,
 	if (in_len == 0)
 		return CKR_OK;
 
-	CK_ULONG new_len = data->buffer_len + in_len;
+	CK_ULONG new_len;
+	if (__builtin_uaddl_overflow(data->buffer_len, in_len, &new_len))
+		return CKR_ARGUMENTS_BAD;
 	CK_BYTE *new_buffer = sc_mem_secure_alloc(new_len);
 	if (!new_buffer)
 		return CKR_HOST_MEMORY;
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -39,7 +39,7 @@ struct operation_data {
 	struct hash_signature_info *info;
 	sc_pkcs11_operation_t *md;
 	CK_BYTE			*buffer;
-	unsigned int	buffer_len;
+	CK_ULONG		buffer_len;
 };
 
 static struct operation_data *
@@ -67,7 +67,7 @@ signature_data_buffer_append(struct operation_data *data,
 	if (in_len == 0)
 		return CKR_OK;
 
-	unsigned int new_len = data->buffer_len + in_len;
+	CK_ULONG new_len = data->buffer_len + in_len;
 	CK_BYTE *new_buffer = sc_mem_secure_alloc(new_len);
 	if (!new_buffer)
 		return CKR_HOST_MEMORY;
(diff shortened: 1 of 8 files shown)


src/pkcs11/mechanism.c
```diff
@@ -67,8 +67,7 @@ signature_data_buffer_append(struct operation_data *data,
 	if (in_len == 0)
 		return CKR_OK;
 
-	unsigned int new_len = data->buffer_len + in_len;
-	CK_BYTE *new_buffer = sc_mem_secure_alloc(new_len);
+	CK_BYTE *new_buffer = sc_mem_secure_alloc(data->buffer_len + in_len);
```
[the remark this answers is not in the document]
frankmorgner: Down the rabbit hole you can see a weird mix of length types that propagates to several places. Here, you have a mix of `unsigned int` and `CK_ULONG` (`unsigned long int`), `sc_mem_secure_alloc` takes `size_t`. Then you have `struct operation_data` whose `buffer_len` is `unsigned int`, but gets casted to `CK_ULONG` in `sc_pkcs11_signature_final`. Then there is `sc_pkcs11_verify_data` which uses `unsigned int`s, which themselves need a cast in `sc_pkcs11_verify_final`.

This looks like a big mess... Maybe we should stick to `CK_ULONG` on the pkcs#11 layer by
- changing `struct operation_data`'s `buffer_len` to `CK_ULONG` and
- changing `sc_pkcs11_verify_data`'s types to `CK_ULONG`

This would be my first suggestion, but I'm not exactly sure. In any case, I'd like to avoid fiddeling around at small pieces of the code while the big picture is not clear..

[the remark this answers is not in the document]
Jakuje: Yeah. I think we can never avoid the casts. Even if we will sync on unsigned long in the pkcs11 layer, there will be cases where openssl or system calls will require or return `int`s, `size_t` or others and we will have to cast them back and forth. The point is that we should cast them in a sane way and copying the code back and forth will not solve this.

Any idea why these do not show up in either of the Linux builds? Is the MacOS build in master using some special build flags?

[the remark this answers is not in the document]
frankmorgner: I've updated the code to unify the pkcs#11 length/buffer types (a) and to explicitly check for an integer overflow (b). The second part may come handy for other pieces of the code as well.

Jakuje:
I pulled the commit 5fc6d9de0ce6e50632c0b1e666f32e5a5ff71667 fixing typos to master separately to get rid of the codespell errors as the osx build still looks like failing here.

frankmorgner:
Hmm, I was pushing to the wrong remote. :see_no_evil:

src/pkcs11/openssl.c
```diff
@@ -499,9 +499,10 @@ static CK_RV gostr3410_verify_data(const unsigned char *pubkey, unsigned int pub
 		}
 #endif
 		r = -1;
-		if (group)
-			octet = d2i_ASN1_OCTET_STRING(NULL, &pubkey, (long)pubkey_len);
-		if (group && octet) {
+		if (group && pubkey_len <= LONG_MAX) {
+			const unsigned char *p = pubkey;
+			octet = d2i_ASN1_OCTET_STRING(NULL, &p, (long)pubkey_len);
+		} if (group && octet) {
```
Jakuje, line 505:
```suggestion
		}
		if (group && octet) {
```

src/pkcs11/openssl.c
```diff
@@ -636,15 +637,15 @@ CK_RV sc_pkcs11_verify_data(const unsigned char *pubkey, unsigned int pubkey_len
 				int r;
 				r = sc_asn1_sig_value_rs_to_sequence(NULL, signat,
 						signat_len, &signat_tmp, &signat_len_tmp);
-				if (r == 0) {
-					res = EVP_VerifyFinal(md_ctx, signat_tmp, signat_len_tmp, pkey);
+				if (r == 0 && signat_len_tmp < UINT_MAX) {
+					res = EVP_VerifyFinal(md_ctx, signat_tmp, (unsigned int) signat_len_tmp, pkey);
 				} else {
 					sc_log(context, "sc_asn1_sig_value_rs_to_sequence failed r:%d",r);
```
Jakuje, line 643: either update the log message here to reflect the mention the overflow or create a separate error condition.

Jakuje:
Please, rebase also on top of master so we can get rid of 5d3657f76b0f0645a41bde948aa5adb6d1826995 (typos in oberthur).

src/pkcs11/openssl.c
frankmorgner: done

src/pkcs11/openssl.c
frankmorgner: done

Jakuje (approved):

PR #2732, opened 2023-03-16
Fix loop in sc_hsm_write_ef, separate offset into buf and into EF

istr:
As discovered in #2730, the inner loop needs to track the offset into the buffer and into the card's file separately.
It could be a good idea to update the signature of the functions to disambiguate the parameter name `idx`.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-sc-hsm.c
+++ b/src/libopensc/card-sc-hsm.c
@@ -782,7 +782,7 @@ static int sc_hsm_logout(sc_card_t * card)
 }
 
 
-
+/* NOTE: idx is an offset into the card's file, not into buf */
 static int sc_hsm_read_binary(sc_card_t *card,
 			       unsigned int idx, u8 *buf, size_t count,
 			       unsigned long flags)
@@ -823,7 +823,7 @@ static int sc_hsm_read_binary(sc_card_t *card,
 }
 
 
-
+/* NOTE: idx is an offset into the card's file, not into buf */
 static int sc_hsm_write_ef(sc_card_t *card,
 			       int fid,
 			       unsigned int idx, const u8 *buf, size_t count)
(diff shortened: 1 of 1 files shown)


istr:
@frankmorgner @Jakuje as discussed in #2730 here is a draft PR to address the issue with `idx`. I'd rather go with `size_t` for the offsets as well, but you might prefer to use `unsigned int` like in the signature of the function. Just let me know.

To prevent errors like these I would recommend updating the signature of these functions and/or add some comments to explain the parameters  (at least in  `card.c`).

Jakuje:
Changes look good to me. As discussed in #2731, I think the `unsigned int` is here as most of the stuff in pkcs11 is unsigned int and we wanted to avoid casts.

Not sure if we need the comments in each of the function though as this really maps to the terminology used in the iso operations, but I understand it might be confusing when there are two different offsets or indexes. And not everyone is familiar with all the corner cases of the (non-public) iso specifications ...

The most important question, does it work with your card?

istr:
@Jakuje 
> The most important question, does it work with your card?

Yes, it (still) works with the REINER SCT cyberJack RFID komfort and a SmartCard-HSM 3.5.

The failure in Codespell seems to be unrelated, so I will set this to "Ready for review" now.

istr:
Rebased to 34f6ca6ed1cdd6eeb0faac4087634467ff604a26.

istr:
The appveyor build failed due to an artifact upload problem.
```
Add new data to archive: 1 file, 6541312 bytes (6388 KiB)
Files read from disk: 1
Archive size: 58227770 bytes (56 MiB)
Everything is Ok
appveyor PushArtifact %ARTIFACT%.msi
Uploading artifact OpenSC-0.23.0_win64.msi (16,761,356 bytes)...1%Error uploading artifact to the storage: One or more errors occurred.
Command exited with code 2
```
Someone with sufficient privileges should try to restart the build, if deemed necessary.

frankmorgner:
thank you

PR #2733, opened 2023-03-17
fixed setting CKP_PUBLIC_CERTIFICATES_TOKEN

frankmorgner:
Due to a bug in checking the object bitmask, a *private* data object would be considered to be a private certificate and the CKP_PUBLIC_CERTIFICATES_TOKEN would not be set to this slot.

##### Checklist

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1538,8 +1538,8 @@ _add_public_objects(struct sc_pkcs11_slot *slot, struct pkcs15_fw_data *fw_data)
 		if (obj->p15_object->flags & SC_PKCS15_CO_FLAG_PRIVATE) {
 			/* If we found some non-accessible public object,
 			 * we can no longer claim Public Ceritificate Token conformance */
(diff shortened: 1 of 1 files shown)


frankmorgner:
This PR now sets the flag correctly for my token. However, this doesn't result in the friendlyness I was expecting. I am experimenting with having a soft-certificate loaded directly into firefox and having OpenSC loaded in parallel with a token inserted. When connecting to a server that requests a signature from the softkey, the smart card PIN is still requested (it's possible to abort the PIN prompt multiple times to proceed). Debugging in Firefox's workflow, it looks like this behavior is intended and is not related to `CKP_PUBLIC_CERTIFICATES_TOKEN` .

In https://github.com/OpenSC/OpenSC/pull/2681 it was stated, that the missing "friendlyness" was the original reason for having the onepin module. At least this should be fixed now. @metsma , could you please check whether there are still issues with multiple PINs after applying this PR? This maybe allows avoiding some of the complexity of https://github.com/OpenSC/OpenSC/pull/2681.

Jakuje (approved):

Issue #2707, opened 2023-02-14
Firefox: pthread deadlock after hardware error


frankmorgner:
### Problem Description

I'm using Yubikey 4 OTP+U2F+CCID and Firefox 109 on Ubuntu (pcscd 1.8.26, libccid 1.4.31) and after hibernation I'm sporadically observing that the card doesn't respond anymore, which causes OpenSC/Firefox to run in a deadlock. This makes the browser unusable since it doesn't react anymore.

The last line of the OpenSC log shows that it doesn't return from `sc_lock_mutex`. Debugging into the process, I observe that two threads are actually requesting a lock:
```
Thread 7 (Thread 0x7f5adab55700 (LWP 430926)):
#0  __lll_lock_wait (futex=futex@entry=0x7f5ab90613a0, private=0) at lowlevellock.c:52
#1  0x00007f5aecf31235 in __GI___pthread_mutex_lock (mutex=0x7f5ab90613a0) at ../nptl/pthread_mutex_lock.c:135
#2  0x00007f5aec9bb2f9 in PR_Lock (lock=0x7f5ab90613a0) at /build/firefox-gaF6Cx/firefox-109.0.1+build1/nsprpub/pr/src/pthreads/ptsynch.c:176
#3  0x00007f5aebeb4309 in secmodLockMutext (mutext=0x7f5ab90613a0) at /build/firefox-gaF6Cx/firefox-109.0.1+build1/security/nss/lib/pk11wrap/pk11load.c:54
#4  0x00007f5abe669942 in sc_pkcs11_lock () at pkcs11-global.c:973
#5  0x00007f5abe6687b8 in C_GetSlotInfo (slotID=0, pInfo=0x7f5adab53f28) at pkcs11-global.c:603
#6  0x00007f5aebecc02c in PK11_GetSlotInfo (slot=0x7f5abd1a4c00, info=info@entry=0x7f5adab53f28) at /build/firefox-gaF6Cx/firefox-109.0.1+build1/security/nss/lib/pk11wrap/pk11slot.c:1926
#7  0x00007f5aebe90859 in nssSlot_IsTokenPresent (slot=0x7f5ad768a030) at /build/firefox-gaF6Cx/firefox-109.0.1+build1/security/nss/lib/dev/devslot.c:175
[...]

PR #2735, opened 2023-03-23
Fix several race conditions

frankmorgner:
https://github.com/OpenSC/OpenSC/issues/2707 was haunting me for months now and I was finally able to track the problem down to a race condition in sc_lock().

On my way, I learned about valgrind's [Helgrind](https://valgrind.org/docs/manual/hg-manual.html), which may be used like this:
```sh
valgrind --tool=helgrind --error-limit=no --free-is-write=yes pkcs11-tool --test-threads ILGISL  --test-threads ILGISL --test-threads ILGISL --test-threads ILGISL
```
This still shows some problem in pkcs11-tool, which I am ignoring for now.

Unfortunately, running Firefox with helgrind is not possible, at least on my machine, which caused me a lot of headache while finally fixing the real problem. I think it would be very useful if we could extend pkcs11-tool's thread tests with the tests we already have in place by using `--test --pin=123456`. Currently `--test-threads` only calls C_Initialize (`IL`), C_GetInfo (`GI`) or C_GetSlotList (`SL`).

##### Checklist

### DIFF
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -291,27 +291,31 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
 {
 	CK_RV rv;
 #if !defined(_WIN32)
-	pid_t current_pid = getpid();
+	pid_t current_pid;
 #endif
 	int rc;
 	sc_context_param_t ctx_opts;
 
 #if !defined(_WIN32)
 	/* Handle fork() exception */
+	C_INITIALIZE_M_LOCK
+	current_pid = getpid();
 	if (current_pid != initialized_pid) {
-		if (context)
+		if (context && CKR_OK == sc_pkcs11_lock()) {
 			context->flags |= SC_CTX_FLAG_TERMINATE;
+			sc_pkcs11_unlock();
+		}
 		C_Finalize(NULL_PTR);
 	}
 	initialized_pid = current_pid;
 	in_finalize = 0;
+	C_INITIALIZE_M_UNLOCK
 #endif
 
 	/* protect from nesting */
 	C_INITIALIZE_M_LOCK
 	nesting++;
 	if (nesting > 1) {
-		sc_log(context, "C_Initialize(): Nested init detected");
 		nesting--;
 		C_INITIALIZE_M_UNLOCK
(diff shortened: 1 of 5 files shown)


frankmorgner:
@dengert , could you check whether ignoring the result of card_reader_lock_obtained in sc_lock is OK for you?

src/tools/pkcs11-tool.c
```diff
@@ -8635,6 +8632,22 @@ static void * test_threads_run(void * pttd)
 			}
 		}
 
+		else if (*pctest == 'L' && *(pctest + 1) == 'T') {
```
Jakuje, line 8621: oneline description of the LT flag would be good here in a comment. Also note about this new operation in the man page/help would be useful

frankmorgner: those options are currently not documented at all, the help message states to look at the source code... :roll_eyes:

src/pkcs11/pkcs11-global.c
```diff
@@ -291,27 +291,31 @@ CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
...
 	initialized_pid = current_pid;
 	in_finalize = 0;
+	C_INITIALIZE_M_UNLOCK
 #endif
 
 	/* protect from nesting */
 	C_INITIALIZE_M_LOCK
 	nesting++;
 	if (nesting > 1) {
-		sc_log(context, "C_Initialize(): Nested init detected");
```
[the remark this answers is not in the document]
frankmorgner: yes, it is intentional. otherwise there is a race condition writing to the log file. normally, we have mutex in place to prevent this (`sc_pkcs11_lock()`), but in case of C_Initialize this may not yet be initialized. Hence, I just removed the logging here.

src/tools/pkcs11-tool.c
Jakuje: yeah. But in this case, it is not even documented in the source code, so please add a comment similarly as with the other options

src/pkcs11/pkcs11-global.c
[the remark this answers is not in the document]
Jakuje: Thanks for clarification.

Jakuje (approved):

PR #2745, opened 2023-04-10
support the latest generation of Slovak eID

jurajsarinay:
I propose to add drivers for Slovak eID cards "version 4" (as numbered by the issuing authorities) based on IDEMIA Cosmo 9.2 operating over the contact interface.

My [skeid](https://github.com/OpenSC/OpenSC/pull/2672) driver deals with an earlier generation of the card. The current one has been issued since December 2022. The platforms are different, yet there are several similarities. I prefer to maintain separate drivers and pkcs15 emulators, would attemt to unite parts of the drivers if deemed appropriate. I am opening a draft PR "early" in the hope of help with shaping the driver(s).

The driver is modelled on the proprietary [eID Klient](https://eidas.minv.sk/download) distributed by the government.
There appear to be no pkcs15 information available on the card. There are three 3072-bit RSA keys and corresponding certificates, the operations with all the three keys appear to work well.

The driver uses a global PIN (called _BOK_) and a local signature PIN (called _Podpisový PIN_). VERIFY and INFO work, I have not been able to test any PIN changes.

There is also a PUK that has been omitted on purpose and a CAN that has been of no use so far. 

The card has a contactless interface that will likely require PACE.
```npa-tool --can``` & ```npa-tool --pin``` are known to work with the card, support for wireless operation (if at all desired) should be well within reach.

The card seems to only sign 32-byte (SHA256) hashes, that is probably one of the reasons ```pkcs11-tool --test``` fails.

##### Checklist

- [ ] macOS tokend is tested

### CODE REVIEW

**Comment by frankmorgner** [comment]
In this case, I think, it's a good idea to start a new internal card driver, because it is based on a different chip OS.

File: src/libopensc/card-skeid4.c [hunk 1]
```diff
@@ -0,0 +1,461 @@
...
+			if (pathlen == 2) {	/* only 3F00 supplied */
+				select_mf = 1;
+				break;
+			}
+			path += 2;
+			pathlen -= 2;
+		}
+		break;
+	case SC_PATH_TYPE_FROM_CURRENT:
+		apdu.p1 = 2;
```
frankmorgner, line 261: In ISO/IEC 7816, 0x02 only references a file identifier under the current DF, so this may not work with a longer path, which is why the iso7816 uses 0x09

src/libopensc/card-skeid4.c
```diff
@@ -0,0 +1,461 @@
...
+					|| pathtype == SC_PATH_TYPE_DF_NAME)
+				pathtype = SC_PATH_TYPE_FROM_CURRENT;
+		}
+	}
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0, 0);
+
+	switch (pathtype) {
+	case SC_PATH_TYPE_FILE_ID:
+		apdu.p1 = 2;
```
frankmorgner, line 241: Selecting a DF with FID may not work, because 0x02 ties this to EFs only, which is why the iso7816 driver uses 0x00

Issue #2738, opened 2023-03-31
Current master broken with CardOS 5.3 card: CKR_USER_PIN_NOT_INITIALIZED


fbezdeka:
### Problem Description

In short: pkcs11-tool --login --test fails with current master

card: CardOS 5.3

Reverting the last commit (5e06acb3f2e9 ("run update_record/read_record multiple times with offset if needed")) resolves this problem.

### Proposed Resolution

I have no proposal yet, but I can tell that reverting the mentioned commit helps.

### Steps to reproduce
git checkout master
./configure --prefix=/usr/local
make
sudo make install
/usr/local/bin/pkcs11-tool --login --test
Using slot 0 with a present token (0x0)
Logging in to "Siemens Corporate ID Card (V5)".
Please enter User PIN: 
error: PKCS11 function C_Login failed: rv = CKR_USER_PIN_NOT_INITIALIZED (0x102)
Aborting.

### Logs

This is the last part of the log. Everything else looks ok (lots of Success messages). If you need the full log, let me know.

```
    P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] apdu.c:539:sc_transmit: returning with: 0 (Success)
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] card.c:530:sc_unlock: called
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] iso7816.c:276:iso7816_read_record: returning with: 66
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] iso7816.c:195:encode_do_data: sc_asn1_encode() failed: -1401 (Invalid ASN.1 object)
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] iso7816.c:240:iso7816_read_record: Could not encode data objects: -1401 (Invalid ASN.1 object)
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] iso7816.c:276:iso7816_read_record: returning with: -1401 (Invalid ASN.1 object)
P:291296; T:0x140365321798272 12:30:00.105 [opensc-pkcs11] card.c:530:sc_unlock: called
[...]

PR #2746, opened 2023-04-11
iso7816: Fixed reading with discretionary data object

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2738

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card.c
+++ b/src/libopensc/card.c
@@ -658,14 +658,14 @@ int sc_read_binary(sc_card_t *card, unsigned int idx,
 			sc_log(card->ctx, "Subsequent read failed with %d, returning what was read successfully.", r);
 			break;
 		}
-		if ((idx > SIZE_MAX - (size_t) r)
-				|| (size_t) r > todo) {
-			/* `idx + r` or `todo - r` would overflow */
-			r = SC_ERROR_OFFSET_TOO_LARGE;
-		}
 		if (r < 0) {
 			sc_unlock(card);
 			LOG_FUNC_RETURN(card->ctx, r);
+		}
+		if ((idx > SIZE_MAX - (size_t) r) || (size_t) r > todo) {
+			/* `idx + r` or `todo - r` would overflow */
+			sc_unlock(card);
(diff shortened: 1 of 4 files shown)


src/libopensc/card.c
```diff
@@ -659,15 +659,20 @@ int sc_read_binary(sc_card_t *card, unsigned int idx,
 		r = card->ops->read_binary(card, idx, buf, chunk, flags);
 		if (r == 0 || r == SC_ERROR_FILE_END_REACHED)
 			break;
-		if ((idx > SIZE_MAX - (size_t) r)
-				|| (size_t) r > todo) {
-			/* `idx + r` or `todo - r` would overflow */
-			r = SC_ERROR_OFFSET_TOO_LARGE;
+		if (r < 0 && todo != count) {
+			/* the last command failed, but previous ones succeeded.
+			 * Let's just return what we've successfully read. */
```
Jakuje, line 664: I would propose to convert this comment into a debug log as it might be unexpected that after a read fails, this function would return success with some bytes.

src/libopensc/iso7816.c
```diff
@@ -213,6 +213,9 @@ int decode_do_data(struct sc_context *ctx,
 			sc_asn1_decode(ctx, asn1_do_data, encoded_data, encoded_data_len, NULL, NULL),
 			"sc_asn1_decode() failed");
 
+	if (!(asn1_do_data[1].flags & SC_ASN1_PRESENT))
+		return SC_ERROR_INVALID_ASN1_OBJECT;
+
```
Jakuje, line 218: From the commit message and change I do not see how this is solving the issue. You mark the field optional, but if it is not present, you fail here?

frankmorgner: When receiving the data from the card, it is mandatory, because we're expecting data. This is what is tested here. Additionally, we use the same prototype for encoding the request, where we don't include data (just the offset). That's why the ASN.1 object is *optional* in the prototype.

PR #2749, opened 2023-04-13
pkcs11-tool: accumulate all errors when testing C_GenerateRandom

frankmorgner:
fixes warning about unused variable

##### Checklist

- [ ] macOS tokend is tested


src/tools/pkcs11-tool.c
```diff
@@ -7368,48 +7368,49 @@ static int test_random(CK_SESSION_HANDLE session)
 	rv = p11->C_SeedRandom(session, seed1, 100);
 	if (rv == CKR_RANDOM_NO_RNG) {
 		printf("  RNG not available\n");
-		return 0;
+		errors++;
```
Jakuje, line 7371: When we get the error here, it means the RNG is not available and all subsequent calls should fail in the same way (if the pkcs11 module is sane) so I am not sure if it is worth trying in this case.

Jakuje (approved):

PR #2753, opened 2023-04-14
Fix OpenPGP select data

frankmorgner:
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -1780,12 +1780,17 @@ pgp_get_pubkey_pem(sc_card_t *card, unsigned int tag, u8 *buf, size_t buf_len)
  * p1: number of an instance (DO 7F21: 0x00 for AUT, 0x01 for DEC and 0x02 for SIG)
  */
 static int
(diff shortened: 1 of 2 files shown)


Jakuje (approved):
good catch. Please, remind me with the next release that I should test the Nitrokey/Gnuk token I have in my drawer. For some reason, I missed that one while testing the 0.23.0: https://github.com/OpenSC/OpenSC/wiki/Smart-Card-Release-Testing#opensc-0230 :facepalm:

ya-isakov:
Hello! I found a bug, in this change, while I was debugging related issue on Yubikey (https://github.com/OpenSC/OpenSC/pull/2969). It seems that this change is ignoring any errors from SELECT DATA, but, such errors might indicate that select failed, so, next GET_DATA will read same cert with id 0, as it would read without SELECT DATA. I found this issue in XCA - when I was importing attestation certs from my Yubikey, first (AUT) imported successfully, but when I was importing SIG, it complained, that such cert was already imported as AUT - I compared both certs, and they were the same.

PR #2754, opened 2023-04-17
Add Meson build system

frankmorgner:
- [ ] macOS tokend is tested


meson.build
```diff
@@ -0,0 +1,111 @@
...
+dl_dep = cc.find_library('dl', required : false)
+openpace_dep = dependency('libeac', required: false)
+
+cdata.set('ENABLE_READLINE', readline_dep.found())
+cdata.set('ENABLE_ZLIB', zlib_dep.found())
+cdata.set('HAVE_ZLIB_H', zlib_dep.found())
+cdata.set('ENABLE_OPENSSL', openssl_dep.found())
+cdata.set('HAVE_OPENSSL_CRYPTO_H', openssl_dep.found())
+cdata.set('ENABLE_OPENPACE', openpace_dep.found())
+cdata.set('ENABLE_SM', get_option('sm').enabled())
```
jpakkane, line 81: The type of `sm` is a boolean so this should probably be `cdata.set('ENABLE_SM', get_option('sm'))`. The `enabled` is used when the type of the option is `feature`.

meson.build
```diff
@@ -0,0 +1,111 @@
+project('opensc', 'c',
+    version: '0.23.165'
```
Jakuje, line 2: note that for the final version, the version will be best to be moved away to some separate file/script to make sure we do not have to update it on several places for a release. I have only experience with libcacard, where this is done with git-version-gen from autotools, but I believe there will be other/better ways.

https://gitlab.freedesktop.org/spice/libcacard/-/blob/master/meson.build#L2
https://gitlab.freedesktop.org/spice/libcacard/-/blob/master/build-aux/git-version-gen

src/pkcs11/meson.build
```diff
@@ -0,0 +1,27 @@
...
+)
+
+pkcs11_mod = shared_module('opensc-pkcs11',
+    pkcs11_sources,
+    include_directories: core_inc,
+    dependencies: [openssl_dep, openpace_dep],
+    install: true,
+)
+
+pkcs11_mod = shared_module('onepin-opensc-pkcs11',
```
Jakuje, line 21:
```suggestion
pkcs11_onepin_mod = shared_module('onepin-opensc-pkcs11',
```

meson.build
eli-schwartz: If it's acceptable to update to a minimum required meson >=0.57 then you can do this with:
```suggestion
    version: files('VERSION.txt')
```

and meson will read that file as a version file. For scripts, the libcacard approach seems fine.

From local testing, this build file currently supports even meson 0.49.0, so bumping the minimum aggressively may not be desired...

meson_options
eli-schwartz, line 1: File is misnamed, should end in .txt

src/tools/meson.build
```diff
@@ -0,0 +1,159 @@
...
+executable('pkcs15-crypt', 'pkcs15-crypt.c', 'util.c',
+    include_directories: core_inc,
+    link_with: [opensc_lib],
+    dependencies: [openssl_dep],
+    install: true,
+)
+
+executable('pkcs11-tool', 'pkcs11-tool.c', 'util.c',
+    include_directories: core_inc,
+    link_with: [opensc_lib, pkcs11_lib, pkcs11_mod],
```
eli-schwartz, line 35: This prints a deprecation warning because pkcs11_mod is a dlopen() style `shared_module()`.

meson.build
Jakuje: I would not be afraid to bump meson requirement to 0.57. I do not expect we will need to build it on some old systems, for example RHEL8 already has 0.58.

eli-schwartz: At any rate, permitting version 0.61.0 is sufficient to guarantee that people stuck on operating systems that only ship python 3.6 but desperately want bleeding edge OpenSC, can still manually acquire meson (by git cloning that too, or by using `pip install`) regardless of which version of meson is packaged by that distro repository.

The python requirement is a much harder one to ask people to manually compile the build system, but fortunately there are no more operating systems where python 3.5 is relevant. Even 3.6 is end of life and no longer receiving security updates by the Python Software Foundation (that's a baseline requirement for meson to drop support for an old version of python).

frankmorgner:
closed in favor of https://github.com/OpenSC/OpenSC/pull/3128

PR #2756, opened 2023-04-21
Run tests also under valgrind

### LINKED ISSUE #2736: Integrate valgrind into CI tests
### Problem Description

valgrind can exit with a non-zero exitcode for easy integration into CI scripts by using `--error-exitcode=1`. This would allow automatically detecting race conditions and memory leaks with something like this:
```
valgrind --error-exitcode=1 --leak-check=full pkcs11-tool --test --login --pin=123456
valgrind --error-exitcode=1 --tool=helgrind --error-limit=no --free-is-write=yes pkcs11-tool --use-locking --pin=123456 --test-threads LT --test-threads LT  --test-threads LT  --test-threads LT  --test-threads LT  --test-threads LT
```

### Proposed Resolution

- **Modify and integrate `runtest.sh`?** `src/tests/p11test/runtest.sh` already uses some `VALGRIND` environment variable if it is set, when running pkcs11-tool with tests. However, that script does not seem to be integrated into CI.
- **Modify `test-{isoapplet,openpgp,piv}.sh`?** Some Test scripts in `.github`  already use `pkcs11-tool -l -t -p 123456`, which could be run with valgrind as well.

@Jakuje , do you want to have a look at this?

Jakuje:
Closes #2736

~The leaks in valgrind are not yet "enforced" with `--error-exitcode=1` as it fails at several occasions that need to be fixed.~

##### Checklist

- [ ] macOS tokend is tested


containers/opensc-test-piv/Containerfile
```diff
@@ -11,3 +11,4 @@ WORKDIR /src
 RUN /src/.github/setup-linux.sh piv
 RUN /src/.github/build.sh
 CMD /src/.github/test-piv.sh
+CMD /src/.github/test-piv.sh valgrind
```
frankmorgner, line 14: I'm not an expert, but I think only exactly one CMD is executed, i.e. the last CMD will overwrite the previous one

frankmorgner (changes requested):

containers/opensc-test-piv/Containerfile
Jakuje: Correct. In other cases, I commented out the first one as this is used only for local testing to avoid wasting GH actions minutes. Will fix on the next rebase.

Jakuje:
This is getting to be ready for reviews. Now, valgrind is executed over the emulator tests as well as `make check` unit tests or shell tests, which surfaced several issues including reproducing the issue reported in #2775.

@frankmorgner can you have a look?

@swissbit-csteuer Can you have a look that the fixes for the isoapplet make sense?

I plan to remove the suppression file as there is nothing to suppress now (when we got rid of the pcsclite issues and issues brought in by glib/gobject/notifications), there will likely be some more issues from the isoapplet/openpgp, but I think this should be close to final version.

src/libopensc/card-isoApplet.c
```diff
@@ -1218,8 +1218,13 @@ isoApplet_compute_signature(struct sc_card *card,
...
 		size_t tmplen = sizeof(tmp);
-		sc_pkcs1_strip_digest_info_prefix(NULL, data, datalen, tmp, &tmplen);
-		r = iso_ops->compute_signature(card, tmp, tmplen, out, outlen);
+		r = sc_pkcs1_strip_digest_info_prefix(NULL, data, datalen, tmp, &tmplen);
+		if (r == SC_SUCCESS) {
+			r = iso_ops->compute_signature(card, tmp, tmplen, out, outlen);
+		} else {
+			/* No digest info present? Use the value as it is */
+			r = iso_ops->compute_signature(card, data, datalen, out, outlen);
+		}
```
swissbit-csteuer, line 1227: I think its better to return an error if the digest info is missing.

Jakuje: There is no digestinfo in case the `RSA_PKCS_PSS` mechanism is used (as visible in the modified p11test results). If you use the `SHA*_RSA_PKCS_PSS` mechanism, the input is either hashed inside of the card or in OpenSC, depending on the card capabilities, but with `RSA_PKCS_PSS`, the OpenSC gets the digest only with the length of `hLen` so there is no digestinfo involved (the digest is part of the PSS_PARAMS structure though):
https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061139

At least that is how I covered this mechanism in the p11test and how it works with other cards so far. If you reading of the specs is different, feel free to correct me.

src/pkcs15init/pkcs15-isoApplet.c
```diff
@@ -548,6 +548,7 @@ isoApplet_generate_key_ec(const sc_pkcs15_prkey_info_t *key_info, sc_card_t *car
 	memcpy(pubkey->u.ec.ecpointQ.value, args.pubkey.ec.ecPointQ.value, args.pubkey.ec.ecPointQ.len);
 
 	/* The OID is also written to the pubkey->u.ec.params */
+	free(pubkey->u.ec.params.der.value);
```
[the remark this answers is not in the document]
Jakuje: It would help if this would be a bit more described. In the past, the EC params were sometimes very wildly assigned to various structures without much care. I hope this has improved a bit over last year with the fuzzing and with this PR.

For this case, the `pubkey` comes from `sc_pkcs15init_generate_key()` function, where it is memset to zeroes and then the EC params are copied to that structure from the keygen params using `sc_copy_ec_params()` so it should not really be anything dangerous (unless somebody starts assigning here some dangerous values on the way).

src/tests/p11test/p11test_case_common.c
```diff
@@ -413,9 +413,11 @@ int callback_public_keys(test_certs_t *objects,
 			RSA *rsa = RSA_new();
 			if (RSA_set0_key(rsa, n, e, NULL) != 1 ||
 			    EVP_PKEY_set1_RSA(o->key, rsa) != 1) {
+				RSA_free(rsa);
 				fail_msg("Unable to set key params");
 				return -1;
 			}
+			RSA_free(rsa);
```
xhanulik, line 420: When `EVP_PKEY_set1_RSA` calls internally `EVP_PKEY_assign_RSA`, I think it should not be freed here (but I may not understand the [documentation](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_assign_RSA.html) correctly).

Jakuje: The `*set1*`/`*get1*` functions set/get values by incrementing/decrementing their internal reference counter (in comparison to the `*set0*`/`*get0*` functions which hand over the ownership. Unfortunately documentation is not much helpful here, but source code helps noticing the difference between these two:
https://github.com/openssl/openssl/blob/b77826877be3bdd56e3e86887cb78ea010db90be/crypto/evp/p_legacy.c#L27-L31
Indeed, using `EVP_PKEY_assign_RSA()` directly might be simpler as it will help us avoid this free, but then I would have to break the condition to capture `RSA_set0_key()` failure separately (and free the RSA) and not free it in case of success of the `EVP_PKEY_set1_RSA()`. But I see that the EVP and RSA allocations are not checked so I will rewrite it anyway.

src/libopensc/card-openpgp.c
```diff
@@ -2837,10 +2845,14 @@ pgp_update_pubkey_blob(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t *key_in
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
 
 	r = sc_pkcs15_encode_pubkey(card->ctx, &p15pubkey, &data, &len);
-	LOG_TEST_RET(card->ctx, r, "Cannot encode pubkey");
+	if (r != 0) {
+		free(data);
```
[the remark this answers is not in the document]
Jakuje: Correct. On error, this should not allocate the buffer in the first place. I will drop this part. Thanks!

src/libopensc/card-openpgp.c
```diff
@@ -2837,10 +2845,14 @@ pgp_update_pubkey_blob(sc_card_t *card, sc_cardctl_openpgp_keygen_info_t *key_in
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
 
 	r = sc_pkcs15_encode_pubkey(card->ctx, &p15pubkey, &data, &len);
-	LOG_TEST_RET(card->ctx, r, "Cannot encode pubkey");
+	if (r != 0) {
```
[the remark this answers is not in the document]
Jakuje: removed this part

src/tests/p11test/p11test_case_common.c
xhanulik: Alright, thanks for clarification.

src/pkcs15init/pkcs15-isoApplet.c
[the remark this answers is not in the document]
Jakuje: Should I drop some of the patches/changes from the PR to get it merged (without these changes, the CI will most likely fail)? Or do you mean that we should just create a follow-up issue to review how the ec->params are handled in various cases?

src/libopensc/card-isoApplet.c
swissbit-csteuer: I'm reading the spec the same way, but I still do not understand how this has worked before where the return code of `sc_pkcs1_strip_digest_info_prefix` was not checked. Shouldn't the `RSA_PKCS_PSS` test cases in p11test have failed before that change?

Jakuje: I think they were failing, but given the vast amount of the combination for the RSA-PSS and that really no driver supports everything, the failures are silently ignored ...

swissbit-csteuer: Understood. Forwarding the data as-is to the card if no digest is present is the correct way then :+1:

swissbit-csteuer (approved):

Jakuje:
@frankmorgner can you clarify what you mean with the "good stuff" so we can move on with this? I see approvals but also some concerns in your last comment. Or should I merge it and open a follow-up issue to review the EC parameters ownership throughout the code?

frankmorgner:
> merge it and open a follow-up issue to review the EC parameters ownership throughout the code

Yes, that's what I wanted to say. thank you!

Jakuje:
Thank you for the reviews! Merging. The new issue is #2790

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1936,6 +1923,15 @@ sc_pkcs15init_store_public_key(struct sc_pkcs15_card *p15card, struct sc_profile
 			key_info->params.len = key.u.ec.params.der.len;
 			memcpy(key_info->params.data, key.u.ec.params.der.value, key.u.ec.params.der.len);
 		}
+		if (keyargs->key.u.ec.ecpointQ.value) {
+			key.u.ec.ecpointQ.value = malloc(keyargs->key.u.ec.ecpointQ.len);
+			if (!key.u.ec.ecpointQ.value) {
+				r = SC_ERROR_OUT_OF_MEMORY;
+				LOG_TEST_GOTO_ERR(ctx, r, "Cannot allocate EC params");
+			}
+			key.u.ec.ecpointQ.len = key.u.ec.ecpointQ.len;
```
frankmorgner, line 1932: CID 385152:  Incorrect expression  (NO_EFFECT)
    Assigning "key.u.ec.ecpointQ.len" to itself has no effect.

Issue #2725, opened 2023-03-06
NSS does not like colons in token label


Jakuje:
### Problem Description

I was made aware of the following NSS bug https://bugzilla.mozilla.org/show_bug.cgi?id=1817553 with new CAC cards containing colons in CN of certificates, which is being propagated to the token labels.

From there the NSS composes the certificate "nicknames" in format of "token label":"certificate label" and depends on the "token label" not having any colons in it.

### Proposed Resolution
We are discussing some workarounds in NSS, but other option might be to provide a bit more flexibility using configuration options (on NSS quirks similar to #2681) to either replace colons from CN or use more generic "Common Access Card" or something.

### Steps to reproduce
* Have a card with token name containing colon
* Try to add it and use it from Firefox/Thunderbird or any NSS application

### Logs

no logs available

mouse07410:
Replacing colons with e.g. "_" seems the best. Using more generic "CAC" instead seems acceptable, though not as good.

PR #2760, opened 2023-04-24
pkcs11: Do not use colons in the token info label

Jakuje:
The NSS can not handle tokens with colons in token info labels because it is using colon as a separator for token info label and object label.

Fixes: #2725

More information available in the following NSS bug:

https://bugzilla.mozilla.org/show_bug.cgi?id=1817553

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1176,7 +1176,6 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 			size_t pin_len = 0;
 			if (auth->label[0] && strncmp(auth->label, "PIN", 4) != 0)
 				pin_len = strlen(auth->label);
-
 			if (pin_len) {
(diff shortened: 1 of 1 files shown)


src/pkcs11/framework-pkcs15.c
```diff
@@ -1208,6 +1207,13 @@ pkcs15_init_slot(struct sc_pkcs15_card *p15card, struct sc_pkcs11_slot *slot,
 						p15card->tokeninfo ? p15card->tokeninfo->label : "",
 						32);
 			}
+			/* Some applications (NSS) do not like the colons in the
+			 * TOKEN_INFO label so replace them here */
+			for (int i = 0; i < 32; i++) {
+				if (slot->token_info.label[i] == ':') {
+					slot->token_info.label[i] = '_';
```
frankmorgner, line 1214: I think underscores are very unpleasant in user messages. Could we maybe agree on replacing that just with a space (my preferred option) or with `.`?

frankmorgner (changes requested):

PR #2765, opened 2023-04-28
Fix pkcs11-tool unwrap / incorrect CKA_ID

popovec:
"object_id[]" and "id_len" must be allocated so that it is not deallocated or overwritten (on the stack) at the time of the C_UnwrapKey() call.


frankmorgner (approved):

Jakuje (approved):
Thanks! This should be something we should be able to catch with  #2756 when we will get rid of all the false positives.

PR #2766, opened 2023-04-28
epass2003: use true random number for mutual authentication for SM.

popovec:

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -39,9 +39,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include <openssl/evp.h>
-#include <openssl/sha.h>
 #include <openssl/cmac.h>
+#include <openssl/rand.h>
+#include <openssl/sha.h>
 
 #include "internal.h"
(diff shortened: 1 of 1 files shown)


src/libopensc/card-epass2003.c
```diff
@@ -39,9 +39,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <openssl/cmac.h>
```
frankmorgner, line 42: this include does not seem to be needed

popovec: We need this #include <openssl/cmac.h> for function aes128_encrypt_cmac() - epass2003 uses CMAC-AES-128 for SM in FIPS certified mode. (I changed the order of includes in this file, as git-clang-format had a problem with another order.)

popovec: Ops, incorrect "include"... The PR is already fixed, modified, the unnecessary #include <openssl/evp.h>" is removed.

frankmorgner (approved):

Jakuje (approved):
Good catch!

PR #2768, opened 2023-05-01
Add support for --private flag when writing certificates

### LINKED ISSUE #2767: --private flag is not available when writing certificates using --write-object in pkcs11-tool
### Problem Description

I used pkcs11-tool to write a certificate to YubiHSM 2 but am having problems with the write failing because CKA_PRIVATE is set to false.

https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L4042

The --private flag is available as an option to set CKA_PRIVATE to true, but it seems to only work for writing [CKO_PUBLIC_KEY](https://github.com/OpenSC/OpenSC/blob/e58bca0a33e1e972728902566ee2641c5f1430d0/src/tools/pkcs11-tool.c#L4196-L4199), [CKO_SECRET_KEY](https://github.com/OpenSC/OpenSC/blob/e58bca0a33e1e972728902566ee2641c5f1430d0/src/tools/pkcs11-tool.c#L4292-L4295) and [CKO_DATA](https://github.com/OpenSC/OpenSC/blob/e58bca0a33e1e972728902566ee2641c5f1430d0/src/tools/pkcs11-tool.c#L4356-L4359).

Referring to the [PKCS#11 specification](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.pdf), it seems correct that CKA_PRIVATE is a common attribute and can therefore also be set on certificate objects.

When writing a CKO_CERTIFICATE object, CKA_PRIVATE seems to be forced to be set to false, does anyone know the background behind such an implementation?

Would it also be acceptable to allow the value of CKA_PRIVATE, which is set when the certificate is written, to be changed by a flag like any other object?

### Proposed Resolution
When writing certificates, switch CKA_PRIVATE true/false according to the value of opt_is_private

### Logs

https://gist.github.com/misodengaku/dd84e59d1403b2b01613860ec4afd3a0

[Here](https://github.com/Yubico/yubihsm-shell/blob/master/pkcs11/yubihsm_pkcs11.c#L1684-L1692) is a link to the line causing the error in yubihsm_pkcs11.c.

misodengaku:
fixes #2767 

##### Checklist

- [x] tested with YubiHSM 2

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -4039,7 +4039,11 @@ static int write_object(CK_SESSION_HANDLE session)
 		FILL_ATTR(cert_templ[1], CKA_VALUE, contents, contents_len);
 		FILL_ATTR(cert_templ[2], CKA_CLASS, &clazz, sizeof(clazz));
 		FILL_ATTR(cert_templ[3], CKA_CERTIFICATE_TYPE, &cert_type, sizeof(cert_type));
-		FILL_ATTR(cert_templ[4], CKA_PRIVATE, &_false, sizeof(_false));
+		if (opt_is_private == 1) {
+			FILL_ATTR(cert_templ[4], CKA_PRIVATE, &_true, sizeof(_true));
+		} else {
+			FILL_ATTR(cert_templ[4], CKA_PRIVATE, &_false, sizeof(_false));
+		}
 		n_cert_attr = 5;
 
 		if (opt_object_label != NULL) {


dengert (approved):

frankmorgner (approved):

Jakuje (approved):

Jakuje:
Thank you!

PR #2770, opened 2023-05-03
cardos-tool: correctly identify CardOS v5.4 cards

hamarituc:
Card name:

```
$ opensc-tool -n
Using reader with a card: Alcor Micro AU9540 00 00
Atos CardOS
```

##### Checklist

Simple fix, Nothing applies.

### DIFF
--- a/src/tools/cardos-tool.c
+++ b/src/tools/cardos-tool.c
@@ -198,6 +198,8 @@ static int cardos_info(void)
 	} else if (apdu.resp[0] == 0xc9 &&
 			(apdu.resp[1] == 0x02 || apdu.resp[1] == 0x03)) {
 		printf(" (that's CardOS V5.3)\n");
+	} else if (apdu.resp[0] == 0xc9 && apdu.resp[1] == 0x04) {
+		printf(" (that's CardOS V5.4)\n");
 	} else {
 		printf(" (unknown Version)\n");
 	}


Jakuje (approved):

frankmorgner:
So CardOS V5.5, 5.1 and 5.2 use the same scheme?

hamarituc:
> So CardOS V5.5, 5.1 and 5.2 use the same scheme?

What do you mean by "scheme"?

This PR just addresses an identification issue. I got a D-Trust Signature Card 4.1 and noticed it uses CardOS 5.4, which was not recognized by `cardos-tool`.

```
$ cardos-tool -i
Using reader with a card: Alcor Micro AU9540 00 00
3b:d2:18:00:81:31:fe:58:c9:04:11
Info : CardOS V5.4     2019
Serial number: 0a 00 33 06 00 0f 3f 51
OS Version: 201.4 (unknown Version)
Current life cycle: 16 (operational)
Security Status of current DF:
Free memory : 1402
ATR Status: 0x0 ROM-ATR
Packages installed:
E1 0B 53 06 03 04 13 02 C9 04 8F 01 01 ..S..........
Ram size: 7, Eeprom size: 118, cpu type: 78, chip config: 63, chip manufacturer: 5
Free eeprom memory: 69120
Current Maximum Data Field Length: 437
Complete chip production data:
CC 78 33 13 03 00 01 00 98 00 00 06 0C 0A 00 00 .x3.............
00 00 00 00 00 00 00 61 75 38 37 67 FF FF FF 78 .......au87g...x
01 51 82 78 05 1E 07 00 00 83 3C 05 E7 55 2D 86 .Q.x......<..U-.
01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ................
00 00 00 00 00 00 00 00 00 00 00 00             ............
System keys: PackageLoadKey (version 0x00, retries 10)
System keys: StartKey (version 0xff, retries 10)
Path to current DF:
01 01 ..
```

I cannot judge whether or not CardOS 5.4 behaves like CardOS 5.3. At the moment I wasn't able to create a signature with this Card using neither the OpenSC PKCS#11 module nor GnuPG. So it seems the card behaves different.

I even cannot say anything about CardOS 5.5. According to https://www.youcard.de/betriebssystem-cardos-v5-4-und-v5-5-vorgestellt/ it seems they differ only in certification.

frankmorgner:
Indeed, I was only referring to the identification process in cardos-tool.

hamarituc:
> Indeed, I was only referring to the identification process in cardos-tool.

I don't have an older Card to compare, but I assume they use the same scheme for identification.

Jakuje:
@frankmorgner good to merge? This is trivial identification change.

If we would like to discuss the functionality of the new CardOS cards, I think this should be handled in separate issue/PR. If you will have a changeset or some debug logs you would like to help with, please open a new PR/Issue.

hamarituc:
> @frankmorgner good to merge? This is trivial identification change.
> 
> If we would like to discuss the functionality of the new CardOS cards, I think this should be handled in separate issue/PR. If you will have a changeset or some debug logs you would like to help with, please open a new PR/Issue.

See #2784 for the issue about the card functionality.

frankmorgner:
Thank you

PR #2772, opened 2023-05-05
epass2003: SM checking (MAC and SW)

popovec:
This patch introduces basic verification of data coming from the token in the secure messaging mode:

1. compare raw SW and SW encoded in SM
2. verify MAC

MAC verification is working only for DES and AES-CBC-MAC.  Code for checking MAC in FIPS mode is still missing.

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -1330,6 +1330,109 @@ epass2003_sm_wrap_apdu(struct sc_card *card, struct sc_apdu *plain, struct sc_ap
 	return SC_SUCCESS;
 }
 
+static int
+epass2003_check_response_mac_and_sw(struct sc_card *card, struct sc_apdu *sm)
+{
+	unsigned char iv[16];
+	unsigned char *data = NULL, *mac = NULL;
+	size_t blocksize, mac_len;
+	int ret = -1;
+	size_t taglen;
+	const u8 *tag;
+	epass2003_exdata *exdata;
+	unsigned char *in = sm->resp;
+	unsigned char *alt_in;
+	size_t inlen = sm->resplen;
+	size_t len_correction;
+
+	/* card/ctx/drv_data is already checked by caller */
+	exdata = (epass2003_exdata *)card->drv_data;
+
+	/* The SM must contain at least TLV encoded SW and MAC fields. */
+	if (inlen < 14 )
+		return ret;
+
+	/* compare BER-TLV encoded SW (TAG 0x99) and raw SW */
+	alt_in = in;
+	tag = sc_asn1_find_tag(card->ctx, alt_in, inlen, 0x99, &taglen);
+	if (tag == NULL || taglen != 2) {
+		/*
+		 * It seems that the EPASS2003 firmware has some problem with BER-TLV encoding.
+		 * Instead of (correct) TLV 87 81 81 [01 .. ..] incorrect TLV 87 81 [01 .. ..]
+		 * is returned. There seems to be some proprietary fix for the faulty encoding
+		 * in the decrypt_response() function, similar fix here:
+		 */
+		if (0x01 == in[2] && 0x82 != in[1]) {
+			sc_log(card->ctx, "Workaround, wrong BER-TLV ?");
+			len_correction = in[1] + 2;
+			if (inlen < len_correction)
+				return ret;
+			inlen -= len_correction;
+			alt_in += len_correction;
+			tag = sc_asn1_find_tag(card->ctx, alt_in, inlen, 0x99, &taglen);
+			if (tag == NULL || taglen != 2)
+				return ret;
+		} else
+			return ret;
+	}
+	if (sm->sw1 != tag[0] || sm->sw2 != tag[1])
+		return ret;
+
+	/* no documentation/real hardware to test, the response is accepted without MAC check */
+	if (exdata->bFipsCertification) {
+		sc_log(card->ctx, "Warning, MAC is not checked");
+		return 0;
+	}
+	tag = sc_asn1_find_tag(card->ctx, alt_in, inlen, 0x8e, &taglen);
+	if (tag == NULL || taglen != 8)
+		return ret;
+
+	if (KEY_TYPE_AES == exdata->smtype)
+		blocksize = 16;
+	else
+		blocksize = 8;
+
+	mac_len = tag - in - 2;
+	if (NULL == (data = calloc(1, mac_len + blocksize)))
+		goto end;
+	if (NULL == (mac = malloc(mac_len + blocksize)))
+		goto end;
+
+	/* copy response to buffer and append padding */
+	memcpy(data, in, mac_len);
+	data[mac_len++] = 0x80;
+
+	if (mac_len % blocksize)
+		mac_len += (blocksize - (mac_len % blocksize));
+
+	/* calculate MAC */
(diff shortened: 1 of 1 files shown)


popovec:
Just for completeness: AES-CBC-MAC was tested on a real epass2003 token.  I don't have a real token using DES, so the code for DES was written according to the construct_mac_tlv() function.  This code was verified only in the simulation -
https://github.com/popovec/epass2003_sim

src/libopensc/card-epass2003.c
```diff
@@ -1330,6 +1330,100 @@ epass2003_sm_wrap_apdu(struct sc_card *card, struct sc_apdu *plain, struct sc_ap
...
+			if (tag == NULL || taglen != 2)
+				return ret;
+		} else
+			return ret;
+	}
+	if (sm->sw1 != tag[0] || sm->sw2 != tag[1])
+		return ret;
+
+	if (exdata->bFipsCertification) {
+		sc_log(card->ctx, "Warning, MAC is not checked");
```
frankmorgner, line 1373: So FIPS forbids authenticated responses for cards!?

src/libopensc/card-epass2003.c
```diff
@@ -1330,6 +1330,100 @@ epass2003_sm_wrap_apdu(struct sc_card *card, struct sc_apdu *plain, struct sc_ap
...
+
+	/* compare BER-TLV encoded SW (TAG 0x99) and raw SW */
+	alt_in = in;
+	tag = sc_asn1_find_tag(card->ctx, alt_in, inlen, 0x99, &taglen);
+	if (tag == NULL || taglen != 2) {
+		/*
+		 * It seems that the EPASS2003 firmware has some problem with BER-TLV encoding.
+		 * There seems to be some proprietary fix for the faulty encoding in the
+		 * decrypt_response() function.  Instead of (correct) TLV 87 81 81 [01 ...]
+		 * incorrect TLV 87 81 [01 ...  ] is returned.
```
frankmorgner, line 1358: I am confused... both cases are valid BER-TLV structures...
- 87 81 81 [01 ...] is Tag 87, length 129, data starting with 01
- 87 81 [01 ... ] is tag 87, length 1, data is 01

Also, the `if` clause below doesn't seem to be be a workaround for the problem you're describing in the comment.

src/libopensc/card-epass2003.c
```diff
@@ -1330,6 +1330,100 @@ epass2003_sm_wrap_apdu(struct sc_card *card, struct sc_apdu *plain, struct sc_ap
...
+		 */
+		if (0x01 == in[2] && 0x82 != in[1]) {
```
frankmorgner, line 1360: when reaching this code, `inlen` is not checked and taglen is different from 2. This means that you could cause an access violation accessing the buffer with those offsets.

src/libopensc/card-epass2003.c
```diff
@@ -1330,6 +1330,100 @@ epass2003_sm_wrap_apdu(struct sc_card *card, struct sc_apdu *plain, struct sc_ap
...
+
+	mac_len = tag - in - 2;
+	if (NULL == (data = malloc(mac_len + blocksize)))
+		goto end;
+	if (NULL == (mac = malloc(mac_len + blocksize)))
+		goto end;
+
+	/* copy response to buffer and append padding */
+	memcpy(data, in, mac_len);
+	memset(data + mac_len, 0, blocksize);
```
frankmorgner, line 1393: just use calloc instead of malloc above and you don't need to memset here

frankmorgner (changes requested):

src/libopensc/card-epass2003.c
popovec: Currently, the SM message is not checked at all - the driver also accepts the SM message with an invalid MAC. After applying this PR, the check will be performed for the MAC generated using DES_CBC_MAC or AES_CBC_MAC. If the token uses AES_CMAC (FIPS certified mode), any SM is accepted as before (without MAC check).   I will write the code that uses AES CMAC later (as a separate PR), it is necessary to solve several workarounds where apparently IV is not used in the MAC generation for some APDUs.

src/libopensc/card-epass2003.c
popovec: The error appears only and only for the case where the "value" field starts with byte 0x01. (0x01 - padding indicator, and and it is followed by 128 bytes of the answer, therefore the correct value of "L" has the value 129 (encoded 0x81 0x81). Here example (this returns my real epass2003 token):
`````
Incoming APDU (147 bytes):                                                                                                                                        
87 81 01 EA 4E 99 AD 5C 7F 6C 22 23 07 99 7A D2 ....N..\.l"#..z.  
5C 7E 2E 34 5F 13 9B 7A CD C3 86 4E BB D2 7C C9 \~.4_..z...N..|.                                                                                                  
05 9F CF 58 D6 2E 01 F2 60 30 3E F2 0D 4B 89 68 ...X....`0>..K.h  
FD 91 DD 3E 0F D0 EB 4C 0E C0 94 59 E5 2E 7A B6 ...>...L...Y..z.                                                                                                  
E5 15 6F C7 F9 44 0D D4 23 84 CF 37 62 AD 27 6E ..o..D..#..7b.'n  
5A 26 04 82 DD 66 C9 0D 36 F9 37 AF 0C 71 BA 19 Z&...f..6.7..q..                                                                                                  
91 54 60 85 F3 A4 98 2A 78 CF A8 68 3B 6F A0 15 .T`....*x..h;o..  
B0 C0 8A 94 08 D2 23 DD 24 6F FD E0 9C 2C 80 8F ......#.$o...,..                                                                                                  
15 5F 45 99 02 90 00 8E 08 C9 9E 68 58 CE 10 EF ._E........hX...  
A8 90 00                                        ...                                                                                                               
`````

I don't fully understand the problem of why epass2003 returns data incorrectly like this, but the author of the code probably knew why it should be checked this way:

`00b069a08 (Viktor Tarasov       2012-06-04 11:52:50 +0200 1458)         if (0x01 == in[2] && 0x82 != in[1]) {`

src/libopensc/card-epass2003.c
popovec: Thanks, it really is necessary to check `inlen' here.

src/libopensc/card-epass2003.c
popovec: It would be best to use `alloca` (let's save `free` maintenance). However, I did not find `alloc' anywhere in the OpenSC code, I understand that there may be a problem with portability.

frankmorgner: alloca is not supported on Windows, afaik. Also, I don't like it personally because you can't do any error handling (a failure in memory allocation with alloca is just undefined behavior).

src/libopensc/card-epass2003.c
frankmorgner: OK, then lets postpone this for later.

src/libopensc/card-epass2003.c
dengert: >  both cases are valid BER-TLV structures...

Both can not be correct BER. It could be BER TLV vs SIMPLE-TLV. The comments indicate it should be BER but in error.

popovec:
Rebased. Access outside the input buffer fixed by checking the `inlen` variable. `malloc + memset` replaced by `calloc`. Thanks again for the review.

frankmorgner (approved):

PR #2779, opened 2023-05-10
Recent OSS-Fuzz issues

xhanulik:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-oberthur.c
+++ b/src/libopensc/card-oberthur.c
@@ -554,12 +554,15 @@ auth_select_file(struct sc_card *card, const struct sc_path *in_path,
 			}
 		}
 		else if (path.len - offs == 0 && file_out)  {
-			if (sc_compare_path(&path, &auth_current_df->path))
+			if (sc_compare_path(&path, &auth_current_df->path) && file_out) {
+				sc_file_free(*file_out);
 				sc_file_dup(file_out, auth_current_df);
-			else  if (auth_current_ef)
+			} else  if (auth_current_ef && file_out) {
+				sc_file_free(*file_out);
 				sc_file_dup(file_out, auth_current_ef);
-			else
+			} else {
 				LOG_TEST_RET(card->ctx, SC_ERROR_INTERNAL, "No current EF");
+			}
 		}
 	}
 
--- a/src/libopensc/card-rtecp.c
+++ b/src/libopensc/card-rtecp.c
@@ -527,15 +527,19 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
 	rsf_length[0] = (newlen >> 8) & 0xFF;
 	rsf_length[1] = newlen & 0xFF;
 
-	if (buf_end - p < (int)(2 + sizeof(rsf_length)))
+	if (buf_end - p < (int)(2 + sizeof(rsf_length))) {
+		free(buf);
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
+	}
 
 	sc_asn1_put_tag(0x80, rsf_length, sizeof(rsf_length), p, buf_end - p, &p);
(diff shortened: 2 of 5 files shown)


src/libopensc/card-oberthur.c
```diff
@@ -548,7 +548,8 @@ auth_select_file(struct sc_card *card, const struct sc_path *in_path,
 
 			for (ii=0; ii < path.len - offs; ii+=2)   {
 				memcpy(tmp_path.value, path.value + offs + ii, 2);
-
+				if (file_out)
+					sc_file_free(*file_out);
```
frankmorgner, line 552: I think this should be the more appropriate change to fix this problem:
```diff
diff --git a/src/libopensc/card-oberthur.c b/src/libopensc/card-oberthur.c
index cf39fec5f..4080df97f 100644
--- a/src/libopensc/card-oberthur.c
+++ b/src/libopensc/card-oberthur.c
@@ -554,11 +554,13 @@ auth_select_file(struct sc_card *card, const struct sc_path *in_path,
                        }
                }
                else if (path.len - offs == 0 && file_out)  {
-                       if (sc_compare_path(&path, &auth_current_df->path))
+                       if (sc_compare_path(&path, &auth_current_df->path) && file_out) {
+                               sc_file_free(*file_out);
                                sc_file_dup(file_out, auth_current_df);
-                       else  if (auth_current_ef)
+                       } else  if (auth_current_ef && file_out) {
+                               sc_file_free(*file_out);
                                sc_file_dup(file_out, auth_current_ef);
-                       else
+                       } else
                                LOG_TEST_RET(card->ctx, SC_ERROR_INTERNAL, "No current EF");
                }
        }
```

src/libopensc/card-rtecp.c
```diff
@@ -217,9 +224,12 @@ static int set_sec_attr_from_acl(sc_card_t *card, sc_file_t *file)
...
 
-	assert(card && card->ctx && file);
-	assert(!file->sec_attr  &&  file->sec_attr_len == 0);
-	assert(1 + 6 < sizeof(sec_attr));
+	if (!card || !card->ctx || !file
+		|| file->sec_attr || file->sec_attr_len != 0)
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_INVALID_ARGUMENTS);
+
+	if (1 + 6 >= sizeof(sec_attr))
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_INTERNAL);
```
frankmorgner, line 232: I think this check can be removed, because SC_RTECP_SEC_ATTR_SIZE is 15

frankmorgner (changes requested):
There is some general problem: When checking input parameters, you may use card->ctx `if (card && card->ctx)`

src/libopensc/card-rtecp.c
```diff
@@ -515,20 +526,25 @@ static int rtecp_change_reference_data(sc_card_t *card, unsigned int type,
 	/* put 0x80 TLV */
 	rsf_length[0] = (newlen >> 8) & 0xFF;
 	rsf_length[1] = newlen & 0xFF;
-	assert(buf_end - p >= (int)(2 + sizeof(rsf_length)));
+
+	if (buf_end - p < (int)(2 + sizeof(rsf_length)))
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
```
Jakuje, line 531: `free(buf)` on error here and in the following `returns`. Sorry, missed this on the first round.

xhanulik: Thanks for catching.

Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you. Merging!

PR #2780, opened 2023-05-11
pkcs11-tool: add cipher test mode

dlegaultbbry:
Add support for non-AEAD ciphers to the tool test
mode to assert that Encrypt/Decrypt APIs work correctly using established test vectors.

Tested using a softhsm supporting the algorithms on a QNX target

```
# pkcs11-tool --module=/system/lib/dll/pkcs11-module.so --test
  seems to be OK
Digests:
  SHA256: OK
Ciphers:
  AES-ECB: OK
  AES-CBC: OK
Signatures: not implemented
Verify: not implemented
No errors

```

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6295,7 +6295,7 @@ static int test_cipher(CK_SESSION_HANDLE session)
 
 		for (CK_ULONG ctlen = 0; ctlen < cipher_algs[i].ctsz;) {
 
-			CK_ULONG isize = min(cipher_algs[i].ctsz - ctlen, CIPHER_CHUNK);
+			CK_ULONG isize = MIN(cipher_algs[i].ctsz - ctlen, CIPHER_CHUNK);
 			CK_ULONG osize = sizeof(ptext1) - poff;
 
 			fct = "C_DecryptUpdate";
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -99,6 +99,8 @@ extern CK_FUNCTION_LIST_3_0 pkcs11_function_list_3_0;
 #define NEED_SESSION_RO	0x01
 #define NEED_SESSION_RW	0x02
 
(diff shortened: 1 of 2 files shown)


frankmorgner:
I know we have some problems with the CI pipeline, but please look at the individual messages, e.g.
https://github.com/OpenSC/OpenSC/actions/runs/4947970899/jobs/8848073478?pr=2780

dlegaultbbry:
I'm a bit confused. Looks like the pipeline is not using the latest code or it expects things to be squashed?

frankmorgner:
you may rebase onto https://github.com/OpenSC/OpenSC/pull/2781, which fixes the CI problem

src/tools/pkcs11-tool.c
```diff
@@ -96,9 +96,15 @@ extern CK_FUNCTION_LIST_3_0 pkcs11_function_list_3_0;
 #define MAX_TEST_THREADS 10
 #endif
 
+#ifndef MIN
+# define MIN(a, b)	(((a) < (b))? (a) : (b))
+#endif
```
Jakuje, line 101: We have this defined three times throught the codebase already:
```
src/libopensc/internal.h:#define MIN(x, y) (((x) < (y)) ? (x) : (y))
src/pkcs15init/pkcs15-cardos.c:# define MIN(a, b)       (((a) < (b))? (a) : (b))
src/pkcs15init/pkcs15-incrypto34.c:# define MIN(a, b)   (((a) < (b))? (a) : (b))
```
could the `internal.h` implementation be reused? Many other tools and places already include the `internal.h`.

Jakuje (commented):
During `make check`, there are some basic tests executed against the softhsm pkcs11 module. The `pkcs11-tool --test` is one of them, but is now expected to fail for now. Can you check how it is running with your changes?

https://github.com/OpenSC/OpenSC/blob/master/tests/test-pkcs11-tool-test.sh

From the CI side, it would be probably good to add some `cat tests/*.log` to see the results also in CI after this line:

https://github.com/OpenSC/OpenSC/blob/master/.github/build.sh#L54

frankmorgner (approved):

PR #2781, opened 2023-05-12
CI: tie ix86 to ubuntu-20.04 to avoid dependency problem in github image

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


Jakuje:
Sorry for a delay. I do not think this is the best solution, but probably good enough for now.

I think we should follow some of the changes from https://github.com/actions/runner-images/issues/4589 -- I was reading through the mails, but did not get to propose a change here yet.
The bad thing is that github keeps breaking this over and over again ...

PR #2795, opened 2023-06-07
Fix one oss-fuzz report, document the way to get better traces from fuzzing and fix coverity warning from recent IDPrime

Jakuje:
##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):
thank you

PR #2798, opened 2023-06-08
enable use_file_cache for MyEID card

popovec:
This PR implements new card_ctl (SC_CARDCTL_GET_CHANGE_COUNTER) which makes it possible to identify every change on the card. This information allows us to safely use the file cache even if the contents of the card is changed on another computer. In this PR, this operation is implemented for the MyEID driver and the file cache for the MyEID card is turned on by default.

##### Checklist


src/libopensc/types.h
```diff
@@ -374,6 +374,8 @@ typedef struct sc_serial_number {
 	struct sc_iin iin;
 } sc_serial_number_t;
 
+typedef int sc_change_counter;
```
frankmorgner, line 377: I think this typedef is a little too much. Just using a size_t to reflect the state would be enough, I think.

frankmorgner (changes requested):

popovec:
Ok "size_t" is perfectly sufficient.

frankmorgner (approved):
looks good, thank you

PR #2801, opened 2023-06-12
misc: Update address in the LGPL 2.1 license headers

Jakuje:
Based on https://spdx.org/licenses/LGPL-2.1-or-later.html

Some of the linters in Fedora pointed out we use the old FSF address in most of our headers. This is mass edit to include the correct one. This was already done in the COPYING file in #1902 some years back.

This is not only in the old files, but it is also copied to the new ones that "inherit" (or copy&paste) the headers from the old ones.

This should not have any licensing impact, but only clean up existing files and making sure we will not carry them forever.

For the long term we should probably consider usage of SPDX license identifiers in headers only as recommended in https://reuse.software/ (these have also quite nice CI tools to keep this up to date and usable in the long term). I will probably open separate issue for this to get this done if there would not be any huge pushback.

### DIFF
--- a/src/common/compat_overflow.c
+++ b/src/common/compat_overflow.c
@@ -1,5 +1,5 @@
 /*
- * compat_overflow.h: Reimplementation of GCC/Clang's built-in
+ * compat_overflow.c: Reimplementation of GCC/Clang's built-in
  * functions to perform arithmetic with overflow checking
  *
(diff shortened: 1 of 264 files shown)


frankmorgner:
Thank you.

Although I do not immediately see the benefit for the SPDX identifiers and the CI integration of REUSE, feel free to explore.

PR #2803, opened 2023-06-15
Handle short-lived removals of reader in reader driver for PC/SC

xhanulik:
As discussed in #2740, the `refresh_attributes()` function in `reader-pcsc.c` does not correctly handle the state when the reader is removed and inserted between two subsequent calls to `refresh_attributes()`. In that case, the PCSC `SCardGetStatusChange()` returns `SCARD_E_TIMEOUT`, handled as the card is present and did not change.

In https://github.com/OpenSC/OpenSC/pull/2740#issuecomment-1551641784, there was proposed to use one more call to `SCardBeginTransaction()`, it seems to be also working with `SCardStatus()` - when timeout is received, the invalid card handle should signalize recently removed reader and `SC_READER_CARD_CHANGED` is set.

The flag `SC_READER_CARD_CHANGED` is also set when `SCARD_E_UNKNOWN_READER` is first received from `SCardGetStatusChange()`.

The changes regarding slot handling remove the infinite loop from `card_detect()` when `SC_READER_CARD_CHANGED` is repeatedly acquired. It happened when the reader was removed & inserted back, so the timeout was reached, but no reconnection between the `refresh_attributes()` in `pcsc-reader.c` happens - the card handle is invalid, so the flag about changed is repeatedly set. This should probably be handled differently - any insight would be appreciated.

The changes were tested with Yubikey according to https://github.com/OpenSC/OpenSC/pull/2740#issuecomment-1534926539.


Jakuje:
Nits:
 * Typo in the commit message: `the change is not reported. Tthe card handle should` in f16e69842ca0068b81674ebf4b75717d5a7ff792
 * Typo in the commit message: `reader-pscs:` in d8e661df827e401bb60d9ea3baef8df0b7b38a9d -- I think this also needs a bit more explanation why this is done (to avoid cycling somewhere?)

Otherwise looks reasonable.

xhanulik:
> * Typo in the commit message: `reader-pscs:` in [d8e661d](https://github.com/OpenSC/OpenSC/commit/d8e661df827e401bb60d9ea3baef8df0b7b38a9d) -- I think this also needs a bit more explanation why this is done (to avoid cycling somewhere?)

This was done only to reflect the change in the flags. Before this change, when the reader was removed, the `reader->flags` in `refresh_attributes()` was updated from `0x00000001 (SC_READER_CARD_PRESENT)` to `0x00000020 (SC_READER_REMOVED)`. However, when the reader was still removed by another call to `refresh_attributes()`, the flags remained `0x00000020 (SC_READER_REMOVED)`.

Now the flags should change as `0x00000001 (SC_READER_CARD_PRESENT) → 0x00000022 (SC_READER_REMOVED | SC_READER_CARD_CHANGED)` and then  `0x00000022 (SC_READER_REMOVED | SC_READER_CARD_CHANGED) → 0x00000020 (SC_READER_REMOVED)`.

It is reflected in the commit message now. Thanks for the review!

frankmorgner:
[Ludovic stated that](https://github.com/OpenSC/OpenSC/pull/2740#issuecomment-1551641784)
> I note that, on Windows, the SCardBeginTransaction() does not fail after reader removal/re-insertion. So you should also check the result of the next PC/SC call.
> I have not checked under macOS.

So if I understand correctly, then using SCardStatus for removal/reinsertion may only work in pcsc-lite, right?

xhanulik:
> So if I understand correctly, then using SCardStatus for removal/reinsertion may only work in pcsc-lite, right?

I understood it as `SCardBeginTransaction` would not detect it; however, I'm not sure about `SCardStatus` - unfortunately, I cannot test it myself.

xhanulik:
I also measured possible time overhead regarding the changes in PCSC calling. Using the `time` command and `pkcs11-tool --test`, I did measurements for the PIV card, when the time increased on average (4 tries) from 46.352s to 46.576s and IDPrime card, when the increase was approximately from 7.382s to 7.445s.

src/libopensc/reader-pcsc.c
```diff
@@ -393,7 +407,13 @@ static int refresh_attributes(sc_reader_t *reader)
 				|| rv == (LONG)SCARD_E_NO_READERS_AVAILABLE
```
Jakuje, line 407: nit: on the above line, there is a typo `prevoiusly` -- not sure why this was not caught by codespell ...

Jakuje (approved):

src/libopensc/reader-pcsc.c
xhanulik: Thanks, fixed.

Jakuje:
@LudovicRousseau can you have a look if this looks reasonable to you?

src/libopensc/reader-pcsc.c
```diff
@@ -376,24 +376,44 @@ static int refresh_attributes(sc_reader_t *reader)
...
+			unsigned char atr[SC_MAX_ATR_SIZE];
+
 			/* Make sure to preserve the CARD_PRESENT flag if the reader was
 			 * reattached and we called the refresh_attributes too recently */
 			if (priv->reader_state.dwEventState & SCARD_STATE_PRESENT) {
 				reader->flags |= SC_READER_CARD_PRESENT;
 			}
+
+			rv = priv->gpriv->SCardStatus(priv->pcsc_card, NULL, &readers_len, &cstate, &prot, atr, &atr_len);
+			if (priv->pcsc_card != 0
```
LudovicRousseau, line 389: Maybe you can test if `priv->pcsc_card != 0` **before** the call to `SCardStatus()` to note waste CPU cycles?

A `hCard` set to 0 is invalid for pcsc-lite https://github.com/LudovicRousseau/PCSC/blob/master/src/winscard.c#L1254-L1256. But I don't know for the other WinSCard implementations (Windows and macOS).

LudovicRousseau:
No objection. But I am not an OpenSC internals expert.

I suggest to make some tests on all the platforms (GNU/Linux, macOS, Windows, etc.) supported by OpenSC to be sure to capture all the different behaviors of `SCardStatus()`. I would use https://github.com/LudovicRousseau/PCSC/blob/master/UnitaryTests/SCardStatus.py for that.

src/libopensc/reader-pcsc.c
```diff
@@ -375,14 +375,29 @@ static int refresh_attributes(sc_reader_t *reader)
 
 	if (rv != SCARD_S_SUCCESS) {
 		if (rv == (LONG)SCARD_E_TIMEOUT) {
-			/* Timeout, no change from previous recorded state. Make sure that
-			 * changed flag is not set. */
-			reader->flags &= ~SC_READER_CARD_CHANGED;
+			DWORD readers_len = 0, cstate = 0, prot, atr_len = SC_MAX_ATR_SIZE;
+			unsigned char atr[SC_MAX_ATR_SIZE];
```
Jakuje, line 379: nit: I would probably move these variables into the inner block where you use them.

src/pkcs11/slot.c
```diff
@@ -427,6 +428,17 @@ card_detect_all(void)
 				}
 			}
 			card_detect(reader);
+
+			/* If reader was removed, card_removed() was already called in card_detect(),
+			 * remove only relation between reader and slot. */
+			if (reader->flags & SC_READER_REMOVED) {
+				for (j = 0; j<list_size(&virtual_slots); j++) {
+					sc_pkcs11_slot_t *slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, j);
```
Jakuje, line 436: there are few minor formatting suggests from clang-format, that I would like to bring in too.

Jakuje (approved):
minor nits and format, but I think this is good to go.

xhanulik:
There were problems with oseid and isoapplet CI tests since previously the changes contained the check for card change as this:

```
rv = priv->gpriv->SCardStatus(priv->pcsc_card, NULL, &readers_len, &cstate, &prot, atr, &atr_len);
if (rv == (LONG)SCARD_W_REMOVED_CARD || rv == (LONG)SCARD_E_INVALID_VALUE || rv == (LONG)SCARD_E_INVALID_HANDLE)
	reader->flags |= SC_READER_CARD_CHANGED;
```
For some reason, in tests, the `SCARD_E_INVALID_HANDLE` value was returned by `SCardStatus,` which causes the `C_GetTokenInfo()` return `CKR_TOKEN_NOT_PRESENT` and tests fail.

With my Yubikey, when I reattach the Yubikey in between calling to `refresh_attributes(),` then `SCardStatus` returns `SCARD_E_INVALID_VALUE.` The `SC_READER_CARD_CHANGED` flag is reported back to `card_detect()`, and `card_removed()` is called. After this, `sc_detect_card_presence()` and `refresh_attributes()` are called again and `SCARD_E_INVALID_HANDLE` is returned by `SCardStatus()`. From this, I think it should be fine to use just this

```
rv = priv->gpriv->SCardStatus(priv->pcsc_card, NULL, &readers_len, &cstate, &prot, atr, &atr_len);
if (rv == (LONG)SCARD_W_REMOVED_CARD || rv == (LONG)SCARD_E_INVALID_VALUE)
	reader->flags |= SC_READER_CARD_CHANGED;
```

for checking that _card has changed_. Just not sure whether the `SCARD_E_INVALID_HANDLE` from `SCardStatus()` should be reported to higher levels because when _no reconnection is done afterward_, `SCardGetStatusChange()` will still return timeout and card handle will be I invalid I.

src/pkcs11/slot.c
```diff
@@ -427,6 +428,17 @@ card_detect_all(void)
 				}
 			}
 			card_detect(reader);
+
+			/* If reader was removed, card_removed() was already called in card_detect(),
+			 * remove only relation between reader and slot. */
+			if (reader->flags & SC_READER_REMOVED) {
```
frankmorgner, line 434: I assume this special case would happen if the reader was removed while being in this very `else` branch. However, is this special treatment really needed? I feel this is not a relevant issue, because it is very unlikely to happen at all *and* the bad consequences are unproblematic (i.e. some slot that was already removed may be visible to the application until card_detect_all is called the next time).

frankmorgner (changes requested):

src/pkcs11/slot.c
```diff
@@ -235,10 +236,10 @@ CK_RV card_detect(sc_reader_t *reader)
 		sc_log(context, "%s: Card changed", reader->name);
 		/* The following should never happen - but if it
 		 * does we'll be stuck in an endless loop.
-		 * So better be fussy.
-		if (!retry--)
-			return CKR_TOKEN_NOT_PRESENT; */
+		 * So better be fussy.*/
 		card_removed(reader);
+		if (!retry--)
+			return CKR_TOKEN_NOT_PRESENT;
```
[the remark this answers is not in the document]
xhanulik: > The changes regarding slot handling remove the infinite loop from card_detect() when SC_READER_CARD_CHANGED is repeatedly acquired. It happened when the reader was removed & inserted back, so the timeout was reached, but no reconnection between the refresh_attributes() in pcsc-reader.c happens - the card handle is invalid, so the flag about changed is repeatedly set. This should probably be handled differently - any insight would be appreciated.

This was the initial problem, however I suppose it is really a rare case.

src/pkcs11/slot.c
xhanulik: It doesn't only cover the case, when reader is removed during the `else` branch,. The `card_detect_all()` does not update the the state of readers when looping over them:

https://github.com/OpenSC/OpenSC/blob/f7070f4c9f9c6e1cb8d2efe838f73d54c78553da/src/pkcs11/slot.c#L393-L398

So it catches the case, when reader is removed _after_ calling `refresh_attributes()`, so `reader->flags` are not yet updated.

src/pkcs11/slot.c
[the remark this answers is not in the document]
frankmorgner: If it is a siginificant problem even in rare cases, I don't mind having a prevention in the code, but as written above, the "prevention" currently doesn't work because `retry` is currently reset every time you call the function.

src/pkcs11/slot.c
frankmorgner: If that is your intention, then I think this code should not be added at all, because modifying the virtual slots would result in a different slotlist that are communicated to the application. that list, however needs to be static until the application calls C_GetSlotList, see https://github.com/OpenSC/OpenSC/issues/3241#issuecomment-2418834920

xhanulik: I actually didn't realize this won't align with the C_GetSlotList behavior, thanks for pointing out. In that case, this PR will just add the correct check for `SC_READER_CARD_CHANGED` and will not touch the connection between slot and reader.

src/pkcs11/slot.c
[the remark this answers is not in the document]
xhanulik: Yes, I understand that it should be `static` to prevent the behavior in between calls to that function, but the problem I've run into was cycling just inside one call, since it was stucked here, calling `sc_detect_card_presence`, getting `SC_READER_CARD_CHANGED` and going back to `again`:

https://github.com/OpenSC/OpenSC/blob/f7070f4c9f9c6e1cb8d2efe838f73d54c78553da/src/pkcs11/slot.c#L221-L242

However, the only case I can hit this problem is now is with the oseid, where I'm trying to find out what also causes the problem described here https://github.com/OpenSC/OpenSC/pull/2803#issuecomment-2404556863. So I'll fix the `static` or remove the whole thing, if I'll find out what is causing it.

Jakuje (approved):
just one nit

src/pkcs11/slot.c
frankmorgner, line 242:
```suggestion
		if (!retry--) {
			retry = 3;
			return CKR_TOKEN_NOT_PRESENT;
		}
```

frankmorgner (changes requested):
`retry` needs to be reset explicitly so that looping works when the function is called a second time

src/libopensc/reader-pcsc.c
```diff
@@ -420,7 +440,7 @@ static int refresh_attributes(sc_reader_t *reader)
 		SC_FUNC_RETURN(reader->ctx, SC_LOG_DEBUG_VERBOSE, SC_SUCCESS);
 	}
 
-	reader->flags &= ~(SC_READER_CARD_CHANGED|SC_READER_CARD_INUSE|SC_READER_CARD_EXCLUSIVE);
+	reader->flags &= ~(SC_READER_CARD_CHANGED | SC_READER_CARD_INUSE | SC_READER_CARD_EXCLUSIVE | SC_READER_REMOVED);
```
frankmorgner, line 443: Is this change really needed? `SC_READER_REMOVED` basically means that its ressources are invalid and should not be accessed anymore. This flag is reset in pcsc_disconnect, where also the card handle is freed. Resetting this flag here without renewing the ressources may lead to problems.

xhanulik: Makes sense, fixed.

frankmorgner (approved):

PR #2807, opened 2023-06-21
Implemented logout for some card drivers

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-cac.c
+++ b/src/libopensc/card-cac.c
@@ -1852,6 +1852,12 @@ static int cac_init(sc_card_t *card)
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
 
+static int cac_logout(sc_card_t *card)
+{
+	int index;
+	return cac_find_first_pki_applet(card, &index);
+}
+
 static int cac_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 {
 	/* CAC, like PIV needs Extra validation of (new) PIN during
@@ -1923,6 +1929,7 @@ static struct sc_card_driver * sc_get_driver(void)
 	cac_ops.decipher =  cac_decipher;
 	cac_ops.card_ctl = cac_card_ctl;
 	cac_ops.pin_cmd = cac_pin_cmd;
+	cac_ops.logout = cac_logout;
 
 	return &cac_drv;
 }
--- a/src/libopensc/card-cac.c
+++ b/src/libopensc/card-cac.c
@@ -1821,9 +1821,6 @@ static int cac_match_card(sc_card_t *card)
 {
 	int r;
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
-	/* Since we send an APDU, the card's logout function may be called...
-	 * however it may be in dirty memory */
-	card->ops->logout = NULL;
 
 	r = cac_find_and_initialize(card, 0);
 	return (r == SC_SUCCESS); /* never match */
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -856,6 +856,19 @@ static int muscle_card_reader_lock_obtained(sc_card_t *card, int was_reset)
 	LOG_FUNC_RETURN(card->ctx, r);
 }
 
+static int muscle_logout(sc_card_t *card)
+{
+	int r = SC_ERROR_NOT_SUPPORTED;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (msc_select_applet(card, muscleAppletId, sizeof muscleAppletId) == 1) {
+		r = SC_SUCCESS;
+	}
+
+	LOG_FUNC_RETURN(card->ctx, r);
+}
+
 
 static struct sc_card_driver * sc_get_driver(void)
 {
@@ -884,6 +897,7 @@ static struct sc_card_driver * sc_get_driver(void)
 	muscle_ops.delete_file = muscle_delete_file;
 	muscle_ops.list_files = muscle_list_files;
 	muscle_ops.card_reader_lock_obtained = muscle_card_reader_lock_obtained;
+	muscle_ops.logout = muscle_logout;
 
 	return &muscle_drv;
 }
--- a/src/libopensc/card-muscle.c
+++ b/src/libopensc/card-muscle.c
@@ -81,10 +81,6 @@ static int muscle_match_card(sc_card_t *card)
 	u8 response[64];
 	int r;
 
-	/* Since we send an APDU, the card's logout function may be called...
-	 * however it's not always properly nulled out... */
-	card->ops->logout = NULL;
-
 	if (msc_select_applet(card, muscleAppletId, sizeof muscleAppletId) == 1) {
 		/* Muscle applet is present, check the protocol version to be sure */
 		sc_format_apdu(card, &apdu, SC_APDU_CASE_2, 0x3C, 0x00, 0x00);
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -3089,10 +3089,6 @@ static int piv_match_card_continued(sc_card_t *card)
 	piv_private_data_t *priv = NULL;
 	int saved_type = card->type;
 
-	/* Since we send an APDU, the card's logout function may be called...
-	 * however it may be in dirty memory */
-	card->ops->logout = NULL;
-
 	/* piv_match_card may be called with card->type, set by opensc.conf */
 	/* user provide card type must be one we know */
 	switch (card->type) {
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -2189,11 +2189,11 @@ static int piv_is_object_present(sc_card_t *card, u8 *ptr)
  * or the global pin for the card 0x00. Look at Discovery object to get this.
  * called by pkcs15-piv.c  via cardctl when setting up the pins.
  */
-static int piv_get_pin_preference(sc_card_t *card, int *ptr)
+static int piv_get_pin_preference(sc_card_t *card, int *pin_ref)
 {
 	piv_private_data_t * priv = PIV_DATA(card);
 
-	*ptr = priv->pin_preference;
+	*pin_ref = priv->pin_preference;
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
 
@@ -3760,12 +3760,18 @@ piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 
 static int piv_logout(sc_card_t *card)
 {
-	int r = SC_ERROR_NOT_SUPPORTED; /* TODO Some PIV cards may support a logout */
-	/* piv_private_data_t * priv = PIV_DATA(card); */
+	int r = SC_ERROR_NOT_SUPPORTED;
+	piv_private_data_t * priv = PIV_DATA(card);
 
 	LOG_FUNC_CALLED(card->ctx);
 
-	/* TODO 800-73-3 does not define a logout, 800-73-4 does */
+	if (priv) {
+		/* logout defined since 800-73-4 */
+		r = iso7816_logout(card, priv->pin_preference);
+		if (r == SC_SUCCESS) {
+			priv->logged_in = SC_PIN_STATE_LOGGED_OUT;
+		}
+	}
 
 	LOG_FUNC_RETURN(card->ctx, r);
 }
--- a/src/libopensc/card-starcos.c
+++ b/src/libopensc/card-starcos.c
@@ -2160,18 +2160,9 @@ static int starcos_card_ctl(sc_card_t *card, unsigned long cmd, void *ptr)
 	}
 }
 
-/**
- * starcos_logout_v3_x()
- * StarCOS 3.x cards will not clear the security status by selecting MF.
- * Returning NOT_SUPPORTED would cause card reset, effectively invalidating
- * the security status.
- */
 static int starcos_logout_v3_x(sc_card_t *card)
 {
-	int r = SC_ERROR_NOT_SUPPORTED;
-	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_NORMAL);
-
-	SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, r);
+	return SC_ERROR_NOT_SUPPORTED;
 }
 
 static int starcos_logout(sc_card_t *card)
--- a/src/libopensc/card-asepcos.c
+++ b/src/libopensc/card-asepcos.c
@@ -1050,6 +1050,20 @@ static int asepcos_card_reader_lock_obtained(sc_card_t *card, int was_reset)
 	LOG_FUNC_RETURN(card->ctx, r);
 }
 
+static int asepcos_logout(sc_card_t *card)
+{
+	int r = SC_ERROR_NOT_SUPPORTED;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (card->type == SC_CARD_TYPE_ASEPCOS_JAVA) {
+		/* in case of a Java card try to select the ASEPCOS applet */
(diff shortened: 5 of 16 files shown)


src/libopensc/card-piv.c
```diff
@@ -3760,12 +3760,15 @@ piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
...
 {
-	int r = SC_ERROR_NOT_SUPPORTED; /* TODO Some PIV cards may support a logout */
-	/* piv_private_data_t * priv = PIV_DATA(card); */
+	int r, pin_ref;
 
 	LOG_FUNC_CALLED(card->ctx);
 
-	/* TODO 800-73-3 does not define a logout, 800-73-4 does */
+	r = piv_get_pin_preference(card, &pin_ref);
+	LOG_TEST_RET(card->ctx, r, "Failure retrieving PIN reference");
```
Jakuje, line 3768: should the failure here cause the `sc_reset()` on cards not supporting the -4 specs?

src/libopensc/card-cac.c
```diff
@@ -1852,6 +1852,12 @@ static int cac_init(sc_card_t *card)
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
 
+static int cac_logout(sc_card_t *card)
+{
+	int index;
+	return cac_find_first_pki_applet(card, &index);
+}
+
```
Jakuje, line 1860: This is the same for CAC and CAC1 so I would propose to move it to card-cac-common.{c,h} to avoid code duplication (even though its just couple of lines)

src/libopensc/card-cac.c
```diff
@@ -1923,6 +1929,7 @@ static struct sc_card_driver * sc_get_driver(void)
 	cac_ops.decipher =  cac_decipher;
 	cac_ops.card_ctl = cac_card_ctl;
 	cac_ops.pin_cmd = cac_pin_cmd;
+	cac_ops.logout = cac_logout;
```
Jakuje, line 1932: For the CAC, there is `card->ops->logout = NULL;` in `cac_match_card()`. I am no longer sure what was the intention of that part initially, but it looks like it makes the changes here no-op?

Very similar code is in muscle, piv, coolkey and cac1 drivers now.

src/libopensc/card-starcos.c
```diff
@@ -2160,18 +2160,9 @@ static int starcos_card_ctl(sc_card_t *card, unsigned long cmd, void *ptr)
 	}
 }
 
-/**
- * starcos_logout_v3_x()
- * StarCOS 3.x cards will not clear the security status by selecting MF.
- * Returning NOT_SUPPORTED would cause card reset, effectively invalidating
- * the security status.
- */
```
Jakuje, line 2168: This comment looks like it was never true as returning `NOT_SUPPORTED` is mostly ignored by the callers of the logout function. OTOH, would it make sense to make it this way to avoid calling the `sc_reset()` in most of the readers and just return `NOT_SUPPORTED` and let the something outside of the driver handle the situation (call the `sc_reset()`)?

src/libopensc/card-cac.c
frankmorgner: cac_find_first_pki_applet is implemented independent from each other (not in card-cac-common.{c,h}). If you confirm, I would move this to the common part as well

src/libopensc/card-piv.c
frankmorgner: Doing a reset for PIV is generally not a good idea, because a PIV *applet* may then not be selected anymore. I realised that @dengert has an even more relaxed handling of errors (i.e. [returning SUCCESS if logout fails](https://github.com/OpenSC/OpenSC/blob/c853dd4df834e461500b015eabfd1ea49a44b757/src/libopensc/card-piv.c#L6147-L6148)). I think I would go with Doug's implementation, but return the error to the upper layer if it fails. For example, in CTK we can handle this error by still treating the card as *sensitive* (i.e. to be separated from other processes).

src/libopensc/card-cac.c
frankmorgner: I need to check on this...

src/libopensc/card-starcos.c
frankmorgner: Indeed, a reset was never done if logout was *not* implemented, so the comment is and has always been wrong. And I think it would be a bad idea to implement it this way.

I agree that propagating the error would be the most sane thing to do as it may be handled in some situations (see comment above about handling it in CTK).

frankmorgner: That being said, the purpose of this PR is to actually *do a reset* if we don't know better what to do and if we know it is safe to reset.

src/libopensc/card-piv.c
dengert: For the record the PIV driver at the start of **every** transaction will either do GET_DATA for the DISCOVERY object and/or a  SELECT AID. (If DISCOVERY is not found or useless it will do a SELECT AID.)  Some "PIV like" cards would lose login in state even if SELECT AID for PIV was used!

From day one in 2004, the goal was to keep the PIV login state active so it could be used by multiple applications especially with PIN PAD reader  even after an application ends. With cards with multiple applet, this is a problem, as NIST 800-73 specially says if an other applet that is on the card is selected, the PIV applet login state is reset. Some "PIV like" cards will also reset the login state if a SELECT_AID is done even if the card does have that applet. So the goal is not possible with some cards or with some non-OpenSC applications. 

NIST only defined the logout in 800-73-4 But does not require it be used.

The use of `opensc.conf` `disconnect_action = leave;` is really the way a user can say to leave the card in a logged in state, and not do a reset or logout when application is done. `disconnect_action = reset;` 

IMHO the decision to do a reset or logout should only be done when `disconnect_action = reset;`  or `disconnect_action = unpower;`

src/libopensc/card-starcos.c
dengert: If you don't know what to do, look at `disconnect_action`

src/libopensc/card-cac.c
Jakuje: Oh, my bad. It was long time since I touched the CAC driver and forgot there are these differences in implementation.

As far as I know, selecting any applet broke the login state in the CAC so maybe just calling select applet might do, but I would have to check or try. Let me get back to that next week.

src/libopensc/card-starcos.c
Jakuje: > If you don't know what to do, look at `disconnect_action`

I do not think the disconnect action should be executed on logout without any warning (even as a fallback).

dengert: What I was implying is if `opensc.conf` has `disconnect_action = reset;` or `disconnect_action = unpower;` then user has modified  the parameter from default of `disconnect_action = leave;` in response to:
> That being said, the purpose of this PR is to actually do a reset if we don't know better what to do and if we know it is safe to reset.

Note that pkcs11 3.0 says: "C_Logout logs a user out from a token."  "Depending on the current user type, if the call succeeds, each of the application’s sessions will enter either the “R/W Public Session” state or the “R/O Public Session” state."  The login state is then controlled by the PKCS11 middleware, just as when   

I read this as this application still has access to the token with "Public Sessions" so reset should not be used. And other applications may still continue to have access to the token wit a logged in state.

PKCS11, PCSC and ISO 7816 don't match when multiple applications have their own PKCS11 and are using shared access to a token via PC/SC. 

But OpenSC does define the  `disconnect_action` that imply (leave) if login state should be not be changed or (reset) which would cause the token to lose the login state.    There could be a 4th state `disconnect_action = logout;`   which would then call the iso7816 logout if user wanted to change the token login state.

mouse07410: Let me express the strongest agreement with the point @dengert made.

src/libopensc/card-piv.c
frankmorgner: If I understood correctly, you want to use the logout command from 800-73-4 and if that fails we should do a reset of the card? That would be easy to do and we would just have to reselect the PIV AID to make sure the applet is active.

src/libopensc/card-starcos.c
frankmorgner: `disconnect_action` does not yet have this security related connotation, I doubt that using this for logout would be an easy to get change for most users.

I understand that we may return an error to the application when a logout is requested (currently, we actually don't). But I don't think we need to put this burden onto the application. (Also note that an application cannot make  a reset via PKCS#11) My thought process e.g. for PIV is the following:
1. If we know how to manually log out via an APDU, go ahead and do so
2. if that fails and it's an applet based card, reselect the applet, which *typically* resets the security state (essentially I'm performing the action from the card's `match_card` and `init` is performed).
3. if that fails or cannot be applied, use `sc_reset` as last resort, which, in particular, is safe to use on static card that don't have volatile data loaded at `init` and `match_card`

I don't think that this would cause problems for applet based cards that have multiple opened sessions to different applications. For one, I cannot imagine a use case scenario where OpenSC is used somewhere and we don't have any information about anyone actually using our middleware this way. Second, assume some middleware supports concurrent access *and* is used together with OpenSC, then we are still safe to use a reset, because that other middleware would need to have some support for recovering from a reset (what we have implemented in `card_reader_lock_obtained`). If it doesn't have such thing implemented, then that middleware likely cannot really be used in parallel with OpenSC or anything else.

In conclusion, we are choosing a reset only if we don't have a any other choice and it should be safe to use. If you see any problems arising from this, please elaborate on the use case and on the software/hardware involved in this.

dengert: > disconnect_action does not yet have this security related connotation, I doubt that using this for logout would be an easy to get change for most users.

Well then add something like `logout_action = yes|NO`  Give the users a choice to continue using the cards as they have done since 2004.  See https://github.com/OpenSC/OpenSC/pull/2807#discussion_r1240997601 Don't break what has been working. Give users a choice even if it means another `opensc.conf` option. 

My concern is you will be turning over an application the choice to reset login state or not. Applications like FireFox, Thunderbird or others. Any one could reset the login state at any time.    

> My thought process e.g. for PIV is the following:

Why PIV only?

>If we know how to manually log out via an APDU, go ahead and do so 

NIST did not introduce logout until 800-73-4 in 2016 and the use of "leave" has worked well to not require pin caching and allow the use of PIN PAD readers where a pin can't be cached.     

>if that fails and it's an applet based card, reselect the applet, which typically resets the security state (essentially I'm performing the action from the card's match_card and init is performed).

PIV are always an "applet based card" and NIST has always said reselecting the PIV applet does not reset the login state. (But some "PIV Like" cards have not followed the standard.) Only if some other applet that exists on the card is selected, then the PIV login state is reset. So you would have to know what applets are on the card to use this method. 

>if that fails or cannot be applied, use sc_reset as last resort, which, in particular, is safe to use on static card that don't have volatile data loaded at init and match_card 

Like `disconnect_action = reset;` vs  `disconnect_action = unpower;`  Which cards does OpenSC support that are not static? 

In regard to cards with multiple applets used concurrent: We have that now with OpenPGP and PIV cards, or CAC and PIV cards. Its just the OpenSC can only support one of the applets from the same process, but users such as @mouse07410  I believe used other software for one of the applets. Starting an application with OPENSC_DRIVER= can also cause OpenSC in different processes to different applets.  So this is a case for pin caching by each long running application.

frankmorgner: AFAIK, nobody is using multiple card applications *in parallel*. Multiple long running processes accessing two different card applets at the same time are technically possible, but that's just an illusion. Organizations and people are fully occupied handling crypto tokens with just a single key, they don't want it to be even more complex than that. At least, that's my experience. If you, however, find some use case that actually breaks by accepting this pull request, then please let me know and we can think of some more complex solution.

CTK, MD and PKCS#11 created complex infrastructure to separate processes from each other that are accessing a single card at the same time to avoid a misuse of the card and you want to defeat all this by just leaving the card authenticated even after explicit logout? Look, here is the super secret key to your house, but anyone pressing the door handle can enter as well. I would consider this a security problem!

What NIST thought in their initial specification? Maybe they just forgot, maybe they thought a reset is good enough. I don't know and I don't care. I used PIV just as an example, the security issue I'm talking about, touches **every single one of those 22 different types of tokens from this pull request**.

dengert: @frankmorgner  I have given you my arguments, and options to give uses/administrators a choice. Do wat you want.

src/libopensc/card-cac.c
frankmorgner: I agree, that the code in match_card resets logout to NULL. A similar check in match_card appeared first in b3822c349b9b4c3e837298cd0d1e333faa09a604 for the muscle driver and seemed to be copied to other drivers later. I don't know what it is trying to achieve, but it looks like that it is not needed anymore.

src/libopensc/card-muscle.c
```diff
@@ -856,6 +856,19 @@ static int muscle_card_reader_lock_obtained(sc_card_t *card, int was_reset)
...
 }
 
+static int muscle_logout(sc_card_t *card)
+{
+	int r = SC_SUCCESS;
+
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (msc_select_applet(card, muscleAppletId, sizeof muscleAppletId) != 1) {
+		r = SC_ERROR_INVALID_CARD;
```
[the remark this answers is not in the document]
frankmorgner: done

src/libopensc/card-starcos.c
frankmorgner: OK, understood.

@Jakuje , what's your opinion on using `sc_reset()` for logout for some of the cards and/or adding a new configuration option for this (and its default value)?

Jakuje: Using multiple applets on a card/token is broken and will likely be broken forever. Unfortunately, there is yubikey which presents openpgp and piv applet. The GnuPG using OpenPGP and the rest of the world is using OpenSC with PIV default will be breaking always until either GnuPG will start accessing the card through PKCS#11 or I dont know what.

It is correct that in this case or in case several application will access the single applet on a card/token the reset will forcibly logout any other application, which is something the applications might not be expecting and might not handle correctly and which could cause quite huge influx of issues and we need some fallback for. Hardcoding `sc_reset()` into the card drivers will not make this easier for us.

On the other hand from my experience, the `C_Logout` is not called very frequently by applications. They usually call `C_Login` and then keep the login state until they call `C_Finalize` so I believe most of the users and applications will really not notice this change as their application will got through `disconnect_action` only.

My proposal would be to go ahead with returning `SC_ERROR_NOT_SUPPORTED` instead of calling the `sc_reset()` from the card drivers and handle this in one place.

How, it is another question. We need sane/secure defaults and we need "compatible" option which sounds like the only way to achieve this is with another opensc.conf option, `logout_fallback = [reset|leave|...]` that would allow use to configure this behavior separately from the disconnection action.

frankmorgner: OK, I've removed all manual calls to `sc_reset()`.

For later enhancement for logging out, please note the following:
- In Minidriver, Windows will perform a reset on its own if [CardDeauthenticateEx](https://learn.microsoft.com/en-us/previous-versions/dn468713(v=vs.85)) (`sc_logout()`) fails.
- In PKCS#11 we're calling `sc_logout()` also when a session terminates.

frankmorgner:
I think all comments are covered now. In particulare, sc_reset() is not called to logout the card.

If there are no other objections, I will merge this PR as is.

src/libopensc/card-piv.c
frankmorgner: card-piv.c changes were removed from this PR. However, I think we should implement piv_logout correctly if possible, i.e. with explicit logout.

I don't know if disconnect_action is still needed in 2023, but if you think we should respect this when shutting down the minidriver or the PKCS#11 module, then this should be removed in [‎src/minidriver/minidriver.c‎](https://github.com/OpenSC/OpenSC/blob/6e3908fb655d3266397bfebdb328ff8f3c7206e8/src/minidriver/minidriver.c#L4091) or [‎src/pkcs11/pkcs11-session.c‎](https://github.com/OpenSC/OpenSC/blob/6e3908fb655d3266397bfebdb328ff8f3c7206e8/src/pkcs11/pkcs11-session.c#L126) respectively.

Issue #2790, opened 2023-05-31
Review, clarify and doucment ownership of  `ec->params` and similar throught the code


Jakuje:
> Regarding the `ec->params`, we need to get a common understanding about memory owner ship and proper initialization/cleanup this is ture. But I'd like to postpone this for now to move on with all the good stuff in this PR.

_Originally posted by @frankmorgner in https://github.com/OpenSC/OpenSC/pull/2756#discussion_r1206367123_

As this looks like the EC support is still kind of hackish in some places, the memory used to be passed back end forth without much care for correctness.

This is brought up from time to time when addressing some invalid memory access/use-after-free/double-free/memory leaks reported by oss-fuzz/coverity/valgrind.

PR #2810, opened 2023-06-22
Fix OSS-Fuzz issues

xhanulik:
This PR contains fixes for recent issues reported by OSS-Fuzz

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -1664,17 +1664,11 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 
 err:
 	sc_pkcs15_free_object(object);
-	free(pubkey_args.key.alg_id);
+	sc_pkcs15_erase_pubkey(&pubkey_args.key);
 	if (algorithm == SC_ALGORITHM_EC) {
-		/* Allocated in sc_copy_ec_params() */
-		free(pubkey_args.key.u.ec.params.named_curve);
-		free(pubkey_args.key.u.ec.params.der.value);
 		/* allocated in check_keygen_params_consistency() */
 		free(keygen_args->prkey_args.key.u.ec.params.der.value);
 		keygen_args->prkey_args.key.u.ec.params.der.value = NULL;
-		/* can be allocated in driver-specific generate_key() */
-		free(pubkey_args.key.u.ec.ecpointQ.value);
-		pubkey_args.key.u.ec.ecpointQ.value = NULL;
 	}
 	LOG_FUNC_RETURN(ctx, r);
 }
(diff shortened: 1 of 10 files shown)


src/pkcs15init/profile.c
```diff
@@ -1985,24 +1985,40 @@ static struct block	root_ops = {
    "root",		process_block,	NULL,		root_blocks
 };
 
+static void
+get_inner_word(char *str, char word[64]) {
```
xhanulik, line 1989: I'm not sure whether the macros can be nested inside another expression; the problematic one found by OSS-Fuzz looks like `odf-size=($odf-size�v$` (with some non-alphanumeric byte).

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1674,6 +1675,8 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 		/* can be allocated in driver-specific generate_key() */
 		free(pubkey_args.key.u.ec.ecpointQ.value);
 		pubkey_args.key.u.ec.ecpointQ.value = NULL;
+		if (r < 0 && alg_id_free)
+			free(pubkey->alg_id);
```
Jakuje, line 1679: 1) I think this should not be in the condition `(algorithm == SC_ALGORITHM_EC)` as this can happen also for other algorithms.
2) I think in regards to the recent changes to this code I did with valgrind, I think the `alg_id` should be copied in all cases and free here in all cases to simplify the memory management, similarly as done with the other parts of pubkey structure.

src/pkcs15init/profile.c
Jakuje: I think this is mostly valid use case for example in the myeid profile object, there are macros used as part of the strings (indeed unless it is recursive).

src/pkcs15init/profile.c
```diff
@@ -1985,24 +1985,40 @@ static struct block	root_ops = {
...
+	char *inner = NULL;
+	size_t len = 0;
+
+	inner = str;
+	while (isalnum((unsigned char)*inner) || *inner == '-' || *inner == '_') {
```
Jakuje, line 1994: This condition is already used in `__expr_get()` to filter the macro names. I would prefer to move the condition to separate function to have these two uses in sync and make sure the do not diverge.

src/pkcs15init/profile.c
```diff
@@ -1985,24 +1985,40 @@ static struct block	root_ops = {
...
+		inner++;
+		len++;
+	}
+	len = len >= 64 ? 63 : len;
```
Jakuje, line 1998: using `sizeof(word)` instead of magic number would it make a bit more readable.

src/pkcs15init/profile.c
```diff
@@ -1985,24 +1985,40 @@ static struct block	root_ops = {
...
+	memcpy(word, str, len);
+	word[len] = '\0';
+}
+
 static int
-check_macro_reference_loop(char *start, char *str, struct state *cur, int depth) {
+check_macro_reference_loop(const char *start_name, char *value, sc_profile_t *profile, int depth) {
```
Jakuje, line 2004: I would propose a short doc string here with at least information what return values we can expect in which cases.
Ideally introducing also some unit tests to check the corner cases if possible as it is getting more complicated.

src/pkcs15init/profile.c
xhanulik: This would return only size of `char *`, but will use macro instead, thanks.

src/pkcs15init/profile.c
xhanulik: Ok, thanks for checking.

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1664,6 +1664,7 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 
 err:
 	sc_pkcs15_free_object(object);
+	free(pubkey->alg_id);
```
Jakuje, line 1667: This is not what I meant -- now you are freeing it even if it was not copied, which might cause double frees.

I had in my mind modifying the block from line 1629 to happen in any case to make sure we do not modify the existing sc_algorith_id structure pointed by `pubkey_args` if it is provided by caller and not share it with the newly created `pubkey` object to avoid the double-free.

Jakuje: My bad. You are right!

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1664,6 +1665,8 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 
 err:
 	sc_pkcs15_free_object(object);
+	if (r < 0 && alg_id_free)
+		free(pubkey->alg_id);
```
Jakuje, line 1669: I think the `pubkey->alg_id` should be freed regardless the return of this function as it is pointer to the `pubkey_args`, which is local to this function. It is passed to the `sc_pkcs15init_store_public_key()`, but if it needs it for something, it should copy it out similarly as with the ec params discussed in #2790.

And given that the `pubkey_args` is correctly zeroized in the beginning of this function, I think the `alg_id_free` is not needed either. It will be allocated either here or in the generate_key provided by the card driver and there is no card-specific free after this ...

src/pkcs15init/pkcs15-lib.c
```diff
@@ -1664,6 +1664,7 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 
 err:
 	sc_pkcs15_free_object(object);
+	free(pubkey_args.key.alg_id);
```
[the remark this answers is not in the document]
frankmorgner: Sorry, I just noticed your above discussion on this. I'll just wait a little while to see if you get to a conclusion...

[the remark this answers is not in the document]
Jakuje: I think in this case it is a not the caller, but the `generate_key()` callback of the pkcs15 emulator.

I did not go checking deeper if and when this is initialized in the drivers, but it looks like there is some different cleaning of this -- it should be easy to get traces with valgrind now (On the side note, it is now not executed in CI now because the regular run already crashes -- we might want to either execute the valgrind in the separate job or even if it would fail -- not sure what would be the best).

I think the it should be investigated slightly deeper rather than implementing fast solution using local variables like proposed initially. On the other hand if it is double free, setting just NULL after the free might solve the issue.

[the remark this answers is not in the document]
xhanulik: The double-free happens because `sc_pkcs15init_store_public_key()` in https://github.com/OpenSC/OpenSC/blob/b031789dccbdc46d151354719f0d527fe100cf76/src/pkcs15init/pkcs15-lib.c#L1654-L1667 releases the `alg_id` via `sc_pkcs15_erase_pubkey()`. Setting `NULL` there should work fine.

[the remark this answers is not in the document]
xhanulik: If I understand correctly, `sc_pkcs15_erase_pubkey()` in `sc_pkcs15init_store_public_key()` is called on _(shallow) copy_,
https://github.com/OpenSC/OpenSC/blob/b031789dccbdc46d151354719f0d527fe100cf76/src/pkcs15init/pkcs15-lib.c#L1858-L1860
so the `0/NULL`s are not propagated outside. The second call to  `sc_pkcs15_erase_pubkey()` then causes another double-free.

Jakuje (approved):

frankmorgner:
Thank you!

PR #2813, opened 2023-06-26
Fix some issues reported by Coverity scan

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -2098,6 +2098,8 @@ sc_der_copy(sc_pkcs15_der_t *dst, const sc_pkcs15_der_t *src)
 		return SC_ERROR_INVALID_ARGUMENTS;
 	memset(dst, 0, sizeof(*dst));
 	if (src->len) {
+		if (!src->value)
+			return SC_ERROR_INVALID_ARGUMENTS;
 		dst->value = malloc(src->len);
 		if (!dst->value)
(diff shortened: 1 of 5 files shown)


frankmorgner:
@popovec , could you please verify the fix in card-myeid.c?

src/smm/sm-cwa14890.c
```diff
@@ -139,6 +139,8 @@ sm_cwa_decode_authentication_data(struct sc_context *ctx, struct sm_cwa_keyset *
 		LOG_FUNC_RETURN(ctx, SC_ERROR_SM_AUTHENTICATION_FAILED);
 
 	rv = sm_decrypt_des_cbc3(ctx, keyset->enc, session_data->mdata, session_data->mdata_len, &decrypted, &decrypted_len);
+	if (rv != SC_SUCCESS)
+		free(decrypted);
```
Jakuje, line 143: would not be better if the `sm_decrypt_des_cbc3()` would not allocate/free the output parameter if it fails? With this you will have to fix the same issue in `sm-card-iasecc.c`.

frankmorgner: I didn't want to refactor this too much, but I can try to.

frankmorgner: done

Jakuje (approved):

PR #2814, opened 2023-06-27
myeid: fixed CID 380538  Out-of-bounds read (OVERRUN)

popovec:
also fixes output buffer size checking


Jakuje (approved):
looks correct, but the byte arithmetic might need some comment in the code as it is already hard to follow.

PR #2816, opened 2023-06-29
LibreSSL fixes

klemensn:
- LibreSSL: Use OPENSSL_cleanup()
- LibreSSL does provide EVP_sha3_*() after 3.7.3

Two fixes coming out of OpenBSD's official `security/opensc` port.
Tested with latest LibreSSL and

```
$ opensc-tool -n
Using reader with a card: Generic Smart Card Reader Interface [Smart Card Reader Interface] (20070818000000000) 00 00
EstEID 2018
```

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/sc-ossl-compat.h
+++ b/src/libopensc/sc-ossl-compat.h
@@ -50,10 +50,12 @@ extern "C" {
 #if LIBRESSL_VERSION_NUMBER < 0x30500000L
 #define FIPS_mode()                             (0)
 #endif
+#ifndef EVP_sha3_224
 #define EVP_sha3_224()                          (NULL)
 #define EVP_sha3_256()                          (NULL)
 #define EVP_sha3_384()                          (NULL)
 #define EVP_sha3_512()                          (NULL)
+#endif
 #if LIBRESSL_VERSION_NUMBER < 0x3070000fL
 #define EVP_PKEY_new_raw_public_key(t, e, p, l) (NULL)
 #define EVP_PKEY_get_raw_public_key(p, pu, l)   (0)
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -7230,8 +7230,12 @@ BOOL APIENTRY DllMain( HINSTANCE hinstDLL,
 	case DLL_PROCESS_DETACH:
 		sc_notify_close();
 		if (lpReserved == NULL) {
-#if defined(ENABLE_OPENSSL) && defined(OPENSSL_SECURE_MALLOC_SIZE) && !defined(LIBRESSL_VERSION_NUMBER)
+#if defined(ENABLE_OPENSSL)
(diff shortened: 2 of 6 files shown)


Jakuje:
The CI is stuck to 3.6.1. Can you also bump that version here:

https://github.com/OpenSC/OpenSC/blob/master/.github/setup-libressl.sh

Jakuje (approved):

Jakuje:

If you would have some good idea how to keep up with the libressl releases we are all ears. Unfortunately, dependabot does not look like supporting anything c-based so we either:

 * keep a fixed stable version until somebody notices an issue
 * remember to check for new version manually from time to time
 * track master and resolve intermittent master issues as they come

I am not happy with either of the options, but as I do not have any use case with libressl, its not a top priority for me now.

klemensn:
Perhaps replace LibreSSL (version) macro guards with actual feature guards where possible?

I see lots of checks for `#ifdef LIBRESSL_VERSION_NUMBER < ...`, but you could often just check for the actual feature in use, like this PR did in one hunk.

Then at least future LibreSSL versions automatically get a chance for new code.
This still does not solve the problem of CI being stuck on old versions, but I think it would also help package maintainers as they no longer have to correlate features with LibreSSL features, but instead "just" need to make it build with whatever version they use.

src/libopensc/sc-ossl-compat.h
```diff
@@ -50,10 +50,12 @@ extern "C" {
 #if LIBRESSL_VERSION_NUMBER < 0x30500000L
 #define FIPS_mode()                             (0)
 #endif
+#ifndef EVP_sha3_224
 #define EVP_sha3_224()                          (NULL)
 #define EVP_sha3_256()                          (NULL)
 #define EVP_sha3_384()                          (NULL)
 #define EVP_sha3_512()                          (NULL)
+#endif
 #if LIBRESSL_VERSION_NUMBER < 0x3070000fL
```
klemensn, line 59: These could probably be
```
#ifndef EVP_PKEY_new_raw_public_key
...
#endif /* ! EVP_PKEY_new_raw_public_key */
```
as a better test for [EVP_PKEY_NEW(3)](https://man.openbsd.org/EVP_PKEY_new_raw_public_key.3).

src/minidriver/minidriver.c
```diff
@@ -7230,8 +7230,12 @@ BOOL APIENTRY DllMain( HINSTANCE hinstDLL,
 	case DLL_PROCESS_DETACH:
 		sc_notify_close();
 		if (lpReserved == NULL) {
-#if defined(ENABLE_OPENSSL) && defined(OPENSSL_SECURE_MALLOC_SIZE) && !defined(LIBRESSL_VERSION_NUMBER)
+#if defined(ENABLE_OPENSSL)
+#if defined(OPENSSL_SECURE_MALLOC_SIZE) && !defined(LIBRESSL_VERSION_NUMBER)
```
klemensn, line 7234: Here ` && !defined(LIBRESSL_VERSION_NUMBER)` is useless since you're already using a feature guard.

PR #2817, opened 2023-07-03
Attempt to squash some more coverity reports.

Jakuje:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/apdu.c
+++ b/src/libopensc/apdu.c
@@ -402,8 +402,8 @@ sc_set_le_and_transmit(struct sc_card *card, struct sc_apdu *apdu, size_t olen)
 	apdu->resplen = olen;
 	apdu->le      = nlen;
 #ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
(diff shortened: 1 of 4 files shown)


src/libopensc/apdu.c
```diff
@@ -587,9 +587,8 @@ int sc_transmit_apdu(sc_card_t *card, sc_apdu_t *apdu)
 				 *      secure messaging is used */
 				plen          = max_send_size;
 				tapdu.cla    |= 0x10;
-				tapdu.le      = 0;
 				/* the intermediate APDU don't expect data */
```
xhanulik, line 590: Maybe rewriting the comment would prevent confusion

```suggestion
				/* the intermediate APDU don't expect response data */
```

xhanulik (commented):
Otherwise looks good to me.

src/libopensc/apdu.c
Jakuje: Thanks for the suggestion. Applied!

frankmorgner (approved):

PR #2830, opened 2023-08-11
Fixing OSS-Fuzz issues

xhanulik:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/muscle.c
+++ b/src/libopensc/muscle.c
@@ -627,8 +627,10 @@ int msc_extract_rsa_public_key(sc_card_t *card,
 	if(!*modulus) LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
 	memcpy(*modulus, buffer, *modLength);
 	*expLength = (buffer[*modLength] << 8) | buffer[*modLength + 1];
-	if (*expLength > sizeof buffer)
+	if (*expLength > sizeof buffer) {
+		free(*modulus); *modulus = NULL;
 		return SC_ERROR_OUT_OF_MEMORY;
+	}
 	r = msc_read_object(card, inputId, fileLocation, buffer, *expLength);
 	if(r < 0) {
 		free(*modulus); *modulus = NULL;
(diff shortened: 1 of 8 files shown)


src/pkcs15init/pkcs15-oberthur.c
```diff
@@ -637,24 +645,39 @@ cosm_generate_key(struct sc_profile *profile, struct sc_pkcs15_card *p15card,
...
 	LOG_TEST_RET(ctx, rv, "cosm_generate_key() CARDCTL_OBERTHUR_GENERATE_KEY failed");
 
 	/* extract public key */
 	pubkey->algorithm = SC_ALGORITHM_RSA;
 	pubkey->u.rsa.modulus.len   = key_info->modulus_length / 8;
 	pubkey->u.rsa.modulus.data  = malloc(key_info->modulus_length / 8);
-	if (!pubkey->u.rsa.modulus.data)
+	if (!pubkey->u.rsa.modulus.data) {
+		sc_file_free(prkf);
+		sc_file_free(tmpf);
```
Jakuje, line 668: missing `free(args.pubkey);` I think. Same in the following block.

Jakuje (approved):

src/libopensc/muscle.c
```diff
@@ -627,8 +627,10 @@ int msc_extract_rsa_public_key(sc_card_t *card,
 	if(!*modulus) LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
 	memcpy(*modulus, buffer, *modLength);
 	*expLength = (buffer[*modLength] << 8) | buffer[*modLength + 1];
-	if (*expLength > sizeof buffer)
+	if (*expLength > sizeof buffer) {
+		free(*modulus); *modulus = NULL;
```
frankmorgner, line 631: a line break would be nice, here.

frankmorgner (approved):
Does the security advisory need to be extended with this? AFAIKT, this only touches possible memory leaks.

frankmorgner:
@xhanulik , could you check Packit's RPM build? It seems to be failing consistently due to some dependency problem.

Jakuje:
> Does the security advisory need to be extended with this? AFAIKT, this only touches possible memory leaks.

Correct. The only recent security-related reports from oss-fuzz are already in verified state.

Jakuje:
> @xhanulik , could you check Packit's RPM build? It seems to be failing consistently due to some dependency problem.

Verca is away so I will have a look into this issue in separate PR.

PR #2831, opened 2023-08-14
According configure.ac minumum OpenSSL is 1.1.1

metsma:
##### Checklist

- [ ] macOS tokend is tested


Jakuje:
I see we have this minimum version already in the configure, but it is always hard to keep these in sync. Can you add a comment here and also in the configure to make sure these are updated together next time we will be bumping a openssl version?

https://github.com/OpenSC/OpenSC/blob/master/configure.ac#L24

metsma:
Updated commit

frankmorgner (approved):

Jakuje (approved):

PR #2834, opened 2023-08-14
ci: Fix building packit action and improve error reporting

Jakuje:
The `sudo` is not needed in the github action context anymore. The original issue was hidden by the fact the bash did not have options set to exit on errors so added also that part to unbreak Packit RPM action.


packaging/opensc.spec
```diff
@@ -120,6 +120,10 @@ rm -rf %{buildroot}%{_mandir}/man1/pkcs11-register.1*
 rm %{buildroot}%{_datadir}/applications/org.opensc.notify.desktop
 rm %{buildroot}%{_mandir}/man1/opensc-notify.1*
 
+# remove the onepin pkcs11 module as it is only a symlink
+rm %{buildroot}%{_libdir}/pkcs11/onepin-opensc-pkcs11.so
+rm %{buildroot}%{_libdir}/onepin-opensc-pkcs11.so
```
frankmorgner, line 125: I specifically added the symlink to keep old configurations working, but feel free to remove it in the RPM, if you think that's not a problem in your case.

frankmorgner (approved):

packaging/opensc.spec
Jakuje: The problem was how the symlink is created in when building the RPM. Its absolute symlink, which will point into the buildroot and will not work after the installation. I think it would be more suitable if the symlink would be relative to the original file, what do you think?

Jakuje: The linked page propose to use only
```
        cd $(DESTDIR)$(bindir) && \
          mv -f prog$(EXEEXT) prog-$(VERSION)$(EXEEXT) && \
          $(LN_S) prog-$(VERSION)$(EXEEXT) prog$(EXEEXT)
```
> Note that we rename the program so that a new version will erase the symbolic link, not the real binary. **Also we cd into the destination directory in order to create relative links.**

which is relative symlink while you use hard-links. I did not expect that to be an issue, but with this guideline, it is something that we should change too. I will update the PR as I think it is not worth separate PR.

PR #2838, opened 2023-08-17
Fix remaining fuzzing & coverity issues

Jakuje:
This should close recently reported oss-fuzz and coverity issues.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-setcos.c
+++ b/src/pkcs15init/pkcs15-setcos.c
@@ -364,6 +364,7 @@ setcos_create_key(sc_profile_t *profile, sc_pkcs15_card_t *p15card,
 		LOG_TEST_RET(ctx, r, "Failed to delete private key file");
 	} else if (r != SC_ERROR_FILE_NOT_FOUND) {
 		sc_file_free(file);
+		file = NULL;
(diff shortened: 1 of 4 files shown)


src/pkcs15init/pkcs15-setcos.c
```diff
@@ -349,22 +349,25 @@ setcos_create_key(sc_profile_t *profile, sc_pkcs15_card_t *p15card,
 
 	/* Replace the path of instantiated key template by the path from the object data. */
 	memcpy(&file->path, &key_info->path, sizeof(file->path));
+	if (file->path.len < 2) {
+		LOG_TEST_RET(ctx, SC_ERROR_INVALID_DATA, "Invalid path");
+	}
```
xhanulik, line 354: Shouldn't the `file` be also freed here?

Jakuje: Yes. And the check can be probably moved before the allocation to save the trouble here. Will fix!

xhanulik (approved):
lgtm

frankmorgner (approved):

PR #2844, opened 2023-08-25
Fix rest of memory leaks reported by OSS-Fuzz

xhanulik:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/iasecc-sdo.c
+++ b/src/libopensc/iasecc-sdo.c
@@ -204,7 +204,7 @@ iasecc_sdo_free(struct sc_card *card, struct iasecc_sdo *sdo)
 
 
 static int
-iasecc_crt_parse(struct sc_card *card, unsigned char *data, struct iasecc_se_info *se)
+iasecc_crt_parse(struct sc_card *card, unsigned char *data, size_t data_len, struct iasecc_se_info *se)
 {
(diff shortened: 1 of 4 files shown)


Jakuje (approved):

src/libopensc/iasecc-sdo.c
```diff
@@ -204,19 +204,26 @@ iasecc_sdo_free(struct sc_card *card, struct iasecc_sdo *sdo)
...
 	sc_log(ctx, "iasecc_crt_parse(0x%X) called", *data);
 
+	if (data_len < 2)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_DATA);
+
 	memset(&crt, 0, sizeof(crt));
 	crt.tag = *(data + 0);
 	len = *(data + 1);
+	if ((size_t) len != data_len - 2)
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_DATA);
```
Jakuje, line 222: does this work the way that the fixed block is parsed always or there can be several follow-up blocks to parse? I would say the second as it is called in cycle so I think this check is not correct.

xhanulik: Makes sense, thanks for review. Will fix

frankmorgner (approved):
thank you

PR #2845, opened 2023-08-28
Various fixes for ECDSA key operations with ePass 2003 token

Jakuje:
While setting up a token for automated CI, I noticed couple of failures that I am trying to fix here.

@FeitianSmartcardReader please, have a look. If the ePass should support also other digest operations with ECDSA, please point us to some documentation.

##### Checklist
- [ ] macOS tokend is tested


FeitianSmartcardReader:
let me take a look, thanks

Jakuje:
Dropped the commit ab533d0 removing the raw ECDSA mechanism as this is something I do not completely understand and just announce additional mechanism that works only with specific input sizes (SHA1 and SHA256). The others stayed and should land before the 0.24.0 as they cause memory issues.

frankmorgner (approved):

dengert:
> Dropped the commit https://github.com/OpenSC/OpenSC/commit/ab533d0b59270391c60bbba6d459ac40861f3602 removing the raw ECDSA mechanism as this is something I do not completely understand and just announce additional mechanism that works only with specific input sizes (SHA1 and SHA256). The others stayed and should land before the 0.24.0 as they cause memory issues.

By setting `flags = SC_ALGORITHM_ONBOARD_KEY_GEN | SC_ALGORITHM_ECDSA_HASH_SHA1 | SC_ALGORITHM_ECDSA_HASH_SHA256`  the driver is saying the card (or driver in this case) can do the SHA1 or SHA256 on whatever data is passed to it.

The driver sets `epass2003_ops.compute_signature = epass2003_decipher;`  which is strange but lines:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-epass2003.c#L2282-L2301
do the hash in the driver and set the SHA1 length to  0x14 or SHA256 len to 0x20. 
And the lines after: https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-epass2003.c#L2302-L2308 which would allow any length. So should this be removed since they also dropped SC_ALGORITHM_ECDSA_HASH_NONE?

By dropping the SC_ALGORITHM_ECDSA_RAW, upper layers in OpenSC will not do other hashes.

The card appears to only accept length 0x14 or 0x20 requiring strange code.

dengert:
My old epass2003 without this mods shows:
```
  ECDSA, keySize={256,256}, hw, sign, verify
  ECDSA-SHA224, keySize={256,256}, sign, verify
  ECDSA-SHA384, keySize={256,256}, sign, verify
  ECDSA-SHA512, keySize={256,256}, sign, verify
  ECDSA-SHA1, keySize={256,256}, hw, sign, verify
  ECDSA-SHA256, keySize={256,256}, hw, sign, verify
```

I would expect with this it mod it would only show:
```
  ECDSA-SHA1, keySize={256,256}, hw, sign, verify
  ECDSA-SHA256, keySize={256,256}, hw, sign, verify
```

The "hw" is misleading because the driver is doing the hash.

popovec:
@Jakuje can you please test this patch? This should allow ECDSA to be used with SHA384 and SHA512.

`````
diff --git a/src/libopensc/card-epass2003.c b/src/libopensc/card-epass2003.c
index ed2ec649b..2e1ff31a7 100644
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -2202,6 +2202,11 @@ epass2003_set_security_env(struct sc_card *card, const sc_security_env_t * env,
                        sbuf[2] = 0x92;
                        exdata->ecAlgFlags = SC_ALGORITHM_ECDSA_HASH_NONE;
                }
+               else if (env->algorithm_flags & SC_ALGORITHM_ECDSA_RAW)
+               {
+                       sbuf[2] = 0x92;
+                       exdata->ecAlgFlags = SC_ALGORITHM_ECDSA_HASH_NONE;
+               }
                else
                {
                        r = SC_ERROR_NOT_SUPPORTED;
`````
If we supplement epass2003_decipher() with the function of padding zeros before the shorter input, ECDSA with SHA224 will also work. I have only tested it in the simulator https://github.com/popovec/epass2003_sim/, as my epass2003 does not know anything about EC encryption.

popovec:
I have prepared a patch that fixes all ECDSA-SHA* and RAW EDCSA mechanisms. It also removes the incorrectly flagged "hw" attribute. Please try to check if it works on a real token, I only have a simulation.

https://github.com/popovec/OpenSC/tree/epass2003_ec_fix1

dengert:
Does anyone have a epass2003 with EC that could test  this PR?

My epass2003 is from Gooze in 2011 and does not support EC.

popovec:
As far as I know, @jakuje has the possibility to test it. The simulator test is available here:

https://github.com/popovec/epass2003_sim/actions/runs/6458752219

The error that is there comes from the diff of the reference and the new json file.

Jakuje:
Sorry, I will be able to test this manually only during last week of October. The most I can help with now is just to merge your changes and see what they will do in the gitlab CI, where I have the epass2003 plugged in. I do not have direct access to that machine now.

popovec:
I'll leave it up to you, I won't generate a PR from it yet, as I would hate to disrupt the current state of epass2003, which seems to work well at least for ECDSA-SHA1 and ECDSA-SHA256. However, it would be good if 0.24 final already contained functional code for ECDSA and ECDSA-SHA384 and ECDSA-SHA512.

PR #2848, opened 2023-08-29
sc_pkcs11_en/decrypt: Do not add to null pointer

kunkku:
These functions may be called with a NULL output buffer, which gets passed to the card driver. Now it is possible that NULL gets incremented and passed again to the driver in the finalization phase. This problem causing a segmentation fault was discovered by running `p11test` on a `myeid` card.

##### Checklist
- [ ] macOS tokend is tested


Jakuje (approved):

popovec (approved):

Jakuje:
Thank you!

PR #2849, opened 2023-08-29
p11test: Fix elliptic curve tests

kunkku:
This PR fixes various errors and eliminates unnecessary complexity mostly in elliptic curve–related code of `p11test`. In particular, it fixes:
* OpenSSL 3 support (apparently not tested earlier)
* memory leak when successfully saving an EC key (with OpenSSL 3)
* ECDH test case (mandatory attribute added)
* spelling errors

Testing was done with a `myeid` card.

##### Checklist

### DIFF
--- a/src/tests/p11test/p11test_case_common.c
+++ b/src/tests/p11test/p11test_case_common.c
@@ -557,16 +557,11 @@ int callback_public_keys(test_certs_t *objects,
 			EC_KEY_free(ec);
 #else
 			ctx = EVP_PKEY_CTX_new_from_name(0, "EC", 0);
-			unsigned char pubkey[80]; size_t pubkey_len = 0;
-			if (EVP_PKEY_get_octet_string_param(o->key, OSSL_PKEY_PARAM_PUB_KEY, pubkey, sizeof(pubkey), &pubkey_len) != 1) {
-				debug_print(" [WARN %s ] Cannot get params from EVP_PKEY", o->id_str);
-				goto ec_out;
-			}
 
 			const char *curve_name = OBJ_nid2sn(nid);
 			if (!(bld = OSSL_PARAM_BLD_new()) ||
 				OSSL_PARAM_BLD_push_utf8_string(bld, "group", curve_name, strlen(curve_name)) != 1 ||
-				OSSL_PARAM_BLD_push_octet_string(bld, "pub", pubkey, pubkey_len) != 1 ||
+				OSSL_PARAM_BLD_push_octet_string(bld, "pub", pub, pub_len) != 1 ||
 				!(params = OSSL_PARAM_BLD_to_param(bld))) {
 				debug_print(" [WARN %s ] Cannot set params from EVP_PKEY", o->id_str);
 				goto ec_out;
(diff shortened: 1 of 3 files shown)


src/tests/p11test/p11test_case_common.c
```diff
@@ -577,49 +551,50 @@ int callback_public_keys(test_certs_t *objects,
...
+	ec_out:
+		if (pub_asn1) ASN1_STRING_free(pub_asn1);
+		if (ecgroup) EC_GROUP_free(ecgroup);
+		if (ecpoint) EC_POINT_free(ecpoint);
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+		if (ctx) EVP_PKEY_CTX_free(ctx);
+		if (bld) OSSL_PARAM_BLD_free(bld);
+		if (params) OSSL_PARAM_free(params);
+		if (cert_group) EC_GROUP_free(cert_group);
+		if (cert_point) EC_POINT_free(cert_point);
```
Jakuje, line 593: The `if` should not be needed if the pointer is initially initialized to NULL as all of these function should correctly handle the NULL arguments (and when it was already freed, the `if` condition will not save you).

Jakuje (commented):
I did something similar in #2845 while making the ePass working with ECC. I did not compare it line by line, but looks reasonable.

Would you be interested in putting together some tests that could run in CI? We already have some MyEID coverage with the emulator, but it might make sense to run also the p11test on the MyEID emulator to be able to catch changes in behavior.

src/tests/p11test/p11test_case_common.c
kunkku: Unnecessary `if`s removed.

src/tests/p11test/p11test_case_common.c
```diff
@@ -585,39 +575,45 @@ int callback_public_keys(test_certs_t *objects,
...
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+		EVP_PKEY_CTX_free(ctx);
+		OSSL_PARAM_BLD_free(bld);
+		OSSL_PARAM_free(params);
+		EC_GROUP_free(cert_group);
+		EC_POINT_free(cert_point);
+#endif
+
+		if (ec_error) return -1;
+
```
Jakuje: please move the `int ec_error` definition to the beginning of the block. I was thinking how to make this less jumping as we are not much happy to use `goto`/labels in the new code. What would you say to the following suggestion that would use only one label?
```
		int ec_error = 1; /* initially */

		/* the code above with the `goto ec_out` jumps */

		ec_error = 0;

ec_out:
		EC_GROUP_free(ecgroup);
		EC_POINT_free(ecpoint);
		EVP_PKEY_CTX_free(ctx);
		OSSL_PARAM_BLD_free(bld);
		OSSL_PARAM_free(params);
		EC_GROUP_free(cert_group);
		EC_POINT_free(cert_point);
#endif

		if (ec_error)
			return -1;

```

Jakuje (commented):
just a nit regarding to the goto/label usage. Otherwise it looks good to me.

src/tests/p11test/p11test_case_common.c
kunkku: Excellent suggestion applied.

Jakuje (approved):
thank you! Looks good!

PR #2850, opened 2023-08-29
p11test: Fix myeid initialization

kunkku:
This PR fixes the `myeid` part of `runtest.sh` by adding the missing `$` signs to indicate shell variables. It also enables more versatile testing by
* allowing ECDH and
* adding another AES key to test key wrapping.

##### Checklist


Jakuje (approved):
Thanks! Not sure how I managed to commit it in this state.

Issue #2841, opened 2023-08-24
CVE-2021-34193 reference


abergmann:
## MITRE published CVE-2021-34193 
NVD Published Date: 08/22/2023

### Description
Stack overflow vulnerability in OpenSC smart card middleware before 0.23 via crafted responses to APDUs.

### Links
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27719
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28185  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28383  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28768  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28843  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28855  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29912  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30112  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30800  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31448  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31540  
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32149

### References
https://nvd.nist.gov/vuln/detail/CVE-2021-34193

abergmann:
Is this problem known upstream? And if yes, do we have a commit reference?

carnil:
Not upstream here, but looking at the "new" CVE I tired to get some mapping about the fixes, apart one I have not found a reference in the commits, but which seems still to indicate that the CVE is addressed in 0.22.0-rc1:

* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28185                                                           
  https://github.com/OpenSC/OpenSC/commit/f015746d22d249642c19674298a18ad824db0ed7 (0.22.0-rc1)
* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28843
  https://github.com/OpenSC/OpenSC/commit/1db88374bb7706a115d5c3617c6f16115c33bf27 (0.22.0-rc1)
* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28383
  https://github.com/OpenSC/OpenSC/commit/5df913b7f57ad89b9832555d24c08d23a534311e (0.22.0-rc1)
[...]

PR #2855, opened 2023-09-01
Added missing CVEs to NEWS

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2841

##### Checklist

- [x] Contact Mitre for removing CVE-2021-34193 as duplicate

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,13 @@
 NEWS for OpenSC -- History of user visible changes
 
+# Addendum for 0.22.0; 2023-09-01
+* fixed security problems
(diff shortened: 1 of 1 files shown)


Jakuje:
I would even mention CVE-2021-34193 in the news file with a note that it is a duplicate of these otherwise people will be asking and we will have to direct them to the issue with this discussion.
Otherwise looks good

frankmorgner:
done, thanks for the review

Jakuje (approved):

PR #2859, opened 2023-09-12
Fix new coverity issues + issues noticed while setting up sc-hsm token

Jakuje:
Fix coverity issues introduced with merge of #2053 and while testing a new sc-hsm token.

Please, take a look @CardContact and @dengert (I don't have the cards at hand now to test -- can you verify the the PIV changes work ok?).

##### Checklist

- [ ] macOS tokend is tested


dengert:
@Jakuje 
Building from master which included #2053 is working as expected.

Note that d43a199524728266c51c35b854d52c8f21f90bb1 which you had pushed on my branch updates coveritry.yml

Jakuje:
Seems like this will need some work as some changes seem to break the myeid unwrapping and I did not replace all occurences of the IDshlen.

dengert:
I do see that this PR makes changes to card-piv.c in https://github.com/OpenSC/OpenSC/pull/2859/commits/46d6993fb45fbb5493be4fc9ac9985c059725b60 

There are problems in   [46d6993fb45fbb5493be4fc9ac9985c059725b60 
](https://github.com/OpenSC/OpenSC/pull/2859/commits/46d6993fb45fbb5493be4fc9ac9985c059725b60#diff-14d802e467876e00ece49c8d6621ec74eaf7d02384642ac0694de83397d9bbfcR1524)

 if (buf == NULL || *buf = NULL) {

Jakuje:
> I do see that this PR makes changes to card-piv.c in [46d6993](https://github.com/OpenSC/OpenSC/commit/46d6993fb45fbb5493be4fc9ac9985c059725b60)
> 
> There are problems in [46d6993fb45fbb5493be4fc9ac9985c059725b60 ](https://github.com/OpenSC/OpenSC/pull/2859/commits/46d6993fb45fbb5493be4fc9ac9985c059725b60#diff-14d802e467876e00ece49c8d6621ec74eaf7d02384642ac0694de83397d9bbfcR1524)
> 
> if (buf == NULL || *buf = NULL) {

Fixed in current version (along with some more memory-handling issues mostly in pkcs15-init).

src/pkcs11/framework-pkcs15.c
```diff
@@ -865,7 +865,7 @@ __pkcs15_create_secret_key_object(struct pkcs15_fw_data *fw_data,
 
 	rv = __pkcs15_create_object(fw_data, &any_object,
 			object, &pkcs15_skey_ops, sizeof(struct pkcs15_skey_object));
-        skey = (struct pkcs15_skey_object *) any_object;
+	skey = (struct pkcs15_skey_object *)any_object;
 	if (rv >= 0)
 		skey->info = (struct sc_pkcs15_skey_info *) object->data;
```
xhanulik, line 870:
```suggestion
	skey = (struct pkcs15_skey_object *)any_object;
	if (rv >= 0)
		skey->info = (struct sc_pkcs15_skey_info *)object->data;
```

xhanulik (commented):
Nitpicks only, otherwise looks good.

src/pkcs11/framework-pkcs15.c
Jakuje: fixed

Jakuje:
I think this is good to go. There is one more memory leak that I was not able to remove so far:
```
==24841== 26,068 (22,208 direct, 3,860 indirect) bytes in 8 blocks are definitely lost in loss record 76 of 76
==24841==    at 0x484DA83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==24841==    by 0x54684C5: sc_pkcs15init_new_object (pkcs15-lib.c:3382)
==24841==    by 0x5468F6A: sc_pkcs15init_init_skdf (pkcs15-lib.c:1413)
==24841==    by 0x5470673: sc_pkcs15init_store_secret_key (pkcs15-lib.c:2075)
==24841==    by 0x5314B6A: pkcs15_create_secret_key (framework-pkcs15.c:2615)
==24841==    by 0x53174DF: pkcs15_create_object (framework-pkcs15.c:3044)
==24841==    by 0x52FE8F5: sc_create_object_int (pkcs11-object.c:146)
==24841==    by 0x5301D6D: C_DeriveKey (pkcs11-object.c:1394)
==24841==    by 0x1132EB: pkcs11_derive (p11test_case_ec_derive.c:60)
==24841==    by 0x113E54: test_derive (p11test_case_ec_derive.c:359)
==24841==    by 0x1140CA: derive_tests (p11test_case_ec_derive.c:402)
```
any ideas welcomed.

dengert:
Looks like:
https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L5308-L5314
was never finished. 

For example: 
https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L3909-L3917
frees object specific data.

Jakuje:
> Looks like: https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L5308-L5314 was never finished.
> 
> For example: https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L3909-L3917 frees object specific data.

Sounds like that. I was digging into that, trying to free some members, but what worked for SC-HSM was crashing on double free for OsEID so after some hours testing back and forth I gave this up.

src/libopensc/pkcs15-prkey.c
```diff
@@ -582,6 +578,14 @@ sc_pkcs15_free_prkey(struct sc_pkcs15_prkey *key)
 	}
```
frankmorgner, line 578: please add `sc_mem_clear(key, sizeof *key);`. This may resolve a possible double free. (This is already present in `sc_pkcs15_erase_pubkey()`)

frankmorgner:
I pushed my suggestion and the skey-change you were playing around a bit, I guess. I really think the latter is needed and I'll try to resolve the valgrind issues from CI.

Jakuje:
> I pushed my suggestion and the skey-change you were playing around a bit, I guess. I really think the latter is needed and I'll try to resolve the valgrind issues from CI.

That is exactly what I was trying in the past and that is crashing the oseid test now for some reason :) If you run it with valgrind, you should be able to the backtraces if you add `valgrind` to the lines `.github/test-oseid.sh` in `.github/workflows/linux.yml`.

Jakuje:
Seems like I finally came up to solution that sounds like working. Force-pushing on top of current changes and adding a fix to more recent coverity issues.

src/pkcs11/framework-pkcs15.c
```diff
@@ -5310,7 +5310,16 @@ struct sc_pkcs11_object_ops pkcs15_profile_ops = {
 static void
 pkcs15_skey_release(void *object)
 {
-	__pkcs15_release_object((struct pkcs15_any_object *) object);
+	struct pkcs15_skey_object *skey = (struct pkcs15_skey_object *)object;
+	struct sc_pkcs15_skey_info *skey_info = skey->info;
+	struct sc_pkcs15_object *p15obj = skey->skey_p15obj;
+
+	if (__pkcs15_release_object((struct pkcs15_any_object *)skey) == 0) {
+		if (p15obj->session_object) {
```
frankmorgner, line 5318: I really wonder why checking for session_object here is necessary. since skey_info is either NULL or created with calloc, which in both cases means that calling `free` on the pointer or the members of the data structure should be OK.

src/pkcs11/framework-pkcs15.c
```diff
@@ -5310,7 +5310,16 @@ struct sc_pkcs11_object_ops pkcs15_profile_ops = {
...
+			sc_pkcs15_free_skey_info(skey_info);
+			free(p15obj);
```
frankmorgner, line 5320: Why is this step needed? skey_p15obj is a define for base.p15_object just like, for example, prv_p15obj, but none of the other are calling `free` on base.p15_object in their release functions...

frankmorgner (commented):
Nice to see this working, but I don't quite understand the difference in workflow...

src/pkcs11/framework-pkcs15.c
Jakuje: I tried to describe this in the commit message. The `sc_pkcs15init_store_secret_key()` checks the `session_object` flag and if it is not set, it calls `sc_pkcs15init_add_object()`, which adds the p15 object to the directory structure that is freed during cleanup. Otherwise the structure is referenced only from the pkcs11 layer and need to be freed from that side.

The confusing thing with the skeys is that the skey_info

Looking at it more, we might also need some tweaks form the cards without onboard session objects (`SC_CARD_CAP_ONCARD_SESSION_OBJECTS` -- I think sc-hsm might be a bit different there and I probably did not check well).

src/pkcs11/framework-pkcs15.c
Jakuje: The answer here is basically the same as in the above comment -- because it is not stored in the p15card's directory structure so it can not be freed from there.

Initially I considered using some reference counter, but unfortunately the `sc_pkcs15_object` does not have reference counter so I did it this ugly way.

frankmorgner:
@CardContact , do you have any objections?

Jakuje:
The sc-hsm changes are quite minimal.

I pushed two more oss-fuzz issues that I missed last time going through the list. And skipped the installation of debuginfo if not needed as it is quite flaky.

After we will merge this, I think we should send out the rc ...

xhanulik (approved):
The fixes for OSS-Fuzz looks good.

Issue #2670, opened 2022-12-19
Keep up with the libressl releases


Jakuje:
We have fixed LibreSSL version hardcoded in our tests scripts. They are releasing a new versions infrequently and our workarounds sometimes stop working as last time fixed in the below PR.

It would be great to implement some automated check for the latest version in the script to find out the issues earlier than after the release.

> add some script to check what is the latest file in the https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/ directory to notify us about newest version?

If you think that could be quickly implemented, I'm happy to dump the stable release for an automated solution.

_Originally posted by @frankmorgner in https://github.com/OpenSC/OpenSC/issues/2665#issuecomment-1355146399_

dengert:
One way to do this: git clone or pull their master branch then list tags and checkout that latest tag. 
 https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.7.0-relnotes.txt 11-Dec-2022 15:04 says:
"We have released LibreSSL 3.7.0, which will be arriving in the
LibreSSL directory of your local OpenBSD mirror soon. This is a
development release for the 3.7.x branch, and we appreciate additional testing
and feedback before the final release coming soon with OpenBSD 7.1."

But tag 3.7.0 in not yet  in "git@github.com:libressl-portable/portable.git" with 3.6.1 being the latest release. 
But latest commit https://github.com/libressl-portable/portable/commit/320501da40a6b43965775a68b5019d86a146e21a
was updated yesterday 2022-12-17 So it may contain 3.7.0 but without a tag for it.

PR #2866, opened 2023-09-19
Bump libressl to 3.8.1

Jakuje:
For now, just to check if it works as we do not have any automated mechanism to keep track of libressl releases (#2670).

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
ok, this seems like working ok. I just wanted to make sure before the release.

PR #2874, opened 2023-09-22
Add badges for more cards covered in CI

Jakuje:
I plugged SC-HSM and ePass tokens to my old machine and the CI on gitlab now runs on them. This PR adds them to readme.

Machine is slow so it will take some time to finish (and valgrind will likely fail). The new machine with to test more cards is on the way.

Any ideas what more should be covered in the p11tests appreciated.

##### Checklist


frankmorgner (approved):

frankmorgner (approved):

Jakuje:
I did not add a links to the readme, but it should be these two. I will certainly add more full names to the release testing page:
 * https://www.ftsafe.com/store/product/epass2003-pki-token/
 * https://www.cardomatic.de/en/p/smartcard-hsm-4k-usb-token

not sure if we want to link these (or others) from the main page directly.

Before merging, I will probably disable the valgrind jobs.

frankmorgner:
Did you take a look at the reported memory leak?

I would only add the names of the card so that users get a feeling of what was tested. Smart card shops could be listed here https://github.com/OpenSC/Wiki/blob/31d5516dbc01897d573573f437cad501fc29d673/Frequently-Asked-Questions.textile#where-can-i-buy-smart-cards-

dengert:
Amozon.com in US and amazon.co.uk  and maybe other EU countries sells some smart cards too.

Jakuje:
I think this one is ready now too. The ePass is still failing (there is outstanding PR), the SC-HSM should pass now.

PR #2876, opened 2023-09-23
Run coverity scan only if the secret token is defined

LudovicRousseau:
No need to try and fail on forks of OpenSC repository.

Without this patch the github action coverity.yml (in my fork repo) fails with:
```
Username/Password Authentication Failed.
Error: Process completed with exit code 6.
```

With no objection I plan to merge this patch.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -12,6 +12,9 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


Jakuje (approved):
looks reasonable

LudovicRousseau:
Argh! My patch is incorrect and the action fails.

I am working on a fix.

LudovicRousseau:
bug fixed in https://github.com/OpenSC/OpenSC/commit/b9529d944537f8c1e40b798909b71517a609c870

Issue #2877, opened 2023-09-25
IDPrime 940: Failed to connect to card: Invalid arguments


dannydotpoint:
On a freshly initialized IDPrime 940 I'm seeing:

```
$ pkcs15-tool -D
Using reader with a card: Alcor Micro AU9540 00 00
Failed to connect to card: Invalid arguments
```

The problem occurs on opensc 0.24.0-rc1. On previous versions (opensc 0.22 and 0.23) I  used to see:

```
$  pkcs15-tool -D
Using reader with a card: Alcor Micro AU9540 00 00
PKCS#15 Card [My Token                       ]:
	Version        : 0
	Serial number  : a7c000cbf5a83879a7c000cbf5a83879
	Manufacturer ID: Gemalto
	Flags          : 


PIN [PIN]
	Object Flags   : [0x01], private
	ID             : 11
	Flags          : [0x30], initialized, needs-padding
	Length         : min_len:4, max_len:16, stored_len:16
	Pad char       : 0x00
	Reference      : 17 (0x11)
	Type           : ascii-numeric
```

In debug mode the issue appears to originate in card-idprime.c:
```

P:378900; T:0x139805394267008 17:48:35.178 [pkcs15-tool] card-idprime.c:581:idprime_init: Index file found
P:378900; T:0x139805394267008 17:48:35.178 [pkcs15-tool] card-idprime.c:385:idprime_process_index: called
P:378900; T:0x139805394267008 17:48:35.179 [pkcs15-tool] card-idprime.c:388:idprime_process_index: returning with: -1300 (Invalid arguments)
P:378900; T:0x139805394267008 17:48:35.179 [pkcs15-tool] card-idprime.c:586:idprime_init: returning with: -1300 (Invalid arguments)
P:378900; T:0x139805394267008 17:48:35.180 [pkcs15-tool] card.c:360:sc_connect_card: driver 'Gemalto IDPrime' init() failed: Invalid arguments
P:378900; T:0x139805394267008 17:48:35.181 [pkcs15-tool] reader-pcsc.c:675:pcsc_disconnect: Alcor Micro AU9540 00 00:SCardDisconnect returned: 0x00000000
P:378900; T:0x139805394267008 17:48:35.181 [pkcs15-tool] card.c:403:sc_connect_card: returning with: -1300 (Invalid arguments)
Failed to connect to card: Invalid arguments
```

[...]

PR #2878, opened 2023-09-27
Minor fixes for Idprime 940

Jakuje:
Fixes: #2877

The keyrefmap was mandatory to start processing index file, which was a regression since 0.23.0, where users reported the card worked for them already.

Additionally, the signature pin was label was not propagated to the PKCS#15 objects and only printed in debug messages.

##### Checklist

- [ ] macOS tokend is tested


xhanulik (approved):
Looks good, thanks for fixing.

PR #2882, opened 2023-09-29
MinGW-w64 cross-compile fixes

llogar:
A series of trivial patches that fixes cross-compiling OpenSC (tested on Ubuntu 22.04) for Windows.
It looks like this hasn't been tested for a while if at all?


frankmorgner (approved):

Jakuje (approved):
We have mingw builds as part of CI. Are these something that can be caught earlier as part of CI, either by adjusting the pipelines, adding new ones or enabling some build flags?

Jakuje:
I think whatever works. I am not an legal/license expert but the header sounds to me like "do whatever you want"-type of license. But if it can be pulled during the CI/build time, it works for me.

Jakuje (approved):

frankmorgner:
I think keeping a copy of cardmod.h should be OK.

PR #2391, opened 2021-08-18
Disable old card drivers and PKCS#15 emulators

xhanulik:
This PR disables old and recently unmodified card drivers (`card-*.c`) and their counterpart PKCS#15 emulators (`pkcs15-*.c`), and adds boilerplate for disabling old emulators.

Developer activity below tracks active developing and maintaining a card driver/emulator, excluding generic changes and fuzzing/static analysis fixes:
- `card-akis.c`: modified [2007](https://github.com/OpenSC/OpenSC/commit/f9476144182dcc1568518f436ec8e5368841902a)
- `card-asepcos.c`: modified [2010](https://github.com/OpenSC/OpenSC/commit/02c35be138d67b290e0a3fe239a1db739c1c6fe3)
- `card-atrust-acos.c` (`pkcs15-atrust-acos.c`): modified [2008](https://github.com/OpenSC/OpenSC/commit/a4bad4452e7d6acdb75c129fa28c5291f4606b79)
- `card-flex.c`: modified [2010](https://github.com/OpenSC/OpenSC/commit/7d935df1bc65022ef80a40f8721f0fa8e3709289)
- `card-gpk.c` (`pkcs15-gemsafeGPK.c`): modified [2007](https://github.com/OpenSC/OpenSC/commit/a8908b8548e02d320376844b3d9668f0f89b3c29)
- `card-incrypto34.c`: modified [2007](https://github.com/OpenSC/OpenSC/commit/a2f622a21521cb350541894d80e3266b5f2f5612)
- `card-westcos.c` (`pkcs15-westcos.c`): modified [2010](https://github.com/OpenSC/OpenSC/commit/c3de15d2d08061cd6b2a0fabbdaaa7b8a6ede1fa)
- `pkcs15-actalis.c` and `pkcs15-tccardos.c`: Emulators detect card according to card name, which was changed in `card-cardos.c` in eee4964 (2017) and they were not updated since this change accordingly, which implies no latest use.

There was no recent user activity (issues, bug reports) related to the mentioned cards.
Disabled drivers and emulators can be enabled in `opensc.conf`.
Filtering of PKCS#15 emulators in `pkcs15-emulator-filter.c` enables emulators according to `opensc.conf` file.

##### Checklist

- [x] Documentation is added or updated
- [x] New files have a LGPL 2.1 license statement

Issue #2832, opened 2023-08-14
Athena ASE III gets CKR_TOKEN_NOT_RECOGNIZED


manu0401:
### Problem Description

Athena ASE IIIe USB token ceased to work  between OpenSC 0.22.0 and 0.23.0

With OpenSC 0.23.0, pkcs11-tool -vL gets
```
pkcs11-tool -L gets
Available slots:
Slot 0 (0x0): Athena ASE IIIe 00 00
  (token not recognized)
```
With prior OpenSC the token works. pkcs11-tool -vL says
```
Available slots:
Slot 0 (0x0): Athena ASE IIIe 00 00
  manufacturer:  Athena
  hardware ver:  16.2
  firmware ver:  0.0
  flags:         token present, removable device, hardware slot
  token label        : OpenSC Card (manu)
  token manufacturer : OpenSC Project
  token model        : PKCS#15
  token flags        : login required, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 0105523190192129
  pin min/max        : 4/16
```
### Proposed Resolution

This change caused the regression: https://github.com/OpenSC/OpenSC/commit/f44564b2e9cdc9baf67914f1e5b7c2da6f007862

Moving back asepcos from old_card_drivers to internal_card_drivers  fixes the problem

```
--- src/libopensc/ctx.c.orig    2023-08-14 10:34:40.772689358 +0200
+++ src/libopensc/ctx.c 2023-08-14 10:34:47.435926078 +0200
@@ -100,8 +100,9 @@
 static const struct _sc_driver_entry internal_card_drivers[] = {
        { "cardos",     (void *(*)(void)) sc_get_cardos_driver },
        { "cyberflex",  (void *(*)(void)) sc_get_cyberflex_driver },
        { "gemsafeV1",  (void *(*)(void)) sc_get_gemsafeV1_driver },
+       { "asepcos",    (void *(*)(void)) sc_get_asepcos_driver },
        { "starcos",    (void *(*)(void)) sc_get_starcos_driver },
        { "tcos",       (void *(*)(void)) sc_get_tcos_driver },
 #ifdef ENABLE_OPENSSL
        { "oberthur",   (void *(*)(void)) sc_get_oberthur_driver },
@@ -157,9 +158,8 @@
 };
 
 static const struct _sc_driver_entry old_card_drivers[] = {
        { "akis",       (void *(*)(void)) sc_get_akis_driver },
[...]

PR #2885, opened 2023-10-03
Removal of old card drivers

xhanulik:
This PR proposes a removal of old card drivers with no recent user or developer activity. All of the following drivers were disabled in #2391.
The process of disabling and removing of old drivers is described on [wiki](https://github.com/OpenSC/OpenSC/wiki/Removal-of-Old-Card-Drivers).

The discussed drivers are:
- `card-akis.c`, last developer modification f947614 (2007),
- `card-flex.c`, 7d935df (2010),
- `card-gpk.c`, a8908b8 (2007),
- `card-incrypto34.c`, a2f622a (2007),
- `card-westcos.c`, c3de15d (2010).

The decision regarding whether all or some of the drivers should be removed, or if they should remain disabled, is open for discussion.


frankmorgner:
Before removing the driver completely, we should put it into the list of old card drivers first.

Also, for GPK there has recently been user activity (https://github.com/OpenSC/OpenSC/issues/2832), so we should still keep this in the list of old drivers

xhanulik:
> Before removing the driver completely, we should put it into the list of old card drivers first.

All of the mentioned drivers were disabled in #2391, so they are on the list of old card drivers:

https://github.com/OpenSC/OpenSC/blob/33dd3a237225d1adf65a9c503fe81d5f6b3b8649/src/libopensc/ctx.c#L170-L179

> Also, for GPK there has recently been user activity (#2832), so we should still keep this in the list of old drivers

I think this issue was regarding the disabled `asepcos` driver, which is not part of this PR (and the user activity should be already updated on wiki).

frankmorgner:
Thanks for the clearification. However, it looks like `card-flex.c` was still reachable via the "`cyberflex`" driver, because only the predecessor Cryptoflex (`flex`) was moved to the old drivers previously. I think it would be nice to keep both in the list of old drivers for an other release cycle. The removal of the other drivers could be scheduled when the final release of 0.24.0 is out.

xhanulik:
Alright, I will move `cyberflex` into old drivers then.

frankmorgner (approved):
LGTM

libopensc will be ABI incompatible to previous versions due to the change in defines, we need to make sure to change the library version accordingly before a release.

Jakuje (approved):

frankmorgner:
> Good point! We should probably note it somewhere or start a draft release issue/pr with the notes so it wont get lost.

I extended the note for the release how to in the wiki, I think that should be enough:
>configure.ac : Update the [LT version number](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html), which is required with changes to, for example, opensc.h and libopensc.exports.

> Just to be clear, are we good to include this with the release in 2 months? I would say yes and lets merge this soon so we can get feedback early if somebody will miss these.

Yes, good for me.

PR #2888, opened 2023-10-03
Add printout of signRecover and verifyRecover key usages to pkcs11-tool

### LINKED ISSUE #2886: pkcs11-tool: Missing printout of verifyRecover and signRecover key usages
### Problem Description
pkcs11-tool does not print out signRecover or verifyRecover key usages for keys. On the other hand, pkcs15-tool does.

### Proposed Resolution
Add printout of signRecover and verifyRecover to pkcs11-tool.

### Steps to reproduce
Public key on smart card printed with pkcs15-tool (verifyRecover shown):

```
$ pkcs15-tool --list-public-keys
Public RSA Key [My Label]
        Usage          : [0xD0], wrap, verify, verifyRecover
        Access Flags   : [0x00]
        ModLength      : 2048
        Key ref        : 0 (0x00)
        Native         : no
        Path           : 3f0050155501
        ID             : 0bed4e
```

Same public key on the same smart card printed with pkcs11-tool (verifyRecover not shown):

```
$ pkcs11-tool --list-objects --type pubkey
Public Key Object; RSA 2048 bits
  label:      My Label
  ID:         0bed4e
  Usage:      verify, wrap
  Access:     none
```

Same problem exist for the private key but with signRecover missing instead.

johanhedin:
Fixes #2886

Card used during testing:

```
$ opensc-tool -n
Using reader with a card: ACS ACR 38U-CCID 00 00
MyEID 4.5.5
```

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -5205,7 +5205,7 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 		printf("%ssign", sepa);
 		sepa = ", ";
 	}
-	if ((!pub || sec) && getSIGN_RECOVER(sess, obj)) {
+	if (!pub && getSIGN_RECOVER(sess, obj)) {
 		printf("%ssignRecover", sepa);
 		sepa = ", ";
 	}
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -674,7 +674,9 @@ ATTR_METHOD(MODIFIABLE, CK_BBOOL);			/* getMODIFIABLE */
 ATTR_METHOD(ENCRYPT, CK_BBOOL);				/* getENCRYPT */
 ATTR_METHOD(DECRYPT, CK_BBOOL);				/* getDECRYPT */
 ATTR_METHOD(SIGN, CK_BBOOL);				/* getSIGN */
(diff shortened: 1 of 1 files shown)


frankmorgner (approved):

src/tools/pkcs11-tool.c
```diff
@@ -5203,6 +5205,10 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 		printf("%ssign", sepa);
 		sepa = ", ";
 	}
+	if ((!pub || sec) && getSIGN_RECOVER(sess, obj)) {
```
Jakuje, line 5208: the `SIGN_RECOVER` is not defined for secret keys (unlike the SIGN, which is):
```suggestion
	if (!pub && getSIGN_RECOVER(sess, obj)) {
```

Jakuje (changes requested):

src/tools/pkcs11-tool.c
Jakuje: OK, actually its defined for couple of  `CKM_SEED_*` mechanisms but I hope nobody implements these so I would propose to not print this for generic secret keys.

http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cs01/pkcs11-curr-v2.40-cs01.html#_Toc399399104

johanhedin: Thank you for the clarification! Change committed.

Jakuje (approved):

PR #2893, opened 2023-10-06
More memory issues from pkcs15init fuzzer

Jakuje:
##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-entersafe.c
+++ b/src/libopensc/card-entersafe.c
@@ -31,15 +31,15 @@
 #include "cardctl.h"
 
 static const struct sc_atr_table entersafe_atrs[] = {
-	{ 
-		 "3b:0f:00:65:46:53:05:19:05:71:df:00:00:00:00:00:00", 
-		 "ff:ff:ff:ff:ff:ff:ff:00:ff:ff:ff:00:00:00:00:00:00", 
+	{
+		 "3b:0f:00:65:46:53:05:19:05:71:df:00:00:00:00:00:00",
+		 "ff:ff:ff:ff:ff:ff:ff:00:ff:ff:ff:00:00:00:00:00:00",
 		 "ePass3000", SC_CARD_TYPE_ENTERSAFE_3K, 0, NULL },
-	{ 
+	{
 		 "3b:9f:95:81:31:fe:9f:00:65:46:53:05:30:06:71:df:00:00:00:80:6a:82:5e",
 		 "FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:00:FF:FF:FF:FF:FF:FF:00:00:00:00",
 		 "FTCOS/PK-01C", SC_CARD_TYPE_ENTERSAFE_FTCOS_PK_01C, 0, NULL },
-	 { 
+	 {
 		"3b:fc:18:00:00:81:31:80:45:90:67:46:4a:00:64:18:14:00:00:00:00:02",
 		"ff:00:00:00:00:00:00:00:00:ff:ff:ff:ff:00:00:00:00:ff:ff:ff:ff:00",
 		"EJAVA/PK-01C", SC_CARD_TYPE_ENTERSAFE_EJAVA_PK_01C, 0, NULL },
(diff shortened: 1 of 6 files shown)


Jakuje:
The https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61965 is not yet completely fixed as I do not follow where the leak gets lost (but there were too so I got rid at least of one so far).

src/pkcs15init/pkcs15-oberthur.c
```diff
@@ -625,20 +625,25 @@ cosm_generate_key(struct sc_profile *profile, struct sc_pkcs15_card *p15card,
...
 	rv = sc_select_file(p15card->card, &key_info->path, &prkf);
 	LOG_TEST_RET(ctx, rv, "Failed to generate key: cannot select private key file");
 
 	/* In the private key DF create the temporary public RSA file. */
 	rv = cosm_get_temporary_public_key_file(p15card->card, prkf, &tmpf);
-	if (rv != SC_SUCCESS)
+	if (rv != SC_SUCCESS) {
 		sc_file_free(prkf);
-	LOG_TEST_RET(ctx, rv, "Error while getting temporary public key file");
+		sc_file_free(tmpf);
```
xhanulik, line 637: I think this `sc_file_free` is not needed here. When error occurs in `cosm_get_temporary_public_key_file`, the allocated file should be freed there and `tmpf` remains `NULL`.

Jakuje: Right. Removed that line. Thanks!

src/libopensc/card-entersafe.c
```diff
@@ -1418,40 +1419,49 @@ static int entersafe_gen_key(sc_card_t *card, sc_entersafe_gen_key_data *data)
...
-
-	p=rbuf;
-	if (*p!='E') {
-		free(data->modulus);
-		data->modulus = NULL;
+	p = rbuf;
+	plen = apdu.resplen;
+	if (* p!= 'E') {
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_VERBOSE, SC_ERROR_INVALID_DATA);
+	}
```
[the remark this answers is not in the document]
Jakuje: This is mostly matter of taste. But without braces, adding another statement breaks (fortunately, compilers can detect this issue quite well these days), so I prefer adding braces everywhere. So I left that.
But noticed that there is misplaced space between asterisk and `p`, which should have come after `p` which I fixed.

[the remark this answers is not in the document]
xhanulik: Sure, I just noticed when scrolling through the file.

xhanulik (approved):

Jakuje:
@vjardin do you still have the iasecc card? There are some modifications fixing issues reported by fuzzing. It would be great if you could give it a test to make sure they make sense and keep working (the same for the 0.24.0-rc1)

src/libopensc/iasecc-sdo.c
```diff
@@ -143,20 +143,22 @@ iasecc_sdo_convert_acl(struct sc_card *card, struct iasecc_sdo *sdo,
...
-		free(sdo->docp.non_repudiation.value);
-	if (sdo->docp.acls_contact.value)
-		free(sdo->docp.acls_contact.value);
-	if (sdo->docp.size.value)
-		free(sdo->docp.size.value);
-	if (sdo->docp.name.value)
-		free(sdo->docp.name.value);
-	if (sdo->docp.issuer_data.value)
-		free(sdo->docp.issuer_data.value);
+	free(sdo->docp.tries_maximum.value);
```
frankmorgner, line 146: In iasecc_sdo_allocate_and_parse, struct iasecc_sdo is created with calloc(). Since you're adding resets to NULL to some members, you may want to extend this to the whole structure by using memset(sdo, 0, sizeof *sdo);. This would also destroy the `magic` value, which I think is a good thing, because the structure is not valid anymore.

Jakuje:
Rebased and addressed last comments

Jakuje:
Frank, is this good to go to 0.24.0?

frankmorgner (approved):

frankmorgner:
Yes, thank you.

PR #2898, opened 2023-10-12
Sets the 'hw' flag for the 'RSA-PKCS-KEY-PAIR-GEN'

popovec:
If the card announces an RSA key generation operation, let's also announce it in the corresponding flag for the RSA-PKCS-KEY-PAIR-GEN mechanism.


popovec:
This patch will not pass the test for the following cards:

test-isoapplet-v1
test-isoapplet-v0
test-openpgp

It can be fixed either by changing the relevant json files or by editing `src/tests/p11test/p11test_case_mechs.c` .  

We currently have a difference when listing the mechanisms if they have and do not have the HW flag:
`````
        [
                "EC_KEY_PAIR_GEN",
                "256",
                "256",
                "0x00010001"
        ],
`````
`````
        [
                "RSA_PKCS_KEY_PAIR_GEN",
                "512",
                "2048",
                "CKF_GENERATE_KEY_PAIR"
        ]],
`````
For this reason, I would rather fix `src/tests/p11test/p11test_case_mechs.c` to include the HW flag in the output.

Alternatively, does anyone have a suggestion for another solution?

**Comment by Jakuje** [comment]
Good catch! I think this should certainly be in! The p11test_case_mechs is kind of dummy, where the last field are the flags (if only one is set, it is expanded to its name if I remember well) so my proposal would be just updating the reference json files, but if you are up to some larger changes in p11test, I am all ears :)

**Comment by popovec** [comment]
One more test needs to be completed in  b7e2581:
`````
+ if (strlen(flag_buffer))
	flag_buffer[strlen(flag_buffer) - 1] = '\0';
`````
but before I complete it, I have other things that I noticed about CKF_HW.

For the MyEID/OsEID card, I have the `hw` flag for both RSA_PKCS_OAEP and RSA_PKCS_PSS, the same happens with the epass2003 token.  This information is incorrect. I think that it will be enough for the correction:
`````
index 3988c062d..3e679d3d7 100644
@@ -6607,7 +6607,9 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
 
        if (rsa_flags & SC_ALGORITHM_RSA_PAD_PSS) {
                CK_FLAGS old_flags = mech_info.flags;
-               mech_info.flags &= ~(CKF_DECRYPT|CKF_ENCRYPT);
+               mech_info.flags &= ~(CKF_HW|CKF_DECRYPT|CKF_ENCRYPT);
+               if ((card->caps & SC_ALGORITHM_RSA_PAD_PSS) == SC_ALGORITHM_RSA_PAD_PSS)
+                       mech_info.flags |= CKF_HW;
                mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS_PSS, &mech_info, CKK_RSA, NULL, NULL, NULL);
                rc = sc_pkcs11_register_mechanism(p11card, mt, &registered_mt);
                sc_pkcs11_free_mechanism(&mt);
@@ -6649,7 +6651,9 @@ register_mechanisms(struct sc_pkcs11_card *p11card)
 
        if (rsa_flags & SC_ALGORITHM_RSA_PAD_OAEP) {
                CK_FLAGS old_flags = mech_info.flags;
-               mech_info.flags &= ~(CKF_SIGN|CKF_VERIFY|CKF_SIGN_RECOVER|CKF_VERIFY_RECOVER);
+               mech_info.flags &= ~(CKF_HW|CKF_SIGN|CKF_VERIFY|CKF_SIGN_RECOVER|CKF_VERIFY_RECOVER);
+               if ((card->caps & SC_ALGORITHM_RSA_PAD_OAEP) == SC_ALGORITHM_RSA_PAD_OAEP)
+                       mech_info.flags |= CKF_HW;
                mt = sc_pkcs11_new_fw_mechanism(CKM_RSA_PKCS_OAEP, &mech_info, CKK_RSA, NULL, NULL, NULL);
                rc = sc_pkcs11_register_mechanism(p11card, mt, NULL);
                sc_pkcs11_free_mechanism(&mt);
`````

File: src/tests/p11test/p11test_case_common.c [hunk 1]
```diff
@@ -1213,11 +1213,23 @@ const char *get_mechanism_flag_name(int mech_id)
 		case CKF_EC_ECPARAMETERS:
 			return "CKF_EC_ECPARAMETERS";
 		default:
-			sprintf(flag_buffer, "0x%.8X", mech_id);
-			return flag_buffer;
+			return "unknown";
```
Jakuje, line 1216: I think we have most of the flags already covered now, but if there is some new, this hugely simplified adding the missing flag, while now, the word "unknown" wont be much useful. I would propose to keep the old behavior (while it will need a new static buffer).

src/tests/p11test/p11test_case_common.c
```diff
@@ -1213,11 +1213,23 @@ const char *get_mechanism_flag_name(int mech_id)
...
+const char *
+get_mechanism_all_flag_name(int mech_id)
+{
+	CK_FLAGS j;
+
+	flag_buffer[0] = '\0';
+	for (j = 1; j <= CKF_EC_COMPRESS; j = j << 1)
+		if ((mech_id & j) != 0)
+			snprintf(flag_buffer + strlen(flag_buffer), sizeof(flag_buffer) - strlen(flag_buffer), "%s,", get_mechanism_flag_name(j));
+	flag_buffer[strlen(flag_buffer) - 1] = '\0';
```
Jakuje, line 1229: I think this line is not needed. The `snprintf` always terminate the string with null byte so here you only override that NULL byte (as `strlen` already searches for the NULL byte).

Jakuje (commented):
Thank you for the work on this! There is just two minor nits.

Regarding the CKF_HW flags, the problem is that in case of combined mechanisms (Digest&sign, PSS, OAEP, ...), it is not clear whether all of the parts need to be done in HW to have the CKF_HW flag. This was discussed in the past and the consensus was, as far as I remember, that if the the private key operation is with HW, the HW flag should be there.

popovec:
Ok, in this PR I will no longer comment on the HW flag for PSS and OAEP.   The `get_mechanism_flag_name()` function  remains unchanged, the new `get_mechanism_all_flag_name()` function uses a local static buffer.

The output  of the `p11test` has changed slightly, the flags are listed as hexadecimal number and at the same time its name is listed for each known mechanism:
`````
[RSA_X_509            ] [ 512][2048] [0x00002A01] CKF_HW,CKF_DECRYPT,CKF_SIGN,CKF_VERIFY
[RSA_PKCS             ] [ 512][2048] [0x00002A01] CKF_HW,CKF_DECRYPT,CKF_SIGN,CKF_VERIFY
[SHA1_RSA_PKCS        ] [ 512][2048] [0x00002800] CKF_SIGN,CKF_VERIFY
`````
The json file is also slightly modified:
`````
                "ECDSA_SHA256",
                "256",
                "256",
-               "0x00002801"
+               "CKF_HW,CKF_SIGN,CKF_VERIFY"
`````
Wouldn't it be better to use output similar to what goes to stdout in the "json" output?
`````
                "ECDSA_SHA256",
                "256",
                "256",
-               "0x00002801"
+               "0x00002801 CKF_HW,CKF_SIGN,CKF_VERIFY"
`````

Jakuje (approved):

Jakuje:
> Wouldn't it be better to use output similar to what goes to stdout in the "json" output?

I wanted to keep the JSON simple. As the information is basically duplicate information, I would prefer to keep there only one of the representation. But given that you exclude the unknown flags, we might be missing some information though. But I do not think it is a huge deal

frankmorgner (approved):

frankmorgner:
thanks

PR #2901, opened 2023-10-16
opensc-notify: Fix wrong popup menu popup position

llogar:
If monitor resolution scaling is used, opensc-notify.exe popup menu pops up at the wrong position.

### DIFF
--- a/src/ui/notify.c
+++ b/src/ui/notify.c
@@ -84,6 +84,26 @@ void ShowContextMenu(HWND hwnd, POINT pt);
 // we need commctrl v6 for LoadIconMetric()
 #include <commctrl.h>
 
+static int GetMonitorScalingRatio(void)
+{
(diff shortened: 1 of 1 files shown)


frankmorgner (commented):
Looks good.

If we rely on `EnumDisplaySettings` to be successful, then `info.rcMonitor.right - info.rcMonitor.left` should not be `0`. Do you think it makes sense to test for this anyway in order to prevent a possible devision by zero?

llogar:
I also share the opinion the division by zero shouldn't occur, but I guess one more check doesn't hurt...

frankmorgner (approved):
thank you

Jakuje:
Thank you for the contribution and reviews! I do not have a way to test it, but I think it is good to go for 0.24.0 unless there would be some concerns against.

Issue #2870, opened 2023-09-20
is_nss_browser process name check is brittle and doesn't include Microsoft Edge


erik-anderson:
### Problem Description

In #2681, a check for the process name was added to check for "chrome", "chromium", and "firefox". As far as I can tell, this is unlikely to cover other Chromium-based browsers like Microsoft Edge ("msedge").

### Proposed Resolution
Add other Chromium-based browsers that are available on Linux.

### Other context
I was looking at this code because the Microsoft Edge team received a report from a customer that a crash happens in Microsoft Edge when they move to `0.23.0-0.1ubuntu1` from `0.20.0-3ubuntu0.1~esm1`. That crash looks like heap corruption and the customer doesn't see the issue in Google Chrome, so I decided to see if there were any clear browser detection code paths. I don't have any smoking gun yet on where the heap corruption is happening, unfortunately.

Since #2681 was added after that release, I don't think it's pertinent to the customer-reported problem, but it would presumably cause Chrome vs. Edge behavior differences in 0.24.0, so I'm proactively filing this issue.

frankmorgner:
We would be happy to detect Edge. If `msedge` is indeed reliable, feel free to open a pull request adding this to the list.

The only reported crash of Firefox (as NSS browser) that is still present in 0.23.0, that I am aware of, is https://github.com/OpenSC/OpenSC/issues/2419. This is fixed in `master` and may be worth trying...

frankmorgner:
Also note that you may always set `slots_per_card = 1` (the default for NSS browsers) via opensc.conf without any changes to the code.

PR #2902, opened 2023-10-17
Add MS Edge as nss browser

jbrummer-vw:
Fixes: https://github.com/OpenSC/OpenSC/issues/2870

### DIFF
--- a/src/pkcs11/misc.c
+++ b/src/pkcs11/misc.c
@@ -467,7 +467,7 @@ static int is_nss_browser(sc_context_t * ctx)
 		basename += sizeof(char);
 
 	if (strstr(basename, "chromium") || strstr(basename, "chrome")
-			|| strstr(basename, "firefox"))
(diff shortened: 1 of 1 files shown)


Jakuje (approved):

dengert:
Does MS edge use Mozilla NSS?  I don't think so: https://stackoverflow.com/questions/441194/does-microsoft-provides-a-pkcs-11-provider-that-connects-to-its-own-ms-cryptoap

OpenSC provides a "minidriver" for Windows which should be used with Edge.  This calls the OpenSC pkcs15 API which does not use PKCS11.  

https://learn.microsoft.com/en-us/windows-hardware/drivers/smartcard/smart-card-minidrivers


So IMHO,`strstr(basename, "msedge")`  will never be true.

metsma:
Yes on linux it does use NSS

frankmorgner (approved):

frankmorgner:
thank you

PR #2904, opened 2023-10-19
pkcs11: Fix C_InitToken() pLabel past-end-of-buffer read

llogar:
`C_InitToken()` `pLabel` parameter is not required to be NUL-terminated. But underlying functions use regular C string.h functions to manipulate a label which leads to random trailing characters being appended.

##### Checklist

### DIFF
--- a/src/common/compat_strndup.c
+++ b/src/common/compat_strndup.c
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2023 Luka Logar <luka.logar@iname.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifndef HAVE_STRNDUP
+#include <stdlib.h>
+#include <string.h>
+#include "common/compat_strnlen.h"
+
+char *_strndup(const char *str, size_t n)
+{
+	char *dst;
+	if (!str)
+		return NULL;
+	n = strnlen(str, n);
+	dst = (char*)malloc(n + 1);
+	if (!dst)
+		return NULL;
+	memcpy(dst, str, n);
+	dst[n] = '\0';
+	return dst;
+};
+
+#endif
--- a/configure.ac
+++ b/configure.ac
@@ -445,7 +445,7 @@ AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([ \
 	getpass gettimeofday getline memset mkdir \
 	strdup strerror memset_s explicit_bzero \
-	strnlen sigaction
+	strnlen strndup sigaction
 ])
 
 #
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -8,7 +8,7 @@ dist_noinst_DATA = \
 	LICENSE.compat_getopt compat_getopt.txt \
 	compat_getopt_main.c \
 	README.compat_strlcpy compat_strlcpy.3
-noinst_HEADERS = compat_strlcat.h compat_strlcpy.h compat_strnlen.h compat_getpass.h compat_getopt.h simclist.h libpkcs11.h libscdl.h compat_overflow.h
+noinst_HEADERS = compat_strlcat.h compat_strlcpy.h compat_strnlen.h compat_strndup.h compat_getpass.h compat_getopt.h simclist.h libpkcs11.h libscdl.h compat_overflow.h
 
 AM_CPPFLAGS = -I$(top_srcdir)/src
 
@@ -17,6 +17,7 @@ libcompat_la_SOURCES = \
 	compat_strlcat.c \
 	compat_strlcpy.c \
 	compat_strnlen.c \
+	compat_strndup.c \
 	compat_getpass.c \
 	compat_getopt.c \
 	compat_report_rangecheckfailure.c \
@@ -37,6 +38,7 @@ TIDY_FILES = \
 	compat_strlcat.h compat_strlcat.c \
 	compat_strlcpy.h compat_strlcpy.c \
 	compat_strnlen.h compat_strnlen.c \
+	compat_strndup.h compat_strndup.c \
 	compat_getpass.h compat_getpass.c \
 	compat_getopt.h compat_getopt.c \
 	compat_report_rangecheckfailure.c \
--- a/src/common/compat_strndup.h
+++ b/src/common/compat_strndup.h
@@ -1,0 +1,22 @@
+/**
+ * @file
+ * @brief prototype of strndup()
+ */
+
+#ifndef __COMPAT_STRNDUP_H
+#define __COMPAT_STRNDUP_H
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifndef HAVE_STRNDUP
+#include <stddef.h>
+/* Workaround for mingw gcc nonnull-compare error */
+#define strndup _strndup
+char *_strndup(const char *str, size_t n);
+#else
+#include <string.h>
+#endif
+
+#endif
--- a/src/pkcs11/pkcs11-global.c
+++ b/src/pkcs11/pkcs11-global.c
@@ -36,6 +36,7 @@
 
 #include "sc-pkcs11.h"
 #include "ui/notify.h"
+#include "common/compat_strndup.h"
 
 #ifdef ENABLE_OPENSSL
 #include <openssl/crypto.h>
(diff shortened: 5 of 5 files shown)


src/common/compat_strndup.c
```diff
@@ -0,0 +1,36 @@
...
+#include <string.h>
+#include "common/compat_strnlen.h"
+
+char *strndup(const char *str, size_t n)
+{
+	char *dst;
+	n = strnlen(str, n);
+	dst = (char*)malloc(n + 1);
+	memcpy(dst, str, n);
+	dst[n] = 0;
```
frankmorgner, line 32:
```suggestion
	dst[n] = '\0';
```

frankmorgner (changes requested):
your `strndup` implementation lacks error checking.

src/common/compat_strndup.c
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef HAVE_STRNDUP
+#include <stdlib.h>
+#include <string.h>
+#include "common/compat_strnlen.h"
+
+char *strndup(const char *str, size_t n)
+{
+	char *dst;
+	n = strnlen(str, n);
```
frankmorgner, line 29: please check `str != NULL`

frankmorgner (changes requested):

src/common/compat_strndup.c
llogar: Since `str` is `const char *` I get an `error: ‘nonnull’ argument ‘str’ compared to NULL [-Werror=nonnull-compare]` error during compiling?

I could drop the const qualifier or probably get around this by using a hackish cast, but other string helper functions here also don't do any kind of input parameter checking?

frankmorgner: Could it be that you're using some other header file for strndup, which specifies `__attribute__((nonnull))` for `str`?

llogar: Nope. 
It turned out, `gcc` has some (lots of, see [Other-Builtins](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)) built-in functions with it's own declarations which in case of `strndup` specify ` __attribute__((nonnull))` for `str`. 
But oddly enough, at least in the case of MinGW `x86_64-w64-mingw32-gcc`, there is only a built-in `strndup` declaration and not also accompanying definition.
Using  `-fno-builtin` gets rid of this error, but using this looks sub-optimal to me? 
Or we can just leave it as it is, as I mentioned before, no other `compat_str*` function does any parameter checking (quite possibly due to the same `nonnull-compare` error)?

frankmorgner: normally, this should only be compiled in if ./configure doesn't detect `strndup`. If it is not detected while being built into gcc/mingw, then the detection mechanism is faulty.

llogar: I think the `./configure` `strndup()` detection works just fine.
It detects `strndup()` on Linux & co. and doesn't detect `strndup()` on MinGW (as it should).

I think on both platforms `gcc` has built-in `strndup()` declaration but on Linux it then links to `strndup()` from `glibc` (or something like that) and since on Windows there is no system defined `strndup()` (or standard C library) the `strndup()` declaration was simply removed from `string.h` and `strndup()` not implemented (but the built-in `strndup()` declaration was kept in `gcc`)

In any case even the "standard" `glibc` [strndup()](https://github.com/bminor/glibc/blob/master/string/strndup.c) implementation doesn't check for `str != NULL`, so I suggest we leave it as it is...

frankmorgner: Not sure if we need to redo the mistakes of the past. I cannot find any documentation stating that this input must be nonnull.

Maybe you can check for strndup by linking against it instead of checking the definition in configure.ac

llogar: A check for `strndup()` works just fine. And `strndup()`  definitely doesn't exist under MinGW-w64 but the compiler nonetheless throws an error as it has a leftover built-in `strndup()` declaration. At least that's how I see it.

So I just renamed `strndup()` to `_strndup()` and use that instead...

Jakuje:
I am not sure if this is the right solution. The PKCS#11 specs say:

> pLabel points to the 32-byte label of the token (which MUST be padded with blank characters, and which MUST not be null-terminated).

http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/pkcs11-base-v2.40.html

I find your proposal over-complicated and only suboptimaly solving the original issue. It does the following:
 * Only extend the 32 bytes to 33 with null byte (there are simpler way to do the same)
 * The logging functions still print the spaces up to 32 characters (this could be done with the different format string, see below)
 * the internal functions behind the `init_token` callback in card drivers might expect the input to come with the trailing whitespace, such as [sc_hsm_init_token()](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-sc-hsm.c#L1626) (actually, I think no other card driver supports this operation than the sc-hsm -- please correct me if I am wrong).

My proposal would be to either simply:
 * strip the trailing blank characters for printing (in separate variable) and pass the original the driver

or
 * change the log format to log max of 32 bytes when priniting the label ('%.32s` format string) in the two cases.

llogar:
> I am not sure if this is the right solution. The PKCS#11 specs say:
> 
> > pLabel points to the 32-byte label of the token (which MUST be padded with blank characters, and which MUST not be null-terminated).
> 
> http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/pkcs11-base-v2.40.html
> 

I read the specs as there shouldn't be any `nul` characters in the first 32 bytes (and the rest are ignored anyway).

It's not just the printout. The `pLabel` provided is copied to the PKCS#15 `EF(TokenInfo)` `label` during initialization (when using, for example, _IsoApplet_) and without this patch invalid (with the garbage characters appended as the `pLabel` is not nul-terminated) value is written to the `EF(TokenInfo)` `label` (there is no 32-byte or nul-termination requirement for it?). Which then breaks PKCS#11 `label` formatting in [framework-pkcs15.c](https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L1189) and possibly elsewhere.

Jakuje:
> I read the specs as there shouldn't be any nul characters in the first 32 bytes (and the rest are ignored anyway).

correct.

> It's not just the printout. The pLabel provided is copied to the PKCS#15 EF(TokenInfo) label during initialization (when using, for example, IsoApplet).

Oh, I see now that the for the IsoApplet this goes through `sc_pkcs15init_add_app()` which sets the tokeninfo label here.

In any case, I would still strip the trailing spaces in the pkcs11 layer rather than just null terminate the trailing spaces.

llogar:
Stripping the trailing spaces won't help in a (I admit, unlikely) scenario where a label is 32 characters long (with no trailing spaces)...

Jakuje:
> Stripping the trailing spaces won't help in a (I admit, unlikely) scenario where a label is 32 characters long (with no trailing spaces)...

Indeed, both stripping and null terminating the buffer should be done from my point of view to keep this PKCS#11 specific semantics in the pkcs11 layer.

Jakuje:
> At least for `pkcs11-tool` stripping is already done in `p11_utf8_to_local()`, where else would you do it?

this converts the pkcs11 formatting to the user facing format. This is ok. But I think we need to do the same when we convert the data on the "other side" of the pkcs11 module, when we pass data from the pkcs11 layer to pkcs15/emulators/drivers. Probably right there where you do your change.

> The reason `p11_utf8_to_local()` didn't correctly strip the label (before this patch) is that the `CK_TOKEN_INFO.label` was (due to the reasons explained above) wrongly formatted as (for example) "`some label ...`" (notice the ending ellipsis, added by `strcpy_bp()`) instead of "`some label `"

Can you point me where? I do not follow this part.

llogar:
[Here](https://github.com/OpenSC/OpenSC/blob/6d7fad101812dbcbde4a3ff069281a359fc54d1f/src/pkcs11/framework-pkcs15.c#L1183) or [here ](https://github.com/OpenSC/OpenSC/blob/6d7fad101812dbcbde4a3ff069281a359fc54d1f/src/pkcs11/framework-pkcs15.c#L1189). As `EF(TokenInfo)` `label` is more than 32 characters long (due to random characters beeing appended during initialization), `strcpy_bp()` adds ellipsis because 

https://github.com/OpenSC/OpenSC/blob/6d7fad101812dbcbde4a3ff069281a359fc54d1f/src/pkcs11/misc.c#L52-L56

where `src_len > 32` (first 32 characters are correctly taken from `C_InitToken()` `pLabel` (non nul-terminated, space padded) followed by some random characters beeing appended during initialization) and `dstsize = 32`

Jakuje:
I think this code is ok. When the tokeninfo will be initialized and null terminated correctly, this won't happen. But with given input, the `strcpy_bp()` does the best it can I think.

llogar:
> I think this code is ok. When the tokeninfo will be initialized and null terminated correctly, this won't happen. But with given input, the `strcpy_bp()` does the best it can I think.

Yes, the code is ok. I thought you were asking why the label wasn't stripped of the trailing spaces...

So, do I add the stripping code, as you suggested, to this PR or do we do it with another one?

Jakuje:
I think this PR is ok as it is related to the change you discuss here.

frankmorgner (approved):

Jakuje:
@frankmorgner what do you think about the discussion during last week? I think we should also strip the trailing spaces while in the pkcs11 layer to avoid the drivers need to mess with that or store bogus trailing whitespace, which is not handled in this PR yet.

Indeed, it can be handled as a separate PR, but I would prefer to do it now as it changes the same code.

I also do not like the overriding of the `pLabel` function argument, but I think I can live with that.

frankmorgner:
Actually, I just now checked [PKCS#11 v3.0 ](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976619)and the specification is quite clear about this:
>C_InitToken initializes a token. slotID is the ID of the token’s slot; pPin points to the SO’s initial PIN (which need not be null-terminated); ulPinLen is the length in bytes of the PIN; pLabel points to the 32-byte label of the token (which MUST be padded with blank characters, and which MUST not be null-terminated). This standard allows PIN values to contain any valid UTF8 character, but the token may impose subset restrictions.
```
Example:

CK_SLOT_ID slotID;

CK_UTF8CHAR pin[] = {“MyPIN”};

CK_UTF8CHAR label[32];

CK_RV rv;

 

.

.

memset(label, ‘ ’, sizeof(label));

memcpy(label, “My first token”, strlen(“My first token”));

rv = C_InitToken(slotID, pin, strlen(pin), label);

if (rv == CKR_OK) {

  .

  .

}

```
Having this in mind, it seems that any label below 32 characters should be rejected...

A fix to reading beyond the input buffer would be to
1. check the input length of 32
2. malloc 32 character
3. memcpy the label

no strndup needed...

frankmorgner:
Just as a side note, the [official pkcs11t.h ](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/include/pkcs11-v3.0/pkcs11t.h)defines the type as follows:
```
/* an 8-bit UTF-8 character */
typedef CK_BYTE           CK_UTF8CHAR;
```
which means that the label is still exactly 32 bytes long!

Jakuje:
The important part of this was fixed in 4562a385ffd40552a839133141d1a19e0ff03f62 (#2945) earlier last year so closing this one. Please, have a look if it fixes the issue for you. If not, please open a new PR.

PR #2905, opened 2023-10-19

popovec:

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -7067,7 +7067,7 @@ static int test_verify(CK_SESSION_HANDLE sess)
 	return errors;
 }
 
-#if OPENSC_VERSION_MAJOR == 0 && OPENSC_VERSION_MINOR <= 23
+#if OPENSC_VERSION_MAJOR == 0 && OPENSC_VERSION_MINOR <= 24
 #else
 #ifdef ENABLE_OPENSSL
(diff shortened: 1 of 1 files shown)


Jakuje (approved):

src/tools/pkcs11-tool.c
```diff
@@ -7067,7 +7067,8 @@ static int test_verify(CK_SESSION_HANDLE sess)
 	return errors;
 }
 
-#if OPENSC_VERSION_MAJOR == 0 && OPENSC_VERSION_MINOR <= 23
+#if 1
```
frankmorgner, line 7070:
```suggestion
#if OPENSC_VERSION_MAJOR == 0 && OPENSC_VERSION_MINOR <= 24
```

frankmorgner (changes requested):

src/tools/pkcs11-tool.c
popovec: Ok, I'll change it so that it checks the opensc version like before. I'm afraid that in a year I will put this patch again on the next version of opensc :-)

I tried to fix this code to pass to pass this test. The problem is that a different template must be used for MyEID than for softhsm. Another problem is that this test will generate additional keys on the card, which must then be deleted. Deleting is not a problem, but it might be appropriate to make the "pkcs11-tool --test" tests so that they run in read only mode. It is not good to write to the card repeatedly during tests (due to the lifetime of the EEPROM on the card).

PR #2906, opened 2023-10-20
pkcs11: use secure memory for puk if configured for `init_pin_in_so_s…

frankmorgner:
…ession`

##### Checklist

- [ ] macOS tokend is tested


Jakuje (approved):
looks good. Do you want this for the 0.24.0? I do not see a reason why it should not be there, but rather checking anyway.

frankmorgner:
as you prefer. I think it can be included in 0.24.0

Jakuje:
I think it would be better merged earlier to get it tested rather than later. It should be not cause any issues though. Merging then. Thank you!

Issue #2903, opened 2023-10-17
Asking for user PIN when already logged in.


larssilven:
### Problem Description
0.24.0-rc1 asks for for PIN all the time regardless if the user is logged in or not. According to PKCS#11 PIN must only be provided to the token after C_Login.

I am using a PIN pad reader so I recognizes all logins.

Version 0.23.0 is working as expected. master branch and 0.24.0-rc1 has this problem.

### Proposed Resolution

If this new behavior is wanted for some application it must be enabled by the configuration.
Default must be to have it working according to PKCS#11 (as in version 0.23.0).

### Steps to reproduce

Just run any application that has "private" objects and you will have to provide PIN all the time.

### Logs
Logs from a test that show what is happening. The name of the file tells opensc log level (\_l**level**) and version of opensc (\_**version**.log).
Here are the files that have been produced
* [pksc11_l3_0.24.0-rc1.log](http://pkimodules.com/logs/pkcs11_l3_0.24.0-rc1.log)
* [opensc_l3_0.24.0-rc1.log](http://pkimodules.com/logs/opensc_l3_0.24.0-rc1.log)
* [pksc11_l8_0.24.0-rc1.log](http://pkimodules.com/logs/pkcs11_l8_0.24.0-rc1.log)
* [opensc_l8_0.24.0-rc1.log](http://pkimodules.com/logs/opensc_l8_0.24.0-rc1.log)
* [pksc11_l8_0.23.0.log](http://pkimodules.com/logs/pkcs11_l8_0.23.0.log)
* [opensc_l8_0.23.0.log](http://pkimodules.com/logs/opensc_l8_0.23.0.log)

PIN is provided at these lines in  pksc11_l**level**_0.24.0-rc1.log (yes it was the same lines for both levels):
60	89	234	308	383	693	905	1237	1836	2319	2392	2423	

In pkcs11_l8_0.23.0.log PIN is just provided 2 times each time after C_Login: 60 1237

In the opensc logs you can see when PIN was provided (all lines with "sc_pin_cmd"). It was done 32 times for 0.24.0-rc1. So some pkcs#11 functions generates several PIN authentications.

C_Login was just called 2 times so it should just have been 2 PIN authentications.

When not using PIN pad everything is working as expected if PIN caching is enabled. PIN is only provided by the application 2 times by using C_Login:
[...]

PR #2907, opened 2023-10-20
pkcs11: don't logout to preserve context of concurrent processes

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


Jakuje (approved):
looks reasonable. It is though a question if we should make it a build time (like you propose) or runtime option (as Doug proposed in #2903).

frankmorgner:
> looks reasonable. It is though a question if we should make it a build time (like you propose) or runtime option (as Doug proposed in #2903).

Indeed. As you can see, I even propose to not make it an option for `./configure`, but to still keep it in the source code. I suspect that making this optional (build time/ runtime) would not be used by anyone, but I'm open to other arguments.

Jakuje:
That is a good point. It does not make sense to make it more complicated with configure/configuration options if this would be something that we can hardly support.

Jakuje:
Any other concerns or are we good to merge this change for 0.24.0?

dengert:
Looks good.

PR #2913, opened 2023-10-24
pkcs11-tool.c - test CLASS for CKO_PRIVATE_KEY

dengert:
In tests for "CKO_PRIVATE_KEY" Use "getCLASS"  not "getKEY_TYPE"

which caused login(session,CKU_CONTEXT_SPECIFIC); to never be called. 

 On branch object-class
 Changes to be committed:

##### Checklist


Jakuje (approved):
noticed the same couple of hours back. Looks good!

PR #2915, opened 2023-10-24
pkcs11-tool: configure AES-CMAC

### LINKED ISSUE #2914: Add AES-CMAC support to pkcs11-tool
### Problem Description

AES-CMAC doesn't work using pkcs11-tool as it's missing basic configuration similar to HMAC

NOTE: There are many algorithms which seem to not be setup properly and I'm not sure if one wouldn't need to go through all of them? I don't have a HSM capable of all of them for this task though.

### Proposed Resolution

Add proper flags in to recognize that the algorithm requires an secret key and not a private one. Add the flags that it supports sign/verify as per signature algorithm definitions used for MAC algorithms.

### Steps to reproduce

See below

The Private key error message is because the algorithm doesn't have MF_CKO_SECRET_KEY set so it fails in find_object_flags and falls back to default of private key for sign.

### Logs

```
# echo -n "0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b" | xxd -r -p - > /data/key.bin
# pkcs11-tool --module=module.so --write-object /data/key.bin --usage-sign --type secrkey --key-type AES:16 --label cmackey --id 1234
Created secret key:
Secret Key Object; AES length 16
  VALUE:      0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b
  label:      cmackey
  ID:         1234
  Usage:      verify
  Access:     none
# echo -n "Hi There" | pkcs11-tool --module=module.so --sign --mechanism AES-CMAC --id 1234 | xxd -p -c 0
error: Private key not found
Aborting.
```

dlegaultbbry:
Fixes #2914

Testing: 
```
# echo -n "0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b" | xxd -r -p - > /data/key.bin
# pkcs11-tool --module=module.so --write-object /data/key.bin --usage-sign --type secrkey --key-type=AES:16 --label cmackey --id 1234
Created secret key:
Secret Key Object; AES length 16
  VALUE:      0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b
  label:      cmackey
  ID:         1234
  Usage:      verify
  Access:     none
# echo -n "Hi There" | pkcs11-tool --module=module.so --sign --mechanism AES-CMAC --id 1234 | xxd -p -c 0
Using signature algorithm AES-CMAC
...
```
##### Checklist

- [x] PKCS#11 module is tested (tested my own module)


Jakuje (approved):

Jakuje:

PR #2919, opened 2023-10-29
pkcs11: Fix C_WaitForSlotEvent() not working in Windows

llogar:
If there are any readers present when `C_WaitForSlotEvent()` is called, `C_WaitForSlotEvent()` returns immediately and does not wait for the card/slot event to happen. Apparently the MSDN documentation for `SCardGetStatusChange()` is inaccurate/incomplete and the `dwCurrentState` for `\\?PnP?\Notification` reader should contain the number of readers in it's HiWord to work properly. See [https://stackoverflow.com/questions/16370909](https://stackoverflow.com/questions/16370909).

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -1640,6 +1640,11 @@ static int pcsc_wait_for_event(sc_context_t *ctx, unsigned int event_mask, sc_re
 #else
 			rgReaderStates[num_watch].szReader = "\\\\?PnP?\\Notification";
 			rgReaderStates[num_watch].dwCurrentState = SCARD_STATE_UNAWARE;
+#ifdef _WIN32
+			/* Windows expects number of readers in HiWord of dwCurrentState.
+			 * See https://stackoverflow.com/questions/16370909. */
+			rgReaderStates[num_watch].dwCurrentState |= (num_watch << 16);
+#endif
 			rgReaderStates[num_watch].dwEventState = SCARD_STATE_UNAWARE;
 			num_watch++;
 			sc_log(ctx, "Trying to detect new readers");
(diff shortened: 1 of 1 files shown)


Jakuje:
@LudovicRousseau are you aware of this issue of windows implementation of pcsc? Is it known issue? Would pcsc-lite handle this input too to avoid the ifdefs?

LudovicRousseau:
I confirm that Windows includes the number of readers in the high 16 bits of `.dwEventState`.

The line https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L1708 should be enough to copy the state in the expected format:
```
rsp->dwCurrentState = rsp->dwEventState;
```

I do not need the proposed patch in `pcsc_scan` for Windows for example.

The first call to `SCardGetStatusChange()` WILL return if `.dwCurrentState` contains `SCARD_STATE_UNAWARE`. This should happen both on Windows and Linux and is the expected behaviour.
I don't see why this patch is needed.

llogar:
> I confirm that Windows includes the number of readers in the high 16 bits of `.dwEventState`.

Hmm, the only reference to the number of readers I've [found](https://stackoverflow.com/questions/16370909) and mentioned in my PR  suggested that the number of readers should be put into the high 16 bits of `.dwCurrentState`. And if done so, the `.dwEventState` (after `SCardGetStatusChange()` returns) does not hold the number of readers in it's high 16 bits. Hence the need to also update `rsp->dwCurrentState`.

If I put (per your suggestion) number of readers into the high 16 bits of `.dwEventState` then updating `rsp->dwCurrentState` is not necessary, as the number of readers is carried over from the `.dwEventState` with the assignment you quoted.

In both cases one has to set the number of readers (one way or another) at least once in the beginning, otherwise `SCardGetStatusChange()` won't block and `pcsc_wait_for_event()` will return immediately with `SC_ERROR_EVENT_TIMEOUT`?

PR #2920, opened 2023-10-30
pkcs11: Fix onepin for thunderbird

fbezdeka:
I did some tests on current master (0.24.0-rc1) to make sure that all of my use cases are covered by the next release. There is one fallout due to the removal of onepin-opensc-pkcs11.so:

Thunderbird is quite similar to Firefox and asked several times for the same pin. Adding "thunderbird" to the list of NSS browsers fixes that issue.

Tested with:

```
opensc-tool -n       
Using reader with a card: Alcor Micro AU9540 00 00
Atos CardOS
``` 

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs11/misc.c
+++ b/src/pkcs11/misc.c
@@ -467,7 +467,8 @@ static int is_nss_browser(sc_context_t * ctx)
 		basename += sizeof(char);
 
 	if (strstr(basename, "chromium") || strstr(basename, "chrome")
-			|| strstr(basename, "firefox") || strstr(basename, "msedge"))
+			|| strstr(basename, "firefox") || strstr(basename, "msedge")
+			|| strstr(basename, "thunderbird"))
 		return 1;
 
 	return 0;


Jakuje:
Do all the other features of CardOS work, including the always-authenticate signature pin with this modification?

Do you have opensc debug trace from the PIN prompts? Could this be related to the other issues we have in rc1 regarding to the tracking of the login state?

dengert:
@fbezdeka  and @metsma 
What is output from: `pkcs15-tool --list-pins`  for your cards?

metsma:
Nss thinks all certificates are protected and it logins every slot before reading certificate.

Jakuje:
> Nss thinks all certificates are protected and it logins every slot before reading certificate.

This is different issue and that should have been solved by the profile object "Public Certificates Token" from the PKCS#11 3.0 but for some reason it wont work in some cases. But I was not able to get back down to that issue with reliable reproducer ... The code is there for both opensc and nss, but it does not work for some reason.

https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.0/os/pkcs11-profiles-v3.0-os.html#_Toc10196664

dengert:
> This is different issue and that should have been solved by the profile object "Public Certificates Token" from the PKCS#11 3.0 but for some reason it wont work in some cases.

[3.7.2 Conformance of a Public Certificates Token](https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.0/os/pkcs11-profiles-v3.0-os.html)  in  "8 Supports the following object location requirements:"
```
c. One or more of the following conditions must be met:
        i.    The matching private key for a certificate can be found via C_FindObjects using the matching CKA_ID value without a login having been performed;
        ii.   The matching public key for a certificate can be found via C_FindObjects using the matching CKA_ID value without a  login having been performed.
```
Is it possible NSS ignored the "One or more". 

`pkcs11-tool -O` shows OpenSC does not do "i." only "ii.".  Some spy trace should show if NSS requests the profile object  and what it does next. 

Another possible problem could be some card drivers (PIV for example) make the CKA_ID unique for the token, so all 9A keys have CKA_ID=01. Where as PKCS11 implies CKA_ID should be unique across all tokens.  So would work if only 1 token is present but might fail if to or more are present.

Jakuje:
> Is it possible NSS ignored the "One or more".

I do not think so the code just checks the profile object, which should be present (if the token does not have "hidden" private certificates):

https://github.com/nss-dev/nss/blob/5832b5cf4ad8ac24cd5893a214324bbb438934a6/lib/pk11wrap/pk11slot.c#L1799-L1810

But to see whats really going on in NSS, we would need a debug log or pkcs11 trace.

> `pkcs11-tool -O` shows OpenSC does not do "i." only "ii.". Some spy trace should show if NSS requests the profile object and what it does next.

Correct. I think I never noticed that part that you might provide also public private keys, but yeah, this option probably makes sense if the token has limitation of objects number and public key can be derived from the private one.

> Another possible problem could be some card drivers (PIV for example) make the CKA_ID unique for the token, so all 9A keys have CKA_ID=01. Where as PKCS11 implies CKA_ID should be unique across all tokens. So would work if only 1 token is present but might fail if to or more are present.

I do not think this should be the case. CKA_ID uniqueness across tokens does not make much sense. Can you point me to where from the PKCS#11 this comes from?

dengert:
https://github.com/nss-dev/nss/blob/5832b5cf4ad8ac24cd5893a214324bbb438934a6/lib/pk11wrap/pk11slot.c#L1196-L1213
uses template for CKA_CLASS = CKO_PROFILE and CKA_TOKEN = CK_TRUE. 

looks like OpenSC needs this untested fix:
```
index 3988c062d..7bde6145b 100644
@@ -5280,6 +5280,10 @@ pkcs15_profile_get_attribute(struct sc_pkcs11_session *session, void *object, CK
                check_attribute_buffer(attr, sizeof(CK_ULONG));
                *(CK_ULONG*)attr->pValue = pobj->profile_id;
                break;
+       case CKA_TOKEN;
+               check_attribute_buffer(attr, sizeof(CK_BBOOL));
+               *(CK_BBOOL*)attr->pValue = CK_TRUE;
+               break;
        default:
                return CKR_ATTRIBUTE_TYPE_INVALID;
        }
```

( Idon't see how a profile would ever have CKA_TOKEN == CK_FALSE.)

Jakuje:
Good catch! Well, technically, the profile objects are not on the card as we just emulate them in "opensc memory", which would fit a description of CKA_TOKEN=FALSE. It is question why NSS insist on having these as a token objects, but it is certainly worth trying.

Can you propose a PR with this change so the reporter (or anyone interested) can try it with the test builds?

dengert:
@fbezdeka  can you try  #2924

dengert:
> I do not think this should be the case. CKA_ID uniqueness across tokens does not make much sense. Can you point me to where from the PKCS#11 this comes from?

pkcs11-base-v3.0-os says:
"The CKA_ID attribute is intended as a means of distinguishing multiple public-key/private-key pairs held by the same subject (whether stored in the same token or not). (Since the keys are distinguished by  subject name as well as identifier, it is possible that keys for different subjects may have the same CKA_ID value without introducing any ambiguity.)"
And:
"CKA_ID value be identical to the key identifier in such a certificate extension, although this will not be enforced by Cryptoki.

More of intent and not a requirement.  So we should be OK.

Jakuje:
Thanks for the pointer. Yeah, it kind of makes sense, but it does not look like requirement as you mention.

The solution might be to generate the ID based on the public key parameters as it is done when the ID is not provided in the pkcs11 tool (or in other tools). But the drawback would be that the IDs would not be predictable and would be very long to be useful in configuration.

dengert:
You lost me on this: "The solution might be to generate the ID based on the public key parameters as it is done when the ID is not provided in the pkcs11 tool (or in other tools). "

Are you saying for the PIV card at least, to use: "CKA_ID value be identical to the key identifier in such a certificate extension". It could do that but could have consequences as PKCS11 URIs or things like minidriver container ID are based on PIV using CKA_IDs:
 01, 02, 03, 04  for keys: 9A, 0C, 9D, 9E.

Jakuje:
No, the idea was to use ID as a sha hash of the public key, but as mentioned it would cause more trouble than gain (but it should be indeed unique).

fbezdeka:
Hi all,

starting from the top:

> @fbezdeka and @metsma What is output from: `pkcs15-tool --list-pins` for your cards?

```
$ pkcs15-tool --list-pins
Using reader with a card: Alcor Micro AU9540 00 00
PIN [PIN]
	Object Flags   : [0x03], private, modifiable
	Auth ID        : 02
	ID             : 01
	Flags          : [0x11], case-sensitive, initialized
	Length         : min_len:8, max_len:8, stored_len:0
	Pad char       : 0x00
	Reference      : 129 (0x81)
	Type           : UTF-8

PIN [SO-PIN]
	Object Flags   : [0x03], private, modifiable
	ID             : 02
	Flags          : [0x99], case-sensitive, unblock-disabled, initialized, soPin
	Length         : min_len:8, max_len:8, stored_len:0
	Pad char       : 0x00
	Reference      : 130 (0x82)
	Type           : UTF-8

AuthKey [Admin Key]
	Object Flags   : [0x03], private, modifiable
	ID             : 08
	Derived        : 1
	SecretKeyID    : 43524b3031

PIN [Extra PIN #1]
	Object Flags   : [0x03], private, modifiable
	Auth ID        : 02
	ID             : 04
	Flags          : [0x13], case-sensitive, local, initialized
	Length         : min_len:4, max_len:16, stored_len:0
	Pad char       : 0x00
	Reference      : 133 (0x85)
	Type           : UTF-8
	Path           : 3f005015

PIN [Extra PIN #0]
	Object Flags   : [0x03], private, modifiable
	Auth ID        : 02
	ID             : 03
	Flags          : [0x13], case-sensitive, local, initialized
	Length         : min_len:4, max_len:16, stored_len:0
	Pad char       : 0x00
	Reference      : 132 (0x84)
	Type           : UTF-8
	Path           : 3f005015
```

> @fbezdeka can you try https://github.com/OpenSC/OpenSC/pull/2924

I did. The problem is still visible, I have to insert my PIN several times.


Debug logs (level=3), one generated on top of #2924, one ontop of #2920:
- [opensc-debug-2924.txt](https://github.com/OpenSC/OpenSC/files/13255274/opensc-debug-2924.txt)
- [opensc-debug-with-nss-onepin-fix.txt](https://github.com/OpenSC/OpenSC/files/13255275/opensc-debug-with-nss-onepin-fix.txt)
 
Hope that helps, let me know if you need more.

fbezdeka:
Just for the records: I continued testing and realized that evolution (another mail client on Linux) has the same issue (multiple PIN requests).

Jakuje:
The different pins are represented as different slots in pkcs11. Can you show us the listing of the different slot objects? Is there the profile object in all of them or just in the first one?

fbezdeka:
> Can you show us the listing of the different slot objects?

```
$ pkcs11-tool --list-slots
Available slots:
Slot 0 (0x0): Alcor Micro AU9540 00 00
  token label        : Siemens Corporate ID Card (V5)
  token manufacturer : www.atos.net/cardos
  token flags        : login required, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 324554444b5a3737
  pin min/max        : 8/8
Slot 1 (0x1): Alcor Micro AU9540 00 00
  token label        : Siemens Corpor... (Extra PIN #1)
  token manufacturer : www.atos.net/cardos
  token flags        : login required, token initialized, PIN initialized, user PIN locked
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 324554444b5a3737
  pin min/max        : 4/16
Slot 2 (0x2): Alcor Micro AU9540 00 00
  token label        : Siemens Corpor... (Extra PIN #0)
  token manufacturer : www.atos.net/cardos
  token flags        : login required, token initialized, PIN initialized, user PIN locked
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 324554444b5a3737
  pin min/max        : 4/16
```

My understanding is that the profile object is only in slot 0:

```
$ pkcs11-tool --list-objects --slot 0
<snip>
Profile object 2229691136
  profile_id:          CKP_AUTHENTICATION_TOKEN (3)
$
$ pkcs11-tool --list-objects --slot 1
$ pkcs11-tool --list-objects --slot 2
$
```

Jakuje:
> My understanding is that the profile object is only in slot 0:

So that might be the issue, which we missed initially.

Yeah, using onepin semantics will work too as it will effectively hide all the other PINs, but I think the solution should be to add a profile object to all the slots.

I see your card reports `CKP_AUTHENTICATION_TOKEN` which means neither of the conditions for `CKP_PUBLIC_CERTIFICATES_TOKEN` is likely not satisfied (can you confirm the certificates are not readable before login on your token with `pkcs11-tool -O`?).

I will have a look if I can put together some patch to populate all slots with profile objects.

Jakuje:
Filled my proposal in #2928. It might need also the changes from #2924. I can check at least with some multi-slot tokens tomorrow, but feedback would be welcomed.

fbezdeka:
I can confirm that the combination of #2928 and #2924 works (so far).

metsma:
EstEID card
```
pkcs15-tool --list-pins
Using reader with a card: Generic Smart Card Reader Interface
PIN [PIN1]
	Object Flags   : [0x30]
	Auth ID        : 03
	ID             : 01
	Flags          : [0x30], initialized, needs-padding
	Length         : min_len:4, max_len:12, stored_len:12
	Pad char       : 0xFF
	Reference      : 1 (0x01)
	Path           : 3f00
	Tries left     : 3

PIN [PIN2]
	Object Flags   : [0x32], modifiable
	Auth ID        : 03
	ID             : 02
	Flags          : [0x32], local, initialized, needs-padding
	Length         : min_len:5, max_len:12, stored_len:12
	Pad char       : 0xFF
	Reference      : 133 (0x85)
	Path           : 3f00adf2
	Tries left     : 3

PIN [PUK]
	Object Flags   : [0x70]
	ID             : 03
	Flags          : [0x70], initialized, needs-padding, unblockingPin
	Length         : min_len:8, max_len:12, stored_len:12
	Pad char       : 0xFF
	Reference      : 2 (0x02)
	Path           : 3f00
	Tries left     : 3
```
```
pkcs11-tool --list-slots
Available slots:
Slot 0 (0x0): Generic Smart Card Reader Interface
  token label        : JÕEORG,JAAK-KRISTJAN,... (PIN1)
  token manufacturer : IDEMIA
  token flags        : login required, token initialized, PIN initialized, readonly
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : AS0250086
  pin min/max        : 4/12
Slot 1 (0x1): Generic Smart Card Reader Interface
  token label        : JÕEORG,JAAK-KRISTJAN,... (PIN2)
  token manufacturer : IDEMIA
  token flags        : login required, token initialized, PIN initialized, readonly
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : AS0250086
  pin min/max        : 5/12
```

frankmorgner:
> I can confirm that the combination of #2928 and #2924 works (so far).

Could this mean, that the proposed changes also deprecate the special treatment of all the other NSS based applications as well (i.e. the quirk that used to be part of the onepin module only)?

metsma:
> > I can confirm that the combination of #2928 and #2924 works (so far).
> 
> Could this mean, that the proposed changes also deprecate the special treatment of all the other NSS based applications as well (i.e. the quirk that used to be part of the onepin module only)?

I don't think google has recent NSS code. But I could look wrong place.
https://chromium.googlesource.com/chromium/third_party/nss/+/refs/heads/master/mozilla/security/nss/lib/pk11wrap/pk11slot.c#1592

Jakuje:
> > I can confirm that the combination of #2928 and #2924 works (so far).
> 
> Could this mean, that the proposed changes also deprecate the special treatment of all the other NSS based applications as well (i.e. the quirk that used to be part of the onepin module only)?

I believe mostly yes. It solve the PIN prompting issues when:
 * the token has public readable certs/keys -- the prompt for this slot wont happen before certificate selection
 * the "virtual" slots that are technically SO PINs/PUKs wit no objects assigned -- the prompt wont happen at all.

But it won't solve the case when the certs/public keys are NOT publicly readable. In that case, both of the solution have to prompt for the pin. But in both of the ways, it should be just one prompt (onepin semantics will show only one slot and the SO PINs will show as "friendly" for NSS so it will not try to log into them.

So I think for the current NSS/Firefox, we could drop the one-pin semantics in long term (while having it in should not hurt), but if some browsers use older NSS, it might be still relevant. Some more testing is probably needed.

frankmorgner:
Thanks for the clarification. Then, I think, we can move on with https://github.com/OpenSC/OpenSC/pull/2928 and https://github.com/OpenSC/OpenSC/pull/2924 and close this PR, right?

Jakuje:
> Thanks for the clarification. Then, I think, we can move on with #2928 and #2924 and close this PR, right?

Yes. That would be my proposal short term.

PR #2921, opened 2023-10-31
Workaround for unwrapping using 2K RSA key with MyEID. 

hhonkanen:
In MyEID 4.5.5 APDU chaining doesn't work correctly in the unwrap operation. This PR checks for MyEID version, and calls myeid_transmit_decipher_pi_split to use the pre 4.5 way of splitting 2048 bit data for the unwrap operation, if the version is 4..5.x.

##### Checklist

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -1383,7 +1383,8 @@ static int myeid_transmit_decipher(struct sc_card *card, u8 p1, u8 p2,
 		apdu.le = MIN(card->max_recv_size, crgram_len);
 	}
 
-	if (p2 == 0x86 && crgram_len == 256 && priv && !priv->cap_chaining) {
+	/* In MyEID 4.5.x, unwrapping with 2K RSA using APDU chaining doesn't work properly. Split the APDU in the old way in this case. */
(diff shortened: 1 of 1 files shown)


Jakuje:
I don't have any test coverage for unwrap operation in p11test now so I do not have a simple way to test this. But the p11test could be (hopefully) simple to extended and I hope I will be able to set up machine to run tests automatically soon.

Do you have some specific commands/scripts/tests you used for testing that I can reproduce?

Jakuje (approved):

frankmorgner (approved):

hhonkanen:
@Jakuje: @larssilven tested the C_UnwrapKey function with MyEID 4.5.5 before and after this PR, and reported to me that that it fixes the issue. The issue occurred when hUnwrappingKey was a 2048 bit RSA key. I don't currenlty have an automated test for this either. In my test program at https://github.com/hhonkanen/WrapTest/blob/master/main.cpp C_UnwrapKey is called with similar configuration from line 470. You need to prepare the card with a 2048 bit RSA key to test it.

popovec:
Wrap/unwrap is tested in CI (OsEID emulation). RSA unwrap uses a 1024-bit key in the test. We can change it to a 2048-bit key, or perform a test with both a 1024-bit and a 2048-bit key. However, the emulation would not detect this bug, as it uses a different APDU processing (supports APDU chaining, extended APDU and using split - i.e. two APDUs)

popovec (approved):

Jakuje:
I started playing with this a bit, but I did not have much success so far. So lets not block on me and lets have this merged. Thank you!

PR #2922, opened 2023-11-01
Normalize logging of OpenSSL errors 

### LINKED ISSUE #2847: Normalize loggging OpenSSL errors througout the code
> On one hand, I agree, on the other I know that when we will want to debug openssl calls, we will have to add it back. Given the openssl error codes are very low level, would it make sense to create a new debug level for these and log them somehow in cleaner way?

_Originally posted by @Jakuje in https://github.com/OpenSC/OpenSC/pull/2053#discussion_r753443058_
            
The PIV driver introduces a debug log of OpenSSL errors in #2053, there is similar logging in westcos driver:

https://github.com/OpenSC/OpenSC/blob/38da7e1062d4eb3beefa50a2c4838b8ac4ae7066/src/libopensc/card-westcos.c#L58-L69

Having this in verbose/debug logs might be useful to debug issues with OpenSSL.

xhanulik:
Fixes #2847

This pull request enhances and unifies the logging of OpenSSL errors throughout the OpenSC codebase. The newly introduced `sc_log_openssl()` function internally utilizes OpenSSL's `ERR_print_errors_cb()` to log and clear the error queue.

Most of the error handling code in `libopensc/`, `sm/`, `pkcs15init/` and `tools`  (where `sc_context_t` is accessible) is adjusted to incorporate the logging. Additionally, some error handling code has been added in places where it was previously missing.

##### Checklist
- [ ] macOS tokend is tested


src/pkcs11/openssl.c
```diff
@@ -888,10 +909,12 @@ CK_RV sc_pkcs11_verify_data(const CK_BYTE_PTR pubkey, CK_ULONG pubkey_len,
 				unsigned int tmp_len;
 
 				if (!md_ctx || !EVP_DigestFinal(md_ctx, tmp, &tmp_len)) {
+					sc_log_openssl(context);
 					sc_evp_md_free(mgf_md);
 					sc_evp_md_free(pss_md);
 					free(rsa_out);
 					EVP_PKEY_free(pkey);
+					sc_log_openssl(context);
```
Jakuje, line 917: duplicate `sc_log_openssl(context);`

src/pkcs15init/pkcs15-lib.c
```diff
@@ -2625,32 +2626,38 @@ prkey_fixup_rsa(struct sc_pkcs15_card *p15card, struct sc_pkcs15_prkey_rsa *key)
...
+			r = SC_ERROR_INTERNAL;
+			goto end;
+		}
 
-		BN_mod_inverse(rsa_iqmp, rsa_q, rsa_p, bn_ctx);
+		if (!(rsa_dmp1 = BN_new()) ||
+				!(rsa_dmq1 = BN_new()) ||
+				!(rsa_iqmp = BN_new()) ||
+				!(aux = BN_new()) ||
+				!(bn_ctx = BN_CTX_new())) {
```
Jakuje, line 2648: nit: I would rather see the assignments outside of the `if` condition and then separate if condition checking the results instead of fitting everything into the multiline `if` condition as these results are not dependent on each other (unlike the following part where the calculation is already done).

src/pkcs15init/pkcs15-lib.c
```diff
@@ -2683,7 +2697,8 @@ prkey_fixup_rsa(struct sc_pkcs15_card *p15card, struct sc_pkcs15_prkey_rsa *key)
 				key->iqmp.len = 0;
 			}
 		}
-
+end:
+		sc_log_openssl(ctx);
```
Jakuje, line 2701: I think this log should not be here as all the `goto end` have logged the errors already and this code path is executed also in case of success.

src/libopensc/padding.c
```diff
@@ -559,6 +568,7 @@ static int sc_pkcs1_add_pss_padding(sc_context_t *scctx, unsigned int hash, unsi
 	rv = SC_SUCCESS;
 
 done:
+	sc_log_openssl(scctx);
```
Jakuje, line 571: I think this should be before the `goto done` as this code path is used also for the success.

src/libopensc/pkcs15-oberthur.c
```diff
@@ -734,6 +733,9 @@ sc_pkcs15emu_oberthur_add_cert(struct sc_pkcs15_card *p15card, unsigned int file
 
 	rv = sc_oberthur_get_certificate_authority(&cinfo.value, &cinfo.authority);
 	if (rv != SC_SUCCESS) {
+#ifdef ENABLE_OPENSSL
+		sc_log_openssl(ctx);
+#endif
```
Jakuje, line 738: is the ifdef needed? I think the `sc_log_openssl()` is defined also when built without openssl as noop. This is used in few more places.

frankmorgner (changes requested):
Thanks for also having a look at the OpenSSL related error handling. The code looks good, I found only some minor issues.

I think it would be useful to additionally have a dedicated debug level for OpenSC dependencies, e.g.
```diff
diff --git a/src/libopensc/log.c b/src/libopensc/log.c
index cee59bcd9..2764efd83 100644
--- a/src/libopensc/log.c
+++ b/src/libopensc/log.c
@@ -197,7 +197,7 @@ void _sc_log(struct sc_context *ctx, const char *format, ...)
 
 void _sc_log_openssl(struct sc_context *ctx)
 {
-       sc_do_log_openssl(ctx, SC_LOG_DEBUG_NORMAL, NULL, 0, NULL);
+       sc_do_log_openssl(ctx, SC_LOG_DEBUG_DEPS, NULL, 0, NULL);
 }
 
 static int is_a_tty(FILE *fp)
diff --git a/src/libopensc/log.h b/src/libopensc/log.h
index 8179faa6c..97082e0d4 100644
--- a/src/libopensc/log.h
+++ b/src/libopensc/log.h
@@ -37,6 +37,8 @@ enum {
        SC_LOG_DEBUG_SM,                /* secure messaging */
        SC_LOG_DEBUG_ASN1,              /* asn1.c */
        SC_LOG_DEBUG_MATCH,             /* card matching */
+       SC_LOG_DEBUG_DEPS,              /* debugging of dependencies, e.g. OpenSSL */
+       SC_LOG_DEBUG_RFU3,              /* RFU */
        SC_LOG_DEBUG_PIN,               /* PIN commands */
 };
```

src/libopensc/log.c
```diff
@@ -68,6 +71,26 @@ void sc_do_log_color(sc_context_t *ctx, int level, const char *file, int line, c
 	va_end(ap);
 }
 
+#ifdef ENABLE_OPENSSL
+static int log_openssl_error_line(const char *str, size_t len, void *u) {
+	struct log_info_openssl *info = (struct log_info_openssl *) u;
+	// print error message without newline character at the end
+	sc_do_log(info->ctx, info->level, info->file, info->line, info->func, "OpenSSL error: %.*s", (int) len - 2, str);
+	return 1;
```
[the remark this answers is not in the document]
xhanulik: For example when calling `EVP_PKEY_fromdata_init(NULL)`, the OpenSSL `ERR_print_errors_cb()` function gives me error string

`00D8CDCD007F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354`

That was the main reason for prepending `OpenSSL error`.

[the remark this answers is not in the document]
frankmorgner: I was wondering how this looks like in an opensc debug log. if this always has multiple long lines, then, for example, it could be more useful to have an indication on top, with multiple openssl debug log errors following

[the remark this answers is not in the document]
xhanulik: Indeed, the resulting line in the debug log is quite long. Also, when the `sc_log_openssl` is called, the whole OpenSSL error stack is printed out, so even though it would usually be 1 error, having something like

```
P:74501; T:0x140574661335040 10:59:29.103 [opensc-tool] card-piv.c:5698:piv_init: OpenSSL error
00D8C016DA7F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354
```

what is shorter and probably more readable.

xhanulik:
I have changed the OpenSSL function for printing errors from `ERR_print_errors_cb` to `ERR_print_errors_fp`, so the particular errors are printed on separate lines, for example as

```
P:73509; T:0x139793080543232 10:20:17.214 [opensc-tool] card-piv.c:5700:piv_init: OpenSSL error
0068E81C247F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354:
0068E81C247F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354:
0068E81C247F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354:
```

src/libopensc/card-authentic.c
Jakuje, line 1: nit to the commit message -- this does not remove libraries, but headers so I would prefer to reword to "Remove unused openssl headers" or something like that. Similarly in the other commits.

src/libopensc/card-entersafe.c
```diff
@@ -345,14 +378,11 @@ static int entersafe_mac_apdu(sc_card_t *card, sc_apdu_t *apdu,
...
 out:
-	 if(tmp)
-		  free(tmp);
-	 if(tmp_rounded)
-		  free(tmp_rounded);
-	 if (ctx) {
-		sc_evp_cipher_free(alg);
-		EVP_CIPHER_CTX_free(ctx);
-	}
+	sc_log_openssl(card->ctx);
```
Jakuje, line 353: should this be called only if the `r != SC_OK`? Or is the code now quite when there is no error to report?

src/libopensc/card-epass2003.c
```diff
@@ -646,6 +658,7 @@ sha256_digest(struct sc_card *card, const unsigned char *input, size_t length, u
 
 	r = openssl_dig(md, input, length, output);
 	sc_evp_md_free(md);
+	sc_log_openssl(card->ctx);
```
Jakuje, line 661: same here in most of the changes to epass -- this code path is executed in both positive and negative cases

src/tools/netkey-tool.c
```diff
@@ -144,12 +145,25 @@ static void show_certs(sc_card_t *card)
...
+					X509_free(c);
+					return;
+				}
 				printf("  Subject-CN: %s\n", buf2);
-				X509_NAME_get_text_by_NID(X509_get_issuer_name(c), NID_commonName, buf2,sizeof(buf2));
+				if (X509_NAME_get_text_by_NID(X509_get_issuer_name(c), NID_commonName, buf2, sizeof(buf2)) < 0) {
+					sc_log_openssl(card->ctx);
+					printf("  Invalid Issuer-CN\n");
+					X509_free(c);
+					return;
```
Jakuje, line 159: should we return here or `continue` to the next cert? Same in the above case.

src/tools/piv-tool.c
```diff
@@ -214,9 +215,9 @@ static int load_cert(const char * cert_id, const char * cert_file,
 	} else {
 		cert = PEM_read_X509(fp, &cert, NULL, NULL);
     	if(cert == NULL){
-        	printf("file %s does not contain PEM-encoded certificate\n",
-				 cert_file);
-        	goto err;
+			sc_log_openssl(ctx);
+			printf("file %s does not contain PEM-encoded certificate\n", cert_file);
+			goto err;
     	}
```
Jakuje: I think the `if` line should be indented one more tab

src/tools/pkcs15-init.c
```diff
@@ -2330,7 +2341,8 @@ do_read_pkcs12_private_key(const char *filename, const char *passphrase,
 	*key = user_key;
 	return ncerts;
 
-error:	ossl_print_errors();
+error:
+	ossl_print_errors();
```
Jakuje, line 2345: should we convert this to `sc_log_openssl()` and remove the custom `ossl_print_errors()` altogether?

src/libopensc/log.h
```diff
@@ -105,6 +108,10 @@ void _sc_debug(struct sc_context *ctx, int level, const char *format, ...);
 void _sc_log(struct sc_context *ctx, const char *format, ...);
 int sc_color_fprintf(int colors, struct sc_context *ctx, FILE * stream, const char * format, ...);
 #endif
+
+void sc_do_log_openssl(struct sc_context *ctx, int level, const char *file, int line,
+	       const char *func);
```
Jakuje, line 113:
```suggestion
		const char *func);
```
from clang-format:
```
@@ -110,7 +110,7 @@ int sc_color_fprintf(int colors, struct sc_context *ctx, FILE * stream, const ch
 #endif
 
 void sc_do_log_openssl(struct sc_context *ctx, int level, const char *file, int line,
-	       const char *func);
+		const char *func);
 
 /** 
  * @brief Log binary data to a sc context
```

src/libopensc/pkcs15-oberthur.c
```diff
@@ -734,6 +733,7 @@ sc_pkcs15emu_oberthur_add_cert(struct sc_pkcs15_card *p15card, unsigned int file
 
 	rv = sc_oberthur_get_certificate_authority(&cinfo.value, &cinfo.authority);
 	if (rv != SC_SUCCESS) {
+		sc_log_openssl(ctx);
```
Jakuje, line 736: the windows build is missing something which still needs some fixes:
```
pkcs15-oberthur.c(736): error C2220: warning treated as error - no 'object' file generated
pkcs15-oberthur.c(736): warning C4013: '_sc_log_openssl' undefined; assuming extern returning int
```

src/libopensc/card-entersafe.c
xhanulik: Yes, thanks for noticing - after the change to `ERR_print_errors` it would print OpenSSL logging error when no errors are in the stack.

src/tools/pkcs15-init.c
xhanulik: Yes, updated with `sc_log_openssl`

src/libopensc/card-entersafe.c
```diff
@@ -345,14 +378,12 @@ static int entersafe_mac_apdu(sc_card_t *card, sc_apdu_t *apdu,
...
-	 if(tmp)
-		  free(tmp);
-	 if(tmp_rounded)
-		  free(tmp_rounded);
-	 if (ctx) {
-		sc_evp_cipher_free(alg);
-		EVP_CIPHER_CTX_free(ctx);
-	}
+	if (r != SC_SUCCESS)
+		sc_log_openssl(card->ctx);
```
Jakuje, line 354: Calling the log function on the particular line where some of the openssl functions failed instead of doing it after the `goto out` has the advantage that you will see in the error log the file and line on which this macro was called, which simplifies debugging. When calling this from the `goto out`, one needs to guess from the error message which function failed so I would prefer to call this function before the `goto out` for example in this case.

src/tools/piv-tool.c
```diff
@@ -337,27 +338,33 @@ static int gen_key(const char * key_info)
...
 		}
 #else
-		EVP_PKEY_CTX *ctx = NULL;
+		EVP_PKEY_CTX *cctx = NULL;
 		OSSL_PARAM_BLD *bld = NULL;
 		OSSL_PARAM *params = NULL;
 #endif
 
 		if (!keydata.pubkey || !keydata.exponent) {
+			sc_log_openssl(ctx);
```
Jakuje, line 366: I think in this case no openssl operation is involved. These data come from the card.

Jakuje (approved):
two last nits and I think we are good to go.

src/libopensc/card-entersafe.c
xhanulik: Okay, I'll break it down to the individual cases.

Jakuje (approved):

xhanulik:
By now, all review comments should be fixed, just cleaned the commits.

frankmorgner:
Thank you.

PR #2924, opened 2023-11-02
pkcs15-framework.c NSS tests profile for CKA_TOKEN

dengert:
Fixes #2920  Needs test using NSS 

NSS use template with CKO_CLASS = CKO_PROFILE and CKA_TOKEN = CK_TRUE  to find profiles.
pkcs15_framwork.c  now responds withy CKA_TOKEN = CK_TRUE. 
Previously would return  CKR_ATTRIBUTE_TYPE_INVALID

 Date:      Thu Nov 2 09:00:17 2023 -0500
 On branch pkcs11-profile-fix
 Changes to be committed:
	modified:   framework-pkcs15.c

##### Checklist


Jakuje (approved):

Jakuje:
Discussed briefly with Bob about the NSS implementation and really the consensus should be that this is the right solution. If we do not provide the CKA_TOKEN=CK_TRUE, the specification defaults to FALSE, which means that the object is session-only, which really does not make much sense.

Would be glad for comments on #2928 which is related to the fix already verified in #2920.

src/pkcs11/framework-pkcs15.c
```diff
@@ -5280,6 +5280,11 @@ pkcs15_profile_get_attribute(struct sc_pkcs11_session *session, void *object, CK
 		check_attribute_buffer(attr, sizeof(CK_ULONG));
 		*(CK_ULONG*)attr->pValue = pobj->profile_id;
 		break;
+	case CKA_TOKEN:
+		/* NSS test for this */
```
frankmorgner, line 5284: I think we can drop the NSS specific comment here, as we agree that this property is described (and enforced) well enough in the standard.

frankmorgner (approved):

src/pkcs11/framework-pkcs15.c
Jakuje: @dengert can you update the PR removing this comment so we can merge this?

dengert: Forced push without comment.

Jakuje:

PR #2927, opened 2023-11-05
pkcs11-tool: add support for AES/GCM encryption and decryption

jclc:
Adds support for `--encrypt` and `--decrypt` when using `-m AES-GCM`.

Note that due to AES-GCM being a cipher method with authenticated data, multipart decryption doesn't return any data until C_DecryptFinal. This implementation decrypts the entire ciphertext in memory and uses a single C_Decrypt call for inputs of all sizes.

Tested (manually) on the latest version of SoftHSMv2

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -212,7 +212,7 @@ enum {
 	OPT_LIST_INTERFACES,
 	OPT_IV,
 	OPT_AAD_FILE,
-	OPT_TAG_SIZE
+	OPT_TAG_LEN
 };
 
 static const struct option options[] = {
@@ -300,7 +300,7 @@ static const struct option options[] = {
 	{ "allow-sw",		0, NULL,		OPT_ALLOW_SW },
 	{ "iv",			1, NULL,			OPT_IV },
 	{ "aad-file",		1, NULL,		OPT_AAD_FILE },
-	{ "tag-size",	1, NULL,			OPT_TAG_SIZE },
+	{ "tag-len",	1, NULL,			OPT_TAG_LEN },
 
 	{ NULL, 0, NULL, 0 }
 };
@@ -390,7 +390,7 @@ static const char *option_help[] = {
 	"Allow using software mechanisms (without CKF_HW)",
 	"Initialization vector",
 	"Additional authenticated data for AEAD methods",
-	"Authentication tag size in bytes (default: 16)",
+	"Authentication tag length in bytes (default: 16)",
 };
 
 static const char *	app_name = "pkcs11-tool"; /* for utils.c */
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -2577,6 +2602,24 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 		mech.pParameter = iv;
 		mech.ulParameterLen = iv_size;
 		break;
+	case CKM_AES_GCM:
+		use_aead_buffer = TRUE;
+		iv_size = 16;
+		iv = get_iv(opt_iv, &iv_size);
+		gcm_params.pIv = iv;
+		gcm_params.ulIvLen = iv_size;
+		gcm_params.ulIvBits = iv_size * 8; // no one seems to know what this is for
```
Jakuje, line 2611: The section 2.13.5 AES GCM and CCM Mechanism parameters says:

> `ulIvBits`        length of initialization vector in bits. Do no use ulIvBits to specify the length of the initialization vector, but ulIvLen instead.

so I would drop this.

https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061261

src/tools/pkcs11-tool.c
```diff
@@ -2577,6 +2602,24 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+
+		aad_size = 0;
+		aad = get_aad(opt_aad_file, &aad_size);
+		gcm_params.pAAD = aad;
+		gcm_params.ulAADLen = aad_size;
+
+		gcm_params.ulTagBits = opt_tag_len * 8;
```
Jakuje, line 2618: I would sanity-check that the value is in the bounds required by the specs. From the above linked section, it should be between 0 and 128

src/tools/pkcs11-tool.c
```diff
@@ -2617,14 +2660,37 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
 
 	rv = CKR_CANCEL;
-	if (r < (int) sizeof(in_buffer)) {
+	if (use_aead_buffer) {
+		rv = p11->C_DecryptInit(session, &mech, key);
+		if (rv != CKR_OK)
+			p11_fatal("C_DecryptInit", rv);
+		if ((getCLASS(session, key) == CKO_PRIVATE_KEY) && getALWAYS_AUTHENTICATE(session, key))
+			login(session, CKU_CONTEXT_SPECIFIC);
+		out_len = sizeof(out_buffer);
```
Jakuje, line 2685: There is still a lot of duplicated code. In both of these conditions, this block is the same. Would it be possible to rewrite it to avoid the code duplication by placing this part before the newly introduced condition?

Jakuje:
Regarding testing, I would consider to run some automated against some software token. I think NSS softoken supports GCM ciphers and given that it can now take a path to NSS db in the environment variable [1], it should not be very complicated to integrated into the CI (see for example pkcs11-provider [2]).

[1] https://github.com/nss-dev/nss/commit/c0bbdc4ed7b3520c44600f0e254887d2ae938325
[2] https://github.com/latchset/pkcs11-provider/blob/main/tests/setup-softokn.sh

Jakuje:
Merged #3195 as an alternative implementation of the same as there was was no follow-up.

PR #2799, opened 2023-06-08
Windows: Enable MSI signing via Signpath CI integration

frankmorgner:
##### Checklist

- [ ] Documentation is added or updated
- [ ] New files have a LGPL 2.1 license statement
- [ ] PKCS#11 module is tested
- [ ] Windows minidriver is tested
- [ ] macOS tokend is tested

PR #2930, opened 2023-11-07
OpenSSL 1.1.1 is deprecated

metsma:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/MacOSX/build-package.in
+++ b/MacOSX/build-package.in
@@ -29,16 +29,16 @@ if test "$FORCE_OPENSSL_BUILD" == "1" || ! pkg-config libcrypto --atleast-versio
 	if ! test -e $BUILDPATH/openssl_bin/$PREFIX/lib/pkgconfig; then
 		# Build OpenSSL manually, because Apple's binaries are deprecated
 		if ! test -e openssl; then
-			git clone --depth=1 https://github.com/openssl/openssl.git -b OpenSSL_1_1_1-stable
+			git clone --depth=1 https://github.com/openssl/openssl.git -b openssl-3.0
 		fi
 		cd openssl
-		MACHINE=x86_64 ./config no-shared --prefix=$PREFIX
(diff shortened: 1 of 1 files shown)


frankmorgner:
I think it would be good to schedule this when after the upcoming version was released.

I agree with the suggested change, although I'd like to note that extended support for 1.1.1 is [still available](https://www.openssl.org/policies/releasestrat.html)

metsma:
> I think it would be good to schedule this when after the upcoming version was released.
> 

I agree

> I agree with the suggested change, although I'd like to note that extended support for 1.1.1 is [still available](https://www.openssl.org/policies/releasestrat.html)

Thats why I did not touch configure script. It is still possible build with 1.1.1. But default build should produce "suppored" binary

frankmorgner:
Agreed. [Our AppVeyor configuration ](https://github.com/OpenSC/OpenSC/blob/master/.appveyor.yml)should be changed to use the [pre installed 3.0 version ](https://www.appveyor.com/docs/windows-images-software/)as well then.

metsma:
Huh haven't used AppVeyor recently. Is there reason that OpenSC does not use GitHub Actions windows environments?

frankmorgner:
1. We're using AppVeyor for historical reasons
2. My long term plan is to integrate [Signpath](https://github.com/OpenSC/OpenSC/pull/2799), which currently doesn't support GHA

Jakuje:
@metsma can you please rebase so we can re-triger the CI and make sure everything works before merging? Release is out so I would like to give this some more testing.

metsma:
Done

Jakuje:
Looking at the https://www.openssl.org/source/ I am wondering if we should keep the 3.0 or jump to 3.2 directly, which is latest stable.

Jakuje (approved):

frankmorgner:
I think it is better to stick to the LTS version.

PR #2936, opened 2023-11-14
pkcs15_gen_keypair: propagate SENSITIVE/EXTRACTABLE flag from PKCS11 to PKCS15

popovec:
When generating the private key, we transfer the request for the CKA_SENSITIVE and CKA_EXTRACTABLE key attributes up to the pkcs#15 layer.
    
When reading attributes from the PKCS#15 layer, set the CKA_EXTRACTABLE attribute for the pkcs#11 layer according to the SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE attribute.

### DIFF
--- a/doc/tools/pkcs11-tool.1.xml
+++ b/doc/tools/pkcs11-tool.1.xml
@@ -714,9 +714,21 @@
 			in DER format from smart card:
 				<programlisting>pkcs11-tool --read-object --id KEY_ID --type cert --output-file cert.der</programlisting>
 
+			Write certificate to token:
+				<programlisting> pkcs11-tool --login --write-object certificate.der --type cert</programlisting>
+
 			To convert the certificate in DER format to PEM format, use OpenSSL
 			tools:
 				<programlisting>openssl x509 -inform DER -in cert.der -outform PEM > cert.pem</programlisting>
+
+			Generate new RSA Key pair:
+				<programlisting>pkcs11-tool --login --keypairgen --key-type RSA:2048</programlisting>
+
+			Generate new extractable RSA Key pair:
+				<programlisting>pkcs11-tool --login --keypairgen --key-type RSA:2048 --extractable</programlisting>
+
+			List private keys:
+				<programlisting>pkcs11-tool --login --list-objects --type privkey</programlisting>
 
 			To sign some data stored in file <replaceable>data</replaceable>
 			using the private key with ID <replaceable>ID</replaceable> and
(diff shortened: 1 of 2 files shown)


doc/tools/pkcs11-tool.1.xml
```diff
@@ -714,30 +714,50 @@
...
 				<programlisting>pkcs11-tool --read-object --id KEY_ID --type cert --output-file cert.der</programlisting>
 
+			Write certificate to token:
+				<programlisting> pkcs11-tool --login --write-object certificate.der --type cert</programlisting>
+
 			To convert the certificate in DER format to PEM format, use OpenSSL
 			tools:
 				<programlisting>openssl x509 -inform DER -in cert.der -outform PEM > cert.pem</programlisting>
 
+			Generate new key:
```
Jakuje, line 724: I would be more explicit here like "generate new RSA Key pair" to be obvious both of the keys are generated and to avoid confusion with generating symmetric key. Same below.

src/pkcs11/framework-pkcs15.c
```diff
@@ -3410,6 +3410,22 @@ pkcs15_gen_keypair(struct sc_pkcs11_slot *slot, CK_MECHANISM_PTR pMechanism,
...
 
+	/* set EXTRACTABLE and SENSITIVESE flags */
+	rv = attr_find(pPrivTpl, ulPrivCnt, CKA_EXTRACTABLE, &always_auth, &len);
+	if (rv == CKR_OK && always_auth == CK_TRUE)
+		keygen_args.prkey_args.access_flags |= SC_PKCS15_PRKEY_ACCESS_EXTRACTABLE;
+	else
+		keygen_args.prkey_args.access_flags |= SC_PKCS15_PRKEY_ACCESS_NEVEREXTRACTABLE;
+
+	rv = attr_find(pPrivTpl, ulPrivCnt, CKA_SENSITIVE, &always_auth, &len);
+	if (rv == CKR_OK && always_auth == CK_TRUE) {
```
Jakuje, line 3421: please, use a different variable than `always_auth` here. It makes the code very confusing. Either separate variable for both of these or just `bool_value` for all of them together.

doc/tools/pkcs11-tool.1.xml
popovec: Ok, I'll change it..

popovec:
I prepared a test branch, but I have some strange problem with compilation on CI..

make[4]: *** No rule to make target 'org.opensc-project.mac.pkcs11-register.plist.in', needed by 'org.opensc-project.mac.pkcs11-register.plist'.  Stop.


https://github.com/popovec/OpenSC/actions/runs/6886758679 
https://github.com/popovec/OpenSC/actions/runs/6886758676

src/pkcs11/framework-pkcs15.c
popovec: I will make separate variables here.

Jakuje:
Running the logs through diff shows that all dependencies versions are the same. But going through the logs shows the following issue which goes undetected during the `make dist`, which prevents packing these files into the release tarball:

```
tar: opensc-0.24.0PKCS11_EXTRACTABLE_SENSITIVE_test/src/tools/org.opensc-project.mac.pkcs11-register.plist.in: file name is too long (max 99); not dumped   
tar: opensc-0.24.0PKCS11_EXTRACTABLE_SENSITIVE_test/src/tools/org.opensc-project.mac.opensc-notify.plist.in: file name is too long (max 99); not dumped     
```
in short, given that the branch name is used as part of the version, too long file paths are created, which are not handled by the `tar`. I think just using shorter branch name should solve the issue.

popovec:
Thanks for the tip. I did not expect such a restriction at all.

Jakuje (approved):

Jakuje:
> Thanks for the hint. I did not expect such a restriction at all.

Me neither. Learning something new every day :)

metsma:
There are ustar and pax extensions for tar to allow longer filenames. Default is 100 chars

frankmorgner (approved):

PR #2940, opened 2023-11-18
pkcs15-piv.c placeholder for Yubikey key Attestation certificate CK_ID

dengert:
Yubikey PKCS11 module uses CKA_ID "25" for its Attestation certificate which is read using a Yubikey APDU. Other PIV certificates are stored in PIV objects and the certificate is extracted from the object.

OpenSC code was using the same CKA_ID for "Secure Messaging Certificate Signer" which is used with PIV SM.

 So to avoid any confusion, the "Secure Messaging Certificate Signer"
 will now use "81".

 On branch placeholer-for-Attestation
 Changes to be committed:
	modified:   pkcs15-piv.c

##### Checklist

- [ ] macOS tokend is tested


dengert:
I would like to see this simple PR in 0.24.0

Jakuje (approved):
looks reasonable. I think this is good to go for the final release.

frankmorgner (approved):

PR #2941, opened 2023-11-23
card-idprime.c : add support for Gemalto IDPrime 940C

vbonamy:
Define Gemalto IDPrime 940C as SC_CARD_TYPE_IDPRIME_940

- [ ] macOS tokend is tested

 We have recently encountered an issue with OpenSC when using the Gemalto IDPrime 940C card, which is a newer variant of the Gemalto IDPrime 940. After investigating the problem, we have made modifications to OpenSC to add support for Gemalto IDPrime 940C.

The ATR of our Gemalto IDPrime 940C card is 3b:7f:96:00:00:80:31:80:65:b0:85:05:00:39:12:0f:fe:82:90:00.

Without our patch, attempting to use Gemalto IDPrime 940C (via `pkcs11-tool --login --test` ) resulted in the following error:
```
error: PKCS11 function C_SignFinal failed: rv = CKR_FUNCTION_NOT_SUPPORTED (0x54)
```

**With the applied patch, we have successfully authenticated using the PIN code and signed documents without encountering any issues.**

We acknowledge that we may not be deeply familiar with the OpenSC codebase and the subtleties regarding ATRs. However, with the provided patch, we have successfully resolved the issue and verified that Gemalto IDPrime 940C works as expected.

Thank you for your time.


src/libopensc/card-idprime.c
```diff
@@ -61,6 +61,10 @@ static const struct sc_atr_table idprime_atrs[] = {
 	  "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:ff:ff:ff",
 	  "Gemalto IDPrime 940",
 	  SC_CARD_TYPE_IDPRIME_940, 0, NULL },
+	{ "3b:7f:96:00:00:80:31:80:65:b0:85:05:00:39:12:0f:fe:82:90:00",
+	  "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:ff:ff:ff",
+	  "Gemalto IDPrime 940C",
+	  SC_CARD_TYPE_IDPRIME_940, 0, NULL },
```
Jakuje, line 67: Passing the ATR into the ATR parser, the block with difference is a block `pre-issuing data`, but it looks like the rest is the same for also the IDPrime 840 so I am ok with the change.

https://smartcard-atr.apdu.fr/parse?ATR=3b:7f:96:00:00:80:31:80:65:b0:85:05:00:39:12:0f:fe:82:90:00

Jakuje (approved):
looks good! Thank you for your contribution and confirmation it works for you!

frankmorgner (approved):

Jakuje:
This one is trivial enough so we can get it into the 0.24.0 too.

PR #2945, opened 2023-11-27
Fix more fuzzing issues and alternative proposal for #2904 (which looks stuck)

Jakuje:
I included also fix for #2904 as the fuzzer bumped into that after fixing the https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64522 with the same input.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-epass2003.c
+++ b/src/pkcs15init/pkcs15-epass2003.c
@@ -172,7 +172,7 @@ static int epass2003_pkcs15_create_dir(struct sc_profile *profile,
 
 		for (i = 0; create_efs[i]; ++i) {
 			if (sc_profile_get_file(profile, create_efs[i], &file)) {
-				sc_log(card->ctx, 
+				sc_log(card->ctx,
 					 "Inconsistent profile: cannot find %s",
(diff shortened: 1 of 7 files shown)


src/pkcs15init/pkcs15-epass2003.c
```diff
@@ -578,17 +578,17 @@ static int epass2003_pkcs15_generate_key(struct sc_profile *profile,
...
 	{
-	
+
 		r = cosm_new_file(profile, card, SC_PKCS15_TYPE_PUBKEY_EC, idx, &pukf);
 	}
 	else
 	{
-		
+
 		r = cosm_new_file(profile, card, SC_PKCS15_TYPE_PUBKEY_RSA, idx, &pukf);
```
xhanulik, line 587: nit: the empty lines can be removed too

Jakuje: removed. Thanks.

xhanulik (approved):

Jakuje:
Thanks for review! Lets see what next batch of fuzzing will bring us ...

PR #2948, opened 2023-11-30
Constant time RSA PKCS#1 v1.5 depadding

xhanulik:
This PR implements constant-time PKCS#1 v1.5 depadding and return value handling after RSA decryption. The returned value and depadded message are handled in the same way, both for valid and invalid padding, not to disclose the result of the depadding operation in the time side-channel.

- The details on the related Marvin attack can be found on https://people.redhat.com/~hkario/marvin/. 
- CVE 2023 5992 (https://github.com/OpenSC/OpenSC/wiki/CVE-2023-5992)

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/internal.h
+++ b/src/libopensc/internal.h
@@ -175,8 +175,8 @@ int _sc_card_add_xeddsa_alg(struct sc_card *card, size_t key_length,
 
 int sc_pkcs1_strip_01_padding(struct sc_context *ctx, const u8 *in_dat, size_t in_len,
 		u8 *out_dat, size_t *out_len);
-int sc_pkcs1_strip_02_padding(struct sc_context *ctx, const u8 *data, size_t len,
-		u8 *out_dat, size_t *out_len);
+int sc_pkcs1_strip_02_padding_constant_time(sc_context_t *ctx, unsigned int n, const u8 *data,
+		unsigned int data_len, u8 *out, unsigned int *out_len);
 int sc_pkcs1_strip_digest_info_prefix(unsigned int *algorithm,
 		const u8 *in_dat, size_t in_len, u8 *out_dat, size_t *out_len);
 #ifdef ENABLE_OPENSSL
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -308,10 +308,10 @@ int sc_pkcs15_decipher(struct sc_pkcs15_card *p15card,
 
 	/* Strip any padding */
 	if (pad_flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
-		int s = r;
-		int key_size = alg_info->key_length;
+		unsigned int s = r;
+		unsigned int key_size = (unsigned int)alg_info->key_length;
 		r = sc_pkcs1_strip_02_padding_constant_time(ctx, key_size / 8, out, s, out, &s);
-		LOG_TEST_RET(ctx, r, "Invalid PKCS#1 padding");
+		/* for keeping PKCS#1 v1.5 depadding constant-time, do not log error here */
 	}
 #ifdef ENABLE_OPENSSL
 	if (pad_flags & SC_ALGORITHM_RSA_PAD_OAEP)
@@ -333,7 +333,8 @@ int sc_pkcs15_decipher(struct sc_pkcs15_card *p15card,
 		LOG_TEST_RET(ctx, r, "Invalid OAEP padding");
 	}
 #endif
-	LOG_FUNC_RETURN(ctx, r);
+	/* do not log error code to prevent side channel attack */
+	return r;
 }
 
 /* derive one key from another. RSA can use decipher, so this is for only ECDH
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -308,8 +308,9 @@ int sc_pkcs15_decipher(struct sc_pkcs15_card *p15card,
 
 	/* Strip any padding */
 	if (pad_flags & SC_ALGORITHM_RSA_PAD_PKCS1) {
-		size_t s = r;
-		r = sc_pkcs1_strip_02_padding(ctx, out, s, out, &s);
+		int s = r;
+		int key_size = alg_info->key_length;
+		r = sc_pkcs1_strip_02_padding_constant_time(ctx, key_size / 8, out, s, out, &s);
 		LOG_TEST_RET(ctx, r, "Invalid PKCS#1 padding");
 	}
 #ifdef ENABLE_OPENSSL
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -23,8 +23,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include "sc-pkcs11.h"
 #include "common/compat_overflow.h"
+#include "common/constant-time.h"
+#include "sc-pkcs11.h"
 
 /* Also used for verification data */
 struct hash_signature_info {
@@ -1093,7 +1094,9 @@ sc_pkcs11_decr(struct sc_pkcs11_session *session,
 	rv = op->type->decrypt(op, pEncryptedData, ulEncryptedDataLen,
 	                       pData, pulDataLen);
 
-	if (rv != CKR_BUFFER_TOO_SMALL && pData != NULL)
+	/* terminate session for any return value except CKR_BUFFER_TOO_SMALL,
+	 * perform check in time side-channel free way to prevent Marvin attack */
+	if (!constant_time_eq_s(rv, CKR_BUFFER_TOO_SMALL) && pData != NULL)
 		session_stop_operation(session, SC_PKCS11_OPERATION_DECRYPT);
 
 	return rv;
@@ -1114,10 +1117,12 @@ sc_pkcs11_decr_update(struct sc_pkcs11_session *session,
 	rv = op->type->decrypt_update(op, pEncryptedData, ulEncryptedDataLen,
 			pData, pulDataLen);
 
-	/* terminate session for any error except CKR_BUFFER_TOO_SMALL */
-	if (rv != CKR_OK && rv != CKR_BUFFER_TOO_SMALL)
+	/* terminate session for any return value except CKR_BUFFER_TOO_SMALL,
+	 * perform check in time side-channel free way to prevent Marvin attack */
+	if (~constant_time_eq_s(rv, CKR_OK) & ~constant_time_eq_s(rv, CKR_BUFFER_TOO_SMALL))
 		session_stop_operation(session, SC_PKCS11_OPERATION_DECRYPT);
-	LOG_FUNC_RETURN(context, (int)rv);
+	/* do not log error code to prevent side channel attack */
+	return rv;
 }
 
 CK_RV
@@ -1535,6 +1540,10 @@ sc_pkcs11_decrypt(sc_pkcs11_operation_t *operation,
 
 	if (pulDataLen)
 		*pulDataLen = ulDataLen;
+
+	/* Skip DecryptFinalize for PKCS#1 v1.5 padding to prevent time side-channel leakage */
+	if (((CK_MECHANISM_PTR)&operation->mechanism)->mechanism == CKM_RSA_PKCS)
+		return rv;
 
 	if (rv != CKR_OK)
 		return rv;
--- a/src/tests/unittests/Makefile.mak
+++ b/src/tests/unittests/Makefile.mak
@@ -1,11 +1,12 @@
 TOPDIR = ..\..\..
 
-TARGETS = asn1 compression pkcs15filter check_macro_reference_loop
+TARGETS = asn1 compression pkcs15filter check_macro_reference_loop strip_pkcs1_2_padding
 
 OBJECTS = asn1.obj \
 	compression.obj \
 	pkcs15-emulator-filter.obj \
-	check_macro_reference_loop.obj
+	check_macro_reference_loop.obj \
+	strip_pkcs1_2_padding.obj \
 	$(TOPDIR)\win32\versioninfo.res
 
 all: $(TARGETS)
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -8,7 +8,8 @@ dist_noinst_DATA = \
 	LICENSE.compat_getopt compat_getopt.txt \
 	compat_getopt_main.c \
 	README.compat_strlcpy compat_strlcpy.3
-noinst_HEADERS = compat_strlcat.h compat_strlcpy.h compat_strnlen.h compat_getpass.h compat_getopt.h simclist.h libpkcs11.h libscdl.h compat_overflow.h
+noinst_HEADERS = compat_strlcat.h compat_strlcpy.h compat_strnlen.h compat_getpass.h \
+	compat_getopt.h simclist.h libpkcs11.h libscdl.h compat_overflow.h constant-time.h
 
 AM_CPPFLAGS = -I$(top_srcdir)/src
 
@@ -43,7 +44,8 @@ TIDY_FILES = \
 	compat___iob_func.c \
 	compat_overflow.h compat_overflow.c \
 	simclist.c simclist.h \
-	libpkcs11.c libscdl.c
+	libpkcs11.c libscdl.c \
+	constant-time.h
 
 check-local:
 	if [ -x "$(CLANGTIDY)" ]; then clang-tidy -config='' --checks='$(TIDY_CHECKS)' --warnings-as-errors='$(TIDY_CHECKS)' -header-filter=.* $(addprefix $(srcdir)/,$(TIDY_FILES)) -- $(TIDY_FLAGS); fi
--- a/src/common/constant-time.h
+++ b/src/common/constant-time.h
@@ -1,0 +1,128 @@
+/* Original source: https://github.com/openssl/openssl/blob/9890cc42daff5e2d0cad01ac4bf78c391f599a6e/include/internal/constant_time.h */
+
+#ifndef CONSTANT_TIME_H
+#define CONSTANT_TIME_H
+
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(inline)
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+#define constant_inline inline
+#elif defined(__GNUC__) && __GNUC__ >= 2
+#elif defined(__GNUC__) && __GNUC__ >= 2
+#elif defined(_MSC_VER)
+#define constant_inline __inline
+#else
+#define constant_inline
+#endif
+#else			       /* use what caller wants as inline  may be from config.h */
+#define constant_inline inline /* inline */
+#endif
+
+/*-
+ * The boolean methods return a bitmask of all ones (0xff...f) for true
+ * and 0 for false. For example,
+ *      if (a < b) {
+ *        c = a;
+ *      } else {
+ *        c = b;
+ *      }
+ * can be written as
+ *      unsigned int lt = constant_time_lt(a, b);
+ *      c = constant_time_select(lt, a, b);
+ */
+
+static constant_inline unsigned int
+value_barrier(unsigned int a)
+{
+	volatile unsigned int r = a;
+	return r;
+}
+
+static constant_inline size_t
+value_barrier_s(size_t a)
+{
+	volatile size_t r = a;
+	return r;
+}
+
+/* MSB */
+static constant_inline size_t
+constant_time_msb_s(size_t a)
+{
+	return 0 - (a >> (sizeof(a) * 8 - 1));
+}
+
+static constant_inline unsigned int
+constant_time_msb(unsigned int a)
+{
+	return 0 - (a >> (sizeof(a) * 8 - 1));
+}
+
+/* Select */
+static constant_inline unsigned int
+constant_time_select(unsigned int mask, unsigned int a, unsigned int b)
+{
+	return (value_barrier(mask) & a) | (value_barrier(~mask) & b);
+}
+
+static constant_inline unsigned char
+constant_time_select_8(unsigned char mask, unsigned char a, unsigned char b)
+{
+	return (unsigned char)constant_time_select(mask, a, b);
+}
+
+static constant_inline size_t
+constant_time_select_s(size_t mask, size_t a, size_t b)
+{
+	return (value_barrier_s(mask) & a) | (value_barrier_s(~mask) & b);
+}
+
+/* Zero */
+static constant_inline unsigned int
+constant_time_is_zero(unsigned int a)
+{
+	return constant_time_msb(~a & (a - 1));
+}
+
+static constant_inline size_t
+constant_time_is_zero_s(size_t a)
+{
+	return constant_time_msb_s(~a & (a - 1));
+}
+
+/* Comparison*/
+static constant_inline size_t
+constant_time_lt_s(size_t a, size_t b)
+{
+	return constant_time_msb_s(a ^ ((a ^ b) | ((a - b) ^ b)));
+}
+
+static constant_inline unsigned int
+constant_time_lt(unsigned int a, unsigned int b)
+{
+	return constant_time_msb(a ^ ((a ^ b) | ((a - b) ^ b)));
+}
+
+static constant_inline unsigned int
+constant_time_ge(unsigned int a, unsigned int b)
+{
+	return ~constant_time_lt(a, b);
+}
+
+/* Equality*/
+
+static constant_inline unsigned int
+constant_time_eq(unsigned int a, unsigned int b)
+{
+	return constant_time_is_zero(a ^ b);
+}
+
+static constant_inline size_t
+constant_time_eq_s(size_t a, size_t b)
+{
+	return constant_time_is_zero_s(a ^ b);
+}
+
+#endif /* CONSTANT_TIME_H */
--- a/src/libopensc/padding.c
+++ b/src/libopensc/padding.c
@@ -32,9 +32,12 @@
 #include <string.h>
 #include <stdlib.h>
 
+#include "common/constant-time.h"
 #include "internal.h"
 #include "pkcs11/pkcs11.h"
 /* TODO doxygen comments */
+
+#define SC_PKCS1_PADDING_MIN_SIZE 11
 
 /*
  * Prefixes for pkcs-v1 signatures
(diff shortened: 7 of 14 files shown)


frankmorgner (approved):
We can see that it is very hard to avoid time leakage throughout the whole software stack. It is not simply a matter of implementing `sc_pkcs1_strip_02_padding_constant_time()` and being done with it. Obviously, some files needed to be touched that seem unrelated at first glance. I have some doubts that we can keep up this effort without accidentally introducing new leakage.

That's why we should think again about the impact of deactivating RSA PKCS#1 v1.5 depadding *in software*. If the impact should be too heavy, then a legacy option could be introduced to enable this via opensc.conf. For now, this PR seems to solve the issue, so I think we can discuss this seperately without pressure.

popovec:
I will try to draw my view on this PR. Let's assume that from the moment the token returns the APDU, OpenSC processes the APDU and returns the return value to the application in constant time. The attacker can find out from the timing attack on the application whether the decrypt operation was completed correctly or ended with an error. 
For this reason, it is not necessary that all OpenSC operations take place in constant time. The only place where we have to prevent a side channel attack is how it is decided whether the decryption went correctly, i.e. Is the padding okay?
We only need to ensure that the removal (check) of the padding takes a constant time for arbitrary input data. For incorrect padding, the operation must take time T1, and for correct padding time T2, and T1 may or may not be the same as T2.
I believe that many changes in this PR are unnecessary. We don't need constant time for buffer size tests, and testing whether the user is logged in.. we don't need to adjust the logging of return codes. I would minimize this PR so that checking/removing the padding takes time T1 for correct padding and time T2 for incorrect padding.

frankmorgner:
The matter is complicated as even a shortcut due to some error handling may lead to a side channel. If you want to dive into this, you may want to check out the marvin page and read the [technical paper](https://people.redhat.com/~hkario/marvin/marvin-attack-paper.pdf). I believe, the changes in this PR were guided by the [toolkit](https://github.com/tomato42/marvin-toolkit) for detecting time based side channels...

popovec:
As I suggested, even if we make the OpenSC code so robust that it still maintains a constant time during decryption (for both correct and incorrect resulting plaintext), the resulting processing in the application that called the pkcs11 opensc service will still be vulnerable to a side channel timing attack. Even if the pkcs1 type 2 padding is removed directly in the token, we can still determine from a side channel attack on the application whether the plaintext met / did not meet the pkcs1 type 2 conditions (because it is possible to determine from the timing whether the application is processing the decrypted result, or is going the other way - an error path). 

The only thing left is to consider how much we plan to complicate the OpenSC code, when a possible attacker calmly focuses on a timing attack that reveals the necessary information from the application calling opensc.

frankmorgner:
We should avoid leak of information that allows the attacker to *distinguish* a padding error from any other error, because this allows to gain information about the plaintext. This is also an API related problem, because the timing of handling buffers of different lengths are different. That's why changes in the PKCS#11/Minidriver/CTK code are needed. Above that, the attacker will only know that decryption failed without getting information about the plaintext.

It could be possible that I'm explaining this not well enough or even incorrectly. If you like, we can get you in touch with the author of the marvin paper.

popovec:
Yes, I understand that we mask absolutely everything... the attacker finds out that the decrypt was wrong, but he doesn't know if it was a mistake in the padding or another mistake. Although in my opinion, if the attacker finds out from the side channel that there was a decryption error, he can almost certainly rely on the fact that it was an error when removing the padding. You're right, we can't do more. It is OK.

src/libopensc/pkcs15-sec.c
```diff
@@ -332,7 +333,7 @@ int sc_pkcs15_decipher(struct sc_pkcs15_card *p15card,
 		LOG_TEST_RET(ctx, r, "Invalid OAEP padding");
 	}
 #endif
-	LOG_FUNC_RETURN(ctx, r);
```
popovec, line 335: /* do not log error code to prevent side channel attack */

tomato42:
> That's why we should think again about the impact of deactivating RSA PKCS#1 v1.5 depadding in software.

even if it's done in hardware, the error must be returned to the calling application in constant time

> As I suggested, even if we make the OpenSC code so robust that it still maintains a constant time during decryption (for both correct and incorrect resulting plaintext), the resulting processing in the application that called the pkcs11 opensc service will still be vulnerable to a side channel timing attack.

Correct, and applications like openssl should already be doing that; so, while it's hard, it's not impossible. For a library like opensc there are only two options: don't create a side-channel where there wasn't one, or don't support PKCS#1v1.5 decryption, anything else is a vulnerability in OpenSC. It won't matter if the calling code is processing the result in side-channel free way if the returned value already has the side-channel baked-in.

src/libopensc/internal.h
```diff
@@ -168,6 +168,8 @@ int sc_pkcs1_strip_01_padding(struct sc_context *ctx, const u8 *in_dat, size_t i
 		u8 *out_dat, size_t *out_len);
 int sc_pkcs1_strip_02_padding(struct sc_context *ctx, const u8 *data, size_t len,
```
tomato42, line 169: why old code is retained and not removed or replaced?

tomato42 (approved):

xhanulik:
The above comments should by fixed by now.

popovec:
There is error in this implementation for missing message - openssl correctly returns an empty file for an empty message, test  script:

`````
#!/bin/bash
set -e
openssl genrsa -out rsa1024-key.pem 1024
openssl rsa -in rsa1024-key.pem -pubout -out rsa1024-pub.pem

T0="\
000210ff101010101fffffffffffffffffffffffffffffffffffffffffffffff\
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff\
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff\
ffffffffffffffffffffffffffffffffffffffffffffffffffffffd010101000\
"
echo -n $T0|xxd -p -r > rsa_decrypt_padding_testfile.txt
openssl pkeyutl -encrypt -pkeyopt rsa_padding_mode:none -pubin -inkey rsa1024-pub.pem  -in rsa_decrypt_padding_testfile.txt -out rsa_encrypted_testfile_check.txt
echo $?
openssl pkeyutl -decrypt -inkey rsa1024-key.pem -in rsa_encrypted_testfile_check.txt -out rsa_decrypted_testfile_check.txt
echo $?
hd rsa_decrypted_testfile_check.txt
`````

MyEID card or OsEID simulator, padding removal turned off in opensc.conf:
`````
app default {
        card_driver myeid {
                disable_hw_pkcs1_padding = 1;
        }
}
`````
Using this PR ...
`````
P:409665; T:0x140574651076096 15:46:05.315338 [opensc-pkcs11] reader-pcsc.c:244:pcsc_internal_transmit: called
P:409665; T:0x140574651076096 15:46:05.336470 [opensc-pkcs11] reader-pcsc.c:335:pcsc_transmit: 
Incoming APDU (130 bytes):
00 02 10 FF 10 10 10 10 1F FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF D0 10 10 10 00 ................
90 00                                           ..

P:409665; T:0x140574651076096 15:46:05.336557 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:409665; T:0x140574651076096 15:46:05.336605 [opensc-pkcs11] apdu.c:539:sc_transmit: returning with: 0 (Success)
P:409665; T:0x140574651076096 15:46:05.336649 [opensc-pkcs11] card.c:523:sc_unlock: called
P:409665; T:0x140574651076096 15:46:05.336694 [opensc-pkcs11] card-myeid.c:1404:myeid_transmit_decipher: returning with: 128
P:409665; T:0x140574651076096 15:46:05.336739 [opensc-pkcs11] card-myeid.c:1429:myeid_decipher: returning with: 128
P:409665; T:0x140574651076096 15:46:05.336783 [opensc-pkcs11] sec.c:50:sc_decipher: returning with: 128
P:409665; T:0x140574651076096 15:46:05.336827 [opensc-pkcs11] card.c:523:sc_unlock: called
P:409665; T:0x140574651076096 15:46:05.336871 [opensc-pkcs11] pkcs15-sec.c:169:use_key: returning with: 128
P:409665; T:0x140574651076096 15:46:05.336915 [opensc-pkcs11] padding.c:160:sc_pkcs1_strip_02_padding_constant_time: called
P:409665; T:0x140574651076096 15:46:05.336965 [opensc-pkcs11] card.c:523:sc_unlock: called
P:409665; T:0x140574651076096 15:46:05.337010 [opensc-pkcs11] reader-pcsc.c:742:pcsc_unlock: called
P:409665; T:0x140574651076096 15:46:05.337191 [opensc-pkcs11] framework-pkcs15.c:4620:pkcs15_prkey_decrypt: Decryption complete.
P:409665; T:0x140574651076096 15:46:05.337281 [opensc-pkcs11] mechanism.c:1123:sc_pkcs11_decr_update: returning with: -1412 (Wrong padding)
`````

**Comment by xhanulik** [comment]
Thanks for catching, there was a redundant check for a empty message in the depadding function. Added  also the corresponding unit test.

**Comment by popovec** [comment]
Thank you, at the moment the tests are going without any problems.

Just for fun, I measured the approximate depadding time (average of 10 tests, i5-3470, Linux version 6.1.0-15-rt-amd64) -  time  from APDU response to C_finalize() , with PR/without this PR: 

- empty message .. depadding time 575uS/1267uS
- correct padding about 100 bytes of message ... 1121uS/1331uS
- short padding   (00 inside 8 bytes after initial  00:02:) 458 uS/947uS
- wrong init bytes (00:01:..) 826uS/564uS
- wrong init bytes (01:02:...)  829uS/650uS

The truth is that the attacker has worse options for measuring this time, but thanks to other bugs in current processors, we have no way to prevent him from accessing this information. 

Anyway, I agree with the fact that we try to disable the possible attack as much as possible, and this PR contributes to that.

**Comment by tomato42** [comment]
@popovec Those results don't look right to me... Veronika has been running tests to verify that there are no differences in behaviour on the nanosecond level, not µs level. If, in your testing, you're running the same ciphertext over and over, and just averaging out the time over such repeated measurements, then you're not measuring the differences in the behaviour of the code, you're measuring the average frequency of the CPU that was performing the test. Please read the Out of the Box Testing paper linked from the Marvin page for details, and the [marvin-toolkit](https://github.com/tomato42/marvin-toolkit/) for a practical example of applying conclusion from that paper.

**Comment by popovec** [comment]
Let's describe once again the methodology I use:
We use an empty message (as I mentioned above) as the first test and the exact same message with the opening byte changed to 01. We measure the time from the arrival of the APDU to the return of C_Finalize. The measurement is performed only by editing OpenSC src/libopebsc/log.c, it prints the time in uS. 

Here is an example for some  measurements that I quickly made:
|Ok|wrong|
|---|---|
|0.000457 | 0.000377
|0.000854 | 0.000425
|0.000477 |0.000727
|0.000898 |0.000654
|0.000813 |0.000666
|0.000471 |0.00066
|0.000774 |0.000616
|0.000826 |0.000367
|0.000812 |0.000334
|0.00087 |0.000666
|0.00046 |0.000363
|0.000873| 0.000382

Yes, sometimes it is reversed, i.e. incorrect padding takes longer than correct padding, but  it seems to me that it makes it possible to identify quite precisely which message has faulty padding..

**Comment by xhanulik** [comment]
I have measured the whole `C_Decrypt` function with the virtual card but without the APDU calls and actual decryption (to speed up measurements and eliminate noise) - with the input as padded plaintext, copied into the output buffer. The modified `pkcs11-tool` for the testing is [here](https://github.com/xhanulik/OpenSC/commits/constant-pkcs1-test/).

For generating the input plaintexts and performing the time analysis, I used [marvin-toolkit](https://github.com/tomato42/marvin-toolkit) (slightly adjusted to generate plaintexts instead of ciphertexts). The final [report.txt](https://github.com/OpenSC/OpenSC/files/13750064/report.txt) looks ok. However, I can rerun the tests with the current state of PR to double-check.

**Comment by popovec** [comment]
Your testing method seems perfectly fine to me. I looked through the entire PR and it also seems fine. I still don't know the reason why the measurement I made gives such a good indication of the nature of the padding.

I will try to find out more precisely where the problem arises. The log shows which lines I use to measure the time difference:
`````
Incoming APDU (130 bytes):
00 02 10 FF 10 10 10 10 1F FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF ................
FF FF FF FF FF FF FF FF FF FF FF D0 10 10 10 00 ................
90 00                                           ..

P:541144; T:0x140353344187904 08:33:38.583479 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
P:541144; T:0x140353344187904 08:33:38.583500 [opensc-pkcs11] apdu.c:539:sc_transmit: returning with: 0 (Success)
P:541144; T:0x140353344187904 08:33:38.583520 [opensc-pkcs11] card.c:523:sc_unlock: called 
P:541144; T:0x140353344187904 08:33:38.583539 [opensc-pkcs11] card-myeid.c:1404:myeid_transmit_decipher: returning with: 128
>>>>>P:541144; T:0x140353344187904 08:33:38.583559 [opensc-pkcs11] card-myeid.c:1429:myeid_decipher: returning with: 128
P:541144; T:0x140353344187904 08:33:38.583578 [opensc-pkcs11] sec.c:50:sc_decipher: returning with: 128
P:541144; T:0x140353344187904 08:33:38.583597 [opensc-pkcs11] card.c:523:sc_unlock: called
P:541144; T:0x140353344187904 08:33:38.583616 [opensc-pkcs11] pkcs15-sec.c:169:use_key: returning with: 128
P:541144; T:0x140353344187904 08:33:38.583636 [opensc-pkcs11] padding.c:160:sc_pkcs1_strip_02_padding_constant_time: called
P:541144; T:0x140353344187904 08:33:38.583659 [opensc-pkcs11] card.c:523:sc_unlock: called
P:541144; T:0x140353344187904 08:33:38.583679 [opensc-pkcs11] reader-pcsc.c:742:pcsc_unlock: called
P:541144; T:0x140353344187904 08:33:38.583838 [opensc-pkcs11] framework-pkcs15.c:4620:pkcs15_prkey_decrypt: Decryption complete.
P:541144; T:0x140353344187904 08:33:38.583891 [opensc-pkcs11] pkcs11-object.c:1038:C_Decrypt: C_Decrypt()
P:541144; T:0x140353344187904 08:33:38.583920 [opensc-pkcs11] pkcs11-session.c:163:C_CloseSession: C_CloseSession(0x556b58289a50)
P:541144; T:0x140353344187904 08:33:38.583942 [opensc-pkcs11] pkcs11-session.c:109:sc_pkcs11_close_session: real C_CloseSession(0x556b58289a50)
P:541144; T:0x140353344187904 08:33:38.583966 [opensc-pkcs11] pkcs15-pin.c:851:sc_pkcs15_pincache_clear: called
P:541144; T:0x140353344187904 08:33:38.584003 [opensc-pkcs11] framework-pkcs15.c:1973:pkcs15_logout: Clearing PIN state without calling sc_logout()
>>>>>P:541144; T:0x140353344187904 08:33:38.584027 [opensc-pkcs11] pkcs11-global.c:415:C_Finalize: C_Finalize()
`````

**Comment by popovec** [comment]
One more thing, pkcs11-tool tries to call C_Decrypt first, if it fails (it ends with an error) it tries to call C_DecryptUpdate (even if it is non-standard, it can also be used for asymmetric decryption),  however, this ends in the case of incorrect padding as follows:

`error: PKCS11 function C_DecryptUpdate failed: rv = unknown PKCS11 error (0xfffffa7c)`
`````
P:542859; T:0x140565198802432 11:44:26.632504 [opensc-pkcs11] padding.c:160:sc_pkcs1_strip_02_padding_constant_time: called
P:542859; T:0x140565198802432 11:44:26.632546 [opensc-pkcs11] card.c:523:sc_unlock: called
P:542859; T:0x140565198802432 11:44:26.632574 [opensc-pkcs11] reader-pcsc.c:742:pcsc_unlock: called
P:542859; T:0x140565198802432 11:44:26.632739 [opensc-pkcs11] framework-pkcs15.c:4620:pkcs15_prkey_decrypt: Decryption complete.
P:542859; T:0x140565198802432 11:44:26.632793 [opensc-pkcs11] mechanism.c:1123:sc_pkcs11_decr_update: returning with: -1412 (Wrong padding)
P:542859; T:0x140565198802432 11:44:26.632826 [opensc-pkcs11] pkcs11-object.c:1062:C_DecryptUpdate: C_DecryptUpdate() = 0xFFFFFFFFFFFFFA7C
P:542859; T:0x140565198802432 11:44:26.632858 [opensc-pkcs11] pkcs11-global.c:415:C_Finalize: C_Finalize()
`````
When I conducted my measurements, I was actually measuring part of C_DecryptUpdate and not C_Decrypt itself. 
I measured it again on a completely different processor - Quad-Core AMD Opteron(tm) Processor 2352, the log shows the part that was measured:
`````
P:544673; T:0x140544698334720 12:17:30.697226 [opensc-pkcs11] card-myeid.c:1429:myeid_decipher: returning with: 128
P:544673; T:0x140544698334720 12:17:30.697245 [opensc-pkcs11] sec.c:50:sc_decipher: returning with: 128
P:544673; T:0x140544698334720 12:17:30.697265 [opensc-pkcs11] card.c:523:sc_unlock: called
P:544673; T:0x140544698334720 12:17:30.697284 [opensc-pkcs11] pkcs15-sec.c:169:use_key: returning with: 128
P:544673; T:0x140544698334720 12:17:30.697303 [opensc-pkcs11] padding.c:160:sc_pkcs1_strip_02_padding_constant_time: called
P:544673; T:0x140544698334720 12:17:30.697327 [opensc-pkcs11] card.c:523:sc_unlock: called
P:544673; T:0x140544698334720 12:17:30.697346 [opensc-pkcs11] reader-pcsc.c:742:pcsc_unlock: called
P:544673; T:0x140544698334720 12:17:30.697501 [opensc-pkcs11] framework-pkcs15.c:4620:pkcs15_prkey_decrypt: Decryption complete.
`````
|correct| fail(Decrypt)|fail(decrypt update)
|------|----|------
|0.000436|0.000465|0.000467
|0.000476|0.000466|0.000465
|0.000454|0.000491|0.000507
|0.000458|0.000418|0.000495
|0.000276|0.000255|0.000284
|0.000465|0.000279|0.000273
|0.000456|0.000473|0.000472
|0.000467|0.000469|0.000445
|0.000277|0.00047|0.000533
|0.000458|0.000499|0.000482

I don't dare to judge whether there is any correlation between padding and time, rather no than yes. 
Unless someone has a different opinion, I consider this PR suitable for acceptance.

**Comment by tomato42** [comment]
hmm, isn't it actually incorrect to call `C_DecryptUpdate` on a context returned by `C_DecryptInit` when `C_Decrypt` was already called?

Or the behaviour you're talking about is for handling a buggy PKCS#11 driver, that doesn't implement `C_Decrypt` but does `C_DecryptInit`? That's probably fine for a command line tool, but will create a side-channel when done on library level.

The measurements you've quoted are too variable to be able to say anything about presence or absence of a side-channel:

> Yes, sometimes it is reversed, i.e. incorrect padding takes longer than correct padding, but it seems to me that it makes it possible to identify quite precisely which message has faulty padding

unless there's a consistent difference, then no, it can't be used like that

**Comment by popovec** [comment]
> hmm, isn't it actually incorrect to call `C_DecryptUpdate` on a context returned by `C_DecryptInit` when `C_Decrypt` was already called?

`pkcs11-tool` is also used for symmetric decryption. In this tool, we cannot say in advance whether the pkcs11 module to be used has support for C_Decrypt for symmetric ciphers. Therefore, a decryption attempt is first made via C_Decrypt, and in case of failure, the attempt is repeated via C_DecryptUpdate (at the same time, a new initialization will take place using C_DecrytpInit). Yes, we can modify the `pkcs11 tool` to no longer decrypt with C_DecryptUpdate for certain return codes from C-Decrypt. Feel free to submit a PR.

> unless there's a consistent difference, then no, it can't be used like that

When I took the measurements on the i5, the correlation of time and type of padding was strangely very accurate... I don't know why. It was probably a coincidence. However, the measurements on the Opteron are far less convincing. 

@xhanulik: please check if we don't have a condition in C_DecryptUpdate that should also be executed in constant time.
Thank you.

**Comment by tomato42** [comment]
Let me ask it the other way: `pkcs11-tool` is a debugging and management tool, it's not the intended interface for any automation around encryption and decryption?

> When I took the measurements on the i5, the correlation of time and type of padding was strangely very accurate

@xhanulik please correct me if you have the data, but based on the necessary changes I wouldn't expect the side-channel to be larger than 10 µs. With a processing time in the order of 500 µs, st.dev of 90µs, with just 10 measurements even Student t test is unlikely to have a statistically significant result with normally distributed data. So yes, very likely a coincidence.

**Comment by popovec** [comment]
> Let me ask it the other way: `pkcs11-tool` is a debugging and management tool, it's not the intended interface for any automation around encryption and decryption?

We can safely expect that in programming languages that do not have pkcs11 support in their libraries, anyone can call pkcs11-tool via execve() or a similar call. An example can be a bash script where someone uses the pkcs11-tool to decrypt some data. Even if the relevant programming language has pkcs11 support in it, someone can call pkcs11-tool as a decryption tool...

Much more dangerous things happen on the internet, people download an unverified docker file and run it…

**Comment by popovec** [comment]
src/libopensc/pkcs15-sec.c
`````
-       LOG_FUNC_RETURN(ctx, r);
+       /* do not log error code to prevent side channel attack */
+       return r;
`````
I was particularly interested in this place in this PR, why not record the return value to opensc if a few places later in the user application there is a change in the behavior of the program flow, and here one way or another the error type will show up.

But I accept it, in that case I have to modify similarly:
`````
diff --git a/src/pkcs11/pkcs11-object.c b/src/pkcs11/pkcs11-object.c
index 99219eeb3..b02391121 100644
--- a/src/pkcs11/pkcs11-object.c
+++ b/src/pkcs11/pkcs11-object.c
@@ -1059,7 +1059,8 @@ C_DecryptUpdate(CK_SESSION_HANDLE hSession,  /* the session's handle */
                rv = sc_pkcs11_decr_update(session, pEncryptedPart, ulEncryptedPartLen,
                                pPart, pulPartLen);
 
-       SC_LOG_RV("C_DecryptUpdate() = %s", rv);
+       /* do not log error code to prevent side channel attack */
+       SC_LOG("C_DecryptUpdate()");
        sc_pkcs11_unlock();
        return rv;
 }
@@ -1087,7 +1088,8 @@ C_DecryptFinal(CK_SESSION_HANDLE hSession,   /* the session's handle */
                rv = reset_login_state(session->slot, rv);
        }
 
-       SC_LOG_RV("C_DecryptFinal() = %s", rv);
+       /* do not log error code to prevent side channel attack */
+       SC_LOG("C_DecryptFinal()");
        sc_pkcs11_unlock();
        return rv;
 }
`````

**Comment by popovec** [comment]
sc_pkcs11_decrypt():
`````
???        if (pulDataLen)
                *pulDataLen = ulDataLen;

        /* Skip DecryptFinalize for PKCS#1 v1.5 padding to prevent time side-channel leakage */
        if (((CK_MECHANISM_PTR) &operation->mechanism)->mechanism == CKM_RSA_PKCS) {
???                if (pulDataLen)
                        *pulDataLen = ulDataLen;
                return rv;
        }
`````
Isn't it a duplicate condition?

**Comment by popovec** [comment]
pkcs15_prkey_decrypt(): 

We get a mix of error codes (CKR_  and SC_ERROR_) :

`rv_pkcs11 = constant_time_select_s(good, CKR_OK, SC_ERROR_WRONG_PADDING);`

`SC_ERROR WRONG PADDING` should first be converted to the corresponding CKR error:

`rv_pkcs11 = constant_time_select_s(good, CKR_OK, sc_to_cryptoki_error(SC_ERROR_WRONG_PADDING, "C_Decrypt"));`

**Comment by xhanulik** [comment]
> @xhanulik: please check if we don't have a condition in C_DecryptUpdate that should also be executed in constant time. Thank you.

I'll take a look.

> I was particularly interested in this place in this PR, why not record the return value to opensc if a few places later in the user application there is a change in the behavior of the program flow, and here one way or another the error type will show up.

The return value should not be logged, since processing different error codes might take different time. I'll check also the other paths, thanks for pointing out.

**Comment by xhanulik** [comment]
According to https://github.com/OpenSC/OpenSC/pull/2948#issuecomment-1867568003 and https://github.com/OpenSC/OpenSC/pull/2948#issuecomment-1868224785, I have adjusted the logging of return values and conditions to be constant-time also in `C_DecryptUpdate` and `C_DecryptFinal`.

However, as the `C_DecryptUpdate` is called in `pkcs11-tool` when `C_Decrypt` fails, it would need some more adjustments, such as skipping the `C_DecryptUpdate` for some error return values or something like that.

**Comment by popovec** [comment]
I will also look at how C_DecryptFinal could be masked.

I have one more proposal, currently we have created a branch in `pkcs15_prkey_decrypt()`, which processes the error code/buffer in constant time for SC_ALGORITHM_RSA_PAD_PKCS1. The question is, wouldn't it be easier to manage everything in constant time, i.e. also the output of the RSA raw operation or of other depadding (OAEP). The time impact is insignificant here (the longest operation takes a token and data transfer). The code would be simplified to one branch, which would always execute in constant time.

**Comment by tomato42** [comment]
Unless `pkcs11-tool` works very differently than most other command line tools, I don't think it's fixable: we can't write different amount of bytes to an output file or print them to stdout in the same amount of time. (it still should follow good programming practice so people that use it for example code don't shoot themselves in the foot, but I don't expect we can make it actually secure)

**Comment by popovec** [comment]
We don't need to fix pkcs11-tool, this is an application that is part of opensc, but de facto only uses the pkcs11 opensc module, i.e. let's focus only on the fact that the opensc pkcs11 module is masked in such a way that it does not allow the misuse of information from the faulty depadding pkcs#1 v1.5 type2. (if someone writes his own application calling opensc, we will not prevent him from calling RSA decrypt in any way).

We need to ensure that a user who writes code that calls the pkcs11 interface of the opensc module and uses C_DecryptInit, C_DecryptUpdate, C_DecryptFinal has the same masked execution of depadding as a user who uses C_DecryptInit, C_Decrypt.

And at the same time, I still don't have an answer to the question, what about the depadding that was done on the token? Can we really mask  him in properly? If no, for all cards that support RSA_RAW is better to turn depadding on card off and do depadding in opensc,  it will be safer than depadding on the card. If some card does not support RSA_RAW, we have to leave the depadding on the token, is it safe?

**Comment by tomato42** [comment]
> And at the same time, I still don't have an answer to the question, what about the depadding that was done on the token? Can we really mask him in properly?

If the token did everything properly, then we need to pass in to buffers and error values up the stack without inspecting them. That will ensure that OpenSC is not contributing to the side-channel: if the token doesn't leak, then it won't leak when combined with opensc, if it does, then there are limits on what OpenSC can do. Yes, doing RAW RSA and doing depadding in sofware is one option, but that requires support for RAW RSA in the token.

I don't know side-channel free the tokens are, I haven't tested any.

**Comment by popovec** [comment]
As I wrote above, we can measure very precisely how long a message from a token takes (practically, we can say with great certainty how many bytes the token sent - only two bytes - either an error code or code 0x9000 = an empty message, or a response of length 2 + 1 byte to 2+ XY bytes - according to the length of the message after depadding. 

If you need to know where there was a problem in depadding (i.e. if the token didn't send 00 at the beginning or didn't send the next 02 or then there were less than 8 random non-zero bytes) you simply can't find out from a PC environment. A possible DPA/SPA attack on the token would be much more effective.

Consider also the fact that if you try the token using specific ciphertexts, you will not be able to do more than, say, 10,000 tests in 1 hour .. (or maybe less).

**Comment by tomato42** [comment]
There's no way to pad the messages between the smartcard and the PC?

> Consider also the fact that if you try the token using specific ciphertexts, you will not be able to do more than, say, 10,000 tests in 1 hour .. (or maybe less).

So, the attack according to [Bardou et.al](https://eprint.iacr.org/2012/417.pdf) would take just under 2 hours (see Table 1), provided that the measurement would be precise enough?

**Comment by popovec** [comment]
Let's assume that we can accurately measure how long it takes to copy the message (this is exactly the place that Verionika masks in the code pkcs15_prkey_decrypt() .. behind the note /* move everything from decrypted into out buffer, if rv is ok */)  Without this PR, you can measure it quite accurately - otherwise we wouldn't mask it. 

Similar to how Veronika masks the copying of the resulting  message to the output buffer, we would have to mask all the places where the data coming from the token is handled. This concerns the part of the opensc code that communicates with the card (pcsc or other) .. then the pcscd code itself, then the part of the code in the OS core that ensures the transfer from the reader (either USB or via serial). 

Even though USB transfer can mess up this "timing" data quite unpleasantly (for example, USB transfers data in chunks of 64 bytes, i.e. we can say with great certainty whether the transfer was under 64 bytes or several times 64 bytes).

**Comment by tomato42** [comment]
Yes, I understand. But the fix on the level of OpenSSL API is to always return 256 bytes long buffer when a 2048 bit key is used, and in separate variable return the length of actual message.

Is such behaviour impossible on the level of the pc/sc protocol?

**Comment by popovec** [comment]
What to do with a card reader that has a hardware limit of 64 bytes for one  USB transfer? .. see an example from lsusb -v:
`````
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
`````
For a depadded message, let's say 180 bytes, you measure at least 3 USB transfers, for an error 1 transfer .. and you can't do anything with that.  The T1 card protocol will be another place that will be difficult to mask.

Anytime later you try to mask this limitation, you will not succeed.

**Comment by tomato42** [comment]
Then, yes, sounds like it may be unfixable without implementing Marvin workaround in the card itself...

Do you have contacts to OEMs, so that we can inform them of the vulnerability? You can send the contacts to my work email: hkario at redhat.com

**Comment by popovec** [comment]
I'm afraid that here we would have to push these things through ISO7816-8. 

I think that from OpenSC's point of view it will be best to prepare (after accepting this PR) patches for each of the supported cards, which will make it possible to turn off hardware depadding on the cards (if possible). 

MyEID card already has support for turning off depadding in the card, but depadding is turned on by default. This will need to be changed.

**Comment by tomato42** [comment]
Yes, I don't think it needs to be in a single PR.

Speaking of wider efforts, I've been talking with Robert Relyea (co-chair of the OASIS PKCS#11 committee) about recommending Marvin workaround/implicit rejection on PKCS#11 level. I'm also working on an update to the PKCS#1 standard (through CFRG) here: https://datatracker.ietf.org/doc/draft-kario-rsa-guidance/ . I don't think I have the necessary contacts, or time, to influence ISO standards directly.

**Comment by xhanulik** [comment]
> I have one more proposal, currently we have created a branch in `pkcs15_prkey_decrypt()`, which processes the error code/buffer in constant time for SC_ALGORITHM_RSA_PAD_PKCS1. The question is, wouldn't it be easier to manage everything in constant time, i.e. also the output of the RSA raw operation or of other depadding (OAEP). The time impact is insignificant here (the longest operation takes a token and data transfer). The code would be simplified to one branch, which would always execute in constant time.

This should be fixed by now.

**Comment by popovec** [comment]
The final code is fine, but the readability of this PR (17 commits) is rather unclear. I would like it if the PR could be simplified somehow, after years it will be very difficult to read what was done and why. Don't fix your own code with new commits, use rebase and  fix the commit where the error in your PR originated. For example: commit `pkcs15_prkey_decrypt: Fix error code mismatch` should be removed and error should be fixed in commit `framework-pkcs15.c: Handle PKCS# 1 v1.5 depadding constant time`.

And one whitespace error:
````
$ git rebase master
$ git diff master --check
src/tests/unittests/strip_pkcs1_2_padding.c:152: trailing whitespace.
+                        0x00, 
````
coding style:
````
+# define CONSTANT_TIME_H
````
It will be better if we write it down without that space..  Try using clang-format - more in #2017

**Comment by xhanulik** [comment]
Sure, will do, I was just waiting to see if there were any other comments, thanks.

File: src/pkcs11/mechanism.c [hunk 3]
```diff
@@ -1536,6 +1541,10 @@ sc_pkcs11_decrypt(sc_pkcs11_operation_t *operation,
 	if (pulDataLen)
 		*pulDataLen = ulDataLen;
 
+	/* Skip DecryptFinalize for PKCS#1 v1.5 padding to prevent time side-channel leakage */
+	if (((CK_MECHANISM_PTR) &operation->mechanism)->mechanism == CKM_RSA_PKCS)
```
Jakuje, line 1545: nit: Cast type should be adjacent to the variable:
```suggestion
	if (((CK_MECHANISM_PTR)&operation->mechanism)->mechanism == CKM_RSA_PKCS)
```

Jakuje (approved):
just a minor nits.
I think for the pkcs11 we have test coverage. Could somebody test the minidriver works too? Or even measure the timing difference if possible?

src/tests/unittests/Makefile.mak
```diff
@@ -1,11 +1,12 @@
...
 
-TARGETS = asn1 compression pkcs15filter check_macro_reference_loop
+TARGETS = asn1 compression pkcs15filter check_macro_reference_loop strip_pkcs1_2_padding
 
 OBJECTS = asn1.obj \
 	compression.obj \
 	pkcs15-emulator-filter.obj \
-	check_macro_reference_loop.obj
+	check_macro_reference_loop.obj \
+	strip_pkcs1_2_padding.obj
```
Jakuje, line 9: should this line be terminated with the `\`  so the versioninfo is also part of the objects list?

Jakuje (approved):

frankmorgner (approved):
Looks good now

xhanulik:
Added minor fixes for includes order.

xhanulik:
I have also checked how the returned decrypted buffer is handled in the card drivers (supporting RSA_RAW operation) before the depadding is done. In most drivers, I haven't found any code dependent on the particular bytes in the buffer, so that should be ok.

There is only `card-tcos.c` where `tcos_decipher()` attempts to remove padding from the received buffer
https://github.com/OpenSC/OpenSC/blob/b21d01eee0817ae8130914e24cb8e2bc552dae09/src/libopensc/card-tcos.c#L657-L671
I don't know whether the card checks the padding, and this stripping is done only when the padding is correct, so I'm not sure whether this needs more care or not.

Jakuje:
Thank you! Merging to move along with the release.

PR #2959, opened 2023-12-14
Update Security policy to match current practices

Jakuje:
I do not know how exactly the workflow looks for non-owner, but this is taken from the pkcs11-provider security policy, where you get a choice to report security issue after clicking to "New Issue".

##### Checklist

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,18 +5,20 @@
 OpenSC releases are made roughly once a year, unless important security is discovered.
 
 OpenSC does not release micro updates for previously released versions and does not
-backport security fixes into them.
+backport security fixes into them. Only the last release is supported.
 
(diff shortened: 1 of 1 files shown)


tomato42:
I see in the Security tab that there's an option to file a private security bug, wouldn't it be better to recommend use of those?

Jakuje:
> I see in the Security tab that there's an option to file a private security bug, wouldn't it be better to recommend use of those?

Do you have a screenshot how it looks from your side? For me the Security tab has all the configuration and no button to report security vulnerability.

frankmorgner:
grafik

Jakuje:
Thanks. I was not expecting the button will be there for not logged in users. Updated.

tomato42 (approved):

PR #2960, opened 2023-12-14
Fixes mostly in pkcs11 layer and pkcs11-tool

Jakuje:
* Missing user constent check when doing signature
 * Not announcing WRAP/UNWRAP flags for RSA-PSS mechanisms
 * pkcs11-spy: Avoid null dereference
 * Simplify printing EC keys parameters

##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):

Jakuje:
Thank you! Merging.

PR #2968, opened 2023-12-23
MyEID: Avoid memory leaks

popovec:
Thanks oss-fuzz

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65299


Jakuje (approved):

Jakuje:
Thank you!

Issue #2952, opened 2023-12-01
Montgomery and (Edwards) Key Generation, Use, and Interoperability


dengert:
### Problem Description

These are comments while experimenting with OpenPGP on a Nitro Start (GUNK) and a Yubikey 5 NFC and OpenSSL 3.1.2 generated MONTGOMERY keys and OpenSC 0.24.0-rc2. The Nitro and Yubikey initialized using GnuPG 2.2.7  gpg-card command.  

Running  derive commands between any two of the above produce the same generic secret key as expected.

Example using Yubikey token and the Nitro public key with output to YK10-Nitro.der: 
```
/opt/ossl-3.1.2/bin/pkcs11-tool --slot 0 --module /opt/ossl-3.1.2/lib/opensc-pkcs11.so \
  -l --slot 0 --derive -m ECDH1-DERIVE -O -d 2 -i /tmp/derive.509765.other.pubkey.der -o YK10-Nitro.der
```

Example of OpenSSL private key and the Nitro public key:
```
openssl pkeyutl -derive -keyform der -inkey openssl.pkey.der  -peerform der \
     -peerkey NitroStart.pubkey.02.der > openssl-Nitro.der
```

The problem stems from curve "friendly" names and OIDs used by different programs, RFCs and other documents.

OpenSSL is following [RFC 8410](https://datatracker.ietf.org/doc/html/rfc8410)  using "OBJECT IDENTIFIER ::= { 1 3 101 110 }" I 

gpg-card uses "cv25519" as algorithm and does not list the OID.

[OpenPGP Card Specification 3.4.1](https://www.gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.1.pdf) appears to be the latest, but it does says anything about Edwards or Montgomery curves. 

Internally the Yubikey and Nitro store on card  `C2 0B 12 2B 06 01 04 01 97 55 01 05 01`  in the DO 6E. Labeled "Algorithm attributes decryption" which is:  Blob 'C2', length '0B', algorithm '12' and OID (minus the tag and length) '2B 06 01 04 01 97 55 01 05 01'

This problem of inconsistent names and OIDs is known. and [pkcs11-curr-v3.0-os](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.pdf)) says:
[...]

PR #2969, opened 2023-12-25
Workaround Yubikey bug with SELECT DATA

ya-isakov:
There is a known issue with Yubikey tokens, that they're requiring another length byte in APDU for SELECT DATA:
https://github.com/Yubico/yubikey-manager/issues/403

Let's workaround this, by checking return value of SELECT DATA, and if error indicates that parameters are wrong, let's try Yubikey-specific ones.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -71,7 +71,7 @@ static const struct sc_atr_table pgp_atrs[] = {
 	},
 	{ "3b:fc:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:4e:45:4f:72:33:e1", NULL, "Yubikey NEO (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ "3b:f8:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:34:d4", NULL, "Yubikey 4 (OpenPGP v2.1)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
-	{ "3b:fd:13:00:00:81:31:fe:15:80:73:c0:21:c0:57:59:75:62:69:4b:65:79:40", NULL, "Yubikey 5 (OpenPGP v3.4)", SC_CARD_TYPE_OPENPGP_V3, 0, NULL },
+	{ "3b:fd:13:00:00:81:31:fe:15:80:73:c0:21:c0:57:59:75:62:69:4b:65:79:40", NULL, "Yubikey 5 (OpenPGP v3.4)", SC_CARD_TYPE_OPENPGP_V3, SC_CARD_FLAG_YUBIKEY_SELECT, NULL },
 	{ "3b:da:18:ff:81:b1:fe:75:1f:03:00:31:f5:73:c0:01:60:00:90:00:1c", NULL, default_cardname_v3, SC_CARD_TYPE_OPENPGP_V3, 0, NULL },
 	{ NULL, NULL, NULL, 0, 0, NULL }
 };
@@ -311,6 +311,7 @@ pgp_match_card(sc_card_t *card)
 	i = _sc_match_atr(card, pgp_atrs, &card->type);
 	if (i >= 0) {
 		card->name = pgp_atrs[i].name;
+		card->flags = pgp_atrs[i].flags;
 		LOG_FUNC_RETURN(card->ctx, 1);
 	}
 	else {
@@ -1772,43 +1773,6 @@ pgp_get_pubkey_pem(sc_card_t *card, unsigned int tag, u8 *buf, size_t buf_len)
 }
 
 /**
- * Internal: Same as pgp_select_data, but with Yubikey workaround
- * https://github.com/Yubico/yubikey-manager/issues/403
- *
- * p1: number of an instance (DO 7F21: 0x00 for AUT, 0x01 for DEC and 0x02 for SIG)
- */
-static int
-pgp_select_data_yubikey(sc_card_t *card, u8 p1)
-{
-	sc_apdu_t	apdu;
-	u8	apdu_data[7];
-	int	r;
-
-	LOG_FUNC_CALLED(card->ctx);
-
-	sc_log(card->ctx, "select data with: %u (Yubikey workaround)", p1);
-	// Yubikey wants another length
-	apdu_data[0] = 0x06;
-	// create apdu data (taken from spec: SELECT DATA 7.2.5.)
-	apdu_data[1] = 0x60;
-	apdu_data[2] = 0x04;
-	apdu_data[3] = 0x5c;
-	apdu_data[4] = 0x02;
-	apdu_data[5] = 0x7f;
-	apdu_data[6] = 0x21;
-
-	// apdu, cla, ins, p1, p2, data, datalen, resp, resplen
-	sc_format_apdu_ex(&apdu, 0x00, 0xA5, p1, 0x04, apdu_data, sizeof(apdu_data), NULL, 0);
-
-	// transmit apdu
-	r = sc_transmit_apdu(card, &apdu);
-	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
-	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
-	LOG_TEST_RET(card->ctx, r, "Card returned error");
-	LOG_FUNC_RETURN(card->ctx, r);
-}
-
-/**
  * Internal: SELECT DATA - selects a DO within a DO tag with several instances
  * (supported since OpenPGP Card v3 for DO 7F21 only, see section 7.2.5 of the specification;
  *  this enables us to store multiple Card holder certificates in DO 7F21)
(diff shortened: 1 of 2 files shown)


Jakuje:
Thanks for the PR! Isn't there a way to detect yubikey before we will get the error to send the right APDU right away?

ya-isakov:
Very good question - I tried to find the way. It seems that in yubikey-manager, they use Yubikey-specific command to detect firmware version (https://github.com/Yubico/yubikey-manager/blob/main/yubikit/openpgp.py#L1018), which, I thought, is not the best approach for OpenSC. I cannot find any generic way to determine vendor, unfortunately.

dengert:
Using a Yubikey with OpenPGP applet to get ATR and version:

```
./opensc-tool --card-driver openpgp -a -s "00:f1:00:00:03"
Using reader with a card: Yubico YubiKey CCID 00 00
3b:f8:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:34:d4
Sending: `00 F1 00 00 03`
04 02 06 ...
```

Paste the ATR into https://smartcard-atr.apdu.fr/ 
The historic bytes `59 75 62 69 6B 65 79 34` spell "Yubikey4" The "Y" is also "proprietary format"
Then sending `00 F1 00 00 03` returns version number.


With PIV applet can also get version number sending `00 FD 00 00 03`

Jakuje:
Well. We already have a yubico_version as part of the PIV private structure:

https://github.com/OpenSC/OpenSC/blob/23f3df96883123aa0e52932119c71c818799e742/src/libopensc/card-piv.c#L424

Would using that to detect yubikey work?

dengert:
I believe it would for OpenPGP applet. Just a different command. Yubikey PIV applet uses `00 FD 00 00`  Yubikey OpenPGP applett uses `00 F1 00 00` as shown in example in https://github.com/OpenSC/OpenSC/pull/2969#issuecomment-1872432827

ya-isakov:
Hmm, is this ATR standard command? I did not want to use vendor-specific commands to detect version etc, to not cause issues with other tokens.

Jakuje:
> Hmm, is this ATR standard command? I did not want to use vendor-specific commands to detect version etc, to not cause issues with other tokens.

I would say it should be safe to assume the token is yubico if it is in the ATR historic bytes.

ya-isakov:
I was thinking, that there is already ATR matching, by pgp_atrs, and there is even flags field - so, I can add flag there, to allow Yubikey-specific commands/workarounds. But, I cannot find where these flags are read or passed.

dengert:
> I was thinking, that there is already ATR matching, by pgp_atrs, and there is even flags field - so, I can add flag there, to allow Yubikey-specific commands/workarounds. 

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L305-L315 will match a card by ATR, or in lines https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L320-L325 will match a card if it has a OpenPGP applet. In either case,  The ATR Historical bytes can be found at: `card->reader->atr_info.hist_bytes`

The ATR is for the card. A Yubikey card can have both a openpgp applet and a piv applet. (OpenSC can only use one at a time. 
based on first driver to find its applet will be used. See `OPENSC_DRIVER=` env or in `opensc.conf`) 

card-piv.c checks the Historical bytes here:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L5354-L5385

It the sets a card type. But with OpenPGP the card type is based on version of OpenPGP applet.  You could just set a flag saying card is a Yubikey and later call a routing like 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L5481-L5497 to save the Yubikey version. 
Note that with PIV applet, line https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L5485 it uses "0xFD"
but with OpenPGP applet you would use "0xF1" to get the version.

ya-isakov:
Please, check new commit - I added new flag, used it to select proper pgp_select_data..., and in pgp_select_data_yubikey I'm checking for Yubikey version, to see if it should try Yubikey-specific ATR. I'm a bit confused, that my APDU seems to be different from what @dengert shown - I'm sending `00 F1 00 00 01 03 03` (this is what sc_format_apdu_ex created, and I cannot find way to put data in sc_format_apdu), not `00:f1:00:00:03`, but it seems, this one is working, too - it returns proper versions for both of my Yubikey tokens.

dengert:
`SC_CARD_FLAG_YUBIKEY_SELECT` should not be in opensc.h. It should not be a new card type.

The point of reading the Yubikey version, is that Yubikey as fixed problem in 5.4.3 and OpenSC could then handle old and new cards as needed.

You could simplify the code by saving the version in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.h#L171-L190
In openpgp_init() test ATR historical bytes for `Yubikey`  and call pgp_yubikey_get_version(card, &priv->version);

Then test in  `pgp_select_data` `if (priv->yubikey_version > 0  &&   priv->yubikey_version < 0x050403)` change the string to be sent (and the length). 

 You would not need SC_CARD_FLAG_YUBIKEY_SELECT or special `pgp_select_data_yubikey`

ya-isakov:
@dengert 
> It should not be a new card type.
It is not a card type, it's a flag

> In openpgp_init() test ATR historical bytes for Yubikey and call pgp_yubikey_get_version(card, &priv->version);
Why to test for historical bytes 2nd time, if there is already historical bytes check via pgp_match_card?

ya-isakov:
So, yea, in Yubikey 5, last bytes of ATR are 59:75:62:69:4b:65:79:40 instead of 59:75:62:69:6b:65:79:34:d4 - so, YubiKey (with big `K`), instead of Yubikey4

dengert:
> It is not a card type, it's a flag

Then add a flag  that is only used in OpenPGP driver in `card-openpgp.h`

If you did want to have a new card type for OpenPGP card driver, add it in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/cards.h#L107
But note that the card types refer to the OpenPGP applet versions. The problem is Yubico had a bug in there version
of applet. 

> Why to test for historical bytes 2nd time,

Because if the ATRs is not matched, the code then tests for OpenPGP applet at 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L320-L325
This will then work with newer cards or OpenPGP applets on other cards not listed in the driver.  

> The whole historical bytes for Yubikey5 looks like this 80:73:c0:21:c0:57:59:75:62:69:4b:65:79 - I'm not sure, if I should compare with this whole line? It might have some settings in it.

You could almost copy this code to test for Yubikey cards:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L5348-L5385

Also look at https://smartcard-atr.apdu.fr/ and paste in the ATR to see what is in it. 

What cards do you have?

dengert:
@ya-isakov   While looking at `card-openpgp.c` for #2952  I notice `pgp_get_card_features` called from `pgp_init`  parses the ATR historical bytes expecting the card to follow the OpenPGP standards and provide some additional information. 

But of the 3 Yubikey devices listed in `pgp_atrs` on the last  `3b:fd:13:00:00:81:31:fe:15:80:73:c0:21:c0:57:59:75:62:69:4b:65:79:40` Yubikey 5 NFC  does not use the proprietary format, but uses the  `compact TLV data object` which includes some extra features, and  `card issuer's data` of "YubiKey".  The code should be able to handle future Yubikey devices with different ATRs even if not listed in  `pgp_atrs`

I would suggest the code be modified here to parse further for Yubikey and to then also read the Yubikey version number to handle the bug you are trying to address.

frankmorgner:
@ya-isakov will you have a second look at the suggestions above?

ya-isakov:
@frankmorgner Sorry, I don't have time now to do it as suggested.

frankmorgner:
OK, will mark this as draft then.

frankmorgner:
@ya-isakov , @dengert do you plan to continue the work on this pull request? If not, I'd like to close it.

dengert:
What it sounds like is some versions of Yubikey openpgp have a bug, with reading certificates. And it may be fixed in newer cards. 
I have not looked at it in a long time. From PIV experience the Yubikey version had many issues, and fixed at various times. 
I would rewrite it to always identify if this is a Yubikey from ATR, then read version number from pgp_init, and set a flag internal to card-openpgp.c (not in opensc.h). 

Do we know when the bug was introduced, and when it was fixed? 
Then in pgp_select_file test for version number, of 0, its not a yubikey then check version number.

Another approach is if the normal select file fails, then try the fixed version, and save flag that card has the bug.

frankmorgner:
I looked deeper into the implementation and I think @ya-isakov took care of the previous comments. Regardless of the question whether we already do have the yubikey version available, the detection mechanism @ya-isakov used should work as expected. The only change I'd like to see is that the flag for the select workaround is cleared during initialization if it is the wrong version of yubikey rather than checking this every time a select is made.

frankmorgner:
I have rebased this PR onto master and implemented the suggested change here https://github.com/OpenSC/OpenSC/pull/3332

frankmorgner:

PR #2974, opened 2024-01-07
iasecc: Avoid memory leaks

popovec:
Thanks oss-fuzz

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65562

### DIFF
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1544,6 +1544,7 @@ iasecc_delete_file(struct sc_card *card, const struct sc_path *path)
 	struct sc_apdu apdu;
 	struct sc_file *file = NULL;
 	int rv;
+	unsigned int file_id;
 
 	LOG_FUNC_CALLED(ctx);
 	sc_print_cache(card);
@@ -1554,13 +1555,17 @@ iasecc_delete_file(struct sc_card *card, const struct sc_path *path)
 	LOG_TEST_RET(ctx, rv, "Cannot select file to delete");
 
 	entry = sc_file_get_acl_entry(file, SC_AC_OP_DELETE);
+
+	file_id = file->id;
+	sc_file_free(file);
+
 	if (!entry)
 		LOG_TEST_RET(ctx, SC_ERROR_OBJECT_NOT_FOUND, "Cannot delete file: no 'DELETE' acl");
 
 	sc_log(ctx, "DELETE method/reference %X/%X", entry->method, entry->key_ref);
 	if (entry->method == SC_AC_SCB && (entry->key_ref & IASECC_SCB_METHOD_SM))   {
 		unsigned char se_num = entry->key_ref & IASECC_SCB_METHOD_MASK_REF;
-		rv = iasecc_sm_delete_file(card, se_num, file->id);
+		rv = iasecc_sm_delete_file(card, se_num, file_id);
 	}
 	else   {
 		sc_format_apdu(card, &apdu, SC_APDU_CASE_1, 0xE4, 0x00, 0x00);
@@ -1576,7 +1581,6 @@ iasecc_delete_file(struct sc_card *card, const struct sc_path *path)
 		card->cache.current_ef = NULL;
 	}
 
-	sc_file_free(file);
 	LOG_FUNC_RETURN(ctx, rv);
 }
 


popovec:
When I was looking at the code of the `iasecc_delete_file()` function, I found one thing, after deleting the file using SM, it does not update `card->cache.current_ef`. Can someone who knows details about the "iaescc" code confirm whether the following patch should be applied?
`````
index b20123d65..7ec1fd575 100644
@@ -1573,13 +1573,11 @@ iasecc_delete_file(struct sc_card *card, const struct sc_path *path)
                rv = sc_transmit_apdu(card, &apdu);
                LOG_TEST_RET(ctx, rv, "APDU transmit failed");
                rv = sc_check_sw(card, apdu.sw1, apdu.sw2);
-               LOG_TEST_RET(ctx, rv, "Delete file failed");
-
-               if (card->cache.valid) {
-                       sc_file_free(card->cache.current_ef);
-               }
-               card->cache.current_ef = NULL;
        }
+       LOG_TEST_RET(ctx, rv, "Delete file failed");
+       if (card->cache.valid)
+               sc_file_free(card->cache.current_ef);
+       card->cache.current_ef = NULL;
 
        LOG_FUNC_RETURN(ctx, rv);
 }

`````

File: src/libopensc/card-iasecc.c [hunk 1]
```diff
@@ -1544,6 +1544,7 @@ iasecc_delete_file(struct sc_card *card, const struct sc_path *path)
 	struct sc_apdu apdu;
 	struct sc_file *file = NULL;
 	int rv;
+	unsigned int file_id;
```
Jakuje, line 1547: `struct sc_file` member `id` is defined as a signed int so I think we should keep that (technically it should not matter as it should be just 2 bytes, but some future compiler might start complaining):
https://github.com/OpenSC/OpenSC/blob/c2860b0501a2f44307a063d4e4886ea37ec8306e/src/libopensc/types.h#L263

Jakuje:
> When I was looking at the code of the `iasecc_delete_file()` function, I found one thing, after deleting the file using SM, it does not update `card->cache.current_ef`. Can someone who knows details about the "iaescc" code confirm whether the following patch should be applied?

I would say yes, but I am not that much familiar with the iasecc. Maybe @vjardin ?

src/libopensc/card-iasecc.c
popovec: In that case, we have a problem in the following function prototype (where `file_id` is used):
`````
src/libopensc/iasecc-sm.c:iasecc_sm_delete_file(struct sc_card *card, unsigned se_num, unsigned int file_id)
src/libopensc/iasecc-sdo.h:int iasecc_sm_delete_file(struct sc_card *card, unsigned se_num, unsigned int file_id);
`````
We even find another conversion there:
`sm_info->cmd_data = (void *)(uintptr_t)file_id;`

Should I fix it everywhere?

vjardin:
> I would say yes, but I am not that much familiar with the iasecc. Maybe @vjardin ?

I would need to run some tests with some IASECC cards, but I do not have time for some short terms ; can it wait a bit ?

src/libopensc/card-iasecc.c
Jakuje: Oh. If it would be just this one internal function, I would go ahead and fix it to avoid needless conversion.
The other cast is weird. The cast back is in `src/smm/sm-card-iasecc.c`, but only for logging if I see right and I do not see if it is used for something else deeper in the code so I would probably not touch it and keep it as it is.

popovec: At this point, it doesn't really matter if I use `unsigned int file_id;` or `int file_id;` because the conversion will happen either way and this conversion is completely unambiguous in the case of a 16-bit number.

I think we should leave this PR as it is. Any fix  related to `file_id` (unsigned/signed) will take place in a separate PR .. and also a separate PR will be used for the problem with `card->cache.valid` and `card->cache.current_ef`.

Jakuje: Agreed. Thanks for looking into this comment!

Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you! Merging to move on with fuzzing. @popovec Please move the discussion regarding to the delete file either to new issue or draft PR so it wont get lost.

PR #2977, opened 2024-01-10
Minimal updated subset of changes from #2017 (clang-format)

Jakuje:
Based on last comment in #2017 from @popovec I pulled the .clang-format and some minimal changes to the separate branch, that can be merged if we will agree on that format.

I played with it a bit more, but I am still not completely satisfied with the current configuration as it sometimes formats some complicated constructs in non-intuitive way. But it is something we can try to tune later (or rewrite the code to avoid these constructs).

In any case, review here just to make sure the .clang-format, .editorconfig and description of the format in contributing.md matches would be welcomed. Similarly if we should add some more examples or some look odd.

##### Checklist


src/libopensc/card-openpgp.c
```diff
@@ -272,6 +273,7 @@ static pgp_do_info_t	pgp34_objects[] = {	/**** OpenPGP card spec 3.4 ****/
 	{ DO_ENCR_SYM, SIMPLE,      READ_ALWAYS | WRITE_PIN3,  pgp_get_pubkey_pem, NULL   },
 	{ 0, 0, 0, NULL, NULL },
 };
+// clang-format on
```
frankmorgner, line 276: please look at the formatting report for a single remark in this file:
https://github.com/OpenSC/OpenSC/actions/runs/7479501161/job/20356826376?pr=2977

frankmorgner (changes requested):
If you fix the remark from the new formatting CI workflow, this is good to go

src/libopensc/card-openpgp.c
Jakuje: Resolved the outstanding issues.

popovec:
I noticed that we currently have `TabWidth 4' defined. I don't have a problem with this setting, but I'll just remind you of the years-old debate about TAB in the Linux kernel.

https://public-inbox.org/git/634393B0-734A-4884-93E3-42F7D3CB157F@mit.edu/

probably the most explanatory email from the mentioned debate:

https://public-inbox.org/git/alpine.LFD.0.999.0710161722320.26902@woody.linux-foundation.org/#t

Jakuje:
> I noticed that we currently have `TabWidth 4' defined. I don't have a problem with this setting, but I'll just remind you of the years-old debate about TAB in the Linux kernel.

Oh, The editorconfig has tab width 8 and clang 4. This is clearly an inconsistency I wanted to catch before merging this. But it should not have any effect in the resulting formatting itself.

> https://public-inbox.org/git/634393B0-734A-4884-93E3-42F7D3CB157F@mit.edu/
> 
> probably the most explanatory email from the mentioned debate:
> 
> https://public-inbox.org/git/alpine.LFD.0.999.0710161722320.26902@woody.linux-foundation.org/#t

The indentation debates will never be over. I read through this mail, but it really did does not bring much insights to what is the reason and what are advantages. There was already a lot of discussion in the #2017.

I was used to tab size 4 most of the time, but I agree that tab width 8 have some advantages such as avoiding too much indentation or better readability, especially if we use spaces for continuation indentation. So I do not have strong opinion.

While trying to reformat some code with current configuration, there are still things I do not like I mentioned in the previous PR, so I will try again here if  there will be some other insights:

 * The `ContinuationIndentWidth: 16` with `IndentWidth: 8` and `TabWidth: 8 ` indents the structures with two tabs making it needlessly deep. The alignment of "columns" looks off as it uses mixture of tabs and spaces without any sense (or at least I do not see it there). Example:
```diff
 	CK_ATTRIBUTE template[] = {
-		{ CKA_CLASS, &class, sizeof(class) },
-		{ CKA_KEY_TYPE, &keyType, sizeof(keyType) },
-		{ CKA_LABEL, label, sizeof(label)-1 },
-		{ CKA_ENCRYPT, &_true, sizeof(_true) },
-		{ CKA_DECRYPT, &_true, sizeof(_true) },
-		{ CKA_VALUE, key, keysize },
-		{ CKA_VALUE_LEN, &keylen, sizeof(keylen) },
+			{CKA_CLASS,	    &class,   sizeof(class)	 },
+			{CKA_KEY_TYPE,  &keyType, sizeof(keyType)	},
+			{CKA_LABEL,	    label,	   sizeof(label) - 1},
+			{CKA_ENCRYPT,   &_true,   sizeof(_true)	   },
+			{CKA_DECRYPT,   &_true,   sizeof(_true)	   },
+			{CKA_VALUE,	    key,	 keysize		},
+			{CKA_VALUE_LEN, &keylen,	 sizeof(keylen)   },
 	};
```
* I do not like the continuation indentation of parameters of function call inside of `if` such as demonstrated here. Regardless the `UseTab: Always` it indents with a combination of 2 tabs and 4 spaces:
```diff
 	if (!find_object(session, CKO_PUBLIC_KEY, &hWrappingKey,
-			 opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
+			    opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
 		if (!find_object(session, CKO_SECRET_KEY, &hWrappingKey,
-				 opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
+				    opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
 			util_fatal("Public/secret key (wrapping key) not found");
```
 * With `else if` its 3 tabs and 3 spaces: 
```diff
 		} else if (!find_object(session, CKO_SECRET_KEY, &object,
-				 opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
+					   opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
 			util_fatal("Secret key not found");
```
 * some continuation is still a mystery for me. The split string is indented with one tab and 3 spaces and the remaining argument with normal 2 tabs:
```diff
 				util_fatal("Unknown key pair type %s, valid key types for mechanism GOSTR3410 are GOSTR3410-2001:{A,B,C},"
-					" GOSTR3410-2012-256:{A,B,C,D}, GOSTR3410-2012-512:{A,B,C}", type);
+					   " GOSTR3410-2012-256:{A,B,C,D}, GOSTR3410-2012-512:{A,B,C}",
+						type);
 
```

(now it tries to reformat also the json files with p11test results, but thats something we can handle)

frankmorgner:
I think most of the time we avoided a mixture of spaces and tabs by replacing 8 spaces with a tab so that IndentWidth and TabWidth should both be `8`

> The ContinuationIndentWidth: 16 with IndentWidth: 8 and TabWidth: 8 indents the structures with two tabs making it needlessly deep

I think ContinuationIndentWidth should be `8`, making it one additional indent level.

> - I do not like the continuation indentation of parameters of function call inside of if
> - With else if its 3 tabs and 3 spaces:

Your examples above originally contain a mixture of some tabs and a single leading space. That single space is superfluous and may confuse clang-format. could you please try removing this space first and then apply clang-format?

> - some continuation is still a mystery for me

That particular example looks strange, indeed.

Jakuje:
> > The ContinuationIndentWidth: 16 with IndentWidth: 8 and TabWidth: 8 indents the structures with two tabs making it needlessly deep
> 
> I think ContinuationIndentWidth should be `8`, making it one additional indent level.

My reading of the documentation is the value as used in the clang-format is not additional indent, but the whole indent so we need 16 if we want to use two tabs for continuation:

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#continuationindentwidth

The problem is that, I think, also the struct initialization is considered a continuation here. It should be solved with `BracedInitializerIndentWidth: 8`, but its only in clang 17 (i dont have it locally now to test).

> > * I do not like the continuation indentation of parameters of function call inside of if
> > * With else if its 3 tabs and 3 spaces:
> 
> Your examples above originally contain a mixture of some tabs and a single leading space. That single space is superfluous and may confuse clang-format. could you please try removing this space first and then apply clang-format?

It did not help. I think it tries to align something somewhere, but I do not see what and where.

I tried also different values for `UseTab`, but no improvement:
```diff
 		} else if (!find_object(session, CKO_PRIVATE_KEY, &object,
-				opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
+					   opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
 			if (!find_object(session, CKO_SECRET_KEY, &object,
-					opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
+					    opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, 0))
 				util_fatal("Private/secret key not found");
```

> > * some continuation is still a mystery for me
> 
> That particular example looks strange, indeed.

I though we could get around this with `AlwaysBreakBeforeMultilineStrings` by breaking before that long string, but it the output still looks the same -- broken string aligned with tab and spaces to the opening quote and the rest with the continuation indent ...

frankmorgner:
I don't see an easy way out of the formatting problems of clang-format. What about adding the configuration files as suggested right now, but allowing the workflow to fail?

frankmorgner (approved):
OK, fine. Thanks for the clarification.

PR #2981, opened 2024-01-12
Fix Heap-use-after-free in iasecc_delete_file

popovec:
Patch 773fcc6bd0d047aa629dd297e87244ca3add5b16 unfortunately introduced a new bug.  Here is a fix for both the old and the new bug.

Thanks oss-fuzz

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65562
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65684


Jakuje (approved):
Completely missed that the `sc_file_get_acl_entry()` keeps reference to file.

popovec:
I didn't notice it, I apologize. The `card->cache` issue was distracting me.

Jakuje:
Thank you!

Issue #2867, opened 2023-09-20
Given an SO-PIN, how do you change the PIN?


minfrin:
### Problem Description

I have a smartcard where the PIN and PUK have been lost. I have the SO-PIN, and would like to re-set the PIN. There does not seem to be a way to do this.

### Steps to reproduce

Test is good:

```
blackadder ~ # pkcs11-tool --login-type so --login --test
Using slot 0 with a present token (0x0)
Logging in to "xxx (Smartcard PIN)".
Please enter SO PIN: 
C_SeedRandom() and C_GenerateRandom():
  seeding (C_SeedRandom) not supported
  seems to be OK
Digests:
  all 4 digest functions seem to work
  MD5: OK
  SHA-1: OK
  RIPEMD160: OK
Signature: not a R/W session, skipping signature tests
Verify: not a R/W session, skipping verify tests
Decryption: not a R/W session, skipping decryption tests
No errors
```

Change PIN is no good, asks for the existing PIN:

```
blackadder ~ # pkcs11-tool --login-type so --login --change-pin
Using slot 0 with a present token (0x0)
Logging in to "xxx (Smartcard PIN)".
Please enter SO PIN: 
Please enter the current PIN: 
^C
```

Init PIN no good, fails complaining about the PIN length (with no way to indicate which PIN, I'm assuming this might be trying to change the SO PIN, which I do not want to do):

```
blackadder ~ # pkcs11-tool --login-type so --login --init-pin
Using slot 0 with a present token (0x0)
Logging in to "xxx (Smartcard PIN)".
Please enter SO PIN: 
Please enter the new PIN: 
Please enter the new PIN again: 
error: PKCS11 function C_InitPIN failed: rv = CKR_PIN_LEN_RANGE (0xa2)
Aborting.
```

The pkcs15-tool also appears to not work, we try change the PIN, and the SO-PIN is being changed (no):

```
blackadder ~ # pkcs15-tool --change-pin
Using reader with a card: ACS ACR39U ICC Reader 00 00
Enter old PIN [Security Officer PIN]: 
^C
```

Try change the PIN by specifying the auth-id, we're asked for the old PIN we don't have:

```
blackadder ~ # pkcs15-tool --auth-id 01 --change-pin
Using reader with a card: ACS ACR39U ICC Reader 00 00
Enter old PIN [Smartcard PIN]: 
^C
```

Is this even possible?

Jakuje:
[...]

PR #2983, opened 2024-01-12
docs: add example to initialize PIN with pkcs11-tool

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2867

##### Checklist

- [ ] macOS tokend is tested


src/tools/pkcs11-tool.c
```diff
@@ -2985,9 +2986,6 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 			size_t mtypes_num = sizeof(mtypes)/sizeof(mtypes[0]);
 			const char *p_param_set = type + strlen("GOSTR3410");
 
-			if (p_param_set == NULL)
-				util_fatal("Unknown key pair type %s, expecting GOSTR3410:<nbytes>", type);
-
```
Jakuje, line 2990: I think you did not want to remove this check as this could lead to using of NULL value in strcmp below and there is no default involved in the gost case.

src/tools/pkcs11-tool.c
```diff
@@ -2863,11 +2863,12 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 				if (!find_mechanism(slot, CKF_GENERATE_KEY_PAIR, mtypes, mtypes_num, &opt_mechanism))
 					util_fatal("Generate RSA mechanism not supported");
 
-			if (size == NULL)
-				util_fatal("Unknown key pair type %s, expecting RSA:<nbytes>", type);
 			key_length = (unsigned long)atol(size);
```
Jakuje, line 2866: Does the atol handle NULL parameters correctly? I did not find this confirmed in the manual pages.

frankmorgner:
No, this was not an accident. Note the previous check for `type != NULL`, which makes any value of kind `type + X` also non-NULL. All that could happen is that the derived value points to`'\0'`, but strcmp will not match that to a gost parameter set and atol in the RSA/AES case will return `0` so that the subsequent check of the key_length will fail.

As written in the commit message [ac77fc8](https://github.com/OpenSC/OpenSC/pull/2983/commits/ac77fc82088be604698f6c10241b33d8b25f1d18), this disallows the use of default key length. This was implemented for RSA, for example, but the user had to use `RSA:` (note the colon at the end). As User, I would have wanted a default value only without the colon. Additionally, if the user supplies `RSA:abcde`, then atol also returns `0`, which currently results in the default key length rather than being treated as error. That's why, I changed `--key-type` to not use any default values.

src/tools/pkcs11-tool.c
Jakuje: My bad. You are right.

src/tools/pkcs11-tool.c
```diff
@@ -2863,11 +2863,12 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
-			if (key_length != 0)
+			if (key_length != 0) {
+				if (*size != '\0')
+					util_fatal("Unknown key pair type %s, expecting RSA:<nbytes>", type);
 				modulusBits = key_length;
+			}
```
Jakuje, line 2871: so in that case, this logic of this block is somehow flawed as it fails the tests setup now with an error like:
```
Setup SoftHSM
The token has been initialized and is reassigned to slot 148272278
Using slot 0 with a present token (0x8d67496)
error: Unknown key pair type RSA:1024, expecting RSA:<nbytes>
Aborting.
Couldn't generate RSA:1024 key pair
====> ERROR: Failed to set up card
```

frankmorgner: obviously it should have been `if (key_length == 0) {`

I fixed that and additionally removed the default 1024 for the reasons described above

Jakuje: Thanks. This is resolved. Now just one more tests depends on the defaults:

https://github.com/OpenSC/OpenSC/blob/master/tests/test-pkcs11-tool-allowed-mechanisms.sh#L24

Please, fix also that one so we can merge this.

frankmorgner: Is there an option to view tests/test-suite.log from the CI pipeline?

Jakuje: Its uploaded as part of artifacts (can be downloaded from the workflow summary):

https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/linux.yml#L27

The other option that looks like it does not work now is dumping the logs into the github output (but with the set -e it will never be executed as the execution ends with the failed `make check`):

https://github.com/OpenSC/OpenSC/blob/c3545018d059b4debde33b9f34de719dd41e5531/.github/build.sh#L67
https://github.com/OpenSC/OpenSC/blob/master/.github/dump-logs.sh

If you would find it useful to see the logs in the output, I can try to get that working.

frankmorgner: fixed now.

The output of the logs in case of failure would be useful for future debugging, if you want to look at this again later.

src/tools/pkcs11-tool.c
```diff
@@ -2919,8 +2918,13 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 				if (!strcmp(ec_curve_infos[ii].oid, type + 3))
 					break;
 			}
-			if (!ec_curve_infos[ii].name)
-				util_fatal("Unknown EC key params '%s'", type + 3);
+			if (!ec_curve_infos[ii].name) {
+				fprintf(stderr, "EC key parameters may be specified by their canonic name or object identifier. Possible values are:\n");
+				for (ii=0; ec_curve_infos[ii].name; ii++)   {
```
Jakuje, line 2923: Can you fix the spacing here as suggested by the clang-format:
```
index 8478f41..9c68081 100644
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2920,7 +2920,7 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 			}
 			if (!ec_curve_infos[ii].name) {
 				fprintf(stderr, "EC key parameters may be specified by their canonic name or object identifier. Possible values are:\n");
-				for (ii=0; ec_curve_infos[ii].name; ii++)   {
+				for (ii = 0; ec_curve_infos[ii].name; ii++) {
 					fprintf(stderr, "%s (%s)\n", ec_curve_infos[ii].name, ec_curve_infos[ii].oid);
 				}
 				util_fatal("Unknown EC key parameter '%s'", type + 3);
```

Jakuje (approved):

src/tools/pkcs11-tool.c
Jakuje: For dumping logs, I filled #3005.

src/tools/pkcs11-tool.c
Jakuje: @frankmorgner can you squash the style issue so we can merge this?

frankmorgner: done and rebased on top of main branch

Jakuje (approved):

Jakuje:
Thank you!

PR #2985, opened 2024-01-15
pkcs11-tool: add support for importing Ed25518/448 keys

dlegaultbbry:
Related #2952

NOTE: I do not have this patch in my version of the tool so the EC_POINT bit value is off for 448: https://github.com/OpenSC/OpenSC/commit/5493770996ae31ee113af9be39ae35783e9b93cd

OIDs are printed properly regardless.

Tested this with pkcs11-tool compiled with both openssl 1.1.1 and 3.0.

```
# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --write-object /data/keystore/ed25519_privkey.pem --type privkey --id 1 --usage-sign --label ed25519key
Created private key:
Private Key Object; EC_EDWARDS
  label:      ed25519key
  ID:         01
  Usage:      sign
  Access:     sensitive, always sensitive, never extractable

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --write-object /data/keystore/ed25519_pubkey.pem --type pubkey --id 1 --usage-sign --label ed25519key
Created public key:
Public Key Object; EC_EDWARDS  EC_POINT 255 bits
  EC_POINT:   cac46187cb18b72fce802f34d70598f23e7a82dc6adfce79fae215e6ec686ae2
  EC_PARAMS:  06032b6570 (OID 1.3.101.112)
  label:      ed25519key
  ID:         01
  Usage:      verify
  Access:     none

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --write-object /data/keystore/ed448_privkey.pem --type privkey --id 2 --usage-sign --label ed448key
Created private key:
Private Key Object; EC_EDWARDS
  label:      ed448key
  ID:         02
  Usage:      sign
  Access:     sensitive, always sensitive, never extractable

# pkcs11-tool --module=/system/lib/dll/pkcs11-qkeystore.so --write-object /data/keystore/ed448_pubkey.pem --type pubkey --id 2 --usage-sign --label ed448key
Created public key:
Public Key Object; EC_EDWARDS  EC_POINT 255 bits
  EC_POINT:   fedfd388faf107519be6f06526f5fc72019178eb7b7b153aa95114592c24955c274964103b6829c702997ef271e5291f655eef06d1caf65c00
  EC_PARAMS:  06032b6571 (OID 1.3.101.113)
  label:      ed448key
  ID:         02
  Usage:      verify
  Access:     none
```

##### Checklist
- [x] PKCS#11 module is tested (custom soft HSM module)


src/tools/pkcs11-tool.c
```diff
@@ -3913,6 +3913,55 @@ parse_ec_pkey(EVP_PKEY *pkey, int private, struct gostkey_info *gost)
...
+	}
+
+	key = malloc(key_size);
+	if (key == NULL) {
+		return -1;
+	}
+
+	if (private) {
+		if (EVP_PKEY_get_raw_private_key(pkey, key, &key_size) != 1) {
+			return -1;
```
Jakuje, line 3951: free the `key` here otherwise you will leak the memory. The same in the following case

Jakuje (commented):
just one nit inline.

Does this require some modified softhsm or could this be intergrated in the tests in tests/ directory?

src/tools/pkcs11-tool.c
dlegaultbbry: I can do that. I'll be honest, I didn't quite know what the rules were given this is a command line tool which dies after it's done its work.

dlegaultbbry:
> just one nit inline.
> 
> Does this require some modified softhsm or could this be intergrated in the tests in tests/ directory?

I'll be honest, I haven't looked at adding any tests so far. I develop against a QNX custom key management solution at the moment with my own custom software HSM (using openssl 1.1.1 or 3.0).

I could look into adding ED key generation and EDDSA tests as it seems softhsm2 does support them. It would take me some time to figure out the test setup and modifications to do so.

Jakuje (approved):
lgtm.

Tests would be great though. I think something simple as modifying the following test (that now tries to generate and import RSA and EC keys) to do also ED25519 and ED448 should do:

https://github.com/OpenSC/OpenSC/blob/master/tests/test-pkcs11-tool-import.sh

dlegaultbbry:
Yes that's what I was looking at. Feel free to hold this up until I add the test or I can push another PR for the test only separately once I can get things working. I can also do the same for EDDSA at the same time if done in a separate PR.

frankmorgner (changes requested):
`gost_info_free()` uses `OPENSSL_free()` to release the structure's members. I'm not sure if we are relying on some OpenSSL internal initialization somewhere, but if so, we should use `OPENSSL_malloc()` to create the members by hand. If we don't use some internal OpenSSL allocation, then we can stick to `malloc()`, but should switch to `free()` when releasing the memory.

frankmorgner (approved):

Issue #2987, opened 2024-01-16
OpenSC 0.24.0 PKCS#11 + Firefox + TLS client-side auth via PIV = connection timeout


SCjona:
### System & OpenSC Information

Component: opensc-pkcs11.so
Smartcard(s): 2x Yubico YubiKey OTP+FIDO+CCID 00 00 (YubiKey 5). Also happens with just one card connected.
Operating System: Arch Linux
Package: https://archlinux.org/packages/extra/x86_64/opensc/
Version number: 0.24.0 ([package is compiled from Github releases](https://gitlab.archlinux.org/archlinux/packaging/packages/opensc/-/blob/de535bdfecf68f09f28df4ef9d4c0932f6d285c5/PKGBUILD))
Also happened to other coworkers on MacOS with brew and its OpenSC package.

### Problem Description & Steps to reproduce

This bug only occurs in Firefox. Chromium works fine.

> The following is copied from Mozilla Bugtracker: https://bugzilla.mozilla.org/show_bug.cgi?id=1872998

User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:121.0) Gecko/20100101 Firefox/121.0

Steps to reproduce:

I'm using a YubiKey to authenticate myself to a web server which requires client-side TLS certificates for authentication. This worked fine with OpenSC version 0.23.0 (on MacOS and Linux (Windows untested)), but yields a timeout error on OpenSC version 0.24.0.

You can work around this issue by going to Settings -> Security Devices (under Certificates) -> Select the Smartcard under the OpenSC module -> Log In.
This will prompt the pin and after completing the login everything works as expected.

Actual results:

Opening an website which is protected by client-side TLS certificate authentication yields an network timeout error after 5 to 10 seconds. No PIN/Password prompt for the smart card is shown.

Expected results:

A PIN/Password prompt for the smartcard should be shown. After entering the correct PIN the page should load.

### Logs

https://gist.github.com/SCjona/ea342d3bf429a7fcf5faea8917e1c23b

SCjona:
I found the culprit using `git bisect`: https://github.com/OpenSC/OpenSC/commit/1e625a3824c6a639e4c1590fe20899ade93242d0

I can get a working build using:
1. `git checkout 0.24.0`
2. `git revert 1e625a3824c6a639e4c1590fe20899ade93242d0`
3. `make`

dengert:
[...]

PR #2988, opened 2024-01-18
pkcs11-spy.c check return code

dengert:
The po->C_GetInterface is passed the callers ppInterface where *ppInterface may not be valid.

if the po->C_GetInterface may not update the *ppInterface and return an error. In this case  spy_interface_function_list should not be called, as it assumes the *ppInterface has been modified.

Found debugging FireFox version 121 where FireFox passes a ppInterface where *ppInterface is not a valid pointer, causing a segfault in spy_interface_function_list.

FireFox calls C_GetInterface twice with flags = CKF_INTERFACE_FORK_SAFE twice then on third time requests with flag = 0  where po->GetInterface can support and it updates the *ppInterface  with valid data.

Found with debugging #2987 

 On branch pkcs11-spy-segfault
 Changes to be committed:
	modified:   pkcs11-spy.c

##### Checklist


Jakuje (approved):
looks reasonable. I think we should run the spy as part of the ci somehow too to prevent issues like this ...

PR #2989, opened 2024-01-18
Fix memory leaks discovered by OSS-Fuzz

xhanulik:
Recently reported OSS-Fuzz issues

##### Checklist


src/pkcs15init/pkcs15-cflex.c
```diff
@@ -571,6 +571,8 @@ cflex_create_pin_file(sc_profile_t *profile, sc_pkcs15_card_t *p15card,
 	 * necessary */
 	ndummies = cflex_create_dummy_chvs(profile, p15card,
 				file, SC_AC_OP_UPDATE, dummies);
+	if (ndummies == 0)
+		sc_file_free(file);
 	LOG_TEST_RET(ctx, ndummies, "Unable to create dummy CHV file");
```
Jakuje, line 576: The `LOG_TEST_RET` returns only if `_ret < 0` so I this branch should be written the same otherwise the fuzzer will find similar leak again (when the `ndummies` will be negative as an error code).

Jakuje (changes requested):

Jakuje (approved):

frankmorgner (approved):

PR #2992, opened 2024-01-19
doc: Correctly forward IO to xsltproc to correctly replace paths in manual pages

Jakuje:
I accidentally noticed that some of the generated manual pages had non-translated replacement strings, which was caused by wrong invocation of the xsltproc.

##### Checklist


frankmorgner (approved):

Jakuje:
Thanks for review! Merging.

Issue #2953, opened 2023-12-05
C_SignFinal failed: rv = CKR_GENERAL_ERROR with pkcs11-tool --test --login --pin XXXX 


knecht:
### Problem Description

`pkcs11-tool --test --login --pin XXXX`
results in
```
Using slot 0 with a present token (0x0)
C_SeedRandom() and C_GenerateRandom():
  seeding (C_SeedRandom) not supported
  seems to be OK
Digests:
  all 4 digest functions seem to work
  MD5: OK
  RIPEMD160: OK
  SHA-1: OK
  SHA256: OK
Ciphers: not implemented
Signatures (currently only for RSA)
  testing key 0 (Private key 1) 
error: PKCS11 function C_SignFinal failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.

```

```
OpenSC-0.23.0-562-g258bfc8a, rev: 258bfc8a, commit-time: 2023-12-01 01:19:06 +0100
Using reader with a card: Gemalto PC Twin Reader (A7D1F9E2) 00 00
Gemalto IDPrime MD 830
```

```

Outgoing APDU (11 bytes):
00 22 41 B6 06 80 01 02 84 01 41 ."A.......A
Incoming APDU (2 bytes):
90 00 ..
P:17858; T:0x140064411954048 22:30:09.598 [opensc-pkcs11] card-idprime.c:1004:idprime_set_security_env: returning with: 0 (Success)

P:17858; T:0x140064411954048 22:30:09.598 [opensc-pkcs11] card-idprime.c:1020:idprime_compute_signature: called
P:17858; T:0x140064411954048 22:30:09.566 [opensc-pkcs11] apdu.c:367:sc_single_transmit: CLA:0, INS:2A, P1:90, P2:A0, data(37) 0x7ffe06e803f0
Outgoing APDU (42 bytes):
00 2A 90 A0 25 90 23 30 21 30 09 06 05 2B 0E 03 .*..%.#0!0...+..
02 1A 05 00 04 14 29 B0 E7 87 82 71 64 5F FF B7 ......)....qd_..
EE C7 DB 4A 74 73 A1 C0 0B C1                   ...Jts....

Incoming APDU (2 bytes):
6A 88 j.

P:17858; T:0x140064411954048 22:30:09.584 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
[...]

PR #3003, opened 2024-01-30
idprime: Print some more logs from index file parsing

Jakuje:
This should simplify log comparison and interpretation of index files such as in #2953.

##### Checklist

- [ ] macOS tokend is tested


xhanulik (approved):

frankmorgner (approved):

Jakuje:
Thank you for the reviews!

PR #3005, opened 2024-01-31
ci: Dump logs on error

Jakuje:
The logs should be grouped in expandable sections as described here for better readability:

https://github.com/actions/toolkit/blob/main/docs/commands.md#group-and-ungroup-log-lines

This also attempts to dump the logs from make distcheck as they will be in different path.

Example of failed run:

https://github.com/Jakuje/OpenSC/actions/runs/7723818954/job/21054680420

This allows fast review of the failed logs. Unfortunately, loading of long sections takes some time in the webui ...

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -11,17 +11,20 @@ fi
 if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
 	PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
 	if [ "$GITHUB_BASE_REF" == "master" ]; then
-		./bootstrap.ci -s "-pr$PR_NUMBER"
+		SUFFIX="-pr$PR_NUMBER"
 	else
-		./bootstrap.ci -s "$GITHUB_BASE_REF-pr$PR_NUMBER"
+		SUFFIX="$GITHUB_BASE_REF-pr$PR_NUMBER"
 	fi
 else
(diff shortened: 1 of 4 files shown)


frankmorgner:
Expandability looks nice, although it not really helps if the UI is very slow

The Artifact naming looks not OK (see https://github.com/Jakuje/OpenSC/actions/runs/7723818954):
>debian-%{{ matrix.compiler }}-strict-test-logs
	1.45 MB 	
fedora-%{{ matrix.compiler }}-strict-test-logs
	115 KB 	
ubuntu-%{{ matrix.compiler }}-strict-test-logs
	1.45 MB

Jakuje:
> Expandability looks nice, although it not really helps if the UI is very slow

I hope it will help at lest for fast glimpse of the logs. Or should I change it to not expandable? I am worried that it would be too much of logs on the bottom that it would be quite hard to see the actual test summary on the fast review just to see what log file to look for.

> The Artifact naming looks not OK (see https://github.com/Jakuje/OpenSC/actions/runs/7723818954):
> 
> > debian-%{{ matrix.compiler }}-strict-test-logs
> > 1.45 MB
> > fedora-%{{ matrix.compiler }}-strict-test-logs
> > 115 KB
> > ubuntu-%{{ matrix.compiler }}-strict-test-logs
> > 1.45 MB

Indeed. I was likely writing too much jinja2 recently ... Should be fixed in the top commit now.

frankmorgner (approved):
having it expandable is fine, thank you

Jakuje:
Thanks for comments! Merging.

Issue #3006, opened 2024-01-31
reader configuration partially skipped If SCardControl missing


frankmorgner:
When 
`SCardControl(card_handle, CM_IOCTL_GET_FEATURE_REQUEST,...` fails at:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L1153

the 178 lines of code: 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/reader-pcsc.c#L1158-L1337
are not executed, which includes:

- code to test for specific readers
- pin pad info (probably why pin pad readers don't work in Windows) 
- `part10_detect_pace_capabilities(reader, card_handle);`
- `max_send_size = max_recv_size = part10_detect_max_data(reader, card_handle);`
- other code to `set max_send_size` and `max_recv_size` which may end in "Assuming that the reader supports sending short length APDUs only"

This results in a card driver like `card-sc-hsm.c` expects the reader in the token to support extended APDUs that the reader does not.  

This lack of the features from PCSClite in the Windows implementation could be addressed if OpenSC took some of the changes from PCSClite to fill in info for readers. Or at least look in `opensc.conf`

This problem may be the same problem as in #3004  and other card drivers may have similar problems. 

(I actually found that building OpenSC from github using  Cygwin was easy and could debug using `gdb`)

_Originally posted by @dengert in https://github.com/OpenSC/OpenSC/issues/2944#issuecomment-1918938363_

metsma:
For the record apple has switched to their CCID driver and they did not support SCardControl in 14.0/14.1.
https://blog.apdu.fr/posts/2023/11/apple-own-ccid-driver-in-sonoma/
In the 14.2 they switched back to @LudovicRousseau CCID driver.
In 14.3 they implemented SCardControl in some extent and switched back to their own CCID implementation.
Least PinPAD is working but I am not sure about APDU size parameters. (@LudovicRousseau maybe you can verify this also their driver misidentifies some card readers as pinpads #1438)

frankmorgner:
does dlopen on macOS' pcsc implementation yield NULL when requesting SCardControl or do they implement it by always returning an error? In the former case, this may be the reason for #2788 ...

metsma:
> does dlopen on macOS' pcsc implementation yield NULL when requesting SCardControl or do they implement it by always returning an error? In the former case, this may be the reason for #2788 ...

[...]

PR #3007, opened 2024-02-01
Fix reader initialization without SCardControl

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/3006

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -238,7 +238,7 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 	struct pcsc_private_data *priv = reader->drv_data;
 	SCARD_IO_REQUEST sSendPci, sRecvPci;
 	DWORD dwSendLength, dwRecvLength;
-	LONG rv;
+	LONG rv = SCARD_E_INVALID_VALUE;
 	SCARDHANDLE card;
 
 	LOG_FUNC_CALLED(reader->ctx);
@@ -264,8 +264,10 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 				  recvbuf, &dwRecvLength);
 		}
 		else {
-			rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
-				  recvbuf, dwRecvLength, &dwRecvLength);
+			if (priv->gpriv->SCardControl != NULL) {
+				rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
+						recvbuf, dwRecvLength, &dwRecvLength);
+			}
 		}
 	}
 
@@ -1016,7 +1018,7 @@ static unsigned long part10_detect_pace_capabilities(sc_reader_t *reader, SCARDH
 	if (!priv)
 		goto err;
 
-	if (priv->pace_ioctl && priv->gpriv) {
+	if (priv->pace_ioctl && priv->gpriv && priv->gpriv->SCardControl) {
 		if (SCARD_S_SUCCESS != priv->gpriv->SCardControl(card_handle,
 					priv->pace_ioctl, pace_capabilities_buf,
 					sizeof pace_capabilities_buf, rbuf, sizeof(rbuf),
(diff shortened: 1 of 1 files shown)


src/libopensc/reader-pcsc.c
```diff
@@ -264,8 +264,10 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 				  recvbuf, &dwRecvLength);
 		}
 		else {
-			rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
-				  recvbuf, dwRecvLength, &dwRecvLength);
+			if (priv->gpriv->SCardControl != NULL) {
+				rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
+						recvbuf, dwRecvLength, &dwRecvLength);
+			}
```
Jakuje, line 270: nit: I would put it in `else if` branch here to avoid too much indentation:
```
		} else if (priv->gpriv->SCardControl != NULL) {
			rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
					recvbuf, dwRecvLength, &dwRecvLength);
		}
```

src/libopensc/reader-pcsc.c
```diff
@@ -1147,13 +1149,12 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
...
-		return;
+	if (gpriv->SCardControl != NULL) {
+		rv = gpriv->SCardControl(card_handle, CM_IOCTL_GET_FEATURE_REQUEST, NULL, 0, buf, sizeof(buf), &rcount);
+		if (rv != SCARD_S_SUCCESS) {
+			PCSC_TRACE(reader, "SCardControl failed", rv);
+			return;
+		}
 	}
 
 	if ((rcount % sizeof(PCSC_TLV_STRUCTURE)) != 0
```
Jakuje, line 1160: This makes an unitializer read of `rcount` if there is no SCardControl. I think initialization should initialize it to 0, which should do the job.

src/libopensc/reader-pcsc.c
```diff
@@ -1221,7 +1222,7 @@ static void detect_reader_features(sc_reader_t *reader, SCARDHANDLE card_handle)
 	}
 
 	/* Detect display */
-	if (priv->pin_properties_ioctl) {
+	if (priv->pin_properties_ioctl && gpriv->SCardControl) {
```
Jakuje, line 1225: nit. In some cases in this commit you test `gpriv->SCardControl` and in some `gpriv->SCardControl != NULL`. Its usually consistent with the surrounding code, but not consistent together. Would be great to stick to either of these. I dont have strong opinion on which, but likely the first one as I see it as more commonly used and shorter.

frankmorgner: I was looking more on the surrounding code. I'll see what I can do

src/libopensc/reader-pcsc.c
frankmorgner: AFAIK, that is used as output parameter only. I'll add the initialization anyway, it doesn't hurt.

src/libopensc/reader-pcsc.c
```diff
@@ -259,13 +259,13 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
...
 				   &sRecvPci, recvbuf, &dwRecvLength);
 	} else {
-		if (priv->gpriv->SCardControlOLD != NULL) {
+		if (!priv->gpriv->SCardControlOLD) {
 			rv = priv->gpriv->SCardControlOLD(card, sendbuf, dwSendLength,
 				  recvbuf, &dwRecvLength);
 		}
-		else {
+		else if (!priv->gpriv->SCardControl) {
 			rv = priv->gpriv->SCardControl(card, (DWORD) control, sendbuf, dwSendLength,
```
Jakuje, line 267: last nit from the clang is to merge these two lines to not have closing brace alone on the line and connect the cast with the variable name:
```suggestion
		} else if (!priv->gpriv->SCardControl) {
			rv = priv->gpriv->SCardControl(card, (DWORD)control, sendbuf, dwSendLength,
```
then we I think we are good

frankmorgner:
thank you

Jakuje:
I fixed the stray closing brace in 22e4ebcfc5d22fd07a4a98af5e78f2e16da25c11 (sorry, the github does not allow me to do suggestions over modifed and unmodified lines).

PR #3008, opened 2024-02-01
Fix error handling on strdup

frankmorgner:
I have reviewed all issues that were brought up by CodeQL. Unfortunately, almost everyone of them was a false positive. It seems that CodeQL doesn't understand it when we're correctly handling some error in the line of the following:
```
p = calloc(5);
if (p == NULL)
   LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
```
Anyway, one CodeQL issue at least led me to harden some other error handling around the referenced code.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -923,8 +923,7 @@ sc_pkcs15init_add_app(struct sc_card *card, struct sc_profile *profile,
 	}
 
 	if (args->label) {
-		if (p15card->tokeninfo->label)
-			free(p15card->tokeninfo->label);
+		free(p15card->tokeninfo->label);
 		p15card->tokeninfo->label = strdup(args->label);
 	}
 	if (p15card->tokeninfo->label)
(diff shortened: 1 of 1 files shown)


src/pkcs15init/pkcs15-lib.c
```diff
@@ -927,7 +927,10 @@ sc_pkcs15init_add_app(struct sc_card *card, struct sc_profile *profile,
 			free(p15card->tokeninfo->label);
```
Jakuje, line 927: nit: once we touch this, I would replace the above two lines with just the free (removing the NULL check `if (p15card->tokeninfo->label)`):
```suggestion
		free(p15card->tokeninfo->label);
```

Jakuje (approved):
otherwise lgtm

src/pkcs15init/pkcs15-lib.c
frankmorgner: well, did that first and decided that it is too far off. anyway, I'll add that in :wink:

Jakuje (approved):

PR #3010, opened 2024-02-01
CI: add worker for M1 macOS runner

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -18,7 +18,10 @@ permissions:
 
 jobs:
   build:
-    runs-on: macos-12
+    strategy:
+      matrix:
+        os: [macos-12, macos-14]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
       - run: .github/setup-macos.sh
@@ -33,7 +36,7 @@ jobs:
       - name: Cache build artifacts
         uses: actions/upload-artifact@v3
         with:
-          name: opensc-build-macos
+          name: opensc-build-${{ matrix.os }}
           path:
             OpenSC*.dmg
       - run: .github/cleanup-macos.sh
@@ -49,7 +52,7 @@ jobs:
       - name: Pull build artifacts
         uses: actions/download-artifact@v3
         with:
-          name: opensc-build-macos
+          name: opensc-build-macos-12
       - run: git config --global user.email "builds@github.com"
       - run: git config --global user.name "Github Actions";
       - run: .github/push_artifacts.sh "Github Actions ${GITHUB_REF}"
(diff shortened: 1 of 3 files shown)


.github/workflows/macos.yml
```diff
@@ -41,6 +41,24 @@ jobs:
...
+          PASS_SECRETS_TAR_ENC: ${{ secrets.PASS_SECRETS_TAR_ENC }}
+      - run: .github/build.sh
+        env:
+          CODE_SIGN_IDENTITY: ${{ secrets.CODE_SIGN_IDENTITY }}
+          DEVELOPMENT_TEAM: ${{ secrets.DEVELOPMENT_TEAM }}
+          INSTALLER_SIGN_IDENTITY: ${{ secrets.INSTALLER_SIGN_IDENTITY }}
+      - run: .github/cleanup-macos.sh
+        env:
+          KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
+          PASS_SECRETS_TAR_ENC: ${{ secrets.PASS_SECRETS_TAR_ENC }}
```
Jakuje, line 60: would it be possible to avoid copy&past configuration given that its the same for the previous build target using the matrix similarly as used in https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/linux-strict.yml#L55 ?

I think something like this should do (untested):
```
  build-m1:
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
        os: [macos-12, macos-14]
    steps:
      - uses: actions/checkout@v3
      - run: .github/setup-macos.sh
        env:
          KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
          PASS_SECRETS_TAR_ENC: ${{ secrets.PASS_SECRETS_TAR_ENC }}
      - run: .github/build.sh
        env:
          CODE_SIGN_IDENTITY: ${{ secrets.CODE_SIGN_IDENTITY }}
          DEVELOPMENT_TEAM: ${{ secrets.DEVELOPMENT_TEAM }}
          INSTALLER_SIGN_IDENTITY: ${{ secrets.INSTALLER_SIGN_IDENTITY }}
      - run: .github/cleanup-macos.sh
        env:
          KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
          PASS_SECRETS_TAR_ENC: ${{ secrets.PASS_SECRETS_TAR_ENC }}
```

frankmorgner: done

Jakuje (approved):
thanks! I think this way its slightly more readable, at least in the long term when we will need to change or add something.

PR #3013, opened 2024-02-01
p11test: Avoid more memory leaks

Jakuje:
Running the tests against epass2003 token revealed few more memory leaks in the p11test itself that for some reason do not demonstrate with other cards/drivers/pkcs11 modules (likely the verification does not work on card for some reason so the tests fall back to openssl verification).

##### Checklist
- [ ] macOS tokend is tested


Jakuje:
Thank you!

PR #3015, opened 2024-02-02
idrpime: Get applet version based on (#2953)

Jakuje:
Print the applet version reported by the official driver.

##### Checklist

- [ ] macOS tokend is tested


xhanulik (approved):

Jakuje:
Thank you!

PR #3016, opened 2024-02-05
Coverity issues

xhanulik:
Fixes for recently reported Coverity issues

##### Checklist


src/pkcs11/framework-pkcs15.c
```diff
@@ -4627,11 +4627,11 @@ pkcs15_prkey_decrypt(struct sc_pkcs11_session *session, void *obj,
 
 	/* only padding error must be handled in constant-time way,
 	 * other error can be returned straight away */
-	if ((~constant_time_eq_s(rv, SC_ERROR_WRONG_PADDING) & constant_time_lt_s(sizeof(decrypted), rv)))
+	if ((~constant_time_eq_s((size_t)rv, SC_ERROR_WRONG_PADDING) & constant_time_lt_s(sizeof(decrypted), (size_t)rv)))
```
Jakuje, line 4630: The `SC_ERROR_WRONG_PADDING` is negative so I am wondering if we should not do either:
 * have a constant time functions to compare signed ints, such as `constant_time_eq_i`?
 * have temporary variable size_rv to avoid casting it on too many places?

After the first bit operation, the types does not matter anymore so we might not need to go all the way to the bottom of the call stack and just cast it to the `size_t` or something, but I would find it nicer to be done in one place than in most of the places where this function is used.

xhanulik: Good point, as both `rv` and `SC_ERROR_WRONG_PADDING` are signed, I will make `constant_time_eq_i` for integers.

Jakuje (approved):
thanks.

Issue #2963, opened 2023-12-18
Invalid ASN.1 object error when using myeid profile in pkcs15-init -C command


hhonkanen:
### Problem Description

When using the MyEID profile file, pkcs-init -C command fails with error message "Failed to create PKCS #15 meta structure: Invalid ASN.1 object"

Tested with MyEID 4.5.5 and OpenSC 0.24.0 on Windows.

### Steps to reproduce

pkcs15-init -C --profile myeid --pin 1111 --puk 1111

### Logs

P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init] card.c:885:sc_select_file: returning with: 0 (Success)
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init] encoding 'dirRecord'
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init] type=129, tag=0x60000001, parm=000000EBBDAFED50, len=0
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init]  encoding 'aid'
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init]  type=4, tag=0x4000000f, parm=000000EBBDAFEC68, len=0
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init] cannot encode empty non-optional ASN.1 object
P:92432; T:81956 2023-12-18 12:08:58.703 [pkcs15-init] encoding of ASN.1 object 'dirRecord' failed: Invalid ASN.1 object
P:92432; T:81956 2023-12-18 12:08:58.704 [pkcs15-init] dir.c:303:encode_dir_record: Encode DIR record error: -1401 (Invalid ASN.1 object)
P:92432; T:81956 2023-12-18 12:08:58.704 [pkcs15-init] pkcs15-lib.c:3154:sc_pkcs15init_update_dir: returning with: -1401 (Invalid ASN.1 object)

popovec:
Can you please try it with the file cache disabled?

hhonkanen:
Tried with the following opensc.conf, but still got the same error.

app default {
	debug = 9;
	debug_file = c:\debug\opensc-debug64.txt;
	framework pkcs15 {
		use_file_caching = false;
	}
}

popovec:
The same thing happens on linux, it doesn't depend on the file cache. 

Fail:
`````
pkcs15-init -C --profile myeid --pin 1111 --puk 1111  
`````

OK:
[...]

PR #3020, opened 2024-02-07
MyEID: For the private key, use the ACLs that are defined in the prof…

popovec:
…ile.

Another issue regarding setting the ACL to a private key was raised in issue #2963.  This patch removes a part of the code that makes it impossible to set the ACL according to the selected profile.

##### Checklist

### DIFF
--- a/src/pkcs15init/pkcs15-myeid.c
+++ b/src/pkcs15init/pkcs15-myeid.c
@@ -537,9 +537,6 @@ myeid_create_key(struct sc_profile *profile, struct sc_pkcs15_card *p15card,
 	struct sc_path *path;
 	int *key_reference;
 	struct sc_file *file = NULL;
-	struct sc_pkcs15_object *pin_object = NULL;
-	struct sc_pkcs15_auth_info *pkcs15_auth_info = NULL;
-	unsigned char sec_attrs[] = {0xFF, 0xFF, 0xFF};
 	int r, ef_structure = 0, pin_reference = -1;
 	size_t keybits = 0;
 	unsigned char prop_info[] = {0x00, 0x00};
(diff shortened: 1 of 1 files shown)


Jakuje:
This exact code wast introduced in 8cf68bce1aef671937eb3c6c99f58724515bd10c describing it was fixing an issue that the key files were always created under PIN 1 regardless what was requested by the user. Is the original issue resolved in some other way or do we need a follow-up issue or fixes for the original issue?

popovec:
8cf68bc introduced the use of '--id' as information under which ACL the key should be created. This breaks the normal way where ACLs are specified in the profile file.

Let's wait for the approval of this PR from @hhonkanen.

hhonkanen:
The purpose of 8cf68bc is to be able to dynamically select, that for example PIN 1 is used with the authentication key, e.g. for login, and PIN 2 is used for e.g. signing documents, and associated with the signing key. This is still necessary, so seems like we must find a way to combine this functionality with the ability to customize the profile. Or does OpenSC now handle what 8cf68bc does in some generic way for all cards? 

8cf68bc sets all security attributes to the value set in -a. Would it make sense to set only "USE" security attribute to the PIN set in object.auth_id, and take the rest (PUT DATA, GENERATE, DELETE) from the profile? This would at least resolve the use case we have now, where it is important that only SO can delete keys, while keys can be used with PIN 1. In pkcs15-init -a option is mandatory for key generation, unless -insecure switch is used to allow USE without a pin at all, therefore I think the logic would be ok: USE attribute is always provided by caller or user, and other security attributes would come from the profile.

Could something like this work?

		```
if (file->sec_attrs_len >= 3)
    memcpy(sec_attrs, file->sec_attrs, 3);  /* take the security attributes set by the profile */

pin_reference = pkcs15_auth_info->attrs.pin.reference;

if (pin_reference >= 1 && pin_reference < MYEID_MAX_PINS) {
	sec_attrs[0] |= (pin_reference << 4);    /* set only USE ac to object.auth_id*/
	sc_file_set_sec_attr(file, sec_attrs, sizeof(sec_attrs));
```

I have some challenges building OpenSC, so I cannot test this right away. I don't have a Linux environment anymore, and I am trying to build it on Windows according to https://github.com/OpenSC/OpenSC/wiki/Compiling-on-Windows
but I still get various errors. 

I can create a PR with the code above, if you agree that the logic is correct.

popovec:
The user can set his own profile when generating/importing the key:

`````
cp /usr/share/opensc/pkcs15.profile .
cp /usr/share/opensc/myeid.profile  myeid_local.profile 
echo -e "app default {\n profile_dir = .\n}\n "> opensc_test.conf
`````
Edit `myeid_local.profile` as needed... and import the key with ACL according to the local profile:
`````
OPENSC_CONF=./opensc_test.conf pkcs15-init -c myeid_local --store-private-key "keys/rsa2048-key.pem" --key-usage sign,decrypt  --auth-id=1 --pin 1111 --so-pin 0000
`````
I admit, it's a bit of an overcomplicated solution.

Using the `--auth-id` switch could partially simplify the ACL setting for the generated/imported key, but this probably introduces inconsistency with respect to other cards supported in OpenSC.  I would welcome more contributors to comment on this topic.

hhonkanen:
pkcs15-init's help shows: 
-a, --auth-id <arg>           Specify ID of PIN to use/create
I think when creating a key, "use" can be understood to associate the pin with the new key. If we keep the feature, it could be made more clear in the help, though.

The feature to associate the auth-id with the key may be used by existing MyEID users, and removing it could cause problems to them. I know it is a very typical scenario among MyEID users to have a separate key+certificate combination for authentication and signing, and to associate them with different pins.  Creating two profiles and switching between them would work too, but I agree users might find it a bit complicated, and its not obvious that you need to do this.

My code from 2017 seems faulty though, management permissions should be separate from the USE permission.

Jakuje:
I agree that creating a separate profiles for setting different auth objects to created keys is probably overkill (unless we would provide some alternative myeid profile that would create this layout by default?).

I used pkcs15init only with the MyEID card, with the ePass (I think this one does not support multiple auth objects?) and sc-hsm (all mostly for testing) so I do not have huge experience with the smart card enrollment.

I see there are hints to use `--auth-id` in the wiki for Rutoken when creating keys as well as for OpenPGP (and in readme and in manual pages) so I would consider this as a supported use case that needs a better documentation (and some fix for MyEID, likely something like proposed in https://github.com/OpenSC/OpenSC/pull/3020#issuecomment-1938504732 ) :

 * https://github.com/OpenSC/OpenSC/wiki/Aktiv-Co.-Rutoken-ECP
 * https://github.com/OpenSC/OpenSC/wiki/OpenPGP-card
 * https://github.com/OpenSC/OpenSC/blob/a007f3a04ab00eddcc63e34f8592cf9a0ba3ee67/src/pkcs15init/README#L13
 * https://github.com/OpenSC/OpenSC/blob/a007f3a04ab00eddcc63e34f8592cf9a0ba3ee67/doc/tools/pkcs15-init.1.xml#L136

hhonkanen:
I got my build environment mostly working, but I have still issues with pkcs15-init and OpenSSL. 

I created a branch for the proposed changes here: https://github.com/hhonkanen/OpenSC/tree/key_permissions_from_profile

I will be on vacation next week, and may not be able to test it today, as I have to fix the build issues first to get pkcs15-init working. Feel free to test it meanwhile, if anybody has a chance. 

This might not be the optimal solution. I think it would be better to handle it in higher level in a common way for all cards, but it would require some more work. myeid_create_key calls myeid_new_file, which calls sc_profile_get_file to get the sc_file object with ACLs set as defined in the profile. I couldn't find any card-independent function which would replace the USE permission with the auth-id.

hhonkanen:
I finally had time to get my build environment working, and tested the above-mentioned patch, but it doesn't set the ACLs correctly. I have to debug it to understand how it works.

frankmorgner:
The feedback from @hhonkanen was that the ACLs were not correct. Is there any update on this in the meantime?

frankmorgner:

PR #3024, opened 2024-02-12
OpenPGP: fixed state tracking after erasing card

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/3018

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -1654,6 +1654,8 @@ sc_pkcs15init_generate_key(struct sc_pkcs15_card *p15card, struct sc_profile *pr
 		r = profile->ops->emu_store_data(p15card, profile, object, NULL, NULL);
 		if (r == SC_ERROR_NOT_IMPLEMENTED)
 			r = SC_SUCCESS;
+		if (r < 0)
+			sc_pkcs15_remove_object(p15card, object);
 		LOG_TEST_GOTO_ERR(ctx, r, "Card specific 'store data' failed");
 	}
 


src/libopensc/card-openpgp.c
```diff
@@ -357,6 +357,25 @@ pgp_match_card(sc_card_t *card)
...
 
+/* populate MF - add matching blobs listed in the pgp_objects table */
+int populate_blobs_to_mf(sc_card_t *card, struct pgp_priv_data *priv)
+{
+	pgp_do_info_t	*info;
+	for (info = priv->pgp_objects; (info != NULL) && (info->id > 0); info++) {
+		if (((info->access & READ_MASK) != READ_NEVER) && (info->get_fn != NULL)) {
+			pgp_blob_t *child = NULL;
+
+			child = pgp_new_blob(card, priv->mf, info->id, sc_file_new());
```
Jakuje, line 368: If the `pgp_new_blob()` fails, the memory allocated by `sc_file_new()` is leaked (yeah, it was in the original code too but as I noticed it, it would be good to fix it).

src/libopensc/card-openpgp.c
```diff
@@ -3551,6 +3561,15 @@ pgp_erase_card(sc_card_t *card)
 					"Card does not offer life cycle management");
 	}
 
+	if (r == SC_SUCCESS && priv->mf) {
+		new_mf = pgp_new_blob(card, NULL, priv->mf->id, priv->mf->file);
+		if (new_mf) {
+			pgp_free_blobs(priv->mf);
+			priv->mf = new_mf;
+			populate_blobs_to_mf(card, priv);
+		}
```
Jakuje, line 3570: I would propose to report some error when the allocation fails instead of silently skipping the rest of the code.

Something like this could do:
```suggestion
		new_mf = pgp_new_blob(card, NULL, priv->mf->id, priv->mf->file);
		if (new_mf == NULL) {
			LOG_TEST_RET(card->ctx, SC_ERROR_INTERNAL, "Failed to allocate the new MF blob");
		}
		pgp_free_blobs(priv->mf);
		priv->mf = new_mf;
		populate_blobs_to_mf(card, priv);
```

frankmorgner:
thanks for the review. I'll look at the suggestions, when @arrowd confirms that it actually fixes the problem

arrowd:
I will not be able to test this during this week, sorry, but I will certainly check this out later.

If I'm reading it right, this change would allow me to just do `sc_disconnect_card + sc_connect_card` to refresh the state?

frankmorgner:
> If I'm reading it right, this change would allow me to just do sc_disconnect_card + sc_connect_card to refresh the state?

No. If this works as expected, then you will get refreshed objects right after erasing the card without the need of reconnecting.

arrowd:
The problem is not only with erasing.

The first post of https://github.com/OpenSC/OpenSC/issues/3018 shows a code that generates a keypair twice in a row, but it seemingly doesn't change until the reinsertion (physical or emulated with my hacks).

frankmorgner:
please check if it works in case of erasing the card. If so, one case is fixed and you know where to poke.

arrowd:
Sorry for taking this too long. The proposed change indeed improves things for both `opensc-explorer` and my daemon program. Erasing the card causes OPENPGP_ACCOUNT field to be reset immediately without the need to reinsert the token or to programmatically reset the reader via hacks.

The key information isn't updated, though.

dengert:
> What does 00 44 00 00 do in this case?

7.2.17 ACTIVATE FILE
This optional command (announced in Life Cycle Status indicator in Historical bytes) can
be used to reset all values (DOs, Keys, PWs etc.) to their default values (after production/
initialisation). The command has effect only, if the life cycle status is in initialisation state
(indicator byte set to 03). If the command is used in operational state, it will end with status
bytes 9000, but nothing in the application will be changed (dummy command). The command
can be used directly after TERMINATE DF without resetting the card.
Command:
CLA 00 / 0C
INS 44
P1 00
P2 00
Lc Empty
Data
field
Empty
Le Empty
Response:
Data field Empty
SW1-SW2 9000 or specific status bytes
Example:
CLA INS P1 P2
00 44 00 00
SW1SW2
9000

frankmorgner:
I fixed the memory leaks and found some other minor fix. I'd like to leave this fix as is, because I doubt that we will get more insights without debugging.

frankmorgner:
It is not essential, so we can postpone this to the next release

PR #3025, opened 2024-02-12
fixed strict build

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


Jakuje (approved):
thanks! Looks reasonable.

PR #3026, opened 2024-02-14
dtrust: add dtrust-tool

hamarituc:
`dtrust-tool` can be used to query the PIN status and to remove the initial transport portection.

As my card for #2784 was already unlocked, I was not able to provide such a tool in the referenced pull request. I received more test cards yesterday just right before #2784 was merged, so it need a further pull request.

##### Checklist

### DIFF
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -30,7 +30,8 @@ noinst_HEADERS = util.h fread_to_eof.h \
 	openpgp-tool-helpers.h
 bin_PROGRAMS = opensc-tool opensc-explorer opensc-asn1 \
 	pkcs15-tool pkcs15-crypt pkcs11-tool pkcs11-register \
-	cardos-tool eidenv openpgp-tool iasecc-tool egk-tool goid-tool
+	cardos-tool eidenv openpgp-tool iasecc-tool egk-tool goid-tool \
+	dtrust-tool
 if ENABLE_OPENSSL
 bin_PROGRAMS += cryptoflex-tool pkcs15-init netkey-tool piv-tool \
 	westcos-tool sc-hsm-tool dnie-tool gids-tool
@@ -126,6 +127,8 @@ goid_tool_CFLAGS = $(OPENPACE_CFLAGS)
 if HAVE_UNKNOWN_WARNING_OPTION
 goid_tool_CFLAGS += -Wno-unknown-warning-option
 endif
+
+dtrust_tool_SOURCES = dtrust-tool.c util.c
 
 opensc_asn1_SOURCES = opensc-asn1.c fread_to_eof.c opensc-asn1-cmdline.c
 if HAVE_UNKNOWN_WARNING_OPTION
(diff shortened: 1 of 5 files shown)


src/tools/Makefile.am
```diff
@@ -30,7 +30,8 @@ noinst_HEADERS = util.h fread_to_eof.h \
 	openpgp-tool-helpers.h
 bin_PROGRAMS = opensc-tool opensc-explorer opensc-asn1 \
 	pkcs15-tool pkcs15-crypt pkcs11-tool pkcs11-register \
-	cardos-tool eidenv openpgp-tool iasecc-tool egk-tool goid-tool
+	cardos-tool eidenv openpgp-tool iasecc-tool egk-tool goid-tool \
+	dtrust-tool
```
Jakuje, line 34: Please, add the tool also to the windows makefile:

https://github.com/OpenSC/OpenSC/blob/master/src/tools/Makefile.mak

src/tools/dtrust-tool.c
```diff
@@ -0,0 +1,316 @@
...
+#include "config.h"
+
+#include <stdio.h>
+#include <string.h>
+
+// clang-format off
+#include "libopensc/opensc.h"
+#include "libopensc/cards.h"
+#include "libopensc/errors.h"
+// clang-format on
```
Jakuje, line 30: if some of the includes need to come before others, its easier to separate them by a newline than turning off the clang format altogether.
If it is the `opensc.h`, something like this should do:
```suggestion
#include "libopensc/opensc.h"

#include "libopensc/cards.h"
#include "libopensc/errors.h"
```

src/tools/dtrust-tool.c
```diff
@@ -0,0 +1,316 @@
...
+static const char *option_help[] = {
+	"Uses reader number <arg> [0]",
+	"Wait for card insertion",
+	"Show PIN status",
+	"Check transport protection",
+	"Unlock transport protection",
+	"This message",
+	"Verbose operation, may be used several times",
+};
+// clang-format on
```
Jakuje, line 57: I would prefer to format this part according to the clang. I think it should be just one more tab indented, isn't it?

Jakuje (commented):
Thank you. Looks reasonable. Added just minor comments.

src/tools/Makefile.am
hamarituc: fixed

src/tools/dtrust-tool.c
hamarituc: fixed ... and learned that empty lines matter in this case.

src/tools/dtrust-tool.c
hamarituc: I fixed it. Becaus all the other programs used one tab, I would like to introduce some oddness here. But if you like two tabs, I am fine with that.

Jakuje: Not that I would like it much, but its better to be consistent.

Please, keep the `options` with the clang-format off -- the indentation of the elements is a weirdness that even I can not handle, do not know why it happens nor how to easily fix it without breaking something else in formatting.

hamarituc: OK, I reverted the code-style changes.

Jakuje (approved):

xhanulik (approved):

Jakuje:
Thank you!

PR #3042, opened 2024-02-21
pkcs15-openpgp.c  Authentication key for decrypt requires MSE

dengert:
pkcs11-tool --test calls "test_decrypt" and test any RSA key that supports decryption.

OpenPGP can do this for the Authentication key, but requires the optional MANAGE SECURITY ENVIRONMENT (MSE) command.

Do not set decrypt or wrap usage bits unless MSE is supported for the card.

Found using YubiKey NFC and Nitro start that do not support MSE.

 On branch X25519-improvements-2
 Changes to be committed:
	modified:   libopensc/pkcs15-openpgp.c

##### Checklist

- [ ] macOS tokend is tested


dengert:
The [test-openpgp](https://github.com/OpenSC/OpenSC/actions/runs/7990369165/job/21819490312?pr=3042) fails with this because the `src/tests/p11test/openpgp_s0_ref.json`  file is expecting YES for the four responses which are are not set if MSE is not available.

It is not clear if the `softhsm2` used in this test needs some work to follow OpenPGP V3.4.1.

dengert:
I did try updating the `src/tests/p11test/openpgp_s0_ref.json`:

```
diff --git a/src/tests/p11test/openpgp_s0_ref.json b/src/tests/p11test/openpgp_s0_ref.json
index 6d7bf66da..d91a67cf9 100644
--- a/src/tests/p11test/openpgp_s0_ref.json
+++ b/src/tests/p11test/openpgp_s0_ref.json
@@ -284,10 +284,10 @@
                "",
                "YES",
                "YES",
-               "",
-               "",
-               "",
-               "",
+               "YES",
+               "YES",
+               "YES",
+               "YES",
                "",
                "",
                ""
```
Either github actions was not running with the updated file  or the above was not the correct update.

> These are the Encrypt/Decrypt/Wrap/Unwrap usage flags on Authentication key 03. Given that the key can be used only for Signatures, this is expected.

The point is OpenPGP 3.4.1 (and introduced in 3.3) allows for keys 02 and 03 to have alternate usages, but only if the the card supports MSE:
"7.2.18 MANAGE SECURITY ENVIRONMENT
This optional command (announced in Extended Capabilities) assigns a specific key to a
command. After selecting the OpenPGP application the private keys are linked to a default
command:
• Signature key (Key-Ref 1) binded to PSO:COMPUTE DIGITAL SIGNATURE
• Decryption key (Key-Ref 2) binded to PSO:DECIPHER
• Authentication key (Key-Ref 3) binded to INTERNAL AUTHENTICATE
The DEC-key (Key-Ref 2) can be assigned to the command INTERNAL AUTHENTICATE
and the AUT-Key (Key.Ref 3) can be linked to the command PSO:DECIPHER also. Before
running these commands the assignments can be SET with the MSE command by referencing
the functionality in P2 (A4 for INT-AUT and B8 for PSO:DEC) and addressing the
new key in a Control Reference Template (CRT) with his reference number (Key-Ref) in
the command data.

The problem was the OpenSC card-openpgp.c code was setting usage for 03 without checking if the card supported it. 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-openpgp.c#L2211-L2218 Note comment says "key 2  (auth key)" but the code checks for not 01 or not 02 which is key 03.

It looks like that was the difference between the Nitro Start and Nitro Pro. Start does not do MSE, but Pro does. This problem only showed up when pkcs11-tool assumed an RSA key could do DECIPHER  no mater what,  but the token could not. 

IMHO, OpenPGP mixed up "key ref" and "algo" from version 1.0 assuming a key could only do one algo.

Jakuje:
@dengert Can you add the changes to the ref json file to this PR so we can see what you see?

Jakuje:
> But Check Code Style fails: https://github.com/OpenSC/OpenSC/actions/runs/8007038450/job/21870157954?pr=3042
> It suggests to indent most of the file with an extra tab?

Yeah. I saw it last time I was updating them. Not sure what would be the best. We can update them to be more indented when they are geneterated, but given that this indentation does not add much readability, I would propose to make the clang ignore the json files.

But good to see that with the updated ref file the test works.

Jakuje (approved):
the changes look good. The current test-openpgp log is terribly long, but I think we can live for it until we will debug the issue.

.github/restart-pcscd.sh
```diff
@@ -3,6 +3,8 @@
 # This file is made to be sourced into other test scripts and not executed
 # manually because it sets trap to restore pcscd to working state
 
+# Set PCSCD_DEBUG="-d -a" to debug APDUs before sourcing this file
+
```
Killernate520, line 7: Duplicate of #

PR #3046, opened 2024-02-23

popovec:


frankmorgner (approved):

Jakuje:
Thanks for thinking about this for every new release :)

Jakuje (approved):

frankmorgner:
Even better if we could review the actual issue after the release, again 😉

PR #2812, opened 2023-06-26
include ATR of SafeNet eToken 5110 in the IDPrime driver

jurajsarinay:
I have a USB token labelled *SafeNet 5110* (cf. https://github.com/OpenSC/OpenSC/issues/1320) known to be an IDPrime "card". It turns out to work well with the IDPrime driver (possibly thanks to  #2666) after adding the appropriate ATR (also listed [here](https://github.com/LudovicRousseau/pcsc-tools/blob/d202099ef9344ad3930980cdf70573639f2befc3/smartcard_list.txt#L13287-L13288)) to [card-idprime.c](https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-idprime.c).

``` text
$./pkcs11-tool --login --test
Using slot 1 with a present token (0x4)
Logging in to "QES JS".
Please enter User PIN: 
C_SeedRandom() and C_GenerateRandom():
  seeding (C_SeedRandom) not supported
  seems to be OK
Digests:
  all 4 digest functions seem to work
  MD5: OK
  RIPEMD160: OK
  SHA-1: OK
  SHA256: OK
Signatures (currently only for RSA)
  testing key 0 (Private key 1) 
  all 4 signature functions seem to work
  testing signature mechanisms:
    RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
Verify (currently only for RSA)
  testing key 0 (Private key 1)
    RSA-PKCS: OK
Decryption (currently only for RSA)
  testing key 0 (Private key 1) -- can't be used to decrypt, skipping
No errors
```
I only have a single (production) token. I am therefore not very keen on testing any other (advanced) features of the driver.

It appears that none of the IDPrime cards are included in the Windows minidriver. I guess this must have been deliberate. An attempt to include the token resulted in an error.
``` text
>certutil /scinfo
=======================================================
Analyzing card in reader: SafeNet eToken 5100 [eToken 5110 SC] 01 00
Microsoft Base Smart Card Crypto Provider: Missing stored keyset
Microsoft Smart Card Key Storage Provider: Missing stored keyset

--------------===========================--------------
CertUtil: -SCInfo command FAILED: 0x80090016 (-2146893802 NTE_BAD_KEYSET)
CertUtil: Keyset does not exist
```

Feel free to change `SC_CARD_TYPE_IDPRIME_GENERIC` (or the description) to a more appropriate value.

##### Checklist

- [x] PKCS#11 module is tested

PR #3048, opened 2024-02-26
Add SafeNet 5110+ FIPS token support

metsma:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-idprime.c
+++ b/src/libopensc/card-idprime.c
@@ -57,6 +57,10 @@ static const struct sc_atr_table idprime_atrs[] = {
 	  "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:ff:ff:ff",
 	  "Gemalto IDPrime 930/3930",
 	  SC_CARD_TYPE_IDPRIME_930, 0, NULL },
+	{ "3b:ff:96:00:00:81:31:fe:43:80:31:80:65:b0:84:65:66:fb:12:01:78:82:90:00:85",
+	  "ff:ff:00:ff:ff:ff:ff:00:ff:ff:ff:ff:ff:ff:00:00:00:00:ff:ff:ff:ff:ff:ff:00",
+	  "based Gemalto IDPrime 930",
+	  SC_CARD_TYPE_IDPRIME_930, 0, NULL },
 	{ "3b:7f:96:00:00:80:31:80:65:b0:85:59:56:fb:12:0f:fe:82:90:00",
 	  "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:00:00:00:ff:00:00:ff:ff:ff",
 	  "Gemalto IDPrime 940",
@@ -648,10 +652,8 @@ static int idprime_init(sc_card_t *card)
 	if (card->type == SC_CARD_TYPE_IDPRIME_930
 	    || card->type == SC_CARD_TYPE_IDPRIME_940) {
 		_sc_card_add_rsa_alg(card, 4096, flags, 0);
-	}
 
-	/* Set up algorithm info for EC */
-	if (card->type == SC_CARD_TYPE_IDPRIME_940) {
+		/* Set up algorithm info for EC */
 		flags = SC_ALGORITHM_ECDSA_RAW | SC_ALGORITHM_ECDSA_HASH_NONE;
 		ext_flags = SC_ALGORITHM_EXT_EC_F_P
 			| SC_ALGORITHM_EXT_EC_ECPARAMETERS
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,5 @@ src/tests/fuzzing/fuzz_pkcs15init
 src/tests/fuzzing/fuzz_scconf_parse_string
 
 version.m4.ci
+openssl_arm64
+openssl_bin


src/libopensc/card-idprime.c
```diff
@@ -57,6 +57,10 @@ static const struct sc_atr_table idprime_atrs[] = {
 	  "ff:ff:00:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:ff:ff:ff",
 	  "Gemalto IDPrime 930/3930",
 	  SC_CARD_TYPE_IDPRIME_930, 0, NULL },
+	{ "3b:ff:96:00:00:81:31:fe:43:80:31:80:65:b0:84:65:66:fb:12:01:78:82:90:00:85",
+	  "ff:ff:00:ff:ff:ff:ff:00:ff:ff:ff:ff:ff:ff:00:00:00:00:ff:ff:ff:ff:ff:ff:00",
```
metsma, line 61: I am not sure how the mask is selected?

src/libopensc/card-idprime.c
```diff
@@ -452,6 +456,7 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
 				new_object.key_reference = 0x41 + cert_id;
 				break;
 			case SC_CARD_TYPE_IDPRIME_930:
+			case SC_CARD_TYPE_IDPRIME_930_2:
```
metsma, line 459: Dunno if this is correct case? If it works then it is?

Jakuje: Correct.

src/libopensc/cards.h
```diff
@@ -244,6 +244,7 @@ enum {
 	SC_CARD_TYPE_IDPRIME_3810,
 	SC_CARD_TYPE_IDPRIME_830,
 	SC_CARD_TYPE_IDPRIME_930,
+	SC_CARD_TYPE_IDPRIME_930_2,
```
Jakuje, line 247: If I read it right, the new ATR matches basically the IDPrime 930, but it does support EC?

In that case, I would propose to use some better name to differentiate from the "short" IDPrime ATRs. Your ATR almost matches the `SC_CARD_TYPE_IDPRIME_GENERIC`, which was added as part of #2812, where the token is referenced as `eToken 5110`. Is there a way to make it more generic ATR mask to match both of them? Tagging @jurajsarinay to have a look or get some clarification if his eToken supports EC or his ATR can be used for some simplification here.

src/libopensc/card-idprime.c
```diff
@@ -646,12 +652,14 @@ static int idprime_init(sc_card_t *card)
...
 	_sc_card_add_rsa_alg(card, 2048, flags, 0);
 	if (card->type == SC_CARD_TYPE_IDPRIME_930
+	    || card->type == SC_CARD_TYPE_IDPRIME_930_2
 	    || card->type == SC_CARD_TYPE_IDPRIME_940) {
 		_sc_card_add_rsa_alg(card, 4096, flags, 0);
 	}
 
 	/* Set up algorithm info for EC */
-	if (card->type == SC_CARD_TYPE_IDPRIME_940) {
+	if (card->type == SC_CARD_TYPE_IDPRIME_930
```
Jakuje, line 661: should this one be the `SC_CARD_TYPE_IDPRIME_930_2`?

metsma: My bad, fixed

metsma:
I received today ECC certificates for this device and I can confirm ECC signing works.
Started looking the code and seems like ECC dervice is not implemented?

src/libopensc/cards.h
metsma: First I tried to match SC_CARD_TYPE_IDPRIME_940 atr, but there are couple cases where it breaks.
This device does not have  0005 file
https://github.com/OpenSC/OpenSC/pull/3048/files#diff-1892ee089e440f54f73abfe799748d7cdbdf540419ae7898086382caeb395b43L582

And also I am not sure the key_reference arithmetic? how I can verify my device correct key_reference
https://github.com/OpenSC/OpenSC/pull/3048/files#diff-1892ee089e440f54f73abfe799748d7cdbdf540419ae7898086382caeb395b43L455
Maybe I can use the 940 key_reference case here
https://github.com/OpenSC/OpenSC/pull/3048/files#diff-1892ee089e440f54f73abfe799748d7cdbdf540419ae7898086382caeb395b43L457

metsma:
```
Token category: Hardware
Reader name: SafeNet eToken 5110+ FIPS 0
Serial number (PKCS#11): ....
Free space: 53788
Card ID (GUID): 0x...
Product name: eToken 5110+ FIPS
Card type: IDPrime
Applet Version: IDPrime Java Applet 4.5.0.F
Mask version: G286
Token Password: Present
Token Password retries remaining: 5
Maximum Token Password retries: 5
Token Password expiration: No expiration
Administrator Password: Present
Administrator Password retries remaining: 5
Maximum administrator Password retries: 5
FIPS: FIPS 140-2 L2
Full Secure Messaging (SM): No
Sign padding on-board: Yes
Supported key size: 4096 bits
ECC: Supported
CSP: eToken Base Cryptographic Provider
KSP: SafeNet Smart Card Key Storage Provider
```

Jakuje: > And also I am not sure the key_reference arithmetic? how I can verify my device correct key_reference

Usually by running the `pkcs11-tool --test` or `p11test` with the token and having more than one keys on the token loaded. The tools do signature+verification process  and if the key identifier would not match the calculation, the signature would either fail (if wrong key reference would be used) or the signature verification would fail (if the key reference matching different key than we think would be used).

metsma: Seems like this test is RSA only
```
pkcs11-tool --test
  seems to be OK
Digests:
Failed to load legacy provider
Ciphers: not implemented
Signatures: no private key found in this slot
  No private key found for testing
Unwrap: not implemented
No errors
```

Jakuje: Yes, pkcs11-tool test supports only RSA. If you want more extensive tests, I can recommend the p11test in `src/tests/p11test` (not packaged, needs to be built from source).

metsma:
Seems like signing works in my computer. I will let test this our test team.
Regarding ECC key derive, it does not work with SAC tools either.

Jakuje:
Whats is the status of this PR? Is there some simplification that can be implemented, or should we consider this as final?

If I see right, the only difference from `SC_CARD_TYPE_IDPRIME_930` to `SC_CARD_TYPE_IDPRIME_930_2` is that for the _2 you add the EC support. I think I had only one 930 without EC keys, but they are likely supported so I would propose to avoid adding the _2 and just use the `SC_CARD_TYPE_IDPRIME_930` and add the EC support to that.

metsma:
> Whats is the status of this PR? Is there some simplification that can be implemented, or should we consider this as final?
> 
> If I see right, the only difference from `SC_CARD_TYPE_IDPRIME_930` to `SC_CARD_TYPE_IDPRIME_930_2` is that for the _2 you add the EC support. I think I had only one 930 without EC keys, but they are likely supported so I would propose to avoid adding the _2 and just use the `SC_CARD_TYPE_IDPRIME_930` and add the EC support to that.

Does it support RSA 4k? If so I can merge these two

Jakuje:
> > Whats is the status of this PR? Is there some simplification that can be implemented, or should we consider this as final?
> > If I see right, the only difference from `SC_CARD_TYPE_IDPRIME_930` to `SC_CARD_TYPE_IDPRIME_930_2` is that for the _2 you add the EC support. I think I had only one 930 without EC keys, but they are likely supported so I would propose to avoid adding the _2 and just use the `SC_CARD_TYPE_IDPRIME_930` and add the EC support to that.
> 
> Does it support RSA 4k? If so I can merge these two

Looking at the card I have, we already have there some EC keys so EC keys is ok to add. We do not have 4k keys there, but I believe it should support them too. So lets merge them.

metsma:
> > > Whats is the status of this PR? Is there some simplification that can be implemented, or should we consider this as final?
> > > If I see right, the only difference from `SC_CARD_TYPE_IDPRIME_930` to `SC_CARD_TYPE_IDPRIME_930_2` is that for the _2 you add the EC support. I think I had only one 930 without EC keys, but they are likely supported so I would propose to avoid adding the _2 and just use the `SC_CARD_TYPE_IDPRIME_930` and add the EC support to that.
> > 
> > 
> > Does it support RSA 4k? If so I can merge these two
> 
> Looking at the card I have, we already have there some EC keys so EC keys is ok to add. We do not have 4k keys there, but I believe it should support them too. So lets merge them.

Done

Jakuje (approved):
@xhanulik is this good to go to the 0.25.0?

xhanulik (approved):
> @xhanulik is this good to go to the 0.25.0?

Yes, I think so.

Jakuje:
> Seems currently opensc doesn't support key object creation and key generation. Are there any plans to support it? Thanks!

If you can contribute the code or documentation, somebody can have a look into that. But with the lack of both of these, it is very complicated. I think it might even involve secure messaging, which would require getting the transport encryption keys from somewhere.

PR #3051, opened 2024-02-27
dtrust-tool: Fix NULL dereference with pinpad

Jakuje:
Thanks coverity:

CID 414970
CID 414971
CID 414972
CID 414973

Fixup #3026

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/dtrust-tool.c
+++ b/src/tools/dtrust-tool.c
@@ -177,12 +177,11 @@ unlock_transport_protection(sc_card_t *card)
 			fprintf(stderr, "New signature PINs doesn't match.\n");
 			goto fail_qespin2;
 		}
+		data.pin1.data = (u8 *)tpin;
+		data.pin1.len = strlen(tpin);
+		data.pin2.data = (u8 *)qespin1;
+		data.pin2.len = strlen(qespin1);
 	}
-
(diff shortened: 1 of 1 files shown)


Jakuje:
@hamarituc can you have a look if this works for you? Can you test the tool with pinpad reader?

hamarituc:
> @hamarituc can you have a look if this works for you? Can you test the tool with pinpad reader?

Good spot. That was a pratfall from my side, which shouldn't have happened :see_no_evil:.

I can confirm your patch fixes the SEGFAULT. You may cherry-pick https://github.com/hamarituc/OpenSC/commit/694ad25e5a9b3dc4214d2091d7ddacd14007f7c2 as the pin pad entry requires to specify minimum and maxium PIN length, which is reasonable for non pin pad cases, too.

But I was not able to successfully test it with my pin pad reader. It's a Cherry Secure Board with an integrated smart card readers. But I wasn't able to use the pin pad function of this keyboard in the past. I suppose it's an issue with my reader and not the `dtrust-tool`.

Jakuje:
Thanks. Pulled your commit. I also had issues with pinpad readers so I do not have a good one to recommend.

frankmorgner (approved):

xhanulik (approved):

Issue #3035, opened 2024-02-18
Update Links in README.md before making a release


msetina:
README file contains:
Do NOT use any links from wiki to download the OpenSC because wiki can be modified by anybody, see
[#2554](https://github.com/OpenSC/OpenSC/issues/2554). For downloading OpenSC, use the links here in README.

# Downloads

[OpenSC 0.23.0](https://github.com/OpenSC/OpenSC/releases/tag/0.23.0) is the latest stable version released on
29.11.2022. It is available as

 * Windows installer
   * [OpenSC-0.23.0_win64.msi](https://github.com/OpenSC/OpenSC/releases/download/0.23.0/OpenSC-0.23.0_win64.msi) for 64 bit programs
   * [OpenSC-0.23.0_win32.msi](https://github.com/OpenSC/OpenSC/releases/download/0.23.0/OpenSC-0.23.0_win32.msi) for 32 bit programs
 * [OpenSC-0.23.0.dmg](https://github.com/OpenSC/OpenSC/releases/download/0.23.0/OpenSC-0.23.0.dmg): macOS installer
 * [opensc-0.23.0.tar.gz](https://github.com/OpenSC/OpenSC/releases/download/0.23.0/opensc-0.23.0.tar.gz): Source code distribution

frankmorgner:
We will not change the 0.24.0 artifacts anymore, but with your comment in mind, we'll try to make sure that 0.25.0 files contain the correct links

Jakuje:
Thank you for the suggestion! I did it after the release last time as files exist only after we tag the release, the artifacts are built and uploaded to the release page itself. The paths are fixed so it can be certainly done before the release. But I wanted to avoid to have a dead links on the main page of the project before all the stuff will be ready.
Other option might be going back to the wiki given that we did not open it for modification of non-members (and I do not think we will do it any time soon).

PR #3052, opened 2024-02-27
doc: generalize README.md

frankmorgner:
Note the use of http://htmlpreview.github.io for rendering HTML pages from Github. From now on, no more modification to README.md should be needed on releases, as we use the Github sticky link https://github.com/OpenSC/OpenSC/releases/latest

fixes https://github.com/OpenSC/OpenSC/issues/3035

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,23 @@
 # OpenSC documentation
 
+Manual pages for the [OpenSC command line tools](htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/master/doc/tools/tools.html) as well as for the [OpenSC configuration files](http://htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/master/doc/files/files.html) are available online and typically distributed along with your installation.
 
-Wiki is [available online](https://github.com/OpenSC/OpenSC/wiki)
-
-Please take a look at the documentation before trying to use OpenSC.
-
-Do NOT use any links from wiki to download the OpenSC because wiki can be modified by anybody, see
-[#2554](https://github.com/OpenSC/OpenSC/issues/2554). For downloading OpenSC, use the links here in README.
+The [OpenSC Wiki](https://github.com/OpenSC/OpenSC/wiki) includes, among others, information for:
+ * [Windows Quick Start](https://github.com/OpenSC/OpenSC/wiki/Windows-Quick-Start)
+ * [macOS Quick Start](https://github.com/OpenSC/OpenSC/wiki/macOS-Quick-Start)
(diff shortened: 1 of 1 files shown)


frankmorgner:
@Jakuje could you please regenerate files.html and tools.html for the new release? on my machine, this unfortunately creates non-utf8 characters that cannot be processed by xsltproc. Thank you.

Jakuje:
> @Jakuje could you please regenerate files.html and tools.html for the new release? on my machine, this unfortunately creates non-utf8 characters that cannot be processed by xsltproc. Thank you.

Sure. I thought this is somehow automatic, but it does not look like it is. Should we add it also to the release checklist? I think adding it as part of the build would cause some unrelated changes as the paths are part of these files. On the other hand, I would like to avoid keeping these files outdated forever.

Jakuje:
The html files have `ISO-8859-1` encoding/charset in the HTML header. It looks like this is part of the `docbook.xsl`. I changed the xsltproc to build utf8 and rebuilt the html pages is filled as #3054.

Regarding the automatic rebuild of HTML pages as you propose, it sounds reasonable.

frankmorgner:
I tested the additional rule to generate files.html and tools.html and it doesn't work as I like to. The basic problem is that, for example, it is possible that someone modifies the XML based documentation based files without updating the HTML files, because he doesn't actually use `--enable-doc`. When this commit is checked in, `git` doesn't restore the modification time of the XML file and when running `make dist` (even with documentation enabled), the build process doesn't pick up the change  to rebuild the HTML files.

I think we can move forward with the following options:
1. update the release guide to re-generate the HTML files manually
2. set up a CI job that checks if regenerating of documentation is required. A developer will have to use `--enable-doc` to fix the failed build.  (Theoretically, the CI job could generate and check in the HTML changes  automatically, but I don't really like automatic changes of the source code for security reasons.)
3. use a phony Makefile target for *always* generating the HTML files if possible so that someone at some time will stumble upon the necessary rebuild.

I think, I favor going with the first option, because this causes the fewest work and gives the developer the most freedom in extending the documentation (an updated HTML files is really only needed when a new release is done).

Jakuje:
Thanks for checking this. I think the release checklist is already long enough so I would like to avoid adding more manual stuff.

I think we can do similar job as the formatting job, that it will try to generate the html files (with the right prefix/paths and everything) and if it will not match what is in the repository, it can print a diff, that can be either manually pulled by the contributor to fix the HTML or the contributor can regenerate the HTML files with `make ...` (given that we would provide reasonable description in the CI job).

I can probably take a look into that some time later.

frankmorgner:
Agreed, I removed the change to Makefile.am and created a seperate new issue so that we don't forget.

Issue #2577, opened 2022-07-12
Support meson build system


Jakuje:
### Problem Description

As proposed by @jpakkane in https://github.com/OpenSC/OpenSC/pull/2567#issuecomment-1170234236 the autotools build system is outdated, inflexible and can cause a barrier for new users interested in contributing to opensc.

Right now, we have autotools and some windows Makefile.mak which are used to build the windows binaries on appveyor.

### Proposed Resolution

Implement meson build system to simplify building opensc.

We would probably have to keep both of them for some transition period.

metsma:
I vote for CMake.
Better support with IDEs.

eli-schwartz:
> However, for most of my projects I still stuck to autotools, because the tools are easily available everywhere except on Windows. So when switching to some new build system we need to make sure that it still exists in 10 years from now.

Scons, meson, and cmake are all at least a decade old, there's quite some staying power with that. But scons isn't "like meson, written in python". The language implementation of a program is not the distinction of that program.

Scons is *literally* a library toolkit for writing your own build system in python, while Meson is a DSL that isn't python.

Not relying on a programming language to write build definitions is a good thing. It means that people can more easily recreate that build system themselves, if they need to, to make sure it still exists for their use cases. In fact, there's [a c99 implementation of Meson](https://sr.ht/~lattis/muon), created by someone who just wanted a version of Meson that is easily bootstrappable on operating systems that don't have python.


As far as whether Meson exists in 10 years, the GNOME ecosystem has made it a goal to use Meson everywhere, and is largely successful in that goal. Mesa, systemd, gstreamer, xorg and wayland, and a variety of Freedesktop software use it. To give some particularly notable names. OpenSC would not be a trend-setter in choosing Meson, or running the risk of betting on some obscure new build system no one uses.

While it's certainly true that "which build system to prefer" can be deeply personal and I hesitate to call anyone "wrong" for preferring cmake because they like it better... I don't think it's fair to claim that cmake is somehow the best choice ***because only cmake is an established build system***. ????

It's certainly worth considering to use a well-established build system called "Meson" if someone else puts in the work to port it, as I believe @jpakkane was intending to do...

> It is one tool, that can configure, build and install your software

No it isn't. Just like Meson, cmake is a generator, not a builder. It creates Makefiles or ninja files or Visual Studio / Xcode solutions, and then after running cmake, you run make/ninja/whatever.

[...]

PR #3053, opened 2024-02-27
Fixed pkcs11-register defaults on macOS and Windows

frankmorgner:
pkcs11-register command line defaults are based on the platform. On macOS/Windows, we don't want to register pkcs#11 modules to avoid conflict with the native API. For this to be effective, we need to build cmdline explicitly on those platforms.

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
The generated commandline files are causing problems with `-Wshorten-64-to-32`, which is why https://github.com/OpenSC/OpenSC/pull/3053/commits/e59586dd9f946b00a3f42d8b370b558ad8bcbf6c detects and disables this warning. The strict build, however, explicitly enables this warning - overwriting the local setting for the command line file:
```
clang -DHAVE_CONFIG_H -I. -I../..  -I../../src -D'DEFAULT_PKCS11_PROVIDER="/usr/local/lib/opensc-pkcs11.so"' -D'DEFAULT_ONEPIN_PKCS11_PROVIDER="/usr/local/lib/onepin-opensc-pkcs11.so"'   -Wno-unknown-warning-option -Wno-shorten-64-to-32 -Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing=2 -DDEBUG_PROFILE=1 -pedantic -Werror -Wall -Wno-strict-prototypes -Wno-empty-translation-unit -Wno-incompatible-library-redeclaration -Wno-variadic-macros -Wno-unused-function -Wno-incompatible-pointer-types-discards-qualifiers -Wno-gnu-zero-variadic-macro-arguments -Wno-format-pedantic -Wno-pedantic -Wno-incompatible-function-pointer-types -Wshorten-64-to-32 -c -o goid_tool-goid-tool-cmdline.o `test -f 'goid-tool-cmdline.c' || echo './'`goid-tool-cmdline.c
```
@Jakuje , do you know how to circumvent this?

Jakuje:
> @Jakuje , do you know how to circumvent this?

I was fixing this manually in https://github.com/OpenSC/OpenSC/commit/b21d01eee0817ae8130914e24cb8e2bc552dae09#diff-b9804e986581c6f7b6926feede5d5abddf4af651138eca115894fef96632e309

Together with the typo that I managed to push also to upstream: 

https://savannah.gnu.org/bugs/?64743
https://git.savannah.gnu.org/cgit/gengetopt.git/commit/?id=bef6597d1d077fe751a89307ea4f7cc5701653ef

I can try to fix also the cast in gengetopt, but given that this commit was first in 4 years since last release, I am not sure how active development is and when we can expect this to get pulled by distributions. I filled this here:

https://savannah.gnu.org/bugs/index.php?65417

For the time being, I believe we should not regenerate the cmdline files as part of the build process (at least not for the goid that is having this issue).

Side note: Once you do the detection of the `-Wshorten-64-to-32` support in configure, I think we can remove it from the list of forced CFLAGS in strict build as it becomes and it should be pulled automatically.

Other option to avoid this issue might be not to use the `int` argument in the goid tool. The short/long already have the casts or do not need it so they should not demonstrate this issue and code change should be quite minimal.

frankmorgner:
Thanks for tracking this to the original source.

In CI, we're currently not using `--enable-strict` to set default CFLAGS so I think the detection in configure.ac will not help.

I pushed a commit with the suggested workaround of using a short.

Jakuje:
One more thing. Once more regenerate the cmdline files during the build it is a question, whether the resulting files should be in git at all. Indeed, it helps to keep track of changes in the gengetopt tool (if there are some), but otherwise it will just cause these issues. If we add gengetopt as a dependency, we can generate them during the build (and during the `make dist`).

If we want to keep them in git, I think we should not codespell and code style check them as there is really nothing the PR contributor could do about typos in them and formatting issues. Unfortunately, I do not see an option in https://github.com/yshui/git-clang-format-lint to exclude files (but that should not be an issue), but codespell has an exclude_file, which can contain the files to exclude so thats what 

https://github.com/codespell-project/actions-codespell?tab=readme-ov-file#parameter-exclude_file

frankmorgner:
There is no easy way to get gengetopt on Windows - we would have to compile this ourselves. Yes, we are already doing this for OpenSSL and zlib, but those libraries are popular enough to have compiled binaries available somewhere. Basically, developer on-boarding would be a little more complicated on Windows. #2577 would fix that to some extent, but l currently don't have time to take a deeper look into this.

packaging/opensc.spec
```diff
@@ -17,6 +17,7 @@ BuildRequires:  docbook-style-xsl
 BuildRequires:  autoconf automake libtool gcc
 BuildRequires:  bash-completion
 BuildRequires:  zlib-devel
+BuildRequires:  gengetopt
```
Jakuje, line 20: Is this needed for the build from the dist tarball? If so, it might be a problem as we do not have the gengetopt in RHEL (only in EPEL).

frankmorgner: You can build without gengetopt, but obviously OpenSC would not include egk-tool goid-tool opensc-asn1 pkcs11-register, opensc-notify and npa-tool.

frankmorgner:
Finally... after many hickups and corner cases, gengetopt is now an optional build dependency and we don't need to handle cmdline files in the source code distribution.

src/tools/Makefile.am
```diff
@@ -22,26 +22,42 @@ do_subst = $(SED) \
...
 bin_PROGRAMS += cryptoflex-tool pkcs15-init netkey-tool piv-tool \
 	westcos-tool sc-hsm-tool dnie-tool gids-tool
+endif
+
+if ENABLE_GENGETOPT
+bin_PROGRAMS += egk-tool goid-tool opensc-asn1 pkcs11-register
+
+if ENABLE_NOTIFY
+bin_PROGRAMS += opensc-notify
+endif
```
Jakuje, line 43: Can you move these two blocks below the `if ENABLE_OPENSSL` to avoid closing and opening the `ENABLE_OPENSSL` block before and after this?

src/tools/Makefile.am
```diff
@@ -101,51 +117,76 @@ dnie_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 gids_tool_SOURCES = gids-tool.c util.c
 gids_tool_LDADD = $(OPTIONAL_OPENSSL_LIBS)
 
-npa_tool_SOURCES = npa-tool.c fread_to_eof.c util.c npa-tool-cmdline.c
+npa_tool_SOURCES = npa-tool.c fread_to_eof.c util.c
+nodist_npa_tool_SOURCES = npa-tool-cmdline.c
```
Jakuje, line 121: should this be nodist? I would say that we should generate these files during dist and include them in the release tarball, similarly as we include the `configure`.

.appveyor.yml
```diff
@@ -62,6 +62,14 @@ install:
...
           appveyor DownloadFile "https://download.microsoft.com/download/1/7/6/176909B0-50F2-4DF3-B29B-830A17EA7E38/CPDK_RELEASE_UPDATE/cpdksetup.exe"
       }
+  - wget https://ftp.gnu.org/gnu/gengetopt/gengetopt-2.23.tar.xz
+  - tar xJf gengetopt-2.23.tar.xz
+  - cd gengetopt-2.23
+  - bash -c "exec 0</dev/null && ./configure --prefix=/ || cat config.log"
+  - bash -c "exec 0</dev/null && make"
+  - cd ..
+    #- curl -sSf -o C:\cygwin64\setup-x86_64.exe https://cygwin.com/setup-x86_64.exe
+    #- C:\cygwin64\setup-x86_64.exe -q -P gengetopt
```
Jakuje, line 72: I see this was likely an attempt to install it through cygwin. Is it still needed or is is just an artifact of failed attempt to avoid people trying? I would propose to remove the commented out lines or added some explanation why it did not work.

.appveyor.yml
```diff
@@ -62,6 +62,14 @@ install:
       If (!(Test-Path cpdksetup.exe )) {
           appveyor DownloadFile "https://download.microsoft.com/download/1/7/6/176909B0-50F2-4DF3-B29B-830A17EA7E38/CPDK_RELEASE_UPDATE/cpdksetup.exe"
       }
+  - wget https://ftp.gnu.org/gnu/gengetopt/gengetopt-2.23.tar.xz
+  - tar xJf gengetopt-2.23.tar.xz
+  - cd gengetopt-2.23
+  - bash -c "exec 0</dev/null && ./configure --prefix=/ || cat config.log"
+  - bash -c "exec 0</dev/null && make"
+  - cd ..
```
Jakuje, line 70: Looking at the above block, I see installation of other dependencies (openpace, zlib) is done in `c:\` directly, which makes the paths slightly nicer. Whats the reason for installing the gengeopt in different path? Can the `appveyor DownloadFile` be used (I assume it can do some caching)?

frankmorgner:
currently, this PR is designed to integrate gengetopt as mandatory build requirement. There is no other way to change default options based on the platform, I think. For example, if you generate a distribution tarball with cmdline files in it, they will contain the default values for your Linux machine. On macOS, we want different defaults when installing from the distribution tarball.

frankmorgner:
We could do the following:
- include the cmdline files in the distribution,
- but always with macOS/Windows defaults (safe for Linux, but not ideal).
- modify src/tools/pkcs11-register.desktop.in to explicitly add `--skip-chrome=off --skip-firefox=off`, which "restores" the Linux defaults

configure.ac
```diff
@@ -148,6 +148,8 @@ AX_CODE_COVERAGE()
 
 AX_CHECK_COMPILE_FLAG([-Wunknown-warning-option], [have_unknown_warning_option="yes"], [have_unknown_warning_option="no"])
 AM_CONDITIONAL([HAVE_UNKNOWN_WARNING_OPTION], [test "${have_unknown_warning_option}" = "yes"])
+AX_CHECK_COMPILE_FLAG([-Wshorten-64-to-32], [have_shorten_warning_option="yes"], [have_shorten_warning_option="no"])
+AM_CONDITIONAL([HAVE_SHORTEN_WARNING_OPTION], [test "${have_shorten_warning_option}" = "yes"])
```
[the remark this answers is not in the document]
frankmorgner: I will remove the shorten warning from the strict build. However, I'm not sure what you mean with the second part.

frankmorgner:
> We could do the following:
> 
>     * include the cmdline files in the distribution,
> 
>     * but always with macOS/Windows defaults (safe for Linux, but not ideal).
> 
>     * modify src/tools/pkcs11-register.desktop.in to explicitly add `--skip-chrome=off --skip-firefox=off`, which "restores" the Linux defaults

I have reverted my changes to the above approach. This now means, that we don't strictly need gengetopt at compile time and the cmdline options have the defaults for macos/windows. On Linux, where we more likely have gengetopt available, the defaults can be changed with the availability of the tool. Also, the XDG autostart of pkcs11-register enables firefox/chrome integration.

This change also resolves most of the review comments.

configure.ac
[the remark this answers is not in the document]
Jakuje: the yaml changed since this comment and as I did not provide the permalink, it points to different place. The original line meant is this one:

https://github.com/OpenSC/OpenSC/blob/f0a3f67ab1ee3aa8e763859fbcddde79e1c3d411/.github/workflows/linux-strict.yml#L72

as we will not have the `CLANG_CFLAGS` anymore, this line can be simplified too.

Jakuje (approved):
After the last part of the comment is resolved, I think we are good to go for 0.26.0

configure.ac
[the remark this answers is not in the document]
frankmorgner: so I remove this line entirely!?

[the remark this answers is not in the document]
Jakuje: Yes, I think that should do.

xhanulik:
After discussion with @Jakuje, I'll merge this one to have the release candidate out today. Please, create additional PR with changes from https://github.com/OpenSC/OpenSC/pull/3053#discussion_r1522781295.

src/tools/goid-tool-cmdline.c
```diff
@@ -428,7 +428,7 @@ check_possible_values(const char *val, const char *values[])
           ++found;
           last = i;
           if (strlen(values[i]) == len)
-            return i; /* exact match no need to check more */
+            return i; /* exact macth no need to check more */
```
xhanulik, line 431:
```suggestion
            return i; /* exact match no need to check more */
```

xhanulik (approved):

PR #3054, opened 2024-02-29
Regenerate html pages for #3052

Jakuje:
The HTML pages contain also some paths that were sometimes pointing in wrong direction. I used the standard prefix "/usr"` for the configure to get the sane paths:

```
./configure --prefix="/usr"
make -j8
cd doc/tools
make tools.html
cd ../files
make files.html
```

I also removed trailing whitespace from source documentation xmls before generating the HTML.

Note, that the HTML diff might look weird when there are non-ascii chararacters in the UTF8 terminals, but the HTML page has `ISO-8859-1` defined in the HTML Header so it should render ok.

##### Checklist

### DIFF
--- a/doc/docbook-utf8.xsl
+++ b/doc/docbook-utf8.xsl
@@ -1,0 +1,10 @@
+<?xml version='1.0'?>
+<xsl:stylesheet  xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+                 version="1.0">
+
+<xsl:import href="docbook.xsl"/>
+<xsl:output method="html"
+            encoding="UTF-8"
+            indent="no"/>
+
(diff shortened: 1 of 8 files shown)


Jakuje:
The original files had forced default ISO encoding, which did not play well with the javascript page written in utf8 forcing this encoding. In 21st century everything should be able to handle UTF8. The changes needed to do this are described in 

http://www.sagehill.net/docbookxsl/OutputEncoding.html

Applying them to the build process generates valid UTF8 files:

http://htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/9c8aeaabc964f8875ad272686db09f1f073454c6/doc/tools/tools.html

PR #3067, opened 2024-03-12
Remove reselection of DF after error in decipher, sign and derive

### LINKED ISSUE #3044: Reselection of DF after failure in `sc_pkcs15_decipher` function
### Problem Description

When running `p11test` with the PIV test card and RSA2048 keys, I encountered a problem: the `pss_oaep_test` decryption test fails with `ERROR:  [ SKIP 02 ] Re-authentication failed`. I've discovered that the `sc_pkcs15_decipher()` function is called twice from `pkcs15_prkey_decrypt()` in `framework-pkcs15.c`:

https://github.com/OpenSC/OpenSC/blob/c153e2fe979b05851ab06b45799a9358cdde6fe3/src/pkcs11/framework-pkcs15.c#L4611-L4619

In that failed test, the first `sc_pkcs15_decipher()` returns `SC_ERROR_WRONG_PADDING,` so it is called for a second time. Still, since the token requires authentication before doing the decipher operation, it returns `SC_ERROR_SECURITY_STATUS_NOT_SATISFIED.`

This error code is then converted into `CKR_USER_NOT_LOGGED_IN,` the login state is then reset, and `p11test` cannot continue since `always_authenticate()` in the next test case fails.

This problem also shows up by only decryption of data with invalid padding with `pkcs11-tool` as the error is
```
error: PKCS11 function C_Login failed: rv = CKR_OPERATION_NOT_INITIALIZED (0x91)
```
(probably) instead of
```
error: PKCS11 function C_DecryptUpdate failed: rv = CKR_ENCRYPTED_DATA_INVALID (0x40)
```

### Proposed Resolution

Not sure; I think that the reselection and second call to `sc_pkcs15_decipher()` in `pkcs15_prkey_decrypt()`
 shouldn't be done after every error code.

### Steps to reproduce

Create some bogus data to be decrypted (so after decryption, there isn't valid padding):

```
openssl rand -out random.txt 256
```
Decrypt the data with a token requiring always authenticate:

```
pkcs11-tool --decrypt --pin xxxxxx --id xx -m RSA-PKCS-OAEP --mgf MGF1-SHA256 --hash-algorithm SHA-1 -i random.txt
```

Output:

```
Using the decrypta algorithm RSA-PKCS-OAEP
OAEP parameters: hashAlg=SHA-1, mgf=MGF1-SHA256, source_type=0, source_ptr=(nil), source_len=0
error: PKCS11 function C_Login failed: rv = CKR_OPERATION_NOT_INITIALIZED (0x91)
Aborting.
```

### Logs

Here is the log starting at the first call to `sc_pkcs15_decrypt()`; it fails due to invalid padding, and then the second attempt to `sc_pkcs15_decrypt()` fails with `Security status not satisfied,` which is the error propagated outside.

[log.txt](https://github.com/OpenSC/OpenSC/files/14372944/log.txt)

xhanulik:
Fixes #3044

After the error in `sc_pkcs15_decipher`/`sc_pkcs15_derive`/`sc_pkcs15_derive`, the DF is reselected and the operation is called for second time. This is problematic for cards which require authentication before operations with private key. 
The code for reselecting the DF and second call for operations is removed, based on discussion in #3044.

##### Checklist

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -218,7 +218,6 @@ static CK_RV	get_ec_pubkey_point(struct sc_pkcs15_pubkey *, CK_ATTRIBUTE_PTR);
 static CK_RV	get_ec_pubkey_params(struct sc_pkcs15_pubkey *, CK_ATTRIBUTE_PTR);
 static int	lock_card(struct pkcs15_fw_data *);
 static int	unlock_card(struct pkcs15_fw_data *);
-static int	reselect_app_df(sc_pkcs15_card_t *p15card);
 
 #ifdef USE_PKCS15_INIT
 static CK_RV	set_gost3410_params(struct sc_pkcs15init_prkeyargs *,
@@ -4263,7 +4262,7 @@ pkcs15_prkey_sign(struct sc_pkcs11_session *session, void *obj,
 	struct sc_pkcs11_card *p11card = session->slot->p11card;
 	struct pkcs15_fw_data *fw_data = NULL;
 	CK_RV rv;
-	int flags = 0, prkey_has_path = 0, rc;
+	int flags = 0, rc;
 	unsigned sign_flags = SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_SIGNRECOVER
 			| SC_PKCS15_PRKEY_USAGE_NONREPUDIATION;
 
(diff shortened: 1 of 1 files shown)


frankmorgner (approved):

xhanulik:
I tried to test this with several cards we have, but unfortunately, none hit the reselection case.

Jakuje (approved):

PR #3071, opened 2024-03-14
Yubikey 5.3+: Add support for PIN policy and touch policy

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested


Jakuje (changes requested):
This wont work as the pin/touch policy is set on the key files and not the pin files. The current code (after going through the issues described inline), I am getting the following response:
```
P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] apdu.c:367:sc_single_transmit: CLA:0, INS:F7, P1:0, P2:80, data(0) (nil)                                                                                                         
P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] reader-pcsc.c:324:pcsc_transmit: reader 'Yubico YubiKey OTP+FIDO+CCID 00 00'
P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] reader-pcsc.c:325:pcsc_transmit:•
Outgoing APDU (5 bytes):
00 F7 00 80 64 ....d

P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] reader-pcsc.c:244:pcsc_internal_transmit: called
P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] reader-pcsc.c:334:pcsc_transmit:•
Incoming APDU (12 bytes):                                                                                          
01 01 FF 05 01 01 06 02 03 03 90 00 ............                                                                                                                                                                                             

P:2605154; T:0x140137054143232 13:55:26.355 [opensc-pkcs11] apdu.c:382:sc_single_transmit: returning with: 0 (Success)
```
The docs say "PIN and Touch policy of the key (keys only)" which means this policy lives only on the key/cert objects and it needs to be touched/pin verified.

The touching requirement on the yubikey worked though so after updating it, I will be able to test it further.

dengert:
For whatever it is worth, I wrote this 4 years ago, but no one was interested: https://github.com/dengert/OpenSC/tree/YubiKey5

src/libopensc/card-piv.c
```diff
@@ -5682,6 +5705,49 @@ static int piv_match_card_continued(sc_card_t *card)
...
+}
+
+static void
+piv_yk_init_metadata(sc_card_t *card)
+{
+	sc_apdu_t apdu;
+	u8 resp[100];
+	piv_private_data_t *priv = PIV_DATA(card);
+
+	if (priv->yubico_version < 0x00050300)
```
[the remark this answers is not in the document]
frankmorgner: done

src/libopensc/card-piv.c
```diff
@@ -5682,6 +5705,49 @@ static int piv_match_card_continued(sc_card_t *card)
...
+			*pin = policy[0];
+		if (touch)
+			*touch = policy[1];
+	} else {
+		/* The default behaviour is used */
+		if (pin)
+			*pin = 0x00;
+		if (touch)
+			*touch = 0x00;
+	}
```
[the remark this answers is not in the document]
frankmorgner: done

frankmorgner:
I fixed the inline comments. Feel free to add more commits on top of this PR (which should work as repo owner).

> This wont work as the pin/touch policy is set on the key files and not the pin files

My reading of the documentation is a little different:
- The policy is tied to the PIN object, that's why it can be read by referencing the *slot id*, i.e. PIN reference
- The policy is only evaluated when the key is used. For example, touching the token is only required when some APDU for signing some data has been sent.

Anyway, it should be easy to test.

> For whatever it is worth, I wrote this 4 years ago, but no one was interested: https://github.com/dengert/OpenSC/tree/YubiKey5

I don't think that the attestation certificate is the best place to check for the policies. My reading of the documentation is that the initial cert doesn't change when the policies are updated and that users may overwrite the initial cert with custom ones.

Jakuje:
> I fixed the inline comments. Feel free to add more commits on top of this PR (which should work as repo owner).

Thanks. I can try to do that, but first we should agree on the concept of the pin policy, see below. The way how I understand it from reading the manual pages and testing with the yubikey sounds incompatible with what is implemented here.

> > This wont work as the pin/touch policy is set on the key files and not the pin files
> 
> My reading of the documentation is a little different:
> 
>     * The policy is tied to the PIN object, that's why it can be read by referencing the _slot id_, i.e. PIN reference

I did not find any way to set the pin/touch policy for pins. The only operation that allows setting pin/touch policy is key generation or import as described in the manual page for yubico-piv-tool, which was where I started:

https://developers.yubico.com/yubico-piv-tool/Manuals/yubico-piv-tool.1.html

If you have some reference to the documentation that allows setting policies for PIN objects, I would be happy to test it.

>     * The policy is only evaluated when the key is used. For example, touching the token is only required when some APDU for signing some data has been sent.
> 
> Anyway, it should be easy to test.

Indeed, I tested this and it worked for me, as the operation was not allowed until I touched the token (with the touch policy), but there was no touch policy tag on the pin objects (and I believe it would be on the cert/key object, but I did not test it).

dengert:
@Jakuje I agree that the touch policy is set on the key, much like CKA_ALWAYS_AUTHENTICATE  method used with the PIV 9C  key. 

The code from 4 years ago: https://github.com/dengert/OpenSC/tree/YubiKey5  would in effect let user know they may need to touch the token: https://github.com/OpenSC/OpenSC/commit/06a967b8ab5f27206d7d75285773afc5f8423993#diff-c951b6f8d354170c2b32ab1f3cc4ae962b09dbef33734ba7fdca5e23c425bf56R1255-R1258
```
if (has_yubikey_touch_policy) {
			int offset = strlen(label);
			strncpy(label + offset, ",Yubikey has touch policy", SC_PKCS15_MAX_LABEL_SIZE - offset - 1);
		}
```
This is more of a friendly reminder to look at the token and if blinking, touch it whenever entering a PIN. 
Tested using Yubikey 5 NFC version 5.2.6 that Yubico has sent me at that time.

frankmorgner:
Sorry, I was misled by the documentation available at https://developers.yubico.com/PIV/Introduction/Yubico_extensions.html#_get_metadata. I now found https://docs.yubico.com/yesdk/users-manual/application-piv/slots.html which makes it more clear what they define to be slots. I'll rewrite the code to attach the policies to the keys instead of the PINs.

I find notifications more usable than changing the PIN's label, but your suggestion, Doug, is a good idea if notifications are not available.

src/libopensc/card-piv.c
```diff
@@ -4389,12 +4384,72 @@ static int piv_get_pin_preference(sc_card_t *card, int *pin_ref)
...
+
+	if (priv->yk_metadata[i].slot == 0x00) {
+		/* initialize this entry */
+		sc_format_apdu_ex(&apdu, 0x00, 0xF7, 0x00, slot, NULL, 0, resp, sizeof resp);
+		if (SC_SUCCESS == sc_transmit_apdu(card, &apdu) && SC_SUCCESS == sc_check_sw(card, apdu.sw1, apdu.sw2)) {
+			piv_yk_metadata_get_policy(card->ctx, resp, apdu.resplen,
+					&priv->yk_metadata[i].pin_policy,
+					&priv->yk_metadata[i].touch_policy);
+			sc_log(card->ctx, "PIN policy for slot 0x%02X: 0x%02X (touch 0x%02X)",
+					slot, priv->yk_metadata[i].pin_policy, priv->yk_metadata[i].touch_policy);
```
Jakuje, line 4435: I would propose to have here some return value to be able to distinguish (at least in the debug logs without parsing the ASN1 output manually), if the policy was not found or if it is set to 0 (default). Right now, the log will show up the same for these two cases.

src/libopensc/card-piv.c
```diff
@@ -4686,6 +4754,20 @@ piv_decipher(sc_card_t *card, const u8 * data, size_t datalen, u8 * out, size_t
...
+
+	piv_yk_get_metadata(card, priv->key_ref, NULL, &touch_policy);
+	switch (touch_policy) {
+		case 0x02:
+			sc_notify(title, "Touching the token is required for deciphering");
+			break;
+		case 0x03:
+			sc_notify(title, "Touching the token is required for deciphering (needed again after 15 seconds)");
+			break;
+	}
```
Jakuje, line 4769: Now, this code is the same for both sign/decipher. Would it be possible to somehow share it to avoid copy&paste?

src/libopensc/card-piv.c
```diff
@@ -4378,6 +4384,75 @@ static int piv_get_pin_preference(sc_card_t *card, int *pin_ref)
...
+		if (priv->yk_metadata[i].slot == 0x00)
+			/* reached the last initialized entry */
+			break;
+
+		if (priv->yk_metadata[i].slot == slot)
+			/* metadata already initialized */
+			break;
+	}
+
+	if (priv->yk_metadata[i].slot == 0x00) {
```
Jakuje, line 4427: if we fill the whole 26 slots in the metedata array, you will access memory behind the buffer. Even though well behaving yubikey should not have more than 26 slots (I think), it would be good to add a sanity check the `i` is smaller than the length of the buffer here.

src/libopensc/card-piv.c
frankmorgner: Yes, if we accept that the notification message is the same for both cases

src/libopensc/card-piv.c
frankmorgner: I added a log message

src/libopensc/card-piv.c
frankmorgner: the plan was to stop when `i` is exactly the last index. i changed the loop to fix this, thanks

src/libopensc/card-piv.c
Jakuje: Oh, i missed the difference in the message. But given that its just one word, I think it could be an argument.

Jakuje:
I will give it a try with a yubikey after the fuzzing issues are resolved.

src/libopensc/card-piv.c
frankmorgner: i reworded and unified the message

src/libopensc/card-piv.c
```diff
@@ -4378,6 +4384,77 @@ static int piv_get_pin_preference(sc_card_t *card, int *pin_ref)
...
+					slot, priv->yk_metadata[i].pin_policy, priv->yk_metadata[i].touch_policy);
+			priv->yk_metadata[i].slot = slot;
+		}
+	}
+
+	if (priv->yk_metadata[i].slot == slot) {
+		if (pin_policy)
+			*pin_policy = priv->yk_metadata[i].pin_policy;
+		if (touch_policy)
+			*pin_policy = priv->yk_metadata[i].touch_policy;
```
Jakuje, line 4446: there is still a crash on invalid write:
```suggestion
			*touch_policy = priv->yk_metadata[i].touch_policy;
```

src/libopensc/pkcs15-piv.c
```diff
@@ -1166,7 +1167,18 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
...
+			/* PIN is checked once for the session */
+			prkey_obj.user_consent = 0;
+			break;
+		case 0x03:
+			/* PIN is verified just before operation */
+			prkey_obj.user_consent = 1;
+		}
+		/* PIN is never checked for operations if PIN policy is set to 0x01 */
+		if (pin_policy != 0x01)
 			sc_pkcs15_format_id(prkeys[i].auth_id, &prkey_obj.auth_id);
```
Jakuje, line 1182: I think this will need also some more love. If I keep this part in, the pkcs11 fails to find my key, while it is visible ok in the pkcs15 layer without the `Auth ID` field:
```
Private EC Key [KEY MAN key]
	Usage          : [0x100], derive
	Access Flags   : [0x1D], sensitive, alwaysSensitive, neverExtract, local
	Algo_refs      : 0
	FieldLength    : 256
	Key ref        : 157 (0x9D)
	Native         : yes
	ID             : 03
	MD:guid        : 0x'30333862646435356335303039396362306465303666333530343033343931320000000000000000'
```

I do not know where did it go and was not able to tweak the above condition to show the key on the pkcs11 layer without the Auth ID.

Looking at the `CARD AUTH` key, it does not have the `SC_PKCS15_CO_FLAG_PRIVATE` flag. What I used and worked for me was:

```suggestion
		if (prkeys[i].auth_id && pin_policy != 0x01)
			sc_pkcs15_format_id(prkeys[i].auth_id, &prkey_obj.auth_id);
		else 
			prkey_obj.flags &= ~SC_PKCS15_CO_FLAG_PRIVATE;
```

I did not  try with the card auth key yet if the default policy will be 0x00 or 0x01. If we will be able to override the policy, we will need to be able to set the reverse of this for this key (add auth_id + `SC_PKCS15_CO_FLAG_PRIVATE` for this key).

frankmorgner: my intention was to create a key that doesn't need verification of the PIN. I thought that leaving auth_id would be all that's needed. I'll have a second look.

dengert: The Yubikey 5 NFCs tokens I have are version 5.2.6, but the metadata feature requires 5.3.0 which is per key(slot) and not in the Attestation cert, as there is only one. The metadata feature is what I was referring to. 

What looks strange is in the code  `piv_yk_get_metadata`  the `rbuf` is only 100 bytes.  The metadata returned could have an RSA pub key which will not fit. So 

https://docs.yubico.com/yesdk/users-manual/application-piv/apdu/metadata.html
says Le absent, which does not make sense, I assume it should be 0x00 with a large `rbuf`  and get responses will be done to return the data.

 The call to: `sc_card_ctl(p15card->card, SC_CARDCTL_PIV_YK_PIN_POLICY, &pin_policy);` does not have any way to return an error.

dengert:
The Yubico metadata has a lot of useful info. If you where to issue the Yubico get metadata command from `card-piv.c` and cache the data in `card-piv.c` it could be used to verify that the card has a key and the type and size as well as the public key, even if no certificate is present.  And if certificate is present, that the public keys match from metadata and certificate. If the metadata returns that no key is present, then read the certificate code could be skipped.  

Could be just before the  comment: `/* get the cert or the pub key out and into the cache too */`
The `piv_objects` structure could have the key id to use as the "slot" in the metadata and some other flags too.  

Standard PIV cards still need to read a certificate to get key type and size.

src/libopensc/card-piv.c
```diff
@@ -5685,6 +5708,55 @@ static int piv_match_card_continued(sc_card_t *card)
...
+}
+
+static void
+piv_yk_init_metadata(sc_card_t *card)
+{
+	sc_apdu_t apdu;
+	u8 resp[100];
+	piv_private_data_t *priv = PIV_DATA(card);
+
+	if (priv->yubico_version < 0x00050300) {
```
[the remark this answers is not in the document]
frankmorgner: sure, np

src/libopensc/pkcs15-piv.c
frankmorgner: removing SC_PKCS15_CO_FLAG_PRIVATE sounds reasonable, I committed that. thank you.

Scrolling down to
https://docs.yubico.com/yesdk/users-manual/application-piv/apdu/metadata.html#examples
we can see some examples for the command (generated with opensc-tool!) and it seems that a) the public key may indeed be present and be quite long and b) tags appear in sorted order so that the PIN policy appears early. So that, at elast should work as expected.

I think we can avoid propagating (and managing) some error code from SC_CARDCTL_PIV_YK_PIN_POLICY, because we do have safe defaults.

frankmorgner:
Reviewing https://docs.yubico.com/yesdk/users-manual/application-piv/attestation.html, it seems that the attestation statement is a good fallback mechanism in order to test for the PIN policy. From the documentation it is unclear whether the attestation statement is always newly created (i.e. a slow signature). Anyway, I'll check if I can copy that to the driver level... (feel free to jump in if you want to push some commit on top of this PR).

frankmorgner:
I added debug messages and error codes.

I'd also like to add support for the attestation statements as fall back, because is already implemented. I converted this PR into a draft until I have time to look at it.

Jakuje (commented):
was there any progress on the attestation? The current code looks good to go after resolving the conflict.

frankmorgner:
> was there any progress on the attestation?

Testing would be needed for
[ ] yubikey without support of touch policy
[ ] yubikey with support for attestation statements only
[ ] yubikey with support for commands for touch policy (5.3+)

I can only look at the first category.

Jakuje:
I have some old and new yubikeys. It would be sad to let this not get merged so I will give it a try. I rebased the branch and I will give it some more testing before opening a new PR

https://github.com/Jakuje/OpenSC/pull/new/yk_policy

frankmorgner:
thanks for caring. I'm happy to take a look once you're ready.

PR #3082, opened 2024-03-21
Coverity issues

xhanulik:
Fix for recently found CID 424593, 424594.

##### Checklist

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -293,8 +293,10 @@ openssl_enc(const EVP_CIPHER * cipher, const unsigned char *key, const unsigned 
 	ctx = EVP_CIPHER_CTX_new();
 	if (ctx == NULL)
 		goto out;
-	EVP_EncryptInit_ex(ctx, cipher, NULL, key, iv_tmp);
-	EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+	if (!EVP_EncryptInit_ex(ctx, cipher, NULL, key, iv_tmp) ||
+			!EVP_CIPHER_CTX_set_padding(ctx, 0))
+		goto out;
 
 	if (!EVP_EncryptUpdate(ctx, output, &outl, input, (int)length))
 		goto out;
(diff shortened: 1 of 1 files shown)


frankmorgner (approved):

Jakuje (commented):
nit: In the ways `out:` the check for non-NULL `ctx` are not needed

And once we touch it, we should probably check the return value of `EVP_CIPHER_CTX_set_padding` too.

xhanulik:
> And once we touch it, we should probably check the return value of EVP_CIPHER_CTX_set_padding too.

From documentation, `EVP_CIPHER_CTX_set_padding() always returns 1`. But I will remove the unneeded ifs.

Jakuje:
> From documentation, `EVP_CIPHER_CTX_set_padding() always returns 1`. But I will remove the unneeded ifs.

Good point. But it might change  so sanity-checking the return value should not hurt.

src/libopensc/card-epass2003.c
```diff
@@ -323,8 +324,10 @@ openssl_dec(const EVP_CIPHER * cipher, const unsigned char *key, const unsigned
 	ctx = EVP_CIPHER_CTX_new();
 	if (ctx == NULL)
 		goto out;
-	EVP_DecryptInit_ex(ctx, cipher, NULL, key, iv_tmp);
-	EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+	if (!EVP_DecryptInit_ex(ctx, cipher, NULL, key, iv_tmp)
+			|| EVP_CIPHER_CTX_set_padding(ctx, 0))
```
Jakuje, line 329: this should be negated. If you put the || on the first line, clang should not complain about it:
```suggestion
	if (!EVP_DecryptInit_ex(ctx, cipher, NULL, key, iv_tmp) ||
			EVP_CIPHER_CTX_set_padding(ctx, 0) != 1)
```

xhanulik: Yes, you're just too quick reviewing...

xhanulik:
Thanks for review, merging.

Issue #3108, opened 2024-04-10
Avoid non SELECT(AID) commands for probing


CardContact:
Further digging into #3107 I've extracted probing APDUs that are not SELECT(AID) APDUs:

00000010 [140479840638656] APDU: 00 CA DF 30 05 
00008056 [140479840638656] SW: 6A 80 
00000018 [140479840638656] APDU: 00 CB 3F FF 03 5C 01 7E 00 
00005369 [140479840638656] SW: 6E 00 
00000020 [140479840638656] APDU: 00 A4 00 0C 02 3F 00 
00005586 [140479840638656] SW: 6A 86 
00000011 [140479840638656] APDU: 00 A4 00 00 02 3F 00 
00004883 [140479840638656] SW: 6A 86 

The INS CA seems to be issued by the setcos driver.
The INS CB comes from the PIV driver.
The INS A4 seems to be related to the JPKI driver.

The issue with these APDUs is, that they potentially interfere with an application that is not supported by OpenSC.

In our case, the probing APDUs end up in a running applet that has an establish secure messaging channel. That channel is interrupted at plain APDUs for security reasons. SELECT(AID) for probing is fine, as those APDUs are handled by the JCVM and don't reach the applet. But all other APDUs are passed to the applet that uses the card in non-exclusive mode.

To have better coexistence with other card aware applications, we should avoid sending non SELECT(AID) APDUs for probing, unless we match the ATR on a white list. And that whitelist should be specific enough and not catch just all JavaCards.

dengert:
We live in a smart card jungle: 
- applets that can be loaded on almost any card, so the way to identify the applet is not by ATR, but by select AID. 
- cards or card vendors that support multiple applets on the same card. PIV and OpenPGP for example. 
- applets that treat any select AID as just a select file and trick a card driver that did not check the response to select AID for the AID in the response, Muscle card for example. 
- cards that will allow non-APDUs to be done in between SM protected APDUs without requiring reestablishment of SM. NIST PIV SM for example. 
- cards that are not supported (yet) by OpenSC, that if they were, would not have the problem in #3107 or #3108.


Some remedies to the above:
- provide OpenSC  support for the device. (I assume you will be doing that?)
- ignore, treat as error or allow non-SM APDU while in a the middle of a SM session  
- card driver can lock the card to avoid interference from other applications during some critical sequences of APDUs.
- whitelists or blacklists
[...]

PR #3112, opened 2024-04-12
card-piv.c - Card matching fix for unknown card

dengert:
During card matching, only use GET DATA of Discovery object before or in place of SELECT AID, if the ATR matches a known card or user has forced the card type or forced the use of a PIV driver.

This avoids possible state change in the unknown card if used by other processes.

partial fox for #3108

 On branch piv-fix-#3108
 Changes to be committed:
	modified:   card-piv.c

##### Checklist

- [ ] macOS tokend is tested


Jakuje (approved):
looks good!

xhanulik (approved):
LGTM, thank you!

Issue #3131, opened 2024-04-30
Support D-Trust Card 5.1 (Std. RSA CardOS6.0) with CAN


janknieling:
### Problem Description

The D-Trust Signature Card 5.1 Std. with CardOS6.0 is currently not supported by OpenSC.
To connect to the card/use the card a CAN number is needed (the CAN is printed on the card).
It seems that OpenSC hasn't the possibility to provide/submit a CAN.

### Proposed Resolution

It would be great if this card type is supported by OpenSC. This is merely a feature request than a bug report.
[log.txt](https://github.com/OpenSC/OpenSC/files/15170795/log.txt)

### Steps to reproduce
`echo "Sign me" | OPENSC_DEBUG=9 pkcs11-tool --slot 1 -a Signaturzertifikat -s -m RSA-PKCS -p XXXX &> log.txt`
### Logs

```
➜  ~ opensc-tool --version
OpenSC-<version not available>, rev: 0a4b772, commit-time: 2024-04-04 16:21:38 +0200
P:94404; T:0x140704682469312 22:40:04.528 [opensc-tool] ctx.c:747:process_config_file: scconf_parse failed: Line 2: not expecting 'default'

P:94404; T:0x140704682469312 22:40:04.529 [opensc-tool] ctx.c:981:sc_context_create: ===================================
P:94404; T:0x140704682469312 22:40:04.529 [opensc-tool] ctx.c:982:sc_context_create: OpenSC version: 0.25.1
P:94404; T:0x140704682469312 22:40:04.529 [opensc-tool] ctx.c:983:sc_context_create: Configured for opensc-tool (/Library/OpenSC/bin/opensc-tool)
P:94404; T:0x140704682469312 22:40:04.530 [opensc-tool] ctx.c:858:sc_openssl3_init: Failed to load OpenSSL Legacy provider
P:94404; T:0x140704682469312 22:40:04.530 [opensc-tool] reader-pcsc.c:897:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 reconnect_action=0 enable_pinpad=1 enable_pace=1
P:94404; T:0x140704682469312 22:40:04.546 [opensc-tool] reader-pcsc.c:1397:pcsc_detect_readers: called
P:94404; T:0x140704682469312 22:40:04.546 [opensc-tool] reader-pcsc.c:1410:pcsc_detect_readers: Probing PC/SC readers
[...]

PR #3137, opened 2024-05-08
Add support for D-Trust Card 5.1 & 5.4

hamarituc:
This PR adds support for D-Trust Cards 5.1 & 5.4 (see #3131).

There are a lot of changes introduced by this PR:

* [x] Card identification
* [x] Establish PACE channel
  * Some of the operations require to establish a secure channel between card reader an card. The `pin_cmd`-handler accounts for this and establishes a secure channel when necessary.
  * The state of the PACE channel is tracked by the driver to omit subsequent PACE operations when not necessary.
  * The drivers allows establishing a PACE channel explicitly by the `SC_PIN_CMD_VERIFY` command.
* [x] Query Card Access Number (CAN)
  * As most applications doesn't support prompting for the CAN, several sources are considered by the driver: static configuration, environment variables, manual input on the card readers pin pad
* [x] Cache CAN
  * Once a CAN was provided to the driver, it is saved in a synthetic PKCS#15 cache file and reused later.
* [x] Implement card initialization in `dtrust-tool`
* [x] Implement PIN management commands in `dtrust-tool` (change PIN, resume PIN, unblock PIN)

The following feature is postponed to a future PR:

* ~~Query interactively for CAN via Windows dialog or `pinentry`~~
  * This feature highly depends on the underlying operating system. I decided to postpone this feature to a future PR and to focus on using the cache feature in the meantime.

##### Checklist

- [ ] macOS tokend is tested


src/libopensc/card-dtrust.c
```diff
@@ -213,10 +271,44 @@ dtrust_init(sc_card_t *card)
...
+	if (drv_data == NULL)
 		return SC_ERROR_OUT_OF_MEMORY;
 
+	drv_data->pace = 0;
+	drv_data->can = 0;
+	drv_data->can_value = NULL;
+	drv_data->can_cache = 1;
+	drv_data->p15card = NULL;
+
+	drv_data->can_value = can_env = getenv("DTRUST_CAN");
```
frankmorgner, line 284: This buffer is not safe for later use. better use `getenv_s` or, at the very least, make a copy from the value.

src/libopensc/card-dtrust.c
```diff
@@ -521,13 +999,34 @@ dtrust_decipher(struct sc_card *card, const u8 *data,
...
-	sc_format_path("3F00", &path);
-	r = sc_select_file(card, &path, NULL);
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
-	return r;
+	drv_data = card->drv_data;
+
+	sc_sm_stop(card);
+	drv_data->pace = 0;
+	drv_data->can = 0;
```
frankmorgner, line 1011: I would prefer using NULL here

src/tools/dtrust-tool.c
```diff
@@ -197,29 +333,470 @@ unlock_transport_protection(sc_card_t *card)
...
+	/* We only have one chance to set the new PIN. Once the Transport PIN
+	 * is verified, it is not usable anymore. For pin pad readers we
+	 * continue as long as the new PIN is set successfully or the user
+	 * aborts the program and renders its card unusable as a consequence. */
+	do {
+		r = sc_pin_cmd(card, &data, NULL);
+		if (r == SC_SUCCESS) {
+			printf("Transport protection removed. You can now use your %s.\n", pin_label);
+			break;
+		}
```
frankmorgner, line 431: This is an endless loop, if the user always aborts the insertion of the PIN

src/tools/dtrust-tool.c
```diff
@@ -262,50 +857,254 @@ main(int argc, char *argv[])
...
-			pin_status(card, 0x03, "Card Holder PIN");
-		pin_status(card, 0x04, "Card Holder PUK");
-		pin_status(card, 0x87, "Signature PIN");
-		pin_status(card, 0x8B, "Transport PIN");
+		switch (card->type) {
+		case SC_CARD_TYPE_DTRUST_V4_1_STD:
+		case SC_CARD_TYPE_DTRUST_V4_1_MULTI:
+		case SC_CARD_TYPE_DTRUST_V4_1_M100:
+			pin_status(card, DTRUST4_PIN_ID_PIN_CH, "Card Holder PIN", 0);
+			/* FALLTHRU */
```
frankmorgner, line 940: fall through

src/tools/dtrust-tool.c
```diff
@@ -262,50 +857,254 @@ main(int argc, char *argv[])
...
+			pin_status(card, DTRUST5_PIN_ID_PIN_T_AUT, "Transport PIN (Authentication)", 1);
+
+			/* We have to select the eSign app to verify and change the Authentication PIN. */
+			sc_format_path("3F000102", &path);
+			r = sc_select_file(card, &path, NULL);
+			if (r)
+				goto out;
+
+			pin_status(card, DTRUST5_PIN_ID_AUT, "Authentication PIN", 0);
+			/* FALLTHRU */
```
frankmorgner, line 973: fall through is recognized as well and it is gramatically correct

src/tools/dtrust-tool.c
```diff
@@ -262,50 +857,254 @@ main(int argc, char *argv[])
...
+		case SC_CARD_TYPE_DTRUST_V4_4_STD:
+		case SC_CARD_TYPE_DTRUST_V4_4_MULTI:
+			check_transport_protection(card, DTRUST4_PIN_ID_PIN_T, "Signature PIN");
+			break;
+
+		case SC_CARD_TYPE_DTRUST_V5_1_STD:
+		case SC_CARD_TYPE_DTRUST_V5_1_MULTI:
+		case SC_CARD_TYPE_DTRUST_V5_1_M100:
+			check_transport_protection(card, DTRUST5_PIN_ID_PIN_T_AUT, "Authentication PIN");
+			/* FALLTHRU */
```
frankmorgner, line 1009: fall through

src/tools/dtrust-tool.c
```diff
@@ -262,50 +857,254 @@ main(int argc, char *argv[])
...
 
 	if (opt_unlock) {
-		r = check_transport_protection(card);
-		if (r)
-			printf("Cannot remove transport protection.\n");
-		else
-			unlock_transport_protection(card);
+		/* Warn the user he must not abort the unlocking process on
+		 * pin pad readers, as the transport pin has already been used
+		 * and there is not next attempt. */
```
frankmorgner, line 1021: no next attempt?

frankmorgner (changes requested):
The code looks very promising. Apart from the comments above, here are some general suggestions:

- Did you test caching of CAN in Windows Edge/IE? Last time I used this trick (I only have experience with IE), the browser was executed in a temporary user account, which had a different (temporary) home directory, which in turn is used by the PKCS#15 layer for caching files. I'm not sure if this security measure (?) is used by default or if it is triggered by some GPO, nor am I sure whether it is also used for Edge. When I ran into this problem I tied the caching directory to some path that was globally readable/writable. In less menaged environments, you may want to check if a registry entry works better for this purpose. (We already have some PIN related stuff in `HKCU\\Software\\OpenSC Project\\OpenSC\\md_pinpad_dlg_enable_cancel`
- You may want to consider `sc_mem_secure_alloc()` for caching secrets.
- You may want to check if some of the other PINs can be used within the SM channel that was established via CAN. If so, you can possibly avoid re-validating CAN (and save around 1s in execution time)

src/libopensc/card-dtrust.c
hamarituc: These are bit flags and not pointers. Are you sure to favor `NULL` here?

src/tools/dtrust-tool.c
hamarituc: fixed

src/tools/dtrust-tool.c
hamarituc: fixed

src/tools/dtrust-tool.c
hamarituc: fixed

src/tools/dtrust-tool.c
hamarituc: fixed

src/libopensc/card-dtrust.c
hamarituc: To be portable between Linux and Windows I decided to stay with `getenv()` here, but to copy the CAN value into a secure buffer. This also addresses your `sc_mem_secure_alloc()`-remark from above.

src/tools/dtrust-tool.c
hamarituc: This fail-operational pattern is intentional. When verifying the Transport PIN its use counter is set and cannot be used anymore. But the signature PIN still need to be changed. So whatever happens, the user should be prevented from accidentally aborting the program at this stage as it virtually bricks the card.

Nevertheless the user always has the possibility to abort the program by CTRL+C.

hamarituc:
> The code looks very promising. Apart from the comments above, here are some general suggestions:
> 
>     * Did you test caching of CAN in Windows Edge/IE? Last time I used this trick (I only have experience with IE), the browser was executed in a temporary user account, which had a different (temporary) home directory, which in turn is used by the PKCS#15 layer for caching files. I'm not sure if this security measure (?) is used by default or if it is triggered by some GPO, nor am I sure whether it is also used for Edge. When I ran into this problem I tied the caching directory to some path that was globally readable/writable. In less menaged environments, you may want to check if a registry entry works better for this purpose. (We already have some PIN related stuff in `HKCU\\Software\\OpenSC Project\\OpenSC\\md_pinpad_dlg_enable_cancel`

I didn't test with browsers, just document signing with Adobe and the command-line utils, of course. Can you provide a brief description what to test in the browser and how to do it?

>     * You may want to consider `sc_mem_secure_alloc()` for caching secrets.

Fixed.

>     * You may want to check if some of the other PINs can be used within the SM channel that was established via CAN. If so, you can possibly avoid re-validating CAN (and save around 1s in execution time)

I tested it for the PUK, but it could only be validated via PACE authentication not inside an already established PACE channel. I didn't test it for the Transport PIN as verifying the Transport PIN is only possible for one time. But the specifications states that only PACE authentication is supported for PUK and transport PINs and all my tests where consistent with that.

frankmorgner:
> Can you provide a brief description what to test in the browser and how to do it?

It's simple client authentication with a smart card certificate. If you do this with IE (Edge?), you should notice that the user cache files will not be written. Instead, you should find those files in an obscure filename which was created for a temporary user account. 

I believe using the registry as database for a saved CAN should be better in this case. Right now, I think you can leave it as is and consider this suggestion as improvement if you encounter any problems.

Thanks for having a look at the possibilities around using the PACE SM channel. I think there is no easy solution, but if the card is fast enough, it should not be a problem reevaluating some secret.

hamarituc:
> What's the status of this topic, is there anything to do?

From my point of view it's ready to merge. Just saving the CAN into the registry under windows is still an open topic I will solve in an other PR. Getting the driver for 5.1/5.4 cards into the master branch would be very gratifying as I already received the specs for the successor cards (expected to be issued in 2026) and would like start working on it.

frankmorgner:
OK, sounds great!

I'll have a look through the code soon. Please have a look at the build problems listed above.

frankmorgner (approved):

src/libopensc/card-dtrust.c
```diff
@@ -57,18 +84,26 @@ static const struct sc_atr_table dtrust_atrs[] = {
...
-static struct dtrust_supported_ec_curves {
-	struct sc_object_id oid;
-	size_t size;
-} dtrust_curves[] = {
-	{ .oid = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }}, .size = 256 },	/* secp256r1 */
-	{ .oid = {{ -1 }},                            .size =   0 },
-};
+static struct sc_object_id oid_secp256r1 = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }};
+static struct sc_object_id oid_secp384r1 = {{ 1, 3, 132, 0, 34, -1 }};
 // clang-format on
```
Jakuje: does this need to be enclosed with the `clang-format off`?

hamarituc: It's just for whitespace issues.

```
src/libopensc/card-dtrust.c:104:45: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp256r1 = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }};
                                            ^
src/libopensc/card-dtrust.c:104:46: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp256r1 = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }};
                                             ^
src/libopensc/card-dtrust.c:104:76: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp256r1 = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }};
                                                                           ^
src/libopensc/card-dtrust.c:104:78: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp256r1 = {{ 1, 2, 840, 10045, 3, 1, 7, -1 }};
                                                                             ^
src/libopensc/card-dtrust.c:105:45: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp384r1 = {{ 1, 3, 132, 0, 34, -1 }};
                                            ^
src/libopensc/card-dtrust.c:105:46: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp384r1 = {{ 1, 3, 132, 0, 34, -1 }};
                                             ^
src/libopensc/card-dtrust.c:105:67: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp384r1 = {{ 1, 3, 132, 0, 34, -1 }};
                                                                  ^
src/libopensc/card-dtrust.c:105:69: warning: code should be clang-formatted [-Wclang-format-violations]
static struct sc_object_id oid_secp384r1 = {{ 1, 3, 132, 0, 34, -1 }};
```

I you prefer, I can remove the whitespace (and `clang-format` directives).

Jakuje: I would prefer to stick to the clang-format, unless it makes it significantly less readable. In this case, where it is just a whitespace inside of the quotes, I would prefer to use that rather than turning it off and on like this.

hamarituc: done

Jakuje:
Thank you! I think we are good to go. Merging then

PR #3141, opened 2024-05-09
Extend the p11test to support kryoptic

Jakuje:
This mostly fixes some corner case of the p11test that were caught while running it against [kryoptic](https://github.com/latchset/kryoptic), adding new mechanisms and verifying the work as expected (while fixing some bugs on the kryoptic side too).


src/pkcs11/pkcs11-spy.c
```diff
@@ -1020,7 +1038,7 @@ C_Encrypt(CK_SESSION_HANDLE hSession, CK_BYTE_PTR pData, CK_ULONG  ulDataLen,
 	spy_dump_ulong_in("hSession", hSession);
 	spy_dump_string_in("pData[ulDataLen]", pData, ulDataLen);
 	rv = po->C_Encrypt(hSession, pData, ulDataLen, pEncryptedData, pulEncryptedDataLen);
-	if (rv == CKR_OK)
+	if (rv == CKR_OK || rv == CKR_BUFFER_TOO_SMALL)
```
frankmorgner, line 1041: If I understand correctly, CKR_BUFFER_TOO_SMALL may set the output length, but do you really think that in this case the output buffer should be filled by the module so that we dump it here?

simo5: if CKR_BUFFER_TOO_SMALL is returned then pEncryptedData may have any random data in it, C_Encrypt will probably not touch it at all.

src/pkcs11/pkcs11-spy.c
```diff
@@ -1462,7 +1501,7 @@ C_WrapKey(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechanism,
 	spy_dump_ulong_in("hWrappingKey", hWrappingKey);
 	spy_dump_ulong_in("hKey", hKey);
 	rv = po->C_WrapKey(hSession, pMechanism, hWrappingKey, hKey, pWrappedKey, pulWrappedKeyLen);
-	if (rv == CKR_OK)
+	if (rv == CKR_OK || rv == CKR_BUFFER_TOO_SMALL)
```
frankmorgner, line 1504: here is one instance that you missed

frankmorgner (changes requested):

src/pkcs11/pkcs11-spy.c
Jakuje: Thanks . Changed also this one (and rebased on master to fix Fedora CI).

frankmorgner (approved):

xhanulik (approved):
lgtm

Jakuje:
Merging. Thank you for the reviews!

PR #3142, opened 2024-05-10
Use zlib-ng for Fedora

Jakuje:
It should be ABI compatible with the old zlib

https://fedoraproject.org/wiki/Changes/ZlibNGTransition

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/setup-fedora.sh
+++ b/.github/setup-fedora.sh
@@ -13,9 +13,9 @@ fi
 
 # 64bit or 32bit dependencies
 if [ "$1" == "ix86" ]; then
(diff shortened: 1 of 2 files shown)


frankmorgner:
If zlib-ng is THE default within fedora nowadays, I am OK with switching. Does it make sense to do this switch only for recent versions of fedora?

Jakuje:
Correct. The older versions of Fedora do not need this change as the old zlib is still available in Fedora 39 (and we will likely not rebuild it):

https://src.fedoraproject.org/rpms/zlib

Actually the Fedora 40 has the compat package with the "old" name so technically this change is not needed as for now, but we saw the failure on 32b arch, where the resolution was probably not working correctly and switching to -ng solve it.

frankmorgner (approved):

Jakuje:
Merging then to make the CI green again. Thanks for reviews.

PR #3146, opened 2024-05-13
Changes to properly identify IDPrime MD 840

### LINKED ISSUE #3106: Card misidentification
Running OpenSC 0.25.1:
Got two cards:
ATR: 3b:7f:96:00:00:80:31:80:65:b0:85:59:56:fb:12:0f:fe:82:90:00
and
ATR: 3b:7f:96:00:00:80:31:80:65:b0:85:03:00:ef:12:0f:fe:82:90:00

Both identify as Gemalto IDPrime 940.

Code

msetina:
Previously MD 840 was identified as 940.
Also changed the part where mechanism are assigned to the card, so that EC is also added as by spec.

Fixes #3106


msetina:
Did you change code style lately? I followed the file.... It wants different indents.

Jakuje:
No, the clang-format does not capture all the corner cases. Feel free to ignore the proposa of changes to the array of structs.

The long condition should be wrapped but the line needs to end with the logical operator, not start with it and the continuation should be 2 tabs.

Jakuje (approved):
tested with the bunch of cards I have and did not notice any obvious issue. Thank you. Merging

PR #3148, opened 2024-05-15
Remove legacy EstEID codepaths

metsma:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-mcrd.c
+++ b/src/libopensc/card-mcrd.c
@@ -102,24 +102,6 @@ struct mcrd_priv_data {
 
 #define DRVDATA(card) ((struct mcrd_priv_data *) ((card)->drv_data))
 
-// Control Reference Template Tag for Key Agreement (ISO 7816-4:2013 Table 54)
-static const struct sc_asn1_entry c_asn1_control[] = {
-	{ "control", SC_ASN1_STRUCT, SC_ASN1_CONS | SC_ASN1_CTX | 0xA6, 0, NULL, NULL },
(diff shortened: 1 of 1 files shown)


martinpaljak:
I'm pretty sure that the whole mcrd driver could be removed in that case, I don't recall any other active users for it and I doubt new ones have emerged. If the rationale is "unused/maintained old card, lets remove". Yet this specific path (decryption of old documents) might be a reason to keep it around (so that things could be decrypted without manually having to build older versions)

metsma:
> I'm pretty sure that the whole mcrd driver could be removed in that case, I don't recall any other active users for it and I doubt new ones have emerged. If the rationale is "unused/maintained old card, lets remove". Yet this specific path (decryption of old documents) might be a reason to keep it around (so that things could be decrypted without manually having to build older versions)

The only use case in this codepath was for EstEID old cards and this support has been already removed. This is followup cleanup commit.
I cannot comment the other driver support.

frankmorgner (approved):
thank you.

removal of mcrd can be planned as described here:
https://github.com/OpenSC/OpenSC/wiki/Removal-of-Old-Card-Drivers

Jakuje (approved):

xhanulik:
As the EstEID support was removed, I updated https://github.com/OpenSC/OpenSC/wiki/Removal-of-Old-Card-Drivers and since there was not any recent user activity either, we can deactivate the driver. I will do it in separate PR.

xhanulik (approved):

PR #3153, opened 2024-05-23
Fixes for issues found by static analysis

xhanulik:
Fixes for following bugs:
1. _integer overflow_ in `piv_put_data()`, when return value of `sc_asn1_put_tag()` function of type `int` was stored into variable of type `size_t`
2. _buffer overflow_ in `iso7816_set_security_env()`, when `file_ref.len` is checked against `0xFF` instead of `SC_MAX_PATH_SIZE`
3. _buffer overflow_ in `iso7816_set_security_env()`, when `key_ref_len` is checked against `0xFF` instead of `8` (`SC_MAX_KEYREF_SIZE`)

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -3408,11 +3408,11 @@ piv_put_data(sc_card_t *card, int tag, const u8 *buf, size_t buf_len)
 	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
 
 	tag_len = piv_objects[tag].tag_len;
-	sbuflen = sc_asn1_put_tag(0x5c, piv_objects[tag].tag_value, tag_len, NULL, 0, NULL);
-	if (sbuflen <= 0) {
+	r = sc_asn1_put_tag(0x5c, piv_objects[tag].tag_value, tag_len, NULL, 0, NULL);
+	if (r <= 0) {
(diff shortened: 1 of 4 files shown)


src/libopensc/card-piv.c
```diff
@@ -3408,10 +3408,11 @@ piv_put_data(sc_card_t *card, int tag, const u8 *buf, size_t buf_len)
...
 
 	tag_len = piv_objects[tag].tag_len;
-	sbuflen = sc_asn1_put_tag(0x5c, piv_objects[tag].tag_value, tag_len, NULL, 0, NULL);
-	if (sbuflen <= 0) {
+	r = sc_asn1_put_tag(0x5c, piv_objects[tag].tag_value, tag_len, NULL, 0, NULL);
+	if (r <= 0) {
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
 	}
+	sbuflen = r;
 	sbuflen += buf_len;
```
Jakuje, line 3416: would it make sense to squeeze these two into something like this to make code less complicated?
```suggestion
	sbuflen = r + buf_len;
```

Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you! Merging.

PR #3158, opened 2024-05-25
piv-tool: Fix RSA key generation

### LINKED ISSUE #3156: Unable to generate RSA key using piv-tool
### Problem Description

Unable to generate RSA key using piv-tool since version 0.23.0.
Works with OpenSC version 0.22.0 on Ubuntu 22.04.
ECC key also works.

Tested with OpenSC versions:
|Distro|Version of OpenSC|Able to generate RSA key|Note|
|-|-|-|-|
|OpenSUSE Tumbleweed|0.25.1(0.25.1-1.1)|N||
|Ubuntu 24.04|0.25.0(0.25.0~rc1-1build2)|N||
|OpenSUSE Tumbleweed|0.24.0|N|First encountered this issue, this is the old version number I remembered. Could be inaccruate.|
|Debian 12.5|0.23.0(0.23.0-0.3+deb12u1)|N|with `-A M:9B:0C` to avoid admin error|
|Ubuntu 22.04|0.22.0(0.22.0-1ubuntu2)|Y|with `-A M:9B:0C` to avoid admin error|

Tested with:

- NXP J3H145 with [arekinath/PivApplet v0.9.0](https://github.com/arekinath/PivApplet/releases/tag/v0.9.0) with AES256 9B key
- NXP J3R180 with [makinako/OpenFIPS201 v1.10.2](https://github.com/makinako/OpenFIPS201/releases/tag/v1.10.2) with AES256 9B key
- YubiKey 5 NFC with default 3DES 9B key (only tested with 0.25.1 on Tumbleweed)

### Steps to reproduce

1. Setup keys through `PIV_EXT_AUTH_KEY`
2. `piv-tool -v -A A:9B:0C -G 9E:07`

### Logs

```
$ piv-tool -v -A A:9B:0C -G 9E:07
Using reader with a card: Generic EMV Smartcard Reader (ÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿ) 00 00
Connecting to card in reader Generic EMV Smartcard Reader (ÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿ) 00 00...
Using card driver Personal Identity Verification Card.
gen_key unable to gen RSA
```

invis-z:
`params = OSSL_PARAM_BLD_to_param(bld)` was called twice, once inside `if` condition and once after the `if`.  Removed the second one to resolve issue generating RSA keys with OpenSSL 3.0.

Someone else please verify this is the correct fix, I am far from understanding all the code and only did some preliminary tests to check if it is working after the fix. 

Also, maybe consider adding a test for `piv-tool` inside the piv test routine (with PivApplet)?

Fixes #3156

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/piv-tool.c
+++ b/src/tools/piv-tool.c
@@ -423,7 +423,7 @@ static int gen_key(const char * key_info)
 			BN_free(newkey_e);
 			return -1;
 		}
-		params = OSSL_PARAM_BLD_to_param(bld);
+
 		BN_free(newkey_n);
 		BN_free(newkey_e);
 


dengert (approved):

dengert:
Tested using Yubikey 4 with PIV applet with OpenSC master and OpenSSL 3.2.1 on Ubuntu 22.04.

Jakuje (approved):
good catch! Thanks!

You mentioned you tested this with Yubikey, does the piv-took key generation work also with the applet in CI, so we could use the piv-tool for enrollment instead of (or in addition to) the `yubico-piv-tool`?

dengert:
I did use my enrolment scripts which call `piv-tool` to test this patch. 

The OpenSC piv-tool could be used, but many of the operations such as deleting a cert or setting a pin are done using `-s "apdu"` option. Creating some objects such as a CHUID are done via a separate program which has not been generally published. 

(There used to be a WIKI pages with all of this, but I can't find it. Any way to find deleted WIKI pages?) 

The piv-tool when generating a key saves the public key to a file and scripts must be used to create a certificate either self signed or via a certificate request signed using libp11.  if `pkcs15-piv.c` does not find a certificate on the card it will look for an environment variable pointing at the saved public key. https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c#L1023-L1072

Yubikey piv tool when generating a key writes a dummy certificate to the card as part of the key gen process, which simplifies the process. 

One main difference between Yubikey piv tool and piv-tool  is piv-tool expects objects to be written (other then a certificate) to be contained in the  `7E` or `5C` tag and length. Yubikey  piv tool wraps the input in `7E` or `5C`. 

piv-tool can take a certificate and optionally gzip it and place in a certificate object and write the certificate object to the card.

The piv-tool was written in 2004 to do the minimal needed to write to test cards, not to be a general support tool, as PIV cards at that time were only issued by government agencies and card vendors provided the card management systems. 

I have no real interest it trying to make any substantial changes to piv-tool.     

I did provide my enrolment scripts to my employer at the time to be used to generate temporary PIV cards, for new hires while waiting for a gov issued card, or a card that broke, was lost, or was left at home. The PIV cards tmp or gov issued were used to authenticate to Active Directory, so these people could still get some work done.

dengert:
One way to  generate a certificate or certificate request is by using PKCS11. The CKA_ID of the certificate, public key and private key are the same. In my test scripts I used the `openssl` ,  `CA.pl`  and `openssl.conf`  with a pkcs11 engine or provider.
Since the PIV card did not have a certificate, the public key was passed by the env variable as outlined in:  https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c#L1023-L1072 

The test scripts would make a copy of `openssl.conf` and using `sed` make changes based on the username, email, which CKA_ID and other extensions to sign a CSR which was then sent to an AD CA  or my test CA which used  `CA.pl` to call `openssl` to sign the CSR and create a certificate using a test CA certificate.   The certificate was then loaded onto the card using piv-tool. 
All of this was run on Linux.

Jakuje:
> (There used to be a WIKI pages with all of this, but I can't find it. Any way to find deleted WIKI pages?)

The whole wiki pages history should be here: https://github.com/OpenSC/Wiki. Searching through git I managed to find
`PivTool.textile` which was destroyed with https://github.com/OpenSC/Wiki/commit/58193e34ca74f50211732cabfbfe6bce9fa985fb

I vaguely remember we discussed that this was more confusing users than helping as it could not do all the operations the Yubico piv tool could do, but the commit message is not very descriptive.

> So, if I am understanding correctly, since there is no deletion needed in test-piv.sh, the following code provides the same functionality in the test (if openssl is already configured to use pkcs11 provider):

Thanks! This looks reasonable. But if I am right, the pkcs11-provider is not yet in standard ubuntu repositories (especially not in the old images in GH actions) so we would need to build it as part of the CI (instead of the `yubico-piv-tool`).

The only thing I found is ppc64el version in universe: https://packages.ubuntu.com/noble/ppc64el/libs/pkcs11-provider

I think this does not have to happen now in this PR, but at very least, we should create an issue with the notes from here as they are certainly helpful for writing the test.

xhanulik (approved):
Please fix the spelling errors.

frankmorgner (approved):

Jakuje:
Created separate issue for the tests #3181 and merging this not to delay the fix even mode.

PR #3161, opened 2024-05-28
 log: Null terminate the errors from OpenSSL 

Jakuje:
Should resolve the following oss-fuzz issue:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67807

##### Checklist
- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/log.c
+++ b/src/libopensc/log.c
@@ -75,7 +75,7 @@ void sc_do_log_color(sc_context_t *ctx, int level, const char *file, int line, c
 void sc_do_log_openssl(sc_context_t *ctx, int level, const char *file, int line, const char *func)
 {
 	BIO *bio = NULL;
-	int length = 0;
+	int length, rc;
 	char *buffer = NULL;
 
 	if ((bio = BIO_new(BIO_s_mem())) == NULL) {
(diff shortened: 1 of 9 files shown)


xhanulik (approved):

src/libopensc/log.c
```diff
@@ -84,10 +84,21 @@ void sc_do_log_openssl(sc_context_t *ctx, int level, const char *file, int line,
...
+	/* trailing null byte */
+	buffer = malloc(length + 1);
+	if (buffer == NULL) {
+		sc_do_log(ctx, level, file, line, func, "No memory!");
+		goto end;
+	}
+	rc = BIO_read(bio, buffer, length);
+	buffer[length] = '\n';
+	if (rc <= 0) {
+		sc_do_log(ctx, level, file, line, func, "Cannot read OpenSSL error");
```
frankmorgner, line 101: I think it would be useful to keep the old error message "Cannot log OpenSSL error"

src/libopensc/log.c
```diff
@@ -84,10 +84,21 @@ void sc_do_log_openssl(sc_context_t *ctx, int level, const char *file, int line,
...
+		goto end;
+	}
+	/* trailing null byte */
+	buffer = malloc(length + 1);
+	if (buffer == NULL) {
+		sc_do_log(ctx, level, file, line, func, "No memory!");
+		goto end;
+	}
+	rc = BIO_read(bio, buffer, length);
+	buffer[length] = '\n';
```
frankmorgner, line 99: This is not NUL terminated, this is a newline!

frankmorgner (changes requested):

src/libopensc/log.c
Jakuje: :facepalm: will fix!

src/libopensc/log.c
Jakuje: The thing was that the same log message was here for different things and I find it useful when they are as specific as possible so one does not have to go through the code and line offsets to figure out which one of these same messages failed. The old message `Cannot log OpenSSL error` is still on line 82 (which I would consider to rewrite also to something about memory issues).

frankmorgner: thanks for the explanation 👍

frankmorgner (approved):

Jakuje:
Thanks for reviews! Merging to get rid of the spellcheck and fuzzing issues.

PR #3167, opened 2024-06-01
minidriver.c pkcs15-piv.c - Support PinCacheAlwaysPrompt

### LINKED ISSUE #3159: OpenSC Minidriver with PIVApplet + ECC keys on Win11: error on slot 9c - public key does not match private key
Hi there,

### Problem Description

after using @vletoux GIDS applet for quite a while (thanks :)!) I wanted to migrate to ECC and PIV because I need integration with
the OS - such as logon, VPN, TLS, SSH. Unfortunately I could only manage to get a Yubikey running with Active Directory CA
credentials in a manual process with a custom CA template - so I could verify AD and client are prepared to handle at least ECCP256 and do workstation logon. (enrollment for ECC doesn't really seem to work anyway ...?)
I am using the Yubico piv-tool below because I need to generate CSRs and the OpenSC toolchain lacks functionality for that.

So I chose https://github.com/arekinath/PivApplet and compiled it with the settings:

```
<property name="PIV_SUPPORT_RSA" value="false"/>
<property name="PIV_SUPPORT_EC" value="true"/>
<property name="PIV_SUPPORT_ECCP384" value="false"/>
<property name="PIV_SUPPORT_AES" value="true"/>
<property name="PIV_SUPPORT_3DES" value="false"/>
<property name="PIV_STRICT_CONTACTLESS" value="false"/>
<property name="YKPIV_ATTESTATION" value="true"/>
<property name="APPLET_EXTLEN" value="true"/>
<property name="APPLET_USE_RESET_MEM" value="false"/>
<property name="APPLET_LOW_TRANSIENT" value="false"/>
```

put it on a NXP P71D321 / J3R180 variant:
`...\tools> .\gp.exe -dvil -install C:\Users\ck\git\PivApplet\bin\PivApplet.cap
`

check the status
```
...\PIV>piv-init
STATUS:
Version:        5.4.0
Serial Number:  -1654144648
CHUID:  30190000000000000000000000000000000000000000000000000034105db884ca12cbf24fe16120aa25809fa5350832303530303130313e00fe00
CCC:    f015a000000116ff0204345082ddf1aea2203f92e803b8f10121f20121f300f400f50110f600f700fa00fb00fc00fd00fe00
PIN tries left: 5
```

initialize the attestation, card authentication and key management keys
```
...\PIV>piv-init initcard
Successfully generated a new private key.
Successfully generated a new private key.
Successfully verified PIN.
Successfully generated a new self signed certificate.
Successfully verified PIN.
Successfully imported a new certificate.
Successfully verified PIN.
Successful ECDSA verification.
Successfully generated a new private key.
Successfully verified PIN.
Successfully generated a new self signed certificate.
Successfully verified PIN.
Successfully imported a new certificate.
Successfully verified PIN.
Successful ECDSA verification.
```

generate keypairs for PIV authentication and digital signature (testcase here, simulating the exact same problem with real CA)
```
...\PIV>piv-init selfsign
Successfully generated a new private key.
Successfully verified PIN.
Successfully generated a new self signed certificate.
Successfully verified PIN.
Successfully imported a new certificate.
Successfully verified PIN.
Successful ECDSA verification.
Successfully generated a new private key.
Successfully verified PIN.
Successfully generated a new self signed certificate.
Successfully verified PIN.
Successfully imported a new certificate.
Successfully verified PIN.
Successful ECDSA verification.
Successfully verified PIN.
Successfully set new CCC.
Successfully set new CHUID.
```

check the results and getting errors, resembling what I observed before:

```
...\PIV>certutil -scinfo -v -privatekey
Die Microsoft Smartcard-Ressourcenverwaltung wird ausgeführt.
...
Karte im Leser wird analysiert: ACS ACR1255U-J1 PICC Reader 0
Microsoft Base Smart Card Crypto Provider: Gespeicherter Schlüsselsatz fehlt
...
ERROR: Der öffentliche Schlüssel des Zertifikats konnte nicht gegen den privaten Schlüssel überprüft werden.
Microsoft Smart Card Key Storage Provider: KeySpec=0
AES256+RSAES_OAEP(ECC:CNG) Test übersprungen
....
--------------===========================--------------
CertUtil: -SCInfo-Befehl ist fehlgeschlagen: 0x80090016 (-2146893802 NTE_BAD_KEYSET)
CertUtil: Der Schlüsselsatz ist nicht vorhanden.
```

The certificates show up nicely:

image

I tracked the issue down to the second key slot, being assigned to the 9c ("digital signature") slot in the PIV applet.

I'm attaching all I've got from the logs.

### Proposed Resolution

Slot 9c needs to work to do smartcard logon - while TLS client authenticaton may work with slot 9a many other things (VPN, RDP) do not.

### Steps to reproduce

For piv-init initcard I do:

```
:initcard
%PIVTOOL% -r %READER% -agenerate -s f9 -A%ALG% > att-key.pem

REM card authentication key
%PIVTOOL% -r %READER% -agenerate -s 9e -A%ALG% > tmp-key.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aselfsign-certificate -s 9e -S"/CN=%CN%-CA/" < tmp-key.pem > tmp-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aimport-certificate -s 9e < tmp-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -atest-signature -s 9e < tmp-cert.pem

REM key management key
%PIVTOOL% -r %READER% -agenerate -s 9d -A%ALG% > tmp-key.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aselfsign-certificate -s 9d -S"/CN=%CN%-KM/" < tmp-key.pem > tmp-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aimport-certificate -s 9d < tmp-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -atest-signature -s 9d < tmp-cert.pem
goto status
```

For piv-init selfsign I do:

```
:selfsign
REM PIV authentication key
%PIVTOOL% -r %READER% -agenerate -s 9a -A%ALG% > auth-key.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aselfsign-certificate -s 9a -S"/CN=%CN%-PA/" < auth-key.pem > auth-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aimport-certificate -s 9a < auth-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -atest-signature -s 9a < auth-cert.pem

REM digital signature key 
%PIVTOOL% -r %READER% -agenerate -s 9c -A%ALG% > sign-key.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aselfsign-certificate -s 9c -S"/CN=%CN%-DS/" < sign-key.pem > sign-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -aimport-certificate -s 9c < sign-cert.pem
%PIVTOOL% -r %READER% -P %PIN% -averify -atest-signature -s 9c < sign-cert.pem

REM init card capabilities container and UID
%PIVTOOL% -r %READER% -P %PIN% -averify -aset-ccc -aset-chuid
goto status

```

### Logs

see https://gist.github.com/ckahlo/bc00b88d911364a584069dd237601967 for scinfo -v -privatekey

see opensc-debug.txt attached

[opensc-debug.txt](https://github.com/OpenSC/OpenSC/files/15448844/opensc-debug.txt)

Thank you!

dengert:
(WIP may fix #3159 waiting to download the MSI artifacts for testing.)

PKCS11 supports CKA_ALWAYS_AUTHENTICATE and PKCS15 user_consent Windows minidriver supports PinCacheAlwaysPrompt

Mindriver has MD_ROLE_USER_SIGN for a pin which is taken to be a second user local pin to be used for signing.

A second user local pin was defined in pkcs15-piv.c which is a duplicate of the user pin except  the sc_pkcs15_auth_info_auth_method set is set SC_AC_CONTEXT_SPECIFIC So when a key is used that requires "ALWAYS" authentication it will be be handled like framework-pkcs15.c handles CKA_ALWAYS_AUTHENTICATE

 On branch minidriver-PinCacheAlwaysPrompt
 Changes to be committed:
	modified:   libopensc/pkcs15-piv.c
	modified:   minidriver/minidriver.c

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-piv.c
+++ b/src/libopensc/pkcs15-piv.c
@@ -71,6 +71,7 @@ typedef struct pdata_st {
 	int         tries_left;
 	const unsigned char  pad_char;
 	int         obj_flags;
+	int         cardmod; /* only use with cardmod minidriver */
 } pindata;
 
 typedef struct pubdata_st {
@@ -385,17 +386,30 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		  SC_PKCS15_PIN_FLAG_INITIALIZED |
 		  SC_PKCS15_PIN_FLAG_LOCAL,
 		  -1, 0xFF,
-		  SC_PKCS15_CO_FLAG_PRIVATE },
+		  SC_PKCS15_CO_FLAG_PRIVATE, 0},
+
 		{ "02", "PIV PUK", "", 0x81,
 		  SC_PKCS15_PIN_TYPE_ASCII_NUMERIC,
 		  8, 4, 8,
 		  SC_PKCS15_PIN_FLAG_NEEDS_PADDING |
 		  SC_PKCS15_PIN_FLAG_INITIALIZED |
-		  SC_PKCS15_PIN_FLAG_LOCAL | SC_PKCS15_PIN_FLAG_SO_PIN |
+		  SC_PKCS15_PIN_FLAG_SO_PIN |
 		  SC_PKCS15_PIN_FLAG_UNBLOCKING_PIN,
 		  -1, 0xFF,
-		  SC_PKCS15_CO_FLAG_PRIVATE },
-		{ NULL, NULL, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+		  SC_PKCS15_CO_FLAG_PRIVATE, 0},
+
+		/* only used with minidriver */
+		{ "03", "PIN", "", 0x80,
+		  /* used in minidriver as the sign key and for 9C key */
+		  /* label, flag  and ref will change if using global pin */
+		  SC_PKCS15_PIN_TYPE_ASCII_NUMERIC,
+		  8, 4, 8,
+		  SC_PKCS15_PIN_FLAG_NEEDS_PADDING |
+		  SC_PKCS15_PIN_FLAG_INITIALIZED |
+		  SC_PKCS15_PIN_FLAG_LOCAL,
+		  -1, 0xFF,
+		  SC_PKCS15_CO_FLAG_PRIVATE, 1}, /* only use if cardmod */
+		{ NULL, NULL, NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
 	};
 	// clang-format on
 
@@ -540,7 +554,7 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 					SC_PKCS15_PRKEY_USAGE_NONREPUDIATION,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_SIGN |
 					SC_PKCS15_PRKEY_USAGE_NONREPUDIATION,
-			"", 0x9C, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1},
+			"", 0x9C, "01", SC_PKCS15_CO_FLAG_PRIVATE, 1}, /* use sign pin and user_consent */
 		{ "03", "KEY MAN key",
 				/*RSA*/SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_UNWRAP,
 				/*EC*/SC_PKCS15_PRKEY_USAGE_DERIVE,
@@ -952,6 +966,10 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		const char * label;
 		int pin_ref;
 
+		/* the SignPIN is only used with minidriver */
+		if (pins[i].cardmod && (strcmp(card->ctx->app_name, "cardmod") != 0))
+			continue;
+
 		memset(&pin_info, 0, sizeof(pin_info));
 		memset(&pin_obj,  0, sizeof(pin_obj));
 
@@ -968,17 +986,17 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		sc_format_path(pins[i].path, &pin_info.path);
 
 		label = pins[i].label;
-		if (i == 0 &&
+		if ((i == 0 || pins[i].cardmod) &&
 			sc_card_ctl(card, SC_CARDCTL_PIV_PIN_PREFERENCE,
 					&pin_ref) == 0 &&
 				pin_ref == 0x00) { /* must be 80 for PIV pin, or 00 for Global PIN */
 			pin_info.attrs.pin.reference = pin_ref;
-			pin_info.attrs.pin.flags &= ~SC_PKCS15_PIN_FLAG_LOCAL;
 			label = "Global PIN";
 		}
+
 		strncpy(pin_obj.label, label, SC_PKCS15_MAX_LABEL_SIZE - 1);
 		pin_obj.flags = pins[i].obj_flags;
-		if (i == 0 && pin_info.attrs.pin.reference == 0x80) {
+		if ((i == 0 || pins[i].cardmod)) {
 			/*
 			 * according to description of "RESET RETRY COUNTER"
 			 * command in specs PUK can only unblock PIV PIN
@@ -1015,7 +1033,6 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 		strncpy(pubkey_obj.label, pubkeys[i].label, SC_PKCS15_MAX_LABEL_SIZE - 1);
 
 		pubkey_obj.flags = pubkeys[i].obj_flags;
-
 
 		if (pubkeys[i].auth_id)
 			sc_pkcs15_format_id(pubkeys[i].auth_id, &pubkey_obj.auth_id);
@@ -1168,6 +1185,10 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 
 		if (prkeys[i].auth_id)
 			sc_pkcs15_format_id(prkeys[i].auth_id, &prkey_obj.auth_id);
+
+		/* If using minidriver, use Sign PIN  for 9C key */
+		if (prkey_obj.user_consent && (strcmp(card->ctx->app_name, "cardmod") == 0))
+			sc_pkcs15_format_id("03", &prkey_obj.auth_id);
 
 		/*
 		 * When no cert is present and a pubkey in a file was found,
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -1893,9 +1893,10 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
 
 					/* set flag that at least one key that uses the sign key needs PinCacheAlwaysPrompt */
 					logprintf(pCardData, 7, "key_obj->user_consent: %d\n", (int) key_obj->user_consent);
-					if (key_obj->user_consent)
+					if (key_obj->user_consent) {
 						vs->need_pin_always = 1;
-					logprintf(pCardData, 7, "vs->need_pin_always %d\n", (int) vs->need_pin_always);
+						logprintf(pCardData, 7, "vs->need_pin_always %d\n", (int) vs->need_pin_always);
+					}
 
 					if (pin_mode < pin_mode_n) {
 						pin_mode = pin_mode_n;
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -233,6 +233,9 @@ typedef struct _VENDOR_SPECIFIC
 	struct md_dh_agreement* dh_agreements;
 	BYTE allocatedAgreements;
 
+	/* if any key used with the MD_ROLE_USER_SIGN has user_consent set PinCacheAlwaysPrompt */
+	int need_pin_always;
+
 	CRITICAL_SECTION hScard_lock;
 } VENDOR_SPECIFIC;
 
@@ -1887,6 +1890,12 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
 						  ii,
 						  cont->flags & CONTAINER_MAP_DEFAULT_CONTAINER ?
 						  " (default)" : "");
+
+					/* set flag that at least one key that uses the sign key needs PinCacheAlwaysPrompt */
+					logprintf(pCardData, 7, "key_obj->user_consent: %d\n", (int) key_obj->user_consent);
+					if (key_obj->user_consent)
+						vs->need_pin_always = 1;
+					logprintf(pCardData, 7, "vs->need_pin_always %d\n", (int) vs->need_pin_always);
 
 					if (pin_mode < pin_mode_n) {
 						pin_mode = pin_mode_n;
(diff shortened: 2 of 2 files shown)


dengert:
This should be a fix for #3159  when PIV is used with the minidriver. 
The code in `pkcs15-piv.c` defines a third pin, which is the same as the first pin, accept the `auth_method` is set to  `SC_AC_CONTEXT_SPECIFIC` rather then `SC_AC_CHV`. The 9C key is listed as using the third pin in `pkcs15-piv.c` 

 Mindriver looks for a `user_pin`  which is the default first pin in `pkcs15-piv.c`. It also looks for a `sign_pin` as the second type of `user_pin` and finds the new third pin in `pkcs15-piv.c`  Because it is a "sign_pin" the minidriver sets `PinCacheAlwaysPrompt` for this pin. (This is one area which maybe a problem for other cards.)  

So when an application like `certutil` wants to use the certificate which uses the 9C key, the CSP will always pops up a PIN which says: "Please enter your digital signature pin."  The CSP then passes this PIN to minidriver  to use the new third pin that has the `auth_method = SC_AC_CONTEXT_SPECIFIC;`.

(For PKCS11, framework-pkcs15.c also sets  `auth_method = SC_AC_CONTEXT_SPECIFIC;` so that has been around for some time.)  

The `pkcs15-pin.c` routines pass pin request to the `card-piv.c` which triggers  `card-piv.c` to obtain an additional PC/SC lock so the verify and whatever APDU the application requests next (which should be the crypto instruction)  to be in the same SCardTransaction. The PIV card keep track of the last verify it sees is just before the crypto operation as required by the specs. 

This has been tested on Windows 11 PRO using an IDEMIA ID-One PIV 2.4.1 test card with ECDSA and RSA keys using `certutil -v -scinfo`  And `certutil popup  says: ""Please enter your digital signature pin." and  certutil says: `Private key verifies` for all keys on the card. 

I have been using opesc.conf so as to avoid any OpenSC caching and get PIN debugging: 
```
app default {
	 debug = 9;
	 debug_file = \tmp\opensc-debug.txt;
		framework pkcs15 {
		use_pin_caching = false;
		use_file_caching = false;
	}
}
```
registry was  modified like what   https://github.com/ckahlo did to use the OpenSC minidriver for a PIV with the correct ATR. 

Note that Windows is caching the `user_pin` and used it several times. There was only one use for the 'sign_pin` 

What I have not done is test with others cards, especially with cards that have a `user_pin` and a `sign_pin` I don't expect any problems as then have the same restrictions as the PIV specs have.

src/minidriver/minidriver.c
```diff
@@ -6477,8 +6477,11 @@ DWORD WINAPI CardGetProperty(__in PCARD_DATA pCardData,
 
 				if (dwFlags == ROLE_USER)
 					p->PinPurpose = PrimaryCardPin;
-				else if (dwFlags == MD_ROLE_USER_SIGN)
+				else if (dwFlags == MD_ROLE_USER_SIGN) {
 					p->PinPurpose = DigitalSignaturePin;
+					/* may need to check pin auto_info for auth_method SC_AC_CONTEXT_SPECIFIC */
+					p->PinCachePolicy.PinCachePolicyType = PinCacheAlwaysPrompt;
```
maciejsszmigiero, line 6484: Won't this force PIN re-prompts for every `MD_ROLE_USER_SIGN` PIN, not only the new PIV type?

Minidriver spec says that:
> Note: Windows logon may not work properly if a PIN is not cached. This behavior is by design. 
> Therefore, careful consideration should be given when setting a PIN cache mode to any value other than `PinCacheNormal`.

I guess it really should be a special PKCS15 API PIN flag that says that this PIN needs re-sending before each operation. Or some other way to retrieve this information about PIN from the particular OpenSC card driver.

And then when this OpenSC API PIN flag is set the Minidriver can return `PinCacheAlwaysPrompt` for this PIN.

dengert: > Won't this force PIN re-prompts for every MD_ROLE_USER_SIGN PIN, not only the new PIV type?

Maybe I have taken a different approach with the  force push today: 

       minidriver.c - Create shadow  MD_ROLE_USER_ALWAYS and MD_ROLE_USER_SIGN_ALWAYS
    
       PKCS11 defines CKA_ALWAYS_AUTHENTICATE attribute for private keys, which is converted
       to `user_consent` for PKCS15 for private keys. Windows has `PinCacheAlwaysPrompt`
       on PINS to accomplish the same thing - prompt user before using a key.
    
       But a key "is secured by" only one pin, but a pin may secure multiple keys where only
       a subset of keys need `PinCacheAlwaysPrompt`
    
       Two new pin roles are defined and use a sc_pkcs15_id auth_id of
       auth_id_md_role_user_always = {"MD_ROLE_USER_ALWAYS", 18};
       and
       auth_id_md_role_user_sign_always = {"MD_ROLE_USER_SIGN_ALWAYS", 18};
    
       When building containers for pkcs15 key objects, if user_consent > 0, the auth_id
       is set to one of the above. Thus when Windows CSP selects a key, it will find
       a pin that matches the correct role for the key.

This is still being tested. But should work for any card the has PKCS15 `user_consent > 0` for 
the User key and/or the Sign key.

dengert:
Version 3 of this PR... `minidriver.c`  was modified by bd9cdd243d1c35da2e83be6792565c27c685f17d in 2017 to support multiple pins for a card based on `framework-pkcs15.c`  10e1ad001d from 2012 which tries to create UserPIN, and SignPIN and some others based on pins define in PKCS15 . The concept of Global PIN and Local PIN come into play. 

But none of these tried to address the PKCS11 `CKA_ALWAYS_AUTHENTICATE` attribute of a private key which is true or false, or the PKCS15 `user_consent` on an object which can define the number of times a pin can be used before it needs to be verified again.  

Cards may have multiple pins, which may or may not have a `user_consent` limit. The general logic above assumes if the card supports more then one normal PIN i.e. not a PUK or SO_USER type PIN, then the first normal PIN is the UserPIN also  
referred to as Authentication PIN  and never or rarely has a `user_consent` limit.  

`pkcs15-esteid2018.c`, `pkcs15-idprime.c`, `pkcs15-jpki.c`, `pkcs15-skeid.c` and  `pkcs15-piv.c` all set `user_consent`  for some key which `is secured by`  the second pin i.e. SignPIN which may have a different value the the UserPIN.   

The PIV card has just one PIN but if the 9C key is used the verify ADPU must immediately proceed the crytpo ADPU. So it needs `CKA_ALWAYS_AUTHENTICATE` and `user_consent` to signal the application to  prompt for PIN. With Windows that is the  
`PinCacheAlwaysPrompt` flag of a PIN.

`pkcs15-piv.c` defined 2 pins, the normal user pin and a PUK/SO_USER pin, but also set the `SC_PKCS15_PIN_FLAG_LOCAL`
flag on the PUK pin, which caused the `minidriver` to assume it was a SignPIN which is a bug.

In this version of the PR:

-  `pkcs15-piv.c` now defines 3 pins, the UserPIN, a SignPIN and a PUK pin.   

- `minidriver.c` while looking at keys and which PIN "is secured by" the SignPIN, will test if `user_consent > 0`.and set 
`vs->need_pin_alway`. Then when CSP picks a key to use, which `is secured by` the SignPIN i.e. `MD_ROLE_USER_SIGN`  the `minidriver.c` will set `PinCacheAlwaysPrompt` causing the CSP to prompt the user for the PIN.  

This should also work for the other 4 cards, but they should be tested. 

"is secured by" is used in the debug logs with level 7. 

@maciejsszmigiero  can you have a look at this too?

Should fix #3156 and #3169

dengert:
The PR has been rebased to a single commit, and tested with two PIV type cards with:

-  `certutil -scinfo` on Windows 11 Pro
-  `pkcs15-tool`  on Windows with various list options
-  `pkcs11-tool --test --login` on Windows and Ubuntu

Card drivers that do not use `user_consent` are not affected.   `PinCacheAlwaysPrompt` is only set on the `Sign PIN` if there is a key that requires `user_consent`. There are 5 drivers that use `user_consent`:

- `pkcs15-piv.c` when not being called from minidriver sitll defines a `User PIN` or `Global PIN` (based on the cards preference for `Global` or `Local`) and a `PUK PIN`.  When `ctx->app_name == "cardmod"` i.e. `minidriver` a `Sign PIN` is defined to be used only with the 9C key so Windows  `PinCacheAlwaysPrompt` can be set on the `Sign PIN`.
- `pkcs15-esteid.c` defines a `User PIN` and `Sign PIN` but no `PUK`. 
- `pkcs15-idprime.c` defines a `User PIN` and for 940 type cards a `Sign PIN`. 
- `pkcs15-dnie.c`  defines a `User PIN` but has its own `dnie_ask_user_consent` routine, so should not be affected by this PR.
- `pkcs15-jpki.c`  defines a `User PIN` and  `Sign PIN` but there is still the problem that the the public key is private and needs to be read and placed in the `CCP_CONTAINER_INFO` before the key is selected and PIN verified to allow the public key to be read. 

So this PR will be useful for the 5 drivers.  But registry entries for OpenSC with PIV are not installed be default and the JPKI still has problems. No bug reports (that I know of) with `pkcs15-esteid.c` or `pkcs15-idprime.c` may indicate a lack of use of these drivers with Windows.

src/libopensc/pkcs15-piv.c
```diff
@@ -968,17 +986,17 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
...
 				pin_ref == 0x00) { /* must be 80 for PIV pin, or 00 for Global PIN */
 			pin_info.attrs.pin.reference = pin_ref;
-			pin_info.attrs.pin.flags &= ~SC_PKCS15_PIN_FLAG_LOCAL;
 			label = "Global PIN";
 		}
+
 		strncpy(pin_obj.label, label, SC_PKCS15_MAX_LABEL_SIZE - 1);
 		pin_obj.flags = pins[i].obj_flags;
-		if (i == 0 && pin_info.attrs.pin.reference == 0x80) {
+		if ((i == 0 || pins[i].cardmod)) {
```
maciejsszmigiero, line 999: This seems to change the logic condition from (`i == 0` **AND** `pin.reference == 0x80`) to (`i == 0` **OR** `cardmod`).
Is this change intentional?

src/minidriver/minidriver.c
```diff
@@ -1888,6 +1891,12 @@ md_set_cmapfile(PCARD_DATA pCardData, struct md_file *file)
 						  cont->flags & CONTAINER_MAP_DEFAULT_CONTAINER ?
 						  " (default)" : "");
 
+					/* set flag that at least one key that uses the sign key needs PinCacheAlwaysPrompt */
+					logprintf(pCardData, 7, "key_obj->user_consent: %d\n", (int) key_obj->user_consent);
+					if (key_obj->user_consent)
+						vs->need_pin_always = 1;
+					logprintf(pCardData, 7, "vs->need_pin_always %d\n", (int) vs->need_pin_always);
```
maciejsszmigiero, line 1898: I guess this should print `need_pin_always` only when it is being set (inside that `if` block)?

This value can be deducted from the above `user_consent` printout anyway.

src/libopensc/pkcs15-piv.c
dengert: It logic may need some changes.  The intent is when run by minidriver, the pins[i].cardmod is set on pin[2] which will be treated as the Sign PIN by minisriver with the same pin reference as PIN[0].  With PKCS11 or any application other than minidriver, lines 969-971 will cut the loop short and only two pins are defined: pin[0] as the auth pin, and pin[1] as the PUK PIN/Admin PIN. 

`((i == 0 || pins[i].cardmod)) {`  is really `if (i == 0 || i == 2)` set the auth_id of the PUK/ADMIN PIN on the other pin(s).
sp800-73-X (all version) only defines one PIN which when used with the 9C key, must be verified just before the APDU (00 87 XX 9C) crypto operation.       

sp800-73-X also allows for a Global PIN  with reference 00 or a local pin with reference 80 to be use for pin[0] It defines which is preferred or which is the only one to use from the DISCOVERY object.  (The prompt to the user
will say USER PIN or Global PIN, just incase they are different.   

NIST says only the local PUK PIN (key_ref 81) can be used by the user to unlock/change the pin. Some PIV like cards may allow a Global PUK/Admin pin (key_ref 01) to be used as a PUK. This restriction was relaxed by removing `pin_info.attrs.pin.reference == 0x80`   The card should enforce this.    

The point is when not run with minidriver, pkcs15 tools and routines will list only two pins. Only the minidriver sees a copy of the pin[0] as pin[2] so minidriver uses pin[2] when the 9C key is being used: `usrr_consent > 0`. The minidriver tests for `need_pin_always`.

src/minidriver/minidriver.c
dengert: I left the `logprintf` outside to show value of `need_pin_always` after the test.
the `need_pin_always` could be an array for each ROLE, if there is ever a third type of Sign Pin.

Jakuje (approved):
What is the status of this PR? I do not have a way to test the windows changes, but the rest looks reasonable and I think we should get it merged for the next release.

dengert:
I thing it is ready. And I tested it on Windows 11. using PIV card. Note that the registry entries for PIV  are not normally added, because most vendors provide their own PIV minidriver via plug-and-play and don't need OpenSC.  But there are people loading piv applets on different cards  which is the case in #3159

Jakuje:
Thank you! Merging then.

PR #3171, opened 2024-06-03
dtrust-tool: fix memory leak

hamarituc:
If `qespin1` and `qespin2` doesn't match, `qespin2` was neither cleared nor freed;

##### Checklist

Just a minor fix. No items apply.

### DIFF
--- a/src/tools/dtrust-tool.c
+++ b/src/tools/dtrust-tool.c
@@ -167,19 +167,19 @@ unlock_transport_protection(sc_card_t *card)
 		r = util_getpass(&qespin1, &qespin1_len, stdin);
 		if (r < 0 || qespin1 == NULL) {
 			fprintf(stderr, "Unable to get PIN");
-			goto fail_qespin1;
+			goto fail;
 		}
 
 		printf("Enter new Signature PIN again:");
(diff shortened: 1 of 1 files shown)


src/tools/dtrust-tool.c
```diff
@@ -196,6 +196,7 @@ unlock_transport_protection(sc_card_t *card)
 	else
 		printf("Can't change pin: %s\n", sc_strerror(r));
 
+fail_qespincmp:
```
Jakuje, line 199: I am not a huge fan of the huge amount of `goto`/`labels` for different stages of fail. The code looks ok, but I would prefer to have only one label `fail`.
Given that the variables are always compared with NULL and they are initialized to NULL when declared it is safe to run anything from here below at any failure case, isn't it?

So my preference would be to define here only one label `fail`, remove the others and change all `goto`s to `fail`. I think that should not make any difference in behavior, but the code would be much easier to understand.

Jakuje (approved):

Jakuje (approved):
Thanks! Looks good!

frankmorgner (approved):

xhanulik (approved):
Thank you, merging

PR #3172, opened 2024-06-05
[0.25] pkcs15-crypt: Use correct PKCS#1 v1.5 padding type

kunkku:
This PR backports a fix from `master` to the `stable-0.25` branch, making `pkcs15-crypt --sign` work.


frankmorgner:
we only support the latest version (which includes the requested change). if you require any backports, the please refer to your distribution support

PR #3174, opened 2024-06-11
feat(pkcs11-tool): don't limit object size to 5000 bytes

AlexandreGonzalo:
Hi,
The goal of this PR is to remove the limit of 5000 bytes when an object is imported.
In the current version, the object size is silently truncated to 5000 bytes.
Objects such as certificates or CKO_DATA can be larger than 5000 bytes and I suggest that we should dynamically allocate the buffers instead of using a hard coded 5000 bytes buffer.
I checked the patch with the Trustonic TEE HSM.
Regards,
Alexandre.


src/tools/pkcs11-tool.c
```diff
@@ -4575,6 +4594,10 @@ static CK_RV write_object(CK_SESSION_HANDLE session)
 	EVP_PKEY_free(evp_key);
 #endif /* ENABLE_OPENSSL */
 
+	if (contents)
+		free(contents);
+	if (certdata)
+		free(certdata);
```
Jakuje, line 4600: nit: No need to check for null here. If it is null, the `free()` is no-op anyway (and if it was freed or is uninitialized you are screwed anyway).
```suggestion
	free(contents);
	free(certdata);
```

src/tools/pkcs11-tool.c
```diff
@@ -4050,25 +4048,43 @@ static CK_RV write_object(CK_SESSION_HANDLE session)
...
+		util_fatal("Couldn't set file position to the end of the file \"%s\"", opt_file_to_write);
+	contents_len = ftell(f);
 	if (contents_len < 0)
+		util_fatal("Couldn't get file position \"%s\"", opt_file_to_write);
+	contents = malloc(contents_len + 1);
+	if (contents == NULL)
+		util_fatal("malloc() failure\n");
+	if (fseek(f, 0L, SEEK_SET) != 0)
+		util_fatal("Couldn't set file position to the beginning of the file \"%s\"", opt_file_to_write);
+	size_t ret = fread(contents, 1, contents_len, f);
```
Jakuje, line 4064: I would prefer to have the declaration at the beginning of the block. Sounds like no CI we run enforces this, but it is a good practice to avoid shadowing the variable declarations (such as you do on line 4084) and getting unexpected errors.

Jakuje (approved):

frankmorgner (approved):

xhanulik (approved):
lgtm

jluebbe:
OpenEmbedded/Yocto [uses pkscs11-tool --write-object /proc/self/fd/0](https://git.openembedded.org/meta-openembedded/tree/meta-oe/classes/signing.bbclass?id=4a05d9a185df41e3b84f4d821db831836fa77e8d#n234) to convert PEM files to DER format while importing keys into SoftHSM. Since this change, it fails with `error: Couldn't set file position to the end of the file "/proc/self/fd/0"`, as seeking is not possible in a pipe.

Do you consider importing objects from stdin (to avoid writing the to a temporary file) as a valid use-case?

Jakuje:
> OpenEmbedded/Yocto [uses pkscs11-tool --write-object /proc/self/fd/0](https://git.openembedded.org/meta-openembedded/tree/meta-oe/classes/signing.bbclass?id=4a05d9a185df41e3b84f4d821db831836fa77e8d#n234) to convert PEM files to DER format while importing keys into SoftHSM. Since this change, it fails with `error: Couldn't set file position to the end of the file "/proc/self/fd/0"`, as seeking is not possible in a pipe.

AFAIK the pkcs11-tool already knows to convert the PEM files to DER when writing them to the object so the use case could be simplified. The only issue might be if the PEM is password protected (not sure how this is exactly handled by pkcs11-tool).

> Do you consider importing objects from stdin (to avoid writing the to a temporary file) as a valid use-case?

I think it is valid use case, but I am not sure if there is a good way to have both flexibility to handle large objects and not doing `seek`s. It might work with reallocs though. If this is a important use case for you, I think this change would be accepted here.

PR #3184, opened 2024-06-20
[pkcs11-tool] Add support for CKM_AES_CMAC_GENERAL

b-ori:
Hi,
This pull request adds the support for the CKM_AES_CMAC_GENERAL mechanism in pkcs11-tool.
The CK_MAC_GENERAL_PARAMS and CK_MAC_GENERAL_PARAMS_PTR data types were added as they are needed by this mechanism.
The changes were validated using the Trustonic TEE HSM.
Best regards,
Rémy


src/tools/pkcs11-tool.c
```diff
@@ -2332,10 +2344,20 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
 
+	CK_MAC_GENERAL_PARAMS mac_gen_param;
 	if (opt_mechanism == CKM_RSA_PKCS_PSS && (size_t)sz != hashlen) {
 		util_fatal("For %s mechanism, message size (got %z bytes) "
 			"must be equal to specified digest length (%lu)\n",
 			p11_mechanism_to_name(opt_mechanism), sz, hashlen);
+	} else if (opt_mechanism == CKM_AES_CMAC_GENERAL) {
+		if (opt_mac_gen_param == 0 || opt_mac_gen_param > 16) {
+			util_fatal("For %s mechanism, the value CK_MAC_GENERAL_PARAMS"
+				"must be comprised between 1 and 16 (>=8 recommended).\n",
```
Jakuje, line 2355: in this error message, I would prefer to reference the cli option name used instead of the PKCS#11 value (or both) as this is what user provided and what was wrong. User reading this error needs to backtrack from the parameter name to the cli option, which might not be straightforward.

src/tools/pkcs11-tool.c
```diff
@@ -2332,10 +2344,20 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 	if (sz < 0)
 		util_fatal("Cannot read from %s: %m", opt_input);
 
+	CK_MAC_GENERAL_PARAMS mac_gen_param;
```
Jakuje, line 2347: please, move this to the beginning of the function to the other declarations.

Jakuje (commented):
Does this work with softhsm? If so, it would be great to get some test coverage for this.

b-ori:
https://github.com/opendnssec/SoftHSMv2/blob/913e7bfd463194fadcdd28f578087cc9c15643ee/src/lib/SoftHSM.cpp#L792

It does not look like the algorithm is supported in softhsm

src/tools/pkcs11-tool.c
```diff
@@ -300,6 +301,7 @@ static const struct option options[] = {
 	{ "generate-random",	1, NULL,		OPT_GENERATE_RANDOM },
 	{ "allow-sw",		0, NULL,		OPT_ALLOW_SW },
 	{ "iv",			1, NULL,		OPT_IV },
+	{ "mac-general-param",	1, NULL, 		OPT_MAC_GEN_PARAM},
```
Jakuje, line 304: last things:
 * The new cli switches need to be added to the manual pages
 * there are some slight formatting suggestions. You do not need to reformat all of the tables, but the new code should follow the formatting if possible.

b-ori:
@Jakuje just addressed your last comment, tell me if you want me to correct anything else

Jakuje (approved):

Jakuje:

PR #3186, opened 2024-06-24
Fix disable flag for application conf block

llogar:
Currently the mere presence of 'disable=xx' disables the application. Even the 'disable=false', which imho can be confusing. 
Treat it as a boolan instead...

##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):
Interestingly, this is already documented in opensc.conf.5.xml.in as being a bool.

xhanulik (approved):

PR #3188, opened 2024-06-24
Only append PIN name to token label if necessary

llogar:
There is no need to append PIN name to token label, if the token has only one PIN object (occupies only one virtual slot).

##### Checklist

- [x] PKCS#11 module is tested (with **eOI** card, which contains 2 apps, one with 2 PIN objects and another one without any PIN objects)
- [ ] macOS tokend is tested


Jakuje (approved):
This is a good idea. Can you adjust the formatting according to the clang-format?

Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you! Merging.

PR #3189, opened 2024-06-24
Introduce SC_CARDCTL_GET_MODEL

llogar:
Card model can now alternatively be set by card driver via the` sc_card_ctl()` `SC_CARDCTL_GET_MODEL` call and not just via the configuration file.

If no card model is specified in `opensc.conf` and/or via the `sc_card_ctl()` then default `PKCS#15`/`PKCS#15 emulated` names will be used.

For now only **eOI** driver uses this, but this can can easily be extended to other drivers as well.

##### Checklist

- [x] PKCS#11 module is tested (with **eOI** card)
- [ ] macOS tokend is tested


frankmorgner (approved):

Jakuje:
Thank you for the contribution! Merging.

Issue #2716, opened 2023-02-22
Implement the Official PKCS#11 Test Cases from v3.1


frankmorgner:
### Problem Description

With PKCS#11 3.1, OASIS has created some test cases, which would be nice to pass with OpenSC.

- https://www.oasis-open.org/2022/07/25/pkcs-11-profiles-v3-1-from-pkcs-11-tc-approved-as-a-committee-specification/
- https://youtu.be/vgQiYptgzh8?t=964

frankmorgner:
I'd love to see that. From what I read from the specification the task would be to
1. set up some testing environment with a virtual card just like we have in CI
2. create some new program or use pkcs11-tool to help with bootstrapping the process
3. map the [XML specified test cases](https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.1/cs01/test-cases/pkcs11-v3.1/) to some actual PKCS#11 calls

As alternative we also have some test cases in `src/tests/p11test`. But using the pkcs#11 test cases could provide some benefit also to other pkcs#11 modules than just OpenSC, which is why I think such a test program should be shipped seperately.

bharsaklemukesh975:
May I know the steps to set up some testing environment with a virtual card just like we have in CI?

bharsaklemukesh975:
I am getting the following error while calling ./.github/test-piv.sh

card-openpgp.c:1164:7: error: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]
                func(blob);
                    ^

frankmorgner:
Thanks for pointing that out, I've created a fix here https://github.com/OpenSC/OpenSC/pull/2748

bharsaklemukesh975:
I will start this once we have #2748 completed.

frankmorgner:
Done

bharsaklemukesh975:
Thanks @frankmorgner , any idea about the following compilation error, I see used after set?

pkcs11-tool.c:7364:6: error: variable 'errors' set but not used [-Werror,-Wunused-but-set-variable]
        int errors = 0;
            ^

frankmorgner:
I've opened an other fixup https://github.com/OpenSC/OpenSC/pull/2749

If you stumble on more issues, feel free to create a fixup yourself :wink:

bharsaklemukesh975:
Thanks understood!

frankmorgner:
[...]

PR #3192, opened 2024-06-26
Implement testing tool for PKCS#11 Test Cases from v3.1

xhanulik:
Related #2716

This PR contains a base implementation of the tool for running official PKCS#11 test cases from v3.1. The tool accepts XML files with specified test cases (details of the XML representation are [here](https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.1/cs01/pkcs11-profiles-v3.1-cs01.html#_Toc98325423)) and PKCS#11 module, it parser the XML file into function calls and input parameters and finally it should test return value and output parameters.

At the moment, it covers the functionality used in mandatory test cases (https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.1/cs01/test-cases/pkcs11-v3.1/mandatory/). After finalization, it should be possible to use the tool to test the PKCS#11 module with the user's specified test cases.

For now, it is still WIP, as it needs at least:

- [ ] setting up the testing environment with a virtual card for running the tests against it,
- [ ] test XML parsing for proper mapping between function calls and input/output parameters.

Any comments are welcome.

##### Checklist
- [ ] macOS tokend is tested


frankmorgner:
Nice to see someone caring for this kind of test as well (#2716). I started a similar project a couple of months ago when I was learning some Rust https://github.com/frankmorgner/pkcs11-test. Unfortunately, I never found the time to finalize my project (remove hard coded paths, add some documentation, beautify the code with more rust magic), which is the reason for this sitting around in a private repository until now. My code is fully functional, meaning that it is capable of running and performing all the mandatory tests from the standard.

As my code uses the rust module cryptoki, it cannot replicate the very exact sequence of a few PKCS#11 calls (e.g. regarding the C_FindObject*). That could only be done when handling the PKCS#11 call backs in unsafe code blocks, which I wanted to avoid in my learning project. Looking back to this now, I think this was a good decision as it heavily simplifies the coding and still allows most tests to be executed exactly as they are intended. This limitation would be one advantage of implementing this in pure C rather than in rust.

From an implementation point of view, some trickier things I found where the following:
1. implement the support for variable input/output, which are expressed with shell-variable-like expressions in the XML definition is not trivial. As far as I see, your code does not support this, right? I implemented this using a hash map with the variable name as key
https://github.com/frankmorgner/pkcs11-test/blob/591a58ba24e3713aadccfb11c2d88bdd1a96503c/src/main.rs#L114
2. Also, when checking if a test succeeds, you currently only look at the return value rather than the returned data. In my implementation, I found that checking the actual data sometimes requires to be a little more relaxed to reflect the behavior that is propably intended (reflected by the `STRICT` setting in my code).
3. There is an obvious error in the official test cases, did you stumble upon this as well?
https://github.com/frankmorgner/pkcs11-test/blob/591a58ba24e3713aadccfb11c2d88bdd1a96503c/src/test-cases/pkcs11-v3.1/mandatory/BL-M-1-31.xml#L67
4. What is still unknown to me is how it is possible to select the testing profile in OpenSC or any other PKCS#11 module.

From an organizational point of view, I think it would be good to have a testing tool in src/tools rather than to hide it in an internal (and not installed) folder for tests (if the tool is capable of handling other modules as well). I will keep my implementation as seperated project for this reason.

xhanulik:
Hi @frankmorgner, I'm returning to this PR again as I want to see the testing tool run in OpenSC CI. From a quick look at the implementation in Rust, it's probably cleaner than the pure C implementation, primarily because of the XML parsing part, which is quite heavy in C. But on the other side, this processing in C brings some flexibility, as it allows to call the PKCS#11 functions out of the sequence order. The intention of the code in this PR is to be flexible, so it should work not only on the official test cases but also on custom test cases.

As I'm not skilled in Rust, I'm curious to know how much work would need to be done in  https://github.com/frankmorgner/pkcs11-test to be finalized, but I would probably be less of a help there. However, I would like to set up a virtual environment to run the tool, which would be useful in CI. What do you think about the Rust vs C implementation? As I would like to make the code in this PR work, I don't want to work unnecessarily on something already implemented and usable, so I could at least help with some integration with the Rust project.

frankmorgner:
I'm not skilled in Rust, too. I used this project to learn, but I agree that Rust's infrastructure offers many simplifications. However, you may have noticed that my project is using a PKCS#11 wrapping library to do the heavy lifting of handling `unsave` c code (the testee) mapped to rust. The drawback is that I don't always have *full* control over the exact flow of calls (in particular of functions that are split up in multiple commands C_*Init, C_*Final), but for testing purposes, I think, this would be an acceptable limitation.

A C implementation inside OpenSC, on the other hand, would make the testing self contained and it would allow exact control of the test cases. Due to the high complexity and the limited applicability, I believe that not many developers would notice this tool, let alone using this tool elsewhere. At the very least, I would recommend extracting pkcs11test to a seperate project to make it available (and visible) to other PKCS#11 providers. (Maybe this step could also be done for p11test, but my understanding is that p11test is not a general purpose tool, but instead relies very heavily on the OpenSC implementation of the supported cards.)

My recommendation would be a) create a seperate project (possibly in the organization OpenSC). For using this inside OpenSC CI, just use the tool that *you* think is most appropriate.

Item 4 from [above](https://github.com/OpenSC/OpenSC/pull/3192#issuecomment-2192889839) is still unsolved: I don't know how to switch to some particular (testing) profile via PKCS#11 interface. This is needed, to, for example, change some static labels to match the test case requirements.

frankmorgner:
For my part, I'll keep the rust implementation as a seperate "project".

What's the status of your implementation, is there anything to do? Will you polish this a little more to integrate in OpenSC or will you use it as seperate project?

xhanulik:
I'll take a look and see how much polishing it needs. However, it looks like it already adds quite accessive amount of file & code so it would probably make sense to move it into separate project.

frankmorgner:
I'm now confident enough that with the current state of https://github.com/frankmorgner/pkcs11-test we don't need the tests duplicated in OpenSC.

PR #3193, opened 2024-07-01
feat(pkcs11-tool): add support for HKDF

AlexandreGonzalo:
Hi,
We are adding the support for HKDF in our Trustonic HSM and I thought that it is a good opportunity to add it to pkcs11-tool.
I can't test with the SoftHSM since this algo is not supported.
I added two parameters salt and info for the HKDF derivation and I would like to discuss if we pass the salt and info by file or as a string (hexa string).
I also added a new function fill_attributes_seckey() to fill the template with the optional attributes. The idea is that it could be used as well for the key creation and generation instead of duplicating the code.
Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -20,7 +20,6 @@
 
 #include "config.h"
 
-#include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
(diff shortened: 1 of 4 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -5048,6 +5103,155 @@ derive_ec_key(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key, CK_MECHANISM_TYPE
...
 }
 
+static CK_BBOOL s_true = TRUE;
+static CK_BBOOL s_false = FALSE;
+
+static void fill_attributes_seckey(CK_ATTRIBUTE *template, int *n_attrs, int max_attrs)
+{
+	assert(*n_attrs < max_attrs);
+	FILL_ATTR(template[*n_attrs], CKA_PRIVATE, opt_is_private?&s_true:&s_false, sizeof(CK_BBOOL));
+	(*n_attrs)++;
```
AlexandreGonzalo, line 5113: Ideally, I would create a new macro which is checking that the template is large enough when a new attribute is added.

Jakuje: Could the `FILL_ATTR` be extended to do this checking?

Jakuje (commented):
>  I added two parameters salt and info for the HKDF derivation and I would like to discuss if we pass the salt and info by file or as a string (hexa string).

My preference would be the hex string input, but I am not sure of the HKDF use case you have in your mind for pkcs11-tool for now so if files are more convinient, I think its ok.

frankmorgner (changes requested):
Please also add documentation of the new CLI switches to pkcs11-tool's manual page

src/tools/pkcs11-tool.c
```diff
@@ -3295,6 +3311,25 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
 			FILL_ATTR(keyTemplate[n_attr], CKA_KEY_TYPE, &key_type, sizeof(key_type));
 			n_attr++;
 		}
+
```
xhanulik, line 3314: nit: trailing whitespace

src/tools/pkcs11-tool.c
```diff
@@ -5131,6 +5185,139 @@ derive_ec_key(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key, CK_MECHANISM_TYPE
 #endif /* ENABLE_OPENSSL  && !OPENSSL_NO_EC && !OPENSSL_NO_ECDSA */
 }
 
+static CK_BBOOL s_true = TRUE;
+static CK_BBOOL s_false = FALSE;
+
+#define FILL_ATTR_EX(attr, index, max, typ, val, len) \
+	{ \
+		assert(*(index) < max); \
```
frankmorgner, line 5193: Would it be possible to use some error handling instead of `assert()`?

Jakuje (approved):

src/tools/pkcs11-tool.c
```diff
@@ -3378,7 +3393,25 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
 			FILL_ATTR(keyTemplate[n_attr], CKA_KEY_TYPE, &key_type, sizeof(key_type));
 			n_attr++;
 		}
-		else {
+
+		else if (strncasecmp(type, "HKDF:", strlen("HKDF:")) == 0) {
```
sarroutbi, line 3397: HKDF: seems to be defined in different places. May be a const char* for it would allow easy changing it

src/tools/pkcs11-tool.c
```diff
@@ -3378,7 +3393,25 @@ gen_key(CK_SLOT_ID slot, CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *hSecretKey
...
+			CK_MECHANISM_TYPE mtypes[] = {CKM_HKDF_KEY_GEN};
+			size_t mtypes_num = sizeof(mtypes) / sizeof(mtypes[0]);
```
sarroutbi, line 3399: You could use ARRAY_SIZE macro for this

sarroutbi (commented):
Minor stuff

src/tools/pkcs11-tool.c
AlexandreGonzalo: The code I added for HKDF is a copy/paste of the GENERIC case.
We would have to do the same for the other key types otherwise the code would not be consistent.
I will address your remark in a new PR.

src/tools/pkcs11-tool.c
AlexandreGonzalo: The code I added for HKDF is a copy/paste of the GENERIC case.
We would have to do the same for the other key types otherwise the code would not be consistent.
I will address your remark in a new PR.

AlexandreGonzalo:
> Minor stuff

Thanks for your review.
I will address your remarks in a new PR.

Jakuje:

PR #3194, opened 2024-07-02
Allow OpenSSL engine header only for version <3.0

xhanulik:
This PR adds check for OpenSSL version when including `engine.h` and `OPENSSL_NO_ENGINE` is not defined.

In Fedora rawhide, the engine related files were moved to separate subpackage `devel-engine`, so the include for `engine.h` is problematic.
However, in OpenSC `openssl.c` is engine API used only for OpenSSL version < 3.0, so the version check is added also to include directive (this particular include of `engine.h` is there probably for https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/openssl.c#L261-L307 which is also already wrapped in `#if OPENSSL_VERSION_NUMBER < 0x30000000L` and `#if !defined(OPENSSL_NO_ENGINE)`). 

##### Checklist

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -33,6 +33,7 @@
 #include <openssl/conf.h>
 #include <openssl/opensslconf.h> /* for OPENSSL_NO_* */
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#define OPENSSL_NO_ENGINE
 #include <openssl/core_names.h>
 #include <openssl/param_build.h>
 #endif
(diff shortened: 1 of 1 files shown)


Jakuje:
This does not have anything to do with the libp11/pkcs11-provider. Both of them can load any generic pkcs11 module ragardless of this change.

The engines in OpenSC are/were used to load gost engine and because of the engines deprecation, this is no longer built with OpenSSL 3.0 if I see the code right. The only remaining part that was missed is the header file, which was uncovered by the change in Fedora removing the header file from the default openssl.

In the following PR we got suggestion to simplify this by setting `OPENSSL_NO_ENGINE` at the beginning of the file for OpenSSL 3+ instead of duplicating both conditions on all cases, which I  think is a good idea:

https://src.fedoraproject.org/rpms/opensc/pull-request/21#request_diff

Jakuje:
Maybe having this configurable would be even better, but I think we did not hear anyone who would miss this since 56df8f520bd91c2fc1c9d5367e7a2160003a9af9 (2 years ago)

dengert:
OK

Jakuje (approved):
Thank you. This looks much cleaner than wrapping every use of `OPENSSL_NO_ENGINE` with version check.

Jakuje:
Thank you. Merged to unbreak master build.

PR #3195, opened 2024-07-03
[pkcs11-tool] Add support for CKM_AES_GCM encryption

b-ori:
Hi,
This pull request adds the support for the CKM_AES_GCM mechanism for encryption and decryption in pkcs11-tool.
I also took the opportunity to fix a few coding style issues reported by the the CI.
The changes were validated using the Trustonic TEE HSM.
Best regards,
Rémy

Issue #3000, opened 2024-01-27
Incorrect  Encoding of EC and EDDSA public keys  


dengert:
### Problem Description

OpenSC has two ways to encode EC, EDDSA and XEDDSA public keys: 

- The SPKI method used in PKCS15 and X.509 public keys in certificates which is a sequence with the curve OID and the public key encoded in a ASN.1 BIT STRING. (Which is correct.)
- The other method is to just encode and decode the pubkey data in an ASN.1 OCTET STRING.  ~~Which is wrong~~ and used mostly internally in OpenSC.) _This edit of the original problem description, now believes it is not wrong._ 

_After further review of a **draft** version of ANSI X9.62  from 11/17/1997 the following is somewhat correct:_
ANSI X9.62 says: "The elliptic curve public key (an ECPoint which is an OCTET STRING) is mapped to
a subjectPublicKey (a BIT STRING)"  
_The above is from a section describing  ASN.1 when used in `X.509 type SubjectPublicKeyInfo` What: "is mapped to" means in not clear, but RFC 8410 and other RFCs define the, subjectPublicKey format for EC,  EDDSA and XEDDSA and it  appears to be the "octet string" i.e. "An octet string is an ordered sequence of octets" vs an ASN.1 OCTET STRING representing the ECPoint is used in the BIT STRING. The use of  the capitalized OCTET STRING could be misinterpreted to mean an ASN.1  OCTET STRING, and may have been changed in later drafts or releases of the standard._

PKCS11 says: `CKA_EC_POINT | Byte array | DER-encoding of ANSI X9.62 ECPoint value Q". 

_In another section in ANSI X9.62 says: "An elliptic curve point shall be represented by a value of type ECPoint:
ECPoint ::= OCTET STRING The value of ECPoint shall be the octet string representation of an elliptic curve point following the conversion routine in Section 4.4.2.a." This is an ANS.1 OCTET STRING._

EC keys where first added to OpenSC in c34caeb6625fe80550b22f8e5740f9d8a6f198da in 2010-11-30 as:
```
+static struct sc_asn1_entry c_asn1_ec_pointQ[2] = {
+       { "ecpointQ", SC_ASN1_OCTET_STRING, SC_ASN1_TAG_OCTET_STRING, SC_ASN1_ALLOC, NULL, NULL },
+       { NULL, 0, 0, 0, NULL, NULL }
+};
```
[...]

PR #3198, opened 2024-07-05
[pkcs11-tool] Improve coverage of key wrapping and unwrapping

b-ori:
Hi,
This pull request adds a few features to improve the coverage of key wrapping and unwrapping in pkcs11-tool:
- CKM_AES_GCM wrapping/unwrapping is now supported,
- CKM_AES_KEY_WRAP_PAD is added to the list of supported mechanisms,
- RSA and EC keys can now be wrapped/unwrapped.

The changes were validated using the Trustonic TEE HSM.
Best regards,
Rémy


b-ori:
This PR stacks above the changes made in [PR 3195](https://github.com/OpenSC/OpenSC/pull/3195). I will wait for it to be merged to open it for review.

src/tools/pkcs11-tool.c
```diff
@@ -3469,64 +3521,99 @@ unwrap_key(CK_SESSION_HANDLE session)
...
+	}
+
+	if (opt_is_extractable != 0) {
+		FILL_ATTR(keyTemplate[n_attr], CKA_EXTRACTABLE, &_true, sizeof(_true));
+	} else {
+		FILL_ATTR(keyTemplate[n_attr], CKA_EXTRACTABLE, &_false, sizeof(_true));
+	}
+	n_attr++;
+
+	/* softhsm2 does not allow to attribute CKA_VALUE_LEN, but MyEID card must have this attribute
```
b-ori, line 3611: I'm conflicted regarding this approach. The PKCS11 spec explicitly mentions that CKA_VALUE_LEN **must** be absent from the template when unwrapping an AES key. Do MyEID cards not comply to PKCS11?

It should at least be mentioned somewhere that a PKCS11 module will return an error if anything other than `--type AES:` is present in an AES key unwrap command.

Jakuje: @popovec can you have a look to this comment regarding the myeid driver and the need for the CKA_VALUE_LEN for AES keys?

src/tools/pkcs11-tool.c
```diff
@@ -3545,57 +3545,75 @@ unwrap_key(CK_SESSION_HANDLE session)
...
 			n_attr++;
 		}
+		FILL_ATTR(keyTemplate[n_attr], CKA_SIGN, &_true, sizeof(_true));
+		n_attr++;
+	}
+
+	if (opt_is_extractable != 0) {
+		FILL_ATTR(keyTemplate[n_attr], CKA_EXTRACTABLE, &_true, sizeof(_true));
+	} else {
+		FILL_ATTR(keyTemplate[n_attr], CKA_EXTRACTABLE, &_false, sizeof(_true));
```
Jakuje, line 3607: technically, this should be `sizeof(_false)`, but given the size is same, it does not matter. It might be just confusing for a reader or less clever static analyzers ...

Jakuje (commented):
Merged the underlying PR and left some minor notes. Generally looks good.

src/tools/pkcs11-tool.c
popovec: First, the technical reasons why MyEID needs to know the length of the unwrapped key before the operation itself - before the "unwrap" operation, MyEID needs to create a file in which the corresponding AES key is stored. The length of the file must correspond to the length of the AES key.

In a certain way, pkcs#11 (ver.2.40, 2.8.5 AES-CBC or ver.3.0, 2.10.5 AES-CBC) explains it, where, as far as I understand, CKA_VALUE_LEN can also be part of template:

"For unwrapping, the mechanism decrypts the wrapped key, and truncates the result according to the
CKA_KEY_TYPE attribute of the template and, if it has one, and the key type supports it, the
CKA_VALUE_LEN attribute of the template. The mechanism contributes the result as the CKA_VALUE
attribute of the new key; other attributes required by the key type must be specified in the template"

The description of pkcs#11 is a bit imprecise for me, the table in the relevant pkcs#11 section shows it more clearly: C_UnwrapKey -> Output length -> determined by type of key being unwrapped or CKA_VALUE_LEN. 

From this I conclude that the implementation can use CKA_VALUE_LEN specified in template.

b-ori: The quote is correct, however AES keys in fact do not support it. In [ver.3.0, 2.10.2 AES secret key objects]/[ver.2.40, 2.8.2 AES secret key objects] the CKA_VALUE_LEN attribute has a footnote that specifies that for AES keys it must be absent from the template when unwrapping.

image
image

popovec: As I already mentioned, the pkcs#11 specification is sometimes rather difficult to read for me, it is difficult for me to find out which case applies to what. I am looking at the document: http://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/pkcs11-curr-v3.0.pdf and specifically on the AES-CBC method (or AES-ECB, which MyEID also has implemented). The tables for the mentioned methods (table 71 and 72) do not mention footnotes for CKA_VALUE_LEN, (in contrast to table 70, which contains the footnotes you mentioned). For this reason, I believe that the AES-CBC/AES_ECB __method__ (used for unwrap operation) already allows the use of CKA_VALUE_LEN while "AES_secret key" __objects__ do not allow the use of CKA_VALUE_LEN.

b-ori: The spec is indeed difficult to read.
My understanding is that when unwrapping a key whose key type supports CKA_VALUE_LEN in the unwrapping template, AES-ECB/AES-CBC will make use of this attribute and otherwise will only use the CKA_KEY_TYPE attribute (as well as the length of the wrapped key) to determine the length of the key, hence
> "the mechanism [...] truncates the result according to [...], if it has one _(e.g. in the template)_ and the key type supports it, the CKA_VALUE_LEN attribute of the template."
AES key objects have a CKA_VALUE_LEN attribute **but** do not support specifying it in the unwrapping template.

popovec: I think OpenSC needs to work in "benevolent" mode and occasionally allow things not allowed in the spec. Currently, the pkcs11-tool allows the entry of the AES key type without specifying the length, or with the entry of the key length, and accordingly the corresponding field CKA_VALUE_LEN in the template will be filled or not filled. If one day it turns out that this behavior is not allowed, I will try to modify the unwrap function for the MyEID driver so that this driver also does not need to know the length of the unwrapped key in advance.

b-ori: I completely agree, I never meant to prevent specifying the length altogether.
I'll just write a note in the documentation to make the user know that in some cases, they will need to discard the length in the key type.

b-ori:
I am writing some tests to cover the mechanisms I added

src/tools/pkcs11-tool.c
dengert: I have been trying to follow this discussion. 

> First, the technical reasons why MyEID needs to know the length of the unwrapped key before the operation itself - before the "unwrap" operation, MyEID needs to create a file in which the corresponding AES key is stored. The length of the file must correspond to the length of the AES key.

As pointed out PKCS11 table 11 (6)  "6 MUST not be specified when object is unwrapped with C_UnwrapKey."

But if the MyEID is trying to unwrap an AES key, PKCS11 3.0 current says:
```
2960 For unwrapping, the mechanism decrypts the wrapped key, and truncates the result according to the 
2961 CKA_KEY_TYPE attribute of the template and, if it has one, and the key type supports it, the 
2962 CKA_VALUE_LEN attribute of the template. The mechanism contributes the result as the CKA_VALUE 
2963 attribute of the new key; other attributes required by the key type must be specified in the template.
```

The encrypted length of an encrypted AES key is the same size as the unencrypted form, as they are all multiples of the blocksize.  So in this case the encrypted size would need to present in the  `CK_UnwrapKey` `CK_ULONG ulWrappedKeyLen` and the  `CK_ATTRIBUTE_PTR pTemplate` would contain `CK_KEY_TYPE keyType = CKK_AES`,
`CK_VOID_PTR pValue;` and `CK_ULONG ulValueLen;`

which could then be use to in place of  the "MUST not be present" `CKA_VALUE_LEN`" in question.  

The `ulWrappedKeyLen` would be the same as the The resulting `ulValueLen`  in `CKA_VALUE`  `CK_ATTRIBUTE` 
```
CK_ATTRIBUTE_TYPE type;
CK_VOID_PTR pValue; 
CK_ULONG ulValueLen;
```
So in effect MyEID  "know**S** the length of the unwrapped key in advance."
Did I get this right?

b-ori: > The encrypted length of an encrypted AES key is the same size as the unencrypted form

Sadly this is rarely true... RSA wrapping mechanisms, PKCS#7 padded AES wrapping mechanisms, AES GCM (among others) will always output a wrapped key that is longer than the initial key.

popovec: > So in effect MyEID "knowS the length of the unwrapped key in advance."
> Did I get this right?

Yes, you are right, with the AES-CBC and AES-ECB mechanisms it is probably still possible to determine the length of the resulting key from the length of the encrypted data. Padding does not apply here. This of course means that AES-CBC/AES-ECB does not allow wrapping a 192-bit key using a 256-bit key or a 128-bit key either, as the resulting key does not have a length in multiples of the block size, so AES-CBC-PAD must be used in such cases.

For cases where it is possible to derive the length of the resulting key without CKA_VALUE_LEN information, it will probably be appropriate to correct the unwrapping code so that it derives the relevant key length from the input data.

popovec: The object into which the key is unwrapped is created in the function CK_RV C_UnwrapKey(), specially by calling 
`rv = sc_create_object_int(hSession, pTemplate, ulAttributeCount, phKey, 0);` . At this point, CKA_VALUE_LEN must be present in the template, so "adding" CKA_VALUE_LEN to the template for specific cases should happen before this call.

b-ori:
Tests added, I faced a few limitations and could not add asymmetric key wrapping to the list.

The last version of the apt package of SoftHSM2 is pretty far behind their master branch and they supposedly have corrected issues with the PKCS#8 encoding of RSA private keys since then. I built the SoftHSM2 locally and I can confirm that I could correctly wrap RSA keys with CKM_AES_CBC_PAD and CKM_AES_KEY_WRAP_PAD.
Should a new release happen, I'll come back to these tests and add the ones that are missing.

doc/tools/pkcs11-tool.1.xml
```diff
@@ -146,9 +146,9 @@
 					<term>
 						<option>--key-type</option> <replaceable>specification</replaceable>
 					</term>
-					<listitem><para>Specify the type and length (bytes if symmetric) of the key to create,
+					<listitem><para>Specify the type and (optional) flavour of the key to create,
```
Jakuje, line 149: I think this is still not completely accurate as I think the `GENERIC` secret always need the length if I remember well. I would probably keep the length naming as "length" is usually more understandable for the users than the flavour. I think the original note `(bytes if symmetric)` was meant as "bits for asymmetric and bytes for symmetric), which got lost by this change. Do we have some some other key type where the length is optional than AES and for unwrapping? If not, I would keep it as specific as possible:
```suggestion
					<listitem><para>Specify the colon separated type and length (optional for AES unwrapping) of the key to create,
```

b-ori: If we strictly stick to the PKCS#11 spec, the presence of CKA_VALUE_LEN in the template is:
- for unwrapping **never** _compulsory_, **sometimes** _forbidden_,
- for key creation (C_CreateObject) **always** _forbidden_,
- for key generation **always** _compulsory_.

Also the reason I changed "length" to "flavour" is because in many cases the "thing" that follows the key type is not a length, but an elliptic curve identifier or another kind of parameter.
I agree that we are loosing the byte/bit length differentiation, but, now that we're on this road, it feels weird to mention an "AES:16" key. The more traditional 128 bit-wise length could be processed in the program to send the expected byte-wise value to the P11 module.

I feel like this CLI parameter fills too big of a role and also needs to work with plugins that do not strictly follow the same conventions so I find it very difficult to give it an exhaustive, concise description.
IMO the correct thing to do would be to separate it, but it's hardly ideal since it would break retrocompatibility.
For example have ```key type``` alone, a ```key length``` that would result in a CKA_VALUE_LEN for symmetric key objects and CKA_MODULUS_BITS for RSA keys and a ```key flavour``` for everything else. Or merge ```key length``` and ```key flavour```.

Would something of this extent be more acceptable ? It feels awkward but at least its somehow complete...
> Specify the type and (not always compulsory) flavour (byte-wise symmetric key length, bit-wise asymmetric key length, elliptic curve identifier, etc.) of the key to create, for example RSA:1024, EC:prime256v1, GOSTR3410-2012-256:B, DES:8, DES3:24, AES:16, AES: or GENERIC:64.

Jakuje: Yes, I think this should be enough. Thank you for having a look into that!

Once we touch it, it might make sense to use some better example for RSA, as people tend to copy from manual pages and use that. Having 2k in examples might work better.

dengert: > key flavour for everything else. 
Since pkcs11-tool is for PKCS11 we should stick to PKCS11 terms.  As you pointed out  CKA_*_LEN is in bytes and CKA_*_BITS is in bits.

All EC mechanisms use CKA_EC_PARAMS which has 4 choices: ECParameters, OID curveNames, implicitlyCA and  curveName as PrintableString.   "The use of oId or curveName is recommended over the choice ecParameters. The choice implicitlyCA must not be used in Cryptoki." 

standardized as OID curve names.  was added as some curves names where bases on IETF experimental curve names,  and implemented in software and hardware but were later defined with standardized as OID curveNames. 
See https://github.com/OpenSC/OpenSC/issues/3000  and https://github.com/OpenSC/OpenSC/pull/3090  
(which is not complete), but has several changes for pkcs11-tool to use standardized OID curve names and address the use of standardized as OID curveNames. 

I have not looked at all the PKCS11 defined mechanisms but DES, AES, RSA, EC and maybe GOST are the common ones supported by OpenSC.  So "flavor" comes down to OID curveNames.

Jakuje (approved):

b-ori:
@Jakuje the last changes resolved the conflicts introduced by the merge of [PR 3193](https://github.com/OpenSC/OpenSC/pull/3193), fundamentally nothing changed

Jakuje:
Thank you for taking care of the conflict. The PIV tests are broken now (fixed in another PR). Given there were no other comments/concerns. Merging.

PR #3175, opened 2024-06-11
feat(pkcs11-tool): add support for RSA OAEP encryption

AlexandreGonzalo:
Hi,
In this PR, I propose to add the support for RSA OAEP for the "--encrypt" option.
The tool already supports decryption with the "--decrypt" option  or both encryption and decryption with the "--test" option.
I checked the patch with the Trustonic TEE HSM.
Regards,
Alexandre.

PR #3202, opened 2024-07-16
Avoid dead code

Jakuje:
Merging #3175 introduced dead code that I missed, but coverity was fast to point it out.

I will try to extend the pkcs11-tool tests with kryoptic, which has much wider selection of algorithms to catch these issues earlier.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -7528,7 +7528,6 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
 	int             failed;
 	CK_RV           rv;
 	int pad;
-	CK_RSA_PKCS_MGF_TYPE mgf = CKG_MGF1_SHA256;
 	CK_RSA_PKCS_OAEP_PARAMS oaep_params;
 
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -7651,17 +7650,13 @@ static int encrypt_decrypt(CK_SESSION_HANDLE session,
...
 		case CKG_MGF1_SHA1:
 			md = EVP_sha1();
 			break;
 		case CKG_MGF1_SHA224:
 			md = EVP_sha224();
 			break;
-		default:
-			printf("mgf %s unknown, defaulting to CKG_MGF1_SHA256\n", p11_mgf_to_name(mgf));
-			mgf = CKG_MGF1_SHA256;
-			/* fall through */
```
xhanulik, line 7664: Shouldn't the `default` case for `CKG_MGF1_SHA256` be just moved at the end with setting `md = EVP_sha256()`?

Jakuje: The `oaep_params.mgf` is always set by `build_rsa_oaep_params` so I think there is no need to set the default here.

PR #3208, opened 2024-07-25
Update min-length of user-pin to support new tokens

YuliaKolobok:
Rutoken ECP BIO requires min-length param for user/so PIN to be greater than 6 bytes


Jakuje (approved):

frankmorgner (approved):

Jakuje:
Thank you!

PR #3209, opened 2024-07-25
ci: Fix PIV test after yubico-piv-tool update

Jakuje:
The recent CI runs started failing with errors like this causing the CI failures:

    yubico-piv-tool: ambiguous argument, "verify", for option `--action' (`-a')

The new option is called "verify-pin" so updated the scripts to use that.

### DIFF
--- a/.github/test-piv.sh
+++ b/.github/test-piv.sh
@@ -55,19 +55,19 @@ $VALGRIND opensc-tool -n
 
 PIN="123456"
 yubico-piv-tool -v 9999 -r 'Virtual PCD 00 00' -P "$PIN" -s 9e -a generate -A RSA2048 | tee 9e.pub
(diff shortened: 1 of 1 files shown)


Jakuje:
> could you please mention the version of the tool in the commit message, that we are now compliant with?

Its not released version yet -- the code is so far only in the master, likely merged from Yubico/yubico-piv-tool#492. I can just guess it will be something after `yubico-piv-tool-2.5.2` (latest release) -- either 2.5.3, but more likely 2.6.x or even larger as this sounds like break of API.

xhanulik (approved):
The `verify-pin` option is there for some time (https://github.com/Yubico/yubico-piv-tool/commit/ca45b6a5653b45c33e793636b97538bf9d1644ff).

Issue #3217, opened 2024-08-08
A command prompt window briefly appears on login when using the 64-bit version with Windows


Reisenbrg:
### Problem Description
A command prompt window briefly appears on login when using the 64-bit version with Windows.

### Proposed Resolution
Create a Visual Basic script at "tools/pkcs11-launcher.vbs" that launches "[tools/pkcs11_register.exe](https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-register.c)" invisibly and modify the autorun entry created by the [WiX installer](https://github.com/OpenSC/OpenSC/blob/master/win32/OpenSC.wxs.in) to reflect this change.

Note:
- It has to be a standalone Visual Basic script because the one-liner autorun entry triggers a false positive with Windows Defender.

### Steps to reproduce
Install the latest  64-bit version version of OpenSC in Windows and log out and in again.

PR #3218, opened 2024-08-08
Stopped a command prompt window from briefly appearing on login when using the 64-bit version with Windows.

Reisenbrg:
Stopped a command prompt window from briefly appearing on login when using the 64-bit version with Windows by creating a Visual Basic script at "[tools/pkcs11-launcher.vbs](https://github.com/Reisenbrg/OpenSC-Contributions/blob/windows-x64-ui-bugfix/src/tools/pkcs11-launcher.vbs)" that launches "[tools/pkcs11_register.exe](https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-register.c)" invisibly and modifying the autorun entry created by the [WiX installer](https://github.com/OpenSC/OpenSC/blob/master/win32/OpenSC.wxs.in) to reflect this change.

Fixes: #3217

### DIFF
--- a/src/tools/pkcs11-launcher.vbs
+++ b/src/tools/pkcs11-launcher.vbs
@@ -1,0 +1,1 @@
+CreateObject("Wscript.Shell").Run """" & CreateObject("Scripting.FileSystemObject").GetParentFolderName(WScript.ScriptFullName) + "/" + "pkcs11-register.exe" & """", 0, False
--- a/win32/OpenSC.wxs.in
+++ b/win32/OpenSC.wxs.in
@@ -151,6 +151,9 @@
               <Component Id="pkcs11_register.exe" Guid="*" Win64="$(var.Win64YesNo)">
                 <File Source="$(var.SOURCE_DIR)\src\tools\pkcs11-register.exe" Vital="yes"/>
               </Component>
+              <Component Id="pkcs11_launcher.vbs" Guid="*" Win64="$(var.Win64YesNo)">
+                <File Source="$(var.SOURCE_DIR)\src\tools\pkcs11-launcher.vbs" Vital="yes"/>
+              </Component>
               <Component Id="cardos_tool.exe" Guid="*" Win64="$(var.Win64YesNo)">
                 <File Source="$(var.SOURCE_DIR)\src\tools\cardos-tool.exe" Vital="yes"/>
               </Component>
@@ -209,7 +212,7 @@
               <?endif ?>
               <Component Id="Autostart_tools" Guid="*" Win64="$(var.Win64YesNo)">
                 <RegistryKey Root="HKMU" Key="Software\Microsoft\Windows\CurrentVersion\Run">
-                  <RegistryValue Type="string" Name="pkcs11-register.exe" Value="[INSTALLDIR_TOOLS]pkcs11-register.exe" />
+                  <RegistryValue Type="string" Name="pkcs11-launcher.vbs" Value='wscript.exe "[INSTALLDIR_TOOLS]pkcs11-launcher.vbs"' />
                 </RegistryKey>
               </Component>
               <Component Id="Autostart_native_tools" Guid="*" Win64="$(var.Win64YesNo)">
@@ -377,6 +380,7 @@
         <ComponentRef Id="opensc_notify.exe"/>
         <ComponentRef Id="pkcs11_tool.exe"/>
         <ComponentRef Id="pkcs11_register.exe"/>
+        <ComponentRef Id="pkcs11_launcher.vbs"/>
         <ComponentRef Id="cardos_tool.exe"/>
         <ComponentRef Id="egk_tool.exe"/>
         <ComponentRef Id="goid_tool.exe"/>


frankmorgner (approved):
I'm not familiar with VBS, but the change looks reasonable. thank you.

H5117:
I think you are solving the problem wrongly. As I remember, the command prompt window automatically created only for applications with `Windows subsystem` set to `console` in the PE header of executable. So the right solution is to set it to `windows` for `pkcs11-register.exe`.

Meson allows to control the subsystem with the [`win_subsystem`](https://mesonbuild.com/Reference-manual_functions.html#executable_win_subsystem) flag. You can find out in the sources what compiler flags it translates to. For MSVC it is [/SUBSYSTEM](https://learn.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=msvc-170).

<details><summary>$ file OpenSC/tools/*</summary>

```
cardos-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
cryptoflex-tool.exe: PE32+ executable (console) x86-64, for MS Windows, 6 sections
dnie-tool.exe:       PE32+ executable (console) x86-64, for MS Windows, 6 sections
egk-tool.exe:        PE32+ executable (console) x86-64, for MS Windows, 6 sections
eidenv.exe:          PE32+ executable (console) x86-64, for MS Windows, 6 sections
gids-tool.exe:       PE32+ executable (console) x86-64, for MS Windows, 6 sections
goid-tool.exe:       PE32+ executable (console) x86-64, for MS Windows, 6 sections
iasecc-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
netkey-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
npa-tool.exe:        PE32+ executable (console) x86-64, for MS Windows, 6 sections
openpgp-tool.exe:    PE32+ executable (console) x86-64, for MS Windows, 6 sections
opensc-asn1.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
opensc.dll:          PE32+ executable (DLL) (GUI) x86-64, for MS Windows, 6 sections
opensc-explorer.exe: PE32+ executable (console) x86-64, for MS Windows, 6 sections
opensc-notify.exe:   PE32+ executable (GUI) x86-64, for MS Windows, 6 sections
opensc-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
piv-tool.exe:        PE32+ executable (console) x86-64, for MS Windows, 6 sections
pkcs11-register.exe: PE32+ executable (console) x86-64, for MS Windows, 6 sections
pkcs11-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
pkcs15-crypt.exe:    PE32+ executable (console) x86-64, for MS Windows, 6 sections
pkcs15-init.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
pkcs15-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
sc-hsm-tool.exe:     PE32+ executable (console) x86-64, for MS Windows, 6 sections
smm-local.dll:       PE32+ executable (DLL) (GUI) x86-64, for MS Windows, 6 sections
westcos-tool.exe:    PE32+ executable (console) x86-64, for MS Windows, 6 sections
```
</details>

frankmorgner:
The thing is that pkcs11-register.exe can *also* be used from the command line and we would like to keep it that way. In opensc-notify.exe we have achieved this by defining `WinMain()` instead of `main()` (see https://learn.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=msvc-170). From the code in opensc-notify.c you can see that program initialization and CLI handling works differently. It needs to be tested if everything works as expected if WinMain is used or if `/subsystem:WINDOWS` was used. Having a wrapping VBS (or a wrapping app that uses `WinMain` and `exec` to call the original program) seems like less work to achieve this goal.

H5117:
OK, but I vote to try adding `/SUBSYSTEM:WINDOWS` first. Don't see a reason why a console program would not work with it.

xhanulik:
@frankmorgner @H5117 what is the status of this PR? Will it be ready for 0.26.0 or should we postpone it for the next release?

frankmorgner:
Avoiding vbs would be nice, but I don't have a system to test `/SUBSYSTEM:WINDOWS `. So I'd like to postpone this change until we can try out the suggestion.

hamarituc:
> Avoiding vbs would be nice, but I don't have a system to test `/SUBSYSTEM:WINDOWS `. So I'd like to postpone this change until we can try out the suggestion.

Using `/SUBSYSTEM:WINDOWS` effectively disconnects the application from the terminal at all, even if invoked from a command shell. The output (`Added ... to ...`), commandline help (`--help`), etc. will not be shown. I tried it in #3336. A possible solution is to compile `pkcs11-register` twice, one time for autostart usage and one time for CLI usage. Using the VBS script doesn't require to duplicate `pkcs11-register`.

H5117:
> Using /SUBSYSTEM:WINDOWS effectively disconnects the application from the terminal at all

[This page](https://learn.microsoft.com/en-us/windows/console/allocconsole) says that `AllocConsole()` solves this problem (will try to test it soon). So here is the updated plan:

1. Introduce a new argument `--noconsole` for `pkcs11-register.exe`;
2. Add this block to the `main()`:

```c
#ifdef _WIN32
    if (!noconsole) {
        AllocConsole();
    }
#endif
```

3. Compile `pkcs11-register.exe` with `/SUBSYSTEM:WINDOWS`;
4. Start `pkcs11-register.exe` with `--noconsole` on login.

H5117:
OK, after some experiments I can say that there is no ideal solution on Windows. It is not a problem to attach the standard output of a GUI application to an existing console with `AttachConsole()` or create a new one with `AllocConsole()`. Particularly, this code in `pkcs11-register.exe` works for me:

```
int
main(int argc, char **argv)
{
#ifdef _WIN32
        if (AttachConsole(ATTACH_PARENT_PROCESS)) {
            FILE *stream;
            freopen_s(&stream, "CONIN$", "r", stdin);
            freopen_s(&stream, "CONOUT$", "w", stderr);
            freopen_s(&stream, "CONOUT$", "w", stdout);
        }
#endif
```

We can even don't introduce the argument `--noconsole`, because the program outputs to a parent's console only and don't create its own (exactly as we want). But the problem is that a command processor on Windows does not wait for GUI programs to complete, so it outputs the prompt before the program's output:

<details><summary>Example</summary>

```
D:\OpenSC-build>src\tools\pkcs11-register.exe --help

D:\OpenSC-build>Usage: pkcs11-register [OPTION]...

Install a PKCS#11 module to known applications.

  -h, --help              Print help and exit
  -V, --version           Print version and exit
  -m, --module=FILENAME   Specify the module to load  (default=`OpenSC's
                            PKCS#11 module')
      --skip-chrome       Don't install module to Chrome  (default=on)
      --skip-firefox      Don't install module to Firefox  (default=on)
      --skip-thunderbird  Don't install module to Thunderbird  (default=off)
      --skip-seamonkey    Don't install module to SeaMonkey  (default=off)

Report bugs to https://github.com/OpenSC/OpenSC/issues

Written by Frank Morgner <frankmorgner@gmail.com>
```
</details>

The problem is described also [here](https://stackoverflow.com/a/15955276). Particularly, there is a workaround to start a GUI program as `start /wait program`, but it is not a user-friendly way. So, after all, I think wrapping with VB script is the acceptable solution. Duplicating `pkcs11-register.exe` with different subsystems looks redundant for me.

H5117:
Visual Basic scripts are deprecated and will be disabled by default in 2027 ([link](https://techcommunity.microsoft.com/blog/windows-itpro-blog/vbscript-deprecation-timelines-and-next-steps/4148301)).

frankmorgner:
I believe that the most prominent use case of pkcs11-register was Firefox, but since on Windows and macOS FF switched to the OS' native method for hardware tokens, FF integration is already disabled by default on those platforms in pkcs11-register.

What I'm thinking about is that we could remove pkcs11-register as (default) autostart item. We have already done this on Linux due to the many different platform configurations which caused unwanted side effects. @OpenSC/maintainers , what do you think about removing pkcs11-register from autostart in general?

PR #3219, opened 2024-08-12
openpgp: Oss-fuzz fixes

Jakuje:
When generating RSA key pair using PKCS#15 init, the driver could accept responses relevant to ECC keys, which made further processing in the pkcs15-init failing/accessing invalid parts of structures.

and

Avoid buffer overflow when writing fingerprint 

Thanks oss-fuzz!

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=71010

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=70933

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2903,6 +2903,9 @@ pgp_parse_and_set_pubkey_output(sc_card_t *card, u8* data, size_t data_len,
 
 		/* RSA modulus */
 		if (tag == 0x0081) {
+			if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA) {
+				LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN_DATA_RECEIVED);
+			}
 			if ((BYTES4BITS(key_info->u.rsa.modulus_len) < len)  /* modulus_len is in bits */
 				|| key_info->u.rsa.modulus == NULL) {
 
@@ -2918,6 +2921,9 @@ pgp_parse_and_set_pubkey_output(sc_card_t *card, u8* data, size_t data_len,
 		}
 		/* RSA public exponent */
 		else if (tag == 0x0082) {
+			if (key_info->algorithm != SC_OPENPGP_KEYALGO_RSA) {
+				LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN_DATA_RECEIVED);
+			}
 			if ((BYTES4BITS(key_info->u.rsa.exponent_len) < len)  /* exponent_len is in bits */
 				|| key_info->u.rsa.exponent == NULL) {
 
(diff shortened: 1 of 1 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -2933,6 +2946,10 @@ pgp_parse_and_set_pubkey_output(sc_card_t *card, u8* data, size_t data_len,
 		}
 		/* ECC public key */
 		else if (tag == 0x0086) {
+			if (key_info->algorithm != SC_OPENPGP_KEYALGO_ECDSA &&
+					key_info->algorithm != SC_OPENPGP_KEYALGO_ECDH) {
+				LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN_DATA_RECEIVED);
+			}
 			/* set the output data */
 			/* len is ecpoint length + format byte
 			 * see section 7.2.14 of 3.3.1 specs */
```
xhanulik, line 2955: Shouldn't also lines below `key_info->u.ec.ecpoint = malloc(len);` set the `key_info->u.ec.ecpoint_len` to the `len`?

src/libopensc/card-openpgp.c
```diff
@@ -2781,14 +2781,21 @@ pgp_calculate_and_store_fingerprint(sc_card_t *card, time_t ctime,
 	/* update the blob containing fingerprints (00C5) */
 	sc_log(card->ctx, "Updating fingerprint blob 00C5.");
 	fpseq_blob = pgp_find_blob(card, 0x00C5);
-	if (fpseq_blob == NULL)
-		LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_OUT_OF_MEMORY, "Cannot find blob 00C5");
+	if (fpseq_blob == NULL) {
+		r = SC_ERROR_OUT_OF_MEMORY;
+		LOG_TEST_GOTO_ERR(card->ctx, r, "Cannot find blob 00C5");
+	}
+	if (20 * key_info->key_id > fpseq_blob->len) {
```
xhanulik, line 2788: This check covers line `p = newdata + 20 * (key_info->key_id - 1);`, but  `memcpy(p, fingerprint, 20);` is accessing another 20 bytes.

src/libopensc/card-openpgp.c
Jakuje: Good point.  In `src/pkcs15init/pkcs15-openpgp.c` where the overrun happened, the `ecpoint_len` is coming in the `required->field_length` and that one comes from the algorithm/mechanism specification.

This code on the following lines checks if the length is different than expected and I think it should really fail if the length is not the expected one (2z + 1) from specs in https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.pdf

Jakuje: Having further look in the processing, you are right and the length should be set here otherwise we will overrun the buffers somewhere later.

xhanulik (approved):

src/libopensc/card-openpgp.c
Jakuje: Turns out the windows compilers are not happy about this signed/unsigned comparison so this is something I will have to fix too.

Jakuje: ugh ... all of these are unsigned on any sane platform. Unless the `20*` would covert the u8 to signed or the u8 would be signed on windows. Any ideas?
```
card-openpgp.c(2788): error C2220: warning treated as error - no 'object' file generated
card-openpgp.c(2788): warning C4018: '>': signed/unsigned mismatch
```

xhanulik: not sure, isn't maybe `20` (when not `20U`) considered to by signed?

Jakuje: Lets try

Jakuje: Seems like it helped. So lets wait for the CI to finish.

frankmorgner (approved):

Jakuje:
Thank you for reviews! Merging.

Issue #3201, opened 2024-07-13
Gemalto IDPrime 940 no longer lists private key after renewal


ignisf:
### Problem Description

Context:

I am using a qualified e-signature by Borica's B-Trust service https://www.b-trust.bg/en/electronic-signatures/qualified-certificates . Initially the smart card and the e-signature worked flawlessly with opensc and without any proprietary pkcs11 libraries. The certificate on the card expired and I went through the issuer's online renewal process. After the renewal the token is not usable with OpenSC and only with the proprietary pkcs11 module.

### Proposed Resolution

N/A

My assumption is that the employees at the office use a different software to provision the smart cards which sets them up in a compatible with OpenSC manner. The renewal process though goes through the proprietary pkcs11 module which results in an unusable card.

### Steps to reproduce

TBD, as this might be reproducible by using the proprietary `libIDPrimePKCS11.so` library to renew a certificate. However, currently the steps to repro are:

1. Acquire a qualified certificate from https://www.b-trust.bg/en/electronic-signatures/qualified-certificates
2. Request a renewal upon approach of expiration date via https://store.borica.bg/en/qualified-signature-renewal.
3. Renew via https://store.borica.bg/en/qualified-signature-renewal which calls out to their proprietary java-based signing helper software https://www.b-trust.bg/attachments/BtrustPrivateFile/24/docs/B-TrustBISS.tar . This ships and uses an old version of `libIDPrimePKCS11.so`.
4. Observe that *OpenSC no longer sees the private key on the card*.

### Logs

https://gist.github.com/ignisf/2d537ab82585bc9a2d505b42e6246fbc

ignisf:
I am holding off going to the office to get the certificate re-issued in hope of helping out to debug this. Let me know if I can provide more data.

ignisf:
Logs from a similar but still working smart card:

https://gist.github.com/ignisf/bedc398c2625ef961118b971ea2887c8

ignisf:
@xhanulik, thank you for looking into this, here is the requested `pkcs11-tool --module /path/to/libIDPrimePKCS11.so -l -O` output for the broken smart card:

```
22280259 [140413637756608] APDU: 00 A6 00 00 15 
00010551 [140413637756608] SW: 79 6D 3F 3A 0F 59 74 0A A0 00 00 00 18 80 00 00 00 06 62 40 51 90 00 
00000197 [140413637756608] APDU: 00 A4 00 00 02 00 29 
[...]

PR #3220, opened 2024-08-21
Check key references instead of container map for IDPrime 940

xhanulik:
The IDPrime 940 card in https://github.com/OpenSC/OpenSC/issues/3201 has an empty container map file, even with a private key on the card.

The IDPrime driver uses a container map to identify whether there is a corresponding key to the certificate on the card. When not, the certificate object is stored without a key reference. However, the IDPrime 940 card also has a key reference file containing key references and the certificate ID to which they belong. The reference file can be used instead of the container map to check the existence of the key, even when the container map file is, for some reason, empty.

This cannot be used for other IDPrime card types, as they do not have any key reference file, and so a container map still needs to be used.

Changes were tested only with IDPrime 940 with a non-empty container map file. It still needs to check with the card without a container map file whether the key references are correct.

##### Checklist

### DIFF
--- a/src/libopensc/card-idprime.c
+++ b/src/libopensc/card-idprime.c
@@ -441,11 +441,16 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, 
 
 			container = (idprime_container_t *) list_seek(&priv->containers, &cert_id);
 			if (!container) {
-				/* Object is added, but missing private key */
+				/* Container map missing container with certificate ID */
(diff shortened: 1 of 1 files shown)


ignisf:
Test results here: https://github.com/OpenSC/OpenSC/issues/3201#issuecomment-2303083273

Jakuje (approved):
Thank you for taking care of this one! Looks good to me!

xhanulik:
Merging, thanks for reviews!

PR #3240, opened 2024-10-02
Add Support for ECC Signatures on D-Trust Card 4.1/4.4

### LINKED ISSUE #3236: Unable to perform ECDSA signature when `field_size` extension is not present
### Problem Description

I was about to implement the ECDSA signature for D-Trust 4.1 signature card and observed the following issue. I used the code in https://github.com/hamarituc/OpenSC/tree/dtrust-ecc. The hash to be signed is prepared by OpenSC and the card does the signature. The private key descriptor file doesn't contain the proprietary `field_size` extension. This leads to `prkey->field_length` beeing `0` in `src/libopensc/pkcs15-sec.c`:

```c
/* PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures
 * If the length of the hash value is larger than the bit length of n, only
 * the leftmost bits of the hash up to the length of n will be used. Any
 * truncation is done by the token.
 * But if card is going to do the hash, pass in all the data
 */
        inlen = MIN(inlen, (prkey->field_length+7)/8);
} 
```

The variable `inlen` initially contains the size of the hash to be signed, but now gets set to `0`. This late issues a signing command with empty data later which fails.

This is the content of the cryptographic information application:

```
OpenSC [3F00/0104]> asn1 5032
30 SEQUENCE (68 bytes)
   02 INTEGER (1 byte): 1
   04 OCTET STRING (10 bytes)
      92 76 00 32 11 86 00 05 45 0F .v.2....E.
   0C UTF8String (16 bytes): D-TRUST GmbH (C)
   80 Context 0  (29 bytes)
      44 2D 54 52 55 53 54 20 43 61 72 64 20 34 2E 31 D-TRUST Card 4.1
      20 4D 31 30 30 20 45 43 43 20 32 63 61           M100 ECC 2ca
   03 BIT STRING (2 bytes): 110
```

This is the content of the private key directory:

```
OpenSC [3F00/0104]> asn1 5001
A0 Context 0  (62 bytes)
   30 SEQUENCE (35 bytes)
      0C UTF8String (26 bytes): Authentisierungsschluessel
      03 BIT STRING (2 bytes): 01
      04 OCTET STRING (1 byte): 03 .
   30 SEQUENCE (11 bytes)
      04 OCTET STRING (1 byte): 03 .
      03 BIT STRING (3 bytes): 100000100
      02 INTEGER (1 byte): 131
   A1 Context 1  (10 bytes)
      30 SEQUENCE (8 bytes)
         30 SEQUENCE (6 bytes)
            04 OCTET STRING (4 bytes): 3F 00 01 02 ?...
A0 Context 0  (57 bytes)
   30 SEQUENCE (30 bytes)
      0C UTF8String (18 bytes): Signaturschluessel
      03 BIT STRING (2 bytes): 01
      04 OCTET STRING (1 byte): 07 .
      02 INTEGER (1 byte): 100
   30 SEQUENCE (11 bytes)
      04 OCTET STRING (1 byte): 02 .
      03 BIT STRING (3 bytes): 1000000000
      02 INTEGER (1 byte): 130
   A1 Context 1  (10 bytes)
      30 SEQUENCE (8 bytes)
         30 SEQUENCE (6 bytes)
            04 OCTET STRING (4 bytes): 3F 00 01 01 ?...
```

### Proposed Resolution

Commenting out the above cited C code, resolves the issue. But of course it is there for a specific reason, so this just demonstrates the error originates there. The case that the `field_size` extension is not present should be handled properly. But as I don't know how to do it, I chosen to open this issue to seeking for assistance.

### Steps to reproduce

Execute the following command on the D-Trust 4.1/4.4 multisignature card.

```
$ echo "Sign me" | pkcs11-tool --slot 1 -a Signaturzertifikat -s -m ECDSA-SHA256 -p 1234567890
```

### Logs

* Relevant part of the log: [opensc.txt](https://github.com/user-attachments/files/17040931/opensc.txt)

hamarituc:
This PR adds support for ECDSA Signatures of the german D-Trust Signatures Cards 4.1/4.4 issued by the german Bundesdruckerei. To set the `field_length` attribute in the private key data structure, a PKCS#15 emulation wrapper for D-Trust Cards had to be introduced.

To use the authentication certificate for ECDH key derivation, there is still more work to do, which will be addressed eventually in its own PR.

Tested on:

* D-TRUST Card 4.1 M100 ECC 2ca
* D-TRUST Card 4.1 Multi ECC 2ca

* Fixes #3236

##### Checklist

- [ ] macOS tokend is tested (unable to test, help needed)

### DIFF
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -58,7 +58,7 @@ libopensc_la_SOURCES_BASE = \
 	pkcs15-oberthur.c pkcs15-itacns.c pkcs15-gemsafeV1.c pkcs15-sc-hsm.c \
 	pkcs15-coolkey.c pkcs15-din-66291.c pkcs15-idprime.c pkcs15-nqApplet.c \
 	pkcs15-dnie.c pkcs15-gids.c pkcs15-iasecc.c pkcs15-jpki.c pkcs15-esteid2018.c \
-	pkcs15-starcos-esign.c pkcs15-skeid.c pkcs15-eoi.c compression.c sm.c \
+	pkcs15-starcos-esign.c pkcs15-skeid.c pkcs15-eoi.c pkcs15-dtrust.c compression.c sm.c \
 	aux-data.c
 
 if ENABLE_CRYPTOTOKENKIT
@@ -140,7 +140,7 @@ TIDY_FILES = \
 	pkcs15-oberthur.c pkcs15-itacns.c pkcs15-sc-hsm.c \
 	pkcs15-coolkey.c pkcs15-din-66291.c pkcs15-idprime.c pkcs15-nqApplet.c \
 	pkcs15-dnie.c pkcs15-gids.c pkcs15-iasecc.c pkcs15-jpki.c pkcs15-esteid2018.c \
-	pkcs15-starcos-esign.c pkcs15-skeid.c compression.c sm.c \
+	pkcs15-starcos-esign.c pkcs15-skeid.c pkcs15-dtrust.c compression.c sm.c \
 	aux-data.c \
 	#$(SOURCES)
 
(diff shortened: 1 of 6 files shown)


frankmorgner (approved):
LGTM, thank you

Jakuje (approved):
looks good! I think this would be good for 0.26.0

PR #3248, opened 2024-10-16
Add full ECC Support for D-Trust Card 4.1/4.4

### LINKED ISSUE #3236: Unable to perform ECDSA signature when `field_size` extension is not present

hamarituc:
This PR adds support for ECDSA signatures and ECDH key agreement for D-Trust Signatures Cards 4.1/4.4 issued by the german Bundesdruckerei.

This PR supersedes #3240. But since @frankmorgner already reviewed that PR I wouldn't like to amend it afterwards. Feel free to continue reviewing #3240 and to postpone this one or just to review this PR instead. I am unsure which way do you prefer. Sorry the noise in case you preferred just amending an already reviewed PR.

Tested on:

* D-TRUST Card 4.1 M100 ECC 2ca
* D-TRUST Card 4.1 Multi ECC 2ca
* Fixes #3236

##### Checklist

- [x] Windows minidriver is tested (only for signature, but not for key derivation)
- [ ] macOS tokend is tested (unable to test, help needed)

### DIFF
--- a/src/libopensc/card-cardos-common.h
+++ b/src/libopensc/card-cardos-common.h
@@ -1,0 +1,43 @@
+/*
+ * card-cardos-common.c: Common code for CardOS based cards
+ *
+ * Copyright (C) 2024 Mario Haustein <mario.haustein@hrz.tu-chemnitz.de>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifndef HAVE_CARD_CARDOS_COMMON_H
+#define HAVE_CARD_CARDOS_COMMON_H
+
+#include "libopensc/opensc.h"
+
+/**
+ * @brief compute a shared value from the peers public ECC key
+ *
+ * Key agreement on ECC keys requires a CardOS-specific command.
+ *
+ * @param  card[in]       struct sc_card object on which to issue the command
+ * @param  crgram[in]     public key point coordinates of the peer party in uncompressed format
+ * @param  crgram_len[in] size of the public key point
+ * @param  out[out]       output buffer for the shared value
+ * @param  outlen[in]     size of the output buffer
+ * @return number of bytes of the shared value or an error code
+ */
+int
+cardos_ec_compute_shared_value(struct sc_card *card,
+		const u8 *crgram, size_t crgram_len,
+		u8 *out, size_t outlen);
+
+#endif /* HAVE_CARD_CARDOS_COMMON_H */
--- a/src/libopensc/card-dtrust.c
+++ b/src/libopensc/card-dtrust.c
@@ -28,6 +28,7 @@
 #include <string.h>
 
 #include "asn1.h"
+#include "card-cardos-common.h"
 #include "internal.h"
 
 static const struct sc_card_operations *iso_ops = NULL;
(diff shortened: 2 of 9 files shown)


src/libopensc/card-dtrust.c
```diff
@@ -70,6 +70,27 @@ static struct dtrust_supported_ec_curves {
 };
 // clang-format on
 
+/* copied from iso7816.c */
```
Jakuje, line 73: This is already copied in cardos and gids driver. Can we make it public function in some common file to avoid the need to copy it to one more place? It makes in unmaintainable.

src/libopensc/card-dtrust.c
```diff
@@ -438,6 +472,86 @@ dtrust_compute_signature(struct sc_card *card, const u8 *data,
...
+{
+	int r;
+	struct sc_apdu apdu;
+	u8 *sbuf = NULL;
+
+	if (card == NULL || crgram == NULL || out == NULL) {
+		return SC_ERROR_INVALID_ARGUMENTS;
+	}
+	LOG_FUNC_CALLED(card->ctx);
+	sc_log(card->ctx, "CardOS compute shared value: in-len %" SC_FORMAT_LEN_SIZE_T "u, out-len %" SC_FORMAT_LEN_SIZE_T "u", crgram_len, outlen);
```
Jakuje, line 488:
```suggestion
	sc_log(card->ctx, "D-Trust compute shared value: in-len %" SC_FORMAT_LEN_SIZE_T "u, out-len %" SC_FORMAT_LEN_SIZE_T "u", crgram_len, outlen);
```

Jakuje: or if the function matches what is in the cardos, it might be better to move it to some shared file too.

src/libopensc/card-dtrust.c
hamarituc: This was my initial thought, but I didn't want to throw it in question as other drivers already copied that code. But now I mad this function public and removed all duplicate code.

src/libopensc/card-dtrust.c
hamarituc: I don't have access to CardOS specification. Just the documentation of the D-Trust card specifies it should be a CardOS-specific command.

Do you have any suggestion where to place this function?

I doubt that the CardOS driver would be the right place as it introduces a dependency between two different card drivers. The CardOS driver is intended to be used for plain CardOS cards (at least as I understand the code). In contrast the D-Trust card is just a pre-initialized PKCS#15 card built on top of CardOS.

Jakuje: Something like `card-caros-common.c` might be a good place. We use this semantics when sharing code between different drivers or their versions.

hamarituc: > Something like `card-caros-common.c` might be a good place. We use this semantics when sharing code between different drivers or their versions.

:white_check_mark: thanks, fixed

Jakuje (approved):
Thank you! Looks good to me!

src/libopensc/card-cardos-common.h
```diff
@@ -0,0 +1,43 @@
...
+ * Key agreement on ECC keys requires a CardOS-specific command.
+ *
+ * @param  card[in]       struct sc_card object on which to issue the command
+ * @param  crgram[in]     public key point coordinates of the peer party in uncompressed format
+ * @param  crgram_len[in] size of the public key point
+ * @param  out[out]       output buffer for the shared value
+ * @param  outlen[in]     size of the output buffer
+ * @return number of bytes of the shared value or an error code
+ */
+extern int
```
frankmorgner, line 38:
```suggestion
int
```

frankmorgner (changes requested):

frankmorgner:
Also, the [linter ](https://github.com/OpenSC/OpenSC/actions/runs/11442963644/job/31834521048?pr=3248)doesn't seem to like the extra empty line:
```diff
diff --git a/src/libopensc/iso7816.c b/src/libopensc/iso7816.c
index 30cebb4..e17549c 100644
--- a/src/libopensc/iso7816.c
+++ b/src/libopensc/iso7816.c
@@ -30,7 +30,6 @@
 #include "iso7816.h"
 #include "sm/sm-iso.h"
 
-
 void
 iso7816_fixup_transceive_length(const struct sc_card *card,
 		struct sc_apdu *apdu)
```

src/libopensc/card-cardos-common.h
hamarituc: fixed

frankmorgner (approved):

xhanulik (approved):

frankmorgner (approved):

xhanulik:
@frankmorgner  @Jakuje  Is this good to go to 0.26.0 or should we postpone?

Jakuje:
I think this is good to go for 0.26. It is changing only the dtrust driver. The changes in other files are only refactoring and should not introduce any changes in behavior.

PR #3249, opened 2024-10-16
Update CI runners

frankmorgner:
Both, Github and Signpath updated the list of their supported build infrastructure.

##### Checklist

- [ ] macOS tokend is tested


xhanulik (approved):

frankmorgner:
maybe we wait with a merge until the release is finalized so that we switch to VS 2017 only for the release thereafter

PR #3251, opened 2024-10-22
Run Windows CI on Github Actions

metsma:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,120 @@
+name: Windows
+
+on:
+  pull_request:
+    paths:
+      - '**.c'
+      - '**.h'
+      - '**.sh'
+      - .github/workflows/windows.yml
+      - '**.am'
+      - '**.mak'
+      - configure.ac
+  push:
+
+permissions:
+  contents: read  #  to fetch code (actions/checkout)
+
+jobs:
+  build:
+    runs-on: ${{ matrix.image }}
+    strategy:
+      matrix:
+        platform: [x86, x64]
+        configuration: [Light, Release]
+        image: [windows-2019, windows-2022]
+    env:
+      OPENPACE_VER: 1.1.3
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Git describe
+        id: ghd
+        uses: proudust/gh-describe@v2
+      - name: Package name
+        shell: bash
+        run: |
+          echo PACKAGE_NAME=OpenSC-${{ steps.ghd.outputs.tag }} >> $GITHUB_ENV
+          echo ARTIFACT=OpenSC-${{ steps.ghd.outputs.tag }}_${{ matrix.platform == 'x86' && 'win32' || 'win64' }}${{ matrix.configuration == 'Light' && '-Light' || '' }} >> $GITHUB_ENV
+      - name: Install CPDK
+        run: choco install windows-cryptographic-provider-development-kit -y > $null
+      - name: Install autotools
+        uses: msys2/setup-msys2@v2
+        with:
+          install: autotools mingw-w64-x86_64-pkg-config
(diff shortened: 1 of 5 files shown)


metsma:
I moved Appveyor build 1 to 1 github.
@frankmorgner Can you please add signpath.
I left only 2019 and 2022 VC

frankmorgner (commented):
Thank you very much for taking your time and making this work!

This PR removes AppVeyor entirely. (I am not against this step, I just want to highlight this explicitly.)

As the metadata for instructing some CI runner to do things changes from platform to platform, we have chosen in the other workflow files for Github Actions to put most of the commands into scripts in the `.github` folder. Would it be possible to do the same with the Windows commands, e.g. as powershell scripts? This approach would be more flexible if we need to switch away from Github some day.

You have also simplified [OpenSC.wxs.in](https://github.com/OpenSC/OpenSC/pull/3251/files#diff-a143744d1ceaef543502b75259cf9f9e7d91e91d558039a167a8c29d61ee0213). Did you test if local builds on a 64 bit platform still support building the 32 bit installer?

You are very generous of including new build dependencies (lukka/run-vcpkg@v7, ilammy/msvc-dev-cmd@v1, msys2/setup-msys2@v2, proudust/gh-describe@v2) for things that previously have been done with small sniplets. I am a bit sceptical of this approach as it adds an attack vector to the build infrastructure. (Also, having dedicated build scripts for the runner as suggested above, those dependencies might not be needed anymore.) Do other @OpenSC/maintainers have an opinion about this?

metsma:
> Thank you very much for taking your time and making this work!
> 
> This PR removes AppVeyor entirely. (I am not against this step, I just want to highlight this explicitly.)

I left this for now

> As the metadata for instructing some CI runner to do things changes from platform to platform, we have chosen in the other workflow files for Github Actions to put most of the commands into scripts in the `.github` folder. Would it be possible to do the same with the Windows commands, e.g. as powershell scripts? This approach would be more flexible if we need to switch away from Github some day.

I really don't want invest to much time for this because most the complexity can be solved with meson build system.

> You have also simplified [OpenSC.wxs.in](https://github.com/OpenSC/OpenSC/pull/3251/files#diff-a143744d1ceaef543502b75259cf9f9e7d91e91d558039a167a8c29d61ee0213). Did you test if local builds on a 64 bit platform still support building the 32 bit installer?

I reverted this to make follow up PR for switching to WIX 5. The change did not affect the 32 build only cleanup wix code.

> You are very generous of including new build dependencies (lukka/run-vcpkg@v7, ilammy/msvc-dev-cmd@v1, msys2/setup-msys2@v2, proudust/gh-describe@v2) for things that previously have been done with small sniplets. I am a bit sceptical of this approach as it adds an attack vector to the build infrastructure. (Also, having dedicated build scripts for the runner as suggested above, those dependencies might not be needed anymore.) Do other @OpenSC/maintainers have an opinion about this?

Like I said most of the helpers are for autotools bootstraping.
Only one is lukka/run-vcpkg@v7 that is for building OpenSSL with vcpkg and it uses caching for github actions.

frankmorgner (changes requested):
AppVeyor is reporting the following problem:
```
C:\projects\opensc\win32\OpenSC.wxs(30) : error CNDL0265 : The Platform attribute has an invalid value X64.  Possible values are x86, x64, or ia64.
```
Could you try changing this to `x64`, which may be compatible between different versions of wxs

metsma:
> AppVeyor is reporting the following problem:
> 
> ```
> C:\projects\opensc\win32\OpenSC.wxs(30) : error CNDL0265 : The Platform attribute has an invalid value X64.  Possible values are x86, x64, or ia64.
> ```
> 
> Could you try changing this to `x64`, which may be compatible between different versions of wxs

Resolved

Issue #3258, opened 2024-11-12
Minidriver doesn't wotk with EC 256


ugochirico:
### Problem Description

I have a card with EC 256 private key and the related certificate.
certutil -SCInfo doen't work and gives the error:

 0x80090016 (-2146893802 NTE_BAD_KEYSET)

The same card with an RSA key-certificate works like a charm

### Steps to reproduce

Run certutil -SCInfo with a card containing EC 256 private key and the related certificate.

dengert:
What  Windows version, and what OpenSC version?  

Can you post output of `certutil -v -scinfo` or at least the first part of it?

With the ATR, and look for a line after a certificate that looks like:  
`Card: Identity Device (NIST SP 800-73 [PIV])`  This would be for a PIV card using the Windows built in PIV code.

Look in registry `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\Calais\SmartCards`
find the entry  and look for name 80000001  and Value has the minidriver dll been used. 
This is the OpenSC `C:\Program Files\OpenSC Project\OpenSC\minidriver\opensc-minidriver.dll`

Also run `certutil -displayEccCurve` and see if you curve is one or both lists.

This might also help, its only 2 weeks old:
https://learn.microsoft.com/en-us/windows/win32/seccng/cng-named-elliptic-curves#remarks
 
  






Card: Identity Device (NIST SP 800-73 [PIV])

cyberneid:
I'm on Windows 11, OpenSC 0.25.1.
The registry is ok, the curve is nistP256 (1.2.840.10045.3.1.7) and it is supported by certutil.
RSA works like a charm.
The OS is in Italian, so,  this is the output

C:\Users\me\certutil -v -SCInfo
Gestione risorse smart card in esecuzione.
Stato corrente lettore/scheda:
Lettori: 1
  0: Circle CIR135 ICC 0
--- Lettore: Circle CIR135 ICC 0
--- Stato: SCARD_STATE_PRESENT | SCARD_STATE_INUSE
--- Stato: La scheda è condivisa da un processo.
---   Scheda: XXXXXXX
---    ATR:
        3b ff 18 00 ff 81 31 fe  4d 00 6b 04 04 b8 5b 01   ;.....1.M.k...[.
        f1 01 43 43 43 10 30 80  9d                        ..CCC.0..


=======================================================
Analisi della scheda nel lettore: Circle CIR135 ICC 0
[...]

PR #3262, opened 2024-11-15
minidriver.c - possible fix for  #3258

dengert:
EC pubkey should be 04|X|Y, but old code was setting X|Y

##### Checklist

- [ ] macOS tokend is tested


dengert:
It does not fix the problem.

PR #3269, opened 2024-11-21
sc_mem_secure_alloc: only lock requested length

### LINKED ISSUE #3267: sc_mem_secure_alloc/free leave unaligned allocations locked in memory
### Problem Description

Context: Fedora 41 / OpenSC 0.25.1 (but the affected code is completely unchanged on master).

`sc_mem_secure_alloc` and `sc_mem_secure_free` use the `mlock` and `munlock` syscalls to lock pages in memory (to mitigate the chances of secret data being written to swap, I assume). Because `mlock` and `munlock` operate at page granularity, `sc_mem_secure_alloc` makes an attempt to get an aligned block of memory from `calloc` by rounding up the allocation size to the system page size. I do not know if POSIX mandates that `calloc` *always* return a page-aligned address for a page-size allocation, but one situation where it empirically does not is inside the Chromium browser, which uses its own PartitionAlloc allocator instead of glibc's.

`sc_mem_secure_alloc` behaves badly in a couple ways when `calloc` returns a pointer is that isn't page-aligned:

1. Any data on the page at lower addresses than what `calloc` returned gets `mlock`ed along with the intended buffer (not terrible, but probably not what you intended)
2. Because `sc_mem_secure_alloc` passes the rounded-up-to-page-size length to `mlock`, the page *after* the intended page also gets locked, even if the intended buffer does not touch that page
3. Because `sc_mem_secure_free` passes the *not* rounded-up-to-page-size length to `munlock`, the range of pages covered by the `munlock` call may be one less than the range of pages previously locked by `mlock`

I found it difficult to wrap my head around the issue at first, so here are some logs I added that helped me understand:

```
mlock(0x32d403f53600, 4096)
<<< PAGE TO BE LOCKED: 32d403f53000
<<< PAGE TO BE LOCKED: 32d403f54000
munlock((nil), 0)
mlock(0x32d402cfc800, 4096)
<<< PAGE TO BE LOCKED: 32d402cfc000
<<< PAGE TO BE LOCKED: 32d402cfd000
munlock(0x32d402cfc800, 307)
<<< PAGE TO BE UNLOCKED: 32d402cfc000
munlock(0x32d403f53600, 51)
<<< PAGE TO BE UNLOCKED: 32d403f53000
```

As you can see, the calls to `mlock` lock page N *and* N+1, but the calls to `munlock` only unlock page N. This actually causes Chromium to crash because it may later try to `madvise(MADV_DONTNEED)` the page that OpenSC left locked. Passing `madvise(MADV_DONTNEED)` a page that is currently locked in memory fails with EINVAL, and Chromium dies here.

### Proposed Resolution

`calloc` is probably not the right function to *guarantee* you get a whole page to yourself that you can `mlock`/`munlock` as you need for your secure alloc function. I recommend calling `mmap`/`VirtualAlloc` directly, bypassing the libc allocator.

From my reading of the man pages that ship on Fedora as well as MSVC's documentation, the alignment behavior that you're currently relying on is not necessarily guaranteed. Even if the alignment behavior is reasonably common, if what you need is your own page, bypassing the libc allocator is a less brittle way of doing that.

If calling `mmap`/`VirtualAlloc` sounds OK to you, let me know and I'll send a patch.

If that's unacceptable for some reason, I found that C11's `aligned_alloc` also worked around the issue. (If C11 is too new, there are older functions in POSIX that do the same thing).

### Steps to reproduce

The log snippet posted above was captured while using a smart card for TLS client authentication to a web service in a debug build of the Chromium browser from the tip of the master branch. I built OpenSC 0.25.1 from source (that is the version Fedora 41 ships), but as noted above, the implementation of `sc_mem_secure_alloc` and `sc_mem_secure_free` are unchanged on master.

dgalling:
sc_mem_secure_alloc uses calloc to allocate memory, but then uses mlock to try to lock the allocated memory. calloc may not necessarily return a page-aligned pointer, even if we request a whole page of memory.

Instead of relying on the libc allocator behavior, we directly call mmap (VirtualAlloc on Windows) to ensure we get whole page(s) to ourselves for sc_mem_secure_alloc.

Fixes #3267

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -916,6 +916,7 @@ static void init_page_size()
 void *sc_mem_secure_alloc(size_t len)
 {
 	void *p;
+	size_t requested_len = len;
 
 	init_page_size();
 	if (page_size > 0) {
@@ -928,9 +929,9 @@ void *sc_mem_secure_alloc(size_t len)
 		return NULL;
 	}
 #ifdef _WIN32
-	VirtualLock(p, len);
+	VirtualLock(p, requested_len);
 #else
-	mlock(p, len);
+	mlock(p, requested_len);
 #endif
 
 	return p;


dgalling:
I'm still working on getting a Windows dev environment up to test there... (maybe there's CI to help confirm that it builds and runs there also?)

dgalling:
I confirmed that my branch builds & installs on Windows, and I followed the steps to test with both Windows's `certutil` and OpenSC's `pkcs11-tool`. Both seem to be able to log in to my card with my changes to the memory allocator.

frankmorgner:
May not be related, but is interesting anyway:
```
Fuzzing logs:
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1337
INFO: Loaded 1 modules   (40956 inline 8-bit counters): 40956 [0x5576d4393990, 0x5576d439d98c), 
INFO: Loaded 1 PC tables (40956 PCs): 40956 [0x5576d439d990,0x5576d443d950), 
INFO:    10608 files found in /github/workspace/cifuzz-corpus/fuzz_pkcs15init
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
INFO: seed corpus: files: 10608 min: 1b max: 5242857b total: 53961424b rss: 59Mb

==23==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 3 byte(s) in 1 object(s) allocated from:
    #0 0x5576d3c04dac in realloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:82:3
    #1 0x5576d3cbe752 in asn1_encode /src/opensc/src/libopensc/asn1.c:2056:16
    #2 0x5576d3cbe3d7 in sc_asn1_encode /src/opensc/src/libopensc/asn1.c:2078:9
    #3 0x5576d3d34f66 in sc_pkcs15_encode_pubkey_ec /src/opensc/src/libopensc/pkcs15-pubkey.c:688:2
    #4 0x5576d3d2e44b in sc_pkcs15_encode_pubkey /src/opensc/src/libopensc/pkcs15-pubkey.c:743:10
    #5 0x5576d3c5b7ee in sc_pkcs15init_generate_key /src/opensc/src/tests/fuzzing/../../../src/pkcs15init/pkcs15-lib.c:1647:6
    #6 0x5576d3c89654 in do_generate_key /src/opensc/src/tests/fuzzing/fuzz_pkcs15init.c:208:9
    #7 0x5576d3c8a836 in LLVMFuzzerTestOneInput /src/opensc/src/tests/fuzzing/fuzz_pkcs15init.c:338:5
    #8 0x5576d3af8bf0 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:13
    #9 0x5576d3af8415 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:7
2024-11-22 11:58:29,457 - root - INFO - Fuzzer: fuzz_pkcs15init. Detected bug.
2024-11-22 11:58:29,457 - root - INFO - Trying to reproduce crash using: /tmp/tmpkwdip3bi/leak-5aad2bb61ed81c4b8ff7334431927ed1254a23cc.
    #10 0x5576d3afa3a2 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile>>&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:829:7
    #11 0x5576d3afa692 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile>>&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:867:3
    #12 0x5576d3ae97cb in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:914:6
    #13 0x5576d3b14ba2 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #14 0x7fdf7b478082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 0702430aef5fa3dda43986563e9ffcc47efbd75e)

DEDUP_TOKEN: __interceptor_realloc--asn1_encode--sc_asn1_encode
SUMMARY: AddressSanitizer: 3 byte(s) leaked in 1 allocation(s).

INFO: a leak has been found in the initial corpus.

INFO: to ignore leaks on libFuzzer side use -detect_leaks=0.

MS: 0 ; base unit: 0000000000000000000000000000000000000000
artifact_prefix='/tmp/tmpkwdip3bi/'; Test unit written to /tmp/tmpkwdip3bi/leak-5aad2bb61ed81c4b8ff7334431927ed1254a23cc
stat::number_of_executed_units: 7775
stat::average_exec_per_sec:     1943
stat::new_units_added:          0
stat::slowest_unit_time_sec:    0
stat::peak_rss_mb:              84
2024-11-22 11:58:29,791 - root - INFO - Reproduce command returned: 77. Reproducible on /github/workspace/build-out/fuzz_pkcs15init.
2024-11-22 11:58:29,791 - root - INFO - Crash is reproducible.
2024-11-22 11:58:29,911 - root - INFO - Downloading latest build.
2024-11-22 11:58:29,913 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): storage.googleapis.com:443
2024-11-22 11:58:30,072 - urllib3.connectionpool - DEBUG - [https://storage.googleapis.com:443](https://storage.googleapis.com/) "GET /clusterfuzz-builds/opensc/opensc-address-202411220604.zip HTTP/1.1" 200 73018291
2024-11-22 11:58:32,020 - root - INFO - Done downloading latest build.
2024-11-22 11:58:32,020 - root - INFO - Trying to reproduce crash using: /tmp/tmpkwdip3bi/leak-5aad2bb61ed81c4b8ff7334431927ed1254a23cc.
2024-11-22 11:58:33,172 - root - INFO - Reproduce command returned: 0. Not reproducible on /github/workspace/cifuzz-prev-build/fuzz_pkcs15init.
2024-11-22 11:58:33,172 - root - INFO - The crash is not reproducible on previous build. Code change (pr/commit) introduced crash.
2024-11-22 11:58:33,172 - root - INFO - SAVING CRASH
2024-11-22 11:58:33,173 - root - INFO - NOT MINIMIZED
2024-11-22 11:58:33,173 - root - INFO - Deleting corpus and seed corpus of fuzz_pkcs15init to save disk.
2024-11-22 11:58:33,457 - root - INFO - Deleting fuzz target: fuzz_pkcs15init.
2024-11-22 11:58:33,461 - root - INFO - Done deleting.
2024-11-22 11:58:33,461 - root - INFO - Bug found. Stopping fuzzing.
2024-11-22 11:58:33,461 - root - INFO - Not uploading crashes because on OSS-Fuzz.
2024-11-22 11:58:33,461 - root - INFO - Bug found.
```

dgalling:
Ah, I bet those failures are related: perhaps a case of a buffer allocated by `sc_mem_secure_alloc` later being passed to `realloc`. This would have not leaked when `sc_mem_secure_alloc` was backed by `calloc`, although it may have been vulnerable code in the event that `realloc` copied sensitive data away from locked pages.

I will investigate today, thanks for having good CI.

dgalling:
This was challenging to track down--I had to write a custom heap debugger!

Similar to my suspicion from earlier, the pkcs15 library is allocating a buffer using a direct call to `realloc` and then later passing that buffer to `sc_mem_secure_free`. This would have worked okay before, although it would have been subtly incorrect to `munlock` a page that you hadn't previously `mlock`ed.

Here's the output from my heap debugger:

```
Tried to sc_mem_secure_free a buffer allocated by libc: heap entry { addr: 0x6238fd06b390, len: 0x6a, allocated at: 
/usr/local/lib/libopensc.so.12(+0x23987)[0x70ed1e823987]
/usr/local/lib/libopensc.so.12(+0x23d6f)[0x70ed1e823d6f]
/usr/local/lib/libopensc.so.12(+0x2be5e)[0x70ed1e82be5e]
/usr/local/lib/libopensc.so.12(sc_pkcs15_encode_pubkey_rsa+0xdc)[0x70ed1e84c62c]
/usr/local/lib/libopensc.so.12(+0x12a11f)[0x70ed1e92a11f]
/usr/local/lib/libopensc.so.12(sc_pkcs15init_store_private_key+0x328)[0x70ed1e932088]
pkcs15-init(+0x8e3e)[0x6238fc2e9e3e]
pkcs15-init(+0x61b7)[0x6238fc2e71b7]
/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x70ed1e42a1ca]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x70ed1e42a28b]
pkcs15-init(+0x68a5)[0x6238fc2e78a5]
}
Attempted to free at: 
/usr/local/lib/libopensc.so.12(sc_mem_secure_free+0xa9)[0x70ed1e823f68]
/usr/local/lib/libopensc.so.12(+0x42600)[0x70ed1e842600]
/usr/local/lib/libopensc.so.12(sc_pkcs15init_store_private_key+0x3c2)[0x70ed1e932122]
pkcs15-init(+0x8e3e)[0x6238fc2e9e3e]
pkcs15-init(+0x61b7)[0x6238fc2e71b7]
/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x70ed1e42a1ca]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x70ed1e42a28b]
pkcs15-init(+0x68a5)[0x6238fc2e78a5]
```

On the allocation path, part of storing an RSA private key involves using the ASN1-encoding functions intended for RSA public keys. The ASN1 encoder allocates memory using a call to `realloc`, but because the overall PKCS15 object is marked as a private key, its memory is released with a call to `sc_mem_secure_free`.

It may be difficult to pull apart the uses of the standard libc heap and the `sc_mem_secure` functions, because when it was all backed by the libc heap anyway, it would have been mostly ok (although subtly incorrect as noted above) to allocate from one and free by the other.

One option I noted in #3267 was to use `alligned_alloc` to keep using the libc heap while being more explicit about the alignment requirements to use `mlock`. Doing that instead of `mmap` would fix the failures here, but it would not fix the potential imbalance in calls to `mlock` and `munlock`... @frankmorgner do you have any preference for a direction here?

dgalling:
On further thought, we could go in the complete other direction and ignore the alignment: the code on the master branch today that attempts to align the allocation to a page boundary does not necessarily work, and that's okay. If instead we change the calls to `mlock` and `munlock` to cover the same ranges of addresses (they do not on master) that should sufficiently clean up the process memory state to not crash a memory allocator that tries to discard pages with `MADV_DONTNEED`.

frankmorgner:
I cannot decide for some strategy right now, as I need to check dig into the documentation first.

But you're right, I have identified the problem which is causing the memory allocation problem in pkcs15init. However, I'm still thinking about what should be the best solution for this. In any case, I'll open a separate pull request.

frankmorgner (approved):
I didn't realize that your alternative solution would be as simple as this. If it works for you and if you like, we can merge this as solution for now.

However, if possible, please open an issue (or a seperate pull request) with what you had implemented previously to fix the caluclation of the page size. Thank you!

src/libopensc/sc.c
```diff
@@ -897,40 +896,22 @@ int _sc_parse_atr(sc_reader_t *reader)
...
-		len = pages * page_size;
-	}
-
-	p = calloc(1, len);
-	if (p == NULL) {
-		return NULL;
-	}
 #ifdef _WIN32
-	VirtualLock(p, len);
+	p = VirtualAlloc(NULL, len, MEM_COMMIT, PAGE_READWRITE);
```
frankmorgner: The [example code from MSDN](https://learn.microsoft.com/en-us/windows/win32/Memory/memory-protection-constants) still uses the pagesize from sysinfo for the allocation. Aren't you  missing that part in your code?

frankmorgner: you're right, sorry for the noise and thanks for the clarification

frankmorgner:
I reviewed your original fix and this looks good. I think we can merge https://github.com/OpenSC/OpenSC/pull/3281 directly.

dgalling:
We should not merge #3281 as-is because of the heap corruption issues discussed above on this PR. 

I recommend merging this PR because it fixes the Chromium crashes (and is also quite easy to backport to older branches), and then you/we can work to untangle the allocation paths to make merging #3281 possible.

frankmorgner:
The bad free above should be fixed with https://github.com/OpenSC/OpenSC/pull/3282 - together with https://github.com/OpenSC/OpenSC/pull/3281 the memory problems hopefully go away.

dgalling:
Agreed, #3282 looks like a good change.

I don't expect to have time to test anything until the weekend, but this weekend I can try running both #3281 and #3282 together under the custom heap debugger to see if that prevents the heap corruption issues we were seeing. Valgrind should also catch most of the issues if you wanna just press ahead.

Jakuje (approved):
this looks reasonable

frankmorgner:
fixed with https://github.com/OpenSC/OpenSC/pull/3281

Issue #3264, opened 2024-11-18
Minidriver and Usage


dengert:
### Problem Description
Both #3244 and #3258 report problems with certificates with multiple keyUsage bits set when run on Windows using minidriver.

#3244 has two RSA certificate one with "Digital Signature" and the other with  "Digital Signature, Key Encipherment, Data Encipherment' on a PIV type device. 
 
#3258 has a problem with one certificate with  "Digital Signature, Key Encipherment" using an external OpenSC driver.

OpenSC will map certificate `keyUsage`  using `sc_pkcs15_map_usage` which is called from `  ./pkcs15-cac.c` and  `./pkcs15-idprime.c` or most other driver do their own mapping to `SC_PKCS15_PRKEY_USAGE_*` flags.

`minidriver.c` defines `md_pkcs15_container` from  pkcs15 objects, then in 
https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L1933-L1952
will assign either  `AT_KEYEXCHANGE` or `AT_SIGNATURE`  but not both.

So if any  pkcs15 `USAGE_ANY_DECIPHER` bits are set there will be no  `AT_SIGNATURE` for this key.

Questions and comments:
- Microsoft requires there be a user auth type key one with `AT_SIGNATURE`.
- Are the certificates is question generated by some CA with both sign and encryption flags?
- If there is only one key and certificate, should it be allowed to be used for both sign and encryption?
- Microsoft  requires there be a user auth type key, 
- `pkcs15-piv.c` sets to many flags, and the lines with `/* extra */` should be modified. 
- Should `minidriver.c` parse the certificate for keyUsage flags (it has access to the certificate) and if needed return two containers one for  `AT_KEYEXCHANGE` and one for `AT_SIGNATURE` ?
- There is a large section of code in `minidriver.c` within "#if 0...#endif" 

### Proposed Resolution

- https://github.com/dengert/OpenSC/tree/Minidriver-2 is one attempt to not add the `/* extra */` flags and do something if only one key is on the card. 
- If user has set conflicting `keyUsage` flags, can they get new certificate(s) so same key is not used for both.

Anyone want to take on this issue?
 
### Logs
 See #3244 and #3258

dengert:
#3258 is using EC  key with  "Digital Signature, Key Encipherment" 
EC does not support Key Encipherment but rather "Key Agreement"  using ECDH.
Best practices are to use two keys rather then one.

[...]

PR #3270, opened 2024-11-21
Minidriver tests for creation of AT_SIGNATURE and AT_KEYEXCHANGE containers.

dengert:
minidriver.c - Testing logic of minidriver usage flags and creation of AT_SIGNATURE and AT_KEYEXCHANGE containers.
    
Current code assumes key will only be used for AT_SIGNATURE  **OR** AT_KEYEXCHANGE  but not both. But minidriver spec allow this as `CONTAINER_MAP_RECORD` allows both with wSigKeySizeBits and wSigKeySizeBits

But in issue #3264 we see two examples of cards with certificates with keyUsage that imply  both AT_SIGNATURE  **AND** AT_KEYEXCHANGE for the same key. And also misuse "Key Encipherment" with ECC. 

`pkcs15-piv.c is also modified as it was setting extra PKCS15 usage bits, compounding the problem.

To test this, `opensc.conf` needs something like this with your ATR:
```    
    card_atr 3b:f0:0d:ca:fe {
    	# a key can have both  SIGN and EXCHANGE
    	md_sign_and_exchange = true;
    
    	# AN ECC cert which include "Key Encipherment" is not valid,
    	# ECC keys should have "Key Agreement". To accept either:
    	md_ec_usage_accept_decrypt = true;
   }
```` 
So several test can be run using the same minidriver.

Even if we get this part to work, it is not clear what else Windows is doing to associate User Auth, Signature and Key Agreement keys/certs. For example it may require a match Email extension in the Signature and Key Agreement keys/certs.
And may require the correct KeyUsage bits with ECC keys. 
 
I am using the CI to build the Windows MSI files for my testing and hope I can get others to test as well. 

##### Checklist

- [ ] macOS tokend is tested

### CODE REVIEW

**Comment by frankmorgner** [comment]
this change shows the same issue as described in https://github.com/OpenSC/OpenSC/issues/3244

**Comment by dengert** [comment]
@frankmorgner  thanks for the input, but can you run it and get a md.log with log level 15?
 You should see a number of lines like this:
`flag_name:md_ec_usage_accept_decrypt:TRUE` and `flag_name:md_sign_and_exchange:TRUE` 

If the lines shows FALSE, it did not find  `card_atr` section or the `ec_usage_accept_decrypt = true;` or `md_sign_and_exchange = true;` lines. 

If you don't see any of these it not not the using the modifed minidriver. 

A diff certutil command of the microsoft anf OpenSC drivers in original showed:

Are you still getting " Test NICHT erfolgreich: Das Zertifikat und der private Schlüssel für die Entschlüsselung wurden nicht gefunden 0x8009200c (-2146885620 CRYPT_E_NO_DECRYPT_CERT)

Are you still seeing these?

**Comment by dengert** [comment]
One other possible issue when switching between drivers, is that Containers are stored in the cert store, differently by the OpenSC PIV driver and the Microsoft driver. So if you do not clear out the cert store of certificates when switching, a containerID may not be found by the other driver. Microsoft may be looking up the public key from the certificate to get a containerID  look to see if the card is the one it is testing. 

OpenSC takes the GUID from the CHUID. Microsoft format:    5fcdc2e3-ee6c-4605-991b-7f43551e071c
to use as the card serial number.
 
For a certificate OpenSC sets the container (from you card): Schl�sselcontainer = 01c2cd5f6cee0546991b7f43551e071c
where the pkcs15/pkcs11 id  replaces the first byte. In this case  01. 

where as Microsoft assigns  Schl�sselcontainer = 5fcdc2e3-ee6c-4605-991b-7f43555fc105
and replaced the last 3 bytes with the  `5fc105`  which is from NIST sp800-73-4  Table 3: "Object Identifiers of the PIV Data Objects for Interoperable Use"  "X.509 Certificate for PIV Authentication|2.16.840.1.101.3.7.2.1.1|'5FC105'"

When OpenSC piv driver was written, every one was expected to have a FASC-N. Overlaying the first byte as was chosen to avoid a duplicate ContainerIDs  since employee number as the last bytes. If two employees used the same computer and differed in only the last byte of the FASC-N duplicate ContainerIDs were possible. 

Since as far as I know no one really use the OpenSC minidriver for PIV cards, as Microsoft and card vendors provide their
own minidrivers, and the OpenSC install does not add registry entries for the PIV driver. It would not be too hard to change `card-piv.c` to do Microsoft does.  

Have you tried using the Yubikey minidriver?

**Comment by dengert** [comment]
https://github.com/ugochirico

The minidriver does not implement CCP_ASSOCIATED_ECDH_KEY:
"In this situation, pbData points to a container index for the ECDH key of an ECDSA key container. This creates an association between an ECDSA key and an ECDH key for logon."

It should be added here:
https://github.com/OpenSC/OpenSC/blob/master/src/minidriver/minidriver.c#L6296-L6297

And at least it should the  log the  wszProperty before the return with `SCARD_E_INVALID_PARAMETER` which maybe the 

This may be the problem with  in #3258  0x80090016 (-2146893802 NTE_BAD_KEYSET)

**Comment by spurglegooge** [comment]
A data point for the `CCP_ASSOCIATED_ECDH_KEY` comment above, from a PIV card rather than a
test harness.

Card is OpenFIPS201 v2.0 on a JavaCard, 9A holding EC P-384, Windows 11 Pro 10.0.26200, OpenSC
minidriver ATR-registered because the inbox PIV driver cannot read the P-384 keyset.

**Symptom.** The logon screen showed "no valid certificates" and never offered the card, while
`certutil -scinfo` reported the chain validates — that command does not check the ECDH
association, so everything looked correct everywhere except the one place that mattered. The
container layout the minidriver exposed at the time:

```
0  9A  EC P-384   signature       (the logon key)
1  9C  EC P-256   signature
2  9D  RSA-2048   key-exchange
3  9E  EC P-256   signature
4  82  EC P-384   key-exchange    (cbKeyExPublicKey=104, cbSigPublicKey=0)
```

**What fixed it.** Provisioning an EC P-384 key-agreement key and certificate into retired slot
82. Online smart-card logon worked immediately afterwards. The RSA-2048 key-exchange container
in 9D was left in place throughout, and 9A was not re-keyed.

I cannot tell you which container Windows associated, or whether it consulted
`CCP_ASSOCIATED_ECDH_KEY` first 

**On your second suggestion, which I would like to second.**
`CardGetContainerProperty` returns `SCARD_E_INVALID_PARAMETER` for unhandled properties without
logging `wszProperty`, so `md.log` gives no indication that Windows asked for the association at
all. From the outside the failure is indistinguishable from a certificate or trust problem,
which is where we spent most of our time. Logging the requested property before the return
would have pointed straight at this.

**Comment by dengert** [comment]
@spurglegooge @frankmorgner
 I forgot all about this, basically because all the PIV and "PIV like" card vendors provide their own minidriver, so there was no need for using the OpenSC PIV card driver with the OpenSC minidriver. (I don't like having to deal with the minidriver.) 
With you loading a PIV applet on a card you have become a private "PIV-like" card vendor.   

But what both of you may be seeing, is certificates with keyUsage that do not follow the usage that NIST expects for the 9A, 9C and 9D keys.  And mixing P-256 and P-384 on the same card without having a key agreement key with the same curve as the 9A key. 

The OpenSC pkcs15-piv.c was modified to map any keyUsage (which is not NIST approved) for any key for "PIV-like" cards which can lead to problems like this.  (9A is P-384, and 9D is RSA.  so no P-384 to use for key agreement.)

By adding the 82 key which NIST says is the first of the "retired key management keys"  Windows found a key to use for key agreement with the 9A  keu. 

@spurglegooge @frankmorgner 
I am not interested in debugging this problem, but I can rebuild the Minidriver-2 branch. One of the artifacts is the MSI files to install on Windows and if you see some other changes I can add to the branch. Or one of you can take over this branch or submit your own PR.

**Comment by frankmorgner** [comment]
I think the isolated change mentioned in https://github.com/OpenSC/OpenSC/issues/3244#issuecomment-5338053770 is needed, at least for me. I don't think all the configuratbility is required.

**Comment by dengert** [comment]
@spurglegooge @frankmorgner @Jakuje 
I rebased this PR on master yesterday so @spurglegooge could use the MSI files to test on his system.
Look at the below list of "successful checks" for [Windows / build (x64, Release, windows-2025)](https://github.com/OpenSC/OpenSC/actions/runs/34871849979/job/104069408717?pr=3270)
 and [Windows / build (x86, Release, windows-2025) (pull_request)](https://github.com/OpenSC/OpenSC/actions/runs/34871849979/job/104069408525?pr=3270)

under each look for the "Archive artifacts" click on it, and at the bottom (This is the x64:)
Artifact download URL: https://github.com/OpenSC/OpenSC/actions/runs/34871849979/artifacts/10359114103
This is x86: 
Artifact download URL: https://github.com/OpenSC/OpenSC/actions/runs/34871849979/artifacts/10360045436

The URL will change after ever change to the PR and the "Windows / build..." will point to a new version

The zip files have the MSI. (I have not tested them.)

But to me, the main problem is the **cards @spurglegooge is using are not provisioned correctly.** He has already shown
by adding a "82" P-384 key and certificate, Windows will find it and work correctly. And the problem is Windows expects to find an EC key with the same curve as the signing key it is using (either 9A or 9C). If it does not find one it falls back to using no hash or a hash this is 32 bytes long. The PIV applet he is using says this is an error. Which meets the NIST standards. 

PIV specifically defines 3 keys, auth key signing key and key agreement to be used for CMS type messages i.e. encrypted email.   

I don't believe we have seen what is in the 32 bytes. Hopeful it looks random and uses all 32 bits.     

I don't like the idea of zero padding on the right as this just covers up what maybe a security issue if what is being sent is not a hash.

**Comment by spurglegooge** [comment]
@dengert — per the suggestion in https://github.com/OpenSC/OpenSC/pull/3270#issuecomment-2495232730, we implemented `CCP_ASSOCIATED_ECDH_KEY`: a handler in `CardGetContainerProperty` (one commit on master `bfbf27b6a6b1`), tested on two card layouts, Windows 11 Pro 10.0.26200, logging every container-property request by name.

- **Finding:** Windows requests the property only when a card has more than one key-exchange container. On a card with a single EC key-exchange container it never asked — 48 property queries, all `PIN Identifier`. On a card with an RSA key-exchange container at index 2 and an EC P-384 one at index 4, it asked, the handler returned index 4, and Windows used it:
  ```
  CardGetContainerProperty bContainerIndex=0, wszProperty=Associated ECDH Key, cbData=1
  associated ECDH key: container 4, curve matches (384 bits)
  ```
- **Important caveat:** on this Windows build, the fallback with the property unimplemented already picks the EC container correctly. So this doesn't fix a live failure — it makes the choice deterministic instead of dependent on undocumented behavior that may differ on other builds.
- **Ask:** happy to open the handler as a PR, or as a commit on this one — say which. Either way, worth logging the requested property name before the `SCARD_E_INVALID_PARAMETER` return, so the next person can tell "Windows never asked" from "OpenSC refused to answer" without building a custom DLL.

This is separate from #3817 and governs which container the cached-credential ECDH uses; it doesn't change the 32-byte signing challenge, which is refused independently. Full method and both cards' numbers in the attached file.

[40-ATTACHMENT-ecdh-association-detail.md](https://github.com/user-attachments/files/32265329/40-ATTACHMENT-ecdh-association-detail.md)

**Comment by dengert** [comment]
I don't recall how you are provisioning the cards. On #3817 comments you talked about provisioned a fresh card which might imply you may change just one key/cert on an existing card.   

NIST never defined a "serial number" for a PIV card.  But serial numbers are used in OpenSC. And the minidriver needs some info to distinguish between cards and creates containerIDs. 

My concern is Microsoft may still be using some ContainerIDs left around from one card or previous version of a certificate (with different key) and getting confused. `certutil .exe -scinfo` can show ContainerIDs or ask you to insert a specific card. I don't know how to clear out all ContainerIDs. 

Note NIST PIV does not store a copy of public key, but expect the SPKI of the certificate to provide it. The public key is returned when a key is generated on the card and software like piv-tool use it in a CSR to get it into a certificate.

Have you ever tried a new card with 9A, 9C, 9D using P-384 on a machine that has never seen a PIV card before? 

Both OpenSC and  Microsoft builtin PIV driver create ContainerIDs  for the card and certificates using the PIV CHUID. Do you create a CHUID object on the card? The default is all zeros, which works if only one card is ever used on a machine.    
 https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L4365-L4447
originally there was only the FASC-N to use and OpenSC and Microsoft used it, but in slightly different ways to derive ContainerIDs from the FASC-N.  (I got to visit with Microsoft developers while at an IETF meeting in 2005 or so. I told them what we were doing for linux systems and was asking if they were implementing PIV support for Windows. They never said yes but their smiles implied they where working on it.)      

NIST left it up to vendors on how to provision a PIV card. But did define how to generate a key on the card and how to write objects to a card, but not how to write a key to the card, or move a key management key to the list of retired keys.  Many PIV-like cards can use `yubico-piv-tool` for provisioning.

As I always have said, the OpenSC piv-tool was for developers and only on cards which supported the basic commands defined by NIST including a 9B key and did not require writing the other keys.   

Note NIST PIV does not store a copy of public key, but expect the SPKI of the certificate to provide it. The public key is returned when a key is generated on the card and `piv-tool` can save it as a file. `card-piv.c` while looking for certificates on the card and if none, will check env for the saved public key i.e. the SPKI so key can be used to sign a CSR to get a certificate which is then load on card. Windows does something similar saving  SPKI on disc when a key is generated. `yubico-piv-tool` does save a copy of public key on card. There is #3822 which will reads these but only usable on Yubico-piv-like cards.     

If we really have to change OpenSC there is a way for to check if called by minidriver, i.e. "cardmod` and the card type is PIV thus limiting the change which may have consequences for other card drivers.

**Comment by spurglegooge** [comment]
@dengert — good questions;

**How the cards are provisioned.** card2 is a genuinely blank NXP chip provisioned from
scratch, not a one-key edit of an existing card. It has its own CHUID/GUID
(`A04F5343731001F9AA6E34537BEE6FC1`), distinct from the original card, so the two do not derive
the same ContainerIDs. 9A/9C/9E and the 9D key-agreement key were all generated on the card in
this one provisioning pass.

**On the ContainerID / stale-cache concern.** It is a real Windows-side worry and worth
raising, but it doesnt explain the failure this PR fixes, because the failure reproduces with
Windows entirely out of the loop. Raw `GENERAL AUTHENTICATE` APDUs via `opensc-tool -s` to 9A
holding P-384 (posted on #3817, repeated here):

```
20 bytes -> 6A80      48 bytes -> 9000
32 bytes -> 6A80      64 bytes -> 9000   (longer than the 48-byte block, still signed)
33 bytes -> 6A80      65 bytes -> 6A80
```

No minidriver, no ContainerID, no CHUID, no credential cache — just the card refusing a
32-byte input to a P-384 key and signing a 48-byte one. Whatever ContainerID state Windows
carries, it selects *which* container to talk to; it cannot change what the card returns to a
raw APDU. That is the behaviour the patch accommodates.

I'm happy to run `certutil -scinfo` and, if useful, repeat the Windows offline test on a
machine that has never seen a PIV card — but I'd offer the raw-APDU result is a valid
isolation, since it removes Windows from the question rather than resetting its state.

**Your last paragraph — scoping the change.** This is the part I'd like to build on:

> If we really have to change OpenSC there is a way to check if called by minidriver (cardmod)
> and the card type is PIV, thus limiting the change which may have consequences for other card
> drivers.

Yes — that addresses @Jakuje's concern about `pkcs15-sec.c` being common to every driver, and
I'd rather gate it narrowly than pad EC input globally. I can rework the discriminator to fire
only when the card is PIV *and* the call came through the minidriver, and keep the
non-repudiation exclusion inside that so 9C is still never padded (auth key yes, signing key
no). Point me at the cardmod signal you have in mind (I see `card->type` for the PIV side; what
do you use to detect the cardmod caller?) and I'll put it on #3817 as a replacement for the
current `NONREPUDIATION`-only clause. An `opensc.conf` toggle is the other option if you'd
prefer the behaviour off by default — either works for me; your call on which is more
acceptable upstream.

**Comment by dengert** [comment]
So what you are asking of OpenSC is make a change to accommodate a Microsoft bug when using P-384 on a card that enforces the NIST standards of ECDSA to used a hash with the same size as the field size the EC P-384. even though other PIV cards may not test for this NIST standard.

>  It has its own CHUID/GUID
(A04F5343731001F9AA6E34537BEE6FC1), distinct from the original card, so the two do not derive
the same ContainerIDs. 9A/9C/9E and the 9D

I would like to see the full CHUID, with the  9A/9C/9E and the 9D all using P-384.

I would also like to see the response to SELECT AID as this has info about the applet that would be useful in identifying the applet as https://github.com/makinako/OpenFIPS201/ for any future mods. 

Also the ATR  and serial as returned by `opensc-tool -a --serial`

**Comment by spurglegooge** [comment]
@dengert — data below, and one correction to the card's curves first.

**Curves.** This card is not all-P-384: **9A and 9D are P-384; 9C and 9E are P-256.** 9C/9E
are P-256 deliberately (a separate platform constraint); this is the
layout we actually ran. 9A (the logon/auth key this PR is about) and 9D (the key-agreement
key) are both P-384.

**ATR + serial** (`opensc-tool -a --serial`):
```
ATR:    3b:fa:18:00:00:91:01:31:fe:45:4a:33:52:33:31:30:2d:33:35:35:ff
serial: D0 42 10 D8 21 08 6C 10 84 21 0D 83 68 58 21 08 42 10 84 21 C8 42 10 C3 EB
```
(the ATR tail `4A 33 52 33 31 30 2D 33 35 35` is ASCII `J3R310-355`, the NXP JCOP part. The
"serial" OpenSC prints is the CHUID FASC-N — see below.)

**SELECT PIV AID** (`00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00`) — the application property
template identifying the applet:
```
61 6F 4F 0B A0 00 00 03 08 00 00 10 00 01 00 79
07 4F 05 A0 00 00 03 08 50 0B 4F 70 65 6E 46 49
50 53 32 30 31 5F 50 49 68 74 74 70 3A 2F 2F 6E
76 6C 70 75 62 73 2E 6E 69 73 74 2E 67 6F 76 2F
6E 69 73 74 70 75 62 73 2F 53 70 65 63 69 61 6C
50 75 62 6C 69 63 61 74 69 6F 6E 73 2F 4E 49 53
54 2E 53 50 2E 38 30 30 2D 37 33 2D 34 2E 70 64
66
```
Decoded: AID `A0 00 00 03 08 00 00 10 00 01 00`; application label (tag 50) **`OpenFIPS201`**;
URL (tag 5F50) `http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf`. So
the applet is [makinako/OpenFIPS201](https://github.com/makinako/OpenFIPS201), as you expected.

**Full CHUID** (GET DATA `5FC102`):
```
53 39 30 19 D0 42 10 D8 21 08 6C 10 84 21 0D 83
68 58 21 08 42 10 84 21 C8 42 10 C3 EB 34 10 A0
4F 53 43 73 10 01 F9 AA 6E 34 53 7B EE 6F C1 35
08 32 30 33 36 30 39 31 32 3E 00
```
Decoded (`53` wrapper, len 0x39):
- FASC-N (tag 30, 25 bytes): `D0 42 10 D8 21 08 6C 10 84 21 0D 83 68 58 21 08 42 10 84 21 C8 42 10 C3 EB` — populated, **not the all-zero default**.
- GUID (tag 34, 16 bytes): `A0 4F 53 43 73 10 01 F9 AA 6E 34 53 7B EE 6F C1` = `A04F5343731001F9AA6E34537BEE6FC1`.
- Expiration (tag 35): `20360912` (2036-09-12).
- Issuer signature (tag 3E): empty — the CHUID is unsigned, as expected for a lab card.

On 800-73-4 cards the GUID (tag 34) is what drives the minidriver's ContainerIDs, and it is
populated and card-specific here, so this card does not fall into the all-zeros collision case
you flagged.

**On the framing** — "a change to accommodate a Microsoft bug … even though other PIV cards
may not test for this NIST standard." That is good justification to gate it
on the minidriver caller (`ctx->app_name == "cardmod"`) **and** a PIV card type, keeping the
non-repudiation exclusion so 9C is never touched. Then a card/driver that isn't PIV-via-Windows
sees no change at all — the accommodation is confined to precisely the case that needs it, and
a PIV card that *doesn't* enforce the pairing (your IDEMIA trace, for instance) is unaffected
because it already signs the 32 bytes. If you'd rather it be off unless explicitly enabled, an
`opensc.conf` toggle does the same job; your call which is more acceptable upstream and I'll
send it as a single commit on #3817.

**Comment by dengert** [comment]
@spurglegooge 

OK, the  Application 16 (11 byte) OpenFIPS201 can be used to to identify this applet on any card. 

Here is a proposed change to `card-piv.c`  and cards.h` to do the padding. It compiles, but not tested. 
Please check if the padding is correct.
Also look at the `if`  is restrictive enough or to much.

It defines the OpenFIPS201 as special PIV card type based on response to SELECT AID 
This applet may appearon other cards so they may have the same problem.
so It adds the `card_issue`   `CI_ENFORCES_EC_HASH_EQ_KEYSIZE`

[openfips2001.patch.txt](https://github.com/user-attachments/files/32345587/openfips2001.patch.txt)

**Comment by spurglegooge** [comment]
@dengert — This is a great solution: identifying the applet from the SELECT AID label
and hanging a `card_issue` flag off it targets exactly the cards that enforce the pairing,
instead of every PIV card. Thanks for writing it. Two things —  one is the
answer to the "restrictive enough?" question , the other is a small typo — plus a couple of minor notes.

**1. "Is the `if` restrictive enough?" — exclude the Digital Signature key (9C).**

Right now it pads any EC key when `indatalen < nLen`. That's correct for the *authentication*
keys (9A, 9E), where the input is a challenge, not a hash. But it also pads the *Digital
Signature* key (9C), where the input genuinely is a message hash and the size-matches-curve
guard is legitimate — padding a 32-byte SHA-256 hash up to 48 so a P-384 9C signs it defeats
the guard exactly where it should apply.

`priv->key_ref` is saved from `set_security_env`, so the driver can tell them apart. Gate on
the authentication keys only:

```c
if (priv->card_issues & CI_ENFORCES_EC_HASH_EQ_KEYSIZE &&
        !strcmp(card->ctx->app_name, "cardmod") &&
        (priv->key_ref == 0x9A || priv->key_ref == 0x9E) &&   /* auth keys, not 9C */
        (indatalen < nLen)) {
    padlen = nLen - indatalen;
}
```

(This is the card-native version of the `NONREPUDIATION` exclusion I had in `pkcs15-sec.c`;
your `key_ref` check is cleaner. On our specific cards 9C is P-256 so it never triggers either
way, but it's worth gating for a P-384 signing key.)

**2. Typo: `memcpy` second arg should be indata**

```diff
data = calloc(1, indatalen + padlen);
...
- memcpy(data + padlen, data, indatalen);  
+ memcpy(data + padlen, indata, indatalen);
```

**Minor, take or leave:**

- `nLen` is computed twice now (once in the new block, once in the existing one) — harmless,
  could fold.
- The `calloc`/`memcpy` runs for RSA too (padlen 0), an extra alloc+copy on every RSA
  signature. Could wrap it in `if (padlen)` and otherwise sign from `indata` directly, but then
  the `free(data)` at `err:` has to know not to free `indata` — your unconditional-copy version
  is simpler and fine once the source is fixed.
- I can vouch for `CI_ENFORCES_EC_HASH_EQ_KEYSIZE` on OpenFIPS201; I haven't verified the other
  two flags you set for it (`CI_OTHER_AID_LOSE_STATE`, `CI_LEAKS_FILE_NOT_FOUND`) — those are
  your call.

Happy to test the corrected version on hardware 
(card2: P-384 9A signs the 32-byte challenge, 9C P-256 unaffected)

dengert:
A few  comments:

You CHUID decodes to: 
Expiration: 20360912
FASCN: ;0000=0000=000000=0=1=0000000000300001?:
FASCN-HEX: D04210D821086C1084210D836858210842108421C84210C3EB
GUID: a04f5343731001f9aa6e34537bee6fc1

The serial number:  D0 42 10 D8 21 08 6C 10 84 21 0D 83 68 58 21 08 42 10 84 21 C8 42 10 C3 EB
is being based on the FASCN, not the GUID.

This could be a problem if multiple cards have the same FASC-N

Also see:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L4397-L4437

You FASC-N is present and does not start with ;9999 which was reserved for non government issued cards.  
Since you have a "gov" FASC-N The keyUsage in certificates is ignored,and the flags builtin to the driver are used. 

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c#L681-L712
Line 709:  /* US gov issued PIVs have CHUID with a FASCN that does not start with 9999 */ 

Also related:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c#L116-L233

The PIV driver was written to follow NST 800-73-* .  Extensions to use with non gov cards were added to use the keyUsage from certificates as a way to use the 20 "retired" key/certs for any purpose.  

Helpful to build and decode a CHUID:
[chuid.c.txt](https://github.com/user-attachments/files/32395225/chuid.c.txt)

dengert:
The OpenFIPS201  wiki says  provisioning can be done by multiple methods include Yubico
It is not clear how much of the Yubikey docs are followed. 

Can you run `opensc-tool -s "00FD000003" which reads a Yubico like version number. Some cards follow Yubico some not and  return their own version number.  What does   OpenFIPS201?
https://docs.yubico.com/yesdk/users-manual/application-piv/apdu/version.html
https://docs.yubico.com/yesdk/users-manual/application-piv/commands.html

spurglegooge:
@dengert — answers to both, and a correction to something I wrote earlier.

**GET VERSION.** `00 FD 00 00 03` returns `6D 00` (INS not supported) on both of our
OpenFIPS201 cards. The applet does not implement the Yubico version command, so anything
keyed off that response will see "not a Yubico-style card" here.

**FASC-N / serial — you are right, and I was wrong earlier.** I described the FASC-N in the
CHUID above as "populated, not the all-zero default". Decoding it (5-bit BCD) gives
`;0000=0000=000000=0=1=0000000000300001?` — it *is* the all-zero default that pivy's `init`
writes. I misread the encoded bytes. Sorry for missing that.

Consequences, checked on hardware:

- Both cards (card1 and card2) carry that identical FASC-N, so `opensc-tool --serial` and
  `pkcs15-tool --list-info` report the **same serial** for both:
  `d04210d821086c1084210d836858210842108421c84210c3eb`. Their GUIDs differ
  (`A04F5343…6FC1` vs `E603124B…B240`).
- Neither FASC-N starts with `9999`, so per `pkcs15-piv.c` the driver treats them as
  government-issued and uses its built-in slot flags rather than the certificates' keyUsage.

Both are our provisioning's doing, not the driver's. We have fixed card2 (the card in this
thread) and left card0/card1 as they are for now, so those two still collide with each other;
that is pending on our side.

**card2, before / after** (one `PUT DATA 5FC102`; tag 30 replaced, tags 34/35/3E preserved
byte-for-byte; keys and certificates untouched; every APDU `90 00`):

```
FASC-N before  D0 42 10 D8 21 08 6C 10 84 21 0D 83 68 58 21 08 42 10 84 21 C8 42 10 C3 EB
               ;0000=0000=000000=0=1=0000000000300001?
FASC-N after   D4 E7 39 D8 21 0C 2C 10 84 21 45 A1 68 58 21 08 42 10 84 28 84 E7 39 C3 E2
               ;9999=0001=000002=1=1=0000000002199991?
GUID           A04F5343731001F9AA6E34537BEE6FC1   (unchanged)

opensc-tool --serial   before  D0 42 10 D8 ... C3 EB   (25 bytes, the FASC-N)
                       after   A0 4F 53 43 73 10 01 F9 AA 6E 34 53 7B EE 6F C1   (16 bytes, the GUID)
```

So, reading `card-piv.c:4418-4430` and confirmed on the card: with the `;9999` prefix the
serial becomes the GUID, and that is where the uniqueness comes from, not the credential
number. card2's serial is now `a04f5343731001f9aa6e34537bee6fc1` and we will not touch its
CHUID again, so it is safe to reference.

Two consequences we checked, since they matter for #3817 and for anyone doing the same:

- **keyUsage is now authoritative** (`pkcs15-piv.c:685/831`). `pkcs15-tool --list-keys` on
  card2 after the write: 9A sign, 9C sign, 9D derive, 9E sign, matching the certificates'
  keyUsage exactly. card2's 9C carries `digitalSignature` alone (checked on-card before the
  write), so 9C does **not** surface as a key-exchange container ahead of 9D and card2 stays a
  clean canonical key-management card for the #3817 comparison. A 9C with `keyAgreement` in
  its keyUsage would have flipped that, hidden until now by the all-zero FASC-N.
- **Per-object GUIDs changed** although tag 34 did not: `piv_get_guid()` selects its
  derivation by serial length, so the minidriver's container ids on Windows moved from the
  FASC-N-digit form to the GUID form. Windows treats it as a new card and re-enrols; a smart
  card logon with card2 against our Samba KDC succeeded right after the rewrite, and we re-ran
  the stock-minidriver capture from #3817 on the rewritten card: same container map (9D is the
  only key-exchange container), same 32-byte challenge refused offline, same 48-byte sign and
  ECDH cache write online. Nothing in those results depended on the shared serial.

The offer to test the corrected patch (with the `indata` memcpy fix) on card2 still stands.

<details>
<summary>APDUs (card2, OpenFIPS201 on NXP J3R310). Two GENERAL AUTHENTICATE APDUs (9B, AES-256) between 2 and 3 omitted.</summary>

```
# Yubico GET VERSION (same on card1)
>> 00 FD 00 00 03
<< SW=6D00

# 1. SELECT PIV
>> 00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00
<< SW=9000, 113 bytes  (application label "OpenFIPS201")

# 2. GET DATA 5FC102 (CHUID, before)
>> 00 CB 3F FF 05 5C 03 5F C1 02 00
<< SW=9000, 59 bytes
   53 39
     30 19 D0 42 10 D8 21 08 6C 10 84 21 0D 83 68 58 21 08 42 10 84 21 C8 42 10 C3 EB
     34 10 A0 4F 53 43 73 10 01 F9 AA 6E 34 53 7B EE 6F C1
     35 08 32 30 33 36 30 39 31 32
     3E 00

# 3. PUT DATA 5FC102 (CHUID, new: tag 30 replaced, 34/35/3E unchanged)
>> 00 DB 3F FF 40 5C 03 5F C1 02 53 39
     30 19 D4 E7 39 D8 21 0C 2C 10 84 21 45 A1 68 58 21 08 42 10 84 28 84 E7 39 C3 E2
     34 10 A0 4F 53 43 73 10 01 F9 AA 6E 34 53 7B EE 6F C1
     35 08 32 30 33 36 30 39 31 32
     3E 00
<< SW=9000

# 4. GET DATA 5FC102 (read-back)
>> 00 CB 3F FF 05 5C 03 5F C1 02 00
<< SW=9000, 59 bytes  (identical to the PUT DATA payload)
```
</details>

dengert:
Two more things. 
 
The patch tests for both P-256 and P-384 before padding. Did you see any problems when using just P-256? If so I will change the patch to drop the P-256 check.   

>  ;0000=0000=000000=0=1=0000000000300001? — it is the all-zero default that pivy's init writes.

Well  it decodes as having some special non digits, and has a 1, 3, and 1. The zero FASC-N the OpenSC PIV is testing for is the encoded being 25 all 0x00 bytes.  i.e. absent FASC-N. It then checks the first 3 bytes and one bit of the encoded FASC-N  to see if its the encoded ;9999 as non gov FASC-N. I could also check for ;0000 
 Before I retired we would put a user's badge number in when issuing a temporary card -  if a user left their real card at home or for new employees waiting to get their official PIV card. 

The way a containerID is based on what the minidriver can use from a card. Windows builtin and OpenSC also differed and I assume any vendor of a PIV card who provides a minidriver does something different too. This has always been a problem if minidrivers are changed. 

Also not that each certificate gets a containerID too. certutil -v -scinfo shows these for each certificate.

dengert:
I assume you are in the US. Might I also suggestthat you look at:

PIVkey based on NIST sp800-73-3 and have a minidriver.
https://pivkey.com/index.html from https://www.taglio.com/  "Taglio LLC was founded in Austin Texas in 2011" https://pivkey.zendesk.com/hc/en-us
https://store.taglio.com/
They say they are on Amazon but not currently available.  

Yubico has the Yubikey but are tokens not cards, also based on NIST sp800-73-3 with extensions  have a minidriver and many other features and are very active. 
https://www.yubico.com/support/download/smart-card-drivers-tools/
They are on Amazon.

OpenSC has support for the above but via the OpenSC minidriver.

spurglegooge:
@dengert —

P-256: we never ran that case, but I think the answer is keep it

The issue this patch addresses happens in the cached-credential logon
path, where Windows hands a 32-byte value to the *logon* key. On every card we have, the logon
key (9A) is P-384, so we have never had a P-256 logon key to try it against. Our captures contain
three P-256 signatures, but they are 9C rather than logon and the input was 32 bytes, which equals
the P-256 field, so they say nothing about a short input. The shortest input we have ever seen
Windows send, on any key, is 32 bytes.

Sent raw APDUs to 9E (P-256,
PIN policy never, so no PIN and no retry counter), `GENERAL AUTHENTICATE` alg `0x11`:

```
20B  refused 6A80     SHA-1 length
28B  refused 6A80     SHA-224 length
31B  refused 6A80
32B  signed  9000     equals the field
33B  refused 6A80
48B  signed  9000
64B  signed  9000
```

Which I think answers your question, by a different route than the one you asked about. The
P-256 arm can only ever fire for a digest shorter than 32 bytes, and the only real ones are SHA-1
at 20 and **SHA-224 at 28**. The applet refuses both with the same `6A80` it gives the
32-byte-to-P-384 case. SHA-224 with P-256 is an approved pairing, so the arm is not merely
rescuing deprecated SHA-1 — dropping it would leave a legitimate combination broken on this
applet.

The padding is also semantically right rather than a workaround: `calloc` plus
`memcpy(data + padlen, indata, indatalen)` puts the zeros in front, and left-padding preserves
the big-endian integer, so the card signs exactly the value a correct implementation would use.

So: keep it, on the strength of the applet's rule rather than of any traffic we have observed. We
have still never seen a host send a short input to a P-256 key.

**FASC-N — you are right and my wording was wrong.** "All-zero default" described the *decoded*
digits; the encoded bytes are `D0 42 10 D8 …`, so `fbits` is set and OpenSC correctly treated it
as a present, well-formed FASC-N. The zero case in the driver is 25 literal `0x00` bytes, i.e. no
FASC-N at all, which is not what we had. Nothing in the driver misbehaved — the card carried a
default template and OpenSC did exactly what it documents.

On possibly testing for `;0000`: the only fact I can add is that this is not a one-off. `pivy-tool
init` writes that same template on every card unless told otherwise, so any two cards provisioned
that way carry the same FASC-N and therefore the same serial. Not a request — ours are fixed, and
your temporary-badge history is a better guide than our one data point to whether `;0000` can
safely be treated that way.

**ContainerIDs.** Understood, and that matches what we saw: preserving the GUID did not preserve
them, because the derivation follows the serial's length. We have not captured `certutil -v
-scinfo` per certificate yet; we will at the next session with the cards.

**The cards, closing an earlier loop.** card0 and card1 have since had the same CHUID fix, so all
three now present distinct serials, each the card's GUID. No certificate needed re-issuing.

**On your second note:** we do run YubiKeys alongside these JavaCards and they work great. We
needed the card form factor as an option as well. Thank you for the recommendation and research.

dengert:
https://learn.microsoft.com/en-us/windows/security/identity-protection/smart-cards/smart-card-certificate-requirements-and-enumeration does say Windows 11 but keeps saying CSP and not KSP which i think is needed for P-384 

You says Yubico works and I assume you have tried with P-384  so that points at the OpenFIPS201.
Have you heard anything from them? I get their github notifications and have not seen any.

Google: NIST PIV requirements 2031
PIV Authentication Key: RSA requirement transitions to a minimum of 3072 bits (disallowing 2048-bit RSA for this tier); ECDSA remains valid using Curve P-256 or P-384 but also dropping RSA 2048. 
Yubico also supports 3072, and Microsoft will also have to support 3072. 

So you could stick with P-256. Note OpenFIPS201 looks like it has not been changed in 4 years. The PIVKey  also appears to be based on NIST 900-73-3, and they too may need updates for the future.

PR #3275, opened 2024-11-26
pkcs11-spy: Avoid crash while spying C_GetInterface()

Jakuje:
When calling the C_GetInterface(), the spy was trying to change the memory returned by the underlying pkcs11 module. In cases where it was pointing to the readonly/static memory, the application crashed.

This changes the C_GetInterface() in a way that it does not return the original memory of underlying pkcs11 module, but local one with our pointers.

##### Checklist


xhanulik (approved):

frankmorgner:
thank you

PR #3281, opened 2024-12-02
Allocate whole page(s) for each sc_mem_secure_alloc

### LINKED ISSUE #3267: sc_mem_secure_alloc/free leave unaligned allocations locked in memory

dgalling:
sc_mem_secure_alloc uses calloc to allocate memory, but then uses mlock to try to lock the allocated memory. calloc may not necessarily return a page-aligned pointer, even if we request a whole page of memory.

Instead of relying on the libc allocator behavior, we directly call mmap (VirtualAlloc on Windows) to ensure we get whole page(s) to ourselves for sc_mem_secure_alloc.

Fixes #3267

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -2736,12 +2736,9 @@ sc_pkcs15_make_absolute_path(const struct sc_path *parent, struct sc_path *child
 
 void sc_pkcs15_free_object_content(struct sc_pkcs15_object *obj)
 {
-	if (obj->content.value && obj->content.len)   {
-		if (SC_PKCS15_TYPE_AUTH & obj->type
-			|| SC_PKCS15_TYPE_SKEY & obj->type
-			|| SC_PKCS15_TYPE_PRKEY & obj->type) {
-			/* clean everything that potentially contains a secret */
-			sc_mem_secure_clear_free(obj->content.value, obj->content.len);
+	if (obj->content.value && obj->content.len) {
+		if (obj->content_free) {
+			obj->content_free(obj->content.value, obj->content.len);
 		} else {
 			free(obj->content.value);
 		}
@@ -2772,6 +2769,7 @@ sc_pkcs15_allocate_object_content(struct sc_context *ctx, struct sc_pkcs15_objec
 			|| SC_PKCS15_TYPE_SKEY & obj->type
 			|| SC_PKCS15_TYPE_PRKEY & obj->type) {
 		tmp_buf = sc_mem_secure_alloc(len);
+		obj->content_free = sc_mem_secure_free;
 	} else {
 		tmp_buf = malloc(len);
 	}
--- a/src/libopensc/pkcs15.h
+++ b/src/libopensc/pkcs15.h
@@ -486,6 +486,10 @@ struct sc_pkcs15_object {
 	struct sc_pkcs15_object *next, *prev; /* used only internally */
 
 	struct sc_pkcs15_der content;
+	/* Method for deallocating the object's content.value.
+	 * If no specific function for deallocation is given, then free() is used
+	 * to release content.value */
+	void (*content_free)(void *, size_t);
 
 	int session_object;	/* used internally. if nonzero, object is a session object. */
 };
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -902,14 +902,12 @@ void *sc_mem_secure_alloc(size_t len)
 
 #ifdef _WIN32
 	p = VirtualAlloc(NULL, len, MEM_COMMIT, PAGE_READWRITE);
-	if (p != NULL)
-	{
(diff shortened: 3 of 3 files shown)


frankmorgner (approved):
Looks great! Thanks again for the bugfix.

frankmorgner:
please just check the code style: the opening curly bracket should go into the line of the `if`

frankmorgner:
https://github.com/OpenSC/OpenSC/actions/runs/12123642629/job/33836613737?pr=3281

dgalling:
I cherry-picked 92f53e63cdbb867bfc3d2e0b5bb87320cfd3ad5a to this branch because it doesn't make sense to merge this PR without your changes there.

The oseid and openpgp tests now pass on my machine under both valgrind and my custom heap debugger, and it looks like all of the tests are happy in CI as well.

Also, the PKCS11/SC is working on my machine in the Chromium browser with no crashes so far.

Nice job on the fixes to the PKCS15 allocation, I did not think it would be such a clean patch. One further question I had here was whether you wanted me to contribute my heap tracking code. It would allow you to detect cases where a memory block is allocated with the libc allocator and then freed with the `secure_free` (but not vice versa without overriding the libc `free`). There is of course some performance cost, but it might be worth it to allow the library to crash instead of corrupting the heap. Just a thought, definitely not a blocker for this PR.

frankmorgner:
Thanks for your help and the excellent analysis!

PR #3282, opened 2024-12-03
fixed usage of free() on data that was created with sc_mem_secure_alloc

frankmorgner:
This will crash applications, which, for example are importing a private key as reported here
https://github.com/OpenSC/OpenSC/pull/3269#issuecomment-2495074344

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -2736,12 +2736,9 @@ sc_pkcs15_make_absolute_path(const struct sc_path *parent, struct sc_path *child
 
 void sc_pkcs15_free_object_content(struct sc_pkcs15_object *obj)
 {
-	if (obj->content.value && obj->content.len)   {
-		if (SC_PKCS15_TYPE_AUTH & obj->type
-			|| SC_PKCS15_TYPE_SKEY & obj->type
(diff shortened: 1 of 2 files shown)


Jakuje (approved):
This will avoid the crash by calling the `sc_mem_secure_clear_free()` only on the stuff allocated with `sc_mem_secure_alloc()`.

But will it do the right thing for the content imported from the ASN1 parsers that are marked as secure? If I read it right, these data allocated from `src/libopensc/pkcs15-pubkey.c` are marked as secure, but allocated in the non-secure memory (which might be a different issue to solve).

frankmorgner:
Yes, correct. Only when `sc_pkcs15_allocate_object_content` is called "by hand" (i.g. not for the ASN.1 encoder), then `sc_mem_secure_alloc` is used and the dedicated `sc_mem_secure_free` will be used as replacement for `free`.

A subsequent issue, we might want to solve later, is when pkcs15init imports some key and initializes the object with some private key material, then it uses `malloc` and similar, which means that the key material may be swapped out...

frankmorgner:
merged with https://github.com/OpenSC/OpenSC/pull/3281

PR #3284, opened 2024-12-04
Add new appveyor image combinations

metsma:
* Use OpenSSL 3.0 in new images
* Use Wix 3.14 in new images
* Update ZLib to 1.3.1

##### Checklist

- [ ] macOS tokend is tested


metsma:
If we want minimise the build matrix in Appveyor, we should move least to Visual Studio 2019 image.
Then we have up to date OpenSSL 3.0.
And default build artifacts should come also from this image.

frankmorgner:
I think it is good to test all the different combinations. Having an Github action with Windows now enables developers to quickly see problems on Windows without waiting for every job to complete in AppVeyor.

frankmorgner (approved):

metsma:
> I think it is good to test all the different combinations. Having an Github action with Windows now enables developers to quickly see problems on Windows without waiting for every job to complete in AppVeyor.

I was thinking more the Appveyor build matrix.
What is the plan next release?
Do we still produce final product from appveyor?
I am worried that Appveyor 2017 image uses deprecated OpenSSL 1.1.1

frankmorgner:
For the release, we can take the newer image, no problem. The signpath hook works for all AppVeyor builds, we only need to select the correct one for release signing in their web front end when requesting the release package

PR #3285, opened 2024-12-12
Remove obsolete tokend support

metsma:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,4 +14,4 @@ run this command: `opensc-tool -n`
 - [ ] New files have a LGPL 2.1 license statement
 - [ ] PKCS#11 module is tested
 - [ ] Windows minidriver is tested
-- [ ] macOS tokend is tested
+- [ ] macOS token is tested
--- a/.gitignore
+++ b/.gitignore
@@ -109,7 +109,6 @@ MacOSX/resources/Welcome.html
 
 *.dmg
 *.pkg
-OpenSC.tokend/
 build/
 engine_pkcs11/
 libp11/
--- a/MacOSX/Distribution.xml.in
+++ b/MacOSX/Distribution.xml.in
@@ -11,32 +11,15 @@ https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/I
     <welcome file="Welcome.html" mime-type="text/html"/>
     <title>@PACKAGE_STRING@</title>
     <options customize="allow" require-scripts="false" rootVolumeOnly="true"/>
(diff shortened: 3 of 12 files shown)


metsma:
I think it is time to get rid of this. Nobody sane can build this today.

frankmorgner (commented):
In general I agree. Project description (Github metadata) needs to be updated. Also, OpenSC.tokend should get a note that it is not used anymore.

There are some more references in
- src/libopensc/pkcs15.c
- src/libopensc/reader-pcsc.c
- src/ui/notify.c

metsma:
> In general I agree. Project description (Github metadata) needs to be updated. Also, OpenSC.tokend should get a note that it is not used anymore.
> 
> There are some more references in
> 
> * src/libopensc/pkcs15.c
> * src/libopensc/reader-pcsc.c
> * src/ui/notify.c

I am not sure how to handle pkcs15-piv and notify comment

Jakuje (approved):
Thank you for cleaning this up!
Just a nit regarding the fuzzer input names.  I can not coment on how this is relevant for the macos or not as I do not use that.

frankmorgner (approved):

Jakuje:
@metsma I see we have tokend still mentioned in the "about" and tags on the main page https://github.com/opensc/opensc

I am not familiar with what happend with the tokend and if it should be dropped, but it should be dropped from these places too. Can you take care of that?

Then there is also the project itself that we are not using anymore and that should have at least a warning that its unmaintained: https://github.com/OpenSC/OpenSC.tokend/ (or should be archived altogether?)

metsma:
> @metsma I see we have tokend still mentioned in the "about" and tags on the main page https://github.com/opensc/opensc
> 
> I am not familiar with what happend with the tokend and if it should be dropped, but it should be dropped from these places too. Can you take care of that?
> 
> Then there is also the project itself that we are not using anymore and that should have at least a warning that its unmaintained: https://github.com/OpenSC/OpenSC.tokend/ (or should be archived altogether?)

This can probably archived.

I am not sure what is the correct branding of its successor. CTK Token?
I do not see that I have permission to alter About info.
And it is hosted by @frankmorgner 
https://github.com/frankmorgner/OpenSCToken

Jakuje:
I removed the TokenD tag and description from the "About" section. Frank, please adjust if there is some name to put there for Mac.

I also filled frankmorgner/OpenSCToken#54 to clarify the state of OpenSCToken.

metsma:
Sorry for confusion TokenD was replaced by CTK Token in macOS.
And its successor implementation is https://github.com/frankmorgner/OpenSCToken

PR #3289, opened 2024-12-16
Add support for PKCS#11 URI option into pkcs11-tool

xhanulik:
This PR adds new `--uri` option into  `pkcs11-tool` according to  [The PKCS #11 URI Scheme](https://www.rfc-editor.org/rfc/rfc7512.html). After parsing the URI string, it uses the already implemented `opt_*` variables for choosing module, slot/token, PIN and object.

It does not yet support:
- `file:` option for `pin-source`,
- specification module via `module-name`,
- `library-description`,
- `library-manufacturer`,
- `library-version`.

##### Checklist


Jakuje:
The failing CI does not look like related to your changes.

xhanulik:
> The failing CI does not look like related to your changes.

Yes, but there is still some problem with compilation of `pkcs11-tool` on win.

src/tools/pkcs11_uri.c
```diff
@@ -0,0 +1,332 @@
...
+#include "config.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "pkcs11_uri.h"
+
+static struct pkcs11_uri_attr path_attr[] = {
+		{"id", 0},
```
Jakuje, line 33: please, use enums instead of the numbers like this. It makes is hard to read.

src/tools/pkcs11-tool.c
```diff
@@ -1400,22 +1469,37 @@ int main(int argc, char * argv[])
...
 		 * if it's not known fall back to the old code always assuming it was a
 		 * CKO_PRIVATE_KEY.
 		 */
 		if (mf_flags != MF_UNKNOWN) {
 			/* this function dies on error via util_fatal */
-			find_object_flags(session, mf_flags, &object,
-				opt_object_id_len ? opt_object_id : NULL,
-				opt_object_id_len, 0);
+			if (opt_uri) {
+				find_object_flags_and_type_or_id_or_label(session, mf_flags, &object,
```
Jakuje, line 1480: can we do the search here directly by the URI, instead of limiting only to class, id, label?

xhanulik: In `struct pkcs11_uri`, there are 13 attributes (`id`, `library_description`, `library_manufacturer`, `library_version`, `token_manufacturer`, `token_model`, `object`, `serial`, `slot_description`, `slot_id`, `slot_manufacturer`, `token_label`, `type`).

In `pkcs11-tool`,  some of the attributes are unused ( `library_description`, `library_manufacturer`, `library_version`).
The `token_manufacturer`, `token_model`, `slot_description`, `slot_id`, `slot_manufacturer`, `token_label`, `serial` are used for finding a slot and opening new session

https://github.com/OpenSC/OpenSC/blob/58599a75c1138f1638afe54ebf5e73f8d8473799/src/tools/pkcs11-tool.c#L1355-L1358

For object there are `id`, `object`, `type` left used for finding the objects within the session. What do you mean by doing the search directly by URI?

Jakuje: I meant to pass the URI to the `find_object_flags()` function directly and search by the parts of the URI, but looking better now I see that you use just the class from uri here so it probably does not make sense.

frankmorgner (changes requested):
The URI parsing is quite complicated... looks like an idea fuzzing target!

xhanulik:
> The URI parsing is quite complicated... looks like an idea fuzzing target!

Good idea, I'll create one

frankmorgner (approved):
Thank you

frankmorgner:
Are there any use cases we could support now that didn't work previously? something to test for during CI?

xhanulik:
> Are there any use cases we could support now that didn't work previously? something to test for during CI?

I don't think so, I just mapped the PKCS#11 URI to current pkcs11-tool code functionality.

src/tools/pkcs11_uri.c
```diff
@@ -0,0 +1,342 @@
...
+	/* parse path */
+	token = strtok(path, ";");
+	while (token != NULL) {
+		char *argument = NULL, **result_ptr = NULL;
+		int *result_len = NULL;
+		int max_len = -1;
+		int id = get_attr(path_attr, token);
+		if (id == -1) {
+			rv = 1;
+			fprintf(stderr, "Invalid PKCS#11 token\n");
```
Jakuje, line 197:
```suggestion
			fprintf(stderr, "Invalid PKCS#11 uri atrribute\n");
```

src/tools/pkcs11_uri.c
```diff
@@ -0,0 +1,342 @@
...
+		int id = get_attr(query_attr, token);
+		if (id == -1) {
+			rv = 1;
+			fprintf(stderr, "Invalid PKCS#11 URI query\n");
+			goto end;
+		}
+		argument = token + strlen(query_attr[id].name) + 1;
+
+		switch (id) {
+		case 0:
```
Jakuje, line 277: same here -- please use the enum values

src/tools/pkcs11_uri.c
```diff
@@ -0,0 +1,342 @@
...
+				rv = 1;
+				fprintf(stderr, "A pin and pin-source cannot be specified together in PKCS#11 URI\n");
+				goto end;
+			}
+			result_ptr = &result->pin_source;
+			break;
+		case 1:
+			if (result->pin_source != NULL) {
+				rv = 1;
+				fprintf(stderr, "A pin and pin-source cannot be specified together in PKCS#11 URI\n");
```
Jakuje, line 288:
```suggestion
				fprintf(stderr, "A pin-value and pin-source cannot be specified together in PKCS#11 URI\n");
				goto end;
			}
			result_ptr = &result->pin_source;
			break;
		case 1:
			if (result->pin_source != NULL) {
				rv = 1;
				fprintf(stderr, "A pin-value and pin-source cannot be specified together in PKCS#11 URI\n");
```

src/tools/pkcs11-tool.c
```diff
@@ -1233,6 +1254,53 @@ int main(int argc, char * argv[])
...
+			} else {
+				fprintf(stderr, "Unsupported object type \"%s\" specified in PKCS#11 URI\n", opt_uri->type);
+				util_print_usage_and_die(app_name, options, option_help, NULL);
+			}
+		}
+		if (opt_uri->id) {
+			opt_object_id_len = opt_uri->id_len;
+			memcpy(opt_object_id, opt_uri->id, opt_object_id_len);
+		}
+	}
```
Jakuje, line 1299: Doing this allows us to have the change in the rest of the code quite small, but I thought that this could be done the other way round -- when uri is not provided, it can be constructed from the old options. By having the uri structure as the uniform exchange format in the pkcs11-tool accepted by all the operations, we could get rid of the number of different parameters to each of the function and just operate on the uri.

Indeed this would take more work and more changes to do that in all places so keeping this as a suggestion. If you believe it is a bad idea, feel free to discard ;)

frankmorgner:

xhanulik:
> What's the status of this topic, is there anything to do?

I'll resolve the comments above.

I like the idea desribed in https://github.com/OpenSC/OpenSC/pull/3289#discussion_r2035289786, however it looks like it'll need bit more work, and I'm not sure, whether I'll have time for it next month or so.

src/tools/pkcs11-tool.c
```diff
@@ -1233,6 +1254,53 @@ int main(int argc, char * argv[])
 	if (action_count == 0)
 		util_print_usage_and_die(app_name, options, option_help, NULL);
 
+	if (opt_uri) {
+		/* Check that no interfering options were set */
+		if (opt_module || opt_slot_set || opt_slot_index_set || opt_slot_description ||
+				opt_pin || opt_object_label || opt_object_class_str || opt_object_id_len) {
+			fprintf(stderr, "Invalid combination of arguments when --uri used\n");
```
Jakuje, line 1261: I think it would help to have this error message a bit more verbose explaining what arguments can not be used with the uri, otherwise it is hard to track from this error which of the options is the problematic one.

src/tools/pkcs11_uri.c
```diff
@@ -0,0 +1,342 @@
...
+		int id = get_attr(query_attr, token);
+		if (id == -1) {
+			rv = 1;
+			fprintf(stderr, "Invalid PKCS#11 URI query\n");
+			goto end;
+		}
+		argument = token + strlen(query_attr[id].name) + 1;
+
+		switch (id) {
+		case PKCS11_ID:
```
Jakuje, line 277:
```suggestion
		case PKCS11_PIN_SOURCE:
```
same in the cases below

frankmorgner (approved):

frankmorgner:
@xhanulik if you commit the formatting changes from `git clang-format`, I think we would be good for merging

xhanulik:
> @xhanulik if you commit the formatting changes from `git clang-format`, I think we would be good for merging

Resolved, I will create a separate issue for https://github.com/OpenSC/OpenSC/pull/3289#discussion_r2035289786 after merging.

Jakuje:
Thank you! Please open the separate follow-up issue then

PR #3291, opened 2024-12-17
Fix some recent CI issues 

Jakuje:
This is attempt to fix recent CI failures:
 * ubuntu changes broke valgrind builds
 * Fedora spec file is out of sync
 
Remaining issues look like related to libressl

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.github/setup-libressl.sh
+++ b/.github/setup-libressl.sh
@@ -2,7 +2,8 @@
 
 set -ex -o xtrace
 
-V=libressl-3.8.1
+# WARNING: Change this also in .github/workflows/linux.yml
+V=libressl-4.0.0
 
 sudo apt-get remove -y libssl-dev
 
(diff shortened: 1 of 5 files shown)


Jakuje:
The valgrind still fails and I still do not understand why the libressl tests keep failing. Will have to figure out what is going on there locally later.

xhanulik:
Can we merge this PR or do you want to look into the libressl problem?

Jakuje:
I did not have time to look into what is going on there with libressl. I would go ahead and merge it as it fixes some of the issues we are seeing in CI, but unfortunately not all.

frankmorgner (approved):

dengert:
> The valgrind still fails and I still do not understand why the libressl tests keep failing. Will have to figure out what is going on there locally later.

See PR #3310  that fixes valgrind and partially fixes the libressl tests. It appears that libressl 4.0.0 has a problem with saltlen,
or its the way we build it from ftp, rather than github. When built from  https://github.com/libressl/portable the autogen.sh will cause the patches in https://github.com/libressl/portable/tree/master/patches to be appied to the BSD version. One of then deals with "long long" "ssize_t".

PR #3303, opened 2025-01-06

tpetazzoni:
Similar to ab74fae4d71d1705b77b9459141987a95dcfc91e ("pkcs11-tool: disable wrap/unwrap test until OpenSC#1796 is resolved"), but for 0.26, since OpenSC#1796 is still open.

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -7681,7 +7681,7 @@ static int test_verify(CK_SESSION_HANDLE sess)
 	return errors;
 }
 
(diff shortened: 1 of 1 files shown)


Jakuje (approved):
thank you! Indeed we missed this with the release. After all these years, I am a bit skeptic if we will be able to fix it in the foreseeable future as the wrapping/unwrapping is not that common and some cards or drivers interpret some of the specs on their own. So I would probably propose to skip it altogether with `#if 1` or something like that.

xhanulik (approved):

frankmorgner (approved):

xhanulik:
Thanks, merging.

PR #3306, opened 2025-01-08
Use X509_NAME_print_ex() instead of X509_NAME_print()

botovq:
The third argument of X509_NAME_print() is obase and is intended as indentation level. This never worked though. From the flag value XN_FLAG_RFC2253 it is clear that the use of X509_NAME_print_ex() was intended. Use this instead.

https://docs.openssl.org/3.4/man3/X509_NAME_print_ex/


Jakuje (approved):
good catch! Looks good

metsma (approved):

xhanulik:
Thanks for the contribution!

PR #3307, opened 2025-01-09
Filter out unsupported algorithms by FIPS mode from p11test

xhanulik:
Unsupported algorithms are not tested by p11test, the `test-p11test.sh` should work also when FIPS is enabled.

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
Please, rebase on top of current master + fix the formatting issues. Otherwise it looks good.

Jakuje:
One more thing, the `softhsm_ref_fisp.json` needs to be added to the `tests/Makefile.am` to make sure it is part of the release tarball. Otherwise I think we are good and would be great to have this in 0.26.1.

Jakuje (approved):

Issue #3308, opened 2025-01-10
Implement applet version 1.8 for belpic card


bigon:
Hello,

I just my new Belgian ID card today and it seems that it's currently not supported by opensc

I see that the ATR for that card has been disabled in fe2c1c86c3826685be80a1c70aa526ae79477064

I've reenable the ATR and I can list the certificates and keys, but signature fails, as mentioned in the commit.

~Please find here the logs: [opensc-debug.txt](https://github.com/user-attachments/files/18375612/opensc-debug.txt)~

An idea how we should proceed?

Jakuje:
Do you have the specification or ATR trace of the part that does the signature? The ATR was added in #2455, which links some documentation, but I did not investigate it.

bigon:
Yes indeed #2455 contains some documentation and the APDU

I see page 45 of https://github.com/Fedict/eid-mw/blob/master/doc/sdk/documentation/Applet%201.8%20eID%20Cards/Public_Belpic_Applet_v1%208_Ref_Manual_v1%20.pdf that the applet 1.8 only supports ECDSA signature algorithm, could it be that?

bigon:
Oupsie, it seems that I didn't give the proper logfile, here's the correct one: 
[opensc-debug2.txt](https://github.com/user-attachments/files/18381642/opensc-debug2.txt)

bigon:
Hello,

With the following patch, it seems I get something that works.

On my card, the key is 384 bits, but the docs says it supports more sizes

But:
1. I don't have SHA3 it seems?
2. I see in other drivers that `curve_oid` parameter of `_sc_card_add_ec_alg()` is set, not sure whether it's OK to set it to NULL
3. `SHA1` and `SHA224`  (and other digests) appears in the list of supported mechanisms, is that OK?
 
``` patch
diff --git a/src/libopensc/card-belpic.c b/src/libopensc/card-belpic.c
index 3029c9c7c..37671082c 100644
--- a/src/libopensc/card-belpic.c
+++ b/src/libopensc/card-belpic.c
@@ -146,8 +146,8 @@ static long t1, t2, tot_read = 0, tot_dur = 0, dur;
 static size_t next_idx = (size_t)-1;
 
 static const struct sc_atr_table belpic_atrs[] = {
-	/* Applet V1.8 -- disabled, as it requires driver updates which are not yet implemented */
[...]

PR #3309, opened 2025-01-11
Add supports for belpic applet version 1.8

bigon:
This version of the applet only supports ECDSA keys

Documentation of the applet can be found at:
https://github.com/Fedict/eid-mw/tree/master/doc/sdk/documentation/Applet%201.8%20eID%20Cards

Fixes: #3308

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-belpic.c
+++ b/src/libopensc/card-belpic.c
@@ -146,15 +146,15 @@ static long t1, t2, tot_read = 0, tot_dur = 0, dur;
 static size_t next_idx = (size_t)-1;
 
 static const struct sc_atr_table belpic_atrs[] = {
-	/* Applet V1.8 */
-	{ "3B:7F:96:00:00:80:31:80:65:B0:85:04:01:20:12:0F:FF:82:90:00", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
-	/* Applet V1.1 */
-	{ "3B:98:13:40:0A:A5:03:01:01:01:AD:13:11", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
-	/* Applet V1.0 with new EMV-compatible ATR */
-	{ "3B:98:94:40:0A:A5:03:01:01:01:AD:13:10", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
-	/* Applet beta 5 + V1.0 */
-	{ "3B:98:94:40:FF:A5:03:01:01:01:AD:13:10", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL },
-	{ NULL, NULL, NULL, 0, 0, NULL }
+		/* Applet V1.8 */
+		{"3B:7F:96:00:00:80:31:80:65:B0:85:04:01:20:12:0F:FF:82:90:00", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL},
+		/* Applet V1.1 */
+		{"3B:98:13:40:0A:A5:03:01:01:01:AD:13:11",			   NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL},
+		/* Applet V1.0 with new EMV-compatible ATR */
+		{"3B:98:94:40:0A:A5:03:01:01:01:AD:13:10",			   NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL},
+		/* Applet beta 5 + V1.0 */
+		{"3B:98:94:40:FF:A5:03:01:01:01:AD:13:10",			   NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL},
+		{NULL,							  NULL, NULL, 0,			      0, NULL}
 };
 
 static struct sc_card_operations belpic_ops;
--- a/src/libopensc/card-belpic.c
+++ b/src/libopensc/card-belpic.c
@@ -146,8 +146,8 @@ static long t1, t2, tot_read = 0, tot_dur = 0, dur;
 static size_t next_idx = (size_t)-1;
 
 static const struct sc_atr_table belpic_atrs[] = {
-	/* Applet V1.8 -- disabled, as it requires driver updates which are not yet implemented */
-	/* { "3B:7F:96:00:00:80:31:80:65:B0:85:04:01:20:12:0F:FF:82:90:00", NULL, NULL, SC_CARD_TYPE_BELPIC_EID, 0, NULL }, */
+	/* Applet V1.8 */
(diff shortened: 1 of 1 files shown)


bigon:
I still need to look at the parameters of the curves I guess

dengert (approved):

src/libopensc/card-belpic.c
```diff
@@ -231,11 +232,19 @@ static int belpic_init(sc_card_t *card)
 		if(get_carddata(card, carddata, sizeof(carddata)) < 0) {
 			return SC_ERROR_INVALID_CARD;
 		}
-		if (carddata[BELPIC_CARDDATA_OFF_APPLETVERS] >= 0x17) {
-			key_size = 2048;
+		if (carddata[BELPIC_CARDDATA_OFF_APPLETVERS] >= 0x18) {
+			flags = SC_ALGORITHM_ECDSA_HASH_NONE | SC_ALGORITHM_ECDSA_RAW;
+			_sc_card_add_ec_alg(card, 256, flags, 0, NULL);
+			_sc_card_add_ec_alg(card, 384, flags, 0, &(struct sc_object_id){{1, 3, 132, 0, 34, -1}});
```
Jakuje, line 238: please, do the oids similarly as the PIV -- create global static list of the oids:

https://github.com/OpenSC/OpenSC/blob/8c1a26023c0da489577f253d596ce0a488252faa/src/libopensc/card-piv.c#L548-L549

and then pass them to this function:

https://github.com/OpenSC/OpenSC/blob/8c1a26023c0da489577f253d596ce0a488252faa/src/libopensc/card-piv.c#L5755-L5758

using the cycle might make it more readable too.

src/libopensc/card-belpic.c
```diff
@@ -231,11 +232,19 @@ static int belpic_init(sc_card_t *card)
...
+			_sc_card_add_ec_alg(card, 512, flags, 0, NULL);
```
Jakuje, line 239: what curve has 512 bits?

Jakuje (commented):
please, clean up the OIDs/ec curves as mentioned inline. Otherwise it looks good.

src/libopensc/card-belpic.c
dengert: Most cards do not add the curve to _sc_card_add_ec_alg.  And if you don't have documentation on what the others are, I would suggest you don't add any, since it appears to be a read only card and the curve is not really needed in _sc_card_add_ec_alg, as it is in public key and public key in certificates. 
The curve was added to PIV for the Windows  minidriver when creating keys for ECDH and Windows only supports 3 at this time.   `grep curve minidriver/minidriver.c`

Jakuje: Ok, then. So lets not add any OIDs. But just make sure the sizes make sense, see the comment about the 512b one

bigon: I don't know, the 512 bits is mentioned in the doc as a supported key size

bigon:
OK I've dropped the curve in the `_sc_card_add_ec_alg()` call

Jakuje (approved):
ok then. Its certainly better if it works than not.

dengert:
I don't see a 512 bit curve either,  but are used here, but  "Reference manual- Belpic V1.8 eID card" has:  "[ER10] NIST-FIPS-186-4,
ECDSA
ECDSA as specified by NIST in FIPS 186-4,July 2013
http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf"

which is replaced by:  
https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf and NIST SP 800-186 https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf does list curves but not the OIDs.
But talks about Edwards and Montgomery curves as well as brainpool curves. 


OpenSC source does show a few "nistp" curves:
```	{{{1, 2, 840, 10045, 3, 1, 1, -1}},     192, 0x0000}, /* secp192r1, nistp192, prime192v1, ansiX9p192r1 */
	{{{1, 3, 132, 0, 33, -1}},              224, 0x0000}, /* secp224r1, nistp224 */
	{{{1, 2, 840, 10045, 3, 1, 7, -1}},     256, 0x0000}, /* secp256r1, nistp256, prime256v1, ansiX9p256r1 */
	{{{1, 3, 132, 0, 34, -1}},              384, 0x0000}, /* secp384r1, nistp384, prime384v1, ansiX9p384r1 */
./libopensc/card-isoApplet.c
		{"nistp192",		"1.2.840.10045.3.1.1", "06082A8648CE3D030101", 192},
		{"nistp224",		"1.3.132.0.33", "06052b81040021", 224},
		{"nistp256",		"1.2.840.10045.3.1.7", "06082A8648CE3D030107", 256},
		{"nistp384",		"1.3.132.0.34", "06052B81040022", 384},
		{"nistp521",		"1.3.132.0.35", "06052B81040023", 521},
./libopensc/pkcs15-pubkey.c
	{{{1, 2, 840, 10045, 3, 1, 7, -1}},     256}, /* secp256r1, nistp256, prime256v1, ansiX9p256r1 */
	{{{1, 3, 132, 0, 34, -1}},              384}, /* secp384r1, nistp384, prime384v1, ansiX9p384r1 */
./libopensc/card-piv.c
		len = snprintf(alg, sizeof alg, "ecdsa-sha2-nistp%zu", n);
		tmp = snprintf((char *) buf+len+1, 9, "nistp%zu", n);
./tools/pkcs15-tool.c
	{"nistp192",     "1.2.840.10045.3.1.1", "06082A8648CE3D030101", 192, 0},
	{"nistp224",  "1.3.132.0.33", "06052b81040021", 224, 0},
	{"nistp521",  "1.3.132.0.35", "06052B81040023", 521, 0},
./tools/pkcs11-tool.c
		/* prime224v1, nistp224 */
./pkcs15init/pkcs15-isoApplet.c
```

bigon:
What am I doing then here?

Should I try to get clarification about the 512bits curves?

bigon:
On the Fedict/eid-mw driver, I see some OID for 384, 512 and 521: https://github.com/Fedict/eid-mw/blob/820fc69a5389fc95e80dc94b3133016bf203e3b2/cardcomm/pkcs11/src/asn1.h#L142

dengert:
Knowing what curves are used is probably not important at this time. The card issuer will issue certificates with whatever curve they want and since both the client and server must use the same curve, they will keep it simple. 

Since they assign an algorithm ID based on size (so they only support one curve per size), all the OpenSC has to do as take the size of the public key and map to the algorithm ID.  The public key will also have the OID if really need it.

bigon:
OK so is there anything blocking here?

Jakuje:
Nothing blocks. Just waiting if some others will have some more comments, but given that this is approved for over a week, I think its time to merge. Thank you for your contribution!

PR #3310, opened 2025-01-11
setup-linux.sh - get package name  to load the -dbgsym libs for OpenSSL 

dengert:
Use the same OpenSSL package names as installed for libs. Newer versions of debian based OS use a different naming convention for libs then older versions. 

`dpkg -S` is used to find the package name for the OpenSSL libs to load the `<package>-dbgsym` files.
The libcrypto and libssl are both in the same package. 

This also fixes the other actions that depend on setup-linux.sh  loading the dbgsym libs.
 
The `test-oseid-libressl` still fails but this appears to be problem within libressl with PSS saltlen 

 On branch github-setup
 Changes to be committed:
	modified:   setup-linux.sh

##### Checklist

- [ ] macOS tokend is tested


Jakuje:
Thank you for having a look into that! Looks good.

frankmorgner (approved):

PR #3312, opened 2025-01-14
OpenSC 0.26.1

xhanulik:
The stable branch for 0.26.1: https://github.com/OpenSC/OpenSC/tree/0.26.1

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-version: 0.26.0.{build}
+version: 0.26.1.{build}
 
 platform:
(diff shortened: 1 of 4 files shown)


dengert:
@xhanulik 

Can you fix the commit message in 643d4c1a0c8faf85341d099d5197f5a984020fcc It does not have a title and is taking the what should be the body of the message as the title, which caused things like   `git log --oneline to print the 4000 bytes of the commit  as a oneline.

Issue #2450, opened 2021-11-24
ARM64 Windows support


tommy2d:
First of all, excuse me for pushing this feature request through this route. If I should post this elsewhere I will gladly do so.

### Problem Description
Installing the current x86_64 build on Windows ARM64 2200 yields working tools on ARM64 Windows (e.g. opensc-tool.exe --list readers works), but the windows driver does not, resulting in a none-working putty-CAC etc. This is probably only fixed when a native ARM64 windows build becomes available. Is this something that is on the roadmap?

frankmorgner:
Windows on ARM was something people were wishing few years ago, but it looked like WIntel was too much of a burden to make it real. Nice to hear that some devices finally made it to production...

The build configuration should not be too complicated, but I don't think I can test that. Also, I'm unsure if I find some time soon for looking at the changes.

In x86 64 bit systems, we have a 32 bit subsystem which can be used in parallel. Could it be that Windows also brings such a emulation subsystem to run non-arm software on an actual arm system?

Also, an other way of solving your issue is to use the linux subsystem an run ssh directly with the "linux version" of OpenSC. But I'm not sure if the card reader will be available to the linux subsystem...

Jakuje:
As mentioned by Frank, we have limited capacity to work on opensc and to support various configurations of different OS versions. I dont have anything that runs windows arm64 anywhere around and I am not familiar with the technology (but there should not be any issue as the code builds ok on arm on linux) so this is mostly about somebody putting things together and figuring how this can be build and shipped to our users.

PR #3315, opened 2025-01-16
Build Windows Arm64 binaries

metsma:
https://github.com/OpenSC/OpenSC/issues/2450, Depends https://github.com/OpenSC/OpenSC/pull/3283

##### Checklist

- [ ] macOS tokend is tested


frankmorgner (approved):

frankmorgner:
I'm puzzled about the force-pushes... do you think this PR is ready for review/inclusion?

metsma:
I think so

metsma:
@frankmorgner is there anything blocking to get this merged?

.github/vcpkg.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "opensc",
```
frankmorgner, line 2: does this mean that we have an opensc vcspkg with this? would this be a new distribution path?

metsma: No it is just manifest file to help install dependencies with vcpkg.

PR #3316, opened 2025-01-16
Fix error found when compiling with gcc 15

xhanulik:
##### Checklist


src/libopensc/card-epass2003.c
```diff
@@ -3050,7 +3050,7 @@ epass2003_erase_card(struct sc_card *card)
 		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
 		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
 	};
-	static const unsigned char magic_pin[16] = "1234567812345678";
+	static const unsigned char magic_pin[17] = "1234567812345678";
```
Jakuje, line 3053: I do not think this is correct. The APDU will likely not expect the trailing null byte so either it needs to be stripped from the sizeof(magic_pin) or it needs to be defined explicitly without trailing null byte , similarly as the `install_magic_pin` variable above.

Jakuje (changes requested):

src/libopensc/card-epass2003.c
xhanulik: Thanks, I missed there is `sizeof` instead of `strlen` for getting the length, I will fixed that.

src/libopensc/card-piv.c
```diff
@@ -594,8 +594,8 @@ struct piv_object {
 	unsigned int resp_tag;
 	const char * oidstring;
 	size_t tag_len;
-	u8  tag_value[3];
-	u8  containerid[2];	/* will use as relative paths for simulation */
+	u8  tag_value[4];
+	u8  containerid[3];	/* will use as relative paths for simulation */
```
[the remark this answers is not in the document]
xhanulik: In this case, for `tag_value` there is explictly set and used `tag_len` so the terminating zero should be alright. For `containerid` there is no checking of length but it's always used as exactly two bytes. But for perevnting future problems, I will rewrite this as well.

src/libopensc/card-epass2003.c
```diff
@@ -3050,7 +3050,10 @@ epass2003_erase_card(struct sc_card *card)
 		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
 		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
 	};
-	static const unsigned char magic_pin[16] = "1234567812345678";
+	static const unsigned char magic_pin[16] = {
+		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38,
+		0x31,0x32,0x33,0x34, 0x35,0x36,0x37,0x38
+	};
```
Jakuje, line 3056: the only nit would be to make the clang format happy:
```suggestion
			0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38,
			0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38,
	};
```

Jakuje (approved):
only nit about formatting, but you can consider it approved

frankmorgner (approved):

xhanulik:
Thanks for review, merging.

PR #3299, opened 2024-12-28
Fix -O3 gcc optimization failure on amd64 and ppc64el

LocutusOfBorg:
convert int bMatch = -1 into size_t bMatch = 0, this fixes errors such as: card-entersafe.c: In function ‘entersafe_select_file’: card-entersafe.c:649:28: error: ‘__memcpy_chk’ writing between 18446744073709551608 and 18446744073709551613 bytes into a region of size 64 [-Werror=stringop-overflow=]
  649 |                            memcpy(new_path.value, &(path[bMatch+2]), new_path.len);
      |                            ^
card-entersafe.c:639:38: note: destination object ‘new_path’ of size 64
  639 |                            sc_path_t new_path;
      |                                      ^~~~~~~~
/bin/bash ../../libtool  --tag=CC --tag CC  --mode=compile gcc -DHAVE_CONFIG_H -I. -I../..  -D'OPENSC_CONF_PATH="/etc/opensc/opensc.conf"' -D'DEFAULT_SM_MODULE_PATH="/usr/lib/powerpc64le-linux-gnu"' -D'DEFAULT_SM_MODULE="libsmm-local.so"' -I../../src -Wdate-time -D_FORTIFY_SOURCE=3    -I/usr/include/PCSC   -Wall -Wextra -Wno-unused-parameter -Werror -Wstrict-aliasing=2 -g -O3 -Werror=implicit-function-declaration -ffile-prefix-map=/<<PKGBUILDDIR>>=. -flto=auto -ffat-lto-objects -fstack-protector-strong -Wformat -Werror=format-security -fno-stack-clash-protection -fdebug-prefix-map=/<<PKGBUILDDIR>>=/usr/src/opensc-0.25.1-2 -Wno-error=maybe-uninitialized -c -o libopensc_la-card-coolkey.lo `test -f 'card-coolkey.c' || echo './'`card-coolkey.c
[...]

PR #3318, opened 2025-01-16
From  Jakub Jelen <jakuje@gmail.com>

LocutusOfBorg:
on line 437 we assert the pathlen is 2 or 4 or 6
on line 440 we assert that if pathlen is 6, the first two bytes are 0x3F00 on line 444 we extend the 2 or 4 bytes paths to 4 or 6 bytes starting with 0x3F00 the line 459 is no-operation with your change. It should be removed if it keeps working as it should This line change semantics as previously we were heading here also when we had a cache, but first two bytes of path did not match. OTOH I think this should never happen as the path should always start with 0x3F00 so I believe this change is ok.

Just remove the line 459 to avoid confusion. Same in the other files, where similar construct exist.

https://github.com/OpenSC/OpenSC/pull/3299/files#r1899506106

### DIFF
--- a/src/libopensc/card-atrust-acos.c
+++ b/src/libopensc/card-atrust-acos.c
@@ -456,7 +456,6 @@ static int atrust_acos_select_file(struct sc_card *card,
 		    && card->cache.current_path.len >= 2
 		    && card->cache.current_path.len <= pathlen )
 		{
-			bMatch = 0;
 			for (i=0; i < card->cache.current_path.len; i+=2)
 				if (card->cache.current_path.value[i] == path[i]
 				    && card->cache.current_path.value[i+1] == path[i+1] )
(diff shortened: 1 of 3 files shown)


Jakuje:
Please, adjust the commit message as described in some of the articles like "How to Write a Git Commit Message", for example https://cbea.ms/git-commit/

LocutusOfBorg:
done

Jakuje (approved):

Jakuje:
The CI failure is not your fault, should be fixed with #3319.

frankmorgner (approved):

xhanulik (approved):

LocutusOfBorg:
rebased to fix some test failures

Jakuje:
Thank you! Merging!

LocutusOfBorg:
thanks!

PR #3319, opened 2025-01-16
ci: Update awalsh128/cache-apt-pkgs-action to v1.4.3

Jakuje:
The latest branch was not updated and still contains the use of deprecated actions/upload-artifact@v3 which no longer works.

https://github.com/awalsh128/cache-apt-pkgs-action/issues/145


.github/workflows/linux.yml
```diff
@@ -570,7 +570,7 @@ jobs:
     needs: [build-libressl]
     steps:
       - uses: actions/checkout@v4
-      - uses: awalsh128/cache-apt-pkgs-action@latest
+      - uses: awalsh128/ncache-apt-pkgs-action@v1.4.3
```
metsma, line 573:
```suggestion
      - uses: awalsh128/cache-apt-pkgs-action@v1.4.3
```

Jakuje: Fixed! Thanks for prompt checking!

dengert:
@Jakuje  https://github.com/OpenSC/OpenSC/actions/caches says we are at 10.1GB of 10GB limit for caches. Should some of these be deleted? How do we get people to rebase any PRs so then can bet the latest versions of the actions and thus the caches.

metsma:
> @Jakuje https://github.com/OpenSC/OpenSC/actions/caches says we are at 10.1GB of 10GB limit for caches. Should some of these be deleted? How do we get people to rebase any PRs so then can bet the latest versions of the actions and thus the caches.

I think the largest caches are from Windows vcpkg artifacts. I will look into ways to reduce them.

Jakuje:
> @Jakuje https://github.com/OpenSC/OpenSC/actions/caches says we are at 10.1GB of 10GB limit for caches. Should some of these be deleted?

The page says that the oldest unused get deleted automatically so this should not be an issue, but indeed it makes the builds slower if the caches get deleted too frequently.

> How do we get people to rebase any PRs so then can bet the latest versions of the actions and thus the caches.

I do not think that this should be an issue. The caches are used only when people push. The ubuntu images and packages should be the same for everyone so they should get used without any issues. Then we have some caches that are internal for each workflow which account for ~110MB per PR.

> I think the largest caches are from Windows vcpkg artifacts. I will look into ways to reduce them.

These do 400MB per PR, which is likely the largest chunk. I am not sure what they are used for and if there is space to shrink them. If it would be possible, it would be helpful.

Anyway, both of this is not directly related to this change so please, review/approve/merge so we can unbreak the builds if possible.

metsma (approved):

dengert:
I was just pointing out the limit, and not following the PR closely.   So am not in position to review/approve/merge, leaving that up to you and and Frank.

PR #3329, opened 2025-01-28
Issues reported by coverity

xhanulik:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/common/libpkcs11.c
+++ b/src/common/libpkcs11.c
@@ -99,6 +99,8 @@ C_LoadModule(const char *mspec, CK_FUNCTION_LIST_PTR_PTR funcs)
 			mod = NULL; /* already freed */
 	}
 failed:
+	if (mod && mod->handle)
+		sc_dlclose(mod->handle);
 	free(mod);
 	return NULL;
 }
--- a/src/libopensc/card-authentic.c
+++ b/src/libopensc/card-authentic.c
@@ -2295,6 +2295,7 @@ authentic_sm_get_wrapped_apdu(struct sc_card *card, struct sc_apdu *plain, struc
 
 	apdu->resp = calloc (1, plain->resplen + 32);
 	if (!apdu->resp) {
+		free((unsigned char *) apdu->data);
 		free(apdu);
 		LOG_FUNC_RETURN(ctx, SC_ERROR_OUT_OF_MEMORY);
 	}
@@ -2305,6 +2306,7 @@ authentic_sm_get_wrapped_apdu(struct sc_card *card, struct sc_apdu *plain, struc
 	rv = card->sm_ctx.module.ops.get_apdus(ctx, &card->sm_ctx.info, NULL, 0, NULL);
 	if (rv < 0) {
 		free(apdu->resp);
+		free((unsigned char *) apdu->data);
 		free(apdu);
 	}
 	LOG_TEST_RET(ctx, rv, "SM: GET_APDUS failed");
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -709,29 +709,34 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
 
 int
 _pgp_handle_curve25519(sc_card_t *card,
-		sc_cardctl_openpgp_key_gen_store_info_t key_info, unsigned int do_num)
+		sc_cardctl_openpgp_key_gen_store_info_t *key_info, unsigned int do_num)
 {
-	if (!sc_compare_oid(&key_info.u.ec.oid, &curve25519_oid) &&
-			!sc_compare_oid(&key_info.u.ec.oid, &X25519_oid))
+	if (!card || !key_info)
+		return 0;
+	if (!sc_compare_oid(&key_info->u.ec.oid, &curve25519_oid) &&
+			!sc_compare_oid(&key_info->u.ec.oid, &X25519_oid))
 		return 0;
 
 	/* CKM_XEDDSA supports both Sign and Derive, but
 	* OpenPGP card supports only derivation using these
 	* keys as far as I know */
-	_sc_card_add_xeddsa_alg(card, key_info.u.ec.key_length,
-			SC_ALGORITHM_ECDH_CDH_RAW | SC_ALGORITHM_ONBOARD_KEY_GEN, 0, &key_info.u.ec.oid);
+	_sc_card_add_xeddsa_alg(card, key_info->u.ec.key_length,
+			SC_ALGORITHM_ECDH_CDH_RAW | SC_ALGORITHM_ONBOARD_KEY_GEN, 0, &key_info->u.ec.oid);
 	sc_log(card->ctx, "DO %uX: Added XEDDSA algorithm (%d), mod_len = %zu",
-	    do_num, SC_ALGORITHM_XEDDSA, key_info.u.ec.key_length);
+	    do_num, SC_ALGORITHM_XEDDSA, key_info->u.ec.key_length);
 	return 1;
 }
 
 int
-_pgp_add_algo(sc_card_t *card, sc_cardctl_openpgp_key_gen_store_info_t key_info, unsigned int do_num)
+_pgp_add_algo(sc_card_t *card, sc_cardctl_openpgp_key_gen_store_info_t *key_info, unsigned int do_num)
 {
 	unsigned long flags = 0, ext_flags = 0;
 
+	if (!card || !key_info)
+		return 0;
+
 	/* [RFC 4880], [draft-ietf-openpgp-crypto-refresh] */
-	switch (key_info.algorithm) {
+	switch (key_info->algorithm) {
 	case SC_OPENPGP_KEYALGO_RSA:
 		/* OpenPGP card spec 1.1 & 2.x, section 7.2.9 & 7.2.10 /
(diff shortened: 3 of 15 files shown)


src/tools/goid-tool-cmdline.c
```diff
@@ -1038,7 +1038,7 @@ void free_list(struct generic_list *list, short string_arg)
  * @brief updates a multiple option starting from the passed list
  */
 static
-void update_multiple_arg(void *field, char ***orig_field,
+int update_multiple_arg(void *field, char ***orig_field,
```
frankmorgner, line 1041: I suggest to ignore the cmdline files and instead report this problem to gengetopt

frankmorgner (changes requested):
Maybe you could also rename indx ==> index to silence the spelling annotation.

src/common/libpkcs11.c
```diff
@@ -99,6 +99,8 @@ C_LoadModule(const char *mspec, CK_FUNCTION_LIST_PTR_PTR funcs)
 			mod = NULL; /* already freed */
 	}
 failed:
+	if (mod && mod->handle)
```
Jakuje, line 102: is the `mod` null check needed? I think at every code path we get here, the `mod` is not null. OTOH it should not hurt checking this as the code could change in the future.

src/libopensc/card-authentic.c
```diff
@@ -2305,6 +2306,7 @@ authentic_sm_get_wrapped_apdu(struct sc_card *card, struct sc_apdu *plain, struc
 	rv = card->sm_ctx.module.ops.get_apdus(ctx, &card->sm_ctx.info, NULL, 0, NULL);
 	if (rv < 0) {
 		free(apdu->resp);
+		free((unsigned char *) apdu->data);
```
Jakuje, line 2309: why is the cast needed?

src/libopensc/pkcs15-coolkey.c
```diff
@@ -679,7 +679,7 @@ static int sc_pkcs15emu_coolkey_init(sc_pkcs15_card_t *p15card)
 			sc_log(card->ctx, "sc_pkcs15emu_object_add() returned %d", r);
 fail:
 		if (key) { sc_pkcs15_free_pubkey(key); }
-
+		if (r < 0) { free(cert_info.value.value); }
```
Jakuje, line 682: even if we have old ugly code around, it does not mean the new code needs to be ugly.
```suggestion
		if (r < 0) {
			free(cert_info.value.value);
		}
```

src/libopensc/pkcs15-sc-hsm.c
```diff
@@ -1028,10 +1028,16 @@ static int sc_pkcs15emu_sc_hsm_add_pubkey(sc_pkcs15_card_t *p15card, u8 *efbin,
...
+		sc_pkcs15_erase_pubkey(&pubkey);
 	LOG_TEST_RET(ctx, r, "Could not encode public key");
 	r = sc_pkcs15_encode_pubkey(ctx, &pubkey, &pubkey_info.direct.raw.value, &pubkey_info.direct.raw.len);
+	if (r != SC_SUCCESS)
+		sc_pkcs15_erase_pubkey(&pubkey);
 	LOG_TEST_RET(ctx, r, "Could not encode public key");
 	r = sc_pkcs15_encode_pubkey_as_spki(ctx, &pubkey, &pubkey_info.direct.spki.value, &pubkey_info.direct.spki.len);
+	if (r != SC_SUCCESS)
+		sc_pkcs15_erase_pubkey(&pubkey);
 	LOG_TEST_RET(ctx, r, "Could not encode public key");
```
Jakuje, line 1041: I would suggest to keep both in the condition to be clear on the first sight both of the commands are error handling and they return.
```suggestion
	if (r != SC_SUCCESS) {
		sc_pkcs15_erase_pubkey(&pubkey);
		LOG_TEST_RET(ctx, r, "Could not encode public key");
	}
	r = sc_pkcs15_encode_pubkey(ctx, &pubkey, &pubkey_info.direct.raw.value, &pubkey_info.direct.raw.len);
	if (r != SC_SUCCESS) {
		sc_pkcs15_erase_pubkey(&pubkey);
		LOG_TEST_RET(ctx, r, "Could not encode public key");
	}
	r = sc_pkcs15_encode_pubkey_as_spki(ctx, &pubkey, &pubkey_info.direct.spki.value, &pubkey_info.direct.spki.len);
	if (r != SC_SUCCESS) {
		sc_pkcs15_erase_pubkey(&pubkey);
		LOG_TEST_RET(ctx, r, "Could not encode public key");
	}
```

src/tools/egk-tool.c
```diff
@@ -66,12 +66,14 @@ int uncompress_gzip(void* uncompressed, size_t *uncompressed_len,
 	stream.next_out = (Bytef *) uncompressed;
 
 	/* 15 window bits, and the +32 tells zlib to to detect if using gzip or zlib */
-	if (Z_OK == inflateInit2(&stream, (15 + 32))
-			&& Z_STREAM_END == inflate(&stream, Z_FINISH)) {
-		*uncompressed_len = stream.total_out;
-	} else {
+	if (inflateInit2(&stream, (15 + 32)) != Z_OK) {
+		return SC_ERROR_INVALID_DATA;;
```
Jakuje, line 70:
```suggestion
		return SC_ERROR_INVALID_DATA;
```

src/common/libpkcs11.c
xhanulik: Just the `else` branch above `failed` sets `mod = NULL;` without explicit return.

src/libopensc/card-authentic.c
xhanulik: The `apdu->data` is `const` so the compiler complains about discarding `const` when passed to `free` without cast.

src/pkcs15init/pkcs15-sc-hsm.c
```diff
@@ -282,6 +282,8 @@ static int sc_hsm_generate_key(struct sc_profile *profile, struct sc_pkcs15_card
 		r = SC_ERROR_NOT_IMPLEMENTED;
 		break;
 	}
+	if (r != SC_SUCCESS)
+		sc_pkcs15emu_sc_hsm_free_cvc(&cvc);
 	LOG_TEST_RET(p15card->card->ctx, r, "Could not encode GAKP cdata");
```
Jakuje, line 287: The same here as pointed in the previous comments, it would be more understandable if the return was in the came block:
```suggestion
	if (r != SC_SUCCESS) {
		sc_pkcs15emu_sc_hsm_free_cvc(&cvc);
		LOG_TEST_RET(p15card->card->ctx, r, "Could not encode GAKP cdata");
	}
```

src/tools/pkcs11-tool.c
```diff
@@ -2625,9 +2625,9 @@ static void verify_signature(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 			 * something like: "04 41 04||x||y"
 			 * Do simple size calculation based on DER encoding
 			 */
-			if ((len - 2) <= 127)
+			if (len >= 3 && (len - 2) <= 127)
 				rs_len = len - 3;
-			else if ((len - 3) <= 255)
+			else if (len >= 4 && (len - 3) <= 255)
```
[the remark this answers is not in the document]
dengert: pkcs11-tool.c can call some of the opensc routines, with context == NULL.  But you need to look at the routine.
It is also possible to use OpenSSL routines. in #3090 now committed: https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L6566-L6588 uses OpenSSL to get the value of an octet string or bit string or raw string from ECPOINT.  note `ASN1_OCTET_STRING *os;` is one of the few structures that the caller can access.  line 6588  uses `os->length` to get the length of the EDwards or Montgomery curve.  

Something like this might work.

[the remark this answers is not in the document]
dengert: Also  `ec_curve_by_name` and `ec_curve_by_parms`  return  ec_curve_info which contains the field size in bits.
So no need to get the ec_point for EC curves. 
Code like:
https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-tool.c#L2449-L2455
looks up the curve the Edwards and Montgomery curves to get size. (PKCS11 defines  ec_parms for EC,  Edwards and Montgomery curves.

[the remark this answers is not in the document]
frankmorgner: sc_asn1_read_tag should be suitable

Jakuje (approved):

src/libopensc/card-openpgp.c
```diff
@@ -709,40 +709,40 @@ pgp_parse_algo_attr_blob(sc_card_t *card, const pgp_blob_t *blob,
 
 int
 _pgp_handle_curve25519(sc_card_t *card,
-		sc_cardctl_openpgp_key_gen_store_info_t key_info, unsigned int do_num)
+		sc_cardctl_openpgp_key_gen_store_info_t *key_info, unsigned int do_num)
 {
-	if (!sc_compare_oid(&key_info.u.ec.oid, &curve25519_oid) &&
```
Jakuje, line 714: given that we now pass the struct by reference, we should really make sure it is not NULL before dereferencing here.  The same in the other function.

Jakuje (approved):

frankmorgner (approved):

PR #3332, opened 2025-02-01
 Workaround Yubikey bug with SELECT DATA amended #2969 

frankmorgner:
##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -71,7 +71,7 @@ static const struct sc_atr_table pgp_atrs[] = {
 	},
 	{ "3b:fc:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:4e:45:4f:72:33:e1", NULL, "Yubikey NEO (OpenPGP v2.0)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
 	{ "3b:f8:13:00:00:81:31:fe:15:59:75:62:69:6b:65:79:34:d4", NULL, "Yubikey 4 (OpenPGP v2.1)", SC_CARD_TYPE_OPENPGP_V2, 0, NULL },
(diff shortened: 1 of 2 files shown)


src/libopensc/card-openpgp.c
```diff
@@ -1856,6 +1894,42 @@ pgp_select_data(sc_card_t *card, u8 p1)
...
+{
+	sc_apdu_t	apdu;
+	u8	apdu_data[7];
+	int	r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_log(card->ctx, "select data with: %u (Yubikey workaround)", p1);
+	// Yubikey wants another length
+	apdu_data[0] = 0x06;
```
Jakuje, line 1914: given that the code change is basically just this line, would it make sense to have the check for the flag checked directly in `pgp_select_data()` instead of duplicating the whole function? it will need a slight change in how the data is constructed to something like
```
if (card->flags & SC_CARD_FLAG_YUBIKEY_SELECT) { 
	apdu_data[n++] = 0x06;
}
// the rest of assgined in format of apdu_data[n++] = 0x06;
```
and then use `n` instead of `sizeof(adu_data)` below?

src/libopensc/card-openpgp.c
```diff
@@ -584,6 +615,14 @@ pgp_init(sc_card_t *card)
 		}
 	}
 
+	if (card->reader->atr_info.hist_bytes_len >= 7 &&
+			!(memcmp(card->reader->atr_info.hist_bytes, "Yubikey", 7)) &&
+			SC_SUCCESS == pgp_yubikey_get_version(card, &yubikey_version) &&
+			yubikey_version <= 0x050403) {
```
dengert, line 621: the compare should test a range of version number. Do you know when it started and when it is fixed?

Issue #3334, opened 2025-02-07
Invalid write on repeated C_WaitForSlotEvent


3v1n0:
When using opensc pkcs via [gnome-settings-daemon plugin](https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/blob/master/plugins/smartcard/gsd-smartcard-manager.c#L143) and doing quick plug/unplug of a token we may get this memory error.

It seems to happen on the next call to `C_WaitForSlotEvent` just after the slot has been removed

This is with 0.25.0~rc1, but there are no significant changes to the slot code from what I saw, so it shouldn't matter.

```
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.456: Detected smartcard insertion event in slot 8
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.457: Waiting for event on 0

(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.484: ===============================
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.484:  Token 'MARCO TREVISAN (PIN CNS1)'
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.484:  Inserted: yes
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.484:  Previously used to login: no
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:13.484: ===============================

(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.338: DOne waiting for event on 8

(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.349: Detected smartcard removal event in slot 8
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.349: Waiting for event on 0

(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.358: ===============================
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.358:  Token 'MARCO TREVISAN (PIN CNS1)'
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.358:  Inserted: no
(gsd-smartcard:1750487): smartcard-plugin-DEBUG: 16:56:14.358:  Previously used to login: no
[...]

PR #3337, opened 2025-02-11
Avoid use after free when handling hotplug events in WaitForSlotEvent

Jakuje:
Fixes: #3334

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/cwa14890.c
+++ b/src/libopensc/cwa14890.c
@@ -305,7 +305,7 @@ static int cwa_parse_tlv(sc_card_t * card,
 			}
 		}
 		tlv->data = buffer + n + j;
-		tlv->buflen = j + tlv->len;;
+		tlv->buflen = j + tlv->len;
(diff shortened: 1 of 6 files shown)


3v1n0:
Thanks I can confirm that the write error mentioned in #3334 disappears with this.

Jakuje:
Thank you for prompt verification!

The follow-up question for the @OpenSC/maintainers is if this should be considered a security issue, given that it involves use-after-free. 

We probably missed that as we usually end up using p11-kit-proxy, which does not take this code path (but might end up having the same issue too).

xhanulik (approved):
Since you've started removing trailing spaces, could you please remove also the ones CI found? Otherwise looks good.

Jakuje:
> Since you've started removing trailing spaces, could you please remove also the ones CI found? Otherwise looks good.

Editor did that for me :) Afaik they are just duplicate newlines that should go away too, but once I touch it, did that.

frankmorgner (approved):

PR #3338, opened 2025-02-11
feat(pkcs11-tool): add support for AES CTR

AlexandreGonzalo:
Hi OpenSC team!

In this PR, I am adding the support for CKM_AES_CTR for the encrypt and decrypt options.
As you will see, the ulCounterBits is hard coded to 128 bits which is the only value supported by the HSM that I am using.
I could test the changes with the Trustonic TEE HSM and I could double check with OpenSSL3.

Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2819,6 +2819,7 @@ typedef union
 	CK_GCM_PARAMS gcm;
 	CK_CHACHA20_PARAMS chacha20;
 	CK_SALSA20_CHACHA20_POLY1305_PARAMS chacha20poly1305;
+	CK_AES_CTR_PARAMS ctr;
 } params_t;
 
 static void
@@ -2859,6 +2860,16 @@ build_params(
 		params->gcm.ulTagBits = opt_tag_bits;
 		mech->pParameter = &params->gcm;
 		mech->ulParameterLen = sizeof(params->gcm);
+		break;
+	case CKM_AES_CTR:
+		*iv = hex_string_to_byte_array(opt_iv, &iv_size, "IV");
+		if (iv_size != 16U) {
+			util_fatal("Invalid IV length %zu", iv_size);
+		}
+		memcpy(&params->ctr.cb[0], *iv, sizeof(params->ctr.cb));
+		params->ctr.ulCounterBits = 128U;
+		mech->pParameter = &params->ctr;
+		mech->ulParameterLen = sizeof(params->ctr);
 		break;
 	case CKM_CHACHA20:
 		*iv = hex_string_to_byte_array(opt_iv, &iv_size, "IV");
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -2830,7 +2833,7 @@ static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 	/* set "default" MGF and hash algorithms. We can overwrite MGF later */
 	switch (opt_mechanism) {
 	case CKM_RSA_PKCS_OAEP:
-		build_rsa_oaep_params(&oaep_params, &mech, NULL, 0);
+		build_rsa_oaep_params(&params.oaep, &mech, NULL, 0);
```
Jakuje, line 2836: As this is growing to more mechanisms, I am wondering what are the difference in encryption and decryption case so it could be all moved into separate function `build_mechanism` or something to avoid code duplication.

Jakuje (commented):
Code wise it looks good, but I would prefer less duplication.

I would also prefer to have the following test extended with the new mode. It should be pretty easy to add another section with the CTR modes:

https://github.com/OpenSC/OpenSC/blob/master/tests/test-pkcs11-tool-sym-crypt-test.sh

src/tools/pkcs11-tool.c
AlexandreGonzalo: This PR and the Chacha20 PRs are conflicting.
I propose to add the `build_mechanism` function in the Chacha20 PR and once it is merged, I will rework this one.
Is it OK for you?

tests/test-pkcs11-tool-sym-crypt-test.sh
```diff
@@ -181,6 +181,26 @@ assert $? "Fail/pkcs11-tool decrypt"
 cmp gcm_vector_plain.data gcm_test_plain.data >/dev/null 2>&1
 assert $? "Fail, AES-GCM - wrong decrypt"
 
+echo "======================================================="
+echo " AES-CTR"
+echo " OpenSSL encrypt, pkcs11-tool decrypt"
+echo " pkcs11-tool encrypt, compare to openssl encrypt"
+echo "======================================================="
+
+openssl enc -aes-128-CTR -nopad -in aes_plain.data -out aes_ciphertext_openssl.data -iv "${VECTOR}" -K "70707070707070707070707070707070"
```
Jakuje, line 190: sounds like libressl needs the ctr in lower case from the failure in the build-libressl job?
https://github.com/OpenSC/OpenSC/actions/runs/13284043279/job/37088623747?pr=3338

Jakuje (commented):
Thank you. Now the change looks much more digestible as it is only on one place!
The only concerning failure is the libressl -- can you  try to fix it with the suggestion inline? The other tests need some more love, but they fail also in master.

Jakuje (approved):

PR #3339, opened 2025-02-11
feat(pkcs11-tool): add support for ChaCha20 Poly1305

AlexandreGonzalo:
Hi OpenSC team!

In this PR, I am adding the support for Chacha20 and Poly1305.
This is adding 2 new key types: CKK_CHACHA20 and CKK_POLY1305.
These keys can be either generated with CKM_CHACHA20_KEY_GEN and CKM_POLY1305_KEY_GEN or imported.
CKM_CHACHA20 is for encryption, CKK_POLY1305 is for message authentication and CKM_CHACHA20_POLY1305 is for authenticated encryption.
For CKM_CHACHA20, the nonce and initial counter value are passed via the IV option (128-bit).
I decided to use by default 64-bit for the nonce and 64-bit for the initial counter value (Bernstein's original variant), like it is done in OpenSSL.
For CKM_CHACHA20_POLY1305, the nonce is given via the IV option and the AAD via the AAD option like we did for AES GCM.

I could test the changes with the Trustonic TEE HSM and I could double check with OpenSSL3.

Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2813,29 +2813,91 @@ build_chacha20_params(
 	mech->ulParameterLen = sizeof(*params);
 }
 
-static void decrypt_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
-		CK_OBJECT_HANDLE key)
+typedef union
 {
-	unsigned char	in_buffer[1024], out_buffer[1024];
-	CK_MECHANISM	mech;
-	CK_RV		rv;
-	union
-	{
-		CK_RSA_PKCS_OAEP_PARAMS oaep;
-		CK_GCM_PARAMS gcm;
-		CK_CHACHA20_PARAMS chacha20;
-		CK_SALSA20_CHACHA20_POLY1305_PARAMS chacha20poly1305;
-	} params = {0};
-	CK_ULONG	in_len, out_len;
-	int		fd_in, fd_out;
-	CK_BYTE_PTR	iv = NULL;
-	size_t		iv_size = 0;
-	ssize_t sz;
-	CK_BYTE_PTR aad = NULL;
+	CK_RSA_PKCS_OAEP_PARAMS oaep;
+	CK_GCM_PARAMS gcm;
+	CK_CHACHA20_PARAMS chacha20;
+	CK_SALSA20_CHACHA20_POLY1305_PARAMS chacha20poly1305;
+} params_t;
+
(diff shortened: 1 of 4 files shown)


Jakuje:
Thank you! Looks good to me. I would prefer to make the clang-format a bit more happy. Do not apply all the reformat changes in pkcs11.h for example, but the spacing around operators and `else` can really make the code easier to read!

Jakuje (approved):

src/tools/pkcs11-tool.c
```diff
@@ -4822,6 +4808,10 @@ static CK_RV write_object(CK_SESSION_HANDLE session)
 				type = CKK_GENERIC_SECRET;
 			else if (strncasecmp(opt_key_type, "HKDF:", strlen("HKDF:")) == 0)
 				type = CKK_HKDF;
+			else if (strncasecmp(opt_key_type, "CHACHA20", strlen("CHACHA20")) == 0)
+				type = CKK_CHACHA20;
+			else if (strncasecmp(opt_key_type, "POLY1305", strlen("POLY1305")) == 0)
+				type = CKK_POLY1305;
```
frankmorgner, line 4814: Please add documentation of the new values here:
https://github.com/OpenSC/OpenSC/blob/9b0313d3caf154b1506e90a03d92cd34d176b749/doc/tools/pkcs11-tool.1.xml#L146-L151

src/tools/pkcs11-tool.c
```diff
@@ -4822,6 +4808,10 @@ static CK_RV write_object(CK_SESSION_HANDLE session)
...
 			else
 				util_fatal("Unknown key type: 0x%lX", type);
```
frankmorgner, line 4816: This error message should contain the possible values that are recognized

frankmorgner (changes requested):

Jakuje (commented):
Thank you! Looks good to me. The last thing would be to run `make tools.html` in `doc/tools` as we do not have automatic update of HTML pages, as suggested here:

https://github.com/OpenSC/OpenSC/actions/runs/13280992311/job/37079146770?pr=3339

frankmorgner (approved):
Thank you, looks good

Jakuje:
Thank you! Merging to let you move forward with #3338.

AlexandreGonzalo:
> Thank you! Merging to let you move forward with #3338.

Thank you both for your review!

PR #3340, opened 2025-02-12
feat(pkcs11-tool): add --private-privkey option

AlexandreGonzalo:
Hi guys,

I am restarting this https://github.com/OpenSC/OpenSC/pull/3177.


src/tools/pkcs11-tool.c
```diff
@@ -3410,6 +3415,16 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+	if (opt_is_private_privkey != 0) {
+		FILL_ATTR(privateKeyTemplate[n_privkey_attr], CKA_PRIVATE,
+				&_true, sizeof(_true));
+		n_privkey_attr++;
+	} else {
+		FILL_ATTR(privateKeyTemplate[n_privkey_attr], CKA_PRIVATE,
+				&_false, sizeof(_false));
+		n_privkey_attr++;
+	}
+
```
Jakuje, line 3427: This changes default, which might come unexpected for some of our users. I would say the private key should have PRIVATE attribute by default so instead of changing this, I would probably propose to go with the negative option, either `--no-private-privkey` to keep backward compatibility and allow generating (and writing) non-private private keys, which is the aim if I see right.

AlexandreGonzalo: I simply implemented what you wanted in the previous PR ;-)
But I agree that I don't want to change the default behavior.
What about simplifying and use `--no-private` and both the public and private keys will have the CKA_PRIVATE=false ?

Jakuje: Yes, no private sounds reasonable to me.

AlexandreGonzalo: Hi,
I was wondering that this PR is probably useless and I should decline it.

Let me explain.
In the past our HSM didn't support login and CKA_PRIVATE.
So it was an issue that when I was using the pkcs11-tool, by default the key pairs were generated with the CKA_PRIVATE attribute.
I had a patch to change this default behavior.

Last year, I added the support for login and CKA_PRIVATE, so the default behavior should no longer be an issue and I don't need this patch anymore.

So last week, I started to update our tests to take into account this new behavior.
Whenever a private key is involved, we have to login as user because the key has the CKA_PRIVATE attribute.
It is working fine when I am using the pkcs11-tool, for example for signatures but it does not work with OpenSSL.
I tried with both the libp11 engine with OpenSSL v1.1.1 and the pkcs11-provider with OpenSSL v3.
The issue is that even if I am providing the "pin-value" as part of the URL, the engine or provider is not calling C_Login.
My understanding is that it does not try to log in because the "login is not required by the token".
True, our HSM is not setting the flag CKF_LOGIN_REQUIRED.

There is not much information about the flag in the specifications.
_CKF_LOGIN_REQUIRED: True if there are some cryptographic functions that a user MUST be logged in to
perform._

It is really vague and subject to various interpretations.

I thought, why not setting this flag?
Then, I tried to set this flag but I faced a new issue.
When this flag is set, we need to log in and provide a PIN for operations that involve public or secret keys which don't have the CKA_PRIVATE attribute.
It does not seem to be a proper solution.

What am I missing?
You probably have more experience in this than me, any help is much appreciated.

dengert: A user pin and CKF_LOGIN_REQUIRED are used on 99% of smart cards. If your HSM does not require a PIN how do you control access to the HSM on a multiuser system?

One example of the other 1% is a PIV card has the 9E key which does not require a login, as its intent is use it to provide proof of position of the card, such as a door lock.  (Yubikey has a PIV applet, you could try using one for testing.) 

As you noted, the pkcs11-tool and OpenSC PKCS11 module are smart enough to not do a login if the only key being used does not require a login, even if CKF_LOGIN_REQUIRED is set. 

You should be able to do public key operations, but for 99% of smart cards the public key (or certificate) can be read and used in software, which is much faster.  (It is not clear if allowing symmetric key operations to be done without a login is a security risk.)

Also note that PKCS11 has CKS_RO_PUBLIC_SESSION,  CKS_RO_USER_FUNCTIONS, CKS_RW_PUBLIC_SESSION,  CKS_RW_USER_FUNCTIONS and CKS_RW_SO_FUNCTIONS. which are changed when a C_Login is done from _PUBLIC to _USER. or if C_LOGIN  is done with the SO PIN. 

It may be libp11 and provider do no support the concept of a private key which does not require a login.

dengert: Libp11 has a CMD_FORCE_LOGIN you may want to look at. Provider may have something similar.

PR #3342, opened 2025-02-13
fix(pkcs11-tool): fixes for Ed448 and X448

AlexandreGonzalo:
Hi OpenSC team,

In the past, I was maintaining my own set of patches for the Edwards curves.
I have noticed that you recently made many changes for these curves and I wanted to give it a try.
Please have a look at the changes I had to to make my tests pass again.

Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6556,9 +6556,31 @@ static int read_object(CK_SESSION_HANDLE session)
 						ASN1_PRINTABLESTRING_free(curve);
 					} else if (d2i_ASN1_OBJECT(&obj, &a, (long)len) != NULL) {
 						int nid = OBJ_obj2nid(obj);
-						if (nid != NID_ED25519 && nid != NID_X25519) {
-							util_fatal("Unknown curve OID, expected NID_ED25519 (%d), got %d",
-									NID_ED25519, nid);
+						if (nid != NID_ED25519 &&
+								nid != NID_X25519
+#if defined(EVP_PKEY_ED448)
+								&& nid != NID_ED448
+#endif
+#if defined(EVP_PKEY_X448)
+								&& nid != NID_X448
+#endif
+						) {
+							util_fatal("Unknown curve OID, expected NID_ED25519 (%d), NID_X25519 (%d), "
+#if defined(EVP_PKEY_ED448)
+								   "NID_ED448 (%d), "
+#endif
+#if defined(EVP_PKEY_X448)
+								   "NID_X448 (%d), "
+#endif
+								   "got %d",
+									NID_ED25519, NID_X25519,
+#if defined(EVP_PKEY_ED448)
+									NID_ED448,
+#endif
+#if defined(EVP_PKEY_X448)
+									NID_X448,
+#endif
+									nid);
 						}
 						ASN1_OBJECT_free(obj);
 					} else {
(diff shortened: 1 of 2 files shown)


src/tools/pkcs11-tool.c
```diff
@@ -6556,9 +6556,13 @@ static int read_object(CK_SESSION_HANDLE session)
...
-							util_fatal("Unknown curve OID, expected NID_ED25519 (%d), got %d",
-									NID_ED25519, nid);
+						if (nid != NID_ED25519 &&
+								nid != NID_X25519 &&
+								nid != NID_ED448 &&
+								nid != NID_X448) {
+							util_fatal("Unknown curve OID, expected NID_ED25519 (%d), NID_X25519 (%d), NID_ED448 (%d), NID_X448 (%d), got %d",
+									NID_ED25519, NID_X25519,
+									NID_ED448, NID_X448, nid);
 						}
```
dengert, line 6566: This may be a problem if using libressl and they did not define all the nids. May need some ifdefs.

Note that OpenSC currently does not support any tokens that can do ED448 or X448. SO if you have one that would be a first. 

Code in pkcs11-tool for these was added but never tested.   

This looks reasonable.

dengert:
Are you saying that even though the name *448 implies 56 bytes, it really need 57?
Can you show where that is documented? Did you get a failure because of this?

There are a lot of places in the new code for ED448 were the bytes are set at 448 or using BYTES4BITS(448). that will need to be changed.

dengert:
Good catch, Looking closer at PKCS11 3.1 and RFC8032 the EC_POINT for the ED25512 is 32 bytes, and for the ED448 it is 57 bytes, and the signature is 114 bytes.

src/tools/pkcs11-tool.c
AlexandreGonzalo: I added the missing #idef for libressl.
Let's see how it goes with the automatic checks.

src/tools/pkcs11-tool.c
```diff
@@ -6556,9 +6556,31 @@ static int read_object(CK_SESSION_HANDLE session)
...
-							util_fatal("Unknown curve OID, expected NID_ED25519 (%d), got %d",
-									NID_ED25519, nid);
+						if (nid != NID_ED25519 &&
+								nid != NID_X25519
+#if defined(EVP_PKEY_ED448)
+								&& nid != NID_ED448
+#endif
+#if defined(EVP_PKEY_X448)
+								&& nid != NID_X448
+#endif
```
Jakuje, line 6566: Do we know about some implementation that would have only one of these which would prevent us writing something like this?
```suggestion
#if defined(EVP_PKEY_ED448) && defined(EVP_PKEY_X448)
								&& nid != NID_ED448 && nid != NID_X448
#endif
```

AlexandreGonzalo: I can't tell you, I am only use OpenSSL v1.1.1 and OpenSSL v3.
Please, let me know if I take your suggested changes.

dengert:
@AlexandreGonzalo @Jakuje 

Here is a diff  (untested) that  changes the field_length which is only  to calculate bytes need for Edwards and Montgomery curves
for EC_POINT and signatures. I think I have them all. 

In curent code these we 255 and 448 In new code use 255 + 1 and 448 + 1.  It works for 25519 as BYTES4BITS(256) is same as BYTES4BITS(256 + 1)  And used the +1 mor as a comment to then reader. 

[ed448.extra-byte.git.patch.txt](https://github.com/user-attachments/files/18801623/ed448.extra-byte.git.patch.txt)

Please give this a try,and add to your PR as needed.

AlexandreGonzalo:
> The previous patch https://github.com/user-attachments/files/18801623/ed448.extra-byte.git.patch.txt is incorrect and makes too many changes.
> 
> This patch is much simpler defining SC_X448_OUTPUT_BYTES 57 and using in place of BYTES4BITS(448) for X448:
> 
> [X448.patch.txt](https://github.com/user-attachments/files/18810915/X448.patch.txt) It compiles, but untested as OpenSC does not have any tokens that support X448.

Hi Doug,
The key size for X448 is 56 bytes and for ED448 it is 57 bytes.
I just update this PR and I am now using the definition ED448_KEY_SIZE_BYTES instead of a magic number.

Jakuje (approved):
I think we are good here. Than you!

PR #3344, opened 2025-02-13
feat(pkcs11-tool): CKA_ALLOWED_MECHANISMS is also for public and secr…

AlexandreGonzalo:
Hi guys,

I recently added the support for the CKA_ALLOWED_MECHANISMS attribute in the Trustonic TEE HSM and I wanted to use the pkcs11-tool for the testing of the feature.
This PR is allowing to set this attribute for public and secret keys.

Regards,
Alexandre.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -3423,6 +3423,10 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 	}
 
 	if (opt_allowed_mechanisms_len > 0) {
+		FILL_ATTR(publicKeyTemplate[n_pubkey_attr],
+				CKA_ALLOWED_MECHANISMS, opt_allowed_mechanisms,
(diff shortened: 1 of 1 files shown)


Jakuje:
Even though the allowed mechanisms are specified as common key attributes, I find it they do not make much sense for the public key attributes in practice, as the public keys can be easily exported from the token/hsm and used to whatever mechanism the user wants.

I do not see any (unexpected) failure so I think we are good to take this.

Thank you for your continuous contributions!

AlexandreGonzalo:
> Even though the allowed mechanisms are specified as common key attributes, I find it they do not make much sense for the public key attributes in practice, as the public keys can be easily exported from the token/hsm and used to whatever mechanism the user wants.
> 
> I do not see any (unexpected) failure so I think we are good to take this.
> 
> Thank you for your continuous contributions!

I can't tell you if there is such a use case for public keys, anyway I had to test my implementation and I found it useful to add it to pkcs11-tool.
You're welcome!

PR #3348, opened 2025-02-19
Fix Edwards and Montgomery "field_length"

dengert:
The Edwards and Montgomery "field_length" is changed from 255 to 256 and for Ed448, 448 is changed to 456 as it needs an extra byte.

The field_length in bits, and is used to caculate number of bytes.

The BYTES4BITS(255) returns 32 and happens to work but should be BTYES4BITS(256) so the 25519 curves have been working but not as expected. But BYTES4BITS(448) returns 56, but Ed448 needed an extra byte so will use BYTES4BITS(456) to return 57 as required in the RFCs.

 On branch ED448-X448-fix
 Changes to be committed:
	modified:   src/libopensc/pkcs15-pubkey.c

##### Checklist

- [ ] macOS tokend is tested

### DIFF
--- a/src/libopensc/pkcs15-pubkey.c
+++ b/src/libopensc/pkcs15-pubkey.c
@@ -1524,17 +1524,18 @@ static struct ec_curve_info {
 
 		/* OpenPGP extensions by Yubikey and GNUK are not defined in RFCs but we know the oid written to card */
 
-		{"edwards25519",	"1.3.6.1.4.1.11591.15.1", {(u8 *)"\x06\x09\x2B\x06\x01\x04\x01\xDA\x47\x0F\x01", 11}, 255, SC_ALGORITHM_EDDSA},
-		{"curve25519",		"1.3.6.1.4.1.3029.1.5.1", {(u8 *)"\x06\x0A\x2B\x06\x01\x04\x01\x97\x55\x01\x05\x01", 12}, 255, SC_ALGORITHM_XEDDSA},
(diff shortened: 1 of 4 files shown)


Jakuje:
I am not completely sure this is the right solution. If we consider the value to be a field length than it is not correct as the field length of ed25519 is really 255 and not 256. If we consider it an output length of full bytes converted to bits, its odd, but it might be right. But then we should change also the 521 -> 528 for P521 curve and maybe some others in the list.

The thing is that the 255->32 is converted correctly with the BYTES4BITS macro and that is what the macro is supposed to be doing. The only case where it does not work is the oddball Ed448 so I would prefer to change only that one with appropriate comment as you have there. I would prefer to keep the rest as it was.

dengert:
So do you want to leave the 25519 at 255 and only change Ed448 entries to use something like `#define SC_ED448_BITS 456` ?

Jakuje:
Yes, that would be my suggestion, but I am open for discussion if there are other/better ideas how to handle, especially from @AlexandreGonzalo or others using the keys.

dengert:
@Jakuje  Rebased as requested.
@AlexandreGonzalo  Please comment on this PR, and any suggestions as noted by @Jakuje and myself above.

frankmorgner:

Jakuje (approved):

frankmorgner (approved):

Jakuje:
I think this one is pretty straight-forward and do not expect this would break so merging even though the broken CI fails to report many jobs.

PR #3354, opened 2025-02-25
win: remove pkcs11-register from autostart

frankmorgner:
Since on Windows Firefox switched to the OS' native method for hardware tokens, FF integration is already disabled by default on those platforms in pkcs11-register. There now is not much to register anymore.

Closes https://github.com/OpenSC/OpenSC/pull/3218
Closes https://github.com/OpenSC/OpenSC/pull/3336
Fixes  https://github.com/OpenSC/OpenSC/issues/3217

##### Checklist

- [ ] macOS tokend is tested


frankmorgner:
I'd like to merge this if there are no objections

PR #3355, opened 2025-02-25
Configure --enable-piv-sm does not work with LibreSSL

dengert:
Configuring with --enable-piv-sm and LibreSSL fails when compiling.

card-piv.c checks for this and other requirements and does "#undef ENABLE_PIV_SM" to turn off ENABLE_PIV_SM.
But config.h, is include a second time by internal.h which redefines ENABLE_PIV_SM.

The commit reorders the includes so the above does not happen.

 On branch PIV-SM-LibreSSL
 Changes to be committed:

##### Checklist
Tested with and without --enable-piv-sm with OpenSSL-3.4.1 and LibreSSL-4.0.0

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -49,18 +49,17 @@
 #endif
 #endif
 
+#include "internal.h"
+
 /* 800-73-4 SM and VCI need: ECC, SM and real OpenSSL >= 1.1 */
 #if defined(ENABLE_OPENSSL) && defined(ENABLE_SM) && !defined(OPENSSL_NO_EC) && !defined(LIBRESSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+#include <openssl/cmac.h>
 #else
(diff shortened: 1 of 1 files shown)


src/libopensc/card-piv.c
```diff
@@ -49,18 +49,17 @@
 #endif
 #endif
 
+#include "internal.h"
+
 /* 800-73-4 SM and VCI need: ECC, SM and real OpenSSL >= 1.1 */
 #if defined(ENABLE_OPENSSL) && defined(ENABLE_SM) && !defined(OPENSSL_NO_EC) && !defined(LIBRESSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+#include <openssl/cmac.h>
```
frankmorgner, line 56: This kind of check for prequisites should be done at configure time, when the user explicitly selects `--enable-piv-sm`. At compile time the user expects the feature to be included.

frankmorgner (changes requested):

PR #3358, opened 2025-02-26
Configure.ac - --enable-piv-sm now test for --enable_sm and --enable-openssl 

dengert:
card-piv.c now relies on configure tests for SM and OpenSSL which are needed for PIV-SM.   

--enable-piv-sm can also use LibreSSL as well as OpenSSL.

 Changes to be committed:
	modified:   configure.ac

Tested with Idemia test PIV card and the PIV SM works with OpenSSL-3.4.1 and Libressl-4.0.0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -861,7 +861,11 @@ else
 fi
 
 if  test "${enable_piv_sm}" = "yes"; then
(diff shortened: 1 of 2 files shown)


frankmorgner (approved):

Jakuje:
Thank you! Looks good!

PR #3360, opened 2025-03-01
CI: remove SoftHSM from OpenSC CI tests

frankmorgner:
SoftHSM suppressed errors show up as new errors in updated cmocka library

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -39,16 +39,6 @@ dist_noinst_SCRIPTS = common.sh \
 .NOTPARALLEL:
 TESTS = \
         test-manpage.sh \
-        test-duplicate-symbols.sh \
-        test-pkcs11-tool-test-threads.sh \
-        test-pkcs11-tool-allowed-mechanisms.sh \
-        test-pkcs11-tool-sym-crypt-test.sh
-if ENABLE_OPENSSL
-TESTS += \
-        test-pkcs11-tool-test.sh \
-        test-pkcs11-tool-sign-verify.sh \
-        test-pkcs11-tool-unwrap-wrap-test.sh \
-        test-pkcs11-tool-import.sh
-endif
+        test-duplicate-symbols.sh
 # no tests expected to fail
 #XFAIL_TESTS =
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -49,11 +49,6 @@ TESTS += \
         test-pkcs11-tool-sign-verify.sh \
         test-pkcs11-tool-unwrap-wrap-test.sh \
         test-pkcs11-tool-import.sh
-if ENABLE_TESTS
-if ENABLE_SHARED
-TESTS += test-p11test.sh
-endif
-endif
 endif
 # no tests expected to fail
 #XFAIL_TESTS =


Jakuje:
The CI still dumps two uninitialized memory issues which are missing:
```
  {
     <insert_a_suppression_name_here>
     Memcheck:Value8
     fun:_itoa_word
     fun:__printf_buffer
     fun:__vfprintf_internal
     fun:printf
     fun:supported_mechanisms_test
     obj:/usr/lib/x86_64-linux-gnu/libcmocka.so.0.8.0
     fun:_cmocka_run_group_tests
     fun:main
  }
  {
     <insert_a_suppression_name_here>
     Memcheck:Cond
     fun:_itoa_word
     fun:__printf_buffer
     fun:__vfprintf_internal
     fun:printf
     fun:supported_mechanisms_test
     obj:/usr/lib/x86_64-linux-gnu/libcmocka.so.0.8.0
     fun:_cmocka_run_group_tests
     fun:main
  }
```

frankmorgner:
I've updated the workflow file and realized that the log contains a ton of other problems that seem to origin in SoftHSM. I've opened a ticket here, https://github.com/softhsm/SoftHSMv2/issues/785.

I wonder how useful a valgrind analysis of SoftHSM within the OpenSC pipeline actually is. This basically only tests pkcs11-tool as part of OpenSC, which should already have good coverage. My suggestion would be to remove those valgrind runs involving SoftHSM to avoid this kind of problems.

Jakuje:
Yeah, the softhsm is mess. I think the uninitialized memory in mechanism list was actually fixed upstream for years, but the new version was never released nor ubuntu pulled the changes from upstream. I was trying to fix it some years back in fedora by backport, but it whole exploded so I gave up.

We test by this mostly handling of the test and tools themselves, which I still find useuful.

@xhanulik is already working on testing against kryoptic, which should not have these issues.

dengert:
https://github.com/softhsm/SoftHSMv2 with tag 2.6.1 from 2020-04-29, but the develop branch has 150 or so commits as late as 2025-01-28  and a master branch 6 years old  and 127 open issues including Franks from today. 

Would it make any sense to try the develop branch instead of 2.6.1?

Jakuje:
> My preferred way of moving forward would be to exclude SoftHSM from `make check`. The code in this PR was modified accordingly.

That wont solve the valgrind issue as the softhsm is used as part of other tests under `tests/`.

frankmorgner:
I disabled the tests of pkcs11-tool with softhsm as well.

The valgrind workflow is now functional again, even though it doesn't do a lot anymore. However, we still have a lot of valgrind testing going on in `.github/test-*.sh` during CI.

xhanulik:
I'm currently working on adding some CI tests similar to those in `tests/` with kryoptic and NSS softokn. I think the softokn could be used instead of softhsm in `make check`.

frankmorgner:
The target `check-valgrind-memcheck` comes directly from `m4/ax_valgrind_check.m4` (similar to `check-code-coverage` from `m4/ax_code_coverage.m4`), I don't think it reflects some kind of flexibility or automatism.

In Github's CI we put different tests into different jobs so that we can easily locate a possible problem and who might be responsible for fixing this problem. If the job `test-piv` fails, then I assume Doug would be most interested. If some generic `build` job fails, then I assume something internally goes wrong, which lets me assume that I should have a look. If we include too much external dependencies into `make check`, which is often called as part of the build job, then that job may fail due to many other reasons, that I often don't feel much responsibility for.

Adding a command line switch to `configure.ac` (and `tests/Makefile.am`) which is triggered by some specific job is certainly possible, but it adds complexity to the build system, which is not needed. This specific job can directly invoke the test needed without a customized build pipeline.

dengert:
Yes I am interested if test-piv fails, but when I run make check on my own machine, it works, but when  run on a CI machine, some usually fails.

Jakuje:
You are right. We still want to run the unit tests with valgrind so my initial suggestion wont work. But introducing a configure switch to skip the "integration" tests under valgrind should not add that much complexity that we could not bear. With your change, we would be left with running the manpage and duplicate symbols tests and really neither of them make sense with valgrind. Can you have a look at #3370 to share your thoughts?

frankmorgner:
Well, if you insist, we can move forward with #3370 instead of the change proposed here, but please note that we do have plenty of valgrind checking going on when running the CI tests with the simulated cards, even though their job is not named someting with valgrind.

PR #3362, opened 2025-03-03
card-itacns: Try to get the card name from the card itself

3v1n0:
##### Checklist
- [ ] macOS token is tested

The card name may vary on these cards even though is currently set
to 'Carta Nazionale dei Servizi', but let's not hardcode it, while
get instead it from a path that is also shared by other IDEMIA cards.


src/libopensc/card-itacns.c
```diff
@@ -81,6 +81,39 @@ static void itacns_init_cns_card(sc_card_t *card)
...
+}
+
+static void itacns_init_idemia_cns_card(sc_card_t *card)
+{
+	u8 buf[512] = {0};
+
+	if (itacns_get_card_name(card, buf, sizeof(buf)) == SC_SUCCESS) {
+		DRVDATA(card)->card_name = strndup ((const char*) buf,
+						    sizeof (buf));
+		card->name = DRVDATA(card)->card_name;
```
frankmorgner, line 113: please make sure that the card name is a printable string (`isprint()`)

frankmorgner:
There are also build errors in mingw and formatting changes needed

frankmorgner:

frankmorgner:

PR #3343, opened 2025-02-13
feat(pkcs11-tool): allow EC keys to have CKA_ENCRYPT or CKA_DECRYPT

AlexandreGonzalo:
Hi again!

The Trustonic TEE HSM has support for SM2 PKE which is an asymmetric cipher.
We need to be able to generate SM2 key pairs with the CKA_ENCRYPT and CKA_DECRYPT attributes.
(that's already supported for key import)

Regards,
Alexandre.

PR #3364, opened 2025-03-05
Do not add CKA_ENCRYPT or CKA_DECRYPT for EC keys by default

xhanulik:
Setting CKA_ENCRYPT or CKA_DECRYPT by default for new keys in `pkcs11-tool` breaks the EC key generation.
With these changes, the CKA_ENCRYPT or CKA_DECRYPT can be set using `--usage-decrypt` option for `pkcs11-tool`.
Tested with NSS softokn.


xhanulik:
@AlexandreGonzalo as this change was originally added in #3343, does this work for you?

Jakuje (approved):

Jakuje:
Thank you! Merged.

PR #3365, opened 2025-03-06
Add CI tests for pkcs11-tool with kryoptic and softokn

xhanulik:
This PR adds new CI jobs for testing pkcs11-tool with kryoptic and softokn SW tokens.
(The c56b998f1f09a80f08918fb5c8a6ca63c005bf1c from #3364 unblocks the EC generation for softokn)

##### Checklist

- [ ] macOS token is tested


.github/workflows/pkcs11.yaml
```diff
@@ -0,0 +1,66 @@
+name: pkcs11-tool tests
```
frankmorgner, line 1:
```suggestion
name: tests of external pkcs11 modules
```

frankmorgner, line 1: maybe rename pkcs11.yaml into external-pkcs11.yaml

frankmorgner (changes requested):
1. I like having the tests of other PKCS#11 modules inside a dedicated job and outside of OpenSC's `make check`. I think this could be extended into testing softhsm as well (see my comments in #3360).
2. I am confused why you put all your scripts into `.github`. Given the current structure, I would suggest to split this into scripts for the software setup, which live in `.github`, and scripts for testing and initializing the modules, which live in `tests`.
3. Please review `pkcs11-common.sh` with `tests/common.sh`, there is likely some overlap that can be reused.

xhanulik:
Thanks @frankmorgner for review, I've moved the test scripts into `tests/` and _somehow_ unified it with the ones already there. Now there is possibility to choose token type by the particular test, and the token is set up and the tests are run on it. I'll summarize it into some README and fix the failing libressl build.

frankmorgner (approved):
LGTM and should go well together with #3360. Thanks for the refactoring.

.github/setup-kryoptic.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# build kryoptic
+if [ ! -d "kryoptic" ]; then
+	git clone https://github.com/latchset/kryoptic.git
+fi
+pushd kryoptic
+git submodule init
+git submodule update
```
Jakuje, line 9: nit: When we do dynamic build, we should not need to pull submodules.
Also please add a note to remove the custom build once we will have final package in Fedora.

.github/workflows/external-pkcs11.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Setup Kryoptic
+        run: .github/setup-kryoptic.sh
+      - name: Test Sign & Verify with Kryoptic
+        run: tests/test-pkcs11-tool-sign-verify.sh kryoptic
+      - name: Test Import with Kryoptic
+        run: tests/test-pkcs11-tool-import.sh kryoptic kryoptic
+      - name: Test Allowed Mechanisms with Kryoptic
+        run: tests/test-pkcs11-tool-allowed-mechanisms.sh kryoptic
+      - name: Test pkcs11-tool test with Kryoptic
+        run: tests/test-pkcs11-tool-test.sh kryoptic
```
Jakuje, line 56: I would rather have this in separate script that could be executed locally easily

tests/test-p11test.sh
```diff
@@ -11,8 +11,10 @@ if [[ ! -f $P11LIB ]]; then
     exit 77;
 fi
 
-# The Ubuntu has old softhsm version not supporting this feature
-grep "Ubuntu 18.04" /etc/issue && echo "WARNING: Not supported on Ubuntu 18.04" && exit 77
+if [ "${TOKENTYPE}" == "softhsm" ]; then
+    # The Ubuntu has old softhsm version not supporting this feature
+    grep "Ubuntu 18.04" /etc/issue && echo "WARNING: Not supported on Ubuntu 18.04" && exit 77
+fi
```
Jakuje, line 17: not sure if this is relevant anymore. Ubuntu 18 is EOL (except for Extended support), but certainly for the CI use so I would just drop this condition.

.github/workflows/external-pkcs11.yaml
```diff
@@ -0,0 +1,114 @@
...
+      BUILD_PATH: ./
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Setup Fedora
+        run: .github/setup-fedora.sh kryoptic
+      - name: Install OpenSC
+        run: .github/build.sh
+      - name: Setup Kryoptic
+        run: .github/setup-kryoptic.sh
```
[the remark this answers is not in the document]
xhanulik: This one is only about installing Kryoptic module, the oen called from `common.sh` is about initialization of the token (as done for softokn and softhsm). I will rename it to prevent confusion.

tests/common.sh
```diff
@@ -70,46 +63,31 @@ function generate_key() {
...
-		echo "Couldn't generate GENERIC key"
-		return 1
-	fi
-}
 
-function softhsm_cleanup() {
-	rm .softhsm2.conf
-	rm -rf ".tokens"
-	sleep 1
+	if [[ ${TOKENTYPE} == "softhsm" ]]; then
```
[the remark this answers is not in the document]
xhanulik: For softokn, the key generation returns
```
error: PKCS11 function C_Login failed: rv = CKR_USER_TYPE_INVALID (0x103)
```
Kryoptic can generate the key, but using it for signature for all `SHA_-HMAC` ends with
```
error: PKCS11 function C_SignInit failed: rv = CKR_KEY_TYPE_INCONSISTENT (0x63)
```
I haven't found the reason for these yet.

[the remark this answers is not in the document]
Jakuje: Thanks for clarification. Please report the issues with reproducers to kryoptic so we can have a look into that eventually.

Jakuje (approved):

xhanulik:
Added tests for wrap/unwrap and symmetric encrypt/decrypt for kryoptic.

frankmorgner:
> Added tests for wrap/unwrap and symmetric encrypt/decrypt for kryoptic.

Is this a step forward for resolving https://github.com/OpenSC/OpenSC/issues/1796?

Jakuje:
> > Added tests for wrap/unwrap and symmetric encrypt/decrypt for kryoptic.
> 
> Is this a step forward for resolving #1796?

I do not think so. The problem in #1796 is that different pkcs11 modules have different expectation or interpretation of the specification and choke on different inputs

frankmorgner:
So I think this is good to be merged.

frankmorgner:
Oh, but the decision to #3370, you may actually want to integrate the pkcs#11 tests from this PR into `make check` (similar to the tests with softhsm)

xhanulik:
> > > Added tests for wrap/unwrap and symmetric encrypt/decrypt for kryoptic.
> > 
> > 
> > Is this a step forward for resolving #1796?
> 
> I do not think so. The problem in #1796 is that different pkcs11 modules have different expectation or interpretation of the specification and choke on different inputs

Yes, I just enabled the already existing tests for SoftHSM to work also with Kryoptic.

> Oh, but the decision to #3370, you may actually want to integrate the pkcs#11 tests from this PR into `make check` (similar to the tests with softhsm)

I can do that, but there should probably be some more sophisticated search for softokn and cryoptic PKCS#11 modules. I can proceed with it in another PR if you want to merge this one.

xhanulik:
Rebased to get rid of failing CI tests results. Otherwise I think it's good to merge. I would add the tests for kryoptic and softokn into `make check` in separate PR.

Jakuje:
Thank you for the work on this! CI is good so merging.

PR #3371, opened 2025-03-14
Fix memory leaks in `piv-tool` key generation

xhanulik:
Recently reported issue from OSS-Fuzz (https://issues.oss-fuzz.com/issues/397209407).

##### Checklist
- [ ] macOS token is tested

### DIFF
--- a/src/tools/piv-tool.c
+++ b/src/tools/piv-tool.c
@@ -374,8 +374,8 @@ static int gen_key(const char * key_info)
 			fprintf(stderr, "gen_key failed %d\n", r);
 			free(keydata.pubkey);
 			free(keydata.exponent);
-			EVP_PKEY_free(evpkey);
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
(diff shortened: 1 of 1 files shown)


dengert (approved):

src/tools/piv-tool.c
```diff
@@ -503,29 +508,25 @@ static int gen_key(const char * key_info)
...
 
-		free(keydata.ecpoint);
-		keydata.ecpoint_len = 0;
 		BN_free(x);
 		BN_free(y);
 
 		if (r == 0) {
 			sc_log_openssl(ctx);
 			fprintf(stderr, "EC_POINT_set_affine_coordinates_GFp failed\n");
-			EVP_PKEY_free(evpkey);
```
Jakuje, line 528: I think this free is still needed as the evpkey is allocated with the openssl < 3 unconditionally. Same above.

Jakuje (changes requested):

Jakuje (approved):

PR #3128, opened 2024-04-26
build: add support for meson

H5117:
Fixes #3110, https://github.com/OpenSC/OpenSC/issues/2577.

PR #3378, opened 2025-03-20
Automatically create Software Bill of Material

frankmorgner:
This PR builds on #2754 and utilizes [cdxgen](https://github.com/CycloneDX/cdxgen)'s capabilities of dependency tracking via the meson build files. The workflow linux-meson is modified so that the SBOM is distributed along with opensc-pkcs11.so.

I am personally more interested in generating a [cryptographic BOM](https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-CBOM-en.pdf), but that doesn't currently seem to be supported.

##### Checklist


frankmorgner:
An example run can be found here
https://github.com/frankmorgner/OpenSC/actions/runs/13969926949
with a sbom.json included in this archive
https://github.com/frankmorgner/OpenSC/actions/runs/13969926949/artifacts/2788231172

subprojects/packagefiles/openpace/meson.build
```diff
@@ -0,0 +1,135 @@
...
+    conf.set('HAVE_DH_SET0_KEY', true)
+    conf.set('HAVE_DH_GET0_PQG', true)
+    conf.set('HAVE_DH_SET0_PQG', true)
+    conf.set('HAVE_ECDSA_SIG_GET0', true)
+    conf.set('HAVE_ECDSA_SIG_SET0', true)
+    conf.set('HAVE_EC_POINT_GET_AFFINE_COORDINATES', true)
+    conf.set('HAVE_EC_POINT_SET_AFFINE_COORDINATES', true)
+    conf.set('HAVE_ASN1_STRING_GET0_DATA', true)
+else
+    conf.set10('HAVE_DECL_OPENSSL_ZALLOC',
```
hardening, line 41: You should do something like `cc = meson.get_compiler('c')` and use cc afterward

frankmorgner: I think this comment should go to #3128

PR #3380, opened 2025-03-21
Build on Linux arm64

metsma:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -32,55 +32,46 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-24.04
+    name: build-${{ matrix.dist }}-${{ matrix.arch }}-${{ matrix.build }}
+    strategy:
+      matrix:
+        dist: ['22.04', '24.04']
+        arch: ['amd64', 'arm64']
+        build: ['dist', 'valgrind']
+        exclude:
+          - build: dist
+            arch: arm64
+    runs-on: ubuntu-${{ matrix.dist }}${{ matrix.arch == 'arm64' && '-arm' || '' }}
     steps:
       - uses: actions/checkout@v4
       - uses: awalsh128/cache-apt-pkgs-action@v1.4.3
         with:
           packages: ${{ env.BASE_DEPS }}
-          version: apt-20-base
+          version: apt-${{ matrix.dist }}-${{ matrix.arch }}-base
           execute_install_scripts: true
-      - run: .github/setup-linux.sh
-      - run: .github/build.sh dist
+      - run: .github/setup-linux.sh ${{ matrix.build == 'valgrind' && 'debug' || '' }}
+      - run: .github/build.sh ${{ matrix.build }}
       - name: Upload test logs
         uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: ubuntu-test-logs
+          name: ubuntu-${{ matrix.dist }}-${{ matrix.arch }}-test-logs
           path: |
             tests/*.log
             src/tests/unittests/*.log
       - uses: actions/cache@v4
         id: cache-build
+        if: matrix.build == 'dist'
         with:
           path: ./*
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -62,6 +62,42 @@ jobs:
...
+        with:
+          name: ubuntu-${{ matrix.dist }}-${{ matrix.arch }}-test-logs
+          path: |
+            tests/*.log
+            src/tests/unittests/*.log
+      - uses: actions/cache@v4
+        id: cache-build
+        with:
+          path: ./*
+          key: ${{ runner.os }}-${{ matrix.dist }}-${{ matrix.arch }}-${{ github.sha }}
```
Jakuje, line 93: Do we need this step? In existing workflows it is used to cache the build between the stages (tests after build) and I do not think we have here any state following up

.github/workflows/linux.yml
```diff
@@ -62,6 +62,42 @@ jobs:
...
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: opensc-build-${{ matrix.dist }}-${{ matrix.arch }}
+          path:
+            opensc*.tar.gz
```
Jakuje, line 99: Similarly here. This is done in the main dist job to publish the release tarball (which should be architecture independent anyway). So in theory, it might be used afterward to do some checks they really match, but there will be different timestamps anyway so I would probably drop this step too.

.github/workflows/linux.yml
```diff
@@ -62,6 +62,42 @@ jobs:
           path:
             opensc*.tar.gz
 
+  build-arm64:
+    name: build-${{ matrix.dist }}-${{ matrix.arch }}
+    strategy:
+      matrix:
+        dist: ['22.04', '24.04']
+        arch: ['arm64']
+    runs-on: ubuntu-${{ matrix.dist }}${{ matrix.arch == 'arm64' && '-arm' || '' }}
```
Jakuje, line 71: nit: We have the arch = arm64 here in all the cases so the condition is not needed, unless you would like to use this job to run some more x86_64 jobs and simplify this workflow file.

.github/workflows/linux.yml
metsma: I was thinking maybe make this thing more general. Maybe unify some other builds also eg: x64.
I see lot of duplication already there.
I did not figure out how to depend specifically on one matrix combination. Maybe it is ok to wait until whole matrix ends?

Jakuje: I think so. As long as they will run similar time, it should be ok to wait for all of them.

metsma: Done

.github/workflows/linux.yml
metsma: Runs only on amd64 build

.github/workflows/linux.yml
```diff
@@ -32,56 +32,43 @@ env:
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: awalsh128/cache-apt-pkgs-action@v1.4.3
-        with:
-          packages: ${{ env.BASE_DEPS }}
-          version: apt-latest-base
-          execute_install_scripts: true
-      - run: .github/setup-linux.sh debug
-      - run: .github/build.sh valgrind
```
Jakuje, line 75: The separate valgrind job had the `debug` argument to setup, which installed debuginfo to have reasonable reports of errors. It also runs very slowly, which was a reason why we did not enable it for the main jobs.

metsma: oh I did not notice, ubuntu 22.04 did not have the debug option

metsma: Resolved

Jakuje (approved):

xhanulik (approved):

PR #3382, opened 2025-03-25
More strict code checking

metsma:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -43,6 +43,10 @@
 
 #include "pace.h"
 
+#ifndef MAXDWORD
+#define MAXDWORD UINT32_MAX
+#endif
+
 #ifdef HAVE_PCSCLITE_H
 #if !defined (__MAC_OS_X_VERSION_MIN_REQUIRED) || __MAC_OS_X_VERSION_MIN_REQUIRED < 101000
 #define HAVE_PCSCLITE 1
@@ -245,6 +249,11 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 	LOG_FUNC_CALLED(reader->ctx);
 	card = priv->pcsc_card;
 
+	// Ensure send buffer fits into a DWORD
+	if (sendsize > MAXDWORD)
+		return SC_ERROR_INVALID_ARGUMENTS;
+	if (*recvsize > MAXDWORD)
+		return SC_ERROR_INVALID_ARGUMENTS;
 	if (reader->ctx->flags & SC_CTX_FLAG_TERMINATE)
 		return SC_ERROR_NOT_ALLOWED;
 
@@ -253,8 +262,8 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 	sRecvPci.dwProtocol = opensc_proto_to_pcsc(reader->active_protocol);
 	sRecvPci.cbPciLength = sizeof(sRecvPci);
 
-	dwSendLength = sendsize;
-	dwRecvLength = *recvsize;
+	dwSendLength = (DWORD)sendsize;
+	dwRecvLength = (DWORD)*recvsize;
 
 	if (!control) {
 		rv = priv->gpriv->SCardTransmit(card, &sSendPci, sendbuf, dwSendLength,
--- a/.github/workflows/linux-strict.yml
+++ b/.github/workflows/linux-strict.yml
@@ -7,7 +7,7 @@ on:
       - '**.c'
       - '**.h'
       - '**.sh'
-      - .github/workflows/linux.yml
+      - .github/workflows/linux-strict.yml
       - '**.am'
       - doc/**
       - configure.ac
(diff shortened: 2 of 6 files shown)


src/libopensc/reader-pcsc.c
```diff
@@ -253,8 +253,8 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 	sRecvPci.dwProtocol = opensc_proto_to_pcsc(reader->active_protocol);
 	sRecvPci.cbPciLength = sizeof(sRecvPci);
 
-	dwSendLength = sendsize;
-	dwRecvLength = *recvsize;
+	dwSendLength = (DWORD)sendsize;
+	dwRecvLength = (DWORD)*recvsize;
```
Jakuje, line 257: In theory, converting the size_t (8B on x86_64), while `DWORD` is `uint32_t` (4B) if I see right. Most of the callers call this with much smaller sizes though (i did not check all though). If we know that the buffers are always smaller, I think it should have a comment describing this. If it could be larger than uin32_t max, it should have sanity check that the conversion wont overflow.

Jakuje (approved):

src/libopensc/reader-pcsc.c
metsma: Maybe modify function signature to let callers to know?

Jakuje: That works ok for the recvsize, but I feel like it just moves the problem elsewhere where you need to do the casts to compare with some other types or return the lengths. But still looks reasonable.

src/libopensc/reader-pcsc.c
```diff
@@ -303,7 +304,8 @@ static int pcsc_internal_transmit(sc_reader_t *reader,
 
 static int pcsc_transmit(sc_reader_t *reader, sc_apdu_t *apdu)
 {
-	size_t ssize, rsize, rbuflen = 0;
+	size_t ssize, rbuflen = 0;
+	DWORD rsize;
```
frankmorgner, line 308: rsize is used as input to pcsc_internal_transmit. all other instances allow implicit casts. For pcsc_internal_transmit, however, size_t is required as type

src/libopensc/reader-pcsc.c
```diff
@@ -2146,7 +2147,7 @@ part10_check_pin_min_max(sc_reader_t *reader, struct sc_pin_cmd_data *data)
 {
 	int r;
 	unsigned char buffer[256];
-	size_t length = sizeof buffer;
+	DWORD length = sizeof buffer;
```
frankmorgner, line 2150: Again, when used as input for pcsc_internal_transmit this potentially leads to the wrong type

src/libopensc/reader-pcsc.c
```diff
@@ -2202,7 +2203,8 @@ pcsc_pin_cmd(sc_reader_t *reader, struct sc_pin_cmd_data *data)
 	u8 sbuf[sizeof(PIN_VERIFY_STRUCTURE)>sizeof(PIN_MODIFY_STRUCTURE)?
 		sizeof(PIN_VERIFY_STRUCTURE)+SC_MAX_APDU_BUFFER_SIZE:
 		sizeof(PIN_MODIFY_STRUCTURE)+SC_MAX_APDU_BUFFER_SIZE];
-	size_t rcount = sizeof(rbuf), scount = 0;
+	DWORD rcount = sizeof(rbuf);
```
frankmorgner, line 2206: Again, when used as input for pcsc_internal_transmit this potentially leads to the wrong type

frankmorgner (changes requested):

src/libopensc/reader-pcsc.c
metsma: I probably do not follow. Can you explain more?
```
static int
pcsc_internal_transmit(sc_reader_t *reader,
		const u8 *sendbuf, DWORD sendsize,
		u8 *recvbuf, DWORD *recvsize,
		unsigned long control)
{
...
	DWORD dwSendLength, dwRecvLength;
...
	dwSendLength = sendsize;
	dwRecvLength = *recvsize;
...
		rv = priv->gpriv->SCardTransmit(card, &sSendPci, sendbuf, dwSendLength,
				   &sRecvPci, recvbuf, &dwRecvLength);
...
typedef LONG (PCSC_API *SCardTransmit_t)(SCARDHANDLE hCard, LPCSCARD_IO_REQUEST pioSendPci,
	LPCBYTE pbSendBuffer, DWORD cbSendLength, LPSCARD_IO_REQUEST pioRecvPci,
```

frankmorgner: sorry, I missed the change here which also modifies the signature of pcsc_internal_transmit to accept a DWORD.

However, I believe the implicit approach here historically was that pcsc_internal_transmit offers only types that are generic (i.e. not specific to winscard.dll). I think the better approach would be to convert those types within pcsc_internal_transmit instead of changing the function's signature.

frankmorgner (changes requested):

frankmorgner (approved):
thank you

Jakuje:
Please, rebase on top of recent changes that fixed the Fedora build and for which you have fix in your branch too.

frankmorgner:
thank you

PR #3386, opened 2025-04-02
pkcs11-tool: Don't clutter stdout with OAEP params

D4ryus:
Use stderr for info messages instead.

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2762,7 +2762,7 @@ build_rsa_oaep_params(
 		oaep_params->mgf = CKG_MGF1_SHA3_512;
 		break;
 	default:
-		printf("hash-algorithm %s unknown, defaulting to CKM_SHA256\n", p11_mechanism_to_name(opt_hash_alg));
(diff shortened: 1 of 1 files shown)


Jakuje:
In the same function, there is a bit more outputs, that should go into the stderr. Can you change these too? Other than that, it looks good.

popovec:
Is it even necessary to print these parameters? In my opinion, it is not a good idea to flood stderr with them.

Jakuje:
> Is it even necessary to print these parameters? In my opinion, it is not a good idea to flood stderr with them.

I think they are quite useful when trying to figure out what is actually being used by the application and verifying one set all the required switches right (which might not work on the first try).

Indeed, better would be to print these only if we have some "verbose" mode. There is always way to get them from the pkcs11-spy output, but that is not exactly simple for new users.

Jakuje:
The reasoning for having this in stderr is that people might redirect the output of encryption/decryption to the file and if there is an output like this, it messes things up.

D4ryus:
> In the same function, there is a bit more outputs, that should go into the stderr. Can you change these too? Other than that, it looks good.

Done

Jakuje (approved):

Issue #3110, opened 2024-04-11
Install p11-kit configuration file


H5117:
As OpenSC now provides a single PKCS#11 library, it would be nice to make it available out-of-the-box by installing a p11-kit configuration file ([specification](https://man.archlinux.org/man/pkcs11.conf.5)).

I think it should be just a single line:
```
module: opensc-pkcs11.so
```

Arch Linux
OpenSC 0.25.1-1

Jakuje:
We have the installation in Fedora, but it looks like only in the spec file outside of the build system:
https://github.com/OpenSC/OpenSC/blob/master/packaging/opensc.spec

The p11-kit configuration file itself is trivial so the main issue would be massaging the autotools to install it at right time at right place.

PR #3389, opened 2025-04-04
Add P11-Kit configuration for OpenSC

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/3110

##### Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -956,6 +956,17 @@ if test "${completiondir}" = "detect"; then
 fi
 AC_SUBST([completiondir])
 
+AC_ARG_ENABLE(p11_system_config_modules,
+              AS_HELP_STRING([--enable-p11_system_config_modules=DIR],
+							 [directory containing P11-Kit cnfiguration files (default is determined by libp11-kit)]),
+                             [p11_system_config_modules="${enableval}"],
+                             [p11_system_config_modules=false])
+if test "${p11_system_config_modules}" = false ; then
+    p11_system_config_modules="`$PKG_CONFIG p11-kit-1 --variable=p11_system_config_modules`"
+fi
+AM_CONDITIONAL([ENABLE_P11KIT], [test -n "${p11_system_config_modules}"])
+P11_SYSTEM_CONFIG_MODULES="${p11_system_config_modules}"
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -956,6 +956,17 @@ if test "${completiondir}" = "detect"; then
 fi
 AC_SUBST([completiondir])
 
+AC_ARG_ENABLE(p11_system_config_modules,
+              AS_HELP_STRING([--enable-p11_system_config_modules=DIR],
+							 [directory containing P11-Kit cnfiguration files (default is determined by libp11-kit)]),
```
xhanulik, line 961: typo
```suggestion
							 [directory containing P11-Kit configuration files (default is determined by libp11-kit)]),
```

packaging/opensc.spec
```diff
@@ -211,9 +209,6 @@ rm %{buildroot}%{_mandir}/man1/opensc-notify.1*
 
 %config(noreplace) %{_sysconfdir}/opensc-%{_arch}.conf
 # Co-owned with p11-kit so it is not hard dependency
-%dir %{_datadir}/p11-kit
-%dir %{_datadir}/p11-kit/modules
-%{_datadir}/p11-kit/modules/opensc.module
```
Jakuje, line 216: this lists the files in the RPM so this should stay in if the `make install` will correctly install the file

Jakuje: I think we need to add a p11-kit-devel into the `BuildRequires` and then add the configure switch to make sure this module is correctly configured and installed?

frankmorgner: Yes, a build requirement would be the pkg-config files of p11-kit.

Jakuje: Now it looks like it is installed correctly and the Packit ci fails with:
```
error: Installed (but unpackaged) file(s) found:
   /usr/share/p11-kit/modules/opensc.module
```
So please, get back these three lines. Then I think we are good to go.

frankmorgner: done

Jakuje (approved):
thank you!

PR #3398, opened 2025-04-15
openpgp: Supports key derived PIN format (KDF-DO)

MkfsSion:
KDF-DO feature does not make sense in security aspect but support it would improve compatibility for cards with it enabled for some reason.

##### Checklist
- [ ] macOS token is tested


src/libopensc/card-openpgp.c
```diff
@@ -956,6 +960,83 @@ pgp_get_card_features(sc_card_t *card)
...
+		if (priv->ext_caps & EXT_CAP_KDF_DO) {
+			pgp_blob_t *kdf_blob = NULL;
+			if ((pgp_get_blob(card, priv->mf, DO_KDF, &kdf_blob) == SC_SUCCESS) && (kdf_blob->len >= 3)) {
+				uint8_t *data = kdf_blob->data;
+				priv->pin_kdf_info = NULL;
+				/* KDF_ITERSALTED_S2K */
+				size_t i = 0;
+				pgp_pin_kdf_info *pin_kdf_info = (pgp_pin_kdf_info *)calloc(1, sizeof(pgp_pin_kdf_info));
+				if (!priv) {
+					free(priv);
```
frankmorgner, line 973:
```suggestion
				if (!pin_kdf_info) {
```

src/libopensc/card-openpgp.c
```diff
@@ -956,6 +960,83 @@ pgp_get_card_features(sc_card_t *card)
...
+					pgp_free_blob(kdf_blob);
```
frankmorgner, line 974: I think this does not need to be freed as it will still be "cached" in the priv data

src/libopensc/card-openpgp.c
```diff
@@ -956,6 +960,83 @@ pgp_get_card_features(sc_card_t *card)
...
+				}
+
+				if (valid && (i + 6 <= kdf_blob->len)) {
+					valid = 0;
+					if (data[i] == 0x83 && data[i + 1] == 0x04) {
+						pin_kdf_info->iterations = (uint32_t)bebytes2ulong(&data[i + 2]);
+						valid = 1;
+					}
+					i += 6;
+				}
```
frankmorgner, line 1005: please use the parsing functions from asn1.h to avoid doing all the above by hand.

src/libopensc/card-openpgp.c
```diff
@@ -997,6 +1078,18 @@ pgp_finish(sc_card_t *card)
 		struct pgp_priv_data *priv = DRVDATA(card);
 
 		if (priv != NULL) {
+
```
frankmorgner, line 1081:
```suggestion
```

src/libopensc/card-openpgp.c
```diff
@@ -997,6 +1078,18 @@ pgp_finish(sc_card_t *card)
...
+			if (priv->pin_kdf_info != NULL) {
+				if (priv->pin_kdf_info->userpw_salt != NULL) {
+					free(priv->pin_kdf_info->userpw_salt);
+				}
+				if (priv->pin_kdf_info->adminpw_salt != NULL) {
+					free(priv->pin_kdf_info->adminpw_salt);
+				}
```
frankmorgner, line 1088:
```suggestion
				free(priv->pin_kdf_info->userpw_salt);
				free(priv->pin_kdf_info->adminpw_salt);
```

src/libopensc/card-openpgp.c
```diff
@@ -2082,6 +2175,152 @@ pgp_put_data(sc_card_t *card, unsigned int tag, const u8 *buf, size_t buf_len)
...
+		EVP_MD_CTX_free(hctx);
+		sc_evp_md_free(md);
+		return SC_ERROR_NOT_ENOUGH_MEMORY;
+	}
+	size_t count = iterations;
+	if (count < pinlen + saltlen) {
+		count = pinlen + saltlen;
+	}
+	while (count >= pinlen + saltlen) {
+		EVP_DigestUpdate(hctx, salt, saltlen);
```
frankmorgner, line 2219: missign error checking

src/libopensc/card-openpgp.c
```diff
@@ -2082,6 +2175,152 @@ pgp_put_data(sc_card_t *card, unsigned int tag, const u8 *buf, size_t buf_len)
...
+	if (count <= saltlen) {
+		EVP_DigestUpdate(hctx, salt, count);
+	} else {
+		EVP_DigestUpdate(hctx, salt, saltlen);
+		count -= saltlen;
+		EVP_DigestUpdate(hctx, pin, count);
+	}
+
+	unsigned int outsize = keysize;
+	EVP_DigestFinal(hctx, outkey, &outsize);
```
frankmorgner, line 2232: missign error checking

src/libopensc/card-openpgp.c
```diff
@@ -2082,6 +2175,152 @@ pgp_put_data(sc_card_t *card, unsigned int tag, const u8 *buf, size_t buf_len)
...
+	const uint8_t *pin1 = data->pin1.data;
+	size_t pin1len = data->pin1.len;
+	const uint8_t *pin2 = data->pin2.data;
+	size_t pin2len = data->pin2.len;
+	uint8_t *salt = NULL;
+	size_t saltlen = 0;
+	switch (data->pin_reference) {
+	case 0x81:
+	case 0x82:
+		salt = info->userpw_salt;
```
frankmorgner, line 2270: need to check if `info != NULL`

src/libopensc/card-openpgp.h
```diff
@@ -66,17 +66,18 @@ typedef enum _pgp_access {		/* access flags for the respective DO/file */
...
-	EXT_CAP_C4_CHANGEABLE		= 0x0010,
-	EXT_CAP_KEY_IMPORT		= 0x0020,
-	EXT_CAP_GET_CHALLENGE		= 0x0040,
-	EXT_CAP_SM			= 0x0080,
-	EXT_CAP_LCS			= 0x0100,
-	EXT_CAP_CHAINING		= 0x1000,
-	EXT_CAP_APDU_EXT		= 0x2000,
-	EXT_CAP_MSE			= 0x4000
+typedef enum _pgp_ext_caps { /* extended capabilities/features: bit flags */
+	EXT_CAP_KDF_DO = 0x0001,
```
frankmorgner, line 70: not sure why you're chaning the alignment here...

frankmorgner (changes requested):

src/libopensc/card-openpgp.h
MkfsSion: because clang-format-linter complains about the format.

src/libopensc/card-openpgp.c
MkfsSion: It's already checked by the caller.

MkfsSion:
I am confused that what the return value of pgp_get_card_features actually means. 
```
	/* get card_features from ATR & DOs */
	if (pgp_get_card_features(card)) {
	}
```
For the caller, i guess it means the number of algorithm decoded.
but
```
			/* OpenPGP card spec 1.1 & 2.x section 4.3.3.6 / v3.x section 4.4.3.7 */
			if ((pgp_get_blob(card, blob73, i, &blob) >= 0) &&
			    (pgp_parse_algo_attr_blob(card, blob, &key_info) >= 0)) {
				if (!_pgp_add_algo(card, &key_info, i))
					LOG_TEST_RET(card->ctx, SC_ERROR_INTERNAL, "Cannot add algorithm");
			}
```
It return a error code, and the caller think it's  the number of algorithm decoded.
Does this means pgp_get_card_features can not fail?

src/libopensc/card-openpgp.h
frankmorgner: i see

frankmorgner:
about pgp_get_card_features(), I think it should be changed to something like this:
```diff
diff --git a/src/libopensc/card-openpgp.c b/src/libopensc/card-openpgp.c
index 3abc5b41f..11cc118c6 100644
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -975,9 +975,8 @@ pgp_get_card_features(sc_card_t *card)
 
                        /* OpenPGP card spec 1.1 & 2.x section 4.3.3.6 / v3.x section 4.4.3.7 */
                        if ((pgp_get_blob(card, blob73, i, &blob) >= 0) &&
-                           (pgp_parse_algo_attr_blob(card, blob, &key_info) >= 0)) {
-                               if (!_pgp_add_algo(card, &key_info, i))
-                                       LOG_TEST_RET(card->ctx, SC_ERROR_INTERNAL, "Cannot add algorithm");
+                                       (pgp_parse_algo_attr_blob(card, blob, &key_info) >= 0)) {
+                               handled_algos += _pgp_add_algo(card, &key_info, i);
                        }
                }
```

src/libopensc/card-openpgp.c
frankmorgner: That's incorrect. if the initial calloc fails, then this is not fatal to the initialization of the openpgp driver and the priv's member will be NULL. please be generous with sanity checks - if they are not needed then they are either optimized out by the compiler or they are extremely cheap given modern CPUs.

MkfsSion: `pgp_kdf_do_pin_cmd()` will only be called if `priv->pin_kdf_info` is not NULL. But I'll add a check anyway.

src/libopensc/card-openpgp.c
```diff
@@ -2082,6 +2194,166 @@ pgp_put_data(sc_card_t *card, unsigned int tag, const u8 *buf, size_t buf_len)
 	LOG_FUNC_RETURN(card->ctx, (int)buf_len);
 }
 
+#ifdef ENABLE_OPENSSL
+/* KDF_ITERSALTED_S2K algorithm, a simplified version of openpgp_s2k from libgcrypt[1]
+ * that supports one-pass only and outputs the same key length as the hash length of
+ * the specified hash algorithm.
+ * [1]: https://github.com/gpg/libgcrypt/blob/libgcrypt-1.11.0/cipher/kdf.c#L32-L109
```
frankmorgner, line 2206: you need to reference a specific commit if you want to get a url that is permanently accessible

frankmorgner (approved):

src/libopensc/card-openpgp.c
Jakuje: This references a tag, `libgcrypt-1.11.0` , which should be stable too.

src/libopensc/card-openpgp.c
```diff
@@ -2082,6 +2194,166 @@ pgp_put_data(sc_card_t *card, unsigned int tag, const u8 *buf, size_t buf_len)
...
+		r = SC_SUCCESS;
+	}
+out3:
+	if (r != SC_SUCCESS) {
+		sc_mem_secure_clear_free(outkey, keysize);
+	}
+out2:
+	EVP_MD_CTX_free(hctx);
+out1:
+	sc_evp_md_free(md);
```
Jakuje: I would prefer to have just one label for a goto. More jumping makes the code harder to read.

If you initialize the pointers to NULL at the beginning, its safe to call on them the `*_free()`. regardless you passed some point in the code or not.

This also involves declaring the variables at the beginning of the block rather than through the function, which might sounds old-fashined, but very helpful to get the code that is easy to follow.

I would suggest you to do it also in the other functions you did as mixing up declarations and code does not make the code easier to read.

Jakuje (commented):
please, also rebase on top of current master. It should fix the CI.

Jakuje (approved):

frankmorgner:
thank you

PR #3401, opened 2025-04-18
piv_tool: Refactor the keygen cleanup

Jakuje:
to avoid possible memory leaks reported by oss-fuzz

https://issues.oss-fuzz.com/u/1/issues/397209407

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/piv-tool.c
+++ b/src/tools/piv-tool.c
@@ -109,8 +109,7 @@ static const char *option_help[] = {
 
 static sc_context_t *ctx = NULL;
 static sc_card_t *card = NULL;
-static BIO * bp = NULL;
-static EVP_PKEY * evpkey = NULL;
+static BIO *bp = NULL;
 
 static int load_object(const char * object_id, const char * object_file)
 {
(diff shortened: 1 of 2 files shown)


src/tools/piv-tool.c
```diff
@@ -284,14 +283,16 @@ static int admin_mode(const char* admin_info)
 /* generate a new key pair, and save public key in newkey */
 static int gen_key(const char * key_info)
 {
-	int r;
+	int r = -1;
```
frankmorgner, line 286: This is a little odd: `main()` directly takes the error code from `gen_key()` and outputs it if it is non-zero. From this, I would expect `gen_key()` to return a positive number. Other internal functions that are called from `main()` are restricting themselves to small positive values... I think the refactoring done here would give a good opportunity to fix the return value from `gen_key()`. Do you agree, @dengert ?

frankmorgner (changes requested):
CI suggests the following formatting change:
```diff
 diff --git a/src/tools/piv-tool.c b/src/tools/piv-tool.c
index 5318e97..2b3b312 100644
--- a/src/tools/piv-tool.c
+++ b/src/tools/piv-tool.c
@@ -109,7 +109,7 @@ static const char *option_help[] = {
 
 static sc_context_t *ctx = NULL;
 static sc_card_t *card = NULL;
-static BIO * bp = NULL;
+static BIO *bp = NULL;
 
 static int load_object(const char * object_id, const char * object_file)
 {
@@ -383,7 +383,7 @@ static int gen_key(const char * key_info)
 		newkey = RSA_new();
 		if (!newkey) {
 			sc_log_openssl(ctx);
-			fprintf(stderr, "gen_key RSA_new failed %d\n",r);
+			fprintf(stderr, "gen_key RSA_new failed %d\n", r);
 			r = -1;
 			goto out;
 		}
```

src/tools/piv-tool.c
Jakuje: Right. It was mess already and I did not want to mess with it too much. But looks like I already did so lets fix this a bit.

Jakuje:
> CI suggests the following formatting change:

Yes, I have this fixed locally, but did not push as I wanted to wait for reviews and comments to save one CI run. Will include these changes.

frankmorgner (approved):
thank you

PR #3402, opened 2025-04-20
Coverity fixes

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/compression.c
+++ b/src/libopensc/compression.c
@@ -91,9 +91,12 @@ static int sc_compress_gzip(u8* out, size_t* outLen, const u8* in, size_t inLen)
 	gz.avail_out = (unsigned)*outLen;
 
 	err = deflateInit2(&gz, Z_BEST_COMPRESSION, Z_DEFLATED, window_size, 9, Z_DEFAULT_STRATEGY);
-	if(err != Z_OK) return zerr_to_opensc(err);
+	if (err != Z_OK) {
+		deflateEnd(&gz);
+		return zerr_to_opensc(err);
+	}
 	err = deflate(&gz, Z_FINISH);
-	if(err != Z_STREAM_END) {
+	if (err != Z_STREAM_END) {
 		deflateEnd(&gz);
 		return zerr_to_opensc(err);
 	}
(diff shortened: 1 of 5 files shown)


src/libopensc/compression.c
```diff
@@ -90,14 +90,9 @@ static int sc_compress_gzip(u8* out, size_t* outLen, const u8* in, size_t inLen)
...
 	gz.avail_out = (unsigned)*outLen;
 
-	err = deflateInit2(&gz, Z_BEST_COMPRESSION, Z_DEFLATED, window_size, 9, Z_DEFAULT_STRATEGY);
-	if(err != Z_OK) return zerr_to_opensc(err);
-	err = deflate(&gz, Z_FINISH);
-	if(err != Z_STREAM_END) {
-		deflateEnd(&gz);
-		return zerr_to_opensc(err);
+	if ((err = deflateInit2(&gz, Z_BEST_COMPRESSION, Z_DEFLATED, window_size, 9, Z_DEFAULT_STRATEGY)) == Z_OK && (err = deflate(&gz, Z_FINISH)) == Z_STREAM_END) {
+		*outLen = gz.total_out;
```
Jakuje, line 94: I do not think this makes it easier to read. At least, I would break it into more lines to avoid the very long line, but better keep it as it was and just call the `deflateEnd()` in case the `deflateInit2` fails:
```
	err = deflateInit2(&gz, Z_BEST_COMPRESSION, Z_DEFLATED, window_size, 9, Z_DEFAULT_STRATEGY);
	if (err != Z_OK) {
		deflateEnd(&gz);
		return zerr_to_opensc(err);
	}
	err = deflate(&gz, Z_FINISH);
	if (err != Z_STREAM_END) {
		deflateEnd(&gz);
		return zerr_to_opensc(err);
	}
	*outLen = gz.total_out;
```
or rewrite it a bit with the `goto err`, but for freeing one resource, this sounds like an overkill.

frankmorgner: done

Jakuje (approved):

metsma (approved):

xhanulik (approved):
looks good

my12345677 (changes requested):
Hi everyone

PR #3403, opened 2025-04-20
Removed internal caching for current EF/DF

frankmorgner:
- [ ] macOS token is tested


frankmorgner:
This fixes concurrent access to the card from different threads or processes at the cost of potentially causing more card interactions than necessary.

Since the card driver for IAS/ECC uses the cached path additionally for managing the ACLs in SM context, this state tracking was moved to a card driver private scope.

Jakuje (approved):
I don't have anything against simplifying the code like this! Thank you!

PR #3405, opened 2025-04-24
sc-hsm-tool: Use dynamic allocations where possible

hendrikdonner:
Main motivation is that i encountered a PRKD blob of size 262, larger than the fixed size buffer of 256 bytes. That meant that `--wrap-key` created a structure with the wrapped private key and the certificate only, skipping the PRKD blob. But using `--unwrap-key` ends up trying to parse the certificate as the PRKD blob, so `sc-hsm-tool` erroneously reports that it found a key blob, PRKD and no certificate, ultimately failing to import anything meaningful.

Get rid of most fixed size buffers.


Jakuje:
Its kind of two-way. If there is dynamic allocation for everything and card returns garbage, you might end up allocating too much, running out of memory and so on so I would prefer to keep some limits in places. But generally, it looks reasonable.

@CardContact any thoughts?

src/tools/sc-hsm-tool.c
```diff
@@ -1467,20 +1482,28 @@ static int wrap_key(sc_context_t *ctx, sc_card_t *card, int keyid, const char *o
...
-		fclose(out);
-		return -1;
+		r = -1;
 	}
 
-	free(key);
 	fclose(out);
-	return 0;
+err:
+	if (key)
```
frankmorgner, line 1496: this check is not required, free does recognize NULL

src/tools/sc-hsm-tool.c
```diff
@@ -1329,16 +1326,17 @@ static int wrap_key(sc_context_t *ctx, sc_card_t *card, int keyid, const char *o
...
 	u8 wrapped_key_buff[MAX_KEY];
-	u8 keyblob[MAX_WRAPPED_KEY];
-	u8 *key;
-	u8 *ptr;
+	u8 *keyblob = NULL;
+	u8 *key = NULL;
+	u8 *ptr = NULL;
 	char *lpin = NULL;
-	size_t key_len;
+	size_t key_len, keyblob_len = MAX_KEY;
```
frankmorgner, line 1339: If you initialize all the new pointers to `NULL`, I suggest you also initialize their length values to `0`

src/tools/sc-hsm-tool.c
```diff
@@ -1551,17 +1575,23 @@ static int unwrap_key(sc_card_t *card, int keyid, const char *inf, const char *p
 	}
 
 	in = fopen(inf, "rb");
-
 	if (in == NULL) {
 		perror(inf);
 		return -1;
 	}
 
-	sz = fread(keyblob, 1, sizeof(keyblob), in);
```
frankmorgner, line 1560: you could use `file_read_to_eof()`

src/tools/sc-hsm-tool.c
```diff
@@ -1809,16 +1867,23 @@ static int export_key(sc_card_t *card, int keyid, const char *outf)
...
 		free(data);
+	if (outfp)
+		fclose(outfp);
+	if (dica)
+		free(dica);
+	if (dev_aut_cert)
+		free(dev_aut_cert);
+	if (ef_cert)
+		free(ef_cert);
+	if (file)
```
frankmorgner, line 1885: sc_file_free can also handle NULL input

frankmorgner (changes requested):

src/tools/sc-hsm-tool.c
hendrikdonner: True, i somewhat mindlessly followed the existing pattern.

Jakuje (approved):
lgtm

xhanulik (approved):

frankmorgner:
I pushed a commit to fix the formatting, hopefully all signs will become green now

hendrikdonner:
> I pushed a commit to fix the formatting, hopefully all signs will become green now

Right, i tried to fix the last 2 lines and sqashed into one commit.

frankmorgner (approved):

PR #3406, opened 2025-04-24
edo2021: initial support for Polish eID v2.0 cards

piotro888:
Hello,
This PR adds initial support for Polish eID v2.0 cards that are issued from 2021

Unfortunately there is still no published documentation for any version...

Verified with `pkcs15-tool` and `pkcs11-tool` - login and signing works with all `ECDSA*` mechanisms for keys 0 (Authorization) and 2 (Authentication).
Other feature support may be not ideal. 

Changes are based on card-edo v1 (PACE CAN authentication, unlock, and curves are the same). 
v2 cards use different chip and protocol (as mentioned in #1831), but more features seem to work with standard handles.
Security env APDU turned out to be similar as in `card-eoi` - it is dependent on hash length, so I used similar trick as there to call secure env from signing function, but a bit cleaner.

##### Checklist
- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-edo2021.c
+++ b/src/libopensc/card-edo2021.c
@@ -1,0 +1,206 @@
+/*
+ * Driver for Polish eID v2.0 cards issued from 2021.
+ *
+ * Copyright (C) 2025 Piotr Wegrzyn <piotro@piotro.eu>
+ *
+ * This file is part of OpenSC.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#if defined(ENABLE_SM) && defined(ENABLE_OPENPACE)
+
+#include "libopensc/asn1.h"
+#include "libopensc/card-edo.h"
+#include "libopensc/card-edo2021.h"
+#include "libopensc/internal.h"
+#include "libopensc/opensc.h"
+#include "libopensc/pace.h"
+#include "libopensc/sm.h"
+#include "sm/sm-eac.h"
+#include <openssl/sha.h>
+#include <stdlib.h>
+#include <string.h>
+
+static struct sc_card_operations edo2021_ops;
+
+static struct sc_card_driver edo2021_drv = {
+		"Polish eID card issued from 2021 (e-dowód, eDO)",
+		"edo2021",
+		&edo2021_ops,
+		NULL, 0, NULL};
+
+static const struct sc_atr_table edo2021_atrs[] = {
+		{"3b:89:80:01:02:4d:4b:4d:57:4b:53:4b:54:11", NULL, NULL, SC_CARD_TYPE_EDO2021, 0, NULL},
+		{NULL,					NULL, NULL, 0,		      0, NULL}
+};
+
+static struct {
+	int len;
+	struct sc_object_id oid;
+} edo2021_curves[] = {
+		// secp384r1
+		{384, {{1, 3, 132, 0, 34, -1}}}
+};
+
+static int
+edo2021_match_card(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	if (_sc_match_atr(card, edo2021_atrs, &card->type) >= 0) {
+		sc_log(card->ctx, "ATR recognized as Polish eID card (edo2021 driver).");
+		LOG_FUNC_RETURN(card->ctx, 1);
+	}
+	LOG_FUNC_RETURN(card->ctx, 0);
+}
+
+static int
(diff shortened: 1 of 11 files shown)


src/libopensc/pkcs15.c
```diff
@@ -150,8 +150,8 @@ int sc_pkcs15_parse_tokeninfo(sc_context_t *ctx,
 	u8 last_update[32], profile_indication[SC_PKCS15_MAX_LABEL_SIZE];
 	size_t lupdate_len = sizeof(last_update) - 1, pi_len = sizeof(profile_indication) - 1;
 	size_t flags_len   = sizeof(ti->flags);
-	u8 preferred_language[3];
-	size_t lang_length = sizeof(preferred_language);
+	u8 preferred_language[SC_PKCS15_MAX_LABEL_SIZE];
```
frankmorgner, line 153: What kind of problem are you trying to fix here?

By definition, I think, this should be a two letter ISO code of the localization. We are using this, for example, in `src/ui/strings.c` to show text in the correct language.

src/libopensc/card-edo2021.c
```diff
@@ -0,0 +1,275 @@
...
+	return SC_SUCCESS;
+}
+
+static int
+edo_init(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	struct edo2021_privdata *privdata = (struct edo2021_privdata *)card->drv_data;
+	privdata = sc_mem_secure_alloc(sizeof(struct edo2021_privdata));
```
frankmorgner, line 214: I don't think this actually contains sensitive data so I suggest you simply use `calloc()`.

src/libopensc/card-edo2021.c
```diff
@@ -0,0 +1,275 @@
...
+	LOG_TEST_RET(card->ctx, edo_set_security_env(card, &privdata->sec_env, data_len), "set_security_env failed");
+
+	/* Call default iso7816 handler, we need this card op only for security_env. */
+	LOG_FUNC_RETURN(card->ctx, sc_get_iso7816_driver()->ops->compute_signature(card, data, data_len, out, outlen));
+}
+
+static int
+edo_logout(sc_card_t *card)
+{
+	sc_sm_stop(card);
```
frankmorgner, line 204: edo driver uses edo_unlock here additionally presumably because private key operations are only allowed in an active SM channel. That SM channel, however, would not be available here anymore. As simple solution, I suggest you add edo_unlock(card) here as well.

As more advanced solution (for both edo and edo2021) you should, check *before* sc_login whether a SM channel is active or not. That can happen, for example also when some concurrent App is accessing the card without SM. Currently, both drivers cannot recover from this problem. With this more advanced check, you would verify the CAN when needed, which would allow concurrent access to the card.

frankmorgner (changes requested):
please use `edo2021` as prefix for your driver callbacks

please use shared code between edo2021 and edo to simultanously use edo_get_can, edo_unlock and likely edo_logout

have you also tested encryption/key derivation?

src/libopensc/pkcs15.c
piotro888: This card returns `pl-PL` as a `preferredLanguage`, and it breaks pkcs15 ASN.1 parser with `buffer too small`, causing failing to read the card completely, as unsupported card.

I've read the RFCs now, and it turns out this response is compliant with PCKS15.
It should return RFC1766 language tag which is:
```
2.  The Language tag

   The language tag is composed of 1 or more parts: A primary language
   tag and a (possibly empty) series of subtags.

   The syntax of this tag in RFC-822 EBNF is:

    Language-Tag = Primary-tag *( "-" Subtag )
    Primary-tag = 1*8ALPHA
    Subtag = 1*8ALPHA
```
So, it can be up to 17 characters long.

Should I set the buffer to 17 characters? Should it return the full tag or only primary-tag? Or maybe pkcs15 behaviour should be changed to not fail on optional tags?
What do you think?

src/libopensc/card-edo2021.c
piotro888: > As more advanced solution (for both edo and edo2021) you should, check before sc_login 

This card without CAN fails as early as reading `EF(ODF)` file in `pkcs15_bind_internal`. What card operation would be the correct place to check for SM channel? I'm not sure if the card is usable for any case without SM and is it worth adding.

frankmorgner: card_reader_lock_obtained would be the best place for such a check and for establishing SM if needed, I think. You may want to try this out in different applications, because the additional interactions with the card may cause quite some delay...

piotro888: done, seems to work fine

src/libopensc/pkcs15.c
piotro888: switched to increased buffer size

piotro888:
> please use edo2021 as prefix for your driver callbacks
> please use shared code between edo2021 and edo to simultanously use edo_get_can, edo_unlock and likely edo_logout

fixed

> have you also tested encryption/key derivation?

They don't work. I'm only able to observe how signature operation works from "official" software, which is the main use anyway.

piotro888:
Unfortunately I found the problem that signature verification fails both with `pkcs11-tool` and `openssl dgst`.
Public key read from card's X.509 certificate matches known correct one.
However, when I tested it with some python script that used `opensc-libpkcs11` to sign a document with XAdES using the same smartcard key, it is correctly verified by official gov software.
I have no idea where to look for problems, do you have some suggestions?

metsma:
pkcs11-tool has options:
-f, --signature-format <arg>  Format for ECDSA signature <arg>: 'rs' (default), 'sequence', 'openssl'

src/libopensc/pkcs15.c
frankmorgner: Using SC_PKCS15_MAX_LABEL_SIZE is good, I think. I have added an issue that the ui text needs to ignore the subtags

piotro888:
Status:
Card returns correct length of signature data for RS pair, but the values turn out to be invalid.
I will look more into original signing communication when I have a bit more free time.

frankmorgner:

PR #3411, opened 2025-05-05
pkcs11-spy: Fall back to PKCS#11 2.* interfaces rather than returning…

Jakuje:
… NULL function list pointer


frankmorgner (approved):

Jakuje:
Thank you for the review! Merging.

PR #3415, opened 2025-05-08
idprime: IDPrime 940 supports 3072 RSA keys

vssldmtrv:
Such is the key on the Gemalto IDPrime 940B card I got from the bulgarian eID provider B-Trust [1].
Tested and confirmed to work.

Its ATR is as follows:
`3b:7f:96:00:00:80:31:80:65:b0:85:05:00:39:12:0f:fe:82:90:00`

I.e. identical to the one already defined for IDPrime 940.

[1] https://www.b-trust.bg/en/electronic-signatures/cards-and-readers

##### Checklist

- [ ] macOS token is tested


Jakuje (approved):
thank you! Looks good!

metsma (approved):

PR #3420, opened 2025-05-14
Coverity fixes

xhanulik:
Recently found issues by Coverity

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-esteid2025.c
+++ b/src/libopensc/pkcs15-esteid2025.c
@@ -50,13 +50,16 @@ sc_pkcs15emu_esteid2025_init(sc_pkcs15_card_t *p15card)
 
 	/* Read document number to be used as serial */
 	sc_format_path("DFDD5007", &tmppath);
-	LOG_TEST_RET(card->ctx, sc_select_file(card, &tmppath, NULL), "SELECT docnr");
+	r = sc_select_file(card, &tmppath, NULL);
+	if (r < 0) {
+		free(buff);
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
+	}
 	r = sc_read_binary(card, 0, buff, buflen, 0);
 	if (r < 0) {
 		free(buff);
 		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
 	}
-	LOG_TEST_RET(card->ctx, r, "read document number failed");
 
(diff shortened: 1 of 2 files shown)


src/libopensc/pkcs15-esteid2025.c
```diff
@@ -50,7 +50,11 @@ sc_pkcs15emu_esteid2025_init(sc_pkcs15_card_t *p15card)
 
 	/* Read document number to be used as serial */
 	sc_format_path("DFDD5007", &tmppath);
-	LOG_TEST_RET(card->ctx, sc_select_file(card, &tmppath, NULL), "SELECT docnr");
+	r = sc_select_file(card, &tmppath, NULL);
+	if (r < 0) {
+		free(buff);
+	}
+	LOG_TEST_RET(card->ctx, r, "SELECT docnr");
```
Jakuje, line 57: I would change it like this to avoid two different conditions checking the same failure (similarly as in the following block):
```suggestion
		free(buff);
		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
	}
```

Jakuje (approved):
otherwise lgtm

metsma (approved):

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -50,7 +50,11 @@ sc_pkcs15emu_esteid2025_init(sc_pkcs15_card_t *p15card)
 
 	/* Read document number to be used as serial */
 	sc_format_path("DFDD5007", &tmppath);
-	LOG_TEST_RET(card->ctx, sc_select_file(card, &tmppath, NULL), "SELECT docnr");
+	r = sc_select_file(card, &tmppath, NULL);
+	if (r < 0) {
+		free(buff);
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
+	}
```
frankmorgner, line 57: The test in line 59 cannot be true
```
LOG_TEST_RET(card->ctx, r, "read document number failed");
```
could you please remove that line even though it is out of scope of the memory leak?

frankmorgner (changes requested):

src/libopensc/pkcs15-esteid2025.c
Jakuje: Goo catch! So I think we are good here. Than you! Merging then

Issue #3404, opened 2025-04-24
pkcs11-register does not work on linux


metsma:
Ubuntu 25.05 includes OpenSC version 0.26.1. Our testers found that pkcs11-register no longer enables opensc-pkcs11 for Chrome or Firefox by default.

This behavior appears to stem from the following change:
https://github.com/OpenSC/OpenSC/commit/71959c7105537ae8a84d747e5605fcef6cad3f17#diff-b954244a44e6aa1482b45288afaecdad21eb3c5b4353a8b5c102480361fa47feR84

Could you clarify the rationale behind this change? Chrome/Firefox is set to skip_chrome_flag 1 on linux.
And command options only offer skip_firefox/skip_chrome that also turns this to 1?

frankmorgner:
If you look at `src/tools/pkcs11-register.ggo.in`, you can see that the default actually depends on the platform. To restore the original default, gengetopt needs to be available during compile time, which will regenerate the source files (with the platform's default). You would need to ask the Ubuntu package manager for that. We tried to compensate this change to some respect by adding `--skip-chrome=off --skip-firefox=off` to the xdg autostart item of pkcs11-register (`src/tools/pkcs11-register.desktop.in`). Those options could also be used when being called via the command line.

The reason behind this change is that we prefer distributing the generated source files for Windows, where gengetopt requires more effort to be installed.

metsma:
kcs11-register --skip-chrome=off --skip-firefox=off
pkcs11-register: option '--skip-chrome' doesn't allow an argument

frankmorgner:
The commandline options are *toggling* the default action, so just using `pkcs11-register --skip-chrome --skip-firefox` would enable both settings in your case.

I understand that the naming of those options don't match their intended use, in particular when we have different defaults for different OSs. I can think of the following two solutions:
1. use one default for every OS, this would allow renaming the option to exactly match the intended use (--no-skip-chrome and --no-skip-firefox)
2. refactor the code to make it an option with two possible values "on" and "off" (see https://www.gnu.org/software/gengetopt/gengetopt.html#Options-with-enumerated-values). This would allow setting the value explicitly with the support of OS specific defaults.

metsma:
https://github.com/OpenSC/OpenSC/blob/master/src/tools/pkcs11-register.desktop.in#L4
Then this should be fixed?

frankmorgner:
[...]

PR #3426, opened 2025-05-22
pkcs11-register: allow explicitly specifying an app to configure

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/3404

##### Checklist

- [ ] macOS token is tested


Jakuje:
Thank you!

PR #3427, opened 2025-05-22
fixed Stack-buffer-overflow WRITE in GET RESPONSE

frankmorgner:
The do-while loop in apdu.c requires the output data to be set in any case, otherwise non existent data may be copied to the output data.

fixes https://issues.oss-fuzz.com/issues/416351800 fixes https://issues.oss-fuzz.com/issues/416295951

##### Checklist

- [ ] macOS token is tested


Jakuje (approved):

frankmorgner:
good to know that this would be possible 😉

PR #3428, opened 2025-05-22
pkcs11-tool: allocate for C_Sign/C_SignFinal

antimeme:
Addresses issue 3425.  Quantum resistant algorithms like LMS and ML-DSA have large signature sizes.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2516,9 +2516,11 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 			sz = read(fd, in_buffer, sizeof(in_buffer));
 		} while (sz > 0);
 
-		sig_len = sizeof(sig_buffer);
+		/* Signature buffer may have been allocated above */
+		free(sig_buffer);
+		sig_buffer = NULL;
 		rv = p11->C_SignFinal(session, sig_buffer, &sig_len);
-		if ((rv == CKR_OK) && !sig_buffer) {
+		if (rv == CKR_OK) {
 			sig_buffer = malloc(sig_len);
 			if (!sig_buffer)
 				util_fatal("malloc() failure\n");
@@ -2552,10 +2554,9 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 				util_fatal("Failed to convert signature to ASN.1 sequence format");
 			}
 
-			memcpy(sig_buffer, seq, seqlen);
+			free(sig_buffer);
+			sig_buffer = seq;
 			sig_len = seqlen;
-
-			free(seq);
 		}
 	}
 	sz = write(fd, sig_buffer, sig_len);
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -2520,12 +2520,12 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 		free(sig_buffer);
 		sig_buffer = NULL;
 		rv = p11->C_SignFinal(session, sig_buffer, &sig_len);
-		if (rv == CKR_OK) {
-			sig_buffer = malloc(sig_len);
-			if (!sig_buffer)
-				util_fatal("malloc() failure\n");
-			rv = p11->C_SignFinal(session, sig_buffer, &sig_len);
-		}
+		if (rv != CKR_OK)
+			p11_fatal("C_SignFinal", rv);
+		sig_buffer = malloc(sig_len);
+		if (!sig_buffer)
+			util_fatal("malloc() failure\n");
+		rv = p11->C_SignFinal(session, sig_buffer, &sig_len);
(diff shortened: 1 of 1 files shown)


frankmorgner:
I assume that key derivation with e.g. ML-KEM is not yet ready for practical usage, right?

frankmorgner (changes requested):
The same modifications need to be done in `verify_signature()`

antimeme:
> I assume that key derivation with e.g. ML-KEM is not yet ready for practical usage, right?

ML-KEM provides a key encapsulation primitive, which isn't exactly encryption.  I'm not sure whether PKCS#11 currently supports this (by current I mean version 3.1).  It may be necessary to introduce C_Encapsulate/C_Decapsulate.  I don't see these in the 3.1 header, but it's possible they're just called something else.

Jakuje:
> > I assume that key derivation with e.g. ML-KEM is not yet ready for practical usage, right?
> 
> ML-KEM provides a key encapsulation primitive, which isn't exactly encryption. I'm not sure whether PKCS#11 currently supports this (by current I mean version 3.1). It may be necessary to introduce C_Encapsulate/C_Decapsulate. I don't see these in the 3.1 header, but it's possible they're just called something else.

The PKCS#11 3.2 brings new functions C_Encapsulate/C_Decapsulate, which is something that will be needed to add to the pkcs11-tool anyway, but likely once the headers will be officially out.

frankmorgner:
> The PKCS#11 3.2 brings new functions C_Encapsulate/C_Decapsulate, which is something that will be needed to add to the pkcs11-tool anyway, but likely once the headers will be officially out.

Interesting... because dencapsulation/decapsulation is functionally equivalent to key derivation, e.g. with eph. DH.

frankmorgner (approved):
LGTM, thanks

frankmorgner (approved):

src/tools/pkcs11-tool.c
```diff
@@ -2491,8 +2492,13 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
 			login(session,CKU_CONTEXT_SPECIFIC);
 
-		sig_len = sizeof(sig_buffer);
 		rv = p11->C_Sign(session, in_buffer, sz, sig_buffer, &sig_len);
+		if (rv == CKR_OK) {
+			sig_buffer = malloc(sig_len);
+			if (!sig_buffer)
+				util_fatal("malloc() failure\n");
+			rv = p11->C_Sign(session, in_buffer, sz, sig_buffer, &sig_len);
+		}
```
Jakuje, line 2501: I think this check should be written the other way round to fail in case this function call fails here (and to avoid needless nesting):
```suggestion
		if (rv != CKR_OK) {
			util_fatal("Failed to get signature length\n");
		}
		sig_buffer = malloc(sig_len);
		if (!sig_buffer) {
			util_fatal("malloc() failure\n");
		}
		rv = p11->C_Sign(session, in_buffer, sz, sig_buffer, &sig_len);
```

Jakuje: same for the following block, which is similar.

Jakuje (changes requested):

src/tools/pkcs11-tool.c
antimeme: This patch changes the intent of the original code (before my proposed changes).  When the input is below a certain size threshold C_Sign is used.  If it fails a C_SignUpdate loop followed by C_SignFinal are attempted.  So the indentation is not actually unnecessary in the first section, unless you want to change the behavior.  My intention is only to make LMS signatures possible, not to alter any semantics.  The second section can be changed as you suggest though, so I'll do that.

frankmorgner (approved):

src/tools/pkcs11-tool.c
```diff
@@ -2510,7 +2516,15 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
 		} while (sz > 0);
 
-		sig_len = sizeof(sig_buffer);
+		/* Signature buffer may have been allocated above */
+		free(sig_buffer);
+		sig_buffer = NULL;
+		rv = p11->C_SignFinal(session, sig_buffer, &sig_len);
+		if (rv != CKR_OK)
+			p11_fatal("C_SignFinal", rv);
+		sig_buffer = malloc(sig_len);
```
Jakuje, line 2525: nit: Do we need to free and allocate the `sig_buffer` here? Given that this is the same operation with the same key, I would expect the signature size would be same for both `C_Sign` and `C_SignFinal` operation. 
But this is really just a corner case. Making sure the length of the old buffer is right would complicate the code probably eve more.

src/tools/pkcs11-tool.c
Jakuje: thank you for the clarification. I missed that part of the logic.

Jakuje (approved):

Jakuje:
Thank you for your contribution! Merged

PR #3430, opened 2025-05-24
EstEID 2018 v2 fixes with some readers

metsma:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/cards.h
+++ b/src/libopensc/cards.h
@@ -231,6 +231,7 @@ enum {
 
 	/* EstEID cards */
 	SC_CARD_TYPE_ESTEID_2018 = 35000,
+	SC_CARD_TYPE_ESTEID_2018_V2_2025,
 	SC_CARD_TYPE_ESTEID_2025,
 
 	/* Rutoken cards */
--- a/src/libopensc/card-esteid2018.c
+++ b/src/libopensc/card-esteid2018.c
@@ -39,9 +39,9 @@
 #define PUK_REF		       0x02
 
 static const struct sc_atr_table esteid_atrs[] = {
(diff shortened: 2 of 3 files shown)


src/libopensc/cards.h
```diff
@@ -231,6 +231,7 @@ enum {
 
 	/* EstEID cards */
 	SC_CARD_TYPE_ESTEID_2018 = 35000,
+	SC_CARD_TYPE_ESTEID_2018_V2,
```
frankmorgner, line 234: we are not very strict with this, but I think you should add it behind the 2025 variant...

metsma: There will be 2025 card also in spring.
SC_CARD_TYPE_ESTEID_2018_V2_2025 ?

frankmorgner (changes requested):
If this is a generic problem with the reader, then I think this should be fixed at the reader level rather than at the card driver level. We already have such option available here:

https://github.com/OpenSC/OpenSC/blob/953986f65db61871bbbff72788d861d67d5140c6/src/libopensc/reader-pcsc.c#L1298-L1306

metsma:
> If this is a generic problem with the reader, then I think this should be fixed at the reader level rather than at the card driver level. We already have such option available here:
> 
> https://github.com/OpenSC/OpenSC/blob/953986f65db61871bbbff72788d861d67d5140c6/src/libopensc/reader-pcsc.c#L1298-L1306

I am not sure if it is readers fault. Right now it only affects the new Cosmo X cards.
Previous Cosmo 8 cards are working with this reader just fine and also upcoming Thales cards.

src/libopensc/cards.h
frankmorgner: This would assure that the number for `SC_CARD_TYPE_ESTEID_2025` doesn't change:
```
	SC_CARD_TYPE_ESTEID_2018 = 35000,
	SC_CARD_TYPE_ESTEID_2025,
	SC_CARD_TYPE_ESTEID_2018_V2,
```

frankmorgner (approved):

PR #3435, opened 2025-06-03
Upgrade Wix 6.0.0

metsma:
* Fix PKCS11-Spy location in arm64
* Make minidriver conditional
* Remove deprecated GithubActions Windows 2019 runner

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/setup-wix.ps1
+++ b/.github/setup-wix.ps1
@@ -1,5 +1,5 @@
 param(
-  [string]$version = '5.0.2'
+  [string]$version = '6.0.0'
 )
 & dotnet tool install -g --version $version wix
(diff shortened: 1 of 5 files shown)


metsma:
There are still some issues with Arm64 on windows. Arm64 and X64 use same registry space and minidriver will only point to single dll. I will investigate how to use https://learn.microsoft.com/en-us/windows/arm/arm64x-pe in separate PR

Jakuje (approved):
Removing of the w2019 runners is certainly good. I can not talk about all the other windows configurations though. But did not find there anything odd.

frankmorgner (approved):

PR #3436, opened 2025-06-03
fixed oberthur Stack-buffer-overflow WRITE

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-oberthur.c
+++ b/src/libopensc/card-oberthur.c
@@ -1127,7 +1127,7 @@ auth_compute_signature(struct sc_card *card, const unsigned char *in, size_t ile
 	apdu.lc = ilen;
 	apdu.le = olen > 256 ? 256 : olen;
 	apdu.resp = resp;
-	apdu.resplen = olen;
+	apdu.resplen = SC_MAX_APDU_BUFFER_SIZE;
 
 	rv = sc_transmit_apdu(card, &apdu);
 	LOG_TEST_RET(card->ctx, rv, "APDU transmit failed");
--- a/src/libopensc/card-oberthur.c
+++ b/src/libopensc/card-oberthur.c
@@ -227,7 +227,7 @@ auth_init(struct sc_card *card)
 	card->caps |= SC_CARD_CAP_RNG;
 	card->caps |= SC_CARD_CAP_USE_FCI_AC;
 
-	if (auth_select_aid(card))   {
+	if (auth_select_aid(card)) {
 		sc_log(card->ctx, "Failed to initialize %s", card->name);
 		rv = SC_ERROR_INVALID_CARD;
 		LOG_TEST_GOTO_ERR(card->ctx, SC_ERROR_INVALID_CARD, "Failed to initialize");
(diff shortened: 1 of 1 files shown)


src/libopensc/card-oberthur.c
```diff
@@ -2245,14 +2245,16 @@ auth_read_record(struct sc_card *card, unsigned int nr_rec, unsigned int idx,
 	if (flags & SC_RECORD_BY_REC_NR)
 		apdu.p2 |= 0x04;
 
-	apdu.le = count;
-	apdu.resplen = count;
+	apdu.le = count > SC_MAX_APDU_BUFFER_SIZE ? SC_MAX_APDU_BUFFER_SIZE : count;
```
Jakuje: please, use the MIN() macro here for easier readability.

src/libopensc/card-oberthur.c
```diff
@@ -2245,14 +2245,16 @@ auth_read_record(struct sc_card *card, unsigned int nr_rec, unsigned int idx,
 	if (flags & SC_RECORD_BY_REC_NR)
 		apdu.p2 |= 0x04;
 
-	apdu.le = count;
-	apdu.resplen = count;
+	apdu.le = MIN(count, SC_MAX_APDU_BUFFER_SIZE);
+	apdu.resplen = SC_MAX_APDU_BUFFER_SIZE;
```
Jakuje, line 2218: AFAIK this (including the `recvbuf` definition above) should have the length of `SC_MAX_APDU_RESP_SIZE`, instead of `SC_MAX_APDU_BUFFER_SIZE` currently. At least according to the comment next to its definition, this size is for encoding the whole APDU, not just the response buffer:

https://github.com/OpenSC/OpenSC/blob/f041813c3fdf76d80605647f34606827a8178784/src/libopensc/types.h#L34

But I am not sure if it is worth fixing it here now, given that this was not problem for last 20 years. Also not sure if this could break something (likely not).

src/libopensc/card-oberthur.c
```diff
@@ -1125,9 +1125,9 @@ auth_compute_signature(struct sc_card *card, const unsigned char *in, size_t ile
 	apdu.datalen = ilen;
 	apdu.data = in;
 	apdu.lc = ilen;
-	apdu.le = olen > 256 ? 256 : olen;
+	apdu.le = MIN(olen, 256);
```
Jakuje: I would use the `SC_MAX_APDU_RESP_SIZE` here too, instead of the magic number. And likely too for the `resp` buffer definition above as it has the same issue as the `read_record()` below.

src/libopensc/card-oberthur.c
frankmorgner: I only want to make sure that the buffer is not exceeded. Other semantic optimizations may be possible, but that needs to be tested, which I am unable to do.

Jakuje (approved):
thank you! The only nit is the formatting of one line pointed out by the clang. Otherwise lgtm

Issue #3418, opened 2025-05-14
global buffer overflow caught by ASAN


citypw:
Reproducible steps:
```
shawn@raspberrypi:~/repos/OpenSC $ file ./src/tools/.libs/pkcs11-tool
./src/tools/.libs/pkcs11-tool: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=f108d297bffc662df94e48c3a3008f1f42d985d9, for GNU/Linux 3.7.0, with debug_info, not stripped
shawn@raspberrypi:~/repos/OpenSC $ ./src/tools/.libs/pkcs11-tool
...
...
...
==2360==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00556b1915e8 at pc 0x00556b14fe14 bp 0x007fcfe1a840 sp 0x007fcfe1a858
READ of size 8 at 0x00556b1915e8 thread T0
    #0 0x556b14fe10 in util_print_usage /home/shawn/repos/OpenSC/src/tools/util.c:274
    #1 0x556b1503e0 in util_print_usage_and_die /home/shawn/repos/OpenSC/src/tools/util.c:310
    #2 0x556b10ea80 in main /home/shawn/repos/OpenSC/src/tools/pkcs11-tool.c:1235
    #3 0x7f890c773c in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #4 0x7f890c7814 in __libc_start_main_impl ../csu/libc-start.c:360
    #5 0x556b104cac in _start (/home/shawn/repos/OpenSC/src/tools/.libs/pkcs11-tool+0x24cac)

0x00556b1915e8 is located 0 bytes to the right of global variable 'option_help' defined in 'pkcs11-tool.c:339:20' (0x556b191360) of size 648
0x00556b1915e8 is located 56 bytes to the left of global variable 'app_name' defined in 'pkcs11-tool.c:431:21' (0x556b191620) of size 8
SUMMARY: AddressSanitizer: global-buffer-overflow /home/shawn/repos/OpenSC/src/tools/util.c:274 in util_print_usage
Shadow bytes around the buggy address:
  0x001aad632260: 00 00 00 00 00 00 f9 f9 f9 f9 f9 f9 00 00 00 00
  0x001aad632270: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[...]

PR #3419, opened 2025-05-14
Fix #3418: buffer overflow issue

citypw:
Fix #3418: buffer overflow issue

PR #3437, opened 2025-06-05
pkcs11-tool: Add missing help message to avoid buffer overrun

Jakuje:
This was uninitentionally omitted in 175ac15c37817d0641ecd87066158365e49c6418 and later reported as #3418. The PR #3419 was stale so fixing this in separate PR.

Fixes: #3418
Supersedes: #3419

##### Checklist


Jakuje:
This is also pretty non-controversial so merging now.

PR #3438, opened 2025-06-05
Coverity 2025

Jakuje:
Fix recent coverity reports.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-coolkey.c
+++ b/src/libopensc/pkcs15-coolkey.c
@@ -681,7 +681,7 @@ fail:
 		if (key) {
 			sc_pkcs15_free_pubkey(key);
 		}
-		if (r < 0) {
+		if (r != SC_SUCCESS) {
 			free(cert_info.value.value);
 		}
 	}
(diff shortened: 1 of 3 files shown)


Jakuje:
Sounds like the coolkey change will need some more love. I finally found a working coolkey card and this version is crashing.

src/libopensc/pkcs15-coolkey.c
```diff
@@ -246,6 +246,31 @@ coolkey_get_attribute_lv(sc_card_t *card, sc_cardctl_coolkey_object_t *obj, CK_A
 	return coolkey_get_attribute_bytes_alloc(card, obj, type, &item->data, &item->data_len);
 }
 
+/*
+ * A number of opensc structure have the same layout, use a common function to free them
```
frankmorgner: This won't work, because all those structures may be organized differently in memory by the compiler. Strictly speaking, this should also be undefined behavior.

If you make this a macro, this would work:

```
#define coolkey_free_attribute_lv(ptr) do { if (ptr) free(ptr->data) } while (0)
```

Jakuje: I took this mostly from the above `coolkey_get_attribute_lv()` which was already there (and working to some extent).

The problem with the macro is that the field is named differently in different variants (therefore the cast). But looking into that more, the `data`  in `sc_pkcs15_data` is the same member as `value` in `sc_pkcs15_der` so this should not really be necessary, as the coverity was having two separate reports for `value` and `data` fields. So hopefully just changing the condition will make it easier to understand for coverity ...

PR #3439, opened 2025-06-05
Oss-fuzz reported issue in OpenPGP

Jakuje:
The heavy rewrite of OpenPGP driver showed some issues in fuzzer, including missing checking of return values that could cause the driver to crash when card does not reply with the expected values.

This needs to be fixed before next release. I am worried that there will be more.

##### Checklist

- [ ] macOS token is tested


frankmorgner (approved):

Jakuje:
Sounds like the builds on ubuntu 22 fail to install debuginfo packages. It sounds like terribly fragile thing. Yesterday it failed to sync with mirrors so rerunning now, but it still fails. But I think most of the tests we run do not depend on this job, but would be good to get it fixed somehow eventually.

Jakuje:
From the run in #3348 it looks like this is an issue only for the arm64 images as the `build-22.04-amd64-valgrind` passed.  We should probably disable the `fail-fast` option so at least the amd jobs work, as it now cancels all the other jobs in the same matrix:

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Jakuje:
This should fix the CI problems we are having for last weeks by excluding the failing valgrind job. Please, have a look and merge so we can run the tests on new incoming PRs. @metsma @frankmorgner @xhanulik

Jakuje:
Thank you for the review! Merging so we can continue testing on other PRs

PR #3443, opened 2025-06-18
Coverity new fixes

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-esteid2018.c
+++ b/src/libopensc/card-esteid2018.c
@@ -122,9 +122,10 @@ static int esteid_select_file(struct sc_card *card, const struct sc_path *in_pat
 				r = iso_ops->process_fci(card, file, resp, resplen);
 				if (r != SC_SUCCESS) {
 					sc_file_free(file);
+				} else {
+					*file_out = file;
 				}
 				LOG_TEST_RET(card->ctx, r, "Process fci failed");
-				*file_out = file;
 			}
 			break;
 		}
--- a/src/pkcs15init/pkcs15-lib.c
+++ b/src/pkcs15init/pkcs15-lib.c
@@ -3945,7 +3945,7 @@ sc_pkcs15init_verify_secret(struct sc_profile *profile, struct sc_pkcs15_card *p
 		struct sc_path tmp_path = *path;
 		int iter;
 		r = SC_ERROR_OBJECT_NOT_FOUND;
-		for (iter = (int)tmp_path.len/2; iter >= 0 && r == SC_ERROR_OBJECT_NOT_FOUND; iter--, tmp_path.len -= 2) {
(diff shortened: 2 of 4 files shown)


metsma (approved):

xhanulik (approved):

src/libopensc/card-esteid2018.c
```diff
@@ -122,9 +122,10 @@ static int esteid_select_file(struct sc_card *card, const struct sc_path *in_pat
 				r = iso_ops->process_fci(card, file, resp, resplen);
 				if (r != SC_SUCCESS) {
 					sc_file_free(file);
+				} else {
+					*file_out = file;
 				}
 				LOG_TEST_RET(card->ctx, r, "Process fci failed");
```
Jakuje, line 128: I would not mind moving the LOG_TEST_RET into the branch too for clarity
```suggestion
					LOG_TEST_RET(card->ctx, r, "Process fci failed");
				} else {
					*file_out = file;
				}
```

Jakuje (approved):

Jakuje:
Hmm. The current master CI works, but runs in this PR fail to create pkcs15 structures in tests:
```
$ pkcs15-init --create-pkcs15 --so-pin 123456 --so-puk 0123456789abcdef
[...]
Failed to read PIN from user: Internal error
Failed to create PKCS #15 meta structure: Internal error
```
for all isoapplet and oseid so it sounds like there will be likely some issue that we missed. I will try to have a look what might have gone wrong here.

src/pkcs15init/pkcs15-lib.c
```diff
@@ -3945,7 +3945,7 @@ sc_pkcs15init_verify_secret(struct sc_profile *profile, struct sc_pkcs15_card *p
 		struct sc_path tmp_path = *path;
 		int iter;
 		r = SC_ERROR_OBJECT_NOT_FOUND;
-		for (iter = (int)tmp_path.len/2; iter >= 0 && r == SC_ERROR_OBJECT_NOT_FOUND; iter--, tmp_path.len -= 2) {
+		for (iter = (int)tmp_path.len/2; iter >= 0 && r == SC_ERROR_OBJECT_NOT_FOUND && tmp_path.len >= 2; iter--, tmp_path.len -= 2) {
```
Jakuje: my guess is this change is trigeering the failures as I guess that the `sc_pkcs15_find_pin_by_type_and_reference` could be called with the NULL path when the len is 0.

Jakuje: I think this comment still applies. Can you rebase on top of the current master to rule out the CI failure is some kind of fluke?

Jakuje (changes requested):

src/pkcs15init/pkcs15-lib.c
frankmorgner: I removed the commit regarding "427417 Overflowed constant" as coverity's description isn't clear to me anymore on second view.

Jakuje (approved):

PR #3445, opened 2025-06-20
initialize keyblob

### LINKED ISSUE #3444: sc-hsm-tool: --unwrap-key randomly fails with "realloc(): invalid pointer"
during some test runs while unwrapping keys with 
sc-hsm-tool --unwrap-key <filename> --key-reference <keyNo>

I ran into random errors with the following message:
realloc(): invalid pointer
Aborted

traced this back to 
static int unwrap_key(...)
where 
u8 *keyblob;
is uninitialized

marcwillert:
keyblob is a local variable 
as local variables are not initialized, its value is indeterminate

keyblob is passed into fread_to_eof(...) which will do a realloc on keyblob

Fixes #3444

Tested with:
Identive SCT3522CC token [CCID Interface]


Jakuje:
Thank you for the digging into the issue and providing fix! Looks pretty straightforward. Merging.

PR #3454, opened 2025-07-21
nqApplet Update: new cards and profiles

jozsefd:
This is an nqApplet card driver and PKCS15 emulator update for new card types and profiles. Besides the current JCOP4 based cards, new JCOP4.52 based cards are produced with nqApplet. The same applet is deployed on these cards, but a new ATR is used. Additionally, the RFID interface of the card is also enabled with its own ATR and card type.

The cards may include additional applets, which may interfere with the standard PIN CMD, so this PR offers a fallback by selecting the nqApplet if necessary. The encryption container may be missing from the applet, the PKCS15 emulator is adjusted to handle this option correctly.

Additionally, the existing and new nqApplet ATRs are registered for Minidriver support in the Windows Installer.

##### Checklist


jozsefd:
Thank you for the swift review

PR #3458, opened 2025-07-23
Fix bash-completion function name

xhanulik:
The `_split_longopt` is renamed to `_comp__split_longopt` from bash-completion v2.12.0.

### DIFF
--- a/doc/tools/completion-template
+++ b/doc/tools/completion-template
@@ -5,7 +5,11 @@ _FUNCTION_NAME()
     local cur prev split=false
     _get_comp_words_by_ref -n : cur prev
 
-    _split_longopt && split=true
+    if type _init_completion &>/dev/null; then
+        _init_completion -s && split=true
+    else
+        _split_longopt && split=true
+    fi
 
     opts="ALLOPTS"
 


doc/tools/completion-template
```diff
@@ -5,7 +5,11 @@ _FUNCTION_NAME()
     local cur prev split=false
     _get_comp_words_by_ref -n : cur prev
 
-    _split_longopt && split=true
+    if type _comp__split_longopt &>/dev/null; then
+        _comp__split_longopt && split=true
```
TheComputerGuy96: See https://github.com/scop/bash-completion/issues/1135#issuecomment-1998483710

xhanulik: Yes, they also mention in the conversation that also _`_split_longopt` has been an internal function_. Do you have some suggestion for another way how to fix this issue?

frankmorgner (approved):

doc/tools/completion-template
TheComputerGuy96: The comment linked by me above mentions a commit that adds a replacement for `_split_longopt` (and inside that commit many of the example scripts were switched from `_split_longopt` to `_init_completion -s` command) so that seems like a viable replacement (possibly with a `_split_longopt` fallback for ancient Linux distros)

TheComputerGuy96: Also the updated change shouldn't really check the unstable function name (`_init_completion` would be a better function to check)

doc/tools/completion-template
```diff
@@ -5,7 +5,11 @@ _FUNCTION_NAME()
     local cur prev split=false
     _get_comp_words_by_ref -n : cur prev
 
-    _split_longopt && split=true
+    if type _comp__split_longopt &>/dev/null; then
```
Jakuje: also here it should be changed, I think:
```suggestion
    if type _init_completion &>/dev/null; then
```

TheComputerGuy96 (approved):

PR #3462, opened 2025-08-04
Fix signing with IDPrime 5110 FIPS+

metsma:

##### Checklist

- [ ] macOS token is tested


Jakuje:
Please, de-indent the switch/case as suggested by the clang-format.

For the `idprime_atrs`, I would rather turn off the formatter for that part for the resulting output has a lot of needless tabs after the` SC_CARD_TYPE_IDPRIME_*` field, similarly as in the PIV driver:
https://github.com/OpenSC/OpenSC/blob/7f3dd28866fe96461519f7bafae5851b649d96e7/src/libopensc/card-piv.c#L214-L237

 Also the ATR and mask does not align under each other in the code making it is virtually impossible to manually figure out what bits are significant and what are not.

Otherwise, the change itself looks good. I do not have this specific card though.

frankmorgner (approved):

PR #3469, opened 2025-08-15
Fix new coverity issues 

Jakuje:
@hamarituc , can you check the dtrust change works as expected?

There is also one issue reported, which I am not completely sure how to fix:
```
*** CID 487695:           (FORWARD_NULL)
/src/tools/dtrust-tool.c: 875             in main()
869     		pin_resume = parse_pin(card, opt_resume, "Resume", NULL);
870     		if (pin_resume < 0)
871     			goto out;
872     	}
873     
874     	if (opt_unblock != NULL) {
>>>     CID 487695:           (FORWARD_NULL)
>>>     Passing null pointer "card" to "parse_pin", which dereferences it.
```
The tool is first reading the pins, but the function already needs to know the type of the card. But you delay the connect after that, which means it will likely not work as it is written.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-dtrust.c
+++ b/src/libopensc/card-dtrust.c
@@ -314,7 +314,7 @@ dtrust_init(sc_card_t *card)
 	if (can_value != NULL) {
 		size_t can_len;
 
-		can_len = strlen(can_env);
+		can_len = strlen(can_value);
 		drv_data->can_value = sc_mem_secure_alloc(can_len + 1);
 		if (drv_data->can_value == NULL) {
 			LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
--- a/src/tests/fuzzing/fuzz_pkcs11_uri.c
+++ b/src/tests/fuzzing/fuzz_pkcs11_uri.c
@@ -44,8 +44,10 @@ LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 	memcpy(input_string, Data, Size);
 	input_string[Size] = 0;
 
-	if ((uri = pkcs11_uri_new()) == NULL)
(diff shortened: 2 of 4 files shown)


hamarituc:
> There is also one issue reported, which I am not completely sure how to fix:
> 
> ```
> *** CID 487695:           (FORWARD_NULL)
> /src/tools/dtrust-tool.c: 875             in main()
> 869     		pin_resume = parse_pin(card, opt_resume, "Resume", NULL);
> 870     		if (pin_resume < 0)
> 871     			goto out;
> 872     	}
> 873     
> 874     	if (opt_unblock != NULL) {
> >>>     CID 487695:           (FORWARD_NULL)
> >>>     Passing null pointer "card" to "parse_pin", which dereferences it.
> ```
> 
> The tool is first reading the pins, but the function already needs to know the type of the card. But you delay the connect after that, which means it will likely not work as it is written.

This approach was to circumvent a Windows issue, there the card is locked for only 5 seconds. This is unacceptable when it is necessary to query the PIN interactively. The goal was to first input the PINs and then connect to the card. But the code just reads the PIN identifiers and not the PIN values.

Thus `dtrust-tool` needs to be reworked to
 * connect to the card
 * parse PIN identifiers
 * determines which PIN values to be queried
 * disconnects the card
 * query the PIN values
 * reconnect to the card
 * do the things with the card, which have to be done

#3480 fixes the NULL-pointer issues detected by coverity, but the fix for the above mentioned issue still needs to be implemented.

hamarituc:
> @hamarituc , can you check the dtrust change works as expected?

And yes, I checked it and I can confirm it works as expected.

frankmorgner:
closed with #3480

PR #3470, opened 2025-08-22
openpgp: Fix retry counter for card does not support query PIN status…

MkfsSion:
… by VERIFY command

According to OpenPGP spec, DO C4 is a simple DO which means response of GET DATA is not TLV encoded data. So retry counter bytes starts from byte 5 (count from 1) of buffer. Since 81 <= PIN reference <= 83, access to error counter would start from byte 6 of buffer and get wrong data.

Tested with opensc-explorer
Before this change:
```
OpenSC Explorer version 0.26.1
Using reader with a card: Yubico YubiKey OTP+FIDO+CCID 00 00
OpenSC [3F00]> pin_info CHV1
Login status unknown.
3 tries left.
OpenSC [3F00]> pin_info CHV2
Login status unknown.
3 tries left.
OpenSC [3F00]> pin_info CHV3
Login status unknown.
113 tries left.
```

After this change:
```
OpenSC Explorer version 0.26.1
Using reader with a card: Yubico YubiKey OTP+FIDO+CCID 00 00
OpenSC [3F00]> pin_info CHV1
Login status unknown.
3 tries left.
OpenSC [3F00]> pin_info CHV2
Login status unknown.
3 tries left.
OpenSC [3F00]> pin_info CHV3
Login status unknown.
3 tries left.
```

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2172,9 +2172,9 @@ pgp_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 			LOG_TEST_RET(card->ctx, SC_ERROR_OBJECT_NOT_VALID,
 				"CHV status bytes have unexpected length");
 
-                data->pin1.tries_left = c4data[4 + (data->pin_reference & 0x0F)];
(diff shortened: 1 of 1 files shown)


frankmorgner:
thank you

PR #3126, opened 2024-04-24
pkcs11: leverage sc_wait_for_event to avoid unneccesary queries to card

frankmorgner:
card_detect_all() is now used as global entry point even to reader specific card detection (card_detect() is removed). This enforces a constant update of reader_states so that no events are silently dropped. This removes the need for having a timer in C_GetSlotInfo() for avoiding too many queries. If the reader implementation (i.e. reader-*.c) does not implement wait_for_event, then all slots will be queried on request just like it used to be.

fixes https://github.com/OpenSC/OpenSC/issues/3107

##### Checklist

- [x] PKCS#11 module is tested
- [ ] Windows is tested (pkcs11-tool with --test-hotplug, --test, --test-threads)
- [ ] Linux is tested (pkcs11-tool with --test-hotplug, --test, --test-threads)
- [ ] macOS is tested (pkcs11-tool with --test-hotplug, --test, --test-threads)

PR #3472, opened 2025-08-27
pkcs11/slot.c  If first  sc_card_connect fails, mark reader with SC_READER_CARD_INVALID

### LINKED ISSUE #3471: OpenSC pkcs11 module making Firefox ESR unusable on Debian Trixie when Croatian eID or Certilia cards are connected to smartcard reader
### Problem Description

dengert:
Fixes #3471

With pkcs11, If a card in a reader fails in `sc_card_connect`, `reader->flags |= SC_READER_CARD_INVALID;`  is set. If at a later time `C_GetSlotList`  is called by the application like FireFox, and SC_READER_CARD_INVALID is set, OpenSC will not call sc_card_connect again and will avoid all the overhead of trying to connect to the same card.

The SC_READER_CARD_INVALID will be reset if the reader reports SC_READER_CARD_CHANGED as a different card may have been inserted.

 On branch pkcs11-invalid-card
 Changes to be committed:
	modified:   libopensc/opensc.h
	modified:   pkcs11/slot.c

Partially tested on Ubuntu 24.04 with FireFox ESR  simulating an invalid card, by 
`export OPENSC_DRIVER=openpgp`  was set so any non-openpgp card will be considered invalid. 
 

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/opensc.h
+++ b/src/libopensc/opensc.h
@@ -377,6 +377,7 @@ struct sc_reader_driver {
 #define SC_READER_HAS_WAITING_AREA	0x00000010
 #define SC_READER_REMOVED			0x00000020
 #define SC_READER_ENABLE_ESCAPE		0x00000040
+#define SC_READER_CARD_INVALID			0x00000080
 
 /* reader capabilities */
 #define SC_READER_CAP_DISPLAY	0x00000001
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -424,6 +424,7 @@ static int refresh_attributes(sc_reader_t *reader)
 			} else {
 				reader->flags |= SC_READER_CARD_CHANGED;
 			}
+			reader->flags &= ~SC_READER_CARD_INVALID;
 			reader->flags &= ~SC_READER_CARD_PRESENT;
 			reader->flags |= SC_READER_REMOVED;
 			priv->gpriv->removed_reader = reader;
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -1654,7 +1654,8 @@ pkcs15_create_tokens(struct sc_pkcs11_card *p11card, struct sc_app_info *app_inf
 	}
 
 	/* Find out framework data corresponding to the given application */
-	fw_data = get_fw_data(p11card, app_info, &idx);
+	if (p11card)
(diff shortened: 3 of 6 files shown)


psadaic:
Everything looks good to me.

frankmorgner:
What could be problematic, is that you're defining a reader flag at the lowest layer possible, which will only be used at the PKCS#11 level. This could lead to problems when lower layers are accessing or modifying the reader flags.

I think SC_READER_CARD_CHANGED is returned from sc_detect_card_presence, when reader-pcsc.c detects some change. In the past, we have seen that this change might get swallowed in some other code path so that the signal of a change never arrives at the PKCS#11 layer.

I think #3126 will achieve the same goal as this PR, with the advantage of having the state tracking exclusively in the PKCS#11 layer so that it cannot be accidentally overwritten somewhere else. Maybe it is worth having a second look there.

dengert:
With further testing this is still not correct. The SC_READER_CARD_INVALID gets set, but removing the card and reinserting is not doing what I would suspect.

dengert:
I have a working version Using Firefox-esr on Ubuntu 24.04 Firefox 142.0.1 installed in ~/firefox
`doug@ubuntu-24-04:~$ OPENSC_DRIVER=PIV-II gdb --args firefox/firefox` 

I can start with or without a card in the  reader and switch back in forth between  a PIV card or another card which does not have a PIV applet will be considered invalid because OpenSC is only going to match a PIV card in my test case. 

An attempt to match the invalid card is only done once for each insertion. 

@frankmorgner as you pointed out setting the reader flag did need a change in `reader-pcsc.c` to turn it off.

The `sc_pkcs11_card`  `p11card` is not freed.  Note `rc = sc_connect_card(reader, &pkcs11->card);` would normally fill in 
`pkcs11->card` but not when there is no driver for the card. This allows a slot to point at the `p11card`Oter wise it would not add  a slot for it.  https://github.com/OpenSC/OpenSC/pull/3472/commits/ece943eb2528eba7b8087394372a0bee8822a9f1 -  everywhere where `p11card->card->ctx` was replaced with `context` which is the same ctx that is used by PKCS11 when  `card` is not available.    

This is a problem with applications like FireFox that polls by calling
C_GetSlotList once a second and a card not supported by OpenSC is used
which could be handled by another PKCS11 module when using NSS or p11-kit.

frankmorgner:
closing with https://github.com/OpenSC/OpenSC/pull/3126

Issue #3107, opened 2024-04-10
OpenSC in Firefox constantly sending SELECT(AID) APDUs for probing


CardContact:
### Problem Description

Environment: OpenSC 0.25 on Debian Testing with Firefox ESR.

If the card inserted into the reader is not recognized by OpenSC, then SELECT(AID) APDUs are issued at a high frequency, probably for probing. This interferes with other applications accessing the card.

Most APDUs contain the ActivCard AID (A0 00 00 00 79):

    00000016 [140675220883136] APDU: 00 CB 3F FF 03 5C 01 7E 00 
    00005039 [140675220883136] SW: 6E 00 
    00000021 [140675220883136] APDU: 00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00 
    00007490 [140675220883136] SW: 6A 82 
    00000012 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 01 16 DB 00 00 
    00006727 [140675220883136] SW: 6A 82 
    00000011 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 10 00 00 
    00006848 [140675220883136] SW: 6A 82 
    00000010 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 00 00 
    00006763 [140675220883136] SW: 6A 82 
    00000007 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 01 00 
    00007016 [140675220883136] SW: 6A 82 
    00000013 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 02 00 
    00006937 [140675220883136] SW: 6A 82 
    00000021 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 03 00 
    00007193 [140675220883136] SW: 6A 82 
    00000034 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 04 00 
    00006927 [140675220883136] SW: 6A 82 
    00000012 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 05 00 
    00006790 [140675220883136] SW: 6A 82 
    00000014 [140675220883136] APDU: 00 A4 04 0C 07 A0 00 00 00 79 01 06 00 
    00006715 [140675220883136] SW: 6A 82 
[...]

PR #3473, opened 2025-08-28
pkcs11: perform driver matching with invalid card only once while card is present in reader

### LINKED ISSUE #3471: OpenSC pkcs11 module making Firefox ESR unusable on Debian Trixie when Croatian eID or Certilia cards are connected to smartcard reader

psadaic:
sc_connect_card is modified to allow invalid card to remain connected. Connection between unsupported card handle and pkcs11 slot is now preserved, allowing correct status detection. Side effect of this modification is that sc_disconnect_card is now required to be called later on to perform card cleanup.

On top of that SC_PKCS11_CARD_INVALID flag is introduced for sc_pkcs11_card structure. This flag is set in case sc_connect_card returns SC_ERROR_INVALID_CARD on card_detect allowing us to preserve info about card support for the duration of card presence in reader.

Set SC_PKCS11_CARD_INVALID flag prevents running additional unnecessary driver matching actions on unsupported cards.

Other pksc11 methods utilizing sc_pkcs11_card objects are updated to include SC_PKCS11_CARD_INVALID flag checks.

fixes #3471 , possibly also #3107 , #3108

@dengert @frankmorgner @Jakuje please have a look, would appreciate your feedback on this.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card.c
+++ b/src/libopensc/card.c
@@ -242,7 +242,7 @@ int sc_connect_card(sc_reader_t *reader, sc_card_t **card_out)
 	sc_card_t *card;
 	sc_context_t *ctx;
 	struct sc_card_driver *driver;
-	int i, r = 0, idx, connected = 0;
+	int i, r = 0, idx;
 
 	if (card_out == NULL || reader == NULL)
 		return SC_ERROR_INVALID_ARGUMENTS;
@@ -258,7 +258,6 @@ int sc_connect_card(sc_reader_t *reader, sc_card_t **card_out)
 	if (r)
 		goto err;
 
-	connected = 1;
 	card->reader = reader;
 	card->ctx = ctx;
 
@@ -392,10 +391,7 @@ int sc_connect_card(sc_reader_t *reader, sc_card_t **card_out)
 
 	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
 err:
-	if (connected)
-		reader->ops->disconnect(reader);
-	if (card != NULL)
-		sc_card_free(card);
+	*card_out = card;
 	LOG_FUNC_RETURN(ctx, r);
 }
 
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -7173,6 +7173,7 @@ static DWORD associate_card(PCARD_DATA pCardData)
 	r = sc_connect_card(vs->reader, &(vs->card));
 	if (r != SC_SUCCESS) {
 		logprintf(pCardData, 0, "Cannot connect card in reader '%s'\n", NULLSTR(vs->reader->name));
+		sc_disconnect_card(vs->card);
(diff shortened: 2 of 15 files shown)


frankmorgner:
In https://github.com/OpenSC/OpenSC/pull/3472 [I have argued](https://github.com/OpenSC/OpenSC/pull/3472#issuecomment-3227698543) that using `sc_detect_card_presence` to detect a new card event for a PC/SC reader is problematic, because it uses one single low level monitoring state that is accessed at multiple places. In reader-pcsc.c this is done in refresh_attributes, which is called at multiple occasions. Yes, in refresh_attributes it will be detected, when an invalid card is changed for a different card, but we cannot be sure that this event also propagates to the PKCS#11 layer (i.e. to `card_detect` in slot.c).

However, I reviewed the argument and the possible occasions where `refresh_attributes` might be called directly or indirectly. And I believe in its current implementation it should work OK, because if the PKCS#11 layer won't touch the invalid card, then `refresh_attributes` will only be called in `detect_card` (slot.c via sc_detect_card_presence) so that a changed card will correctly be detected.

I am still hesitant, because the suggested design seems a little fragile to me. We might forget about this problem in the future and accidentally use `sc_detect_card_presence` at some other place introducing the problem that a changed card can sometimes not be detected anymore. Additionally this PR makes very low level changes which might break other things.

What I like about this PR is that it tries to solve the problem at the PKCS#11 layer, but I'd additionally like to see the monitoring state also to be managed at this layer instead of using the internal state from reader-pcsc.c. Essentially this would mean to use sc_wait_for_event with a local monitoring state in slot.c instead of using sc_detect_card_presence (see https://github.com/OpenSC/OpenSC/pull/3126).

psadaic:
@frankmorgner thank you for a very detailed reply. You're right, `sc_detect_card_presence` could be fragile in the future and I like your approach in #3126 , will take another look at that.

I believe the main issue why invalid cards are getting checked repeatedly is here: https://github.com/OpenSC/OpenSC/blob/972d19aeb72e7f9af5a6d000c40b03d4ec83bf4d/src/libopensc/card.c#L396

With every failed `sc_connect_card` call due to driver not being matched new card event is created that will trigger another card check on the next call. I do not know if we can prevent that with a change on the pkcs11 layer only. I'm unable to assess how much of an issue change in `sc_connect_card` could cause with other parts of OpenSC but I'm open to suggestions if you believe I should take a different approach in solving this.

frankmorgner:
As said before, sc_wait_for_event (SCardGetStatusChange) uses a global state which is provided by PC/SC. Those events are provided in a bit mask, which allow sorting out actual insertion/removel events - independant from sc_connect_card that we are doing in OpenSC.

I think it is essential to keep the code you have linked above. If a card is not recognized (invalid) in OpenSC, then you want to disconnect from that card so that other processes can still use this card exclusively, for example.

psadaic:
Okay, thanks for clarifying that out. Will focus on #3126 then since that seems like the better approach.

Issue #3466, opened 2025-08-07
Failure when using `dnie-tool`


frankmorgner:
When I connect my Spanish DNIe, and try to run dnie-tool, it works fine. But it crashes if I provide my PIN. Similarly, Firefox crashes if I try to Log In the SmartCard reader in the Security Devices page.

> I've just test and I can confirm that I got the same error as https://github.com/OpenSC/OpenSC/issues/3390#issuecomment-3102073809 
> 
> `dnie-tool --pin XXXXXXX --all -vvv`
> 
> ```
> P:36709; T:0x4529550848 02:59:26.478 [dnie-tool] ctx.c:975:sc_context_create: ===================================
> P:36709; T:0x4529550848 02:59:26.478 [dnie-tool] ctx.c:976:sc_context_create: OpenSC version: 0.26.1
> P:36709; T:0x4529550848 02:59:26.478 [dnie-tool] ctx.c:977:sc_context_create: Configured for dnie-tool (/Library/OpenSC/bin/dnie-tool)
> P:36709; T:0x4529550848 02:59:26.479 [dnie-tool] ctx.c:852:sc_openssl3_init: Failed to load OpenSSL Legacy provider
> P:36709; T:0x4529550848 02:59:26.479 [dnie-tool] reader-pcsc.c:915:pcsc_init: PC/SC options: connect_exclusive=0 disconnect_action=0 transaction_end_action=0 reconnect_action=0 enable_pinpad=1 enable_pace=1
> P:36709; T:0x4529550848 02:59:26.480 [dnie-tool] reader-pcsc.c:1415:pcsc_detect_readers: called
> P:36709; T:0x4529550848 02:59:26.480 [dnie-tool] reader-pcsc.c:1428:pcsc_detect_readers: Probing PC/SC readers
> P:36709; T:0x4529550848 02:59:26.480 [dnie-tool] reader-pcsc.c:1481:pcsc_detect_readers: Establish PC/SC context
> P:36709; T:0x4529550848 02:59:26.497 [dnie-tool] reader-pcsc.c:1364:pcsc_add_reader: Adding new PC/SC reader 'MSI StarReader SMART'
> P:36709; T:0x4529550848 02:59:26.497 [dnie-tool] reader-pcsc.c:373:refresh_attributes: MSI StarReader SMART check
> P:36709; T:0x4529550848 02:59:26.499 [dnie-tool] reader-pcsc.c:439:refresh_attributes: current  state: 0x00000122
[...]

PR #3476, opened 2025-09-01
DNIe: Always select MF without SM to delete session at card

frankmorgner:
Explicitly disabling SM is, for example, done at logout, finish and init which sets the card into a usable state when being used concurrently

fixes https://github.com/OpenSC/OpenSC/issues/3466

##### Checklist

- [ ] macOS token is tested


rickyepoderi (approved):
Thanks @frankmorgner! The change worked OK for me.

frankmorgner:
thanks for testing

PR #3477, opened 2025-09-03
pkcs11-tool: Add id/slot-id to PKCS#11 URIs

sarroutbi:
Extend pkcs11-tool to include additional identifier parameters in PKCS#11 URIs to improve object and token identification:

For -L option (list slots):
- Add slot-id parameter to token URIs to identify specific slots
- Before: pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=84914bc36f04a55d;token=clevis
- After:  pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=84914bc36f04a55d;token=clevis;slot-id=0

For -O option (list objects):
- Add id parameter to data object URIs when CKA_ID attribute is available
- Previously only keys and certificates included id parameters
- Data objects now consistently show id parameters using getID() function

Changes made:
1. Created get_uri_with_slot() function to generate token URIs with slot-id parameter
2. Modified show_token() to use new URI generation function with slot identification
3. Enhanced show_dobj() to retrieve and include object IDs for data objects
4. Added proper ID handling using existing getID() function for PKCS#11 objects

The implementation correctly distinguishes between:
- Token identifiers: slot-id parameter for slot/token identification
- Object identifiers: id parameter from CKA_ID attribute for PKCS#11 objects

Card used for testing:
```bash
opensc-tool -n
Using reader with a card: Yubico YubiKey OTP+CCID 00 00
```

###### Checklist

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -1899,6 +1899,24 @@ get_uri(CK_TOKEN_INFO_PTR info)
 	return copy_key_value_to_uri("token=", token, CK_TRUE);
 }
 
+static const char *
+get_uri_with_slot(CK_TOKEN_INFO_PTR info, CK_SLOT_ID slot)
+{
+	copy_key_value_to_uri("pkcs11:", NULL, CK_FALSE);
+	const char *model = percent_encode(info->model, sizeof(info->model));
+	copy_key_value_to_uri("model=", model, CK_FALSE);
+	const char *manufacturer = percent_encode(info->manufacturerID, sizeof(info->manufacturerID));
+	copy_key_value_to_uri("manufacturer=", manufacturer, CK_FALSE);
+	const char *serial = percent_encode(info->serialNumber, sizeof(info->serialNumber));
+	copy_key_value_to_uri("serial=", serial, CK_FALSE);
+	const char *token = percent_encode(info->label, sizeof(info->label));
+	copy_key_value_to_uri("token=", token, CK_FALSE);
+
+	static char slot_id_str[32];
+	snprintf(slot_id_str, sizeof(slot_id_str), "%lu", slot);
+	return copy_key_value_to_uri("slot-id=", slot_id_str, CK_TRUE);
+}
+
 static void show_token(CK_SLOT_ID slot)
 {
 	CK_TOKEN_INFO	info;
@@ -1933,7 +1951,7 @@ static void show_token(CK_SLOT_ID slot)
 	printf("  serial num         : %s\n", p11_utf8_to_local(info.serialNumber,
 			sizeof(info.serialNumber)));
 	printf("  pin min/max        : %lu/%lu\n", info.ulMinPinLen, info.ulMaxPinLen);
-	printf("  uri                : %s", get_uri(&info));
+	printf("  uri                : %s", get_uri_with_slot(&info, slot));
 	printf("\n");
 }
 
@@ -6419,9 +6437,11 @@ static void show_cert(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 static void show_dobj(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 {
 	unsigned char *oid_buf;
+	unsigned char *id;
 	char *label;
 	char *application;
 	CK_ULONG    size = 0;
+	CK_ULONG idsize = 0;
 	CK_TOKEN_INFO info;
 	CK_BBOOL modifiable = 0;
 	CK_BBOOL private = 0;
@@ -6476,6 +6496,12 @@ static void show_dobj(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 
 	get_token_info(opt_slot, &info);
 	printf("  uri:            %s", get_uri(&info));
+	if ((id = getID(sess, obj, &idsize)) != NULL && idsize) {
+		printf(";id=");
+		for (unsigned int n = 0; n < idsize; n++)
+			printf("%%%02x", id[n]);
+		free(id);
+	}
 	if (label != NULL) {
 		const char *pelabel = percent_encode((unsigned char *)label, strlen(label));
 		printf(";object=%s", pelabel);


dengert:
PKCS11 only define CKA_ID for certificates and keys. Not data objects.
The CKA_ID can be used to associate certificate, public key and private key.
Data objects as well as certificates, public, private and pins have labels.
PKCS15 does have a ID for data objects. 

NIST sp800-73-* defines the names and OIDs and these are in  https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c
Internally card-piv.c and pkcs15-piv.c use the ` const char *id;`

sarroutbi:
> PKCS11 only define CKA_ID for certificates and keys. Not data objects. The CKA_ID can be used to associate certificate, public key and private key. Data objects as well as certificates, public, private and pins have labels. PKCS15 does have a ID for data objects.
> 
> NIST sp800-73-* defines the names and OIDs and these are in https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-piv.c Internally card-piv.c and pkcs15-piv.c use the ` const char *id;`

Thanks for clarification. May I keep slot-id at least for the token URI?

dengert:
See [RFC7515](https://datatracker.ietf.org/doc/html/rfc7512) to see how URI component names map to PKCS11 defines.
for example "object" maps to CKA_LABEL and `slot-id` map to `CK_SLOT_ID'.  But `CK_SLOT_ID` may not stay the same, if there are more than one readers or tokens available.

Latest is [PKCS11-3.2-draft](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html)

Jakuje:
As @dengert already mentioned, the use of slot ID in PKCS#11 URI is discouraged so I would like to avoid including it. See 

https://www.rfc-editor.org/rfc/rfc7512.html#page-9

For the CKA_ID for data objects, even though OpenSC might provide them in the Yubikey tokens, they are not defined by the standard so I do not think it is a good idea to include them in the URI either.

Does either of these solve some problem you had with some particular use case where you needed these?

mouse07410:
**Use** of slot IDs is discouraged (though not prohibited). That is reasonable. 

The **ability** to use slot IDs, however, needs to be there, for those (hopefully) rate cases when it's really necessary. 

No evens, buts, etc.

Jakuje:
> **Use** of slot IDs is discouraged (though not prohibited). That is reasonable.
> 
> The **ability** to use slot IDs, however, needs to be there, for those (hopefully) rate cases when it's really necessary.

I do not discourage the ability to use the slot IDs. I think it is ok to use them if they are provided as part of the uri to a user. But this change is about printing them by default for all objects, which, to my understanding, will bring more harm than use as people tend to copy the uri without inspection./understanding and then end up with something that wont work if kernel decides to loads usb readers in different order.

mouse07410:
> this change is about printing them by default for all objects, which, to my understanding, will bring more harm than use 

You have a point. 

I guess, as long as there is a way to print all the objects with their slot IDs, it doesn't need to be **by default**.

src/tools/pkcs11-tool.c
```diff
@@ -1899,6 +1906,24 @@ get_uri(CK_TOKEN_INFO_PTR info)
...
+{
+	copy_key_value_to_uri("pkcs11:", NULL, CK_FALSE);
+	const char *model = percent_encode(info->model, sizeof(info->model));
+	copy_key_value_to_uri("model=", model, CK_FALSE);
+	const char *manufacturer = percent_encode(info->manufacturerID, sizeof(info->manufacturerID));
+	copy_key_value_to_uri("manufacturer=", manufacturer, CK_FALSE);
+	const char *serial = percent_encode(info->serialNumber, sizeof(info->serialNumber));
+	copy_key_value_to_uri("serial=", serial, CK_FALSE);
+	const char *token = percent_encode(info->label, sizeof(info->label));
+	copy_key_value_to_uri("token=", token, CK_FALSE);
```
Jakuje: There is a lot of duplicate information.

Can we move the condition for `opt_uri_with_slot_id` into this function (given that the variable is already global) to avoid code duplication when calling the function.

On a side note, this will work only when the slot id is specified on the cli too. I think for the code to be robust, it should add the slot ID into the URI also when there is no `--slot-id` on CLI as every token needs to be in some slot.

src/tools/pkcs11-tool.c
```diff
@@ -1896,6 +1903,14 @@ get_uri(CK_TOKEN_INFO_PTR info)
...
 	const char *token = percent_encode(info->label, sizeof(info->label));
+
+	if (opt_uri_with_slot_id) {
+		copy_key_value_to_uri("token=", token, CK_FALSE);
+		static char slot_id_str[32];
+		snprintf(slot_id_str, sizeof(slot_id_str), "%lu", slot);
+		return copy_key_value_to_uri("slot-id=", slot_id_str, CK_TRUE);
+	}
+
 	return copy_key_value_to_uri("token=", token, CK_TRUE);
```
Jakuje: the order of the parameters in the uri really does not matter so I would rather move it to the beginning than duplicating the adding the token label in two places. It makes it hard to follow and understand.
```suggestion

	if (opt_uri_with_slot_id) {
		static char slot_id_str[32];
		snprintf(slot_id_str, sizeof(slot_id_str), "%lu", slot);
		copy_key_value_to_uri("slot-id=", slot_id_str, CK_FALSE);
	}

	const char *token = percent_encode(info->label, sizeof(info->label));
	return copy_key_value_to_uri("token=", token, CK_TRUE);
```
(or even before all the other parameters)

src/tools/pkcs11-tool.c
```diff
@@ -6475,7 +6491,13 @@ static void show_dobj(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
...
 
 	get_token_info(opt_slot, &info);
-	printf("  uri:            %s", get_uri(&info));
+	printf("  uri:            %s", get_uri(&info, opt_slot));
+	if ((id = getID(sess, obj, &idsize)) != NULL && idsize) {
+		printf(";id=");
+		for (unsigned int n = 0; n < idsize; n++)
+			printf("%%%02x", id[n]);
+		free(id);
+	}
```
Jakuje: I think we discussed somewhere previously, that the Data objects do not have the CKA_ID attribute defined. So even if it would work for some custom implementations, it would result in false errors/warnings on compliant implementations, that I would like to avoid.

https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203218

sarroutbi: Sorry. You are totally right. Removing it.

Jakuje (approved):

sarroutbi:
@Jakuje , @frankmorgner : is it possible for some of you to re-run `Linux / test-cac-22.04 (pull_request)` job?

frankmorgner (approved):
thank you.

all CI jobs passed successfully

Jakuje:
Thank you! Merging then

PR #3480, opened 2025-09-10
Fix coverity issues in dtrust driver

hamarituc:
This PR fixes coverity issued in the `dtrust` driver. It already contains the fixes of @Jakuje in #3469 and furthermore fixes the issue which was left open there.

##### Checklist
- [ ] macOS token is tested


Jakuje (approved):
Thank you for having a look into this!

PR #3485, opened 2025-09-24
Change meta data

swissbit-csteuer:
Change product names, URLs, and other metadata that is, e.g., compiled into the properties of DLLs to indicate that this is not a build of the actual OpenSC project but a modified version from Swissbit.


swissbit-csteuer:
Sorry. Wrong target repo.

Issue #2875, opened 2023-09-22
C_Initialize() calls itself in a loop when pkcs-spy is used with SoftHSM from Homebrew (on macOS)


LudovicRousseau:
### Problem Description

I try to use pkcs11-spy on macOS Ventura 13.6 (22G120). 
My problem is that `C_Initialize()` from pkcs-spy calls itself recursively until the stack is full and the application crashes.
Instead `C_Initialize()` from pkcs-spy should call `C_Initialize()` from the PKCS#11 lib I configured in PKCS11SPY.

I tried with the current git version of OpenSC.
```
PKCS11SPY=/usr/local/lib/softhsm/libsofthsm2.so pkcs11-tool -I --module=/usr/local/lib/pkcs11-spy.so


*************** OpenSC PKCS#11 spy *****************
Loaded: "/usr/local/lib/softhsm/libsofthsm2.so"

0: C_GetInterface
P:49666; T:0x140704475666176 2023-09-22 23:20:36.564
[compat]
[in] pInterfaceName 000000010a03bb6c / 7
    00000000  50 4B 43 53 20 31 31                             PKCS 11         
[in] pVersion = NULL
[in] flags = 
Returned:  0 CKR_OK

1: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:36.564
[in] pInitArgs = 0x0

2: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:36.564
[in] pInitArgs = 0x0

3: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:36.565
[in] pInitArgs = 0x0

4: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:36.565
[in] pInitArgs = 0x0

[...]

174597: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:37.798
[in] pInitArgs = 0x0

174598: C_Initialize
P:49666; T:0x140704475666176 2023-09-22 23:20:37.798
[in] pInitArgs = 0x0
Segmentation fault: 11
```

### Proposed Resolution

I have no solution.
I am pretty sure I already got this problem a few years back. But I do not remember how I solved it.

It looks like a symbol resolution issue between `C_Initialize()` from pkcs-spy and `C_Initialize()` from the PKCS#11 library I want to spy.

I guess the issue is specific to macOS.

### Steps to reproduce

```
[...]

PR #3487, opened 2025-09-26
fixed overwriting global module functions, e.g. C_Initialize

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2875

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/common/libscdl.c
+++ b/src/common/libscdl.c
@@ -70,7 +70,11 @@ int sc_dlclose(void *handle)
 
 void *sc_dlopen(const char *filename)
 {
-	return dlopen(filename, RTLD_LAZY | RTLD_LOCAL);
+	return dlopen(filename, RTLD_LAZY | RTLD_LOCAL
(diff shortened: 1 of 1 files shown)


Jakuje:
I am not sure if the Ubuntu 22 with valgrind failures are relevant to the DEEPBIND change. I know it messed up something with address sanitizer, but it might be also messing up the valgrind.

Jakuje:
Thanks for the clarification in #2875! I think its just the ubuntu 22 + valgrind broken again. It fails in other PRs too.

frankmorgner:
Your comment came after I rewrote this to use dlmopen. Let's give it a try in CI. The manual lists some bugs for dlmopen, that are a little worrysome. So I'm not 100% sure if dlmopen should be preferred over RTLD_DEEPBIND

Jakuje (approved):
this sounds like the safest option to do right now.

PR #3489, opened 2025-09-28
docker-image.yml

m7mdragab:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,18 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


frankmorgner:
We don't have a Dockerfile

Jakuje:
That sounds to me like a spam PR from a bot account hoping somebody will get merge the PRs and then they will be able to run in their CI some malicious code. Given that the PR s completely phony and there was really no follow-up, I would just go ahead and close it and likely report it as spam, if you do not mind.

frankmorgner:
Do you want to imply that once we enable CI run for a pull request, that additional (or modified) commits in this pull requests would also be executed (i.e. in this case _with_ a malicious Dockerfile)?

Jakuje:
> Do you want to imply that once we enable CI run for a pull request, that additional (or modified) commits in this pull requests would also be executed (i.e. in this case _with_ a malicious Dockerfile)?

Probably not, but I think it depends on sloppiness of people approving every pipeline. It could run automatically only if we would get some code from this user merged. But some people having Dockerfile in the root of repo might just see it as an useful addition and merge it.

PR #3490, opened 2025-09-30
WIP: Check if PIN is locked and hint CKF_USER_PIN_TO_BE_CHANGED

metsma:
##### Checklist

- [ ] macOS token is tested


metsma:
I am not sure if this is correct and best approach to notifying this in PKCS#11.

metsma:
> I do not have a way to test it and I can not think of any better way to do this.

I did not figure out how to map the internal flag to PKCS11 flag without defining custom PKCS15 flag.
/* vendor-specific flags */
#define SC_PKCS15_PIN_FLAG_VENDOR_NEEDS_CHANGE 0x1000

frankmorgner:
If a PIN needs to be changed, then this information should be associated with a specific PIN, right? In `opensc.h` we do already have `sc_pin_cmd_pin.logged_in`:
```
/** Values for sc_pin_cmd_pin.logged_in */
#define SC_PIN_STATE_UNKNOWN	-1
#define SC_PIN_STATE_LOGGED_OUT 0
#define SC_PIN_STATE_LOGGED_IN  1

/* A card driver receives the sc_pin_cmd_data and sc_pin_cmd_pin structures filled in by the
 * caller, with the exception of the fields returned by the driver for SC_PIN_CMD_GET_INFO.
 * It may use and update any of the fields before passing the structure to the ISO 7816 layer for
 * processing.
 */
struct sc_pin_cmd_pin {
	const char *prompt;	/* Prompt to display */

	const unsigned char *data; /* PIN, set to NULL when using pin pad */
	size_t len;		/* set to 0 when using pin pad */

	size_t min_length;	/* min length of PIN */
	size_t max_length;	/* max length of PIN */

	unsigned int encoding;	/* ASCII-numeric, BCD, etc */

	size_t pad_length;	/* PIN padding options, used with SC_PIN_CMD_NEED_PADDING */
	unsigned char pad_char;

	size_t offset;		/* PIN offset in the APDU when using pin pad */

	int max_tries;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
	int tries_left;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
	int logged_in;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
};
```
Would maybe `#define SC_PIN_STATE_NEEDS_CHANGE  2` a better option? This would not require any modification in `pkcs15.h` and instead `logged_in` could be accessed directly from the PKCS#11 layer (as currently done in `slot.c`).

Maybe the naming `logged_in` would then not be appropriate anymore and could be replaceced by `state`, for example. Also, a bit mask may be more useful than a number of distinct values.

metsma:
> /* A card driver receives the sc_pin_cmd_data and sc_pin_cmd_pin structures filled in by the
>  * caller, with the exception of the fields returned by the driver for SC_PIN_CMD_GET_INFO.
>  * It may use and update any of the fields before passing the structure to the ISO 7816 layer for
>  * processing.
>  */
> struct sc_pin_cmd_pin {
> 	const char *prompt;	/* Prompt to display */
> 
> 	const unsigned char *data; /* PIN, set to NULL when using pin pad */
> 	size_t len;		/* set to 0 when using pin pad */
> 
> 	size_t min_length;	/* min length of PIN */
> 	size_t max_length;	/* max length of PIN */
> 
> 	unsigned int encoding;	/* ASCII-numeric, BCD, etc */
> 
> 	size_t pad_length;	/* PIN padding options, used with SC_PIN_CMD_NEED_PADDING */
> 	unsigned char pad_char;
> 
> 	size_t offset;		/* PIN offset in the APDU when using pin pad */
> 
> 	int max_tries;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
> 	int tries_left;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
> 	int logged_in;		/* Used for signaling back from SC_PIN_CMD_GET_INFO */
> };

Maybe I am missing something.
The sc_pin_cmd_pin is mapped to sc_pkcs15_auth_info here:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-pin.c#L686-L731
and later sc_pkcs15_auth_info is mapped to PKCS#11 info.
https://github.com/OpenSC/OpenSC/blob/master/src/pkcs11/framework-pkcs15.c#L583-L605
I do not see where of how I can access sc_pin_cmd_pin struct there? Or should I alter sc_pkcs15_auth_info?

frankmorgner:
`sc_pkcs15_auth_info.logged_in` is directly taken from `sc_pin_cmd_pin.pin1.logged_in`:

https://github.com/OpenSC/OpenSC/blob/6db171bcb6fd7cb3b51098fefbb3b28e44f0a79c/src/libopensc/pkcs15-pin.c#L725

Adding an additional state for `logged_in` in opensc.h would be enough to be passed through the PKCS#15 layer.

Please consider this as an suggestion; there may be other reasons why this suggestion may not be the best option.

The way you implemented it, reminded me of the elements `tries_left`, `max_tries`, `logged_in;` which all three were added in both `sc_pkcs15_auth_info` and `sc_pin_cmd_pin` to allow passing information up to the application.

metsma:
Maybe something like this? I really do not want rename sc_pin_cmd_pin logged_in, it is used all over place.

frankmorgner:
Almost... CKF_USER_PIN_COUNT_LOW, CKF_USER_PIN_FINAL_TRY, CKF_USER_PIN_LOCKED and CKF_USER_PIN_TO_BE_CHANGED seem to be mutially exclusive, which is why I suggest to simplify this a little:
https://github.com/frankmorgner/OpenSC/commit/4dc7e7798f19f68039463a91f1a7c4a32a7e5a92

(this should also fix the compilation error 😉)

metsma:
> CKF_USER_PIN_TO_BE_CHANGED
> 0x00080000
> True if the user PIN value is the default value set by token initialization or manufacturing, or the PIN has been expired by the card.

I do not think it is mutually exclusive. You can still login to token. In our use case just it does not let use key before it has changed least once.

frankmorgner:
I see. Then I think you should make it clear in the definition of either `logged_in` and/or the states themselves, that the values may be ORed together.

src/libopensc/opensc.h
```diff
@@ -433,10 +433,11 @@ typedef struct sc_reader {
 #define SC_PIN_ENCODING_BCD	1
 #define SC_PIN_ENCODING_GLP	2 /* Global Platform - Card Specification v2.0.1 */
 
-/** Values for sc_pin_cmd_pin.logged_in */
+/** Values for sc_pin_cmd_pin.logged_in, can be ored together */
```
Jakuje: I would rather describe it as a bitmap. The word "ored" looks odd ;)

metsma: Fixed

PR #3492, opened 2025-10-01
Remove autodools dependency from NMake build

metsma:
##### Checklist

- [ ] macOS token is tested


metsma:
I do not understand why tools print OPENSC_SCM_REVISION.
It is undefined in github actions builds because the checkout does fetch git gistory.

VERSION.mk
```diff
@@ -0,0 +1,17 @@
+PRODUCT_NAME=OpenSC
+PRODUCT_TARNAME=opensc
+PRODUCT_BUGREPORT=https://github.com/OpenSC/OpenSC/issues
+PRODUCT_URL=https://github.com/OpenSC/OpenSC
+PACKAGE_VERSION_MAJOR=0
+PACKAGE_VERSION_MINOR=26
+PACKAGE_VERSION_FIX=1
```
Jakuje, line 7: This changes the place where we need to update the version information. When we merge this, please update the wiki with the release checklist:

https://github.com/OpenSC/OpenSC/wiki/OpenSC-Release-Howto

metsma: I know, will do. I hope the filename is ok?
Plus side, it optimises greatly windows CI build times.

Jakuje: The downside of the .inc extension is that it is very generic and the clang-format tries to reformat it as a c file (which could be probably somehow excluded). One hint is to use `.clang-format-ignore`, but it looks like it does not work in the gh action: https://github.com/yshui/git-clang-format-lint/issues/5

But otherwise I do not have any other concerns.

metsma: renamed to VERSION.mk, makes more sense

Jakuje:
Thank you! Merging.

Issue #3495, opened 2025-10-02
pkcs11-tool is sending null digest to Openssl on FIPS enabled ubuntu 24.04


casual-lemon:
### Description 

On Ubuntu24.04 with FIPS enabled, openssl is segfaulting when using the pkcs11-tool -L command to list slots.

```
user1@ubuntu:~$ sudo pkcs11-tool -L
Segmentation fault
```

### Steps to reproduce
On Ubuntu vm (lxd or qemu) with fips enabled. 
```
sudo pro attach <token uuid>
sudo pro enable fips-updates
sudo apt-get install opensc 
```
Reboot vm

after logging in again, run the command
`sudo pkcs11-tool -L`
and we see the error Segmentation fault.

### Expected Output 
```
Available slots: /usr/bin/pkcs11-tool --module=/usr/lib/x86_64-linux-gnu/opensc-pkcs11.so -L
Slot 0 (0x0): Gemalto Gemplus USB SmartCard Reader 433-Swap [CCID Interface...
  token label        : John Doe
  token manufacturer : Common Access Card
  token model        : PKCS#15 emulated
  token flags        : login required, PIN pad present, rng, token initialized, PIN initialized
  hardware version   : 0.0
  firmware version   : 0.0
  serial num         : 000058bd002c19b5
  pin min/max        : 4/8
```
Seg fault bt

```
Thread 1 "pkcs11-tool" received signal SIGSEGV, Segmentation fault.
EVP_MD_up_ref (md=0x0) at ../crypto/evp/digest.c:1083
warning: 1083    ../crypto/evp/digest.c: No such file or directory
(gdb) bt
#0  EVP_MD_up_ref (md=0x0) at ../crypto/evp/digest.c:1083
```

### Environment

-OS: Ubuntu 24.04 with FIPS enabled
-Openssl version: user1@ubuntu:~$ openssl version
`OpenSSL 3.0.8 7 Feb 2023 (Library: OpenSSL 3.0.13 30 Jan 2024)`

casual-lemon:
Original issue opened with openssl: https://github.com/openssl/openssl/issues/28708

Jakuje:
Can you share the full backtrace where does the call come from? From this information, which just shows the last function that is called with NULL pointer, it is impossible to figure out what is going on there

casual-lemon:
[full-bt-output.txt](https://github.com/user-attachments/files/22746013/full-bt-output.txt)

Jakuje:
[...]

PR #3496, opened 2025-10-07
Do not register SHA1-based mechanisms in FIPS mode

Jakuje:
SHA1 is no longer accepted in FIPS mode.

Fixes: #3495 

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -424,7 +424,7 @@ jobs:
           distribution: 'semeru'
           java-version: '8'
           cache: 'maven'
-      - run: .github/setup-linux.sh piv
+      - run: .github/setup-linux.sh piv debug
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -388,7 +388,7 @@ jobs:
           version: apt-22-java
           execute_install_scripts: true
       - run: .github/setup-linux.sh
-      - run: .github/build.sh piv-sm dist valgrind
+      - run: .github/build.sh piv-sm dist
```
metsma: I had already somethin similar  queued https://github.com/OpenSC/OpenSC/pull/3497

Jakuje: Sorry for a confusion, this line was a mistake and should not be there. Removed.

PR #3497, opened 2025-10-08
Merge some linux actions

metsma:
##### Checklist

- [ ] macOS token is tested


Jakuje:
I think this is orthogonal to what I am trying to fix in the #3496.

.github/workflows/linux.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       matrix:
         dist: ['22.04', '24.04']
         arch: ['amd64', 'arm64']
-        build: ['dist', 'valgrind']
+        build: ['dist', 'valgrind', 'piv-sm']
```
Jakuje: I am not sure if this is exactly what we want -- we want to have the combination of valgrind and piv-sm builds to be able to test for leaks in the piv-sm builds (even though we do not have the piv-sm cards in CI, testing with PivApplet should give some idea).

metsma:
> This has 33 jobs, while the previous runs have 60. In the changeset it is not exactly clear where the jobs are disappearing, why and if there was some duplication or we are losing some test coverage.
If I compare these job I see that there is some additional jobs. Where did you get these numbers?
https://github.com/OpenSC/OpenSC/actions/runs/18338921252
https://github.com/OpenSC/OpenSC/actions/runs/18339381856

.github/workflows/linux.yml
metsma: seems like there is mismatch with different config
  build-piv-sm:
      - run: .github/setup-linux.sh
      - run: .github/build.sh piv-sm dist
vs
  build-ubuntu-22-piv-sm:
      - run: .github/setup-linux.sh
      - run: .github/build.sh piv-sm dist valgrind

metsma: Isnt the PivApplet already executed with regular valgrind task?

Jakuje:
I was just looking at the CI overview of the last PRs:
image
from #3496
vs.
image
from this PR

.github/workflows/linux.yml
Jakuje: i think these two things are different things. The piv-sm is the build configuration option, while PivApplet runs with the default build configuration.

I think we want both to catch possible issues in both build time configurations.

metsma:
> I was just looking at the CI overview of the last PRs: image from #3496 vs. image from this PR

I do not see any windows/mac builds, but I did not change any of their scripts and change did not affect their workloads.
https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/windows.yml#L4-L13

.github/workflows/linux.yml
metsma: Maybe then simplify matrix. I propose dist and piv-sm and with valgrind parameter

Jakuje:
> I do not see any windows/mac builds, but I did not change any of their scripts and change did not affect their workloads. https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/windows.yml#L4-L13

Oh, that might be relevant as I forgot the CI runs just the parts that are changing. So I will try to dig into that deeper

.github/workflows/linux.yml
metsma: Added condition that pig-sm adds valgrind parameter

Jakuje:
I do not see any more concerns so merging. Thank you!

PR #3501, opened 2025-10-16
fixed overwriting global module functions when loading PKCS#11 modules

### LINKED ISSUE #2875: C_Initialize() calls itself in a loop when pkcs-spy is used with SoftHSM from Homebrew (on macOS)

frankmorgner:
PKCS#11 modules are loaded into seperate namespaces, where they don't conflict with global symbols and don't overwrite global symbols. We only do this for seperating PKCS#11 modules, but not all all sorts of shared library, because the number of namespaces is limited to 16 in total and there are bugs around the glibc implementation.

fixes #2875

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -11,7 +11,7 @@ fi
 
 export SOPIN="12345678"
 export PIN="123456"
-PKCS11_TOOL="$BUILD_PATH/libtool --mode=execute gdb -ex run --args $BUILD_PATH/src/tools/pkcs11-tool"
+PKCS11_TOOL="$BUILD_PATH/libtool --mode=execute gdb -ex run -ex bt --args $BUILD_PATH/src/tools/pkcs11-tool"
 
 if [ "${TOKENTYPE}" == "softhsm" ]; then
     source "${BUILD_PATH}/tests/init-softhsm.sh"


frankmorgner:
CI still fails all tests pkcs11-tool, but I cannot reproduce this locally. This is one of the error report, which I do not undertand:

```
No tokentype provided, running with SoftHSM
Testing /usr/local/lib/softhsm/libsofthsm2.so
Testing /usr/lib/softhsm/libsofthsm2.so
Setting P11LIB=/usr/lib/softhsm/libsofthsm2.so
Setup softhsm
The token has been initialized and is reassigned to slot 1424859635
Generate key-pair with CKA_ALLOWED_MECHANISMS
./test-pkcs11-tool-allowed-mechanisms.sh: line 34: 34408 Segmentation fault         (core dumped) $PKCS11_TOOL --keypairgen --key-type="RSA:1024" --login --pin=$PIN --module="$P11LIB" --label="test" --id="$ID" --allowed-mechanisms="$MECHANISMS,SHA384-RSA-PKCS"
====> ERROR: Failed to Generate RSA key pair
./test-pkcs11-tool-allowed-mechanisms.sh: line 39: 34422 Segmentation fault         (core dumped) $PKCS11_TOOL --list-objects --login --pin=$PIN --module="$P11LIB" --id=$ID > objects.list
====> ERROR: Failed to list objects
====> ERROR: Allowed mechanisms not in the object list
====> ERROR: The RSA-PKCS,SHA1-RSA-PKCS,RSA-PKCS-PSS is not in the list
./test-pkcs11-tool-allowed-mechanisms.sh: line 49: 34438 Segmentation fault         (core dumped) $PKCS11_TOOL --id $ID -s -p $PIN -m SHA256-RSA-PKCS --module $P11LIB --input-file data --output-file data.sig &> sign.log
====> ERROR: It was possible to sign using non-allowed mechanism
Cleanup
FAIL test-pkcs11-tool-allowed-mechanisms.sh (exit status: 1)
```

src/common/libscdl.c
```diff
@@ -64,10 +64,22 @@ int sc_dlclose(void *handle)
...
 }
 #else
+#ifdef HAVE_DLMOPEN
+#define _GNU_SOURCE
+#endif
 #include <dlfcn.h>
 void *sc_dlopen(const char *filename)
 {
-	return dlopen(filename, RTLD_LAZY);
+	return dlmopen(LM_ID_NEWLM, filename, RTLD_LAZY | RTLD_LOCAL);
```
Jakuje: this should stay dlopen, should not it?
```suggestion
	return dlopen(LM_ID_NEWLM, filename, RTLD_LAZY | RTLD_LOCAL);
```

frankmorgner:
I rebased the dlmopen commit onto #3487, because the changes there will still be helpful, in particular when dlmopen is not available.

frankmorgner:
Here is one backtrace that I was able to generate locally:
```
Starting program: /home/fm/Dokumente/OpenSC/src/tools/.libs/pkcs11-tool --module=/usr/lib/softhsm/libsofthsm2.so --read-object --id 01 --type pubkey --output-file 01.der
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
get_and_lock (global=global@entry=0x17, class_index=class_index@entry=10)
    at ../crypto/ex_data.c:39
Warnung: 39	../crypto/ex_data.c: Datei oder Verzeichnis nicht gefunden
#0  get_and_lock (global=global@entry=0x17, class_index=class_index@entry=10)
    at ../crypto/ex_data.c:39
#1  0x00007ffff6e21a4a in ossl_crypto_new_ex_data_ex (ctx=0x0, class_index=10, 
    obj=0x7ffff742d690, ad=0x7ffff742d738) at ../crypto/ex_data.c:226
#2  0x00007ffff6dd0f1f in ENGINE_new () at ../crypto/engine/eng_lib.c:38
#3  0x00007ffff6dd0532 in engine_dynamic () at ../crypto/engine/eng_dyn.c:242
#4  engine_load_dynamic_int () at ../crypto/engine/eng_dyn.c:260
#5  0x00007ffff6e2256d in ossl_init_engine_dynamic () at ../crypto/init.c:301
#6  ossl_init_engine_dynamic_ossl_ () at ../crypto/init.c:298
#7  0x00007ffff64a1ed3 in __pthread_once_slow (
    once_control=0x7ffff7110394 <engine_dynamic>, 
    init_routine=0x7ffff6e22560 <ossl_init_engine_dynamic_ossl_>)
    at ./nptl/pthread_once.c:116
#8  0x00007ffff6e306ad in CRYPTO_THREAD_run_once (once=<optimized out>, 
    init=<optimized out>) at ../crypto/threads_pthread.c:154
#9  0x00007ffff6e22d88 in OPENSSL_init_crypto (opts=30208, 
    settings=<optimized out>) at ../crypto/init.c:618
#10 0x00007ffff6dd1f60 in ENGINE_by_id (id=0x7ffff79d2ff4 "rdrand")
    at ../crypto/engine/eng_list.c:381
#11 0x00007ffff7997654 in OSSLCryptoFactory::OSSLCryptoFactory (
    this=0x7ffff7429480) at OSSLCryptoFactory.cpp:149
#12 0x00007ffff7997735 in OSSLCryptoFactory::i () at OSSLCryptoFactory.cpp:284
#13 0x00007ffff79900b9 in CryptoFactory::i () at CryptoFactory.cpp:45
#14 0x00007ffff797772b in SoftHSM::C_Initialize (this=0x7ffff7429330, 
    pInitArgs=pInitArgs@entry=0x0) at SoftHSM.cpp:512
#15 0x00007ffff7943600 in C_Initialize (pInitArgs=0x0) at main.cpp:133
#16 0x0000555555561568 in main (argc=9, argv=0x7fffffffdf68)
    at pkcs11-tool.c:1339
```
The `global` data is non null, but obviously invalid (`0x17`) and cannot be dereferenced:
```
/*
 * Return the EX_CALLBACKS from the |ex_data| array that corresponds to
 * a given class.  On success, *holds the lock.*
 * The |global| parameter is assumed to be non null (checked by the caller).
 */
static EX_CALLBACKS *get_and_lock(OSSL_EX_DATA_GLOBAL *global, int class_index)
{
    EX_CALLBACKS *ip;

    if (class_index < 0 || class_index >= CRYPTO_EX_INDEX__COUNT) {
        ERR_raise(ERR_LIB_CRYPTO, ERR_R_PASSED_INVALID_ARGUMENT);
        return NULL;
    }

    if (global->ex_data_lock == NULL) {
        /*
         * If we get here, someone (who?) cleaned up the lock, so just
         * treat it as an error.
         */
         return NULL;
    }

    if (!CRYPTO_THREAD_write_lock(global->ex_data_lock))
        return NULL;
    ip = &global->ex_data[class_index];
    return ip;
}
```
It seems that `dlmopen()` leads to some inconsistencies inside of openssl, which is used in both, `pkcs11-tool` and the `dlmopen`ed SoftHSM...

Jakuje:
The OpenSSL 3.0 is using the dlopen to load providers (and engines). And from what I can see, it fails to load the engine correctly when the openssl is initialized inside of the dlopen namespace.
This particular error comes from loading rdrand, which is something we hit previously in the pkcs11-provider tests at various occasions both in Debian and in RHEL:

 *  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1104425
 * https://issues.redhat.com/browse/RHEL-107058

In the previous runs, I see that also the kryoptic tests were crashing so it really looks like its not the softhsm-only issue, but the way how openssl is (not)able to handle the dlmopen. I found similar issue in https://github.com/openssl/openssl/issues/26265#issuecomment-2564800667 which is suggesting that the atexit handlers are to blame.

That said, I think we will not be able to use the dlmopen safely, as long as the openssl registers the atexit handlers (should not be the case in Fedora anymore, but it still crashes there in the previous CI runs -- the current one is missing gdb).

frankmorgner:
Thanks for looking into this. Given the current situation, I'd like to proceed with #3487 only. If (or when) you see signs that OpenSSL can be used with `dlmopen()` as needed here, feel free to clone this pull request and reopen it.

Issue #1348, opened 2018-05-01
New Finish Identity Card not regonized


joni73:
### Problem Description

joni@mpi1:~$ pkcs15-tool -D
Using reader with a card: SCM Microsystems Inc. SCR 3310 [CCID Interface] (21120503106222) 00 00
PKCS#15 binding failed: Unsupported card

### Proposed Resolution
Must meet standards:

https://eevertti.vrk.fi/standardit

as well settings:

https://eevertti.vrk.fi/fineid-maaritykset

frankmorgner:
If you're planning to make a contribution to the project, we'd love to add a new card to OpenSC. We would be happy to get you going and to comment on your code. However, please have in mind that in lack of hardware, documentation and manpower, we can't do this on our own.

frankmorgner:
Closing this issue due to inactivity. Please re-open the ticket if more input is available.

PR #3505, opened 2025-10-27
Implement FinEID 4.0/4.1 support

metsma:
##### Checklist

- [ ] macOS token is tested


metsma:
Only ECC keys are currently loaded.

Related #1348

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -24,67 +24,108 @@
...
+		for (size_t j = 0; j < taglen; j++) {
+			if (!isalnum(tag[j]))
+				LOG_TEST_RET(card->ctx, SC_ERROR_INVALID_CARD, "Invalid character in document number");
+		}
+
+		free(p15card->tokeninfo->serial_number);
+		p15card->tokeninfo->serial_number = malloc(taglen + 1);
+		if (!p15card->tokeninfo->serial_number)
+			LOG_TEST_RET(card->ctx, SC_ERROR_OUT_OF_MEMORY, "Allocating memory failed");
+		p15card->tokeninfo->serial_number = memcpy(p15card->tokeninfo->serial_number, tag, taglen);
```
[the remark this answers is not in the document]
metsma: I do not understand why it ignores this line
```
		if (!p15card->tokeninfo->serial_number)
			LOG_TEST_RET(card->ctx, SC_ERROR_OUT_OF_MEMORY, "Allocating memory failed");
```

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -24,67 +24,108 @@
...
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
-	}
-	r = sc_read_binary(card, 0, buff, buflen, 0);
-	if (r < 0) {
-		free(buff);
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
-	}
+		r = sc_read_binary(p15card->card, 4, buf, buflen, 0);
+		if (r < 0)
+			LOG_TEST_RET(card->ctx, r, "Reading tokeninfo file failed");
```
frankmorgner:
```suggestion
		LOG_TEST_RET(card->ctx, r, "Reading tokeninfo file failed");
```

frankmorgner (approved):

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -182,12 +233,12 @@ sc_pkcs15emu_esteid2025_init(sc_pkcs15_card_t *p15card)
...
 				.id = {.len = 1, .value[0] = prkey_id[i]},
 				.native = 1,
-				.key_reference = prkey_ref[i],
+				.key_reference = prkey_ref[is_fin_eid(p15card)][i],
 				.field_length = field_length,
 				.usage = prkey_usage[i]
 		       };
 		struct sc_pkcs15_object prkey_obj = {
-				.auth_id = {.len = 1, .value[0] = prkey_pin[i]},
+				.auth_id = {.len = 1, .value[0] = pin_authid[i]},
```
Jakuje, line 241: just a nit that this mapping is really just mapping `i` -> `i + 1`. But given that it is used at least at two places, its probably ok to use it as it is.

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -24,67 +24,107 @@
...
-	}
-	r = sc_read_binary(card, 0, buff, buflen, 0);
-	if (r < 0) {
-		free(buff);
-		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_CARD);
-	}
+		r = sc_read_binary(p15card->card, 4, buf, buflen, 0);
+		LOG_TEST_RET(card->ctx, r, "Reading tokeninfo file failed");
+
+		tag = sc_asn1_find_tag(card->ctx, buf, buflen, 0x04, &taglen);
```
Jakuje: The `sc_read_binary()` returns the length read as a return value so using the `buflen` for bounds check in the `sc_asn1_find_tag()` could cause use of uninitialized memory.

src/libopensc/pkcs15-esteid2025.c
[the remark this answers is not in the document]
Jakuje: Could we use the `LOG_FUNC_RETURN()` instead to make it more direct for the analyzer that the return happens unconditionally? The message is really just a generic one and `SC_FUNC_RETURN()` converts it already to strerror which should print very similar message.

src/libopensc/pkcs15-esteid2025.c
```diff
@@ -24,67 +24,107 @@
...
+		buf = malloc(buflen + 1);
+		if (!buf)
+			LOG_TEST_RET(card->ctx, SC_ERROR_OUT_OF_MEMORY, "Allocating memory failed");
+		r = sc_read_binary(card, 0, buf, buflen, 0);
+		if (r < 0) {
+			free(buf);
+			LOG_TEST_GOTO_ERR(card->ctx, r, "Reading document number failed");
+		}
+
+		for (size_t j = 0; j < buflen; j++) {
```
Jakuje: this has the same assumptions that we read the full `buflen`. The card can return less data and then we will use uninitialized memory in `buf`, which is UB.
(I know this is just moved from the previous version, but I think its worth fixing once we touch it)

Jakuje:
Thank you! Merging.

PR #3506, opened 2025-10-28
openpgp: Fix retry counter of encryption PIN in SC_PIN_CMD_GET_INFO e…

MkfsSion:
…mulation for card supporting OpenPGP card spec v2.0 and later

The meaning of byte 6 of DO C4 has changed from the retry counter of the encryption PIN (CHV2) to that of resetting code since OpenPGP card spec v2.0, and PW1 now serves as both the signature and encryption PIN. Map PIN reference from 82 to 81 to correct retry counter of encryption PIN for these cards.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -2468,6 +2468,14 @@ pgp_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
 			LOG_TEST_RET(card->ctx, SC_ERROR_OBJECT_NOT_VALID,
 				"CHV status bytes have unexpected length");
 
+		/* The definition of fields of DO C4 changed between OpenPGP
+		 * card specification v1.1 and v2.0. There is no longer a separate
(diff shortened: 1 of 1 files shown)


frankmorgner:
Is your change explained with the general comment at the top of this block?
```
	 * The PIN references (P2 in APDU) for "VERIFY" are the same in both versions:
	 * 81 (CHV1 or PW1), 82 (CHV2 or PW1-mode 2), 83 (CHV3 or PW3),
	 * On the other hand from version 2.0 "CHANGE REFERENCE DATA" and
	 * "RESET RETRY COUNTER" don't support PW1-mode 2 (82) and need this
	 * value changed to PW1 (81).
```

MkfsSion:
> Is your change explained with the general comment at the top of this block?

I would rather leave a comment at the top of the block as it's only related to this emulation implementation and data layout of DO C4.

frankmorgner:
thank you

PR #3508, opened 2025-11-02
Restore valid Edwards and Montgomery curve lengths

mtrojnar:
##### PR Description

Commit 72125917695362de3306ebfeb06ce518b6fc42e6 updated curve lengths instead of only the internal storage lengths, violating [the PKCS#11 3.2 specification](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html):
###### 6.3.10 Edwards Elliptic Curve key pair generation
[...]
For this mechanism, the `ulMinKeySize` and `ulMaxKeySize` fields of the `CK_MECHANISM_INFO` structure specify the minimum and maximum supported number of bits in the field sizes, respectively. **For this mechanism, the only allowed values are 255 and 448 as [RFC 8032] only defines curves of these two sizes. A Cryptoki implementation may support one or both of these curves and should set the `ulMinKeySize` and `ulMaxKeySize` fields accordingly.**
###### 6.3.11 Montgomery Elliptic Curve key pair generation
[...]
For this mechanism, the `ulMinKeySize` and `ulMaxKeySize` fields of the `CK_MECHANISM_INFO` structure specify the minimum and maximum supported number of bits in the field sizes, respectively. **For this mechanism, the only allowed values are 255 and 448 as [RFC 7748] only defines curves of these two sizes. A Cryptoki implementation may support one or both of these curves and should set the `ulMinKeySize` and `ulMaxKeySize` fields accordingly.**

##### PR testing results

Without this PR, `pkcs11-tool` fails at least with the following modules:
 - libsofthsm2.so
 - libykcs11.so.2
 - opensc-pkcs11.so
    
The returned error was:
```
P:5669; T:0x139760863991616 09:18:11.899 [opensc-pkcs11] pkcs15-sec.c:221:format_senv: Card does not support EDDSA with field_size 256
```

This PR was successfully tested with SoftHSM2, YubiKey, and OpenSC modules.

##### Checklist
- [ ] macOS token is tested

### DIFF
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -2866,15 +2866,15 @@ pkcs15_create_public_key(struct sc_pkcs11_slot *slot, struct sc_profile *profile
 	}
 
 	if (key_type == CKK_EC_EDWARDS) {
-		if (ec->ecpointQ.len == BYTES4BITS(255))
+		if (ec->ecpointQ.len == BYTES4BITS(256)) /* note extra bit */
 			ec->params.id = oid_ED25519;
-		else if (ec->ecpointQ.len == BYTES4BITS(448))
+		else if (ec->ecpointQ.len == ED448_KEY_SIZE_BYTES)
 			ec->params.id = oid_ED448;
 		else
 			return CKR_ATTRIBUTE_VALUE_INVALID;
 
 	} else if (key_type == CKK_EC_MONTGOMERY) {
-		if (ec->ecpointQ.len == BYTES4BITS(255))
+		if (ec->ecpointQ.len == BYTES4BITS(256)) /* note extra bit */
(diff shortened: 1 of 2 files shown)


Jakuje:
@dengert can you have a look? We had quite a lengthy discussion around this over last year

dengert:
Should work, the issue was size in bytes need to hold the number of bits, and Ed448 needed 57 bytes.


The BYTES4BITS(255) returns 32 and happens to work but should be BTYES4BITS(256)
so the 25519 curves have been working but not as expected. But
BYTES4BITS(448) returns 56, but Ed448 needed an extra byte so will use
BYTES4BITS(456) to return 57 as required in the RFCs.

Jakuje:
No more comments. Thank you for catching this! Merging.

PR #3510, opened 2025-11-03
Import PKCS#11 3.2 header and add support for the new API and types in pkcs11-spy and pkcs11-tool

Jakuje:
Based on the work from @xhanulik which was started earlier this year.

I will need to do some more testing with PQC algorithms before it will be ready, but it should build now and be ready for initial testing if somebody is interested.

##### Checklist
- [ ] macOS token is tested


Jakuje:
@OpenSC/opensc-maintainers this should be ready for reviews. I would like to have this included in the next release 0.27

Tested locally against kryoptic with and without PQC mechanism with Fedora 43 (the build images are not yet updated so they fail -- should resolve by itself in couple of days).

The only failing job should be the Ubuntu 24 as there seems to be some broken images or something.

src/pkcs11/pkcs11-display.c
```diff
@@ -848,36 +1238,92 @@ type_spec ck_attribute_specs[] = {
   { CKA_REQUIRED_CMS_ATTRIBUTES, "CKA_REQUIRED_CMS_ATTRIBUTES ", print_generic, NULL },
   { CKA_DEFAULT_CMS_ATTRIBUTES, "CKA_DEFAULT_CMS_ATTRIBUTES ", print_generic, NULL },
   { CKA_SUPPORTED_CMS_ATTRIBUTES, "CKA_SUPPORTED_CMS_ATTRIBUTES ", print_generic, NULL },
+
+  // NEW
```
frankmorgner:
```suggestion
  // NEW in PKCS#11 v3.2
```

src/pkcs11/pkcs11-spy.c
```diff
@@ -178,17 +182,37 @@ allocate_function_list(int v3)
 	list_3_0->C_VerifyMessageNext = C_VerifyMessageNext;
 	list_3_0->C_MessageVerifyFinal = C_MessageVerifyFinal;
 
-	return list_3_0;
+	if (v < 32) {
```
frankmorgner, line 185: could this lead to semantical problems, when we get version 4 (with 4<32)?

src/pkcs11/pkcs11.h
frankmorgner, line 1: > This file is a modified implementation of the PKCS #11 standard

In your commit, you simply removed the copyright notice without explaining the change. I don't mind copying pkcs11.h from OASIS if their version is published with a permissible license. Still, that very license needs to be mentioned/written down somewhere.

src/tests/p11test/p11test_case_common.c
```diff
@@ -187,9 +191,13 @@ add_supported_mechs(test_cert_t *o)
 			o->mechs[0].result_flags = 0;
 			o->mechs[0].usage_flags = CKF_SIGN | CKF_VERIFY;
 		}
+		break;
 #ifdef EVP_PKEY_ED25519
-	} else if (o->type == EVP_PKEY_ED25519) {
-		if (token.num_ed_mechs > 0 ) {
+	case EVP_PKEY_ED25519:
```
frankmorgner, line 196: This could be more correct regarding the different configurations of EVP_PKEY_ED25519 and EVP_PKEY_ED448:
```
#ifdef EVP_PKEY_ED25519
	case EVP_PKEY_ED25519:
#endif
#ifdef EVP_PKEY_ED448
	case EVP_PKEY_ED448:
#endif
#if defined(EVP_PKEY_ED25519) || defined(EVP_PKEY_ED448)
		if (token.num_ed_mechs > 0) {
			o->num_mechs = token.num_ed_mechs;
			for (i = 0; i < token.num_ed_mechs; i++) {
				o->mechs[i].mech = token.ed_mechs[i].mech;
			o->mechs[0].result_flags = 0;
			o->mechs[0].usage_flags = CKF_SIGN | CKF_VERIFY;
		}
		break;
#endif /* defined(EVP_PKEY_ED25519) || defined(EVP_PKEY_ED448) */
```

The same can be applied when handling EVP_PKEY_X25519 and EVP_PKEY_X448

src/tests/p11test/p11test_case_common.c
```diff
@@ -754,12 +936,161 @@ int callback_public_keys(test_certs_t *objects,
...
+				debug_print(" [WARN %s ] Got different key type than certificate "
+					    "(got %d, expected %d)",
+						o->id_str, base_id, o->type);
+			}
+
+			if ((EVP_PKEY_get_octet_string_param(o->key, "pub", NULL, 0, &cert_pubkey_len) != 1)) {
+				fprintf(stderr, "Failed to extract public key length");
+				return -1;
+			}
+			if ((cert_pubkey = malloc(cert_pubkey_len)) == NULL) {
```
frankmorgner, line 1055: `cert_pubkey` is not free'd anywhere, is it?

src/tests/p11test/p11test_case_readonly.c
```diff
@@ -576,8 +578,14 @@ int verify_message_openssl(test_cert_t *o, token_info_t *info, CK_BYTE *message,
...
 			return -1;
 		}
+		break;
+	}
 #ifdef EVP_PKEY_ED25519
-	} else if (o->type == EVP_PKEY_ED25519) {
+	case EVP_PKEY_ED25519:
+#ifdef EVP_PKEY_ED448
+	case EVP_PKEY_ED448:
+#endif
```
frankmorgner, line 588: See above comment on handling EVP_PKEY_ED25519 and EVP_PKEY_ED448

src/tests/p11test/p11test_case_readonly.c
```diff
@@ -806,26 +867,27 @@ void readonly_tests(void **state) {
 		test_cert_t *o = &objects.data[i];
 
 		if (o->key_type != CKK_RSA &&
-		    o->key_type != CKK_EC &&
-		    o->key_type != CKK_EC_EDWARDS &&
-		    o->key_type != CKK_EC_MONTGOMERY)
+				o->key_type != CKK_EC &&
+				o->key_type != CKK_EC_EDWARDS &&
+				o->key_type != CKK_ML_DSA &&
+				o->key_type != CKK_SLH_DSA)
```
frankmorgner, line 875: did you miss/forget CKK_EC_MONTGOMERY?

src/tools/pkcs11-tool.c
```diff
@@ -2589,8 +2608,18 @@ static void sign_data(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
 	mech.mechanism = opt_mechanism;
 	hashlen = parse_pss_params(session, key, &mech, &pss_params);
 
-	/* support pure EdDSA only */
-	if (opt_mechanism == CKM_EDDSA) {
+	if (opt_input == NULL)
+		fd = 0;
+	else if ((fd = open(opt_input, O_RDONLY | O_BINARY)) < 0)
```
frankmorgner, line 2613: fd is not closed, is it?

src/tools/pkcs11-tool.c
```diff
@@ -3566,8 +3690,133 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
 			}
-		}
-		else {
+		} else if (strncasecmp(type, "ML-DSA-", strlen("ML-DSA-")) == 0) {
+			CK_MECHANISM_TYPE mtypes[] = {CKM_ML_DSA_KEY_PAIR_GEN};
+			size_t mtypes_num = sizeof(mtypes) / sizeof(mtypes[0]);
+
+			key_type = CKK_ML_DSA;
+
+			if (strlen(type) != strlen("ML-DSA-") + 2) {
```
frankmorgner: You are actually catching this also in the else branch below.

src/tools/pkcs11-tool.c
```diff
@@ -3566,8 +3690,133 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+				n_privkey_attr++;
+			}
+
+		} else if (strncasecmp(type, "ML-KEM-", strlen("ML-KEM-")) == 0) {
+			CK_MECHANISM_TYPE mtypes[] = {CKM_ML_KEM_KEY_PAIR_GEN};
+			size_t mtypes_num = sizeof(mtypes) / sizeof(mtypes[0]);
+
+			key_type = CKK_ML_KEM;
+
+			if (strlen(type) != strlen("ML-KEM-") + 3 && strlen(type) != strlen("ML-KEM-") + 4) {
```
frankmorgner: already catched in the else branch below

src/tools/pkcs11-tool.c
```diff
@@ -3566,8 +3690,133 @@ static int gen_keypair(CK_SLOT_ID slot, CK_SESSION_HANDLE session,
...
+				n_privkey_attr++;
+			}
+
+		} else if (strncasecmp(type, "SLH-DSA-", strlen("SLH-DSA-")) == 0) {
+			CK_MECHANISM_TYPE mtypes[] = {CKM_SLH_DSA_KEY_PAIR_GEN};
+			size_t mtypes_num = sizeof(mtypes) / sizeof(mtypes[0]);
+
+			key_type = CKK_SLH_DSA;
+
+			if (strlen(type) != strlen("SLH-DSA-SHA2-NNNN") && strlen(type) != strlen("SLH-DSA-SHAKE-NNNN") + 7) {
```
frankmorgner: already catched in the else branch below

src/tools/pkcs11-tool.c
```diff
@@ -3946,6 +4195,7 @@ unwrap_key(CK_SESSION_HANDLE session)
 	} else {
 		util_fatal("Unsupported key type %s", opt_key_type);
 	}
+	/* PQC: Add*/
```
frankmorgner: TODO?

src/tools/pkcs11-tool.c
```diff
@@ -4442,25 +4692,140 @@ parse_ec_pkey(EVP_PKEY *pkey, int private, struct gostkey_info *gost)
...
+		ec->header_len = header_len;
 #endif
 	}
 
 	return 0;
 }
 
+#if OPENSSL_VERSION_NUMBER >= 0x30500000L
+static int
+parse_pqc_pkey(EVP_PKEY *pkey, CK_KEY_TYPE type, int private, struct pqckey_info *pqc)
```
frankmorgner, line 4710: not sure if this is the best way of grouping.

maybe separating lattice crypto from hash based crypto could be useful.

src/pkcs11/pkcs11-spy.c
Jakuje: Indeed. But given this is just an internal variable, I do not think its a big issue. But I can change the 3 to 30 easily to make it more future proof.

src/pkcs11/pkcs11.h
Jakuje: The problem is the OASIS PKCS#11 header is not clearly open-source compatible license, as you can see in:

https://github.com/oasis-tcs/pkcs11/blob/master/published/3-02/pkcs11.h

But maintentance of compatible header file based on some ancient project in opensc is very painful and error prone.

The commit "Update PKCS#11 header file to v3.2" already mentions the source of the headers is coming from:

https://github.com/latchset/pkcs11-headers/

But I agree that alread the first commit should say this. Will update.

src/tests/p11test/p11test_case_common.c
Jakuje: Correct. This was written with the assumption that the implementation of `EVP_PKEY_ED448`  will have also the `EVP_PKEY_ED25519`, which is true for both older openssl versions as well as libressl. I just wanted to avoid too many ifdefs. But I can change this.

src/tests/p11test/p11test_case_readonly.c
Jakuje: no. these are just the signature algorithms (regardless of the confusing name "readonly". The mongomery curves are usable only with the derive operation.

src/tools/pkcs11-tool.c
Jakuje: well. Technically none of the cases where we call the `util_fatal()` between this line and the line calling the `close(fd);` do close the FD. But the OS should take care of that on the error condition. In function code, I could consider this an issue, but I think there were several pre-existing case like this before so I would not consider it an issue in test program like pkcs11-tool.

src/tools/pkcs11-tool.c
Jakuje: Good point! Will remove these.

src/tools/pkcs11-tool.c
Jakuje: Same as for ML-DSA. Will remove

src/tools/pkcs11-tool.c
Jakuje: same as previous. Will remove

src/tools/pkcs11-tool.c
Jakuje: True. Will reword to make it more clear. I think we do not support wrapping/unwrapping in kryoptic yet so there is no way to test this for now.

src/tools/pkcs11-tool.c
Jakuje: The PQC algorithms (at least the NIST ones that I implement here) have very the same syntax in the PKCS11 specification (except for the SEED attribute). I started writing them separately, but I wanted to avoid the code duplication and they ended up the same. I would prefer to keep them together and if we will need some other algorithm, split that later.

Jakuje:
Rebased on top of #3515.

src/tests/p11test/p11test_case_common.c
Jakuje: Thanks! You are right. Will fix this!

src/pkcs11/pkcs11-display.c
Jakuje: Fixed. Thanks!

src/pkcs11/pkcs11.h
frankmorgner: https://github.com/latchset/pkcs11-headers/blob/main/public-domain/3.2/pkcs11.h could be included in OpenSC as standalone file with its own copyright notice (e.g. as pkcs11_ls_32.h). This extra file can be included from our pkcs11.h and it could be updated more easily in the future. Also, we are free to keep the license notice in our file.

Anyway, I'm open for suggestions.

Jakuje: True. I think keeping the unmodified file from pkcs11-headers would be the cleanest way as it will make it easy to update it.

Jakuje: Technically, the copyright in pkcs11.h was from scute, but after moving the public domain pkcs11.h to separate file there is nothing left from that (except for a few macros that I can not really attribute to scute) so I restored there the opensc copyright header that was there before bringing in the scute.

src/pkcs11/pkcs11.h
```diff
@@ -1,54 +1,22 @@
...
-   <stdbool.h>.
-
-   If CRYPTOKI_COMPAT is defined before including this header file,
-   then none of the API changes above take place, and the API is the
-   one defined by the PKCS #11 standard.  */
+/*
+ * pkcs11.h: OpenSC project's PKCS#11 header
+ *
+ * Copyright (C) 2002  Timo Teräs <timo.teras@iki.fi>
+ *
```
frankmorgner:
```suggestion
```

src/pkcs11/pd-pkcs11.h
```diff
@@ -0,0 +1,2773 @@
+/* This file is in the Public Domain */
+// clang-format off
```
frankmorgner:
```suggestion
/* https://github.com/latchset/pkcs11-headers/blob/main/public-domain/3.2/pkcs11.h */
// clang-format off

/* This file is in the Public Domain */
```

Jakuje: Adjusted. Thanks!

src/pkcs11/pkcs11.h
Jakuje: Removed. Thanks!

frankmorgner (approved):

Jakuje:
Thank you for the review! I will merge this one and open follow-up for the kryoptic tests to get fixed.

PR #3514, opened 2025-11-12
Fix windows unblock pin bug

swissbit-dnutz:
Changed the management key algorithm to select the correct one in the function PIV get challenge.

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -4519,7 +4519,7 @@ static int piv_get_challenge(sc_card_t *card, u8 *rnd, size_t len)
 	}
 
 	/* NIST 800-73-3 says use 9B, previous versions used 00 */
-	r = piv_general_io(card, 0x87, 0x00, 0x9B, sbuf, sizeof sbuf, rbuf, sizeof rbuf);
+	r = piv_general_io(card, 0x87, priv->mgmt_key_alg, 0x9B, sbuf, sizeof sbuf, rbuf, sizeof rbuf);
 	/*
 	 * piv_get_challenge is called in a loop.
 	 * some cards may allow 1 challenge expecting it to be part of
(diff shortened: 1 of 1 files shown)


dengert:
This is more than just a  "Fix windows unblock pin bug". Hits 41 files!
 @frankmorgner and @Jakuje please have a look at this some low level routines have changes that will affect other cards.  Especially apdu.c and ctx.c 

Had a quick look at this.

In regards to PIV changes:
- Have you built this with the configure option  --"enable_piv_sm"? 
- Can you send the response to the SELECT AID?  I would like to see the AC entries. This is a problem area. 
- You are not the only vendor wanting to make additional changes to the PIV. 
- Have you tested the changes with real PIV cards or Yubikey cards to see if they still work?
- Have you looked at PR #3098 
- On Windows, the minidriver is not configured for PIV. This is done as all other vendors provide their own minidrivers. The PIV driver can be configured if needed. 
- What is difference between SC_CARD_TYPE_PIV_II_SWISSBIT  and SC_CARD_TYPE_PIV_II_SWISSBIT2?
- Is any documentation available?
- Can you send me a pre initialized demo card?

swissbit-dnutz:
> This is more than just a "Fix windows unblock pin bug". Hits 41 files! @frankmorgner and @Jakuje please have a look at this some low level routines have changes that will affect other cards. Especially apdu.c and ctx.c
> 
> Had a quick look at this.
> 
> In regards to PIV changes:
> 
>     * Have you built this with the configure option  --"enable_piv_sm"?
> 
>     * Can you send the response to the SELECT AID?  I would like to see the AC entries. This is a problem area.
> 
>     * You are not the only vendor wanting to make additional changes to the PIV.
> 
>     * Have you tested the changes with real PIV cards or Yubikey cards to see if they still work?
> 
>     * Have you looked at PR [NIST Secure Messaging moved from PIV to separate sm/sm-nist.c  #3098](https://github.com/OpenSC/OpenSC/pull/3098)
> 
>     * On Windows, the minidriver is not configured for PIV. This is done as all other vendors provide their own minidrivers. The PIV driver can be configured if needed.
> 
>     * What is difference between SC_CARD_TYPE_PIV_II_SWISSBIT  and SC_CARD_TYPE_PIV_II_SWISSBIT2?
> 
>     * Is any documentation available?
> 
>     * Can you send me a pre initialized demo card?

You're right, there are more changes. I accidentally opened the pull request with the wrong destination branch. That's the reason why I closed it right afterwards.
I apologise for any inconvenience caused.

PR #3521, opened 2025-11-21
Add Latvian Cosmo 8.2 card support

metsma:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-esteid2018.c
+++ b/src/libopensc/pkcs15-esteid2018.c
@@ -43,8 +43,8 @@ is_latvian_eid(sc_pkcs15_card_t *p15card)
 static int sc_pkcs15emu_esteid2018_init(sc_pkcs15_card_t *p15card) {
 	sc_card_t *card = p15card->card;
 	u8 buff[11];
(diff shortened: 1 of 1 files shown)


Jakuje (approved):

src/libopensc/cards.h
```diff
@@ -231,6 +231,7 @@ enum {
 
 	/* EstEID cards */
 	SC_CARD_TYPE_ESTEID_2018 = 35000,
+	SC_CARD_TYPE_LATEID_2018,
```
frankmorgner: This should go to the end of the list, shouldn't it? This way, the identifiers for the other types stay as they were before this change.

Jakuje: is it important given that we do not support direct linking to libopensc?

frankmorgner: not important, but possible. consider it as an optional suggestion.

metsma: Fixed

Jakuje:
Thank you! Merging!

PR #3523, opened 2025-11-22
CI: added upload to Signpath.io

frankmorgner:
##### Checklist

- [ ] macOS token is tested


frankmorgner:
I adapted the Artifact Configurations in Signpath.io to Github's format and successfully verified that artifacts can be submitted (https://github.com/OpenSC/OpenSC/actions/runs/20799292001) and signed (via manual authorization in the Signpath portal).

frankmorgner:
Sigh, it seems that `if: ${{ github.repository == 'OpenSC/OpenSC' }}` is not enough to only issue the signing request when we're on OpenSC/OpenSC...

metsma:
> Sigh, it seems that `if: ${{ github.repository == 'OpenSC/OpenSC' }}` is not enough to only issue the signing request when we're on OpenSC/OpenSC...

Maybe you need to re-enter the API token?

frankmorgner:
The link I did post above came from a dedicated branch in OpenSC/OpenSC (https://github.com/OpenSC/OpenSC/actions/runs/20799292001). What doesn't work is *this* pull request here, because the repository's secrets aren't available. For one, that's expected and I'd like to avoid release signing for pull requests (hence the limitation of the scope to OpenSC/OpenSC).

Ideally, I'd like to have the release signing only for commits that have a (pre-) release tag on it so that we don't flood Signpath (or Apple's notarization) with too many requests... Unfortunately, I do not currently have enough knowlegdge on how to code this into the workflow files.

metsma:
> The link I did post above came from a dedicated branch in OpenSC/OpenSC (https://github.com/OpenSC/OpenSC/actions/runs/20799292001). What doesn't work is _this_ pull request here, because the repository's secrets aren't available. For one, that's expected and I'd like to avoid release signing for pull requests (hence the limitation of the scope to OpenSC/OpenSC).
> 
> Ideally, I'd like to have the release signing only for commits that have a (pre-) release tag on it so that we don't flood Signpath (or Apple's notarization) with too many requests... Unfortunately, I do not currently have enough knowlegdge on how to code this into the workflow files.

Maybe something like this:

     if: ${{ startsWith(github.ref, 'refs/tags/') && github.repository == 'OpenSC/OpenSC' }}

frankmorgner:
I have added a new commit with your suggestion so that external verification should now only be done for (release) tags in the OpenSC repository.

frankmorgner:
I checked with a dedicated branch in the OpenSC repository, which works as expected for both, tagged commits and without tags. https://github.com/OpenSC/OpenSC/actions/runs/21357098586

frankmorgner:
I removed the AppVeyor configuration for release signing and disabled the AppVeyor integration in Signpath. I think we are ready now

.github/workflows/windows.yml
metsma:
```suggestion
        uses: actions/cache@v5
```

metsma:
```suggestion
        uses: actions/cache@v5
```

.github/workflows/macos.yml
metsma:
```suggestion
      - uses: actions/cache@v5
```

metsma:
```suggestion
        uses: actions/upload-artifact@v6
```

metsma:
```suggestion
        uses: actions/download-artifact@v7
```

metsma:
```suggestion
        uses: actions/checkout@v6
```

metsma:
```suggestion
      - uses: actions/checkout@v6
```

metsma:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/windows.yml
metsma:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/macos.yml
metsma:
```suggestion
        uses: actions/upload-artifact@v6
```

metsma:
```suggestion
        uses: actions/upload-artifact@v6
```

.github/workflows/windows.yml
```diff
@@ -115,10 +115,21 @@ jobs:
       - name: Build OpenSC
         run: nmake /nologo /f Makefile.mak opensc.msi
       - name: Archive artifacts
+        id: upload-unsigned-artifact
         uses: actions/upload-artifact@v5
```
metsma:
```suggestion
        uses: actions/upload-artifact@v6
```

.github/workflows/windows.yml
```diff
@@ -115,10 +115,21 @@ jobs:
...
+        uses: signpath/github-action-submit-signing-request@v2
+        with:
+          api-token: '${{ secrets.SIGNPATH_API_TOKEN }}'
+          organization-id: '8d2463fe-39bd-4a41-bb72-f008b4b1fe17'
+          project-slug: 'OpenSC'
+          signing-policy-slug: 'release-signing'
+          github-artifact-id: '${{ steps.upload-unsigned-artifact.outputs.artifact-id }}'
+          wait-for-completion: false
       - name: Archive debug artifacts
         uses: actions/upload-artifact@v5
```
metsma:
```suggestion
        uses: actions/upload-artifact@v6
```

.github/workflows/macos.yml
metsma:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/macos.yml
```diff
@@ -46,23 +46,30 @@ jobs:
...
           DEV_ID_APPLICATION: ${{ secrets.DEV_ID_APPLICATION }}
       - run: .github/build.sh
+        if: ${{   startsWith(github.ref, 'refs/tags/') && github.repository == 'OpenSC/OpenSC' }}
         env:
           CODE_SIGN_IDENTITY: ${{ secrets.CODE_SIGN_IDENTITY }}
           DEVELOPMENT_TEAM: ${{ secrets.DEVELOPMENT_TEAM }}
           INSTALLER_SIGN_IDENTITY: ${{ secrets.INSTALLER_SIGN_IDENTITY }}
           NOTARIZATION_PASSWORD: ${{ secrets.NOTARIZATION_PASSWORD }}
           APPLE_ID: ${{ secrets.APPLE_ID }}
+      - run: .github/build.sh
```
metsma: I wonder if it better to separate the notarization from regular build script, to avoid this duplication?https://github.com/OpenSC/OpenSC/blob/master/MacOSX/build#L193-L196

metsma:
The fixpath branch macOS build looks good.
Is there way to download signed windows artifacts?

frankmorgner:
With a test branch and tag on OpenSC i successfully verified that signed (and notarized) binaries are created on releases, but not without a (release) tag. The signed Windows installer needs to be downloaded from Signpath.io. We don't currently have configured any automation for this, because currently signing requires manual approval via the Signpath dashboard.

Jakuje:
Than you for the work on this! So if I understand it right, we have #3536 which signs now using @metsma certificate and is uploaded to nightly repo and is supposed to be the test builds and this change, which pushes the artifacts to signpath, needs manual confirmation and download of the artifacts and is the place where we should take the final release artifacts?

I think this needs to be reflected in the release page once merged:

https://github.com/OpenSC/OpenSC/wiki/OpenSC-Release-Howto

metsma:
> So if I understand it right, we have https://github.com/OpenSC/OpenSC/pull/3536 which signs now using @metsma certificate and is uploaded to nightly repo

And if it is built from tag it is also notarized.
Is the nightly repo needed used ? Every github action has is own artifact attached.

Jakuje:
>  Is the nightly repo needed used ? Every github action has is own artifact attached.

That is a good point. I think the Nightly repo was created before we had GH actions at all and mostly to collect artifacts from different CI systems. I think the downside of artifacts from GH Action directly is that they expire at some point, but it can be configured and I do not think we really need one year old nightlies.

So if we do not have a good use for that I would probably stop using the Nightly.

frankmorgner:
Yes correct, the Nightly repository contains all build artifacts that were generated in OpenSC/OpenSC (master, branches, tags, no PR). We used this to combine build artifacts from different sources (AppVeyor, GH Actions). Signed Windows artifacts can only be downloaded manually via Signpath, but the signed and notarized binaries from the OSX build will be available through both, GH artifacts and Nightly. In my opinion, the Nightly repository is still useful as it is independant from the availability of GH action artifacts. The drawback is that Nightly needs to be cleaned up manually from time to time.

PR #3524, opened 2025-11-25
Fix external tests

Jakuje:
This is reworking external tests to make their output reasonable.

Also fixes outstanding coverity issue introduced  from #3510.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tests/p11test/isoapplet_ref_v0.json
+++ b/src/tests/p11test/isoapplet_ref_v0.json
@@ -269,6 +269,8 @@
 		"DECRYPT",
 		"WRAP",
 		"UNWRAP",
+		"ENCAPSULATE",
+		"DECAPSULATE",
 		"DERIVE PUBLIC",
 		"DERIVE PRIVATE",
 		"ALWAYS AUTH"
@@ -287,6 +289,8 @@
 		"YES",
 		"",
 		"",
+		"",
+		"",
 		""
 	],
 	[
@@ -303,6 +307,8 @@
 		"YES",
 		"",
 		"",
+		"",
+		"",
 		""
 	],
 	[
@@ -313,6 +319,8 @@
 		"",
 		"YES",
 		"YES",
+		"",
+		"",
 		"",
 		"",
 		"",
--- a/src/tests/p11test/isoapplet_ref_v1.json
+++ b/src/tests/p11test/isoapplet_ref_v1.json
@@ -359,6 +359,8 @@
 		"DECRYPT",
 		"WRAP",
 		"UNWRAP",
+		"ENCAPSULATE",
+		"DECAPSULATE",
 		"DERIVE PUBLIC",
 		"DERIVE PRIVATE",
 		"ALWAYS AUTH"
@@ -377,6 +379,8 @@
 		"YES",
 		"",
 		"",
+		"",
+		"",
 		""
 	],
 	[
@@ -393,6 +397,8 @@
 		"YES",
 		"",
 		"",
+		"",
+		"",
 		""
 	],
 	[
@@ -403,6 +409,8 @@
 		"",
 		"YES",
 		"YES",
+		"",
+		"",
 		"",
 		"",
 		"",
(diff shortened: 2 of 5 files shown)


Jakuje:
Sigh. We still can not get away from building kryoptic as the github fedora image is still on the f42.

Jakuje:
It took me much more effort than I was hoping for, but the external tests should now run as expected including the kryoptic and interoperability tests.

src/pkcs11/Makefile.mak
frankmorgner, line 1: If the idea was to trigger [the rule](https://github.com/OpenSC/OpenSC/blob/220bc82e08e71d720d742ce2e8713dc9af14dade/win32/Make.rules.mak#L167) for `.exports.def`, then I think this needs to be named `pkcs11.exports.def`

Jakuje: Yes. See the discussion in 11a7777bbf65694d51e310c6f39ba24ad0192da0. Looking at the build log, it looks like the pkcs11.def is generated, but ignored as we need also the OUT to match the object name:
```
opensc-pkcs11.exp : warning LNK4070: /OUT:pkcs11.dll directive in .EXP differs from output filename 'opensc-pkcs11.dll'; ignoring directive
```
so this part will need some more love.

Jakuje:
Ugh. This sounds like finally solved the CI issues on build pipeline. Now I just need to do the same on the tests pipeline to have all green.

src/pkcs11/Makefile.mak
Jakuje: @jurajsarinay @metsma can you have a look of this change is fixing the issue introduced in 11a7777bbf65694d51e310c6f39ba24ad0192da0 or it needs some adjustments? I do not have any windows machine around.

metsma: pkcs11-tool lists objects with github x86/x64/arm64 builds

jurajsarinay: I have compiled the branch on a 64-bit Windows 11 system using VS 2022. The DLL exports what it should and `pkcs11-tool.exe` is happy.

Jakuje: Thank you for testing both of you! The only issue is now the warning above, which likely does not have any functional drawback, but solving it would likely require to have two separate exports for two separate libraries or doing some other custom `.def` file generation so unless there are some strong opinions against, I would consider this finalized.

Jakuje:
This grew a bit from the initial attempt to fix simple small thing, but should make the CI happy (If the mirrors will cooperate).

Jakuje:
Thank you for the review! I will merge this so we can get some CI results in the remaining PRs. If there would be some further issues, please let me know.

Jakuje:
I have some idea how to tackle the random apt update issues, but its for separate PR.

src/pkcs11/Makefile.mak
jurajsarinay: If the name of the library is dropped from the first line of the `.def` file, the warnings disappear. The use of `\OUT` [is sufficient](https://learn.microsoft.com/en-us/cpp/build/reference/library?view=msvc-170). Because of that, the following change would not impact any of the DLLs we build:

```diff
diff --git a/win32/Make.rules.mak b/win32/Make.rules.mak
index 820560d73..dd2871cdb 100644
--- a/win32/Make.rules.mak
+++ b/win32/Make.rules.mak
@@ -165,7 +165,7 @@ COPTS = /O1 /$(BUILD_TYPE) $(COPTS)
        rc $(DEFS) /l 0x0409 /I$(TOPDIR)\win32 $<

 .exports.def:
-       echo LIBRARY $* > $*.def
+       echo LIBRARY > $*.def
        echo EXPORTS >> $*.def
        type $*.exports >> $*.def

```

Jakuje: Thanks for looking in that! Can you drop here a simple PR with this change?

Jakuje:
> I have some idea how to tackle the random apt update issues, but its for separate PR.

Seems like using debuginfod did the trick: #3527 No failed valgrinds anymore

PR #3527, opened 2025-12-01
ci: Try to use debuginfod instead of manual download of ddebs

Jakuje:
... which is fragile given that ubuntu does not have predictable naming pattern for the debuginfo packages and the repositories are most of the time being updated which results in random failures in CI installing these packages.

##### Checklist

- [ ] macOS token is tested


metsma (approved):

Jakuje:
Thank you for the review!

PR #3531, opened 2025-12-09
pkcs15init: Avoid buffer overrun

Jakuje:
The maximum path length is 16 bytes and if we want to extend the path, we need to make sure its not too long.

Thanks oss-fuzz.

https://issues.oss-fuzz.com/issues/467161860

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/pkcs15init/pkcs15-cflex.c
+++ b/src/pkcs15init/pkcs15-cflex.c
@@ -539,6 +539,9 @@ cflex_create_pin_file(sc_profile_t *profile, sc_pkcs15_card_t *p15card,
 
 	/* Build the CHV path */
 	path = *df_path;
+	if (path.len > SC_MAX_PATH_SIZE - 2) {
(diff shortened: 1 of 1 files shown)


src/pkcs15init/pkcs15-cflex.c
```diff
@@ -539,6 +539,9 @@ cflex_create_pin_file(sc_profile_t *profile, sc_pkcs15_card_t *p15card,
 
 	/* Build the CHV path */
 	path = *df_path;
+	if (path.len + 2 > SC_MAX_PATH_SIZE) {
```
frankmorgner:
```suggestion
	if (path.len > SC_MAX_PATH_SIZE - 2) {
```

Jakuje: Well, given that the `SC_MAX_PATH_SIZE` is 16, which means the path we get can not simply overrun (unless we would get already broken path object), but I can change that.

Issue #3534, opened 2025-12-13
pkcs11-tool --help  out of alignment


dengert:
looks like  --usage-encapsulate  is missing its text, has the text for -w. 
``` 
     --usage-derive            Specify 'derive' key usage flag (EC only)
      --usage-wrap              Specify 'wrap' key usage flag
      --usage-encapsulate       Write an object (key, cert, data) to the card
  -w, --write-object <arg>      Get object's CKA_VALUE attribute (use with --type)
  -r, --read-object             Delete an object (use with --type cert/data/privkey/pubkey/secrkey)
  -b, --delete-object           Specify the application label of the data object (use with --type data)
```

Jakuje:
Oh, we messed this up again! Good catch! Will fix and try to add some sanity check to make sure we avoid this in the future.

PR #3535, opened 2025-12-15
Fix pkcs11-tool options and add guardrails to prevent this common issue next time

Jakuje:
Fixes: #3534

##### Checklist

- [~] macOS token is tested


frankmorgner:
If you want to get fancy, this could also be done by a [static assertion](https://en.cppreference.com/w/c/language/_Static_assert.html), which would be evaluated and checked at compile time. This would give an immediate feedback on the error without running the program.

Jakuje:
Thanks for the suggestions! I was not aware of this exists! Will change to the static assert as this is more suitable! I tried initially something with ifdefs, but it did not work

frankmorgner:
I was jus learning about this myself and never used it in practice 😉

PR #3536, opened 2025-12-19
macOS code signing

metsma:
##### Checklist

- [ ] macOS token is tested


frankmorgner (commented):
I pushed the changes to an [OpenSC branch](https://github.com/OpenSC/OpenSC/tree/refs/heads/pr_3536) to see if this works here as well. Let's see if we see green lights here as well:
https://github.com/OpenSC/OpenSC/actions?query=branch%3Apr_3536

Jakuje:
The pipeline looks ok and the artifacts made it to the nightly so all looks good to me:
https://github.com/OpenSC/Nightly/tree/2026-01-02_797926fe

If either of you can try that the artifacts work as expected, I think we are good to go!

metsma:
> The pipeline looks ok and the artifacts made it to the nightly so all looks good to me: https://github.com/OpenSC/Nightly/tree/2026-01-02_797926fe
> 
> If either of you can try that the artifacts work as expected, I think we are good to go!

Looks good to me

Jakuje:
Thank you! Merging then.

PR #3549, opened 2026-01-13
configure: Drop AC_C_CONST and softhsm 2.7.0 compatibility

Jakuje:
The autoconf in Fedora rawhide does not seem to correctly detect the support for `const`:

```
checking for an ANSI C-conforming const... no
```

which is resulting in weird errors where we are passing const to non-const arguments, while having `const` in place.

The autoconf documentation says this macro is obsolescent so lets drop it before it will cause more issues:

https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/C-Compiler.html

##### Checklist

- [~] macOS token is tested

### DIFF
--- a/src/pkcs11/pkcs11-spy.c
+++ b/src/pkcs11/pkcs11-spy.c
@@ -337,15 +337,32 @@ init_spy(void)
 		return CKR_DEVICE_ERROR;
 	}
 	modhandle = C_LoadModule(module, &po_v2);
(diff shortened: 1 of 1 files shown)


Jakuje:
and now, we have failures in the tests due to the fedora rawhide having softhsm 2.7.0.rc1 with new functionalities. Cute!

metsma (approved):

Jakuje:
Thank you for the review! Merging

Issue #3500, opened 2025-10-16
pkcs11-tool: sc_evp_md returns 0 for all SHA* algorithm in FIPS mode.


dargad:
This issue is a follow-up to #3495.

After applying #3496 (Ubuntu 24.04 test build available in my [ppa:dgadomski/lp2127205-opensc](https://launchpad.net/~dgadomski/+archive/ubuntu/lp2127205-opensc)) I am still getting almost identical stack, but this time with sha224:
```
#0  EVP_MD_up_ref (md=md@entry=0x0) at ../crypto/evp/digest.c:1083
#1  0x00007ffff7041e86 in ossl_md_copy (src=0x0, dst=0x5555555e58e0) at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/openssl.c:240
#2  0x00007ffff703ed69 in sc_pkcs11_copy_mechanism (new_mt=<synthetic pointer>, mt=0x5555555e56e0)
    at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/mechanism.c:113
#3  sc_pkcs11_register_mechanism (p11card=p11card@entry=0x5555555c1f00, mt=mt@entry=0x5555555e56e0, result_mt=result_mt@entry=0x0)
    at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/mechanism.c:170
#4  0x00007ffff7042417 in sc_pkcs11_register_openssl_mechanisms (p11card=0x5555555c1f00)
    at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/openssl.c:323
#5  sc_pkcs11_register_generic_mechanisms (p11card=0x5555555c1f00) at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/mechanism.c:1749
#6  register_mechanisms (p11card=0x5555555c1f00) at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/framework-pkcs15.c:6511
#7  pkcs15_bind (p11card=0x5555555c1f00, app_info=<optimized out>) at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/framework-pkcs15.c:345
#8  0x00007ffff703ae26 in card_detect (reader=reader@entry=0x5555555c1100) at /build/opensc-CNVldA/opensc-0.26.1/src/pkcs11/slot.c:323
[...]

PR #3551, opened 2026-01-14
Adjust FIPS mode detection even further to work with RHEL and Ubuntu (hopefully)

Jakuje:
Follow-up from the discussion in #3500, it looks like the Ubuntu FIPS mode works the same way as the RHEL one so when the global openssl context identifies FIPS Mode, we should load FIPS provider.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6469,6 +6469,7 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 	int		sec = 0;
 	CK_TOKEN_INFO info;
 	CK_ULONG parameter_set;
+	int kem = 0;
 
 	switch(getCLASS(sess, obj)) {
 		case CKO_PRIVATE_KEY:
@@ -6678,6 +6679,7 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 			}
 			break;
 		case CKK_ML_KEM:
+			kem = 1;
 			switch (parameter_set) {
 			case CKP_ML_KEM_512:
 				printf("; ML-KEM-512\n");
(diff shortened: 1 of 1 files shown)


Jakuje:
> test-p11test.sh is failing at https://github.com/OpenSC/OpenSC/actions/runs/21032922793/job/60473195612?pr=3551 for some reason. The [build errors](https://github.com/OpenSC/OpenSC/actions/runs/21032922793/job/60473195568?pr=3551) in AppVeyor and build-strict seem to be unrelated.

I think these builds were failing due to unrelated changes that should have been fixed with recently merged #3549. I will rebase and see.

> As sidenote, who manages (the security of) the docker images that we are using from quai.io? If I am reading the documentation correctly, then the CI runner's secrets are not exposed to the docker images by default and have to be enabled explicitly, right?

Its a Red Hat managed container registry: https://quay.io/security/ -- AFAIK the centos images are not on dockerhub (or the github default registry) for some reason. I think pushes to the fedora/centos official repositories are part of the release/update pipelines, but I was never digging into that deeper.

Regarding secrets, my reading is that this applies only for the reusable workflows. What we do here in our CI is mostly calling the "run", which executes the shell scripts we have in our repository.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets

> In external-pkcs11.yaml we are using quay.io/fedora/fedora:latest, but in fedora.yml we use fedora:latest. Is there any reason for this?

I know in other projects we had some issues the default registry (dockerhub?) had some outdated Fedora images or caching on the way so we were switching to quay directly. Technically, for Fedora the same images should be in both places, I believe. But I never digged into this deeper. I can switch to either if you prefer.

frankmorgner (commented):
Thank you, the CI jobs are looking better now.

Regarding the containers, I don't have a preference regarding the mirror, both seem credible enough. I'd like to keep a single mirror consistently to remind us that there was something special with the container when we need to use a different mirror.

You're right with the secrets, I think. As long as we don't configure something special, they will only be available in the workflow without being accessible by the programs we run.

src/tools/pkcs11-tool.c
```diff
@@ -6818,11 +6820,11 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 		printf("%sderive", sepa);
 		sepa = ", ";
 	}
-	if (pub && getENCAPSULATE(sess, obj)) {
+	if (pub && kem && getENCAPSULATE(sess, obj)) {
```
frankmorgner, line 6823: Are there modules which cannot handle this request for ML-KEM public keys? If implemented correctly, we should not need this additional check, right?

Jakuje: Technically, these are general public/private key attributes in pkcs11 3.2. But they really make sense only on the keys that can do KEM. The modules should handle that gracefully, likely but returning `CKR_ATTRIBUTE_TYPE_INVALID` is quite noisy in the pkcs11 tool:
```
warning: PKCS11 function C_GetAttributeValue(ENCAPSULATE) failed: rv = CKR_ATTRIBUTE_TYPE_INVALID (0x12)
```
so I tend to avoid that to prevent needless bug reports.

PR #3555, opened 2026-01-19
Bare minimum Lithuanian Identity Card driver

tadas-s:
Mostly copy/paste work from card-edo.c - I have no idea how PACE works, so this a guesswork.

opensc-explorer connects with a correct CAN number. And it fails with incorrect one.

I'm able to read some EFs, eg: 011C (which is something to do with PACE).

I'm able to change dir into DF02 and then back up to 3F00.

After changing to DF02, I'm able to read some records which look like public keys.

##### Checklist

- [ ] macOS token is tested


tadas-s:
Sorry, did not mean to open PR yet.

PR #3557, opened 2026-01-21
Fixed CID 501272: Resource leaks

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/pkcs11/pkcs11-spy.c
+++ b/src/pkcs11/pkcs11-spy.c
@@ -223,7 +223,7 @@ init_spy(void)
 {
 	CK_FUNCTION_LIST_PTR po_v2 = NULL;
 	const char *output, *module;
-	CK_RV rv = CKR_OK;
+	CK_RV rv = CKR_GENERAL_ERROR;
 #ifdef _WIN32
         char temp_path[PATH_MAX], expanded_path[PATH_MAX];
         DWORD temp_len, expanded_len;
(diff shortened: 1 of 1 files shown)


src/pkcs11/pkcs11-spy.c
```diff
@@ -326,8 +329,8 @@ init_spy(void)
 		}
 	}
 #endif
-	if (module == NULL) {
-		fprintf(spy_output, "Error: no module specified. Please set PKCS11SPY environment.\n");
+	if (module == NULL || (modhandle = C_LoadModule(module, &po_v2)) == NULL) {
+		fprintf(spy_output, "Error: '%s' cannot be loaded. Please set PKCS11SPY environment.\n", module);
```
Jakuje: I would probably keep this as separate errors as this makes it a bit confusing.

This check is for `module == NULL` which means the environment variable was not set up. The next check should be for valid `modhandle`, which means the variable is set, but it does not look like exposing the pkcs11 API.

src/pkcs11/pkcs11-spy.c
```diff
@@ -350,12 +353,9 @@ init_spy(void)
...
-	if (modhandle && po) {
-		fprintf(spy_output, "Loaded: \"%s\"\n", module);
-	}
-	else {
+	} else {
+		memcpy(po, po_v2, sizeof(CK_FUNCTION_LIST));
+		fprintf(spy_output, "Loaded: '%s'\n", module);
 		po = NULL;
 		free(pkcs11_spy);
 		pkcs11_spy = NULL;
```
Jakuje: I think you wanted to remove these lines (and the following ones) as they are "failure" ones.

But removing the whole "else" branch and keeping just the two lines (memcpy, fprintf) outside would be also ok as all the previous failure paths return.

frankmorgner: Oh, right 🤦

src/pkcs11/pkcs11-spy.c
frankmorgner: due to the heavy error handling, I would actually like to keep this together...

frankmorgner:
I have pushed an additional commit to address the comments. This unifies the error handling in the function to keep the code more readable.

PR #3558, opened 2026-01-22
card-piv.c - fix history object

dengert:
Improve handling of History object
 On branch piv-history-fix
 Changes to be committed:
	modified:   card-piv.c

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -5135,14 +5135,14 @@ piv_process_history(sc_card_t *card)
 			goto err;
 		}
 		fp++;
-		if (strlen(fp) != 64) {  /* ASCII-HEX encoded SHA-256 */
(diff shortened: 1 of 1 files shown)


Jakuje:
Actually, it would make sense to me to add an attribution to Nicholas who reported this issue to us.

src/libopensc/card-piv.c
```diff
@@ -5130,6 +5135,11 @@ piv_process_history(sc_card_t *card)
 			goto err;
 		}
 		fp++;
+		if (strlen(fp) != 64) {  /* ASCII-HEX encoded SHA-256 */
```
frankmorgner: I think it would make sense to sanitize the string with `isxdigit()`. This would avoid potential future problems like directory traversels etc.

If you do sanitization character by character, then you can check for the max length of 64 at the same time.

PR #3561, opened 2026-01-25
ui: fixed detecting default language

### LINKED ISSUE #2008: Notify shows messages in the wrong language
### Problem Description

I recently updated OpenSC using the MSI installer.  I think my older version didn't have the notification service, or it wasn't enabled, so I selected the Notify option during install.  Today, when I logged in to my computer, the Windows Action Center had two messages from "OpenSC Notify", and both were in German.  (I don't remember the exact wording, but something to the effect of "card reader was recognized" and "card is ready".)

As far as I can tell, there is nothing on the wiki about the Notify service, and I don't see any relevant lines in `opensc.conf`.  I actually can't find a binary with `notify` in the name anywhere in the install directory either?

### Proposed Resolution

Notifications should use the OS language (English, in my case) if possible, or fall back to a sensible default (again, English?) if not.

### Steps to reproduce

No idea, honestly.

### Logs

I don't know where Notify logs would show up, if there are any -- again, I don't see any documentation.

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/2008

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/ui/char_str_from_wchar.h
+++ b/src/ui/char_str_from_wchar.h
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
+#ifdef _WIN32
 #include <windows.h>
 
 static char *char_str_from_wchar(const WCHAR *in)
@@ -44,3 +45,4 @@ static char *char_str_from_wchar(const WCHAR *in)
 
 	return out;
 }
+#endif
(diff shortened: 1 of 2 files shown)


Jakuje (approved):
untested, but looks sensible

Jakuje:
@frankmorgner this is also good to go for 0.27?

frankmorgner:
after release

PR #3562, opened 2026-01-26
switched to Github issue forms

frankmorgner:
##### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -19,6 +19,18 @@ body:
       required: true
 
   - type: textarea
+    id: steps
+    attributes:
+      label: Steps to reproduce
+      description: How can we reproduce the issue?
+      placeholder: |
+        1.
+        2.
+        3.
+    validations:
+      required: true
+
+  - type: textarea
     id: expected
     attributes:
       label: Expected behavior
@@ -31,18 +43,6 @@ body:
     attributes:
       label: Actual behavior
       description: What actually happened.
-    validations:
-      required: true
-
-  - type: textarea
-    id: steps
-    attributes:
-      label: Steps to reproduce
-      description: How can we reproduce the issue?
-      placeholder: |
-        1.
-        2.
-        3.
     validations:
       required: true
 


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+    id: steps
+    attributes:
+      label: Steps to reproduce
+      description: How can we reproduce the issue?
+      placeholder: |
+        1.
+        2.
+        3.
+    validations:
+      required: true
```
Jakuje: I would prefer to have the steps to reproduce first and what happened/what is expected to happen later. This is somehow in line with the flow of how things are happening (first the input, then the output) and in-line with the bugzilla template I was used to (for exmaple here):

https://bugzilla.redhat.com/show_bug.cgi?id=2433079

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logs and debug output
+      description: |
+        Please include relevant logs or debug output.
+        You can enable verbose output setting the environment variable `OPENSC_DEBUG` to `3`.
+        - `set OPENSC_DEBUG="3"` (Windows CMD)
+        - `export OPENSC_DEBUG="3"` (Unix shell)
```
Jakuje, line 86: Should we suggest log level 8 (all minus PIN commands) by default?

.github/ISSUE_TEMPLATE/bug_report.yml
frankmorgner: done

.github/ISSUE_TEMPLATE/bug_report.yml
frankmorgner: Level 3 was alway sufficient for me. I used to be distracted from OpenSSL or ASN.1 debugging, for example...

Jakuje: ok, so lets keep it.

jluebbe:
It seems it's no longer possible to open a generic bug report:
image
Is that intentional?

Jakuje:
Sounds like a bug. This PR has a `.github/ISSUE_TEMPLATE/bug_report.yml` but for some reason its not rendered ...

Jakuje:
OK, Github has a linter in there:

https://github.com/OpenSC/OpenSC/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml
```
Critical

There are some problems with this template

Body must have unique ids. [Learn more about error 1.](https://docs.github.com/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#body-must-have-unique-ids)

Checkboxes must have unique labels. [Learn more about error 2.](https://docs.github.com/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#checkboxes-must-have-unique-labels)
```
Let me fix that.

Jakuje:
Filled #3568. The linter no longer complains so I think we should be good:

https://github.com/Jakuje/OpenSC/blob/gh-forms/.github/ISSUE_TEMPLATE/bug_report.yml

frankmorgner:
Unfortunately, I did not find the linter in the documentation or anywhere else... 🤷‍♂️

Also, I added #3569 for some more cleanup

PR #3564, opened 2026-01-29
CI: updated github actions

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -208,7 +208,7 @@ jobs:
           version: apt-latest-base
           execute_install_scripts: true
       - name: Restore LibreSSL build
-        uses: actions/cache/restore@v4
+        uses: actions/cache/restore@v5
         id: libressl-cache
         with:
           path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/centos.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       image: quay.io/centos/centos:stream10
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
```
metsma: Can you update to v6

> Improved credential security: persist-credentials now stores credentials in a separate file under $RUNNER_TEMP instead of directly in .git/config

frankmorgner: done

.github/workflows/macos.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     runs-on: macos-latest
     needs: [build]
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Pull build artifacts
         uses: actions/download-artifact@v4
```
metsma: Maybe update the download/upload artifacts also?

Jakuje (approved):
not sure whats up with the doc pipeline though

PR #3571, opened 2026-02-12
OpenSC 0.27.0-rc1

Jakuje:
##### Checklist

- [ ] macOS token is tested


configure.ac
```diff
@@ -39,9 +39,9 @@ OPENSC_VS_FF_PRODUCT_UPDATES="VS_FF_PRODUCT_UPDATES"
 #   (Code changed:                      REVISION++)
 #   (Oldest interface changed/removed:  OLDEST++)
 #   (Interfaces added:                  CURRENT++, REVISION=0)
-OPENSC_LT_CURRENT="12"
+OPENSC_LT_CURRENT="13"
 OPENSC_LT_OLDEST="12"
```
frankmorgner:
```suggestion
OPENSC_LT_OLDEST="13"
```
This needs to be set to the CURRENT value, because we added `key_type` to `struct sc_ec_parameters`  and we removed `cache` from `sc_card_t`. This means the input data for all interfaces using these types is incompatible with previous versions.

Jakuje: Okay. I missed the removal.

But do we still consider the libopensc API as externally-facing?

frankmorgner: By policy, we don't. However, the versioning guidance only considers the techical view
https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

There is the possibility to use dedicated library names for every new version for internal libraries, but we aren't using that right now
https://autotools.info/libtool/version.html

And although we are discouraging to link against libopensc, there are still projects that are doing this.
https://github.com/carblue/acos5

Jakuje: Thanks for double-checking! Fixed!

SECURITY.md
```diff
@@ -9,8 +9,8 @@ backport security fixes into them. Only the last release is supported.
 
 | Version  | Supported          |
 | -------- | ------------------ |
-| 0.26.1   | :white_check_mark: |
-| < 0.26.1 | :x:                |
+| 0.27.0   | :white_check_mark: |
+| < 0.27.0 | :x:                |
```
Jakuje: Actually, I think we should update this only with the final release. Will drop this from this rc commit.

Issue #3566, opened 2026-02-11
CKF_USER_PIN_TO_BE_CHANGED get set for card that report logged in state as PIN_STATE_UNKNOWN


MkfsSion:
### Problem statement

PR [#3490](https://github.com/OpenSC/OpenSC/pull/3490) turns `logged_in` field into a bitmap, but `PIN_STATE_UNKNOWN` has a value of `-1` which means it has all 32 bits set. This include `PIN_STATE_NEEDS_CHANGE`, which makes card incorrectly marked with `CKF_USER_PIN_TO_BE_CHANGED` when card driver reports logged in state as PIN_STATE_UNKNOWN and causes issues with applications checking this flag (e.g., OpenSSL PKCS11 provider).

This can be reproduced with OpenPGP driver by forcing it using `SC_PIN_CMD_GET_INFO` emulation.
A quick search of codebase indicates that there are also other drivers may report logged in state as `PIN_STATE_UNKNOWN` (e.g. gids driver), they might also be affected by the issue.

### Proposed solution

Before checking  `PIN_STATE_NEEDS_CHANGE`, check `PIN_STATE_UNKNOWN`  first.

### Additional context

OpenSSL PKCS11 provider log:
```
2026-02-11 17:44:31.018] Slot Info:
  ID: 18
  Description:      [Canokeys Canokey [OpenPGP PIV OATH] (*) 00 00]
  Manufacturer ID:  [canokeys.org]
  Flags (0x000007):

[2026-02-11 17:44:31.018]     CKF_TOKEN_PRESENT         (0x000001)
[2026-02-11 17:44:31.018]     CKF_REMOVABLE_DEVICE      (0x000002)
[2026-02-11 17:44:31.018]     CKF_HW_SLOT               (0x000004)
[2026-02-11 17:44:31.018]   Hardware Version: 0.0
  Firmware Version: 0.0

[2026-02-11 17:44:31.018] Token Info:
  Label:            [OpenPGP card (User PIN (sig))]
  Manufacturer ID:  [OpenPGP project]
  Model:            [PKCS#15 emulated]
  Serial Number:    [*]
  Flags (0x08040c):

[2026-02-11 17:44:31.018]     CKF_LOGIN_REQUIRED                  (0x000004)
[2026-02-11 17:44:31.018]     CKF_USER_PIN_INITIALIZED            (0x000008)
[2026-02-11 17:44:31.018]     CKF_TOKEN_INITIALIZED               (0x000400)
[...]

PR #3573, opened 2026-02-12
Change usage of pin logged_in state to bitfield

Jakuje:
Fixes: #3566

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/opensc.h
+++ b/src/libopensc/opensc.h
@@ -434,10 +434,10 @@ typedef struct sc_reader {
 #define SC_PIN_ENCODING_GLP	2 /* Global Platform - Card Specification v2.0.1 */
 
 /** Values for sc_pin_cmd_pin.logged_in, can be bitmapped together */
-#define SC_PIN_STATE_UNKNOWN	-1
-#define SC_PIN_STATE_LOGGED_OUT 0
-#define SC_PIN_STATE_LOGGED_IN  1
-#define SC_PIN_STATE_NEEDS_CHANGE 2
+#define SC_PIN_STATE_UNKNOWN	0
+#define SC_PIN_STATE_LOGGED_OUT 1
+#define SC_PIN_STATE_LOGGED_IN  2
+#define SC_PIN_STATE_NEEDS_CHANGE 4
 
 /* A card driver receives the sc_pin_cmd_data and sc_pin_cmd_pin structures filled in by the
  * caller, with the exception of the fields returned by the driver for SC_PIN_CMD_GET_INFO.
--- a/src/libopensc/card-dtrust.c
+++ b/src/libopensc/card-dtrust.c
@@ -734,7 +734,7 @@ dtrust_pin_cmd(struct sc_card *card,
 		data2.pin_type = data->pin_type;
 		r = dtrust_pin_cmd(card, &data2, tries_left);
 
-		if (data2.pin1.logged_in == SC_PIN_STATE_LOGGED_IN) {
+		if (data2.pin1.logged_in & SC_PIN_STATE_LOGGED_IN) {
 			/* Return if we are already authenticated */
 			sc_log(card->ctx, "PIN 0x%02x already verified. Skipping authentication.", data->pin_reference);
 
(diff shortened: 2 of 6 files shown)


src/libopensc/opensc.h
```diff
@@ -434,10 +434,10 @@ typedef struct sc_reader {
 #define SC_PIN_ENCODING_GLP	2 /* Global Platform - Card Specification v2.0.1 */
 
 /** Values for sc_pin_cmd_pin.logged_in, can be bitmapped together */
-#define SC_PIN_STATE_UNKNOWN	-1
-#define SC_PIN_STATE_LOGGED_OUT 0
-#define SC_PIN_STATE_LOGGED_IN  1
-#define SC_PIN_STATE_NEEDS_CHANGE 2
+#define SC_PIN_STATE_UNKNOWN	0
+#define SC_PIN_STATE_LOGGED_OUT 1
+#define SC_PIN_STATE_LOGGED_IN  2
```
metsma, line 439: Isn't there now be risk that card can be at same time be logged in and logged out?

metsma: Seems like code uses only assignments and will clear previous states.
There is one minor problem that I discovered.
In iso7816.c:iso7816_pin_cmd
```
	} else if (r == SC_ERROR_PIN_CODE_INCORRECT) {
		data->pin1.tries_left = apdu->sw2 & 0xF;
		data->pin1.logged_in = SC_PIN_STATE_LOGGED_OUT;
		if (data->cmd == SC_PIN_CMD_GET_INFO)
			r = SC_SUCCESS;
	} else if (r == SC_ERROR_AUTH_METHOD_BLOCKED) {
		data->pin1.tries_left = 0;
		data->pin1.logged_in = SC_PIN_STATE_LOGGED_OUT;
		if (data->cmd == SC_PIN_CMD_GET_INFO)
			r = SC_SUCCESS;
	}
```
it will clear SC_PIN_STATE_NEEDS_CHANGE status
I am not sure if the correct solution is bitwise alter status or even prevent login this case.
This probably something I should handle in esteid2025 esteid_pin_cmd and will create followup PR

metsma (approved):

src/libopensc/opensc.h
frankmorgner: The iso driver is used in many places, so it should be more cautious than other drivers. So feel free to change this to something like
```
data->pin1.logged_in &= ~(SC_PIN_STATE_LOGGED_IN | SC_PIN_STATE_UNKNOWN);
data->pin1.logged_in |= SC_PIN_STATE_LOGGED_OUT;
```

MkfsSion: > it will clear SC_PIN_STATE_NEEDS_CHANGE status I am not sure if the correct solution is bitwise alter status or even prevent login this case. This probably something I should handle in esteid2025 esteid_pin_cmd and will create followup PR

Seems `SC_PIN_STATE_NEEDS_CHANGE ` flag will only be set with `SC_PIN_CMD_GET_INFO` pin cmd, this flag won't propagate to commands like `SC_PIN_CMD_VERIFY`.  Should this flag be cached, it's more appropriate to be done with card driver with this capability.

MkfsSion: > ```
> data->pin1.logged_in &= !(SC_PIN_STATE_LOGGED_IN | SC_PIN_STATE_UNKNOWN);
> ```

`!` seems to be a typo here, a bitwise NOT should be used here.

metsma: `data->pin1.logged_in &= ~(SC_PIN_STATE_LOGGED_IN | SC_PIN_STATE_UNKNOWN);`

Jakuje: > Isn't there now be risk that card can be at same time be logged in and logged out?

in theory, it could,  but I did not find a better way to fight this bitfield, if we want to keep the "unknown" state (otherwise the LOGGED_OUT would be 0). But if we only set only the LOGGED_IN or LOGGED_OUT without using bitmask, they should never be set at the same time.

Regarding the possible override of `SC_PIN_STATE_NEEDS_CHANGE`, feel free to take it to follow-up PR, if this looks good otherwise. I will likely not get to fix that today.

metsma: Looks good to me for now

PR #3574, opened 2026-02-14
Piv tool  support generating retired keys 

dengert:
piv-tool.c  and pkcs15-piv.c have been modified to support generating any of the 20  PIV retired keys.

piv-tool.1.xml documents this and shows an example of creating a History object so these can be accessed. 

A Token2 R3.3  token with a PIV applet was used in the testing, which included RSA 4096, ED25519 and X25519 keys. 
Other "PIV-like" devices also support these 

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/doc/files/files.html
+++ b/doc/files/files.html
@@ -57,7 +57,7 @@ span.errortext {
 						<code class="literal">HKEY_LOCAL_MACHINE</code> (if available)
 				</p></li><li class="listitem"><p>
 						system-wide configuration file
-						(<code class="literal">/opt/ossl-3.5.4/etc/opensc.conf</code>)
+						(<code class="literal">/etc/opensc.conf</code>)
 				</p></li></ol></div><p>
 		</p><p>
 			The configuration file, <code class="literal">opensc.conf</code>, is composed
@@ -146,7 +146,7 @@ app <em class="replaceable"><code>application</code></em> {
 						directory for
 						<span class="citerefentry"><span class="refentrytitle">pkcs15-init</span>(1).
 						</span>
-						(Default: <code class="literal">/opt/ossl-3.5.4/share/opensc</code>).
+						(Default: <code class="literal">/usr/share/opensc</code>).
 					</p><p>
 						If this configuration value is not found on
 						Windows, the registry key
@@ -304,7 +304,7 @@ app <em class="replaceable"><code>application</code></em> {
 								<code class="option">module_path = <em class="replaceable"><code>filename</code></em>;</code>
 							</span></dt><dd><p>
 									Directory with external SM module
-									(Default: /opt/ossl-3.5.4/lib).
+									(Default: /usr/lib64).
 								</p><p>
 									If this configuration value is not
 									found on Windows, the registry key
@@ -469,7 +469,7 @@ app <em class="replaceable"><code>application</code></em> {
 						</span></dt><dd><p>
 								Use specific PC/SC provider
 								(Default:
-								<code class="literal">/usr/lib/x86_64-linux-gnu/libpcsclite.so</code>).
+								<code class="literal">libpcsclite.so.1</code>).
 						</p></dd></dl></div></div><div class="refsect3"><a name="openct"></a><h4>Configuration of OpenCT Readers</h4><div class="variablelist"><dl class="variablelist"><dt><span class="term">
 							<code class="option">readers = <em class="replaceable"><code>num</code></em>;</code>
 						</span></dt><dd><p>
@@ -1233,11 +1233,11 @@ app <em class="replaceable"><code>application</code></em> {
 						PIV configuration during initialization
 						See Configuration Options for PIV Card.
 				</p></dd></dl></div></div><div class="refsect1"><a name="id-1.2.6"></a><h2>Files</h2><div class="variablelist"><dl class="variablelist"><dt><span class="term">
-					<code class="filename">/opt/ossl-3.5.4/etc/opensc.conf</code>
+					<code class="filename">/etc/opensc.conf</code>
 				</span></dt><dd><p>
 						System-wide configuration file
 				</p></dd><dt><span class="term">
-					<code class="filename">/opt/ossl-3.5.4/share/doc/opensc/opensc.conf</code>
+					<code class="filename">/usr/share/doc/opensc/opensc.conf</code>
 				</span></dt><dd><p>
 						Extended example configuration file
 				</p></dd></dl></div></div></div><div class="refentry"><div class="refentry.separator"><hr></div><a name="pkcs15-profile"></a><div class="titlepage"></div><div class="refnamediv"><h2>Name</h2><p>pkcs15-profile — format of profile for <span class="command"><strong>pkcs15-init</strong></span></p></div><div class="refsect1"><a name="id-1.3.3"></a><h2>Description</h2><p>
@@ -1255,7 +1255,7 @@ app <em class="replaceable"><code>application</code></em> {
 		</p><p>
 			The card specific profile contains additional information required during
 			card initialization, such as location of PIN files, key references etc.
-			Profiles currently reside in <code class="filename">/opt/ossl-3.5.4/share/opensc</code>
+			Profiles currently reside in <code class="filename">/usr/share/opensc</code>
 		</p><p>
 			Basic PKCS#15 terminology:
 			</p><div class="orderedlist"><ol class="orderedlist" type="1"><li class="listitem"><p>


src/tools/piv-tool.c
```diff
@@ -295,18 +295,38 @@ static int gen_key(const char * key_info)
 
 	sc_hex_to_bin(key_info, buf, &buflen);
 	if (buflen != 2) {
-		fprintf(stderr, "<keyref>:<algid> invalid, example: 9A:06\n");
+		fprintf(stderr, "<keyref>:<algid> invalid, example: 9A:07 9A ker with RSA 2048\n");
```
frankmorgner: `ker`?

doc/tools/piv-tool.1.xml
```diff
@@ -186,6 +195,12 @@
 		</para>
 	</refsect1>
 
+	<refsect1>
+		<title>Examples</title>
+		<para>create History Object to use the 20 retired keys</para>
```
frankmorgner:
```suggestion
		<para>Authenticate to the card and create History Object to use the 20 retired keys</para>
```

src/tools/piv-tool.c
dengert: ker should be key, see 74c43c431208c4eee5ab25c2294dad8dd4826c4b

doc/tools/piv-tool.1.xml
dengert: fixed as suggested.

dengert:
@Jakuje   I do my builds in an out of source directory. It looks like `doc/tools/tools.html` and `doc/files/files.html` are generated by Makefile in the build directory, but  these files are  traced by git:
```
git log tools.html
commit ad95c14b110f9c35933fe51f371ba4dc5d467872
Author: Jakub Jelen <jakuje@gmail.com>
Date:   Thu Oct 23 14:07:49 2025 +0200
...
```
So  I get: [Make sure HTML files are updated / build (pull_request)](https://github.com/OpenSC/OpenSC/actions/runs/22044786005/job/63691581923?pr=3574)Failing after 1m

Question:
Should these 2 files be tracked by git? as these are not original source.
Are these used after a PR is committed for the Wiki?

If the need to be tracked, it looks like all I have to do is copy the files back to the source and commit them. 
with a waning when doing out of source builds.

Jakuje:
> @Jakuje I do my builds in an out of source directory. It looks like doc/tools/tools.html and doc/files/files.html are generated by Makefile in the build directory, but these files are traced by git:

This is all is kind of hackish. Generally, the build should not modify anything in source directory, but if it matches the build directory, they are updated. We could change the target to be in source directory, but it will be similarly odd. Third option is just to copy the generated files from the build directory to source directory after build manually. But its even more annoying that the generated files have some hardcoded system paths which depend on configure flags (and which need to match the one in CI).

>  Should these 2 files be tracked by git? as these are not original source.
>  Are these used after a PR is committed for the Wiki?

Yes and No, they are reference from the README.md:

 https://github.com/opensc/opensc?tab=readme-ov-file#opensc-documentation

and rendered like this as a handy manual page:

https://htmlpreview.github.io/?https://github.com/OpenSC/OpenSC/blob/master/doc/tools/tools.html

> If the need to be tracked, it looks like all I have to do is copy the files back to the source and commit them.

yes please. The other option is to apply the patch/diff generated by the CI job.

dengert:
@Jakuje The "Make sure HTML files are updated / build (pull_request)is still failing.

My prefix = "/opt/ossl-3.5.4/" is being added to the the opensc.conf.5,  opensc.conf.xml, pkcs15-profile.5 and pkcs15-profile.xml files  from the Makefile which then does not compare to the CI. 

This appear to happen all time.   

A possible fix  is for these files, that use "prefix"  is to use "comment_prefix" which matches what you want in the CI configuration. It would only be changed by a destro build or user build.  Something like "DESTRO_PATH" so this would show up in the WIKI as "DESTRO_PATH" 

```
grep ossl *
Makefile:LDFLAGS =  -g -Wl,-rpath,/opt/ossl-3.5.4/lib
Makefile:OPENSSL_CFLAGS = -I/opt/ossl-3.5.4/include
Makefile:OPENSSL_LIBS = -L/opt/ossl-3.5.4/lib -lcrypto
Makefile:OPTIONAL_OPENSSL_CFLAGS = -I/opt/ossl-3.5.4/include
Makefile:OPTIONAL_OPENSSL_LIBS = -L/opt/ossl-3.5.4/lib -lcrypto
Makefile:PKG_CONFIG_PATH = /opt/ossl-3.5.4/lib/pkgconfig
Makefile:abs_builddir = /a/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-ossl-3.5.4/doc/files
Makefile:abs_srcdir = /a/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-ossl-3.5.4/../src/doc/files
Makefile:abs_top_builddir = /a/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-ossl-3.5.4
Makefile:abs_top_srcdir = /a/appl/OpenSC-dev/build/opensc-git-my/amd64_linux26-ossl-3.5.4/../src
Makefile:completiondir = /opt/ossl-3.5.4/etc/bash_completion.d
Makefile:prefix = /opt/ossl-3.5.4
opensc.conf.5:system\-wide configuration file (/opt/ossl\-3\&.5\&.4/etc/opensc\&.conf)
opensc.conf.5:/opt/ossl\-3\&.5\&.4/share/opensc)\&.
opensc.conf.5:Directory with external SM module (Default: /opt/ossl\-3\&.5\&.4/lib)\&.
opensc.conf.5:/opt/ossl\-3\&.5\&.4/etc/opensc\&.conf
opensc.conf.5:/opt/ossl\-3\&.5\&.4/share/doc/opensc/opensc\&.conf
opensc.conf.5.xml:						(<literal>/opt/ossl-3.5.4/etc/opensc.conf</literal>)
opensc.conf.5.xml:						(Default: <literal>/opt/ossl-3.5.4/share/opensc</literal>).
opensc.conf.5.xml:									(Default: /opt/ossl-3.5.4/lib).
opensc.conf.5.xml:					<filename>/opt/ossl-3.5.4/etc/opensc.conf</filename>
opensc.conf.5.xml:					<filename>/opt/ossl-3.5.4/share/doc/opensc/opensc.conf</filename>
pkcs15-profile.5:/opt/ossl\-3\&.5\&.4/share/opensc
```

The `Makefile.in`  in both `doc/files` and `doc/tools` needs to be looked at closer if we go this way.

Jakuje:
I think this is a good idea to use some fixed prefix for the html page generation given that they are really just for the website/repo and not manual page for users.

But I will likely not have time to toy with it right now so please, either apply the changes manually or use configure wit the prefix as in CI to build the html pages so we can get this merged.

dengert:
This PR can go in next release. Most PIV-like card vendors provide their own provisioning tools or can use Yubikey PIV tool.

dengert:
> But I will likely not have time to toy with it right now so please, either apply the changes manually or use configure with the prefix as in CI to build the html pages so we can get this merged.

I have both. The PR passes all 67 checks including the "Make sure HTML files are updated / build (pull_request)Successful in 1m" 

I have also been looking at how to build with the values from the CI as best I can tell:   
```
 #!/bin/bash
# the tmp build is at same level as OpenSC
set -xv

	# based on host  DYN_LIB_EXT  from configure ".dll" or ".so"
	#DEFAULT_SM_MODULE from enable_sm
	#PROFILE_DIR_DEFAULT from pkgdatadir
	#DEFAULT_SM_MODULE from enable_sm
	
../OpenSC/configure --enable-doc --prefix=/usr
cd doc/files
#
# default should work:
#   DYN_LIB_EXT=.so

make --debug=b sysconfdir=/etc \
	docdir=/usr/share/doc/opensc \
	libdir=/usr/lib64 \
	DYN_LIB_EXT=.so \
	DEFAULT_PCSC_PROVIDER=libpcsclite.so.1 \
	PROFILE_DIR_DEFAULT=/usr/share/opensc \
	DEFAULT_SM_MODULE=libsmm-local.so \
	pkgdatadir=/usr/share/opensc \
	files.html
	

cd ../tools
make --debug=b \
	pkgdatadir=/usr/share/opensc \
	tools.html
```

If file.html and tools.html are really only for the web site,  values could be turned into generic paths.  

This can all wait till after 0.27 is out.

Jakuje:
Thank you. With the release 0.27.1 out, I think we should be good to merge this

PR #997, opened 2017-03-16
pkcs11-tool: add GOSTR3410 keypair generation

konstantinpersidskiy:
Hello

Recently I found out that pkcs11-tool cannot generate keypairs of mechanism GOSTR3410 (although this mechanism is shown as supported in `pkcs11-tool -M`). This mechanism is a standard mechanism specified by laws in some countries, so I think it is important to make it work  properly in OpenSC.

This commit finishes implementing GOSTR3410 keypair generation feature; now pairs can be generated just like they are with other mechanisms.

Example of keypair generation using native OpenSC pkcs15 framework:
```
$ pkcs11-tool -m GOSTR3410-KEY-PAIR-GEN -k --key-type GOSTR3410:C -d 1234 -l -p '12345678'
Using slot 0 with a present token (0x0)
Key pair generated:
Private Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022303
  label:      Private Key
  ID:         1234
  Usage:      sign
Public Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022303
  VALUE:      044084a5de6a1bfd68ff6b106f6921d4da163244b15ea27dd3a87968d40959a6
              e4170df7674bc7ba098080d7ac2cfe80b29b4ed4369dfedf6e13e42ae32eb7b3
              3233
  label:      Private Key
  ID:         1234
  Usage:      verify
```

Example of keypair generation using another pkcs module:
```
$ pkcs11-tool --module "/home/user/Downloads/librtpkcs11ecp.so" -m GOSTR3410-KEY-PAIR-GEN -k --key-type GOSTR3410:B -d 4321 -l -p '12345678'
Using slot 0 with a present token (0x0)
Key pair generated:
Private Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022302
  label:      
  ID:         4321
  Usage:      sign
Public Key Object; GOSTR3410 
  PARAMS OID: 06072a850302022302
  VALUE:      4c0a3a792a9c03cf227be5b4f94d668a11ce34071b29d9da2585fa1d7cbd1c74
              eb874e468245882c31a66a590cec62699b73b225b69e69ed298ff104be7a8c4f
  label:      
  ID:         4321
  Usage:      verify
```
I believe this commit will be useful. Looking forward for reviews and comments.
Regards, Konstantin Persidskiy

PR #3576, opened 2026-02-18
Add generate and delete operations to keys Magma and Kuznechik

AktivCoAndreyTimofeev:
##### Checklist

- [ ] macOS token is tested


popovec:
According to https://en.wikipedia.org/wiki/Kuznyechik, I quote:

"Later Leo Perrin and Aleksei Udovenko published two alternative decompositions of the S-box and proved its connection to the S-box of the Belarusian cipher BelT.[10] The authors of the paper note that while the reason for using such a structure remains unclear, generating S-boxes by a hidden algorithm contradicts the concept of nothing-up-my-sleeve numbers which could prove that no weaknesses were intentionally introduced in their design." 

Just as openssl gets rid of older, insecure ciphers, we should also not introduce outdated things into OpenSC at all.

frankmorgner (changes requested):
I agree that we should not promote the use of grasshopper. It has been rejected in other places before.

Jakuje:
As already mentioned above, I do not think we should include vendor-specific mechanisms without very good reasoning, which is clearly not provided in this no-description PR.

Mironenko:
@Jakuje , the Magma and Kuznyechik (“Grasshopper”) ciphers are specified in the National Standard of the Russian Federation GOST R 34.12-2015. It is the current national standard for block ciphers in the Russian Federation.

PKCS#11 extensions for implementing these standards are defined in the [following guidelines](https://digitalcryptography.ru/upload/iblock/3d4/y8zw1jb7hhbelqf5qbypvo0hmm4yfl47.pdf) published by the Russian [National Digital Cryptography Technology Center](https://digitalcryptography.ru/) ([about](https://tadviser.com/index.php/Company:National_Digital_Cryptography_Technology_Center)).

These extensions are implemented by major Russian smart-card vendors, including [Aktiv Company](https://aktiv-company.com/) ([Rutoken](https://www.rutoken.ru/)) and [ESMART](https://esmart.ru/) ([ESMART Token](https://token.esmart.ru/)).

Support for such smart cards in widely used, internationally adopted tools such as OpenSC is important for Russian smart-card users.

The Kuznyechik block cipher is implemented in the [Bouncy Castle](github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/engines/GOST3412_2015Engine.java) and [Botan](https://botan.randombit.net/handbook/api_ref/block_cipher.html#kuznyechik) cryptographic libraries.

Jakuje:
Crypto-quality aside, is there some way to test the code introduced in this PR? Is there some software token that implements these operations? None of the current maintainers have a way to test this code, which makes it  hard to justify for us to step up and maintain it if we do not have a way to execute it.

AktivCoAndreyTimofeev:
@Jakuje 

> Crypto-quality aside, is there some way to test the code introduced in this PR? Is there some software token that implements these operations? None of the current maintainers have a way to test this code, which makes it hard to justify for us to step up and maintain it if we do not have a way to execute it.

Unfortunately, there is no publicly available software token that can work with these algorithms. 
We can prepare pkcs11-tool logs for you, for example, through pkcs11-spy or something similar for demonstration.

popovec:
> Unfortunately, there is no publicly available software token that can work with these algorithms. We can prepare pkcs11-tool logs for you, for example, through pkcs11-spy or something similar for demonstration.

Maybe you should start by adding these algorithms to softhsm first. Then we could consider adding support to  pkcs11-tool and doing the appropriate tests in CI.

Code that doesn't have integration tests today can very quickly (by a random error in another PR) break down, and we don't want that.

Jakuje:
> Maybe you should start by adding these algorithms to softhsm first. Then we could consider adding support to pkcs11-tool and doing the appropriate tests in CI.

SoftHSM is mostly dead from my experience. Yes, they made a release this year and merged a lot of stuff, but there is still a lot of PRs and issues that are unhandled and the complexity of adding new crypto to all their backends is something that stopped me from finalizing the last PR I had there.

Even if it would be implemented in softhsm, we would need separate build of softhsm with botan.

AktivCoAndreyTimofeev:
@Jakuje 

How was the previous work on adding the GOST 3410 generation verified?

popovec:
> How was the previous work on adding the GOST 3410 generation verified?

#3577 does not yet meet the conditions for acceptance (at least due to disabling deprecated-declaration).
Of course, I would expect some integration tests with this PR as well.

AktivCoAndreyTimofeev:
> > How was the previous work on adding the GOST 3410 generation verified?
> 
> #3577 does not yet meet the conditions for acceptance (at least due to disabling deprecated-declaration). Of course, I would expect some integration tests with this PR as well.


You misunderstood me. My question concerns the already existing GOST code in the main project branch. How did he pass the checks without a soft token?

popovec:
> You misunderstood me. My question concerns the already existing GOST code in the main project branch. How did he pass the checks without a soft token?

OpenSC didn't use integration tests for everything years ago. The use of integration tests started gradually, first on the "travis" platform and later on "github". The patch from your question was created in 2017...

Date:   Fri Jun 9 11:30:40 2017 +0300
pkcs11-tool: add GOSTR3410 keypair generation (#997)

I don't think the requirement for integration tests is a problem these days. CI tests are common in other projects as well. And OpenSC uses not only softhsm for testing, but also simulations of various real smart cards.

DenisSolomatinDev:
Thank you for the detailed comments and for testing the library.

Could you please clarify what materials or artifacts (for example, logs) we could provide to confirm that the program behaves correctly with the introduced changes and that these changes are being used in your code?

On our side, we can offer a logging mechanism that intercepts calls to our PKCS#11 library at the entry points and reports the mechanisms being used, as well as any transformations of input and output data. If needed, we are also ready to provide video materials demonstrating the expected behavior when generating and deleting Magma and Kuznechik keys.

Jakuje:
I am still failing to see the usefulness of this addition. Right now, what is added is the key generation of these two algorithm and names when they are found on tokens, but nothing else.
For the your use case, you use opensc just to do this and some other tools to do the operations using these keys?

popovec:
Since the cipher uses a 64-bit block and the "Botan" library also warns about a dubious S-box, I do not recommend adding such a cipher to pkcs11-tool. If so, I suggest adding the "--allow-insecure-ciphers" switch to pkcs11-tool and only then the switches that enable not so secure (deprecated) ciphers could be used. This should of course also apply to the DES or triple DES algorithm and perhaps also to RSA shorter than 2048 or EC shorter than 256 bits. In the future, this switch will probably be needed for all algorithms except PQC.

DenisSolomatinDev:
At the moment, we would like to add support in pkcs11-tool for generating and deleting Magma and Kuznechik keys.
This is needed for use through our rtPKCS11ECP library in third-party software, because such software can only work with pre-created keys.

frankmorgner:

PR #3577, opened 2026-02-18
Add derive on GOSTR-3410-2012

AktivCoAndreyTimofeev:
- Add reading GOSTR-3410-2012 public key

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -178,6 +178,10 @@ static const struct sc_aid GOST_HASH2001_PARAMSET_OID = { { 0x06, 0x07, 0x2a, 0x
 static const struct sc_aid GOST_HASH2012_256_PARAMSET_OID = { { 0x06, 0x08, 0x2A, 0x85, 0x03, 0x07, 0x01, 0x01, 0x02, 0x02 }, 10 };
 static const struct sc_aid GOST_HASH2012_512_PARAMSET_OID = { { 0x06, 0x08, 0x2A, 0x85, 0x03, 0x07, 0x01, 0x01, 0x02, 0x03 }, 10 };
 
+/* id-tc26-gost3410-12-256 and id-tc26-gost3410-12-512 (RFC 9215) */
+static const unsigned char OID_GOST3410_12_256[] = {0x06, 0x08, 0x2a, 0x85, 0x03, 0x07, 0x01, 0x01, 0x01, 0x01};
+static const unsigned char OID_GOST3410_12_512[] = {0x06, 0x08, 0x2a, 0x85, 0x03, 0x07, 0x01, 0x01, 0x01, 0x02};
+
 enum {
 	OPT_MODULE = 0x100,
 	OPT_SLOT,
@@ -430,7 +434,7 @@ static const char *option_help[] = {
 		"Specify the value <arg> of the mechanism parameter CK_MAC_GENERAL_PARAMS",
 		"Specify additional authenticated data for AEAD ciphers as a hex string",
 		"Specify the required length (in bits) for the authentication tag for AEAD ciphers",
-		"Specify the file containing the salt for HKDF (optional)",
+		"Specify the file containing the salt for HKDF (optional) / UKM for GOSTR3410-12-DERIVE (required; 256: 8 bytes, 512: 8-16 bytes)",
 		"Specify the file containing the info for HKDF (optional)",
 		"When reading a public key, try to read PUBLIC_KEY_INFO (DER encoding of SPKI)",
 		"Specify the PKCS#11 URI for module, slot, token or object",
@@ -6127,8 +6131,6 @@ derive_ec_key(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key, CK_MECHANISM_TYPE
 	char name[256]; size_t len = 0;
 	int nid = 0;
 #endif
-
-	printf("Using derive algorithm 0x%8.8lx %s\n", mech_mech, p11_mechanism_to_name(mech_mech));
 	memset(&mech, 0, sizeof(mech));
 	mech.mechanism = mech_mech;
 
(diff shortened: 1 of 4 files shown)


popovec:
One comprehensive note, not just for this PR.

I would not like to use deprecated functions in OpenSC code anymore. Sooner or later the "openssl engine" will become unusable and everything will need to be rewritten to the provider API. Given that openssl 4.0 will no longer support "engine", I recommend not using any patches that depend on the openssl engine anymore. (Please note that openssl has already deprecated the engine API in version 3.0)

AktivCoAndreyTimofeev:
We are currently developing a provider with support for GOST algorithms.
As soon as it is ready, this code will be fixed. 
The current work with engine is a temporary measure.

src/tools/pkcs11-tool.c
```diff
@@ -6414,6 +6416,295 @@ derive_hkdf(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
...
+			util_fatal("UKM length for GOSTR3410-2012-512 must be 8-16 bytes (got %ld).", (long)ukm_len);
+		}
+	} else {
+		if (ukm_len != 8) {
+			free(ukm);
+			util_fatal("UKM length for GOSTR3410-2012-256 must be 8 bytes (got %ld).", (long)ukm_len);
+		}
+	}
+
+	read_binary_file(opt_input, &file_buf, &file_len);
```
frankmorgner, line 6738: There are already options `--salt-file` and `--input-file` for `pkcs11-tool`.

src/tools/pkcs11-tool.c
```diff
@@ -6414,6 +6416,295 @@ derive_hkdf(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
 	return newkey;
 }
 
+#if defined(ENABLE_OPENSSL) && !defined(OPENSSL_NO_ENGINE)
+
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
frankmorgner: I don't think we want to disable the deprecation warnings just to support a temporary solution. Please have a look at `sc_pkcs11_register_openssl_mechanisms()` from `src/libopensc/openssl.c` if you can reuse the gost initialization.

frankmorgner (changes requested):

src/tools/pkcs11-tool.c
AktivCoAndreyTimofeev: Thanks, I'll pay attention to that.

src/tools/pkcs11-tool.c
DenisSolomatinDev: Thanks for the note — I understand the concern about adding yet another file option when pkcs11-tool already provides --input-file and --salt-file.

Just to confirm I got your point right: you’re suggesting we should reuse the existing --input-file for the peer public key and reuse --salt-file for the UKM/salt, instead of introducing any new dedicated options for this feature.

In the current version of the patch that’s exactly what I did: GOSTR3410-12 derive reads the peer key from --input-file and requires UKM via --salt-file, and I updated the help text to reflect the dual use of --salt-file (HKDF salt / GOST UKM). If you’d prefer different semantics or naming (e.g., stricter validation or a different help wording), I’m happy to adjust.

src/tools/pkcs11-tool.c
DenisSolomatinDev: Thanks for the review. Your comments have been taken into account, and the corresponding changes have been incorporated into the codebase.

src/tools/pkcs11-tool.c
frankmorgner: Thank you for the clarification, I may have confused this previously

src/tools/pkcs11-tool.c
```diff
@@ -6414,6 +6416,397 @@ derive_hkdf(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
...
+{
+	buffer[0] = (CK_BYTE)(value & 0xFF);
+	buffer[1] = (CK_BYTE)((value >> 8) & 0xFF);
+	buffer[2] = (CK_BYTE)((value >> 16) & 0xFF);
+	buffer[3] = (CK_BYTE)((value >> 24) & 0xFF);
+}
+
+/* Read UKM (User Key Material) from file for GOST derive operations */
+static void
+read_binary_file(const char *filename, CK_BYTE **out_buf, ssize_t *out_len)
```
frankmorgner, line 6433: Did you think about using fread_to_eof.c?

src/tools/pkcs11-tool.c
```diff
@@ -6414,6 +6416,397 @@ derive_hkdf(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE key)
...
+	*out_len = len;
+}
+
+/*
+ * Build SubjectPublicKeyInfo (DER) for GOST R 34.10-2012 from PKCS#11 attributes.
+ * Uses CKA_VALUE, CKA_GOSTR3410_PARAMS, CKA_GOSTR3411_PARAMS.
+ * Returns allocated buffer and length, or NULL on failure.
+ */
+static unsigned char *
+build_gost_spki_der(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE obj, CK_KEY_TYPE key_type,
```
frankmorgner, line 6459: This parsing stuff is quite complicated and may easily lead to bad memory handling, but I assume that there are test vectors available. Could you please add unittest (see src/tests/unittests) and a fuzzer target (see src/tests/fuzzing)?

frankmorgner (changes requested):

src/tools/pkcs11-tool.c
DenisSolomatinDev: I replaced the custom read_binary_file() helper with the existing fread_to_eof() utility to avoid duplicating file-reading logic. pkcs11-tool now links fread_to_eof.c and includes fread_to_eof.h.

This should not change the intended behavior for the GOST derive flow, since the inputs we read (UKM and peer public key/SPKI) are small. Error handling remains fatal for read failures, and filename == NULL is still treated the same way as before.

frankmorgner:
What's the status of this topic?

PR #3579, opened 2026-02-19
Lithuanian Identity Card driver

tadas-s:
Hello all,

So here's the result of the discussion I started about Lithuanian electronic ID cards: https://github.com/OpenSC/OpenSC/discussions/3563

I hope I could get some early feedback if possible.

As far as I know there's no official/public documentation available for the card. So this is based on all kinds of guesswork, looking at other drivers, looking at Finnish, Estonian eIDs documentation.

Card uses PACE authentication, so to get even the most basic details we need 6 digit card access number. 

Officially it is documented to have a single PIN code, but under the hood it's two PINs that are kept in sync. One pin is for authentication key, other - for signatures. PIN change and unblock is implemented in the 'lteid-tool'.

Card also includes cardholder details like full name, date of birth, national identification number, black and white photo. But for now that's not implemented. My goal here is authentication with public gov services.

Confirmed working with:
 - TLS client certificates - was able to sign in to a gov website.
 - SSH, both key for signatures and for auth. SSH working is not really a goal of this, but it's a reasonable test.

## Demo / test

First run of `lteid-tool`:

Screenshot From 2026-02-19 13-27-32

After re-inserting the card, Firefox security devices dialog shows:

Screenshot From 2026-02-19 13-28-18

I'm now able to sign in to some* gov websites:

Screenshot From 2026-02-19 13-30-47

Screenshot From 2026-02-19 13-30-58

`pkcs15-tool -D` output:

```
Using reader with a card: Gemalto PC Twin Reader (AD010CEC) 00 00
Connecting to card in reader Gemalto PC Twin Reader (AD010CEC) 00 00...
Using card driver Lithuanian eID card (asmens tapatybės kortelė).
PKCS#15 Card [TADAS SASNAUSKAS]:
        Version        : 1
        Serial number  : d461ee13bdb8d752
        Manufacturer ID: MaskTech International GmbH
        Flags          : Read-only, Login required, PRN generation
                 sc_supported_algo_info[0]:
                         reference  : 1 (0x01)
                         mechanism  : [0x03] CKM_RSA_X_509
                         operations : [0x02], compute_signature
                         algo_id    : 1.2.840.113549.1.1.1
                         algo_ref   : [0x0f]
                 sc_supported_algo_info[1]:
                         reference  : 2 (0x02)
                         mechanism  : [0x03] CKM_RSA_X_509
                         operations : [0x20], decipher
                         algo_id    : 1.2.840.113549.1.1.1
                         algo_ref   : [0x0f]
                 sc_supported_algo_info[2]:
                         reference  : 3 (0x03)
                         mechanism  : [0x40] CKM_SHA256_RSA_PKCS
                         operations : [0x02], compute_signature
                         algo_id    : 1.2.840.113549.1.1.11
                         algo_ref   : [0x11]
                 sc_supported_algo_info[3]:
                         reference  : 4 (0x04)
                         mechanism  : [0x43] CKM_SHA256_RSA_PKCS_PSS
                         operations : [0x02], compute_signature
                         algo_id    : 1.2.840.113549.1.1.10
                         algo_ref   : [0x21]
                 sc_supported_algo_info[4]:
                         reference  : 5 (0x05)
                         mechanism  : [0x1041] CKM_ECDSA
                         operations : [0x02], compute_signature
                         algo_id    : 1.2.840.10045.2.1
                         algo_ref   : [0x0f]


PIN [PACE-PUK]
        Object Flags   : [0x00]
        ID             : 04
        Flags          : [0x805D], case-sensitive, change-disabled, unblock-disabled, initialized, unblockingPin
        Length         : min_len:8, max_len:12, stored_len:0
        Pad char       : 0x00
        Reference      : 4 (0x04)
        Path           : 3f00
        Tries left     : 3

PIN [PACE-PIN]
        Auth ID        : 04
        ID             : 03
        Flags          : [0x8111], case-sensitive, initialized, disable_allowed
        Length         : min_len:6, max_len:12, stored_len:0
        Pad char       : 0x00
        Reference      : 3 (0x03)
        Path           : 3f00
        Tries left     : 3

PIN [PACE-CAN]
        Object Flags   : [0x00]
        ID             : 02
        Flags          : [0x801D], case-sensitive, change-disabled, unblock-disabled, initialized
        Length         : min_len:6, max_len:12, stored_len:0
        Pad char       : 0x00
        Reference      : 2 (0x02)
        Path           : 3f00

PIN [PIN.QES]
        Object Flags   : [0x00]
        Auth ID        : 04
        ID             : 81
        Flags          : [0x213], case-sensitive, local, initialized, integrity-protected
        Length         : min_len:6, max_len:12, stored_len:0
        Pad char       : 0x00
        Reference      : 129 (0x81)
        Path           : 3f00df02
        Tries left     : 0

Private EC Key [Signature Key 1 QES]
        Usage          : [0x204], sign, nonRepudiation
        Access Flags   : [0x09], sensitive, neverExtract
        Algo_refs      : 5
        FieldLength    : 384
        Key ref        : 136 (0x88)
        Native         : yes
        Path           : 3f00df02
        Auth ID        : 81
        ID             : 01
        MD:guid        : 6e4e5217-43fe-c208-d516-7d77be6f4ce4

Private EC Key [Signature Key 2]
        Usage          : [0x04], sign
        Access Flags   : [0x09], sensitive, neverExtract
        Algo_refs      : 5
        FieldLength    : 384
        Key ref        : 140 (0x8C)
        Native         : yes
        Path           : 3f00df02
        Auth ID        : 03
        ID             : 02
        MD:guid        : 9ddbf5df-d102-10ed-4b86-31496bbb653e

Public EC Key [Signature Key 1 QES]
        Usage          : [0x40], verify
        Access Flags   : [0x00]
        FieldLength    : 384
        Key ref        : 136 (0x88)
        Native         : no
        Path           : 3f00df021f07
        ID             : 01

Public EC Key [Signature Key 2]
        Usage          : [0x40], verify
        Access Flags   : [0x00]
        FieldLength    : 384
        Key ref        : 140 (0x8C)
        Native         : no
        Path           : 3f00df021f0b
        ID             : 02

X.509 Certificate [Signature Key 1 QES]
        Object Flags   : [0x00]
        Authority      : no
        Path           : 3f00df021f06
        ID             : 01
        Encoded serial : 02 0E 699B525DEC76E52E0000000061B5

X.509 Certificate [Signature Key 2]
        Object Flags   : [0x00]
        Authority      : no
        Path           : 3f00df021f0a
        ID             : 02
        Encoded serial : 02 0E 699B525DEC76E52E0000000061B6
```

`pkcs11-tool -M` output:

```
Supported mechanisms:
  SHA-1, digest
  SHA224, digest
  SHA256, digest
  SHA384, digest
  SHA512, digest
  MD5, digest
  RIPEMD160, digest
  GOSTR3411, digest
  ECDSA, keySize={384,384}, hw, sign, verify
  ECDSA-SHA1, keySize={384,384}, sign, verify
  ECDSA-SHA224, keySize={384,384}, sign, verify
  ECDSA-SHA256, keySize={384,384}, sign, verify
  ECDSA-SHA384, keySize={384,384}, sign, verify
  ECDSA-SHA512, keySize={384,384}, sign, verify
```

`lteid-tool` allows resuming, unblocking and changing PINs:

image

That part needs more testing as soon as I get a replacement card, managed to completely lock electronic signature PIN.

It works for me, but I'm kinda newbie in this + not exactly experienced with C.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-lteid.c
+++ b/src/libopensc/card-lteid.c
@@ -1,0 +1,591 @@
+/*
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#if defined(ENABLE_SM) && defined(ENABLE_OPENPACE)
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "asn1.h"
+#include "common/compat_strlcat.h"
+#include "internal.h"
+#include "opensc.h"
+#include "sm/sm-eac.h"
+
+static const struct sc_card_operations *iso_ops = NULL;
+static struct sc_card_operations lteid_ops;
+
+static struct sc_card_driver lteid_drv = {
+		"Lithuanian eID card (asmens tapatybės kortelė)", "lteid",
+		&lteid_ops, NULL, 0, NULL};
+
+#define DRVDATA(card)	 ((struct lteid_drv_data *)((card)->drv_data))
+#define LTEID_CAN_LENGTH 6
+
+struct lteid_drv_data {
+	unsigned char pace;
+	unsigned char pace_pin_ref;
+	unsigned char can[LTEID_CAN_LENGTH];
+	unsigned char can_from_cache;
+};
+
+void
+lteid_get_can_cache_path(sc_card_t *card, char *buf, size_t buf_len)
+{
+	sc_get_cache_dir(card->ctx, buf, buf_len);
+
+#ifdef _WIN32
+	strlcat(buf, "\\", buf_len);
+#else
+	strlcat(buf, "/", buf_len);
+#endif
+
+	strlcat(buf, "lteid_can", buf_len);
+}
+
+static int
+lteid_get_cached_can(sc_card_t *card, unsigned char *can)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	char path[PATH_MAX];
+
+	lteid_get_can_cache_path(card, path, sizeof(path));
+
+	FILE *fd = fopen(path, "r");
+
+	if (!fd) {
+		LOG_FUNC_RETURN(card->ctx, 0);
+	}
+
+	if (LTEID_CAN_LENGTH != fread(can, 1, LTEID_CAN_LENGTH, fd)) {
+		LOG_FUNC_RETURN(card->ctx, 0);
+	}
+
+	fclose(fd);
+
+	LOG_FUNC_RETURN(card->ctx, 1);
+}
+
+static int
+lteid_cache_can(sc_card_t *card, const char *can)
+{
+	int rv;
+	char path[PATH_MAX];
+
+	lteid_get_can_cache_path(card, path, sizeof(path));
+
+	FILE *fd = fopen(path, "w");
+
+	if (!fd && errno == ENOENT) {
+		if ((rv = sc_make_cache_dir(card->ctx)) < 0)
+			return rv;
+
+		fd = fopen(path, "w");
+	}
+
+	if (!fd) {
+		return SC_ERROR_INTERNAL;
+	}
+
+	fwrite(can, 1, 6, fd);
+	fclose(fd);
+
+	return SC_SUCCESS;
+}
+
+static int
+lteid_clear_cached_can(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	char path[PATH_MAX];
+
+	lteid_get_can_cache_path(card, path, sizeof(path));
+
+	if (!unlink(path)) {
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_get_can(sc_card_t *card, struct establish_pace_channel_input *pace_input)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	struct lteid_drv_data *drv_data = DRVDATA(card);
+	int got_can = 0;
+
+	drv_data->can_from_cache = 0;
+
+	// Try env variables first
+	const char *can_from_env = getenv("LTEID_CAN");
+	if (can_from_env && strlen(can_from_env) == LTEID_CAN_LENGTH) {
+		got_can = 1;
+		memcpy(drv_data->can, can_from_env, LTEID_CAN_LENGTH);
+	}
+
+	// Try getting one from the cache
+	if (!got_can && lteid_get_cached_can(card, drv_data->can)) {
+		got_can = 1;
+		drv_data->can_from_cache = 1;
+	}
+
+	// Finally see if there is a default in configuration
+	if (!got_can) {
+		const char *can_from_config = NULL;
+
+		for (size_t i = 0; card->ctx->conf_blocks[i]; ++i) {
+			scconf_block **blocks = scconf_find_blocks(card->ctx->conf, card->ctx->conf_blocks[i], "card_driver", "lteid");
+			if (!blocks)
+				continue;
+			for (size_t j = 0; blocks[j]; ++j)
+				if ((can_from_config = scconf_get_str(blocks[j], "can", NULL)))
+					break;
+			free(blocks);
+		}
+
+		if (can_from_config && strlen(can_from_config) == LTEID_CAN_LENGTH) {
+			got_can = 1;
+			memcpy(drv_data->can, can_from_config, LTEID_CAN_LENGTH);
+		}
+	}
+
+	if (!got_can) {
+		sc_log(card->ctx, "Missing or invalid CAN. 6 digits required.");
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN);
+	}
+
+	pace_input->pin_id = PACE_PIN_ID_CAN;
+	pace_input->pin = drv_data->can;
+	pace_input->pin_length = LTEID_CAN_LENGTH;
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_perform_pace(struct sc_card *card, const int ref, const unsigned char *pin, size_t pinlen, int *tries_left)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	struct lteid_drv_data *drv_data = DRVDATA(card);
+	struct establish_pace_channel_input pace_input = {0};
+	struct establish_pace_channel_output pace_output = {0};
+
+	if (drv_data->pace) {
+		sc_sm_stop(card);
+		drv_data->pace = 0;
+		drv_data->pace_pin_ref = 0;
+	}
+
+	if (ref == PACE_PIN_ID_CAN && !pin) {
+		if (SC_SUCCESS != lteid_get_can(card, &pace_input)) {
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN);
+		}
+	} else {
+		pace_input.pin_id = ref;
+		pace_input.pin = pin;
+		pace_input.pin_length = pinlen;
+	}
+
+	int rv = perform_pace(card, pace_input, &pace_output, EAC_TR_VERSION_2_02);
+	if (rv != SC_SUCCESS) {
+		sc_log(card->ctx, "Error performing PACE for pin ref 0x%02x.", ref);
+
+		drv_data->pace = 0;
+		drv_data->pace_pin_ref = 0;
+
+		// Special case after entering incorrect PACE PIN twice. Card locks with 1 PIN attempt remaining.
+		if (ref == PACE_PIN_ID_PIN && rv == SC_ERROR_NO_CARD_SUPPORT) {
+			rv = SC_ERROR_AUTH_METHOD_BLOCKED;
+		}
+
+		// When CAN code authentication fails and CAN code comes from cache - clear it.
+		// We don't want to make multiple attempts if code is wrong. User should run lteid-tool
+		// again to set up the card.
+		if (ref == PACE_PIN_ID_CAN && drv_data->can_from_cache && rv != SC_ERROR_INTERNAL) {
+			if (lteid_clear_cached_can(card)) {
+				drv_data->can_from_cache = 0;
+			}
+		}
+
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	// If caller provided CAN is valid - cache it.
+	if (ref == PACE_PIN_ID_CAN && pin) {
+		lteid_cache_can(card, (const char *)pin);
+	}
+
+	// Track PACE status
+	drv_data->pace = 1;
+	drv_data->pace_pin_ref = ref;
+
+	free(pace_output.ef_cardaccess);
+	free(pace_output.recent_car);
+	free(pace_output.previous_car);
+	free(pace_output.id_icc);
+	free(pace_output.id_pcd);
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_unlock(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (SC_SUCCESS != lteid_perform_pace(card, PACE_PIN_ID_CAN, NULL, 0, NULL)) {
+		sc_log(card->ctx, "Unlock with CAN code failed. No CAN found in environment, opensc.conf nor cache.");
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_init(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	int rv;
+
+	struct lteid_drv_data *drv_data = calloc(1, sizeof(struct lteid_drv_data));
+
+	if (drv_data == NULL)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
+
+	drv_data->pace = 0;
+	drv_data->pace_pin_ref = 0;
+	card->drv_data = drv_data;
+
+	memset(&card->sm_ctx, 0, sizeof card->sm_ctx);
+
+	card->max_send_size = 65535;
+	card->max_recv_size = 65535;
+	card->caps |= SC_CARD_CAP_ISO7816_PIN_INFO | SC_CARD_CAP_APDU_EXT;
+
+	_sc_card_add_ec_alg(card, 384, SC_ALGORITHM_ECDSA_HASH_NONE | SC_ALGORITHM_ECDSA_RAW, 0, NULL);
+
+	rv = sc_enum_apps(card);
+	LOG_TEST_RET(card->ctx, rv, "Enumerate apps failed");
+
+	rv = lteid_unlock(card);
+	LOG_TEST_RET(card->ctx, rv, "Unlock card failed");
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_finish(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	sc_sm_stop(card);
+
+	free(card->drv_data);
+	card->drv_data = NULL;
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_logout(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	sc_sm_stop(card);
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_pin_cmd_verify(struct sc_card *card, struct sc_pin_cmd_data *data, int *tries_left)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	int rv;
+
+	// Authentication key refers to PACE PIN -> 0x03
+	// Meanwhile, signing key refers to PIN.QES -> 0x81. This pin is verifiable via regular iso7816 cmd verify call.
+	switch (data->pin_reference) {
+	case PACE_PIN_ID_CAN:
+	case PACE_PIN_ID_PIN:
+	case PACE_PIN_ID_PUK:
+		rv = lteid_perform_pace(card, data->pin_reference, data->pin1.data, data->pin1.len, tries_left);
+		break;
+	default:
+		rv = iso_ops->pin_cmd(card, data, tries_left);
+		break;
+	}
+
+	LOG_FUNC_RETURN(card->ctx, rv);
+}
+
+static int
+lteid_pin_cmd_get_info(struct sc_card *card, struct sc_pin_cmd_data *data, int *tries_left)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	struct lteid_drv_data *drv_data = DRVDATA(card);
+	int rv;
+
+	// PACE CAN code info: as far as we know there's no limit to CAN verification attempts
+	if (data->pin_reference == PACE_PIN_ID_CAN) {
+		data->pin1.max_tries = -1;
+		data->pin1.tries_left = -1;
+
+		if (tries_left) {
+			*tries_left = -1;
+		}
+
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+	}
+
+	// PACE PIN and PUK codes: max and remaining attempts are stored in ACE3 and ACE4 files.
+	if (data->pin_reference == PACE_PIN_ID_PIN || data->pin_reference == PACE_PIN_ID_PUK) {
+		struct sc_apdu apdu;
+		unsigned char buf[0xbe] = {0};
+		u8 id[] = {0xac, 0x00};
+		size_t taglen = 0;
+
+		switch (data->pin_reference) {
+		case PACE_PIN_ID_PIN:
+			id[1] = 0xe3;
+			break;
+		case PACE_PIN_ID_PUK:
+			id[1] = 0xe4;
+			break;
+		default:
+			break;
+		}
+
+		sc_format_apdu_ex(&apdu, 0x00, 0xa4, 0x00, 0x04, id, sizeof(id), buf, sizeof(buf));
+
+		rv = sc_transmit_apdu(card, &apdu);
+		LOG_TEST_RET(card->ctx, rv, "APDU transmit failed");
+
+		const u8 *tag = sc_asn1_find_tag(card->ctx, buf, apdu.resplen, 0x62, &taglen);
+		tag = sc_asn1_find_tag(card->ctx, tag, taglen, 0xa5, &taglen);
+		tag = sc_asn1_find_tag(card->ctx, tag, taglen, 0xa2, &taglen);
+		tag = sc_asn1_find_tag(card->ctx, tag, taglen, 0xa3, &taglen);
+		tag = sc_asn1_find_tag(card->ctx, tag, taglen, 0x82, &taglen);
+
+		if (tag && taglen == 2) {
+			data->pin1.tries_left = tag[0];
+			data->pin1.max_tries = tag[1];
+
+			if (tries_left) {
+				*tries_left = data->pin1.tries_left;
+			}
+		} else {
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_OBJECT_NOT_FOUND);
+		}
+
+		if (drv_data->pace_pin_ref == data->pin_reference) {
+			data->pin1.logged_in = SC_PIN_STATE_LOGGED_IN;
+		} else {
+			data->pin1.logged_in = SC_PIN_STATE_LOGGED_OUT;
+		}
+
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+	}
+
+	// PIN.QES is a regular iso7816 pin
+	if (data->pin_reference == 0x81) {
+		rv = iso_ops->pin_cmd(card, data, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_ERROR_OBJECT_NOT_FOUND);
+}
+
+static int
+lteid_pin_cmd_unblock(struct sc_card *card, struct sc_pin_cmd_data *data, int *tries_left)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	int rv;
+
+	// PACE PIN for unblocking command is known as pin ref 0x07
+	if (data->pin_reference == PACE_PIN_ID_PIN) {
+		struct sc_pin_cmd_data with_changed_pin_ref = *data;
+
+		with_changed_pin_ref.pin_reference = 0x07;
+		rv = iso_ops->pin_cmd(card, &with_changed_pin_ref, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	// PIN.QES is a regular iso7816 pin
+	if (data->pin_reference == 0x81) {
+		rv = iso_ops->pin_cmd(card, data, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_ERROR_OBJECT_NOT_FOUND);
+}
+
+static int
+lteid_pin_cmd_change(struct sc_card *card, struct sc_pin_cmd_data *data, int *tries_left)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	int rv;
+
+	// PACE PIN for unblocking command is known as pin ref 0x07
+	if (data->pin_reference == PACE_PIN_ID_PIN) {
+		struct sc_pin_cmd_data with_changed_pin_ref = *data;
+
+		with_changed_pin_ref.pin_reference = 0x07;
+		rv = iso_ops->pin_cmd(card, &with_changed_pin_ref, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	// PIN.QES is a regular iso7816 pin
+	if (data->pin_reference == 0x81) {
+		rv = iso_ops->pin_cmd(card, data, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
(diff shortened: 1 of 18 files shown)


src/tools/Makefile.am
```diff
@@ -30,7 +30,7 @@ if ENABLE_OPENSSL
 bin_PROGRAMS += cryptoflex-tool pkcs15-init netkey-tool piv-tool \
 	westcos-tool sc-hsm-tool dnie-tool gids-tool
 if ENABLE_OPENPACE
-bin_PROGRAMS += npa-tool
+bin_PROGRAMS += npa-tool lteid-tool
```
tadas-s: Not entirely sure about this one - does using `perform_pace` mean my driver depends on openpace?

metsma:
There are some cards that use alse PACE , maybe the CAN setup can inspired from them?

tadas-s:
@metsma 

> There are some cards that use alse PACE , maybe the CAN setup can inspired from them?

dtrust card does something similar (i.e. uses cache as storage for CAN):
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-dtrust.c#L464-L595

The gov provided proprietary software does something bit like this as well - desktop app shares obfuscated CAN with pkcs#11 module via file in users home folder.

But I haven't seen any drivers (ab)using label to inform what needs to be done to get card working ;)

TheComputerGuy96:
Some notes (based on my local driver work):

1. The card does actually have a functional PKCS#15 structure (but it doesn't contain some stuff like the monochrome ID photo) so appending those files to the existing structure would be needed

2. Could the driver share code with other PACE-based drivers (like npa or edo)? I personally used significant bits of code from the npa driver as a base (so it would make a perfect driver to share code with)

tadas-s:
> Some notes (based on my local driver work):
> 
> 1. The card does actually have a functional PKCS#15 structure (but it doesn't contain some stuff like the monochrome ID photo) so appending those files to the existing structure would be needed
>

Yup, all the PKCS#15 struct is there. Photo and personal details are part of the ICAO or something readable documents thing... but at least for now I don't care about that part too much.

> 2. Could the driver share code with other PACE-based drivers (like npa or edo)? I personally used significant bits of code from the npa driver as a base (so it would make a perfect driver to share code with)

Honestly I don't see much to share between those.. `edo` card uses PACE CAN, but rest of pins seem to be "normal" iso7816 api pins. `npa` has PACE CAN, PIN, PUK. And Lithuanian card: PACE CAN, PIN, PUK *and* one regular PIN (aka PIN.QES). With some extra craziness like PACE PIN reference is 0x03, but when resetting or changing - reference is 0x07 :sweat_smile: .

src/tools/Makefile.am
frankmorgner: if lteid-tool.c doesn't explicitly link to OpenPACE, you should add the binary to the normal bin_PROGRAMS, not here. perform_pace will return a "not supported error" if openpace is missing or the reader itself doesn't support pace.

frankmorgner:
we have cards storing the CAN on the filesystem (see dtrust driver), others are using an entry in opensc.conf (e.g. npa driver). It would also be possible to spawn a dedicated PIN request (independant from PKCS#11), which requires capabilities for creating a window (see dnie driver, search for ENABLE_DNIE_UI).

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,446 @@
...
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	struct lteid_drv_data *drv_data = DRVDATA(card);
+	struct establish_pace_channel_input pace_input = {0};
+	struct establish_pace_channel_output pace_output = {0};
+
+	if (drv_data->pace && drv_data->pace_pin_ref != ref) {
+		sc_log(card->ctx, "Re-opening PACE with pin ref 0x%02x. Previous pin ref: 0x%02x.", ref, drv_data->pace_pin_ref);
+		sc_sm_stop(card);
```
frankmorgner, line 193: I'm not sure if you want it exactly this way, but the cards I know are supporting performing PACE inside an existing PACE channel. So if you want to, you do not need to call sc_sm_stop(). This way you would not have to track whether pace is currently executed (drv_data->pace)

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,446 @@
...
+
+	// Authentication key refers to PACE PIN -> 0x3
+	// Meanwhile, signing key refers to PIN.QES -> 0x81. This pin is verifiable via regular iso7816 cmd verify call.
+	if (data->cmd == SC_PIN_CMD_VERIFY && (data->pin_reference == PACE_PIN_ID_PIN)) {
+		rv = lteid_perform_pace(card, data->pin_reference, data->pin1.data, data->pin1.len, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	// PACE CAN code info: there's ACE2 file, but it does not contain any info about max or remaining attempts.
+	if (data->cmd == SC_PIN_CMD_GET_INFO && (data->pin_reference == PACE_PIN_ID_CAN)) {
```
frankmorgner: all cards I know allow an infinite number of attempts for CAN

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,446 @@
...
+		}
+
+		data->pin1.logged_in = (drv_data->pace_pin_ref == data->pin_reference);
+
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+	}
+
+	// Any other commands - fall back to regular iso7816 methods.
+	// Mostly for PIN.QES(ID 0x81) which is a regular pin.
+	rv = iso_ops->pin_cmd(card, data, tries_left);
```
frankmorgner: did you test SC_PIN_CMD_UNBLOCK or SC_PIN_CMD_CHANGE, e.g. with PACE_PIN_ID_PIN? your current code will use the commands from iso7816.c for this...

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,446 @@
...
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INTERNAL);
+
+	sc_log(card->ctx, "algo: %lu operation: %d keyref: %d", env->algorithm, env->operation, env->key_ref[0]);
+
+	if (env->algorithm == SC_ALGORITHM_EC && env->operation == SC_SEC_OPERATION_SIGN) {
+		const u8 data[] = {0x84, 0x01, env->key_ref[0]};
+		sc_format_apdu_ex(&apdu, 0x00, 0x22, 0x41, 0xB6, data, sizeof(data), NULL, 0);
+	} else {
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+	}
```
frankmorgner:
```suggestion
	if (env->algorithm != SC_ALGORITHM_EC || env->operation != SC_SEC_OPERATION_SIGN)
		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);

	const u8 data[] = {0x84, 0x01, env->key_ref[0]};
	sc_format_apdu_ex(&apdu, 0x00, 0x22, 0x41, 0xB6, data, sizeof(data), NULL, 0);
```
currently, `data` goes out of scope before using the buffer. The above suggestion fixes this behavior.

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,446 @@
...
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	int rv = iso_ops->process_fci(card, file, buf, buflen);
+
+	if (rv != SC_SUCCESS) {
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	// Card does not report most of the file sizes. So if it's zero -
```
frankmorgner: If I remember correctly from the discussion, then your card actually does report the file size, but it reports it as exactly `0` bytes

src/libopensc/pkcs15-lteid.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+static int
+sc_pkcs15emu_lteid_dummy_init(struct sc_pkcs15_card *p15card, struct sc_aid *aid)
+{
+	LOG_FUNC_CALLED(p15card->card->ctx);
+
+	// If a valid PACE CAN code is not available - construct a dummy pkcs15 structure
+	// to help user with the initial set-up process.
+	set_string(&p15card->tokeninfo->label, "!! Run `lteid-tool` to set up !!");
```
frankmorgner: Likely, this will confuse many users. I suggest you simply fail the token initialization to signal an error. If you want a better user experience, I assume you would have to request the CAN with a dedicated PIN prompt.

src/libopensc/card-lteid.c
tadas-s: This one is bit weird. If PACE session is started with CAN and then switched over to PACE session with PIN - I'm getting "not allowed" error:

```
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] sm-eac.c:999:perform_pace: returning with: -1209 (Not allowed)
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] card-lteid.c:180:lteid_perform_pace: Error performing PACE for pin ref 0x03.
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] card-lteid.c:199:lteid_perform_pace: returning with: -1209 (Not allowed)
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] card-lteid.c:291:lteid_pin_cmd: returning with: -1209 (Not allowed)
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] sec.c:259:sc_pin_cmd: returning with: -1209 (Not allowed)
P:1632610; T:0x140703871154240 16:21:17.228 [opensc-pkcs11] pkcs15-pin.c:443:sc_pkcs15_verify_pin_with_session_pin: PIN cmd result -1209
```

Command was:

```
$ ./bin/pkcs11-tool --pin <...skipped...> --sign -i ./test.in -o test.bin
Aborting.
```

Full log:
[opensc-debug-sign-without-sm-close.log](https://github.com/user-attachments/files/25444142/opensc-debug-sign-without-sm-close.log)

There's one exception though - when resuming PACE-PIN use after 2 incorrect attempts - I have to establish PACE CAN immediately followed by PACE PIN session to work.

All found after all kinds of trial and error.

src/libopensc/pkcs15-lteid.c
tadas-s: I'll get rid of this.

Looked at the dnie UI, but I'm not sure if it's good idea. Blocking and waiting for input from pkcs#11 module causes Firefox to lock up and Gnome starts suggesting to force close it. Also, if p11-kit loads opensc module system wide - it gets even worse. Gnome UI may freeze / get stuck.

Early adopters (if there's any :sweat_smile:) will have to follow docs or some guide to set up CAN via opensc.conf OR env variable OR cache via `lteid-tool`.

If this gets picked later by more regular users - maybe I'll be able to produce some tray sitting app. I.e. something that listens for card insert events and if card is lteid - shows popup to enter CAN .. Or something like that.

src/libopensc/card-lteid.c
tadas-s: Clarified the comment, hope it makes more sense.

src/libopensc/pkcs15-lteid.c
frankmorgner: > set up CAN via opensc.conf OR env variable OR cache via lteid-tool.

That's definately a good choice to start with. You could spawn a new thread/process with a UI request for the CAN right on initialization (non-blocking for the actual application). Then, when the user enters the CAN, cache it via the file cache. This way it is retreavable when the PKCS#11 application actually wants to verify the PIN.

Sorry, if you feel overwhelmed with all the choices, but I'm thinking about the corner cases with CAN+PIN already for some time. Unfortunately, I never had the time to implement this 😇. If you think this is too much for you either, just leave it with the initial setup as described above (most users have a very static setup anyway)!

src/libopensc/card-lteid.c
frankmorgner: OK, so I'll leave this up to you.

TheComputerGuy96:
@tadas-s I uploaded my initial (garbage) code that got my Lithuanian MaskTech card working a few months ago (with QES too) here: https://github.com/TheComputerGuy96/OpenSC/tree/lteid-garbage (so hopefully some parts could be useful)

Also your ATR seems strange (can you tell when the card was issued?); both of the E-stonian eID frameworks (libelectronic-id and DigiDoc's own thing) don't have yours listed: https://github.com/web-eid/libelectronic-id/blob/affadb37f3db86e211e06f7fcaa1ddc3169f0de4/src/electronic-id.cpp#L80-L87

(And for reference, mine was issued in July 2025 with the LitEID v2.0 ATR mentioned in the code above)

tadas-s:
> @tadas-s I uploaded my initial (garbage) code that got my Lithuanian MaskTech card working a few months ago (with QES too) here: https://github.com/TheComputerGuy96/OpenSC/tree/lteid-garbage (so hopefully some parts could be useful)

Thanks, I'll check it out.

> Also your ATR seems strange (can you tell when the card was issued?); both of the E-stonian eID frameworks (libelectronic-id and DigiDoc's own thing) don't have yours listed: https://github.com/web-eid/libelectronic-id/blob/affadb37f3db86e211e06f7fcaa1ddc3169f0de4/src/electronic-id.cpp#L80-L87
> 
> (And for reference, mine was issued in July 2025 with the LitEID v2.0 ATR mentioned in the code above)

My card was issued on Jan 2026.

Looks like there's no single ATR, but rather some range. Here's what [this atr parser][1] shows for both of our ATRs side by side:
image

Looks like we'll have to collect more examples and figure out a matching mask.

edit: there's also list of cards - https://pcsc-tools.apdu.fr/smartcard_list.txt .. Search `3B 9D 18 81 31 FC 35 80 31 C0 69 4D 54 43 4F 53 73 02`:

image

Last byte is checksum. It almost looks like mask could be `ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:ff`, but it catches "Serbian vehicle card (Other)" .. I guess we'll just compile a list.

[1]: https://smartcard-atr.apdu.fr/

metsma:
> My card was issued on Jan 2026.
> 
> Looks like there's no single ATR, but rather some range. Here's what [this atr parser](https://smartcard-atr.apdu.fr/) shows for both of our ATRs side by side: image
> 
> Looks like we'll have to collect more examples and figure out a matching mask.

Seems like the version number has changed 2.6.4 -> 2.6.5

tadas-s:
> 
> Seems like the version number has changed 2.6.4 -> 2.6.5

Right! I searched for "mtcos" and there's plenty pages mentioning "MTCOS Pro 2.6". It really is version, and not something unique to Lithuanian ID.

edit:

Unless I'm missing something -

- Can't use ATR based detection. OpenSC is used worldwide and we may end up with confused users asking why their smartcard is identified as Lithuanian ID card.
- Can't claim it's "MTCOS driver" because my driver is not generic at all.
- Can't really query card before PACE is established.

So end users will have to edit /etc/opensc.conf and configure their own card ATR to use `lteid` driver. And that I'll have to document in `opensc.conf` man page.

edit: detection by ATR removed, card works fine with this config bit:
```
    card_atr 3b:9d:18:81:31:fc:35:80:31:c0:69:4d:54:43:4f:53:73:02:00:00:00 {
        atrmask = ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00:00;
        driver = lteid;
        type = 43000;
    }
```

src/libopensc/card-lteid.c
tadas-s: It doesn't. But I can change or unblock PACE PIN by using pin reference 0x07 (wich I'm doing from `lteid-tool`). I should update that comment.

I could also try to remap pin cmds using pin ref PACE_PIN_ID_PIN to use reference 0x07. Can't tell yet if it would work or if it's good idea at all.

src/libopensc/card-lteid.c
```diff
@@ -36,26 +36,12 @@ static struct sc_card_driver lteid_drv = {
...
 
-static int
-lteid_match_card(sc_card_t *card)
-{
-	if (_sc_match_atr(card, lteid_atrs, &card->type) >= 0) {
-		return 1;
-	}
-	return 0;
-}
-
```
TheComputerGuy96: I'm personally NAK'ing the card detection removal change until at least OpenSC gains user-specific configuration file support (ideally `lteid-tool` would have an option to enable the driver for the currently attached card because that program is already present for inputting the CAN; so it wouldn't harm to have more functionality in it)

tadas-s: I think it should be fine. Further down the line, if this picks up at all, we'll need some LT eID specific package which depends on `opensc` and makes all the necessary config. Some sort of gui, docs, etc.. Bit like `qdigidoc`.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,629 @@
...
+{
+	int rv;
+	char path[PATH_MAX];
+
+	sc_get_cache_dir(card->ctx, path, sizeof(path));
+	strcat(path, CAN_STORE_FILE);
+
+	unlink(path);
+
+	FILE *fd = fopen(path, "w");
```
Jakuje: nit: I think unlink is not necessary. Opening for writing truncates potential existing file.

I think for the cases where the opening file for writing will fail, also the unlink will fail so I am not sure what specific use case this operation would solve.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,629 @@
...
+char *
+lteid_get_stored_can(sc_card_t *card)
+{
+	char path[PATH_MAX];
+	char *can;
+
+	can = calloc(7, 1);
+
+	sc_get_cache_dir(card->ctx, path, sizeof(path));
+	strcat(path, CAN_STORE_FILE);
```
Jakuje: I did remove all uses of strcat in #3554. I would prefer not to introduce new ones now and use strlcat instead to make sure the bounds are checked.
Its just a magnet for issues.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,629 @@
...
+
+	if (rv != SC_SUCCESS) {
+		fprintf(stderr, "PIN number blank, too short or too long.\n");
+		return SC_ERROR_PIN_CODE_INCORRECT;
+	}
+
+	input_number("New PIN (repeat)", 6, 12, NULL, &new_pin_repeated);
+
+	if (new_pin_repeated == NULL || strcmp(new_pin, new_pin_repeated) != 0) {
+		fprintf(stderr, "New PIN and repated entry do not match. PIN was not changed.\n");
```
Jakuje:
```suggestion
		fprintf(stderr, "New PIN and repeated entry do not match. PIN was not changed.\n");
```

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,437 @@
...
+
+	card->max_send_size = 65535;
+	card->max_recv_size = 65535;
+	card->caps |= SC_CARD_CAP_ISO7816_PIN_INFO | SC_CARD_CAP_APDU_EXT;
+
+	_sc_card_add_ec_alg(card, 384, SC_ALGORITHM_ECDSA_HASH_NONE | SC_ALGORITHM_ECDSA_RAW, 0, NULL);
+
+	LOG_TEST_RET(card->ctx, sc_enum_apps(card), "Enumerate apps failed");
+
+	LOG_TEST_RET(card->ctx, lteid_unlock(card), "Unlock card failed");
```
Jakuje: While the macros `LOG_TEST_RET` are constructed to support this, I find it very hard to read and follow so I would prefer to break it to separate lines with temporary variable to keep it more readable

```suggestion
    rc = sc_enum_apps(card);
	LOG_TEST_RET(card->ctx, rc, "Enumerate apps failed");

    rc = lteid_unlock(card);
	LOG_TEST_RET(card->ctx, rc, "Unlock card failed");
```

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,437 @@
...
+			break;
+		default:
+			break;
+		}
+
+		sc_format_apdu_ex(&apdu, 0x00, 0xa4, 0x00, 0x04, id, sizeof(id), buf, sizeof(buf));
+
+		LOG_TEST_RET(card->ctx, sc_transmit_apdu(card, &apdu), "APDU transmit failed");
+
+		const u8 *tag = sc_asn1_find_tag(card->ctx, buf, sizeof(buf), 0x62, &taglen);
```
Jakuje: I think we should use the `resplen` to make sure we do not read the empty data behind what was returned by the card:
```suggestion
		const u8 *tag = sc_asn1_find_tag(card->ctx, buf, apdu.resplen, 0x62, &taglen);
```

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,588 @@
...
+		rv = lteid_pin_cmd_unblock(card, data, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	if (data->cmd == SC_PIN_CMD_CHANGE) {
+		rv = lteid_pin_cmd_change(card, data, tries_left);
+		LOG_FUNC_RETURN(card->ctx, rv);
+	}
+
+	LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
```
TheComputerGuy96: `switch` cases would be a perfect idea here

tadas-s: Yup :)

src/libopensc/card-lteid.c
tadas-s: So update on this - I've changed driver so now SC_PIN_CMD_UNBLOCK and SC_PIN_CMD_CHANGE are available via driver using PACE_PIN_ID_PIN reference. Under the hood driver remaps PACE_PIN_ID_PIN 0x03 to 0x07 pin ref.

tadas-s:
Hi @frankmorgner  , could you clarify this comment please?

> Your setting may not work in a multi user environment and it uses an unexpected location for cached data.

I'm placing `lteid_can` file in OpenSC folder which I get from `sc_get_cache_dir`.

frankmorgner:
Basically, I am suggesting to use a generated file caching path for CAN as done in card-dtrust:

Read:
https://github.com/OpenSC/OpenSC/blob/c8b09077e0a6377ebfdac93a4b8da24a45699d3f/src/libopensc/card-dtrust.c#L487-L498

Write:
https://github.com/OpenSC/OpenSC/blob/c8b09077e0a6377ebfdac93a4b8da24a45699d3f/src/libopensc/card-dtrust.c#L588-L592

The caching path would be card dependant (includes serial number) and the logic doesn't require special logic in lteid-tool (verifying CAN is enough for the driver to cache this transperently).

tadas-s:
The trouble with that is I cannot read `EF.CIAInfo` (`df02/5032`) without unlocking card using CAN first. 

So if I stored it using card serial number as cache key prefix - I won't be able to look it up when card is inserted next time.

Screenshot From 2026-03-02 12-23-46

And I can't seem to find any other unique identifiers which are available before card is unlocked using CAN.

Proprietary software works pretty much the same way - it only remembers last used CAN code and it cannot deal with more than one card without re-entering CAN each time you switch (tested it myself - I have two ID copies).

The second part I've implemented now - after successful PIN_CMD_VERIFY for pin ref 0x02 + PIN given by the caller
-> it gets cached.

frankmorgner:
I was not aware that your card doesn't have some globally readable static identifier attached to it. Then, I believe, your effort is good enough, thank you for looking into this.

(Did you check `card->reader->uid` if it stays static? Most cards have a static UID and some readers support reading that from PC/SC. In any, UID as static identifier is a very limited option for a static identifier.)

tadas-s:
Yup, added this at the end of `lteid_init`:

image

And it logs:

image

It's empty.

frankmorgner:
You would need to switch on `enable_escape = true` in `opensc.conf`, but, as I said, this would only be a very limited solution. With your observation about the missing static identifier, I think that is resolved.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,563 @@
...
+	int opt_verify_can = 0;
+
+	int err = 0;
+	sc_context_t *ctx = NULL;
+	sc_context_param_t ctx_param;
+	sc_card_t *card = NULL;
+	struct sc_pkcs15_card *p15card = NULL;
+	int c, rv;
+
+	while ((c = getopt_long(argc, argv, "hr:wc:p:u:vCRU", options, (int *)0)) != -1) {
```
TheComputerGuy96: The `-V` option isn't handled here

TheComputerGuy96:
Anyway my slightly older card works with online authentication and is also even QES'ing without issues (so besides the card detection situation, the driver should be fine to ship)

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,586 @@
...
+			free(blocks);
+		}
+
+		if (can_from_config && strlen(can_from_config) == LTEID_CAN_LENGTH) {
+			got_can = 1;
+			memcpy(drv_data->can, can_from_config, LTEID_CAN_LENGTH);
+		}
+	}
+
+	if (!got_can) {
```
frankmorgner, line 171:
```suggestion
	if (!got_can && !(card->reader->capabilities & SC_READER_CAP_PACE_GENERIC)) {
```
Input at the reader side is also a possibility...

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,586 @@
...
+		sc_log(card->ctx, "Missing or invalid CAN. 6 digits required.");
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN);
+	}
+
+	pace_input->pin_id = PACE_PIN_ID_CAN;
+	pace_input->pin = drv_data->can;
+	pace_input->pin_length = LTEID_CAN_LENGTH;
```
frankmorgner, line 178:
```suggestion
	pace_input->pin_length = got_can ? LTEID_CAN_LENGTH : 0;
```

tadas-s:
After initial review I was mostly piling commits on top for reviewers to see what's changing.

For final review I can squash/reword so everything is in a nice single unit. Shall I do it now @frankmorgner ?

frankmorgner:
Do you want to support card readers with a pin pad that are supporting pace?

tadas-s:
> Do you want to support card readers with a pin pad that are supporting pace?

Oh that's whole another topic. Generally government does not encourage using secure pin entry via readers with pinpads. And I can't seem to find any for purchase locally, especially PACE compatible. So it's not a goal of this first revision, but I do plan to look at it some time later.

src/libopensc/card-lteid.c
tadas-s: I'd almost argue against this until I get my hands on a pace capable reader and can verify it works with the rest of the stuff.

TheComputerGuy96:
@tadas-s 

> Generally government does not encourage using secure pin entry via readers with pinpads.

At least a special reader isn't needed for even QES operations (unlike the CAT-K requirement in the German nPA card) so I can buy an arbitrary card reader and it will just work

> And I can't seem to find any for purchase locally

Technically I found this: https://www.varle.lt/bruksniniu-kodu-barkodu-skaitytuvai/skaitytuvas-acs-usb-viso-dydzio-contact-smart--26265648.html (but it's quite expensive and I don't even know if it's EAC/PACE compatible)

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,586 @@
...
+#endif
+
+static int
+lteid_get_cached_can(sc_card_t *card, unsigned char *can)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+	char path[PATH_MAX];
+
+	sc_get_cache_dir(card->ctx, path, sizeof(path));
+	strlcat(path, CAN_CACHE_FILE, PATH_MAX);
```
Jakuje: nit: This is repeated in the following three functions. Would it make sense to extract it to some helper function `lteid_get_cach_path()` or similar?

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,586 @@
...
+
+	// Track PACE status
+	drv_data->pace = 1;
+	drv_data->pace_pin_ref = ref;
+
+	free(pace_output.ef_cardaccess);
+	free(pace_output.recent_car);
+	free(pace_output.previous_car);
+	free(pace_output.id_icc);
+	free(pace_output.id_pcd);
```
Jakuje, line 245: This is probably comment to general use of PACE than to this triver as this is present in other drivers too. Would it make sense to introduce some `clear_pace_output()` helper to avoid the need to free separate fields like this in each of the drivers? This could be addressed in follow-up PR

src/libopensc/card-lteid.c
```diff
@@ -0,0 +1,586 @@
...
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+lteid_unlock(sc_card_t *card)
+{
+	SC_FUNC_CALLED(card->ctx, SC_LOG_DEBUG_VERBOSE);
+
+	if (SC_SUCCESS != lteid_perform_pace(card, PACE_PIN_ID_CAN, NULL, 0, NULL)) {
+		sc_log(card->ctx, "Unlock with CAN code failed. No CAN found in environment, opensc.conf or cache.");
```
Jakuje: spelling nit:
```suggestion
		sc_log(card->ctx, "Unlock with CAN code failed. No CAN found in environment, opensc.conf nor cache.");
```

src/libopensc/pkcs15-lteid.c
```diff
@@ -0,0 +1,76 @@
...
+	if (df->enumerated)
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+
+	rv = sc_pkcs15_parse_df(p15card, df);
+	LOG_TEST_RET(ctx, rv, "DF parse error");
+
+	if (df->type != SC_PKCS15_PRKDF)
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+
+	rv = sc_pkcs15_get_objects(p15card, SC_PKCS15_TYPE_PRKEY, pkobjs, sizeof(pkobjs) / sizeof(pkobjs[0]));
```
Jakuje, line 45: we have several macros computing this throughout the code, for example
```
#define DIM(v)           (sizeof(v)/(sizeof((v)[0])))
```
or
```
#define NELEMENTS(x)    (sizeof(x)/sizeof((x)[0]))
```
but on short search, I did not find any in headers that could be readily reused.  So this might be something for follow-up PR to unify this throughout the code.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,563 @@
...
+	if (rv != SC_SUCCESS) {
+		fprintf(stderr, "CAN code verification failed: %s\n", sc_strerror(rv));
+		return SC_ERROR_PIN_CODE_INCORRECT;
+	}
+
+	pace_input.pin_id = PACE_PIN_ID_PIN;
+	pace_input.pin = (unsigned char *)pin;
+	pace_input.pin_length = strlen(pin);
+
+	rv = perform_pace(card, pace_input, &pace_output, EAC_TR_VERSION_2_02);
```
Jakuje, line 360: Do we need to free the members of `pace_output` after this as we do in other cases?

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,563 @@
...
+		case 'V':
+			opt_verify_can = 1;
+			break;
+		case 'h':
+		default:
+			util_print_usage_and_die(app_name, options, option_help, NULL);
+		}
+	}
+
+	memset(&ctx_param, 0, sizeof(ctx_param));
```
Jakuje: I would prefer to use the C initializer instead of the memset:
```
	sc_context_param_t ctx_param = {0};
```
using memset just causes unnecessary code.

src/tools/lteid-tool.c
```diff
@@ -0,0 +1,563 @@
...
+
+	if (opt_change_pin) {
+		err = change_pin(card, opt_pin);
+	} else if (opt_resume) {
+		err = resume(p15card, opt_can, opt_pin);
+	} else if (opt_unblock) {
+		err = unblock_using_puk(p15card, opt_puk);
+	}
+
+lteid_tool_end:
```
Jakuje: nit: I would just rename the label `end`/`cleanup` or something shorter to simplify it, given that is just the only label around here.

src/tools/lteid-tool.c
frankmorgner: no. the underlying code should reuse the memory and may even avoid read out of card files.

tadas-s:
Hi,
A few things got merged into master and now there's a small conflict.. Shall I squash/rebase/reword this PR?

Jakuje:
> Hi, A few things got merged into master and now there's a small conflict.. Shall I squash/rebase/reword this PR?

Sorry for that. We did the release so please rebase and then we can get this merged.

tadas-s:
Hi @Jakuje 
No problem at all. I've rebased everything+squashed+reworded, did some quick testing. As far as I'm concerned it should be good to go :slightly_smiling_face:

Jakuje:
No more comments from my side so lets get it merged! Thank you for your contribution!

Issue #3581, opened 2026-02-24
The release tarballs from the tag result in wrong tarballs


Jakuje:
### Describe the bug

The tarballs created in the master branch result in filenames `opensc-0.27.0rc1.tar.gz` (missing dash before -rc).

https://github.com/OpenSC/OpenSC/actions/runs/22353427326

But it looks like I messed this up in VERSION.mk in #3571. Will fix later

But tarballs generated from the 0.27.0-rc1 tag result in two different tarballs generated
 * opensc-0.27.00.27.0-rc1.tar.gz
 * opensc-0.27.0rc1.tar.gz

both of them are uploaded to the artifacts of the GH Actions.

https://github.com/OpenSC/OpenSC/actions/runs/22353524910

Seems like this is caused by this part where the tag is considered as a branch from `$GITHUB_REF`, which causes  the bootstrap called with the whole suffix:
```
+ ./bootstrap.ci -s 0.27.0-rc1
```
and configure getting the double-version:
```
Version:                 0.27.00.27.0-rc1
```

https://github.com/OpenSC/OpenSC/blob/master/.github/build.sh#L19-L21

### Steps to reproduce

inspect GH actions CI results

### Expected behavior

single version should be appended to the package name

### Actual behavior

version is included twice

### OpenSC version

master

### Operating system

GH Actions

### Logs and debug output

```shell

```

### Checklist

- [x] I am using a supported OpenSC version
- [x] I have searched existing issues to avoid duplicates

Jakuje:
Actually, the MacOS tarball is missing the rc1 suffix too:

https://github.com/OpenSC/Nightly/tree/2026-02-24_c7b3d416

This is also something we should fix for the final release

metsma:
Same with windows build OpenSC-0.27.0_arm64-0.27.0-rc1

Jakuje:
> Same with windows build OpenSC-0.27.0_arm64-0.27.0-rc1

I see. The names are correct in appveyor and nightly repo, but the artifacts from GH Actions have wrong names too.  Do you want to fix that or should I look into that? I guess we should do rc2 to make sure we get this right :)

PR #3582, opened 2026-02-24
macos: Include package suffix in package version

Jakuje:
Fixes: #3581

##### Checklist

- [ ] macOS token is tested


Jakuje:
Thanks for the review! Any thoughts on #3581 and the version suffix being pulled from the tag?

PR #3584, opened 2026-02-25
sc_dlsym: ignore symbols defined in dependent shared libs

### LINKED ISSUE #3580: pkcs11-spy: C_LoadModule loads a PKCS11 3.0 module instead of the specified PKCS11 2.0 module
### Describe the bug

I've been trying to debug a mysterious failure in [glib-networking](https://gitlab.gnome.org/GNOME/glib-networking)'s test suite. The error is:

```
**
GLib-Net:ERROR:../../../../../glib-networking/tls/tests/connection.c:1373:test_client_auth_pkcs11_connection: assertion failed (error == NULL): Could not import PKCS #11 certificate URI: The requested data were not available. (g-tls-error-quark, 2)
not ok /tls/gnutls/connection/client-auth-pkcs11 - GLib-Net:ERROR:../../../../../glib-networking/tls/tests/connection.c:1373:test_client_auth_pkcs11_connection: assertion failed (error == NULL): Could not import PKCS #11 certificate URI: The requested data were not available. (g-tls-error-quark, 2)
```

This error occurs on Fedora 44 and other recent versions of Fedora, but not on glib-networking's CI, which is stuck on a very old version of Fedora (probably Fedora 36?). I'm trying to update the CI to a newer image. Suffice to say this is a regression introduced sometime in the past several years. This error indicates that glib-networking's mock-pkcs11 module is not working properly, but it notably only occurs if using pkcs11-spy: the tests succeed if I uninstall Fedora's opensc-libs package, or if I sabotage glib-networking's code that loads pkcs11-spy (which is in tls/tests/connection.c, in `main()`).

After some debugging, I eventually discovered the problem: opensc's `C_LoadModule` is broken when trying to load a module that does not implement `C_GetInterface`, if any other module does implement `C_GetInterface`. The problem is `dlsym()` apparently does not restrict itself to loading symbols only from the library specified by `mod->handle`. glib-networking's mock-pkcs11.so is old and does not implement `C_GetInterface`. In practice, even though `mod->handle` corresponds to glib-networking's mock-pkcs11.so, `dlsym()` sadly loads the `C_GetInterface` from p11-kit, so it returns p11-kit's interface and opensc winds up spying on the p11-kit module by mistake. Now mock-pkcs11.so is not used at all, then the glib-networking tests fail because the mock certificates are unavailable.

This looks similar but not identical to #2875. I am about to try testing the fix for that issue just in case it makes a difference, but I don't think it should make any difference.

I do not know how to fix this. My plan is to work around it by implementing `C_GetInterface` in mock-pkcs11.so, which shouldn't be required.

### Steps to reproduce

1. Clone glib-networking repo
2. `mkdir _build && cd _build`
3. `meson setup ..`, `meson compile`, `meson test`

### Expected behavior

Test suite passes

### Actual behavior

```
GLib-Net:ERROR:../../../../../glib-networking/tls/tests/connection.c:1373:test_client_auth_pkcs11_connection: assertion failed (error == NULL): Could not import PKCS #11 certificate URI: The requested data were not available. (g-tls-error-quark, 2)
```

### OpenSC version

0.26.1-6.fc44

### Operating system

Fedora 44

glib-networking's mock-pkcs11

glib-networking testsuite

### Logs and debug output

Using gdb, I can see it is called with the correct path to glib-networking's mock-pkcs11.so:

```
(gdb) s
C_LoadModule (mspec=0x44de1a "/home/mcatanzaro/.cache/jhbuild/build/glib-networking/tls/tests/mock-pkcs11.so", funcs=0x7fffffffd4c8)
    at ../common/libpkcs11.c:53
53		mod = calloc(1, sizeof(*mod));
```

Next, it calls `sc_dlsym()` to load `C_GetInterface`. If I step some more, I see that `c_get_interface` gets initialized (which should not happen), and step into it:

```
(gdb) n
75			rv = c_get_interface((CK_UTF8CHAR_PTR) "PKCS 11", NULL, &interface, 0);
(gdb) s
C_GetInterface (pInterfaceName=0x7ffff78c7607 "PKCS 11", pVersion=0x0, ppInterface=0x7fffffffd4d0, flags=0) at ../p11-kit/proxy.c:891
891	{
(gdb) c
Continuing.
Loaded: "/home/mcatanzaro/.cache/jhbuild/build/glib-networking/tls/tests/mock-pkcs11.so"

0: C_GetInterface
P:195153; T:0x140737352653440 2026-02-19 10:24:02.325
[compat]
[in] pInterfaceName 00007ffff7352db0 / 7
    00000000  50 4B 43 53 20 31 31                             PKCS 11         
[in] pVersion = NULL
[in] flags = 
Returned:  0 CKR_OK

1: C_Initialize
P:195153; T:0x140737352653440 2026-02-19 10:24:02.325
[in] pInitArgs = 0x7fffffffd5b0
     flags: 3
       CKF_LIBRARY_CANT_CREATE_OS_THREADS
       CKF_OS_LOCKING_OK
Returned:  0 CKR_OK

2: C_GetInfo
P:195153; T:0x140737352653440 2026-02-19 10:24:02.390
[out] pInfo: 
      cryptokiVersion:         3.2
      manufacturerID:         'PKCS#11 Kit                     '
      flags:                   0
      libraryDescription:     'PKCS#11 Kit Proxy Module        '
      libraryVersion:          1.1
Returned:  0 CKR_OK
# Start of tls tests
# End of tls tests
1..0
```

You can see it prints that it is loading mock-pkcs11.so, but it's clear that it actually loads p11-kit-proxy.so.

### Checklist

ueno:
When a shared library is loaded with dlopen, dlsym looks for all symbols defined in the shared libraries linked to it. This behavior is sometimes problematic when loading a PKCS#11 module linked to p11-kit, which exposes functions whose names clash with the standard PKCS#11 functions, such as C_GetInterfaceList.

To avoid that, this patch tries to limit the lookup only to the symbols defined directly in the dlopen'ed shared library, using dladdr and dlinfo defined in glibc.

Fixes: #3580

##### Checklist

- [ ] macOS token is tested


configure.ac
```diff
@@ -23,6 +23,7 @@ AC_CONFIG_AUX_DIR([.])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIR([m4])
 AM_INIT_AUTOMAKE(foreign 1.10 [subdir-objects])
+AC_USE_SYSTEM_EXTENSIONS
```
mcatanzaro, line 24: I think this must be what's causing the CI failure:

```
In file included from fuzz_pkcs15_tool.c:44:
../../../src/tools/pkcs15-tool.c:26:9: error: "_XOPEN_SOURCE" redefined [-Werror]
   26 | #define _XOPEN_SOURCE 500
      |         ^~~~~~~~~~~~~
In file included from /usr/include/bits/libc-header-start.h:33,
                 from /usr/include/stdlib.h:26,
                 from fuzz_pkcs15_tool.c:32:
/usr/include/features.h:216:10: note: this is the location of the previous definition
  216 | # define _XOPEN_SOURCE  700
      |          ^~~~~~~~~~~~~
```

mcatanzaro:
I will test this soon to see whether it works in practice, but I think it should.

src/common/libscdl.c
```diff
@@ -67,19 +67,70 @@ int sc_dlclose(void *handle)
...
-			| RTLD_DEEPBIND
+							    | RTLD_DEEPBIND
+#endif
+	);
+	if (!dlhandle->handle) {
+		free(dlhandle);
+		return NULL;
+	}
+
+#if defined(HAVE_DLINFO) && defined(RTLD_DI_LINKMAP)
```
mcatanzaro: This doesn't work with glibc because `RTLD_DI_LINKMAP` is an enum member, not a preprocessor define. Since `dlinfo()` is a GNU extension, I suppose it's probably safe to assume `RT_DI_LINKMAP` exists if `HAVE_DLINFO` is 1.

src/common/libscdl.c
```diff
@@ -67,19 +67,70 @@ int sc_dlclose(void *handle)
...
+	struct link_map *lm;
+
+	if (dlinfo(dlhandle->handle, RTLD_DI_LINKMAP, &lm) < 0) {
+		dlclose(dlhandle->handle);
+		free(dlhandle);
+		return NULL;
+	}
+	dlhandle->address = lm->l_addr;
```
mcatanzaro:
```
libscdl.c: In function ‘sc_dlopen’:
libscdl.c:90:27: error: assignment to ‘void *’ from ‘Elf64_Addr’ {aka ‘long unsigned int’} makes pointer from integer without a cast [-Wint-conversion]
   90 |         dlhandle->address = lm->l_addr;
      |                           ^
```

We can placate -Werror=int-conversion by adding an explicit cast:


```suggestion
	dlhandle->address = (void *) lm->l_addr;
```

This actually fixes the glib-networking tests on my system, so I _guess_ it's correct? But I'm not sure how you figured it out. `dlinfo(3)` says:

```
       RTLD_DI_LINKMAP (struct link_map **)
              Obtain  a  pointer  to  the  link_map structure corresponding to handle.  The info argument points to a pointer to a link_map
              structure, defined in <link.h> as:

                  struct link_map {
                      ElfW(Addr) l_addr;  /* Difference between the
                                             address in the ELF file and
                                             the address in memory */
```

"Difference between the address in the ELF file and the address in memory"... casting a "difference" directly to a pointer seems _really_ sketchy. And yet, it seems to work, so perhaps I shouldn't complain!

src/common/libscdl.c
mcatanzaro:
```suggestion
#if defined(HAVE_DLINFO)
```

ueno: Good catch, thanks.

src/common/libscdl.c
ueno: According to the [manual](https://sourceware.org/glibc/manual/2.43/html_node/Dynamic-Linker-Introspection.html):
```
l_addr

    This field contains the load address of the object. This is the offset that needs to be applied to unrelocated addresses in the object image (as stored on disk) to form an address that can be used at run time for accessing data or running code.
```
The following sentences are bit worrying though.

Jakuje: Let me commit the change to see how it works in the limited CI environment we have. We have at least one data point that it fixes the glib-networking CI, which is a good thing.

Jakuje:
This sounds like working ok in  our CI  (though we do not run much tests on Mac/Windows, I think). But sounds like working ok with all pkcs11-spy, kryoptic and sofhtms (which does not have the 3.0 API). I am though a bit undecided if we want to include this in the next release. If the fire (glib-networking CI) can be put down by implementing the 3.0 API, this can land after the release and we can give it enough testing. Thoughts?

mcatanzaro:
> If the fire (glib-networking CI) can be put down by implementing the 3.0 API

Already done, so no urgency from my side.

src/common/libscdl.c
mcatanzaro: Staring at the man pages, I thought that checking `dli_fname` from `struct Dl_info`, returned by `dladdr()`, might be a good alternative. But that one has another scary warning at the bottom of `dladdr(3)` man page.

frankmorgner:
As a different workarount has already been implemented in glib-networking's mock-pkcs11, I would prefer to wait with this bugfix.

dengert:
In case this PR or the glib-networking does not work, could opensc -spy linked  with -Bsymbolic?

-Bsymbolic
When creating a shared library, bind references to global symbols to the definition within the shared library, if any. Normally, it is possible for a program linked against a shared library to override the definition within the shared library. This option is only meaningful on ELF platforms which support shared libraries.

I remember seeing something like this with Kerberos modules and PAM 30 years ago.

Jakuje:
If I understand it correctly, the pkcs11-spy linking is not the issue. The issue is the glib-networking mock module linking against gnutls, which brings in p11-kit proxy C_GetInterface. The pkcs11-spy is just the "application" hitting this issue, but it would been hit by any other application dlopening this mock module and going according to the pkcs11 3.x specs.

PR #3585, opened 2026-02-25
Normalize windows package name and strip version number from git tag

metsma:
#3581

##### Checklist

- [ ] macOS token is tested


.github/workflows/windows.yml
```diff
@@ -38,19 +38,25 @@ jobs:
...
         run: |
           if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_name }}" == "master" ]]; then
             SUFFIX=""
           elif [[ "${{ github.event_name }}" == "push" ]]; then
-            SUFFIX="-${{ github.ref_name }}"
+            if [[ "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
+              SUFFIX="${BASH_REMATCH[1]}"
+            else
+              SUFFIX="-${{ github.ref_name }}"
+            fi
```
Jakuje: I think the ref_name refers to the branch name *or* tag name so for pushing release tags, this will result in suffix being `-0.27.0-rc1` which is not correct. I think we need to distinguish the tag and branch push with the `github.ref_type` .

The same will need to be fixed in `.github/build.sh` I think. And likely in the macos too.

metsma: > I think the ref_name refers to the branch name _or_ tag name so for pushing release tags, this will result in suffix being `-0.27.0-rc1` which is not correct. I think we need to distinguish the tag and branch push with the `github.ref_type` .
> 

https://github.com/metsma/OpenSC/actions/runs/22391201096/job/64813380108#step:3:33
It did strip the version part from tag

`PACKAGE_SUFFIX=-rc1.1`

Jakuje: Still I think we should differentiate the tag and branch pushes to make sure we do not match this regex against branch names. My suggestion would be something like this:
```
if [[ "${{ github.event_name }}" == "pull_request" ]]; then
  if [[ "${{ github.base_ref }}" == "master" ]]; then
    SUFFIX="-pr${{ github.event.number }}"
  else
    SUFFIX="-${{ github.base_ref }}-pr${{ github.event.number }}"
  fi
elif [[ "${{ github.event_name }}" == "push" ]]; then
  if [[ "${{ github.ref_type }}" == "branch" && "${{ github.ref_name }} != "master" ]]; then
    SUFFIX="-${{ github.ref_name }}"
  elif [[ "${{ github.ref_type }}" == "tag" && "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
    SUFFIX="${BASH_REMATCH[1]}"
  fi
fi
```
The `.github/build.sh` should be adjusted accordingly, ideally using the same logic so its maintainable.

What do you think?

metsma: Updated code

Jakuje (approved):
thank you! Looks reasonable to me. Could you also update the build.sh in the same manner so its in sync and we have also linux artifacts with same suffix?

frankmorgner:
I'm not sure if relying on github specific variables is the most sustainable approach. historically, we were managing the version information in configure.ac, which is happening in VERSION.mk. I think parsing that information would be good enough to create a package name without relying on github.

Jakuje:
> I'm not sure if relying on github specific variables is the most sustainable approach. historically, we were managing the version information in configure.ac, which is happening in VERSION.mk. I think parsing that information would be good enough to create a package name without relying on github.

This is just about the suffixes, which are not part of the VERSION.mk (except for the -rcX).

frankmorgner:
There is `PACKAGE_SUFFIX=rc1`. What are you missing?

metsma:
> There is `PACKAGE_SUFFIX=rc1`. What are you missing?

it depends if you take artifacts from tag build or from master branch
https://github.com/OpenSC/OpenSC/blob/67278266ec6b69b47a64685dd52936ae96b84b03/.github/build.sh#L21
If from tag then the additional suffix is added eg 0.26.1 or 0.26.1-rc1

frankmorgner:
Having branch and/or tag is optional and may be omitted, in my opinion; the version information in VERSION.mk is good enough. I think we always added the value "rc1" in VERSION.mk/configure.ac in previous release candidates

Jakuje:
> Having branch and/or tag is optional and may be omitted, in my opinion; the version information in VERSION.mk is good enough. I think we always added the value "rc1" in VERSION.mk/configure.ac in previous release candidates

Yes, this was done also this time, but as mentioned in #3581 I noticed that the tag builds add nonsense suffixes, which is what we try to fix here.

frankmorgner:
sorry, maybe I didn't fully understand the issue. I'll try to have a look later with more refreshed eyes.

PR #3589, opened 2026-03-02
 refactored pin commands to avoid tries_left as additional parameter 

frankmorgner:
##### Checklist

- [ ] macOS token is tested


Jakuje (approved):
No issues found. But I merged #3579 which has a new pin_cmd() so this needs to be rebased and the pin_cmd in there adjusted otherwise it will fail to build.

frankmorgner:
I'm on vacation, I'll check later.

frankmorgner:
updated to latest changes of `master`

Jakuje:
I merged also #3595 as it was already due. This still fails to build as it needs some more changes to the lteid driver to work properly. It will likely need some adjustments for serbian driver too.

Jakuje (approved):

PR #3591, opened 2026-03-02
switch to portable bit fields

frankmorgner:
The C standard only guarantees bit fields for `unsigned int`, `signed int`, `int` or `_Bool` (since C99), even though in practice major compilers also support more types.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-coolkey.c
+++ b/src/libopensc/card-coolkey.c
@@ -253,8 +253,8 @@ typedef struct coolkey_attribute_header {
  * Definitions:
  *
 struct coolkey_fixed_attributes_values {
-	uint32_t  cka_id:4;
-	uint32_t  cka_class:3;
+	unsigned int  cka_id:4;
+	unsigned int  cka_class:3;
 	unsigned int  cka_token:1;
 	unsigned int  cka_private:1;
 	unsigned int  cka_modifiable:1;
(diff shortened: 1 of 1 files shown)


src/libopensc/card-coolkey.c
```diff
@@ -255,23 +255,23 @@ typedef struct coolkey_attribute_header {
...
+	unsigned int  cka_unwrap:1;
+	unsigned int  cka_sign:1;
+	unsigned int  cka_sign_recover:1;
+	unsigned int  cka_verify:1;
+	unsigned int  cka_verify_recover:1;
+	unsigned int  cka_sensitive:1;
+	unsigned int  cka_always_sensitive:1;
+	unsigned int  cka_extractable:1;
+	unsigned int  cka_never_extractable:1;
 	uint32_t  reserved:8;
```
Jakuje: Should also the `cka_class` and `cka_id` be converted to the same type? I think so (but the uint32_t type is likely used to express the size how it is stored in the attributes in practice.

But this is all block is commented out and is not used for any real mapping. The mapping to separate attributes is done with bit shifts and masks in the code. So does it even matter?

frankmorgner: Whether to delete this section or not, I am leaving up to you.

I'll change those parts as well, I overlooked this.

Jakuje:

PR #3594, opened 2026-03-06
Rename DBG artifacts

metsma:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -57,6 +57,12 @@ jobs:
           PACKAGE_SUFFIX="${{ matrix.configuration == 'Light' && '-Light' || '' }}${SUFFIX}"
           sed -i.bak "s/^PACKAGE_SUFFIX=\([-~]*[0-9a-zA-Z]*\)$/PACKAGE_SUFFIX=$PACKAGE_SUFFIX/g" VERSION.mk
           cat VERSION.mk
+          while IFS='=' read -r key value; do
+              # Skip empty or comment lines
+              [[ -z "$key" || "$key" =~ ^# ]] && continue
+              export "$key=$value"
+          done < VERSION.mk
+          echo "DBG_NAME=${PRODUCT_NAME}-${PACKAGE_VERSION_MAJOR}.${PACKAGE_VERSION_MINOR}.${PACKAGE_VERSION_FIX}${PACKAGE_SUFFIX}_${{ matrix.platform }}-dbg.zip" >> $GITHUB_ENV
       - name: Install CPDK
         run: choco install windows-cryptographic-provider-development-kit -y > $null
       - name: Setup dev env
@@ -139,7 +145,7 @@ jobs:
       - name: Archive debug artifacts
         uses: actions/upload-artifact@v6
         with:
-          name: debug_${{ matrix.image }}_${{ matrix.platform }}_${{ matrix.configuration }}
+          name: ${{ env.DBG_NAME }}
           path: |
             ./src/**/*.pdb
             ./win32/*.pdb


.github/workflows/windows.yml
```diff
@@ -57,6 +57,12 @@ jobs:
           PACKAGE_SUFFIX="${{ matrix.configuration == 'Light' && '-Light' || '' }}${SUFFIX}"
           sed -i.bak "s/^PACKAGE_SUFFIX=\([-~]*[0-9a-zA-Z]*\)$/PACKAGE_SUFFIX=$PACKAGE_SUFFIX/g" VERSION.mk
           cat VERSION.mk
+          while IFS='=' read -r key value; do
+              # Skip empty or comment lines
+              [[ -z "$key" || "$key" =~ ^# ]] && continue
+              export "$key=$value"
+          done < VERSION.mk
+          echo "DBG_NAME=${PRODUCT_NAME}-${PACKAGE_VERSION_MAJOR}.${PACKAGE_VERSION_MINOR}.${PACKAGE_VERSION_FIX}${PACKAGE_SUFFIX}_${{ matrix.platform }}-dbg.zip" >> $GITHUB_ENV
```
Jakuje, line 65: Thank you

I see now the filename generated looks like this:
```
OpenSC-0.27.0-Light-dbg_arm64-dbg.zip
```
when built from your `dbg` branch (therefore the two dbg suffixes -- its ok):
https://github.com/metsma/OpenSC/actions/runs/22759165017

In the PR its
```
OpenSC-0.27.0-pr3594_arm64-dbg.zip
```

The MSI names look like this (but they are in zip artifacts):
```
OpenSC-0.27.0_arm64-0.27.0-rc1.msi
```
so I think its ok.

My last suggestion would be if we can make the zip archive with the msi to have the same name so its easier to match with the debug symbols. Right now its
```
msi_windows-2025_arm64_Light
```
which is cognitively hard to match to
```
OpenSC-0.27.0-Light-pr3594_arm64-dbg.zip 
```

so just defining the artifact name here and adding the dbg in the `upload-artifact` would be probably easiest:

```suggestion
          echo "ARTIFACT_NAME=${PRODUCT_NAME}-${PACKAGE_VERSION_MAJOR}.${PACKAGE_VERSION_MINOR}.${PACKAGE_VERSION_FIX}${PACKAGE_SUFFIX}_${{ matrix.platform }}" >> $GITHUB_ENV
```
and 
```
          name: ${{ env.DBG_NAME }}-dbg.zip
```
and for the `upload-unsigned-artifact` step:
```
          name: ${{ env.DBG_NAME }}.zip
```

frankmorgner:
I wonder if we have gotten a little bit too crazy with the build artifacts. I mean, it is useful to test many build configurations and test environments running automatically, but as resulting artifacts we only need what we would also distribute on a release. That should simplify the naming of the artifacts and the process of finding the "correct" artifacts (which basically would be all artifacts built).

Jakuje:
I am just trying to make the release process straightforward and artifacts named in some consistent manner for users (and for us doing the releases) .

Right now, if I would be a user and asked to test something from the artifacts https://github.com/OpenSC/OpenSC/actions/runs/22912989586 I have no idea where to start. For macos its a bit better, but still not great. Do I want 14, 15 or 26? https://github.com/OpenSC/OpenSC/actions/runs/22912989566

frankmorgner:
I totally agree. And with "**we** have gotten a little bit too crazy", I should have written that **I** have added a lot of complexity over time without having thinking of releases and users wanting to test some pull request. I think some simplification should be done in the future, not necessarily in this very pull request.

Jakuje:
Merged as part of #3603.

PR #3595, opened 2026-03-06
Add srbeid card driver for Serbian eID and PKS cards

hirashix0:
## Summary

Adds a card driver and PKCS#15 emulator for Serbian government smart cards that use the CardEdge PKI applet:

- **Serbian eID Gemalto (2014+)** — matched by ATR `3B:FF:94`
- **Serbian eID IF2020 Foreigner** — matched by AID `A0 00 00 00 63 50 4B 43 53 2D 31 35`
- **PKS Chamber of Commerce card** — matched by same AID

Health insurance cards (RFZO) sharing the same CardEdge AID are explicitly excluded in `match_card()`.

The PKCS#15 emulator requires zlib (`ENABLE_ZLIB`) for decompressing certificates stored on the card.

## New files

- `src/libopensc/card-srbeid.c` — card driver (ATR/AID matching, APDU ops, PIN, signing)
- `src/libopensc/pkcs15-srbeid.c` — PKCS#15 emulator (cert/key/PIN enumeration from cmapfile)

## Modified files

- `cards.h` — `SC_CARD_TYPE_SRBEID_*` constants at 43000
- `ctx.c` — register driver (`ENABLE_ZLIB` gated)
- `pkcs15-syn.h/.c` — register PKCS#15 emulator
- `Makefile.am/.mak` — add source files

## Test results

Tested on macOS (HID OMNIKEY 5422) and Linux (Manjaro, Gemalto PC Twin Reader):

**Serbian eID (Gemalto):**
- `pkcs15-tool --list-certificates` — 2 certs (Key Exchange + Digital Signature)
- `pkcs15-tool --list-keys` — RSA-2048 (kxc 0x6005, ksc 0x6019)
- `pkcs15-tool --list-pins` — User PIN, ref 0x80, tries=3
- `pkcs15-crypt --sign` — RSA-2048 PKCS#1 v1.5 verified OK

**PKS Chamber of Commerce card:**
- `pkcs15-tool --list-certificates` — 2 certs enumerated
- `pkcs15-tool --list-keys` — RSA-2048 keys enumerated
- `pkcs15-tool --list-pins` — User PIN, ref 0x80, tries=3
- `pkcs15-crypt --sign` — RSA-2048 PKCS#1 v1.5 verified OK

**Health insurance card (RFZO):**
- Correctly rejected — driver detects health card and does not claim it

### DIFF
--- a/src/libopensc/card-srbeid.c
+++ b/src/libopensc/card-srbeid.c
@@ -5,7 +5,7 @@
  * same CardEdge PKCS#15 applet.  Cards are matched either by ATR
  * (Gemalto 2014+ eID) or by AID selection.
  *
- * Copyright (C) 2026  LibreSCRS contributors
+ * Copyright (C) 2026 LibreSCRS contributors
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -28,20 +28,12 @@
 
 #include <string.h>
 
+#include "card-srbeid.h"
 #include "internal.h"
 #include "log.h"
 
-/* CardEdge PKI applet AID  (A0 00 00 00 63 50 4B 43 53 2D 31 35) */
-static const u8 AID_PKCS15[] = {
-		0xA0, 0x00, 0x00, 0x00, 0x63,
-		0x50, 0x4B, 0x43, 0x53, 0x2D, 0x31, 0x35};
-#define AID_PKCS15_LEN (sizeof(AID_PKCS15))
-
 /* MSE algorithm byte for RSA-2048 PKCS#1 v1.5 */
 #define CE_MSE_ALG_RSA2048 0x02u
-
-/* Base address of key files: key FID = CE_KEYS_BASE_FID | container/type bits */
-#define CE_KEYS_BASE_FID 0x6000u
 
 static struct sc_card_operations srbeid_ops;
 static const struct sc_card_operations *iso_ops;
@@ -256,7 +248,7 @@ srbeid_compute_signature(sc_card_t *card,
 	apdu.lc = datalen;
 	apdu.resp = resp;
 	apdu.resplen = sizeof(resp);
-	apdu.le = 256;
+	apdu.le = sizeof(resp);
 
 	r = sc_transmit_apdu(card, &apdu);
 	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
@@ -292,7 +284,7 @@ srbeid_decipher(sc_card_t *card,
 	apdu.lc = crgram_len;
 	apdu.resp = resp;
 	apdu.resplen = sizeof(resp);
-	apdu.le = 256;
+	apdu.le = sizeof(resp);
 
 	r = sc_transmit_apdu(card, &apdu);
 	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
--- a/src/libopensc/pkcs15-srbeid.c
+++ b/src/libopensc/pkcs15-srbeid.c
@@ -2,7 +2,7 @@
  * pkcs15-srbeid.c: PKCS#15 emulation for Serbian cards using the
  *                 CardEdge PKI applet.
  *
- * Copyright (C) 2026  LibreSCRS contributors
+ * Copyright (C) 2026 LibreSCRS contributors
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -29,15 +29,10 @@
 #ifdef ENABLE_ZLIB
 #include "compression.h"
 #endif
+#include "card-srbeid.h"
 #include "internal.h"
 #include "log.h"
 #include "pkcs15.h"
-
-/* CardEdge PKI applet AID  (A0 00 00 00 63 50 4B 43 53 2D 31 35) */
-static const u8 AID_PKCS15[] = {
-		0xA0, 0x00, 0x00, 0x00, 0x63,
-		0x50, 0x4B, 0x43, 0x53, 0x2D, 0x31, 0x35};
-#define AID_PKCS15_LEN (sizeof(AID_PKCS15))
 
 /* CardEdge cmapfile constants. */
 #define CE_CMAP_RECORD_SIZE	86u
@@ -45,7 +40,6 @@ static const u8 AID_PKCS15[] = {
 #define CE_CMAP_SIG_SIZE_OFFSET 82u
 #define CE_CMAP_KX_SIZE_OFFSET	84u
 #define CE_CMAP_VALID_CONTAINER 0x01u
-#define CE_KEYS_BASE_FID	0x6000u
 #define CE_KEY_KIND_PRIVATE	1u
 #define CE_AT_KEYEXCHANGE	1u
 #define CE_AT_SIGNATURE		2u
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -13,7 +13,7 @@ noinst_HEADERS = cards.h ctbcs.h internal.h muscle.h muscle-filesystem.h \
 	authentic.h iasecc.h iasecc-sdo.h sm.h card-sc-hsm.h \
 	pace.h cwa14890.h cwa-dnie.h card-gids.h aux-data.h \
 	jpki.h sc-ossl-compat.h card-npa.h card-openpgp.h \
-	card-eoi.h card-dtrust.h ccid-types.h reader-tr03119.h \
+	card-eoi.h card-dtrust.h card-srbeid.h ccid-types.h reader-tr03119.h \
 	card-cac-common.h card-cardos-common.h
 
 AM_CPPFLAGS = -D'OPENSC_CONF_PATH="$(sysconfdir)/opensc.conf"' \
--- a/src/libopensc/card-srbeid.h
+++ b/src/libopensc/card-srbeid.h
@@ -1,0 +1,35 @@
+/*
+ * card-srbeid.h: Shared definitions for Serbian CardEdge card and PKCS#15 drivers.
+ *
+ * Copyright (C) 2026 LibreSCRS contributors
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifndef CARD_SRBEID_H
+#define CARD_SRBEID_H
+
+#include "types.h"
+
+/* CardEdge PKI applet AID  (A0 00 00 00 63 50 4B 43 53 2D 31 35) */
+static const u8 AID_PKCS15[] = {
+		0xA0, 0x00, 0x00, 0x00, 0x63,
+		0x50, 0x4B, 0x43, 0x53, 0x2D, 0x31, 0x35};
+#define AID_PKCS15_LEN (sizeof(AID_PKCS15))
+
+/* Base address of key files: key FID = CE_KEYS_BASE_FID | container/type bits */
+#define CE_KEYS_BASE_FID 0x6000u
+
+#endif /* CARD_SRBEID_H */


src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+
+	/* ATR hit: Gemalto 2014+ Serbian eID (3B:FF:94 ...) */
+	if (_sc_match_atr(card, srbeid_atrs, &card->type) >= 0)
+		return 1;
+
+	/*
+	 * AID-based match for IF2020 Foreigner and PKS cards.
+	 *
+	 * Health cards (SERVSZK) also accept AID_PKCS15 — reject them first.
+	 * This mirrors the PKSCard::probe() false-positive fix in lib/pkscard/.
```
frankmorgner: What reference code are you using? Which license does it have?

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * Apollo 2008 ATR 3B:B9:18 ... is intentionally absent — no CardEdge applet.
+ */
+static const struct sc_atr_table srbeid_atrs[] = {
+	{ "3B:FF:94", "FF:FF:FF", "Serbian eID (Gemalto 2014+)", SC_CARD_TYPE_SRBEID_GEMALTO, 0, NULL },
+	{ NULL, NULL, NULL, 0, 0, NULL }
+};
+
+/* Select an AID and return the status word (0 on transport error). */
+static int srbeid_select_aid(sc_card_t *card, const u8 *aid, size_t aid_len)
```
frankmorgner: use `iso7816_select_aid(card, aid, aid_len, NULL, NULL);` and check for `SC_SUCCESS`. If you want to make it identical to your current implementation, then supply a response buffer to `iso7816_select_aid()` as well.

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+/* Base address of key files: key FID = CE_KEYS_BASE_FID | container/type bits */
+#define CE_KEYS_BASE_FID	0x6000u
+
+/* CardEdge PIN constants */
+#define CE_PIN_REFERENCE	0x80u
+#define CE_PIN_MAX_LENGTH	8u
+
+/* Per-card private data stored in card->drv_data. */
+typedef struct srbeid_private {
+	int      card_type;
```
frankmorgner: is this hallucinated? card_type is never used

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+	u8       pin_tries;  /* cached tries-left (0 = unknown/uncached) */
```
frankmorgner: never used

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+	free(card->drv_data);
+	card->drv_data = NULL;
+	return SC_SUCCESS;
+}
+
+/* -------------------------------------------------------------------------
+ * restore_security_env — no persistent SE on CardEdge; always succeeds.
+ * OpenSC calls this before set_security_env on some code paths.
+ * ------------------------------------------------------------------------- */
+static int srbeid_restore_security_env(sc_card_t *card, int se_num)
```
frankmorgner: why is this needed?

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+static int srbeid_compute_signature(sc_card_t *card,
+	const u8 *data, size_t datalen, u8 *out, size_t outlen)
+{
+	srbeid_private_t *priv = (srbeid_private_t *)card->drv_data;
+	sc_apdu_t apdu;
+	u8 mse_data[7];
+	u8 resp[512];
+	int r;
+
+	/* MSE SET: tag 0x80 = algorithm (RSA2048), tag 0x84 = key ref (2 bytes BE) */
```
frankmorgner, line 205: MSE:SET should likely be sent in `set_security_env()`, hence the name of the callback. If you implement this change, then `key_ref` in your private data structure is also not used anywhere.

hirashix0:
Thanks for a prompt reply and review. Let me review it and update the PR.

I'll clean up comments and address things you mentioned, re-test and update. Reference code I used is https://github.com/LibreSCRS/LibreMiddleware (LGPL-2.1), which is my own project — the CardEdge protocol was reverse-engineered through APDU tracing and testing on physical cards with help of AI. Some implementation patterns were carried over from there without adapting to OpenSC conventions (e.g., the custom srbeid_select_aid() instead of iso7816_select_aid()). 

Let me clean it up and update...

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+	apdu.datalen = datalen;
+	apdu.lc      = datalen;
+	apdu.resp    = resp;
+	apdu.resplen = sizeof(resp);
+	apdu.le      = 256;
+
+	r = sc_transmit_apdu(card, &apdu);
+	if (r < 0)
+		return r;
+	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
```
frankmorgner: typically, you would use `sc_check_sw()`

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+	u8 pin1_padded[CE_PIN_MAX_LENGTH];
+	u8 pin2_padded[CE_PIN_MAX_LENGTH];
+	u8 change_data[CE_PIN_MAX_LENGTH * 2];
+	int r;
+
+	if (tries_left)
+		*tries_left = -1;
+
+	/* Re-select AID_PKCS15 before any PIN APDU. */
+	if (srbeid_select_aid(card, AID_PKCS15, AID_PKCS15_LEN) != 0x9000) {
```
frankmorgner: If you use the OpenSC convention to provide a return code from `errors.h`, then you can use macros such as `LOG_TEST_RET()`

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+ * Used by TLS client authentication and S/MIME decryption.
+ * Uses confidentiality MSE template (P2=0xB8) and PSO DECIPHER (00 2A 80 86).
+ * ------------------------------------------------------------------------- */
+static int srbeid_decipher(sc_card_t *card,
+	const u8 *crgram, size_t crgram_len, u8 *out, size_t outlen)
+{
+	srbeid_private_t *priv = (srbeid_private_t *)card->drv_data;
+	sc_apdu_t apdu;
+	u8 mse_data[7];
+	u8 resp[512];
```
frankmorgner: I'm curious about why you use a 512 byte buffer, if you're only expecting 256 bytes of response.

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+			data->pin1.tries_left = tl;
+		} else if (apdu.sw1 == 0x69 && apdu.sw2 == 0x83) {
+			if (tries_left)
+				*tries_left = 0;
+			data->pin1.tries_left = 0;
+		}
+		return SC_SUCCESS;
+
+	case SC_PIN_CMD_VERIFY:
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x20, 0x00, CE_PIN_REFERENCE);
```
frankmorgner: PIN-pad readers are not supported. Is this intended? If you use the code from iso7816.c, then you would get PIN-pad support for free.

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,462 @@
...
+ * pin_cmd
+ *
+ * Handles SC_PIN_CMD_GET_INFO, SC_PIN_CMD_VERIFY, SC_PIN_CMD_CHANGE.
+ * PIN is null-padded to CE_PIN_MAX_LENGTH bytes before transmission.
+ * Status word parsing:
+ *   0x9000        → success
+ *   0x6983        → blocked
+ *   0x63 CX       → X tries remaining
+ * ------------------------------------------------------------------------- */
+static int srbeid_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data,
```
frankmorgner: As far as I can see, the PIN commands are identical to the default implementation in iso7816.c, apart from the preceeding selection of the AID. So I think there is a chance that the default implemetntation can be reused: change `srbedi_pin_cmd()` to first call `srbeid_select_aid()` and then call `return iso_drv->ops->pin_cmd()` and you should be done. Also, you could check if you can avoid manually selecting the AID: If the PKCS#15 metadata of the PIN would have the AID attached to it, the PKCS#15 layer would know that it needs to select the AID first.

src/libopensc/cards.h
```diff
@@ -278,6 +278,12 @@ enum {
 	SC_CARD_TYPE_DTRUST_V5_1_M100,
 	SC_CARD_TYPE_DTRUST_V5_4_STD,
 	SC_CARD_TYPE_DTRUST_V5_4_MULTI,
+
+	/* Serbian eID and PKS Chamber of Commerce cards (CardEdge PKI applet) */
+	SC_CARD_TYPE_SRBEID_BASE    = 43000,
+	SC_CARD_TYPE_SRBEID_GEMALTO = 43000,  /* Serbian eID Gemalto (2014+), ATR 3B:FF:94 */
```
frankmorgner: Please stick to the convention to increase the value for every entry.

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+	free(mscp_buf);
+	free(cmap_buf);
+	free(root_entries);
+	free(mscp_entries);
+	free(certs);
+	return r;
+}
+
+/*
+ * Read the raw (possibly zlib-compressed) cert file and return DER bytes.
```
frankmorgner, line 338: If the certificate is only *possibly* compressed, then you may avoid `#ifdef ENABLE_ZLIB` in your driver and just let `sc_decompress()` handle a missing zlib. This would assure that non-compressed certificates could still be read even if zlib is missing.

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+		| ((key_pair_id << 2) & 0x000Cu)
+		| CE_KEY_KIND_PRIVATE;
+}
+
+/* -------------------------------------------------------------------------
+ * Low-level I/O helpers
+ * ------------------------------------------------------------------------- */
+
+/* Select an AID and return the status word (0 on transport error). */
+static int srbeid_select_aid(sc_card_t *card, const u8 *aid, size_t aid_len)
```
frankmorgner: avoid duplicate dimplementation

we expect missing low level commands with APDU exchange to be implemented in the card-srbeid.c

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+ * so we must NOT go through sc_select_file() / iso7816_select_file()
+ * which would try to parse it as tag-0x6F TLV and return
+ * SC_ERROR_UNKNOWN_DATA_RECEIVED.
+ *
+ * CardEdge FCI layout (10 bytes, all big-endian):
+ *   [FID_H FID_L Size_H Size_L ACL*6]
+ *
+ * Returns the file size on success, or a negative SC_ERROR_* code.
+ */
+static int srbeid_select_fid_raw(sc_card_t *card, unsigned int fid)
```
frankmorgner: `sc_read_binary()` should just work.

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+
+	return (int)(((unsigned)fci[2] << 8) | (unsigned)fci[3]);
+}
+
+/*
+ * Select FID and read the entire file into a malloc'd buffer.
+ * *out_len receives the byte count; caller must free() the buffer.
+ * Returns SC_SUCCESS or a negative SC_ERROR_* code.
+ */
+static int srbeid_read_fid(sc_card_t *card, unsigned int fid,
```
frankmorgner: sc_read_binary() also supports multiple APDU exchanges with READ BINARY and GET RESPONSE

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+/* One entry from a CardEdge directory file. */
+typedef struct ce_dir_entry {
+	char     name[9];   /* 8-char name + NUL */
+	unsigned fid;
+	int      is_dir;
+} ce_dir_entry_t;
+
+/* Parse a CardEdge directory file into an array of ce_dir_entry_t.
+ * *entries_out: caller must free().  Returns entry count or -1 on error. */
+static int ce_parse_dir(const u8 *data, size_t len, ce_dir_entry_t **entries_out)
```
frankmorgner: Is this meant to parse EF.DIR as done in `sc_enum_apps()` from `dir.c`?

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,686 @@
...
+
+	/*
+	 * Query PIN tries_left — AID_PKCS15 is still selected from enum_certs.
+	 * VERIFY Case 1 (no data body) checks status without consuming a try.
+	 */
+	{
+		sc_apdu_t tapdu;
+		int rv, pin_tries_left = -1;
+
+		sc_format_apdu(card, &tapdu, SC_APDU_CASE_1, 0x20, 0x00, CE_PIN_REFERENCE);
```
frankmorgner: At this point, you could just use sc_pin_cmd to get the tries levt.

hirashix0:
Thanks. These comments helped a lot. I've now addressed them all. Driver looks more compact and cleaner.

Here's the overview:                                                                                                                                                                                                   
  - Replaced custom SELECT AID with iso7816_select_aid()
  - Removed private data / finish / restore — driver is stateless                                                                                                                                  
  - Moved MSE SET to set_security_env()                                                                                                                                                            
  - sc_check_sw() + LOG_TEST_RET throughout                                                                                                                                                        
  - Comments cleaned and dash separators removed                                                                                                                                                                        
  - iso_ops delegation pattern (same as card-piv, card-openpgp)                                                                                                                                    
  - Added srbeid_select_file() for CardEdge's proprietary 10-byte FCI; DF_NAME delegates to iso_ops                                                                                                
  - SC_PKCS15_PIN_FLAG_NEEDS_PADDING — ISO layer handles null-padding                                                                                                                              
  - Response buffers 512→256                                                                                                                                                                       
  - cards.h enum auto-increment                                                                                                                                                                    
  - sc_select_file() + sc_read_binary() replace raw APDU file I/O                                                                                                                                  
  - sc_pin_cmd(GET_INFO) for tries_left + SC_CARD_CAP_ISO7816_PIN_INFO in init                                                                                                                     
  - Removed #ifdef ENABLE_ZLIB guard — sc_decompress_alloc() called unconditionally, uncompressed certs work regardless of zlib availability                                                       
  - Attached AID to auth_info.path and key_info.path — PKCS#15 layer now selects the applet automatically before PIN and key operations, removing the need for a custom pin_cmd wrapper and manual 
  AID selection in set_security_env(). Should this give PIN-pad reader support for free?                                                                                                      
  - ce_parse_dir() parses CardEdge's proprietary binary directory format, not ISO 7816-4 EF.DIR                                                                                                    
                                                                                                                                                                                                   
  Hardware tested: list-pins (tries_left OK), list-certs, list-keys, sign+verify, PIN change.

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,337 @@
...
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+/*
+ * select_file — handle CardEdge's proprietary 10-byte FCI response.
+ *
+ * CardEdge FCI layout (10 bytes, big-endian):
+ *   [FID_H FID_L Size_H Size_L ACL*6]
+ *
+ * iso7816_select_file() would try to parse this as ISO 7816-4 TLV (tag 0x6F)
```
frankmorgner, line 100: If it is only the FCI that is problematic, but everything else can be reused, please try if an alternative implementation of the card's callback to `process_fci()` would be sufficient to adapt to the proprietary buffer.

src/libopensc/ctx.c
```diff
@@ -143,6 +143,11 @@ static const struct _sc_driver_entry internal_card_drivers[] = {
  * driver's `match_card()` function. */
 	{ "esteid2018",	(void *(*)(void)) sc_get_esteid2018_driver },
 	{ "esteid2025",	(void *(*)(void)) sc_get_esteid2025_driver },
+	/* srbeid: Serbian eID (Gemalto 2014+, IF2020 Foreigner) and PKS card
+	 * need AID selection in match_card() to reject the SERVSZK health card. */
+#ifdef ENABLE_ZLIB
```
frankmorgner: this restriction is not needed anymore

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,590 @@
...
+ * CardEdge directories use a proprietary binary format:
+ *   [10-byte header] [12-byte entries...]
+ * This is NOT ISO 7816-4 EF.DIR (ASN.1 BER-TLV application templates),
+ * so standard sc_enum_apps() / iso7816_read_ef_dir() cannot be used.
+ *
+ * *entries_out: caller must free().  Returns entry count or -1 on error.
+ */
+static int ce_parse_dir(const u8 *data, size_t len, ce_dir_entry_t **entries_out)
+{
+	unsigned count, i;
```
frankmorgner: would it make sense to switch to `size_t` instead of `unsigned`?

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,590 @@
...
+	if (ncerts == 0) {
+		sc_log(card->ctx, "srbeid: no certificates found");
+		goto out;
+	}
+
+	/* Set card label. */
+	free(p15card->tokeninfo->label);
+	p15card->tokeninfo->label = strdup("Serbian eID");
+	free(p15card->tokeninfo->manufacturer_id);
+	p15card->tokeninfo->manufacturer_id = strdup("LibreSCRS");
```
frankmorgner: This should be the name of the manufacturer of the token

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,590 @@
...
+	free(certs);
+	return r;
+}
+
+int sc_pkcs15emu_srbeid_init_ex(sc_pkcs15_card_t *p15card, struct sc_aid *aid)
+{
+	(void)aid;
+
+	if (p15card->card->type != SC_CARD_TYPE_SRBEID_GEMALTO &&
+	    p15card->card->type != SC_CARD_TYPE_SRBEID_IF2020  &&
```
frankmorgner: I think this should give an error with the code formatting check as you're mixing tabs and spaces

src/libopensc/card-srbeid.c
hirashix0: I investigated this including writing a test that sends SELECT FILE with every P1/P2/Le combination to the CardEdge applet. The card always returns the same proprietary 10-byte response (FID_H FID_L Size_H Size_L ACL×6) regardless of P2.

The problem is that iso7816_select_file() checks apdu.resp[0] for tags 0x6F (FCI) or 0x62 (FCP) at line 759 and returns SC_ERROR_UNKNOWN_DATA_RECEIVED for anything else — before process_fci() is ever called. Since the CardEdge response starts with the high byte of the FID (e.g. 0x70 for dir 0x7000), it always fails and process_fci in not reached.

This is output of the test:
Testing FID: root dir (0x7000)                                                                                                                                                                   
                                                                                                                                                                                                   
    P1=0x00 (select by FID) | P2=0x00 (return FCI, expect 0x6F) | Le=0x00 (up to 256)                                                                                                              
      SW=9000 (OK)                                                                                                                                                                               
      Response (10 bytes): 70 00 00 2C 01 01 01 01 01 00                                                                                                                                           
      First byte: 0x70 (raw proprietary, not TLV)                                                                                                                                                  
                                                                                                                                                                                                   
    P1=0x00 (select by FID) | P2=0x04 (return FCP, expect 0x62) | Le=0x00 (up to 256)                                                                                                              
      SW=9000 (OK)                                                                                                                                                                                 
      Response (10 bytes): 70 00 00 2C 01 01 01 01 01 00                                                                                                                                           
      First byte: 0x70 (raw proprietary, not TLV)

src/libopensc/pkcs15-srbeid.c
hirashix0: Hm, it seems we're going around in circles with this one :)

This way, we're matching the pattern with e.g. card-coolkey.c.

The driver registration in ctx.c is no longer guarded — the driver loads unconditionally. Here it works for some operations even without zlib; only certificate reading requires decompression since cards I have access to use zlib-compressed DER certificates.

The guard must be her, not at the registration point, because compression.c is entirely empty when ENABLE_ZLIB is not defined — it provides no stub functions — and any direct reference to sc_decompress_alloc() causes an undefined symbol linker error at build time.

src/libopensc/card-srbeid.c
frankmorgner: thanks for looking into this. You are right, the current implementation of iso7816.c is not flexible enough for this.

src/libopensc/ctx.c
hirashix0: Removed, but back at calling site

src/libopensc/pkcs15-srbeid.c
hirashix0: Definetely. Done

src/libopensc/pkcs15-srbeid.c
hirashix0: I'm not sure what has to go here. CardEdge is propriatery applet produced by (former) NetSeT for eID, PKS and health cards. Those are issued by MoI, Chamber of commetce and health fond. That's why I'm keeping it here as just CardEdge. 
Please advise if it should be something different.

src/libopensc/pkcs15-srbeid.c
hirashix0: Fixed

src/libopensc/pkcs15-srbeid.c
frankmorgner: I mistakenly had the impression that compression.h and compression.c would already be capable of handling a missing zlib. I was wrong, sorry. At least for now, your implementation is good as is. I hope to defer missing zlib support into compression.c in the future to allow simplifying the drivers.

hirashix0:
> From my point of view the code would be good to be merged when the upcoming release is done.

Perfect. Thanks! Is there anything else I should do for the merge process? Sorry, not sure how it works.

> Did you test the higher API layer, e.g. PKCS#11, Windows Minidriver and/or macOS CTK? Is there some card emulation software which would allow performing integration tests without an actual token?

Tested on Linux and macOS in several rounds:                                                                                                                                
                                                                                                                                                                              
  1. Basic enumeration with pkcs15-tool — slots, certificates, keys, PINs                                                                                                     
  2. PKCS#11 interface with pkcs11-tool, including mTLS authentication from Firefox
  3. Signing: pkcs11-tool --sign --mechanism SHA256-RSA-PKCS and pkcs15-crypt --sign --pkcs1 --sha-256

  All tests passed on both platforms. macOS was tested with both PCSC and CryptoTokenKit reader transports.

Unfortunately, I'm not aware that some card emulation software exists for Serbian cards.

frankmorgner:
Just wait and see if someone else has time for a review. If there are no remarks, then everything is fine.

In the meantime, you could check the formatting suggestion from the CI checks https://github.com/OpenSC/OpenSC/actions/runs/22951533367/job/66620066388?pr=3595

hirashix0:
> Just wait and see if someone else has time for a review. If there are no remarks, then everything is fine.

OK. Thanks!

> In the meantime, you could check the formatting suggestion from the CI checks https://github.com/OpenSC/OpenSC/actions/runs/22951533367/job/66620066388?pr=3595

New commit pushed to address formatting.

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,588 @@
...
+{
+	sc_path_t path;
+	sc_file_t *file = NULL;
+	u8 *buf;
+	int r;
+
+	*buf_out = NULL;
+	*out_len = 0;
+
+	memset(&path, 0, sizeof(path));
```
Jakuje: memset should not be needed if you initialize the variable with:
```
sc_path_t path = {0};
```

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,588 @@
...
+	r = sc_select_file(card, &path, &file);
+	if (r < 0)
+		return r;
+
+	if (!file || file->size == 0) {
+		sc_file_free(file);
+		return SC_SUCCESS;
+	}
+
+	buf = malloc(file->size);
```
Jakuje, line 101: Before allocating something based on the direct output from the card like here, its good practice to set some upper bound to prevent memory exhaustion when malfunctioning or malicious card/device is connected to the system

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,588 @@
...
+
+	*entries_out = NULL;
+	if (len < CE_DIR_HEADER_SIZE)
+		return -1;
+
+	count = (size_t)data[6] | ((size_t)data[7] << 8);
+	if (count == 0)
+		return 0;
+
+	entries = calloc(count, sizeof(*entries));
```
Jakuje: nit: I prefer the following syntax as it refers directly to the type and not indirectly through the variable name:
```suggestion
	entries = calloc(count, sizeof(ce_dir_entry_t));
```

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,588 @@
...
+
+		/*
+		 * Key usage flags by type:
+		 *   kxc (AT_KEYEXCHANGE) — encryption / key wrapping / decryption
+		 *   ksc (AT_SIGNATURE)   — digital signature / non-repudiation only
+		 */
+		if (is_kxc) {
+			key_info.usage = SC_PKCS15_PRKEY_USAGE_ENCRYPT | SC_PKCS15_PRKEY_USAGE_DECRYPT | SC_PKCS15_PRKEY_USAGE_WRAP | SC_PKCS15_PRKEY_USAGE_UNWRAP | SC_PKCS15_PRKEY_USAGE_SIGN;
+		} else {
+			key_info.usage = SC_PKCS15_PRKEY_USAGE_SIGN | SC_PKCS15_PRKEY_USAGE_NONREPUDIATION;
```
Jakuje, line 511: the code does not match the comment. The first list has additional `SC_PKCS15_PRKEY_USAGE_SIGN`. Should the ksc have symmetrically `SC_PKCS15_PRKEY_USAGE_VERIFY` too when the first has encrypt?

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,588 @@
...
+		}
+
+		/* ---- Certificate object ---- */
+		if (srbeid_read_cert_der(card, certs[i].cert_fid, &der, &der_len) < 0) {
+			sc_log(card->ctx, "srbeid: could not read cert[%d] DER", i);
+			continue;
+		}
+
+		memset(&cert_info, 0, sizeof(cert_info));
+		memset(&cert_obj, 0, sizeof(cert_obj));
```
Jakuje: Same here -- I would prefer initializers over the memset.

src/libopensc/pkcs15-srbeid.c
hirashix0: Done — replaced memset with initializer here and in all other instances.
 There are much more memsets throughout drivers, so I relied on that, but happy to use {0} going forward

src/libopensc/pkcs15-srbeid.c
hirashix0: Good catch. Added guard.

src/libopensc/pkcs15-srbeid.c
hirashix0: Done. Changed throughout the file.

src/libopensc/pkcs15-srbeid.c
hirashix0: SIGN is needed for kxc because TLS client authentication needs it (to perform a signature using the key exchange). Updated the comment to clarify this. The ksc key doesn't need VERIFY since verification is done with the public key.

Jakuje:
@hirashix0 I merged #3579 which caused conflicts as its a new driver. please, rebase and resolve the conflicts. Note, that the `SC_CARD_TYPE_SRBEID_BASE` will have to be bumped to  `44000` as the 43000 is now taken by Lithuanian eID.

hirashix0:
> @hirashix0 I merged #3579 which caused conflicts as its a new driver. please, rebase and resolve the conflicts. Note, that the `SC_CARD_TYPE_SRBEID_BASE` will have to be bumped to `44000` as the 43000 is now taken by Lithuanian eID.

Rebased, resolved  and squashed to one meaningful commit.

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,322 @@
...
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0x2A, 0x9E, 0x00);
+	apdu.data = data;
+	apdu.datalen = datalen;
+	apdu.lc = datalen;
+	apdu.resp = resp;
+	apdu.resplen = sizeof(resp);
+	apdu.le = 256;
```
Jakuje: nit: use the expression rather than magic numbers:
```suggestion
	apdu.resplen = sizeof(resp);
	apdu.le = sizeof(resp);
```

src/libopensc/card-srbeid.c
```diff
@@ -0,0 +1,322 @@
...
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0x2A, 0x80, 0x86);
+	apdu.data = crgram;
+	apdu.datalen = crgram_len;
+	apdu.lc = crgram_len;
+	apdu.resp = resp;
+	apdu.resplen = sizeof(resp);
+	apdu.le = 256;
```
Jakuje: nit: same here:
```suggestion
	apdu.resplen = sizeof(resp);
	apdu.le = sizeof(resp);
```

src/libopensc/pkcs15-srbeid.c
```diff
@@ -0,0 +1,584 @@
...
+#endif
+#include "internal.h"
+#include "log.h"
+#include "pkcs15.h"
+
+/* CardEdge PKI applet AID  (A0 00 00 00 63 50 4B 43 53 2D 31 35) */
+static const u8 AID_PKCS15[] = {
+		0xA0, 0x00, 0x00, 0x00, 0x63,
+		0x50, 0x4B, 0x43, 0x53, 0x2D, 0x31, 0x35};
+#define AID_PKCS15_LEN (sizeof(AID_PKCS15))
```
Jakuje: We have this in both of the card and pkcs15 files. Would it make sense to move it to separate header file to avoid duplication? Are there some other structures/constants that are reused and that might make sense to be moved there?

src/libopensc/card-srbeid.c
hirashix0: Agree, fixed.

src/libopensc/card-srbeid.c
hirashix0: Agree, fixed

src/libopensc/pkcs15-srbeid.c
hirashix0: Agree, factored out to separate .h file

hirashix0:
> I did one more path and noted few nits. Otherwise it looks good and if no more concerns show up, I will merge it later this week

Hi @Jakuje 

Just a friendly ping - Is there anything else needed before this can be merged? 
Rebase resolved conflicts with #3579 and a few proposed changes added.

Thanks!

Jakuje:
Nothing from my side. I wanted to give others time to have a look into this, but if there are no other comments, lets get it merged. I would slightly prefer to get the fixup commit squashed to the original one, but I am find either way. If there will be no more comments, I will merge it later today.

hirashix0:
Thanks!
Then, let's keep two commits as those are already prepared and no new force-push changes are needed.

Jakuje:
@hirashix0 Coverity complains about the following part of the code after merge:
```
** CID 503167:       Insecure data handling  (TAINTED_SCALAR)
/src/libopensc/pkcs15-srbeid.c: 150           in ce_parse_dir()


_____________________________________________________________________________________________
*** CID 503167:         Insecure data handling  (TAINTED_SCALAR)
/src/libopensc/pkcs15-srbeid.c: 150             in ce_parse_dir()
144     		return -1;
145     
146     	count = (size_t)data[6] | ((size_t)data[7] << 8);
147     	if (count == 0)
148     		return 0;
149     
>>>     CID 503167:         Insecure data handling  (TAINTED_SCALAR)
>>>     Passing tainted expression "count" to "calloc", which uses it as an allocation size.
150     	entries = calloc(count, sizeof(ce_dir_entry_t));
151     	if (!entries)
152     		return -1;
153     
154     	for (i = 0; i < count; i++) {
155     		size_t off = CE_DIR_HEADER_SIZE + i * CE_DIR_ENTRY_SIZE;
```
Is there some upper bound of the number of entries we can expect here to prevent allocation of too much of memory?

hirashix0:
Coverity marked it correctly. There's no particular hard limit on the number of entries (we could assume a reasonable one, but it would be arbitrary), so instead of bounding count directly, maybe the cleanest check is against the buffer size we already have:

if (CE_DIR_HEADER_SIZE + count * CE_DIR_ENTRY_SIZE > len)                                                                                                                                                                               
  return -1;

This way the buffer length itself provides the bound — if the header claims more entries than the data could hold, we reject it before allocating. 

But again, this one can overflow and size_t can be small enough after wrapparound to pass a check.
So, the check can be then:
if (count > (len - CE_DIR_HEADER_SIZE) / CE_DIR_ENTRY_SIZE)                                                                                                                                                                               
      return -1;  
(at this point len - CE_DIR_HEADER_SIZE is always > 0)

Let me refactor and test, then open a new PR from master.

PR #3598, opened 2026-03-10
CI: enable Dependabot for actions and vcpkg

frankmorgner:


Jakuje:
I know that in python/rust project dependabot is very noisy. But for OpenSC if it will be limited to the github actions and vcpkg, I think it should not generate a lot of PRs.

Do we have a way to make some dry-run what it will do? Do we need to install this somehow to the OpenSC organization? Did you already do that or is it still neded to be done after merging this PR?

frankmorgner:
Indeed, the limited scope should not add too much noise. I am not aware of how to perform a dry run, but adding this as is should not have negative side effects even if we decide to revert the change later.

PR #3604, opened 2026-03-12
Remove broken packit integration and build RPMs from scratch with rpmbuild

Jakuje:
The packit pipelines started to fail recently and I have no idea why. We are using some unsupported setup that we put ad-hoc together some years back, but building RPMs is not that hard so I switched to manual rpmbuild.

Currently built on top of #3603, but I do not care if it will land earlier or later.

##### Checklist

- [ ] macOS token is tested


packaging/opensc.spec
```diff
@@ -25,10 +26,10 @@ BuildRequires:  p11-kit-devel
 BuildRequires:  libcmocka-devel
 BuildRequires:  vim-common
 %if ! 0%{?rhel}
-BuildRequires:  softhsm
-BuildRequires:  openssl
 BuildRequires:  openpace-devel
 %endif
+BuildRequires:  softhsm
```
frankmorgner, line 31: is this for testing reasons?

Jakuje: yes. I think in some older version we skipped the tests in rhel, but it works now quite well

frankmorgner (approved):

Jakuje:
Rebased and merging to fix CI pipelines in master.

PR #3608, opened 2026-03-13
ci: Update to latest libressl 4.2.1

Jakuje:
The latest libressl version is 4.2.1 so lets update our CI to that

##### Checklist

- [ ] macOS token is tested


Jakuje:
Filled oseid issue in popovec/oseid#6 as it looks like the only issue with the new version

Jakuje:
CI is happy after rerun so I think we should be ok to let this in before the release as it only touches the tests.

Jakuje:
This is just a CI so I think its ok to merge it before the release.

PR #3601, opened 2026-03-12
Add brainpoolPxxxt1 missing (twisted) curve OIDs definitions

embetrix:
Add OIDs missing definitions for the brainpool twisted curves as defined in RFC 5639

  - brainpoolP192t1 (1.3.36.3.3.2.8.1.1.4)
  - brainpoolP224t1 (1.3.36.3.3.2.8.1.1.6)
  - brainpoolP256t1 (1.3.36.3.3.2.8.1.1.8)
  - brainpoolP320t1 (1.3.36.3.3.2.8.1.1.10)
  - brainpoolP384t1 (1.3.36.3.3.2.8.1.1.12)
  - brainpoolP512t1 (1.3.36.3.3.2.8.1.1.14)

Fixes [#3164](https://github.com/OpenSC/OpenSC/issues/3164)

PR #3611, opened 2026-03-20
pkcs15-sc-hsm.c Comparing Brainpool EC parameters

dengert:
Brainpool regular and twisted curves use the same prime number P. `sc_pkcs15emu_sc_hsm_get_curve_oid` was only comparing P to determine the OID. It now compares P and A because the twisted versions use a different A sometimes called A'. The table of EC parameters used in the compare is in libopensc/pkcs15-sc-hsm.c where #3601 adds the twisted curves.

 Tested on branch PR-3601
 Changes to be committed:
	modified:   libopensc/pkcs15-sc-hsm.c

Fixes: #3601 when SC-HSM tries to use a twisted curve.

It can then be merged:

-  now as existing EC curves all have different P and A and will work  as before. 

- immediately after #3601 which adds the twisted curves and the problem. 

Tested against "SmartCard-HSM version 4.1', type:26000", card with atr:3b:de:96:ff:81:91:fe:1f:c3:80:31:81:54:48:53:4d:31:73:80:21:40:81:07:92
using:
```
sc-hsm-tool --initialize --so-pin 3132333435363738 --pin 123456
./pkcs11-tool -l --pin 123456 --keypairgen --key-type EC:brainpoolP256t1 --label bp256t1 --id 12
./pkcs11-tool --sign -m ECDSA-SHA256 --id 12 --signature-format openssl --input-file /tmp/ecdsa.sign.text --output-file /tmp/ecdsa.sign.text.sig
openssl dgst -verify /tmp/ecdsa.sign.pubkey.12.pem -sha256 -signature /tmp/ecdsa.sign.text.sig /tmp/ecdsa.sign.text
Verified OK
```

##### Checklist

- [X ] PKCS#11 module is tested for key generation and ECDSA signature verified by OpenSSL
- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-sc-hsm.c
+++ b/src/libopensc/pkcs15-sc-hsm.c
@@ -838,7 +838,11 @@ int sc_pkcs15emu_sc_hsm_get_curve_oid(sc_cvc_t *cvc, const struct sc_lv_data **o
 	int i;
 
 	for (i = 0; curves[i].oid.value; i++) {
-		if ((curves[i].prime.len == cvc->primeOrModuluslen) && !memcmp(curves[i].prime.value, cvc->primeOrModulus, cvc->primeOrModuluslen)) {
+		if ((curves[i].prime.len == cvc->primeOrModuluslen) &&
+				!memcmp(curves[i].prime.value, cvc->primeOrModulus, cvc->primeOrModuluslen) &&
+				(curves[i].coefficientA.len == cvc->coefficientAorExponentlen) &&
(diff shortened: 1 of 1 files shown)


dengert:
Forced push with code style line indent.

Jakuje:
Thank you! I believe this is worth including in the next release.

embetrix:
`immediately after https://github.com/OpenSC/OpenSC/pull/3601 which adds the twisted curves and the problem.`

Issue #3613, opened 2026-03-22
DNIe 4.0: `C_Sign` fails with `CKR_USER_NOT_LOGGED_IN` after successful login


cprecioso:
### Describe the bug

After a successful `C_Login`, the second `C_Sign` operation fails with `CKR_USER_NOT_LOGGED_IN (0x101)`. The first signature succeeds, but any subsequent signature in the same session fails.

### Steps to reproduce

1. Insert DNIe 4.0 card
2. Run `pkcs11-tool --test --login --pin <PIN> -v`

### Expected behavior

All signature tests pass.

### Actual behavior

The first signature succeeds, but the second fails:

```
Signatures (currently only for RSA)
  testing key 0 (KprivAutenticacion)
error: PKCS11 function C_Sign failed: rv = CKR_USER_NOT_LOGGED_IN (0x101)
Aborting.
```

### OpenSC version

`master` (at `34d750a`)

### Operating system

macOS Tahoe 26.3.1 (25D2128)

### Smart card model or security token

DNIe 4.0

### Application used with smart card

`pkcs11-tool`

### Logs and debug output

```shell
C_SeedRandom() and C_GenerateRandom():
  seeding (C_SeedRandom) not supported
  seems to be OK
Digests:
  all 4 digest functions seem to work
  MD5: OK
  RIPEMD160: OK
  SHA-1: OK
  SHA256: OK
Signatures (currently only for RSA)
  testing key 0 (KprivAutenticacion) 
error: PKCS11 function C_Sign failed: rv = CKR_USER_NOT_LOGGED_IN (0x101)
Aborting.
```

### Checklist

- [x] I am using a supported OpenSC version
- [x] I have searched existing issues to avoid duplicates

nicolasgutierrezdev:
This might be due to the DNIe wiping it's pin-verified state after signature. With OpenSC failing to re-verify. I don't have a card to test the hypothesis though.

frankmorgner:
If this should really be the problem, then you can check by creating two signatures: the first should succeeed, the second (without re-verification of the PIN) should fail.

DNIe's PKCS#15 profile is parsed from the document itself. The related option would be the `user_consent` on the private key object. It needs to be set to `1` in order to re-verify the PIN every time. You may check this flag on the private key via `pkcs15-tool -D`.

[...]

PR #3614, opened 2026-03-22
card-dnie: Tear down SM channel before PIN re-verification on DNIe 3.0

cprecioso:
- Fixes https://github.com/OpenSC/OpenSC/issues/3613

> [!NOTE]
> This issue was found and fixed through Claude. I understand the code at a generic level but I lack the SmartCard-specific knowledge to get the nuances. I tried the fix locally with my ID, and it solves the issue I was seeing.

Card tested: DNIe 4.0 (Spanish national ID card), identified by `opensc-tool -n` as `dnie`.

## Problem

DNIe 3.0 keys require re-authentication before each private key operation (effectively `CKA_ALWAYS_AUTHENTICATE`), but are not tagged with this attribute in the card's PKCS#15 data. OpenSC handles this via PIN cache revalidation: when a signing operation gets SW 6982, `sc_pkcs15_pincache_revalidate()` re-verifies the cached PIN by calling `dnie_pin_verify()`.

During the initial login, `dnie_pin_verify()` works correctly because the SM channel is already OFF (from `dnie_init()`). However, during revalidation the DATA SM channel is still active. `dnie_pin_verify()` switches the CWA provider to PIN mode and calls `cwa_create_secure_channel(CWA_SM_ON)` without first tearing down the active DATA SM channel. The PIN SM establishment and PIN verification both succeed, but the subsequent DATA SM re-establishment fails — the card returns corrupted responses during certificate reading via GET RESPONSE, causing the entire revalidation to fail with `CKR_USER_NOT_LOGGED_IN`.

## Fix

Add `cwa_create_secure_channel(card, provider, CWA_SM_OFF)` at the start of `dnie_pin_verify()` for DNIe 3.0, before switching to the PIN provider. This mirrors what `dnie_init()` already does during card initialization: set `SM_MODE_NONE` on the host and send a plain SELECT MF to close the card-side SM channel. When SM is already OFF (e.g., during the initial login), the teardown is a harmless no-op.

##### Checklist

### DIFF
--- a/src/libopensc/card-dnie.c
+++ b/src/libopensc/card-dnie.c
@@ -2116,6 +2116,9 @@ static int dnie_pin_verify(struct sc_card *card,
 	if (card->atr.value[15] >= DNIE_30_VERSION) {
 		/* the provider should be prepared for using PIN information */
 		sc_log(card->ctx, "DNIe 3.0 detected doing PIN initialization");
+		/* Tear down any active SM channel first to avoid desynchronization
+		 * when re-establishing SM (e.g. during PIN cache revalidation) */
+		cwa_create_secure_channel(card, GET_DNIE_PRIV_DATA(card)->cwa_provider, CWA_SM_OFF);
 		dnie_change_cwa_provider_to_pin(card);
 	}
 	res = cwa_create_secure_channel(card, GET_DNIE_PRIV_DATA(card)->cwa_provider, CWA_SM_ON);


cprecioso:
Will do! I'm traveling right now so I'll do it next week. Thanks!

cprecioso:
I sat down to get you the logs, but I am having a lot of trouble getting any reproducible behaviour from my ID card. I rebased my patch on top of the latest `master` and now I every time I run `pkcs11-tool` it trips up in a completely different place each time (both with my patch and without).
I think I should withdraw this PR, since I can't confirm anything works right now.

frankmorgner:
Thanks for the feedback

PR #3620, opened 2026-03-23
pkcs11: fix infinite card detection loop for unsupported cards

### LINKED ISSUE #3471: OpenSC pkcs11 module making Firefox ESR unusable on Debian Trixie when Croatian eID or Certilia cards are connected to smartcard reader

### LINKED ISSUE #3616: OpenSC sends infinite SELECT APPLICATION APDUs on unrecognized cards (Italian CIE)
### Describe the bug

When an Italian CIE (Carta d'Identità Elettronica) smart card is inserted in a reader, OpenSC enters an infinite loop sending SELECT APPLICATION APDUs with incrementing AID suffixes. The card does not match any OpenSC driver, but instead of giving up after probing, OpenSC keeps retrying indefinitely.

This monopolizes the PC/SC channel, causing:
- The card reader LED to blink continuously
- Any other PKCS#11 module sharing the same reader to be blocked
- Firefox to freeze (pages never load) when OpenSC is loaded as a security module, because the PC/SC socket is permanently busy

The problem is related to #3108 (probing APDUs interfering with unsupported applications) but is more severe: it's not just interference with a running applet, it's an infinite loop that renders the entire smart card subsystem unusable.

### Steps to reproduce

1. Have a PC/SC reader connected with an Italian CIE card inserted
2. Load `opensc-pkcs11.so` in Firefox (or any application that uses it)
3. Observe the reader LED blinking continuously
4. Firefox UI becomes sluggish and web pages never finish loading

### Expected behavior

OpenSC should probe the card, fail to match any known driver/ATR, and stop. Subsequent calls to `C_GetSlotInfo` should return without attempting further communication with the card.

### Actual behavior

OpenSC sends an endless stream of SELECT APPLICATION APDUs to the card via pcscd. The APDU pattern observed via `strace` on the pcscd socket is:

```
00 A4 04 00 07 A0 00 00 00 79 01 04
00 A4 04 00 07 A0 00 00 00 79 01 05
00 A4 04 00 07 A0 00 00 00 79 01 06
00 A4 04 00 07 A0 00 00 00 79 01 07
00 A4 04 00 07 A0 00 00 00 79 01 08
00 A4 04 00 07 A0 00 00 00 79 01 09
00 A4 04 00 07 A0 00 00 00 79 01 0A
... (continues indefinitely with incrementing last byte)
```

Each APDU is a SELECT APPLICATION (`00 A4 04 00`) for AID `A0 00 00 00 79 01 xx` with an incrementing suffix. The card responds with an error (file/application not found) each time, but OpenSC never stops retrying.

Between each SELECT, OpenSC also performs what appears to be session management (disconnect/reconnect or transaction begin/end), making each cycle take ~15-20ms. This completely saturates the PC/SC channel.

## Impact

This issue affects **any Linux user** who has:
- An Italian CIE in a smart card reader, AND
- OpenSC installed (which is common on Linux systems, often pulled in as a dependency)

The CIE is the Italian national electronic identity card, issued to all Italian citizens. There are tens of millions of these cards in circulation. The card is not supported by OpenSC (see #2486, closed as not planned), but OpenSC should fail gracefully on unsupported cards rather than entering an infinite loop.

The issue also impacts users who have a third-party PKCS#11 module (e.g., the official CIE middleware or custom modules) loaded alongside OpenSC in Firefox: the two modules contend for the same PC/SC channel, and OpenSC's infinite probing loop blocks the other module from functioning.

## Suggested Fix

When card probing fails to match any known driver, OpenSC should:

1. **Stop probing after a bounded number of attempts** — do not retry SELECT APPLICATION indefinitely
2. **Cache the negative result** — remember that this ATR/card was not recognized, so subsequent `C_GetSlotInfo` / `C_GetTokenInfo` calls do not re-trigger probing
3. **Consider adding the CIE ATR to a known-unsupported list** so that probing is skipped entirely for this widely deployed card

Alternatively, an `opensc.conf` workaround using `card_atr` with `flags = "ignore"` could be documented, but this requires manual configuration by end users and does not solve the problem for default installations.

### OpenSC version

0.27.0~rc2-1

### Operating system

Debian/sid

3B 8E 80 01 80 31 80 65 49 54 4E 58 50 12 0F FF 82 90 F0 (can be sligtly different)

Firefox

### Logs and debug output

```shell
00000005 [139975727244992] ../src/ifdhandler.c:1433:IFDHTransmitToICC() usb:ae68/8001:libudev:0:/dev/bus/usb/001/002 (lun: 0)
00023398 [139975727244992] ../src/winscard.c:1650:SCardTransmit() UnrefReader() count was: 2
00000014 [139975727244992] ../src/winscard_svc.c:739:ContextThread() TRANSMIT for client 16, rv=SCARD_S_SUCCESS
00000065 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: END_TRANSACTION from client 16
00000010 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1236:SCardEndTransaction() Status: rv=SCARD_S_SUCCESS
00000004 [139975727244992] ../src/winscard.c:1239:SCardEndTransaction() UnrefReader() count was: 2
00000003 [139975727244992] ../src/winscard_svc.c:634:ContextThread() END_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000085 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: BEGIN_TRANSACTION from client 16
00000008 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1084:SCardBeginTransaction() Status: rv=SCARD_S_SUCCESS
00000004 [139975727244992] ../src/winscard.c:1087:SCardBeginTransaction() UnrefReader() count was: 2
00000004 [139975727244992] ../src/winscard_svc.c:618:ContextThread() BEGIN_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000073 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: TRANSMIT from client 16
00000010 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1597:SCardTransmit() Send Protocol: T=1
00000005 [139975727244992] ../src/ifdhandler.c:1433:IFDHTransmitToICC() usb:ae68/8001:libudev:0:/dev/bus/usb/001/002 (lun: 0)
00087068 [139975727244992] ../src/winscard.c:1650:SCardTransmit() UnrefReader() count was: 2
00000011 [139975727244992] ../src/winscard_svc.c:739:ContextThread() TRANSMIT for client 16, rv=SCARD_S_SUCCESS
00000056 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: END_TRANSACTION from client 16
00000009 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1236:SCardEndTransaction() Status: rv=SCARD_S_SUCCESS
00000004 [139975727244992] ../src/winscard.c:1239:SCardEndTransaction() UnrefReader() count was: 2
00000003 [139975727244992] ../src/winscard_svc.c:634:ContextThread() END_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000086 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: BEGIN_TRANSACTION from client 16
00000007 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1084:SCardBeginTransaction() Status: rv=SCARD_S_SUCCESS
00000003 [139975727244992] ../src/winscard.c:1087:SCardBeginTransaction() UnrefReader() count was: 2
00000004 [139975727244992] ../src/winscard_svc.c:618:ContextThread() BEGIN_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000086 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: TRANSMIT from client 16
00000011 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000003 [139975727244992] ../src/winscard.c:1597:SCardTransmit() Send Protocol: T=1
00000005 [139975727244992] ../src/ifdhandler.c:1433:IFDHTransmitToICC() usb:ae68/8001:libudev:0:/dev/bus/usb/001/002 (lun: 0)
00017193 [139975727244992] ../src/winscard.c:1650:SCardTransmit() UnrefReader() count was: 2
00000012 [139975727244992] ../src/winscard_svc.c:739:ContextThread() TRANSMIT for client 16, rv=SCARD_S_SUCCESS
00000123 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: END_TRANSACTION from client 16
00000011 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000003 [139975727244992] ../src/winscard.c:1236:SCardEndTransaction() Status: rv=SCARD_S_SUCCESS
00000003 [139975727244992] ../src/winscard.c:1239:SCardEndTransaction() UnrefReader() count was: 2
00000003 [139975727244992] ../src/winscard_svc.c:634:ContextThread() END_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000101 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: BEGIN_TRANSACTION from client 16
00000009 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000004 [139975727244992] ../src/winscard.c:1084:SCardBeginTransaction() Status: rv=SCARD_S_SUCCESS
00000003 [139975727244992] ../src/winscard.c:1087:SCardBeginTransaction() UnrefReader() count was: 2
00000005 [139975727244992] ../src/winscard_svc.c:618:ContextThread() BEGIN_TRANSACTION for client 16, rv=SCARD_S_SUCCESS
00000120 [139975727244992] ../src/winscard_svc.c:383:ContextThread() Received command: TRANSMIT from client 16
00000013 [139975727244992] ../src/readerfactory.c:871:RFReaderInfoById() RefReader() count was: 1
00000006 [139975727244992] ../src/winscard.c:1597:SCardTransmit() Send Protocol: T=1
00000006 [139975727244992] ../src/ifdhandler.c:1433:IFDHTransmitToICC() usb:ae68/8001:libudev:0:/dev/bus/usb/001/002 (lun: 0)
```

### Checklist

sherpya:
## Summary

When an unsupported card is inserted (e.g. Italian CIE, Croatian eID, Certilia), `sc_connect_card()` fails because no driver matches. However the reader keeps `SC_READER_CARD_PRESENT` set, so every subsequent call to `card_detect()` — triggered by Firefox polling `C_GetSlotInfo` every ~1 second — re-attempts `sc_connect_card()` on the same card, creating an infinite loop that saturates the PC/SC channel and freezes the application.

This adds a `SC_READER_CARD_INVALID` reader flag to cache the failed connection result:

- **`slot.c`**: When `sc_connect_card()` fails, set `SC_READER_CARD_INVALID` on the reader. On subsequent calls to `card_detect()`, if the flag is present, return `CKR_TOKEN_NOT_RECOGNIZED` immediately without re-attempting the connection.
- **`reader-pcsc.c` / `reader-cryptotokenkit.m`**: Clear the flag when the card is physically changed or removed, so a new card will be properly detected.

`CKR_TOKEN_NOT_RECOGNIZED` is already handled correctly in `C_GetSlotInfo` — it sets `CKF_TOKEN_PRESENT` (card is physically there) and converts the return value to `CKR_OK` (not a fatal error).

## Impact

This affects millions of users with national ID cards (Italian CIE, Croatian eID, etc.) on Linux systems where OpenSC is installed as a dependency and loaded automatically by p11-kit.

Fixes #3471
Fixes #3616

## Test plan

- [ ] Insert an unsupported card → verify single connection attempt, no loop, Firefox remains responsive
- [ ] Remove and re-insert the same card → verify detection is re-attempted
- [ ] Insert a supported card after an unsupported one → verify it works normally
- [ ] Verify no regressions with supported cards

### DIFF
--- a/.github/test-oseid.sh
+++ b/.github/test-oseid.sh
@@ -51,6 +51,8 @@ echo | ./OsEID-tool INIT
 ./OsEID-tool EC-ECDH-TEST
 ./OsEID-tool UNWRAP-WRAP-TEST
 ./OsEID-tool DES-AES-UPLOAD-KEYS
+export OPENSC_DEBUG=255
+export OsEID_DEBUG=-1
 ./OsEID-tool SYM-CRYPT-TEST
 ./OsEID-tool ERASE-CARD
 
--- a/.github/test-oseid.sh
+++ b/.github/test-oseid.sh
@@ -51,9 +51,7 @@ echo | ./OsEID-tool INIT
 ./OsEID-tool EC-ECDH-TEST
 ./OsEID-tool UNWRAP-WRAP-TEST
 ./OsEID-tool DES-AES-UPLOAD-KEYS
-export OPENSC_DEBUG=255
-export OsEID_DEBUG=-1
-./OsEID-tool SYM-CRYPT-TEST
+#./OsEID-tool SYM-CRYPT-TEST
 ./OsEID-tool ERASE-CARD
 
 # initialize card for p11test
--- a/src/pkcs11/framework-pkcs15.c
+++ b/src/pkcs11/framework-pkcs15.c
@@ -5818,7 +5818,12 @@ pkcs15_skey_encrypt(struct sc_pkcs11_session *session, void *obj,
 		return sc_to_cryptoki_error(rv, "C_Encrypt...");
 
 	/* pointer CK_ULONG_PTR to size_t conversion */
-	lpEncryptedDataLen = pulEncryptedDataLen ? &lEncryptedDataLen : NULL;
+	if (pulEncryptedDataLen) {
+		lEncryptedDataLen = *pulEncryptedDataLen;
+		lpEncryptedDataLen = &lEncryptedDataLen;
+	} else {
+		lpEncryptedDataLen = NULL;
+	}
 
 	rv = sc_pkcs15_encrypt_sym(fw_data->p15_card, skey->prv_p15obj, flags,
 			pData, ulDataLen, pEncryptedData, lpEncryptedDataLen,
--- a/src/pkcs11/sc-pkcs11.h
+++ b/src/pkcs11/sc-pkcs11.h
@@ -210,7 +210,8 @@ struct sc_pkcs11_card {
  * slot alive. PKCS#11 2.30 allows allows adding but not removing slots until
  * the application calls `C_GetSlotList` with `NULL`. This flag tracks the
  * visibility to the application */
-#define SC_PKCS11_SLOT_FLAG_SEEN 1
+#define SC_PKCS11_SLOT_FLAG_SEEN	 1
+#define SC_PKCS11_SLOT_FLAG_CARD_INVALID 2
 
 struct sc_pkcs11_slot {
 	CK_SLOT_ID id;			/* ID of the slot */
--- a/src/pkcs11/slot.c
+++ b/src/pkcs11/slot.c
@@ -197,6 +197,7 @@ CK_RV card_removed(sc_reader_t * reader)
 			/* Save the "card" object */
 			if (slot->p11card)
 				p11card = slot->p11card;
+			slot->flags &= ~SC_PKCS11_SLOT_FLAG_CARD_INVALID;
 			slot_token_removed(slot->id);
 		}
 	}
@@ -210,6 +211,7 @@ CK_RV card_removed(sc_reader_t * reader)
 CK_RV card_detect(sc_reader_t *reader)
 {
 	struct sc_pkcs11_card *p11card = NULL;
+	sc_pkcs11_slot_t *first_slot = NULL;
 	int free_p11card = 0;
 	int rc;
 	CK_RV rv;
@@ -249,9 +251,17 @@ again:
 	for (i=0; i<list_size(&virtual_slots); i++) {
 		sc_pkcs11_slot_t *slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, i);
 		if (slot->reader == reader) {
+			if (first_slot == NULL)
+				first_slot = slot;
 			p11card = slot->p11card;
 			break;
 		}
+	}
+
+	/* Card present but previously found to be unsupported */
+	if (first_slot != NULL && first_slot->flags & SC_PKCS11_SLOT_FLAG_CARD_INVALID) {
+		sc_log(context, "%s: card present but unsupported (cached)", reader->name);
+		return CKR_TOKEN_NOT_RECOGNIZED;
 	}
 
 	/* Detect the card if it's not known already */
@@ -269,6 +279,8 @@ again:
 		rc = sc_connect_card(reader, &p11card->card);
 		if (rc != SC_SUCCESS) {
 			sc_log(context, "%s: SC connect card error %i", reader->name, rc);
+			if (first_slot != NULL)
+				first_slot->flags |= SC_PKCS11_SLOT_FLAG_CARD_INVALID;
 			rv = sc_to_cryptoki_error(rc, NULL);
 			goto fail;
 		}


frankmorgner (changes requested):
Given that this functionally actually works, the flag should be. handled on the PKCS#11 layer. Currently it is defined at the low level for card reader (driver). So if a flag helps here, it should be added to the `slot` object rather than to the `reader` object in `slot.c`

frankmorgner:
Did you try #3126 if it fixes your issue? This fixes a state management problem, which leads to the recurring queries of the card. It was not adopted because we have lacking tests on different platforms...

sherpya:
> Did you try #3126 if it fixes your issue? This fixes a state management problem, which leads to the recurring queries of the card. It was not adopted because we have lacking tests on different platforms...

it doesn't, here the log
[opensc.log](https://github.com/user-attachments/files/26185327/opensc.log)

I'll update the PR

sherpya:
new commit is less intrusive, I've tested also using a supported card and I don't see regressions

dengert:
Have you look at: https://github.com/OpenSC/OpenSC/pull/3472 From August, 2005, that has not been commited?
This should fix the problem.

sherpya:
> Have you look at: #3472 From August, 2005, that has not been commited? This should fix the problem.

yes it works, but it's at reader level, you were against it, as long it works, not a problem for me

dengert:
> > Have you look at: https://github.com/OpenSC/OpenSC/pull/3472 From August, 2005, that has not been commited? This should fix the problem.

> yes it works, but it's at reader level, you were against it, as long it works, not a problem for me

@frankmorgner here is proof #3472 works. 

The flag needs to be at the reader level to keep the overhead down, as the flag is only cleared if the card is reinserted no mater what the code above the reader level does.

frankmorgner:
The fundamental problem is that the state tracking with `refresh_attributes` (at reader level) is not bound to the PKCS#11 level, which calls sc_connect_card. The reader level may update its state without the PKCS#11 level noticing it.

#3472 solves this by introducing a massive change that mixes the state machines on all layers. (It was also the first attempt to solve the issue for this very PR.) This approach is overly complicated and hard to maintain in the long run. #3620 now keeps this at the PKCS#11 layer, which seems like a short term fix that can be removed later if state tracking is implemented properly some day. The drawback, however, is that #3620 may miss some corner cases where a changed card in the same reader may not be detected properly, for example.

In my opinion, the best approach to fix state tracking is to create a dedicated PC/SC event state on the PKCS#11 layer, which is only updated when PKCS#11 requests this. This event state must be independant from refresh_attributes. This is what https://github.com/OpenSC/OpenSC/pull/3126 tries to do. Unfortunetely, I never had time to complete this beyond the first draft.

metsma (approved):

frankmorgner (approved):

Jakuje:
This breaks the oseid tests for some reason.

frankmorgner:
yes, oseid test case needs to be fixed before merging. This may indicate a more serious problem that could affect others as well.

is there a way to get some logs?

dengert:
@frankmorgner said:
> https://github.com/OpenSC/OpenSC/pull/3472 solves this by introducing a massive change that mixes the state machines on all layers. (It was also the first attempt to solve the issue for this very PR.) This approach is overly complicated and hard to maintain in the long run.

I would not call it massive or complicated.
 Most of the changes in pkcs11 files are to use the "context" created in  "pkcs11-global.c" which is called by C_Initialize. This would then allow OpenSC log functions to be called, from routines where `p11card == NULL` or `p11card->card == NULL`  I  would call this an improvement. 

The p11card may be provide or allocated in `card_detect`  while trying to setup a slot.  If the card has been removed the SC_READER_CARD_INVALID is reset.

This does not require a PC/SC event state as the problem is not with an event, the problem, is PKCS11 keeps trying to call sc_connect_card for a card that has failed to connect as not supported by OpenSC. The flag is cleared when a new token or card is inserted. So there is no loop other then the user keeps trying to insert an unsupported card or token.

frankmorgner:
I rebased the PR onto current master, but it still fails the oseid test case. This failure seems to be isolated to the changes in this PR, it doesn't occur in other PRs or on `master`. Without solving this issue, the suggested changes cannot be merged.

frankmorgner:
https://github.com/OpenSC/OpenSC/pull/3620 solves the same problem and fails the same tests. I assume that there is some problem in the oseid configuration/initialisation, but in any case this needs to be solved before merging.

frankmorgner:
@sherpya , could you check if the infinite loop is also gone with https://github.com/OpenSC/OpenSC/pull/3126?

sherpya:
> @sherpya , could you check if the infinite loop is also gone with #3126?

look good to me, it spams a bit but then it stops

[pcscd.log](https://github.com/user-attachments/files/29256581/pcscd.log)

frankmorgner:
This kind of *spam* is intended: The application (Firefox) *spams* requests whether or not there is a change with the slot. I am using the dedicated PC/SC functionality to detect this rather than asking the card, which was the lagging I/O operation in your case. I think this approach is a good balance between correctly acting upon the application's request and keeping the performance up by avoiding I/O.

Since your test was successful, we will adopt #3126 then. Thank you.

PR #3622, opened 2026-03-23
ci: Simplify kryoptic tests by using packaged version

Jakuje:
Building kryoptic is slow, but we have recent version in Fedora now so we can use that one to get the CI 

##### Checklist

- [ ] macOS token is tested


Jakuje:
CI only. Merging

PR #3623, opened 2026-03-23
pkcs11-tool: Print more information about RSA keys

### LINKED ISSUE #3617: compact representation of public part
### Problem statement

TODO

### Proposed solution

TODO

### Alternatives considered

_No response_

### Additional context

_No response_

Jakuje:
Fixes #3617

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -6474,7 +6474,18 @@ print_hex(const u8 *bin_input, size_t input_size, int separator, bool newline)
 {
 	char out[BYTES_PER_LINE * 3] = {0};
 	size_t out_len = BYTES_PER_LINE * 3;
-	unsigned int n;
+	unsigned int dec = 0;
+	unsigned int n, i;
+
+	/* If it is small enough number, convert it to deccimal */
+	if (input_size <= sizeof(unsigned int)) {
+		for (i = 0; i < input_size; i++) {
+			dec = (dec << 8) + (bin_input[i] & 0xff);
+		}
+		if (dec != 0) {
+			printf("%u (0x", dec);
+		}
+	}
 
 	for (n = 0; n < input_size; n += BYTES_PER_LINE) {
 		size_t chunk_len = MIN(input_size - n, BYTES_PER_LINE);
@@ -6482,7 +6493,7 @@ print_hex(const u8 *bin_input, size_t input_size, int separator, bool newline)
 				? "\n              " /* continuation of block */
 				: (newline ? "\n" : ""); /* end of the block */
 		sc_bin_to_hex(bin_input + n, chunk_len, out, out_len, separator);
-		printf("%s%s", out, indent);
+		printf("%s%s%s", out, (dec != 0 ? ")" : ""), indent);
 	}
 }
 
@@ -6546,8 +6557,8 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
 			}
 			public_exponent = getPUBLIC_EXPONENT(sess, obj, &public_exponent_len);
 			if (public_exponent) {
-				printf("  Public exponent: ");
-				print_hex(public_exponent, public_exponent_len, ':', true);
+				printf("  Public exp: ");
+				print_hex(public_exponent, public_exponent_len, 0, true);
 				free(public_exponent);
 			}
 		}


src/tools/pkcs11-tool.c
```diff
@@ -6500,11 +6500,45 @@ show_key(CK_SESSION_HANDLE sess, CK_OBJECT_HANDLE obj)
...
+			}
+			public_exponent = getPUBLIC_EXPONENT(sess, obj, &public_exponent_len);
+			if (public_exponent) {
+				unsigned int n;
+
+				printf("  PUBLIC_EXPONENT: ");
+				for (n = 0; n < public_exponent_len; n++) {
+					if (n && (n % 32) == 0)
+						printf("\n              ");
+					printf("%02x", public_exponent[n]);
```
frankmorgner: I would prefer the use of `sc_hex_dump` (prints 16 bytes with ASCII per line) or sc_bin_to_hex (chunk wise for each of the lines).

frankmorgner (changes requested):
Are there other public keys that would be useful to print?
Is the public key maybe too much output for non-verbose invocation (i.e. should we restrict this to -v only as suggested in the issue)?

Jakuje:
> Are there other public keys that would be useful to print?

I think all other public keys are printed.

>  Is the public key maybe too much output for non-verbose invocation (i.e. should we restrict this to -v only as suggested in the issue)?

I do not think its too much information. For 2k key is just like this (much better than the ML-DSA keys):
```
Public Key Object; RSA  2048 bits                                                                              
  MODULUS:    c318e050cca74c1a9076e11620896bf7167400a5898113fd5cadc9b0d1ea9809                                 
              25de98a8921bf09c568e764a73f954b0076a59a22f829411e3c7e45ffadeca32                                 
              803dc50b3c4a748a70f05443a1707529c7bb5b6cd28832bb2f5d0db52ea70dda                                 
              8e63ec156e15bfdce3090c189d84d68025bff31834ceba1cfc5181cdf762c1f3                                 
              43400a12d7f9186fa55476da243fe3db6d660e0e5bd5c974162ac1e45839df33                                 
              4d1e5f4ced07ccb68c44c6249dd1c72943f2f6bdbfd1976569e16399c7029e25                                 
              32f456f4217af19ff0c8d43954d28e951f72126a50c940ee7f18347f2004ed12                                 
              1154154bdfa81fff32c26f951ba8bb5e63545d42f7750394e25e48de2b1ab191                                 
  PUBLIC_EXPONENT: 010001                                                                                      
  label:      RSA2048                                                                                          
  ID:         01                                                                                               
  Usage:      encrypt, verify                                                                                  
  Access:     local                                                                                            
  Unique ID:  5cb8ac7f-155e-4ffb-892b-1ca571742b21                                                             
  uri:        pkcs11:model=v1;manufacturer=Kryoptic%20Project;serial=159dd6d35f7bb1e6;token=Kryoptic%20Token;id=%01;object=RSA2048;type=public
```

src/tools/pkcs11-tool.c
Jakuje: I think the `sc_hex_dump()` is for dumping partially-printable strings (ASN1 buffers, certs) with the right side showing the ASCII values. The modulus is really just "random" so I do not think using this is helpful here.

Using `sc_bin_to_hex()` might work. I copied this snippet from other places dumping buffers so some simplification would be useful. I will give it a try.

Jakuje: Similar method is used to print both short buffers and longs. I would say that the short (OIDs, modulus) could make use of the separators.
The ugly part is that this writes fixed indent, but some uses write some more information behind the string so I am not completely happy with the solution, but at least it removes some duplicated code.


```
Private Key Object; RSA  2048 bits
  MODULUS:    b86caac7842df94115f9a4b6ced816d5637f3f47c8438c3a379634d1f0571bf0
              f0c8d7329515f15e3397698f466d3a7b29289c8642f5be9238018700e8b4881b
              aeb5372a4491d6c928afafa9a3601b64bc6257fcd2e6733333e5a1e405b322c2
              1d891a7dd87c14eb02dbbcfe1761db32ba38b94de4c55eb961657643079c16d5
              26eb093a5d8f034a8a2dfea72580de2870b0c9b94cfde754666b7ceac64bd3dd
              40698f8cd9c1a3129f0f55c565d8485b9b1b1ac0045fc9a9bd695820230d37c3
              7615175f99386df584d4a0778bb3c29d67593ca0da74887c0568a13ad5b0a35b
              13596233a7fc8659fcd9f08846c42ba5d53a0d7727be673c580e8fcf4ab6b155
  PUBLIC_EXPONENT: 01:00:01
  label:      RSA2048                                                                                     
  ID:         01                                                              

```

```
Public Key Object; EC  EC_POINT 521 bits
  EC_POINT:   04818504017d8f4bd5cd507a252c951beb44f06a2106195a9153b0ea161a9aa9
              47e84f9f154d5ee0aa760a95ce4b65ba58f7c88c61077925266d35e4e23274bf
              c60788f5b01900123fdfb6976917b1f70a61d3a44035e4aaed82262c3bc45c6c
              787a22cc137395c8f74124eb583a4fb2d133292db427cd49722844718d2d2820
              31b36a289fd1ff3c
  EC_PARAMS:  06:05:2b:81:04:00:23 ("prime521v1" OID:"1.3.132.0.35")
```

Jakuje:
Looking at the output now, I think I do not have to use all upper case letters as the other labels are formatted normally. Let me just fix that very fast.

frankmorgner:
The public exponent looks weird... would it make sense to display this as normal integer? Maybe add "in hex" if it is *not* in decimal?

frankmorgner:
Also related, the man page from pkcs11-tool currently is not very consistent with *input* format of data: sometimes hex is used with `0x` prefix, sometimes it is not. sometimes input is in decimal.

My preference for the exponent would be this:
```
PUBLIC_EXPONENT: 65537 (0x010001)
```

Jakuje:
Updated:
```
Public Key Object; RSA  2048 bits
  Modulus:    dbc6433a542362d2093cce39839d1f14101005f8d7b28c53f23a2bde6bb172c2
              416c7e2bd2a1eb1ca5d70d294cb752fb685b02a9e6e134ca2417069dd3538a4a
              7f9ee7cdfad480ea689dc1065486e6e7f2c2b83bf7dd10bb17708672fe92df6b
              fa19f5d7373a8421aa98f1e046b24893368ef9bfbfce7da2b2c2d25939b28c15
              0a3ec9deb4ca68fd02f208887f6c26a56f370d331a5a19da8f8a74d1e63a7dda
              89a52f677a55c41ce35da16693b53a7646ffea0be196cf72c7adc651ba7707d6
              f1044b1bdf03b5ff0c2ba4a86074897447a8dd94d4bfc470f4bad20982413c94
              5488d1b16f7bf42723ffe1954ebf0c3c6164b91609d3fc3e6537e0bc03125907
  Public exp: 65537 (0x010001)
  label:      RSA2048
  ID:         1 (0x01)
```
It also prints now IDs in both decimal and hexadecimal (and any small enough number) and public exponent is aligned with the rest of the output.

PR #3625, opened 2026-03-23
MacOSX/build-openssl: Fix infinite loop on missing CLI arg

gkapetanakis:
This PR fixes a minor bug where the arg-checking loop in `MacOSX/build-openssl-macos.sh` could loop infinitely if insufficient arguments are given, because `shift 2` fails and thus does not reduce the value of `$#`.

Example that causes infinite loop: `./MacOSX/build-openssl-macos.sh -b`


Jakuje:
I think this is also ok to include in the release.

Issue #3628, opened 2026-03-28
do not execute p11test when build is without cmocka


petrovr:
### Describe the bug

p11test executable is build with condition TESTS (src/Makefile.am) and OPENSSL+CMOCKA(src/tests/p11test/Makefile.am)

The test-p11test.sh is added with OPENSSL+TESTS (tests/Makefile.am) i.e., makefile lacks statement if ENABLE_CMOCKA .

Because the executable file is missing, the test failed.

### Steps to reproduce

build without (or disable) cmocak
run tests

### Expected behavior

pass 

### Actual behavior

failed

### OpenSC version

any

### Operating system

any

### Logs and debug output

```shell

```

### Checklist

- [x] I am using a supported OpenSC version
- [x] I have searched existing issues to avoid duplicates

Jakuje:
Thanks for the report! Do you want to submit a PR with the fix? It should be quite easy to conditionalize the `test-p11test.sh` with `if ENABLE_CMOCKA` in https://github.com/OpenSC/OpenSC/blob/master/tests/Makefile.am

Jakuje:
Fixed with #3629. Thank you for the report! Should be in the next release due soon

PR #3629, opened 2026-03-30
Minor fixes for the dist scripts

Jakuje:
* Missing scripts that should be included in the release tarballs
* Exclude p11test when we miss cmocka
* Fix some paths to refer to correct SOURCE_PATH instead of BUILD_PATH

No idea how it worked before.

##### Checklist

- [ ] macOS token is tested


Jakuje:
Fixes: #3628.

PR #3631, opened 2026-03-30
OpenSC 0.27.1

Jakuje:
Fixup release to clean up artifact names.

This bumps the release number again and changes the logic so there is no tag name appended to the current version.

##### Checklist

- [ ] macOS token is tested


.github/workflows/windows.yml
```diff
@@ -41,18 +41,24 @@ jobs:
...
             if [[ "${{ github.base_ref }}" == "master" ]]; then
               SUFFIX="-pr${{ github.event.number }}"
             else
               SUFFIX="-${{ github.base_ref }}-pr${{ github.event.number }}"
             fi
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            if [[ "${{ github.ref_type }}" == "tag" ]]; then
+              if [[ "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                SUFFIX="" # regular release tag
+              elif [[ "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
```
metsma, line 54: [0-9]+\.[0-9]+\.[0-9]+(.+)?
Maybe this works to?

Jakuje: I am not sure what you mean here. The `.+` matches 1-n characters. Adding the question mark after that will just convert it to `.*` (0-n) characters which is overlapping with the first regex that as already matched. Removing the escape from the `.` between version parts is also not correct, as it would match any character instead of the `.` exclusively.

metsma: https://regex101.com/?regex=%5B0-9%5D%2B.%5B0-9%5D%2B.%5B0-9%5D%2B%28.%2B%29%3F&testString=0.27.1-rc1&flags=s&flavor=python&delimiter=%22
https://regex101.com/?regex=%5B0-9%5D%2B.%5B0-9%5D%2B.%5B0-9%5D%2B%28.%2B%29%3F&testString=0.27.1&flags=s&flavor=python&delimiter=%22

Jakuje: > https://regex101.com/?regex=%5B0-9%5D%2B.%5B0-9%5D%2B.%5B0-9%5D%2B%28.%2B%29%3F&testString=0.27.1-rc1&flags=s&flavor=python&delimiter=%22 https://regex101.com/?regex=%5B0-9%5D%2B.%5B0-9%5D%2B.%5B0-9%5D%2B%28.%2B%29%3F&testString=0.27.1&flags=s&flavor=python&delimiter=%22

yes, but it will also match `0a27b1`.

Ok then. I will merge it and retry the release.

Issue #3463, opened 2025-08-05
Migrate javacard tests using JCardSim to JCardEngine


Jakuje:
The JCardSim repository looks dead for years and we are already using our custom fork bringing in stuff we cared about in the CI (ECDSA support, test fixes):

https://github.com/Jakuje/jcardsim/commits/master/

@martinpaljak let us know some time ago that he did a more radical fork, which has some potential future as it is something he is using. The migration for the tests should not be complex

https://github.com/martinpaljak/jcardsim/wiki/Migration-from-legacy-jcardsim

frankmorgner:
In the meantime, Oracle has released a simulation kit. Windows only, though
https://docs.oracle.com/en/java/javacard/3.1/guide/install-and-setup-development-kit.html

martinpaljak:
@frankmorgner not really, Linux (x86) too. You can bridge to oracle simulator via jcardengine too. https://github.com/martinpaljak/JCardEngine/wiki/Command-Line-Tutorial,-Advanced#running-oracle-java-card-simulator-on-linux

PR #3632, opened 2026-04-01
Fix IsoApplet hard coded algorithm_ref

Disappear9:
Related to: https://github.com/philipWendland/IsoApplet/issues/45
Tested on card: J3R180 with IsoApplet v1

opensc-tool -n
```
Using reader with a card: Alcor Micro AU9540
Javacard with IsoApplet
```
##### Checklist
- [ ] macOS token is tested


Jakuje:
We have just the test with the 2k key. When we will get this fixed here an in the isoapplet repo, can you adjust also the following test line to generate the 4k key (or create another one):

https://github.com/OpenSC/OpenSC/blob/master/.github/test-isoapplet.sh#L74

It will need adjustment of the reference json file. I will try to keep an eye on that, but please remind me if I will forget:

https://github.com/OpenSC/OpenSC/blob/master/src/tests/p11test/isoapplet_ref_v1.json

Jakuje (approved):

Disappear9:
Now we wait: https://github.com/philipWendland/IsoApplet/pull/46

philipWendland:
(I am the maintainer of IsoApplet)

From my perspective, the changes here look good. The IsoApplet main branch now works with this change, I tested this with a physical card. I do not expect backwards compatibility issues.

However, the CI build here seems to fail and I do not see why.

Jakuje:
I am not sure if I will have time to look into that, but one thing to try might be to migrate from the jCardSim to JCardEngine (as suggested by Martin some time ago, see #3463). The new engine should have some more fixes than the unmaintained jCardSim, but I can not guarantee it will solve these.

frankmorgner:
> However, the CI build here seems to fail and I do not see why.

Our CI fails on IsoApplet v1 to install the applet into jCardSim, but succeeds with v0. Here is what I think could cause the failure (diff from IsoApplet v0/v1):

```diff
diff --git a/build.xml b/build.xml
index 55bcff3..3f2d786 100644
--- a/build.xml
+++ b/build.xml
@@ -7,8 +7,8 @@
     <description>Builds the project. </description>
     <target name="dist" description="generate the distribution">
         <tstamp/>
-        <javacard jckit="ext/sdks/jc320v25.1_kit">
-            <cap targetsdk="3.0.4" aid="f2:76:a2:88:bc:fb:a6:9d:34:f3:10" output="IsoApplet.cap" sources="src" version="1.0">
+        <javacard jckit="ext/sdks/jc310r20210706_kit">
+            <cap targetsdk="ext/sdks/jc304_kit" aid="f2:76:a2:88:bc:fb:a6:9d:34:f3:10" output="IsoApplet.cap" sources="src" version="1.0">
                 <applet class="xyz.wendland.javacard.pki.isoapplet.IsoApplet" aid="f2:76:a2:88:bc:fb:a6:9d:34:f3:10:01"/>
             </cap>
         </javacard>
```

Could it be that jCardSim is not compatible with the targetsdk from v1, @philipWendland ?

frankmorgner:
The date of the last successful run in *our* CI pipeline https://github.com/OpenSC/OpenSC/actions/runs/27008737435 correllates with the SDK change made the next day in https://github.com/philipWendland/IsoApplet/commit/20583f1f6fe3299ff5c44047ecb7c52da14f3a37

philipWendland:
The switch of jckit was intentional, I switched to the new SDK because it supports JDK 21.
The older one requires JDK 17 and openjdk-17 is not even in the current stable debian. If possible, I'd rather not switch back if we find a different solution.

You diff looks incorrect to me @frankmorgner . The targetsdk should not have changed, it is still "targetsdk="3.0.4" " on IsoApplet main. So only the jckit has changed. It is still possible that jcardsim might not be compatible with this version of IsoApplet (v1). I think the newly introduced test is the first one to test v1 of IsoApplet? But the time correlation you mentioned might indicate that is due to the different jckit? I.e. even though only a different jckit has been used (the cap file should be compatible with the same target SDK), jcardsim seems to have a problem with this change.

What would be the best solution here in your opinion @frankmorgner ? 
One part of this PR fixes a known bug in IsoApplet. The other part introduces a new unit test that seems to fail due to the used simulator. Can these issues be split?

frankmorgner:
The last working [version](https://github.com/OpenSC/OpenSC/compare/master...frankmorgner:OpenSC:isoapplet) with the [OpenSC CI](https://github.com/frankmorgner/OpenSC/actions/runs/28320407361/job/83901997996) of IsoApplet v1 is 

https://github.com/philipWendland/IsoApplet/commit/9cf87af45dc2949730a700a52a2dceb87353734c

I also tested the [direct child](https://github.com/philipWendland/IsoApplet/commit/b03379685724bb307c4c6cb25d4584042361955d), which failed in jCardSim. Note that this child is *not* the SDK change, but this part of the code:
```
        if( ! JCSystem.isObjectDeletionSupported()) {
            ISOException.throwIt(ISO7816.SW_FUNC_NOT_SUPPORTED);
        }
```
This exception is then converted to `0x6444` (internal for `SW_APPLET_CREATION_FAILED`). Obviously, the Simulator always returns [`false`](https://github.com/licel/jcardsim/blob/41511ec2a3d3dfa7a5fe2d07f5b4e56bc9c8dd56/src/main/java/com/licel/jcardsim/base/SimulatorRuntime.java#L644) here.

Since everything else works fine, just removing this check may be enough to solve the issue

philipWendland:
It was intentional to block the modern version of IsoApplet for smartcards that do not support RequestObjectDeletion. (I do not agree that it is obvious that JCardSim returns false here. There shouldn't even be persistent garbage in a simulator in a normal JVM, so jcardsim should implement the mechanism as no-op and return true for the check IMO. But this is out of scope for this PR anyway).

Is there a way to fix this in OpenSC's test scripts?

Given that most (or all) cards with JC version >= 3 support this, this is an edge case and I don't need to be pedantic. If you wish @frankmorgner I can revert this guard from the applet code.

frankmorgner:
For OpenSC's purposes, the situation currently is good as is and the jCardSIM problem may be resolved together with #3463. I think I would leave it as is.

PR #3633, opened 2026-04-02
Fix coverity issues reported in the recent lteid tool

Jakuje:
@tadas-s can you double-check this is working as expected?

##### Checklist

- [ ] macOS token is tested


tadas-s:
Hi @Jakuje,
Yup, all works as expected :+1:

PR #3642, opened 2026-04-09
Fix minor issues reported via security issues

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -1439,8 +1439,9 @@ end:
 static int
 decrypt_response(struct sc_card *card, unsigned char *in, size_t inlen, unsigned char *out, size_t * out_len)
 {
+	unsigned int cla = 0, tag = 0;
+	const unsigned char *p = in;
 	size_t cipher_len;
-	size_t i;
 	unsigned char iv[16] = {0};
 	unsigned char plaintext[4096] = {0};
 	epass2003_exdata *exdata = NULL;
@@ -1450,33 +1451,23 @@ decrypt_response(struct sc_card *card, unsigned char *in, size_t inlen, unsigned
 
 	exdata = (epass2003_exdata *)card->drv_data;
 
-	/* no cipher */
-	if (in[0] == 0x99)
-		return 0;
-
-	/* parse cipher length */
-	if (0x01 == in[2] && 0x82 != in[1]) {
-		cipher_len = in[1];
-		i = 3;
-	} else if (0x01 == in[3] && 0x81 == in[1]) {
-		cipher_len = in[2];
-		i = 4;
-	} else if (0x01 == in[4] && 0x82 == in[1]) {
-		cipher_len = in[2] * 0x100;
-		cipher_len += in[3];
-		i = 5;
-	} else {
+	if (SC_SUCCESS != sc_asn1_read_tag(&p, inlen, &cla, &tag, &cipher_len)
+			|| p == NULL) {
 		return -1;
 	}
+	if ((cla | tag) == 0x99)
+		return 0;
+	if ((cla | tag) != 0x87)
+		return -1;
 
-	if (cipher_len < 2 || i + cipher_len > inlen || cipher_len > sizeof plaintext)
+	if (cipher_len < 2 || cipher_len > sizeof plaintext)
 		return -1;
 
 	/* decrypt */
 	if (KEY_TYPE_AES == exdata->smtype)
-		aes128_decrypt_cbc(card, exdata->sk_enc, 16, iv, &in[i], cipher_len - 1, plaintext);
+		aes128_decrypt_cbc(card, exdata->sk_enc, 16, iv, p, cipher_len - 1, plaintext);
 	else
-		des3_decrypt_cbc(card, exdata->sk_enc, 16, iv, &in[i], cipher_len - 1, plaintext);
+		des3_decrypt_cbc(card, exdata->sk_enc, 16, iv, p, cipher_len - 1, plaintext);
 
 	/* unpadding */
 	while (0x80 != plaintext[cipher_len - 2] && (cipher_len > 2))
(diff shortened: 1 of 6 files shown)


src/libopensc/card-sc-hsm.c
```diff
@@ -812,7 +813,8 @@ static int sc_hsm_read_binary(sc_card_t *card,
 	cmdbuff[2] = (idx >> 8) & 0xFF;
 	cmdbuff[3] = idx & 0xFF;
 
-	assert(count <= sc_get_max_recv_size(card));
+	if (count <= sc_get_max_recv_size(card))
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INTERNAL);
```
Jakuje: afaik the logic should be reversed:
```suggestion
	if (count > sc_get_max_recv_size(card)) {
		LOG_FUNC_RETURN(ctx, SC_ERROR_INTERNAL);
	}
```
same for the other cases

frankmorgner: yes, sorry

src/libopensc/card-epass2003.c
```diff
@@ -1450,33 +1451,23 @@ decrypt_response(struct sc_card *card, unsigned char *in, size_t inlen, unsigned
...
-		i = 5;
-	} else {
+	if (SC_SUCCESS != sc_asn1_read_tag(&p, inlen, &cla, &tag, &cipher_len)
+			|| p == NULL) {
 		return -1;
 	}
+	if ((cla | tag) == 0x99)
+		return 0;
+	if ((cla | tag) != 0x87)
+		return -1;
```
Jakuje, line 1460: where does this condition come from? I do not see respective condition in the original code.

src/libopensc/iasecc-sdo.c
```diff
@@ -1263,15 +1263,23 @@ int
...
+	for (offs = 0; offs + 1 < data_len; ) {
+		rc = iasecc_parse_size(data + offs + 1, data_len - offs - 1, &size);
+		if (rc <= 0) {
+			LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_DATA);
+		}
+		size_size = rc;
+		if (offs + 1 + size_size + size > data_len) {
+			LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_DATA);
+		}
 		size_size = iasecc_parse_size(data + 1, data_len - 1, &size);
```
Jakuje: this line should be removed, should not it?

src/libopensc/pkcs15-pubkey.c
```diff
@@ -988,7 +988,7 @@ sc_pkcs15_read_pubkey(struct sc_pkcs15_card *p15card, const struct sc_pkcs15_obj
 		LOG_TEST_GOTO_ERR(ctx, r, "Failed to read public key file.");
 
 		if ((algorithm == SC_ALGORITHM_EC || algorithm == SC_ALGORITHM_EDDSA || algorithm == SC_ALGORITHM_XEDDSA)
-				&& *data == (SC_ASN1_TAG_SEQUENCE | SC_ASN1_TAG_CONSTRUCTED))
+				&& len > 0 && *data == (SC_ASN1_TAG_SEQUENCE | SC_ASN1_TAG_CONSTRUCTED))
```
Jakuje: Can `sc_pkcs15_decode_pubkey()` handle the `len <= 0`? Would not it be better to bail out explicitly before if this length is not valid, rather than falling back to the else branch?

frankmorgner: Yes, I think so. `sc_pkcs15_decode_pubkey()` directs calls to some ASN.1 function in `asn1.c`. Those functions should be able of handling a length of `0` and they are throwing an appropriate error code.

I prefer redirecting the call, because here we are only checking for spki pubkeys and I don't want to presume anything for other types of public keys.

src/libopensc/card-epass2003.c
frankmorgner: It is true that the original implementation only checked the tag to be different from `0x99`, but the comment before this function makes it clear that `0x87` is expected to encode the cipher.

src/libopensc/iasecc-sdo.c
frankmorgner: true. this is a copy/paste error

src/libopensc/pkcs15-pubkey.c
```diff
@@ -987,8 +987,7 @@ sc_pkcs15_read_pubkey(struct sc_pkcs15_card *p15card, const struct sc_pkcs15_obj
 		r = sc_pkcs15_read_file(p15card, &info->path, &data, &len, private_obj);
 		LOG_TEST_GOTO_ERR(ctx, r, "Failed to read public key file.");
 
-		if ((algorithm == SC_ALGORITHM_EC || algorithm == SC_ALGORITHM_EDDSA || algorithm == SC_ALGORITHM_XEDDSA)
-				&& *data == (SC_ASN1_TAG_SEQUENCE | SC_ASN1_TAG_CONSTRUCTED))
+		if ((algorithm == SC_ALGORITHM_EC || algorithm == SC_ALGORITHM_EDDSA || algorithm == SC_ALGORITHM_XEDDSA) && len > 0 && *data == (SC_ASN1_TAG_SEQUENCE | SC_ASN1_TAG_CONSTRUCTED))
```
Jakuje, line 990: nit: I would keep this on two lines for readability.

Jakuje (approved):

frankmorgner:
With https://github.com/OpenSC/OpenSC/pull/3669 merged now, do you think this PR can be closed now?

Jakuje:
> With #3669 merged now, do you think this PR can be closed now?

please, resolve the conflicts. I think we can merge it then. The only nit I had was about the formatting, otherwise it looks good to get merged (I think it was addressing different issues than in 3669).

frankmorgner:
I rebased and fixed the merge conflicts. Formatting changed according to your comment.

Jakuje (approved):
Thank you! Looks good! The CI failures look unrelated. I will have to look into them separately

PR #3645, opened 2026-04-13
pkcs11: Fix SHA1 mechanisms registration in FIPS mode

Jakuje:
With the FIPS mode detection done right, the FIPS provider does not have SHA1 mechanism, which causes registration fail. This skips all SHA1 based mechanisms in FIPS mode.

Unfortunately, I did not catch this before trying the OpenSC internals in fips mode over last couple of weeks.

##### Checklist

- [-] Documentation is added or updated
- [-] New files have a LGPL 2.1 license statement
- [ ] macOS token is tested

### DIFF
--- a/src/pkcs11/mechanism.c
+++ b/src/pkcs11/mechanism.c
@@ -1786,7 +1786,7 @@ sc_pkcs11_register_sign_and_hash_mechanism(struct sc_pkcs11_card *p11card,
 	CK_RV rv;
 
 	LOG_FUNC_CALLED(p11card->card->ctx);
-	sc_log(p11card->card->ctx, "mech = %lx, hash_mech = %lx", mech, hash_mech);
+	sc_log(p11card->card->ctx, "mech = 0x%lx, hash_mech = 0x%lx", mech, hash_mech);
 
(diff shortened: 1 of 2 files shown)


frankmorgner (approved):

Jakuje:
I also wanted to extend the CI to run also the opensc internals in FIPS mode (for example with the libcacard/virt_cacard), but it can land in separate PR later.

Actually thinking about this a bit further, the other option would be reverting the 91e61b9a7667ce0bf62e4b0e872e39a34295eb89 as the SHA1 mechanism is still in the provider (even though we do not like it)... and this problem is directly caused by this commit as the mechanism registration fails because the SHA1 is not found in the mechanism list.

Jakuje:
After a bit more thinking, I decided that going with the revert of 91e61b9a7667ce0bf62e4b0e872e39a34295eb89 will be cleaner. Please take a look.

Jakuje:
@frankmorgner can you review this one? Its just a revert, but should resolve the same issue, #3658 we got reported recently here too.

frankmorgner (approved):

PR #3655, opened 2026-04-15
Changes for OpenSSL-4.0.0

dengert:
Changes to get OpenSC to compile with OpenSSL-4.0.0

OpenSSL says:
"Opaque Structures: ASN1_STRING has been made opaque, and further work has been done to make X509 structures more memory-efficient and const-correct."

ASN1_STRING was made opaque thus requiring the use of ASN1_STRING_get0_data, and ASN1_STRING_length. These routines have been in OpenSSL since 1.1.1 and are also in libressl.

ASN1_STRING_get0_data returns a const pointer, so the data can not be changed. There  are a number of reverse() functions one of which was reversing the bytes in a EC pubkey.

Changes not addressed:
"Global Cleanup: libcrypto no longer uses atexit() for cleanup; OPENSSL_cleanup() now runs in a global destructor."

 On branch OpenSSL-4.0.0
 Changes to be committed:
	modified:   src/libopensc/cwa-dnie.c
	modified:   src/pkcs15init/pkcs15-oberthur-awp.c
	modified:   src/tests/p11test/p11test_case_common.c
	modified:   src/tools/netkey-tool.c

All the Github actions passed. But I have no way to test cwa-dnie.c, pkcs15-oberthur-awp.c, pkcs15-oberthur-awp.c, netkey-tool.c.

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -463,6 +463,7 @@ static CK_RV gostr3410_verify_data(const CK_BYTE_PTR pubkey, CK_ULONG pubkey_len
 	EC_POINT *P;
 	BIGNUM *X, *Y;
 	ASN1_OCTET_STRING *octet = NULL;
+	unsigned char *octet_rev = NULL;
 	char paramset[2] = "A";
 	int r = -1, ret_vrf = 0;
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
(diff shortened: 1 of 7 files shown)


frankmorgner (changes requested):
please import the suggested format changes with something like the following:

```
git apply <<'EOF'
EOF

...
(diff as shown in the CI run https://github.com/OpenSC/OpenSC/actions/runs/24478468760/job/71536794270?pr=3655)
...

EOF
```
The other changes look good.

src/pkcs11/openssl.c
```diff
@@ -516,13 +517,16 @@ static CK_RV gostr3410_verify_data(const CK_BYTE_PTR pubkey, CK_ULONG pubkey_len
 		if (group && pubkey_len <= LONG_MAX) {
 			const unsigned char *p = pubkey;
 			octet = d2i_ASN1_OCTET_STRING(NULL, &p, (long)pubkey_len);
+			if (octet)
+				octet_rev = malloc(ASN1_STRING_length(octet));
```
frankmorgner, line 521: buffer needs to be `free()`d

frankmorgner (changes requested):

src/pkcs11/openssl.c
dengert: octet_rev is freed at line 529 along with octet.

frankmorgner (approved):

Jakuje (approved):

PR #3661, opened 2026-04-19
card-piv.c, fuzz_pkcs11.c fix -Wshorten-64-to-32 errors

dengert:
`card-piv.c` has a "Wshorten-64-to-32" error when built using clang and with --enable-piv_sm. 

An actual error in `piv_inc` would only occur if either of the 16 byte  counters `enc_counter` or  `resp_enc_counter`  exceeded 2^128 secure messages in a single session when the counter would overflow.  

A few other places in card-piv SM code where a 64 bit length is converted to a 32 bit integer are type casted to 'int' as none of the numbers are over 32 bits.    

`src/tests/fuzzing/fuzz_pkcs11.c`  had a few "Wshorten-64-to-32" errors.

 On branch Wshorten-64-to-32
 Changes to be committed:
	modified:   src/tests/fuzzing/fuzz_pkcs11.c

Idemia PIV cards that support NIST Secure Messaging were used in test. 

##### Checklist

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -871,10 +871,10 @@ static void piv_inc(u8 *counter, size_t size)
 	unsigned int c = 1;
 	unsigned int b;
 	int i;
-	for (i = size - 1; c != 0 && i >= 0; i--){
-			b = c + counter[i];
-			counter[i] = b & 0xff;
-			c = b>>8;
+	for (i = (int)size - 1; c != 0 && i >= 0; i--) {
+		b = c + counter[i];
+		counter[i] = b & 0xff;
+		c = b >> 8;
 	}
 }
 
@@ -995,8 +995,8 @@ static int piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 		T87len = 0;
 		padlen = 0;
 	} else {
-		enc_datalen = ((plain->datalen + 15) / 16) * 16; /* may add extra 16 bytes */
-		padlen = enc_datalen - plain->datalen;
+		enc_datalen = (int)(((plain->datalen + 15) / 16) * 16); /* may add extra 16 bytes */
+		padlen = enc_datalen - (int)plain->datalen;
 		r = T87len = sc_asn1_put_tag(0x87, NULL, 1 + enc_datalen, NULL, 0, NULL);
 		if (r < 0)
 			goto err;
@@ -1022,13 +1022,13 @@ static int piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 		*p++ = 0x01; /* padding context indicator */
 
 		/* first round encryptes Enc counter with zero IV, and does not save the output */
-		if (EVP_CIPHER_CTX_reset(ed_ctx) != 1
-				|| EVP_EncryptInit_ex(ed_ctx, (*cs->cipher_cbc)(), NULL, priv->sm_session.SKenc, IV) != 1
-				|| EVP_CIPHER_CTX_set_padding(ed_ctx,0) != 1
-				|| EVP_EncryptUpdate(ed_ctx, p ,&outl, plain->data, plain->datalen) != 1
-				|| EVP_EncryptUpdate(ed_ctx, p + outl, &outll, pad, padlen) != 1
-				|| EVP_EncryptFinal_ex(ed_ctx, discard, &outdl) != 1
-				|| outdl != 0) {  /* should not happen */
+		if (EVP_CIPHER_CTX_reset(ed_ctx) != 1 ||
+				EVP_EncryptInit_ex(ed_ctx, (*cs->cipher_cbc)(), NULL, priv->sm_session.SKenc, IV) != 1 ||
+				EVP_CIPHER_CTX_set_padding(ed_ctx, 0) != 1 ||
+				EVP_EncryptUpdate(ed_ctx, p, &outl, plain->data, (int)plain->datalen) != 1 ||
+				EVP_EncryptUpdate(ed_ctx, p + outl, &outll, pad, padlen) != 1 ||
+				EVP_EncryptFinal_ex(ed_ctx, discard, &outdl) != 1 ||
+				outdl != 0) { /* should not happen */
 			sc_log_openssl(card->ctx);
 			sc_log(card->ctx,"SM _encode failed in OpenSSL");
 			r = SC_ERROR_INTERNAL;
@@ -1042,7 +1042,7 @@ static int piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 		*p++ = 0x01;
 		*p++ = plain->le;
 	}
-	macdatalen = p - sbuf;
+	macdatalen = (int)(p - sbuf);
 
 	memcpy(priv->sm_session.C_MCV_last, priv->sm_session.C_MCV, MCVlen); /* save is case fails */
 
@@ -1272,7 +1272,7 @@ static int piv_decode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 #endif
 
 	/*  MCV is first, then BER TLV Encoded Encrypted PIV Data and Status */
-	macdatalen = status.value + status.len - sm_apdu->resp;
+	macdatalen = (int)(status.value + status.len - sm_apdu->resp);
 
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
 	if (CMAC_Init(cmac_ctx, priv->sm_session.SKrmac, priv->sm_session.aes_size, (*cs->cipher_cbc)(), NULL) != 1
@@ -1324,7 +1324,7 @@ static int piv_decode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 		plain->resplen = 0;
 	} else {
 		p = ee.value;
-		inlen = ee.len;
+		inlen = (int)ee.len;
 		if (inlen < 17 || *p != 0x01) { /*padding and padding indicator are required */
 			sc_log(card->ctx, "SM padding indicator not 0x01");
 			r = SC_ERROR_SM_AUTHENTICATION_FAILED;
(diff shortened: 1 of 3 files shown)


src/libopensc/card-piv.c
```diff
@@ -995,8 +995,8 @@ static int piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
 		T87len = 0;
 		padlen = 0;
 	} else {
-		enc_datalen = ((plain->datalen + 15) / 16) * 16; /* may add extra 16 bytes */
-		padlen = enc_datalen - plain->datalen;
+		enc_datalen = (int)(((plain->datalen + 15) / 16) * 16); /* may add extra 16 bytes */
+		padlen = enc_datalen - (int)plain->datalen;
 		r = T87len = sc_asn1_put_tag(0x87, NULL, 1 + enc_datalen, NULL, 0, NULL);
```
Jakuje, line 1000: The `sc_asn1_put_tag()` takes third parameter size_t so would not it be better to change the `enc_datalen` to `size_t` rahter than casting it to smaller type?

The padlen though is needed to stay on the int type as it goes to openssl API.

src/tests/fuzzing/fuzz_pkcs11.c
```diff
@@ -81,7 +82,7 @@ static int fuzz_card_connect(const uint8_t *data, size_t size, sc_pkcs11_slot_t
 
 	/* Locate a slot related to the reader */
 	for (size_t i = 0; i < list_size(&virtual_slots); i++) {
-		slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, i);
+		slot = (sc_pkcs11_slot_t *) list_get_at(&virtual_slots, (int)i);
```
Jakuje: should we change the `i` to `unsigned int`? `list_size()` returns `unsigned int` so it would be cleaner to use that type instead of casting everywhere.

Jakuje:
And last thing. We have the strict CI build, which sets various cflags including this one, but likely does not build the opensc with the sm support

https://github.com/OpenSC/OpenSC/blob/master/.github/workflows/linux-strict.yml

I think changing the following line:

https://github.com/OpenSC/OpenSC/blob/2b204ecf5938861be98f34d197fd8408cde25671/.github/workflows/linux-strict.yml#L76

to
```
      - run: .github/build.sh piv-sm dist
```
should allow us to catch these issues (and others) in CI

src/libopensc/card-piv.c
dengert: The data lengths are basically limited to  #define MAX_FILE_SIZE 65535  or 256 from `opensc.h` 

Looking at other code that has addressed the Wshorten-64-to-32 problem, appear to cast to `(int)`.
`grep` and `wc` show 729 uses of `(int)` and 260  uses of "(unsigned int)" in OpenSC. 

I started looking at this when I realized that the default for the PIV SM code was disable. All the CI jobs that used "clang" where failing. So I can now build on my computer with gcc or clang.  The CFLAGS are from  `.github/workflows/linux-strict.yml`  

What this showed was  with `--enable-piv-sm` card-piv.c and the fuzz test needed some recasting. 

99b6c85107  line 49  added the `-Wshorten-64-to-32`

I added a few  other options -g and -O1 for gdb.

We could also consider  for ever shorten type case, test if some upper bits would be are lost.

src/tests/fuzzing/fuzz_pkcs11.c
dengert: Since line 84 has `size_t i = 0` it will always be >= 0 

I only addressed this fuzz program, because it did not compile. I really don't want to get involved with the fuzzing.

dengert:
> We have the strict CI build, which sets various cflags including this one, but likely does not build the opensc with the sm support

build.sh line 82 has:
        if [ "$1" == "piv-sm" ]; then
                 CONFIGURE_FLAGS="$CONFIGURE_FLAGS --enable-piv-sm"

and the clang strict action where failing on the lines I changes. So some test were compiling the code, but we do not have any virtual PIV applet that supports PIV SM. https://github.com/makinako/OpenFIPS201 has one, 
"This project has been commissioned and funded by the Australian Department of Defence," But I am no good at setting up a virtual test.

src/libopensc/card-piv.c
Jakuje: > Looking at other code that has addressed the Wshorten-64-to-32 problem, appear to cast to (int).

Indeed, this is the simplest solution/workaround to this warning. But each of the explicit casts needs to be checked to make sure it really is in the bounds.

dengert:
@metsma would you like to comment on how the piv-sm code is tested? 

@Jakuje 
Looking at: 

https://github.com/OpenSC/OpenSC/actions/runs/24636185611/job/72031942477?pr=3661

 Jobs  "*-piv-sm"  Run `.github/build.sh piv-sm valgrind`

and configure is called:
`./configure --disable-dependency-tracking --enable-doc CLANGTIDY=/bin/no-clang-tidy --enable-piv-sm --disable-notify --enable-valgrind --disable-integration-tests`

and print: "PIV SM support:          yes"

But as I said, we don't actually test the SM code. We need PIV applet that supports PIV SM.  See https://github.com/OpenSC/OpenSC/pull/3661#issuecomment-4276983322


A real PIV card will allow crypto and pin commands over contactless **only if PIV SM is used**. This NIST requirement goes back version 1.  Note: Yubikey cards does not enforce this NIST requirement.

A real PIV card will also allow SM to be used over the contact interface. If env `PIV_USE_SM=always` is defined card-piv.c will force the use of SM for everything for both contact and contactless except for some APDUs used to check card state, login state and initial setup of the card. Any failure for other APDUs to use PIV SM is not allowed by the driver honoring the users request to always use PIV_USE_SM. 

Without `PIV_USE_SM=always` one would need to look at debug log of outgoing and incoming ADPUs or other debug lines or pcsc log.

Jakuje:
I was not talking about testing PIV with SM. It would be different story to do. I was talking just about building the PIV SM enabled in the strict build so we can catch these build issues early. The suggestion in https://github.com/OpenSC/OpenSC/pull/3661#issuecomment-4276558016 should do exactly that unless I am missing something.

dengert:
OK, trying https://github.com/OpenSC/OpenSC/pull/3661#issuecomment-4276558016


But I am suggesting to have a virtual PIV card test with applet with PIV SM. 
And I am not very good with setting that up.

dengert:
@Jakuje  all 65 test passed. 
For example https://github.com/OpenSC/OpenSC/actions/runs/24676311708/job/72161145923
configures with --enable-piv-sm

src/libopensc/card-piv.c
```diff
@@ -1022,13 +1022,7 @@ static int piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu
...
 
 		/* first round encryptes Enc counter with zero IV, and does not save the output */
-		if (EVP_CIPHER_CTX_reset(ed_ctx) != 1
-				|| EVP_EncryptInit_ex(ed_ctx, (*cs->cipher_cbc)(), NULL, priv->sm_session.SKenc, IV) != 1
-				|| EVP_CIPHER_CTX_set_padding(ed_ctx,0) != 1
-				|| EVP_EncryptUpdate(ed_ctx, p ,&outl, plain->data, plain->datalen) != 1
-				|| EVP_EncryptUpdate(ed_ctx, p + outl, &outll, pad, padlen) != 1
-				|| EVP_EncryptFinal_ex(ed_ctx, discard, &outdl) != 1
-				|| outdl != 0) {  /* should not happen */
+		if (EVP_CIPHER_CTX_reset(ed_ctx) != 1 || EVP_EncryptInit_ex(ed_ctx, (*cs->cipher_cbc)(), NULL, priv->sm_session.SKenc, IV) != 1 || EVP_CIPHER_CTX_set_padding(ed_ctx, 0) != 1 || EVP_EncryptUpdate(ed_ctx, p, &outl, plain->data, (int)plain->datalen) != 1 || EVP_EncryptUpdate(ed_ctx, p + outl, &outll, pad, padlen) != 1 || EVP_EncryptFinal_ex(ed_ctx, discard, &outdl) != 1 || outdl != 0) { /* should not happen */
```
Jakuje: this super-long line is not readable. Please, keep it broken to the separate commands for readability. clang-format will not complain if the formatting will be otherwise wrong, which I think is the following format, if you want to reformat it:

```suggestion
		if (EVP_CIPHER_CTX_reset(ed_ctx) != 1 ||
				EVP_EncryptInit_ex(ed_ctx, (*cs->cipher_cbc)(), NULL, priv->sm_session.SKenc, IV) != 1 ||
				EVP_CIPHER_CTX_set_padding(ed_ctx, 0) != 1 ||
				EVP_EncryptUpdate(ed_ctx, p, &outl, plain->data, (int)plain->datalen) != 1 ||
				EVP_EncryptUpdate(ed_ctx, p + outl, &outll, pad, padlen) != 1 ||
				EVP_EncryptFinal_ex(ed_ctx, discard, &outdl) != 1 ||
				outdl != 0) { /* should not happen */
```

Jakuje (approved):
agreed. I think testing with SM-enabled PIV applet should be clearly outside of the scope of this PR.

src/libopensc/card-piv.c
dengert: OpenSC has 73 files using `||` at end of line when splitting an "if" statement and 67 files that put the `||` at the beginning of the line. Of these files 29 file are doing both. 

It is interesting that the CI suggested putting the "if" statement on one big line!

and this the only failing action:

https://github.com/OpenSC/OpenSC/actions/runs/24697539588/job/72233482236?pr=3661

Jakuje: Yes, there are still some cases written before we introduced formatting (and we did not refermat the whole code base).
It does that because it does not have any line length limit. If it would have, it would reflow all the multiline conditions to non-readable blocks. This is a known trade-off we discussed when we introduced the clang. Its not ideal, but its best we could get.

dengert:
Forced push with 3 commits that address 3 different files. 

If `clang` with `-Wshorten-64-to-32`  finds a typecast error in a multiline if statement will combine the lines if it does not meet code style. It will keep the lines split if the `||` is at the end of a line and not at the beginning of the next.

> OpenSC has 73 files using `||` at end of line when splitting an "if" statement and 67 files that put the || at the beginning of the line. Of these files 29 file are doing both.

So there are 67 files that may run into this situation.

Jakuje:
> So there are 67 files that may run into this situation.

Yes, but we did not want to do the automatic reformat due to automatic reformat resulting in odd results as you saw here. But I think we are slowly replacing them and getting them in the shape. It would be good to do that  more automatically (and I think I had some huge semi-manual rewrite of many places, but was years ago).

Jakuje (approved):
final approval. If there will not be more comments or you do not plan to add more changes, feel free to merge (or I will do later).

PR #3669, opened 2026-04-23
Fix outstanding coverity reports

Jakuje:
Recent changes introduced new issues noticed by coverity so this is attempting to fix these.

##### Checklist

- [~] macOS token is tested

### DIFF
--- a/src/libopensc/card-openpgp.c
+++ b/src/libopensc/card-openpgp.c
@@ -3570,8 +3570,6 @@ set_taglength_tlv(u8 *buffer, unsigned int tag, size_t length)
 {
 	u8 *p = buffer;
 
-	if (tag > 0xffff)
-		return SC_ERROR_INTERNAL;
 	if (tag > 0xff)
 		*p++ = (tag >> 8) & 0xFF;
 	*p++ = tag;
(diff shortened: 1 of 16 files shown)


Jakuje:
@frankmorgner do we still need to remove the appveyor app or de-register the opensc on the appveyor side? Not it still triggers pipelines, which just fail both in PR sand branches

frankmorgner (approved):

Jakuje:
Updated with bunch of other issues that we overlooked while rewriting asserts.

src/libopensc/card-openpgp.c
```diff
@@ -3726,6 +3724,9 @@ pgp_build_extended_header_list(sc_card_t *card, sc_cardctl_openpgp_key_gen_store
 
 	for (i = 0; i < comp_to_add; i++) {
 		sc_log(ctx, "Set Tag+Length for %s (%X).", componentnames[i], componenttags[i]);
+		if (componenttags[i] > 0xffff) {
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_DATA);
+		}
 		len = set_taglength_tlv(p, componenttags[i], componentlens[i]);
```
frankmorgner, line 3730: If we do it like this, then you may want to change the check on the tag in set_taglength_tlv so that it doesn't return an error code as the function returns size_t

frankmorgner (changes requested):

src/libopensc/card-openpgp.c
Jakuje: The return type of `set_taglength_tlv()` is unsinged so returning error wraps it to positive value. Given that there is just one caller of this function, I find it easier to move the check before the function and failing early than doing casting afterward or changing type of return type to accommodate negative error codes. Is guaranteed that the negative value (`-1400`) cast to `size_t` and back to `int` will be `-1400`?

frankmorgner: good choice, I overlooked that you modified `set_taglength_tlv()` as well

frankmorgner (approved):

PR #3670, opened 2026-04-27
common: Add a way to exclude the RTLD_DEEPBIND from dlopen flags

Jakuje:
The address sanitizer does not work with DEEPBIND so we provide (undocumented?) way to exclude this flag from the OpenSC operations based on environment variable.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/doc/files/files.html
+++ b/doc/files/files.html
@@ -1218,6 +1218,29 @@ app <em class="replaceable"><code>application</code></em> {
 					<code class="option">card_drivers =  <em class="replaceable"><code>name</code></em>... ;</code>
 				</a>
 				</p></dd><dt><span class="term">
+					<code class="envar">OPENSC_DEEPBIND</code>
+				</span></dt><dd><p>
+						The OpenSC and its tools are dynamically loading various
+						libraries (pcsclite, dynamic drivers and other pkcs11
+						modules) using
+						<span class="citerefentry"><span class="refentrytitle">dlopen</span>(3)</span>.
+					</p><p>
+						When this environment variable is set to
+						<code class="literal">1</code>
+						for the process running OpenSC, the calls will include
+						<code class="literal">RTLD_DEEPBIND</code> flag, which should
+						cause better symbol isolation between OpenSC and
+						loaded libraries. On the other hand, this is incompatible
+						with various static analysis tools as well as some
+						browser builds.
+					</p><p>
+						This option may be useful, if OpenSC needs to load a library
+						that pollutes the processes' global namespace and thereby
+						causes the unintended use of a different function with the
+						same name. In particular, this may happen when
+						<code class="literal">pkcs11-spy</code>
+						is used for logging between application and PKCS#11 library.
+				</p></dd><dt><span class="term">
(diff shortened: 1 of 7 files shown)


Jakuje:
We found a way around this without the workaround so I do not think we need that anymore. I will keep it open as draft to keep remind ourselves or other users that this is an issue and might have similar problems in other projects.

frankmorgner:
Thanks for the update. It may be of interest to know what your alternative solution has been...

Jakuje:
> Thanks for the update. It may be of interest to know what your alternative solution has been...

Indeed! The workaround was not to use address sanitizer with the pkcs11-tool as it was really not needed for the pkcs11-provider use:

https://github.com/openssl-projects/pkcs11-provider/pull/729

The openssl config was set to load a pkcs11 provider, which happened also within the pkcs11-tool openssl context. This was not necessary (as the pcks11-provider in pkcs11-tool is not used for anything)

For applications that would like to test their pkcs11 modules under pkcs11-tool (or pkcs11-spy), this workaround will not be usable and the only solution is likely to use this approach of removing the DEEPBIND flag

frankmorgner:
Could you please add documentation for the new environment variable?

Do we need to enable the environment variable in pkcs11-spy by default? With the problems so far it seems to be good to have it optionally only...

Jakuje:
Before writing a documentation, I would rather first agree where do we want to end up. This was mostly initial PoC and something that can be used to verify the root cause of the issues is really the deepbind. We also use the dlopen at three different places -- pkcs11-tool loading the pkcs11 module, the pkcs11-spy loading the pkcs11 module and opensc loading libpcsc.

The last use case dropped the deepbind already before the release (776e4cd2874108acc1b563be573dd96aaab58cdb). The original issue Ludovic saw in #2875 was explicitly with the pkcs11-spy.

My preferred approach would be to not use DEEPBIND by default and provide a OPENSC_DEEPBIND=1 environment variable that would force the use of RTLD_DEEPBIND in all of the three cases. Does it sounds reasonable?

frankmorgner:
Yes, that was my understanding as well. deepbind off by default, enabled via environment variable (and on for pkcs11-spy)

Jakuje:
I am looking that also the SM modules and dynamic drivers are loaded using this API.

I would suggest to disable it by default altogether for all of these uses (including the pkcs11-spy) to make the code and configuration less complicated.

frankmorgner:
Do you want to add the following hint?

OPENSC_DEEPBIND=1 may be useful, if OpenSC needs to load a library that pollutes the processes' global namespace and thereby causes the unintended use of a different function with the same name. In particular, this may happen when pkcs11-spy is used for logging between application and PKCS#11 library.

Jakuje:
Good idea. Added to both of the manual pages.

frankmorgner (approved):
thank you

Jakuje:
I did one more minor fixup in the manual page to avoid two trailing periods after the sentence.

frankmorgner (approved):

frankmorgner:
Thank you

PR #3672, opened 2026-04-29
More openssl 4.0 fixes and some cleanup 

simo5:
One file still fails on openssl 4.0 builds

also while inspecting previous changes I spotted a place in need of improvement so you get a bonus patch that improves correctness.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/pkcs11/openssl.c
+++ b/src/pkcs11/openssl.c
@@ -439,16 +439,23 @@ static void sc_pkcs11_openssl_md_release(sc_pkcs11_operation_t *op)
 
 #if !defined(OPENSSL_NO_EC)
 
-static void reverse(unsigned char *buf, size_t len)
+static unsigned char *
+reverse(const unsigned char *buf, size_t len)
 {
+	unsigned char *out = malloc(len);
 	unsigned char tmp;
 	size_t i;
 
+	if (out == NULL)
+		return NULL;
+
+	memcpy(out, buf, len);
 	for (i = 0; i < len / 2; ++i) {
-		tmp = buf[i];
-		buf[i] = buf[len - 1 - i];
-		buf[len - 1 - i] = tmp;
+		tmp = out[i];
+		out[i] = out[len - 1 - i];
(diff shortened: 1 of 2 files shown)


src/pkcs11/openssl.c
```diff
@@ -515,16 +521,11 @@ static CK_RV gostr3410_verify_data(const CK_BYTE_PTR pubkey, CK_ULONG pubkey_len
...
-				octet_rev = malloc(ASN1_STRING_length(octet));
 		}
-		if (group && octet && octet_rev) {
-			memcpy(octet_rev, ASN1_STRING_get0_data(octet), ASN1_STRING_length(octet));
-			reverse(octet_rev, ASN1_STRING_length(octet));
-			Y = BN_bin2bn(octet_rev, ASN1_STRING_length(octet) / 2, NULL);
-			X = BN_bin2bn(octet_rev + ASN1_STRING_length(octet) / 2, ASN1_STRING_length(octet) / 2, NULL);
+		if (group && octet) {
+			X = BN_lebin2bn(ASN1_STRING_get0_data(octet), ASN1_STRING_length(octet) / 2, NULL);
+			Y = BN_lebin2bn(ASN1_STRING_get0_data(octet) + ASN1_STRING_length(octet) / 2, ASN1_STRING_length(octet) / 2, NULL);
```
Jakuje, line 527: I see you removed the reverse of the octet here. Is it intentional or an overlook?

simo5: intentional, the reverse is superfluous given BN_lebin2bn (instead of BN_bin2bn) already does the right thing, notice the inversion of X and Y as the buffer is not reverse the two halves are taken in the correct order as well.

Jakuje: Thanks for clarification! Missed that `BN_lebin2bn`.

Jakuje (approved):

frankmorgner:
Thank you

PR #3673, opened 2026-05-01
Major PIV changes to identification and additional card provided info

### LINKED ISSUE #3545: Token2 PIN+ Release3.3 security keys
## General information

- [windows minidriver and management app](https://www.token2.com/site/page/piv-minidriver-and-management-tool-windows)
- [store page for my specific key](https://www.token2.eu/shop/product/token2-t2f2-pin-release3.3-piv-fido2-u2f-and-totp-security-key-with-pin-complexity-feature)
  - Functions: FIDO2.1 WebAuthN+CTAP, TOTP/HOTP, OpenPGP, PIV
  - FIDO available cross-platform
  - TOTP available on Android/iOS/Windows
  - I don't know how to test OpenPGP and don't plan to use it
  - PIV only available for Windows (official drivers for Linux are supposedly coming eventually)
- [FIDO2 applet details](https://tools.token2.com/fido2/info/view.php?file=eabb46cce24180bfae9e96fa6d2975cf) (probably not helpful)
- [FIDO2 applet source code](https://github.com/token2/pin_plus_firmware) (probably not helpful either)

The key doesn't seem to be detected by OpenSC on Linux (Arch Linux, `opensc` package installed)

I used a Windows VM to run OpenSC and installed the manufacturer provided minidriver:
```
C:\Program Files\OpenSC Project\OpenSC\tools>opensc-tool.exe --atr
Using reader with a card: TOKEN2 FIDO2 Security Key(0024) 0
3b:8f:80:01:54:4b:00:50:49:56:04:02:38:38:38:38:38:38:38:60 
```

I'm personally mostly interested in the PIV applet for cert storage, if I can do anything more to help investigate I would love to.
I don't have much specific knowledge about PC/SC (yet) but I have a decent background in Rust/Python/Linux if that helps.

## Detailed information

Output of `lsusb -d 349e:0024 --verbose`:
<details><summary>Show more</summary>
<pre>
Bus 001 Device 007: ID 349e:0024 TOKEN2 FIDO2 Security Key(0024)
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x349e TOKEN2
  idProduct          0x0024 FIDO2 Security Key(0024)
  bcdDevice            1.00
  iManufacturer           1 TOKEN2
  iProduct                2 FIDO2 Security Key(0024)
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x007d
    bNumInterfaces          2
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 Report
          wDescriptorLength      34
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               2
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x04  EP 4 OUT
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               2
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           3
      bInterfaceClass        11 Chip/SmartCard
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
      ChipCard Interface Descriptor:
        bLength                54
        bDescriptorType        33
        bcdCCID              1.10
        nMaxSlotIndex           0
        bVoltageSupport         1  5.0V 
        dwProtocols             3  T=0 T=1
        dwDefaultClock       3580
        dwMaxiumumClock      3580
        bNumClockSupported      0
        dwDataRate           9600 bps
        dwMaxDataRate        9600 bps
        bNumDataRatesSupp.      0
        dwMaxIFSD            2344
        dwSyncProtocols  00000000 
        dwMechanical     00000000 
        dwFeatures       00040480
          Auto PPS made by CCID
          Auto IFSD exchange (T=1)
          Short and extended APDU level exchange
        dwMaxCCIDMsgLen      2344
        bClassGetResponse    echo
        bClassEnvelope       echo
        wlcdLayout           none
        bPINSupport             0 
        bMaxCCIDBusySlots       1
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x06  EP 6 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x85  EP 5 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x87  EP 7 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval              32
Device Status:     0x0000
  (Bus Powered)
</pre>
</details>

Output of `fido2-token -I /dev/hidraw1` (probably not helpful):
<details><summary>Show more</summary>
<pre>
proto: 0x02
major: 0x02
minor: 0x01
build: 0x02
caps: 0x05 (wink, cbor, msg)
version strings: FIDO_2_0, FIDO_2_1, FIDO_2_1_PRE
extension strings: credBlob, credProtect, hmac-secret, largeBlobKey, minPinLength
transport strings: usb, nfc
algorithms: es256 (public-key), eddsa (public-key)
aaguid: eabb46cce24180bfae9e96fa6d2975cf
options: rk, up, noplat, alwaysUv, credMgmt, authnrCfg, clientPin, largeBlobs, pinUvAuthToken, setMinPINLength, nomakeCredUvNotRqd, credentialMgmtPreview
certifications: FIDO 3
fwversion: 0x100
maxmsgsiz: 2048
maxcredcntlst: 8
maxcredlen: 96
maxcredblob: 32
maxlargeblob: 2048
maxrpids in minpinlen: 6
remaining rk(s): 300
minpinlen: 6
pin protocols: 2, 1
pin retries: 8
pin change required: false
uv retries: undefined
</pre>
</details>

dengert:
Although the branch name is NitroKey-PIV, It is also for other PIV cards such as Token2 and PivApplet. (MyEID is next.) 

Major changes include:

If response to Select AID includes tag "AC", NIST only required it to indicate card supports NIST Secure Messaging. The required and optional algorithms need not be present.  But if card provides  "AC" entries for asymmetric algorithms They will be used to build the list supported by the card. If card does not provide any algorithms but does support the Yubico version APDU, The version number can be used to include additional algorithms and address card issues. if card does not support Yubico version APDU it will be assumed to support any of the past and present algorithms. 

PIV cards have been identified by the response to Select AID, the full ATR, or the historical data in an ATR. Most vendors add historical data unique to that vendor. This has been used in the past and continues to be used. 

But it was found that at least one PIVKey card  used the ATR of "3b:80:80:01:01" and  the PivApplet when used in the github action  test-piv.sh as a virtual card, used the same  ATR "3b:80:80:01:01" This caused problems  in trying to identify the card type. 
 Although PivApplet follows the Yubico version, it was identified as a PIVKey and the Yubico  version was never used. The PivApplet does set "AC" entries for RSA 2048, and EC 256 and EC 384.

An additional test was added to use the response to select AID tag "50" Application Label to identify the type of card. PivApplet use " PivApplet vn.n.n..." where as Token2 R3.3 used " PivApplet" and Yubikey does not add a tag "50".

When trying to match a PIV card, the card->type starts at -1 or if the user via opensc.conf provided a type defined in "cards.h". The 14001 (GENERIC) is reserved for use from opensc.conf. As features of the card are found, the card->type may change such as for CAC cards, or cards that support NIST  800-73-4 or  NIST  800-73-5 including NIST Secure Messaging.

Much of the matching/identification process is designed to work with unknown cards, as anyone can write a PIV driver. But they must respond to a Select AID with the require NIST fields.  

Fixes #3545 

Cards used in the testing include Nitrokey 3A NFC, Token2 R3.3, Yubikey 4 and 5 cards, PIVKey cards, IDEMIA ID-One demo cards, NIST demo cards from 2010 and PivApplet via test-piv.sh in github actions.  Also via contactless interface.
 

##### Checklist

- [ X] PKCS#11 module is tested


dengert:
The two failing tests appear to be unrelated to any of the changes in this PR

Configure on fedora-ix86 says: "checking whether the C compiler works... no"  

Tests of external pkcs11 modules / test-interoperability says: "Failed to wrap AES key with AES-CBC from SoftHSM"

.github/test-piv.sh
```diff
@@ -48,7 +48,8 @@ sleep 5
 
 # enforce the setting of different PIV type to support EC mechanisms
 # which are disabled for the generic ATR mapping to older Yubico devices
-export OPENSC_CONF="${PWD}/.github/opensc-piv.conf"
+# not needed
+#export OPENSC_CONF="${PWD}/.github/opensc-piv.conf"
```
Jakuje: If this is no longer needed, please remove the whole block and the file.

src/tests/p11test/piv_ref.json
```diff
@@ -112,98 +112,98 @@
 	],
 	[
 		"RSA_X_509",
-		"1024",
-		"3072",
+		"2048",
+		"2048",
```
Jakuje, line 116: This change says the PIV applet no longer supports the 3k RSA keys? Is it correct?

StefanSa:
@dengert 
Thanks for the PR and guidance.

I validated PR #3673 on Windows x64 with two Token2 R3.3 devices and can confirm the following:

* automatic PIV detection now works without forcing:

  * `driver = "PIV-II"`
  * `type = 14004`
* both tokens are detected simultaneously with distinct serial numbers:

  * `9c9e681f915c3fb9`
  * `e3745b56281fa3ae`
* no slot collisions observed in multi-token operation
* RSA4096 mechanisms are exposed correctly
* RSA4096 signing succeeds using `SHA256-RSA-PKCS`
* PIV objects are enumerated correctly:

  * PIV AUTH (9A)
  * SIGN (9C)
  * KEY MAN (9D)
  * CARD AUTH (9E)
* CHUID, Discovery Object and Key History Object are visible

The previous local workaround requiring `type = 14004` is no longer needed with this PR.

The Token2 devices now behave correctly through the normal PIV path on Windows using the PR build.

src/libopensc/card-piv.c
```diff
@@ -5682,12 +5699,13 @@ static int piv_match_card_continued(sc_card_t *card)
...
 
 		case SC_CARD_TYPE_PIV_II_YUBIKEY4:
+		case SC_CARD_TYPE_PIV_II_TOKEN2:
 			priv->card_issues |=  CI_OTHER_AID_LOSE_STATE
 				| CI_LEAKS_FILE_NOT_FOUND;
 			if (priv->yubico_version  < 0x00040302)
 				priv->card_issues |= CI_VERIFY_LC0_FAIL;
 			/* TODO may need to relocate when I get card to test */
-			if (priv->yubico_version >= 0x00050700)
+			if (priv->yubico_version >= 0x00050700) /* Also used by Token2 */
```
Jakuje: I would suggest to rename the yubico_version to something like applet_version given that its now used for three different applet versions already

Jakuje (commented):
Thank you for your work on this!

The last commit changing 1000+ lines of code has a commit message "code reformat + minor fixup". This makes it very unclear what of the 1000 lines of the code change is the minor fixup and makes it hard to review.

The rest of the series has a lot of adding and removing debug messages. I would prefer to squash the fixup changes, remove the adding of the debug messages and keep the reformat commit without the minor fixup to keep the history readable.

Otherwise the change set looks mostly good. Just a few nits inline.

StefanSa:
Additional Linux Mint validation results for PR #3673.

Environment:

* Linux Mint
* OpenSC 0.27.1 + PR3673 build
* Token2 FIDO2 Security Key (VID:PID 349e:0024)

Results:

* automatic Token2 PIV detection works
* no local OpenSC overrides required
* no `driver = "PIV-II"`
* no `type = 14004`

Validated successfully:

* `pkcs11-tool -L`
* `pkcs11-tool -M`
* `pkcs11-tool -O`
* Firefox NSS/OpenSC PKCS#11 integration
* PIV certificate visibility in Firefox
* RSA4096 mechanisms visible
* EDDSA/XEDDSA mechanisms visible

Important Linux-specific finding:
Initial failure on Linux Mint was NOT caused by OpenSC or PR3673.

Root cause was missing `libccid` ProductID support for:

* VID `0x349e`
* PID `0x0024`

Symptoms before libccid adjustment:

* USB device visible via `lsusb`
* CCID interface correctly exposed
* but no PC/SC reader registration
* `pcsc_scan` showed nothing
* OpenSC only saw the internal Broadcom reader
* Token appeared as "card not present"

After adding ProductID `0x0024` to libccid `Info.plist`:

* pcscd recognized the token correctly
* OpenSC enumeration worked
* Firefox PKCS#11 integration worked
* all OpenSC validation tests passed

Conclusion:
PR3673 itself validates successfully on Linux Mint once the lower-level CCID device mapping issue is resolved.

src/tests/p11test/piv_ref.json
dengert: It never did.  

https://github.com/OpenSC/OpenSC/pull/3673/changes/fcac2c97532e6df3487343b19a7d97b4d73cb8f3
```
With the changes to card-piv.c, the PivApplet used in
test-piv.sh is correctly identifying the PivApplet by using
the response to select AID. Previously the ATR did not have
Historical data so it was not possible to identify the applet.

test-piv.sh no longer needs a special opensc-piv.conf
which was forcing the ATR and the card type.  The same
ATR was also used by a PIVKey card listed in the known
ATRs causing the setting of the wrong algorithims.

The select AID response also include "AC" algorithm Identifies
which are used to define the 3 types algorithms supported
by the applet and NIST: RSA 2048, EC P256 and EC 384
```
One of the "PIVkey"  card entry was using the same generic ATR of "3b:80:80:01:01"

src/libopensc/card-piv.c
dengert: Its 4 applets and maybe more... 

What it comes down to, is some "PIV like" cards  are also "Yubico like" in using the Yubico version number APDU as well as the other Yubico features like "0xE0" and "0xE1" for ED25519 and X25519 or Yubico APDUs like: 
https://docs.yubico.com/yesdk/users-manual/application-piv/apdu/serial.html 
```
5672         /*
5673          * Assumes all Yubikey/Nitrokey/Token2/PivApplet cards are all "Yubico like"
5674          * via ATR, ATR Historic bytes or Application Label
5675          * Will also check if BASE cards are Yubikey like card and return a version number
5676          * SC_CARD_TYPE_PIV_II_GENERIC can be set by user to not test for Yubico version
5677          */
```
Note: unknown PIV cards i.e. SC_CARD_TYPE_PIV_II_BASE as they responded to select AID or user forced driver and type in "opensc.conf".  Will check if card is "Yubico like".

If this causes a problem with some card we have not seen the immediate circumvention is for user to force driver and type via  `opensc.conf` to SC_CARD_TYPE_PIV_II_GENERIC until we can get an ATR and response to Select AID to add  to the known type of card.

If this is going to far, I can remove testing for `yubico_version` for SC_CARD_TYPE_PIV_II_BASE.

dengert:
"code reformat + minor fixup".

Will look to see what the minor fixup was.

src/libopensc/card-piv.c
dengert: "PivApplet" is one of the "Yubico like" applets following Yubico version `00050400` 

I would like to leave it as `yubico_version` as its for any "Yubico like" card.

As pointed out over the years and in #3545 anyone can write a PIV applet, adding their own features, not implementing NIST require features, not making their applet identifiable and OpenSC does not find out until some user can't get some new card to work with OpenSC.

dengert:
@StefanSa Thanks for the  Linux Mint info. The problem is not always OpenSC.

dengert:
@Jakuje In regards to the reformat code and "minor fixes, Using:
`git diff -w --ignore-blank-lines 2277c6d90 c051aad > /tmp/ignore.diff`
This help not showing likes with simple spacing changes. 

I moved or added 6 `// clang-format on/off` lines
        `grep -- "[-+]//[ ]*clang-" ignore.diff | wc`

I updated 2 comments.

146 multiline if statements with "||", "&&", "|" or "&" at start of line
        `grep -- "^-[[:space:]]*[|&]" ignore.diff | wc`
clang-format placed reformatted to a single long line some with white space changes.  

I then split most of the long one lines back to multiple lines with breaks after "||", "&&", "|" or "&" or "," in functions
So I really only split lines at the right spot and not change anything else in the line.

clang-format also did:

19 blank lines removed

46 "static xxxxx functions so function name was on its own line.
        `grep "^+static" ignore.diff | wc` 

12 lines  which started with else
        `grep -- "^-[[:space:]]*else" ignore.diff | wc`

11  `switch` `case:` or `default:` with commands on one line were broken up in to multiple lines.

Some number of indenting `switch` statements so `switch` and `case:` at the same offset, which are not shown because of
`git diff -w`. 

Over all 323 lines added, 310 lines removed.

And I do not recall adding any new code just comments and  `// clang-format` changes.

.github/test-piv.sh
dengert: `opensc-piv.conf removed` https://github.com/OpenSC/OpenSC/pull/3673/commits/9f6be8ebbf367d3e953491dad246661495d8f114

src/tests/p11test/piv_ref.json
dengert: For the record: 
From debug info (now removed from `.github/test-piv.sh`) 
PivApplet responds to Select includes "Application Label" and "AC" entries
showing all key algo types. This PR uses the symmetric entries to define supported mechanisms  
Also shows `yubico_version`: `05 04 00` 
```
opensc-tool -a -s '00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00' -s '00 FD 00 00 03' 
Using reader with a card: Virtual PCD 00 00
3b:80:80:01:01
Sending: 00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00
61 81 79 4F 0B A0 00 00 03 08 00 00 10 00 01 00 a.yO............
79 07 4F 05 A0 00 00 03 08 50 18 50 69 76 41 70 y.O......P.PivAp
70 6C 65 74 20 76 30 2E 39 2E 30 2F 52 45 65 41 plet v0.9.0/REeA
78 61 44 5F 50 1E 67 69 74 68 75 62 2E 63 6F 6D xaD_P.github.com
2F 61 72 65 6B 69 6E 61 74 68 2F 50 69 76 41 70 /arekinath/PivAp
70 6C 65 74 AC 06 80 01 03 06 01 00 AC 06 80 01 plet............
0C 06 01 00 AC 06 80 01 07 06 01 00 AC 06 80 01 ................
11 06 01 00 AC 06 80 01 14 06 01 00             ............
Sending: 00 FD 00 00 03
05 04 00 
```

frankmorgner: Yubico's extension for 3072 RSA keys was adopted by other PIV vendors as well. Maybe it would be useful to support it in OpenSC since it seems to be defacto standard nowadays.

src/libopensc/card-piv.c
```diff
@@ -520,8 +525,12 @@ static const struct sc_atr_table piv_atrs[] = {
 	{ "3b:89:80:01:53:50:49:56:4b:45:59:37:30:44", NULL, NULL, SC_CARD_TYPE_PIV_II_PIVKEY, 0, NULL },
 	/* PIVKey SLE78 (57B) */
 	{ "3b:fd:96:00:00:81:31:fe:45:53:4c:4a:35:32:47:44:4c:31:32:38:43:52:57", NULL, NULL, SC_CARD_TYPE_PIV_II_PIVKEY, 0, NULL },
+	/* This ATR is too generic. It will be accepted if Application Label in AID response matches */
+	/* Which looks like there is no Application Label in AID response of these cards */
+	/* PivApplet (at least when run in virtual machine) uses the same ATR */
 	/* PIVKey uTrust (01) ISO 14443 Type B without historical bytes */
-	{ "3b:80:80:01:01", NULL, NULL, SC_CARD_TYPE_PIV_II_PIVKEY, 0, NULL },
+	/* { "3b:80:80:01:01", NULL, NULL, SC_CARD_TYPE_PIV_II_PIVKEY, 0, NULL }, */
```
frankmorgner, line 533: please remove the above lines

src/libopensc/card-piv.c
```diff
@@ -536,6 +545,13 @@ static const struct sc_atr_table piv_atrs[] = {
 	  "ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:ff:00:00", NULL, SC_CARD_TYPE_PIV_II_800_73_4, 0, NULL },
 	{ "3b:86:80:01:80:31:c1:52:41:12:76", NULL, NULL, SC_CARD_TYPE_PIV_II_800_73_4, 0, NULL }, /* contactless */
 
+	/*
+	 * If your card is not listed it maybe identified by ATR Historic bytes: 
+	 * Yubikey, NitroKey, Token2 and PivApplet
+	 * If still not found it will be treated as a NIST compliant card,
+	 * or card that follows the Yubikey version numbers.
+	 * Or you may try using opensc.conf to force a driver and card type 
+	 */
```
frankmorgner: please write it in a style that doesn't talk to a user. instead it should contain useful information for developers. something like this should do:

/* If the card cannot be identified by ATR only, then piv_match_card() will still check if selection of PIV AID works to continue. */

src/libopensc/card-piv.c
```diff
@@ -2876,7 +2876,7 @@ static int piv_find_aid(sc_card_t * card)
 			priv->init_flags |= PIV_INIT_AID_PARSED;
 			/* look for 800-73-4 0xAC for Cipher Suite Algorithm Identifier Table 14 */
 			/* There may be more than one 0xAC tag, loop to find all */
-
+			/* TODO do we need to look for "Nitrokey PIVP" in tag 0x50 length 12 */
```
frankmorgner: what is there todo here?

src/libopensc/card-piv.c
```diff
@@ -5900,7 +5907,6 @@ static int piv_match_card_continued(sc_card_t *card)
 		case SC_CARD_TYPE_PIV_II_PIVKEY:
 			priv->card_issues |= CI_VERIFY_LC0_FAIL
 				| CI_PIV_AID_LOSE_STATE /* be conservative */
-				| CI_NO_EC384 | CI_NO_EC
 				| CI_NO_RANDOM; /* does not have 9B key */
```
frankmorgner: Has this been removed accidentally?

src/libopensc/card-piv.c
```diff
@@ -2218,39 +2272,38 @@ static int piv_sm_open(struct sc_card *card)
...
 	eph_params[eph_params_n] = OSSL_PARAM_construct_end();
-	if (!(eph_ctx = EVP_PKEY_CTX_new_from_name(PIV_LIBCTX, "EC", NULL))  /* TODO should be FIPS */
-			|| !EVP_PKEY_keygen_init(eph_ctx)
-			|| !EVP_PKEY_CTX_set_params(eph_ctx, eph_params)
-			|| !EVP_PKEY_generate(eph_ctx, &eph_pkey)
-			|| !(Qehxlen = EVP_PKEY_get1_encoded_public_key(eph_pkey, &Qehx))
-			|| !Qehx
-			|| Qehxlen > cs->Qlen
-			) {
+	if (!(eph_ctx = EVP_PKEY_CTX_new_from_name(PIV_LIBCTX, "EC", NULL)) || /* TODO should be FIPS */
```
frankmorgner: what is here todo?

src/libopensc/card-piv.c
```diff
@@ -3561,20 +3644,23 @@ static int piv_write_binary(sc_card_t *card, unsigned int idx,
...
-		case 1:
-			r = piv_write_certificate(card, priv->w_buf, priv->w_buf_len, flags & 0x10);
-			break;
-		case 2: /* pubkey to be added to cache, it should have 0x53 and 0x99 tags. */
+	case 1:
+		r = piv_write_certificate(card, priv->w_buf, priv->w_buf_len, flags & 0x10);
+		break;
+
+	case 2: /* pubkey to be added to cache, it should have 0x53 and 0x99 tags. */
 		/* TODO: -DEE this is not fully implemented and not used */
```
frankmorgner: what is here todo? if it is not used (and tested) then it could be deleted

src/libopensc/card-piv.c
```diff
@@ -5495,23 +5600,21 @@ static int piv_match_card_continued(sc_card_t *card)
...
-			&& card->reader->atr.value[3] == 0x01) {
+	/* See if contactless  or run as virtual card */
+	/* PivApplet in .github/test_piv.sh for example */
+	if (card->reader->atr.len >= 4 && card->reader->atr.value[0] == 0x3b && (card->reader->atr.value[1] & 0xF0) == 0x80 && card->reader->atr.value[2] == 0x80 && card->reader->atr.value[3] == 0x01) {
 		priv->init_flags |= PIV_INIT_CONTACTLESS;
 	}
 
-	for (i=0; i < PIV_OBJ_LAST_ENUM -1; i++)
-		if(piv_objects[i].flags & PIV_OBJECT_NOT_PRESENT)
+	/* TODO is this needed? */
```
frankmorgner: please clarify or remove

src/libopensc/card-piv.c
```diff
@@ -5656,76 +5817,74 @@ static int piv_match_card_continued(sc_card_t *card)
...
-			card->type = SC_CARD_TYPE_PIV_II_GENERIC;
+	case SC_CARD_TYPE_PIV_II_GENERIC:
+		priv->card_issues |= CI_VERIFY_LC0_FAIL | CI_OTHER_AID_LOSE_STATE;
+		break;
+
+	case SC_CARD_TYPE_PIV_II_PIVKEY:
+		priv->card_issues |= CI_VERIFY_LC0_FAIL | CI_PIV_AID_LOSE_STATE | /* be conservative */
+				     CI_NO_RANDOM;				  /* does not have 9B key */
+		/* Discovery object returns 6A 82 so is not on card by default */
+		/* TODO may need more research */
```
frankmorgner: what is here todo?

src/libopensc/card-piv.c
```diff
@@ -6118,24 +6280,24 @@ piv_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data, int *tries_left)
...
-			&& card->type == SC_CARD_TYPE_PIV_II_800_73_4) {
-				sc_log(card->ctx, "Token does not support pin verify over contacless reader");
-				/* TODO maybe true for other contactless cards */
-				r = SC_ERROR_NOT_SUPPORTED;
+	if (priv->pin_cmd_verify_sw1 == 0x69 &&
+			priv->pin_cmd_verify_sw2 == 0x82 &&
+			priv->init_flags & PIV_INIT_CONTACTLESS &&
+			card->type == SC_CARD_TYPE_PIV_II_800_73_4) {
+		sc_log(card->ctx, "Token does not support pin verify over contacless reader");
+		/* TODO maybe true for other contactless cards */
```
frankmorgner: I think this comment can be removed

frankmorgner (changes requested):
The functional changes are good, only the presentation looks improvised.

please remove unnecessary comments (todos) and squash this into a few meaningful commits. If you don't have a tool for cherry-picking individual changes, then put all technical changes into a single commit with the last commit for formatting on-top.

src/tests/p11test/piv_ref.json
dengert: 602 /* NIST  cards have or do support these (non SM) algorithm identifiers */
603 #define AI_NIST (AI_RSA_1024 | AI_RSA_2048 | AI_RSA_3072 | AI_EC_256 | AI_EC_384)

RSA 3072 was optional in early versions of NIST 800-73  then dropped and in 800-73-5 is now optional.
So it is not  "defacto standard" it is the standard. 

If card response to SELECT AID response  has "AC" entries: (NitroKey, Token2, PivApplet all do, 
Use them else set NIST defaults:


5780         /* if card did not specify any of the NIST basic alg_ids add them */
5781         if ((priv->alg_ids & AI_NIST) == 0) {
5782                  priv->alg_ids |= AI_NIST;
5783         }

Current Yubico does not add  "AC" entries, so priv->alg_ids |= AI_NIST;
NIST cards have not required a card to specify any "AC" entries, but id 0x27 or 0x2E indicating SM support.
in 800-73-4. and 800-73-5 says OK to adding "AC" entries. 

Then based on the card type from "user forced card type", ATR or ATR historical bytes:
 

5823         case SC_CARD_TYPE_PIV_II_NEO:
5824                 priv->card_issues |= CI_NO_EC384 | CI_VERIFY_630X | CI_OTHER_AID_LOSE_STATE |
5825                                      CI_LEAKS_FILE_NOT_FOUND | CI_NFC_EXPOSE_TOO_MUCH;
5826                 if (priv->yubico_version < 0x00040302)
5827                         priv->card_issues |= CI_VERIFY_LC0_FAIL;
5828                 break;
5829 
5830         case SC_CARD_TYPE_PIV_II_YUBIKEY4:
5831         case SC_CARD_TYPE_PIV_II_PIVAPPLET:
5832         case SC_CARD_TYPE_PIV_II_TOKEN2:    
5833                 priv->card_issues |= CI_OTHER_AID_LOSE_STATE | CI_LEAKS_FILE_NOT_FOUND;
5834                 if (priv->yubico_version < 0x00040302)
5835                         priv->card_issues |= CI_VERIFY_LC0_FAIL;
5836 
5837                 if (priv->yubico_version >= 0x00050700) /* Also used by Token2 */
5838                         priv->alg_ids |= AI_RSA_4096 | AI_25519;
5839                 break;
5840 
5841         case SC_CARD_TYPE_PIV_II_NITROKEY:
5842                 priv->card_issues |= CI_OTHER_AID_LOSE_STATE;
5843                 if (priv->yubico_version >= 0x00010802) /* use for NitroKey too */
5844                         priv->alg_ids |= AI_RSA_4096;
5845                 break;

Note that Nitrokey version number do not follow Yubio version numbers but used the Yubico APDU to get its version. 
Also early NEO did not have EC, so   priv->card_issues |= CI_NO_EC384 which is then applied 

5982         if (!(priv->card_issues & CI_NO_EC)) {

5991                                 if (!(priv->card_issues & CI_NO_EC384 && ec_curves[i].size == 384))

So I thing this answers your question, 3072 is include by default or by knowledge of  card properties.

dengert: ALso note the  PKCS11 mechanisms listed are what the card, card vendor documentation or standard says should be available and not based on what keys are actually on the card as some users have thought.

src/libopensc/card-piv.c
dengert: No, deliberately.  By now, any new PIVkey cards should have EC, and these lines with leading "|" where  reformatted into one line. OpenSC has no support on how to write anything to their cards, so being "conservative" in this case is to assume the card vendor supports the required EC curves. (I do not recall ever hearing any problems with using PIVKey cards.)

src/libopensc/card-piv.c
dengert: I will remove the comment. 
It is not clear if OpenSSL "FIPS"  LIBCTX is needed when creating ephemeral EC key.

src/libopensc/card-piv.c
dengert: When generating a key on the card is the only time the card returns the public key. 
piv_tool can store the public key to be used at a later time. But it could also stored in the PKCS15 session cache
for use in same session, or could be stored in a dummy certificate on the card. 

When driver starts it reads certificates to get SPKI i.e. the public key. if no certificate, but env says where piv-tool stored it it will read the file is the private key can be used t sign a CSR. 

I will remove the comment.

src/libopensc/card-piv.c
dengert: Will remove the comment.

src/libopensc/card-piv.c
dengert: Will remove the comment.

src/libopensc/card-piv.c
dengert: The code is saying if this is a real NIST card SC_CARD_TYPE_PIV_II_800_73_4, the card will not accept a pin verify over contactless without SM .  So OpenSC will not even try as this is the security risk NIST is trying to avoid, as it may expose the pin. 
But "Yubico like" card will or may accept the verify which violates the NIST standard. 

Again anyone can write a PIV applet and claim it supports PIV without getting NIST certification. 

I will remove the comment.

src/libopensc/card-piv.c
Jakuje: No, the PIV driver should use the `PIV_LIBCTX` (macro mapping to `card->ctx->ossl3ctx->libctx`) unconditionally. The function `sc_openssl3_init()` already handles loading the right provider into this context.

src/tests/p11test/piv_ref.json
frankmorgner: You are right, thank you for the clarification. RSA 4096 is not standardized by NIST, but supported by Yubico and others, right?

dengert: Token2 and Nitrokey say via `AC 06 80 01 16 06 01 00` in response to select AID  "16" i.e. RSA 4096 is supported.

"YubiKey 5 Series firmware 5.7+ introduces native support for Ed25519 (EdDSA) and X25519 keys within the PIV."
( TODO I did not have Yubico 5.7+ with XEDDSA, need to add. ) And  Supports 4096.

Nitrokey uses same APDU to get yubico_version but says it is version  010802 so is in separate case statement.
and does not  set in EDDSA or XEDDSA. If newer card include AC entries the code will set the algos.

src/libopensc/card-piv.c
dengert: The comment is still there. It show the problem of using an unidentifiable ATR to select a card. The code will identify it as being  contactless and will do a select AID. I Don't have that exact card  but other PIVKey card show an ATR like 
'3b:89:80:01:53:50:49:56:4b:45:59:37:30:44"  where historical data has "PIVKEY70".

src/libopensc/card-piv.c
dengert: Comments removed.

dengert:
I believe I have addressed all the comments except the "yubico_version" and have force pushed the PR with squashed and reordered commits. The   "yubico_version" issue is really that many card are using the Yubico APDU to get a version, but only some follow Yubico version numbes The Nitrokey does not and has it own `case statement`.

frankmorgner (approved):
Please merge via SQUASHING into a SINGLE commit, because some of the individual commits (e.g. 55822af) are broken with the errors being patched out.

Jakuje (approved):
please, keep at least the reformat commits separately. If this would be one commit, it would be completely impossible to review as there is 1000 lines reformatted and sometimes among them new functionality added.

dengert:
> please, keep at least the reformat commits separately.

Missed that comment. I can go back to previous head and rebase so the formatting is in separate commit on Monday.

dengert:
In response to:
https://github.com/OpenSC/OpenSC/pull/3673#pullrequestreview-4301740982
https://github.com/OpenSC/OpenSC/pull/3673#pullrequestreview-4304126886
https://github.com/OpenSC/OpenSC/pull/3673#issuecomment-4468723231


Broke the PR up into 5 commits by going back in git history, and only squashing the card-piv.c changes.

1. fix `.clang-format` so .json files are not mis formatted
2. cleanup test-piv.sh as this PR will handle the PivApplet correctly
3. card-piv.c and cards.h with the squashed changes 
4. manual update of long if statements  so line breaks are after "&&",  "||", "&" and "|" rather then before so clang-format  will not put the statement on a single line. (Only manual change since previously approved changes to PR.) 
5. The results of running clang-format-19 with no additional changes. 

I think it is ready to be committed.

Jakuje (approved):

frankmorgner (approved):

dengert:
@Jakuje @frankmorgner  I can go ahead and merge the PR if that is ok with you.

frankmorgner:
Thanks everyone for contributions and help!

PR #3683, opened 2026-05-11
oberthur: Avoid memory leak while writing public keys

Jakuje:
Thanks oss-fuzz

https://issues.oss-fuzz.com/issues/510130167

Unfortunately I was not able to make the oss-fuzz tooling working today for me so the change is untested.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/setup-fedora.sh
+++ b/.github/setup-fedora.sh
@@ -13,7 +13,7 @@ fi
 
 # 64bit or 32bit dependencies
 if [ "$1" == "ix86" ]; then
-	DEPS="$DEPS pcsc-lite-devel*.i686 readline-devel*.i686 openssl-devel*.i686 zlib-ng-devel*.i686 libcmocka-devel*.i686 glibc-devel*i686"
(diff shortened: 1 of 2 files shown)


frankmorgner:
What about always using using `sc_pkcs15_erase_pubkey(key)` in case of an error. If some call to `auth_update_component()` is successful, then remove the *comsumed* buffer from key, e.g. `key.modulus.data = NULL; key.modulus.length = 0;`. This way, you don't have to consider all inner components of the public key, but only those that this function is aware of.

Jakuje:
> What about always using using sc_pkcs15_erase_pubkey(key) in case of an error.

I missed this function. But it would mean we would have to wrap the `sc_pkcs15_pubkey_rsa` in incomplete `sc_pkcs15_pubkey` adding even more complexity.

I can change it but I am really not sure if it would help with the readability of this code.

frankmorgner:
I didn't realize that key is of type `sc_pkcs15_pubkey_rsa` rather than `sc_pkcs15_pubkey`. In that case, I would keep your changes as they are.

frankmorgner (approved):

Jakuje:
Thank you for fast review! Will merge to unbreak the 32b builds on Fedora. The interoperability test is still broken likely due to softhms crashing somewhere. I did not manage to investigate it further.

PR #3689, opened 2026-05-19
CI: Windows artifact contain release (candidate) tags or short hash o…

frankmorgner:
…therwise

fixes problem with artifact upload when ref_name contains invalid characters such as "/"

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -8,27 +8,14 @@ if [ -x "/bin/sudo" ]; then
 	SUDO="sudo"
 fi
 
-if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-	PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
-	if [ "$GITHUB_BASE_REF" == "master" ]; then
-		SUFFIX="-pr$PR_NUMBER"
-	else
-		SUFFIX="$GITHUB_BASE_REF-pr$PR_NUMBER"
-	fi
-else
-	TAG_OR_BRANCH=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
-	if [ "$GITHUB_REF_TYPE" == "tag" ]; then
-		if [[ "$TAG_OR_BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-			# Tag matches the version scheme without suffix -- no suffix needed
-			SUFFIX=""
-		elif [[ "$TAG_OR_BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
-			# rc suffix after version. Use the suffix part only
-			SUFFIX="${BASH_REMATCH[1]}"
-		else
-			SUFFIX="-$TAG_OR_BRANCH"
-		fi
-	elif [ "$TAG_OR_BRANCH" != "master" ]; then
-		SUFFIX="-$TAG_OR_BRANCH"
+SUFFIX="-${GITHUB_SHA:0:7}"
+if [ "$GITHUB_REF_TYPE" == "tag" ]; then
+	if [[ "$GITHUB_REF_NAME" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+		# Tag matches the version scheme without suffix -- no suffix needed
+		SUFFIX=""
+	elif [[ "$GITHUB_REF_NAME" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
+		# rc suffix after version. Use the suffix part only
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -41,23 +41,13 @@ jobs:
...
-              fi
-            elif [[ "${{ github.ref_name }}" != "master" ]]; then
-              SUFFIX="-${{ github.ref_name }}"
+          github_sha_hash="${{ github.sha }}"
+          SUFFIX="-${github_sha_hash:0:7}"
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            if [[ "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              SUFFIX="" # regular release tag
+            elif [[ "${{ github.ref_name }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
+              SUFFIX="${BASH_REMATCH[1]}" # rc release tag
```
Jakuje, line 49: I would prefer to keep this in sync with the similar code in the build script for other platforms to make sure we do not have different suffixes in different artifacts from single push:
https://github.com/OpenSC/OpenSC/blob/master/.github/build.sh

frankmorgner: done

.github/build.sh
```diff
@@ -8,26 +8,15 @@ if [ -x "/bin/sudo" ]; then
...
-if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-	PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
-	if [ "$GITHUB_BASE_REF" == "master" ]; then
-		SUFFIX="-pr$PR_NUMBER"
+SUFFIX="-${GITHUB_SHA:0:7}"
+if [ "$GITHUB_REF_TYPE" == "tag" ]; then
+	if [[ "$TAG_OR_BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+		# Tag matches the version scheme without suffix -- no suffix needed
+		SUFFIX=""
+	elif [[ "$TAG_OR_BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
```
Jakuje:
```suggestion
	if [[ "$GITHUB_REF_NAME" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
		# Tag matches the version scheme without suffix -- no suffix needed
		SUFFIX=""
	elif [[ "$GITHUB_REF_NAME" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
```
also on the line 20

.github/build.sh
```diff
@@ -8,26 +8,15 @@ if [ -x "/bin/sudo" ]; then
...
-			# Tag matches the version scheme without suffix -- no suffix needed
-			SUFFIX=""
-		elif [[ "$TAG_OR_BRANCH" =~ ^[0-9]+\.[0-9]+\.[0-9]+(.+)$ ]]; then
-			# rc suffix after version. Use the suffix part only
-			SUFFIX="${BASH_REMATCH[1]}"
-		else
-			SUFFIX="-$TAG_OR_BRANCH"
-		fi
-	elif [ "$TAG_OR_BRANCH" != "master" ]; then
 		SUFFIX="-$TAG_OR_BRANCH"
```
Jakuje: also here:
```suggestion
		SUFFIX="-$GITHUB_REF_NAME"
```

frankmorgner:
I was in a hurry while editing build.sh
Now, it should be identical to windows.yml

.github/build.sh
```diff
@@ -8,27 +8,14 @@ if [ -x "/bin/sudo" ]; then
 	SUDO="sudo"
 fi
 
-if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-	PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
-	if [ "$GITHUB_BASE_REF" == "master" ]; then
-		SUFFIX="-pr$PR_NUMBER"
```
Jakuje, line 14: I am still not sure if we do not want to keep this semantics too -- the `-prXXXX` suffix from the artifacts built from PRs sound to me like a good idea too. Without that, the PR artifacts will be just the same version as the previously release version? This does not sound to me much practical (if anyone tries to install something from PR artifacts).

If we do not consider this an issue, than ack as it makes the suffix logic for the releases more straighforward.

frankmorgner: The short hash is still PR specific, isn't it? I'd prefer to keep this simple

Jakuje: Right. I missed the logic of setting SUFFIX even without tag.

Jakuje (approved):

PR #3690, opened 2026-05-20
asn1: avoid uninitialized read of INTEGER/ENUMERATED debug output

### LINKED ISSUE #3685: asn1_decode_entry logs uninitialized ASN.1 INTEGER output after decode failure
### Describe the bug

`asn1_decode_entry()` logs an ASN.1 INTEGER/ENUMERATED output value even when `sc_asn1_decode_integer()` has already returned an error before writing that output. The bug is in the ASN.1 decoder/debug logging path, not in the fuzz harness.

The source path that triggers this uses uninitialized local integers as ASN.1 output storage:

```c
/* src/libopensc/asn1.c */
static int asn1_decode_path(sc_context_t *ctx, const u8 *in, size_t len,
			    sc_path_t *path, int depth)
{
	int idx, count, r;
	...
	sc_format_asn1_entry(asn1_path + 1, &idx, NULL, 0);
	sc_format_asn1_entry(asn1_path + 2, &count, NULL, 0);
	...
	r = asn1_decode(ctx, asn1_path, in, len, NULL, NULL, 0, depth + 1);
	if (r)
		return r;
	...
}
```

`sc_asn1_decode_integer()` can return before writing `*out`, for example when the input length is unsupported:

```c
/* src/libopensc/asn1.c */
int sc_asn1_decode_integer(const u8 * inbuf, size_t inlen, int *out, int strict)
{
	...
	if (inlen == 0) {
	}
	if (inlen > sizeof(int)) {
		return SC_ERROR_NOT_SUPPORTED;
	}
	...
}
```

The error path in `asn1_decode_entry()` still reads and formats `entry->parm`:

```c
/* src/libopensc/asn1.c */
case SC_ASN1_INTEGER:
case SC_ASN1_ENUMERATED:
	if (parm != NULL) {
		r = sc_asn1_decode_integer(obj, objlen, (int *) entry->parm, 0);
		sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n", depth, depth, "",
				entry->name, *((int *) entry->parm));
	}
	break;
```

This was originally found by DMSAN (Differential Memory Sanitizer) fuzzing and then reduced to the attached native reproducer.

### Steps to reproduce

Build OpenSC with the existing fuzz target enabled. One working native configuration is:

```sh
./bootstrap
./configure --disable-optimization --enable-static --disable-shared --disable-pcsc --enable-ctapi --enable-openpace --enable-fuzzing
make -j"$(nproc)"
```

Use the attached `poc_pkcs15init.bin` as input to the existing `fuzz_pkcs15init` target. The target source is `src/tests/fuzzing/fuzz_pkcs15init.c`; depending on the build layout, the binary may be under the fuzzing test build directory or copied to the fuzz output directory.

Run:

```sh
OPENSC_DEBUG=9 ./fuzz_pkcs15init poc_pkcs15init.bin 2>&1 | grep "decoding 'index' returned"
```

[opensc_asn1_debug_uum_poc.zip](https://github.com/user-attachments/files/27833808/opensc_asn1_debug_uum_poc.zip)

### Expected behavior

If `sc_asn1_decode_integer()` returns an error, the debug log should report the decode failure without reading the output storage passed through `entry->parm`.

The decoded INTEGER/ENUMERATED value should only be printed after a successful decode. A minimal fix would be to log the value only when `r == 0`, and otherwise log the error code:

```c
r = sc_asn1_decode_integer(obj, objlen, (int *) entry->parm, 0);
if (r == 0) {
	sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n",
			depth, depth, "", entry->name, *((int *) entry->parm));
} else {
	sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' failed: %d\n",
			depth, depth, "", entry->name, r);
}
```

### Actual behavior

With high OpenSC debug logging enabled, the native reproducer prints an INTEGER output value from a failed decode path:

```text
asn1.c:1529:asn1_decode_entry:   decoding 'index' returned -11168
```

The exact integer may vary between runs or builds because it comes from an uninitialized stack slot. With default debug logging disabled, we do not see this value formatted.

We did not observe a crash, memory leak, or out-of-bounds access from this input. The observed impact is a debug-log uninitialized read / limited stack value disclosure in the high-debug path.

### OpenSC version

reproduced on 3f3d484c4508d4265d0519fbe5a85b30c1469497; still present on master 814f745b3b6d100295f65f1935edd33d520d33ab

### Operating system

Linux x86_64, Ubuntu 22.04 userspace

no physical card; OpenSC fuzz_pkcs15init virtual fuzz reader input, mulating PKCS#15/MyEID-like data

OpenSC native fuzz target fuzz_pkcs15init

### Logs and debug output

```shell
asn1.c:1529:asn1_decode_entry:   decoding 'index' returned -11168

The exact integer can vary because it is read from an uninitialized stack slot. With default debug logging disabled, this value is not formatted in the log.
```

### Checklist

Carmel0:
Fixes #3685.

Guards the `SC_ASN1_INTEGER` / `SC_ASN1_ENUMERATED` debug log in `asn1_decode_entry()` on the return value of `sc_asn1_decode_integer()`: log the decoded value only on success, otherwise log the error code. The decoder itself is untouched and the success-path log format is unchanged. See the issue for the full root-cause analysis and reproducer.

##### Checklist

Reproduced and tested with the `fuzz_pkcs15init` native target on Linux x86_64; no physical smart card involved.

### DIFF
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -1525,12 +1525,9 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
 	case SC_ASN1_ENUMERATED:
 		if (parm != NULL) {
 			r = sc_asn1_decode_integer(obj, objlen, (int *) entry->parm, 0);
-			if (r == 0) {
+			if (r == SC_SUCCESS) {
 				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n",
 					depth, depth, "", entry->name, *((int *) entry->parm));
-			} else {
-				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' failed: %d\n",
-					depth, depth, "", entry->name, r);
 			}
 		}
 		break;
--- a/src/libopensc/asn1.c
+++ b/src/libopensc/asn1.c
@@ -1492,7 +1492,7 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
 	void *parm = entry->parm;
 	int (*callback_func)(sc_context_t *nctx, void *arg, const u8 *nobj,
 			     size_t nobjlen, int ndepth);
-	size_t *len = (size_t *) entry->arg;
+	size_t *len = (size_t *)entry->arg;
 	int r = 0;
 
 	callback_func = parm;
@@ -1524,10 +1524,10 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
 	case SC_ASN1_INTEGER:
 	case SC_ASN1_ENUMERATED:
 		if (parm != NULL) {
-			r = sc_asn1_decode_integer(obj, objlen, (int *) entry->parm, 0);
+			r = sc_asn1_decode_integer(obj, objlen, (int *)entry->parm, 0);
 			if (r == SC_SUCCESS) {
 				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n",
-					depth, depth, "", entry->name, *((int *) entry->parm));
+					depth, depth, "", entry->name, *((int *)entry->parm));
 			}
 		}
 		break;
@@ -1698,7 +1698,7 @@ static void sc_free_entry(struct sc_asn1_entry *asn1) {
 		switch (entry->type) {
 		case SC_ASN1_CHOICE:
 		case SC_ASN1_STRUCT:
-			sc_free_entry((struct sc_asn1_entry *) entry->parm);
+			sc_free_entry((struct sc_asn1_entry *)entry->parm);
 			break;
 		case SC_ASN1_OCTET_STRING:
 		case SC_ASN1_BIT_STRING_NI:
@@ -1758,7 +1758,7 @@ static int asn1_decode(sc_context_t *ctx, struct sc_asn1_entry *asn1,
 		/* Special case CHOICE has no tag */
 		if (entry->type == SC_ASN1_CHOICE) {
 			r = asn1_decode(ctx,
-				(struct sc_asn1_entry *) entry->parm,
+				(struct sc_asn1_entry *)entry->parm,
 				p, left, &p, &left, 1, depth + 1);
 			/* When the inner call fails it returns before writing
 			 * back to *newp and *len_left, so the caller's p/left are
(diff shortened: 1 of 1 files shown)


src/libopensc/asn1.c
```diff
@@ -1525,8 +1525,13 @@ static int asn1_decode_entry(sc_context_t *ctx,struct sc_asn1_entry *entry,
...
 			r = sc_asn1_decode_integer(obj, objlen, (int *) entry->parm, 0);
-			sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n", depth, depth, "",
-					entry->name, *((int *) entry->parm));
+			if (r == 0) {
+				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n",
+					depth, depth, "", entry->name, *((int *) entry->parm));
+			} else {
+				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' failed: %d\n",
+					depth, depth, "", entry->name, r);
+			}
```
frankmorgner:
```suggestion
			if (r == SC_SUCCESS) {
				sc_debug(ctx, SC_LOG_DEBUG_ASN1, "%*.*sdecoding '%s' returned %d\n",
					depth, depth, "", entry->name, *((int *) entry->parm));
			}
```

frankmorgner (changes requested):
Thanks for your contribution, please have a look at the suggestion. The error case doesn't need to be debugged, this will be done at the end of the function.

Jakuje:
Providing a reproducer input into the fuzzer curpus in

https://github.com/OpenSC/OpenSC/tree/master/src/tests/fuzzing/corpus/fuzz_pkcs15init

or unit test would be also helpful

Carmel0:
> Providing a reproducer input into the fuzzer curpus in
> 
> https://github.com/OpenSC/OpenSC/tree/master/src/tests/fuzzing/corpus/fuzz_pkcs15init
> 
> or unit test would be also helpful

For this PR, I would prefer to keep the change minimal and avoid adding a fuzzer corpus seed or a unit test around debug logging, unless you think it is needed for merging. The issue and PR already document the reproducer and the root cause, and the actual fix is limited to printing the INTEGER and ENUMERATED value only after the decode succeeds.

My concern with adding a corpus seed here is that it would not be a particularly good one. The interesting part of this issue is not a new ASN.1 decoding case. It is a failed decode path where the debug log reads a value that was not produced. A seed for that would be fairly opaque, and it would mostly preserve a very specific log related reproducer rather than a meaningful parser input.

A unit test is possible, but it would also be a bit artificial. It would have to enable ASN.1 debug logging, capture the log output, and check that no decoded integer value is printed after a failed decode. That would make the test depend on internal log messages and implementation details, instead of the behavior of the ASN.1 decoder itself. I am a little concerned that such a test could become brittle without adding much value for this minimal fix.

I am continuing to fuzz OpenSC and will keep sharing useful findings. When I find inputs that make good corpus additions, or cases that are better covered by unit tests, I will provide them in follow up issues or PRs.

Jakuje (approved):

frankmorgner:
thank you!

PR #3694, opened 2026-05-28
Fix epass2003 regression with SM

### LINKED ISSUE #3693: epass2003 token login with SM fails since May 6
Since merging the 5ebb04ded3e8a54a9544a3745b3b5bd566d1e0b2, the epass login fails in the CI. I did not have time to investigate earlier, but it looks like there is some mistake we missed:

https://gitlab.com/redhat-crypto/OpenSC/-/pipelines?ref=epass2003&scope=all

The driver initialization fails and card reports as unrecognized

Jakuje:
There is 0x01 separator between the tag and value, which was breaking the encryption for last couple of weeks.

Fixup of 5ebb04ded3e8a54a9544a3745b3b5bd566d1e0b2

Fixes: #3693

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-epass2003.c
+++ b/src/libopensc/card-epass2003.c
@@ -1462,6 +1462,8 @@ decrypt_response(struct sc_card *card, unsigned char *in, size_t inlen, unsigned
 	if (cipher_len < 2 || cipher_len > sizeof plaintext)
 		return -1;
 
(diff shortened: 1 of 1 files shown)


Jakuje:
Finally the time when having this token in CI pays off!

popovec (approved):

frankmorgner:
thank you

PR #3695, opened 2026-05-30
[0.26.1] backport CVE-2026-40510: bound PIV history URL length in piv_process_history

dkgkdfg65:
The 0.26.1 branch doesn't have the piv_process_history() bounds fix that landed on master (PR #3558). On 0.26.1, after building the cache filename, sc_get_cache_dir() is called with sizeof(filename)-strlen(fp)-2 which underflows when the on-card URL filename is long, then strcat(filename, fp) overruns filename[PATH_MAX]. There's no urllen>118 cap and no SHA-256 filename validation.

I checked it actually reaches the overflow on 0.26.1 rather than just diffing: extracted the function plus sc_get_cache_dir's snprintf-bounded home handling into a harness, built with -fsanitize=address on ubuntu:22.04. A PIV history object with an oversized offCardCertURL gives, pre-fix, an ASan stack-buffer-overflow WRITE in strcat; with the fix applied it returns -2 cleanly.

One note: the upstream fix 3f24f0b4 is a merge, so I cherry-picked with -m 1, which yields the intended 16-insertion/1-deletion change; authors (Frank Morgner / Doug Engert) preserved. Clean, no conflicts. Glad to rebase if you'd prefer.

upstream: https://github.com/OpenSC/OpenSC/commit/3f24f0b48a481a8cf2e46059d8238a283ddc1c13
CVE-2026-40510

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -5057,6 +5057,10 @@ piv_process_history(sc_card_t *card)
 
 			url = sc_asn1_find_tag(card->ctx, body, bodylen, 0xF3, &urllen);
 			if (url) {
+				if (urllen > 118) {
+					r = SC_ERROR_INVALID_ASN1_OBJECT;
+					goto err;
+				}
 				priv->offCardCertURL = calloc(1,urllen+1);
 				if (priv->offCardCertURL == NULL)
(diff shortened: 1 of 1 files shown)


frankmorgner:
Sorry, we don't have the capacity to manage and backport security patches for older versions. Please see https://github.com/OpenSC/OpenSC/security/policy

PR #3696, opened 2026-06-01
pkcs11-tool unwrap_key should read to dynamically allocated buffer instead static one

Wawszczak:
Currently pkcs11-tool assumes in unwrap_key function that all input files are smaller or equal to 2048 bytes. 
But currently used algorithms often exceed this limit.
My patch replaces static buffer with dynamically allocated one of length 8192 because additional 8k bytes is big enough to strip down the exe file.
Patch was compiled and tested on ubuntu 24.04.
Exact changes:
1. changed declaration of in_buffer to the pointer;
2. added variable with in_buffer_length = 8192;
3. changed initialization to malloc();
4. changed read() to use in_buffer_length;
5. added free(in_buffer) at the end of function

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -4136,7 +4136,8 @@ unwrap_key(CK_SESSION_HANDLE session)
 	int n_attr = 2;
 	CK_RV rv;
 	int fd;
-	unsigned char in_buffer[2048];
+	unsigned char* in_buffer = NULL;
+	CK_ULONG in_buffer_length = 8192;
 	CK_ULONG wrapped_key_length;
 	CK_BYTE_PTR pWrappedKey;
 	params_t params = {0};
@@ -4145,11 +4146,15 @@ unwrap_key(CK_SESSION_HANDLE session)
 	CK_OBJECT_HANDLE hUnwrappingKey;
 	ssize_t sz;
 
+	in_buffer = malloc(in_buffer_length);
+	if(in_buffer == NULL)
+		util_fatal("Cannot allocate enough memory for input buffer\n");
+
 	if (!find_object(session, CKO_PRIVATE_KEY, &hUnwrappingKey,
 			 opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, NULL, 0))
 		if (!find_object(session, CKO_SECRET_KEY, &hUnwrappingKey,
 				 opt_object_id_len ? opt_object_id : NULL, opt_object_id_len, NULL, 0))
-			util_fatal("Private/secret key not found");
+			util_fatal("Private/secret key not found\n");
 
 	if (!opt_mechanism_used)
 		util_fatal("Unable to unwrap, no mechanism specified\n");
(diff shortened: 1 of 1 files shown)


popovec:
I'm not sure if this solves anything. You can easily rewrite unsigned `char in_buffer[2048];` to `unsigned char in_buffer[8192];` and it will be much simpler. If we were to use dynamic allocation, it is advisable to write it so that the size of the allocated buffer adapts to the reading of the input file.

Wawszczak:
> You can easily rewrite unsigned `char in_buffer[2048];` to `unsigned char in_buffer[8192];` and it will be much simpler.

You are absolutelly right. It will be simpler, but this approach makes executable bigger and bigger. Why should we enforce users to download or store on drive empty buffers? 
> If we were to use dynamic allocation, it is advisable to write it so that the size of the allocated buffer adapts to the reading of the input file.

Yeap, I have considered this approach, but the code supports (and It's great!) reading input from stdin. This means we don't know size during buffer allocation.
I can write code to reallocate buffers in chunks like 512 bytes during reading from stdin, but this will be a lot more complicated.
First of all, I would like to achieve goal of support large input blobs.

frankmorgner:
you are not dynamically allocating the buffer accordingly to the length of the input. Hence, the flexibility you desire has not been implemented.

Wawszczak:
> you are not dynamically allocating the buffer accordingly to the length of the input. Hence, the flexibility you desire has not been implemented.

Yes, because of the reason I have mentioned about stdin. I am just avoiding presence of buffer inside the binary file, what happens when buffer is declared statically.

popovec:
> > You can easily rewrite unsigned `char in_buffer[2048];` to `unsigned char in_buffer[8192];` and it will be much simpler.
> 
> You are absolutelly right. It will be simpler, but this approach makes executable bigger and bigger. Why should we enforce users to download or store on drive empty buffers?

No, executable is not bigger..  `unsigned char in_buffer[8192];` is allocated on stack. Malloc allocates RAM on heap. Stack and hap is not a part of executable binary file.

popovec:
> Yes, because of the reason I have mentioned about stdin. I am just avoiding presence of buffer inside the binary file, what happens when buffer is declared statically.

When reading STDIN, allocate a smaller buffer, then when `read()` fills it, call `realloc` and you can use another `read()`.

Wawszczak:
@popovec 
I have aligned the code with your suggestions.

popovec:
I remain skeptical about using malloc/realloc in this part of the code. Considering that the maximum length of private keys in PQC today do not exceed 15kbytes (Classic McEliece).  If you are considering supporting, for example, the Rainbow algorithm with a key size of about 1.4 Mbytes, I still don't know of a cryptographic card that would support this. 
I don't know what other OpenSC developers think about this case, I would just increase the buffer to 16kbytes for now.

Wawszczak:
> I remain skeptical about using malloc/realloc in this part of the code. Considering that the maximum length of private keys in PQC today do not exceed 15kbytes (Classic McEliece). If you are considering supporting, for example, the Rainbow algorithm with a key size of about 1.4 Mbytes, I still don't know of a cryptographic card that would support this. I don't know what other OpenSC developers think about this case, I would just increase the buffer to 16kbytes for now.

Dear @popovec ,

If this is better from your perspective, I don't insist. 
Just give a hint - should I close this one and create new pull request, or should I made this change in this thread?

My primary goal is to achieve official pkcs11-tool ready for bigger key sizes.

Best Regards,
Stanislaw

popovec:
Please wait for the reaction of other people from the OpenSC project. Do not close this PR now, if you make a later change, please force push it to this PR, so that this discussion remains available.

frankmorgner (approved):

Jakuje (approved):

Issue #1234, opened 2018-01-08
Idea: use on Windows SCard caching function as replacement for file cache


vletoux:
To not forget this idea:

On Windows SCardReadCache and SCardWriteCache are available starting Windows 7 to cache smart card data in the smart card service. The advantage is that the data is shared among all users and used typically to cache certificates.
(LogonUI read it, then lsass.exe, etc saving a lot of transactions)

Unfortunately this function is not available on Windows Vista, and older OS ...

To compile it and make it available on older OS, the winscard.dll should be load using the "delay load library" mechanism so when this function is requested and not available, the functions are redirected to an empty stub.
(see __pfnDliFailureHook2)

Vincent

frankmorgner:
This could be a drop in replacement for the file cache on disk. If not configured properly, the `file_cache_dir`, may lead to permission problems. To avoid this with `SCardReadCache`, would be great!

On the other hand, replacing the cache on disk with `SCardReadCache`, could be problematic. If I remember correctly, we only have the `CardIdentifier` when the card is completely initialized (i.e. OpenSC has already read all the certificates)...

frankmorgner:
Even more beneficial, caching should be done on the card driver level instead of the PKCS#15 level. This way we could remove any internal card driver caching mechanism.

PR #3697, opened 2026-06-03
CI: Added workflow to manage inactive PR/Issues automatically

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,8 +1,6 @@
 name: 'Close stale issues and PRs'
 
 permissions:
-  actions: write
-  contents: write # only for delete-branch option
   issues: write
   pull-requests: write
 
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,22 @@
+name: 'Close stale issues and PRs'
+
+permissions:
+  actions: write
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v10
+        with:
+          stale-issue-message: 'This issue has been marked as stale due to inactivity. If there is no further activity, comment, or update within the next 2 months, this issue will be automatically closed. If this issue is still relevant, please leave a comment or provide an update to keep it open. Thank you for your contributions.'
+          stale-pr-message: 'This pull request has been marked as stale due to inactivity. If there are outstanding review comments, requested changes, failing tests, or missing information, please address them and provide the necessary updates. If no further activity occurs within the next 2 months, this pull request will be automatically closed. If this pull request is still active, please leave a comment or push an update to keep it open. Thank you for your contribution.'
+          start-date: '2026-07-01T00:00:00Z' # ISO 8601 or RFC 2822
+          days-before-stale: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Close stale issues and PRs'
+
+permissions:
+  actions: write
+  contents: write # only for delete-branch option
```
Jakuje: We do not use delete-branch so I think this permission is not needed

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v10
+        with:
+          stale-issue-message: 'This issue has been marked as stale due to inactivity. If there is no further activity, comment, or update within the next 2 months, this issue will be automatically closed. If this issue is still relevant, please leave a comment or provide an update to keep it open. Thank you for your contributions.'
+          stale-pr-message: 'This pull request has been marked as stale due to inactivity. If there are outstanding review comments, requested changes, failing tests, or missing information, please address them and provide the necessary updates. If no further activity occurs within the next 2 months, this pull request will be automatically closed. If this pull request is still active, please leave a comment or push an update to keep it open. Thank you for your contribution.'
+          start-date: '2026-07-01T00:00:00Z' # ISO 8601 or RFC 2822
+          days-before-stale: 60
```
Jakuje: should we use `exempt-draft-pr`?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Close stale issues and PRs'
+
+permissions:
+  actions: write
```
Jakuje: what is this permission for? There is nothing about that on the action page and the documentation from github is not informative either

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

Jakuje (commented):
otherwise sounds reasonable. I would consider if the default 60 days is the right time for us. We usually close issues and PRs we want earlier, but there are times when we are less responsive and the issues and PRs accumulate. But I think we can tweak it later.

.github/workflows/stale.yml
frankmorgner: I think we also want to see progress from drafts...

frankmorgner:
I removed the permissions - I don't know what they are meant for, either. I guess that over time we will see some more corner cases that will need to get tweaked, but my aim here is to get rid of the big backlog we currently have, even if that means that some things are perfectly addressed. Obviously, most of us can live with the many limitations we currently have...

frankmorgner:
If needed, just suggest other time frames

Jakuje:
> If needed, just suggest other time frames

I do not have specific time in my mind. But looking at our backlog going to 2018 I am worried if thats not too aggressive to require update every 2 months to avoid closing it. Though the oldest issue #1234 was really not touched since it was created. OTOH on the last page I can see still couple of useful issues/ideas that would be sad to see them go. I would probably aim for some more permissive timeline of 180 days of inactivity and 30 days to close, at least from the beginning

frankmorgner:
Then let's say 4 months until marking it as stale and 1 month until close. Note that this policy only applies to everything that will be opened starting at 1st of july, it does not apply to the old issues.

frankmorgner:
I updated the timeframes as discussed

Jakuje (approved):

PR #3698, opened 2026-06-03
Draft AGENTS.md

Jakuje:
This should give rough instruction for the AI agents and tools to match what we expect from them

##### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,25 @@
+# AI Agent Instructions for OpenSC
+
+Welcome, autonomous agents. If you are analyzing the OpenSC repository, finding bugs, or preparing to submit
+an issue or pull request, you MUST adhere strictly to the following rules:
+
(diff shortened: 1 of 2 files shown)


frankmorgner:
please add that the report should augment referenced files and line numbers with github links to source code files, otherwise this looks reasonable

frankmorgner:
security bugs should not be submitted in bulk - each issue should get its own report to fulfil Github's policy for requesting CVEs

Jakuje:
> please add that the report should augment referenced files and line numbers with github links to source code files, otherwise this looks reasonable

Added

> security bugs should not be submitted in bulk - each issue should get its own report to fulfil Github's policy for requesting CVEs

This sounds to me better placed to SECURITY.md as this applies also for human reports. Added

frankmorgner (approved):

PR #3701, opened 2026-06-07
srbeid: support RSA-2048 RAW signing and 256-byte decipher

hirashix0:
# srbeid: support RSA-2048 RAW signing and 256-byte decipher

Adds RSA-2048 RAW signing (`CKM_RSA_X_509`) and makes 256-byte decipher
work on the Serbian CardEdge `srbeid` driver (builds on #3595/#3662/#3665).

## Why / how

Two gaps today: RAW signing is not offered, and 256-byte decipher does not
work — a 256-byte RSA-2048 block does not fit a short APDU, and this applet
has no extended-length APDUs (`6700`) and no command chaining (`6E00`).

Testing showed the applet treats **P2 as the first byte of the command
data** (not as an ISO 7816-8 template), so the block is sent with byte 0 in
P2 and bytes 1..n-1 in the body, which fits the short-APDU Lc.

- Advertise **raw RSA only** (`SC_ALGORITHM_RSA_RAW | SC_ALGORITHM_RSA_HASH_NONE`);
  OpenSC does the PKCS#1 v1.5 / PSS padding and the constant-time depad in
  software, and the card only does a bare modexp.
- `decipher` **reuses `compute_signature`** — for raw RSA, sign and decrypt
  are the same `c^d mod N`, selected by the preceding MSE SET. Both issue
  PSO COMPUTE SIGNATURE (P1=0x9E); with the native PSO DECIPHER (P1=0x80),
  testing showed P2 is treated as an ISO tag and a reserved leading byte
  (e.g. `0x86`) is rejected, so the P1=0x9E path is used instead.

## Testing

Hardware-tested on a Serbian eID (SmartCafe Expert v8.0 C2), correct PIN only:

| Test | Result |
|------|--------|
| `SHA256-RSA-PKCS` sign → `openssl dgst -verify` | Verified OK |
| `RSA-X-509` (raw) sign → `openssl pkeyutl -verifyrecover` | byte-equal |
| `RSA-PKCS` decrypt round-trip, incl. a leading-`0x86` cryptogram | byte-equal |
| `RSA-X-509` decrypt round-trip, incl. a leading-`0x86` block | byte-equal |
| reserved-leading-byte sweep (sign 0x00–0xC0 incl. 0x9A/0x9E; decrypt incl. 0x86) | byte-exact, no collision |

`git-clang-format --diff upstream/master`: no changes; builds clean under
`-Wall -Wextra -Werror -Wstrict-aliasing=2`. One file (`card-srbeid.c`).

## Notes

Advertising `RSA_RAW` makes OpenSC also offer `RSA-PKCS-PSS` / `-OAEP`
(software-encoded over the raw modexp), the same surface raw-only drivers
such as `card-piv` / `card-coolkey` / `card-cac` expose. Measured on a
single card variant; the `srbeid` ATR table spans several.

### DIFF
--- a/src/libopensc/card-srbeid.c
+++ b/src/libopensc/card-srbeid.c
@@ -32,8 +32,9 @@
 #include "internal.h"
 #include "log.h"
 
-/* MSE algorithm byte for RSA-2048 PKCS#1 v1.5 */
-#define CE_MSE_ALG_RSA2048 0x02u
+/* MSE SET algorithm byte (tag 0x80). */
+#define CE_MSE_ALG_RSA_PKCS1 0x02u /* card adds PKCS#1 padding */
+#define CE_MSE_ALG_RSA_RAW   0x00u /* raw RSA; OpenSC pads */
 
 static struct sc_card_operations srbeid_ops;
 static const struct sc_card_operations *iso_ops;
@@ -84,8 +85,12 @@ srbeid_init(sc_card_t *card)
 
 	card->caps |= SC_CARD_CAP_ISO7816_PIN_INFO;
 
+	/* Keep the card's own PKCS#1 signing, and add raw RSA on top: that gives
+	 * CKM_RSA_X_509 and makes 2048-bit decrypt work (OpenSC does the padding). */
 	_sc_card_add_rsa_alg(card, 2048,
-			SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE, 0);
+			SC_ALGORITHM_RSA_PAD_PKCS1_TYPE_01 | SC_ALGORITHM_RSA_RAW |
+					SC_ALGORITHM_RSA_HASH_NONE,
+			0);
 
 	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 }
@@ -201,10 +206,13 @@ srbeid_set_security_env(sc_card_t *card,
 		return SC_ERROR_NOT_SUPPORTED;
 	}
 
-	/* MSE SET: tag 0x80 = algorithm (RSA2048), tag 0x84 = key ref (2 bytes BE) */
+	/* MSE SET: tag 0x80 = algorithm, tag 0x84 = key ref (2 bytes BE).
+	 * The algorithm byte tells the card to pad (PKCS#1) or not (raw RSA). */
 	mse_data[0] = 0x80;
 	mse_data[1] = 0x01;
-	mse_data[2] = CE_MSE_ALG_RSA2048;
+	mse_data[2] = (env->algorithm_flags & SC_ALGORITHM_RSA_RAW)
+				      ? CE_MSE_ALG_RSA_RAW
+				      : CE_MSE_ALG_RSA_PKCS1;
 	mse_data[3] = 0x84;
 	mse_data[4] = 0x02;
 	mse_data[5] = (u8)((key_ref >> 8) & 0xFF);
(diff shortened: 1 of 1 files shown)


frankmorgner (changes requested):
please be transparent about the tool you are using to produce this code.

please change the code so that the existing signing method is left intact

hirashix0:
Currently RSA-2048 decryption does not work on this card (a 256-byte block does not fit in a short APDU, and there is no extended APDU or command chaining), and raw RSA signing (CKM_RSA_X_509) is not offered. This change is an attempt to enable both of them.

About the tooling: sure — similar like in the earlier reverse-engineering of this card and its protocol, I used an AI tool (Claude Code) to help me analyze the APDU/PSO behaviour and draft the code. But I read through everything myself and test it on the real card before I push anything.

About the signing: you are right. I wanted to keep the driver simple and reduce the branching to support this, so I went with raw-only — but I see now that it changed the existing signing path. I'll try to rework it so the card's native PKCS#1 signing stays untouched, and the raw RSA on top. Will get back with the change and testing results.

hirashix0:
Re-verified on hardware after the rework (Serbian eID, SmartCafe Expert v8.0 C2), correct PIN only. Public keys read from the card's certs into kxc.pem (Key Exchange, id 01) and dsc.pem (Digital Signature, id 02).

Signing:
```
echo "hello world" > msg.txt
  pkcs11-tool --module opensc-pkcs11.so --login --pin <PIN> \
      --sign --mechanism SHA256-RSA-PKCS --id 02 --input-file msg.txt --output-file sig.bin
  openssl dgst -sha256 -verify dsc.pem -signature sig.bin msg.txt          # Verified OK
```
Decryption:
  ```
openssl pkeyutl -encrypt -pubin -inkey kxc.pem -in msg.txt -out enc.bin -pkeyopt rsa_padding_mode:pkcs1
  pkcs11-tool --module opensc-pkcs11.so --login --pin <PIN> \
      --decrypt --mechanism RSA-PKCS --id 01 --input-file enc.bin --output-file dec.txt
  cat dec.txt                                                              # hello world
  ```
Raw signing (RSA-X-509) — raw takes a full key-size block, so the SHA-256 digest is wrapped in a PKCS#1 block first:
  ```
printf '\x30\x31\x30\x0d\x06\x09\x60\x86\x48\x01\x65\x03\x04\x02\x01\x05\x00\x04\x20' > di.bin
  openssl dgst -sha256 -binary msg.txt >> di.bin
  printf '\x00\x01' > emsa.bin 
  printf '\xff%.0s' {1..202} >> emsa.bin
  printf '\x00' >> emsa.bin
  dd if=di.bin >> emsa.bin                                                 # emsa.bin = 256 bytes
  pkcs11-tool --module opensc-pkcs11.so --login --pin <PIN> \
   --sign --mechanism RSA-X-509 --id 01 --input-file emsa.bin --output-file raw.sig
  openssl dgst -sha256 -verify kxc.pem -signature raw.sig msg.txt          # Verified OK          
```
 All paths verified:
  - RSA-PKCS sign (DigestInfo, card-side) — Verified OK
  - SHA256-RSA-PKCS sign — Verified OK
  - RSA-X-509 (raw) sign — Verified OK
  - RSA-PKCS decrypt, including a leading-0x86 cryptogram — round-trips
  - RSA-X-509 (raw) decrypt — round-trips

frankmorgner:
What I find irritating, is that before your change `sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0x2A, 0x80, 0x86);` has been used for deciphering, which is the intended PSO for that operation. I understand, that this is technically identical in case of RAW RSA operations. However, would it be possible to make the split of APDUs also for 002a8086, so that the card also gets to know the intended use of that operation? I would also prefer a clean split between intended uses in the opensc card driver.

hirashix0:
OK. I tried to simplify it and keep the original structure: decryption again uses PSO DECIPHER (00 2A 80), so the card is told the operation. The 256-byte block still goes through the P2-split (first byte in P2), and the only addition is a fallback: when that first byte is 0x86 the applet reads P2 as an ISO template and rejects it, so for those I fall back to COMPUTE SIGNATURE (00 2A 9E), which is the same RSA operation on this card. Signing and decryption now share one small srbeid_pso() helper.

dengert:
In srbeid_init  line 91 you say the card supports SC_ALGORITHM_RSA_PAD_PKCS1_TYPE_01 in hardware. If you are willing to stick with RAW mode for both sign and decrypt opensc will do it in software.  Try remove the SC_ALGORITHM_RSA_PAD_PKCS1_TYPE_01 it might simplify you code.

hirashix0:
Thanks. 
RAW-only is where this PR started — the first version advertised SC_ALGORITHM_RSA_RAW | SC_ALGORITHM_RSA_HASH_NONE. @frankmorgner asked me to leave the existing signing method intact, so I added PKCS1_TYPE_01 to keep signing byte-identical to the current driver. Removing PKCS1_TYPE_01 would simplify the code a bit, but it would switch signing back to RAW — exactly what he asked me to avoid. How should we handle it now :) ?

dengert:
Another good question.  
 "@frankmorgner asked me to leave the existing signing method intact."
Some cards support RSA padding on the card and others like the PIV only support RAW RSA. So it is really up to the card issuer if they consider the security to be better with padding on the card or in software.   It looks like you are trying to modify a card driver for government issued cards that others may depend to work on older versions of cards that may not support RAW.
So I agree with Frank.

frankmorgner (approved):

frankmorgner:
I think the code is in good shapre now, making a good spearation between decryption and signature. If it is not clear what whas intended with the card, then we should try to give control to the card, but that's now handled good enough.

hirashix0:
Thanks, both. I'll squash the three commits into one clean commit before merge.

PR #3702, opened 2026-06-07
piv: add ATR for Giesecke+Devrient SCE7

hirashix0:
# piv: add ATR for Giesecke+Devrient SCE7

## What

Adds the ATR of the Giesecke+Devrient SCE7 (a PIV-only card) to the PIV
driver's ATR table. One row.

## Why

The SCE7 uses a vendor-specific PIX, so its tag 0x4F PIX does not match the
NIST PIX in `piv_aids[]`; without an ATR entry it falls back to
`SC_CARD_TYPE_PIV_II_BASE` and is rejected. But the card returns a valid
**Discovery object** whose tag 0x4F carries the full NIST PIV AID, so once
its ATR maps it to `SC_CARD_TYPE_PIV_II_GI_DE` the card is recognized via
the discovery-first path (`piv_find_aid` is not used). The single ATR row
is therefore sufficient.

Reuses the existing `SC_CARD_TYPE_PIV_II_GI_DE` type (no new enum), as #2995
did for an Idemia card.

## Testing

On a real G+D SCE7 (ATR `3B:F9:96:...:46:4E`):

- card binds as PIV; both certificates read (PIV Authentication, Digital
  Signature);
- `SHA256-RSA-PKCS` sign with the PIV Authentication (id 01) and Digital
  Signature (id 02) keys — both `openssl dgst -verify` → **Verified OK**.

## Build

- `git-clang-format --diff upstream/master`: no changes.
- Builds clean under `-Wall -Wextra -Werror -Wstrict-aliasing=2`.
- One file (`src/libopensc/card-piv.c`), +2 lines.

Tested on a single card variant (G+D SCE7).

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -484,6 +484,8 @@ static const struct sc_atr_table piv_atrs[] = {
 	{ "3B:7A:18:00:00:73:66:74:65:20:63:64:31:34:34", NULL, NULL, SC_CARD_TYPE_PIV_II_GI_DE_DUAL_CAC, 0, NULL },
 	/* Giesecke & Devrient (CAC PIV Endpoint) 2019 */
 	{ "3B:F9:18:00:00:00:53:43:45:37:20:03:00:20:46", NULL, NULL, SC_CARD_TYPE_PIV_II_GI_DE_DUAL_CAC, 0, NULL },
+	/* Giesecke & Devrient SCE7 (PIV-only) */
+	{ "3B:F9:96:00:00:80:31:FE:45:53:43:45:37:20:0F:00:20:46:4E", NULL, NULL, SC_CARD_TYPE_PIV_II_GI_DE, 0, NULL },
 
 	/* IDEMIA (new name for Oberthur) (DoD Alternate Token IDEMIA Cosmo V8.0 2019*/
 	{ "3B:D8:18:00:80:B1:FE:45:1F:07:80:31:C1:64:08:06:92:0F:D5", NULL, NULL, SC_CARD_TYPE_PIV_II_OBERTHUR, 0, NULL },


frankmorgner (approved):

dengert:
Since the G&D cards have been used as approved NIST cards in the past. Tthe  NIST standards  says:

NIST sp800-73-5 part 1https://csrc.nist.gov/pubs/sp/800/73/pt1/5/final  and (previous versions too) says:

> 2.2. PIV Card Application AID
The Application IDentifier (AID) of the Personal Identity Verification Card Application (PIV Card Application) SHALL be:
'A0 00 00 03 08 00 00 10 00 01 00'
The AID of the PIV Card Application consists of the NIST RID ('A0 00 00 03 08') followed by the application portion of the NIST PIX indicating the PIV Card Application ('00 00 10 00') and then the version portion of the NIST PIX ('01 00') for the first version of the PIV Card Application. All other PIX sequences on the NIST RID are reserved for future use.
The PIV Card Application CAN be selected as the current application by providing the full AID as listed above or by providing the right-truncated version (i.e., without the two-byte version), as follows:
'A0 00 00 03 08 00 00 10 00'

which may be interpreted as "NIST PIX ('01 00')" could be set by a vendor to some other value.

In the past the G&D cards have also been uses a dual PIC and CAC cards so tread special by OpenSC
so it was necessary to check the ATR. But this newer card may also support optional features such as SM and VCI. 
and maybe other cards may do the same.

Your problem may also imply:
 https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L2979-L3001
may be returning at: 
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L3001
This is in github master, if you can try your card against it.   

Can you also run this script to see what else shows up:
```
opensc-tool -c PIV-II -a \
	-s "00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00" \
	-s "00 FD 00 00 03" \
	-s "00 F8 00 00 00" \
	-s "00 CB 3F FF 05 5C 03 5F C1 0C 00" \
	-s "00 CB 3F FF 05 5C 03 5F C1 02 00" \
	-s "00 CB 3F FF 03 5C 01 7E 00"
```
These list the SELECT AID response, a Yubico version and serial (both may fail),  History object, CHUID object and Discovery object.  

If they can get away with not responding a vaild version number others may too. so a more generic fix then just adding
an ATR will work with other cards too.

dengert:
Looking closer at the code going back  20 years is causing the problem
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-piv.c#L2979-L3001 is in error
And it looks like the error in 2006/2007 when it should have also stopped from testing of PIX.

The original code from b051b08584  around lines 436
```
static struct piv_aid piv_aids[] = {
	{SC_CARD_TYPE_PIV_II_GENERIC, 
		 9, 11, (u8 *) "\xA0\x00\x00\x03\x08\x00\x00\x10\x00\x01\x00" },
	{0,  9, 0, NULL }
```
i.e.  NIST RID ('A0 00 00 03 08')   NIST PIX indicating the PIV Card Application ('00 00 10 00') 9 bytes 
with version portion of the NIST PIX ('01 00') total of 11 bytes.

ca92ec661f  changed it to have 9, 9 so "short" and "long" were the same.


The output of the opensc-tool from above would help a lot.

dengert:
You could try a  setting on opensc.conf something like
```
app default {
       # debug = 3;
       # debug_file = opensc-debug.txt;
       card_atr "3B:F9:96:00:00:80:31:FE:45:53:43:45:37:20:0F:00:20:46:4E" {
            driver = PIV-II;
            type = 14013;   # SC_CARD_TYPE_PIV_II_800_73_4
       } 
}
```
I would not set it to SC_CARD_TYPE_PIV_II_GI_DE as that is an older card  but the newer card looks like it meets all the requirements of sp800-73-4.

hirashix0:
Thanks! I ran the diagnostic and tried your `card_atr` suggestion. Here are results:

- Auto-detect: "Unsupported card".
- `opensc-tool -c PIV-II`: "Card is invalid or cannot be handled".
- With your `card_atr` override (`driver = PIV-II; type = 14013`, and without `-c` on the command line): the card comes up as "Personal Identity Verification Card" and works as PIV.

`opensc-tool` output (the CHUID's FASC-N and GUID redacted; opensc-tool's ASCII column dropped):

```
Using reader with a card: Gemalto USB SmartCard Reader
3b:f9:96:00:00:80:31:fe:45:53:43:45:37:20:0f:00:20:46:4e
Sending: 00 A4 04 00 09 A0 00 00 03 08 00 00 10 00 00
61 0F 4F 06 00 10 00 01 00 00 79 05 A0 00 00 03 08
Sending: 00 FD 00 00 03
Received (SW1=0x6D, SW2=0x00)
Sending: 00 F8 00 00 00
Received (SW1=0x6D, SW2=0x00)
Sending: 00 CB 3F FF 05 5C 03 5F C1 0C 00
53 08 C1 01 00 C2 01 00 FE 00
Sending: 00 CB 3F FF 05 5C 03 5F C1 02 00
53 3B 30 19 <FASC-N + GUID redacted> 35 08 20 20 20 20 20 20 20 20 3E 00 FE 00
Sending: 00 CB 3F FF 03 5C 01 7E 00
7E 12 4F 0B A0 00 00 03 08 00 00 10 00 01 00 5F 2F 02 40 00
```

A quick AI-assisted look flagged that the SELECT tag `4F` (`00 10 00 01 00 00`) doesn't match the standard PIV value, while tag `79` (`A0 00 00 03 08`) and the Discovery AID look standard. The `4F` value looks like the standard PIX `00 00 10 00 01 00` shifted by one byte — a leading `00` dropped and a trailing `00` added — so possibly an off-by-one in how the card builds the SELECT response. 
As a first attempt at the more generic fix you suggested, a fallback in `piv_find_aid()` — when the tag-`4F` PIX doesn't match, fall back to the RID in tag `79` — fixes the issue and the card is auto-detected as PIV. No ATR row needed. 

What would be the shape for the upstream fix, then? Should I polish that first attempt into a proper patch? 
Happy to test anything else against the card.

dengert:
Please try this diff. [260609-GDPIV.diff.txt](https://github.com/user-attachments/files/28760908/260609-GDPIV.diff.txt)


It includes some comment updates  from DOD as of 2025 with you card listed
https://www.cac.mil/Portals/53/Documents/DoD%20Token%20utilziation%20and%20variation%20matrix%20v2.11_April2025.pdf?ver=IlbJ7_p4AMccUFhlKmm-lg%3D%3D

It also includes your ATR setting. It  is not clear of you card has a CAC or not which An opensc debug log would show this.

hirashix0:
Your diff works — the card is identified as "Personal Identity Verification Card", both auto-detected and with `-c PIV-II`. Thanks!

Selecting the CAC applets all return `6A 82`:

```
00 A4 04 00 07 A0 00 00 00 79 02 00  ->  6A 82
00 A4 04 00 07 A0 00 00 01 16 DB 00  ->  6A 82
00 A4 04 00 07 A0 00 00 00 79 10 00  ->  6A 82
```

OpenSC's own `cac` driver rejects the card — `opensc-tool -c cac` → "Card is invalid or cannot be handled".

One small observation: I tried removing the ATR row from the diff and retested — the card is still recognized, both auto-detected and with `-c PIV-II`. I guess the ATR row is redundant here as well?

dengert:
> I guess the ATR row is redundant here as well?

When OpenSC was started smartcards had only one applet written by the vendor and the vendor provided the software drivers. Now  days, the user is more interested one or more applets on the same card for example PIV and OpenPGP or PIV and CAC.  There are also "PIV-like" cards/tokens, that have written their own PIV applets and added extensions, and there are opensource applets that can be installed by by anyone on different cards. 

The PIV driver is designed to handle any card which responds to a SELECT AID of the PIV applet If they really follow the NIST standard they should work. If G&D have followed the standard, we would not be having this discussion. 

But OpenSC also supports some of the "PIV Like" cards by looking at the ATR or the Historical bytes in the ATR or the response to SELECT AID to see if it can use some additional APDUs supported by the card or the card vendor's documentation to add special code.

hirashix0:
Thanks for clarifying! 
How should we proceed now — will this land in main? Should I update this PR with your diff, or would you prefer to continue and commit it yourself?

dengert:
I will commit it. Thanks for identifying this problem, others may have see it but not reported it.

hirashix0:
Superseded by #3703 — thanks for taking care of it. Closing this one.

PR #3709, opened 2026-06-17
Implemented test tool for macOS CryptoTokenKit

frankmorgner:
I used this to fix the known problems on macOS, which are solved, at least for ECC tokens.

##### Checklist

- [x] macOS token is tested


packaging/opensc.spec
```diff
@@ -143,6 +143,9 @@ rm -rf %{buildroot}%{_mandir}/man1/pkcs11-register.1*
 rm %{buildroot}%{_datadir}/applications/org.opensc.notify.desktop
 rm %{buildroot}%{_mandir}/man1/opensc-notify.1*
 
+# ctk-tool is only available on macOS
+rm %{buildroot}%{_mandir}/man1/ctk-tool.1*
+
```
Jakuje, line 148: Should we rather build the manual pages only when we have `ENABLE_CRYPTOTOKENKIT`? Right now everything is built with a wildcard, which makes it easy to add a new manual page, but it is annoying that unrelated manual pages are installed regardless of configutre options.

frankmorgner: theoretically yes, but we have not done this in the past, e.g. for all tools that use OpenSSL. I think it is OK to keep it simple (i.e. keep it as is)...

Jakuje (approved):

PR #3712, opened 2026-06-21
CI: Add Pico HSM

frankmorgner:
##### Checklist

### DIFF
--- a/.github/test-pico-hsm.sh
+++ b/.github/test-pico-hsm.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+set -ex -o xtrace
+
+source .github/setup-valgrind.sh
+
+# install the opensc
+sudo make install
+export LD_LIBRARY_PATH=/usr/local/lib
+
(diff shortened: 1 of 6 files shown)


.github/test-pico-hsm.sh
```diff
@@ -0,0 +1,46 @@
...
+pico-hsm/build/pico_hsm &
+PID=$!
+sleep 2
+
+# run the tests
+pushd src/tests/p11test/
+sleep 5
+$VALGRIND sc-hsm-tool --initialize --so-pin 3537363231383830 --pin 648219
+$VALGRIND ./p11test -s 0 -p 648219 -o pico-hsm.json
+diff -u3 pico-hsm_ref.json pico-hsm.json
```
Jakuje, line 42: right now you run tests without any keys loaded. It shows the basic infrastructure works, but it would be good to push some keys in there so its tested a bit more throughfully

Jakuje (commented):
Great job!

PR #3713, opened 2026-06-21
Add a hint about "openssl error during verification"

mvogt1:
--test gives an openssl error.

```
$ pkcs11-tool -t -l
[....]
testing key 1 (Authentication Certificate) with 1 mechanism
    RSA-PKCS: openssl error during verification: 0xffffffff (-1)
[...]

```

I had this "openssl error during verification" and my question on opensc couldn't be answered.

[Posting to opensc-devel](https://sourceforge.net/p/opensc/mailman/message/59322232/)

It took me some time until I found it out (by experimenting and printf in OpenSSL).
After I found the solution I made this patch which may help someone in the future.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -8435,6 +8435,7 @@ static int sign_verify_openssl(CK_SESSION_HANDLE session,
 		errors++;
 	} else if (err != 1) {
 		printf("openssl error during verification: 0x%0x (%d)\n", err, err);
+		ERR_print_errors_fp(stdout);
 	} else
 		printf("OK\n");
 


frankmorgner (changes requested):
In case of OpenSSL errors (`ERR_print_errors()`), it would be more helpful to actually print that out instead of adding a comment which is only visible (at all) to developers. See `sc_do_log_openssl()` in `log.c` for usage example.

dengert:
Why are you using SHA1 at all?  If  this is from the 2003 driver that was dropped fix the <jcop_init() at:

https://github.com/mvogt1/OpenSC/blob/devel/src/libopensc/card-jcop.c#L194-L196
and remove these lines.

dengert:
Your also looking at 23 year old piece of code from 2003 which says at: https://github.com/mvogt1/OpenSC/blob/devel/src/libopensc/card-jcop.c#L222-L228 it says OpenSC does not support: "7816-4 direct application selection"

The ATR of your card: 3B F9 96 00 00 80 31 FE 45 53 43 45 37 20 00 00 20 20 27

https://smartcard-atr.apdu.fr/ says it is a:
G&D SmartCafe Expert 7 (JavaCard)

OpenSC supports `iso7816_select_aid` and the AID is the standard PKCS15 AID which is in the EF.DIR 
the code following is trying to do PKCS15 calls which should also be handled like any other PKCS15 card. 

The ATR of your card: 3B F9 96 00 00 80 31 FE 45 53 43 45 37 20 00 00 20 20 27

https://smartcard-atr.apdu.fr/ says it is a:
G&D SmartCafe Expert 7 (JavaCard)

mvogt1:
>In case of OpenSSL errors (ERR_print_errors()), it would be more helpful to actually print 
>that out instead of adding a comment which is only visible (at all) to developers. See 
>sc_do_log_openssl() in log.c for usage example.


The new patch prints "ERR_print_errors", but I can't use sc_do_log_openssl because there is not sc_context_t.

The output now looks as follows:

```
[...]
  testing signature mechanisms:
    RSA-PKCS: openssl error during verification: 0xffffffff (-1)
80E6A047657F0000:error:03000098:digital envelope routines:evp_pkey_ctx_set_md:invalid digest:crypto/evp/pmeth_lib.c:977:
[...]

```

src/tools/pkcs11-tool.c
```diff
@@ -8435,6 +8435,14 @@ static int sign_verify_openssl(CK_SESSION_HANDLE session,
...
 	} else if (err != 1) {
 		printf("openssl error during verification: 0x%0x (%d)\n", err, err);
+		if (md_ctx)
+			/*
+                           Certain operating systems (fedora,..) block weak algorithms, eg: SHA1
+                           quickfix: update-crypto-policies --set DEFAULT:SHA1 enables them
+			*/
+			ERR_print_errors_fp ( stdout );
+
+
```
frankmorgner:
```suggestion
		ERR_print_errors_fp(stdout);
```

frankmorgner (changes requested):
If it is an OpenSSL error (what the error message claims), then we can print out the backtrace without further conditions.

mvogt1:
>If it is an OpenSSL error (what the error message claims), then we can print out the backtrace without further conditions.

I removed the comments and the if (md_ctx). 

In case the error occurs, pasting it into google shows the explanation about update-crypto-policies.

frankmorgner:
Thank you. In my suggestion I had also corrected the formatting. This is also what the CI workflow will complain about in a couple of minutes...

mvogt1:
I have totally overlooked the style guide in the last patch. 
Now I installed :

```
$ git-clang-format --diff 
clang-format did not modify any files
```

and checked the "CONTRIBUTING" section. I hope to make lesser mistakes in the future.

frankmorgner:
Thanks for your help

PR #3714, opened 2026-06-21
configure.ac: Fix reader detection

UVV-gh:
The configure.ac tries to make sure that at most 1 reader is selected, but misses the cases when all are disabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -349,8 +349,8 @@ reader_count=""
 for rdriver in "${enable_pcsc}" "${enable_cryptotokenkit}" "${enable_openct}" "${enable_ctapi}"; do
 	test "${rdriver}" = "yes" && reader_count="${reader_count}x"
 done
-if test "${reader_count}" != "x"; then
-	AC_MSG_ERROR([Only one of --enable-pcsc, --enable-cryptotokenkit, --enable-openct, --enable-ctapi can be specified!])
+if test -n "${reader_count}" && test "${reader_count}" != "x"; then
+	AC_MSG_ERROR([At most one of --enable-pcsc, --enable-cryptotokenkit, --enable-openct, --enable-ctapi can be specified!])
 fi
 
 dnl Checks for programs.
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -78,6 +78,9 @@ else
 	if [ "$1" == "no-openssl" ]; then
 		CONFIGURE_FLAGS="$CONFIGURE_FLAGS --disable-openssl"
 	fi
+	if [ "$1" == "no-readers" ]; then
+		CONFIGURE_FLAGS="$CONFIGURE_FLAGS --disable-pcsc --disable-cryptotokenkit --disable-openct --disable-ctapi"
+	fi
 	export CFLAGS="-DDEBUG_PROFILE=1 $CFLAGS"
 	./configure $CONFIGURE_FLAGS
 	make -j 4 V=1
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -80,6 +80,13 @@ jobs:
       - run: .github/setup-linux.sh
       - run: .github/build.sh no-shared valgrind
 
+  build-no-readers:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: .github/setup-linux.sh
+      - run: .github/build.sh no-readers
+
   build-no-openssl:
     runs-on: ubuntu-latest
     steps:
--- a/src/libopensc/Makefile.am
+++ b/src/libopensc/Makefile.am
@@ -35,7 +35,7 @@ libopensc_la_SOURCES_BASE = \
 	\
 	muscle.c muscle-filesystem.c \
 	\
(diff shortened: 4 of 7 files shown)


configure.ac
```diff
@@ -349,7 +349,7 @@ reader_count=""
 for rdriver in "${enable_pcsc}" "${enable_cryptotokenkit}" "${enable_openct}" "${enable_ctapi}"; do
 	test "${rdriver}" = "yes" && reader_count="${reader_count}x"
 done
-if test "${reader_count}" != "x"; then
+if test -n "${reader_count}" && test "${reader_count}" != "x"; then
 	AC_MSG_ERROR([Only one of --enable-pcsc, --enable-cryptotokenkit, --enable-openct, --enable-ctapi can be specified!])
```
Jakuje: I think you should adjust the error message to match what your new code is checking for now.

```suggestion
	AC_MSG_ERROR([Exactly one of --enable-pcsc, --enable-cryptotokenkit, --enable-openct, --enable-ctapi can be specified!])
```

UVV-gh: Actually no, my use case is exactly to allow none of the readers to be selected. I'm interested only in pkcs11-tool and in none of any readers, because I use OP-TEE backend.

Jakuje: oh, my bad. Thanks for clarification. In that case, wording like "At most one ..." might be more appropriate. It would be also helpful to have some CI job to make sure this keeps working.

Jakuje (approved):

configure.ac
UVV-gh: I hope this should do the trick now. Also rebased on the latest master

Jakuje (approved):

UVV-gh:
I believe the tests are not supposed to run without readers. Would it be OK if I add "--disable-integration-tests" to this job?

Jakuje:
I think the tests are ok to run with this configuration. We have mostly tests of tools in place. The two failures in the following version are known issues we did not manage to tackle yet:

https://github.com/OpenSC/OpenSC/commit/6ef8b7b46ffef913b8894b5a46b96aab84080efc

UVV-gh:
Ah, OK, then please let me know if anything else left here to fix.

Jakuje:
oh, this commit was with the tests: https://github.com/OpenSC/OpenSC/commit/586c314335b0761090d10e7d3dddf2f5f4419384 and it indeed crashes in some tests (same as the current version). It will need some investigation as I do not see a reason why sm or asn1 tests should crash when they are built without readers.

UVV-gh:
Here's the reason for the crash:
```C
#ifdef ENABLE_PCSC
    ctx->reader_driver = sc_get_pcsc_driver();
#elif defined(ENABLE_CRYPTOTOKENKIT)
    ctx->reader_driver = sc_get_cryptotokenkit_driver();
#elif defined(ENABLE_CTAPI)
    ctx->reader_driver = sc_get_ctapi_driver();
#elif defined(ENABLE_OPENCT)
    ctx->reader_driver = sc_get_openct_driver();
#endif

    r = ctx->reader_driver->ops->init(ctx);
    if (r != SC_SUCCESS)   {
        del_drvs(&opts);
        sc_release_context(ctx);
        return r;
    }
```

The `sc` context expects a reader and then calls `init` on a null pointer when none is provided. The context is a part of some unit tests setup method.

Do you wanna keep the PR open until it's fixed or rather merge and create a separate issue for that?

Jakuje:
I think it should be changed together with this change otherwise the first one who will want to build opensc without readers will see this crash. Not sure if just skipping the init call if there is no reader would be enough or we will need some more changes.

Jakuje (approved):
Thank you! I think this looks good now!

UVV-gh:
The style checker suggests 2-tab indentation for a struct? Is that intentional or looks like a false-positive? The other suggestions probably mandate 80-character line length. Is that a strict requirement?

Jakuje:
> The style checker suggests 2-tab indentation for a struct? Is that intentional or looks like a false-positive?

I no longer remember where does this come from. Maybe this rule?

https://github.com/OpenSC/OpenSC/blob/master/.clang-format#L25

I know we tweaked the rules to work for all of us, but some of the result in odd artifacts like that. Adhering to the formatter makes the review less noisy.

> The other suggestions probably mandate 80-character line length. 

No, these are matching the following rule:

https://github.com/OpenSC/OpenSC/blob/master/.clang-format#L31

I think we should use the `ExceptShortType` option, but when we wrote it, we did not have new-enough clang:

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakafterreturntype

> Is that a strict requirement?

No. These are mostly suggestions to avoid unreadable mess (which we still have in some files). My approach is as long as it is consistent and readable readable, its ok. But I see there are some cases where the pointers are used inconsistently even on one line (`sc_context_t *ctx`, `void * empty_context_handle`, likely copied from some other file) so please try to adjust them as suggested. Whether the return type is on the same line or not, its not that of a big deal (but having them on separate line makes the orientation in the code easier, I believe).

Jakuje (approved):

UVV-gh:
Is it now good for merge or something else is still missing?

Jakuje:
All good. I was leaving time for others to voice their concerns, but given that there were none, I will merge it. Thank you for your contribution!

PR #3715, opened 2026-06-22
Fixing minor memory issues

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/reader-pcsc.c
+++ b/src/libopensc/reader-pcsc.c
@@ -2182,16 +2182,14 @@ static int part10_build_modify_pin_block(struct sc_reader *reader, u8 * buf, siz
 /* Find a given PCSC v2 part 10 property */
 static int
 part10_find_property_by_tag(unsigned char buffer[], DWORD length,
-	int tag_searched)
+		int tag_searched)
 {
 	unsigned char *p;
 	int found = 0, len, value = -1;
 
 	p = buffer;
-	while (p-buffer < (long)length)
-	{
-		if (*p++ == tag_searched)
-		{
+	while (p - buffer < (long)length) {
+		if (*p++ == tag_searched) {
 			found = 1;
 			break;
 		}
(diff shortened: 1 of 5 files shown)


Jakuje (approved):

src/libopensc/reader-pcsc.c
```diff
@@ -2201,23 +2201,24 @@ part10_find_property_by_tag(unsigned char buffer[], DWORD length,
...
-			case 4:
-				value = *p + (*(p+1)<<8) + (*(p+2)<<16) + (*(p+3)<<24);
-				break;
-			default:
-				value = -1;
+		if (p-buffer+len <= (long)length) {
+			switch(len)
+			{
+				case 1:
+					value = *p;
```
Jakuje: when you reformat this, please fix also the switch/case:

```suggestion
			switch (len) {
			case 1:
				value = *p;
```

src/libopensc/reader-pcsc.c
```diff
@@ -2201,23 +2201,24 @@ part10_find_property_by_tag(unsigned char buffer[], DWORD length,
 		p += len;
 	}
 
-	if (found)
-	{
+	if (found && p-buffer < (long)length) {
```
Jakuje: isn't pointer arithmetic of a `size_t`?
```suggestion
	if (found && p-buffer < (size_t)length) {
```
in any case, I would find something like following easier to follow without the explicit cast, but its probably matter of taste:
```suggestion
	if (found && p >= buffer + length) {
```

frankmorgner:
Yeah, formatting and casting were a choice from the past. I'll see what I can do...

src/libopensc/reader-pcsc.c
frankmorgner: Obviously, p-buffer yields a 'long int' which errors with comparison against 'long unsigned int' (size_t). I think the code is readable enough so I kept the original structure.

Jakuje: my bad. Sorry for a confusion.

Jakuje (approved):

Jakuje:
the CI failures look like flaky tests to me.

PR #3729, opened 2026-06-29
Add support for the Uruguayan national eID card (cédula de identidad)

carlosplanchon:
*(Description updated to reflect the final implementation after review. The first iteration patched `card-iasecc.c` with a card subtype; see the review below for how the implementation evolved.)*

## Summary 🇺🇾

The Uruguayan national identity card (*cédula de identidad electrónica*) is a Gemalto/Thales "Classic V4" IAS/ECC platform card. On Linux, its only available middleware has been the closed-source Thales Classic Client, so there was no open stack for it. As a Uruguayan citizen, my goal is for the card to be usable with open, auditable tooling rather than only that closed-source middleware.

This PR makes OpenSC drive the card end to end over the contact interface: recognition, certificate read, PIN verification, and signing, with the signature verifying against the card's own certificate.

## Design

The subset exposed by the card over the contact interface is plain ISO 7816, so support is implemented as:

- A dedicated card driver, **`card-cedulauy.c`**, built on the generic `iso7816` operations and following the model of `card-esteid2018.c`. It selects the IAS application by AID at initialization; nothing on the card is accessible before that selection. It reselects the application after a card reset, emulates the MF (the card exposes none, so `3F00` maps to the application), and signs using `MSE SET DST`, followed by `PSO HASH` and `PSO COMPUTE DIGITAL SIGNATURE`. The card builds the DigestInfo on-card for SHA-256. For other hash-and-sign mechanisms, OpenSC hashes in software and submits the resulting DigestInfo for raw PKCS#1 v1.5 signing. Random data is read in 8-byte chunks.

- A synthetic PKCS#15 emulator, **`pkcs15-cedulauy.c`**. The card's on-card `EF(TokenInfo)` encodes one `supportedAlgorithms` entry as an empty `SEQUENCE`, where PKCS#15 expects `NULL` or an OID, so the generic binder rejects it. The emulator instead builds the PKCS#15 view, consisting of the Global PIN, signing certificate, and RSA-2048 key, from AGESIC's documented layout. The token label and serial number come from the signing certificate, and the serial number keys the file cache, which is enabled by default because the card profile is static.

Review also surfaced a possible `NULL`-destination dereference in `sc_file_dup()`, which is now guarded in the library itself.

## Card identification

| | |
|---|---|
| Card | Uruguayan eID (*cédula de identidad*), Gemalto "Classic V4" |
| ATR | `3B 7F 94 00 00 80 31 80 65 B0 85 03 00 EF 12 0F FF 82 90 00` |
| ATR mask | `FF FF 00 FF FF FF FF FF FF FF FF 00 00 00 00 00 00 FF FF FF` (TA1 and the version/batch bytes vary between batches) |
| Application AID | `A0 00 00 00 18 40 00 00 01 63 42 00` |
| Signing certificate | EF `B001` |
| Signing key | Reference `0x01`, RSA 2048 |
| Global PIN | Reference `0x11`, zero-padded to 12 bytes |

## Testing

- Two physical cards from different batches: mine (pre-2022) on an ACS ACR38U, and @nicolasgutierrezdev's (March 2026) on an Alcor Link AK9563.
- Recognition, certificate read, both cold and from the file cache, PIN verification, and signing. `pkcs11-tool --test --login` reports no errors on both cards.
- Signatures verify against the certificate's public key with OpenSSL, and @nicolasgutierrezdev signed and validated a PAdES document end to end.

## Scope and follow-ups

This PR supports the contact interface only. The following are planned as separate work:

- ID data and face image as generic PKCS#15 data objects (EFs `7001`, `7002`, `7004`, `700B`, and the SOD in `711D`, as documented by AGESIC).
- Document number as the token serial via `SC_CARDCTL_GET_SERIALNR`, weighing the privacy trade-off discussed in the review.
- Contactless support with PACE. Cards issued since late 2022 support it, and @nicolasgutierrezdev's findings will move to a dedicated issue.

## References

Card conventions come from AGESIC's official, public materials; no reverse-engineering of the closed-source middleware was involved:

- [Información Técnica de Cédula de Identidad con chip](https://www.gub.uy/agencia-gobierno-electronico-sociedad-informacion-conocimiento/comunicacion/publicaciones/documentacion-tecnica-id-uruguay/documentacion-tecnica-id-uruguay-9) (AGESIC).
- [Documentación técnica de la cédula de identidad con chip](https://www.gub.uy/agencia-gobierno-electronico-sociedad-informacion-conocimiento/book/7192/download) (AGESIC, compiled PDF; the APDU signing chapter includes the `AlgoID` table and a worked example).
- [Autenticación y Firma Avanzada a través de PKCS#11](https://www.gub.uy/agencia-gobierno-electronico-sociedad-informacion-conocimiento/comunicacion/publicaciones/documentacion-tecnica-cedula-identidad-chip/documentacion-tecnica-3) (AGESIC).
- [`eIDuy/apdu-services`](https://github.com/eIDuy/apdu-services) (AGESIC reference code using Java `javax.smartcardio`).
- [`agesic-eid/Interfaz-datosci-apdu`](https://github.com/agesic-eid/Interfaz-datosci-apdu) (AGESIC reference code).

## Appendix: signing APDU sequence (contact interface, no Secure Messaging)

All commands use `CLA=00` (plain). For a standard `SHA256withRSA` signature, the driver uses algorithm reference `0x42`, with SHA-256 in the high nibble and PKCS#1 v1.5 in the low nibble. It loads the bare 32-byte SHA-256 digest, and the card builds the DigestInfo, so the result verifies as a standard RSA signature.

| Step | APDU | Notes |
|---|---|---|
| SELECT application | `00 A4 04 00 0C` + AID | Application AID |
| VERIFY PIN | `00 20 00 11 0C` + PIN zero-padded to 12 bytes | Global PIN |
| MSE SET (DST) | `00 22 41 B6 06 84 01 01 80 01 42` | Key ref `01`, algorithm `0x42` |
| PSO HASH | `00 2A 90 A0 22 90 20` + 32-byte SHA-256 digest | Load digest (tag `90`) |
| PSO COMPUTE DIGITAL SIGNATURE | `00 2A 9E 9A 00` | RSA-2048, 256-byte signature |

## Credits

- @frankmorgner contributed the PKCS#15 appinfo and emulated MF commits, and suggested the `sc_file_dup()` robustness fix.
- @nicolasgutierrezdev contributed the ATR mask, tested a second card batch end to end, and investigated the contactless interface.
- @Jakuje reviewed the final implementation and suggested cleanup around the algorithm constants and structure initialization.

Developed with assistance from Claude Code (Opus 4.8), Fable 5, and ChatGPT Sol Extra High.

### DIFF
--- a/src/libopensc/card-cedulauy.c
+++ b/src/libopensc/card-cedulauy.c
@@ -1,0 +1,299 @@
+/*
+ * card-cedulauy.c: Support for the Uruguayan eID card (cédula de identidad)
+ *
+ * Copyright (C) 2026 Carlos Andrés Planchón Prestes <carlosandresplanchonprestes@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * The card is a Gemalto/Thales "Classic V4" IAS/ECC platform card, but the
+ * subset it exposes over the contact interface is plain ISO 7816, so the
+ * driver is built on the generic iso7816 operations.  The IAS application
+ * must be selected by AID before anything on the card is accessible.
+ *
+ * The card conventions come from the public documentation and reference code
+ * published by AGESIC, Uruguay's national e-government agency ("Documentación
+ * técnica de la cédula de identidad con chip" and
+ * https://github.com/eIDuy/apdu-services).
+ *
+ * The PKCS#15 view of the card is provided by the synthetic emulator in
+ * pkcs15-cedulauy.c.
+ */
+
+#include "libopensc/errors.h"
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "internal.h"
+
+/* MSE SET algorithm references (AGESIC "AlgoID" table): the hash is encoded
+ * in the high nibble and the padding scheme in the low nibble. */
+#define CEDULAUY_ALGO_HASH_NONE	  0x00 /* DigestInfo built in software */
+#define CEDULAUY_ALGO_HASH_SHA256 0x40 /* card builds the DigestInfo */
+#define CEDULAUY_ALGO_PAD_PKCS1	  0x02 /* PKCS#1 v1.5 */
+
+#define CEDULAUY_ALGO_RSA_PKCS1	       (CEDULAUY_ALGO_HASH_NONE | CEDULAUY_ALGO_PAD_PKCS1)
+#define CEDULAUY_ALGO_RSA_PKCS1_SHA256 (CEDULAUY_ALGO_HASH_SHA256 | CEDULAUY_ALGO_PAD_PKCS1)
+
+// clang-format off
+static const struct sc_atr_table cedulauy_atrs[] = {
+	/* TA1 and the version/batch historical bytes differ between card
+	 * batches and are masked out (mask contributed by Nicolás Gutiérrez,
+	 * @nicolasgutierrezdev). */
+	{ "3B:7F:94:00:00:80:31:80:65:B0:85:03:00:EF:12:0F:FF:82:90:00",
+	  "FF:FF:00:FF:FF:FF:FF:FF:FF:FF:FF:00:00:00:00:00:00:FF:FF:FF",
+	  "Uruguayan eID (cedula de identidad)", SC_CARD_TYPE_CEDULAUY, 0, NULL },
+	{ NULL, NULL, NULL, 0, 0, NULL }
+};
+// clang-format on
+
+/* IAS application AID (AGESIC, documented) */
+static const unsigned char cedulauy_aid[] = {
+		0xA0, 0x00, 0x00, 0x00, 0x18, 0x40, 0x00, 0x00, 0x01, 0x63, 0x42, 0x00};
+
+static const struct sc_card_operations *iso_ops = NULL;
+static struct sc_card_operations cedulauy_ops;
+
+static struct sc_card_driver cedulauy_drv = {
+		"Uruguayan eID (cedula de identidad)",
+		"cedulauy",
+		&cedulauy_ops,
+		NULL, 0, NULL};
+
+#define SC_TRANSMIT_TEST_RET(card, apdu, text) \
+	do { \
+		LOG_TEST_RET(card->ctx, sc_transmit_apdu(card, &apdu), "APDU transmit failed"); \
+		LOG_TEST_RET(card->ctx, sc_check_sw(card, apdu.sw1, apdu.sw2), text); \
+	} while (0)
+
+static int
+cedulauy_select_app(struct sc_card *card)
+{
+	struct sc_apdu apdu;
+	unsigned char resp[SC_MAX_APDU_RESP_SIZE];
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	sc_format_apdu_ex(&apdu, card->cla, 0xA4, 0x04, 0x00,
+			cedulauy_aid, sizeof cedulauy_aid, resp, sizeof resp);
+	SC_TRANSMIT_TEST_RET(card, apdu, "Cannot select the eID application");
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+cedulauy_match_card(struct sc_card *card)
+{
+	int i = _sc_match_atr(card, cedulauy_atrs, &card->type);
+
+	if (i < 0)
+		return 0;
+	card->name = cedulauy_atrs[i].name;
+	return 1;
+}
+
+static int
+cedulauy_init(struct sc_card *card)
+{
+	unsigned long flags;
+	int r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	card->caps = SC_CARD_CAP_RNG;
+
+	r = cedulauy_select_app(card);
+	LOG_TEST_RET(card->ctx, r, "Cannot select the eID application");
+
+	/* The card builds the DigestInfo on-card only for SHA-256 (algorithm
+	 * reference 0x42).  Everything else is signed as raw PKCS#1 v1.5 over
+	 * a DigestInfo built in software (algorithm reference 0x02). */
+	flags = SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_HASH_SHA256;
+	r = _sc_card_add_rsa_alg(card, 2048, flags, 0);
+
+	LOG_FUNC_RETURN(card->ctx, r);
+}
+
+static int
+cedulauy_card_reader_lock_obtained(struct sc_card *card, int was_reset)
+{
+	if (was_reset)
+		LOG_FUNC_RETURN(card->ctx, cedulauy_select_app(card));
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static const sc_file_t *
+cedulauy_get_mf()
+{
+	static sc_file_t *mf = NULL;
+	if (!mf) {
+		mf = sc_file_new();
+		if (mf) {
+			mf->path = *sc_get_mf_path();
+			mf->id = 0x3F00;
+			mf->type = SC_FILE_TYPE_DF;
+			mf->magic = SC_FILE_MAGIC;
+		}
+	}
+	return mf;
+}
+
+static int
+cedulauy_select_file(struct sc_card *card, const struct sc_path *in_path,
+		struct sc_file **file_out)
+{
+	struct sc_path path = *in_path;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	if (path.aid.len == sizeof cedulauy_aid && 0 == memcmp(path.aid.value, cedulauy_aid, sizeof cedulauy_aid)) {
+		/* swallow cedula AID, it is always selected via init() */
+		path.aid.len = 0;
+	}
+	if (path.type == SC_PATH_TYPE_PATH && path.len >= 2 && path.value[0] == 0x3F && path.value[1] == 0x00) {
+		/* swallow MF, everything resides within the Application DF */
+		memmove(path.value, path.value + 2, path.len - 2);
+		path.len -= 2;
+	}
+	if (path.type == SC_PATH_TYPE_PATH && path.len == 0) {
+		/* Selection of MF was requested */
+		sc_file_dup(file_out, cedulauy_get_mf());
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+	}
+
+	LOG_FUNC_RETURN(card->ctx, iso_ops->select_file(card, &path, file_out));
+}
+
+static int
+cedulauy_set_security_env(struct sc_card *card, const struct sc_security_env *env,
+		int se_num)
+{
+	struct sc_apdu apdu;
+	/* MSE SET, Digital Signature Template: key reference and algorithm */
+	unsigned char mse_data[] = {0x84, 0x01, 0xFF, 0x80, 0x01, CEDULAUY_ALGO_RSA_PKCS1};
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	if (env == NULL)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
+	if (env->operation != SC_SEC_OPERATION_SIGN)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+	if ((env->flags & SC_SEC_ENV_ALG_PRESENT) && env->algorithm != SC_ALGORITHM_RSA)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+	if (env->key_ref_len != 1)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+
+	mse_data[2] = env->key_ref[0];
+	if (env->algorithm_flags & SC_ALGORITHM_RSA_HASH_SHA256)
+		mse_data[5] = CEDULAUY_ALGO_RSA_PKCS1_SHA256;
+
+	sc_format_apdu_ex(&apdu, card->cla, 0x22, 0x41, 0xB6,
+			mse_data, sizeof mse_data, NULL, 0);
+	SC_TRANSMIT_TEST_RET(card, apdu, "MSE SET DST failed");
+
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+static int
+cedulauy_compute_signature(struct sc_card *card, const u8 *data, size_t datalen,
+		u8 *out, size_t outlen)
+{
+	struct sc_apdu apdu;
+	unsigned char sbuf[64];
+	unsigned char rbuf[256]; /* RSA 2048 */
+	size_t offs = 0;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	if (data == NULL || out == NULL)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_INVALID_ARGUMENTS);
+
+	/* What is loaded here matches the algorithm set in set_security_env:
+	 * for 0x42 'data' is the bare SHA-256 digest and the card builds the
+	 * DigestInfo; for 0x02 'data' is a complete DigestInfo built in
+	 * software, which the card pads (PKCS#1 v1.5) as-is. */
+	if (datalen == 0 || datalen > sizeof(sbuf) - 2)
+		LOG_TEST_RET(card->ctx, SC_ERROR_NOT_SUPPORTED,
+				"Unsupported hash/DigestInfo length");
+
+	/* PSO HASH: load the digest */
+	sbuf[offs++] = 0x90;
+	sbuf[offs++] = (unsigned char)datalen;
+	memcpy(sbuf + offs, data, datalen);
+	offs += datalen;
+	sc_format_apdu_ex(&apdu, card->cla, 0x2A, 0x90, 0xA0, sbuf, offs, NULL, 0);
+	SC_TRANSMIT_TEST_RET(card, apdu, "PSO HASH failed");
+
+	/* PSO COMPUTE DIGITAL SIGNATURE */
+	sc_format_apdu_ex(&apdu, card->cla, 0x2A, 0x9E, 0x9A, NULL, 0, rbuf, sizeof rbuf);
+	SC_TRANSMIT_TEST_RET(card, apdu, "PSO COMPUTE DIGITAL SIGNATURE failed");
+
+	if (apdu.resplen > outlen)
+		LOG_TEST_RET(card->ctx, SC_ERROR_BUFFER_TOO_SMALL,
+				"Signature buffer too small");
+	memcpy(out, apdu.resp, apdu.resplen);
+	LOG_FUNC_RETURN(card->ctx, (int)apdu.resplen);
+}
+
+static int
+cedulauy_get_challenge(struct sc_card *card, u8 *rnd, size_t len)
+{
+	/* As on other IAS/ECC cards, GET CHALLENGE only handles a length of 8 */
+	unsigned char rbuf[8];
+	int r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	r = iso_ops->get_challenge(card, rbuf, sizeof rbuf);
+	LOG_TEST_RET(card->ctx, r, "GET CHALLENGE failed");
+
+	if (len < (size_t)r)
+		r = (int)len;
+	memcpy(rnd, rbuf, (size_t)r);
+
+	LOG_FUNC_RETURN(card->ctx, r);
+}
+
+static int
+cedulauy_logout(struct sc_card *card)
+{
+	/* Re-selecting the application resets its security status. */
+	LOG_FUNC_RETURN(card->ctx, cedulauy_select_app(card));
+}
+
+struct sc_card_driver *
+sc_get_cedulauy_driver(void)
+{
+	struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
+
+	if (iso_ops == NULL)
+		iso_ops = iso_drv->ops;
+
+	cedulauy_ops = *iso_ops;
+	cedulauy_ops.match_card = cedulauy_match_card;
+	cedulauy_ops.init = cedulauy_init;
+	cedulauy_ops.select_file = cedulauy_select_file;
+	cedulauy_ops.set_security_env = cedulauy_set_security_env;
+	cedulauy_ops.compute_signature = cedulauy_compute_signature;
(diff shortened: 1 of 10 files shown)


src/libopensc/card-iasecc.c
```diff
@@ -430,8 +434,28 @@ iasecc_init_gemalto(struct sc_card *card)
...
+		 * 0x42). Every other hash is signed as raw PKCS#1 v1.5 over a DigestInfo
+		 * built in software (MSE algo 0x02). It has no on-card SHA-1. */
+		flags = SC_ALGORITHM_RSA_PAD_PKCS1
+			| SC_ALGORITHM_RSA_HASH_NONE
+			| SC_ALGORITHM_RSA_HASH_SHA256;
+
 	card->caps = IASECC_CARD_DEFAULT_CAPS;
 
+	/* Uruguay cedula (Gemalto Classic V4): the IAS-ECC application must be
+	 * selected by AID before the MF/EF.ATR are accessible. */
```
frankmorgner: Is a *single* select AID enough to make the card usable or is it *always* required in order to select MF/EF.ATR. The former indicates, that the select should be placed into the `..._init()` function of that card *and* that you need to implement card_reader_lock_obtained to reselect that applet in case of a context switch from a different application

src/libopensc/card-iasecc.c
```diff
@@ -450,6 +474,12 @@ iasecc_init_gemalto(struct sc_card *card)
 		sc_log(ctx, "rv %i", rv);
 	}
 	sc_log(ctx, "rv %i", rv);
+	if (rv == SC_ERROR_FILE_NOT_FOUND && card->type == SC_CARD_TYPE_IASECC_GEMALTO_UY) {
+		/* Uruguay cedula: no EF.ATR file (card version is exposed via GET DATA,
+		 * not as a file); proceed with default capabilities. */
+		sc_log(ctx, "No EF.ATR; continuing with default capabilities");
+		rv = SC_SUCCESS;
```
frankmorgner: I think you should write a dedicated init() function of your card type to avoid this complication

src/libopensc/card-iasecc.c
```diff
@@ -974,14 +1005,21 @@ iasecc_select_file(struct sc_card *card, const struct sc_path *path,
 			    )   {
 				apdu.p2 = 0x04;
 			}
+			if (card->type == SC_CARD_TYPE_IASECC_GEMALTO_UY)   {
+				/* Uruguay cedula selects an EF by file ID with P1=00 P2=00 */
+				apdu.p1 = 0x00;
+				apdu.p2 = 0x00;
+			}
```
frankmorgner: you have this already set to 0x00 0x00 in line 951. why do you reset this here?

src/libopensc/card-iasecc.c
```diff
@@ -2230,6 +2295,21 @@ iasecc_pin_merge_policy(struct sc_card *card, struct sc_pin_cmd_data *data,
...
+		 * Use a fixed policy: plain ISO VERIFY (no Secure Messaging) over the contact
+		 * interface. PIN padding (12-byte, zero-filled) comes from the PKCS#15 PIN
+		 * object, so no policy-driven padding is needed here. */
+		memset(policy, 0, sizeof(*policy));
+		policy->min_length = -1;
+		policy->max_length = -1;
+		policy->tries_maximum = -1;
+		policy->tries_remaining = -1;
+		policy->stored_length = pin->pad_length ? (int) pin->pad_length : 12;
+		LOG_FUNC_RETURN(ctx, SC_SUCCESS);
```
frankmorgner: If you are defining a PKCS#15 emulator that includes pin policies, why do you need to adjust those again here?

frankmorgner (changes requested):
card-iasecc.c in general and your changes in particular are looking more and more like spaghetti code with many of exceptions from the supposidly generic definition of IAS/ECC.

Since your agent even includes a PKCS#15 emulator, I am wondering how much overlap you actually have with IAS/ECC specification and if it would make sense to define a dedicated card driver independent from card-iasec.c

frankmorgner:
Documentación técnica de la cédula de identidad con chip only has a handful of commands, that could be well represented in a dedicated card driver. Everything looks pretty standard (i.e. could be based on iso7816.c)

what could be useful for other people would be to provide the image data and the ID data with either a dedicated tool for the card or to provide it as generic PKCS#15 data object.

nicolasgutierrezdev:
While doing some testing I realized the ATR in the card type definition, won't match all uruguayan id cards. For instance:

My ATR:  3B 7F 96 00 00 80 31 80 65 B0 85 05 00 11 12 0F FF 82 90 00

PRs:  3B 7F 94 00 00 80 31 80 65 B0 85 03 00 EF 12 0F FF 82 90 00

There's a diff in 3 bytes. byte 2 = TA1 (transmission speed per the iso7816 spec) 
bytes 11/13 = Gemalto/Batch version I guess.

There is more than one batch of uruguayan id, which have small differences in firmware/card specs. But they all work under the same interface.

Fix: apply a mask over the TA1 and some historical bytes.
FF:FF:00:FF:FF:FF:FF:FF:FF:FF:FF:00:00:00:00:00:00:FF:FF:FF

nicolasgutierrezdev:
> Documentación técnica de la cédula de identidad con chip only has a handful of commands, that could be well represented in a dedicated card driver. Everything looks pretty standard (i.e. could be based on iso7816.c)
> 
> what could be useful for other people would be to provide the image data and the ID data with either a dedicated tool for the card or to provide it as generic PKCS#15 data object.

It is iso7816, per the spec also IAS. But it has a lot of quirks per the manufacturer.

nicolasgutierrezdev:
Tested this PR against a physical Uruguayan eID (cédula) and everything works end to end. The new `card-cedulauy.c` module approach works well.

### Environment

- **Card driver:** `Uruguayan eID` (from this PR)
- **ATR:** `3b:7f:96:00:00:80:31:80:65:b0:85:05:00:11:12:0f:ff:82:90:00`

### 1. Reading the token (PKCS#11)

The certificate is exposed correctly through the module:

```
ID:                01
Subject:           [redacted]
Issuer:            Autoridad Certificadora del Ministerio del Interior
Serial:            [redacted]
Valid until:       [redacted]
Digital signature: yes
```

### 2. Signing

Signed a `test.pdf` as PAdES using the on-card key:

```
Signer:              [redacted]
Issuer:              Autoridad Certificadora del Ministerio del Interior
PKCS#11 ID:          01
Certificate serial:  [redacted]
Output:              test_firmado.pdf (PAdES)
```

### 3. Verifying the signed PDF

```
Signer:  [redacted], DNI [redacted]
Issuer:  Autoridad Certificadora del Ministerio del Interior

  ✓ signature intact (covered bytes unmodified)
  ✓ signature cryptographically valid
  ✓ coverage: whole file (ENTIRE_FILE)
  ✓ certificate chain to trusted root

Indication: VALID
```

### Result

The driver behaves correctly against real hardware.

carlosplanchon:
@nicolasgutierrezdev, thx. Here:

```
❯ ./src/tools/pkcs11-tool --module ./src/pkcs11/.libs/opensc-pkcs11.so --test --login
Logging in to "CARLOS ANDRÉS PLANCHÓN PRESTES".
Please enter User PIN: 
  seems to be OK
Digests:
  MD5: OK
  RIPEMD160: OK
  SHA-1: OK
  SHA256: OK
Ciphers: not implemented
  testing key 0 (Clave de Firma) 
  testing signature mechanisms:
    RSA-PKCS: OK
    SHA256-RSA-PKCS: OK
  testing key 0 (Clave de Firma)
    RSA-PKCS: OK
  testing key 0 (Clave de Firma) -- can't be used to decrypt, skipping
No errors
```

frankmorgner:
the separate driver looks simple enough and avoids complications with the existing code, thanks.

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,320 @@
...
+static int
+cedulauy_init(struct sc_card *card)
+{
+	unsigned long flags;
+	int r;
+
+	LOG_FUNC_CALLED(card->ctx);
+
+	card->cla = 0x00;
+	card->caps |= SC_CARD_CAP_RNG;
```
frankmorgner:
```suggestion
	card->caps = SC_CARD_CAP_RNG;
```

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,320 @@
...
+}
+
+
+static int
+cedulauy_card_reader_lock_obtained(struct sc_card *card, int was_reset)
+{
+	/* The card may have been reset, or another application may have
+	 * selected a different applet in the meantime: re-select the eID
+	 * application, without which nothing on the card is accessible. */
+	LOG_FUNC_RETURN(card->ctx, cedulauy_select_app(card));
```
frankmorgner:
```suggestion
	if (was_reset)
		LOG_FUNC_RETURN(card->ctx, cedulauy_select_app(card));
```

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,320 @@
...
+}
+
+
+static int
+cedulauy_select_file(struct sc_card *card, const struct sc_path *in_path,
```
frankmorgner, line 157: I suggest you limit yourself to
- mapping `3F00` ->`cedulauy_aid` (also in the absolute path case)
- forbidding selection of a different AID (to avoid context loss)
Since your PKCS#15 emulator is adapted to those restrictions, this functionality should work.

All other variations of SELECT can be directly *verified by the card's applet* itself (e.g. when using opensc-explorer). This simplifies your code and allows variations in future versions of your card.

frankmorgner (changes requested):
please reduce the verbosity of comments. repeated information is not necessary and if the code is clear about what is done, then you don't need a comment repeating this.

If your card has a static profile and does *not* support re-issuing a certificate, for example, enable file caching for your card in pkcs15_get_default_use_file_cache

have you thought about the suggestion regarding ID data and face image?

carlosplanchon:
Hello!, thanks a lot for your patience @frankmorgner. Regarding the ID data and face image: yes, my idea was to keep this PR limited to the signature/authentication path, and expose those later as generic PKCS#15 data objects in a follow-up PR.

The elementary files and formats are documented by AGESIC, so that part should be doable separately.

I’m still pretty new to this area of OpenSC though, so if you prefer to include them in this PR, I'm happy to adjust.

src/libopensc/card-cedulauy.c
carlosplanchon: Done

src/libopensc/card-cedulauy.c
carlosplanchon: Done.

src/libopensc/card-cedulauy.c
carlosplanchon: Done.

select_file is now limited to the two rules you suggested. It maps the MF to the eID application, including when the MF is the prefix of an absolute path. And it refuses to select a different application, which would lose the application context. Everything else goes to the card unmodified.

carlosplanchon:
Along the same lines, the emulator no longer probes the application with an extra SELECT
 
One detail this surfaced, the PKCS#15 emulator did not set a token serial number, which generate_cache_filename() requires, so the cache never engaged. The emulator now derives the token serial from the signing certificate's serial, and the cache works (verified the round-trip against my cédula de identidad).
 
All changes re-tested end to end against my physical card. Recognition, certificate read (including the cache), PIN verification and signing. pkcs11-tool --test --login reports no errors.

@nicolasgutierrezdev if you can test it again, that'd be great :)

frankmorgner:
Please note that my suggestions are open for discussion. I just try to point out what experiences we had in the past.

I had a second thought about the MF path conversion... In card-sc-hsm.c, for example, path.aid (or SC_PATH_TYPE_DF_NAME) IS actually evaluated via SELECT AID, then `3F00` in `path->value` is swallowed to simulate MF selection. A card's app may evaluate SELECT AID if the javacard OS doesn't intercept it, this approach would allow this flexibility.

In your PKCS#15 emulator you are attaching an AID to the path of the certificates, which (when actually performed with SELECT AID) may lead to context loss of the PIN verification status, which is typically not what is initended. Maybe it would be better to avoid the AID in the path, and just rely on an absolute path (`3F00 XX XX`) given the SELECT logic of the previous paragraph.

Did you think about the option of using the NumeroDeDocumento as `card->serialnr` in the card driver which would then also be available in the PKCS#15 emulator as serial number (and via `SC_CARDCTL_GET_SERIALNR`). This would reduce the data transferred between card and terminal and parsing of the data would be simpler compared to the certificate parsing. Additionally, the user could easier identify her document with the number printed on the document.

Does your document support NFC? The documentation states that cards as of 2022 are supporting PIN verification via PACE. This would be available in OpenSC via `perform_pace()`.

nicolasgutierrezdev:
My document does support NFC, but my iso7816 reader does not.

carlosplanchon:
My document doesn't support NFC :/

Pixeluted:
alternative PR title: add 1984 support for urugayns, made and proudly coded by — claude opus.

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,309 @@
...
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+
+static int
+cedulauy_compute_signature(struct sc_card *card, const u8 *data, size_t datalen,
+		u8 *out, size_t outlen)
+{
+	struct sc_apdu apdu;
+	unsigned char sbuf[64];
```
Pixeluted, line 217: Leaving a uninitialized variable is not a good idea, especically when you will use that memory down the line for formatting which could lead to undefined behaviour.

Reviewed by — Claude Opus, definitively not a human.

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,309 @@
...
+	unsigned char rbuf[256];	/* RSA 2048 */
+	size_t offs = 0;
+
+	LOG_FUNC_CALLED(card->ctx);
```
Pixeluted, line 221: card could be a nullptr or a invalid pointer causing a crash

Reviewed by — Claude Opus, definitively not a human.

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,309 @@
...
+		LOG_FUNC_RETURN(card->ctx, cedulauy_select_app(card));
+	LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+}
+
+
+static int
+cedulauy_select_file(struct sc_card *card, const struct sc_path *in_path,
+		struct sc_file **file_out)
+{
+	struct sc_path path = *in_path;
```
Pixeluted, line 160: Not only you are dereferencing a pointer without checking if it's invalid but also you're causing useless copy for the sc_path structure.

Reviewed by — Claude Opus, definitively not a human.

src/libopensc/pkcs15-cedulauy.c
```diff
@@ -0,0 +1,171 @@
...
+
+#define CEDULAUY_CERT_FID	"B001"	/* signing certificate EF */
+#define CEDULAUY_KEY_REF	0x01	/* signing private key reference */
+#define CEDULAUY_PIN_REF	0x11	/* Global PIN reference */
+#define CEDULAUY_OBJ_ID		0x01	/* links cert <-> prkey */
+
+static int
+sc_pkcs15emu_cedulauy_init(struct sc_pkcs15_card *p15card)
+{
+	struct sc_context *ctx = p15card->card->ctx;
```
Pixeluted, line 56: `p15card` or `card` can be a nullptr, and you're dereferencing it without any checks.

Reviewed by — Claude Opus, definitively not a human.

src/libopensc/card-cedulauy.c
Pixeluted: > Done.
> 
> select_file is now limited to the two rules you suggested. It maps the MF to the eID application, including when the MF is the prefix of an absolute path. And it refuses to select a different application, which would lose the application context. Everything else goes to the card unmodified.

Done — Coded by Claude Opus

src/libopensc/pkcs15-cedulauy.c
carlosplanchon: https://github.com/OpenSC/OpenSC/blob/727fa9caaea9f8926da2b57ee3fb556e1d9f4ce4/src/libopensc/pkcs15.c#L1279

carlosplanchon:
@Pixeluted, I understand your concern about the use of generative AI. That's why I disclosed Claude assistance explicitly in the PR description. That said, the code still has to stand on its own, documented card behavior, real hardware testing, maintainer review, and whatever changes are needed to fit into OpenSC.

Pixeluted:
> @Pixeluted, I understand your concern about the use of generative AI. That's why I disclosed Claude assistance explicitly in the PR description. That said, the code still has to stand on its own, documented card behavior, real hardware testing, maintainer review, and whatever changes are needed to fit into OpenSC.

"Claude assistance", you mean fully made by — Claude right?

frankmorgner:
Please stay objective. The requested changes are simple enough so that AI tools are a good way forward. Please have a look at the two commits I have pushed on top, primarily to handle the path and check if everything works as expected.

carlosplanchon:
One consideration for that follow-up. The cédula number is a lifelong personal identifier in Uruguay, and the token serial may end up in cache file names, debug logs and pasted bug reports. It is already present in the certificate subject, so it is not secret, but the current opaque serial leaks less by default. Worth weighing when we get there.

carlosplanchon:
Tested your two commits against the card. Found one crash. With a warm file cache, sc_pkcs15_read_file() re-selects the parent of the certificate path with file_out == NULL, which reaches the new emulated master file branch, and sc_file_dup() dereferences the NULL destination (sc.c line 651). Pushed a small guard on top. pkcs11-tool --test --login reports no errors.

@nicolasgutierrezdev, try this with your cédula please.

nicolasgutierrezdev:
I tested the new changes against my hardware. Here are the results:

* Reader model: Alcor Link AK9563 (contact reader) (no NFC).
* OS: Arch Linux.
* OpenSC commit tested: 387b588 "In cedulauy, guard the emulated MF against NULL file_out"
* Card issue date: March 2026.
* opensc-tool -l sees the reader: Alcor Link AK9563 00 00.
* opensc-tool -a / pkcs15-tool -D return the following:
  - opensc-tool -n → Uruguayan eID (cedula de identidad).
  - opensc-tool -a → ATR 3b:7f:96:...:90:00.
  - pkcs15-tool -D → exposes PIN + RSA-2048 key + signing certificate. 
* pkcs11-tool --test --login reports no errors

nicolasgutierrezdev:
Regarding NFC, I tested with an ACR122U, but I could not get the official driver working with NFC. I made some experiments with OpenSC (this PR) and got the following findings:

* PACEInfo: id-PACE-ECDH-GM-AES-CBC-CMAC-256 / NIST P-384
* Card accepts PACE-MRZ only. CAN reference (83 01 02) → 6A88; MRZ (83 01 01) → 9000. 
* Contactless ATR differs from contact ATR, so current mask won't work. 
* Contactless ATR is dynamic (bytes 5–8 are randomized per tap), this must be taken into account to do proper masking. 
* Got to work: PACE, PIN login, digests, cert/pubkey readout, MSE: SET, PSO: HASH.
* Failure: retrieving the 256-byte RSA signature over SM. Wrapped reply is 291 B; OpenSC's short-APDU SM buffer capped it at 256.
* Stale-SM quirk: after a successful PACE the card keeps SM until reset; sc_reset() does not clear it on the ACR122U. Workaround is re-tapping the card.

frankmorgner (approved):
Changes look good for me, now.

frankmorgner (changes requested):
Just one more thing: **Instead** of adding a guard against `NULL` dereference, please make `sc_file_dup()` more robust (387b5889fde35c1f4ccd05fac7c1c4e12c1e7dc9) by adding the following patch:
```diff
diff --git a/src/libopensc/sc.c b/src/libopensc/sc.c
index 075259be7..6882cba0f 100644
--- a/src/libopensc/sc.c
+++ b/src/libopensc/sc.c
@@ -648,6 +648,8 @@ void sc_file_dup(sc_file_t **dest, const sc_file_t *src)
        const sc_acl_entry_t *e;
        unsigned int op;
 
+       if (!dest)
+               return;
        *dest = NULL;
        if (!sc_file_valid(src))
                return;
```

frankmorgner:
Regarding your contactless testing:

- you could use the different ATR on NFC to enforce PIN verification via PACE, because over cleartext NFC i suspect that clear text PIN verification is disabled
- to allow an extended response of 291 bytes, just expand the response buffer to something bigger than 256, e.g. 291. This should send an extended SM APDU with enough space for the response. please check your documentation whether extended length is supported or whether you need to employ some kind of chaining.
- stale sm: just run `sc_sm_stop()` to terminate SM

frankmorgner (approved):

carlosplanchon:
Thanks, I moved the guard into sc_file_dup() as suggested. pkcs11-tool --test --login passes, and the contact path still works end to end.

@nicolasgutierrezdev I think the NFC/contactless work should stay separate from this PR, so we don't expand the scope right before landing contact support.

A good next step would be to open a separate issue to help collect those findings.

src/libopensc/card-cedulauy.c
```diff
@@ -0,0 +1,308 @@
...
+#include <string.h>
+
+#include "internal.h"
+
+/* MSE SET algorithm references (AGESIC "AlgoID" table): the hash is encoded
+ * in the high nibble (0 none, 4 SHA-256, 5 SHA-384, 6 SHA-512) and the
+ * padding in the low nibble (1 ISO 9796-2, 2 PKCS#1 v1.5, 5 PSS). */
+#define CEDULAUY_ALGO_RSA_PKCS1		0x02	/* no on-card hash */
+#define CEDULAUY_ALGO_RSA_PKCS1_SHA256	0x42	/* card builds the DigestInfo */
+
```
Jakuje, line 52: nit: should we construct these from the intermediate constants rather than keeping the construction only in the comment?

src/libopensc/pkcs15-cedulauy.c
```diff
@@ -0,0 +1,180 @@
...
+	if (appinfo == NULL) {
+		LOG_FUNC_RETURN(ctx, SC_ERROR_OUT_OF_MEMORY);
+	}
+	appinfo->aid = aid;
+	appinfo->ddo.aid = aid;
+	p15card->app = appinfo;
+
+	/* Global PIN (reference 0x11, ASCII numeric, zero-padded to 12 bytes). */
+	memset(&pin_info, 0, sizeof pin_info);
+	memset(&pin_obj, 0, sizeof pin_obj);
```
Jakuje: Should we intialize the variables with ` = {0}` instead of calling explicit memset on all of the structures?

src/libopensc/pkcs15-cedulauy.c
```diff
@@ -0,0 +1,180 @@
...
+	cert_info.id.value[0] = CEDULAUY_OBJ_ID;
+	cert_info.id.len = 1;
+	strlcpy(cert_obj.label, "Certificado de Firma", sizeof cert_obj.label);
+	r = sc_pkcs15emu_add_x509_cert(p15card, &cert_obj, &cert_info);
+	LOG_TEST_RET(ctx, r, "Cannot add signing certificate object");
+
+	/* Set the token label and serial number from the certificate, when
+	 * readable.  The serial number also keys the file cache. */
+	{
+		struct sc_pkcs15_cert *cert = NULL;
```
Jakuje: why to introduce nested block like this? Can the declaration be moved to the top of the block and the rest de-indented?

carlosplanchon: Done

src/libopensc/pkcs15-cedulauy.c
carlosplanchon: Done

src/libopensc/card-cedulauy.c
carlosplanchon: Done, built from the nibble constants now. Thanks 👍🏻

carlosplanchon:
CI: The test-interoperability failure appears unrelated to this PR. It fails inside SoftHSM during C_WrapKey.

Jakuje (approved):
I am rerunning the timed out jobs, but otherwise lgtm

Jakuje:
@frankmorgner any more comments or good to merge?

Last thing, I would prefer to have the fixup commits squashed into self-contained commits.

carlosplanchon:
Done. Same code, tidy history. Thanks.

carlosplanchon:
Also gave @nicolasgutierrezdev co-author credit on the driver commit, for the ATR mask and the testing on a second card batch. Same code.

PR #3732, opened 2026-07-01
card-piv.c - Fix PIV SM issue when Lc is multiple of 16

dengert:
Fixes #3727 

 On branch piv-sm-lc-fix
 Changes to be committed:

Thanks to @frankmorgner for the fix and @laughingbro from the report

Idemia ID-One PIV 2.4.1 cards were used to test the fix.

There is still a related issue:
sm-iso.c and the current card-piv.c do not treat Lc=0 as a multiple of 16, but NIST sp800-73-5 says: 

" BER-TLV-encoded encrypted message, which consists of tag '87' followed by the length of the encoded encrypted message (Lcc + 1), the padding indicator byte ('01'), and then the encrypted data. Lcc is the length of the encrypted PIV data; it SHALL be a multiple of 16." 

The Idemia ID-One PIV 2.4.1 cards work with or without an '87' for Lc=0. 

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -1044,7 +1044,7 @@ piv_encode_apdu(sc_card_t *card, sc_apdu_t *plain, sc_apdu_t *sm_apdu)
 		T87len = 0;
 		padlen = 0;
 	} else {
-		enc_datalen = (int)(((plain->datalen + 15) / 16) * 16); /* may add extra 16 bytes */
+		enc_datalen = (int)(((plain->datalen + 16) / 16) * 16); /* may add extra 16 bytes */
 		padlen = enc_datalen - (int)plain->datalen;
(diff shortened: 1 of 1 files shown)


dengert:
Google AI  responded to "In PIV SM is '87' needed if Lc=0":

According to the [NIST SP 800-73-4 / 800-73-5 Part 2 specifications](https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-73-4.pdf), tag '87' represents a BER-TLV encoded encrypted message payload. If a command contains zero plaintext bytes to encrypt, the cryptographic payload wrapper is completely omitted from the newly transformed SM data field.

Although I can not find that, it does follow ISO 7816  SM and OpenSC 'sm-iso.c'  follow that.

frankmorgner (approved):

Jakuje (approved):
good catch!

PR #3734, opened 2026-07-03
pkcs15/iasecc: tolerate missing Oberthur SDO metadata

henriquefurlan:
Handle incomplete IAS/ECC metadata on IAS/ECC v1.0.1 Oberthur cards.

Some IAS/ECC v1.0.1 Oberthur cards return `SC_ERROR_DATA_OBJECT_NOT_FOUND` for
the PIN policy SDO and for the RSA private-key SDO/DOCP. In addition, this card
type was not routed through the synthetic PKCS#15 IAS/ECC emulation path, so the
existing PRKDF fixup was never applied.

This patch fixes the issue in three places:

- add `SC_CARD_TYPE_IASECC_OBERTHUR` to the emulation-only path in
  `src/libopensc/pkcs15-syn.c`
- add `SC_CARD_TYPE_IASECC_OBERTHUR` to the existing PRKDF fixup switch in
  `src/libopensc/pkcs15-iasecc.c`
- in `src/libopensc/card-iasecc.c`, fall back when the card returns
  `SC_ERROR_DATA_OBJECT_NOT_FOUND` while reading PIN policy and RSA private-key
  SDO metadata

For the PIN policy fallback, the patch uses the PKCS#15 PIN description to
populate the policy fields. For the RSA private-key fallback, it uses a minimal
security environment so signing can proceed when the private-key DOCP is absent.

Tested card:
- IAS/ECC v1.0.1 Oberthur

Validation:
- PKCS#15 enumeration succeeds
- PKCS#11 login succeeds
- PKCS#11 signing succeeds with `RSA-PKCS`
- the generated signature was verified against the certificate public key
- Firefox certificate usage was tested successfully on macOS

##### Checklist
- [x] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-iasecc.c
+++ b/src/libopensc/pkcs15-iasecc.c
@@ -162,6 +162,7 @@ _iasecc_parse_df(struct sc_pkcs15_card *p15card, struct sc_pkcs15_df *df)
 		case SC_CARD_TYPE_IASECC_GEMALTO:
 		case SC_CARD_TYPE_IASECC_CPX:
 		case SC_CARD_TYPE_IASECC_CPXCL:
+		case SC_CARD_TYPE_IASECC_OBERTHUR:
 			sc_log(ctx, "Warning: the %d card has an invalid DF, hot patch to be applied",
 				p15card->card->type);
 			break;
--- a/src/libopensc/pkcs15-syn.c
+++ b/src/libopensc/pkcs15-syn.c
@@ -103,6 +103,7 @@ int sc_pkcs15_is_emulation_only(sc_card_t *card)
 		case SC_CARD_TYPE_IASECC_GEMALTO:
 		case SC_CARD_TYPE_IASECC_CPX:
 		case SC_CARD_TYPE_IASECC_CPXCL:
+		case SC_CARD_TYPE_IASECC_OBERTHUR:
 		case SC_CARD_TYPE_PIV_II_GENERIC:
 		case SC_CARD_TYPE_PIV_II_HIST:
 		case SC_CARD_TYPE_PIV_II_NEO:
@@ -542,4 +543,3 @@ int sc_pkcs15emu_object_add(sc_pkcs15_card_t *p15card, unsigned int type,
 
 	LOG_FUNC_RETURN(p15card->card->ctx, SC_SUCCESS);
 }
-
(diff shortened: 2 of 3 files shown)


src/libopensc/pkcs15-iasecc.c
```diff
@@ -162,6 +162,7 @@ _iasecc_parse_df(struct sc_pkcs15_card *p15card, struct sc_pkcs15_df *df)
 		case SC_CARD_TYPE_IASECC_GEMALTO:
 		case SC_CARD_TYPE_IASECC_CPX:
 		case SC_CARD_TYPE_IASECC_CPXCL:
+		case SC_CARD_TYPE_IASECC_OBERTHUR:
```
frankmorgner, line 165: You are changing the parsing of all cards that were previously identified as `SC_CARD_TYPE_IASECC_OBERTHUR`. I am wondering if this breaks backward compatibility with those cards...

src/libopensc/pkcs15-syn.c
```diff
@@ -103,6 +103,7 @@ int sc_pkcs15_is_emulation_only(sc_card_t *card)
 		case SC_CARD_TYPE_IASECC_GEMALTO:
 		case SC_CARD_TYPE_IASECC_CPX:
 		case SC_CARD_TYPE_IASECC_CPXCL:
+		case SC_CARD_TYPE_IASECC_OBERTHUR:
```
frankmorgner, line 106: Here, too, backward compatibility may change

src/libopensc/pkcs15-iasecc.c
henriquefurlan: Good point.

So yes, as written, this changes the behavior for all cards identified as `SC_CARD_TYPE_IASECC_OBERTHUR`, not only for the specific card that returns `SC_ERROR_DATA_OBJECT_NOT_FOUND`.

The intent was to validate the missing-metadata/fixup path on the affected IAS/ECC v1.0.1 Oberthur card, but I agree this is broader than ideal for a final change.

A safer follow-up would be to narrow the behavior behind a more specific condition, for example:
- only enable the fallback/fixup after detecting the missing SDO/DOCP reads
- or gate it on a more specific ATR / card profile / broken metadata pattern

I can rework the patch in that direction.

frankmorgner:
What's the status of this topic? Will you move this forward?

PR #3742, opened 2026-07-21
enable caching for zero length files.

### LINKED ISSUE #3741: Performance: Enable caching of zero size files
### Problem statement

Short summary:

Enabling caching, for files of size zero, can increases the
performance on RDP connection of around 33%.

### Proposed solution

Longer description:

OpenSC uses files caching for pkcs15 objects. The performance gain is achived by sparing the rather expensive APDU call / responses on the SCARD reader.

For example, without caching:

```
$ cat /etc/opensc.conf | grep use_file_caching
        use_file_caching = false;
$ time pkcs11-tool --list-objects 
[...]
real    0m14.613s    <----- 14 secs
user    0m0.130s
sys     0m0.093s

```
with caching:

```
$ cat /etc/opensc.conf | grep use_file_caching
        use_file_caching = public;
$ time pkcs11-tool --list-objects 
[...]
real    0m6.075s  <----- 6 secs
user    0m0.117s
sys     0m0.095s

```
Every APDU calls is expensive, especially in this example, where the connection is done over network (RDP).

I have found out, that OpenSC.conf spares the file caching for pkcs15 objects which have a zero length:

https://sourceforge.net/p/opensc/mailman/message/59359512/

Here a record with ten entries is read, each record has a length of zero and the overall file size is zero. This is the result of 10(!) APDU calls.
Subsequent calls always read the 10 records again and again, which decreases performance.

Enabling caching for zero byte length files:

```
./src/libopensc/pkcs15.c:L2630

- if (len && p15card->opts.use_file_cache
+ if (p15card->opts.use_file_cache
```
The result:

```
$ time pkcs11-tool --list-objects 
real    0m3.987s    <------  4 secs
user    0m0.114s
sys     0m0.092s
```
This a nice performance gain of around 33%. 

(I will create a pull request for this.)

### Alternatives considered

_No response_

### Additional context

_No response_

mvogt1:
Currently OpenSC does not cache files of size zero. This has a performance implication on network connections.
(Fixes #3741)

### DIFF
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -2625,8 +2625,7 @@ sc_pkcs15_read_file(struct sc_pkcs15_card *p15card, const struct sc_path *in_pat
 
 		sc_file_free(file);
 
-		if (len && p15card->opts.use_file_cache
-		    && ((p15card->opts.use_file_cache & SC_PKCS15_OPTS_CACHE_ALL_FILES) || !private_data)) {
+		if (p15card->opts.use_file_cache && ((p15card->opts.use_file_cache & SC_PKCS15_OPTS_CACHE_ALL_FILES) || !private_data)) {
 			sc_pkcs15_cache_file(p15card, in_path, data, len);
 		}
 		if (len == 0) {


Jakuje (approved):
I do not see any issue with this. But this check was added in #1132, which is quite unspecific in the reasons so I would like to hear from @frankmorgner first if looks reasonable.

PR #3743, opened 2026-07-23
card-iasecc: add support for the Monaco national eID

serialseb:
Some IAS/ECC cards omit optional data objects the driver treated as
mandatory, leaving otherwise PKCS#15-compliant cards unusable.

Two changes are general robustness fixes, applied to every card:

  - iasecc_parse_ef_atr(): tolerate an EF.ATR that omits the optional
    pre-issuing/issuer data objects (they only fed a log line and the IO
    buffer sizes; fall back to the defaults 0xFF / 0x100).
  - iasecc_pin_get_policy(): when the card has no CHV SDO (GET DATA -> 6A88),
    fall back to a plain, non-SM verify with the PKCS#15 AODF PIN parameters,
    instead of failing before the VERIFY APDU is sent. This is correct for any
    card lacking the SDO, so it is not gated.

Two changes assume values the card declines to provide, and would be a guess
if applied blindly, so they are gated on per-card quirk flags carried in the
ATR table 'flags' column and copied to card->flags at init (vendor-reserved
range, mirroring card-setcos.c / card-flex.c / card-authentic.c):

  - IASECC_QUIRK_NO_KEY_SDO: the card exposes no RSA-private SDO, so
    iasecc_set_security_env() assumes a 2048-bit key and plain PSO access.
  - IASECC_QUIRK_NO_ALGO_REF: supportedAlgorithms is empty, so the standard
    IAS/ECC algorithm references are used for the MSE:SET templates.

Every other card keeps the original strict behaviour: an unflagged card that
returns 6A88 for the key SDO, or advertises no algorithm references, follows
the unchanged error path.

The first card known to need these fallbacks is the Monaco national eID
(Gemalto/Thales IAS/ECC applet on an NXP chip): its ATR is registered as an
"Amos" type with the per-batch bytes masked, and only its row opts in to the
quirk flags. opensc-pkcs11 then enumerates the card's keys and certificates,
logs in, and signs.

Tested card (opensc-tool -n): IAS-ECC
ATR: 3B:DC:18:FF:81:91:FE:1F:C3:80:73:C8:21:13:66:05:03:63:51:00:02:50

## Checklist
- [x] PKCS#11 module is tested (keys + certs enumerate; session opens; certs read; signs end to end)
- [ ] Documentation is added or updated — n/a (ATR-table entry only)
- [ ] New files have an LGPL 2.1 license statement — n/a (no new files)
- [ ] Windows minidriver is tested — n/a (tested on Linux only)
- [ ] macOS tokend is tested — n/a (tested on Linux only)

## Note for maintainers
The "Amos" mapping is by ATR historical-byte similarity to the existing SC_CARD_TYPE_IASECC_AMOS entries; the ATR mask covers only the per-batch bytes. Confirmation from a second Monaco cardholder's ATR would help validate the mask width.


frankmorgner:
Can you point to a iasecc specification, which marks the SDO as optional and which defines RSA 2048 as standard parameters?

Am 23. Juli 2026 14:34:03 MESZ schrieb Sebastien Lambla ***@***.***>:

serialseb:
Apologies for the very slow reply.

Good question — and the direct answer to both halves is no: I can't point to such language. Upfront, I don't have the IAS-ECC specification to hand, and I won't claim a clause I can't quote; where I point at something concrete below, it's ISO 7816 or the existing driver. Let me separate the two, because they differ in kind.

**The two general fixes are ISO 7816, not IAS-ECC quirks:**

- *EF.ATR omitting the pre-issuing / card-issuer DOs.* In ISO/IEC 7816-4 those EF.ATR (2F01) data objects are optional and issuer-defined; a card omitting them is conformant. `iasecc_parse_ef_atr()` treated them as present (they only fed a log line and the IO buffer sizes), so tolerating their absence and falling back to 0xFF / 0x100 is a conformance fix — applied to every card.
- *No CHV SDO → 6A88.* `6A88` ("referenced data not found", ISO 7816-4) is a legitimate answer from a card that does not expose a CHV SDO via GET DATA. The PIN parameters are still fully given by the card's own PKCS#15 AODF (ISO/IEC 7816-15), so `iasecc_pin_get_policy()` falls back to a plain, non-SM VERIFY from the AODF instead of aborting before the APDU — taking the values from the card rather than inventing them. Also ungated.

**The two gated changes are not spec defaults — and I'm not presenting them as such, which is why they're gated. Their basis is the existing driver and the card itself:**

- `IASECC_QUIRK_NO_KEY_SDO` assumes a 2048-bit key when the card exposes no RSA-private SDO to read the modulus from. 2048 is already what card-iasecc.c registers for IAS-ECC RSA (`_sc_card_add_rsa_alg(card, 2048, …)`, several call sites), and it matches this card's actual keys — both on-card RSA keys are 2048-bit per its own PKCS#15 PrKDF and its certificates. Still an assumption, hence per-card.
- `IASECC_QUIRK_NO_ALGO_REF` reuses the driver's own `IASECC_ALGORITHM_*` constants — the same ones the existing MSE:SET templates use — when `supportedAlgorithms` is empty. Reuse of in-tree values, applied only when the card advertises none.

Both live in the vendor-reserved flags range (mirroring card-setcos.c / card-flex.c / card-authentic.c); every unflagged card keeps the unchanged strict path.

On uniformity: this is Monaco's eID for citizens and residents — one issuer profile, so the ATR row and assumed values generalise across cards, not just mine. Verified on one; a second ATR would confirm the mask.

If there is IAS-ECC specification language on either point that I've missed, I would genuinely welcome the pointer. And if you'd prefer, I'm happy to split this pull request into parts: land the two ISO-7816 fixes on their own, and keep the two assumptions scoped strictly to the Monaco ATR row (or carry them out-of-tree) — whichever you're more comfortable merging. Any guidance would be very welcome.

frankmorgner (changes requested):
I am not happy with adding global flags for what seems a workaround for a single card. If some workarounds are needed for your card, please check for your card type an apply the workarounds rather than adding global flags. This makes it more transparent why we need this.

I am OK with adding the possibility of skipping pre-issuing data if needed, but please leave values uninitialized rather than setting it to *your* defaults. If you need to set some value here, do this based on your card type.

If iasecc_sdo_get_data() fails for your card, would it be possible to provide a virtual SDO which matches the configuration of your card? If properly constructed, this should avoid the modifications you made in iasecc_set_security_env(), because your profile should then reflect the capabilities of the card. The missing SDO for the PIN object could be handled similarly, but hard coding this based on your card type (as you just did), should be simpler in this case.

Please be transparent about the helper tools that you used to draft this PR.

frankmorgner (changes requested):
I am not happy with adding global flags for what seems a workaround for a single card. If some workarounds are needed for your card, please check for your card type an apply the workarounds rather than adding global flags. This makes it more transparent why we need this.

I am OK with adding the possibility of skipping pre-issuing data if needed, but please leave values uninitialized rather than setting it to *your* defaults. If you need to set some value here, do this based on your card type.

If iasecc_sdo_get_data() fails for your card, would it be possible to provide a virtual SDO which matches the configuration of your card? If properly constructed, this should avoid the modifications you made in iasecc_set_security_env(), because your profile should then reflect the capabilities of the card. The missing SDO for the PIN object could be handled similarly, but hard coding this based on your card type (as you just did), should be simpler in this case.

Please be transparent about the helper tools that you used to draft this PR.

serialseb:
Morning,

I'll address the tools I use for submitting this PR. As I am handicaped I usually use VoiceInk to prepaare the text of a PR, and use claude or locally qwen 3.5 to add the technical part the large model in VoiceInk doenn't understand. Writing those manually (like in doing now for full transparency) can take a very long time and leads to a lot of errors that make the PR harder to reason about, let alone get merged. 

I use claude to format code in the style of the original repository, with data collection disabled to avoid any code becoming part of a training dataset. 

I have done some research t try to source cards issued under thesame kind of national id contract to governments by the same provider, but as those are national governmet level i don't yet have a wider view, and cannot say if the quirks are a habit of the developer for those kind of cards or a true quirk of this design..

src/libopensc/card-iasecc.c
```diff
@@ -377,42 +379,61 @@ static int iasecc_parse_ef_atr(struct sc_card *card)
...
+		       "EF.ATR: max send/recv sizes %"SC_FORMAT_LEN_SIZE_T"X/%"SC_FORMAT_LEN_SIZE_T"X",
+		       card->max_send_size, card->max_recv_size);
+	}
+	else if (card->type == SC_CARD_TYPE_IASECC_MONACO)   {
+		/* The Monaco eID publishes no issuer data, and IASECC_CARD_DEFAULT_CAPS
+		 * advertises SC_CARD_CAP_APDU_EXT: left at zero, sc_get_max_send_size()
+		 * would offer this card 65535/65536.  Pin it to short APDUs, which is
+		 * what it was verified against. */
+		card->max_send_size = 0xFF;
+		card->max_recv_size = 0x100;
```
frankmorgner, line 420: Maybe you want to make some tests with your card what length is supported an which length is performing well. Choosing a bigger buffer typically results in significant performance boosts (more throughput). You could try with opensc-explorer or opensc-tool to perform READ BINARY with different values and check which ones result in better throughput.

src/libopensc/card-iasecc.c
```diff
@@ -2874,7 +2960,7 @@ iasecc_sdo_get_tagged_data(struct sc_card *card, int sdo_tag, struct iasecc_sdo
 
 
 static int
-iasecc_sdo_get_data(struct sc_card *card, struct iasecc_sdo *sdo)
+iasecc_sdo_read_from_card(struct sc_card *card, struct iasecc_sdo *sdo)
```
frankmorgner: rename should not be necessary.

src/libopensc/card-iasecc.c
```diff
@@ -1769,29 +1821,29 @@ iasecc_set_security_env(struct sc_card *card,
 			LOG_TEST_RET(ctx, SC_ERROR_INVALID_ARGUMENTS, "Need RSA_PKCS1 specified");
 
 		if (env->algorithm_flags & SC_ALGORITHM_RSA_HASH_SHA256)   {
-			algo_ref = iasecc_get_algorithm(ctx, env, SC_PKCS15_ALGO_OP_HASH, CKM_SHA256);
-			if (!algo_ref)
-				LOG_TEST_RET(ctx, SC_ERROR_NOT_SUPPORTED, "Card application do not supports HASH:SHA256");
+			rv = iasecc_get_algorithm_or_default(card, env, SC_PKCS15_ALGO_OP_HASH, CKM_SHA256,
+					IASECC_ALGORITHM_SHA2, &algo_ref);
+			LOG_TEST_RET(ctx, rv, "Card application do not supports HASH:SHA256");
```
frankmorgner: I think it would make more sense to adapt iasecc_get_algorithm in order to handle your hardcoded mechanisms. Could you check if this makes sense here (and for the other variants)?
```diff
diff --git a/src/libopensc/card-iasecc.c b/src/libopensc/card-iasecc.c
index f508f1eb8..664b31578 100644
--- a/src/libopensc/card-iasecc.c
+++ b/src/libopensc/card-iasecc.c
@@ -1517,7 +1517,7 @@ iasecc_check_sw(struct sc_card *card, unsigned int sw1, unsigned int sw2)
 
 
 static unsigned
-iasecc_get_algorithm(struct sc_context *ctx, const struct sc_security_env *env,
+iasecc_get_algorithm(struct sc_card *card, const struct sc_security_env *env,
                unsigned operation, unsigned mechanism)
 {
     const struct sc_supported_algo_info *info = NULL;
@@ -1526,6 +1526,15 @@ iasecc_get_algorithm(struct sc_context *ctx, const struct sc_security_env *env,
     if (!env)
         return 0;
 
+       if (card->type == SC_CARD_TYPE_IASECC_MONACO) {
+               if (operation == SC_PKCS15_ALGO_OP_HASH && mechanism == CKM_SHA256)
+                       return IASECC_ALGORITHM_SHA2;
+               /* add more supported algorithms here */
+
+               /* operation not supported for SC_CARD_TYPE_IASECC_MONACO */
+               return 0;
+       }
+
     for (ii=0;ii<SC_MAX_SUPPORTED_ALGORITHMS && env->supported_algos[ii].reference; ii++)
         if ((env->supported_algos[ii].operations & operation)
                        && (env->supported_algos[ii].mechanism == mechanism))
@@ -1533,11 +1542,11 @@ iasecc_get_algorithm(struct sc_context *ctx, const struct sc_security_env *env,
 
     if (ii < SC_MAX_SUPPORTED_ALGORITHMS && env->supported_algos[ii].reference)   {
         info = &env->supported_algos[ii];
-        sc_log(ctx, "found IAS/ECC algorithm %X:%X:%X:%X",
+        sc_log(card->ctx, "found IAS/ECC algorithm %X:%X:%X:%X",
                        info->reference, info->mechanism, info->operations, info->algo_ref);
     }
     else   {
-        sc_log(ctx, "cannot find IAS/ECC algorithm (operation:%X,mechanism:%X)", operation, mechanism);
+        sc_log(card->ctx, "cannot find IAS/ECC algorithm (operation:%X,mechanism:%X)", operation, mechanism);
     }
 
     return info ? info->algo_ref : 0;
```
On the other hand, I still wonder if `env->supported_algos` will get initialized with everything needed, when you supply a more sophisticated structure in `iasecc_sdo_virtual_monaco()`

frankmorgner (changes requested):

frankmorgner:
> Morning,
> 
> I'll address the tools I use for submitting this PR. As I am handicaped I usually use VoiceInk to prepaare the text of a PR, and use claude or locally qwen 3.5 to add the technical part the large model in VoiceInk doenn't understand. Writing those manually (like in doing now for full transparency) can take a very long time and leads to a lot of errors that make the PR harder to reason about, let alone get merged.
> 
> I use claude to format code in the style of the original repository, with data collection disabled to avoid any code becoming part of a training dataset.
> 
> I have done some research t try to source cards issued under thesame kind of national id contract to governments by the same provider, but as those are national governmet level i don't yet have a wider view, and cannot say if the quirks are a habit of the developer for those kind of cards or a true quirk of this design..

This looks like an interesting setup. Are there any conference talks or similar, that show how the tooling looks like in order to compensate specific disabilities?

In case of AI-assisted coding we need to make sure that not too much noise enters the code base. Also, it is important to verify claims about standard procedures and things like that. Thanks for being transparent.

serialseb:
As promised, I went and looked at the numbers, but the card wouldn't play ball. Whatever length I ask for, `READ BINARY` comes back `9000` with 231 bytes.

That number didn't sit well with me. I let it pass in the initial PR because the code base already worked that way and I needed the card working for my own project, but I find magic numbers distasteful, so I went looking for a way to get rid of it altogether.

Where I got to: the card declares 1286 bytes in both directions in the `7F66` DO of `EF.ATR`, so my patch could read that instead of pinning 255/256. And when a card declares nothing: ask for the data, record the size you got, ask again at that offset. If data comes back, the first read was not the end of the file, and you have the card's maximum response size.

If you don't, you fall back to the quirk.

The trick only measures what the card sends back. For the other direction I'd assume they match, if only to keep my sanity.

Which way would you prefer I take, if any?

<details>
<summary>How I got there</summary>

I tried increasing lengths, then bisection. Same answer every time.

The card won't tell us anything it doesn't have to. The FCI comes back empty, and the keyset and security environment objects answer `6A88`. At that point all I had was 256 minus 231, and that kind of number looked too much like a red flag.

`EF.CardAccess` does exist, but it is for NFC.

Then `EF.ATR`, with its 1286. Optional in ISO, but present on this card. Other cards running the same operating system may well declare the same thing. I can't prove it, I don't have access to more of these cards.

</details>

frankmorgner:
We have `sc_parse_ef_atr()`, if that helps. If there will be no other card variants, however, it should be no problem to use the magic numbers you derived.

serialseb:
That helps a lot, thanks! My card never gets as far as getting the attributes: its capabilities byte is `81`, so the extended length information bit is not set and the parse is skipped.

The same byte explains the original weirdness. The extended length bit is not set either, so the card says it does not do extended APDUs, and the short APDU maximum is the right size, it is what the card asks for. The number has lost its magic and I am happy to leave the code as it is. Should future IAS-ECC cards, or a new batch of this one, change that, the method is in the thread for future committers.

Both changes are pushed: the rename dropped, and `iasecc_get_algorithm` along the lines of your diff, with a caveat, I left `decipher` strict. This card does not support deciphering and the failed lookup is what prevents an attempt reaching it.

Unless you have more feedback, I think I'm pretty much done on my side.

src/libopensc/card-iasecc.c
```diff
@@ -2088,37 +2154,54 @@ iasecc_pin_get_policy (struct sc_card *card, struct sc_pin_cmd_data *data, struc
...
-	pin->tries_remaining = (sdo.docp.tries_remaining.value ? *sdo.docp.tries_remaining.value : -1);
-	if (sdo.docp.size.value && sdo.docp.size.size <= sizeof(int)) {
-		unsigned int n = 0;
-		unsigned int i;
-		for (i=0; i<sdo.docp.size.size; i++)
-			n = (n << 8) + *(sdo.docp.size.value + i);
-		pin->stored_length = n;
-	} else {
+	if (rv == SC_ERROR_DATA_OBJECT_NOT_FOUND && card->type == SC_CARD_TYPE_IASECC_MONACO) {
+		/* The Monaco eID does not expose the CHV SDO.  Its only role here is the
```
frankmorgner:
```suggestion
	if (rv == SC_ERROR_DATA_OBJECT_NOT_FOUND) {
		/* The card, e.g. Monaco eID, does not expose the CHV SDO.  Its only role here is the
```
I just now see that you're initializing the whole structure without card specific values, but rather with "unknown" values. I think this is a good approach also for other cards that are missing the SDO reference for CHV. So in this particular case it makes sense to actually remove the check

frankmorgner (changes requested):
I think the change looks OK now, just one minor comment.

frankmorgner (approved):

serialseb:
Apologies I made a snafu, had prepared the code formatting, and didn't push the right buttons (secondary email not github registered, doesnt match my s/mime, was going to ask if it was useful at all). Let me know if there's a way to fiix it (II'll add the email to my verified ones at least) 

@frankmorgner should i revert the formatting change?

frankmorgner:
Yes, please drop [21fd87b](https://github.com/OpenSC/OpenSC/pull/3743/commits/21fd87bd8323260458d98710a98fdbb98ded1823). I'm waiting for a second maintainer for a review.

src/libopensc/card-iasecc.c
```diff
@@ -377,38 +379,57 @@ static int iasecc_parse_ef_atr(struct sc_card *card)
...
+		sc_log(ctx, "EF.ATR: IC manufacturer/type %X/%X, OS/IasEcc versions %X/%X",
+			version->ic_manufacturer, version->ic_type, version->os_version, version->iasecc_version);
+	}
+	else   {
+		sc_log(ctx, "EF.ATR: no pre-issuing data, skipping version info");
+	}
+
+	/* Issuer data carries the negotiated IO buffer sizes.  When it is absent the
+	 * sizes stay as the framework left them, except for cards known to need a
+	 * concrete value: see the SC_CARD_TYPE_IASECC_MONACO branch below. */
```
frankmorgner:
```suggestion
```
Comment not needed, codes is clear, I think

src/libopensc/card-iasecc.c
```diff
@@ -377,38 +379,57 @@ static int iasecc_parse_ef_atr(struct sc_card *card)
...
+		if (card->max_send_size > 0xFF)
+			card->max_send_size -= 5;
+
+		sc_log(ctx, "EF.ATR: max send/recv sizes %zX/%zX", card->max_send_size, card->max_recv_size);
+	}
+	else if (card->type == SC_CARD_TYPE_IASECC_MONACO)   {
+		/* The Monaco eID publishes no issuer data, and IASECC_CARD_DEFAULT_CAPS
+		 * advertises SC_CARD_CAP_APDU_EXT: left at zero, sc_get_max_send_size()
+		 * would offer this card 65535/65536.  Pin it to short APDUs, which is
+		 * what it was verified against. */
```
frankmorgner:
```suggestion
```
Comment not needed, codes is clear, I think

src/libopensc/card-iasecc.c
```diff
@@ -643,6 +664,13 @@ iasecc_init(struct sc_card *card)
 		rv = iasecc_init_amos_or_sagem(card);
 	else if (card->type == SC_CARD_TYPE_IASECC_AMOS)
 		rv = iasecc_init_amos_or_sagem(card);
+	else if (card->type == SC_CARD_TYPE_IASECC_MONACO)   {
+		rv = iasecc_init_amos_or_sagem(card);
+		/* The card answers 6D 00 to the ISO PIN status query (VERIFY with no
+		 * data), so it does not support it. Drop the capability rather than
+		 * issue a command the card rejects. */
```
frankmorgner:
```suggestion
```
Comment not needed, codes is clear, I think

src/libopensc/card-iasecc.c
```diff
@@ -1970,6 +2029,13 @@ iasecc_pin_verify(struct sc_card *card, struct sc_pin_cmd_data *data)
 			if (iasecc_chv_cache_is_verified(card, &pin_cmd))
 				LOG_FUNC_RETURN(ctx, SC_SUCCESS);
 	}
+	else if (rv == SC_ERROR_NOT_SUPPORTED)   {
+		/* The card cannot report PIN status. That leaves the login state
+		 * unknown, which is a reason to present the PIN rather than a reason to
+		 * give up: without this, no card lacking SC_CARD_CAP_ISO7816_PIN_INFO
+		 * can verify a PIN through this driver at all. */
```
frankmorgner:
```suggestion
```
Comment not needed, codes is clear, I think

src/libopensc/card-iasecc.c
```diff
@@ -2127,7 +2210,17 @@ iasecc_pin_get_info(struct sc_card *card, struct sc_pin_cmd_data *data)
 	 * policy takes precedence.
 	 */
 	rv = iasecc_pin_get_status(card, data);
-	LOG_TEST_RET(ctx, rv, "Failed to get PIN status");
+	if (rv == SC_ERROR_NOT_SUPPORTED)   {
+		/* No status from the card. Say "unknown" explicitly: leaving the
+		 * caller's zeroed fields alone would report a perfectly good PIN as
+		 * having no tries left. */
```
frankmorgner:
```suggestion
```
Comment not needed, codes is clear, I think

frankmorgner (changes requested):
While we still have time, we can drop some of the needless comments.

serialseb:
No comment. :-)

frankmorgner (approved):

serialseb:
Happy thursday ! Master has moved quite a bit since the last approval / run. We're still mergeable but if you think I should rebase just say the word.

frankmorgner (changes requested):
Please check the suggested styling changes:
https://github.com/OpenSC/OpenSC/actions/runs/31946690902/job/95215734162?pr=3743

applying the suggested diff should be good enough.

PR #3744, opened 2026-07-23
minidriver: Add ATRs for Japanese Individual Number Card (My Number Card)

### LINKED ISSUE #3021: Possibility of another JPKI ATR
### Problem Description

I got MyNumber Card recently but this was unable to detect my card.
It has ATR starting with 3b:da:13:ff:81:31:fb:46:80:12:39.
Windows recongizing this card as unknown smart card.
image

hamano:
### Summary

This PR adds the ATRs of the Japanese Individual Number Card (マイナンバーカード / JPKI) to the Windows minidriver registration in `win32/customactions.cpp`, so that Windows can automatically associate these cards with the OpenSC minidriver.

Fixes #3021

### Background

On Windows, the smart card minidriver is selected **primarily by ATR** (via the `Calais\SmartCards` registry entries written by the installer). Until now only a single JPKI ATR was registered, so most My Number Cards were shown as an "unknown smart card" and users had to edit the registry by hand (see #3021).

This change **only affects the Windows minidriver registration**. It does not touch card detection in `card-jpki.c`:

- **PKCS#11 and the OpenSC internal driver detect the card by AID selection**, not by ATR, so they already work with any reader and any card and need no change.
- The ATR table is required solely because Windows has no other way to bind a third-party minidriver to a card (PnP / `SCFILTER` hardware IDs are also derived from the ATR).

### About the contactless interface

The My Number Card is an ISO 14443 Type B card, and the contactless ATR is synthesized by the reader:

- **PC/SC part 3 compliant readers** build the ATR from the ATQB (Application Data / Protocol Info / MBLI). This ATR is **stable per card type**, so it can be registered — these entries are included in this PR.
- Some readers (e.g. ACR122U) embed the card's **PUPI** in the ATR. The My Number Card uses a random PUPI for privacy, so the ATR changes on every tap and **cannot be used for detection**. This is a limitation we have to accept on the minidriver side; users of such readers can still use the card through PKCS#11.

### Source of the ATRs

The added ATRs are taken from the smart card ATR list maintained at:

https://pcsc-tools.apdu.fr/smartcard_list.txt

and from the ATRs reported by users on real cards and readers in #3021.

### Changes

- `win32/customactions.cpp`: add the contact and contactless ATRs of the My Number Card to `minidriver_registration[]` (full `0xFF` masks, unique names, verified TCK).


frankmorgner:
thank you

Issue #3663, opened 2026-04-20
srbeid driver incorrectly identifies Lithuanian eID as Serbian eID


tadas-s:
### Describe the bug

`opensc-tool` output:

```
tadas@box ~/local $ ./bin/opensc-tool --atr
Using reader with a card: Gemalto PC Twin Reader (AD010CEC) 00 00
3b:9d:18:81:31:fc:35:80:31:c0:69:4d:54:43:4f:53:73:02:06:05:d0
tadas@box ~/local $ ./bin/opensc-tool --name
Using reader with a card: Gemalto PC Twin Reader (AD010CEC) 00 00
Serbian CardEdge driver
tadas@box ~/local $
```

Relevant log part:

```
P:74508; T:0x139779372234816 12:49:55.872 [pkcs15-tool] card-srbeid.c:73:srbeid_match_card: srbeid: CardEdge applet found via AID
P:74508; T:0x139779372234816 12:49:55.872 [pkcs15-tool] card.c:351:sc_connect_card: matched: Serbian CardEdge driver
P:74508; T:0x139779372234816 12:49:55.872 [pkcs15-tool] card-srbeid.c:84:srbeid_init: called
P:74508; T:0x139779372234816 12:49:55.872 [pkcs15-tool] card-srbeid.c:91:srbeid_init: returning with: 0 (Success)
P:74508; T:0x139779372234816 12:49:55.872 [pkcs15-tool] card.c:378:sc_connect_card: card info name:'Serbian CardEdge driver', type:44000, flags:0x0, max_send/recv_size:255/256
```

And relevant code:

https://github.com/OpenSC/OpenSC/blob/c1b5267e886177517df0a12c004ff2c3fb720086/src/libopensc/card-srbeid.c#L71-L76

The "applet" exists on Lithuanian eID card:

```
tadas@box ~/local $ ./bin/opensc-explorer
OpenSC Explorer version 0.27.1
Using reader with a card: Gemalto PC Twin Reader (AD010CEC) 00 00
OpenSC [3F00]> cd aid:A0:00:00:00:63:50:4B:43:53:2D:31:35
OpenSC [A000/0000/6350/4B43/532D/3135]>
```

If I had to guess it's pretty common applet in various eIDs and cannot be used to uniquely identify Serbian eID.

Not really huge issue for now, because to use Lithuanian eID user needs to add `card_atr` section to config file:

```
app default {
[...]

PR #3751, opened 2026-07-28
card-chipdocit: support the Bit4id Digital-DNA Key (Italian CNS/eID)

ale-rinaldi:
Adds a card driver and PKCS#15 emulation for the Bit4id Digital-DNA Key, an
Italian CNS/eID built on an NXP ChipDoc chip (same chip family as card-eoi).

- Access is gated by a PACE (BSI TR-03110) channel via OpenPACE, using a
  card-specific CAN stored encrypted on the card (same unwrap scheme and key
  as card-eoi).
- The CNS and qualified-signature (DS3) certificates and their RSA-2048 keys
  are exposed; DS signatures are produced via PSO.
- The certificates live in a compressed container the generic PKCS#15 reader
  cannot decode, so the card is handled as emulation-only and inflated with a
  preset dictionary (bit4id-cert-dict.h).
- The ATR is shared with other Italian CNS cards, so match_card confirms the
  card application is present and declines otherwise.
- PIN change and unblock are mirrored onto the two in-sync PIN objects (the
  CNS-applet PIN and the IAS PIN under PACE).

Guarded by ENABLE_SM/ENABLE_OPENPACE (ENABLE_ZLIB for the emulator). Tested on
real hardware: certificate reading, DS signature, PIN change and unblock.


src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,598 @@
...
+	int i;
+	u8 b = from[0] & 0x80;
+	for (i = 15; i >= 0; i--) {
+		u8 bo = b;
+		b = from[i] & 0x80;
+		to[i] = (from[i] << 1) | (bo ? 1 : 0);
+		if ((i == 15) && bo)
+			to[i] = (to[i] ^ 0x87) | 1;
+	}
+}
```
Jakuje: would it make sense to extract this to some shared file rather than making a copy?

src/libopensc/cards.h
```diff
@@ -270,6 +270,9 @@ enum {
 	SC_CARD_TYPE_EOI = 41000,
 	SC_CARD_TYPE_EOI_CONTACTLESS,
 
+	/* Bit4id Digital-DNA (NXP ChipDoc, Italian CNS/eID) */
+	SC_CARD_TYPE_CHIPDOC_IT = 41100,
```
Jakuje: Given that its separate driver, I would rather see it as next card id at 46000

src/libopensc/card-chipdocit.c
ale-rinaldi: Good point — the CAN unwrap and its key are identical to card-eoi (same NXP ChipDoc chip). Factored the decryption into `eoi_decrypt_can_block()` in card-eoi.c (declared in card-eoi.h); card-chipdocit now calls it and drops its copy. Happy to move it to a dedicated shared file instead if you prefer that placement.

src/libopensc/cards.h
ale-rinaldi: Done — moved to `SC_CARD_TYPE_CHIPDOC_IT = 46000`.

src/libopensc/card-chipdocit.c
Jakuje: I think this is ok. Thank you.

src/libopensc/pkcs15-chipdocit.c
```diff
@@ -0,0 +1,281 @@
...
+		int n = raw[1] & 0x7f;
+		hdr = 2 + n;
+		for (bodylen = 0, i = 0; i < n; i++)
+			bodylen = (bodylen << 8) | raw[2 + i];
+	} else {
+		hdr = 2;
+		bodylen = raw[1];
+	}
+	if (hdr + bodylen > (size_t)rawlen)
+		return SC_SUCCESS;
```
Jakuje, line 106: This sounds like parsing of TLV. Please, use existing ASN1 functions to do that. I think `sc_asn1_read_tag()` should do this with all the error checking

ale-rinaldi: Done — replaced the manual length parsing with `sc_asn1_read_tag()` (commit pushed). Also fixed the CI: the two new files were missing from the win32 Makefile (MSVC link error) and the code is now clang-format clean.

src/libopensc/card-eoi.h
```diff
@@ -39,3 +39,7 @@ struct eoi_privdata {
 	/* PrKey reference to eOI mappings */
 	int prkey_mappings[MAX_OBJECTS][2];
 };
+
+/* Decrypt the 24-byte encrypted CAN blob into its 16-byte plaintext block;
+ * shared with card-chipdocit (same chip and key). */
+int eoi_decrypt_can_block(const u8 *enc_can, size_t enc_can_len, u8 out[AES_BLOCK_SIZE]);
```
frankmorgner: If this is shared among differnt cards, then this may be a mechanism that should be named more generically. However, given how vendors have sneaked this into their proprietary middleware, I doubt that there is public information about this...

src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,543 @@
...
+chipdocit_finish(sc_card_t *card)
+{
+	free(card->drv_data);
+	card->drv_data = NULL;
+	return SC_SUCCESS;
+}
+
+/* SELECT an application by full AID (00 A4 04 04 <len> <aid> 00). */
+static int
+chipdocit_select_aid(sc_card_t *card, const u8 *aid, size_t aid_len)
```
frankmorgner: this should be identical to `iso7816_select_aid()`, which is exposed as generic operation via opensc.h

Jakuje (approved):
I do not have any more comments. I will leave it for couple days for others to have a look, mostly @frankmorgner  who was involved in the initial discussion.

src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,543 @@
...
+chipdocit_send_clear(struct sc_card *card, int ins, int p1, int p2,
+		const u8 *body, size_t len)
+{
+	struct sc_apdu apdu;
+	int r;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, ins, p1, p2);
+	apdu.lc = apdu.datalen = len;
+	apdu.data = body;
+	r = sc_transmit_apdu(card, &apdu);
```
frankmorgner: Are you sure that this will be sent without SM? `sc_send_apdu()` should apply SM transparently. I think you will need to call `sc_sm_stop()` first.

frankmorgner (commented):
I noticed that AID and MF are selected multiple times. Most other card drivers assume that the card's state is somewhat stable across multiple commands to reduce the amount of APDUs sent to the card. If there would be some significant performance-gain you may want to remove some of the selection commands. In that case, `card_reader_lock_obtained()` should be implemented so that you can reselect your Applet/MF in case of a card reset (from some concurrent process).

Please consider if you can use the iso7816 routines in order to perform a PIN verification/modification (after SM channel has been opened). It also supports a customizable `pad_char`. Also, a reader with a PIN-pad is supported there, which your implementation seems to ignore for now. For reference, in `card-npa.c` the QES-PIN is changed within the SM channel with a traditional PIN command in `npa_standard_pin_cmd()`, which may be similar.

src/libopensc/card-chipdocit.c
ale-rinaldi: Done — now using `iso7816_select_aid()` (dropped the private helper).

src/libopensc/card-chipdocit.c
ale-rinaldi: Good catch — `chipdocit_select_cns()` now calls `sc_sm_stop()` before selecting the CNS applet, so the in-clear CNS commands are actually sent without SM (removed the direct `sm_ctx.sm_mode` assignment). Verified on hardware: PIN change updates both PIN objects.

src/libopensc/card-eoi.h
ale-rinaldi: Renamed to `chipdoc_decrypt_can_block()` to reflect the shared NXP ChipDoc chip rather than a specific card. As you note, the scheme is not publicly documented, so a fully generic name is hard — happy to adjust if you have a preference.

ale-rinaldi:
Thanks @frankmorgner — addressed both points:

- **iso7816 PIN routines / pad_char / PIN-pad**: the IAS (primary) PIN already goes through the iso7816 `pin_cmd`, so `pad_char` and a PIN-pad reader are handled there. I have now routed the secondary CNS-applet PIN (verify/change/unblock) through the iso7816 `pin_cmd` too (0xFF padding via `pad_char`), dropping the hand-built APDUs. That CNS-applet PIN is a second object kept byte-synced with the IAS PIN and accessed in the clear, so it inherently needs the plaintext — a PIN-pad cannot drive it; the primary verification on the IAS side is what benefits from the PIN-pad.

- **repeated AID/MF selects**: these correspond to genuine switches between the card applications — IAS-ECC (CAN + PIN), the SSCD applet (PSO signature) and the CNS applet (mirrored PIN) — plus the MF for PACE (EF.CardAccess). Reselecting the target before each operation also keeps the driver correct if another process resets the card mid-session, which is what `card_reader_lock_obtained()` would otherwise handle. The overhead is a few short APDUs, so I kept it explicit for simplicity/robustness rather than caching state; happy to switch to cached state + `card_reader_lock_obtained()` if you would prefer fewer round-trips.

src/libopensc/card-eoi.h
frankmorgner: thanks

src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,506 @@
...
+	/* If the IAS side failed after the CNS PIN was changed, roll the CNS PIN
+	 * back so the two objects do not diverge (CHANGE only). */
+	if (r != SC_SUCCESS && synced && cmd == SC_PIN_CMD_CHANGE) {
+		/* Revert the CNS PIN (new -> old) so the two objects stay in sync. */
+		if (chipdocit_select_cns(card) == SC_SUCCESS)
+			(void)chipdocit_cns_pin_op(card, SC_PIN_CMD_CHANGE,
+					CHIPDOCIT_CNS_PIN_BSO, &p2_save, &p1_save);
+	}
+	LOG_FUNC_RETURN(card->ctx, r);
+}
```
frankmorgner, line 409: Could you verify with a PIN-pad-reader that the syncing process still works? When triggered via Windows MD, the application will NOT show a PIN-dialog, instead it instructs the user to enter the old and new PIN at the reader. I believe you can simulate this behavior with `pkcs15-tool --use-pinpad`.

The problem I see is that, with a PIN pad, the user needs to enter the old and new PIN twice, because OpenSC doesn't see/get the PIN in clear to send it to both applets. This may lead to a desynchronisation between the applets.

I don't see an easy solution: Either accept a possible desynchronisation or decline the use of a PIN-pad. How does the official middleware solve this problem?

src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,506 @@
...
+	chipdocit_ops.match_card = chipdocit_match_card;
+	chipdocit_ops.init = chipdocit_init;
+	chipdocit_ops.finish = chipdocit_finish;
+	chipdocit_ops.pin_cmd = chipdocit_pin_cmd;
+	chipdocit_ops.set_security_env = chipdocit_set_security_env;
+	chipdocit_ops.compute_signature = chipdocit_compute_signature;
+	return &chipdocit_drv;
+}
+
+#else /* ENABLE_SM && ENABLE_OPENPACE */
```
frankmorgner, line 498: Theoretically, you could also enable the driver even if OpenPACE is missing. perform_pace can also be used when the reader itself is capable of performing PACE. perform_pace will return SC_ERROR_NOT_SUPPORTED otherwise, which lets you know that both options (reader-PACE/OpenPACE) cannot be used.

src/libopensc/card-chipdocit.c
```diff
@@ -0,0 +1,506 @@
...
+		d.pin1 = *p1;
+		d.pin1.pad_char = 0xFF;
+		d.pin1.pad_length = 8;
+	}
+	if (p2) {
+		d.pin2 = *p2;
+		d.pin2.pad_char = 0xFF;
+		d.pin2.pad_length = 8;
+	}
+	return sc_get_iso7816_driver()->ops->pin_cmd(card, &d);
```
frankmorgner, line 265: Please check if you can define the properties for pin1 and pin2 directly in the PKCS#15 profile, i.e. in `add_pin()` in `pkcs15-chipdocit.c`. Normally, the PKCS#15 layer should redirect this information properly to pin_cmd

frankmorgner (commented):
@hirashix0 , could you please review and check if eoi still works as intended?

src/libopensc/card-chipdocit.c
ale-rinaldi: The official middleware, in pin-pad mode, changes only the primary (IAS) PIN and leaves the CNS-applet PIN untouched — it does not try to update the second applet when the PIN is entered on the pad. I have matched that: with `SC_PIN_CMD_USE_PINPAD` the driver now skips the CNS mirror and changes only the IAS PIN, while the software path keeps both PINs in sync. So no double entry and no attempt at an impossible replay. I don't have a pin-pad reader to exercise this end to end; the software path (both PINs in sync) is verified on hardware.

src/libopensc/card-chipdocit.c
ale-rinaldi: The `pad_char`/`pad_length` set here are for the secondary CNS-applet PIN (BSO 0x10), which is not exposed as a PKCS#15 object — it is an internal secret kept byte-synced with the user PIN and accessed in the clear. The primary (IAS) PIN is the PKCS#15 object defined in `add_pin()`, and it is sent unpadded, so the profile already drives its properties; only this in-clear mirror needs the fixed 0xFF padding.

src/libopensc/card-chipdocit.c
ale-rinaldi: Good idea. It is a bit larger than it looks from here though: the CAN unwrap shared with card-eoi (`chipdoc_decrypt_can_block`) currently lives under the same `ENABLE_SM && ENABLE_OPENPACE` guard, so enabling the driver without OpenPACE would also mean relaxing card-eoi's guards, and I have no reader-PACE-capable reader to exercise that path. Would you be OK with doing this as a follow-up, or would you prefer it in this PR?

src/libopensc/card-chipdocit.c
frankmorgner: thanks for the clarification

src/libopensc/card-chipdocit.c
frankmorgner: no, for now just keep it as is. This could be done later, but in a seperate PR toghether with the changes for eoi.

frankmorgner (approved):
From my side the changes look good now.

What I'd like to see would be some more documentation in the [OpenSC wiki](https://github.com/OpenSC/OpenSC/wiki/Italian-CNS-and-CIE) about the different types of Italien CNS and CIE. I still find the situation quite confusing why they have so many different implementations. This makes it even more important to add some user-digestable information to check whether the token she owns is actually supported by OpenSC (and to what extent).

ale-rinaldi:
@frankmorgner thank you for the feedback. At the moment, apart from this one for NXP ChipDoc, I have support ready for other 2 kinds of card (JSign4 and COSMO), I will open a separate PR for them when this one lands to avoid mixing too many things in a single PR.

I will happily update the wiki with a separate PR as well: unfortunately I don't have a really extensive knowledge of all the cards that are around, I know only the ones I personally own. However, as I happen to have access to other different cards, I'd love to continue improving support and documentation.

hirashix0:
> @hirashix0 , could you please review and check if eoi still works as intended?

The change looks fine, but I can't confirm it on a card — I don't have a Slovenian eOI. I guess @llogar would be the one to check that. It looks like it is a pure extract without behaviour changes.

I also noticed that `CHIPDOCIT_IAS_AID` is `A0 00 00 02 47 10 01`, which is the same as the ICAO eMRTD application AID (OpenSC calls those bytes `emrtd_aid` in `src/tools/npa-tool.c`). I tried it on an unrelated ICAO document here: the eMRTD AID returns `90 00`, while `CHIPDOCIT_SSCD_AID` and `CHIPDOCIT_CNS_AID` return `6A 82`. Just to double check and make sure the selection is not too broad. It doesn't look like the case I had in #3663, but worth checking.

ale-rinaldi:
Good catch, thanks. You're right that the IAS-ECC AID isn't a real discriminator: `A0 00 00 02 47 10 01` is also the ICAO eMRTD application AID and is present on other IAS-ECC cards, so it answers `90 00` on unrelated documents — on a card that happened to share the ATR the match would have been too broad.

I've tightened `chipdocit_match_card` so the matching is safer: it still selects the IAS-ECC application first (that step is needed to *activate* the card — the other applets aren't reachable until it is selected), but it now confirms the ChipDoc by then selecting the Bit4id SSCD application `E8 28 BD 08 0F 01 4E 58 50 31`, which — as you found — returns `6A 82` on an eMRTD. Verified on a real card: IAS → `90 00`, then SSCD → `90 00` (match); and the SSCD applet is only reachable after the IAS activation, hence the two steps. Pushed as a follow-up commit.

On the `eoi` part: agreed, it's a pure extract with no behaviour change, so I'll leave that one for @llogar to confirm on hardware as you suggested.

llogar:
Can confirm, eOI still works...

frankmorgner (approved):
The code looks good in principle.

Please review the comments code if all are needed. Keep them consice, don't re-iterate on what is already visible in the code. Thank you.

ale-rinaldi:
Done — went through the comments and trimmed the ones that just restated the code (APDU recipes, a duplicated activation note and an SM-teardown note), keeping only those that explain non-obvious card behaviour. Also confirmed clang-format is clean. Thanks for the review (and thanks @llogar for checking eOI).

frankmorgner (approved):

PR #3754, opened 2026-07-30
Add support for PQC in PIV driver

Jakuje:
This adds boilerplate to support new PQC keys in PIV driver (as well as in the opensc internals).

So far tested with Yubikey and ML-DSA. The ML-KEM still need some more wiring to support the new KEM operations throughout the code.

Opening as Draft as its incomplete, but it might be already useful for people to test it (regardless if they have a token supporting PQC, adding this into different drivers).

##### Checklist

- [ ] macOS token is tested


src/libopensc/pkcs15.h
```diff
@@ -219,6 +219,14 @@ struct sc_pkcs15_prkey_gostr3410 {
 	sc_pkcs15_bignum_t d;
 };
 
+struct sc_pkcs15_pubkey_pqc {
```
frankmorgner, line 222: I think I would prefer more specific names. PQC is what now everyone is looking at, but in 5 years from now, PQC will be a standard commodity that should apply to any new algorithm.

src/libopensc/pkcs15.h
```diff
@@ -219,6 +219,14 @@ struct sc_pkcs15_prkey_gostr3410 {
...
+	struct sc_pkcs15_u8 value;
+};
+
+struct sc_pkcs15_prkey_pqc {
+	struct sc_pkcs15_u8 value;
```
frankmorgner, line 227: Lattice based algorithms have the option of storing the private key in form of a seed, see e.g. https://docs.openssl.org/3.5/man7/EVP_PKEY-ML-DSA this may be reflected in the private key's format

src/libopensc/opensc.h
```diff
@@ -197,6 +200,28 @@ extern "C" {
 #define SC_ALGORITHM_EDDSA_RAW		0x00400000
 #define SC_ALGORITHM_XEDDSA_RAW		0x00800000
 
+
+#define SC_ALGORITHM_MLDSA_44		0x00100000
+#define SC_ALGORITHM_MLDSA_65		0x00200000
+#define SC_ALGORITHM_MLDSA_87		0x00400000
```
frankmorgner, line 206: For smart card use, I'm expecting some more variants to be used:
- Pre-Hashed ML-DSA (although being incompatible with standard ML-DSA)
- ML-DSA with pre-defined mu (as it is compatible with ML-DSA and avoids piping all data to the card)

src/libopensc/pkcs15-piv.c
```diff
@@ -902,7 +932,12 @@ static int sc_pkcs15emu_piv_init(sc_pkcs15_card_t *p15card)
 			case SC_ALGORITHM_EC:
 			case SC_ALGORITHM_EDDSA:
 			case SC_ALGORITHM_XEDDSA:
-				ckis[i].pubkey_len = cert_out->key->u.ec.params.field_length;
+			case SC_ALGORITHM_MLDSA:
+				if (cert_out->key->algorithm == SC_ALGORITHM_MLDSA) {
```
frankmorgner, line 936: AFAIK, NIST is working on adoption of ML-KEM/ML-DSA, but right now this is obviously a vender specific extension.

src/libopensc/pkcs15.h
Jakuje: Yes. The private keys at this moment are really just a placeholders as I did not implement enrollment. But for future, it make sense to be able to provide seed. I will add it.

src/libopensc/pkcs15.h
Jakuje: Good point. I was not able to come up with a better naming. If I would go with lattice, I would exclude the SLH-DSA (which has effective same format as the other lattice based from the current triple). I am not following the new algorithms being standardized, but I assume they will try to make the "API" similar to the existing ones (rather than inventing something as crazy as the RSA structures above even if it would involve some serialization of parameters and values).

src/libopensc/pkcs15-piv.c
Jakuje: Yes. But from the PIV changes needed, its really a just the algorithm ID away from the specs (which is the same for the EDDSA).

src/libopensc/opensc.h
Jakuje: Yes. I expect this will be coming in the future. I will need to figure out if this is something supported on the cards at this moment. I just verified the existing implementation is the pure ML-DSA variant. The pre-hashing will need some different communication with the card, which I did not find in specs so far.

Jakuje:
I moved the tool and test fixes to #3760 as this will likely take some more time to get finalized.

src/libopensc/pkcs15-piv.c
dengert: NIST defines the key references in sp800-78   The latest June 6 2026 draft is here: https://csrc.nist.gov/pubs/sp/800/73/pt1/6/iwd and points a github  and readable document.
It looks like the working draft just points out where changes will be needed to 800-73 and 800-78 but does not set the official key refs.

Hopefully "PIV like" card vendors will not define their own key refs, but wait for NIST to add these.

Jakuje:
Another part I ended up implementing in PIV driver was split into #3822. I will be splitting the p11test changes likely too eventually.

Jakuje:
I had to tweak the pkcs11-spy again to avoid crashes with softhsm. I can move more parts to separate PRs for reviewability later

PR #3761, opened 2026-08-03
p11test: fix AES mechanism addition

kunkku:
This PR fixes a regression introduced in commit c29ad3208acaf28b880153f9848025d5f362f3c8, where the `type` and `key_type` fields of `test_cert_t` were mixed up.


Jakuje (approved):
good catch! I guess we miss some secret key test coverage in here!

PR #3762, opened 2026-08-03
cardos: enable native RSA-PSS support on CardOS V5

janbrummer:
Use mechanism reference 0x33 (MSE SET AT, tag 0x80) for RSA-PSS, which was determined empirically with an Atos CardOS 5.3 card

##### Checklist

### DIFF
--- a/src/libopensc/card-cardos.c
+++ b/src/libopensc/card-cardos.c
@@ -60,6 +60,12 @@ static const struct sc_atr_table cardos_atrs[] = {
 	{ NULL, NULL, NULL, 0, 0, NULL }
 };
 
+/*
+ * Cryptographic mechanism reference (TLV tag 0x80) for RSA-PSS in the
+ * MSE SET AT command
+ */
+#define CARDOS_ALGO_RSA_PSS		0x33
+
 /* private data for cardos driver */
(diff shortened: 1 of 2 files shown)


src/libopensc/card-cardos.c
```diff
@@ -69,6 +75,7 @@ typedef struct cardos_data {
 	unsigned long ec_flags;
 	unsigned long ext_flags;
 	int rsa_2048;
+	u8 pss_mech_ref;
```
frankmorgner: why add a private field, when this byte is added unconditionally for PSS APDU data?

frankmorgner (changes requested):

src/libopensc/card-cardos.c
janbrummer: Good point, it's a left over from testing. I've already removed it locally, will push an update soon

janbrummer: Pushed.

janbrummer:
@frankmorgner CI should be good now

Jakuje (approved):
Verified it works with my CardOS card

Jakuje:
Thank you for your contribution and sorry for a delay!

PR #3764, opened 2026-08-06
feat: Add native driver for G&D StarSign CUT S (ICP-Brasil A3)

### LINKED ISSUE #2580: Giesecke & Devrient GmbH StarSign CUT S Unsupported card. Any chance to get it working in a next release ?
### Problem Description
```
kimera:~ # opensc-tool --list-readers --verbose
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Giesecke & Devrient GmbH StarSign CUT S [.] (03211015001257) 00 00
     3b:f9:96:00:00:81:31:fe:45:53:43:45:37:20:0e:00:20:20:28 Unsupported card 
kimera:~ # opensc-tool --serial
Using reader with a card: Giesecke & Devrient GmbH StarSign CUT S [.] (03211015001257) 00 00
sc_card_ctl(*, SC_CARDCTL_GET_SERIALNR, *) failed
kimera:~ # opensc-tool -n 
Using reader with a card: Giesecke & Devrient GmbH StarSign CUT S [.] (03211015001257) 00 00
Unsupported card
kimera:~ # opensc-tool -i  
OpenSC 0.22.0 [gcc  12.1.0]
Enabled features: locking zlib readline openssl pcsc(libpcsclite.so.1)
kimera:~ # opensc-tool --atr
Using reader with a card: Giesecke & Devrient GmbH StarSign CUT S [.] (03211015001257) 00 00
3b:f9:96:00:00:81:31:fe:45:53:43:45:37:20:0e:00:20:20:28
kimera:~ # pkcs15-tool --list-info
Using reader with a card: Giesecke & Devrient GmbH StarSign CUT S [.] (03211015001257) 00 00
```

And this [site](https://smartcard-atr.apdu.fr/parse?ATR=3b:f9:96:00:00:81:31:fe:45:53:43:45:37:20:0e:00:20:20:28
)  correctly identify the usb token based on atr command above
I appreciate any help.

### Proposed Resolution

none 

### Logs

```
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: new full-speed USB device number 10 using ehci-pci
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: New USB device found, idVendor=1059, idProduct=0019, bcdDevice= 1.00
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: Product: StarSign CUT S
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: Manufacturer: Giesecke & Devrient GmbH
Jul 23 10:02:05 kimera kernel: usb 1-1.4.4: SerialNumber: 03211015001257
```

DiegoRibeirodeSouza:
﻿This PR introduces a native driver (`card-starsign.c`) for the G&D StarSign CUT S token, heavily used in Brazil for ICP-Brasil A3 certificates (e.g., judicial systems).

Previously, these tokens required the buggy proprietary SafeSign middleware. During our reverse engineering, we discovered a few peculiarities that prevented OpenSC from natively communicating with the applet:

1. **Initialization Attestation:** The card requires a specific string (`I am A.E.T. Europe B.V. SafeSign or BlueX approved software.`) sent via `PUT DATA` upon reset to unlock complex commands.
2. **Virtual paths for certificates/data objects:** Certificate and data-object EFs are referenced by paths containing a fictitious `0x3FFF` placeholder (e.g. `3F00 3FFF 4302 05A0`). The card answers `SW 6A82` if `0x3FFF` is ever selected directly — it isn't a real file. The directory right after it (e.g. `4302`) is real and must be selected once (cached, since re-selecting it while already current also fails). The real file size is read from the card's own FCP response rather than assumed.
3. **Logical Channels:** It strictly requires operations to run on Channel 1 (`MANAGE CHANNEL`).
4. **Signature padding:** the card pads a raw hash it is handed with PKCS#1 v1.5, but does not prepend the DigestInfo ASN.1/OID header a standards-compliant signature requires. We advertise `SC_ALGORITHM_RSA_HASH_NONE` only, so OpenSC's own crypto layer builds the correct DigestInfo + PKCS#1 block in software before handing the card an already-padded blob for `SC_ALGORITHM_RSA_RAW`.

This driver allows the token to work out-of-the-box on Linux systems, circumventing the proprietary middleware.

> **Note to reviewers:** The full reverse engineering research, APDU traffic dumps, and documentation behind these discoveries are available in my companion repository: https://github.com/DiegoRibeirodeSouza/starsign-driver-en

Fixes #2580

##### Checklist
- [x] Documentation is added or updated (see companion repository linked above)
- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-starsign.c
+++ b/src/libopensc/card-starsign.c
@@ -1,0 +1,304 @@
+/*
+ * Support for G&D StarSign CUT S cards (A.E.T. Europe SafeSign)
+ *
+ * This driver performs the proprietary DRM handshake and logical
+ * channel selection required by the StarSign CUT S token.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <string.h>
+#include <stdlib.h>
+
+#include "internal.h"
+#include "asn1.h"
+#include "log.h"
+
+static const struct sc_atr_table starsign_atrs[] = {
+	{ "3B:F9:96:00:00:81:31:FE:45:53:43:45:37:20:0E:00:20:20:28", NULL, NULL, SC_CARD_TYPE_STARSIGN, 0, NULL },
+	{ NULL, NULL, NULL, 0, 0, NULL }
+};
+
+static struct sc_card_operations starsign_ops;
+static struct sc_card_driver starsign_drv = {
+	"G&D StarSign CUT S",
+	"starsign",
+	&starsign_ops,
+	NULL, 0, NULL
+};
+
+static const u8 starsign_drm_string[] = "I am A.E.T. Europe B.V. SafeSign or BlueX approved software.";
+
+static int starsign_match_card(sc_card_t *card)
+{
+	int i;
+	i = _sc_match_atr(card, starsign_atrs, &card->type);
+	if (i < 0)
+		return 0;
+	return 1;
+}
+
+static int starsign_init(sc_card_t *card)
+{
+	sc_apdu_t apdu;
+	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
+	int r;
+
+	card->name = "G&D StarSign CUT S";
+	card->type = SC_CARD_TYPE_STARSIGN;
+
+	/* 1. First DRM Handshake (ignore result, usually 6D 00) */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xDA, 0x01, 0x00);
+	apdu.data = starsign_drm_string;
+	apdu.datalen = sizeof(starsign_drm_string) - 1;
+	apdu.lc = apdu.datalen;
+	r = sc_transmit_apdu(card, &apdu);
+
+	/* 2. Select PKCS#15 AID (ignore result) */
+	u8 aid[] = { 0xA0, 0x00, 0x00, 0x00, 0x63, 0x50, 0x4B, 0x43, 0x53, 0x2D, 0x31, 0x35 };
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0xA4, 0x04, 0x00);
+	apdu.data = aid;
+	apdu.datalen = sizeof(aid);
+	apdu.lc = sizeof(aid);
+	apdu.le = 256;
+	apdu.resplen = sizeof(rbuf);
+	apdu.resp = rbuf;
+	r = sc_transmit_apdu(card, &apdu);
+
+	/* 3. Second DRM Handshake (ignore result) */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xDA, 0x01, 0x00);
+	apdu.data = starsign_drm_string;
+	apdu.datalen = sizeof(starsign_drm_string) - 1;
+	apdu.lc = apdu.datalen;
+	r = sc_transmit_apdu(card, &apdu);
+
+	/* 4. Open Logical Channel 1 */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, 0x70, 0x00, 0x00);
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+	apdu.le = 1;
+
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r, "Failed to transmit MANAGE CHANNEL");
+	if (apdu.sw1 == 0x6A && apdu.sw2 == 0x81) {
+		/* Channel already open, that's fine */
+		r = SC_SUCCESS;
+	} else {
+		r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		LOG_TEST_RET(card->ctx, r, "Card refused logical channel opening");
+	}
+
+	/* 3. Force CLA=0x01 for all subsequent operations */
+	card->cla = 0x01;
+
+	/* 5. Select PKCS#15 AID again on Channel 1 */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x00);
+	apdu.data = aid;
+	apdu.datalen = sizeof(aid);
+	apdu.lc = sizeof(aid);
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r, "Failed to select AID on channel 1");
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if (r < 0) {
+		sc_log(card->ctx, "Card refused AID selection on channel 1 (SW %02X %02X), ignoring as it might be already selected", apdu.sw1, apdu.sw2);
+	}
+
+	/* Force RAW RSA (software padding) because G&D StarSign CUT cards 
+           often return 67 00 if sent unpadded hashes during C_Sign.
+           We only set SC_ALGORITHM_RSA_RAW so OpenSC pads the data to 256 bytes itself. */
+        card->caps |= SC_CARD_CAP_APDU_EXT;
+        card->max_send_size = 2048;
+        card->max_recv_size = 2048;
+        unsigned long alg_flags = SC_ALGORITHM_RSA_RAW | SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE | SC_ALGORITHM_RSA_HASH_MD5 | SC_ALGORITHM_RSA_HASH_SHA1 | SC_ALGORITHM_RSA_HASH_SHA256;
+				  
+	_sc_card_add_rsa_alg(card, 1024, alg_flags, 0);
+	_sc_card_add_rsa_alg(card, 2048, alg_flags, 0);
+	_sc_card_add_rsa_alg(card, 4096, alg_flags, 0);
+
+	return SC_SUCCESS;
+}
+
+static int starsign_select_file(sc_card_t *card, const sc_path_t *in_path, sc_file_t **file_out)
+{
+	int r = SC_SUCCESS;
+	size_t i;
+	struct sc_apdu apdu;
+
+	if (in_path->type != SC_PATH_TYPE_PATH && in_path->type != SC_PATH_TYPE_FROM_CURRENT && in_path->type != SC_PATH_TYPE_FILE_ID) {
+		sc_log(card->ctx, "STARSIGN SELECT FILE DEFERRING type=%d", in_path->type);
+		return sc_get_iso7816_driver()->ops->select_file(card, in_path, file_out);
+	}
+
+	sc_log(card->ctx, "STARSIGN SELECT FILE EXECUTING type=%d len=%zu", in_path->type, in_path->len);
+
+	if (in_path->len % 2 != 0 || in_path->len == 0)
+		return SC_ERROR_INVALID_ARGUMENTS;
+
+	/* First, try to select the full path at once if it's a full path from MF */
+	if (in_path->len >= 4 && in_path->value[0] == 0x3F && in_path->value[1] == 0x00) {
+		sc_log(card->ctx, "STARSIGN SELECT FILE: Attempting full path selection from MF");
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x08, 0x0C);
+		apdu.data = in_path->value;
+		apdu.datalen = in_path->len;
+		apdu.lc = in_path->len;
+		r = sc_transmit_apdu(card, &apdu);
+		if (r == 0) r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		if (r == 0) {
+			sc_log(card->ctx, "STARSIGN SELECT FILE: Full path selection successful");
+			return 0;
+		}
+		sc_log(card->ctx, "STARSIGN SELECT FILE: Full path selection failed (%d), falling back to component-by-component", r);
+	}
+
+	/* Loop over every 2 bytes of the path */
+	for (i = 0; i < in_path->len; i += 2) {
+		unsigned short fid = (in_path->value[i] << 8) | in_path->value[i + 1];
+
+		/* DO NOT ignore 3F00! We need to select it to reset the current DF to MF.
+		   HOWEVER, G&D StarSign tokens have a bug where some certificates' absolute paths
+		   omit the 5015 DF (e.g. 3F00 3FFF 4302 13AE). If we just select 3F00, the 
+		   subsequent selection of 4302 will fail.
+		   We MUST explicitly select 5015 after 3F00, or we can just ignore 3F00 entirely
+		   AND ignore 5015 entirely, relying on the fact that we are ALREADY in the 5015
+		   applet, BUT we must reset to 5015 to prevent relative selection bugs.
+		   
+		   The safest fix is: when we see 3F00, we select 3F00 AND THEN select 5015 automatically! */
+		if (fid == 0x3F00) {
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
+			apdu.data = (const u8 *)"\x3F\x00";
+			apdu.datalen = 2; apdu.lc = 2;
+			sc_transmit_apdu(card, &apdu);
+			
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
+			apdu.data = (const u8 *)"\x50\x15";
+			apdu.datalen = 2; apdu.lc = 2;
+			sc_transmit_apdu(card, &apdu);
+			continue;
+		}
+		/* Ignore virtual/intermediate DFs that are not selectable directly on G&D StarSign */
+		if (fid == 0x3FFF) {
+			sc_log(card->ctx, "STARSIGN SELECT FILE: IGNORING virtual DF %04X", fid);
+			continue;
+		}
+
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C); /* P2=0x0C means No response/FCI expected */
+		apdu.data = &in_path->value[i];
+		apdu.datalen = 2;
+		apdu.lc = 2;
+
+		r = sc_transmit_apdu(card, &apdu);
+		if (r < 0) return r;
+		r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		
+		if (r == SC_ERROR_FILE_NOT_FOUND) {
+			sc_log(card->ctx, "STARSIGN SELECT FILE: P1=0x00 failed, retrying with P1=0x01 (Select DF) for FID %04X", fid);
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x01, 0x0C);
+			apdu.data = &in_path->value[i];
+			apdu.datalen = 2;
+			apdu.lc = 2;
+
+			r = sc_transmit_apdu(card, &apdu);
+			if (r < 0) return r;
+			r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		}
+
+		if (r == SC_ERROR_FILE_NOT_FOUND) {
+			sc_log(card->ctx, "STARSIGN SELECT FILE: P1=0x01 failed, retrying with P1=0x02 (Select EF) for FID %04X", fid);
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x02, 0x0C);
+			apdu.data = &in_path->value[i];
+			apdu.datalen = 2;
+			apdu.lc = 2;
+
+			r = sc_transmit_apdu(card, &apdu);
+			if (r < 0) return r;
+			r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		}
+
+		if (r < 0) return r;
+	}
+
+	if (file_out) {
+		sc_file_t *file = sc_file_new();
+		if (!file)
+			return SC_ERROR_OUT_OF_MEMORY;
+		file->id = (in_path->value[in_path->len - 2] << 8) | in_path->value[in_path->len - 1];
+		file->type = SC_FILE_TYPE_WORKING_EF;
+		file->ef_structure = SC_FILE_EF_TRANSPARENT;
+		file->size = 0x8000; /* Dummy size so OpenSC reads until EOF */
+		file->magic = SC_FILE_MAGIC;
+		*file_out = file;
+	}
+	return SC_SUCCESS;
+}
+
+static int starsign_set_security_env(sc_card_t *card, const sc_security_env_t *env, int res)
+{
+        sc_apdu_t apdu;
+        u8 data[6];
+        int r;
+
+        sc_log(card->ctx, "STARSIGN: set_security_env called! Operation: %d", env->operation);
+
+        data[0] = 0x84;
+        data[1] = 0x01;
+        data[2] = 0x01;
+        data[3] = 0x80;
+        data[4] = 0x01;
+        data[5] = 0x02;
+
+        sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x22, 0x41, 0xB6);
+        apdu.data = data;
+        apdu.datalen = 6;
+        apdu.lc = 6;
+        apdu.cla = card->cla;
+
+        r = sc_transmit_apdu(card, &apdu);
+        if (r) return r;
+        return sc_check_sw(card, apdu.sw1, apdu.sw2);
+}
+
+
+static int starsign_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data)
+{
+	if (data->cmd == SC_PIN_CMD_VERIFY) {
+		u8 pin_buf[15];
+		sc_apdu_t apdu;
+		int r;
+
+		if (data->pin1.len > 15)
+			return SC_ERROR_INVALID_PIN_LENGTH;
+
+		memset(pin_buf, 0x00, sizeof(pin_buf));
+		memcpy(pin_buf, data->pin1.data, data->pin1.len);
+
+		/* Force P2=02, padded to 15 bytes */
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x20, 0x00, 0x02);
+		apdu.data = pin_buf;
+		apdu.datalen = 15;
+		apdu.lc = 15;
+        /* Inherit card->cla (which we set to 0x01 in init) */
+        apdu.cla = card->cla;
+
+		r = sc_transmit_apdu(card, &apdu);
+		LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+		/* Save tries_left if we want to handle it (via data block maybe?), but not needed for basic verify here. */
+		return sc_check_sw(card, apdu.sw1, apdu.sw2);
+	}
+	
+	/* Fallback to default behavior for other PIN commands (unblock, etc.) */
+	return sc_get_iso7816_driver()->ops->pin_cmd(card, data);
+}
+
+struct sc_card_driver * sc_get_starsign_driver(void)
+{
+	struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
+	starsign_ops = *iso_drv->ops;
+	starsign_ops.init = starsign_init;
+	starsign_ops.select_file = starsign_select_file;
+	starsign_ops.match_card = starsign_match_card;
+	starsign_ops.pin_cmd = starsign_pin_cmd;
+        starsign_ops.set_security_env = starsign_set_security_env;
+	return &starsign_drv;
+}
--- a/src/libopensc/card-starsign.c
+++ b/src/libopensc/card-starsign.c
@@ -335,10 +335,38 @@ starsign_select_file(sc_card_t *card, const sc_path_t *in_path, sc_file_t **file
 		 * initial "3F00 5031" DF transition and for any top-level
 		 * DF/EF reached before that context has been established.
 		 * The card only accepts P1=0x00 (select by file ID) here --
-		 * it answers SW 6A82 for P1=0x01 ("select child DF"). */
-		for (i = 0; i < in_path->len; i += 2) {
+		 * it answers SW 6A82 for P1=0x01 ("select child DF"). Walk
+		 * every component except the last one this way; DF-only
+		 * selects (P2=0x0C) never return an FCI on this card, which
+		 * is fine for intermediate directories. */
+		size_t last = in_path->len - 2;
+
+		for (i = 0; i < last; i += 2) {
 			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
 			apdu.data = &in_path->value[i];
+			apdu.datalen = 2;
+			apdu.lc = 2;
+
+			r = sc_transmit_apdu(card, &apdu);
+			LOG_TEST_GOTO_ERR(ctx, r, "APDU transmit failed (hierarchical SELECT)");
+			r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+			LOG_TEST_GOTO_ERR(ctx, r, "Hierarchical SELECT returned an error status word");
+		}
+
+		/* The final component: a bare "MF + one FID" path (len == 4,
+		 * e.g. "3F00 5031") is always a DF transition -- select it
+		 * the same way as the rest. Anything deeper always bottoms
+		 * out at a real EF on this card's layout, so select it as a
+		 * child EF instead (P1=0x02) to actually capture its FCI/real
+		 * size via starsign_select_ef_child(), instead of silently
+		 * falling back to the placeholder file->size below. */
+		if (in_path->len > 4) {
+			r = starsign_select_ef_child(card, &in_path->value[last], file);
+			LOG_TEST_GOTO_ERR(ctx, r, "SELECT of the final EF (hierarchical path) failed");
+			selected_as_ef = 1;
+		} else {
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
+			apdu.data = &in_path->value[last];
 			apdu.datalen = 2;
 			apdu.lc = 2;
 
--- a/src/libopensc/card-starsign.c
+++ b/src/libopensc/card-starsign.c
@@ -174,6 +174,23 @@ starsign_init(sc_card_t *card)
 	card->caps |= SC_CARD_CAP_APDU_EXT;
 	card->max_send_size = 2048;
 	card->max_recv_size = 256;
+	/*
+	 * KNOWN HARDWARE LIMITATION (not fixable here): on at least one
+	 * StarSign CUT S unit, the token's built-in CCID reader declares a
+	 * firmware-fixed dwMaxCCIDMsgLen of 271 bytes (~261 usable after the
+	 * CCID header) in its USB descriptor -- see `lsusb -v` -- and does
+	 * not advertise "Extended APDU level exchange" in dwFeatures. A raw
+	 * RSA-2048 operation's 256-byte payload does not fit under that
+	 * ceiling in any standard APDU encoding (even the leanest extended
+	 * Case 3 APDU is 263 bytes), and this card separately rejects ISO
+	 * 7816-4 command chaining as a workaround (SW 6E 00). The result is
+	 * intermittent SCardTransmit failures (SCARD_E_INVALID_PARAMETER) on
+	 * raw sign/decipher through that specific reader, typically
+	 * succeeding on retry. We have no software fix for this: the only
+	 * way to avoid the 256-byte payload is the card hashing and padding
+	 * on-chip with a correct DigestInfo, which this hardware does not do
+	 * (see the SC_ALGORITHM_RSA_HASH_NONE comment above).
+	 */
 	alg_flags = SC_ALGORITHM_RSA_RAW | SC_ALGORITHM_RSA_PAD_PKCS1 | SC_ALGORITHM_RSA_HASH_NONE;
 
 	_sc_card_add_rsa_alg(card, 1024, alg_flags, 0);
--- a/src/libopensc/card-starsign.c
+++ b/src/libopensc/card-starsign.c
@@ -263,14 +263,13 @@ static int starsign_set_security_env(sc_card_t *card, const sc_security_env_t *e
 static int starsign_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data)
 {
 	if (data->cmd == SC_PIN_CMD_VERIFY) {
-		u8 pin_buf[15];
+		u8 pin_buf[15] = {0};
 		sc_apdu_t apdu;
 		int r;
 
 		if (data->pin1.len > 15)
 			return SC_ERROR_INVALID_PIN_LENGTH;
 
-		memset(pin_buf, 0x00, sizeof(pin_buf));
 		memcpy(pin_buf, data->pin1.data, data->pin1.len);
 
 		/* Force P2=02, padded to 15 bytes */
--- a/src/libopensc/card-starsign.c
+++ b/src/libopensc/card-starsign.c
@@ -1,8 +1,24 @@
 /*
- * Support for G&D StarSign CUT S cards (A.E.T. Europe SafeSign)
+ * card-starsign.c: Support for G&D StarSign CUT S cards (A.E.T. Europe SafeSign)
+ *
+ * Copyright (C) 2026 Diego Ribeiro de Souza
  *
  * This driver performs the proprietary DRM handshake and logical
  * channel selection required by the StarSign CUT S token.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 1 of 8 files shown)


frankmorgner:
please add the checkmark information that is required for PR

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+
+static struct sc_card_operations starsign_ops;
+static struct sc_card_driver starsign_drv = {
+	"G&D StarSign CUT S",
+	"starsign",
+	&starsign_ops,
+	NULL, 0, NULL
+};
+
+static const u8 starsign_drm_string[] = "I am A.E.T. Europe B.V. SafeSign or BlueX approved software.";
```
frankmorgner, line 48: I wonder how vendors constantly come up with such bollocks 🙃

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+
+	card->name = "G&D StarSign CUT S";
+	card->type = SC_CARD_TYPE_STARSIGN;
+
+	/* 1. First DRM Handshake (ignore result, usually 6D 00) */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xDA, 0x01, 0x00);
+	apdu.data = starsign_drm_string;
+	apdu.datalen = sizeof(starsign_drm_string) - 1;
+	apdu.lc = apdu.datalen;
+	r = sc_transmit_apdu(card, &apdu);
```
frankmorgner, line 108: r should be `SC_SUCCESS` nevertheless to catch transmission errors

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+	/* 5. Select PKCS#15 AID again on Channel 1 */
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x04, 0x00);
+	apdu.data = aid;
+	apdu.datalen = sizeof(aid);
+	apdu.lc = sizeof(aid);
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r, "Failed to select AID on channel 1");
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if (r < 0) {
+		sc_log(card->ctx, "Card refused AID selection on channel 1 (SW %02X %02X), ignoring as it might be already selected", apdu.sw1, apdu.sw2);
```
frankmorgner: if it is already selected, check explicitly for the related SW instead of assuming success on failure

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	/* Force RAW RSA (software padding) because G&D StarSign CUT cards 
+           often return 67 00 if sent unpadded hashes during C_Sign.
+           We only set SC_ALGORITHM_RSA_RAW so OpenSC pads the data to 256 bytes itself. */
+        card->caps |= SC_CARD_CAP_APDU_EXT;
```
frankmorgner: indenting

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+	apdu.lc = sizeof(aid);
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r, "Failed to select AID on channel 1");
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if (r < 0) {
+		sc_log(card->ctx, "Card refused AID selection on channel 1 (SW %02X %02X), ignoring as it might be already selected", apdu.sw1, apdu.sw2);
+	}
+
+	/* Force RAW RSA (software padding) because G&D StarSign CUT cards 
+           often return 67 00 if sent unpadded hashes during C_Sign.
```
frankmorgner: what does it mean that it **often** returns 6700? Typically, cards are working deterministically...

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+
+	/* First, try to select the full path at once if it's a full path from MF */
+	if (in_path->len >= 4 && in_path->value[0] == 0x3F && in_path->value[1] == 0x00) {
+		sc_log(card->ctx, "STARSIGN SELECT FILE: Attempting full path selection from MF");
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x08, 0x0C);
+		apdu.data = in_path->value;
+		apdu.datalen = in_path->len;
+		apdu.lc = in_path->len;
+		r = sc_transmit_apdu(card, &apdu);
+		if (r == 0) r = sc_check_sw(card, apdu.sw1, apdu.sw2);
```
frankmorgner: code style. `0` is `SC_SUCCESS`

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+		if (fid == 0x3F00) {
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
+			apdu.data = (const u8 *)"\x3F\x00";
+			apdu.datalen = 2; apdu.lc = 2;
+			sc_transmit_apdu(card, &apdu);
+			
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x0C);
+			apdu.data = (const u8 *)"\x50\x15";
+			apdu.datalen = 2; apdu.lc = 2;
+			sc_transmit_apdu(card, &apdu);
```
frankmorgner: This also allows a path such as `3f00 3f00`, which is typically forbidden? Is that intended?

frankmorgner: please add error handling

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+		apdu.data = &in_path->value[i];
+		apdu.datalen = 2;
+		apdu.lc = 2;
+
+		r = sc_transmit_apdu(card, &apdu);
+		if (r < 0) return r;
+		r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+		
+		if (r == SC_ERROR_FILE_NOT_FOUND) {
+			sc_log(card->ctx, "STARSIGN SELECT FILE: P1=0x00 failed, retrying with P1=0x01 (Select DF) for FID %04X", fid);
```
frankmorgner: you can avoid an errornous select if you check the path length. If more than two bytes are left over, then the next FID must belong to an DF rather than to a normal file.

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	if (file_out) {
+		sc_file_t *file = sc_file_new();
+		if (!file)
+			return SC_ERROR_OUT_OF_MEMORY;
+		file->id = (in_path->value[in_path->len - 2] << 8) | in_path->value[in_path->len - 1];
+		file->type = SC_FILE_TYPE_WORKING_EF;
+		file->ef_structure = SC_FILE_EF_TRANSPARENT;
+		file->size = 0x8000; /* Dummy size so OpenSC reads until EOF */
```
frankmorgner: file size of `0` is used for unknown values. Please check where exactly such a big size is required, that might be a limitation of OpenSC core code.

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+
+		if (r < 0) return r;
+	}
+
+	if (file_out) {
+		sc_file_t *file = sc_file_new();
+		if (!file)
+			return SC_ERROR_OUT_OF_MEMORY;
+		file->id = (in_path->value[in_path->len - 2] << 8) | in_path->value[in_path->len - 1];
+		file->type = SC_FILE_TYPE_WORKING_EF;
```
frankmorgner: If you previously selected a DF, then this file type is wrong.

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+		file->ef_structure = SC_FILE_EF_TRANSPARENT;
+		file->size = 0x8000; /* Dummy size so OpenSC reads until EOF */
+		file->magic = SC_FILE_MAGIC;
```
frankmorgner: do you want to add the path?

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+		int r;
+
+		if (data->pin1.len > 15)
+			return SC_ERROR_INVALID_PIN_LENGTH;
+
+		memset(pin_buf, 0x00, sizeof(pin_buf));
+		memcpy(pin_buf, data->pin1.data, data->pin1.len);
+
+		/* Force P2=02, padded to 15 bytes */
+		sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x20, 0x00, 0x02);
```
frankmorgner: You should better set the PIN reference to `0x02` in the PKCS#15 profile, which the iso7816 driver uses to build the APDU. Also, the PKCS#15 profile allows you setting PIN length restrictions.

frankmorgner (changes requested):
Happy to hear that you made the card work. However, the PR is in very rough shape. Please have a look at other card drivers, e.g. iso7816.c what macros and coding paradigms they are using. Also, please have a look at other PKCS#15 drivers that are bootstrapping the card profile from existing information in order to fix some missing bit, e.g. pkcs15-lteid.c

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+
+struct sc_card_driver * sc_get_starsign_driver(void)
+{
+	struct sc_card_driver *iso_drv = sc_get_iso7816_driver();
+	starsign_ops = *iso_drv->ops;
+	starsign_ops.init = starsign_init;
+	starsign_ops.select_file = starsign_select_file;
+	starsign_ops.match_card = starsign_match_card;
+	starsign_ops.pin_cmd = starsign_pin_cmd;
+        starsign_ops.set_security_env = starsign_set_security_env;
```
Jakuje: nit: format
```suggestion
	starsign_ops.set_security_env = starsign_set_security_env;
```

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+{
+	if (data->cmd == SC_PIN_CMD_VERIFY) {
+		u8 pin_buf[15];
+		sc_apdu_t apdu;
+		int r;
+
+		if (data->pin1.len > 15)
+			return SC_ERROR_INVALID_PIN_LENGTH;
+
+		memset(pin_buf, 0x00, sizeof(pin_buf));
```
Jakuje: please, use standard initializers instead of memset

```suggestion
		u8 pin_buf[15] = {0};
		sc_apdu_t apdu;
		int r;

		if (data->pin1.len > 15)
			return SC_ERROR_INVALID_PIN_LENGTH;

```

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,304 @@
...
+        sc_log(card->ctx, "STARSIGN: set_security_env called! Operation: %d", env->operation);
+
+        data[0] = 0x84;
+        data[1] = 0x01;
+        data[2] = 0x01;
+        data[3] = 0x80;
+        data[4] = 0x01;
+        data[5] = 0x02;
+
+        sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x22, 0x41, 0xB6);
```
Jakuje: Can you use the `sc_format_apdu_ex()` which allows you setting the CLA directly?

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Tell me about it! 😂 It took us days of capturing APDU logs from the proprietary Windows driver just to figure out this exact byte sequence. If the Key Reference doesn't strictly precede the Algorithm Reference in this exact format, the chip flat out rejects the signature.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: You're right. I will add a proper check for r == SC_SUCCESS to correctly catch and handle any APDU transmission failures.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: You are right, the card is deterministic. We initially observed 67 00 (Wrong Length) when the card was fed unpadded hashes, because it strictly expects exactly 256 bytes for a 2048-bit key. However, we recently discovered that by injecting the SC_ALGORITHM_RSA_HASH_SHA256 flag, the card does successfully accept a 32-byte hash and performs the PKCS#1 padding on-card. I will reword this comment and update the implementation to accurately reflect the hardware behavior.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Good point. I will explicitly check for the specific SW (e.g., 90 00 or the specific 'already selected' warning) instead of indiscriminately ignoring all errors.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: That's a very helpful tip! I will update the logic to check the remaining path length to differentiate between DF and EF selections, avoiding the erroneous blind select.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Thank you for the guidance! Moving this to the PKCS#15 profile layer will allow us to drop the hardcoded PIN logic here entirely. I will look into pkcs15-lteid.c and others to implement this properly via the profile.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Will fix the formatting.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Correction to my earlier reply here: I was wrong. We decrypted a live signature produced this way with the token's own public key, and the padded block contains the bare SHA-256 digest with **no DigestInfo ASN.1/OID header** — so every signature the card produced via `SC_ALGORITHM_RSA_HASH_SHA256` returned `SW 90 00` but was cryptographically invalid (fails standard PKCS#1 verification). Fixed in the latest commit: we now advertise `SC_ALGORITHM_RSA_HASH_NONE` only, so OpenSC's own crypto layer builds the DigestInfo + PKCS#1 padding in software before handing the card an already-padded blob for `SC_ALGORITHM_RSA_RAW`. Verified end-to-end: PDFs signed this way now validate as "Signature is Valid" under `pdfsig` (Poppler).

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Fixed — reformatted the whole file with `clang-format -style=file` per the project's `.clang-format`.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Fixed — replaced the literal `0`/`< 0` comparisons with `SC_SUCCESS` throughout.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Not intended — that was a side effect of the old ad-hoc logic that injected `5015` automatically whenever it saw `0x3F00` anywhere in the path. Replaced with an explicit state machine: the leading `3F00` is only special-cased as the first two bytes of a genuine top-level MF→DF transition; nothing auto-injects a DF selection anymore, so this can't happen in the new implementation.

DiegoRibeirodeSouza: Fixed — every `sc_transmit_apdu`/`sc_check_sw` call now checks its return value via `LOG_TEST_RET` before proceeding.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Implemented in the latest commit, together with a proper fix for the `0x3FFF` placeholder path issue below.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: You were right to flag this — the `0x8000` placeholder was a hack. Fixed: the driver now parses the real size from the card's own FCP response (tag `0x80`) on every SELECT and uses that, falling back to a small generous default only when the card doesn't return one.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Fixed — `file->type` is now set based on whether the last SELECT actually landed on a DF (multi-component hierarchical select) or an EF (direct child select), instead of being hardcoded to `SC_FILE_TYPE_WORKING_EF`.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Yes — fixed, `file_out` now sets `file->path = *in_path`.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: This whole function (`starsign_pin_cmd`) is gone — per your suggestion below about the PKCS#15 profile, the PIN reference/length/pad-char are now injected via a `parse_df` hook in a new `pkcs15-starsign.c`, modeled on `pkcs15-lteid.c`, so the driver relies on the standard `iso7816` `pin_cmd()` unmodified.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Done — `set_security_env` now uses `sc_format_apdu_ex(&apdu, card->cla, 0x22, 0x41, p2, sbuf, sizeof(sbuf), NULL, 0)` instead of formatting the APDU and then overriding `apdu.cla` afterward.

DiegoRibeirodeSouza:
Pushed a new commit addressing all outstanding review feedback:

- Reworked `select_file` around the real card behavior for the `0x3FFF`-placeholder virtual paths (certificates/data objects), confirmed against a genuine SafeSign 4.7 APDU capture — no more ad-hoc `5015` auto-injection or triple-P1 retry.
- Real file sizes read from the card's FCP response instead of a hardcoded placeholder.
- Moved the PIN reference out of a hardcoded `pin_cmd` override into a `pkcs15-starsign.c` PKCS#15 bootstrap hook, following `pkcs15-lteid.c`.
- Corrected a real bug found via further testing: the card omits the `DigestInfo` header when padding on-card, so `SC_ALGORITHM_RSA_HASH_SHA256` produced signatures that returned `SW 90 00` but were cryptographically invalid. Now only `SC_ALGORITHM_RSA_HASH_NONE` is advertised.
- Added the missing `SC_CARD_TYPE_STARSIGN` value to `cards.h` — its absence meant this driver never actually compiled, which explains the CI build failures on this branch.
- Reformatted with `clang-format -style=file`.

Verified end-to-end against physical hardware: all 4 certificates on the token now read correctly (including the personal signing certificate, which previously failed ASN.1 decoding), and PDFs signed through this build validate as "Signature is Valid" under `pdfsig`.

DiegoRibeirodeSouza:
Updated the checklist and the PR description to match the actual current state of the driver (the technical description was stale from the initial submission). Windows minidriver and macOS token are left unchecked — I don't have access to that hardware/OS to test them, only a Linux box with the physical StarSign CUT S token. Happy to update those if someone with access can confirm, or if there's a way to validate the minidriver logic without a live token.

DiegoRibeirodeSouza:
Pushed one more small fix (`60127e6`) found while exercising the driver's `decipher` path (used by a separate hybrid-encryption tool, not part of this PR's judicial/signing use case): `set_security_env` was raising `max_recv_size` to 2048 alongside `max_send_size`, but `iso7816_fixup_transceive_length()` only clamps `apdu.le` *down* to `max_recv_size`, never up. Some PKCS#11 framework callers -- `decipher` in particular, which allocates a generic 512-byte response buffer regardless of actual key size -- request a `Le` far larger than the real RSA-2048 output (256 bytes), and with `max_recv_size` raised to 2048 that oversized `Le` was never clamped, producing an extended APDU the reader's USB/PC-SC transport could not actually deliver.

Keeping `max_recv_size` at 256 (the real 2048-bit RSA output size) fixes that. This is a correctness improvement, not a response to review feedback -- it doesn't touch anything already discussed here, and cannot affect signing (whose `Le` is always <= 256, so it was never subject to the clamp either way).

DiegoRibeirodeSouza:
Documenting a genuine hardware limitation found while stress-testing raw RSA-2048 sign/decipher on physical hardware (pushed as a code comment in `60127e6`/`8279abb`, not a code change — there is no software fix):

The token's *built-in* CCID reader declares a firmware-fixed message ceiling in its USB descriptor:

```
dwMaxCCIDMsgLen   271        (~261 usable bytes after the CCID header)
dwFeatures        ...        "Short APDU level exchange" only —
                              no "Extended APDU level exchange"
dwProtocols       2  T=1
```

`pcscd` surfaces this directly:
```
CmdXfrBlockTPDU_T0() Command too long (265 bytes) for max: 261 bytes
```

A raw RSA-2048 operation needs to move 256 bytes of payload in a single APDU. Even the leanest possible extended encoding (Case 3, no immediate response requested: 4-byte header + 1-byte extended marker + 2-byte extended Lc + 256 bytes of data) comes to 263 bytes — 2 bytes over that ceiling, unavoidably, regardless of mechanism (OAEP/PKCS1/raw all move the same 256-byte block). ISO 7816-4 command chaining, the usual workaround for an oversized payload, is rejected by this card outright (`SW 6E 00`, "Class not supported").

Net effect: on this specific reader, raw sign/decipher operations can fail intermittently with `SCARD_E_INVALID_PARAMETER` right at that byte boundary, and typically succeed on retry. This isn't something `card-starsign.c` can work around — the only way to avoid moving the full 256-byte block would be the card hashing+padding on-chip with a correct DigestInfo (so only a 32-byte hash crosses the wire), and this hardware's on-chip padding is missing that header (see the earlier `SC_ALGORITHM_RSA_HASH_NONE` fix/comment).

Flagging this so it's on record for anyone hitting intermittent transmit failures on this token/reader combination — it's a firmware ceiling of the CCID controller, not a driver bug.

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,448 @@
...
+	case SC_SEC_OPERATION_SIGN:
+		p2 = 0xB6;
+		break;
+	default:
+		LOG_FUNC_RETURN(ctx, SC_ERROR_INVALID_ARGUMENTS);
+	}
+
+	sc_format_apdu_ex(&apdu, card->cla, 0x22, 0x41, p2, sbuf, sizeof(sbuf), NULL, 0);
+
+	sc_log(ctx, "MSE SET constructed payload=84 01 %02X 80 01 %02X", sbuf[2], sbuf[5]);
```
frankmorgner, line 457: you're not altering the bytes 2 and 5 so that's static. You could, however, use 
```
	if (env->flags & SC_SEC_ENV_ALG_REF_PRESENT) {
		sbuf[5] = env->algorithm_ref & 0xFF;
	}
	if (env->flags & SC_SEC_ENV_KEY_REF_PRESENT) {
		sbuf[2] = env->key_ref_len & 0xFF;
	}
```
But I don't know if that changes anything since your profile doesn't change in that regard

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,448 @@
+/*
+ * Support for G&D StarSign CUT S cards (A.E.T. Europe SafeSign)
+ *
+ * This driver performs the proprietary DRM handshake and logical
+ * channel selection required by the StarSign CUT S token.
+ */
```
frankmorgner, line 22: license header missing

src/libopensc/card-starsign.c
```diff
@@ -0,0 +1,448 @@
...
+	_sc_card_add_rsa_alg(card, 4096, alg_flags, 0);
+
+	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
+}
+
+/* Extract the file size from an FCP response of the form
+ * 6F <len> 80 02 <size-hi> <size-lo> ... (tag 0x80 = number of data bytes).
+ * Returns 0 if the tag is not present or the buffer is too short. */
+static size_t
+starsign_parse_fcp_size(const u8 *buf, size_t len)
```
frankmorgner: would it be possible to use `iso_ops->process_fci(card, file, buf, len)`?

src/libopensc/pkcs15-starsign.c
```diff
@@ -0,0 +1,93 @@
+/*
+ * Support for G&D StarSign CUT S tokens.
+ * This handles PKCS#15 bootstrap quirks, specifically injecting
+ * the correct PIN reference which the token's AODF omits.
+ */
```
frankmorgner, line 34: license header missing

src/libopensc/pkcs15-starsign.c
```diff
@@ -0,0 +1,93 @@
...
+					prkey_info->key_reference = 0x01;
+				}
+				/* Force modulus length if 0 */
+				if (prkey_info->modulus_length == 0) {
+					sc_log(ctx, "Patching PRKEY modulus length to 2048");
+					prkey_info->modulus_length = 2048;
+				}
+			}
+		}
+	}
```
frankmorgner, line 178: your code for path SELECTion looks very convoluted and I have the impression that this is due to the PKCS#15 profile supplying confusing information. One option would be to fix this path information in the profile so that special cases are avoided.

But if the card's profile is so broken, it may be simpler as second option to just supply the card's static profile as OpenSC internal data. See `src/libopensc/pkcs15-esteid2025.c` for such an example. Metadata may be filled with what you have already gathered from the existing profile; paths should be chosen so that they can directly be selected, e.g. with the iso driver (note that you have the option to supply both, and AID *and* a (relative) path with sc_path_t. both are handled in e.g. `iso7816.c`)

My main concern here is that the selection logic doesn't look like it can be maintained in the long run.

frankmorgner (changes requested):

DiegoRibeirodeSouza:
Windows minidriver + PKCS#11 module tested end-to-end today, checking off the remaining item on the checklist. Had to go fetch an old notebook from my mother's house to actually have a Windows machine to test on — but it paid off. Full report:

### Build

Followed the official win32 build path (MSVC + `nmake`, dependencies via `vcpkg` for OpenSSL/zlib, CPDK for the minidriver, matching `.github/workflows/windows.yml`).

**Found and fixed a real bug along the way:** `card-starsign.obj` and `pkcs15-starsign.obj` were never added to the `OBJECTS` list in `src/libopensc/Makefile.mak` (the win32/nmake makefile — separate from `Makefile.am`, which does have them). Concretely, `opensc.dll` failed to link with:

```
ctx.obj : error LNK2001: unresolved external symbol sc_get_starsign_driver
pkcs15-syn.obj : error LNK2001: unresolved external symbol sc_pkcs15emu_starsign_init_ex
```

So on `master` as it stands, the StarSign driver silently never got compiled into the Windows build at all — that's presumably why this checklist item couldn't be checked before now. Pushed the two-line `Makefile.mak` fix directly to this branch (`06663b372`).

### PKCS#11 module

With the fix applied, `opensc-pkcs11.dll` builds and works correctly against the physical token: `pkcs11-tool -O`/`-L` list slots, certs and pubkeys correctly, and a real `C_Sign` against the current signing key returns `CKR_OK` with a 256-byte RSA-2048 signature the card accepts (`SW 90 00`).

One quirk worth flagging: `pkcs11-tool --test`'s full-token enumeration walks an orphaned pre-2024 key plus all 4 certs before attempting to sign, which leaves the wrong DF selected for the actual signing key and gets `SW 6985` (Conditions of use not satisfied). Isolating the sign call to just the current key (`pkcs11-tool --sign --id ...`) works fine and returns `CKR_OK`. Not a driver bug in the real usage path — this card reuses the same raw key reference (`Key ref: 1`) for multiple keys, disambiguated by which DF is currently selected rather than by a distinct reference per key, so `--test`'s enumeration order happens to leave the wrong DF selected by the time it gets to signing.

### Windows minidriver

`opensc-minidriver.dll` builds cleanly against the CPDK headers with no changes needed beyond the fix above (the minidriver code itself is generic/ATR-driven, doesn't need per-card changes).

### Real-world validation

- **Authentication**, using the token through the Windows build, succeeded against three separate Brazilian judicial platforms: **TRT 3** (Labor Court), **TJMG/PJe** (state court), and **e-Proc** (federal court system) — the last one talks to the token/minidriver directly, the first two through a companion PKCS#11-based headless client.
- **PDF signing**: signed a real court filing (a multi-page legal petition PDF) using [pyHanko](https://github.com/MatthiasValvekens/pyHanko)'s PKCS#11 signer against `opensc-pkcs11.dll`. One thing worth noting for anyone else integrating against this driver: since it only advertises `SC_ALGORITHM_RSA_HASH_NONE` (see the DigestInfo point in the PR description above), pyHanko's *default* mechanism selection (`CKM_SHA256_RSA_PKCS`, hash-on-token) fails with `CKR_MECHANISM_INVALID` — you need `PKCS11Signer(..., use_raw_mechanism=True)` so pyHanko hashes client-side and hands the card a pre-built DigestInfo blob via plain `CKM_RSA_PKCS`, exactly mirroring what `sc_pkcs15_compute_signature` already does internally.
- The resulting signed PDF was then checked with Brazil's official government signature-conformance validator. For international reviewers: **ITI (Instituto Nacional de Tecnologia da Informação)** is the Brazilian federal agency that operates the country's root PKI (**ICP-Brasil**) — i.e. the top-level authority that the certificate chain used here (`AC OAB G3` → `AC Certisign G7` → `Autoridade Certificadora Raiz Brasileira v5`) ultimately chains up to. ITI publishes the authoritative conformance-checking tool that Brazilian courts and government systems rely on to accept or reject a digital signature — it's the closest Brazilian equivalent to a national eIDAS/PAdES conformance validator, run by the PKI authority itself rather than a third party.

  The validator's report on the signed PDF: `Status de assinatura: Aprovado` (signature status: approved), `Caminho de certificação: Valid` (full chain built and validated up to the ICP-Brasil root), `Cifra assimétrica: Aprovada`, `Resumo criptográfico: true`, `Estrutura: Em conformidade com o padrão` (structure conformant with the standard).

So: PKCS#11 module, minidriver, and end-to-end signature validity are all confirmed on Windows against physical hardware and real judicial systems, with the resulting signatures passing the Brazilian government's own conformance checker. Updated the checklist above accordingly.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Fixed — added the LGPL 2.1 header.

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Done — `starsign_select_ef_child()` now calls `iso_ops->process_fci(card, file, apdu.resp, apdu.resplen)` instead of the hand-rolled tag walk, and `starsign_select_file()` allocates the `sc_file_t` up front so it can be filled in directly during selection rather than reconstructed by hand afterward.

Verified against physical hardware: `pkcs15-tool -D` output is byte-identical to the previous parser for every certificate, key and PIN object on the token, and signing still works end-to-end (`C_Sign() = CKR_OK`).

src/libopensc/card-starsign.c
DiegoRibeirodeSouza: Implemented, with one correction to the snippet: `env->key_ref_len` is the length of the `key_ref` array, not the reference value itself — used `env->key_ref[0]` for `sbuf[2]` instead.

```c
if ((env->flags & SC_SEC_ENV_KEY_REF_PRESENT) && env->key_ref_len >= 1)
	sbuf[2] = env->key_ref[0];
if (env->flags & SC_SEC_ENV_ALG_REF_PRESENT)
	sbuf[5] = env->algorithm_ref & 0xFF;
```

Verified against physical hardware: as you suspected, this card's PKCS#15 profile doesn't currently set either flag, so the payload actually sent to the card is unchanged (`84 01 01 80 01 02`) and signing still succeeds (`C_Sign() = CKR_OK`, resulting PDF signature verified valid). Keeping the override in since it's correct and harmless either way, and makes the driver do the right thing automatically if that ever changes.

src/libopensc/pkcs15-starsign.c
DiegoRibeirodeSouza: Fixed — added the LGPL 2.1 header.

src/libopensc/pkcs15-starsign.c
DiegoRibeirodeSouza: Addressed part of this so far, but not the structural part yet, and I want to be upfront about that rather than claim more than I've done.

What's fixed in the latest commit: `starsign_select_ef_child()` now populates the `sc_file_t` via `iso_ops->process_fci()` instead of a hand-rolled FCP tag walk (verified byte-identical against physical hardware for every object on the token).

What's *not* addressed yet: the actual selection state machine (`has_3fff_placeholder`, `mid_fid`/`mid_fid_valid` caching, `df5031_selected`) is unchanged. I looked at `pkcs15-esteid2025.c` as you suggested — I understand the pattern (supply a static internal PKCS#15 profile with pre-resolved paths instead of parsing/patching the card's own PrKDF/AODF at runtime), and I agree it would let a lot of this go away if the paths could be expressed directly instead of reconstructed from the `0x3FFF` placeholder convention.

I'd rather take the time to do that properly (and re-verify against physical hardware, the same way I verified the changes in this commit) than rush a bigger rewrite in this pass. Will follow up with that as a separate commit once I've worked through it — didn't want to hold up the smaller, verified fixes in the meantime.

DiegoRibeirodeSouza: Following up on the structural part I said I'd tackle separately: pushed (`7a1c513`) a rewrite of `pkcs15-starsign.c` that replaces the dynamic `sc_pkcs15_parse_df()` + `starsign_parse_df()` patching with a static PKCS#15 profile, exactly along the lines you suggested -- modeled on `pkcs15-esteid2025.c`.

- No more `parse_df` hook and no more reading/patching the card's own AODF/PrKDF. `card-starsign.c`'s `select_file()` is untouched -- this only changes where object *metadata* comes from, not how files are physically selected on the card (the `0x3FFF` placeholder handling etc. is still needed for that, independent of this).
- Cardholder-specific fields (TokenInfo label/serial/manufacturer) are still read from the card's real `EF(TokenInfo)` at runtime, so this stays correct for every StarSign CUT S holder, not just mine.
- One thing worth flagging: the card's own CDF/PrKDF store each key/cert's `id` as the cardholder's name + issuance date, ASCII-encoded (e.g. mine literally decodes to `"DIEGO RIBEIRO DE SOUZA 2024-10-09..."`). Since going static means we no longer read that field off the card, I replaced it with small driver-assigned correlation ids (`01`-`05`) instead of hardcoding that string -- hardcoding it would have leaked my own identity into every other user's build of this driver.
- Added `SC_CARD_TYPE_STARSIGN` to `sc_pkcs15_is_emulation_only()` in `pkcs15-syn.c`. Without it this profile would never actually run by default: this card's raw PKCS#15 structure is well-formed enough that the generic dynamic parser (tried first by default) already succeeds on its own, so `sc_pkcs15_bind_synthetic()` would never get reached.

Verified against physical hardware on Windows: `pkcs15-tool -D` output is identical to the pre-refactor dump (paths, key references, PIN reference/flags, usage and access flags all match byte-for-byte) except for the intentional id/label change above. Re-ran the full validation suite from my last update: `C_Sign()` via PKCS#11 returns `CKR_OK`, a real PDF signed through this build (pyHanko + `use_raw_mechanism=True`) was approved by ITI's official conformance validator (`Caminho de certificação: Valid`, full chain to the ICP-Brasil root), and the driver works end-to-end through `pje_headless` against the local PJeOffice-compatible signer flow.

DiegoRibeirodeSouza:
Quick companion-repo update: rebuilt and republished the Debian packages (`starsign-driver-en`, the docs/packaging repo linked in the PR description) against this branch's latest commit (`7a1c51342`, the static PKCS#15 profile refactor) — they were stale since `v0.2.0`, built before that commit.

Retested against physical StarSign CUT S hardware after rebuilding:
- Certificate/key listing correctly reflects the new static profile (new labels, new driver-assigned ids `01`-`05`).
- A signature built the way real callers actually use this driver (SHA-256 hash + DigestInfo assembled client-side, signed via plain `CKM_RSA_PKCS`, matching pyHanko's `use_raw_mechanism=True` path) verifies correctly end-to-end with `openssl dgst -sha256 -verify`.
- `sudo`/PAM auth via `pam_pkcs11` against the same system library also works correctly with the rebuilt package.

Published as [v0.2.1](https://github.com/DiegoRibeirodeSouza/starsign-driver-en/releases/tag/v0.2.1). No driver source changes here — just confirming the already-reviewed code still works correctly once packaged for actual end users.

DiegoRibeirodeSouza:
Pushed a small correctness fix (`b8b6487`), unrelated to prior review comments -- found while investigating a local-only workaround (in the companion `starsign-driver-en` repo's packaged build, not this PR) that turned out to be masking a real bug in code that *is* part of this PR.

**The bug:** `pkcs15-starsign.c`'s static profile declared a native "Legacy Public Key" object paired with the previous certificate cycle's private key. For any holder who has renewed their ICP-Brasil certificate and cleared the old cycle via the standard SafeSign/XCA workflow -- normal, expected end-of-life for a 3-year certificate, not a corner case -- there's no longer any on-card EF backing that key's public half. Declaring it as native anyway makes the generic PKCS#11 layer try to read modulus bytes that don't exist, failing with `SC_ERROR_INTERNAL` (-1400) whenever a client enumerates objects or queries `CKA_MODULUS`. On my own token, which has exactly this history (renewed last year, old certificate cleared), this fired on every single `pkcs11-tool --list-objects` call.

**The fix:** stop declaring the pubkey object for the legacy key. The private key object itself is untouched and still discoverable (still useful for holders who *haven't* cleared their old cycle) -- it just won't surface in PKCS#11 clients that only discover keys via their public counterpart (e.g. XCA), which is an acceptable tradeoff since without a certificate it can't be used for anything verifiable anyway.

**Verified against physical hardware:** rebuilt both `libopensc` and the PKCS#11 module from this exact branch state, confirmed `pkcs15-tool -D` now shows zero orphaned Public Key objects (down from one), and `pkcs11-tool --list-objects --login` lists all 4 certificates and their keys cleanly with the legacy key present but pubkey-less as intended.

I also ran a real `C_Sign` with the active signing key (ID 01) in the same session, which failed with `CKR_GENERAL_ERROR` (`SW 6E00`, "Class not supported"). This diff cannot be the cause -- it only touches the legacy key's declaration, never `prkey_info`/`prkey_obj` for the active key or anything in the signing path. The most likely explanation is the token's already-documented CCID transport ceiling (`card.c`'s `sc_get_max_send_size()` can silently override the driver's `max_send_size` hint with a smaller value reported by the reader, triggering chaining that this card rejects) -- consistent with the known hardware limitation, though I did not confirm `SC_APDU_FLAGS_CHAINING` was actually set in that specific run, so I'm flagging this as the probable cause rather than a fully traced one.

src/libopensc/pkcs15-starsign.c
frankmorgner: For an actual user, I think only the token's name is typically shown (coming form EF.TokenInfo), so it should be fine for you.

If I remember correctly, then some application will show the common name of the certificate, if you need to choose one explicitly (e.g. Firefox). This should not be needed in most cases, due to the typically known root CA.

src/libopensc/pkcs15-starsign.c
```diff
@@ -0,0 +1,275 @@
...
+static const struct sc_pkcs15_id starsign_id_legacy = {.value = {0x02}, .len = 1};
+
+/* Certificate slots: MF -> [virtual 0x3FFF placeholder, never selected] ->
+ * DF 4302 -> EF <final>. This 4-slot layout (leaf, two intermediates, root)
+ * is fixed by the applet; only the DER content of each EF is personal. */
+static const struct {
+	const char *path;
+	const char *label;
+} starsign_certs[4] = {
+		{"3f003fff43020114", "Signature Certificate"},
```
frankmorgner: My idea here was to use a static PKCS#15 profile so that you can simplify the card driver. Why do you need the `3FFF` part here, if you ignore it at the card driver level? Why don't you just omit it here?

If you fully control all the paths of all data objects, doesn't this allow you to avoid your custom implementation of select_file so that you can just use `iso_ops->select_file`?

src/libopensc/pkcs15-starsign.c
DiegoRibeirodeSouza: Makes sense, thanks for the clarification -- that matches what I'd expect too (EF.TokenInfo label is the only thing visible in normal use, and it's genuinely mine).

src/libopensc/pkcs15-starsign.c
DiegoRibeirodeSouza: Tested both parts of this empirically against physical hardware.

Dropping the `0x3FFF` placeholder from these paths works cleanly -- the card genuinely never needed it, it was purely a leftover from the profile-path convention. Removing it exposed a real bug that had been hiding underneath it, though: the hierarchical-navigation branch in `starsign_select_file()` never captured the real file size from the card's FCP response the way the `0x3FFF`-placeholder and direct-child-EF branches do (via `starsign_select_ef_child()`), so certificates reached through this path silently fell back to the hardcoded 1024-byte placeholder instead of their real size, and failed ASN.1 parsing. Fixed that too. Re-verified against physical hardware: all 4 certificates on the token now read correctly (subject/dates check out) using clean paths like `3F0043020114` instead of `3F003FFF43020114`.

I also tried the bigger question -- dropping `starsign_select_file()` entirely in favor of `iso_ops->select_file`. That does not work on this card: it fails immediately with `SW 6A81` ("Function not supported"), because the generic ISO 7816 driver selects intermediate DFs with `P1=0x01` ("select child DF"), and this card only accepts `P1=0x00` (select by plain file ID) for that -- the same restriction already noted in `starsign_select_file()`'s comments, confirmed again here with a fresh test. So the override needs to stay, but I agree the `0x3FFF` convention itself is unnecessary baggage in the profile and can go.

Happy to push the corrected profile (clean paths) plus the FCI fix as a follow-up commit if that's useful, or hold off if you'd rather review the reasoning first.

frankmorgner:
Sorry, I'm tired of talking to a machine. maybe someone else has the time of iterating this into a maintainable state. If that is the case, please let me know and I'll have a second look.

DiegoRibeirodeSouza:
@frankmorgner — fair point, I won't pretend otherwise. I use AI to help write a lot of this text.  I'm a lawyer, not a programmer, and English isn't my first language. But the work behind it is mine: I've spent the last 6 days with the physical token in hand.

I found the DigestInfo bug by decrypting a real signature and seeing it fail standard verification even though the card returned `SW 90 00`, that came from testing, not from AI. Same story for figuring out why PJe (the Brazilian judicial system that motivated this driver) was rejecting the signature: I traced it down to Java's `SunPKCS11` falling back to BouncyCastle because the key is `CKA_SENSITIVE`.

If you'd rather hand review off to someone else, that's fine. I just wanted it on record that the work behind this is mine, even with AI helping write it.

PR #3765, opened 2026-08-06
Fix another bunch of memory issues reported by AISLE

Jakuje:
assorted memory and buffer fixes

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-starcos.c
+++ b/src/libopensc/card-starcos.c
@@ -1504,6 +1504,12 @@ static int starcos_gen_key(sc_card_t *card, sc_starcos_gen_key_data *data)
 	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
 		return sc_check_sw(card, apdu.sw1, apdu.sw2);
 
+	if (len > sizeof(rbuf) - 18 || apdu.resplen < 18 + len) {
+		sc_log(card->ctx, "READ PUBLIC KEY response too short/invalid: got %zu need >= %zu",
+				apdu.resplen, 18 + len);
+		return SC_ERROR_INVALID_DATA;
+	}
+
 	data->modulus = malloc(len);
 	if (!data->modulus)
 		return SC_ERROR_OUT_OF_MEMORY;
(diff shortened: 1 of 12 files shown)


src/libopensc/card-starcos.c
```diff
@@ -1504,6 +1504,12 @@ static int starcos_gen_key(sc_card_t *card, sc_starcos_gen_key_data *data)
 	if (apdu.sw1 != 0x90 || apdu.sw2 != 0x00)
 		return sc_check_sw(card, apdu.sw1, apdu.sw2);
 
+	if (len > sizeof(rbuf) - 18 || apdu.resplen < 18 + len) {
+		sc_log(card->ctx, "READ PUBLIC KEY response too short/invalid: got %zu need >= %zu",
```
frankmorgner, line 1508: You need to use `SC_FORMAT_LEN_SIZE_T` or simply drop the inclusion of the actual length

frankmorgner (changes requested):
LGTM

src/libopensc/card-starcos.c
Jakuje: I wonder why it does not fail the build anywhere now. AFAIK we have this workaround for Windows, but looks like the current windows already support the `%zu`.
This answer says that this is no longer needed since VS2015 so I would say we no longer need to use `SC_FORMAT_LEN_SIZE_T` (unless there is some other reason):
https://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu

frankmorgner: Yes, https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-140#size-prefixes-for-printf-and-wprintf-format-type-specifiers notes %z as equivalent to %I. Great, then we can leave it as is and later adopt this to the rest of the codebase.

frankmorgner (approved):

src/libopensc/card-starcos.c
Jakuje: Thank you for opening the issue for this. I added comment there with some more clarifications.

Jakuje:
Thank you for the review. I just squash the fixup commits  and rebased.

Issue #3736, opened 2026-07-16
NFC/contactless support for Uruguayan eID (PACE-MRZ)


nicolasgutierrezdev:
### Problem statement

The Uruguayan eID driver added in #3729 supports the contact interface only. The card also exposes a contactless interface, but it requires PACE-MRZ and Secure Messaging, which the current driver and OpenSC's SM path don't fully handle. The official closed-source driver did not work over NFC in my testing, so OpenSC is the only viable testing path right now.

### Proposed solution

Extend the driver to work over NFC. Concrete items, from experimenting on this branch, with an ACR122U:

- ATR matching: the contactless ATR differs from the contact ATR and is dynamic (bytes 5–8 randomized per tap). Need an ATR mask that ignores those bytes and distinguishes contactless from contact. The distinct NFC ATR can also gate the PACE/PIN-over-SM path (cleartext PIN verification appears disabled over NFC).
- Extended SM response: RSA signature readout over SM fails: the wrapped reply is 291 bytes but the short-APDU SM buffer caps it at 256. Widen the response buffer past 256 so an extended-length SM APDU is sent; needs confirming against card docs whether extended length is supported or chaining is required. ACR122U does not support extended APDU.
- SM teardown: after PACE the card keeps SM active until reset; sc_reset() doesn't clear it on the ACR122U. Use sc_sm_stop() instead of the re-tap workaround.

### Additional context

Card / PACE parameters:
- PACEInfo: id-PACE-ECDH-GM-AES-CBC-CMAC-256, curve NIST P-384
- PACE-MRZ only: MRZ (83 01 01) → 9000; CAN (83 01 02) → 6A88

Already working over contactless on this branch: PACE, PIN login, digests, cert/pubkey readout, MSE: SET, PSO: HASH.

Related: #3729 (contact support). Suggested fixes above (ATR gating, buffer expansion, sc_sm_stop()) are from @frankmorgner in that thread.

nicolasgutierrezdev:
@frankmorgner I have achieved working NFC flow. But before making a PR, I want to check a modification to `sm-iso.c` I made to allow chaining, since it would affect other drivers:

In line 479, in the short APDU case:
Old: 
```
sm_apdu->cse = SC_APDU_CASE_4_SHORT;
sm_apdu->resplen = 4 + 2 + mac_len + 2 + 2 + ((apdu->resplen+1)/ctx->block_length+1)*ctx->block_length;
[...]

PR #3768, opened 2026-08-10
card-idprime: Add EC mutual authentication / Secure Messaging for FIPS applets

hardening:
Pushed as draft for review, as claude worked on the generated code I'm aware that lots of comments aren't meaningful and should be reworked (claude tends to think that explaining its iterations are useful in code's comment).

Newer/FIPS IDPrime applets (seen on a SafeNet eToken 5300, reported as SC_CARD_TYPE_IDPRIME_930_PLUS) require every APDU to be wrapped in a proprietary AES-CBC/AES-CMAC Secure Messaging channel once an EC-based mutual authentication reference key is present ("full SM" mode). Without this, the driver could detect such cards but not actually talk to them beyond the initial SELECT.

This adds:
  - Detection of the MA reference key and the FIPS/full-SM applet configuration (idprime_probe_ecc_ma, idprime_get_applet_fips_config, idprime_get_ecc_full_sm). Cards without the MA key are unaffected and keep using the existing, unmodified code path.
  - The mutual-authentication handshake itself (EC key agreement via GENERAL AUTHENTICATE, PSO:VERIFY CERTIFICATE against an embedded CV certificate, EXTERNAL/INTERNAL AUTHENTICATE) in card-idprime.c.
  - sm-idprime.c/.h: a new SM transport implementing the derived AES-CBC encryption and AES-CMAC authentication, wired into the existing generic ISO 7816-4 SM encoder/decoder (sm-iso.c) via struct iso_sm_ctx.
  - Post-handshake quirks required by this applet, all confirmed against a real card: the SSC is reseeded to RND.ICC||RND.IFD once the handshake completes; the AID must not be re-selected once authenticated; SELECT FILE must not carry a protected Le under SM.
  - Correct key-reference derivation for private keys on this applet variant, which is keyed off the container's own index rather than the certificate's embedded id, with a fallback for containers that don't line up 1:1 with certificates.

Getting large operations (e.g. an RSA-2048 signature) working under this SM channel exposed four bugs in the shared SM/ISO7816 framework, since GET RESPONSE chaining under active SM had apparently never been exercised before:
  - sm-iso.c: the reassembled-response buffer was capped at the single-transfer short-form limit (256 bytes) instead of the size needed once GET RESPONSE chaining is used to exceed it.
  - sm-iso.c: the wrapped APDU's Le was hardcoded to 0, which made sc_get_response() skip GET RESPONSE entirely regardless of the buffer fix above.
  - sm-iso.c: the response buffer size formula was off by one byte (the padding-content indicator inside DO'87' wasn't counted), silently truncating the last byte of DO'8E' once chaining was involved.
  - iso7816.c: iso7816_get_response() didn't mark its own APDU SC_APDU_FLAGS_NO_SM, so once GET RESPONSE actually started firing it got wrongly re-wrapped in SM instead of going out in the clear.

Also fixes a related pre-existing bug in sm-iso.c: sm_decrypt()'s MAC buffer was sized for 8-byte (DES-class) MACs even though the module's own documented design assumes AES-length MACs; AES-CMAC responses were overflowing it.

Verified end-to-end against real hardware: object enumeration, PIN login, and RSA-2048 signing (signature independently verified against the card's own certificate).

### DIFF
--- a/src/libopensc/card-idprime.c
+++ b/src/libopensc/card-idprime.c
@@ -29,8 +29,23 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "asn1.h"
 #include "cardctl.h"
+#include "iso7816.h"
 #include "pkcs15.h"
+#include "sm/sm-idprime.h"
+
+#ifdef ENABLE_OPENSSL
+#include <openssl/bn.h>
+#include <openssl/ecdsa.h>
+#include <openssl/evp.h>
+#include <openssl/rand.h>
+#include <openssl/sha.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/core_names.h>
+#include <openssl/params.h>
+#endif
+#endif /* ENABLE_OPENSSL */
 
 static const struct sc_card_operations *iso_ops = NULL;
 
@@ -170,6 +185,20 @@ typedef struct idprime_private_data {
 	unsigned long current_op;		/* current operation set by idprime_set_security_env */
 	list_t containers;			/* list of private key containers */
 	list_t keyrefmap;			/* list of key references for private keys */
+
+	/* ECC Mutual Authentication / proprietary Secure Messaging state, mirroring
+	 * the proprietary driver's idp_maType / idp_isFIPS / idp_isFullSM. This is
+	 * detection only for now -- the SM channel itself is not implemented yet,
+	 * see idprime_probe_ecc_ma() / idprime_get_ecc_full_sm(). */
+	unsigned int ma_point_len; /* 0 if no MA reference key present, else EC point length (65/97/133) */
+	unsigned int sm_key_len;   /* AES key length (bytes) the card expects once MA is used (16 or 32) */
+	int is_fips;		   /* applet reports a FIPS-restricted operation mode */
+	int fips_identify;	   /* 0 = not FIPS, 2/3 = the two FIPS applet variants seen in app_opt bits */
+	int full_sm;		   /* mirrors idp_isFullSM: every APDU must be SM-wrapped */
+	int sm_authenticated;	   /* mutual authentication succeeded -- SM channel is active.
+				    * The card rejects re-selecting the IDPrime AID once we're
+				    * already sitting authenticated in it, so once this is set
+				    * idprime_select_idprime() must skip the re-select. */
 } idprime_private_data_t;
 
 /* For SimCList autocopy, we need to know the size of the data elements */
@@ -271,10 +300,78 @@ int idprime_add_object_to_list(list_t *list, const idprime_object_t *object)
 	return SC_SUCCESS;
 }
 
+/* Select a 2-byte file id under full SM. The generic iso7816_select_file()
+ * (used for the non-SM path) requests the FCI back via a protected Le
+ * (DO'97'), but the card rejects any SELECT command carrying that DO once
+ * mutual-authenticated -- it responds with a bare, unprotected SW=6982
+ * (i.e. it doesn't even attempt SM/MAC processing on the command). The
+ * captured proprietary driver trace confirms SELECT is always sent as
+ * Lc+data only (no Le at all) under this SM; the card still returns FCI in
+ * the encrypted response regardless of the absence of a request Le. */
+static int
+idprime_sm_select_file_by_id(sc_card_t *card, const sc_path_t *path, sc_file_t **file_out)
+{
+	sc_apdu_t apdu;
+	u8 rbuf[SC_MAX_APDU_BUFFER_SIZE];
+	const u8 *buffer;
+	unsigned int cla, tag;
+	size_t buffer_len;
+	sc_file_t *file;
+	int r;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0xA4, 0x00, 0x00);
+	apdu.data = path->value;
+	apdu.datalen = apdu.lc = path->len;
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof(rbuf);
+
+	r = sc_transmit_apdu(card, &apdu);
+	LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
+	r = sc_check_sw(card, apdu.sw1, apdu.sw2);
+	if (r != SC_SUCCESS)
+		LOG_FUNC_RETURN(card->ctx, r);
+
+	if (file_out == NULL)
+		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
+
+	if (apdu.resplen < 2)
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN_DATA_RECEIVED);
+	switch (apdu.resp[0]) {
+	case ISO7816_TAG_FCI:
+	case ISO7816_TAG_FCP:
+		file = sc_file_new();
+		if (file == NULL)
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_OUT_OF_MEMORY);
+		file->path = *path;
+		if (card->ops->process_fci == NULL) {
+			sc_file_free(file);
+			LOG_FUNC_RETURN(card->ctx, SC_ERROR_NOT_SUPPORTED);
+		}
+		buffer = apdu.resp;
+		r = sc_asn1_read_tag(&buffer, apdu.resplen, &cla, &tag, &buffer_len);
+		if (r == SC_SUCCESS)
+			card->ops->process_fci(card, file, buffer, buffer_len);
+		*file_out = file;
+		break;
+	default:
+		LOG_FUNC_RETURN(card->ctx, SC_ERROR_UNKNOWN_DATA_RECEIVED);
+	}
+
+	return SC_SUCCESS;
+}
+
 /* This selects main IDPrime AID which is used for communication with
  * the card */
 static int idprime_select_idprime(sc_card_t *card)
 {
+	idprime_private_data_t *priv = (idprime_private_data_t *)card->drv_data;
+
+	/* Once mutual authentication has established the SM channel, the card
+	 * is already sitting authenticated inside the IDPrime AID and rejects
+	 * (SW=6982) a redundant re-select of the same AID. Skip it. */
+	if (priv && priv->sm_authenticated)
+		return SC_SUCCESS;
+
 	return iso_ops->select_file(card, &idprime_path, NULL);
 }
 
@@ -284,6 +381,7 @@ static int idprime_select_file_by_path(sc_card_t *card, const char *str_path)
 	int r;
 	sc_file_t *file = NULL;
 	sc_path_t index_path;
+	idprime_private_data_t *priv = (idprime_private_data_t *)card->drv_data;
 
 	/* First, we need to make sure the IDPrime AID is selected */
 	r = idprime_select_idprime(card);
@@ -293,7 +391,10 @@ static int idprime_select_file_by_path(sc_card_t *card, const char *str_path)
 
 	/* Returns FCI with expected length of data */
 	sc_format_path(str_path, &index_path);
-	r = iso_ops->select_file(card, &index_path, &file);
+	if (priv && priv->sm_authenticated && index_path.len == 2)
+		r = idprime_sm_select_file_by_id(card, &index_path, &file);
+	else
+		r = iso_ops->select_file(card, &index_path, &file);
 
 	if (r != SC_SUCCESS) {
 		LOG_FUNC_RETURN(card->ctx, r);
@@ -463,6 +564,7 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv, 
 		if (((memcmp(&start[4], "ksc", 3) == 0) || memcmp(&start[4], "kxc", 3) == 0)
 			&& (memcmp(&start[12], "mscp", 5) == 0)) {
 			uint8_t cert_id = 0;
+			uint8_t key_index;
 			idprime_container_t *container = NULL;
 
 			if (start[7] >= '0' && start[7] <= '9' && start[8] >= '0' && start[8] <= '9') {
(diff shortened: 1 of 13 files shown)


frankmorgner:
See the discission about GET RESPONSE bug in iso SM in #3736

src/libopensc/card-idprime.c
```diff
@@ -556,6 +691,1159 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
...
+static const char *
+idprime_ecc_ma_curve_name(unsigned int point_len)
+{
+	switch (point_len) {
+	case 65:
+		return "prime256v1";
+	case 97:
+		return "secp384r1";
+	case 133:
+		return "secp521r1";
```
Jakuje, line 896: This sounds to me quite fragile as some encodings allow omitting some leading zeroes. Or are these just list of hardcoded keys that are found in the known cards that always have this encoding?

src/libopensc/card-idprime.c
```diff
@@ -556,6 +691,1159 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
...
+		0x82,
+		0x5f,
+		0x29,
+		0x01,
+		0x70,
+		0x42,
+		0x08,
+		0x46,
+		0x52,
+		0x47,
```
Jakuje, line 930: can you format this a bit less verbosely? Like 8 bytes per line or 16 bytes per line should be still readable and acceptable by the formatter.

src/libopensc/card-idprime.c
```diff
@@ -556,6 +691,1159 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
...
+	return ret;
+}
+
+/*
+ * Completes the proprietary "MA" mutual-authentication handshake on top of
+ * an already-agreed ECDH shared secret: activates Secure Messaging, sends
+ * the embedded terminal certificate (PSO:VERIFY CERTIFICATE), proves
+ * possession of the corresponding private key (EXTERNAL AUTHENTICATE,
+ * signature construction verified byte-for-byte via live dynamic
+ * instrumentation -- see ETOKEN_FIPS_SM_NOTES.md), and completes the
```
Jakuje, line 1597: this comment references some file that does not exist -- If there are some additional notes, feel free to propose them for wiki, or reflect them in the code comments like this.

src/libopensc/card-idprime.c
```diff
@@ -556,6 +691,1159 @@ static int idprime_process_index(sc_card_t *card, idprime_private_data_t *priv,
...
+		if (r == SC_SUCCESS)
+			r = SC_ERROR_WRONG_LENGTH;
+		goto err;
+	}
+	memcpy(rnd_icc, rbuf, sizeof(rnd_icc));
+
+	/* Build and sign the EXTERNAL AUTHENTICATE digest:
+	 *   SHA256(hostX || terminalID || RND.ICC || cardX || domainParams)
+	 * Verified via live signature verification against the card's own
+	 * public key -- see ETOKEN_FIPS_SM_NOTES.md section 7. */
```
Jakuje, line 1730: same here -- please review the comments and adjust them to be self-contained.

src/libopensc/card-idprime.c
```diff
@@ -615,6 +1903,46 @@ static int idprime_init(sc_card_t *card)
...
+	 * Authentication / Secure Messaging channel. Cards without the MA
+	 * reference key are driven exactly as before. Cards that have it get
+	 * their idp_maType/idp_isFIPS/idp_isFullSM state mirrored into priv;
+	 * cards that require full SM go through the mutual-authentication
+	 * handshake so the rest of this function (container map, index file,
+	 * etc.) can proceed transparently over the now-active SM channel. */
+	r = idprime_probe_ecc_ma(card, priv);
+	if (r != SC_SUCCESS) {
+		idprime_free_private_data(priv);
+		LOG_FUNC_RETURN(card->ctx, r);
```
Jakuje, line 1924: I think on errors like this we should assign back the card->drv_data to NULL to avoid possible use after free.

src/libopensc/iso7816.c
```diff
@@ -920,6 +920,19 @@ iso7816_get_response(struct sc_card *card, size_t *count, u8 *buf)
...
+	 * command it's completing, matching how real SM-capable cards expect
+	 * it (confirmed against a captured proprietary Gemalto/IDPrime driver
+	 * session, which always sends GET RESPONSE unprotected even mid an
+	 * active SM session). Without this, once SM is active any response
+	 * too big for a single short-form transfer (e.g. an RSA-2048
+	 * signature) fails: the GET RESPONSE APDU built here would otherwise
+	 * get SM-wrapped by sc_transmit_apdu(), the card would reject/not
+	 * understand that framing, and the SM layer would then fail trying to
+	 * decrypt a reply that was never SM-protected to begin with. */
+	apdu.flags |= SC_APDU_FLAGS_NO_SM;
```
Jakuje, line 935: Is there some wording around this in the ISO standard? Are we sure other SM of other vendors behave same?

src/sm/sm-idprime.c
```diff
@@ -0,0 +1,548 @@
...
+ * (DO'87'/'97'/'8E'/'99', ISO/IEC 9797-1 padding method 2) with:
+ *   - confidentiality: AES-CBC, IV = AES-ECB-Encrypt(KSKEnc, SSC)
+ *   - integrity: AES-CMAC(KSKMac, data), except the CBC-MAC chaining value
+ *     is seeded with SSC instead of zero (SSC is incremented once per APDU
+ *     direction, i.e. twice per command/response round trip)
+ * This file implements exactly that on top of OpenSC's generic SM
+ * encoder/decoder (sm-iso.c), which already produces byte-for-byte the same
+ * DO layout and CLA handling.
+ *
+ * STATUS: scaffolding, not validated against real hardware yet.
```
Jakuje, line 17: is this still the state or some leftover?

src/sm/sm-idprime.c
```diff
@@ -0,0 +1,548 @@
...
+
+static int
+idprime_sm_cmac_subkeys(const u8 *key, size_t key_len,
+		u8 k1[IDPRIME_SM_BLOCK_LEN], u8 k2[IDPRIME_SM_BLOCK_LEN])
+{
+	u8 zero[IDPRIME_SM_BLOCK_LEN];
+	u8 l[IDPRIME_SM_BLOCK_LEN];
+	int ok;
+
+	memset(zero, 0, sizeof zero);
```
Jakuje, line 123: I do not think the memset here is needed if we initialize the buffer:
```suggestion
	u8 zero[IDPRIME_SM_BLOCK_LEN] = {0};
	u8 l[IDPRIME_SM_BLOCK_LEN];
	int ok;

```

src/sm/sm-idprime.c
```diff
@@ -0,0 +1,548 @@
...
+static void
+idprime_sm_incr_ssc(u8 ssc[IDPRIME_SM_BLOCK_LEN])
+{
+	int i;
+
+	for (i = IDPRIME_SM_BLOCK_LEN - 1; i >= 0; i--) {
+		if (++ssc[i] != 0)
+			break;
+	}
+}
```
Jakuje, line 214: There is `sm_incr_ssc()` in sm-common.c -- could it be used to avoid code duplication?

src/sm/sm-idprime.c
```diff
@@ -0,0 +1,548 @@
...
+err:
+	free(out);
+	if (cctx)
+		EVP_CIPHER_CTX_free(cctx);
+	sc_mem_clear(icv, sizeof icv);
+	return r;
+}
+
+static int
+idprime_sm_decrypt(sc_card_t *card, const struct iso_sm_ctx *ctx,
```
Jakuje, line 262: these function (encrypt/decrypt) look also very generic. The sm-common has encrypt/decrypt functions for DES already so I believe these could be reused by some other drivers (if the idprime specifics can be factored out to some small wrappers.

Jakuje (commented):
thank you for going through this! Some inline comments and I am already looking forward to test it with my cards!

src/libopensc/iso7816.c
nicolasgutierrezdev: While working on #3736 I had to write a custom `cedulauy_get_response` (in order to not modify `iso7816.c`) that did exactly this. My current understanding is that since GET RESPONSE is just getting the remaining bytes, it should not be wrapped, but wrap the whole concatenated response instead. Sourcing from [2005 spec](https://www.freecalypso.org/pub/GSM/ISO7816/ISO_7816-4_2005.pdf):

Page 11 Clause 5.1.2
```
Command chaining and / or the use of SW1 set to '61' allow the transmission of data strings too long
for a single command. Such a process may split data objects in data fields consecutively sent as a
sequence in one direction, i.e., while sending no data field in the opposite direction. When chaining
commands and / or using SW1 set to '61', the concatenation of all the consecutive data fields in the
same direction in the same sequence shall be encoded in BER-TLV.
``` 

Page 28 Clause 6.1
```
NOTE Command chaining and / or the use of SW1 set to '61' induces sequences of commands where data fields
(and consequently, data objects) may be split in smaller consecutive data fields. In such a case, when using SM format,
the concatenation of all the consecutive data fields in the same direction in the same sequence is an SM field
```

It might require testing for other vendors though, mine is Gemalto.

frankmorgner: while reviewing #3736 I can also confirm that this is OK by multiple standards
- 7816-4 (2013) contains Figure 1 — Transmission to the card of an oversize command payload using command chaining, followed by the transmission by the card of an oversize response payload using response chaining. depicting this break down of a big response into multiple small ones, but without mentioning SM.
- [BSI TR-03110](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03110/BSI_TR-03110_Part-3-V2_2.pdf?__blob=publicationFile&v=1) (MRTD, eID) Annex E contains clarification on use of ENVELOPE (C-APDU) and response chaining (R-APDU) to support long messages with short length readers. This explicitly states that ENVELOPE/GET RESPONSE should not be encrypted itself.
- [PIV](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73pt2-5.pdf) depicts the same response chaining approach as in BSI TR-03110 in Fig. 7 and this was also implemented in https://github.com/OpenSC/OpenSC/pull/3098

The same "break-down" approach also exists for oversized C-APDUs, but the signalling differs:
- TR-03110 wraps the oversized SM-APDU into multiple ENVELOPE TPDUs. The chip concatenates all TPDU payload and recovers the SM-APDU
- PIV (Fig. 4) uses the SM-APDU header, enables C-APDU chaining in CLA byte and sends the oversized SM-APDU's payload in multiple chained APDUs (also implemented by @dengert )

All approaches are compatible with ISO 7816, so if possible we should try to support those in `sm-iso.c`.

dengert: I would say get response payload is either in the clear or is SM protected. The host sets if the apdu is using SM or not. It the command is SM protected, the card will respond with a SM protected response. 

May times READ BINARY is used multiple times to read a large file. In this case each READ BINARY command is SM protected and the response is also SM protected. 
  
When using a command such as GET DATA, command, there may be multiple GET RESPONSEs. I know that the NIST SM  get response does applies SM over the full response, then uses multiple GET RESPONSE commands to return the SM protected payload.

dengert: > All approaches are compatible with ISO 7816, so if possible we should try to support those in sm-iso.c
#3098 defines sm/sm-iso.h: 
```
        u8 padding_indicator;
        /** @brief If 1 always include padding_indicator even for tag 87 */
        u8 always_add_padding_indicator;
        /** @brief do not add padding on data to be mac'ed */
        u8 skip_mac_padding;
        /** @brief Do sm_encrypt of all the data. If needed use command chaining to send chunks to card.
         * Card will then return response using get_response commands in multiple chunks if needed.
         * Then do the sm decrypt and verify once on the full response. */
        u8 sm_encrypt_once_then_chaining;
        /** @brief Pad to this block length */
        size_t block_length;
```
For the NIST SM, these are set to 1, as these were found to be the different setting. Some other SM could use combinations of these as well or add additional changes if needed.

frankmorgner: Yes Indeed, some more tweaks to sm-iso.c are required for PIV SM, but only looking at sending oversized SM C-APDUs (either by chaining or by ENVELOPE) and looking at receiving oversized SM R-APDUs (GET RESPONSE), this is done in the same way. Anyway, considering the limitations of sm-iso.c found in this PR (and in #3736), the fixes should be moved to the generic implementation rather than being fixed on the driver level.

src/libopensc/card-idprime.c
hardening: Yeah seams to work, I don't know why clang-format changed the array like that when I've run it first, while now it keeps it in a big block. This program really has reproducibility issues.

src/libopensc/card-idprime.c
hardening: Sorry I don't understand the remark, these are just used to request openssl curves by name, can you elaborate ?

src/sm/sm-idprime.c
hardening: leftover

src/libopensc/card-idprime.c
Jakuje: The question is where do the point lengths come from and if we are guaranteed that they will always have this specific amount of bytes. For the prime256 the length is almost  always guaranteed to have the specific length, but for the p521 leading zeroes are more common (for example https://github.com/latchset/kryoptic/pull/470).

frankmorgner:
We just merged https://github.com/OpenSC/OpenSC/pull/3771 which helps with the problem around the bad Le in sm-iso.c and with the response chaining of oversized SM commands. This should simplify this PR a little.

src/sm/sm-idprime.c
hardening: True but if I'm correct that would force us to link `libopensc` with `libsm` which brings another set of troubles...

src/libopensc/iso7816.c
nicolasgutierrezdev: Due to the generality of this change to `iso7816.c`, I created #3777, to discuss the implementation. In order to offload this PR and make the fix self-contained.

PR #3775, opened 2026-08-11
Remove the use of SC_FORMAT_LEN_* macros

### LINKED ISSUE #3769: Remove SC_FORMAT_LEN_SIZE_T
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-140#size-prefixes-for-printf-and-wprintf-format-type-specifiers notes %z as equivalent to %I. This means we can simplify the codebase by removing the use of SC_FORMAT_LEN_SIZE_T making messages easier to read.

_Originally posted by @frankmorgner in https://github.com/OpenSC/OpenSC/pull/3765#discussion_r3751854869_

Jakuje:
This existed only for VS2013, which did not support standard %zu format string for size_t. Since VS2015, this is also supported and there is no more need for this shenanigans.

Replaced mostly mechanically with

    sed -i 's/" *SC_FORMAT_LEN_SIZE_T *"/z/g' file

and cleaned up manually to avoid too much line splitting for short log lines.

Fixes: #3769

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/apdu.c
+++ b/src/libopensc/apdu.c
@@ -364,10 +364,8 @@ sc_single_transmit(struct sc_card *card, struct sc_apdu *apdu)
 	if (card->reader->ops->transmit == NULL)
 		LOG_TEST_RET(card->ctx, SC_ERROR_NOT_SUPPORTED, "cannot transmit APDU");
 
-	sc_log(ctx,
-	       "CLA:%X, INS:%X, P1:%X, P2:%X, data(%"SC_FORMAT_LEN_SIZE_T"u) %p",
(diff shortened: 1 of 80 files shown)


frankmorgner (approved):

metsma (approved):

frankmorgner:
There are some coding style complaints. just applying the suggested diff should be enough for a fix

Jakuje:
Thanks. I will go through the formatting suggestions and apply them.

Jakuje:
This was just a rebase ro resolve conflict with recently merged changes.

frankmorgner:
thank you

PR #3776, opened 2026-08-11
ci: Skip java applets tests under valgrind for now

Jakuje:
For some reason, these two tests time out under valgrind over last couple of weeks so skip this part for now.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -182,7 +182,6 @@ jobs:
           cache: 'maven'
       - run: .github/setup-linux.sh ${{ matrix.target }} debug
       - run: .github/test-${{ matrix.target }}.sh ${{ matrix.ver }}
-      - run: .github/test-${{ matrix.target }}.sh ${{ matrix.ver }} valgrind
 
   test:
     name: test-${{ matrix.target }}-${{ matrix.dist }}
@@ -199,7 +198,6 @@ jobs:
       - *use_cache
       - run: .github/setup-linux.sh ${{ matrix.target }} debug
       - run: .github/test-${{ matrix.target }}.sh
-      - run: .github/test-${{ matrix.target }}.sh valgrind
 
   #######################
   ## LibreSSL pipeline ##


metsma:
Or can we adjust not use the valgrind build for tests?

Jakuje:
> Or can we adjust not use the valgrind build for tests?

I found the valgrind tests against the real applets quite useful as it runtime verifies common code paths with emulated hardware. I think it already caught some real issues (though I agree that its annoyingly slow).

Jakuje:
One more option might be to set some expected timeout (instead of the gracious 6 hours) so the stale jobs are killed earlier. But its more a bandaid to some underlying problem.

frankmorgner:
AI suggests that we should change the debuginfod configuration in setup-linux.sh/setup-valgrind.sh.

>If you look closely at the timestamps in your log:
>Huge delay: The first valgrind command (the sign command) starts at 01:47:36, but the program only prints its first output at 02:10:10. This means it just sat there waiting for over 22 minutes!
>Why does it hang? valgrind is automatically trying to download "debug symbols" from the internet for every system library it uses (like glibc, OpenSSL, libpcsclite). To do this, it runs a background process called debuginfod-find. You can actually see this process being killed as an "orphan" at the very end of your log file.
>The Timeout: Because every single valgrind command wastes 20+ minutes trying to download these files, the total time for all your tests adds up. At 02:15:02, your GitHub Action hits the hard 6-hour time limit (the job originally started at 20:15), so GitHub forcefully cancels it. The verify command itself isn't broken; it was just the unlucky command that was running when the 6 hours ran out.

setting DEBUGINFOD_URLS to an empty string should be enough. @Jakuje , could you check if d64f316c10fc8d5d896cdf94160d5d365cf37f37 was really needed?

Jakuje:
> @Jakuje , could you check if id64f316c10fc8d5d896cdf94160d5d365cf37f37 was really needed?

Interesting. The d64f316c10fc8d5d896cdf94160d5d365cf37f37 was needed because when we tried to install debuginfo packages from repositories, it was failing more often than not (and changing names more often than I liked). Though the `debuginfod.ubuntu.com` server sounds like is having some issues or rate limiting (likely due to AI scrapers  traffic) so I can not connect to that (and looking at the ci here sounds like it really was not related to the specific applets, but more like to the debuginfo server).

> setting DEBUGINFOD_URLS to an empty string should be enough. 

I think if we remove this variable, we will not have debuginfo at all and I think it will report some false positives with unresolved backtraces ...

Jakuje:
Actually it eventually passed, but still some of the jobs taking 4 hours is something annoying. So maybe it might be the best to skip the valgrind on applet tests for now until we will figure out a better way to make it working.

Jakuje:
I can also see that there are jobs `test-piv-24.04-piv-sm` and `test-piv-24.04-valgrind` running exactly the same test. The last parameter `build` (piv-sm x valgrind) is never used in the job itself (except for the title). I guess this dimension of the test matrix should be also dropped. I will do it also here.

Jakuje:
> The last parameter build (piv-sm x valgrind) is never used in the job itself (except for the title). I guess this dimension of the test matrix should be also dropped. I will do it also here.

Oh, its used for the cache key. Retrying again

.github/workflows/linux.yml
metsma, line 165:
```suggestion
        target: [gidsapplet, openpgp]
```

.github/workflows/linux.yml
```diff
@@ -170,6 +170,10 @@ jobs:
           { dist: '24.04', arch: amd64, build: 'valgrind', target: 'isoapplet', ver: 'v0' },
           { dist: '24.04', arch: amd64, build: 'valgrind', target: 'isoapplet', ver: 'v1' }
         ]
+        exclude: [
+          { dist: '22.04', arch: amd64, build: 'valgrind', target: 'piv' },
+          { dist: '24.04', arch: amd64, build: 'valgrind', target: 'piv' },
+        ]
```
metsma:
```suggestion
```

metsma (changes requested):

.github/workflows/linux.yml
Jakuje: Thanks for the suggestions! This looks clearer!

.github/workflows/linux.yml
Jakuje: Applied.

Jakuje:
All cac/pico-hsm/oseid valgrind tests are too slow too. I will probably drop them here so we have reasonably-fast turn-around and open an issue to investigate this and figure out if we can get this fixed somehow else in the long term.

Jakuje:
now it finishes in minutes. Created  #3779 to re-enable the valgrind eventually. I would like to get this one merged before merging other stuff to avoid dozens of timing out jobs in the master (and other PRs).

metsma (approved):

metsma:
Do you want keep the piv taget then?
I understand that there is currently 2 different combinations regular build (valgrind) with piv target test and piv-sm build with piv target test

Jakuje:
> Do you want keep the piv taget then? I understand that there is currently 2 different combinations regular build (valgrind) with piv target test and piv-sm build with piv target test

I think that was addressed by your suggestion to remove the `piv` from the matrix -- now we have just the `piv-sm` build target there (no `test-piv-24.04-piv-sm` anymore in the current run).

metsma:
> > Do you want keep the piv taget then? I understand that there is currently 2 different combinations regular build (valgrind) with piv target test and piv-sm build with piv target test
> 
> I think that was addressed by your suggestion to remove the `piv` from the matrix -- now we have just the `piv-sm` build target there (no `test-piv-24.04-piv-sm` anymore in the current run).

I was thinking if you disable valgrind anyway then we can keep both combinations?
But I do not know if there is any real difference there

Jakuje:
> I was thinking if you disable valgrind anyway then we can keep both combinations?

Good point. But they are running the same tests. I tried to compare the whole logs, but the cache used is the only difference. They both are configured with PIV without SM (this is in both of the logs):

> PIV SM support:          no

But it might be a different issue. Let me exclude this change from this PR and move it to the next issue for further investigation.

Jakuje:
> > I was thinking if you disable valgrind anyway then we can keep both combinations?
> 
> Good point. But they are running the same tests. I tried to compare the whole logs, but the cache used is the only difference. They both are configured with PIV without SM (this is in both of the logs):
> 
> > PIV SM support:          no
> 
> But it might be a different issue. Let me exclude this change from this PR and move it to the next issue for further investigation.

ok, this is caused by the `./configure --disable-notify` in `.github/setup-valgrind.sh` which overrides how the code was built in the first place, but likely only for the `valgrind` case which I removed in this PR now. So lets keep it now. I agree that its still useful to make sure the PIV applet works in both with and without SM configurations.

PR #3781, opened 2026-08-14
NIST Secure Messaging Usable by Other Card Drivers  - Replaces #3098 

dengert:
sm-nist-2

This branch and PR replaces the sm-nist branch i.e. PR #3098.

The PIV SM code is moved to sm-nist.c so it could be used by other card drivers. (A PR for MyEID is under development.)

The new sm-nist.c uses sm-iso.c to simplify the process as suggested.  This required some changes to sm-iso.h and sm-iso.c as NIST sp800-73-4 encodes and adds CMAC for the entire transaction then sends it to the card using command chaining of a single PUT_DATA command. Original sm-iso.c would break up a payload and encode and MAC each part and sent as separate APDUs.

configure.ac - Revised  ENABLE_PIV_SM and ENABLE_SM_NIST.

ENABLE_SM_NIST (enabled by default) is used to build sm/sm-nist.c based on NIST sp800-73-4 Secure Messaging which was moved from card-piv.c..

ENABLE_PIV_SM (enabled by default) is used within card-piv.c to use the sm-nist code if ENABLE_SM_NIST is defined.

If sm-nist is used by other card drivers, data needed such as the SM EC key reference and certificate with the public key and
CVC certificates will need to be provided by the card driver. The card driver will also determine which APDUs will
be protected by SM. 

sm/sm-nist.h sm/sm-nist.c -  define sm_nist_params_t

sm_nist_params_t is used to pass parameters and shared data to sm_nist_start. This allows a card driver to provide
certificates and flags outside of the strict PIV requirements so other card drivers can use the NIST SM as defined
in NIST sp800-73-4 but store the certificates X509 or CVC as PKCS15 files. sm_nist_params_cleanup can be used by a card driver to cleanup the sm_nist_params_t which can be part of a card driver's private data.

card-piv.c - Use the sm_nist_params_t when starting SM

card-piv.c - remove code moved to sm-nist

card-piv.c - Simplify reader lock obtained processing

The original card-piv.c would first select applet then test if SM was active and send a pin test command via SM to test if some other process had setup its own SM which would have different keys and counters. the pin test command would fail with 69 88.

The above test is no longer done by reader_lock_obtained. Instead, if any SM encoded APDU fails because of some SM error,  sc_sm_open is called and the APDU retried once.

sm/sm-nist.c will test for 6988 on all SM commands and return SC_ERROR_SM_RETRY_WITH_NEW_OPEN on SM failure.  sm.c will treat this as a recoverable error, call the SM open routine and then call card->sm_ctx.ops.get_sm_apdu. This retry method could be used by other SM modules in their get_sm_apdu routine. 

sm-iso.h sm-iso.c - updates needed by sm-nist.c

These are need so sm-nist.c can use sm-iso.c routines with 4 added flags in sm-iso.h The default values are all 0 so sm-iso.c should continue to work.

 sm/sm-nist.h sm/sm-nist.c - check for specific SM error codes

Many of the functions, defines, structures and variables have been renamed as part of separating the NIST sp800-73-4 Secure Messaging out of card-piv.c so it can be used by other card drivers. Names with nist* or NIST_* are in sm-nist.h and sm-nist.c to be used by card drivers. Names with piv* or PIV_* are specific to card-piv.c and implement specific SM and non SM features of PIV cards.

Note the names used for the pairing code and VCI have not been changed yet as they are PIV specific.

Export sm_nist_decode_cvc and sm_nist_clear_cvc_content. nist-decode_cvc and nist_clear_cvc_content were renamed and exported via libopensc.exports nist_cvc_t was moved to sm-nist.h

If some other process has accessed the card and set up its own SM  or selected a different applet the shared secret would be
different as the application uses an ephemeral EC key and the card generates a random nonce. SM and Login state are reset
in PIV card if a different applet is selected. 

PIV allows for some non SM protected operation to be done to a card even when SM is active such as get data and and PIN login state. NIST SM was designed for used over contactless interface where  the card will require PIN and crypto operations to be SM protected. This is the original way PIV cards worked before sp800-74-4.  So this still allows for multiple processes to use the same the card when the first process has verified the pin. 

This code can be tested by running  `pkcs11-tool --test --login --pin xxxxx` in two processes both using SM and watching the interactions which include having to reestablish SM authentication. Multiple runs show from 0 to multiple interferences.  

"Idemia Test Cards ID-One PIV 2.4.1 CIV configuration" card were used for testing, as well as MyEID-4.9.10 cards with the myeid-4.9.1 branch which in not part of this PR. 

##### Checklist


src/libopensc/sm.c
```diff
@@ -167,6 +169,20 @@ sc_sm_single_transmit(struct sc_card *card, struct sc_apdu *apdu)
...
+	/*
+	 * Multiple processes may have reset our SM session.
+	 * sm_ctx.ops.free_sm_apdu can detect SM failure
+	 * try again with new SM session using current pcsc lock
+	 */
+	if (rv == SC_ERROR_SM_RETRY_WITH_NEW_OPEN && retries > 0) {
+		sc_log(ctx, "Retry apdu with new SM session");
+		retries--;
+		if (card->sm_ctx.ops.open)
+			card->sm_ctx.ops.open(card);
```
frankmorgner, line 183: I think it should be up to the specific implementation `open()` to make the SM negotiation atomic by using `sc_lock()` or `sc_unlock()`. Also, I wonder in which situation this may actually fail, because all the higher level interfaces (e.g. PKCS#11/MD/PKCS#15) are using a lock on the card. So there should be no application which can actually interfere, shouldn't it?

src/sm/sm-iso.h
```diff
@@ -162,6 +162,14 @@ struct iso_sm_ctx {
 
 	/** @brief Padding-content indicator byte (ISO 7816-4 Table 30) */
 	u8 padding_indicator;
+	/** @brief If 1 always include padding_indicator even for tag 87 */
+	u8 always_add_padding_indicator;
+	/** @brief do not add padding on data to be mac'ed */
+	u8 skip_mac_padding;
```
frankmorgner, line 168: As I've written elsewhere, this behavior is allowed, but not default by 7816-4. As there are many other instances of "unless specified otherwise", I think it would be useful to create a `flag` which carries `ISO_SM_FLAG_SKIP_MAC_PADDING|ISO_SM_FLAG_ALWAYS_ADD_PADDING_INDICATOR` that may carry some more of those deviations.

src/sm/sm-iso.c
```diff
@@ -305,6 +305,7 @@ static int sm_encrypt(const struct iso_sm_ctx *ctx, sc_card_t *card,
 	}
 	sm_apdu->control = apdu->control;
 	sm_apdu->flags = apdu->flags;
+	sm_apdu->flags &= ~SC_APDU_FLAGS_SM_CHAINING;
```
frankmorgner, line 308: I think APDU SM chaining could eventually be ported to sm-iso.c since it seems more widely needed and supported by the ISO standard.

src/libopensc/sm.c
dengert: Yes all the commands are done under a lock on the card i.e. SCardBeginTransaction. The interference comes when one process (A) is done holding its locks  i.e. SCardEndTransaction  and has an  existing SM session. The card knows it has an SM session with the keys and counters from (A).  The card does not know anything about PCSC transactions.

If another process (B)  then locks the card  it can issue SELECT_AID to change applets or if the SELECT_AID happens to be to the same applet as used by (A), it continues but   (B) does not know that (A) has started an  SM session. But the card has the keys and counters from (A).  So when (B) sends a SM protected  APDU using (B) keys and counters, the card is expecting (A) keys and counters  and the decryption on the card fails and responds with 69 88 
  
From some of your comments, you implied some other SM implementations would end up with the same session keys from two different processes. But not with NIST SM.  sp800-73-5 part 2  "4.1. Key Establishment Protocol": "Generate an ephemeral key pair (deH; QeH) from the domain parameters specified in the response to the SELECT command."  And the process and card keep counters which is another failure point if the counters are out of sync.  

The interference occurs between holding the PCSC lock, not while holding  the lock. (But if some process is not using
PCSC locking, all bets are off.)

src/sm/sm-iso.h
dengert: I followed your use of bytes "u8 padding_indicator" they could be flags if that what you want.

src/sm/sm-iso.c
dengert: Maybe, but it also has to do with the size of the data. Does IOS 7816-4 define chaining  with extended apdus?

src/libopensc/sm.c
frankmorgner: What you are explaining is a general situation that can happen, e.g. **between** PKCS#11. This is understood and agreed. My point was that some second process cannot interfere while the first process is, e.g. **inside** a PKCS#11 operation. Reaching the commented code above means that implementing a retry for `card->sm_ctx.ops.open(card)` makes no sense, because the the above code is already guarded by `SCardBeginTransaction`.

dengert: I should point out NIST SM  allows for non SM protected apdu to be interspersed with SM protected apdus. So as long as process (B) does not do any SM  it can still read objects from the card. And when process (A) goes to use  SM the card and process (A) are in sync.

dengert: Here is an example of detecting and recovering from (A) (B) interference caused be running two `pkcs11-tool --login --test --pin 123456` at the same time. Each writing their own log.  I have seen 0 to 4 or more  interferences in multiple runs. Interference also depends on the OS scheduling. 

[recover-from-interference.txt](https://github.com/user-attachments/files/31100899/recover-from-interference.txt)
It includes a mod to show lock and unlock levels 
[lock-unlock.diff.txt](https://github.com/user-attachments/files/31100911/lock-unlock.diff.txt)

First 6 lines:
pkcs11 decrypt requests a lock
lock gets reader lock
piv_reader_lock_obtained  tests if PIV applet is still active by sending a SM protected get Discovery Object.

line 29 the basic SM code calls the piv_sm_nist_pre_transmit_callback which looks at INS
and says do SM on this.

line 33- 85 sm-iso.c encrypt.

line 96 APDU is sent

line 104 card returns 69 88 because while not holding  the lock the other process changed to SM on the card. 

line 110-122 sm_nist_post_transmit  says this can be retried: r=-1613 (Retry apdu with new sm_open)

line 113 sm-iso iso_rm_sm returns the -1613

line 114 sc_sm_single_transmit: Retry apdu with new SM session starts the retry

line 115-139  sm_nist_open starts a new NIST SM session to card
line 143 card responds

line 160-180 card sm_nist_open verifies the response i.e. new SM session is setup.

line 181 -252 the original APDU (discovery object) is reencoded using new SM session and sent to card
line 265  says SM  APDU returned 90 00
lines 266-295 response is decoded 
line 311 lock level is still 1 from framework call
line 312 sc_pkcs15_decipher starts 
line 342 the PIV 87 crypto operation is started

line 463 shows the SM protected data to be sent 297 bytes

line 493 shows the first 255 bytes being sent as command chaining
line 515 shows first part of chain returned 90 00
line 525  shows the last part of the chain being sent
line 532 is first part of the SM protected response which ends with 61 23
line 552-570  is get response
with line 564 showing SM should not be applied to the get response
Line 574  is last part of response
line 589 - 613 show un encrypted response 
line 633- 648 cleanup and return to sc_decipher: returning with: 256

Note: The reader lock was in line 6 from framework  `card.c:482:sc_lock: 0 -> 1`
and the `card.c:525:sc_unlock: 1 -> 0` was from framework 655-657  The interference was cause while the lock was not held because the other process changed the SM keys and counters on the card while it was holding the lock.

Conclusions:

- There can be no interference from other processes while the reader was locked or another card driver in a different thread could not be able to interfere as it could not get the reader lock for the same card.  the interference seen in  when (A) creates an SM session then (A) releases the lock. Then (B) can get the lock ,set its own SM session and unlock. When (A) get the lock and tries to use SM it fails with 69 88 because the card has (B) SM session.  The above example is (A) discovers the SM session on the card has been changed by (B).

- PIV operations are done by single APDU (and with command chaining and multiple GET REPONSE)
The PIV driver does set a lock when initializing holds on to one lock while reading the cert objects to get the pubkeys from the certificates which have private key parameters. This stops interference while initializing.  

- But with PKCS15 cards or other card drivers, that may not be the case, as they 
seam to lock and unlock way to much. But none use SM-NIST at this time.

- My myeid-4.9.1 branch which is based on this PR (my sm-nist-2 branch) will be reviewed as it uses PKCS15 may  additional calls to sc_lock.     

- Surprised sc_lock will call  `card->sm_ctx.ops.open(card)` in a reset. Need to look at that as well.
  
- From this exercise of documenting I noticed that the discover object may not be need, just use a SELECT AID in the clear. The only reason for using Discovery Object over SELECT AID  was because IIRC some old Yubico tokens would lose login state when SELECT AID was issued for these I could test CI_PIV_AID_LOSE_STATE.

- For now I am leaving the sc_lock patch in the code It is very handy.  


   





 






 
























card returns 69 88

frankmorgner: Thanks for the detailed analysis. Maybe due to the hidden lines in GH's review mode, I was assuming that you are `retry`ing within the card driver's sm_open routine. I just now realized, that this is happening inside `sc_sm_single_transmit()`. Here, I think it is indeed a useful feature, because we get the signalling over wrong SM for free (with from the card) rather than testing for it explicitly in card_reader_lock_obtained.

I would only suggest slight changes:
- naming the response code should be `SC_ERROR_INCORRECT_SM_OBJECTS`
- iso7816.c should return that new code on `0x6988`.
- add a new APDU flag `SC_APDU_FLAGS_RETRY_WITH_SM_OPEN`, which your driver can set before `sc_transmit_apdu()`. This kind of retrying mechanism matches what we already have in the non-SM case
- retrying can automatically be performed without special care from the card driver, because `sc_check_sw()` can directly be used to trigger the event

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+ */
+static int
+Q2OS(int fsize, u8 *Q, size_t Qlen, u8 *OS, size_t *OSlen)
+{
+	size_t i;
+	size_t f = fsize / 8;
+
+	i = (Qlen - 1) / 2;
+
+	if (!OS || *OSlen < f * 2 || !Q || i > f)
```
frankmorgner:
```suggestion
	if (!OS || !OSlen || *OSlen < f * 2 || !Q || i > f)
```

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+		apdu.le = (recvbuflen > 256) ? 256 : recvbuflen;
+		apdu.resplen = recvbuflen;
+	} else {
+		apdu.le = 0;
+		apdu.resplen = 0;
+	}
+	apdu.resp = recvbuf;
+
+	saved_sm_mode = card->sm_ctx.sm_mode;
+	if (card->sm_ctx.sm_mode != SM_MODE_NONE) {
```
frankmorgner, line 556: Why don't you use `apdu->flags |= SC_APDU_FLAGS_NO_SM`?

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+
+	/* Step H6  need left most 8 bytes of hash of sm_cvc */
+	{
+		u8 hash[SHA256_DIGEST_LENGTH] = {0};
+		const u8 *tag;
+		size_t taglen;
+		const u8 *tmpder;
+		size_t tmpderlen;
+
+		if ((tag = sc_asn1_find_tag(card->ctx, cvcder, cvclen, 0x7F21, &taglen)) == NULL || *cvcder != 0x7F || *(cvcder + 1) != 0x21) {
```
frankmorgner, line 883: should check cvclen, just to be sure.

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+err:
+	EVP_MD_CTX_free(md_ctx);
+	LOG_FUNC_RETURN(card->ctx, r);
+}
+
+/*
+ * If sm_in_cvc is present, verify NIST_OBJ_SM_CERT_SIGNER signed sm_in_cvc
+ * and sm_in_cvc signed sm_cvc.
+ * If sm_in_cvc is not present verify NIST_OBJ_SM_CERT_SIGNER signed sm_cvc.
+ */
```
frankmorgner, line 381: Where does the trust anchor (`NIST_OBJ_SM_CERT_SIGNER`) come from when verifying the certificates?

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+		return;
+	sc_mem_clear(session, sizeof(nist_sm_session_t));
+	return;
+}
+
+/*
+ * Decode a card verifiable certificate as defined in NIST 800-73-4
+ */
+int
+sm_nist_decode_cvc(sc_context_t *ctx, u8 **buf, size_t *buflen,
```
frankmorgner, line 1232: Just a reminder that CVC code should eventually be merged in a generic implementation shared with sc-hsm (and others)

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+	sctx->padding_indicator = SM_ISO_PADDING;
+	sctx->always_add_padding_indicator = 1;
+	sctx->skip_mac_padding = 1;
+	sctx->sm_encrypt_once_then_chaining = 1;
+	sctx->block_length = 16; /* 800-73-4 uses 16 for both cipher suites */
+
+	r = iso_sm_start(card, sctx);
+
+	/* We want to control if SM is on or not from driver, so set it off. */
+	card->sm_ctx.sm_mode = SM_MODE_NONE;
```
frankmorgner, line 1478: We always start SM to use it, don't we? So I would expect to enable it via `SM_MODE_TRANSMIT`. The driver still has the option to modify this behavior by adding `SC_APDU_FLAGS_NO_SM` or returning `SC_ERROR_SM_NOT_APPLIED`.

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+		SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_SM, SC_ERROR_INVALID_ARGUMENTS);
+
+	priv = (sm_nist_private_data_t *)ctx->priv_data;
+
+	/*
+	 * Interference from other processes is indicated by 6988
+	 * which will need to reauthenticate and reissue the failing command
+	 */
+	priv->params->last_sw1 = sm_apdu->sw1;
+	priv->params->last_sw2 = sm_apdu->sw2;
```
frankmorgner: I think this is not used anymore

src/sm/sm-nist.c
```diff
@@ -0,0 +1,2000 @@
...
+	sc_log(card->ctx, "nist_post_transmit - sw1:0x%X sw2:0x%X", sm_apdu->sw1, sm_apdu->sw2);
+	for (i = 0; nist_sm_errors[i].SWs != 0; i++) {
+		if (nist_sm_errors[i].SWs == ((sm_apdu->sw1 << 8) | sm_apdu->sw2)) {
+			sc_log(card->ctx, "%s", nist_sm_errors[i].errorstr);
+			SC_FUNC_RETURN(card->ctx, SC_LOG_DEBUG_NORMAL, nist_sm_errors[i].errorno);
+		}
+	}
+
+	memcpy(priv->sm_session.enc_counter_last, priv->sm_session.enc_counter, sizeof(priv->sm_session.enc_counter));
+	nist_inc(priv->sm_session.enc_counter, sizeof(priv->sm_session.enc_counter));
```
frankmorgner, line 1913: why do you need both, `enc_counter` and `enc_counter_last`? I think `enc_counter_last` is not actually used anywhere

src/sm/sm-iso.h
frankmorgner: values for `padding_indicator` are defined in 7816-4 (Table 52 — Padding-content indicator byte). If you want to add more information to this information (i.e. omit padding for MAC), I suggest you use `iso_sm_ctx->flag`.

src/sm/sm-iso.c
frankmorgner: Yes, 7816-4 makes no difference between extended and short length APDUs. It only says that fragmentation should not be used if the data fits the advertised limits of the card. Also, if a reader only supports short length (smartphone) then extended APDUs need to be transmitted fragmented.

Also, I believe that if `SC_APDU_FLAGS_CHAINING` is supported across iso7816.c, iso_sm.c and apdu.c, then we don't need SC_APDU_FLAGS_SM_CHAINING anymore

src/libopensc/sm.c
dengert: I forced pushed  Aug 17 seven commits following your suggestion. These could be squashed into two commits common code and the specific code in `sm-nist`. 

The new `SC_APDU_FLAGS_RETRY_WITH_SM_OPEN` is set by the  `sm_nist_pre_transmit` on the plain APDU, because the card driver does not see all APDUs for example PIN commands  where as `sm_nist_pre_transmit` sees all plain ADPUs. For other drivers using PKCS15 there are a lot of APDUs not seen by  the card driver. 

I question the need for  `SC_APDU_FLAGS_RETRY_WITH_SM_OPEN`  as it should be the default as it indicates the card was not able to decode the request and the caller would recover by call `sc_sm_open` some where is the callers code or quit using SM. But I have not looked how any of the other SM code in `src/sm` or card drivers handles failure of  a 69 88.

src/sm/sm-iso.h
dengert: On mt TODO list.

src/sm/sm-iso.c
dengert: Should be on someone's TODO list, But I don't want to address in the SM_NIST specific PR.

src/sm/sm-nist.c
dengert: On my list of commits to be pushed.

src/sm/sm-nist.c
dengert: That code is in `nist_sm_general_io` is only used to send the SM authenticate  APDU from `sm_nist_open `  i.e. called from  `sc_sm_open` which must be in the clear. You may be right, but for now turning off the SM mode makes sure no SM code is apdu.c tries to change this. 

Added to my TODO list.

src/sm/sm-nist.c
dengert: In a round about way it is. 
The response to the SM authentication is:
`7C' L1 { '82' L2 { CBICC || NICC || AuthCryptogramICC || CICC } }`
where the CVC is Cicc as the last part of the response. 

 830         /* payload is CBicc (1) || Nicc (16 or 24) || AuthCryptogram (CMAC 16 or 16) ||Cicc (variable) */
 831         p = (u8 *)payload;

 852         cvclen = len = payloadlen - (p - payload);
 853         if (len) {
 854                 cvcder = p; /* in rbuf */
 855 
 856                 r = sm_nist_decode_cvc(card->ctx, &p, &len, &priv->sm_cvc);
 857                 if (r != SC_SUCCESS) {
 858                         r = SC_ERROR_SM_AUTHENTICATION_FAILED;
 859                         goto err;

And example: CVC is in 3rd line  `7F 21 81 D1 ...`  to end which is cvclen of  213 bytes.

```
Incoming APDU (254 bytes):
7C 81 F9 82 81 F6 00 97 4B 82 D3 BA 64 2C 41 C0 |.......K...d,A.
16 27 03 84 C0 A8 3A 38 D9 0F 8B 9D 7F F2 EF D8 .'....:8........
A5 28 A3 CB 3D CE 01 7F 21 81 D1 5F 29 01 80 42 .(..=...!.._)..B
08 C8 01 ED BB 3E B7 C6 7C 5F 20 10 CC EB C0 C8 .....>..|_ .....
F1 0F 4B 19 8E A8 62 F4 6E 94 CF BB 7F 49 4D 06 ..K...b.n....IM.
08 2A 86 48 CE 3D 03 01 07 86 41 04 6B 0E 13 EA .*.H.=....A.k...
0A C5 CB 00 AD D8 EA B4 D5 7B 9D C4 A7 8F 14 3A .........{.....:
0D 7D 44 90 4F 58 D9 CA 12 CA A2 50 C4 00 08 8F .}D.OX.....P....
CE 48 5D 82 3F FB 6E AD FB AC 18 42 E0 BC F6 83 .H].?.n....B....
75 7F FA A5 11 00 42 DA D8 94 3E FA 5F 4C 01 00 u.....B...>._L..
5F 37 59 30 57 30 0A 06 08 2A 86 48 CE 3D 04 03 _7Y0W0...*.H.=..
02 03 49 00 30 46 02 21 00 BA 33 B5 9D 53 A5 DA ..I.0F.!..3..S..
9F 55 0F C5 B8 29 4F 90 D0 DA 50 11 E3 EA C8 A0 .U...)O...P.....
A4 ED 09 88 36 1D D1 ED BD 02 21 00 B2 A0 03 FB ....6.....!.....
28 71 E8 B6 5C A0 AE 4F FC 45 B2 64 E0 F5 46 03 (q..\..O.E.d..F.
75 37 3A 69 C0 68 A3 14 9B 0C 6B 44 90 00       u7:i.h....kD..
```
```
P:475962; T:0x139621430031360 19:58:29.903 [opensc-pkcs11] sm-nist.c:585:nist_sm_general_io: returning with: 252
P:475962; T:0x139621430031360 19:58:29.903 [opensc-pkcs11] sm-nist.c:1253:sm_nist_decode_cvc: called
P:475962; T:0x139621430031360 19:58:29.903 [opensc-pkcs11] asn1.c:1725:asn1_decode: called, left=213, depth 0
P:475962; T:0x139621430031360 19:58:29.903 [opensc-pkcs11] asn1.c:1748:asn1_decode: Looking for 'CVC certificate', tag 0x60001f21
P:475962; T:0x139621430031360 19:58:29.903 [opensc-pkcs11] asn1.c:1497:asn1_decode_entry: decoding 'CVC certificate', raw data:5F2901804208C801EDBB3EB7C67C5F20..
```
The code could have used the fact `priv->sm_cvc` was decoded instead of using get_tag.

Adding to mt TODO list.

src/sm/sm-nist.c
dengert: It comes from the card `Secure Messaging Certificate Signer 2.16.840.1.101.3.7.2.16.23 '5FC122'`
pkcs11-tool -O shows:

```
...
  label:      Secure Messaging Certificate Signer
  subject:    DN: CN=TEST PIV CONTENT SIGNER RSA FOR OT TEST PIV CARDS,OU=RnD IDENTITY,OU=CAI,OU=0002 34070953400259,O=OBERTHUR TECHNOLOGIES,C=FR
  serial:     2C
  ID:         129 (0x81)
  uri:        pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=de0a61813810c3ed;token=ID-One%20PIV%20TEST%20CARD%201%20L;id=%81;object=Secure%20Messaging%20Certificate%20Signer;type=cert
and later as pubkey, and as data object, as the certificate is contained in the data object (my be gtzipped)
and pubkey is in certificate. 
...
```
The data object may also contain the "intermediate CVC" certificate at the end.

Note that OpenSC verifies the CVCs  cert chain but leaves it up to any calling application to verify the  "Secure Messaging Certificate Signer".
 
```
./pkcs11-tool --read-object --uri "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=de0a61813810c3ed;token=ID-One%20PIV%20TEST%20CARD%201%20L;id=%81;object=Secure%20Messaging%20Certificate%20Signer;type=cert" | openssl x509
-----BEGIN CERTIFICATE-----
MIIFCTCCA/GgAwIBAgIBLDANBgkqhkiG9w0BAQsFADCBtDELMAkGA1UEBhMCRlIx
HjAcBgNVBAoMFU9CRVJUSFVSIFRFQ0hOT0xPR0lFUzEcMBoGA1UECwwTMDAwMiAz
NDA3MDk1MzQwMDI1OTEMMAoGA1UECwwDQ0FJMRUwEwYDVQQLDAxSbkQgSURFTlRJ
VFkxQjBABgNVBAMMOVRFU1QgUlNBIDIwNDggQ0EgRk9SIE9CRVJUSFVSIFRFQ0hO
T0xPR0lFUyBURVNUIFBJViBDQVJEUzAeFw0xNzA1MTYwNzExNDhaFw0zNzA1MTYw
NzExNDhaMIGsMQswCQYDVQQGEwJGUjEeMBwGA1UECgwVT0JFUlRIVVIgVEVDSE5P
TE9HSUVTMRwwGgYDVQQLDBMwMDAyIDM0MDcwOTUzNDAwMjU5MQwwCgYDVQQLDAND
QUkxFTATBgNVBAsMDFJuRCBJREVOVElUWTE6MDgGA1UEAwwxVEVTVCBQSVYgQ09O
VEVOVCBTSUdORVIgUlNBIEZPUiBPVCBURVNUIFBJViBDQVJEUzCCASIwDQYJKoZI
hvcNAQEBBQADggEPADCCAQoCggEBAMr3EWrES919KE0sy+Yii+/JGxc6Dxqveczn
zeJ65WhlTNnBkpu3f7iYxOYbIOA29CUWh8DLKcuJWhbEIqQeEoH0tTufGJmqKjbO
eewGfRLVEtqwdxBX12eRvYfoz58ALGTHXUKAO9dxQkc7XsbEPl8iL90zhg+L1Jvo
qF301hCkI5FhQ6z7KYmGxMUUyFEkni5RhMxw8gBmeB9pV9N0X61FaZvcj1EAfuuj
leEjd6eq0OrLnXM2HGmGukn1Rj2es5oQ1HSank6HzMs+jDdAtIz99ycbsa7zGV8D
NzA8joUj07PMJEFvPKzoVTQzwlrzqfMtRogalm4dsZo3e9g+cKMCAwEAAaOCASow
ggEmMEcGA1UdHwRAMD4wPKA6oDiGNmh0dHBzOi8vZGVtb3Bpdi5henVyZXdlYnNp
dGVzLm5ldC9kZW1vcGl2Y2Fyc2EyMDQ4LmNybDBSBggrBgEFBQcBAQRGMEQwQgYI
KwYBBQUHMAKGNmh0dHBzOi8vZGVtb3Bpdi5henVyZXdlYnNpdGVzLm5ldC9kZW1v
cGl2Y2Fyc2EyMDQ4LnA3YzAdBgNVHQ4EFgQURwC+ngAbbQ7GDP1lx2Br+Yffn8Mw
HwYDVR0jBBgwFoAUliWr0whTaY+fsL0AFdrgz91GFJkwCQYDVR0TBAIwADAOBgNV
HQ8BAf8EBAMCBsAwFwYDVR0gBBAwDjAMBgpghkgBZQMCAQMnMBMGA1UdJQQMMAoG
CGCGSAFlAwYHMA0GCSqGSIb3DQEBCwUAA4IBAQBiWz1AUA3MoMj76wosaXtN4rZI
P8qUXWoJw+RDWXGmiatsO1peOjeTeJAUAlOWJ19UJaCfn84vuYjTo6Xci/nzlAQ9
yifr26izLLeIpCSPcbHJVnRyudkeAjthyVmmVEuh9GX5OyVUqL6RKQ5m3q9KNlOg
uByqyrqx0Y+o1a564fUyncc8SPHruYuS3hwoWYGLpPFZssXDZcz1ncqLf316ZZUd
3LboMZxMVhD4prtYDK9OEO3BHJNHlyw2Q4I3f/bRuC3ELo89MtavH/2FhFXEQfWJ
3W5pDDtlcrktERz3jKk+Uu6iPBTzvpjOCo5DSdaOBAZrP1fXimKGd0MRs0Y2
-----END CERTIFICATE-----
```

src/sm/sm-nist.c
dengert: See the https://github.com/dengert/OpenSC/tree/myeid-4.9.1 that is based on sn-nist-2 that includes  PKCS15 and PKCS11 support to read and write CVC certificates to the card. This is being added because MyEID  has PIV support that can access the same certs and keys by both the MyEID and PIV applets on the card.   

I want to get the sm-nist-2 into master before  going back to the myeid-4.9.1 branch.  It is 64 commits out of date.

src/sm/sm-nist.c
dengert: As I said before, the card driver does not see every APDU.  So the `piv_sm_nist_pre_transmit_callback` is called from sm-nist.c with the plain APDU, and the card can then return 0 or `SC_ERROR_SM_NOT_APPLIED` so the basic sm code will do it in the clear.  The use by other drives is going to be more tricky because of pkcs15 many calls.
So other driver may need to have their own `*_*_ pre_transmit_callback`

The NIST SM was designed to protect PINs and crypto over contactless interfaces and the card enforces it for PINS and crypto over contactless.

Right now I have the `getenv("PIV_USE_SM")` which can set to always or never. more for testing, but could also be used for remote access, or for contact interface. The default right now is PINs and crypto of contact and contactless.

src/libopensc/sm.c
frankmorgner: Agreed, default yes, but with the option to disable it. The most useful place to decide should be in src/libopensc/card.c:1632 where `secure_messaging` options are checked.

src/sm/sm-nist.c
dengert: Not needed.  On my  to be commited list.

src/sm/sm-nist.c
frankmorgner: Do you mean that you prefer to have SM off by default (hence, setting `card->sm_ctx.sm_mode = SM_MODE_NONE`) and only when the PIV driver wishes to, it enables SM by changeing the `sm_mode`?

src/libopensc/sm.c
dengert: Looks like the `sc_card_sm_check` is called after the card driver's`*_init` routine. So that might work. But it is going to take some time. 

NIST SM can only be used if the card response to  SELECT AID for PIV includes the algo or NIST SM and there are lots of PIV like cards that do not. It might be nice if card driver can set flag so  `sc_card_sm_check` is not called.

Also right now `configure.ac` can enable(default)/disable `--disable-sm`, `--disable-sm-nist`, --disable-piv-sm-nist and in branch myeid-4.9.1 `--disable-myeid-sm-nist`o

And if a PIV card and a MyEID card are both using NIST SM, `sc_card_sm_check` would be called for both. I assume that is not a problem, as two different cards could be using sm-iso.

src/sm/sm-nist.c
frankmorgner: That's what I assumed. If the trust anchor comes from the token to be verified, then the security guarantee is very limited; it is more some kind of sanity check, because anyone could provide a token with a custom PKI. That's why we are deploying trust anchors for SC-HSM along with the OpenSC installation (`DESCHSMCVCA00001` and `DESRCACC100001` in the `etc` directory). Only CVCs in that directory are considered trusted (much like OpenSSL's directory of root certs `openssl version -d`).

Would it make sense to distrubute trusted certificates for NIST based tokens as well or are they always vendor specific? If we cannot find a reliable trusted root, then CVC verification is not really needed here, is it?

dengert:
I have withdrawn this comment. See below.

@frankmorgner @Jakuje 
 has there be recent changes to the packages used in `test-piv.sh`  like
`java -noverify -cp PivApplet/bin/:jcardsim/target/jcardsim-3.0.5-SNAPSHOT.jar com.licel.jcardsim.remote.VSmartCard PivApplet/test/jcardsim.cfg >/dev/null &`

I vaguely remember something about enforcing PIV strict requirements in `test-piv.sh`.  If that is the case some change to Virtual card interface or other code may now let the applet think it is contactless.  


PivApplet https://github.com/arekinath/PivApplet has not been updated since 2022 and for jcardsim-3.0.4,
but we are using jcardsim-3.0.5.

PivApplet has a feature:
"PIV_STRICT_CONTACTLESS | S | Block most slots and keys from use over contactless (strictly conform to the PIV spec)"
Looks like it is but not be the full spec. 

PivApplet is a `PIV-like" applet following Yubico, which do not follow: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73pt2-5.pdf  "Table 2. Data Model Containers" See foot notes as these are only allowed with the use of Secure Messaging. (OpenSC leaves it up to the card to enforce the restrictions, so Yubico cards can still violate the standards.)  

Strict NIST commands does not allow PIN or Crypto over a contactless interface, so "strictly conform to the PIV spec)"

The `test-piv.sh` is failing, showing the PIN is verified, but a chained crypto command is failing on the first chained 
APDU at line 10295 in:
https://github.com/dengert/OpenSC/actions/runs/32315748499/job/96275001907

```
Outgoing APDU (260 bytes):
10 87 07 9A FF 7C 82 01 06 82 00 81 82 01 00 00 .....|..........
00 00 00 00 00 00 00 00 00 00 94 75 D8 BE 61 89 ...........u..a.
F9 5C BB A8 99 0F 95 B1 EB F1 B3 05 EF F7 00 E9 .\..............
A1 3A E5 CA 0B CB D0 48 47 64 BD 1F 23 1E A8 1C .:.....HGd..#...
7B 64 C5 14 73 5A C5 5E 4B 79 63 3B 70 64 24 11 {d..sZ.^Kyc;pd$.
9E 09 DC AA D4 AC F2 1B 10 AF 3B 33 CD E3 50 48 ..........;3..PH
47 15 5C BB 6F 22 19 BA 9B 7D F5 0B E1 1A 1C 7F G.\.o"...}......
23 F8 29 F8 A4 1B 13 B5 CA 4E E8 98 32 38 E0 79 #.)......N..28.y
4D 3D 34 BC 5F 4E 77 FA CB 6C 05 AC 86 21 2B AA M=4._Nw..l...!+.
1A 55 A2 BE 70 B5 73 3B 04 5C D3 36 94 B3 AF E2 .U..p.s;.\.6....
F0 E4 9E 4F 32 15 49 FD 82 4E A9 08 70 D4 B2 8A ...O2.I..N..p...
29 54 48 9A 0A BC D5 0E 18 A8 44 AC 5B F3 8E 4C )TH.......D.[..L
D7 2D 9B 09 42 E5 06 C4 33 AF CD A3 84 7F 2D AD .-..B...3.....-.
D4 76 47 DE 32 1C EC 4A C4 30 F6 20 23 85 6C FB .vG.2..J.0. #.l.
B2 07 04 F4 EC 0B B9 20 BA 86 C3 3E 05 F1 EC D9 ....... ...>....
67 33 B7 99 50 A3 E3 14 D3 D9 34 F7 5E A0 F2 10 g3..P.....4.^...
A8 F6 05 94                                     ....

P:11286; T:0x139916241343040 00:50:54.714 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
P:11286; T:0x139916241343040 00:50:54.715 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
Incoming APDU (2 bytes):
69 82 i.

P:11286; T:0x139916241343040 00:50:54.715 [opensc-pkcs11] apdu.c:381:sc_single_transmit: returning with: 0 (Success)
P:11286; T:0x139916241343040 00:50:54.715 [opensc-pkcs11] apdu.c:538:sc_transmit: returning with: 0 (Success)
P:11286; T:0x139916241343040 00:50:54.715 [opensc-pkcs11] card-piv.c:4477:piv_check_sw: called
P:11286; T:0x139916241343040 00:50:54.715 [opensc-pkcs11] iso7816.c:128:iso7816_check_sw: Security status not satisfied
```

I vaguely remember something about enforcing PIV strict requirements in test-piv.sh. 

If that is the case some change to Virtual card interface or other code may now let the applet think it contactless.

src/sm/sm-nist.c
dengert: Commited as:
[sm-nist.h sm-nist.c - remove unused](https://github.com/OpenSC/OpenSC/pull/3781/commits/9ec446085131f6cff2ebd79ad53bbb0d6fef0327)

src/sm/sm-nist.c
dengert: The example certificate above was from Idemia demo cards. From the above certificate I was able to down load a p7c 
https://demopiv.azurewebsites.net/demopivcarsa2048.p7c  and its self signed trust anchor https://demopiv.azurewebsites.net/demopivca.p7c and these are demo cards. 

Yes, The trust  trust anchor should not come from the card, but from the card issuer. There is no common issuer, as PIV cards can come from government agencies that also provide the certificates on the card.  The card manufacture is not involved.  Although with a PIV-like card, they could be and their piv-tools could request a certificate from some CA.

With PIV SM the CVC must be signed and they are signed by x509  certificate signer. There for there are no CVCs
that are stored off the card. (The intermediate CVC is there for CA that can not use EC, to be phased out.)   

In my testing for  PIV, I use the OpenSSL demo CA to issue certificates from CSRs using  pkcs11-provider and OpenSSL utilities. With the MyEID cards using some horrible bash scripts I can create a CVC and sign it using OpenSSL. So I am creating the trust anchor. I use the same CA to create certificates for other keys too.

The process of using certificates from a smartcard over the network is the same as trusting the SM certificate but on the local host. For OpenSC, which depend on OpenSSL, one could add the trust anchor and/or intermediate  
certificates into the host's OpenSSL store. Same for Windows and set flag in OpenSC if we want to enforce verification of the certificate chain or not from the driver.  If NIST SM is used for remote connections, then the trust anchor needs to be on the other end. 

The NIST SM was designed to protect PINS and crypto and sensitive data over contactless interface from being sniffed. (And the PIV pairing code in another layer on top of this.)  

NIST also says NIST SM it is not to be use when provisioning a card.

src/sm/sm-nist.c
dengert: Now the  `sm_nist_pre_transmit` will return `SC_ERROR_SM_NOT_APPLIED` to force the caller to do an APDU in the clear if the sm-nist.c says so, or `sm_nist_pre_transmit` will call `priv->params->sm_nist_pre_transmit_callback)(card, apdu);`  the card driver can return 0 to use SM or `SC_ERROR_SM_NOT_APPLIED` to do it in the clear. 

PIV and PKCS15 cards uses a different set of APDU instructions so each will have a different callback.

dengert:
I have withdrawn this comment. See below

@frankmorgner @Jakuje  
 I have stalled for two weeks now trying to figure out why `.github/test-piv.sh` fails. 
I am questioning if the [vsmartcard](https://github.com/frankmorgner/vsmartcard.git) or the [jcardsim](https://github.com/Jakuje/jcardsim.git) code has changed recently.

NIST says PIV card shall not do crypto over contactless without NIST SM. 

https://github.com/arekinath/PivApplet/blob/master/src/net/cooperi/pivapplet/PivApplet.java#L587-L596
tests if it is using a contact or contactless interface.  It does not have NIST SM support so it should fail using contacless.

It test here:
https://github.com/arekinath/PivApplet/blob/master/src/net/cooperi/pivapplet/PivApplet.java#L2231-L2236

And the failure is at line 2905.   
https://github.com/arekinath/PivApplet/blob/master/src/net/cooperi/pivapplet/PivApplet.java#L2897-L2913
 I tried changing that to return a different code.

So it looks like the PivApplet is enforcing NIST policy. So if the simulator says contact it should work.  

It is also possible some code in `sm-nist-2` is causing the problem, but the failure at line 2905 points at the difference of how the simulator is setup as contactless vs contact. 

I have tried to get the simulator to run on my debian test system rather then using the CI to do the testing. It still fails.

Any insight as  to how to tell simulator how to simulate contact interface would be helpful.

src/sm/sm-nist.c
dengert: Fixed in: b12511f3cd  sm-nist.c - check pointer exists

dengert:
In regard to:  https://github.com/OpenSC/OpenSC/pull/3781#issuecomment-5500472815 and my comment: "I have stalled for two weeks now trying to figure out why .github/test-piv.sh fails."

I have found the problem. I had over simplified the `piv_reader_lock_obtained` removing too much code, so when a reset was done it would not reattach to the applet correctly.   
With recent commit: https://github.com/OpenSC/OpenSC/pull/3781/commits/f5369098a0b6d131872e7ec781cbafd9e7587e25 `card-piv.c - Cleanup piv_card_reader_lock_obtained` the `test-piv.sh` passes. 

I was also able copy the `test-piv.sh`  to my Debian 13 system to run a hacked up version to test locally. The hacked version use other .github/action routines  but uses locally built yubico-piv-tool, system pcscd and my working branch.

src/sm/sm-iso.h
dengert: @frankmorgner as requested;
 20d627a5415557c8ba9eebf6ecec62f4e0246df0  "sm-iso.h sm-iso.c - Use flags for ISO SM "Unless otherwise specified"  `sm-iso.h` defines the flags and `sm-iso.c`  uses them. 


f3f3b7bf121b69c88c5af699a411708310711b1f "sm-nist.c - Set the "Unless otherwise specified" ISO flags" `sm-nist.c` sets then in the `sctx`.

What do you think?

src/sm/sm-nist.c
dengert: For true compliance the calling pkcs11 application can read all the X509 certificates including the  "Secure Messaging Certificate Signer" and verify the trust of the "Secure Messaging Certificate Signer". "card-piv.c" has verified the CVC(s) against the "Secure Messaging Certificate Signer". 

NIST SP 800-73pt1-5 
"3.3.7. Secure Messaging Certificate Signer"
"The X.509 Certificate for Content Signing SHALL be a digital signature certificate issued under the id-fpki-common-piv-contentSigning policy of [COMMON]. The X.509 Certificate for Content Signing SHALL also include an extended key usage (extKeyUsage) extension asserting id-PIV-content-signing. Additional descriptions for the PIV object identifiers are provided in Appendix B of FIPS 201-3. The X.509 Certificate for Content Signing needed to verify the digital signature of a secure messaging CVC or Intermediate CVC of a valid PIV Card 7 SHALL NOT be expired."


I used the term "sniffing" NIST SP 800-73pt1-5 footnote 15 uses the term "Skimming":
"15 Skimming is when data is surreptitiously obtained from a contactless card using a hidden reader that powers, commands, and reads from the card within the maximum read distance (reported as about 25 cm with ISO/IEC 14443 smart cards like the PIV Card)."

dengert: > If the trust anchor comes from the token to be verified, then the security guarantee is very limited; it is more some kind of sanity check. 

The point of the NIST SM is to protect from "skimming". It protect the cards other keys and pins from being miss used. 

>  because anyone could provide a token with a custom PKI. 

Yes they could use OpenSC to do it for SM. But they do not have access to the keys like 9A, 9C or 9D or PINS from some other card they really want to use. They can write a 9A, 9C or 9D on a card, but the certificates for these are to be signed by a CA the remote application trusts. 

So I don't think this is a big trust issue. It is the card protecting itself from skimmig. 

True PIV cards to not allow the use of PINS or keys over contactless without SM. Some "PIV like" cards do.

src/sm/sm-nist.c
dengert: > Do you mean that you prefer to have SM off by default 

Since NIST SM was designed to be used with contactless and then only with PINS and crypto, I would say yes. But it looks like it could be useful over RDC.  Or even avoids sniffing by USB or PCSCD.  It is also useful for testing SM itself, thus the `getenv("PIV_USE_SM")`

Issue #3783, opened 2026-08-15
improve pkcs15 cache


mvogt1:
### Short summary ###

I found out, that if I swap the reading order in `sc_pkcs15_bind_internal` from

1. `5015` AID
2. `5031` EF(ODF)
3. `5032` TokenInfo
 
to

1. `5015` AID
2. `5032` TokenInfo
3. `5031` EF(ODF)

the caching performance improves by around 20%.

### Long description ###

For remote connections every APDU is expensive, additionally if IPA is used for authentication, where two `opensc-pkcs11` modules are run  for every authentication. [SSSD:8743](https://github.com/SSSD/sssd/issues/8743) 
(SSSD figures are local, not remote). But overall interactive usage becomes better for every omitted APDU call, regardless if local or remote.

I have analysed which APDU calls are issued in OpenSC and found out that in pkcs15.c in sc_pkcs15_bind_internal the ordering of the binding process can be optimized. Currently it reads first `5031`, which contains the files of the pkcs15 structure and afterward it reads "TokenInfo" (`5032`) which contains the serial number.
But only if the serial number is known, its is possible to use the cache, thus `5031` never can be read out of the cache!
As a result:

- Swapping the order improves the setup sequence by 20 percent.

(One APDU for AID, One select_file for `5032`,One read_binary on `5032`, and then one select_file on `5031`). If the size is correct the contents of 5031 are read from cache, which is a gain of 20%.

### Proposed solution

I intend to submit a patch / pull request for this.

frankmorgner:
sounds reasonable

mvogt1:
Resolved: Issue is handled in commited pull request #3784

PR #3784, opened 2026-08-15
Swap odf tokeninfo

mvogt1:
This patch increases the pkcs15 caching performance by 20%.
It adds two functions:

- sc_pkcs15_bind_internal_odf
- sc_pkcs15_bind_internal_tokeninfo

both are factored out from `sc_pkcs15_bind_internal`.
In a last step the calling order is reversed, which gets the 20% gain.
This patch refers to Issue #3783

### DIFF
--- a/src/libopensc/pkcs15.c
+++ b/src/libopensc/pkcs15.c
@@ -1072,77 +1072,17 @@ end:
 }
 
 int
-sc_pkcs15_bind_internal(struct sc_pkcs15_card *p15card, struct sc_aid *aid)
+sc_pkcs15_bind_internal_tokeninfo(struct sc_pkcs15_card *p15card)
 {
 	struct sc_path tmppath;
 	struct sc_card *card = p15card->card;
 	struct sc_context *ctx = card->ctx;
 	struct sc_pkcs15_tokeninfo tokeninfo;
-	const struct sc_app_info *info = NULL;
 	unsigned char *buf = NULL;
 	size_t len;
 	int err, ok = 0;
 
 	LOG_FUNC_CALLED(ctx);
-	/* Enumerate apps now */
-	if (card->app_count < 0) {
-		err = sc_enum_apps(card);
-		if (err != SC_SUCCESS)
-			sc_log(ctx, "unable to enumerate apps: %s", sc_strerror(err));
-	}
-	sc_file_free(p15card->file_app);
-	p15card->file_app = sc_file_new();
-	if (p15card->file_app == NULL) {
-		err = SC_ERROR_OUT_OF_MEMORY;
-		goto end;
-	}
-
-	sc_format_path("3F005015", &p15card->file_app->path);
-
-	info = sc_find_app(card, aid);
-	if (info) {
-		sc_log(ctx, "bind to application('%s',aid:'%s')", info->label, sc_dump_hex(info->aid.value, info->aid.len));
-		sc_pkcs15_free_app(p15card);
-		p15card->app = sc_dup_app_info(info);
-		if (!p15card->app) {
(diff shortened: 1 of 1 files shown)


frankmorgner:
I am confused... why don't you leave the structure of `sc_pkcs15_bind_internal()` as it is and just put the code block of tokeninfo parsing before odf parsing?

src/libopensc/pkcs15.c
```diff
@@ -1118,6 +1060,33 @@ sc_pkcs15_bind_internal(struct sc_pkcs15_card *p15card, struct sc_aid *aid)
 		sc_log(ctx, "  DF type %u, path %s, index %u, count %d", df->type,
 				sc_print_path(&df->path), df->path.index, df->path.count);
 
+	ok = 1;
+end:
+	if(buf != NULL)
+		free(buf);
+	if (!ok) {
+		sc_pkcs15_card_clear(p15card);
```
frankmorgner: error transformation and cleanup in p15card should be left to the caller

src/libopensc/pkcs15.c
frankmorgner: error transformation and cleanup in p15card should be left to the caller

frankmorgner (changes requested):

frankmorgner:
> I am confused... why don't you leave the structure of `sc_pkcs15_bind_internal()` as it is and just put the code block of tokeninfo parsing before odf parsing?

Nevermind, I just saw that you want to break the long function into smaller ones, which is a good thing!

frankmorgner:
please also check the coding style CI run

mvogt1:
I think I have luck with the coding style this time (I'm still learning git) and the error handling is left to the caller in the called functions:

- sc_pkcs15_bind_internal_odf
- sc_pkcs15_bind_internal_tokeninfo

there is no err rewrite and no call to `sc_pkcs15_card_clear(p15card)`.
Thus the `end:` sequence looks like:
```
end:
	if (buf != NULL)
		free(buf);
	if (!ok) {
		LOG_FUNC_RETURN(ctx, err);
	}

	LOG_FUNC_RETURN(ctx, SC_SUCCESS);
```

src/libopensc/pkcs15.c
```diff
@@ -1210,7 +1176,95 @@ sc_pkcs15_bind_internal(struct sc_pkcs15_card *p15card, struct sc_aid *aid)
...
+		sc_format_path("3F00", &p15card->file_app->path);
+		err = SC_SUCCESS;
+	}
+
+	if (err < 0) {
+		sc_log(ctx, "Cannot select application path");
+		goto end;
+	}
+
+	/* tokeninfo contains serialnumber, needed for filecache */
```
frankmorgner:
```suggestion
```

mvogt1:
```
Suggested change
	- /* tokeninfo contains serialnumber, needed for filecache */
```
Ok, done. 
(I have re-done the last commit, the last commit now removes the comment.)

frankmorgner (approved):

Jakuje:
Is the 20% caching improvement observed for specific card type or did you test more cards/tokens?

mvogt1:
>Is the 20% caching improvement observed for specific card type or did you test more cards/tokens?

The improvement is for all cards (caching on , odf + tokeninfo present).
But I have only tested mine (jcop driver).

I have measured the 20% improvement by adding a sleep(1) in apdu transmit (apdu.c). 
Reading the tokeninfo enables the cache and on the next run the call to ODF is then read form the cache.

I think, with this patch the optimization on the apdu transmit layer cannot be improved any further.

Jakuje:
Thank you!

PR #3786, opened 2026-08-17
Improve sc-hsm-tool

CardContact:
This pull request improves the sc-hsm-tool to support key domains available starting with version 3.x of the SmartCard-HSM. Improvements are

- Support for all initialization options
    - Disable RESET RETRY COUNTER command with --no-rrc.
    - Disable resetting the User-PIN with --no-pin-reset.
    - Allow to replace a key registered for Public Key Authentication (--replace-key-allowed).
    - Enforce authentication of PIN and PKA (--require-pka-and-pin). The default is PIN or PKA.
    - Enable transport-pin mode (--transport-pin).
- Allow to define the number of key domain slots to be allocated (--key-domain).
- Allow creating a DKEK key domain in a slot (--create-dkek-key-domain).
- Allow deleting a key domain (--delete-key-domain).
- Allow clearing the Key Encryption Key in a key domain (--clear-kek).
- Allow generating keys in a key domain (--generate-key and --key-domain).
- Support the specification of a key use counter (--key-use-counter).
- Identify the key to be wrapped by its label (--wrap-key with --label).
- Let the driver select the key reference in --unwrap-key.

Two changes are required in the OpenSC core code:

- sc_pkcs15init_generate_key() now allows passing aux_data in the new field aux_data in sc_pkcs15init_prkeyargs. Before aux_data was only used between the P15 layer and the card layer to pass minidriver specific information (GUID). The new feature is used to pass card driver specific parameter for key generation, like key domain id, key use counter and algorithm list in the case of a SmartCard-HSM.
- compare_obj_key() is enhanced to allow searching for the label of a PKCS#15 object. So far searching for the label was only supported for data objects.

Some background on key domains: A key domain can span multiple SmartCard-HSM and allows migration of key material. In its simplest form, a key domain is created by SmartCard-HSMs that have been initialized with the same Device Key Encryption Key (DKEK). XKEK key domains further automate establishing a key encryption key by using ECDH and public key attestation (see [docs](https://www.pki-as-a-service.org/km/xkek/) for more details).

SmartCard-HSMs in version 1 and 2 have support for a single DKEK key domain, while version 3 and later can have multiple key domain slots configured during initialization. These pre-allocated slots can then be used to create DKEK or XKEK key domains. Keys on the device are either associated with the default key domain or a key domain referenced by the index on the device. From the PKCS#11/PKCS#11 perspective it does not matter in which key domain a key was generated, as there is simply no attribute to carry the key domain information. This is why support for key generation was added in sc-hsm-tool to allow specifying the desired key domain.

When generating key via PKCS#11, the key is always associated with the default key domain, which might be the single DKEK key domain created during initialization by using --dkek-shares.

##### Checklist

- [ ] macOS token is tested


doc/tools/sc-hsm-tool.1.xml
```diff
@@ -63,26 +67,78 @@
 					</listitem>
 				</varlistentry>
 
+				<varlistentry>
+					<term>
+						<option>--create-dkek-key-domain</option> <replaceable>number-of-shares</replaceable>,
```
Jakuje: needless trailing comma:
```suggestion
						<option>--create-dkek-key-domain</option> <replaceable>number-of-shares</replaceable>
```

doc/tools/sc-hsm-tool.1.xml
```diff
@@ -63,26 +67,78 @@
...
+					</term>
+					<listitem>
+						<para>Clear the key encryption key established in the key domain slot identified by <option>-d</option> (<option>--key-domain</option>).</para>
+						<para>Use <option>--pin</option> to provide the user PIN on the command line.</para>
+					</listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>
+						<option>--delete-key-domain</option>.
```
Jakuje: I do not think the dots after options are necessary:
```suggestion
						<option>--delete-key-domain</option>
```

src/libopensc/card-sc-hsm.h
```diff
@@ -62,6 +62,10 @@
 
 #define INIT_RRC_ENABLED		0x01		/* Bit 1 of initialization options */
 #define INIT_TRANSPORT_PIN		0x02		/* Bit 2 of initialization options */
+#define INIT_SESSION_PIN		0x04		/* Bit 3 of initialization options */
+#define INIT_REPLACE_PKA_KEY	0x08		/* Bit 4 of initialization options */
+#define INIT_REQ_PKA_AND_PIN	0x10		/* Bit 5 of initialization options */
```
Jakuje: nit: these comments are annoying not explanatory. I would prefer if the would describe the flags, rather than describing the value.

src/pkcs15init/pkcs15-sc-hsm.c
```diff
@@ -291,14 +293,58 @@ static int sc_hsm_generate_key(struct sc_profile *profile, struct sc_pkcs15_card
 	sc_pkcs15emu_sc_hsm_free_cvc(&cvc);
 	LOG_TEST_RET(p15card->card->ctx, r, "Could not encode GAKP cdata");
 
-
 	cvcpo = cvcbin;
 	sc_asn1_read_tag((const u8 **)&cvcpo, cvclen, &cla, &tag, &taglen);
 	sc_asn1_read_tag((const u8 **)&cvcpo, cvclen, &cla, &tag, &taglen);
 
+	memcpy(gakp, cvcpo, taglen);
```
Jakuje, line 304: please, make sure the `taglen < sizeof(gakpk)`.  Also check the return value of `sc_asn1_read_tag` if you want to be sure the TLV is correct. It can return `SC_ERROR_ASN1_END_OF_CONTENTS` and much larger `taglen` than you expected.

src/tools/sc-hsm-tool.c
```diff
@@ -803,15 +883,181 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
 
-static int import_dkek_share(sc_card_t *card, const char *inf, int iter, const char *password, int num_of_password_shares)
+static int ensure_login(sc_card_t *card, const char *pin)
 {
-	sc_cardctl_sc_hsm_dkek_t dkekinfo;
+	struct sc_pin_cmd_data data;
+	char *lpin;
+	int r;
+
+	memset(&data, 0, sizeof(data));
```
Jakuje: please, use the initializers instead of explicit memset

```suggestion
	struct sc_pin_cmd_data data = {0};
	char *lpin;
	int r;

```

doc/tools/sc-hsm-tool.1.xml
CardContact: Fixed

doc/tools/sc-hsm-tool.1.xml
CardContact: Fixed

src/libopensc/card-sc-hsm.h
CardContact: Fixed

src/pkcs15init/pkcs15-sc-hsm.c
CardContact: I added a test, even though an overflow should not occur, as the structure is build a couple of lines before.

src/tools/sc-hsm-tool.c
CardContact: I would rather leave the code as is, because data is reset a couple of times in the code. I'd prefer to have a consistent use.

Jakuje (approved):
Thank you! Looks good

src/libopensc/cardctl.h
```diff
@@ -972,6 +976,13 @@ typedef struct sc_cardctl_sc_hsm_pka_register {
     sc_cardctl_sc_hsm_pka_status_t new_status;
 } sc_cardctl_sc_hsm_pka_register_t;
 
+#define SC_MANAGE_KEY_DOMAIN_IMPORT_DKEK_SHARE		1
+#define SC_MANAGE_KEY_DOMAIN_CREATE_DKEK_KEY_DOMAIN	2
+#define SC_MANAGE_KEY_DOMAIN_CREATE_XKEK_KEY_DOMAIN	3
+#define SC_MANAGE_KEY_DOMAIN_DELETE_KEY_DOMAIN		4
+#define SC_MANAGE_KEY_DOMAIN_CLEAR_KEK				5
+#define SC_MANAGE_KEY_DOMAIN_ASSOCIATE_KEY_DOMAIN	6
```
frankmorgner: An SC-HSM specific prefix should be used. Looking at the struct's name, I suggest something like `SC_HSM_KEY_DOMAIN_OPERATION_IMPORT_DKEK_SHARE`

frankmorgner (changes requested):
As the number of CLI arguments for sc-hsm-tool is getting even more overwhelming, could you please also add usage examples for the new options to the Wiki? In particular, I would be interested in uses of the Match-on-card (including links to the product itself).

src/libopensc/cardctl.h
CardContact: Fixed that.

CardContact: Match-on-Card is supported by the [KX901](https://www.keyxentic.com/en/kx901) token from KeyXentic and the GoID card, which you probably know better than me ;-)

frankmorgner (approved):
As said before, I'd like to see more usage examples for the new options in the wiki.

Also, you should link to [KX901](https://www.keyxentic.com/en/kx901) there, since match-on-document is still very unique!

Code-wise we are ready, I think.

src/tools/sc-hsm-tool.c
```diff
@@ -2121,6 +2572,14 @@ int main(int argc, char *argv[])
 		fprintf(stderr, "Option -S (--public-key-auth-status) excludes option -g\n");
 		exit(1);
 	}
+	if ((initopts & INIT_REPLACE_PKA_KEY) && opt_num_of_pub_keys == -1) {
+		fprintf(stderr, "Option --replace-pka-keys requires option --public-key-auth\n");
+		exit(1);
+	}
+	if ((initopts & INIT_REQ_PKA_AND_PIN) && opt_num_of_pub_keys == -1) {
+		fprintf(stderr, "Option --require-pka-and-pin option --public-key-auth\n");
```
Jakuje:
```suggestion
		fprintf(stderr, "Option --require-pka-and-pin requires option --public-key-auth\n");
```

Jakuje (approved):

CardContact:
Rebased to include #3815.

src/libopensc/card-sc-hsm.c
```diff
@@ -1337,15 +1336,32 @@ static int sc_hsm_import_dkek_share(sc_card_t *card, sc_cardctl_sc_hsm_dkek_t *p
 
 	LOG_FUNC_CALLED(card->ctx);
 
-	if (params->importShare) {
-		sc_format_apdu(card, &apdu, SC_APDU_CASE_4_SHORT, 0x52, 0x00, 0x00);
-		apdu.cla = 0x80;
+	switch(params->operation) {
+	case 0:
```
frankmorgner: Please use a named value just like in the other cases, e.g. `SC_MANAGE_KEY_DOMAIN_IMPORT_GET_STATUS`.

src/tools/sc-hsm-tool.c
```diff
@@ -809,15 +889,184 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
 
 
-static int import_dkek_share(sc_card_t *card, const char *inf, int iter, const char *password, int num_of_password_shares)
+static int ensure_login(sc_card_t *card, const char *pin)
+{
+	struct sc_pin_cmd_data data = {0};
+	char *lpin;
+	int r;
+
+	memset(&data, 0, sizeof(data));
```
frankmorgner: this operation duplicates what `struct sc_pin_cmd_data data = {0};` has already done.

src/tools/sc-hsm-tool.c
```diff
@@ -809,15 +889,184 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
+	data.cmd = SC_PIN_CMD_GET_INFO;
+	data.pin_type = SC_AC_CHV;
+	data.pin_reference = ID_USER_PIN;
+
+	r = sc_pin_cmd(card, &data);
```
frankmorgner, line 895: the return value is ignored. is that intended?

src/tools/sc-hsm-tool.c
```diff
@@ -809,15 +889,184 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
+	}
+
+	memset(&dkekinfo, 0, sizeof(dkekinfo));
+	dkekinfo.operation = SC_HSM_MANAGE_KEY_DOMAIN_CREATE_DKEK_KEY_DOMAIN;
+	dkekinfo.dkek_shares = (u8)shares;
+	dkekinfo.key_domain_idx = (u8)(kdidx - 1);
+
+	r = sc_card_ctl(card, SC_CARDCTL_SC_HSM_MANAGE_KEY_DOMAIN, (void *)&dkekinfo);
+
+	if (r == SC_ERROR_INS_NOT_SUPPORTED) {			// Not supported or not initialized for key shares
```
frankmorgner: the comment is redundant considering the error message below.

src/tools/sc-hsm-tool.c
```diff
@@ -809,15 +889,184 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
+		return -1;
+	}
+
+	memset(&dkekinfo, 0, sizeof(dkekinfo));
+	dkekinfo.operation = SC_HSM_MANAGE_KEY_DOMAIN_DELETE_KEY_DOMAIN;
+	dkekinfo.key_domain_idx = (u8)(kdidx - 1);
+
+	r = sc_card_ctl(card, SC_CARDCTL_SC_HSM_MANAGE_KEY_DOMAIN, (void *)&dkekinfo);
+
+	if (r == SC_ERROR_INS_NOT_SUPPORTED) {			// Not supported or not initialized for key shares
```
frankmorgner: again, redundant comment

src/tools/sc-hsm-tool.c
```diff
@@ -809,15 +889,184 @@ static int recreate_password_from_shares(char **pwd, int *pwdlen, int num_of_pas
...
+		return -1;
+	}
+
+	memset(&dkekinfo, 0, sizeof(dkekinfo));
+	dkekinfo.operation = SC_HSM_MANAGE_KEY_DOMAIN_CLEAR_KEK;
+	dkekinfo.key_domain_idx = (u8)(kdidx - 1);
+
+	r = sc_card_ctl(card, SC_CARDCTL_SC_HSM_MANAGE_KEY_DOMAIN, (void *)&dkekinfo);
+
+	if (r == SC_ERROR_INS_NOT_SUPPORTED) {			// Not supported or not initialized for key shares
```
frankmorgner: redundant comment

src/tools/sc-hsm-tool.c
```diff
@@ -991,7 +1243,7 @@ static int print_dkek_share(sc_card_t *card, const char *inf, int iter, const ch
 
 	memset(&dkekinfo, 0, sizeof(dkekinfo));
 	memcpy(dkekinfo.dkek_share, outbuff, sizeof(dkekinfo.dkek_share));
-	dkekinfo.importShare = 1;
+	dkekinfo.operation = 1;
```
frankmorgner: use the define

src/tools/sc-hsm-tool.c
```diff
@@ -1301,6 +1544,138 @@ static int create_dkek_share(sc_card_t *card, const char *outf, int iter, const
...
+};
+
+static int parse_alg_spec(const struct alg_spec *types, const char *spec, unsigned int *keybits, struct sc_pkcs15_prkey *prkey)
+{
+	int i, types_idx = -1, algorithm = -1;
+	unsigned int user_keybits = 0;
+	char *end;
+
+	for (i = 0; types[i].spec; i++) {
+		if (!strncasecmp(spec, types[i].spec, strlen(types[i].spec))) {
```
frankmorgner, line 1553: `spec` is `opt_keyspec` when called in `generate_key()`. This is user input and may be NULL if unspecified (default). Accessing the string must first check the pointer.

src/tools/sc-hsm-tool.c
```diff
@@ -1301,6 +1544,138 @@ static int create_dkek_share(sc_card_t *card, const char *outf, int iter, const
...
+
+	if (*spec != '\0') {
+		if (isalpha((unsigned char)*spec) && algorithm == SC_ALGORITHM_EC && prkey) {
+			if ((prkey->u.ec.params.named_curve = strdup(spec)) == NULL) /* pass EC curve name */
+				return SC_ERROR_OUT_OF_MEMORY;
+		} else if ((algorithm == SC_ALGORITHM_EDDSA || algorithm == SC_ALGORITHM_XEDDSA) && prkey) {
+			if ((prkey->u.ec.params.named_curve = strdup(types[types_idx].spec)) == NULL) /* copy correct case */
+				return SC_ERROR_OUT_OF_MEMORY;
+			user_keybits = (unsigned)strtoul(spec, &end, 10);
+			if (*end) {
```
frankmorgner: according to the man page it is theoretically possible that `*end` is `NULL` if the base isn't supported. I think it is a good idea to (1) initialize end to NULL and to (2) check `end != NULL`.

src/tools/sc-hsm-tool.c
```diff
@@ -1301,6 +1544,138 @@ static int create_dkek_share(sc_card_t *card, const char *outf, int iter, const
...
+				util_error("Invalid number of key bits \"%s\"", spec);
+				return SC_ERROR_INVALID_ARGUMENTS;
+			}
+			if (user_keybits != *keybits) {
+				util_error("If specified, number of key bits must be \"%d\" for \"%s\"", *keybits, types[types_idx].spec);
+				return SC_ERROR_INVALID_ARGUMENTS;
+			}
+		} else { /* rsa or symmetric key */
+			*keybits = (unsigned)strtoul(spec, &end, 10);
+			if (*end) {
```
frankmorgner: same comment regarding *end as above

frankmorgner (changes requested):
your command line options are complex (I commented on some), you should add a fuzzer target for those to check for problems with (missing) user input.

src/tools/sc-hsm-tool.c
CardContact: Also fixed in pkcs15-init from where the code was copied.

frankmorgner:
please check the coding style. CI throws out a diff, that you only need to apply.

CardContact:
How do I do that ?

frankmorgner (approved):
thank you.

the suggestion of a CLI fuzzer for the tool and the extension of the wiki documentation can be added in seperate PRs.

CardContact:
I've already done some changes in the wiki, but that needs certainly more love.

For the CLI fuzzer I first need to understand how that is done.

Jakuje:
Thank you! I think this is good to merge now.

The CLI fuzzer is something that we came up with as an experiment. Its able to detect some logic in the CLI tools. It occasionally finds an issue (we have now one opened for pkcs15-tool).

saper:
@CardContact  How is transport PIN supposed to work here? 

After the tool quits, there is no session anymore available to generate the keys protected under the transport pin?

Is there a practical usage example for this?

I tried this in https://github.com/OpenSC/OpenSC/pull/2431 and I gave up because just setting the flag made no sense to me.

CardContact:
The option is merely provided for completeness.

It can be rarely used with command line tools that loose the session at every invocation.

One application could be, that you initialize a device with a secret SO-PIN and a transport PIN and give that without any keys to a user. The user can then change the transport-PIN with a user selected PIN and start requesting certificates.

Btw. this is the model we use in the PKI-as-a-Service: You first distribute the secure key container to the user and later refer to this secure key container and it's attestation mechanism when requesting a certificate.

The SmartCard-HSM has a provable identify from the production process, that you can rely on when you later do key attestation or just the login process in the PKI-as-a-Service portal.

So an "empty" SmartCard-HSM with a transport-PIN has some value, because of the unique device identity in the device authentication certificate.

If you use an integrated provisioning system, then you can of course access all functions after you initialize the device, as long as you do no end the session.

PR #3787, opened 2026-08-17
Pkcs11 spy

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/pkcs11/pkcs11-spy.c
+++ b/src/pkcs11/pkcs11-spy.c
@@ -256,12 +256,9 @@ init_spy(void)
 
 #ifdef _WIN32
 	if (!spy_output) {
-		/* try for the machine version first, as we may be running
-		 * without a user during login
-		 */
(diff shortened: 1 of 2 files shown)


Jakuje (approved):

frankmorgner:
@dengert are you fine with that change? (You had introduced switching to HKLM as default instead of HKCU previously.)

dengert:
If it fixes the problem, I am OK. IIRC at login there is no local user.

dengert:
You may want to consider that a local user should not be using SPY.  Only a windows user with admin privileges who can also update HKLM. That is how I use it if needed.

frankmorgner:
Technically, there [*always* is a system user](https://devblogs.microsoft.com/oldnewthing/20070302-00/?p=27783#:~:text=It's%20actually%20the%20profile%20for%20the%20Local,and%20services%20that%20run%20as%20Local%20System.) who has some HKCU tree, but I doubt that anyone configures this with some pkcs11-spy.

Anyway, I think we are good here.

PR #3800, opened 2026-08-21
pkcs11-tool: validate the object ID length before storing it

### LINKED ISSUE #3799: pkcs11-tool: stale opt_object_id_len after an oversized CKA_ID is rejected (follow-up to 814f745b)
### Describe the bug

The fix for CVE-2026-10275 (`814f745b`, and the ECC counterpart `e1e11ceb`) writes the token-supplied `CKA_ID` length into the global `opt_object_id_len` *before* range-checking it. The rejection path returns without clearing it, so the oversized value survives in the global and is reused by a later action in the same invocation.

Notably, **the same commit already handles this correctly at one of the three sites.** In `main()` the PKCS#11 URI path uses `util_fatal()`, which terminates, so no stale global can survive there. Only the two test paths use a soft `return`.

Line numbers below are against master `fff55ae307c97b334ead67c3b3f9e5c0bb17ebb5` (2026-08-21). `src/tools/pkcs11-tool.c` is byte-identical at `ef6a1be8` and at that commit.

* `469` — `static CK_BYTE opt_object_id[100], new_object_id[100];`
* `1344-1347` — the URI path from the same fix, which gets it right:

```c
opt_object_id_len = opt_uri->id_len;
if (opt_object_id_len > sizeof(opt_object_id))
        util_fatal("URI's object ID too long");   /* terminates - no stale global */
memcpy(opt_object_id, opt_uri->id, opt_object_id_len);
```

* `9622-9627` — `test_kpgen_certwrite()`, which does not:

```c
opt_object_id_len = (size_t) i;              /* published to the global first */
if (opt_object_id_len > sizeof(opt_object_id)) {
        fprintf(stderr, "ERR: object ID too long\n");
        return session;                      /* global keeps the oversized value */
}
memcpy(opt_object_id, tmp, opt_object_id_len);
```

* `9810-9815` — `test_ec()` has the same ordering (with `return;`).

These two are the only unbounded writes to `opt_object_id_len` in the file. The others are already safe: `960-961` is capped by `sc_hex_to_bin` with the buffer size as an in/out parameter, `1344` is `util_fatal`-guarded as shown, and `9747-9748` uses a fixed 20-byte literal.

The length comes from `getID()`, generated by `VARATTR_METHOD(ID, unsigned char)` (`786`, macro at `726-750`), which passes the module's reported `ulValueLen` back verbatim. Its only constraints are that the value is not `(CK_ULONG)-1` and that `calloc(1, ulValueLen + 1)` succeeds, so the length is caller-chosen across the whole allocatable range.

**Severity:** low, and arguably a hardening/correctness defect rather than a vulnerability. The overflowing write that CVE-2026-10275 was about is genuinely fixed; what remains is a read, and it is performed by the module, which is a shared object already inside `pkcs11-tool`'s address space. The cases where it matters are a **malfunctioning** module reporting a wrong `ulValueLen`, or an **honest** module relaying an oversized `CKA_ID` originating from a hostile card. In both, `pkcs11-tool` is the component reading past a 100-byte static buffer, and those bytes travel outward as an attribute value or into logs.

*On disclosure:* I considered the private advisory channel and judged this non-exploitable, so I am filing publicly. Happy to move it to a private advisory if you disagree.

I could not find an existing report: no results for `"object ID too long"`, and the only `opt_object_id` / `test_kpgen_certwrite` hits are #3682 and PR #3710, both closed.

*Provenance, to be upfront about it:* I maintain a long-term-support backport of opensc 0.23.0 (TuxCare ELS), which is where I first hit this while backporting CVE-2026-10275. I am not asking you to support 0.23.0 -- I checked current master specifically to see whether the same ordering was still there, and it is, at the lines cited above. That backport carries the reordered form shown under "Expected behavior".

*No debug log attached:* this was found by reading master rather than by hitting it in the field, and reproducing it needs a purpose-built PKCS#11 module that reports an oversized `CKA_ID`. I am happy to write a minimal stub module and attach a run if that would help you confirm it.

### Steps to reproduce

With a PKCS#11 module that reports an oversized `CKA_ID` for a freshly generated key:

```
pkcs11-tool --module <module.so> --login --id 01 \
            -z <cert-file> \
            --test-ec --key-type EC:secp256r1
```

1. `-z` (`--moz-cert`) runs `test_kpgen_certwrite()`. The module returns `CKA_ID` with, say, `ulValueLen = 1000000`. Line `9622` stores it in the global, `9623` rejects it, `9625` returns — `opt_object_id_len` is now `1000000`.
2. `--test-ec` runs `test_ec()`. `main()` dispatches the test actions as independent `if` statements (`1744`, `1751`), not mutually exclusive ones, so both run in one invocation; there is no mutual-exclusion check (`action_count` is only tested for `== 0` at `1295` and against `do_list_slots` at `1414`/`1430`).
3. `test_ec()`'s entry guard at `9781` is `if (!opt_object_id_len)`, which the stale non-zero value passes.
4. `test_ec()` calls `gen_keypair()` at `9802`, before it reassigns the length at `9810`.
5. `gen_keypair()` (`3453`) fills both templates at `3839-3846`, with `FILL_ATTR` (`3450`) assigning `ulValueLen = len` directly, so `C_GenerateKeyPair()` receives `CKA_ID` = `{opt_object_id, 1000000}` — a pointer to a 100-byte static buffer described as a megabyte long.

Prerequisites for step 1 to get that far, all trivially satisfiable by the module: it must advertise `CKM_RSA_PKCS_KEY_PAIR_GEN` (`9591-9599`, else early return), the `-z` file must be openable (`9601-9603`, else `util_fatal`), and `gen_keypair()` must find the EC keygen mechanism for step 2.

### Expected behavior

`opt_object_id_len` should only ever hold a length that has been validated against `sizeof(opt_object_id)`. Rejecting an oversized token-supplied ID should leave the global as it was — holding the previously validated `--id` length, which still matches the contents of `opt_object_id[]` — so that a subsequent action cannot inherit a length that does not describe the buffer.

Validating before publishing the length achieves this. Two hunks, since the sites differ slightly in spacing and return value:

```c
/* test_kpgen_certwrite(), ~9622 */
-       opt_object_id_len = (size_t) i;
-       if (opt_object_id_len > sizeof(opt_object_id)) {
+       if ((size_t) i > sizeof(opt_object_id)) {
                fprintf(stderr, "ERR: object ID too long\n");
                return session;
        }
+       opt_object_id_len = (size_t) i;
        memcpy(opt_object_id, tmp, opt_object_id_len);

/* test_ec(), ~9810 */
-       opt_object_id_len = (size_t)i;
-       if (opt_object_id_len > sizeof(opt_object_id)) {
+       if ((size_t) i > sizeof(opt_object_id)) {
                fprintf(stderr, "ERR: object ID too long\n");
                return;
        }
+       opt_object_id_len = (size_t) i;
        memcpy(opt_object_id, tmp, opt_object_id_len);
```

Alternatively `util_fatal("object ID too long")` would match what `814f745b` already does at the URI site and is strictly safer, at the cost of turning a skipped test into an exit. Resetting `opt_object_id_len = 0` on the error path also works, but changes behaviour for any later action that expects the `--id` value to still be set.

### Actual behavior

The oversized length is written to the global before it is checked, and the early return leaves it there. `test_ec()`'s `if (!opt_object_id_len)` guard passes on the stale value, and `gen_keypair()` then hands `C_GenerateKeyPair()` a `CKA_ID` attribute pointing at the 100-byte `opt_object_id[]` with `ulValueLen` set to the rejected length. The module reads past the end of that static buffer, and the bytes it reads leave the process as an attribute value.

The overflowing `memcpy` is correctly prevented, so this is the residual read rather than the original write. Consequences range from leaking adjacent static/`.bss` objects to a crash on a large length. (Object layout in `.bss` is not guaranteed by the language, so exactly what is adjacent is implementation-dependent.)

### OpenSC version

master fff55ae307c97b334ead67c3b3f9e5c0bb17ebb5 (2026-08-21); introduced by 814f745b (0.26.x) and e1e11ceb

### Operating system

Not OS-specific - source-level defect; reviewed on Linux (AlmaLinux 9)

Not card-specific - requires a PKCS#11 module or token reporting an oversized CKA_ID

pkcs11-tool

### Logs and debug output

```shell

```

### Checklist

wtatarski:
814f745b assigns the token-supplied CKA_ID length to the global opt_object_id_len before checking it against the destination size, so the rejection path returns with the oversized value still in the global.

main() runs the requested test actions sequentially rather than exclusively, so a following action reaches gen_keypair(), which pairs that stale length with the 100-byte opt_object_id buffer in both CKA_ID attribute templates. The overflowing memcpy is prevented, but C_GenerateKeyPair() is then handed a 100-byte buffer described as being longer.

Check the length before publishing it to the global, at both sites. On rejection the global keeps the previously validated --id length, which still describes the contents of the buffer.

Same approach as the URI hunk of 814f745b, which avoids the problem by using `util_fatal()`; here a `return` is kept so the behaviour of the test paths is unchanged apart from not leaving a stale length behind. Happy to switch to `util_fatal()` for consistency if you prefer.

Fixes #3799

##### Testing

No card was used: this touches only the `--moz-cert` / `--test-ec` paths in
`pkcs11-tool`, and reaching the affected code needs a PKCS#11 module that reports an
oversized `CKA_ID`, which I have not built. The change was verified by inspection
against master and with `git-clang-format --diff --commit upstream/master`, which
reports no changes.

I am happy to write a minimal stub module and post a before/after run if you would
like the behaviour demonstrated rather than argued.

##### Checklist

### DIFF
--- a/src/tools/pkcs11-tool.c
+++ b/src/tools/pkcs11-tool.c
@@ -9619,11 +9619,11 @@ static CK_SESSION_HANDLE test_kpgen_certwrite(CK_SLOT_ID slot, CK_SESSION_HANDLE
 		fprintf(stderr, "ERR: newly generated private key has no (or an empty) CKA_ID\n");
 		return session;
 	}
-	opt_object_id_len = (size_t) i;
-	if (opt_object_id_len > sizeof(opt_object_id)) {
+	if ((size_t)i > sizeof(opt_object_id)) {
 		fprintf(stderr, "ERR: object ID too long\n");
 		return session;
 	}
+	opt_object_id_len = (size_t)i;
 	memcpy(opt_object_id, tmp, opt_object_id_len);
 
 	/* This is done in NSS */
@@ -9807,11 +9807,11 @@ static void test_ec(CK_SLOT_ID slot, CK_SESSION_HANDLE session)
 		printf("ERR: newly generated private key has no (or an empty) CKA_ID\n");
 		return;
 	}
-	opt_object_id_len = (size_t)i;
-	if (opt_object_id_len > sizeof(opt_object_id)) {
+	if ((size_t)i > sizeof(opt_object_id)) {
 		fprintf(stderr, "ERR: object ID too long\n");
 		return;
 	}
+	opt_object_id_len = (size_t)i;
 	memcpy(opt_object_id, tmp, opt_object_id_len);
 
 	/* This is done in NSS */


frankmorgner:
thank you

PR #3801, opened 2026-08-21
Fix issues reported by Oss-Fuzz

frankmorgner:
##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/common/simclist.c
+++ b/src/common/simclist.c
@@ -209,14 +209,19 @@ static simclist_inline struct list_entry_s *list_findpos(const list_t *simclist_
  * Default to local RNG for user's ease of use.
  */
 
-#ifdef SIMCLIST_SYSTEM_RNG
+#ifndef SIMCLIST_SYSTEM_RNG
 /* keep track whether we initialized already (non-0) or not (0) */
 static unsigned random_seed = 0;
 
 /* use local RNG */
 static simclist_inline void seed_random() {
-    if (random_seed == 0)
+    if (random_seed == 0) {
+#ifndef _WIN32
         random_seed = (unsigned)getpid() ^ (unsigned)time(NULL);
+#else
+        random_seed = (unsigned)GetCurrentProcessId() ^ (unsigned)time(NULL);
+#endif
+    }
 }
 
 static simclist_inline long get_random() {
--- a/src/common/simclist.c
+++ b/src/common/simclist.c
@@ -863,7 +863,7 @@ static void list_sort_quicksort(list_t *simclist_restrict l, int versus,
     /* base of iteration: one element list */
     if (! (last > first)) return;
 
-    pivotid = (get_random() % (last - first + 1));
+    pivotid = (unsigned int)(get_random() % (last - first + 1));
     /* pivotid = (last - first + 1) / 2; */
 
     /* find pivot */
--- a/src/libopensc/log.c
+++ b/src/libopensc/log.c
@@ -341,18 +341,28 @@ int sc_color_fprintf_va(int colors, struct sc_context *ctx, FILE * stream, const
 void _sc_debug_hex(sc_context_t *ctx, int type, const char *file, int line,
 		const char *func, const char *label, const u8 *data, size_t len)
 {
-	size_t blen = len * 5 + 128;
-	char *buf = malloc(blen);
+	size_t blen;
+	char *buf;
+	const char *truncated = "";
+
+	if (len > SC_MAX_EXT_APDU_BUFFER_SIZE) {
+		len = SC_MAX_EXT_APDU_BUFFER_SIZE;
+		truncated = "truncated to ";
+	}
+	blen = len * 5 + 128;
+
+	buf = malloc(blen);
 	if (buf == NULL)
 		return;
 
 	sc_hex_dump(data, len, buf, blen);
 
 	if (label)
-		sc_do_log(ctx, type, file, line, func, "\n%s (%zu byte%s):\n%s", label, len,
-				len == 1 ? "" : "s", buf);
+		sc_do_log(ctx, type, file, line, func, "\n%s (%s%zu byte%s):\n%s",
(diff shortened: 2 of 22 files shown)


frankmorgner:
I piped the open problems from oss-fuzz and coverity scan into AI and got some good results. Also, I took the time and revisited most the coverity reports that we previously dismissed as "intentional" or "false positive".

I proofread all fixes (some needed to be modified). Please review with care.

src/libopensc/pkcs15-coolkey.c
```diff
@@ -675,8 +675,11 @@ static int sc_pkcs15emu_coolkey_init(sc_pkcs15_card_t *p15card)
 		}
 
 		r = sc_pkcs15emu_object_add(p15card, obj_type, &obj_obj, obj_info);
-		if (r != SC_SUCCESS)
+		if (r != SC_SUCCESS) {
 			sc_log(card->ctx, "sc_pkcs15emu_object_add() returned %d", r);
+		} else {
+			cert_info.value.value = NULL; /* Ownership transferred */
+		}
 fail:
```
Jakuje: Do we konw this will solve the coverity issue? It sounds to me odd to set just random pointer in the structure passed into the library function.

src/libopensc/log.c
```diff
@@ -341,8 +341,15 @@ int sc_color_fprintf_va(int colors, struct sc_context *ctx, FILE * stream, const
...
 		const char *func, const char *label, const u8 *data, size_t len)
 {
-	size_t blen = len * 5 + 128;
-	char *buf = malloc(blen);
+	size_t blen;
+	char *buf;
+
+	if (len > SC_MAX_EXT_APDU_BUFFER_SIZE) {
+		len = SC_MAX_EXT_APDU_BUFFER_SIZE;
+	}
```
Jakuje, line 351: if we truncated, should we print some ellipsis after the buffer to make it clear it was truncated in output, rather than somewhere on the wire? I can see that this could result in some very hard to debug issues if some large data is printed in the debug log, but truncated without any warning.

src/libopensc/pkcs15-sc-hsm.c
```diff
@@ -992,8 +992,10 @@ static int sc_pkcs15emu_sc_hsm_get_ec_public_key(struct sc_context *ctx, sc_cvc_
 	memcpy(pubkey->u.ec.params.der.value, ecp->der.value, ecp->der.len);
 	pubkey->u.ec.params.der.len = ecp->der.len;
 
-	/* FIXME: check return value? */
-	sc_pkcs15_fix_ec_parameters(ctx, &pubkey->u.ec.params);
+	r = sc_pkcs15_fix_ec_parameters(ctx, &pubkey->u.ec.params);
+	if (r != SC_SUCCESS) {
+		sc_log(ctx, "Error %d fixing EC parameters, continuing anyway", r);
+	}
```
Jakuje, line 998: the commit message says:

> ensure all allocated memory is properly freed upon failure

but I do not see this happening anywhere in the code. Can you clarify if the message is hallucinated and needs to be updated or the fix is incomplete?

src/common/simclist.c
```diff
@@ -209,14 +209,14 @@ static simclist_inline struct list_entry_s *list_findpos(const list_t *simclist_
...
-#ifdef SIMCLIST_SYSTEM_RNG
+#ifndef SIMCLIST_SYSTEM_RNG
 /* keep track whether we initialized already (non-0) or not (0) */
 static unsigned random_seed = 0;
 
 /* use local RNG */
 static simclist_inline void seed_random() {
     if (random_seed == 0)
-        random_seed = (unsigned)getpid() ^ (unsigned)time(NULL);
+        random_seed = 123456789 ^ (unsigned)time(NULL);
```
Jakuje: is the `getpid()` portability a concern?

Jakuje (commented):
the strict build also needs to be fixed

src/libopensc/log.c
frankmorgner: added note below

src/common/simclist.c
frankmorgner: I added the windows specific callback, though I'm not sure if this level of care is really needed.

src/libopensc/pkcs15-sc-hsm.c
frankmorgner: this was not hallucinated. It was suggested to treat an error in sc_pkcs15_fix_ec_parameters as fatal (and consequently freeing all data), but I changed the error handling to only debug the error and continue anyway to keep the logic that we currently have.

I reworded the commit message

src/libopensc/pkcs15-coolkey.c
frankmorgner: i guess the more readable way to express this would be `memset(&cert_info, 0, sizeof(cert_info));`, but if you prefer then I drop this commit from the PR.

frankmorgner: I drafted a new change, which makes transfer of ownership more transparent and more robust for other data types as well:
https://github.com/frankmorgner/OpenSC/commit/e6dad4d864b9dadd839905752ad468fe244a9b44

Could you check if that makes more sense?

src/libopensc/pkcs15-sc-hsm.c
Jakuje: Thanks for clarification!

src/libopensc/pkcs15-coolkey.c
Jakuje: Yeah, this looks cleaner. Thanks!

frankmorgner:
Modified the commit (messages) as discussed

Jakuje:
still the strict build needs some explicit cast, otherwise lgtm

Jakuje (approved):

Jakuje:
still ack. I would also consider including some formatting changes.

frankmorgner:
done, thanks for the review

PR #3806, opened 2026-08-31
itacns: recognize the ST2022 Italian CNS card over the contactless interface

### LINKED ISSUE #3755: Italian CNS card (ST2022) not recognized
### Describe the bug

I was sent this replacement card a few months ago but it appears to be unsupported by OpenSC.

### Steps to reproduce

1. Attempt any interaction with the card

### Expected behavior

Certificate list or card identification at least.

### Actual behavior

Any attempt returns "invalid" or "unsupported" card error.

### OpenSC version

git master from 20260729.

### Operating system

openSUSE Leap 16.0

CNS ST2022

So far only command line

### Logs and debug output

```shell
$ pkcs11-tool --test --login
error: PKCS11 function C_GetTokenInfo failed: rv = CKR_TOKEN_NOT_RECOGNIZED (0xe1)
Aborting.
```

### Checklist

ale-rinaldi:
Over NFC my ST2022 (TS-CNS) answers a bare ATR `3B 80 80 01 01` with no historical bytes, so `itacns` — which keys on the `CNS` marker in the historical bytes — doesn't recognise it (#3755).

When there are no historical bytes, the driver now identifies the card by selecting its application (`A0 00 00 00 95 01 00 4D`) and confirming the model with `GET DATA 00 CA 01 80` → `JSign4`, so an unrelated contactless card answering the same AID isn't matched. Identifying by the application rather than the ATR follows `card-isoApplet.c`, and validating the select response follows `card-sc-hsm.c` (#1377) — it also matches @dengert's suggestion on the issue.

The application is kept selected in `init` (the CNS files live under it over NFC) and extended APDUs + RSA-2048 are enabled. Everything is gated on `hist_bytes_len == 0`, so the contact path is untouched.

Tested with a real ST2022 over NFC: recognition, enumeration, certificate read, a signature verified against the card's public key, and PIN change. I also checked the match against a few virtual cards (vpcd) — absent/wrong AID, a different model id, and a contact ATR are all rejected or left to the historical-bytes path. Other CNS models of the same family may work too, but ST2022 is the only one I could test.

Closes #3755.

### DIFF
--- a/src/libopensc/card-itacns.c
+++ b/src/libopensc/card-itacns.c
@@ -81,11 +81,49 @@ static void itacns_init_cns_card(sc_card_t *card)
 	}
 }
 
+/* Over the contactless interface the card has no historical bytes; it is
+ * identified by selecting its application instead of by ATR. */
+static const u8 itacns_cns_aid[] = {
+		0xA0, 0x00, 0x00, 0x00, 0x95, 0x01, 0x00, 0x4D};
+
+/* Confirm the card model after selecting the application, so an unrelated card
+ * answering the same AID is not matched (cf. card-sc-hsm.c, issue #1377). */
+static int
+itacns_match_cns_aid(sc_card_t *card)
+{
+	static const u8 model[] = {'J', 'S', 'i', 'g', 'n', '4'};
+	sc_apdu_t apdu;
+	u8 rbuf[16];
+	int r;
+
+	if (iso7816_select_aid(card, itacns_cns_aid, sizeof itacns_cns_aid,
+			    NULL, NULL) != SC_SUCCESS)
+		return 0;
+
+	sc_format_apdu(card, &apdu, SC_APDU_CASE_2_SHORT, 0xCA, 0x01, 0x80);
+	apdu.le = sizeof model;
+	apdu.resp = rbuf;
+	apdu.resplen = sizeof rbuf;
+	r = sc_transmit_apdu(card, &apdu);
+	if (r < 0 || sc_check_sw(card, apdu.sw1, apdu.sw2) != SC_SUCCESS)
+		return 0;
+
+	return apdu.resplen >= sizeof model &&
+	       0 == memcmp(rbuf, model, sizeof model);
+}
+
 static int itacns_match_cns_card(sc_card_t *card)
 {
 	u8 manufacturer_code;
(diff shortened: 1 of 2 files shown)


frankmorgner:
While I do support the detection method, I don't understand why you need to re-select the AID in `init()` again, because the AID should already be selected from `match_card()`.

With what you have implemented, you seem to imply that the card requires additional selections of the AID when communicating via NFC. Could you elaborate on this?

ale-rinaldi:
You're right, that's over-cautious on my part. In the normal flow `init()` runs right after `match_card()` on the same handle, so the AID is already selected and re-selecting it is redundant — the card doesn't need repeated selections over NFC.

The one thing it still covers is a forced driver from `opensc.conf` (a pinned `driver`/`type`, or a `card_atr` mapping): there `match_card()`'s result is non-fatal and `init()` runs anyway, so if the AID hadn't been selected in `match_card()` it wouldn't be selected at all. It's a corner case — `match_card()` selects the AID before any check, so it only bites if the applet isn't selectable to begin with, but other driver's init() may select other AIDs after it.

I'm happy to drop the re-select and just rely on `match_card()`, or keep it for that forced-driver corner. Do you think the case is worth keeping?

ale-rinaldi:
Sorry @frankmorgner , I looked into it a little more and I was wrong. Even in my edge case, only the match_card() for the specific driver I forced runs, so there's no way I can reach init() if not after the right match_card() and nothing else. My fault, I removed it.

frankmorgner:
I think it would be better to keep it simple, because, if I understand correctly, this change is only required in order to detect the card if it only comes with a generic AID. If so, `is_nfc` is not really needed, which simplifies the code.

However, from your explanation above, I don't exactly know if the card requires an AID to be selected in order to be functional and whether that is only the case via the NFC interface or not. If such a state may be required, then there are some other things that need to be done to cope with concurrently running processes that access the card....

frankmorgner (changes requested):
@ale-rinaldi , do you consider removing is_nfc and soly rely on the AID detection?

ale-rinaldi:
Hi @frankmorgner I'm really sorry but I had to move on urgent topics that didn't let me concentrate on this. That's on my todo list, I'll get there in a few days hopefully!

PR #3811, opened 2026-09-01
myeid: alternative way to unblock user PINs using another PIN marked with the unblockingPin flag

hhonkanen:
When creating a PIN to MyEID, a PUK is always created and directly attached to the PIN object. A separate PKCS#15 object is not created for the PUK. 

MyEID also supports unblocking PINs using another PIN. To enable this, the PIN being unblocked must have the "allow global unblocking" flag set and the other PIN must have the "global unblocker" flag set. This configuration is needed, in case there is a need to be able to change the unblocking PIN after initialization and finalization of the card.

This PR adds support for this configuration. 

New profiles have been added for creating the PINs. I considered using --puk-id switch to create the unblocking PIN within the same command when the PIN is created. However, this does not work with MyEID because the directly attached PUK is also created and you would need to set two different PUK values. In a practical usage scenario, the directly attached PUK could be random and stored either nowhere or in some secure location, where the unblocking PIN could be set with a known default value, to be changed later by a card management system. Therefore, the unblocking PIN is created like a normal user PIN, with a special version of the profile file.

This PR can be tested with the following commands (Starting with an empty MyEID card) :

**initialize the card**

**create user PIN 1**
pkcs15-init -P --card-profile myeid_unblock_with_other_pin -a 1 -l "Basic PIN" --pin 1111 --puk 12345678 --so-pin 12345678

**create second user PIN**
pkcs15-init -P --card-profile myeid_unblock_with_other_pin -a 2 -l "Signature PIN" --pin 2222 --puk 12345678 --so-pin 12345678

**create the unblocking PIN**
pkcs15-init -P --card-profile myeid_unblocking_pin -a 4 -l "Unblocking PIN" --pin 44444444 --puk 12345678 --so-pin 123456786

**unblock the PIN. Enter 44444444 when the PUK is requested **
pkcs15-tool -u -a 1 

or using pkcs11-tool:
pkcs11-tool --module ..\pkcs11\opensc-pkcs11.dll --unlock-pin

Testing with pkcs11-tool requires the following option in opensc.conf:

app opensc-pkcs11 {
    pkcs11 {
        # Configures SC_PKCS11_PIN_UNBLOCK_UNLOGGED_SETPIN behavior
        user_pin_unblock_style = set_pin_in_unlogged_session;
    }
}

### About testing:

- verified the the correct flags get set to MyEID native PINs with the commands above
- verified that using the default profile the PINs are initialized like before this PR (with no unblocking-related flags)
- verified that unblocking using pkcs11-tool and pkcs15-tool works both ways: with the classic MyEID configuration, using directly attached PUK, and using the separate unblocker PIN. OpenSC uses the unblocker PIN if it finds **auth-id** from the CommonObjectAttributes of the PIN being unblocked.

This change should not affect how minidriver and tokend work, at least in case the card is personalized using the default profile. I haven't tested the minidriver and don't have a MacOS system to test with.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -693,6 +693,7 @@ static int myeid_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data)
 			global_unblocker_verif_data.cmd = SC_PIN_CMD_VERIFY;
 			global_unblocker_verif_data.pin1 = data->pin1;
 			global_unblocker_verif_data.pin_reference = data->puk_reference;
+			global_unblocker_verif_data.flags = data->flags;
 			r = iso_ops->pin_cmd(card, &global_unblocker_verif_data);
 			LOG_TEST_RET(card->ctx, r, "APDU transmit failed");
 
--- a/src/libopensc/card-myeid.c
+++ b/src/libopensc/card-myeid.c
@@ -1,7 +1,7 @@
 /*
  * card-myeid.c
  *
- * Copyright (C) 2008-2019 Aventra Ltd.
+ * Copyright (C) 2008-2026 Aventra Ltd.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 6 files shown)


src/libopensc/card-myeid.c
```diff
@@ -669,15 +669,63 @@ static int myeid_pin_cmd(sc_card_t *card, struct sc_pin_cmd_data *data)
...
+			/* Build the RESET RETRY COUNTER APDU here as we must format it in a different way in this case */
+			sc_format_apdu(card, &apdu, SC_APDU_CASE_3_SHORT, 0x2C, 0x00, data->pin_reference);
+			if ((r = sc_build_pin(buf, buf_len, &data->pin1, pad)) < 0)
+				return r;
+			apdu.lc = r;
+			apdu.datalen = r;
+			apdu.data = buf;
+			apdu.resplen = 0;
+
+			r = sc_transmit_apdu(card, &apdu);
```
frankmorgner, line 718: can this be done with pinpad support on the reader side, e.g. via `iso_ops->pin_cmd`?

frankmorgner (commented):
The change looks reasonable at first glance. what kind of input do you need to decide whether or not to create a PR out of it?

src/libopensc/card-myeid.c
hhonkanen: It didn't work by calling iso_ops->pin_cmd directly and I found out that's because MyEID's implementation of implicit unblock contradicts with ISO 7816-4 regarding value of P1. MyEID expects it to be zero while the standard says it should be 02h when only new reference data is in the data field. I changed this to use iso7816_build_pin_apdu to build the command, change P1 to 0 and call sc_transmit_apdu from card_myeid. This should now take pinpads into account.

We will fix this into the next MyEID version, keeping also P1=0 allowed for compatibility.

I wanted to get the automated checks passed before releasing this as a PR. Most are passing now (the CentOS builds are not, but looks like the problem might be in the pipeline/environment) so I can make it a PR now.

hhonkanen:
Did more testing and noticed that the unblocking PIN was not padded properly. Resolved this by copying the flags from the original sc_pin_cmd_data.

frankmorgner (changes requested):
please add the new files to the distributed files in `src/pkcs15init/Makefile.am`. 

Preferably, you also add information about the different personalization (and usage) options to https://github.com/OpenSC/OpenSC/wiki/Aventra-MyEID-PKI-card (later)

hhonkanen:
> please add the new files to the distributed files in `src/pkcs15init/Makefile.am`.
> 
> Preferably, you also add information about the different personalization (and usage) options to https://github.com/OpenSC/OpenSC/wiki/Aventra-MyEID-PKI-card (later)

@frankmorgner Thank you for feedback. I updated Makefile.am. I will update the wiki page too after the PR is merged.

frankmorgner (approved):

frankmorgner:
thank you

PR #3815, opened 2026-09-03
tools: Check util_getpass return value in sc-hsm-tool and gids-tool

Strykar:
`util_getpass()` bails out before assigning `*lineptr`:

```c
if (tcgetattr (fileno (stdout), &old) != 0)
    return -1;
```

`tcgetattr` fails whenever stdout is not a TTY, so any redirect or pipe on
stdout makes the function return -1 with the caller's pointer untouched. Every
call site in `sc-hsm-tool.c` and `gids-tool.c` ignored the return value.

In `sc-hsm-tool` the consequence is a crash, because the pointer goes straight
into `strlen()`:

```c
char *lpin = NULL;
util_getpass(&lpin, NULL, stdin);
data.pin1.len = strlen(lpin);
```

### Reproducing

Any prompt reached without the matching command line option, with stdout
redirected:

```console
$ printf 'x\n' | sc-hsm-tool --unwrap-key blob.bin --key-reference 3 > log
$ echo $?
139
```

With this branch:

```console
$ printf 'x\n' | ./src/tools/sc-hsm-tool --unwrap-key blob.bin --key-reference 3 > log
$ echo $?
1
$ tail -1 log
Enter User PIN : Error reading User PIN
```

The crash happens before the PIN is sent to the card, so it consumes no retry
counter and leaves the card untouched.

### The change

Check the return value at all 15 call sites in the two files and fail with a
message. This is what `pkcs11-tool.c` and `opensc-explorer.c` already do, so the
fix follows existing practice rather than introducing a new convention. Both
files report errors with `fprintf(stderr, ...)` and a negative return
throughout, and the patch keeps to that rather than pulling in `util_fatal()`.

`ask_for_password()` was `static void` with no way to report failure, so it gains
an `int` return. It has a single caller. Its cleanup now guards `refpwd` against
NULL, which the new early-exit path can reach.

Two things deliberately left alone:

- `util_getpass()` itself is unchanged. Testing `tcgetattr` on stdout rather
  than stdin looks questionable, but changing it alters behaviour for every
  caller and does not belong in a crash fix.
- Other tools calling `util_getpass` without checking are untouched, since I
  have not reproduced a crash in them.

### Testing

Card: SmartCard-HSM version 4.1 (Nitrokey HSM 2).

- Built clean, no new warnings in either file.
- `sc-hsm-tool --unwrap-key` with redirected stdout: SIGSEGV before, exit 1 with
  a diagnostic after.
- `sc-hsm-tool --unwrap-key` with `--pin` supplied: unchanged, key imports and
  signs.
- `gids-tool` changes are the same mechanical pattern; I have no GIDS card and
  have not exercised them at runtime.

##### Checklist

Nothing on the checklist applies to this change.

### DIFF
--- a/src/tools/gids-tool.c
+++ b/src/tools/gids-tool.c
@@ -102,7 +102,10 @@ static int initialize(sc_card_t *card, const char *so_pin, const char *user_pin,
 	if (so_pin == NULL) {
 		printf("Enter admin key (48 hexadecimal characters) : \n");
 		printf("Press Enter to set the admin key to 00...00\n");
-		util_getpass(&_so_pin, NULL, stdin);
+		if (util_getpass(&_so_pin, NULL, stdin) < 0) {
+			fprintf(stderr, "Error reading admin key\n");
+			return -1;
+		}
 		printf("\n");
 	} else {
 		_so_pin = (char *)so_pin;
@@ -122,7 +125,10 @@ static int initialize(sc_card_t *card, const char *so_pin, const char *user_pin,
 
 	if (user_pin == NULL) {
 		printf("Enter initial User-PIN (4 - 16 characters) : ");
-		util_getpass(&_user_pin, NULL, stdin);
+		if (util_getpass(&_user_pin, NULL, stdin) < 0) {
+			fprintf(stderr, "Error reading User-PIN\n");
+			return -1;
+		}
 		printf("\n");
 	} else {
 		_user_pin = (char *)user_pin;
@@ -131,7 +137,10 @@ static int initialize(sc_card_t *card, const char *so_pin, const char *user_pin,
 	if (serial == NULL) {
 		printf("Enter serial number (32 hexadecimal characters): \n");
 		printf("Press Enter to set a random serial number\n");
-		util_getpass(&_serial, NULL, stdin);
+		if (util_getpass(&_serial, NULL, stdin) < 0) {
+			fprintf(stderr, "Error reading serial number\n");
+			return -1;
+		}
 		printf("\n");
 	} else {
 		_serial = (char *)serial;
@@ -189,7 +198,10 @@ static int unblock(sc_card_t* card, const char *so_pin, const char *user_pin) {
 		printf("WARNING\n");
 		printf("============================================================\n");
 		printf("Enter admin key (48 hexadecimal characters) : ");
-		util_getpass(&_so_pin, NULL, stdin);
+		if (util_getpass(&_so_pin, NULL, stdin) < 0) {
+			fprintf(stderr, "Error reading admin key\n");
+			return -1;
+		}
 		printf("\n");
 	} else {
 		_so_pin = (char *)so_pin;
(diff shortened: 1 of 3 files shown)


frankmorgner (approved):

frankmorgner:
The criticism of modifying `stdout` rather than `stdin` in `util_getpass()` sounds reasonable, but maybe @Jakuje knows better...

CardContact:
When merged, I will rebase #3786 on it.

Jakuje:
> util_getpass() itself is unchanged. Testing tcgetattr on stdout rather
than stdin looks questionable, but changing it alters behaviour for every
caller and does not belong in a crash fix.

> The criticism of modifying `stdout` rather than `stdin` in `util_getpass()` sounds reasonable, but maybe @Jakuje knows better...

Changing terminal modes on stdout is correct way to prevent password being echoed back to the terminal. It also solves the issue that if you redirect the stdout to file, the application can print the prompt for you, but its redirected to a file so you don't know you are prompted.

Failing cleanly in this case is probably the right thing to do. We might be more verbose directly inside of the `util_getpass()`  and print some warning/error like "WARNING: not prompting for password, because stdout is not a TTY`.

I would not affraid to change also the other cases where the return value is not checked. Its just matter of time somebody else will hit this issue with the other tools.

Jakuje (approved):

Strykar:
Thanks both.

On stdout versus stdin: you are right, my PR description was wrong, and `util_getpass()` is unchanged.

I have added the warning you suggested. It goes to stderr, so it is still
visible in precisely the case that triggers it:

```console
$ printf 'x\n' | sc-hsm-tool --unwrap-key blob.bin --key-reference 3 > log
WARNING: not prompting for input, because stdout is not a TTY
Error reading User PIN
$ echo $?
1
```

I wrote "input" rather than "password" because `gids-tool` also uses
`util_getpass()` to read a serial number. Happy to reword.

On extending the return value check to the other tools: I went through every
remaining call site and they all check it already.

| tool | sites | check |
| --- | --- | --- |
| `dtrust-tool.c` | 2 | `if (r < 0 \|\| *pin == NULL)` |
| `lteid-tool.c` | 1 | `if (number_len < min_len ...)`, so -1 fails the range test |
| `opensc-explorer.c` | 1 | `if (r < 0)` |
| `pkcs15-tool.c` | 1 | `if (r < 0)` |
| `pkcs15-init.c` | 5 | `if (r < 0 \|\| !x)` |
| `pkcs11-tool.c` | 11 | `if (r < 0)` then `util_fatal()` or `return 1` |

`sc-hsm-tool.c` and `gids-tool.c` were the only two that did not, so this PR
already covers the sweep. Glad to add anything further if you spot something I
have missed.

frankmorgner:
Thanks, @Jakuje for the explanation and thanks for adding the warning @Strykar 

I approved the CI run to get additional confirmation.

Jakuje (approved):

PR #3817, opened 2026-09-04
pkcs15-sec: zero-pad short EC digests to the field length

### LINKED ISSUE #3813: CardSignData fails with "Incorrect parameters in APDU" (-1205) when Windows signs a 32-byte digest with an EC P-384 key — breaks offline smart-card logon
### Describe the bug

Hi, and thanks for OpenSC — the minidriver made native Windows PIV logon possible for us
with no third-party code in the authentication path, which we really appreciate.

## Component

**Windows Minidriver** (`opensc-minidriver.dll`), but the defect is in **libopensc**
(`src/libopensc/pkcs15-sec.c`), so `opensc-pkcs11` is affected identically — it reproduces
with `pkcs11-tool` alone (step 5 below), no Windows logon needed.

Card — `opensc-tool --name` and `--atr`:

```
Using reader with a card: Identiv SCR3500 C Contact Reader 0

ATR: 3b:fa:18:00:00:91:01:31:fe:45:4a:33:52:33:31:30:2d:33:35:35:ff
```

A JavaCard running the open-source **OpenFIPS201** PIV applet on NXP JCOP, read through
an Identiv SCR3500 contact reader.

## Version number

**OpenSC 0.27.1**, installed from the official Windows MSI. `opensc-tool --version` on
this build prints only:

```
No Git revision info available
```

(no version/revision line — presumably how the release MSI is built, so I am stating the
version from the installer rather than from the tool.) I also built and tested from the
**0.27.1 release tarball**, which is where the line numbers below come from.

**On "test against the latest development build":** I have not run a Nightly binary 
 I did check **master** source, where the relevant branch is
unchanged — `src/libopensc/pkcs15-sec.c` line ~765 on master is byte-identical to 0.27.1 —
so I do not believe this is already fixed. Happy to be corrected.

## Operating system

Windows 11 Pro, **10.0.26200.9168** (x64).

## Steps to reproduce
1. Provision a PIV card with an **ECDSA P-384** key and certificate in slot **9A**
   (Smart Card Logon + Client Auth EKU, UPN SAN). Any CA will do.
2. Install OpenSC on Windows and register the card's ATR so the OpenSC **minidriver**
   handles it (the inbox minidriver cannot read P-384).
3. Join an AD domain that accepts PKINIT and log on **online** with card + PIN, so
   credentials are cached. This succeeds.
4. Disconnect the network, sign out, and log on again with card + PIN.
   → fails; see *Actual behaviour*.
5. Simpler repro, no logon involved — sign a **32-byte** and then a **48-byte** input with
   the P-384 key:
   ```
   pkcs11-tool --login --sign --mechanism ECDSA --id 01 \
               --input-file digest32.bin --output-file sig.bin      # fails
   pkcs11-tool --login --sign --mechanism ECDSA --id 01 \
               --input-file digest48.bin --output-file sig.bin      # succeeds
   ```

## Expected behaviour

Both signatures in step 5 succeed, and the offline logon in step 4 works like the online
logon in step 3.

A digest shorter than the field size is not an error condition: zero-**left**-padding it to
the field size is numerically a no-op (leading zeros do not change the integer), so the
signature still verifies against the verifier's own short digest, per FIPS 186-4 §6.4.

## Actual behaviour

Step 4 fails with *"The domain specified is not available."* Windows Security event 4625
gives sub-status `0xC0000380` (`STATUS_SMARTCARD_WRONG_PIN`) — misleading, as the PIN is
correct and verifies.

Step 5 fails on 32 bytes and succeeds on 48:

| Digest presented to slot 9A (EC P-384) | Result |
|---|---|
| **32 bytes** (SHA-256 size) | **fails** |
| **48 bytes** (SHA-384 size) | succeeds |

Minidriver log, failing case:

```
CardSignData dwVersion=2, bContainerIndex=0, dwKeySpec=4 (AT_ECDSA_P384),
             dwSigningFlags=0x00000000, aiHashAlg=0x00000000
pInfo->pbData(32)          <- 32-byte digest for a P-384 key
cbSignedData = 96 ; lg = 96
Data to sign: 32 bytes
sc_pkcs15_compute_signature return -1205
sc_pkcs15_compute_signature error "Incorrect parameters in APDU"
CardSignData returning 0x80100004 (SCARD_E_INVALID_PARAMETER)
```

Windows signs a **32-byte SHA-256** digest for cached-credential operations regardless of
key curve, so a P-384 key receives 32 bytes where the applet requires 48.

## Cause

`sc_pkcs15_compute_signature()` handles the digest/field mismatch in both directions for
RSA:

```c
    if (inlen < modlen) { ... memmove(tmp+modlen-inlen, tmp, inlen); ... }
    inlen = modlen;
}
```

…but the EC branch immediately below only truncates:

```c
/* PKCS#11 MECHANISMS V2.30: 6.3.1 EC Signatures ... */
    inlen = MIN(inlen, BYTES4BITS(prkey->field_length));   /* long: handled  */
}                                                          /* short: falls through */
```

The code already anticipates a mismatch for EC — it just implements one half of it. So this
looks like a completeness gap in an existing branch rather than a missing feature.

## Suggested fix

Make that branch symmetric:

```c
    size_t flen = BYTES4BITS(prkey->field_length);   /* 48 for P-384 */
    if (inlen > flen) {
        inlen = flen;                                /* existing behaviour */
    } else if (inlen < flen) {
        if (flen > buflen) { r = SC_ERROR_BUFFER_TOO_SMALL; goto err; }
        memmove(tmp + flen - inlen, tmp, inlen);
        memset(tmp, 0, flen - inlen);
        inlen = flen;
    }
}
```

With this applied, offline Windows smart-card logon on a P-384 credential works.

## Scope

**P-256 is unaffected** — its 32-byte field size coincidentally matches the digest Windows
sends, which is probably why this has not surfaced much.

**Tested on P-384 only.** The fix keys off `BYTES4BITS(prkey->field_length)` so it should
apply equally to **P-521** (66-byte field), but I did not exercise that: NIST SP 800-78
permits only P-256 and P-384 for PIV keys, so I had no P-521 card. Treat P-521 as
expected-to-work rather than verified.

Arguably three parties contribute — Windows pairs SHA-256 with a P-384 key, the applet
demands an exact length, and OpenSC forwards without adapting — but OpenSC already solves
exactly this for RSA, so this seemed the natural place to raise it.

Searched the tracker first (`ECDSA pad digest`, `offline logon P-384`, `Incorrect
parameters in APDU`); nothing open covers this. #3441 shares the `-1205` string but is an
unrelated eID/TLS case.

Glad to open a PR, test patches, or supply fuller logs — whichever is easiest for you.

### Steps to reproduce

1. Provision a PIV card with an **ECDSA P-384** key and certificate in slot **9A**
   (Smart Card Logon + Client Auth EKU, UPN SAN). Any CA will do.
2. Install OpenSC on Windows and register the card's ATR so the OpenSC **minidriver**
   handles it (the inbox minidriver cannot read P-384).
3. Join an AD domain that accepts PKINIT and log on **online** with card + PIN, so
   credentials are cached. This succeeds.
4. Disconnect the network, sign out, and log on again with card + PIN.
   → fails; see *Actual behaviour*.
5. Simpler repro, no logon involved — sign a **32-byte** and then a **48-byte** input with
   the P-384 key:
   ```
   pkcs11-tool --login --sign --mechanism ECDSA --id 01 \
               --input-file digest32.bin --output-file sig.bin      # fails
   pkcs11-tool --login --sign --mechanism ECDSA --id 01 \
               --input-file digest48.bin --output-file sig.bin      # succeeds
   ```

### Expected behavior

Both signatures in step 5 succeed, and the offline logon in step 4 works like the online
logon in step 3.

A digest shorter than the field size is not an error condition: zero-**left**-padding it to
the field size is numerically a no-op (leading zeros do not change the integer), so the
signature still verifies against the verifier's own short digest, per FIPS 186-4 §6.4.

### Actual behavior

Step 4 fails with *"The domain specified is not available."* Windows Security event 4625
gives sub-status `0xC0000380` (`STATUS_SMARTCARD_WRONG_PIN`) — misleading, as the PIN is
correct and verifies.

Step 5 fails on 32 bytes and succeeds on 48:

| Digest presented to slot 9A (EC P-384) | Result |
|---|---|
| **32 bytes** (SHA-256 size) | **fails** |
| **48 bytes** (SHA-384 size) | succeeds |

Minidriver log, failing case:

```
CardSignData dwVersion=2, bContainerIndex=0, dwKeySpec=4 (AT_ECDSA_P384),
             dwSigningFlags=0x00000000, aiHashAlg=0x00000000
pInfo->pbData(32)          <- 32-byte digest for a P-384 key
cbSignedData = 96 ; lg = 96
Data to sign: 32 bytes
sc_pkcs15_compute_signature return -1205
sc_pkcs15_compute_signature error "Incorrect parameters in APDU"
CardSignData returning 0x80100004 (SCARD_E_INVALID_PARAMETER)
```

Windows signs a **32-byte SHA-256** digest for cached-credential operations regardless of
key curve, so a P-384 key receives 32 bytes where the applet requires 48.

### OpenSC version

0.27.1

### Operating system

Windows 11 Pro 10.0.26200.9168 (x64)

Identiv uTrust FIDO2 PIV card (NXP JCOP 4.5 / P71), OpenFIPS201 applet, Identiv SCR3500 C reader

Windows Logon

### Logs and debug output

```shell

```

### Checklist

spurglegooge:
Fixes #3813

Windows signs a 32-byte SHA-256 digest with a P-384 key during
cached-credential (offline) logon. `sc_pkcs15_compute_signature()` truncates a
digest that is longer than the field size but forwards a shorter one unchanged,
so the applet receives 32 bytes where it expects 48.

The OpenFIPS201 applet requires the input to equal the field length and rejects
the short one with "Incorrect parameters in APDU" (-1205). Windows surfaces this
as "The domain specified is not available", which is why the cause was not
obvious.

This makes the branch symmetric by zero-padding on the left, mirroring the RSA
`SC_ALGORITHM_RSA_PAD_NONE` branch a few lines above, including its
`SC_ERROR_BUFFER_TOO_SMALL` guard.

Leading zeros do not change the integer value of the digest, so a card that
already accepts a short input produces the same signature as before. Cards that
require the exact field length now work.

### Testing

Reproduced and fixed end to end on Windows 11 24H2 with a P-384 9A key
(OpenFIPS201 applet), against a Samba AD DC: offline smart-card logon fails
before the change and succeeds after it.

P-256 is unaffected because a 32-byte digest already equals the field size.
The P-521 path is unexercised — only P-256 and P-384 available for
PIV keys, so no card was available to test it.

Note that the fix was originally developed against the 0.27.1 tarball and has
been rebased onto master for this PR; the rebased version is compiled but the
card test above was performed with the 0.27.1 build.

##### Checklist
- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -762,23 +762,25 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	 */
 	else if (senv.algorithm == SC_ALGORITHM_EC &&
 			(senv.algorithm_flags & SC_ALGORITHM_ECDSA_HASHES) == 0) {
+		size_t flen = BYTES4BITS(prkey->field_length);
+
 		if (inlen > flen) {
-   			inlen = flen;
-   		} else if (inlen < flen) {
-   			/* Zero-pad short digests: some applets require the input to match
-   			 * the field length exactly. Leading zeros do not change the integer
-   			 * value, so the signature still verifies. */
-   			if (flen > buflen) {
-   				r = SC_ERROR_BUFFER_TOO_SMALL;
-   				goto err;
-   			}
-   			memmove(tmp + flen - inlen, tmp, inlen);
-   			memset(tmp, 0, flen - inlen);
-   			inlen = flen;
-   		}
+			inlen = flen;
+		} else if (inlen < flen) {
+			/* Zero-pad short digests: some applets require the input to match
+			 * the field length exactly. Leading zeros do not change the integer
+			 * value, so the signature still verifies. */
+			if (flen > buflen) {
+				r = SC_ERROR_BUFFER_TOO_SMALL;
+				goto err;
+			}
+			memmove(tmp + flen - inlen, tmp, inlen);
+			memset(tmp, 0, flen - inlen);
+			inlen = flen;
+		}
 	}
 
-	
+
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
 	LOG_TEST_GOTO_ERR(ctx, r, "use_key() failed");
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -759,6 +759,9 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	 * the leftmost bits of the hash up to the length of n will be used. Any
 	 * truncation is done by the token.
 	 * But if card is going to do the hash, pass in all the data
+	 * A hash shorter than the field length is zero-padded on the left, because
+	 * some applets require the input to match the field length exactly. Leading
+	 * zeros do not change the integer value, so the signature is unaffected.
 	 */
 	else if (senv.algorithm == SC_ALGORITHM_EC &&
 			(senv.algorithm_flags & SC_ALGORITHM_ECDSA_HASHES) == 0) {
@@ -767,9 +770,6 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 		if (inlen > flen) {
 			inlen = flen;
 		} else if (inlen < flen) {
-			/* Zero-pad short digests: some applets require the input to match
-			 * the field length exactly. Leading zeros do not change the integer
-			 * value, so the signature still verifies. */
 			if (flen > buflen) {
 				r = SC_ERROR_BUFFER_TOO_SMALL;
 				goto err;
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -762,9 +762,21 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 	 */
 	else if (senv.algorithm == SC_ALGORITHM_EC &&
 			(senv.algorithm_flags & SC_ALGORITHM_ECDSA_HASHES) == 0) {
-		inlen = MIN(inlen, BYTES4BITS(prkey->field_length));
+		if (inlen > flen) {
+   			inlen = flen;
+   		} else if (inlen < flen) {
+   			/* Zero-pad short digests: some applets require the input to match
+   			 * the field length exactly. Leading zeros do not change the integer
+   			 * value, so the signature still verifies. */
+   			if (flen > buflen) {
+   				r = SC_ERROR_BUFFER_TOO_SMALL;
+   				goto err;
+   			}
+   			memmove(tmp + flen - inlen, tmp, inlen);
+   			memset(tmp, 0, flen - inlen);
+   			inlen = flen;
+   		}
 	}
-
 
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -780,7 +780,6 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
 		}
 	}
 
-
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
 	LOG_TEST_GOTO_ERR(ctx, r, "use_key() failed");
--- a/src/libopensc/pkcs15-sec.c
+++ b/src/libopensc/pkcs15-sec.c
@@ -778,6 +778,7 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
    		}
 	}
 
+	
 	r = use_key(p15card, obj, &senv, sc_compute_signature, tmp, inlen,
 			out, outlen);
 	LOG_TEST_GOTO_ERR(ctx, r, "use_key() failed");


metsma:
Related:
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-esteid2018.c#L191-L193
https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-esteid2025.c#L146-L148

src/libopensc/pkcs15-sec.c
```diff
@@ -762,7 +762,22 @@ int sc_pkcs15_compute_signature(struct sc_pkcs15_card *p15card,
...
 			(senv.algorithm_flags & SC_ALGORITHM_ECDSA_HASHES) == 0) {
-		inlen = MIN(inlen, BYTES4BITS(prkey->field_length));
+		size_t flen = BYTES4BITS(prkey->field_length);
+
+		if (inlen > flen) {
+			inlen = flen;
+		} else if (inlen < flen) {
+			/* Zero-pad short digests: some applets require the input to match
+			 * the field length exactly. Leading zeros do not change the integer
+			 * value, so the signature still verifies. */
```
frankmorgner: This comment should be merged and harmonized with the comment above the whole block for ECC.

frankmorgner (changes requested):

frankmorgner:
Maybe we could also add some more exotic configurations in CI, too... (see `.github/test-piv.sh` for existing configurations)

frankmorgner (approved):

spurglegooge:
Merged into the block comment above and dropped the inline one — thanks, that reads better.
Also removed a stray blank line that `git-clang-format-lint` flagged.

@metsma: that is the same fix one layer down, and a useful check that this does not
disturb it. Both drivers use `SIGNATURE_PAYLOAD_SIZE 0x30` with EC-384 keys, so
`BYTES4BITS(prkey->field_length)` is the same 48. After padding here the drivers receive
`datalen == SIGNATURE_PAYLOAD_SIZE`, their `datalen > SIGNATURE_PAYLOAD_SIZE` guard still
passes, and their own left-pad becomes an identity copy — the APDU is byte-identical
either way. I would leave the driver-side padding in place regardless, since those paths
can be reached without going through `pkcs15-sec.c`.

The reason for doing it in the shared layer is that it is not card-specific: the RSA
`SC_ALGORITHM_RSA_PAD_NONE` branch a few lines above already zero-pads to the modulus, and
the OpenFIPS201 applet needs the same for EC — it rejects a 32-byte input to a P-384 key
with `SC_ERROR_INCORRECT_PARAMETERS`.

On CI: adding a P-384 configuration to `.github/test-piv.sh` would regression-test exactly
this, since the current script generates `ECCP256` for 9c and 9d, where a 32-byte digest
already matches the field size and the bug cannot appear. Signing with `ECDSA-SHA256`
against a P-384 key is what triggers it. Happy to put that up as a separate PR if useful —
one thing to confirm first is whether PivApplet supports `ECCP384`, which I have not
checked.

Jakuje:
ECDSA P-384 is supported so I would prefer to have that in the CI. I do not know 

https://github.com/arekinath/PivApplet

Using this curve instead of one of the P-256 would be a good idea:

https://github.com/OpenSC/OpenSC/blob/master/.github/test-piv.sh

Separate PR is ok.

I will have to test this with the other ECC cards that support P-384, but I do not expect it will be an issue.

spurglegooge:
@Jakuje  Thanks — I looked into what that would take.

PivApplet does support it and no build change is needed: `PIV_SUPPORT_ECCP384` already
defaults to `true` in its `build.xml`, so the applet `.github/test-piv.sh` builds today
can already do P-384. `yubico-piv-tool` takes `-A ECCP384` (`YKPIV_ALGO_ECCP384 0x14`,
fixed to secp384r1). So the script change itself is one word, switching slot 9d from
`ECCP256` to `ECCP384` and leaving 9c on P-256 for coverage of both.

The part that is not one word is `piv_ref.json`. The script ends with

```
diff -u3 src/tests/p11test/piv{_ref,}.json
```

and changing a key's curve changes `p11test`'s output across `readonly_tests`,
`usage_test` and `derive_tests`, so the reference has to be regenerated or CI fails on the
diff rather than on the thing being tested.

I do not have that harness running locally, so I would rather ask than guess: is
regenerating `piv_ref.json` something you would want in the same PR, and is there a
preferred way to produce it? If you would rather fold it into a broader "exotic
configurations" change you already have in mind, that is fine too — happy either way.

Also worth saying explicitly: with 9c and 9d both on `ECCP256` today, a 32-byte digest
matches the field size exactly, so the failure in #3813 cannot occur in CI as configured.
Signing with `ECDSA-SHA256` against a P-384 key is what triggers it.

Jakuje:
I am ok to do it in separate PR.

The ECDSA-SHA256 mechanism is tested in the existing p11test so I assume it should reproduce this issue (if the key is not in the key management slot):

https://github.com/OpenSC/OpenSC/blob/master/src/tests/p11test/piv_ref.json#L292

Regarding the updating of the reference file, when the CI fails, it allows you to download artifacts, which should have the full diff of the file. Most of the time, this is useful to regression-test the driver and opensc, but I agree that updating this format is a bit annoying. Unfortunately, I did not figure out easier way, but if you have ideas or suggestions from outside, I would be happy to accept them or look into them.

spurglegooge:
Good catch, and it corrects what I proposed. I suggested switching 9d, but 9d is the KEY
MAN slot — `pkcs15-piv.c` maps id `03` to `0x9D`, and the signing test you linked is id
`02`, which is the SIGN key at `0x9C`. A P-384 key in 9d would sit in the derive/ECDH
tests and never reach the ECDSA path, which is the case your parenthetical is warning
about.

So the change is 9c rather than 9d:

```diff
-yubico-piv-tool ... -s 9c -a generate -A ECCP256 | tee 9c.pub
+yubico-piv-tool ... -s 9c -a generate -A ECCP384 | tee 9c.pub
```

That puts P-384 under the existing `["02", "ECDSA_SHA256", ...]` case, which is exactly
the 32-byte-digest-to-a-48-byte-field combination from #3813, and leaves 9d on P-256 so
both curves stay covered.

Worth flagging on ordering: with that change the `ECDSA_SHA256` case for key 02 should
*fail* on current master and pass with #3817 applied — that is the point of it, but it
means the CI PR wants to land after #3817 rather than before, or it looks like the test
change broke something.

On the reference file: understood, and the artifact download is a reasonable answer. One
small idea rather than a redesign — have the job print the unified diff into the log on
failure, in addition to uploading it. `diff -u3` already produces it; sending it to stdout
as well would mean the common case ("a mechanism moved from NO to YES, that is expected")
is answerable from the job page without downloading and unzipping anything. It does not
help with the actually annoying case of regenerating a large reference, but it makes the
small ones cheap.

I will open the 9c change as a separate PR once #3817 is in.

Jakuje:
> One small idea rather than a redesign — have the job print the unified diff into the log on failure, in addition to uploading it. 

I think its there, but the problem is that when we print all the testsuite logs, it tends to become a lot of data making the github log scrollback UI unusable (thus I printed just tails, I think). Suggestion to print just the failed logs is a good one. We can probably pull the status of the logs from the `.trs` files. This is the one that does the printing (including grouping):

https://github.com/OpenSC/OpenSC/blob/master/.github/dump-logs.sh

> I will open the 9c change as a separate PR once https://github.com/OpenSC/OpenSC/pull/3817 is in.

I would propose to open a PR before this is in so we can easily see it fails and rebase after we merge this to verify it correctly fixes the issue.

spurglegooge:
Before I open that PR — I checked whether it would go red for the right reason, and the
curve change will not. Something smaller will — SHA-224, not curve size — so I am posting
that instead.

`PivApplet` picks the ECDSA signer from the **input length**, not from the key's field
size (`PivApplet.java:2110-2125`, `processGenAuthEcPlain`):

```java
switch (sLen) {
case 20: si = ecdsaSha;    break;
case 32: si = ecdsaSha256; break;
case 48: si = ecdsaSha384; break;
default: ISOException.throwIt(ISO7816.SW_WRONG_DATA);
}
```

There is no check that the input matches the field length — which is the check the
OpenFIPS201 applet in #3813 makes. What PivApplet rejects is a length it does not
*recognise*, and that distinction turns out to matter; see the SHA-224 section below.

For the ECCP384 idea specifically, both legs land on a signer: on master a 32-byte digest
to a P-384 key takes `case 32` (`ecdsaSha256`); with this PR it is padded to 48 and takes
`case 48` (`ecdsaSha384`). Both sign, so `ECDSA_SHA256` on key 02 passes either way — no
discriminating signal, just a different signer object on each side.

That also corrects something I said earlier in this thread. I claimed that on P-256 "a
short digest never reaches the card". It does: `piv_ref.json` records
`["02","ECDSA_SHA1","YES"]`, a 20-byte digest against today's 32-byte-field P-256 key,
passing.

## The padding path is already exercised and green here

`CKM_ECDSA_SHA1` is registered as a sign-and-hash mechanism (`framework-pkcs15.c:6221`),
so OpenSC hashes with SHA-1 in software and sends a 20-byte digest down the raw EC path.
With this patch that 20 becomes 32, which moves PivApplet from `case 20` to `case 32` — a
different `Signature` object — and the result still has to verify against the SHA-1 digest
p11test computed. `test-piv-22.04-piv-sm`, `test-piv-24.04-piv-sm` and
`test-piv-24.04-valgrind` all pass on the current head.

So "leading zeros do not change the integer" is not only an argument here; the 20→32 case
is covered by the existing green run, against a real applet, with OpenSSL verifying the
output. @Jakuje — that may partly answer your "I will have to test this with the other ECC
cards".

One case deserves a sharper caveat than I gave it: **P-521**. `BYTES4BITS(521)` is 66, so
a SHA-512 digest goes 64 → 66 bytes. Every other combination gains leading zeros within a
length the card already expects; that one crosses to a length nothing has sent before. A
card that hardcodes 64 for P-521 would newly fail. SP 800-78 does not permit P-521 for
PIV, so it is not a PIV concern, and I have no card to test it on.

## The discriminating case is SHA-224, not curve size

SHA-224 is 28 bytes, which falls outside the applet's `{20, 32, 48}` switch and lands on
`default:` → `SW_WRONG_DATA`. On the P-256 9c key already in `test-piv.sh`:

- master: 28 bytes on the wire, sign fails
- with this PR: padded to 32, `case 32`, passes

CI misses it only because `p11test` never asks for the mechanism.
`p11test_case_mechs.c:119-127` enumerates `CKM_ECDSA`, `_SHA1`, `_SHA256`, `_SHA384`,
`_SHA512` and the four SHA3 variants but omits `CKM_ECDSA_SHA224`, even though OpenSC
registers it at `framework-pkcs15.c:6230` and it appears in `piv_ref.json`'s mechanism
list with `CKF_SIGN,CKF_VERIFY`. `p11test_case_readonly.c:506` has no `EVP_sha224()` arm
either, so adding the mechanism alone would give "Skip verify of unknown mechanism".

```c
/* p11test_case_mechs.c - EC mechanism list */
mechanism_list[i] == CKM_ECDSA_SHA224 ||

/* p11test_case_readonly.c - verify switch */
case CKM_ECDSA_SHA224:
	md = EVP_sha224();
	break;
```

Red on master, green here — the ordering you asked for, with no `test-piv.sh` change and
no card.

**Caveat worth flagging up front:** that mechanism list is shared across every token
p11test runs, not just PIV. Five references advertise `ECDSA_SHA224` — piv, oseid,
pico-hsm, sc-hsm and epass2003. Three of those have CI jobs here (`test-piv-*`,
`test-oseid-*`, `test-pico-hsm-*`), so three to regenerate in the PR, and OsEID or
pico-hsm may show the same failure on master for the same reason. `sc-hsm_ref.json` and
`epass2003_ref.json` have no CI job, so those would go stale until someone with that
hardware refreshes them. More signal than "two lines", and more churn.

## On the curve change

Switching 9c to `ECCP384` still looks worth doing on its own merits, and there is a better
argument for it than the one I gave: `ecdsaSha384` is instantiated only under
`//#if PIV_SUPPORT_ECCP384` (`PivApplet.java:360-367`) and gates
`pushAlgorithm(PIV_ALG_ECCP384)` at `:939`. Since no P-384 key exists in CI today, nothing
exercises that applet branch at all. The change would be first coverage of it, rather than
only P-384 coverage in OpenSC.

One correction to what I wrote earlier: 9e is `RSA2048` in CI, not EC, so moving the curve
there trades the card-auth RSA path (key `04`) rather than costing nothing. It overlaps
with 9a's RSA coverage, so probably still a good trade — but a trade.

Both of the above are derived from the applet and p11test sources rather than run: the
SHA-224 prediction and the P-384 signer-path prediction are equally unexecuted here. Happy
to open the SHA-224 change as a PR and let CI be the experiment, if it looks right to you.

dengert:
Have you look at:

https://learn.microsoft.com/en-us/answers/questions/1506551/certificate-based-smart-card-logon-to-windows-10-1

It says see (mmc: Local Computer Policy snap-in):
 `Computer Configuration > Administrative Templates > System > Kerberos > Configure hash algorithms for certificate logon`

On my Windows 11 pro,  the comments also say see: 

"This policy setting controls hash or checksum algorithms used by the Kerberos client when performing certificate authentication."

"More information about the hash and checksum algorithms supported by the Windows Kerberos client and their default states can be found at https://go.microsoft.com/fwlink/?linkid=2169037."

So to live withing the  PIV FIPS specs, I would suggest using SHA-384, which should work with P-265 (truncates) and P-384, where. Stay away from padding with zeros.

Jakuje:
Given the recent comments, I am no longer sure this is the right solution, especially for PIV itself.

The PIV specification says in "A.4.2. ECDSA" the Data field consist of the following data:

> '7C' – L1 { '82' '00' '81' L2 {hash value of message}}

https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73pt2-5.pdf

By zero-padding it to field length does not match the PIV specification.

That said, if the OpenFIPS does not work in this use case, its probably a bug in there. Unfortunately, we were not able to run that applet in emulator in CI when we looked into this couple of years bac.

The PIV applet does not really have a way to distinguish the hash type used from anything else than the input length (which is fragile on its own, but something that would need to be handled on the PIV specification level).

Nevertheless, improving our test coverage by  inlcuding the P-384 curve in the tests would be appreciated.

Jakuje:
For the SHA224, I think I did not include it in the p11tests back then due to being very niche thing not adding much value.

spurglegooge:
@dengert 
Thanks for the links, the second one doesn't seem to work for me:
but searching for those terms came up with things like : https://admx.offsec.nl/policy/windows/system/kerberos/configure-hash-algorithms-for-certificate-logon/

I will run some checks on our end today to confirm those settings

dengert:
OpenSC  card drivers tell OpenSC what the card supports for keys types and sizes. OpenSC then adds software support for
any hashes OpenSC support for the keys the card supports. PIV cards do not support doing the hash on the card but PIV specs says P256 must use SHA256 and P384 must use SHA384.  

  https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-78-5.pdf  "Table 2. Signature algorithm and key size requirements for PIV information" 

only:
 ECDSA (Curve P-256)  SHA-256 
 ECDSA (Curve P-384)  SHA-384 


So lets quit trying to use hash functions that are not recommended and stick with  P-256 uses SHA-246 and P-384 uses SHA-384. 

and also shows in 2031, no RSA 2028.   

Yet OpenSC is stuck supporting old cards and non standard PIV like cards.   
OpenSC has dropped old drivers in the past.  We also need a way to deprecate the use of non recommend keys and hashes.

metsma:
Least with our testin openssl validates all combinations (p-256/384 with sha256/384/512)

spurglegooge:
@Jakuje — fair challenge, so I measured the applet. Two corrections to things I have said in this thread
come out of it. The results are interesting. 

## What the card actually does

Raw ECDSA signs against the P-384 key in 9A, over PKCS#11 on Linux/macOS — no Windows
anywhere in the path. Input is a counting sequence, not a digest, so a success cannot be
explained by the card recognising a real hash:

| input bytes | result |
|---|---|
| 20, 28, 31, 32, 33, 47 | `CKR_DATA_INVALID` |
| 48, 49, 64, 66 | signs, 96-byte R‖S every time |

It is a **minimum length, not an equality**. Anything at or above the field length is
accepted; anything below is refused.

That leaves the question of what it *does* with the extra bytes when the input is longer
than the field. A 96-byte signature comes back either way, so the output size alone does
not tell you — it is consistent with truncating, with hashing, or with ignoring the tail.
So rather than infer it, I checked which bytes actually went into the signature.

Take a 64-byte input `B[0..63]` and have the card sign the whole thing. Then take two
different 48-byte slices of that same input:

```
B:        [ 0 .................................................. 63 ]   64 bytes, signed by the card
left48:   [ 0 ........................ 47 ]                             the first 48 bytes
right48:               [ 16 ......................................63 ]  the last 48 bytes
```

and ask OpenSSL to verify the card's signature against each slice **as if that slice were
the digest**, using the public key from the 9A certificate. ECDSA verification recomputes
from the digest it is given and only succeeds if the signer used that exact integer `e`,
so at most one of these can pass, and a pass identifies precisely which bytes the card
used:

```
left48     VERIFIES
right48    no
```

`right48` is the control. It is the plausible alternative — some implementations take the
least-significant end — and it also guards against the test being vacuous: if verification
passed against both slices, the check would be proving nothing and I could not trust the
`left48` result either. It failed, and `left48` passed, so the card signed the **first 48
bytes** of the input and discarded the remaining 16.

That is `e = leftmost min(N, outlen) bits` — FIPS 186-4 §6.4 / SEC1 — for a digest that is
too long. Combined with the sweep: the card truncates an over-long digest and refuses a
short one.

## Why I think that matters here

The card already performs half of this adaptation itself. #3817 supplies the other half,
and the two halves are not equally invasive:

- **truncation is lossy** — it discards digest bits. The card does this today, and
  `pkcs15-sec.c` has done it for over-long input for years, uncontroversially.
- **left-padding with zeros is not lossy at all** — it is the octet-string-to-integer
  conversion. A 32-byte string and a 48-byte string with 16 leading zeros are *the same
  integer*, so the signature is bit-identical to signing the short digest directly.

The objection on the table is to the lossless operation while the lossy one is already
accepted on both sides of the interface. That asymmetry is what the patch removes.

## On the A.4.2 reading

> The PIV specification says in "A.4.2. ECDSA" the Data field consist of the following
> data: '7C' – L1 { '82' '00' '81' L2 {hash value of message}}
> By zero-padding it to field length does not match the PIV specification.

I take the point, and it is the correct reading of the text. But I do not think it can be
the rule the card is enforcing, because a **66-byte input is not "the hash value of the
message" for a P-384 key either** — SP 800-78 pairs P-384 with SHA-384, 48 bytes — and the
card signs it without complaint. If §A.4.2 were operative as written, that would have to
fail too.

The same reading would also rule out `SHA-384 → P-256`, which was suggested earlier in the
thread as the alternative to this patch. That works only by truncating 48 bytes to 32, so
the `81` field there does not contain "the hash value of the message" either.

## Two corrections to what I wrote earlier

**1. The `-1205` is not evidence of a length check.** I have been citing
`SC_ERROR_INCORRECT_PARAMETERS` as though the applet diagnosed a bad length. It does not.
In OpenFIPS201 v2.0, the signing path (`generalAuthenticateCase1B`, "Variant B — Digital
Signatures") declares `PRE-CONDITIONS: NONE` — there is no length check there at all,
unlike Variants C and D which do test `challengeLength != key.getBlockLength()`. It wraps
the call and converts **any** exception:

```java
try {
    length = key.sign(buffer, challengeOffset, challengeLength, buffer, offset);
} catch (Exception e) {
    authenticateReset();
    // Presume that we have a problem with the input data, instead of throwing 6F00.
    ISOException.throwIt(ISO7816.SW_WRONG_DATA);
```

So `6A80` is the applet *presuming*, and the constraint lives below it. The measured
behaviour above is what that layer actually enforces.

**2. I should not have implied the applet enforces the SP 800-78 pairing.** It does not —
it accepts 64 and 66 bytes. It applies a minimum-length rule, which is a different thing
from conformance to the algorithm pairing.

## On "probably a bug in OpenFIPS201"

Worth flagging neutrally: the applet's own comment block states *"In all cases, the
challenge length must be equal to the key/block length"* and lists *"ECC384 - Challenge
block is 48 bytes"*, but the observed rule is `>= 48`. Its documented contract and its
behaviour differ. That is a real inconsistency and I am happy to raise it with that project 
— it is orthogonal to whether OpenSC pads.

Scope: this is OpenFIPS201 v2.0 on one card. PivApplet is different again — it dispatches
on input length `{20, 32, 48}` and rejects anything else, so it accepts 32 to a P-384 key
and refuses 28. Two real applets, two different rules, neither of them "exactly the field
length".

## The other two points

**SHA-224** — understood, dropping it. It was only ever attractive as a way to make CI go
red for the right reason, and if the mechanism is not worth carrying then that argument
does not apply.

**P-384 in CI** — still worth doing and I will open it separately as you suggested, before
this merges, so it can be seen to fail and then rebased. To be explicit about what it will
and will not show: on PivApplet both legs land on a signer (32 bytes takes `case 32`,
padded 48 takes `case 48`), so `ECDSA_SHA256` on key 02 passes either way. The value is
coverage of the `PIV_SUPPORT_ECCP384` applet branch, which nothing exercises today — not a
regression test for this bug. The card that reproduces #3813 is not one CI can run.

dengert:
Here is where OpenSC  truncates the hash if the hash is larger then the field length based on PKCS11 documentation.
If the card can do hashing, it passed the full inlen.

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-sec.c#L757-L766

Here are three examples using a IDEMIA ID-One PIV 2.4.1 approved for gov use.  The is from demo card #3 with 
9A P-256 and 9C  P-384. 
```
ECDSA-SHA256 02  i.e. P-384 9C key  


8907 Outgoing APDU (44 bytes):
8908 00 87 14 9C 26 7C 24 82 00 81 20 D9 01 4C 46 24 ....&|$... ..LF$
8909 84 4A A5 BA C3 14 77 3D 6B 68 9A D4 67 FA 4E 1D .J....w=kh..g.N.
8910 1A 50 A1 B8 A9 9D 5A 95 F7 2F F5 00             .P....Z../..
8911 
8912 P:262114; T:0x140295475540992 14:04:56.788 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
8913 P:262114; T:0x140295475540992 14:04:56.937 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
8914 Incoming APDU (110 bytes):
8915 7C 6A 82 68 30 66 02 31 00 AD E6 62 1C 72 37 CD |j.h0f.1...b.r7.
8916 01 A5 3E 0A 06 F4 2A D3 FC 42 35 56 0C 9B 3D 3A ..>...*..B5V..=:
8917 5E 2B A0 A8 B1 06 01 BE 65 E8 CD 3B BF FC E7 A9 ^+......e..;....
8918 51 03 5F 4C 8B 04 CD 12 00 02 31 00 D1 DD 72 D5 Q._L......1...r.
8919 05 FD B6 71 03 72 D9 9D A8 8A 48 9B E4 1F C9 A8 ...q.r....H.....
8920 74 E9 DC F2 30 98 E2 E5 6F 93 B7 BF 1D CD 12 DC t...0...o.......
8921 A7 D3 FE 95 22 72 DA A8 7A A8 CC 65 90 00       ...."r..z..e..

ECDSA-SHA384 0  i.e.P-384 9C key


907 Outgoing APDU (60 bytes):
8908 00 87 14 9C 36 7C 34 82 00 81 30 79 A7 AE C7 08 ....6|4...0y....
8909 47 C2 42 10 2B 88 9B 29 8A 07 20 80 3B 34 0B 35 G.B.+..).. .;4.5
8910 0C D9 E8 9F 57 4C 68 4D 46 BF B2 32 B9 2D 2D F3 ....WLhMF..2.--.
8911 56 FD 77 E4 D2 04 7C 43 B3 F8 A0 00             V.w...|C....
8912 
8913 P:262264; T:0x140506063152128 14:14:10.318 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
8914 P:262264; T:0x140506063152128 14:14:10.468 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
8915 Incoming APDU (109 bytes):
8916 7C 69 82 67 30 65 02 31 00 F1 6C D0 FC 7A 41 B7 |i.g0e.1..l..zA.
8917 23 93 2C 78 7D BA 4F A1 46 AF 2A EB 06 AB BB 3A #.,x}.O.F.*....:
8918 BA 41 7C A4 B1 01 D1 AB 67 34 6D E0 47 1E 38 75 .A|.....g4m.G.8u
8919 87 03 D3 45 56 47 41 97 3C 02 30 3D C7 59 6D 03 ...EVGA.<.0=.Ym.
8920 FD 05 59 94 37 07 23 13 30 D3 14 1C 39 EB 8E A3 ..Y.7.#.0...9...
8921 3B 6D EF 5F E4 2E 4E 3D 51 DC 49 73 CD 34 1F BB ;m._..N=Q.Is.4..
8922 01 F8 55 CB 4C B4 64 14 5B 94 F1 90 00          ..U.L.d.[....


ECDSA-SHA384 01  i.e.P-256 9A key Software truncated data sent to field size

8761 Outgoing APDU (44 bytes):
8762 00 87 11 9A 26 7C 24 82 00 81 20 79 A7 AE C7 08 ....&|$... y....
8763 47 C2 42 10 2B 88 9B 29 8A 07 20 80 3B 34 0B 35 G.B.+..).. .;4.5
8764 0C D9 E8 9F 57 4C 68 4D 46 BF B2 00             ....WLhMF...
8765 
8766 P:262377; T:0x140698418131968 14:27:35.458 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
8767 P:262377; T:0x140698418131968 14:27:35.572 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
8768 Incoming APDU (76 bytes):
8769 7C 48 82 46 30 44 02 20 47 09 F8 AB 72 DE 67 27 |H.F0D. G...r.g'
8770 BB DA F4 3A 71 87 B9 2A 69 E2 90 FB 04 68 42 44 ...:q..*i....hBD
8771 B8 D2 CE E1 8F 4E EF 53 02 20 08 D1 C4 BF 90 72 .....N.S. .....r
8772 6E B6 BC C5 A0 9F E7 DE B6 A6 46 24 09 6F 0B 31 n.........F$.o.1
8773 A0 AE 6B 01 6E 38 8F DE D9 F2 90 00             ..k.n8......
```

Both OpenSSL and pkcs11-tool verify the signatures.

dengert:
If the OpenSC code https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-sec.c#L757-L766
is removed so the 48 byte  hash is sent to card to be used with a P-256 key
The IDEMIA ID-One PIV 2.4.1 card fails with 6A80:

```
ECDSA-SHA384 01  i.e. P-256 9A where OpenSC sends full 384 hash to a be used with  p-256 key

474 Outgoing APDU (60 bytes):
8475 00 87 11 9A 36 7C 34 82 00 81 30 79 A7 AE C7 08 ....6|4...0y....
8476 47 C2 42 10 2B 88 9B 29 8A 07 20 80 3B 34 0B 35 G.B.+..).. .;4.5
8477 0C D9 E8 9F 57 4C 68 4D 46 BF B2 32 B9 2D 2D F3 ....WLhMF..2.--.
8478 56 FD 77 E4 D2 04 7C 43 B3 F8 A0 00             V.w...|C....
8479 
8480 P:294863; T:0x140269223695360 17:19:56.483 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
8481 P:294863; T:0x140269223695360 17:19:56.532 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
8482 Incoming APDU (2 bytes):
8483 6A 80 j.
```
In all these tests, the says "Hello World!" data was used as input to be hashed 
and the truncation is needed in software and the software is sending the left 32 bytes.

spurglegooge:
@dengert — thanks for both traces. I tested the Kerberos policy you suggested as well;
both results below.

### The IDEMIA traces

The second trace is the one that changed my view. With the truncation at L757-L766
removed, the card returns `6A80` for a 48-byte digest against a P-256 key — so that branch
is doing necessary work today, and this patch leaves it untouched.

Set against the card in #3813, the two fail in opposite directions:

| | short (32 → P-384) | long (48 → P-256) |
|---|---|---|
| IDEMIA ID-One PIV 2.4.1 | accepts | **refuses** `6A80` |
| OpenFIPS201 v2.0 (NXP JCOP J3R310) | **refuses** `6A80` | accepts, truncates to leftmost |

Neither card handles both directions, which is what makes `pkcs15-sec.c` the natural place
for this rather than the drivers. The two adaptations are not equally invasive, though:
truncation discards digest bits, whereas left-padding with zeros is the
octet-string-to-integer conversion — a 32-byte string and a 48-byte string with 16 leading
zeros are the same integer, so the signature produced is bit-identical to signing the short
digest directly.

I verified the OpenFIPS201 truncation rather than inferring it from the 96-byte output:
sign an over-long input, then verify the signature against its leftmost 48 bytes with the
public key, and against the rightmost 48 as a control. Leftmost verifies and rightmost does
not, at 49, 64 and 66 bytes.

This does support @Jakuje's reading that the OpenFIPS201 behavior is a bug in that applet,
posted it there as makinako/OpenFIPS201#89 — with the measurements, since
"refuses a short digest" is not the whole story: the same asymmetry is present, just
mirrored. The cards are already issued, though, and OpenSC is the layer that has to talk to
both.

### The Kerberos hash policy

I set it up and measured it rather than reason about it. Windows 11 `10.0.26200.9445`,
OpenSC minidriver, P-384 key in 9A, Samba AD DC. I did not hand-type the registry values —
the test script parses `Kerberos.admx` on the machine and derives them, so a typo could not
masquerade as "no effect":

```
Software\Microsoft\Windows\CurrentVersion\Policies\System\Kerberos\Parameters
  PKInitHashAlgorithmConfigurationEnabled = 1
  PKInitSHA384 = 3   (Supported)
  PKInitSHA256 = 0   (Not Supported)
states from the ADMX enum: Not Supported=0, Default=1, Audited=2, Supported=3
```

Values read back with `reg query` **after** each logon, so they survived the reboot. The
number below is the digest length the minidriver logs on entry to `CardSignData` (`cbData`,
printed by OpenSC as `pInfo->pbData(n)`) — what Windows emits, before OpenSC touches it.

| run | policy | interactive logon | background ops (silent, cached PIN) |
|---|---|---|---|
| offline (cached creds) | `SHA256=Not Supported` | **32** | 48 |
| online (PKINIT confirmed KDC-side) | `SHA256=Not Supported` | **32** | 48, 48 |
| online (control) | **policy removed** | **32** | 48, 48 |

Identical with the policy on and off. Two details matter more than the table:

- **PKINIT succeeded *using* SHA-256 while the policy declared SHA-256 Not Supported.** The
  KDC logged `PKINIT pre-authentication succeeded`, and the client had just signed 32
  bytes. If the policy were honoured here, Windows should have moved to SHA-384 or refused.
- **The 32/48 split tracks the PIN context, not the policy.** Interactive logons
  (`CardAuthenticateEx dwFlags=0x00000000`) sent 32; silent cached-PIN operations
  (`0x00000040`) sent 48, in every run. So something on the same machine signs 48 against
  the same key — 32 is not a card or driver limit, it is what Windows chooses for the logon
  signature.

Every `CardSignData` in all three runs also carried `aiHashAlg=0x00000000` with
`CARD_PADDING_INFO_PRESENT` not set: Windows declares no hash algorithm for this operation
at all.

**What I cannot claim.** One Windows build, one card, a Samba/Heimdal KDC. Online I cannot
exclude that our KDC's advertised algorithms constrained the choice rather than the policy
being inert — Heimdal does not log the CMS digest at the level I had enabled. That caveat
does not reach the offline case, which performs no Kerberos exchange at all; #3813 is the
offline path, and it produced 32 bytes with the policy set as strictly as it can be. And I
could not identify the component behind the 48-byte signatures, only that it is unaffected
by the policy.

If the configuration is wrong I would rather be corrected than be right — the script
derives everything from the ADMX and the logs are reproducible. If there is a second
registry location, or a KDC-side setting that has to accompany it, I will test that too.

## Conclusion

Your traces settle the general case: cards like the IDEMIA already accept a short digest,
so this patch is only for the minority of applets that refuse one; the existing truncation
stays as it is. For the cards that fail the other way, the user sees `Incorrect parameters
in APDU` at offline Windows logon with no recourse, and the Kerberos policy does not change
what Windows sends. Whether OpenSC adapts to that combination or refuses it is the
project's call. Let me know how I can help — the `sc_log()` on the padding path is still on
offer if the patch goes in, and I can re-run any of the above if it doesn't.

dengert:
**My Conclusion**

NIST says PIV shall use SHA-256 with P-256 keys and SHA-384 with P-384 keys. Both PIV apps support this. 
There is no truncation or padding need for these.

OpenSC advertises it can do any hash with any key. Trying to use mismatched hash and keys just leads to complications for other software that has to figure out how to handle the mismatches.

So why are we discussing this? We should be discussing how to avoid these mismatches. 

The two PIV applets each fail for a different mismatch. There are at least a dozen versions of PIV applets, from approved by NIST to opensource 

Other applets may have the same problems with mismatches and there are other EC  and non-EC curves. 

At least for PIV driver, it could be modified to enforce the NIST rules (with an `opensc.conf` flag to override) for backward compatability.

pkcs11-tool could also warn if a mismatch  and man page and help too.

dengert:
You could  also ask about HASH and KEY mismatches in a Kerberos, Samba, Heimdal or pkinit  mailing list. 

In the 80's and 90's and early 2000's  before retiring I was active in the IETF Kerberos WG.  We where using Kerberos with AFS and eventually were using Microsoft AD servers as the KDC.  Then in the early 2000's we where issued gov PIV cards using only RSA. But we had a lot of non windows system, so I wrote the OpenSC PIV driver. 

But these days I have no real interest in Kerberos.

spurglegooge:
@dengert — I agree with the goal, and the warning and documentation parts look worth doing
regardless of what happens to this PR. A mismatch that works silently is worse than one
that says so. If it helps I will put up the `pkcs11-tool` warning and the man page wording
as a separate PR, independent of this one.

What this call actually is — I had it wrong, but not in the direction I expected

I have been calling the 32 bytes a digest. I went looking for what the operation is, and I
can no longer justify that

Microsoft's [description of cached smart card logon](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/ff404285(v=ws.10))
says that for RSA the cached-credential key is "derived from a random value signed with the
RSA logon key", and that for ECDSA an associated **ECDH** key is used instead, because a key
derived from an ECDSA signature "would likely be different each time it gets computed".
I should be careful with that: it describes a *different* mechanism from the one I am
watching — our key is ECDSA and this is a signature, not a key agreement — and the page is
2008 R2-era. So I am not claiming it documents this call.

What I can say is what we measured here:

1. **It is provably not PKINIT.** With the stock minidriver and a working network, the
   32-byte call fails and the logon still succeeds:

   ```
   32 bytes → sc_pkcs15_compute_signature return -1205
              Incorrect parameters in APDU
              CardSignData returning with: 0x80100004      FAILED
   48 bytes → sc_pkcs15_compute_signature return 96        OK
   ```

   and the KDC logged `PKINIT pre-authentication succeeded` for that same logon. Kerberos
   completed while the 32-byte call was failing, so it cannot be part of the authpack
   signature.
2. **Its consequences track cached-credential logon.** The same failure is survivable
   online and fatal offline, which is why #3813 only shows up offline. Same failure,
   no fallback.
3. **Windows names no hash for it.** Every one of these carries `aiHashAlg=0x00000000` with
   `CARD_PADDING_INFO_PRESENT` not set.
4. **The values differ every time.** Three consecutive logons, three different 32-byte
   values.

That is all *consistent with* a freshness value being signed rather than a message digest.
I would not put it stronger than that. But if it is even approximately right, the SP 800-78
pairing has nothing to attach to here: the pairing constrains the hash used to sign a
message a relying party will verify, and there is no message in this operation.

The padding is verifiably a no-op on the signature

I had been asserting "leading zeros do not change the integer" from the maths. It is
demonstrable on the card.

`X` is a real 32-byte value captured from a failing logon on this machine — a spent
one-shot freshness value, shown without its signature. `P` is what this patch sends instead:

```
X = 4E750E0D19A6A2AF00EB995154A76A2981E0379D820E0F4EE62F3D19AE1AF6EF        (32 bytes)
P = 00000000000000000000000000000000 || X                                   (48 bytes)
```

Sign `P` with the card, then verify the resulting signature against **`X`**, the original 32
bytes, as the digest:

```
verify against P (what was signed)      VERIFIES
verify against X (what Windows asked)   VERIFIES
```

Both, because `bits2int` yields the same `e` either way (FIPS 186-5). So the signature this
patch causes the card to produce **is** a signature over exactly the value Windows supplied
— not an equivalent one, the same one.

- **It cannot change what any signature is a signature of.** ECDSA draws a fresh `k` each
  time, so the bytes differ on every call regardless; what padding leaves untouched is the
  value being signed. The risk to existing deployments is zero rather than small.
- **Strict and permissive applets converge.** An IDEMIA card given 32 bytes directly and an
  OpenFIPS201 card given the padded 48 both produce signatures that verify against the
  original 32 — interoperable, not a third behaviour.

It also retires the §A.4.2 concern on its own terms, I think: the objection is that the `81`
field should carry the hash value of the message, and padding puts something else there.
What the card returns is nonetheless the signature over the unpadded value — there is no
observable difference for the objection to attach to.

Two snags with enforcing the pairing in the PIV driver

**The driver cannot see the hash.** For `CKM_ECDSA_SHA256` and friends, OpenSC hashes in the
PKCS#11 layer and passes a raw digest down, so by the time `card-piv.c` sees it the only
signal left is the input length. Enforcing the pairing there means inferring the hash from
the length — the inference @Jakuje called fragile earlier in this thread, and the thing the
two applets already do differently from each other. A rule would have to live in the
framework layer where the mechanism is still known; and for `CKM_ECDSA` there is no
mechanism-level hash to check even there.

**It would turn a currently-green test red.** `src/tests/p11test/piv_ref.json` records the
entry `"02"` / `"ECDSA_SHA1"` / `"YES"` — a 20-byte digest against the P-256 key in 9C,
which is a mismatch under the pairing and passes in CI today. Strict enforcement fails that
case, so the flag would have to default permissive.

On the Windows side there is also nothing to configure away: I tested the one candidate I
could find, the Kerberos cert-logon hash policy, against a control with it removed, and it
changed nothing on either path. I would not claim no such setting exists — only that I
found none.

Where that leaves the two ideas

They act on different paths, so I do not think they compete:

| caller names the hash | example | your proposal |
|---|---|---|
| yes | `CKM_ECDSA_SHA384` via pkcs11-tool | warn, or refuse under a strict `opensc.conf` flag — reasonable, and I would use it |
| no | `CKM_ECDSA` raw, `aiHashAlg=0` from `lsass` | nothing to check against; this is the path #3813 is about |

If the project wants the strict mode I am happy to help build it — the `pkcs11-tool`
warning, the man page, the `opensc.conf` switch. I would just rather it land alongside the
padding than instead of it, because on the no-hash path a rule has nothing to act on.

spurglegooge:
> Here is where OpenSC  truncates the hash if the hash is larger then the field length based on PKCS11 documentation.
> 
> If the card can do hashing, it passed the full inlen.
> 
> 
> 
> https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/pkcs15-sec.c#L757-L766
> 
> 
> 
> Here are three examples using a IDEMIA ID-One PIV 2.4.1 approved for gov use.  The is from demo card #3 with 
> 
> 9A P-256 and 9C  P-384. 
> 
> ```
> 
> ECDSA-SHA256 02  i.e. P-384 9C key  
> 
> 
> 
> 
> 
> 8907 Outgoing APDU (44 bytes):
> 
> 8908 00 87 14 9C 26 7C 24 82 00 81 20 D9 01 4C 46 24 ....&|$... ..LF$
> 
> 8909 84 4A A5 BA C3 14 77 3D 6B 68 9A D4 67 FA 4E 1D .J....w=kh..g.N.
> 
> 8910 1A 50 A1 B8 A9 9D 5A 95 F7 2F F5 00             .P....Z../..
> 
> 8911 
> 
> 8912 P:262114; T:0x140295475540992 14:04:56.788 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
> 
> 8913 P:262114; T:0x140295475540992 14:04:56.937 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
> 
> 8914 Incoming APDU (110 bytes):
> 
> 8915 7C 6A 82 68 30 66 02 31 00 AD E6 62 1C 72 37 CD |j.h0f.1...b.r7.
> 
> 8916 01 A5 3E 0A 06 F4 2A D3 FC 42 35 56 0C 9B 3D 3A ..>...*..B5V..=:
> 
> 8917 5E 2B A0 A8 B1 06 01 BE 65 E8 CD 3B BF FC E7 A9 ^+......e..;....
> 
> 8918 51 03 5F 4C 8B 04 CD 12 00 02 31 00 D1 DD 72 D5 Q._L......1...r.
> 
> 8919 05 FD B6 71 03 72 D9 9D A8 8A 48 9B E4 1F C9 A8 ...q.r....H.....
> 
> 8920 74 E9 DC F2 30 98 E2 E5 6F 93 B7 BF 1D CD 12 DC t...0...o.......
> 
> 8921 A7 D3 FE 95 22 72 DA A8 7A A8 CC 65 90 00       ...."r..z..e..
> 
> 
> 
> ECDSA-SHA384 0  i.e.P-384 9C key
> 
> 
> 
> 
> 
> 907 Outgoing APDU (60 bytes):
> 
> 8908 00 87 14 9C 36 7C 34 82 00 81 30 79 A7 AE C7 08 ....6|4...0y....
> 
> 8909 47 C2 42 10 2B 88 9B 29 8A 07 20 80 3B 34 0B 35 G.B.+..).. .;4.5
> 
> 8910 0C D9 E8 9F 57 4C 68 4D 46 BF B2 32 B9 2D 2D F3 ....WLhMF..2.--.
> 
> 8911 56 FD 77 E4 D2 04 7C 43 B3 F8 A0 00             V.w...|C....
> 
> 8912 
> 
> 8913 P:262264; T:0x140506063152128 14:14:10.318 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
> 
> 8914 P:262264; T:0x140506063152128 14:14:10.468 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
> 
> 8915 Incoming APDU (109 bytes):
> 
> 8916 7C 69 82 67 30 65 02 31 00 F1 6C D0 FC 7A 41 B7 |i.g0e.1..l..zA.
> 
> 8917 23 93 2C 78 7D BA 4F A1 46 AF 2A EB 06 AB BB 3A #.,x}.O.F.*....:
> 
> 8918 BA 41 7C A4 B1 01 D1 AB 67 34 6D E0 47 1E 38 75 .A|.....g4m.G.8u
> 
> 8919 87 03 D3 45 56 47 41 97 3C 02 30 3D C7 59 6D 03 ...EVGA.<.0=.Ym.
> 
> 8920 FD 05 59 94 37 07 23 13 30 D3 14 1C 39 EB 8E A3 ..Y.7.#.0...9...
> 
> 8921 3B 6D EF 5F E4 2E 4E 3D 51 DC 49 73 CD 34 1F BB ;m._..N=Q.Is.4..
> 
> 8922 01 F8 55 CB 4C B4 64 14 5B 94 F1 90 00          ..U.L.d.[....
> 
> 
> 
> 
> 
> ECDSA-SHA384 01  i.e.P-256 9A key Software truncated data sent to field size
> 
> 
> 
> 8761 Outgoing APDU (44 bytes):
> 
> 8762 00 87 11 9A 26 7C 24 82 00 81 20 79 A7 AE C7 08 ....&|$... y....
> 
> 8763 47 C2 42 10 2B 88 9B 29 8A 07 20 80 3B 34 0B 35 G.B.+..).. .;4.5
> 
> 8764 0C D9 E8 9F 57 4C 68 4D 46 BF B2 00             ....WLhMF...
> 
> 8765 
> 
> 8766 P:262377; T:0x140698418131968 14:27:35.458 [opensc-pkcs11] reader-pcsc.c:260:pcsc_internal_transmit: called
> 
> 8767 P:262377; T:0x140698418131968 14:27:35.572 [opensc-pkcs11] reader-pcsc.c:357:pcsc_transmit: 
> 
> 8768 Incoming APDU (76 bytes):
> 
> 8769 7C 48 82 46 30 44 02 20 47 09 F8 AB 72 DE 67 27 |H.F0D. G...r.g'
> 
> 8770 BB DA F4 3A 71 87 B9 2A 69 E2 90 FB 04 68 42 44 ...:q..*i....hBD
> 
> 8771 B8 D2 CE E1 8F 4E EF 53 02 20 08 D1 C4 BF 90 72 .....N.S. .....r
> 
> 8772 6E B6 BC C5 A0 9F E7 DE B6 A6 46 24 09 6F 0B 31 n.........F$.o.1
> 
> 8773 A0 AE 6B 01 6E 38 8F DE D9 F2 90 00             ..k.n8......
> 
> ```
> 
> 
> 
> Both OpenSSL and pkcs11-tool verify the signatures.
> 
> 
> 
> 
> 
> 
> 
> 

Anyone know how to actually get IDEMIA cards? They wont talk to use cuz we aren't buying 1million.

Jakuje:
> Anyone know how to actually get IDEMIA cards? They wont talk to use cuz we aren't buying 1million.


Is the issue reproducible with the NIST Test Cards? I have bunch and from what I remember, I was writing the truncation check based on them.

https://csrc.nist.gov/projects/piv/nist-piv-test-cards

dengert:
I have not asked Idemia for years. The old PIV demo cards based on NIST-800-73-3 should still work. But no Secure Messaging.  

But there is some progress on @spurglegooge  problems. 
See https://github.com/OpenSC/OpenSC/pull/3270#issuecomment-5659946445
and https://github.com/makinako/OpenFIPS201/issues/89


The problem appears to come from not providing the correct keyUsage and mismatch of EC keys on the same card. This causes Windows to not find P-384 key agreement key. If this is the case, there should be no need to have to pad short hashes.

I have forgotten about #3270

spurglegooge:
@dengert @Jakuje — thanks both, appreciate the feedback this is really helpful.

## "re : …there should be no need to have to pad short hashes"

> The problem appears to come from not providing the correct keyUsage and mismatch of EC keys
> on the same card. … **If this is the case, there should be no need to have to pad short
> hashes.**

You are right about the fault. Cached logon with an ECC credential needs an **ECDH
(key-agreement) key** on the card associated with the ECDSA logon key — Windows derives the
key protecting the cached credential from that agreement, because it cannot derive a stable
one from an ECDSA signature. Our card had 9A EC P-384 and 9C/9E EC P-256, all *signing* keys,
and 9D **RSA** — so there was no EC key-agreement key at all. Windows had nothing to
associate, hid the logon certificate, and the logon screen said "no valid certificates". And
`CardGetContainerProperty` implements no `CCP_ASSOCIATED_ECDH_KEY`, so nothing could tell it
otherwise. That is #3270, and your 2024 reading of it was correct.

But we fixed it three weeks ago, and it is a different failure:

| | |
|---|---|
| **2026-08-31** | EC P-384 key-agreement key provisioned into slot 82 |
| | → logon certificate appears; **online** smart-card logon works |
| | → **offline** logon still fails, `-1205` on a 32-byte input to the P-384 9A key |
| **today** | with this patch: **offline logon succeeds** |

The two are sequential, not the same thing. Once the association existed, Windows got as far
as asking the card to sign, and the card refused. The association fault stops you before any
signature; this stops you at one.

Measured today on Windows 11 Pro 10.0.26200 with the **current** patch (the one already on
this branch), cached credential flushed and rebuilt first so that stale cache state cannot
account for it — `CachedLogonsCount=0`, reboot, online card logon; back to `10`, reboot,
online card logon; then network down, reboot, card + PIN:

```
CardSignData dwVersion=2, bContainerIndex=0, dwKeySpec=4, aiHashAlg=0x00000000
CARD_PADDING_INFO_PRESENT not set
pInfo->pbData(32)
pInfo->cbSignedData = 96
sc_pkcs15_compute_signature return 96
MD_Function:CardSignData:5048 returning with: 0x00000000
```

`dwKeySpec=4` is `AT_ECDSA_P384`, so Windows has the curve right, declares no hash algorithm,
and sends 32 bytes. Without the patch that call returns `-1205` and offline logon fails; with
it the card signs and the logon completes.

And independently of Windows, raw APDUs to the card via `opensc-tool -s`, 9A holding P-384:

| input | SW | |
|---|---|---|
| 20 bytes | `6A80` | |
| 32 bytes | `6A80` | the failure |
| 33 bytes | `6A80` | not a SHA length |
| 48 bytes | `9000` | |
| 64 bytes | `9000` | longer than the 48-byte block, still signed |
| 65 bytes | `6A80` | not a SHA length |

The 65-byte row is there to prove the bytes reach the card unmodified — had anything been
truncating to the field length it would have become 48 and signed. This matters because the
claim I had to retract earlier in this thread came from exactly that confusion.

## On the NIST test cards

> Is the issue reproducible with the NIST Test Cards?

Probably not, and I would rather say so than have you spend time on it. The refusal is an
applet behaviour: OpenFIPS201 v2.0 refuses a SHA-256-length input to a P-384 key, deliberately
and with FIPS 186-5 6.1.1 quoted in the source above the check
(`Platform.Cryptography.sign()`, `release/v2-0-0-fips`). A card that does not enforce that
pairing will sign the 32 bytes and show nothing. @dengert's IDEMIA traces in this thread are
an example — that card signs a 32-byte input against a P-384 key quite happily.

So the question this PR turns on is not whether cards generally refuse short input. It is
whether OpenSC should left-pad a short EC input to the field length for cards that refuse it,
given that the padding does not change the value being signed. Today it passes the short input
through unchanged, which is fine when the card is lenient and fatal when it is not — and the
caller cannot tell which it got.

Whether the applet should refuse is a separate argument and I have it running with @dengert at
makinako/OpenFIPS201#89. This PR does not depend on winning it.

## Tightening the patch — @Jakuje

You were right to be unsure about the scope, and I have had to correct it: **the patch as
written is too broad.** The diff below is not on the branch yet — it changes what the patch
does rather than how it is written, so I would rather you saw the discriminator first. Say
the word and it goes on as a single commit on top of what is already there, no force push.

On the padding itself, which I should have justified better the first time. FIPS 186-5
Appendix B.2.4 reconciles a bit string against a modulus of a different length, and its
declared inputs are "the bit string to be converted" and "an integer modulus" — it does not
mention a hash:

> 1. Truncate or expand the input bit string to the desired length len(N)
>    a. If n < len(N) then add (len(N)-n) bits (of value zero) to the left of the sequence
>    b. If n >= len(N) then the len(N) leftmost bits are kept, and subsequent bits are
>       discarded

That is both halves of what this branch does, and 6.4.1 step 2(a) agrees for the short case
("If len(n) >= hashlen, set E = H"). Neither has a rejection branch. I am not claiming B.2.4
governs this path — it is reached from A.3.3, the deterministic-ECDSA nonce derivation — only
that where NIST specifies this conversion, a short value is left-padded with zeros rather than
refused, and that padding does not change the integer.

It is also what the RSA branch immediately above already does in the same function:

```c
if (inlen < modlen) {
    if (modlen > buflen) { r = SC_ERROR_BUFFER_TOO_SMALL; goto err; }
    memmove(tmp+modlen-inlen, tmp, inlen);
    memset(tmp, 0, modlen-inlen);
}
```

so this is EC catching up with existing behaviour rather than a new idea, and the added code
deliberately mirrors it.

One clause, so the padding applies only where a strength guard would not be legitimate:

```diff
-		inlen = MIN(inlen, BYTES4BITS(prkey->field_length));
+		size_t flen = BYTES4BITS(prkey->field_length);
+
+		if (inlen > flen) {
+			inlen = flen;
+		} else if (inlen < flen &&
+				!(prkey->usage & SC_PKCS15_PRKEY_USAGE_NONREPUDIATION)) {
+			if (flen > buflen) {
+				r = SC_ERROR_BUFFER_TOO_SMALL;
+				goto err;
+			}
+			memmove(tmp + flen - inlen, tmp, inlen);
+			memset(tmp, 0, flen - inlen);
+			inlen = flen;
+		}
```

Scope worth saying out loud: `pkcs15-sec.c` is common to every card driver, so this pads
short input on any EC key without `NONREPUDIATION` set, not only PIV ones. `pkcs15-piv.c` sets
`NONREPUDIATION` on 9C and not on 9A, so on PIV it separates the two cases, but on another
driver the flag may mean less or nothing. It is a proxy rather than the real signal — PKCS#15 has no "authentication key"
usage — and I am not attached to it. If there is a better discriminator, or you would
rather this were an `opensc.conf` option (which would also suit @dengert's enforcement
proposal), say which and I will rework it.

Built and tested against OpenFIPS201 v2.0 on NXP P71D321, P-384 in 9A (usage `0x04`) and
P-256 in 9C (usage `0x204`, i.e. with NONREPUDIATION), stock 0.27.1 side by side with the
branch:

```
                                  stock 0.27.1   this branch
  32B -> 9A P-384 (auth)          FAIL           SIGN 96B
  32B -> 9C P-256 (sig, =field)   SIGN 64B       SIGN 64B
  20B -> 9C P-256 (sig)           FAIL           FAIL
  20B -> 9A P-384 (auth)          FAIL           SIGN 96B
```

Row 1 is the case in #3813. Row 3 is the point of the narrowing: the previous version of
this patch padded 20 bytes to 32, which lands in the SHA-256 case on a P-256 key where no
guard fires, and the card signed it. It no longer does. Row 2 confirms nothing changes
where the input already matches the field length.

The padding itself remains signature-neutral: leading zeros do not change the integer, so
what the card signs is the value the caller supplied. I verified that on hardware — sign
the zero-padded 48 bytes, then verify the signature against the original 32 with the public
key, and both verify.

## The tightened build, also end to end on Windows

The log above is the current patch. I then swapped in a minidriver built from the **narrowed**
version and repeated the same sequence — flush, rebuild online, offline card + PIN — to check
that restricting the padding does not break the case it exists to fix. It does not: offline
logon succeeds again, with the same `pbData(32)` → `cbSignedData = 96` →
`CardSignData returning with: 0x00000000`.

So 9A does not acquire `NONREPUDIATION` through the minidriver path, and the discriminator
holds on hardware rather than only in `pkcs15-piv.c`.

Worth stating plainly because the original issue text did not: the earlier "fails before,
succeeds after" was measured without rebuilding the credential cache in between. Both runs
today were, so cache state is excluded by construction rather than by assumption.

I have also retracted the central claim of makinako/OpenFIPS201#89 — I reported the card as
truncating over-length input, when in fact OpenSC truncated it before the APDU was built,
exactly where @dengert pointed early in this thread.

spurglegooge:
Pushed as a5c740a4c — single commit on top, no force push. Happy to rework the discriminator.

spurglegooge:
@dengert — replying here to [the comment on #3270](https://github.com/OpenSC/OpenSC/pull/3270#issuecomment-5681425959),
since its subject references this PR. Thank you for rebasing #3270 and for the MSI
pointers — genuinely useful. Before replying I built the card described there and tested it,
because the provisioning concern was checkable and it was partly right. Results below; it
bears on both our claims.

## Quick correction, the padding is on the left

You mentioned "zero padding on the right", and the PR description never said which side —
my fault. The patch left-pads: `0x00…00 ‖ X`, the same integer, so the card signs exactly the
value the caller supplied. The RSA branch a few lines above in the same function already left-pads short input the
same way.

## The experiment: a canonical key-management layout

You mentioned the cards were provisioned incorrectly against the standards, you are right:
the EC agreement key was in retired slot 82 (SP 800-73-5 Pt 1 §3.3.3 defines 82–95 for
superseded keys), 9D held RSA, and 9D's certificate asserted both digitalSignature and
keyEncipherment — the ambiguity you opened #3270 about.

To test we provisioned a fresh card with the canonical key-management layout: 9A EC P-384; 9D EC
P-384, `keyAgreement` only; no RSA anywhere; slots 82–95 never created; Key History zeroed.
Same applet (OpenFIPS201 v2.0), same chip family, same CA, same DC, same machine
(Win 11 Pro 10.0.26200), and — this matters — **stock OpenSC 0.27.1 from the official MSI**,
no patch installed. The original card stayed as the control.

**Online logons (two, the second after a clean sign-out):** the ECC cached-credential
machinery works correctly on the canonical 9D. Both logons show the full derivation
fingerprint against the key-exchange container — container 2, the 9D key:

```
CardGetContainerInfo bContainerIndex=2 ... cbKeyExPublicKey=104   (BCRYPT_ECDH_PUBLIC_P384_MAGIC)
CardConstructDHAgreement dwKeySpec=2, cbData=104
sc_pkcs15_derive returned 48
CardDeriveKey  -> 0x00000000   (x2)
CardDestroyDHAgreement -> 0x00000000
```

So Windows finds the same-curve agreement key exactly where you said it should be, and uses it.

The signature calls during those same online logons are mixed: logon 1 sent one 48-byte value
(signed, `cbSignedData=96`) and **no 32-byte call at all**; logon 2 sent **32 bytes first —
refused — then 48, signed**. Both logons succeeded, as always online, because PKINIT carries
them. I cannot yet explain logon 1's missing 32: the one difference is that it was the
account's first-ever logon, with no cached credential yet in existence, while logon 2 had one
— consistent with the 32-byte call being tied to an existing cached credential, but that is
one observation of each and I am not claiming it.

**Offline logons (two attempts):** consuming the cached credentials written by the two online
logons above, same card, same account. 32 bytes only: eight `CardSignData` calls across the
two attempts, every one 32 bytes, every one refused, no 48-byte call ever made, and both
logons failed.

On "refused": `-1205` is OpenSC's mapping of the card's `6A 80`, not an OpenSC decision — the
stock path has no short-input rejection of its own (its only length handling truncates *long*
input), and on the control card we traced this identical failure to the raw APDU response with
OpenSC bypassed entirely, including a 65-byte probe to prove the bytes reach the card
unmodified. Both cards run the same applet binary. The card is the party saying no.

That 32/48 mix is the same split we measured on the control card during the Kerberos-policy
runs: 32 on the interactive path, 48 on the silent one — it tracks the calling context, not
the card. The canonical layout reproduces it unchanged.

## What that means

The provisioning critique was correct about our cards, and the theory built on it does not
hold. The claim was: *"Windows expects to find an EC key with the same curve as the signing
key … If it does not find one it falls back to using no hash or a hash this is 32 bytes
long"* [sic]. On this card Windows **does** find that key — canonical slot, same curve, derivation
demonstrably running against it — and it still sends 32-byte values, on every interactive
call we observed, offline and online (with logon 1's absent call noted above). It looks like 32 is not
a fallback for a missing agreement key: the key was present and in use, and the 32s came
regardless. The offline data alone carries that conclusion.

So offline ECC P-384 logon with an applet that enforces the hash-strength pairing still 
requires the short value to be padded, even on a card with nothing
non-canonical left in its key-management layout.

## Your question about the 32 bytes

> I don't believe we have seen what is in the 32 bytes.

We had, without realizing it: the minidriver's existing `loghex` writes the full buffer one
line below the `pInfo->pbData(n)` header our extraction always stopped at. No new build
needed — the bytes were in every capture. Pulling all of them from the retained logs:
**30 distinct 32-byte values and 26 distinct 48-byte values**, across both cards, three
minidriver builds, two days, online and offline. No 32-byte value is a prefix or suffix of
any 48-byte value. Pooled entropy 7.81 bits/byte (960 bytes; uniform random measures ~7.8 at
that sample size). No DER structure, no ASCII, 28 distinct leading bytes in 30 values.
Representative:

```
32B: AF932DBCF137E39C9A424A038592F91FBDF564E747DABA73D321C1EC22AB07AC
32B: 44CD2119FE3CC6572B807C7DAAEE3480785284FAF829564E324F9CD06B7D75D8
48B: D8E96B8E5AC1361EB6639164C342EFB340D19BD44F1C830B4C5CC26E91557E7C8EEFEB5C4264B88AA4ECB076A6E7D760
```

The two outcomes that would have marked it as a hash — structure, or stability across calls —
did not occur.

What this establishes: the values are fresh on every call, unstructured, arrive **unlabelled**
(`aiHashAlg=0`, no padding info — Windows names no digest for them), and are not truncations
of the 48-byte path. That rules out the missing-agreement-key fallback theory and rules out a
digest of any stable data. What it does not establish: whether Windows hashed something fresh
to produce them — entropy cannot distinguish a raw challenge from a hash of fresh random
material, and I am not claiming it can.

This pokes at interpretation of SP 800-73 A.3, and that argument probably belongs in makinako/OpenFIPS201#89. This PR's position there is narrow: the hash-strength pairing rule governs hash-then-sign of a message
the card can see, not a bare, unlabelled, caller-supplied value.

Issue #3814, opened 2026-09-02
MinGW cross-build of opensc-minidriver.dll produces a DLL that cannot load: LoadLibrary fails with error 127 (TaskDialogIndirect needs a ComCtl32 v6 manifest)


spurglegooge:
### Describe the bug

Hi — follow-up to #3813, this one about the **MinGW/autotools** cross-build of
`opensc-minidriver.dll` rather than the card. The MSVC build is unaffected.

## Component

**Windows Minidriver** build (`src/minidriver/`, `configure.ac`) — autotools cross-build,
not MSVC. No card is involved: the DLL fails to load before it ever talks to one.

## Version number

**0.27.1** release tarball. `src/minidriver/minidriver.c` on **master** still contains the
same single `TaskDialogIndirect` call site, and `src/minidriver/Makefile.am` still has no
manifest handling, so I believe master is affected identically. I have not run a Nightly
binary — this is a cross-build issue, so the Nightly Windows artifacts (built with MSVC)
would not exhibit it.

## Operating system

- **Build host:** Debian 13, mingw-w64 GCC 14 (`x86_64-w64-mingw32`)
- **Target:** Windows 11 Pro 10.0.26200.9168 (x64)

## Steps to reproduce

1. On a Debian 13 host, install `mingw-w64` and place a WDK/CPDK `cardmod.h` where the
   compiler can find it.
2. Configure and build for Windows with the minidriver enabled:
   ```
   ./configure --host=x86_64-w64-mingw32 --enable-minidriver \
               --enable-openssl --enable-zlib
   make
   ```
3. Copy the resulting `opensc-minidriver.dll` to a Windows 11 machine and try to load it:
   ```powershell
   $sig='[DllImport("kernel32", SetLastError=true, CharSet=CharSet.Unicode)] public static extern IntPtr LoadLibraryW(string p);'
   $k = Add-Type -MemberDefinition $sig -Name L -PassThru
   $k::LoadLibraryW("C:\path\to\opensc-minidriver.dll")
   [Runtime.InteropServices.Marshal]::GetLastWin32Error()
   ```

## Expected behavior

`LoadLibrary` returns a non-zero handle, and the minidriver can be registered for a card's
ATR and used for logon — as the MSVC-built DLL does.

## Actual behavior

`LoadLibrary` returns 0 and the error is **127** (`ERROR_PROC_NOT_FOUND`):

```
FAILED : opensc-minidriver.dll (8492467 bytes)
[...]

PR #3818, opened 2026-09-04
minidriver: correct where cardmod.h comes from

spurglegooge:
In #3814 you mentioned:

> We didn't find a reliable source for `cardmod.h`. Maybe WSL with an original
> `cardmod.h` could be a possibility.

We went looking for the same thing and found a chain that works on plain Linux with no
Windows machine involved, which should suit CI better than WSL. The one-line comment fix in this PR
records the source; the full procedure is below in case it is useful for CI.

### What this PR changes

`minidriver.c` said the header comes "from CNG SDK or platform SDK". That is not correct
and it costs time, so this corrects it to name the CPDK and point at the
`CPDK_INCL_DIR` line in `win32/Make.rules.mak` that already encodes the real location.

### Where it is not

Not in the Windows SDK, the WDK, the EWDK, or the `Microsoft.Windows.SDK.CPP` /
`Microsoft.Windows.WDK.*` nuget packages. We downloaded and searched all four, including a
1.8 GB EWDK ISO with a complete 1,612-header Kits tree. We couldn't find it.

`win32/Make.rules.mak` already points `CPDK_INCL_DIR` at the right directory, so the build
system knows even if the comment did not.

### Where it is

Only in the **Cryptographic Provider Development Kit (CPDK)**. The installer is a WiX Burn
bundle, so the payload has to be extracted rather than run. Tools on Debian:
`p7zip-full cabextract msitools`.

1. Landing page (check for a newer version first):
   <https://www.microsoft.com/en-us/download/details.aspx?id=30688>

2. Bootstrapper:
   ```
   https://download.microsoft.com/download/1/7/6/176909b0-50f2-4df3-b29b-830a17ea7e38/CPDK_RELEASE_UPDATE/cpdksetup.exe
   sha256 76d6f7580fce9bb1045b5f0871718342443ec5d99c316489ef8c7bc83d4dd6ed  (1,281,504 bytes)
   ```

3. List the Burn payloads (`7z` treats the bundle as an archive; file `0` is the manifest):
   ```bash
   7z x cpdksetup.exe -oburn
   grep -oE 'FilePath="Installers\\[^"]+"' burn/0
   ```

4. Payloads sit beside the bootstrapper under `.../CPDK_RELEASE_UPDATE/Installers/`
   (URL-encode the spaces):
   ```
   Installers/Windows%20Cryptographic%20Provider%20Development%20Kit-x86_en-us.msi
   sha256 b401010d60c5cb0e4f6e3e12e938f78e159331f83843405d8b9a5bebba803a22  (409,600 bytes)

   Installers/4c9acdde8c6aed110d8dac9ace60b90a.cab
   sha256 84549b70cc6fe2911f936478b546c6fc0766019605d43da8cfc54f660af2296a  (65,929 bytes)
   ```

5. Cab entries have hashed names, so map them through the MSI `File` table:
   ```bash
   cabextract -d cab 4c9acdde8c6aed110d8dac9ace60b90a.cab
   msiinfo export cpdk.msi File | grep -i cardmod
   # fild99c8486373dedc4d3327c73db2af022 ... cardmod.h  66008
   cp cab/fild99c8486373dedc4d3327c73db2af022 cardmod.h
   ```

Result:
```
cardmod.h  66,008 bytes
sha256     a6222ef5ac63d0db989a85810d5b73d96990122db371c9f0cc50a36d4dcdb598
```

The same cab also carries `bcrypt.h`, `bcrypt_provider.h`, `cspdk.h`, `msclmd.h`,
`ncrypt.h`, `ncrypt_provider.h` and `sslprovider.h`.

### One trap worth documenting

mingw-w64 ships its own `cardmod.h` at `/usr/share/mingw-w64/include/cardmod.h`. It is an
older partial subset, roughly 10 KB, and the minidriver fails against it with 103
undefined identifiers. Because a header of the right name is present, this reads as a
source problem rather than a missing dependency. A quick discriminator:

```bash
grep -c "CARD_DATA_VERSION_SEVEN\|CONTAINER_MAP_RECORD\|PIN_ID\|CARD_CACHE_FILE_FORMAT" cardmod.h
# real CPDK header: non-zero (26 in our copy).  mingw-w64 subset: 0
```

### On WSL

To be clear about scope: the above is only about *obtaining* the header. It does not
compete with using WSL in a GitHub Actions job to run the MinGW build — the two fit
together, since a runner still needs the header from somewhere before it can compile.

### What we are not attaching

We are deliberately not attaching the header. Our own reading of the CPDK licence terms is
that internal use is unlimited but redistribution is not, so we are sharing the procedure
and the hashes rather than the file. That is our reading of our own obligations only —
whether OpenSC can vendor or ship it is a question for the project.

### DIFF
--- a/src/minidriver/minidriver.c
+++ b/src/minidriver/minidriver.c
@@ -20,7 +20,9 @@
  */
 
 /*
- * This module requires "cardmod.h" from CNG SDK or platform SDK to build.
+ * This module requires "cardmod.h" from the Microsoft Cryptographic Provider
+ * Development Kit (CPDK) to build. It is not part of the Windows SDK, WDK or
+ * EWDK. See win32/Make.rules.mak (CPDK_INCL_DIR) for the install location.
  */
 
 #include "config.h"


metsma:
Here is mentioned
https://github.com/OpenSC/OpenSC/wiki/Compiling-on-Windows#build-requirements

metsma (approved):

frankmorgner (approved):
My hint regarding WSL was not a suggestion for a general fix, but rather a suggestion on how to test building a cross compiled minidriver in CI. via GH actions, we have the option to copy cardmod.h into WSL, and then to compile here with MINGW.

Anyway, codewise the change looks good. Improvements for CI may be suggested in a seperate PR.

spurglegooge:
Thanks both.

@metsma — useful link, the wiki Build requirements page is a better landing spot for this
than the source comment was. The in-tree comment was still naming the wrong SDK, which is
what this changes.

@frankmorgner — I misread the WSL hint, and I will correct the "On WSL" section of the
description. Copying `cardmod.h` into WSL and building with MinGW inside a GH Actions job
is a different proposition from sourcing the header, and my note was answering a question
you had not asked. The extraction procedure is about getting the header onto the runner in
the first place, so the two fit together rather than competing.

PR #3819, opened 2026-09-04
md: implemented graceful handling if comctl is missing for UI

### LINKED ISSUE #3814: MinGW cross-build of opensc-minidriver.dll produces a DLL that cannot load: LoadLibrary fails with error 127 (TaskDialogIndirect needs a ComCtl32 v6 manifest)

frankmorgner:
fixes https://github.com/OpenSC/OpenSC/issues/3814

##### Checklist

- [ ] macOS token is tested


metsma (approved):

spurglegooge:
Following on from #3823 — I am closing that in favor of this, which is the better fix.

One thing I might look at before merging. `hComctl` is `static`, but it is freed
unconditionally:

```c
static HMODULE hComctl = NULL;

if (displayUI) {
    if (NULL == (hComctl = LoadLibraryA("comctl32.dll")) || NULL == (pfnTaskDialogIndirect = ...)) {
        displayUI = FALSE;
    }
}

if (!displayUI) {
    rv = md_dialog_perform_pin_operation_thread(parameter);
    SecureZeroMemory(parameter, sizeof(parameter));
    if (hComctl)
        FreeLibrary(hComctl);      /* <-- frees what the PREVIOUS call left behind */
    return rv;
}
...
FreeLibrary(hComctl);
```

A call with `displayUI == TRUE` loads the module, uses it, and frees it at the bottom —
but `hComctl` keeps the now-stale handle because it is static. The next call with
`displayUI == FALSE` skips the `LoadLibraryA`, reaches `if (hComctl)`, and frees the same
handle again. That is a refcount underflow on a DLL other parts of the process are using.

It matters because `displayUI == FALSE` is the common path, not a corner case:
`CardAuthenticateEx` only sets it when the caller supplied no PIN *and* the reader is a
pinpad, so every ordinary logon takes the `!displayUI` branch.

Making `hComctl` automatic rather than static fixes it — it is NULL on entry to each call
and the two `FreeLibrary` sites then only ever see a handle this call obtained. If the
intent was to cache the module across calls, then the `FreeLibrary` calls should go
instead, and the handle should be released on unload.

Offer: the checklist has "Windows minidriver is tested" unticked, and I have the setup
that reproduces #3814 — a MinGW cross-build loaded by `lsass` on Windows 11 with a P-384
PIV card, plus a bare no-manifest test binary that reproduces the 127 failure outside
`lsass`. Happy to build this branch and confirm it loads and that PIN operations still
work, if that is useful. Say the word and I will report back with the log.

frankmorgner:
thank you. I had overlooked the dangling pointer when trying to avoid multiple loads of the library. should be fixed now

PR #3822, opened 2026-09-08
piv: Support GET METADATA to discover keys without certificates

Jakuje:
The yubikeys extension GET METADATA allows pulling public keys from token without the need to have sidecar file with the SPKI and environment variable.

##### Checklist

- [ ] macOS token is tested

### DIFF
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -1811,13 +1811,15 @@ piv_general_io(sc_card_t *card, int ins, int p1, int p2,
 			(sendbuf ? SC_APDU_CASE_3_SHORT : SC_APDU_CASE_1);
 
 	sc_format_apdu(card, &apdu, cse, ins, p1, p2);
-	apdu.flags |= SC_APDU_FLAGS_CHAINING;
+	if (sendbuflen > 255) {
+		apdu.flags |= SC_APDU_FLAGS_CHAINING;
 #ifdef ENABLE_PIV_SM
-	if (card->sm_ctx.sm_mode != SM_MODE_NONE && sendbuflen > 255) {
-		/* tell apdu.c to not do the chaining, let the SM get_apdu do it */
-		apdu.flags |= SC_APDU_FLAGS_SM_CHAINING;
-	}
+		if (card->sm_ctx.sm_mode != SM_MODE_NONE) {
+			/* tell apdu.c to not do the chaining, let the SM get_apdu do it */
+			apdu.flags |= SC_APDU_FLAGS_SM_CHAINING;
+		}
 #endif
+	}
 	apdu.lc = sendbuflen;
 	apdu.datalen = sendbuflen;
 	apdu.data = sendbuf;
--- a/src/libopensc/card-piv.c
+++ b/src/libopensc/card-piv.c
@@ -386,6 +386,16 @@ static const struct sc_card_error piv_sm_errors[] = {
 
 #define PIV_PAIRING_CODE_LEN	 8
 
+#define PIV_NUM_YK_SLOT_INFO 26
+
+typedef struct yk_slot_info_st {
+	u8 slot;
+	u8 policy;
+	u8 touch;
+	u8 algorithm;
+	struct sc_lv_data pubkey;
+} yk_slot_info_t;
+
 typedef struct piv_private_data {
 	struct sc_lv_data aid_der; /* previous aid response to compare */
 	int enumtag;
@@ -418,11 +428,7 @@ typedef struct piv_private_data {
 	unsigned int card_issues; /* card_issues flags for this card */
 	int object_test_verify; /* Can test this object to set verification state of card */
 	int yubico_version; /* 3 byte version number of NEO or Yubikey4  as integer */
-	struct {
-		u8 slot;
-		u8 policy;
-		u8 touch;
-	} yk_pin[26];
+	yk_slot_info_t yk_slot_info[PIV_NUM_YK_SLOT_INFO];
 	unsigned int ccc_flags;	    /* From  CCC indicate if CAC card */
 	unsigned int pin_policy; /* from discovery */
 	unsigned int init_flags;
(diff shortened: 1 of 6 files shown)


frankmorgner (approved):

Jakuje:
I see the kryoptic tests still fail. I will double check what is going on there and do a quick fix of the reference file.

Jakuje:
oh, the error is only with pkcs11 spy. I will pull the p11tool and pkcs11-spy fixes from #3754 to get them merged first.

Jakuje:
CI looks happy after rebase so merging this one too. I will rebase the rest on top of these.

dengert:
Where are you going with this? 

Only Yubico and a hand full of other "PIV-like:  cards do this.  It won't work on real PIV cards. 
Are expecting to use this  from pkcs11?

For the non Yubico and even Yubico PIV cards would be to write to the card a dummy certificate. It looks like a real certificate but not usable anywhere other then to sign a CSR it could be self signed. Then replace it with the real certificate.  

As I say every time, NIST let it up to card (applet) vendors on how to provision a PIV card. Every card PIV card vendor provides at least a Windows minidriver that can be used to provision their card. Even the 2 open source PIV applets I know of can use Yubico-piv-tool.   The main missing piece of NIST 800-73-* is how to import private key.

Jakuje:
This is the reason why this is gated behind the yubico version check.

In this particular case, I found it very useful to test the #3754 without the need to mangle with the environment variables. Yes, I could do dummy certificates for signing keys, but that wont work for ML-KEM and XEDDSA.

Yes, for most common use cases you have a certificates,but there are uses without certificates and I think its worth supporting them.

dengert:
OK.  Note that `card-piv.c` and `pkcs15-piv.c` read all the certificates to find a public key, as NIST does not define a public key object on the card accept in the SPKI of matching certificate. 

For XEDDSA, I have my test CA, based on OpenSSL CA.pl script and demoCA  and use something like this in a `gen_req.sh`
Done some testing using Yubikey 5 NFC 5.7.1 with XEDDSA
Might work for ML-KEM too. 

```
331 if [[ "$FORCE" == "YES" || "$keytype" == "XEDDSA" ]] ; then
332         echo "generating cert for $keytype type key using $DEECA"
333 
334         SERIALOLD=`cat $DEECA/serial`
335         SERIAL=`printf '%X' $((0x$SERIALOLD +0x01))`
336         echo SERIALOLD $SERIALOLD
337         echo SERIAL for cards/$CARDN.cert.$KEYID.pem $SERIAL
338 
339         openssl x509 -new -CA $CACERT -CAkey $CAKEY \
340                 -force_pubkey $PUBKEY \
341                 -set_subject "/CN=$USERN" \
342                 $DAYS -set_serial 0x$SERIAL \
343                 -extfile $CNF \
344                 -extensions "v3_req$KEYUSAGE" \
345                 -outform PEM -out cards/$CARDN.cert.$KEYID.pem
346         if [ $? -ne 0 ] ; then
347                 exit $?
348         fi
349         if [ -f $DEECA/newcerts/$SERIAL.pem ] ; then
350                 echo "Problem $DEECA/newcerts/$SERIAL.pem exists, not copying the cert"
351 #               exit 9
352         fi
353         cp cards/$CARDN.cert.$KEYID.pem $DEECA/../valid.cert.pem
354         cp cards/$CARDN.cert.$KEYID.pem $DEECA/newcerts/$SERIAL.pem
355         echo $SERIAL > $DEECA/serial
356         echo $SERIALOLD > $DEECA/serial.old
357         ( cd $DEECA/.. ;  openssl ca -config CA.openssl.cnf -valid valid.cert.pem )
358         echo database serial $(cat < $DEECA/serial)
359         exit
360 fi
361 # end of force
```

The problem with "Yubikey-like" cards, is they may respond to the Yubikey get version APDU  but may not follow all the features at that level or report a level Yubikey does not have yet. 
```
/opensc-tool -s "00FD000000"
Using reader with a card: Nitrokey Nitrokey 3 [CCID/ICCD Interface] 00 00
Sending: 00 FD 00 00 00 
06 06 06 ...
```
Yubikey does not have a 6.6.6 AFAIK

Jakuje:
I do not have Nitrokey 3. Does it support the metadata or do we need some workaround for that? The lastest yubikey version out there is v5.

dengert:
> Does Nitrokey 3 support the metadata or do we need some workaround for that?

Don't know will have to look closer. I was using OpenSC piv-tool in my testing. If using piv-tool and Nitrokey-App on windows it might be different. 
That's the problem of following someone elses "standards" but not all of them.

dengert:
NitroKey 3A NFC with 9A, 9C, 9D keys and certs I created on card.

Here is when it looks for 9E:
[get9E.txt](https://github.com/user-attachments/files/32485411/get9E.txt)

Looks like `00 F7 00 9E 00`  worked as expected. Don't have any where there is a key but no cert.

dengert:
Here is a token2  with 9A 9C 9D 9E 82  and created some others but no cert.
using the "Token2 companion App v2" on windows  

It responded to `00 FD 00 00 03` with  `05 70 00 90 00`

I believe I deleted 83 certs/keys but looks like the pubkey is still there. 
Here is the 83: 
[token2.txt](https://github.com/user-attachments/files/32486853/token2.txt)

Jakuje:
Thank you for testing with these two! So it looks like Nitrokey does not provide the public key, while Token2 does. Looking at the log:
```
P:37422; T:0x139886619816960 14:27:51.212 [pkcs15-tool] card.c:1106:sc_card_ctl: returning with: 0 (Success)
P:37422; T:0x139886619816960 14:27:51.212 [pkcs15-tool] pkcs15-piv.c:1155:sc_pkcs15emu_piv_init: Got unknown algorithm ID 0
```
the `sc_card_ctl` succeeds even though there is no public key information in metadata, which should be fixed. Can you try with #3839?

dengert:
Here is with #3839 looks better.

[with-change.txt](https://github.com/user-attachments/files/32515975/with-change.txt)

Jakuje:
> Here is with #3839 looks better.
> 
> [with-change.txt](https://github.com/user-attachments/files/32515975/with-change.txt)

Thanks for testing! Let me know if you will have some more concerns! Feel free to update the #3839 with the comment.

PR #3825, opened 2026-09-14
cache: Filter out malicious characters

Jakuje:
Found by AISLE in partnership with Red Hat

https://github.com/OpenSC/OpenSC/security/advisories/GHSA-f5jj-f5qm-qq3c

##### Checklist

- [ ] macOS token is tested


frankmorgner (changes requested):
Please allow a static UID to take the place of a possible non-allowed serial number. Also, please allow last_update to fall back to `"NODATE"` in case of a bad value.

e.g. use local variable `char *serial_number = p15card->tokeninfo->serial_number; if (!is_safe_cache_component(serial_number) serial_number = NULL;`. Similar for `last_update`: `if (!last_update || !is_safe_cache_component(last_update)) last_update = "NODATE";`

frankmorgner (approved):

Jakuje:
Fixed the missing opening curly brace to fix the build.

frankmorgner (approved):

frankmorgner:
Thank you

PR #3826, opened 2026-09-15
Extend the C_LoadModule to handle PKCS#11 versions and softhhsm bugs more reliably

Jakuje:
The SoftHSM returns version 3 from the GetFunction list (still) and
when we use our C_LoadModule() wrapper, we lose information about this
in the pkcs11-spy and we can not reliably filter out this case without
risking buffer overruns and crashes.
    
This moves the implementation to the pkcs11-spy init function and
when we get wrong version from the old API, we safely copy just
the part of the old API, instead of doing some guesswork.

Pulled minimal part from #3754 that can be reviewed separately.

##### Checklist

- [ ] macOS token is tested


frankmorgner:
wouldn't it be more useful to extend C_LoadModule to allow specifying the interface version? this would then also be possible to be used in the other PKCS#11 tools...

Jakuje:
> wouldn't it be more useful to extend C_LoadModule to allow specifying the interface version? this would then also be possible to be used in the other PKCS#11 tools...

The `C_LoadModule()` is used at this moment only in pkcs11-tool and pkcs11-register. I agree that the pkcs11-tool could make use of this too, but I had quite hard time figuring out how to improve the API when the pkcs11 modules are broken as softhsm and report v3 version for getFunctionList.

But in the end I did that. Lets see how it will work.

frankmorgner (approved):

Jakuje:
Thank you for the review. I will merge this one so I can rebase the #3822 on top to verify it solves the test issue (otherwise I will have to investigate it further.

PR #3832, opened 2026-09-16
Remove needless include of assert.h

Jakuje:
This is a fixup of the a807d69706ee6d88f7e57f4b455e5631f331d5a1 which removed almost all the asserts from the codebase, but left the includes.

##### Checklist

- [ ] macOS token is tested


src/tools/pkcs11-tool.c
```diff
@@ -20,6 +20,7 @@
 
 #include "config.h"
 
+#include <assert.h>
```
frankmorgner, line 23:
```suggestion
```

frankmorgner (changes requested):

src/tools/pkcs11-tool.c
Jakuje: The pkcs11 tool has static_assert, which comes from the same header file. I removed it from some internal header file so it needs to be included explicitly to build

frankmorgner: thanks for the clarification

frankmorgner (approved):

Jakuje:
Thank you for the review! I think this is pretty non-controversial so I will get it merged.

PR #3833, opened 2026-09-17
SC HSM PKA registration

lrascao:
Two independent bugs on the SmartCard-HSM M-of-N public key registration path,
found while implementing the PKA challenge-response over PC/SC.

### `card-sc-hsm: fixed status word check when probing for a registered public key`

`verify_certificate()` asks the card whether a CHR is already known and treats
`6A88` ("referenced data not found") as "not yet — import it below". The guard
used `&&` where `||` was meant:

```c
if (apdu.sw1 != 0x6A && apdu.sw2 != 0x88) {
```

so any status whose SW1 was `0x6A` *or* whose SW2 was `0x88` satisfied it.
`6A82`, for instance, fell through and the import was attempted against a card
that had reported something else entirely.

### `sc-hsm-tool: log in before registering a public key`

The card requires an authenticated session before accepting a public key —
`PUK_AUTH 'Add'` checks the security status and fails otherwise.
`register_public_key()` never established one, so `-g` could not succeed
regardless of the file passed to it. This calls `ensure_login()` as the other
write operations in the tool already do, which also gives `-g` the same
`--pin` handling and interactive prompt as `--unwrap-key` and
`--import-dkek-share`. The man page entry gains the same `--pin` note those
options carry.

### Testing

`opensc-tool -n` reports `SmartCard-HSM version 6.6` and `6.4`, but these are
**Pico HSM** (`polhenarejos/pico-hsm`), a compatible reimplementation rather
than genuine SmartCard-HSM hardware — flagging that explicitly, since the
driver name doesn't distinguish them.

Neither fix is reproducible end to end on this hardware, and I'd rather say so
than imply otherwise. Registration fails earlier for an unrelated reason: the
device CVC chain on these devices is self-signed, so `verify_certificate()`
cannot import it (the CHR is unknown, and selecting the issuer of a
self-signed CA selects itself). The `&&` branch is therefore never reached
here, and the missing login is masked by that earlier failure.

Both are visible by inspection against the card's documented behaviour, and
the second is confirmed at the firmware level: pico-hsm's `cmd_puk_auth`
rejects `Add` unless `has_session_pin || has_session_sopin`.

##### Checklist
- [ ] macOS token is tested


src/libopensc/card-sc-hsm.c
```diff
@@ -1488,7 +1488,11 @@ static int verify_certificate(sc_card_t *card, sc_cvc_t *cvc,
 		/* already known */
 		LOG_FUNC_RETURN(card->ctx, SC_SUCCESS);
 	}
-	if (apdu.sw1 != 0x6A && apdu.sw2 != 0x88) {
+	/* Only 6A88 ("referenced data not found") means the key is simply not
+	 * known yet and may be imported below; any other status is an error.
+	 * With && this rejected nothing whose SW1 was 0x6A or whose SW2 was
+	 * 0x88, so e.g. 6A82 fell through and the import was attempted anyway. */
```
frankmorgner: can you check here *whether* a key is about to be imported rather than ignoring this error code unconditionally?

Jakuje: I think this comment should be in the commit message only. It does not add any useful context to the code, only to the change itself.

lrascao: Agreed, moved to the commit message, and the remaining comment describes what
the check does rather than what it used to do.

lrascao: `sc_check_sw()` already maps 6A88 to `SC_ERROR_DATA_OBJECT_NOT_FOUND`, so I've
changed this to test that instead of the raw status bytes, and to propagate the
actual error rather than `SC_ERROR_UNKNOWN`. Continuing to the import now
happens only for the one status that means the key isn't there yet; anything
else returns.

I read your comment as objecting to the error being discarded whatever it was —
let me know if you meant something more literal about checking that an import
is going to happen, since the flow always reaches the import past this point.

lrascao:
sorry for the force push but wanted to keep clean history

src/libopensc/card-sc-hsm.c
frankmorgner: thanks for the clarification. I think the change in the code makes that clearer now.

frankmorgner (approved):

Jakuje (approved):

Jakuje:
You will also have to update the generated html pages, see the CI results. but otherwise lgtm

Issue #3834, opened 2026-09-19
Support for ePass2003 tokens with Microsoft minidriver (MSCP) layout instead of PKCS#15


manoelnetoufc:
### Problem statement

Some FEITIAN ePass2003 tokens are not PKCS#15 cards, so OpenSC cannot use them. With 0.27.1 (and current master):

```
$ opensc-tool --atr --name
3b:9f:95:81:31:fe:9f:00:66:46:53:05:01:00:11:71:df:00:00:00:00:00:13
epass2003
$ pkcs15-tool --list-certificates
PKCS#15 binding failed: Unsupported card
$ pkcs11-tool -L
Slot 0 (0x0): Feitian ePass2003
  (token not recognized)
```

The `epass2003` driver matches and secure messaging (SCP01 with the default transport key) works, so files can be read. What is missing is the PKCS#15 layer. There is no DF 5015. The card carries a Microsoft minidriver (MSCP) style personalization instead:

- `3F00/2003` (FCI file name `ENTERSAFE-ESPK`) contains a flat index, EF `6f01`, made of 30-byte records: FID (LE, `FFFF` = directory label), directory name[10], file name[13], ACL byte, used size (LE), allocated size (LE).
- The listed files (`certNNNN`, `pubkNNNN`, `prvkNNNN`, `containermap`, `cmapfile`, `tokeninfo`, `cardid`) are raw streams of PKCS#11 attributes: `u32 type (LE), u32 length (LE), value`. There is no other envelope.
- Vendor attribute `0x80000100` holds the internal key handle (`0xA0xx` for private keys, `0x80xx` for public keys). `containermap` pairs them.
- `MSE:SET` for RSA signing expects that raw handle in tag `0x81`. `epass2003_set_security_env()` currently computes `0x2900 + 0x20 * key_ref`, which does not exist on this card (`SW 6A82`).
- Some objects have no `CKA_ID` of their own (older containers).

The vendor PKCS#11 module works with this token, but applications that cannot load it have no way to use it. In particular macOS CryptoTokenKit (Chrome, Safari) cannot use it through the OpenSC token.

A similar symptom was reported in #2424; I cannot tell whether that card uses the same layout.

### Proposed solution

I have two small patches on top of master, tested on a real token, and I am happy to open a PR:

[...]

PR #3835, opened 2026-09-21
Add PKCS#15 emulator for MSCP-personalized ePass2003

manoelnetoufc:
Add a PKCS#15 emulator for ePass2003 tokens with a Microsoft minidriver (MSCP) personalization, as discussed in #3834.

Some FEITIAN/EnterSafe ePass2003 tokens (COS version bytes `01:00:11` in the ATR) are not PKCS#15 cards: there is no DF 5015. `3F00/2003` (FCI file name `ENTERSAFE-ESPK`) holds a flat index (EF 6f01) of files that store the certificates and keys as raw PKCS#11 attribute streams, plus a container map. Today `sc_pkcs15_bind()` fails with "Unsupported card".

- `pkcs15-entersafe-mscp.c` reads the index and publishes certificates, RSA public/private keys and the PIN as PKCS#15 objects. Objects without `CKA_ID` (older containers) take the id of the object with the same RSA modulus (private keys: of the public key the containermap pairs them with), since CryptoTokenKit rejects duplicate ids.
- `card-epass2003.c`: new `SC_CARDCTL_ENTERSAFE_MSCP_MODE`, set only by the emulator after it identifies the card. In this mode `MSE:SET` uses the raw vendor key handle (0xA0xx) instead of `0x2900 + 0x20*n`. Other ePass2003 cards keep the existing behaviour.
- The emulator only binds when the card type is `epass2003` and the DF 2003 FCI name is `ENTERSAFE-ESPK`.

##### Tested
- Card: ePass2003, ATR `3b:9f:95:81:31:fe:9f:00:66:46:53:05:01:00:11:71:df:00:00:00:00:00:13` (`opensc-tool -n`: epass2003), macOS 26, Apple Silicon.
- `pkcs15-tool` lists certificates, keys and PIN; `pkcs11-tool -O` lists the objects.
- RSA-2048 signature through opensc-pkcs11 (`--login --sign -m RSA-PKCS`), verified against the certificate public key.
- macOS CryptoTokenKit token (OpenSCToken built against this branch): login with the certificate in Chrome and Safari. The stock token app needs the CryptoTokenKit reader build of this library; the extension also needs the PIN format limits from the PKCS#15 PIN object (default CTK format is 8 digits).
- PIN retry counter unchanged after all tests.

Not tested: Windows minidriver.

##### Checklist
- [x] macOS token is tested


src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,868 @@
...
+
+	a_mod = es_find_attr(pubattrs, pubnattrs, P11_CKA_MODULUS);
+	a_handle = es_find_attr(pubattrs, pubnattrs, P11_CKA_VENDOR_KEY_HANDLE);
+	if (!a_mod || a_mod->len == 0 || !a_handle || a_handle->len != 4) {
+		free(pubattrs);
+		return SC_ERROR_OBJECT_NOT_FOUND;
+	}
+
+	*out_modulus_len = a_mod->len;
+	*out_vendor_handle = es_le32(a_handle->value) + ES_MSCP_PUB_TO_PRIV_HANDLE_OFFSET;
```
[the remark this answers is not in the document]
manoelnetoufc: Agreed, fixed in 05178d5. The private handle now comes from the containermap (32 byte header, then 24 byte records with the private and public handle as LE32). The fixed offset is gone; a key without an entry is skipped instead of guessed. Checked against the real containermap bytes of the card.

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,868 @@
...
+
+	a_id = es_find_attr(attrs, nattrs, P11_CKA_ID);
+	if (a_id && a_id->len > 0) {
+		size_t idlen = a_id->len;
+
+		if (idlen > sizeof(prkey_info.id.value))
+			idlen = sizeof(prkey_info.id.value);
+		memcpy(prkey_info.id.value, a_id->value, idlen);
+		prkey_info.id.len = idlen;
+	}
```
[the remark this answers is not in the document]
manoelnetoufc: Agreed, fixed in 05178d5. Private keys are now included: they take the id of the public key the containermap pairs them with, and anything still id-less gets a unique synthetic id. Note that every private key on the card I have carries a CKA_ID, so this path is only covered by review and a unit test of the pairing helpers, not by a real id-less key.

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,868 @@
...
+	 * personalization is literally "ENTERSAFE-ESPK". That, not anything
+	 * inside tokeninfo, is what identifies this specific MSCP layout. */
+	sc_format_path(ES_MSCP_APP_PATH, &path);
+	r = sc_select_file(card, &path, &file);
+	if (r < 0) {
+		sc_file_free(file);
+		LOG_FUNC_RETURN(ctx, SC_ERROR_WRONG_CARD);
+	}
+	if (file == NULL || file->namelen < strlen(ES_MSCP_TOKEN_LABEL) ||
+			memcmp(file->name, ES_MSCP_TOKEN_LABEL, strlen(ES_MSCP_TOKEN_LABEL)) != 0) {
```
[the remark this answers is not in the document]
manoelnetoufc: Good catch on the prefix (`ENTERSAFE-ESPKX`), fixed in 05178d5. A literal exact-length comparison would not work: the FCI name field is 16 bytes, the 14 characters plus two zero bytes. The check is now the 14 characters followed only by zero padding.

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,868 @@
...
+	if (r < 0) {
+		free(entries);
+		LOG_FUNC_RETURN(ctx, SC_ERROR_WRONG_CARD);
+	}
+
+	/* From here on this really is our card: log failures instead of
+	 * silently deferring to the next emulator. */
+
+	/* RSA keys on this personalization are addressed by their raw vendor handle */
+	r = sc_card_ctl(card, SC_CARDCTL_ENTERSAFE_MSCP_MODE, NULL);
```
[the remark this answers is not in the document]
manoelnetoufc: Agreed, fixed in 05178d5. MSCP mode is now enabled last, after everything that can fail has succeeded, so there is nothing to roll back. Verified with an RSA signature on the card after the change.

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,996 @@
...
+
+	n = 32;
+	if (n > len)
+		n = len;
+	while (n > 0 && buf[n - 1] == ' ')
+		n--;
+	if (n >= label_out_size)
+		n = label_out_size - 1;
+	memcpy(label_out, buf, n);
+	label_out[n] = 0;
```
frankmorgner:
```suggestion
	label_out[n] = '\0';
```

frankmorgner (commented):
Looks good, thank you.

please think about using `array = {0};` instead of `memset(array, 0, sizeof array);` for initializing a structure or buffer.

src/libopensc/pkcs15-entersafe-mscp.c
manoelnetoufc: Applied literally in efafa67.

manoelnetoufc:
Thanks. Converted the memset(&x, 0, sizeof x) zero-inits in pkcs15-entersafe-mscp.c to = {0} at declaration in efafa67 (cert_info/cert_obj, pubkey, pubkey_info/pubkey_obj, prkey_info/prkey_obj, pin_info/pin_obj, and the certs/pubs arrays in es_fill_missing_ids).

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,1161 @@
...
+};
+
+/* internal/containermap: a 32 byte header, then 24 byte records holding the
+ * private and public vendor key handles as little-endian 32-bit values (the
+ * rest of a record is zero). Unused records are zero and end the list. */
+#define ES_MSCP_CMAP_HEADER_LEN 0x20
+#define ES_MSCP_CMAP_RECORD_LEN 0x18
+
+static void
+es_parse_containermap(const u8 *buf, size_t len, struct es_containers *cm)
```
Jakuje, line 524: I wrote containermap parsing for idprime driver couple of years back:

https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/card-idprime.c#L311

I think its mostly the same structure. There is also cmap file parsing written in the gids and srbeid drivers. Is there some way to consolidate the parsing into some common code to avoid code duplication instead of creating one more?

manoelnetoufc: Thanks for the pointers. I compared the three, and this emulator cannot share their parser, because it reads a different file:

- idprime (`idprime_process_containermap()`), gids (`CONTAINER_MAP_RECORD` in card-gids.h) and srbeid (`CE_CMAP_*`) all parse `mscp/cmapfile`, the Microsoft `CONTAINER_MAP_RECORD`: 86-byte records with a UTF-16 GUID, flags and key sizes.
- This emulator parses `internal/containermap` (FID 2420), a Feitian-private file: a 32-byte header, then 24-byte records holding the private and public vendor key handles as LE32 (0xA020/0x8000, ...). It carries no GUID or key size. The MSE:SET key reference comes from it; `mscp/cmapfile` has no key handles, so it cannot replace it.

The directory index is not shared either: 30-byte records here, 21-byte objects in idprime, a 10-byte header plus 12-byte entries in srbeid.

I also checked whether the emulator should read this card's `mscp/cmapfile` (3 x 86 bytes), e.g. to attach the container GUIDs to the keys for the minidriver. Nothing on the card links a GUID to a key: the GUIDs appear only in `cmapfile`, and neither the key objects nor `internal/containermap` reference them. The only link would be the record position, which I cannot verify without Windows and the Feitian minidriver, and a wrong guess would point a certificate at another key's container. The flags (all valid) and key sizes (all 2048) add nothing to what the emulator already reads. So I left it out.

The duplication is the `CONTAINER_MAP_RECORD` parsing in gids, idprime and srbeid, which fits the srbeid follow-up @frankmorgner mentioned in #3841. `aux-data.h` already has `struct sc_md_cmap_record` but no parser, so a shared one could live in `aux-data.c`. @hirashix0, if it helps, I can write it or review it in a separate PR; I have no gids, idprime or srbeid card to test with.

Jakuje: Thanks for clarification! From the name I thought it will have something in common, but frankly I never looked too much into that.

src/tests/unittests/Makefile.am
```diff
@@ -16,10 +16,10 @@ endif
 
 noinst_PROGRAMS = asn1 simpletlv cachedir pkcs15filter openpgp-tool hextobin \
 	decode_ecdsa_signature check_macro_reference_loop strip_pkcs1_2_padding \
-	base64 parse_pkcs11_uri
+	base64 parse_pkcs11_uri entersafe-mscp
```
Jakuje: is there a reason why the unit test is not added to Makefile.mak for windows build?

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,1162 @@
...
+ * little-endian u32) in the card's attribute streams. Kept local instead of
+ * pulling in src/pkcs11/pkcs11.h, which belongs to a different module. */
+#define P11_CKA_LABEL		0x00000003U
+#define P11_CKA_VALUE		0x00000011U
+#define P11_CKA_DECRYPT		0x00000105U
+#define P11_CKA_SIGN		0x00000107U
+#define P11_CKA_SIGN_RECOVER	0x00000108U
+#define P11_CKA_ID		0x00000102U
+#define P11_CKA_MODULUS		0x00000120U
+#define P11_CKA_PUBLIC_EXPONENT 0x00000122U
```
Jakuje: we do pull pkcs11.h in some other drivers (coolkey) or other code in libopensc already (pkcs15-sec.c, pkcs15-skey.c) . While I agree that its not ideal and its mixing layers, I would consider using that directly. But this way is also justified (just a bit confusing by the names).

src/libopensc/pkcs15-entersafe-mscp.c
```diff
@@ -0,0 +1,1162 @@
...
+{
+	size_t i;
+
+	cm->n = 0;
+	for (i = ES_MSCP_CMAP_HEADER_LEN; i + 8 <= len && cm->n < ES_MSCP_MAX_CONTAINERS;
+			i += ES_MSCP_CMAP_RECORD_LEN) {
+		unsigned int priv = es_le32(buf + i);
+		unsigned int pub = es_le32(buf + i + 4);
+
+		if (priv < 0xA000 || priv > 0xAFFF ||
```
Jakuje: can we name this range `ES_MSCP_PRIV_HANDLE_MIN` and `ES_MSCP_PRIV_HANDLE_MAX`? Is there some relationship to the `ES_MSCP_PRIV_HANDLE_BASE`?

PR #3837, opened 2026-09-21
card-piv.c, cards.h - OpenFIPS201 ECDSA and  Windows issue

### LINKED ISSUE #3813: CardSignData fails with "Incorrect parameters in APDU" (-1205) when Windows signs a 32-byte digest with an EC P-384 key — breaks offline smart-card logon

dengert:
Fixes #3813 

With additional comments on #3270 , https://github.com/makinako/OpenFIPS201/issues/90 and https://github.com/makinako/OpenFIPS201/issues/89
by  @spurglegooge

OpenFIPS201 is only PIV applet known to enforce
NIST ECDSA restriction -  hash size == field length

Windows offline login using ECDSA on smart card may send something less then field length which will be padded. The result will by the same as if not padded.

This is a special use case:

- Requires the use of the OpenFIPS201 opensource applet being loaded on a card
- Used on Windows worstation
- Samba/Kerberos as the domain controller (not tested using Microsoft AD controllers) 
- Workstation admin must add registry entries for specific PIV card. OpenSC does not add registry entries for PIV ATRs  as Microsoft has built in PIV support and most PIV card vendors provide plug-and-play PIV minidrivers   
- User must have logged on the workstation at least once while connected to network
- Only when user tries to login to workstation when not connected to the network is this code needed
- May eventually be fixed by fixed by Microsoft to used the correct size data sent to card 

The OpenFIPS201 card is detected via response to SELECT AID and type card type = C_CARD_TYPE_PIV_II_OPENFIPS201
 a card_issues flag CI_ENFORCES_EC_HASH_EQ_KEYSIZEis theb set for  C_CARD_TYPE_PIV_II_OPENFIPS201

If other cards are found to fail with this problem, they can be assigned  a card type or temporarily assigned via `opensc.conf` to this card type.  

 On branch piv-ec-windows
 Changes to be committed:
	modified:   src/libopensc/cards.h

##### Checklist

Needs to be tested by @spuglegooge


src/libopensc/card-piv.c
```diff
@@ -5995,6 +6029,13 @@ piv_match_card_continued(sc_card_t *card)
 			priv->card_issues |= CI_VERIFY_LC0_FAIL;
 		break;
 
+	case SC_CARD_TYPE_PIV_II_OPENFIPS201:
+		/* also supports using Yubic-piv-tool */
```
Jakuje:
```suggestion
		/* also supports using Yubico-piv-tool */
```

src/libopensc/card-piv.c
```diff
@@ -4925,14 +4930,36 @@ piv_compute_signature(sc_card_t *card, const u8 *data, size_t datalen,
...
+		if (priv->card_issues & CI_ENFORCES_EC_HASH_EQ_KEYSIZE &&
+				!strcmp(card->ctx->app_name, "cardmod") &&
+				(datalen < nLen)) {
+			ecpadlen = nLen - datalen;
+		} else {
+			ecpadlen = 0;
+		}
+
+		ecdatalen = datalen + ecpadlen;
+		ecdata = calloc(1, ecdatalen);
```
Jakuje:
```suggestion
		ecdata = calloc(ecdatalen, sizeof(u8));
```

Jakuje (approved):
I did not manage to follow all the AI generated comments in the previous PRs and issues so I can not say I understand the whole chain of conditions, but this sounds more reasonable than doing it for all cards. Only few nits inline

src/libopensc/card-piv.c
dengert: Will fix.

Any AI was from @spurglegooge  not me.  I too found them way to long and would rather see the logs.

src/libopensc/card-piv.c
dengert: fixed

src/libopensc/card-piv.c
dengert: fixed

spurglegooge:
@dengert I tested your patch on our hardware and it works. Thanks!

The log lines below come from OpenSC minidriver's log.

The line numbers show the different builds: :5048 is the 0.27.1 build, :5034 is this PR.
Both were offline (cached-credential) logon attempts on the same card,
container 0 = 9A, where Windows sends a 32-byte value to the P-384 key:

```
BROKEN: should fail:
:5048   32B   sc_pkcs15_compute_signature -1205  ->  CardSignData 0x80100004   x3   before: logon fails
PATCH : should pass
:5034   32B   sc_pkcs15_compute_signature 96     ->  CardSignData 0x00000000   x3   with this PR: logon succeeds
```

The 96 is the P-384 signature coming back, so the applet accepted the padded input. 


Config:
Built from branch piv-ec-windows @ 42f41ed.
We triple checked that our #3817 pkcs15-sec.c change is not in the tree.
OpenFIPS201 applet on an NXP J3R310, 9A is EC P-384, Windows 11 26200, Samba KDC.
Logging enabled with MiniDriverDebug=1 under HKLM\SOFTWARE\OpenSC Project\OpenSC.

dengert:
@spurglegooge  says it works. It looks OK to me

frankmorgner (approved):

PR #3839, opened 2026-09-22
piv: Fail when metadata does not contain public key

Jakuje:
Follow-up from #3822 and @dengert testing with different PIV applets.

##### Checklist

- [ ] macOS token is tested


dengert:
Works!

https://github.com/user-attachments/files/32515975/with-change.txt

frankmorgner (approved):

PR #3840, opened 2026-09-22
Fix strict linux build

Jakuje:
The strict build with clang in Fedora Rawhide was failing for last couple of weeks. This changes the CI process to set the strict CFLAGS only for the make command so the buggy configure scripts are not affected. This also allows us to remove some of `-Wno-*` flags that we had to introduce for the configure to pass before.

##### Checklist

- [ ] macOS token is tested


frankmorgner (approved):

metsma (approved):

Jakuje:
Thank you for the reviews!

Verified in #3842 the errors still trigger if I remove the `-Wno-unused-parameter` so I will get this merged so we can make the CI green again.

